PR #269, opened 2021-03-06
Implemented Simple Drain Node Allocation

njtran:
https://github.com/awslabs/karpenter/issues/226

- Refactored labels for monitoring Karpenter
- Added client-go evict API into reallocator
- Refactored the reallocation process into a tree of subcontrollers
- Changed functional package from `f` to `functional` (https://github.com/awslabs/karpenter/issues/258)


pkg/controllers/provisioning/v1alpha1/reallocation/annotate.go
```diff
@@ -71,6 +71,7 @@ func (a *Annotator) CordonNodes(ctx context.Context, nodes []*v1.Node) error {
 	for _, node := range nodes {
 		persisted := node.DeepCopy()
 		node.Spec.Unschedulable = true
+		node.Labels[v1alpha1.ProvisionerDrainingKey] = "draining"
```
ellistarn, line 74: Convention is to use `string(true)`

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -78,7 +79,9 @@ func (c *Controller) Reconcile(object controllers.Object) error {
 	}
 
 	// 3. Get Nodes with Underutilized Label for resource usage reevaluation and Remove TTL if necessary
-	labeledUnderutilized, err := c.filter.GetLabeledUnderutilizedNodes(ctx, provisioner)
+	labeledUnderutilized, err := c.filter.GetLabeledNodes(ctx, provisioner, map[string]string{
+		v1alpha1.ProvisionerUnderutilizedKey: "true",
```
ellistarn, line 83: Be consistent between underutilized and draining label value (i.e. string(true))

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -87,24 +90,39 @@ func (c *Controller) Reconcile(object controllers.Object) error {
...
 	}
 
-	// TODO
-	// 6. Drain Nodes past TTL
+	// 6. Drain Nodes with draining label
+	draining, err := c.filter.GetLabeledNodes(ctx, provisioner, map[string]string{
+		v1alpha1.ProvisionerDrainingKey: "draining",
+	})
+	if err != nil {
+		return fmt.Errorf("listing nodes, %w", err)
```
ellistarn, line 111: clarify "listing drainable nodes"

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -78,7 +79,9 @@ func (c *Controller) Reconcile(object controllers.Object) error {
 	}
 
 	// 3. Get Nodes with Underutilized Label for resource usage reevaluation and Remove TTL if necessary
-	labeledUnderutilized, err := c.filter.GetLabeledUnderutilizedNodes(ctx, provisioner)
+	labeledUnderutilized, err := c.filter.GetLabeledNodes(ctx, provisioner, map[string]string{
```
ellistarn, line 82: Consider encapsulating this inside of the filter function. Something like filter.GetUnderutilizedNodes() seems idiomatic to me. Looking at this more closely, though, How does this filter differ from line 71?

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -87,24 +90,39 @@ func (c *Controller) Reconcile(object controllers.Object) error {
...
+		return fmt.Errorf("getting terminable nodes, %w", err)
 	}
-	if len(expired) == 0 {
+	if len(terminable) == 0 {
 		return nil
 	}
 
-	// 5. Cordon each node
-	if err := c.annotator.CordonNodes(ctx, c.filter.GetCordonableNodes(expired)); err != nil {
+	// 5. Cordon each node and label draining
```
ellistarn, line 101: You don't need to leak the fact that the node is labeled in the comment. Consider this as an implementation detail of cordon.

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -87,24 +90,39 @@ func (c *Controller) Reconcile(object controllers.Object) error {
...
+	if err := c.annotator.CordonNodes(ctx, c.filter.GetCordonableNodes(terminable)); err != nil {
```
ellistarn, line 102: I'm a little thrown by the differences in patterns in how we're filtering nodes.

Sometimes we get them via filter.GetTerminableNodes(ctx, provisioner), sometimes we get them via c.Filter.GetCordonableNodes(nodes...), c.filter.GetLabeledNodes(ctx, provisioner, map). 

It's not clear to my why this `filter` class behaves in so many different ways. Is there a way that we can collapse these three different modes of operation without sacrificing performance?

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -87,24 +90,39 @@ func (c *Controller) Reconcile(object controllers.Object) error {
...
 		return fmt.Errorf("cordoning node, %w", err)
 	}
 
-	// TODO
-	// 6. Drain Nodes past TTL
+	// 6. Drain Nodes with draining label
+	draining, err := c.filter.GetLabeledNodes(ctx, provisioner, map[string]string{
```
ellistarn, line 107: I'd probably make GetDrainingNodes() a first class concept on the filter. Otherwise we're leaking the labeling implementation detail.

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -87,24 +90,39 @@ func (c *Controller) Reconcile(object controllers.Object) error {
...
+
+	if err := c.terminator.DrainNodes(ctx, draining, provisioner); err != nil {
+		return fmt.Errorf("draining nodes, %w", err)
+	}
 
-	// 7. Delete Nodes past TTL
-	if err := c.terminator.DeleteNodes(ctx, expired, &provisioner.Spec); err != nil {
+	// 7. Delete Nodes with drained label
+	drained, err := c.filter.GetLabeledNodes(ctx, provisioner, map[string]string{
+		v1alpha1.ProvisionerDrainingKey: "drained",
```
ellistarn, line 120: Instead of labeling with draining=drained, can we instead just delete the node if the node is in draining=true and there are no pods scheduled to the node?

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -87,24 +90,39 @@ func (c *Controller) Reconcile(object controllers.Object) error {
...
+	})
+	if err != nil {
+		return fmt.Errorf("listing nodes, %w", err)
```
ellistarn, line 123: Can you differentiate this `listing nodes` log line from the others?

pkg/controllers/provisioning/v1alpha1/reallocation/filter.go
```diff
@@ -89,30 +89,16 @@ func (f *Filter) GetCordonableNodes(nodes []*v1.Node) []*v1.Node {
 	return nonCordonedNodes
 }
 
-// GetLabeledUnderutilizedNodes gets the nodes that have been labled underutilized for resource usage reevaluation
-func (f *Filter) GetLabeledUnderutilizedNodes(ctx context.Context, provisioner *v1alpha1.Provisioner) ([]*v1.Node, error) {
+func (f *Filter) GetLabeledNodes(ctx context.Context, provisioner *v1alpha1.Provisioner, additionalLabels map[string]string) ([]*v1.Node, error) {
```
ellistarn, line 92: consider just GetNodes() since, it's already within the context of a karpenter provisioner.

pkg/controllers/provisioning/v1alpha1/reallocation/filter.go
```diff
@@ -89,30 +89,16 @@ func (f *Filter) GetCordonableNodes(nodes []*v1.Node) []*v1.Node {
...
 	nodes := &v1.NodeList{}
 	labelMap := map[string]string{
-		v1alpha1.ProvisionerUnderutilizedKey:  "true",
 		v1alpha1.ProvisionerNameLabelKey:      provisioner.Name,
 		v1alpha1.ProvisionerNamespaceLabelKey: provisioner.Namespace,
 	}
 	for k, v := range provisioner.Spec.Labels {
 		labelMap[k] = v
```
ellistarn, line 99: This code seems actively dangerous. Why is this here?

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,18 +19,70 @@ import (
...
+		// - Disrupts PDB
+		// - Pods owned by controller object
+		// - Pod on Node can't be rescheduled elsewhere
+
+		// 1. Get Pods on Node
+		podList := &v1.PodList{}
+		if err := t.kubeClient.List(ctx, podList, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
+			return fmt.Errorf("listing pods for node %s, %w", node.Name, err)
+		}
+		// 1a. If no pods on node, label as drained
```
ellistarn, line 51: Don't you have this emptiness logic elsewhere? Does something need to be refactored?

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,18 +19,70 @@ import (
...
+				}
+			}
+		}
+		persisted := node.DeepCopy()
+		if !empty {
+			// Mark as drained
+			node.Labels[v1alpha1.ProvisionerDrainingKey] = "draining"
+		} else {
+			// Mark Node as Draining
+			node.Labels[v1alpha1.ProvisionerDrainingKey] = "drained"
```
ellistarn, line 71: Why not just delete the node immediately, rather than mark as drained and then delete?

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,18 +19,70 @@ import (
...
+		if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
+			return fmt.Errorf("patching node %s, %w", node.Name, err)
+		}
+	}
+	return nil
 }
 
 // DeleteNodes will use a cloudprovider implementation to delete a set of nodes
 func (t *Terminator) DeleteNodes(ctx context.Context, nodes []*v1.Node, spec *v1alpha1.ProvisionerSpec) error {
+	if len(nodes) == 0 {
```
ellistarn, line 83: This check shouldn't be necessary. Why doesn't the cloud provider tolerate being called with 
```
t.cloudprovider.CapacityFor(spec).Delete(ctx, [])
```

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -87,24 +90,39 @@ func (c *Controller) Reconcile(object controllers.Object) error {
...
+	draining, err := c.filter.GetLabeledNodes(ctx, provisioner, map[string]string{
+		v1alpha1.ProvisionerDrainingKey: "draining",
+	})
+	if err != nil {
+		return fmt.Errorf("listing nodes, %w", err)
+	}
+
+	if err := c.terminator.DrainNodes(ctx, draining, provisioner); err != nil {
+		return fmt.Errorf("draining nodes, %w", err)
+	}
```
prateekgogia, line 116: Not clear why are we trying to drain nodes which are already labelled `draining`

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
prateekgogia, line 71: The comments are reversed?

pkg/controllers/provisioning/v1alpha1/reallocation/annotate.go
njtran: I wanted to have the draining key be able to filtered on either `draining` or `drained`

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
njtran: I wanted to be able to have the draining label distinguish between whether it was drained or not, while the underutilized label would stutter if the value was underutilized.

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
njtran: I decided to create the `GetLabeledNodes` function because I saw that the filter file was getting extremely long with repeated logic, and wanted to simplify it all. 

`GetLabeledNodes` simply gets nodes with the associated labels of the provisioner and any additional specified labels. 

`GetUnderutilizedNodes` and  `GetTerminableNodes` both require special logic (checking if it is underutilized and checking if is past TTL.

njtran: `GetCordonableNodes` and `GetTTLableNodes` have special cases as well that require a simple check as opposed to a label check.

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
njtran: A lot of the other comments have `label ___` that help the reader understand the flow of the checks and annotations. I think it's important here to help the reader understand what's going on in the code without having to deep dive.

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
njtran: We can discuss offline, but what's wrong with leaking the labeling implementation detail here?

pkg/controllers/provisioning/v1alpha1/reallocation/filter.go
njtran: I don't quite understand what you mean by dangerous. This code gets the labels specified by the user that will be used for the provisioner. It is required that we filter on these labels as well to make sure that we capture all nodes handled by the provisioner.

pkg/controllers/provisioning/v1alpha1/reallocation/filter.go
```diff
@@ -89,30 +89,16 @@ func (f *Filter) GetCordonableNodes(nodes []*v1.Node) []*v1.Node {
...
-}
-
-func (f *Filter) getNodes(ctx context.Context, provisioner *v1alpha1.Provisioner) ([]*v1.Node, error) {
-	nodes := &v1.NodeList{}
-	labelMap := map[string]string{
-		v1alpha1.ProvisionerNameLabelKey:      provisioner.Name,
-		v1alpha1.ProvisionerNamespaceLabelKey: provisioner.Namespace,
-	}
-	for k, v := range provisioner.Spec.Labels {
+	for k, v := range additionalLabels {
```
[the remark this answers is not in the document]
njtran: We aren't mutating any labels here. We are forming a labelMap to use for searching purposes.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
njtran: I don't know where you might be talking about. Here I'm checking to make sure that the node has no pods. If it has no pods, then it becomes labeled drained.

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
njtran: As we are calling this reconcile loop every 5 seconds, we need to constantly be evicting the pods that are on the node, or draining, since this can take any amount of time, and can fail for many reasons.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,18 +19,70 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type Terminator struct {
 	kubeClient    client.Client
 	cloudprovider cloudprovider.Factory
+	coreV1Client  corev1.CoreV1Interface
+}
+
+// DrainNodes takes in a list of expired non-drained nodes and calls drain on them
```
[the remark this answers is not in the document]
njtran: For the refactor, Ive decided to call delete in the drain nodes action. For any node of the iteration, I will only log a debug statement if it fails and continues. Let me know what you think.

pkg/cloudprovider/aws/fleet/instance.go
```diff
@@ -84,6 +84,9 @@ func (p *InstanceProvider) Create(ctx context.Context,
 }
 
 func (p *InstanceProvider) Terminate(ctx context.Context, nodes []*v1.Node) error {
+	if len(nodes) == 0 {
```
ellistarn, line 87: Sorry if I missed the comment. Why is this necessary?

ellistarn: Nvm -- this is clearly a stupid EC2 impl detail. Why EC2 doesn't happily say OK to zero deletions is beyond me.

pkg/controllers/provisioning/v1alpha1/reallocation/collect.go
```diff
@@ -0,0 +1,121 @@
...
+	f "github.com/awslabs/karpenter/pkg/utils/functional"
+	utilsnode "github.com/awslabs/karpenter/pkg/utils/node"
+	"github.com/awslabs/karpenter/pkg/utils/ptr"
+	"go.uber.org/zap"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"time"
+)
+
+type Collector struct {
```
ellistarn, line 16: Consider Utilization or UtilizationController
utilization.Reconcile()

pkg/controllers/provisioning/v1alpha1/reallocation/collect.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return err
+	}
+
+	// 2. Get nodes and pods for each node
+	for _, node := range nodes {
+		pods, err := c.getPodsOnNode(ctx, node.Name)
+		if err != nil {
+			zap.S().Debugf("Unable to get pods for node %s, %s", node.Name, err.Error())
+			continue
```
ellistarn, line 48: We probably don't want to continue here. If we fail to call the API Server (overloaded, or some other reason), we probably don't want to keep hammering it. Best to let the reconcile loop backoff/retry logic kick in and just return an error here.

pkg/controllers/provisioning/v1alpha1/reallocation/collect.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+	}
+
+	// 3. Set TTL for each underutilized node
+	for _, node := range ttlable {
+		persisted := node.DeepCopy()
+		node.Labels[v1alpha1.ProvisionerUnderutilizedKey] = "true"
+		node.Annotations[v1alpha1.ProvisionerTTLKey] = time.Now().Add(time.Duration(*provisioner.Spec.TTLSeconds) * time.Second).Format(time.RFC3339)
+		if err := c.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
+			zap.S().Debugf("Could not add TTL to node %s, %s", node.Name, err.Error())
```
ellistarn, line 63: I think this is an error that we should bubble up for the same reasons as mentioned above.

pkg/controllers/provisioning/v1alpha1/reallocation/collect.go
```diff
@@ -0,0 +1,121 @@
...
+		v1alpha1.ProvisionerUnderutilizedKey: "true",
+	})
+	if err != nil {
+		return fmt.Errorf("listing labeled underutilized nodes, %w", err)
+	}
+
+	// Clear Underutilized Label if node is utilized
+	for _, node := range nodes {
+		pods := &v1.PodList{}
+		if err := c.kubeClient.List(ctx, pods, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
```
ellistarn, line 83: getPodsOnNode()?

pkg/controllers/provisioning/v1alpha1/reallocation/collect.go
```diff
@@ -0,0 +1,121 @@
...
+			return fmt.Errorf("listing pods on node %s, %w", node.Name, err)
+		}
+
+		if !utilsnode.IsUnderutilized(node, ptr.PodListToSlice(pods)) {
+			persisted := node.DeepCopy()
+			delete(node.Labels, v1alpha1.ProvisionerUnderutilizedKey)
+			delete(node.Annotations, v1alpha1.ProvisionerTTLKey)
+			if err := c.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
+				zap.S().Debugf("Could not remove underutilized labels on node %s, %w", node.Name, err)
```
ellistarn, line 92: Same thing w.r.t errors here.

pkg/controllers/provisioning/v1alpha1/reallocation/collect.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+// getNodes returns a list of nodes with the provisioner's labels and given labels
+func (c *Collector) getNodes(ctx context.Context, provisioner *v1alpha1.Provisioner, additionalLabels map[string]string) ([]*v1.Node, error) {
+	nodes := &v1.NodeList{}
+	labelMap := f.UnionStringMaps(map[string]string{
```
ellistarn, line 104: consider inlining

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,21 +19,125 @@ import (
...
 	cloudprovider cloudprovider.Factory
+	coreV1Client  corev1.CoreV1Interface
 }
 
-// DeleteNodes will use a cloudprovider implementation to delete a set of nodes
-func (t *Terminator) DeleteNodes(ctx context.Context, nodes []*v1.Node, spec *v1alpha1.ProvisionerSpec) error {
+func (t *Terminator) ManageTermination(ctx context.Context, provisioner *v1alpha1.Provisioner) error {
+	// 1. Cordon each terminable node
+	if err := t.cordonNodes(ctx, provisioner); err != nil {
+		return fmt.Errorf("cordoning expired nodes, %w", err)
```
ellistarn, line 43: I think this annotation may not age well if we decide to make nodes terminable in other ways via expiry. Consider cordoning terminable node.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -47,3 +151,16 @@ func (t *Terminator) DeleteNodes(ctx context.Context, nodes []*v1.Node, spec *v1
 	}
 	return nil
 }
+
+
+func (t *Terminator) getNodes(ctx context.Context, provisioner *v1alpha1.Provisioner, additionalLabels map[string]string) ([]*v1.Node, error) {
```
ellistarn, line 156: Can you cut an issue to dedup these two identical getNodes functions?

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,21 +19,125 @@ import (
...
+		}
+		zap.S().Debugf("Cordoned node %s", node.Name)
+	}
+	return nil
+}
+
+// drainAndDeleteNodes takes in a list of expired non-drained nodes and calls drain on them
+func (t *Terminator) drainAndDeleteNodes(ctx context.Context, provisioner *v1alpha1.Provisioner) error {
+	drained := []*v1.Node{}
+	draining, err := t.getNodes(ctx, provisioner, map[string]string{
```
ellistarn, line 82: I like how you factored this nodes function -- seems useful in many places.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,21 +19,125 @@ import (
...
+	for _, node := range draining {
+		// (0.) TODO: Check if Node should be drained
+		// - Disrupts PDB
+		// - Pods owned by controller object
+		// - Pod on Node can't be rescheduled elsewhere
+
+		// 1. Get Pods on Node
+		podList := &v1.PodList{}
+		if err := t.kubeClient.List(ctx, podList, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
+			zap.S().Debugf("Continuing after failing to list pods for node %s, %s", node.Name, err.Error())
```
ellistarn, line 98: Same w.r.t. error handling here.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,21 +19,125 @@ import (
...
+		// 3. Patch Node label if it is draining
+		persisted := node.DeepCopy()
+		if !empty {
+			// If not empty, keep it as draining
+			node.Labels[v1alpha1.ProvisionerDrainingKey] = "true"
+		} else {
+			// If empty, mark as drained
+			drained = append(drained, node)
+		}
+		if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
```
ellistarn, line 124: Do we want to patch if the node is already draining? I think we can reduce calls to the API Server.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,21 +19,125 @@ import (
...
+
+	// 4. Delete any nodes that were fully drained
+	if err := t.deleteNodes(ctx, drained, provisioner); err != nil {
+		return fmt.Errorf("deleting %d nodes, %w", len(drained))
+	}
+	return nil
+}
+
+// deleteNode uses a cloudprovider-specific delete to delete a set of nodes
+func (t *Terminator) deleteNodes(ctx context.Context, nodes []*v1.Node, provisioner *v1alpha1.Provisioner) error {
```
ellistarn, line 137: If this function is just a straight passthrough, can we dedup it?

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,21 +19,125 @@ import (
...
+			return fmt.Errorf("patching node %s, %w", node.Name, err)
+		}
+		zap.S().Debugf("Cordoned node %s", node.Name)
+	}
+	return nil
+}
+
+// drainAndDeleteNodes takes in a list of expired non-drained nodes and calls drain on them
+func (t *Terminator) drainAndDeleteNodes(ctx context.Context, provisioner *v1alpha1.Provisioner) error {
+	drained := []*v1.Node{}
```
ellistarn, line 81: move this to just before the for loop where it's used.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,21 +19,125 @@ import (
...
+	for _, node := range nodeList {
+		if utilsnode.IsPastTTL(node) && !node.Spec.Unschedulable {
+			nodes = append(nodes, node)
+		}
+	}
+
+	for _, node := range nodes {
+		persisted := node.DeepCopy()
+		node.Spec.Unschedulable = true
+		node.Labels[v1alpha1.ProvisionerDrainingKey] = "true"
```
ellistarn, line 70: consider string(true) to avoid naked constants

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,21 +19,125 @@ import (
...
+	// 2. Drain Nodes with draining label and delete any drained nodes
+	if err := t.drainAndDeleteNodes(ctx, provisioner); err != nil {
+		return fmt.Errorf("draining nodes, %w", err)
+	}
+	return nil
+}
+
+// cordonNodes takes in a list of expired nodes as input and cordons them
+func (t *Terminator) cordonNodes(ctx context.Context, provisioner *v1alpha1.Provisioner) error {
+	nodeList, err := t.getNodes(ctx, provisioner, map[string]string{})
```
ellistarn, line 55: Do you want to filter by nodes somehow? What's the signal for if a node is ready to be cordoned? 

In my head, there should be a way to mark a node as terminable and then this guy kicks in. Meaning, anyone should be able to add a label to a node which will cause it to be cordoned and drained by a shared controller. I'd expect that the TTL expiry check actually happens in the underutilization logic, not the termination logic.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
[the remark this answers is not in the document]
ellistarn: Added some comments. I think it's pretty rare that you want to continue if these API calls fail. In general, only continue if it would be harmful not to. The main idea behind this is that if the API server is rejecting your request, it probably doesn't help to keep spamming it with more requests. If it's overloaded, it's bad, if it's an authentication issue, it won't help, if it's unable to connect, you're probably stuck anyways.

Controller runtime has some great machinery around tracking reconcile loop failures and plugging them into a backoff/retry function. 

There are some cases where you see "continuing after ...". These are mostly useful because we've already created an EC2 instance, and it would be harmful to not attempt to at least bind the pods we prepared for it. I'm also going to deep dive all of those continue statements and make sure they're all good ideas under the framework mentioned about.

pkg/controllers/provisioning/v1alpha1/reallocation/filter.go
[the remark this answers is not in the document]
ellistarn: Oops. Thanks for reminding me.

ellistarn:
In general, I like the refactor. A few high level ideas.
1. It's important to bubble up our errors
2. I think the TTL expiry check should happen in the underutilization controller, not the termination controller. This will allow the termination controller to be generic and used by other use cases (upgrade/defrag/etc)

pkg/controllers/provisioning/v1alpha1/reallocation/collect.go
JacobGabrielson: Agreed - my comment about sometimes continuing was specifically (only) for certain EC2 API calls (especially Terminate)

pkg/controllers/provisioning/v1alpha1/reallocation/collect.go
JacobGabrielson: Agreed - see my other clarification

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,21 +19,125 @@ import (
...
+			drained = append(drained, node)
+		}
+		if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
+			zap.S().Debugf("Continuing after failing to patch node %s, %s", node.Name, err.Error())
+		}
+	}
+
+	// 4. Delete any nodes that were fully drained
+	if err := t.deleteNodes(ctx, drained, provisioner); err != nil {
+		return fmt.Errorf("deleting %d nodes, %w", len(drained))
```
JacobGabrielson, line 131: This is the one place where I think we need to have custom error handling - a node can fail to terminate due to some kind of hang up on the control plane side (say something got attached to it we're not aware of, etc) and we should still terminate the other nodes. A simple solution would be to randomize the order that we traverse the nodes to terminate (this is my recommendation), but we could also get into handling different error conditions (differentiate between the API server just throttling us or some kind of malformed input vs. this-instance-cannot-be-terminated-because-reason).

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,21 +19,125 @@ import (
...
-// DeleteNodes will use a cloudprovider implementation to delete a set of nodes
-func (t *Terminator) DeleteNodes(ctx context.Context, nodes []*v1.Node, spec *v1alpha1.ProvisionerSpec) error {
+func (t *Terminator) ManageTermination(ctx context.Context, provisioner *v1alpha1.Provisioner) error {
+	// 1. Cordon each terminable node
+	if err := t.cordonNodes(ctx, provisioner); err != nil {
+		return fmt.Errorf("cordoning expired nodes, %w", err)
+	}
+
+	// 2. Drain Nodes with draining label and delete any drained nodes
+	if err := t.drainAndDeleteNodes(ctx, provisioner); err != nil {
```
JacobGabrielson, line 47: nit: I'd try to avoid adding new "verbs" - what is the difference between "terminate" and "delete"? Could we narrow down to just one?

ellistarn: In this case, I think "terminate" refers to the drain/delete workflow. Terminate is async, as it requires all of the pods to be evicted first.

pkg/controllers/provisioning/v1alpha1/reallocation/collect.go
njtran: Good catch

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
njtran: I'll change to terminate

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
njtran: I can totally see that. That's a good idea to call out. I agree, marking it as expired in the utilization controller is a better organization.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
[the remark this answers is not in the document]
njtran: Got it. The EC2 terminate instance call can take in a batch ids so I don't think we should be concerned with this, right?

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
njtran: Will do.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,21 +19,125 @@ import (
...
+
+	for _, node := range draining {
+		// (0.) TODO: Check if Node should be drained
+		// - Disrupts PDB
+		// - Pods owned by controller object
+		// - Pod on Node can't be rescheduled elsewhere
+
+		// 1. Get Pods on Node
+		podList := &v1.PodList{}
+		if err := t.kubeClient.List(ctx, podList, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
```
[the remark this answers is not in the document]
njtran: I have a getPodsOnNode in the other file. I'll add it in here, but it should be dedup'd later, similar to GetNodes.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
njtran: I'd rather keep it separate here. If I add it into the function it makes the `terminateNodes` function a lot longer. It's also a very different logical flow, and could be used elsewhere.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
njtran: I agree. Removed this patch.

pkg/apis/provisioning/v1alpha1/provisioner.go
```diff
@@ -106,8 +106,10 @@ var (
 	// Reserved labels
 	ProvisionerNameLabelKey      = SchemeGroupVersion.Group + "/name"
 	ProvisionerNamespaceLabelKey = SchemeGroupVersion.Group + "/namespace"
-	ProvisionerUnderutilizedKey  = SchemeGroupVersion.Group + "/underutilized"
-	ProvisionerTTLKey            = SchemeGroupVersion.Group + "/ttl"
+	//ProvisionerUnderutilizedKey  = SchemeGroupVersion.Group + "/underutilized"
+	ProvisionerTTLKey = SchemeGroupVersion.Group + "/ttl"
+	//ProvisionerDrainingKey       = SchemeGroupVersion.Group + "/draining"
```
ellistarn, line 111: Did you mean to delete this?

pkg/apis/provisioning/v1alpha1/provisioner.go
```diff
@@ -106,8 +106,10 @@ var (
 	// Reserved labels
 	ProvisionerNameLabelKey      = SchemeGroupVersion.Group + "/name"
 	ProvisionerNamespaceLabelKey = SchemeGroupVersion.Group + "/namespace"
-	ProvisionerUnderutilizedKey  = SchemeGroupVersion.Group + "/underutilized"
-	ProvisionerTTLKey            = SchemeGroupVersion.Group + "/ttl"
+	//ProvisionerUnderutilizedKey  = SchemeGroupVersion.Group + "/underutilized"
+	ProvisionerTTLKey = SchemeGroupVersion.Group + "/ttl"
```
ellistarn, line 110: Can you move this outside of the labels, since it's an annotation? Maybe it should belong in the utilization controller code.

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -52,11 +52,10 @@ func (c *Controller) Name() string {
 }
 
 // NewController constructs a controller instance
-func NewController(kubeClient client.Client, cloudProvider cloudprovider.Factory) *Controller {
+func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.Factory) *Controller {
 	return &Controller{
-		annotator:     &Annotator{kubeClient: kubeClient},
-		filter:        &Filter{kubeClient: kubeClient},
-		terminator:    &Terminator{kubeClient: kubeClient, cloudprovider: cloudProvider},
+		utilization:     &Utilization{kubeClient: kubeClient},
```
ellistarn, line 57: reformat?

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -66,47 +65,20 @@ func (c *Controller) Reconcile(object controllers.Object) error {
...
+	// 3. Reevaluate any existing underutilized nodes
+	if err := c.utilization.Reconcile(ctx, provisioner); err != nil {
+		return fmt.Errorf("reconciling utilization sub-controller, %w", err)
 	}
 
-	// 2. Set TTL and label underutilized nodes
-	if err := c.annotator.MarkUnderutilized(ctx, c.filter.GetTTLableNodes(underutilized), *provisioner.Spec.TTLSeconds); err != nil {
-		return fmt.Errorf("adding ttl and underutilized label, %w", err)
+	// Reconcile
+	// 1. Get Expired Nodes
```
ellistarn, line 77: I might leave these details inside the terminator, rather than stating the steps outside.

ellistarn: If these details ever change, we'll need to remember to update here (likely will get stale).

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,24 +19,115 @@ import (
...
 	}
+	return nil
+}
 
-	// 2. Delete node in APIServer to ensure
+// cordonNodes takes in a list of expired nodes as input and cordons them
+func (t *Terminator) cordonNodes(ctx context.Context, provisioner *v1alpha1.Provisioner) error {
+	// 1. Get terminable nodes
+	nodeList, err := t.getNodes(ctx, provisioner, map[string]string{
+		v1alpha1.ProvisionerPhaseLabel: "terminable",
```
ellistarn, line 56: We should keep track of these states in the v1alpha1 package, since we hardcode the string in multiple places. 

type LifecyclePhase string
const (
  LifecyclePhaseTerminable LifecyclePhase = "Terminable"
)

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
ellistarn: This has now changed to actually have logic. Before it was of them form
```
func foo() {
  return bar()
}
```
Agreed that it now has enough logic to justify its existence :)

pkg/apis/provisioning/v1alpha1/provisioner.go
njtran: Yeah. Thanks

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
njtran: Ah guess my linter and fmt didn't run.

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
njtran: Since I have the details in the controllers, I'll just remove them. Thanks.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
njtran: Ah yes. That's a great idea. Thanks!

pkg/apis/provisioning/v1alpha1/provisioner.go
```diff
@@ -106,14 +106,22 @@ var (
...
+	// Reserved annotations
+	ProvisionerTTLKey = SchemeGroupVersion.Group + "/ttl"
 
 	// Use ProvisionerSpec instead
 	ZoneLabelKey         = "topology.kubernetes.io/zone"
 	InstanceTypeLabelKey = "node.kubernetes.io/instance-type"
 )
 
+const (
+	ProvisionerUnderutilizedPhase = "underutilized"
```
ellistarn, line 120: Do you mind capitalizing these? Better k8s style I think. Check out other resources like Pod and Node

pkg/apis/provisioning/v1alpha1/provisioner.go
```diff
@@ -106,14 +106,22 @@ var (
...
+	ProvisionerTerminablePhase    = "terminable"
+	ProvisionerDrainingPhase      = "draining"
```
ellistarn, line 122: What's the difference between Terminable and Draining? Can this be collapsed into a single state: Terminating? That state is managed by the terminator?

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -19,24 +19,115 @@ import (
...
 	}
+	return nil
+}
 
-	// 2. Delete node in APIServer to ensure
+// cordonNodes takes in a list of expired nodes as input and cordons them
+func (t *Terminator) cordonNodes(ctx context.Context, provisioner *v1alpha1.Provisioner) error {
+	// 1. Get terminable nodes
+	nodeList, err := t.getNodes(ctx, provisioner, map[string]string{
+		v1alpha1.ProvisionerPhaseLabel: v1alpha1.ProvisionerTerminablePhase,
```
ellistarn, line 56: IMO provisioner is redundant here since we're already in the v1alpha1 provisioner code. Perhaps v1alpha1.PhaseLabel and v1alpha1.TerminablePhase

pkg/apis/provisioning/v1alpha1/provisioner.go
njtran: Terminable is when a node is expired, or after the TTL. Draining is when the node is still in the process of draining. If the underutilization predicate is 0 non-daemonset pods, draining doesn't last very long.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
njtran: Other keys and values use the same syntax as above. I wanted to keep it in line with that, and I think that it should be `ProvisionerNameLabelKey` and so on

pkg/apis/provisioning/v1alpha1/provisioner.go
njtran: I think I'd rather keep it lowercase here. Since these are labels, I want to keep it in line with how the other labels are done, which they're all lowercase.

pkg/apis/provisioning/v1alpha1/provisioner.go
njtran: There is no difference in amount of API calls done.

ellistarn (approved):

ellistarn (approved):

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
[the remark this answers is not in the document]
JacobGabrielson: Probably; I would test to see what happens if you pass in a batch but have (say) one of the instances marked not terminatable.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
JacobGabrielson: I'd be fine w/ it happening in the cloud provider.

pkg/apis/provisioning/v1alpha1/provisioner.go
JacobGabrielson: We should probably chat offline about these terms as it is a bit confusing at the moment (even with the explanation), maybe we can come up with some different words?

PR #274, opened 2021-03-10
Fix wrong remaining pods value 

prateekgogia:
*Issue #, if available:*

*Description of changes:*
Problem: we check all the node capacities, if smallest node is checked in the end `remainingPods` gets updated but the `bestPackedPods` already contains higher number of packed pods. So when this function returns it `bestPackedPods` which contains the highest number of packed pods but `remainingPods` contains pods left over for the node with a smaller capacity.

This fix will update the `remainingPods` to the right set of pods everytime we update `bestPackedPods`.

### DIFF
--- a/pkg/cloudprovider/aws/fleet/packing/packing.go
+++ b/pkg/cloudprovider/aws/fleet/packing/packing.go
@@ -87,12 +87,12 @@ func (p *podPacker) getNodeCapacities() []*nodeCapacity {
 func (p *podPacker) packWithLargestPod(pods []*v1.Pod) (*Packing, []*v1.Pod) {
 	bestPackedPods := []*v1.Pod{}
 	bestCapacities := []*nodeCapacity{}
-	remainingPods := []*v1.Pod{}
+	remainingPods := pods
 	// TODO reserve (Kubelet+ daemon sets) overhead for instance types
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/aws/fleet/packing/packing.go
```diff
@@ -87,13 +87,12 @@ func (p *podPacker) getNodeCapacities() []*nodeCapacity {
 func (p *podPacker) packWithLargestPod(pods []*v1.Pod) (*Packing, []*v1.Pod) {
 	bestPackedPods := []*v1.Pod{}
 	bestCapacities := []*nodeCapacity{}
-	var packedPods []*v1.Pod
-	remainingPods := pods
+	remainingPods := []*v1.Pod{}
```
ellistarn, line 90: Do we need to do all of the variable renaming, or can we just keep this line and resolve the bug?

JacobGabrielson: I *think* that wouldn't work because `remainingPods` would still be always set every time `packPodsForCapacity` is called, as opposed to only setting it when a better packing is found?

ellistarn (approved):

PR #278, opened 2021-03-11
Reserve overhead calculations for node capacities

prateekgogia:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/pkg/cloudprovider/aws/packing/packing.go
+++ b/pkg/cloudprovider/aws/packing/packing.go
@@ -17,19 +17,18 @@ package packing
 import (
 	"context"
 	"sort"
-	"sync"
 
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/awslabs/karpenter/pkg/cloudprovider"
 	"github.com/awslabs/karpenter/pkg/utils/binpacking"
+	"github.com/awslabs/karpenter/pkg/utils/resources"
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
 )
 
 type podPacker struct {
 	// TODO use this ec2 API to get the instance types
-	ec2         ec2iface.EC2API
-	reserveOnce sync.Once
+	ec2 ec2iface.EC2API
 }
 
 // Packer helps pack the pods and calculates efficient placement on the instances.
@@ -80,19 +79,16 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
 
 // TODO filter instance types based on node contraints like availability zones etc.
 func (p *podPacker) getNodeCapacities(constraints *cloudprovider.Constraints) []*nodeCapacity {
-	p.reserveOnce.Do(func() {
-		for _, nc := range nodeCapacities {
-			kubeletOverhead := binpacking.CalculateKubeletOverhead(nc.total)
-			if ok := nc.reserve(binpacking.MergeResources(constraints.Overhead, kubeletOverhead)); !ok {
-				zap.S().Errorf("Failed to reserve kubelet overhead for node capacity type %v", nc.instanceType)
-			}
-		}
-	})
-	nodeCapacitiesCopy := make([]*nodeCapacity, 0)
+	result := make([]*nodeCapacity, 0)
 	for _, nc := range nodeCapacities {
-		nodeCapacitiesCopy = append(nodeCapacitiesCopy, &nodeCapacity{nc.instanceType, nc.reserved.DeepCopy(), nc.total.DeepCopy()})
+		ncc := nc.Copy()
+		kubeletOverhead := binpacking.CalculateKubeletOverhead(ncc.total)
+		if ok := ncc.reserve(resources.Merge(constraints.Overhead, kubeletOverhead)); !ok {
+			zap.S().Errorf("Failed to reserve kubelet overhead for node capacity type %v", ncc.instanceType)
+		}
+		result = append(result, ncc)
 	}
-	return nodeCapacitiesCopy
+	return result
 }
 
 type packingResult struct {
@@ -134,7 +130,7 @@ func (p *podPacker) packWithLargestPod(unpackedPods []*v1.Pod, constraints *clou
 
 func (p *podPacker) packPodsForCapacity(capacity *nodeCapacity, pods []*v1.Pod) *packingResult {
 	// start with the largest pod based on resources requested
-	result := &packingResult{packed: []*v1.Pod{}, unpacked: []*v1.Pod{}}
+	result := &packingResult{}
 	for _, pod := range pods {
 		if ok := capacity.reserveForPod(&pod.Spec); ok {
 			result.packed = append(result.packed, pod)
--- a/pkg/cloudprovider/aws/packing/packing.go
+++ b/pkg/cloudprovider/aws/packing/packing.go
@@ -95,7 +95,7 @@ func (p *podPacker) getNodeCapacities(constraints *cloudprovider.Constraints) []
 	return nodeCapacitiesCopy
 }
 
-type packingsPerCapacity struct {
+type packingResult struct {
 	packed   []*v1.Pod
 	unpacked []*v1.Pod
 }
@@ -132,9 +132,9 @@ func (p *podPacker) packWithLargestPod(unpackedPods []*v1.Pod, constraints *clou
 	return &Packing{Pods: bestPackedPods, InstanceTypes: capacityNames}, remainingPods
 }
 
-func (p *podPacker) packPodsForCapacity(capacity *nodeCapacity, pods []*v1.Pod) *packingsPerCapacity {
+func (p *podPacker) packPodsForCapacity(capacity *nodeCapacity, pods []*v1.Pod) *packingResult {
 	// start with the largest pod based on resources requested
-	result := &packingsPerCapacity{packed: []*v1.Pod{}, unpacked: []*v1.Pod{}}
+	result := &packingResult{packed: []*v1.Pod{}, unpacked: []*v1.Pod{}}
 	for _, pod := range pods {
 		if ok := capacity.reserveForPod(&pod.Spec); ok {
 			result.packed = append(result.packed, pod)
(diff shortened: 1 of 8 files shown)


pkg/cloudprovider/aws/fleet/packing/packing.go
```diff
@@ -45,7 +47,7 @@ type Packing struct {
 
 // NewPacker returns a Packer implementation
 func NewPacker(ec2 ec2iface.EC2API) Packer {
-	return &podPacker{ec2: ec2}
+	return &podPacker{ec2: ec2, reserveOnce: sync.Once{}}
```
JacobGabrielson, line 50: You can leave off the `reserveOnce: sync.Once{}` part.

pkg/cloudprovider/aws/fleet/packing/packing.go
```diff
@@ -77,34 +79,46 @@ func (p *podPacker) Pack(ctx context.Context, pods []*v1.Pod) ([]*Packing, error
 }
 
 // TODO filter instance types based on node contraints like availability zones etc.
-func (p *podPacker) getNodeCapacities() []*nodeCapacity {
+func (p *podPacker) getNodeCapacities(constraints *cloudprovider.Constraints) []*nodeCapacity {
+	p.reserveOnce.Do(
```
JacobGabrielson, line 83: I can see how this works but is it idiomatic Go when there isn't (IIUC?) the potential for concurrent calls to this code?

pkg/cloudprovider/aws/fleet/packing/packing.go
```diff
@@ -77,34 +79,46 @@ func (p *podPacker) Pack(ctx context.Context, pods []*v1.Pod) ([]*Packing, error
...
+type packingsPerCapacity struct {
+	packed   []*v1.Pod
+	unpacked []*v1.Pod
+}
+
 // packWithLargestPod will try to pack max number of pods with largest pod in
 // pods across all available node capacities. It returns Packing: max pod count
 // that fit; with their node capacities and list of leftover pods
-func (p *podPacker) packWithLargestPod(pods []*v1.Pod) (*Packing, []*v1.Pod) {
+func (p *podPacker) packWithLargestPod(podsToBePacked []*v1.Pod, constraints *cloudprovider.Constraints) (*Packing, []*v1.Pod) {
```
JacobGabrielson, line 103: I think this might read better as `unpackedPods`

pkg/cloudprovider/aws/fleet/packing/packing.go
```diff
@@ -77,34 +79,46 @@ func (p *podPacker) Pack(ctx context.Context, pods []*v1.Pod) ([]*Packing, error
...
-	remainingPods := pods
-	// TODO reserve (Kubelet+ daemon sets) overhead for instance types
-	// TODO count number of pods created on an instance type
-	for _, nc := range p.getNodeCapacities() {
+	remainingPods := podsToBePacked
+	for _, nc := range p.getNodeCapacities(constraints) {
 		// check how many pods we can fit with the available capacity
-		packedPods, unpackedPods := p.packPodsForCapacity(nc, pods)
-		if len(packedPods) == 0 {
+		pods := p.packPodsForCapacity(nc, podsToBePacked)
```
JacobGabrielson, line 109: It's a bit confusing to call something `pods` unless it's a collection of `v1.Pod` structs.

pkg/cloudprovider/aws/fleet/packing/packing.go
prateekgogia: Yes, there are no concurrent calls but we need to reserve overhead only once and we call `getNodeCapacities` every time we get pending pods.

pkg/cloudprovider/aws/fleet/packing/packing.go
```diff
@@ -115,24 +129,26 @@ func (p *podPacker) packWithLargestPod(pods []*v1.Pod) (*Packing, []*v1.Pod) {
 	return &Packing{Pods: bestPackedPods, InstanceTypes: capacityNames}, remainingPods
 }
 
-func (p *podPacker) packPodsForCapacity(capacity *nodeCapacity, pods []*v1.Pod) (packedPods, remainingPods []*v1.Pod) {
+func (p *podPacker) packPodsForCapacity(capacity *nodeCapacity, pods []*v1.Pod) *packingsPerCapacity {
 	// start with the largest pod based on resources requested
+	resetCapacity := capacity.reserved.DeepCopy()
+	defer func() {
+		capacity.reserved = resetCapacity
```
JacobGabrielson, line 136: This seems pretty complicated and possibly error-prone (especially if there actually were concurrent access, although I realize there shouldn't be). Why not operate on a copy of `capacity` rather than modify the global copy?

prateekgogia: I removed this, also all this logic will be removed very soon.

pkg/utils/binpacking/binpacking.go
```diff
@@ -40,3 +41,69 @@ func (r ByResourcesRequested) Less(a, b int) bool {
 	}
 	return resourcePodA.Cpu().Cmp(*resourcePodB.Cpu()) == -1
 }
+
+func MergeResources(resources ...v1.ResourceList) v1.ResourceList {
```
ellistarn, line 45: This is almost the same as scheduling.GetResources(), can we dedup?

pkg/cloudprovider/aws/fleet/packing/packing.go
```diff
@@ -77,34 +79,49 @@ func (p *podPacker) Pack(ctx context.Context, pods []*v1.Pod) ([]*Packing, error
...
+		}
+	})
+	nodeCapacitiesCopy := make([]*nodeCapacity, 0)
+	for _, nc := range nodeCapacities {
+		nodeCapacitiesCopy = append(nodeCapacitiesCopy, &nodeCapacity{nc.instanceType, nc.reserved.DeepCopy(), nc.total.DeepCopy()})
+	}
+	return nodeCapacitiesCopy
+}
+
+type packingsPerCapacity struct {
```
ellistarn, line 98: nit: joining words like fooPerBar or fooForBar raise naming red flags to me. What do you think about calling this a packingResult? "Result" implies that there's a range of potential outcomes, which makes sense to me w.r.t the fact that pods can be unpacked. 

Alternatively, we could do a diff function (in linear time, so no worse) and only return the packed pods and let the caller sort out the difference.

pkg/utils/binpacking/binpacking.go
prateekgogia: GetResources() is accepting a set of podSpec and returning the `ResourceList` for all the specs. Here we are combining a list of `ResourceList` into one. 
We would still need a function that will accept podSpecs and loop thorough all pod.Containers to sum the Quantities.

pkg/cloudprovider/aws/fleet/packing/packing.go
prateekgogia: Went with the first approach suggested.

pkg/cloudprovider/aws/fleet/capacity.go
```diff
@@ -38,7 +38,7 @@ type Capacity struct {
 // Create a set of nodes given the constraints.
 func (c *Capacity) Create(ctx context.Context, constraints *cloudprovider.Constraints) ([]cloudprovider.Packing, error) {
 	// 1. Compute Packing given the constraints
-	instancePackings, err := c.packer.Pack(ctx, constraints.Pods)
+	instancePackings, err := c.packer.Pack(ctx, constraints)
```
ellistarn, line 41: +1!

pkg/cloudprovider/aws/fleet/packing/nodecapacity.go
```diff
@@ -80,22 +94,20 @@ type nodeCapacity struct {
...
+	if nc.total.Cpu().Cmp(*targetUtilization.Cpu()) >= 0 &&
+		nc.total.Memory().Cmp(*targetUtilization.Memory()) >= 0 &&
+		nc.total.Pods().Cmp(*targetUtilization.Pods()) >= 0 {
+		nc.reserved = targetUtilization
 		return true
 	}
 	return false
 }
+
+func (nc *nodeCapacity) reserveForPod(podSpec *v1.PodSpec) bool {
```
ellistarn, line 109: IMO, it would be cleaner if we always modeled these as pods
```
reserve(daemonpod)
reserve(workloadpod)
reserve(kubelet_overhead_modeled_as_pod)
```
It's a bit of a hack, but WDYT?

Minimally, if we're keeping both, I'd call it

reserve(pod)
reserveOverhead(resources)

since pod is the common case.

prateekgogia:
```
reserve(daemonpod)
reserve(workloadpod)
reserve(kubelet_overhead_modeled_as_pod)
```

The problem with this is, workload pod needs to increment pod quantity in resources, but daemonset pods and kubelet don't need to specify this. Either the caller of this function will need to do this or we will need a wrapper around reserve to do this.

For the second suggestion, reserve(pod) will need to call reserveOverhead(resources)? Are we fine with this?

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -77,34 +78,46 @@ func (p *podPacker) Pack(ctx context.Context, pods []*v1.Pod) ([]*Packing, error
...
-	remainingPods := pods
-	// TODO reserve (Kubelet+ daemon sets) overhead for instance types
-	// TODO count number of pods created on an instance type
-	for _, nc := range p.getNodeCapacities() {
+	remainingPods := unpackedPods
+	for _, nc := range p.getNodeCapacities(constraints) {
 		// check how many pods we can fit with the available capacity
-		packedPods, unpackedPods := p.packPodsForCapacity(nc, pods)
-		if len(packedPods) == 0 {
+		result := p.packPodsForCapacity(nc, unpackedPods)
```
ellistarn, line 108: did you mean to pass remaining pods here?

ellistarn (approved):
Nicely done!

pkg/cloudprovider/aws/packing/packing.go
prateekgogia: Here we are passing all the pods which are unpacked and comparing, so, unpackedPods is fine.

PR #287, opened 2021-03-13
dynamically retrieve instance types

bwagner5:
Issue #, if available:
https://github.com/awslabs/karpenter/issues/286

Description of changes:
 - Fetch instance types dynamically w/ `ec2.DescribeInstanceTypes` and transforms to nodeCapacity information. 
 - Ran `go mod tidy` which updated some go.sum stuff


pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -78,6 +80,42 @@ func (p *podPacker) Pack(ctx context.Context, pods []*v1.Pod) ([]*Packing, error
...
+				Name:   aws.String("supported-virtualization-type"),
+				Values: []*string{aws.String("hvm")},
+			},
+		},
+	}
+	err := p.ec2.DescribeInstanceTypesPagesWithContext(context.TODO(), describeInstanceTypesInput, func(page *ec2.DescribeInstanceTypesOutput, lastPage bool) bool {
+		for _, instanceTypeInfo := range page.InstanceTypes {
+			nodeCapacity, err := instanceTypeInfoToNodeCapacity(*instanceTypeInfo)
+			if err != nil {
+				zap.S().Warnf("Failed to convert instanceTypeInfo to a nodeCapacity, %s", err.Error())
```
ellistarn, line 104: Maybe controversial, but I think we should fail fast and hard here.

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -78,6 +80,42 @@ func (p *podPacker) Pack(ctx context.Context, pods []*v1.Pod) ([]*Packing, error
...
+				continue
+			}
+			nodeCapacities = append(nodeCapacities, nodeCapacity)
+		}
+		return lastPage
+	})
+
+	if err != nil {
+		zap.S().Warnf("Failed to fetch instance types using ec2.DescribeInstanceTypes, %s", err.Error())
```
ellistarn, line 113: Likewise

pkg/cloudprovider/aws/packing/nodecapacity.go
```diff
@@ -74,6 +77,41 @@ var (
 	}
 )
 
+func instanceTypeInfoToNodeCapacity(instanceTypeInfo ec2.InstanceTypeInfo) (*nodeCapacity, error) {
```
ellistarn, line 80: Check out resource.MustParse. You shouldn't need to error afterwards.

pkg/cloudprovider/aws/packing/nodecapacity.go
```diff
@@ -74,6 +77,41 @@ var (
...
+	instanceTypeName := *instanceTypeInfo.InstanceType
+	vcpusInMillicores := fmt.Sprintf("%dm", *instanceTypeInfo.VCpuInfo.DefaultVCpus*1000)
```
ellistarn, line 82: Are cores ever non-integer? Otherwise you can drop the M

pkg/cloudprovider/aws/packing/packing.go
bwagner5: That's a fair point, I can change it. By failing fast and hard I assume you just mean to propagate an error up?

pkg/cloudprovider/aws/packing/nodecapacity.go
bwagner5: yeah I opted to use the error version to avoid the panic, but if we truly can't recover, maybe it's okay in this case? I usually try not to panic outside of pkg main.

pkg/cloudprovider/aws/packing/nodecapacity.go
bwagner5: ah good point, yeah cores are always ints

pkg/cloudprovider/aws/packing/packing.go
ellistarn: Depends when this runs. I see two approaches
1) Call only on startup and panic to fail fast
2) Call occasionally, with caching, similar subnets/launchtemplates, return an error in the reconcile loop on error

pkg/cloudprovider/aws/packing/nodecapacity.go
ellistarn: Yeah I follow the same, but with a bit more flexibility. If you can guarantee (in some way) that the data will be well formed (I think we can trust EC2 here), panicing can help keep the code clean. This is particularly important when go libraries put errors into functions when errors will only occur if you call them incorrectly, but won't error for other reasons (like a network call).

pkg/cloudprovider/aws/packing/packing.go
JacobGabrielson: I think #2 would be more useful in practice - there's no reason this API should fail on any continual basis.

pkg/cloudprovider/aws/packing/nodecapacity.go
```diff
@@ -15,14 +15,17 @@ limitations under the License.
...
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/awslabs/karpenter/pkg/utils/scheduling"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 // TODO get this information from node-instance-selector
 var (
-	nodeCapacities = []*nodeCapacity{
+	fallbackNodeCapacities = []*nodeCapacity{
```
JacobGabrielson, line 28: I think we should remove this, I don't think "fallback" will work in practice (maybe this is just temporary?)

pkg/cloudprovider/aws/packing/nodecapacity.go
JacobGabrielson: I think `MustParse` would be ok here, if EC2 were returning bad results we can't really proceed anyway.

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -78,6 +80,42 @@ func (p *podPacker) Pack(ctx context.Context, pods []*v1.Pod) ([]*Packing, error
 
 // TODO filter instance types based on node contraints like availability zones etc.
 func (p *podPacker) getNodeCapacities() []*nodeCapacity {
+	nodeCapacities := []*nodeCapacity{}
+	describeInstanceTypesInput := &ec2.DescribeInstanceTypesInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("processor-info.supported-architecture"),
+				Values: []*string{aws.String("x86_64")},
```
JacobGabrielson, line 88: we do also support arm, although I can add that in a later PR

pkg/cloudprovider/aws/packing/packing.go
bwagner5: 👍  This is definitely very cachable data. The data shouldn't change at a high velocity.   

Some reasons why this would fail would be rate limiting, which would obviously be temporary, and permissions issues.  I've seen that some paths aren't exercised regularly, so it may be nice to also call on startup all APIs we're using with dry-runs to ensure permissions are good on startup.

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -78,6 +80,42 @@ func (p *podPacker) Pack(ctx context.Context, pods []*v1.Pod) ([]*Packing, error
...
+				Name:   aws.String("supported-usage-class"),
+				Values: []*string{aws.String("on-demand")},
+			},
+			{
+				Name:   aws.String("supported-virtualization-type"),
+				Values: []*string{aws.String("hvm")},
+			},
+		},
+	}
+	err := p.ec2.DescribeInstanceTypesPagesWithContext(context.TODO(), describeInstanceTypesInput, func(page *ec2.DescribeInstanceTypesOutput, lastPage bool) bool {
```
JacobGabrielson, line 100: We should cache this result (similar to launch template cache), as the data changes very rarely and we don't want to spam the API.

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -78,6 +80,42 @@ func (p *podPacker) Pack(ctx context.Context, pods []*v1.Pod) ([]*Packing, error
...
+				continue
+			}
+			nodeCapacities = append(nodeCapacities, nodeCapacity)
+		}
+		return lastPage
+	})
+
+	if err != nil {
+		zap.S().Warnf("Failed to fetch instance types using ec2.DescribeInstanceTypes, %s", err.Error())
+		return fallbackNodeCapacities
```
JacobGabrielson, line 114: I think we should remove this - I don't think there's a reasonable fallback in this case (and in general fallback is a bad idea, see https://aws.amazon.com/builders-library/avoiding-fallback-in-distributed-systems/ for more rationale).

JacobGabrielson (changes requested):

pkg/cloudprovider/aws/packing/packing.go
JacobGabrielson: I think it's better - if there are permissions issues - to just very clearly fail with an error the operator can see and fix, I don't really see how the controller can proceed without this info (for example it doesn't really know what instance types are available in a given region unless it can call this).

bwagner5: ok, I'll make that change 👍

pkg/cloudprovider/aws/packing/packing.go
bwagner5: oh cool, I didn't see that. I can add it in here!

pkg/cloudprovider/aws/packing/packing.go
ellistarn: Also -- in the spirit of reconciliation, if permissions are missing, the controller should eventually recover if they're added -- it shouldn't require a restart or any manual intervention.

pkg/cloudprovider/aws/packing/packing.go
bwagner5: I haven't look at the launch template cache yet. I'll probably do a follow up PR to add that, if it's okay with you?

JacobGabrielson: Sounds good - can you add an issue?

pkg/cloudprovider/aws/packing/nodecapacity.go
```diff
@@ -15,77 +15,37 @@ limitations under the License.
...
-	nodeCapacities = []*nodeCapacity{
+func instanceTypeInfoToNodeCapacity(instanceTypeInfo ec2.InstanceTypeInfo) *nodeCapacity {
+	instanceTypeName := *instanceTypeInfo.InstanceType
+	vcpusInMillicores := resource.MustParse(fmt.Sprint(*instanceTypeInfo.VCpuInfo.DefaultVCpus * 1000))
+	memory := resource.MustParse(fmt.Sprintf("%dMi", *instanceTypeInfo.MemoryInfo.SizeInMiB))
+	// The number of pods per node is calculated using the formula:
+	//   max number of ENIs * (IPv4 Addresses per ENI -1) + 2
+	// https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt#L20
+	podCapacity := *instanceTypeInfo.NetworkInfo.MaximumNetworkInterfaces*(*instanceTypeInfo.NetworkInfo.Ipv4AddressesPerInterface-1) + 2
+	podCapacityResource := resource.MustParse(fmt.Sprint(podCapacity))
```
JacobGabrielson, line 34: nit: I'd probably inline all/most of these (it's kind of the house style for K.)

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -77,33 +90,76 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
 	return packings, nil
 }
 
-// TODO filter instance types based on node contraints like availability zones etc.
-func (p *podPacker) getNodeCapacities(constraints *cloudprovider.Constraints) []*nodeCapacity {
+func (p *podPacker) getNodeCapacities(constraints *cloudprovider.Constraints) ([]*nodeCapacity, error) {
 	result := make([]*nodeCapacity, 0)
```
JacobGabrielson, line 94: Optional: could you change this to not use `make`? I don't think our house style is to use `make` unless there's a reason (like the size needs to be set in advance).

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -77,33 +90,76 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
...
+			Name:   aws.String("supported-usage-class"),
+			Values: []*string{aws.String("on-demand")},
+		},
+		{
+			Name:   aws.String("supported-virtualization-type"),
+			Values: []*string{aws.String("hvm")},
+		},
+	}
+
+	instanceTypeConstraints := make([]*string, 0)
```
JacobGabrielson, line 139: Can you change this to just `var ... []*string`? We're not using `make` in the codebase unless there's some specific reason.

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -77,33 +90,76 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
...
+	instanceTypesFilter := &ec2.Filter{
+		Name:   aws.String("instance-type"),
+		Values: instanceTypeConstraints,
+	}
+	for _, instanceType := range constraints.InstanceTypes {
+		instanceTypesFilter.Values = append(instanceTypesFilter.Values, &instanceType)
+	}
+	filters = append(filters, instanceTypesFilter)
+	return filters
```
JacobGabrielson, line 148: I'd recommend in-lining this

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -77,33 +90,76 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
...
+	return result, nil
 }
 
-type packingResult struct {
-	packed   []*v1.Pod
-	unpacked []*v1.Pod
+func constraintsToDescribeInstanceTypesFilters(constraints *cloudprovider.Constraints) []*ec2.Filter {
+	architecture := "x86_64"
+	if constraints.Architecture != nil || *constraints.Architecture == provisioningv1alpha1.ArchitectureArm64 {
+		architecture = string(*constraints.Architecture)
```
JacobGabrielson, line 121: Have you checked to make sure this string (`"arm64"`) matches what EC2 wants?

pkg/cloudprovider/aws/packing/nodecapacity.go
```diff
@@ -15,77 +15,37 @@ limitations under the License.
...
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
-// TODO get this information from node-instance-selector
-var (
-	nodeCapacities = []*nodeCapacity{
+func instanceTypeInfoToNodeCapacity(instanceTypeInfo ec2.InstanceTypeInfo) *nodeCapacity {
+	instanceTypeName := *instanceTypeInfo.InstanceType
+	vcpusInMillicores := resource.MustParse(fmt.Sprint(*instanceTypeInfo.VCpuInfo.DefaultVCpus * 1000))
```
ellistarn, line 28: Since it's no longer millicores, I think we need to stop multiplying

pkg/cloudprovider/aws/packing/nodecapacity.go
ellistarn, line 34: I might just inline this:
```
pods := resource.MustParse(fmt.Sprint(
*instanceTypeInfo.NetworkInfo.MaximumNetworkInterfaces*(*instanceTypeInfo.NetworkInfo.Ipv4AddressesPerInterface-1) + 2
))
```

ellistarn: Actually -- now that there's no logic at all, you might as well just inline everything and let this be a one-line-returner.

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -77,33 +90,76 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
...
+	describeInstanceTypesInput := &ec2.DescribeInstanceTypesInput{
+		Filters: constraintsToDescribeInstanceTypesFilters(constraints),
+	}
+
+	err := p.ec2.DescribeInstanceTypesPagesWithContext(context.TODO(), describeInstanceTypesInput, func(page *ec2.DescribeInstanceTypesOutput, lastPage bool) bool {
+		for _, instanceTypeInfo := range page.InstanceTypes {
+			nc := instanceTypeInfoToNodeCapacity(*instanceTypeInfo)
+			kubeletOverhead := binpacking.CalculateKubeletOverhead(nc.total)
+			if ok := nc.reserve(resources.Merge(constraints.Overhead, kubeletOverhead)); !ok {
+				zap.S().Errorf("Failed to reserve kubelet overhead for node capacity type %v", nc.instanceType)
```
ellistarn, line 105: Do we want to error here? Is debug better? IIUC, this only errors if the theoretical node is full, which might happen multiple times in simulation.

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -77,33 +90,76 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
...
+		return nil, fmt.Errorf("fetching instance types using ec2.DescribeInstanceTypes, %w", err)
 	}
-	return result
+	return result, nil
 }
 
-type packingResult struct {
-	packed   []*v1.Pod
-	unpacked []*v1.Pod
+func constraintsToDescribeInstanceTypesFilters(constraints *cloudprovider.Constraints) []*ec2.Filter {
```
ellistarn, line 118: re: naming, I'm a fan of this style
```
filterFrom(constarints *cloudprovider.Constraints)
```
inline it will read like
```
filterFrom(constraints)
```

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -77,33 +90,76 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
...
+		{
+			Name:   aws.String("supported-virtualization-type"),
+			Values: []*string{aws.String("hvm")},
+		},
+	}
+
+	instanceTypeConstraints := make([]*string, 0)
+	instanceTypesFilter := &ec2.Filter{
+		Name:   aws.String("instance-type"),
+		Values: instanceTypeConstraints,
```
ellistarn, line 142: any reason to not inline
```
Values: []*string{},
```

ellistarn: Hmm what about 
https://docs.aws.amazon.com/sdk-for-go/api/aws/#StringSlice

pkg/cloudprovider/aws/fake/ec2api.go
```diff
@@ -97,3 +98,93 @@ func (a *EC2API) DescribeSecurityGroupsWithContext(context.Context, *ec2.Describ
 	}
 	return &ec2.DescribeSecurityGroupsOutput{SecurityGroups: []*ec2.SecurityGroup{{GroupId: aws.String("test-group")}}}, nil
 }
+
+func (a *EC2API) DescribeInstanceTypesPagesWithContext(ctx context.Context, input *ec2.DescribeInstanceTypesInput, fn func(*ec2.DescribeInstanceTypesOutput, bool) bool, opts ...request.Option) error {
```
ellistarn, line 102: if you make `fn` an errorable, you can return it directly
```
return fn(..., ...)
```

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -62,9 +71,13 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
 	sort.Sort(sort.Reverse(binpacking.ByResourcesRequested{SortablePods: constraints.Pods}))
 	packings := []*Packing{}
 	var packing *Packing
+	var err error
 	remainingPods := constraints.Pods
```
ellistarn, line 75: For parity, I might make this one a var too, so it reads
```
var foo
var bar
var baz
for ... {
  foo, bar, baz = ...
}
```

pkg/cloudprovider/aws/packing/packing.go
JacobGabrielson: I'd probably not consider it an error even - it means the node cannot handle kubelet overhead at all, so we shouldn't consider it - maybe make this a log info ("excluding ...")?

pkg/cloudprovider/aws/packing/packing.go
ellistarn: We might want to consider an instanceTypeProvider struct that follows the exact code patterns as Launch Template, Subnet, Security Group, etc.

pkg/cloudprovider/aws/packing/nodecapacity.go
bwagner5: that's an excellent point :)  may have rushed these changes a bit, apologies

pkg/cloudprovider/aws/packing/packing.go
bwagner5: Yeah, I actually had it not using make in the first rev and then changed it when I rebased because I saw that more code in that file was using `make`.  I prefer the `slice{}` syntax anyways, so I'll change these back.

pkg/cloudprovider/aws/packing/packing.go
bwagner5: yeah that was silly 👍

pkg/cloudprovider/aws/packing/packing.go
bwagner5: yep

```
processor-info.supported-architecture - The CPU architecture (arm64 | i386 | x86_64).
```

pkg/cloudprovider/aws/packing/packing.go
bwagner5: I think that makes sense for info level. I just brought this in from a rebase, so not sure if there was a reason for it being error.

pkg/cloudprovider/aws/packing/packing.go
bwagner5: I like that naming style, but the problem here is that the filters are specifically for the DescribeInstanceTypes call. So the return type doesn't give you enough information to infer where you can use it. 

Would this be ok? 

```
describeInstanceTypesFiltersFrom(constraints)
```

pkg/cloudprovider/aws/packing/packing.go
bwagner5: ohhhh I didn't realize those functions existed for slices too! Thanks!

pkg/cloudprovider/aws/fake/ec2api.go
bwagner5: by errorable, do you mean have `fn` be:

```
func(*ec2.DescribeInstanceTypesOutput, bool) err
```

instead of

```
func(*ec2.DescribeInstanceTypesOutput, bool) bool
```

If so, I can't do that since the ec2 api defines that func signature.

pkg/cloudprovider/aws/packing/packing.go
bwagner5: Here's the issue for adding the cache: https://github.com/awslabs/karpenter/issues/292

pkg/cloudprovider/aws/fake/ec2api.go
ellistarn: Ahh gotcha. Nevermind.

pkg/cloudprovider/aws/packing/nodecapacity.go
```diff
@@ -15,84 +15,39 @@ limitations under the License.
...
+		total: v1.ResourceList{
+			v1.ResourceCPU:    resource.MustParse(fmt.Sprint(*instanceTypeInfo.VCpuInfo.DefaultVCpus)),
+			v1.ResourceMemory: resource.MustParse(fmt.Sprintf("%dMi", *instanceTypeInfo.MemoryInfo.SizeInMiB)),
+			// The number of pods per node is calculated using the formula:
+			// max number of ENIs * (IPv4 Addresses per ENI -1) + 2
+			// https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt#L20
+			v1.ResourcePods: resource.MustParse(fmt.Sprint(
+				*instanceTypeInfo.NetworkInfo.MaximumNetworkInterfaces*(*instanceTypeInfo.NetworkInfo.Ipv4AddressesPerInterface-1) + 2)),
+		},
+		reserved: v1.ResourceList{
```
ellistarn, line 44: Do we also need to initialize pods to 0?

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -60,11 +68,15 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
...
-	packings := []*Packing{}
+	var packings []*Packing
 	var packing *Packing
+	var err error
 	remainingPods := constraints.Pods
 	for len(remainingPods) > 0 {
-		packing, remainingPods = p.packWithLargestPod(remainingPods, constraints)
+		packing, remainingPods, err = p.packWithLargestPod(remainingPods, constraints)
+		if err != nil {
+			return packings, err
```
ellistarn, line 78: Talk to Nick about our error annotation patterns. He should be able to share some ideas on whether or not this should be annotated.

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -77,33 +89,73 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
...
 // that fit; with their node capacities and list of leftover pods
-func (p *podPacker) packWithLargestPod(unpackedPods []*v1.Pod, constraints *cloudprovider.Constraints) (*Packing, []*v1.Pod) {
+func (p *podPacker) packWithLargestPod(unpackedPods []*v1.Pod, constraints *cloudprovider.Constraints) (*Packing, []*v1.Pod, error) {
 	bestPackedPods := []*v1.Pod{}
 	bestCapacities := []*nodeCapacity{}
 	remainingPods := unpackedPods
-	for _, nc := range p.getNodeCapacities(constraints) {
+	nodeCapacities, err := p.getNodeCapacities(constraints)
+	if err != nil {
+		return nil, bestPackedPods, err
```
ellistarn, line 156: nil, nil, err?

ellistarn (commented):
Nice work. More or less LGTM, a couple nits.

pkg/cloudprovider/aws/packing/packing.go
JacobGabrielson: Thanks, feel free to convert any `make...0` cases back to slice syntax, I think they're just oversights.

pkg/cloudprovider/aws/packing/packing.go
ellistarn: We're really judicious about Info level -- we can talk about the ideology here, maybe debug?

pkg/cloudprovider/aws/packing/packing.go
ellistarn: SGTM. I think once we move this into its own file (i.e. instance_type_provider), that sort of information may become implied by context. In this context, I agree.

pkg/cloudprovider/aws/packing/nodecapacity.go
bwagner5: actually the reserved can be removed all together since the reserve function is now implemented.

pkg/cloudprovider/aws/packing/packing.go
bwagner5: I don't have a super strong opinion, so I can go either way.  I think Jacob is right that it shouldn't ever happen the way the kubelet overhead is being calculated right now since it's a scaled a percentage. But I suppose if there was a weird instance type that was released, it could spam everyone using karpenter if it's at info level, so I'm fine with debug.

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -16,9 +16,13 @@ package packing
 
 import (
 	"context"
+	"fmt"
 	"sort"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+	provisioningv1alpha1 "github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha1"
```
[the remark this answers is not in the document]
bwagner5: sure! I can change the other spots as well to keep things consistent.

pkg/cloudprovider/aws/packing/packing.go
bwagner5: I chatted w/ Nick and annotated this err

pkg/cloudprovider/aws/packing/packing.go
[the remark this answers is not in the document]
bwagner5: fixed!

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -27,10 +31,14 @@ import (
 )
 
 type podPacker struct {
-	// TODO use this ec2 API to get the instance types
 	ec2 ec2iface.EC2API
 }
 
+type packingResult struct {
```
JacobGabrielson, line 37: I'm having a little trouble understanding this struct - it seems very close to a `Packing` (but doesn't include a `Packing`). It's also not always consistently returned by the "helper" functions. E.g., `packPodsForCapacity` returns it but `packWithLargestPod` does not (and, naively at least, it's not 100% clear why). I'm wondering if we could rationalize this a bit? Could put off to another PR, but wanted to bring it up.

pkg/cloudprovider/aws/packing/packing.go
```diff
@@ -77,33 +89,73 @@ func (p *podPacker) Pack(ctx context.Context, constraints *cloudprovider.Constra
...
+			Name:   aws.String("supported-virtualization-type"),
+			Values: []*string{aws.String("hvm")},
+		},
+	}
+	if len(constraints.InstanceTypes) != 0 {
+		instanceTypesFilter := &ec2.Filter{
+			Name:   aws.String("instance-type"),
+			Values: aws.StringSlice(constraints.InstanceTypes),
+		}
+		filters = append(filters, instanceTypesFilter)
```
JacobGabrielson, line 142: nit: I'd suggest inlining here

bwagner5: 👍  I don't know why I keep missing these inlining opportunities :)

pkg/cloudprovider/aws/packing/packing.go
bwagner5: I agree that it's confusing. I'm not entirely sure if's necessary. I'll just note that this PR didn't actually add it, I just moved the struct up to the top to live with the other structs, so I think it's best to hold off on changing it to another PR.

prateekgogia: This was part of the original PR I created for bin packing, I am working on the fix for this as part of the refactoring for this file.

JacobGabrielson (approved):
```
              |    |    |                 
             )_)  )_)  )_)              
            )___))___))___)\            
           )____)____)_____)\\
         _____|____|____|____\\\__
---------\                   /---------
  ^^^^^ ^^^^^^^^^^^^^^^^^^^^^
    ^^^^      ^^^^     ^^^    ^^
         ^^^^      ^^^
```

PR #290, opened 2021-03-15
Included yq in tools and upgraded to v4

ellistarn:
Issue #, if available:

Description of changes:


hack/codegen.sh
```diff
@@ -14,16 +14,11 @@ mv config/provisioning.karpenter.sh_provisioners.yaml config/provisioner-crd.yam
...
-yq w -i -d1 ./config/webhooks.yaml 'metadata.name' karpenter-validating-webhook-configuration
-yq w -i -d1 ./config/webhooks.yaml 'metadata.annotations[cert-manager.io/inject-ca-from]' karpenter/karpenter-serving-cert
-yq w -i -d1 ./config/webhooks.yaml 'webhooks[0].clientConfig.service.name' karpenter-webhook-service
-yq w -i -d1 ./config/webhooks.yaml 'webhooks[0].clientConfig.service.namespace' karpenter
-yq d -i -d1 ./config/webhooks.yaml 'webhooks[0].admissionReviewVersions[0]'
+yq e -i '.metadata.name = "karpenter-" + .metadata.name' ./config/webhooks.yaml
+yq e -i '.metadata.annotations = { "cert-manager.io/inject-ca-from" : "karpenter/karpenter-serving-cert" }' ./config/webhooks.yaml
+yq e -i '.webhooks[].clientConfig.service.name = "karpenter-webhook-service"' ./config/webhooks.yaml
+yq e -i '.webhooks[].clientConfig.service.namespace = "karpenter"' ./config/webhooks.yaml
+yq e -i 'del(.webhooks[].admissionReviewVersions[0])' ./config/webhooks.yaml
```
njtran, line 21: This command assumes that it is run at root, right? Do we run into problems if they don't run this command at root? For instance, if they're in `./hack` and run `./codegen.sh` they will create the yaml at a different file location than intended.

ellistarn: Correct -- nothing we can do about this without having a global concept for package root, which we don't have. In general, you can assume things should be executed from package root if not stated otherwise.

JacobGabrielson (approved):

PR #303, opened 2021-03-19
do not fail on create fleet launch errors

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - Fleet returns back launch errors that may mean that the request was successful eventually. This can happen if one of the instance types results in an ICE, but fleet retries and is able to launch the capacity with another instance type specified in the request.  We see some errors now because some instance types are not available in all AZs. With this change, we only error out if the capacity is also not met. Otherwise, we print these errors as warnings.

### DIFF
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -70,15 +70,15 @@ func (p *InstanceProvider) Create(ctx context.Context,
 	if err != nil {
 		return nil, fmt.Errorf("creating fleet %w", err)
 	}
-	// TODO aggregate errors
-	if count := len(createFleetOutput.Errors); count > 0 {
-		return nil, fmt.Errorf("errors while creating fleet, %v", createFleetOutput.Errors)
-	}
 	if count := len(createFleetOutput.Instances); count != 1 {
-		return nil, fmt.Errorf("expected 1 instance, but got %d", count)
+		return nil, fmt.Errorf("expected 1 instance, but got %d due to errors %v", count, createFleetOutput.Errors)
 	}
 	if count := len(createFleetOutput.Instances[0].InstanceIds); count != 1 {
-		return nil, fmt.Errorf("expected 1 instance ids, but got %d", count)
+		return nil, fmt.Errorf("expected 1 instance ids, but got %d due to errors %v", count, createFleetOutput.Errors)
+	}
+	// TODO aggregate errors
+	if count := len(createFleetOutput.Errors); count > 0 {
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/aws/instance.go
```diff
@@ -70,15 +70,15 @@ func (p *InstanceProvider) Create(ctx context.Context,
 	if err != nil {
```
JacobGabrielson, line 70: Just to double check if this `err` is non-nil, we are _guaranteed_ no instances launched?

pkg/cloudprovider/aws/instance.go
```diff
@@ -70,15 +70,15 @@ func (p *InstanceProvider) Create(ctx context.Context,
...
 		return nil, fmt.Errorf("creating fleet %w", err)
 	}
-	// TODO aggregate errors
-	if count := len(createFleetOutput.Errors); count > 0 {
-		return nil, fmt.Errorf("errors while creating fleet, %v", createFleetOutput.Errors)
-	}
 	if count := len(createFleetOutput.Instances); count != 1 {
-		return nil, fmt.Errorf("expected 1 instance, but got %d", count)
+		return nil, fmt.Errorf("expected 1 instance, but got %d due to errors %v", count, createFleetOutput.Errors)
```
JacobGabrielson, line 74: I think it would be good to have tests that simulate these cases ... since a lot rides on getting this right (both leaking instances and failing to launch them are pretty severe errors).

pkg/cloudprovider/aws/instance.go
bwagner5: There should be no instances launched if an error is thrown from the request.

pkg/cloudprovider/aws/instance.go
bwagner5: Agreed, I think we're lacking some module level test. I can try adding one for the instance provider in this PR or can do in a follow up.

JacobGabrielson: Can do in follow up

JacobGabrielson (approved):

pkg/cloudprovider/aws/instance.go
ellistarn, line 74: What does this error message look like?

bwagner5: ["Your requested instance type (c5a.4xlarge) is not supported in your requested Availability Zone (us-east-2a). Please retry your request by not specifying an Availability Zone or choosing us-east-2b, us-east-2c.", 
.... ]

PR #308, opened 2021-03-23
make easier to run commands if creating two stacks

JacobGabrielson:
Issue #, if available:

Description of changes: I don't think refer to the name of the nodegroup anywhere, so it's easier if we just let `eksctl` create a random name for us.


ellistarn (commented):
Does it randomize the name instead?

ellistarn (approved):

PR #310, opened 2021-03-24
'Simply reoriented infra' to remove unnecessary testing logic

ellistarn:
image


pkg/controllers/provisioning/v1alpha1/allocation/suite_test.go
```diff
@@ -67,127 +66,115 @@ var _ = AfterSuite(func() {
...
+			ExpectEventuallyReconciled(env.Client, provisioner)
 
 			nodes := &v1.NodeList{}
-			Expect(ns.Client.List(context.Background(), nodes)).To(Succeed())
+			Expect(env.Client.List(context.Background(), nodes)).To(Succeed())
 			Expect(len(nodes.Items)).To(Equal(1))
 			for _, object := range pods {
 				pod := v1.Pod{}
-				Expect(ns.Client.Get(context.Background(), client.ObjectKey{Name: object.GetName(), Namespace: object.GetNamespace()}, &pod)).To(Succeed())
+				Expect(env.Client.Get(context.Background(), client.ObjectKey{Name: object.GetName(), Namespace: object.GetNamespace()}, &pod)).To(Succeed())
```
njtran, line 107: You're using `context.Background()` a lot. Consider defining a context.Background() at the top and reusing it

njtran (approved):
One comment, otherwise, LGTM

pkg/controllers/provisioning/v1alpha1/allocation/suite_test.go
ellistarn: I'm wary of this contexts are cancellable, which means that sharing a context between tests may cause odd side effects if we ever choose to test cancellation. I could imagine defining a beforeEach with a context, but I'm not sure that simplifies things.

PR #313, opened 2021-03-25
Create cluster name specific roles

prateekgogia:
Issue #, if available:

Description of changes:


ellistarn (commented):
Assuming this breaks due to naming conflicts when setting up for multiple clusters. In general, I think breaking into multiple stacks makes sense. 

Are we missing two resources that will name collide?

JacobGabrielson (approved):

PR #316, opened 2021-03-26
Replaced context.TODO with actual contexts

njtran:
Looking for advice on what the default timeout for contexts should be. Ideally, any command we do shouldn't take anywhere close to 300 seconds. 

Replaced each context.TODO with proper contexts.
- Contexts will be created at the top level of each controller as a `context.Background()`
- Each time a context is used as a parameter, it only creates a new context if the command is a request (with timeout of 300 seconds)
  - Request to the API Server
  - Request to the AWS CLI 

### DIFF
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -20,13 +20,13 @@ import (
 	"math/rand"
 	"strings"
 
-	"go.uber.org/zap"
-	v1 "k8s.io/api/core/v1"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/awslabs/karpenter/pkg/packing"
+
+	"go.uber.org/zap"
+	v1 "k8s.io/api/core/v1"
 )
 
 type InstanceProvider struct {
@@ -55,7 +55,6 @@ func (p *InstanceProvider) Create(ctx context.Context,
 			})
 		}
 	}
-
 	// 2. Create fleet
 	createFleetOutput, err := p.ec2api.CreateFleetWithContext(ctx, &ec2.CreateFleetInput{
 		Type: aws.String(ec2.FleetTypeInstant),
--- a/pkg/cloudprovider/aws/instanceprofile.go
+++ b/pkg/cloudprovider/aws/instanceprofile.go
@@ -61,10 +61,10 @@ func (p *InstanceProfileProvider) getInstanceProfile(ctx context.Context, cluste
 		InstanceProfileName: aws.String(instanceProfileName),
 	})
 	if err != nil {
-		return nil, fmt.Errorf("retriving instance profile %s, %w", instanceProfileName, err)
+		return nil, fmt.Errorf("retrieving instance profile %s, %w", instanceProfileName, err)
 	}
 	for _, role := range output.InstanceProfile.Roles {
-		if err := p.addToAWSAuthConfigmap(role); err != nil {
+		if err := p.addToAWSAuthConfigmap(ctx, role); err != nil {
 			return nil, fmt.Errorf("adding role %s, %w", *role.RoleName, err)
 		}
 	}
@@ -73,9 +73,9 @@ func (p *InstanceProfileProvider) getInstanceProfile(ctx context.Context, cluste
 	return output.InstanceProfile, nil
 }
 
-func (p *InstanceProfileProvider) addToAWSAuthConfigmap(role *iam.Role) error {
+func (p *InstanceProfileProvider) addToAWSAuthConfigmap(ctx context.Context, role *iam.Role) error {
 	awsAuth := &v1.ConfigMap{}
-	if err := p.kubeClient.Get(context.TODO(), types.NamespacedName{Name: "aws-auth", Namespace: "kube-system"}, awsAuth); err != nil {
+	if err := p.kubeClient.Get(ctx, types.NamespacedName{Name: "aws-auth", Namespace: "kube-system"}, awsAuth); err != nil {
 		return fmt.Errorf("retrieving configmap aws-auth, %w", err)
 	}
 	if strings.Contains(awsAuth.Data["mapRoles"], *role.Arn) {
@@ -89,7 +89,7 @@ func (p *InstanceProfileProvider) addToAWSAuthConfigmap(role *iam.Role) error {
   - system:nodes
   rolearn: %s
   username: system:node:{{EC2PrivateDNSName}}`, *role.Arn)
-	if err := p.kubeClient.Update(context.TODO(), awsAuth); err != nil {
+	if err := p.kubeClient.Update(ctx, awsAuth); err != nil {
 		return fmt.Errorf("updating configmap aws-auth, %w", err)
 	}
 	zap.S().Debugf("Successfully patched configmap aws-auth with roleArn %s", *role.Arn)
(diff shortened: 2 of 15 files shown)


pkg/cloudprovider/aws/instance.go
```diff
@@ -51,9 +52,13 @@ func (p *InstanceProvider) Create(ctx context.Context,
 			})
 		}
 	}
+	cancelCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
 	// 2. Create fleet
-	createFleetOutput, err := p.ec2.CreateFleetWithContext(ctx, &ec2.CreateFleetInput{
```
ellistarn, line 56: We don't want this. We want to use the ctx passed in.

njtran: This uses the ctx passed in, but this file has it done incorrectly, sorry forgot one push.

njtran: Should be good now.

pkg/cloudprovider/aws/instanceprofile.go
```diff
@@ -56,14 +57,17 @@ func (p *InstanceProfileProvider) Get(ctx context.Context, cluster *v1alpha1.Clu
 }
 
 func (p *InstanceProfileProvider) getInstanceProfile(ctx context.Context, cluster *v1alpha1.ClusterSpec) (*iam.InstanceProfile, error) {
-	output, err := p.iam.GetInstanceProfileWithContext(ctx, &iam.GetInstanceProfileInput{
+	timeoutCtx, cancel := context.WithTimeout(ctx, time.Duration(300) * time.Second)
```
JacobGabrielson, line 60: This feels like a fair amount of duplication - eg it'll be hard to change later. I'm wondering if this is the right level to do this kind of timeout-setting, or if this function should just assume a `Context` gets passed in that has the right (overall) timeout set?

pkg/controllers/controller.go
```diff
@@ -17,7 +17,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-
```
ellistarn, line 20: I think your editor's style is fighting VScode here.

pkg/controllers/provisioning/v1alpha1/allocation/suite_test.go
```diff
@@ -77,9 +78,11 @@ var _ = Describe("Allocation", func() {
 				Cluster: &v1alpha1.ClusterSpec{Name: "test-cluster", Endpoint: "http://test-cluster", CABundle: "dGVzdC1jbHVzdGVyCg=="},
 			},
 		}
+		ctx = context.Background()
 	})
 
 	AfterEach(func() {
+		ctx.Done()
```
ellistarn, line 85: No need to do this, I think.

pkg/controllers/provisioning/v1alpha1/allocation/suite_test.go
```diff
@@ -77,9 +78,11 @@ var _ = Describe("Allocation", func() {
 				Cluster: &v1alpha1.ClusterSpec{Name: "test-cluster", Endpoint: "http://test-cluster", CABundle: "dGVzdC1jbHVzdGVyCg=="},
 			},
 		}
+		ctx = context.Background()
```
ellistarn, line 81: I know it might feel ugly, but I might just define these contexts inline. Up to you, but it's fewer lines of code.

pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go
```diff
@@ -86,6 +85,7 @@ func (u *Utilization) markUnderutilized(ctx context.Context, provisioner *v1alph
 		if err := u.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
 			return fmt.Errorf("patching node %s, %w", node.Name, err)
 		}
+
```
ellistarn, line 88: any reason?

ellistarn (changes requested):
Nice! A couple nits.

pkg/controllers/controller.go
njtran: I'm using `make ci` to do this, not using my IDEs formatter.

pkg/controllers/provisioning/v1alpha1/allocation/suite_test.go
njtran: I personally would prefer the two extra lines vs the ugliness of repeating context.Background() lol

pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go
njtran: Nope. I'll remove this.

pkg/cloudprovider/aws/instance.go
```diff
@@ -17,11 +17,10 @@ package aws
 import (
 	"context"
 	"fmt"
-	"math/rand"
-	"strings"
-
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
+	"math/rand"
+	"strings"
```
prateekgogia, line 23: I think we should look into whats causing this, because these are std library packages and the convention is -
```
std library imports

project-specific imports

third-party imports
```
Refer- https://github.com/golang/go/wiki/CodeReviewComments#imports

pkg/cloudprovider/aws/instanceprofile.go
```diff
@@ -28,6 +26,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
```
ellistarn, line 29: This is in the wrong place. It should be up by other golang native imports.

prateekgogia (changes requested):

prateekgogia:
Maybe I missed this, but its not clear to me where are we setting the Deadlines/timeouts?

njtran:
@prateekgogia the api calls we make to the APIServer and calls we use in client-go and controller-runtime each have their own timeouts, so we don't need to implement them. We just need to make sure that we're populating the context down correctly.

pkg/cloudprovider/aws/instanceprofile.go
njtran: I agree. I just ran `make ci` and it moved the strings down there, so that it could sort it alphabetically. In other words, to separate the imports, it's important to have a newline before running, or it won't sort it properly.

pkg/cloudprovider/aws/instance.go
njtran: Looks like `make ci` doesn't sort the imports by separating it like that. It only sorts the imports alphabetically unless there's a newline in between. Maybe this is auto done for VSCode, but not for Goland.

njtran:
Looks like any AWS SDK for Go doesn't implement timeouts. Good callout Prateek. While there is a timeout for the APIGateway API, there doesn't appear to be one for EC2. https://aws.amazon.com/premiumsupport/knowledge-center/lambda-function-retry-timeout-sdk/ This is why some functions have the ability to pass in a context, so that you can implement the timeout yourself.

njtran:
@prateekgogia Given separate offline discussions with @ellistarn and @JacobGabrielson, going to hold off on doing timeouts for AWS SDK API calls.

ellistarn (approved):
Let's look into timeouts later.

prateekgogia (approved):
LGTM!

JacobGabrielson (approved):

PR #317, opened 2021-03-29
cordon and add draining label to nodes that are already cordoned

bwagner5:
Issue #, if available:
N/A

Description of changes:
Cordon and add draining label to nodes that have already been cordoned. This will allow the aws-node-termination-handler (NTH) to work with karpenter. NTH will cordon and drain instance when spot interruptions or rebalance events occur. Before, karpenter would never terminate these nodes because karpenter wasn't the one to cordon them. 

### DIFF
--- a/pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
+++ b/pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
@@ -17,6 +17,7 @@ package reallocation
 import (
 	"context"
 	"fmt"
+
 	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha1"
 	"github.com/awslabs/karpenter/pkg/cloudprovider"
 	"github.com/awslabs/karpenter/pkg/utils/functional"
@@ -60,18 +61,16 @@ func (t *Terminator) cordonNodes(ctx context.Context, provisioner *v1alpha1.Prov
 	}
 	// 2. Cordon nodes
 	for _, node := range nodeList {
-		if !node.Spec.Unschedulable {
-			persisted := node.DeepCopy()
-			node.Spec.Unschedulable = true
-			node.Labels = functional.UnionStringMaps(
-				node.Labels,
-				map[string]string{v1alpha1.ProvisionerPhaseLabel: v1alpha1.ProvisionerDrainingPhase},
-			)
-			if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
-				return fmt.Errorf("patching node %s, %w", node.Name, err)
-			}
-			zap.S().Debugf("Cordoned node %s", node.Name)
+		persisted := node.DeepCopy()
+		node.Spec.Unschedulable = true
+		node.Labels = functional.UnionStringMaps(
+			node.Labels,
+			map[string]string{v1alpha1.ProvisionerPhaseLabel: v1alpha1.ProvisionerDrainingPhase},
+		)
+		if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
+			return fmt.Errorf("patching node %s, %w", node.Name, err)
 		}
+		zap.S().Debugf("Cordoned node %s", node.Name)
 	}
 	return nil
 }


ellistarn (approved):
+1 for this change (I remember commenting something to this effect the first time around). Love the idempotent style of progressing through the workflow. I'll let @njtran drop an approval before merging, too

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -60,18 +61,16 @@ func (t *Terminator) cordonNodes(ctx context.Context, provisioner *v1alpha1.Prov
...
+		persisted := node.DeepCopy()
+		node.Spec.Unschedulable = true
+		node.Labels = functional.UnionStringMaps(
+			node.Labels,
+			map[string]string{v1alpha1.ProvisionerPhaseLabel: v1alpha1.ProvisionerDrainingPhase},
+		)
+		if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
+			return fmt.Errorf("patching node %s, %w", node.Name, err)
 		}
+		zap.S().Debugf("Cordoned node %s", node.Name)
```
ellistarn, line 73: Is it possible that this logs too frequently?

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -60,18 +61,16 @@ func (t *Terminator) cordonNodes(ctx context.Context, provisioner *v1alpha1.Prov
...
-				return fmt.Errorf("patching node %s, %w", node.Name, err)
-			}
-			zap.S().Debugf("Cordoned node %s", node.Name)
+		persisted := node.DeepCopy()
+		node.Spec.Unschedulable = true
+		node.Labels = functional.UnionStringMaps(
+			node.Labels,
+			map[string]string{v1alpha1.ProvisionerPhaseLabel: v1alpha1.ProvisionerDrainingPhase},
+		)
+		if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
```
ellistarn, line 70: Is it possible that this sends unnecessary requests? Should we check if the patch is empty or not?

ellistarn (changes requested):
removing approval

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
bwagner5: I don't think that's necessary, for the same reason as above. After a node is labeled with the `v1alpha1.ProvisionerPhaseLabel` it won't be cordoned again. The cordon should only be logged once per node that is cordoned.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
bwagner5: I don't think so. I think the log makes sense as a debug message and is only logging when necessary.  After a node is labeled with the `v1alpha1.ProvisionerPhaseLabel` it won't be cordoned again. The cordon should only be logged once per node that is cordoned.

JacobGabrielson: This should maybe actually be an Info? Is there going to be any other log line indicating that the node was cordoned by Karpenter?

bwagner5: It could get unnecessarily noisy in a large cluster. wrt to investigating who cordoned the node though, that should be fairly easy with the label that's added when cordoning nodes. In that situation, it really only matters if karpenter cordoned the node but for some reason didn't terminate it. If it's terminated, the log is at info level anyways so I don't think it's a strong reason to move this log up to info.

ellistarn: Agree. I want to be really stingy with Info logs. As we get debugging feedback, we can think about promoting some debug logs to info. We discussed this a month or so back and decided that top level aggregate steps (e.g. "cordoned 5 nodes") would be info level, while individual workflow steps (e.g. "cordoned node foo") would be debug.

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
ellistarn: Ah you're right, since we change the phase to draining, we don't need to filter. At one point, we were discussing collapsing Draining and Terminable into a single label value, so this would've been necessary.

ellistarn (approved):

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
JacobGabrielson: That's fine - as long as there's a way to figure out what entity did the cordoning.

JacobGabrielson (approved):

njtran (approved):
LGTM. Good catch on this! No more comments from me.

PR #318, opened 2021-03-30
Preliminary unit tests for functional package

JacobGabrielson:
Issue #, if available: n/a

Description of changes: preliminary unit tests for `functional` pkg

### DIFF
--- a/pkg/utils/functional/suite_test.go
+++ b/pkg/utils/functional/suite_test.go
@@ -14,61 +14,58 @@ func TestFunctional(t *testing.T) {
 
 var _ = Describe("Functional", func() {
 	Context("UnionStringMaps", func() {
-		empty := make(map[string]string)
+		empty := map[string]string{}
+		original := map[string]string{
+			"a": "b",
+			"c": "d",
+		}
+		overwriter := map[string]string{
+			"a": "y",
+			"c": "z",
+		}
+		disjoiner := map[string]string{
+			"d": "y",
+			"e": "z",
+		}
+		uberwriter := map[string]string{
+			"d": "q",
+			"e": "z",
+		}
 
 		Specify("no args returns empty", func() {
 			Expect(UnionStringMaps()).To(BeEmpty())
 		})
 
-		Specify("all empty returns empty", func() {
+		Specify("multiple empty returns empty", func() {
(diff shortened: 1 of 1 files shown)


ellistarn (dismissed):
I'm not sure I'm sold on this heavy usage of the DLS (e.g. When, Specify, etc).

I'd probably use the minimum necessary, e.g.
```
Describe("Suite", func() {
  It("should be a test case", func() {
  })
})
```
I could see using additional context blocks as necessary to setup variables at different scopes, but if no setup is being done within a block, we might want to collapse it. WDYT?

JacobGabrielson:
> I'm not sure I'm sold on this heavy usage of the DLS (e.g. When, Specify, etc).
> 
> I'd probably use the minimum necessary, e.g.
> 
> ```
> Describe("Suite", func() {
>   It("should be a test case", func() {
>   })
> })
> ```
> 
> I could see using additional context blocks as necessary to setup variables at different scopes, but if no setup is being done within a block, we might want to collapse it. WDYT?

Makes sense - cut down on levels of indirection, LMKWYT?

pkg/utils/functional/suite_test.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+func TestFunctional(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Functional Suite")
+}
+
+var _ = Describe("Functional", func() {
+	Context("UnionStringMaps", func() {
+		empty := make(map[string]string)
```
ellistarn, line 17: I'd probably just define the maps inline. [string]string{} is only a few extra characters and more explicit.

ellistarn, line 17: It would be more logical to me if this were initialized as a sibling to the other canonical maps.

ellistarn (approved):
One optional nit.

ellistarn (dismissed):

pkg/utils/functional/suite_test.go
JacobGabrielson: Fixing

JacobGabrielson:
> One optional nit.

fix'd

ellistarn (approved):

PR #324, opened 2021-03-31
minimize logging and simplify instance types provider

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - Restructured instance type provider to reduce ec2 InstanceTypeOffering API calls
   - `2021-03-31T00:17:09.790Z	DEBUG	Successfully discovered 392 EC2 instance types` is only logged one time per cache miss. 
 - Fixed bad log on normalizeZones, no need to log in that func at all.  
 - Retrieving all zones and caching as instance type with supported zones allowed a simplification of the provider (removing vpc provider dependency).

### DIFF
--- a/pkg/cloudprovider/aws/capacity.go
+++ b/pkg/cloudprovider/aws/capacity.go
@@ -94,7 +94,7 @@ func (c *Capacity) Delete(ctx context.Context, nodes []*v1.Node) error {
 }
 
 func (c *Capacity) GetInstanceTypes(ctx context.Context) ([]string, error) {
-	return c.instanceTypeProvider.GetAllInstanceTypeNames(ctx, c.spec.Cluster.Name)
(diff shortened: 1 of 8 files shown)


pkg/cloudprovider/aws/capacity.go
```diff
@@ -94,7 +94,7 @@ func (c *Capacity) Delete(ctx context.Context, nodes []*v1.Node) error {
 }
 
 func (c *Capacity) GetInstanceTypes(ctx context.Context) ([]string, error) {
-	return c.instanceTypeProvider.GetAllInstanceTypeNames(ctx, c.spec.Cluster.Name)
+	return c.instanceTypeProvider.GetAllInstanceTypeNames(ctx)
```
ellistarn, line 97: yey

PR #328, opened 2021-03-31
Support spot w/ capacityType

bwagner5:
Issue #, if available:
https://github.com/awslabs/karpenter/issues/305

Description of changes:
 - Initial support for capacityTypes which mainly includes spot
 - I implemented capacityType as a cloudprovider specific label
 - Capacity-optimized-prioritized allocation strategy is being used to somewhat mitigate the risk of provisioning large instances that are not efficiently packed. The priorities are currently defined by the euclidean distance from the origin on vcpus and memory dimensions of the instance types. This is a pretty close approximation for price with the instance types that are currently supported (and the capacity-optimized part of the allocation strategy will ensure we get placed in a deep pool for reduced interruptions and likely a lower price anyways).  We may decide that it's advantageous to use the EC2 Spot Pricing History API to prioritize directly by pricing, but that will likely be in a followup PR if needed.


pkg/apis/provisioning/v1alpha1/provisioner.go
```diff
@@ -83,6 +83,10 @@ type Constraints struct {
 	// OperatingSystem constrains the underlying node operating system
 	// +optional
 	OperatingSystem *string `json:"operatingSystem,omitempty"`
+	// CapacityType constrains what type of capacity is launched by the Provisioner
+	// If unspecified, it will launch on-demand nodes.
+	// +optional
+	CapacityType *string `json:"capacityType,omitempty"`
```
ellistarn, line 89: Is there an upstream kubernetes concept for this? If not, we probably want to bake this into the cloud provider instead

pkg/cloudprovider/types.go
```diff
@@ -43,6 +43,10 @@ type Capacity interface {
 	GetArchitectures(ctx context.Context) ([]string, error)
 	// GetOperatingSystems returns the operating systems supported by the cloud provider.
 	GetOperatingSystems(ctx context.Context) ([]string, error)
+	// GetCapacityTypes returns the capacity types supported by the cloud provider.
+	GetCapacityTypes(ctx context.Context) ([]string, error)
+	// DefaultCapacityType returns the default capacity type for the cloud provider
+	DefaultCapacityType(ctx context.Context) (string, error)
```
ellistarn, line 49: If we go this route of making it a first class concept, we could follow a convention of "default first" to avoid extra API calls.

pkg/webhooks/provisioning/v1alpha1/defaulter.go
```diff
@@ -49,20 +53,39 @@ func (v *Defaulter) Handle(ctx context.Context, req admission.Request) admission
 	if err != nil {
 		return admission.Errored(http.StatusBadRequest, err)
 	}
-	v.applyDefaults(&provisioner.Spec)
+	if err := v.applyDefaults(ctx, &provisioner.Spec); err != nil {
+		return admission.Errored(0, fmt.Errorf("applying defaults to provisioner, %w", err))
```
ellistarn, line 57: Can you use a canonical http status code here?

pkg/webhooks/provisioning/v1alpha1/defaulter.go
```diff
@@ -49,20 +53,39 @@ func (v *Defaulter) Handle(ctx context.Context, req admission.Request) admission
...
 	if err != nil {
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
 	return admission.PatchResponseFromRaw(req.Object.Raw, marshaled)
 }
 
-func (v *Defaulter) applyDefaults(spec *provisioning.ProvisionerSpec) {
-	v.defaultTTL(spec)
+func (v *Defaulter) applyDefaults(ctx context.Context, spec *provisioning.ProvisionerSpec) error {
+	return functional.ValidateAll(
```
ellistarn, line 68: If we use this in this way, we should rename that function to something else. It used to be AllSucceed/SucceedAll

pkg/webhooks/provisioning/v1alpha1/defaulter.go
```diff
@@ -49,20 +53,39 @@ func (v *Defaulter) Handle(ctx context.Context, req admission.Request) admission
...
 		spec.TTLSeconds = ptr.Int32(300)
 	}
+	return nil
+}
+
+func (v *Defaulter) defaultCapacityType(ctx context.Context, spec *provisioning.ProvisionerSpec) error {
+	if spec.CapacityType != nil {
+		return nil
+	}
+	capacityType, err := v.CloudProvider.CapacityFor(spec).DefaultCapacityType(ctx)
```
ellistarn, line 85: Do we need to default anything into the provisioner at all? Can we simply let unspecified allow the controller to do what it will?

pkg/webhooks/provisioning/v1alpha1/defaulter.go
```diff
@@ -49,20 +53,39 @@ func (v *Defaulter) Handle(ctx context.Context, req admission.Request) admission
...
 
-func (v *Defaulter) defaultTTL(spec *provisioning.ProvisionerSpec) {
+func (v *Defaulter) defaultTTL(spec *provisioning.ProvisionerSpec) error {
 	if spec.TTLSeconds == nil {
 		spec.TTLSeconds = ptr.Int32(300)
 	}
+	return nil
+}
+
+func (v *Defaulter) defaultCapacityType(ctx context.Context, spec *provisioning.ProvisionerSpec) error {
```
ellistarn, line 81: Would love to figure out how to make these defaulters not errorable.

pkg/apis/provisioning/v1alpha1/provisioner.go
bwagner5: There is not currently... but shouldn't there be?!  All the major cloud providers have this concept, so it feels like it should be top-level and a k8s thing.  

Looks like there was a very brief discussion on adding one: https://github.com/kubernetes/kubernetes/issues/35166 

If there were a "node.kubernetes.io/capacity-type: spot" , it would be possible to add the percentage to different capacity types in the deployment controller.

pkg/cloudprovider/types.go
bwagner5: interesting, are you concerned about bloating the cloud provider interface or of making unnecessary external API calls (since in this case, the types are hardcoded) ?

pkg/webhooks/provisioning/v1alpha1/defaulter.go
bwagner5: ha yeah... 🙈

pkg/webhooks/provisioning/v1alpha1/defaulter.go
bwagner5: 👍

pkg/webhooks/provisioning/v1alpha1/defaulter.go
bwagner5: So the options are:
 1) Default explicitly by calling a DefaultCapacityType()
 2) Default based on first entry in GetCapacityTypes()
 3) Or allow for empty capacityType and let the cloud provider decide 

I'm not sure which one I like the best... I like the explicit nature of 1, 2 requires less fluff in the cloudprovider interface. 

3 seems like it could be confusing. Especially since we've had discussions on spot being default or on-demand being default... so cloudproviders deciding seems too unpredictable.

pkg/webhooks/provisioning/v1alpha1/defaulter.go
bwagner5: I think it's fair to just remove the error. I had it this way in case the cloud provider wanted to dynamically retrieve the possible capacity types.  I don't think they'll change very often... unless we would put placement groups into the capacityType umbrella...

pkg/cloudprovider/aws/capacity.go
```diff
@@ -112,3 +117,14 @@ func (c *Capacity) GetOperatingSystems(ctx context.Context) ([]string, error) {
 		v1alpha1.OperatingSystemLinux,
 	}, nil
 }
+
+func (c *Capacity) GetCapacityTypes(ctx context.Context) ([]string, error) {
```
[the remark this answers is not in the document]
bwagner5: yeah (I'll repost from above comment): I had it this way in case the cloud provider wanted to dynamically retrieve the possible capacity types. I don't think they'll change very often... unless we would put placement groups into the capacityType umbrella which would require the dynamic lookup.

pkg/apis/provisioning/v1alpha1/provisioner.go
```diff
@@ -103,6 +107,7 @@ var (
 	ProvisionerNameLabelKey      = SchemeGroupVersion.Group + "/name"
 	ProvisionerNamespaceLabelKey = SchemeGroupVersion.Group + "/namespace"
 	ProvisionerPhaseLabel        = SchemeGroupVersion.Group + "/lifecycle-phase"
+	CapacityTypeLabelKey         = SchemeGroupVersion.Group + "/capacityType"
```
JacobGabrielson, line 110: `capacity-type` seems like it would be more consistent

pkg/cloudprovider/types.go
JacobGabrielson: I guess I'm not 100% sure what the `DefaultCapacityType` method is needed for?

pkg/webhooks/provisioning/v1alpha1/defaulter.go
JacobGabrielson: I'd say something like #3 makes the most sense. The default should be on-demand and cloud providers should default to on-demand.

pkg/apis/provisioning/v1alpha1/provisioner.go
bwagner5: Based on conversations today and reading Jacob's proposal in https://github.com/awslabs/karpenter/pull/330, I attempted a cloud provider implementation for `capacity-type`.

pkg/cloudprovider/aws/capacity.go
```diff
@@ -112,3 +145,20 @@ func (c *Capacity) GetOperatingSystems(ctx context.Context) ([]string, error) {
...
+	for key, value := range labels {
+		if strings.HasPrefix(key, nodeLabelPrefix) {
+			switch key {
+			case capacityTypeLabel:
+				capacityTypes := []string{capacityTypeSpot, capacityTypeOnDemand}
+				if !functional.ContainsString(capacityTypes, value) {
+					return fmt.Errorf("%s must be one of %v", key, capacityTypes)
+				}
+			default:
+				return fmt.Errorf("%s/* is reserved for AWS cloud provider use", nodeLabelPrefix)
```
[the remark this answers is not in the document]
bwagner5: Yes, this is used to validate the cloud-provider specific labels on the provisioner inside of the webhook validator.

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -146,16 +145,31 @@ func (p *InstanceTypeProvider) getAllInstanceTypes(ctx context.Context) ([]*ec2.
...
 	filtered := []*packing.Instance{}
 	architecture := utils.NormalizeArchitecture(constraints.Architecture)
 
 	for _, instanceTypeInfo := range instanceTypes {
-		if (len(constraints.InstanceTypes) == 0 || functional.ContainsString(constraints.InstanceTypes, *instanceTypeInfo.InstanceType)) &&
-			(len(constraints.InstanceTypes) != 0 || p.isDefaultInstanceType(instanceTypeInfo)) &&
-			(architecture == nil || functional.ContainsString(aws.StringValueSlice(instanceTypeInfo.ProcessorInfo.SupportedArchitectures), *architecture)) &&
-			functional.ContainsString(aws.StringValueSlice(instanceTypeInfo.SupportedUsageClasses), "on-demand") &&
-			(len(zones) == 0 || len(functional.IntersectStringSlice(instanceTypeInfo.Zones, zones)) > 0) {
+		if functional.All(
```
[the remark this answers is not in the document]
bwagner5: yeah... I kind of agree.  If we go back to the `if` we'd need to keep upping the cyclo-complexity check in golangci-lint. It's currently set at 10 and this extra conditional sent it over the threshold.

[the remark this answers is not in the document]
bwagner5: another way to scale this might be to put these funcs as private methods on the InstanceTypeProvider type and go back to a regular `If` or define a func that takes in an arbitrary list of bools:

```
   if isInstanceTypeSupported(constraints.InstanceTypes, instanceTypeInfo) && 
      isArchitectureSupported(architecture, instanceTypeInfo) &&
      isCapacityTypeSupported(constraints.CapacityType, instanceTypeInfo) &&
     isZonesSupported(zones, instanceTypeInfo) {
     ...
     }
   
```

pkg/cloudprovider/aws/capacity.go
```diff
@@ -17,13 +17,26 @@ package aws
...
 	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha1"
 	"github.com/awslabs/karpenter/pkg/cloudprovider"
 	"github.com/awslabs/karpenter/pkg/packing"
+	"github.com/awslabs/karpenter/pkg/utils/functional"
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
 )
 
+const (
+	nodeLabelPrefix = "node.k8s.aws"
```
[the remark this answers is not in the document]
bwagner5: ah yeah, this was left over from the previous impl where cloudprovider.Constraints took a pointer to a capacityType, I'll fix this

pkg/cloudprovider/aws/capacity.go
```diff
@@ -35,34 +48,54 @@ type Capacity struct {
 	instanceTypeProvider   *InstanceTypeProvider
 }
 
+// AWSConstraints are AWS specific constraints
+type AWSConstraints struct {
+	cloudprovider.Constraints
+	CapacityType string
+}
+
+func NewAWSConstraints(constraints *cloudprovider.Constraints) AWSConstraints {
```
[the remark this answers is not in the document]
bwagner5: 👍

pkg/cloudprovider/aws/instancetypes.go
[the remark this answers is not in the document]
JacobGabrielson: Hmm... I'm not sure the cyclo complexity check makes sense then?

pkg/cloudprovider/aws/capacity.go
```diff
@@ -17,13 +17,27 @@ package aws
...
+	"github.com/awslabs/karpenter/pkg/utils/functional"
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
 )
 
+const (
+	nodeLabelPrefix      = "node.k8s.aws"
+	capacityTypeSpot     = "spot"
+	capacityTypeOnDemand = "on-demand"
+	maxInstanceTypes     = 10
```
JacobGabrielson, line 34: What is this for? Could use a comment on what it does and why `10`?

pkg/cloudprovider/aws/capacity.go
```diff
@@ -35,34 +49,57 @@ type Capacity struct {
...
 	}
 
-	// 4. Create Instances
+	// 5. Create Instances
 	var instanceIDs []*string
 	podsForInstance := make(map[string][]*v1.Pod)
 	for _, packing := range instancePackings {
-		instanceID, err := c.instanceProvider.Create(ctx, launchTemplate, packing.InstanceTypes, zonalSubnetOptions)
+		if len(packing.InstanceTypes) > maxInstanceTypes {
+			packing.InstanceTypes = packing.InstanceTypes[:maxInstanceTypes]
```
JacobGabrielson, line 100: Could use a comment explaining why this is needed? Also is it guaranteed that the packings "at the end" are the best ones to discard?

pkg/cloudprovider/aws/instance.go
```diff
@@ -35,33 +35,46 @@ type InstanceProvider struct {
 }
 
 // Create an instance given the constraints.
+// instanceTypeOptions should be sorted by priority for spot capacity type.
```
JacobGabrielson, line 38: What if Spot isn't being used?

pkg/cloudprovider/aws/instance.go
```diff
@@ -35,33 +35,46 @@ type InstanceProvider struct {
...
-			DefaultTargetCapacityType: aws.String(ec2.DefaultTargetCapacityTypeOnDemand),
+			DefaultTargetCapacityType: aws.String(capacityType),
 			TotalTargetCapacity:       aws.Int64(1),
 		},
+		OnDemandOptions: &ec2.OnDemandOptionsRequest{
+			AllocationStrategy: aws.String(ec2.FleetOnDemandAllocationStrategyLowestPrice),
+		},
+		SpotOptions: &ec2.SpotOptionsRequest{
+			AllocationStrategy: aws.String(ec2.SpotAllocationStrategyCapacityOptimizedPrioritized),
+		},
```
JacobGabrielson, line 77: just double-checking - is it ok to specify this when not requesting spot capacity? (ditto for on-demand options when requesting Spot capacity)

pkg/packing/packing.go
```diff
@@ -79,6 +79,8 @@ func (p *packer) Pack(ctx context.Context, pods []*v1.Pod, instanceTypes []*Inst
 			continue
 		}
 		packings = append(packings, packing)
+		// sort instance type packings by vcpus and memory resources
```
JacobGabrielson, line 82: nit: I'd move the comment above the function definition

pkg/webhooks/provisioning/v1alpha1/validator.go
```diff
@@ -97,7 +97,7 @@ func (v *Validator) validateLabels(ctx context.Context, spec *v1alpha1.Provision
 			}
 		}
 	}
-	return nil
+	return v.CloudProvider.CapacityFor(spec).ValidateLabels(spec.Labels)
```
JacobGabrielson, line 100: This might be not a problem, but thought I'd ask. It does seem potentially a bit odd to pass in `spec.Labels` given that the cloud provider already has a copy of the entire `ProvisionerSpec` so could just access the labels from there? Maybe it would make more sense for the cloud provider to just have a chance to validate the ProvisionerSpec?

pkg/cloudprovider/aws/instance.go
bwagner5: Yep, it's been tested and works

pkg/cloudprovider/aws/instance.go
bwagner5: If spot isn't used, then we don't want to prioritize the instance types (since we're using lowest-price), so it doesn't really matter if they're passed in sorted or not.  I can make that more explicit on the func comment.

pkg/cloudprovider/aws/capacity.go
bwagner5: Sure, I'll add a comment explaining why this is needed, but it's basically because the fleet request size can grow to be too large and get rejected by the EC2 API, if we pass in say 200+ instance types.  I started it at 10 (which will pass in `10 instance types * azs to fleet) since that is still quite a flexible request.  I'm not opposed to changing the number though to something else. 

And yes, it is guaranteed that the instanceTypes packings at the end are the best ones to drop from a point-in-time packing efficiency perspective. The packing pkg returns the instance types sorted by ascending resources.

pkg/cloudprovider/aws/instancetypes.go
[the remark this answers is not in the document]
ellistarn: I want to keep the check in place. I like the most recent approach brandon posted.

pkg/cloudprovider/aws/capacity.go
```diff
@@ -35,34 +49,57 @@ type Capacity struct {
 	instanceTypeProvider   *InstanceTypeProvider
 }
 
+// AWSConstraints are AWS specific constraints
+type AWSConstraints struct {
```
ellistarn, line 53: I like this concept. Could we make a new file to hold this struct?

pkg/cloudprovider/aws/capacity.go
```diff
@@ -35,34 +49,57 @@ type Capacity struct {
...
+		CapacityType: capacityType,
+	}
+}
+
 // Create a set of nodes given the constraints.
 func (c *Capacity) Create(ctx context.Context, constraints *cloudprovider.Constraints) ([]cloudprovider.Packing, error) {
-	// 1. Retrieve normalized zones from constraints or all zones that the cluster spans if no zonal constraints are specified
-	zonalSubnetOptions, err := c.vpcProvider.GetZonalSubnets(ctx, constraints, c.spec.Cluster.Name)
+	// 1. Create AWS Cloud Provider constraints and apply defaults
+	awsConstraints := NewAWSConstraints(constraints)
```
ellistarn, line 72: consider calling the other one cloudprovider constraints

pkg/cloudprovider/aws/capacity.go
ellistarn: Ascending resources is a bit tricky since it's high dimensional. Do we want to prioritize CPU? Memory? Euclidean?

pkg/cloudprovider/aws/capacity.go
```diff
@@ -35,34 +49,57 @@ type Capacity struct {
...
 		return nil, fmt.Errorf("getting launch template, %w", err)
 	}
 
-	// 4. Create Instances
+	// 5. Create Instances
 	var instanceIDs []*string
 	podsForInstance := make(map[string][]*v1.Pod)
 	for _, packing := range instancePackings {
-		instanceID, err := c.instanceProvider.Create(ctx, launchTemplate, packing.InstanceTypes, zonalSubnetOptions)
+		if len(packing.InstanceTypes) > maxInstanceTypes {
```
ellistarn, line 99: imo this limiting detail should be encapsulated by the instanceProvider class

pkg/cloudprovider/aws/capacity.go
```diff
@@ -112,3 +149,20 @@ func (c *Capacity) GetOperatingSystems(ctx context.Context) ([]string, error) {
 		v1alpha1.OperatingSystemLinux,
 	}, nil
 }
+
+func (c *Capacity) ValidateLabels(labels map[string]string) error {
```
ellistarn, line 153: Can we just have the method be Validate() and we can validate arbitrary parts of the spec?

ellistarn: Can we break this validator into another file, too?

pkg/cloudprovider/aws/instance.go
```diff
@@ -35,33 +35,46 @@ type InstanceProvider struct {
...
-			})
+			}
+			if capacityType == capacityTypeSpot {
+				overridesRequest.Priority = aws.Float64(float64(i))
+			}
+			overrides = append(overrides, overridesRequest)
 		}
 	}
 	// 2. Create fleet
+
```
ellistarn, line 65: nit delete

pkg/cloudprovider/aws/instance.go
```diff
@@ -35,33 +35,46 @@ type InstanceProvider struct {
...
 				continue
 			}
-			overrides = append(overrides, &ec2.FleetLaunchTemplateOverridesRequest{
+			overridesRequest := &ec2.FleetLaunchTemplateOverridesRequest{
 				InstanceType: aws.String(*instanceType.InstanceType),
 				// FleetAPI cannot span subnets from the same AZ, so randomize.
 				SubnetId: aws.String(*subnets[rand.Intn(len(subnets))].SubnetId),
-			})
+			}
+			if capacityType == capacityTypeSpot {
```
ellistarn, line 58: Can you add a comment here?

pkg/cloudprovider/aws/capacity.go
ellistarn: Another alternative
```
func GetCapacityType(cloudprovider.Constraints) string {
 ...
}
```
IIUC, you only really need to call this once, and it simplifies lots of code changes. This struct doesn't fundamentally hold additional data, it's just formed differently. This makes me feel like we probably just need a new window (i.e. a function) onto the underlying struct.

pkg/utils/functional/functional.go
```diff
@@ -152,6 +152,16 @@ func ValidateAll(errorables ...func() error) error {
 	return err
 }
 
+// All returns true if all the predicate funcs return true and false otherwise.
+func All(predicates ...func() bool) bool {
```
ellistarn, line 156: AllTrue?

pkg/webhooks/provisioning/v1alpha1/validator.go
ellistarn: 💯

pkg/cloudprovider/aws/capacity.go
bwagner5: I had originally thought that as well, but I'm not sure it makes sense there since that would require us to trim the list before pods are packed which means the packer has less options to choose how to pack pods.  

I also considered the packer to do the limiting, but didn't think it made sense there either since it might be cloud provider specific if you even want to limit packing options.

pkg/cloudprovider/aws/capacity.go
bwagner5: Yeah!

pkg/cloudprovider/aws/capacity.go
bwagner5: good call 👍

pkg/cloudprovider/aws/capacity.go
bwagner5: yeah, I think that's similar to what Jacob proposed.

pkg/utils/functional/functional.go
bwagner5: If we go with this route:

```
   if isInstanceTypeSupported(constraints.InstanceTypes, instanceTypeInfo) && 
      isArchitectureSupported(architecture, instanceTypeInfo) &&
      isCapacityTypeSupported(constraints.CapacityType, instanceTypeInfo) &&
      isZonesSupported(zones, instanceTypeInfo) {
     ...
     }
```

I'll just delete this func.  Or we can change the signature to `func AllTrue(bools ...bool) bool`

pkg/cloudprovider/aws/capacity.go
ellistarn: >but I'm not sure it makes sense there since that would require us to trim the list before pods are packed which means the packer has less options to choose how to pack pods.
I'm not sure I agree. The next line is c.instanceProvider.Create(..., instancetype, ...), surely it could happen inside there without impacting the flow?

pkg/utils/functional/functional.go
ellistarn: prefer the former.

pkg/cloudprovider/aws/capacity.go
bwagner5: it only sorts by CPU (primary) and memory (secondary). I'll make the explicit in the comments rather than keeping it generic for now until we add another sort parameter.

pkg/cloudprovider/aws/capacity.go
bwagner5: oh sorry, I read that wrong, thought you were suggesting the instanceTypeProvider... 

I guess I'm neutral on the instanceProvider doing the trimming. I thought it might be surprising to have instanceTypes truncated when passed a list of options. But I suppose it only has one use-case and we don't want it to use all of them so... I guess it doesn't really matter.  As I'm typing I'm coming around to the idea... :)

pkg/cloudprovider/aws/capacity.go
bwagner5: Maybe I'm taking this too far, but are you suggesting getting rid of the validators in `webhooks/provisioning/validator.go`? And instead have the new cloudprovider `Validate()` do those validations as well?

pkg/cloudprovider/aws/capacity.go
ellistarn: I think it's worth some discussion. I think we probably prefer to have tight memory packing and extra CPU, since CPU is more burstable than memory. You could also imagine cases where we're selecting much too large instances if there are multiple ICE among one dimension, so a euclidean could help there.

pkg/cloudprovider/aws/capacity.go
bwagner5: It needs to be called in InstanceTypeProvider too.  I could do a hybrid where AWSConstraints is a struct that embeds a pointer to `cloudprovider.Constraints` and put a method on that struct of `GetCapacityType()`. That pattern will make it easier to add other AWS specific constraints like the launch template overrides coming soon.

awsconstraints.go
```
type AWSConstraints struct {
	*cloudprovider.Constraints
}

func NewAWSConstraints(constraints *cloudprovider.Constraints) *AWSConstraints {
	return &AWSConstraints{
		constraints,
	}
}

func (a *AWSConstraints) GetCapacityType() string {
	capacityType, ok := a.Labels[capacityTypeLabel]
	if !ok {
		capacityType = capacityTypeOnDemand
	}
	return capacityType
}
```

pkg/cloudprovider/aws/capacity.go
ellistarn: I think we probably want both.

pkg/cloudprovider/aws/capacity.go
bwagner5: I implemented a euclidean distance sort on vcpus and memory and it does a fairly good job of mirroring a price sort without the extra complexity of calling the pricing apis.  I'm not sure I understand your point on getting "ICE[d] among one dimension".  I don't think ICEs would be correlated like that.  

I'm also not sure we should be too picky on what we're sorting on since w/ OD, we don't do any sorting. The spot priorities are best effort and still primarily use capacity (i.e. pool depth) when deciding.

ellistarn: Thinking a touch further, if we had pricing data (oof: historical pricing), then we should just price sort. 

Maybe we could implement it generically as a price estimate sort, and then cloud providers can use any mechanism to generate those prices (i.e. calling price history, euclidean distance, etc)

bwagner5: That might make sense, it's not really price it's just an arbitrary prioritization that the cloud provider decides.  I don't know if it's worth adding a plug into the cloud provider though... the cloud provider could just resort based on whatever they want when they get the list back from packing.

pkg/cloudprovider/aws/validation.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/awslabs/karpenter/pkg/utils/functional"
+)
+
+// Validate cloud provider specific components of the cluster spec
+func (c *Capacity) Validate(ctx context.Context) error {
```
bwagner5, line 26: I'm not sure what to do with this, it seems weird to have it be a method on Capacity... but not sure where else to put it.  Maybe it deserves to be its own Type "Validator", but it seems that `Capacity` is the only view into a cloudprovider right now so, not sure if there's an intended design for other types or not.

pkg/cloudprovider/aws/awsconstraints.go
```diff
@@ -0,0 +1,50 @@
...
+	capacityTypeSpot     = "spot"
+	capacityTypeOnDemand = "on-demand"
+)
+
+var (
+	capacityTypeLabel = fmt.Sprintf("%s/capacity-type", nodeLabelPrefix)
+)
+
+// AWSConstraints are AWS specific constraints
+type AWSConstraints struct {
```
JacobGabrielson, line 34: Since you don't need to add any fields, you can just do:

```go
type AWSConstraints cloudprovider.Contraints
```

pkg/cloudprovider/aws/capacity.go
```diff
@@ -37,33 +37,36 @@ type Capacity struct {
 }
 
 // Create a set of nodes given the constraints.
-func (c *Capacity) Create(ctx context.Context, constraints *cloudprovider.Constraints) ([]cloudprovider.Packing, error) {
-	// 1. Retrieve normalized zones from constraints or all zones that the cluster spans if no zonal constraints are specified
-	zonalSubnetOptions, err := c.vpcProvider.GetZonalSubnets(ctx, constraints, c.spec.Cluster.Name)
+func (c *Capacity) Create(ctx context.Context, cloudProviderConstraints *cloudprovider.Constraints) ([]cloudprovider.Packing, error) {
+	// 1. Create AWS Cloud Provider constraints and apply defaults
+	awsConstraints := NewAWSConstraints(cloudProviderConstraints)
```
JacobGabrielson, line 42: if you make earlier suggestion, this would just be `... := AWSConstraints(cloudProviderConstraints)`

pkg/cloudprovider/aws/instance.go
```diff
@@ -29,39 +29,67 @@ import (
...
+// If spot is not used, the instanceTypeOptions are not required to be sorted
+// because we are using ec2 fleet's lowest-price OD allocation strategy
 func (p *InstanceProvider) Create(ctx context.Context,
 	launchTemplate *ec2.LaunchTemplate,
 	instanceTypeOptions []*packing.Instance,
 	zonalSubnetOptions map[string][]*ec2.Subnet,
+	capacityType string,
 ) (*string, error) {
-	// 1. Construct override options.
+	// 1. Trim the instanceTypeOptions so that the fleet request doesn't get too large
```
JacobGabrielson, line 52: Could you add why it can't get too large?

pkg/cloudprovider/aws/instance.go
JacobGabrielson: Can you add a comment to that effect?

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -146,22 +145,25 @@ func (p *InstanceTypeProvider) getAllInstanceTypes(ctx context.Context) ([]*ec2.
...
+				p.isZonesSupported(zones, instanceTypeInfo)) {
 			filtered = append(filtered, instanceTypeInfo)
 		}
 	}
 	return filtered
 }
 
+func (p *InstanceTypeProvider) isInstanceTypeSupported(instanceTypeConstraints []string, instance *packing.Instance) bool {
+	return (len(instanceTypeConstraints) != 0 || p.isDefaultInstanceType(instance)) &&
+		len(instanceTypeConstraints) == 0 || functional.ContainsString(instanceTypeConstraints, *instance.InstanceType)
```
JacobGabrielson, line 164: This might just be go fmt but the spacing looks off (misleading) here. Are you sure you don't need another set of parens around the 2nd line? (Normally I'm not a big fan of extra parens, but if I don't have to remember the precedence of `||` and `&&` it might be worth it?)

pkg/cloudprovider/aws/validation.go
JacobGabrielson: I think ultimately a `CloudProvider.ValidatorFor()` method might be warranted, but seems like we could wait on that?

pkg/cloudprovider/aws/capacity.go
ellistarn, line 42: I'd just call this variable constraints to avoid code changes below.

pkg/cloudprovider/aws/awsconstraints.go
ellistarn, line 34: I'd just alias this type like
`type Constraints cloudprovider.Constraints` instead of adding an aws concept. We're already in the aws package, which means this type is really aws.AWSConstraints.

pkg/cloudprovider/aws/instance.go
```diff
@@ -29,39 +29,67 @@ import (
...
 	var overrides []*ec2.FleetLaunchTemplateOverridesRequest
-	for _, instanceType := range instanceTypeOptions {
+	for i, instanceType := range instanceTypeOptions {
 		for _, zone := range instanceType.Zones {
 			subnets := zonalSubnetOptions[zone]
 			if len(subnets) == 0 {
 				continue
 			}
-			overrides = append(overrides, &ec2.FleetLaunchTemplateOverridesRequest{
+			overridesRequest := &ec2.FleetLaunchTemplateOverridesRequest{
```
ellistarn, line 66: nit: I'd probably either do (override, overrides), or (overridesRequest, overridesRequests), just to keep things consistent.

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -146,22 +145,25 @@ func (p *InstanceTypeProvider) getAllInstanceTypes(ctx context.Context) ([]*ec2.
...
 	filtered := []*packing.Instance{}
-	architecture := utils.NormalizeArchitecture(constraints.Architecture)
-
 	for _, instanceTypeInfo := range instanceTypes {
-		if (len(constraints.InstanceTypes) == 0 || functional.ContainsString(constraints.InstanceTypes, *instanceTypeInfo.InstanceType)) &&
-			(len(constraints.InstanceTypes) != 0 || p.isDefaultInstanceType(instanceTypeInfo)) &&
-			(architecture == nil || functional.ContainsString(aws.StringValueSlice(instanceTypeInfo.ProcessorInfo.SupportedArchitectures), *architecture)) &&
-			functional.ContainsString(aws.StringValueSlice(instanceTypeInfo.SupportedUsageClasses), "on-demand") &&
-			(len(zones) == 0 || len(functional.IntersectStringSlice(instanceTypeInfo.Zones, zones)) > 0) {
+		if constraints.Constraints == nil ||
```
ellistarn, line 151: Can we just create an invariant where constraints is never nil?

pkg/cloudprovider/aws/instancetypes.go
ellistarn: As a rule of thumb, if the if statement can't be expressed via a single operator 
```foo && bar && baz```
or 
```foo || bar || baz```
then you should separate the logic into a series of if statements
```
if shortcircuitPredicate {
  return false
}
if otherShortcircuit {
  return true
}
return false

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -170,3 +172,17 @@ func (p *InstanceTypeProvider) isDefaultInstanceType(instanceTypeInfo *packing.I
 		!*instanceTypeInfo.BareMetal &&
 		functional.HasAnyPrefix(*instanceTypeInfo.InstanceType, "m", "c", "r", "a", "t3", "t4")
 }
+
+func (p *InstanceTypeProvider) isArchitectureSupported(architecture *string, instance *packing.Instance) bool {
+	return architecture == nil ||
```
ellistarn, line 177: Aren't we guaranteed that architecture has been defaulted?

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -170,3 +172,17 @@ func (p *InstanceTypeProvider) isDefaultInstanceType(instanceTypeInfo *packing.I
...
+		functional.ContainsString(aws.StringValueSlice(instance.ProcessorInfo.SupportedArchitectures), *architecture)
+}
+
+func (p *InstanceTypeProvider) isCapacityTypeSupported(capacityType string, instance *packing.Instance) bool {
+	return capacityType == "" ||
```
ellistarn, line 182: aren't we guaranteed that capacityType has been defaulted?

pkg/cloudprovider/aws/validation.go
ellistarn: Hard agree.

pkg/cloudprovider/aws/instancetypes_test.go
```diff
@@ -60,7 +60,7 @@ var _ = Describe("InstanceTypes", func() {
 			ec2api := getInstanceTypeProviderMocks([]string{testZone}, []string{"m5.large"})
 			instanceTypeProvider := cloudprovideraws.NewInstanceTypeProvider(ec2api)
 			zonalSubnetOptions := map[string][]*ec2.Subnet{testZone: nil}
-			constraints := &cloudprovider.Constraints{}
+			constraints := &cloudprovideraws.AWSConstraints{Constraints: &cloudprovider.Constraints{}}
```
ellistarn, line 63: Here's that stutter I was talking about :)

pkg/cloudprovider/aws/validation.go
bwagner5: Okay, we'll leave as is for now and potentially refactor later.

pkg/cloudprovider/aws/awsconstraints.go
bwagner5: Much cleaner, thanks :D

pkg/cloudprovider/aws/instancetypes.go
bwagner5: No, only because of `GetAllInstanceTypeNames` in InstanceTypeProvider. It's used to get all the instance types for validating the provisioner spec, so it needs no filtering, but I would prefer to use the same overall logic to retrieve.

pkg/cloudprovider/aws/instancetypes.go
bwagner5: same answer as above.

pkg/cloudprovider/aws/instancetypes_test.go
bwagner5: ha so much

pkg/cloudprovider/aws/instancetypes.go
bwagner5: fixed

pkg/cloudprovider/aws/instance.go
bwagner5: fixed

pkg/cloudprovider/aws/awsconstraints.go
bwagner5: fixed

pkg/cloudprovider/aws/capacity.go
bwagner5: fixed

pkg/cloudprovider/aws/instancetypes.go
bwagner5: fixed

pkg/cloudprovider/aws/instance.go
bwagner5: added

pkg/cloudprovider/aws/capacity.go
bwagner5: fixed

pkg/cloudprovider/aws/constraints.go
```diff
@@ -0,0 +1,42 @@
...
+	nodeLabelPrefix      = "node.k8s.aws"
+	capacityTypeSpot     = "spot"
+	capacityTypeOnDemand = "on-demand"
+)
+
+var (
+	capacityTypeLabel = fmt.Sprintf("%s/capacity-type", nodeLabelPrefix)
+)
+
+// AWSConstraints are AWS specific constraints
```
JacobGabrielson, line 33: Comment needs an update to reflect new name

JacobGabrielson (commented):
Looks good to me (only a minor comment), other than resolving remaining @ellistarn feedback - thanks!!

pkg/cloudprovider/aws/constraints.go
bwagner5: 👍  fixed

pkg/cloudprovider/aws/capacity.go
```diff
@@ -37,33 +37,36 @@ type Capacity struct {
 }
 
 // Create a set of nodes given the constraints.
-func (c *Capacity) Create(ctx context.Context, constraints *cloudprovider.Constraints) ([]cloudprovider.Packing, error) {
-	// 1. Retrieve normalized zones from constraints or all zones that the cluster spans if no zonal constraints are specified
+func (c *Capacity) Create(ctx context.Context, cloudProviderConstraints *cloudprovider.Constraints) ([]cloudprovider.Packing, error) {
+	// 1. Create AWS Cloud Provider constraints and apply defaults
```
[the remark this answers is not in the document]
bwagner5: 👍  fixed

JacobGabrielson (approved):

ellistarn (approved):

PR #330, opened 2021-03-31
Proposal for node labels/architecture and launch templates in Provisioner

JacobGabrielson:
Issue #, if available: n/a

Description of changes:

This is a rough proposal for how to handle architecture overriding and launch templates in the Provisioner.


docs/aws/provisioner-launch-templates-options.md
```diff
@@ -0,0 +1,213 @@
...
+`node.<suffix>/something` rather than `<suffix>/something`?
+
+Lastly, a casual survey of other widely-used Kubernetes labels would
+suggest that using dashes is more common than camel case, so the
+portion following the `/` should most likely look like
+`launch-template-id`.
+
+### Launch Templates and Architecture
+
+At present, Launch Templates must specify an AMI-ID (ImageId)
```
prateekgogia, line 63: Also, we need user-data for bootstrapping which can differ based on the desired AMI.

docs/aws/provisioner-launch-templates-options.md
```diff
@@ -0,0 +1,213 @@
...
+we should use `node.k8s.aws/` as the prefix for any node labels we
+choose to use.
+
+Lastly, in keeping with another Kubernetes convention (such as
+`kubernetes.io/service-name`), we should use `lower-case-with-hypens`,
+not `camelCase` for words following the `/`.
+
+## Architecture
+
+One paritial solution to the architecture issue brought up above would
```
prateekgogia, line 127: nit: partial

docs/aws/provisioner-launch-templates-options.md
JacobGabrielson: In the case of cpu architecture though we wouldn't need differing UserData for different architectures.

docs/aws/provisioner-launch-templates-options.md
JacobGabrielson: fixed

docs/aws/provisioner-launch-templates-options.md
```diff
@@ -0,0 +1,213 @@
...
+
+This might be very non-standard however and defy expectations.
+
+## Recommendation
+
+For now the recommendation is to support the following in provisioner
+`spec.labels`:
+
+- `node.k8s.aws/launch-template-id`: id of launch template
+- `node.k8s.aws/launch-template-version`: version number or `$LATEST`
```
prateekgogia, line 188: In this case, we will not know the architecture of the AMI user has provided in template (unless, there is a way to go check the template > AMI > architecture). Now, if a pod has a different architecture specified in the selector we will not fail and request for a node with this template ID and that will fail without Karpenter knowing about it.

IMO, supporting this might not be much different than supporting 3rd proposal with these labels `node.k8s.aws/launch-template-id/arm64: id-of-arm64-based-ami`. In the latter, we get the name and architecture and use the correct template based on node selector(or default to amd64)?

ellistarn:
Discussed in working group. We can catch these edge cases using Validating Webhooks for the Provisioners, and simply ignoring (logging) at the pod level.

Reference:
```
kind: Pod
spec:
  nodeSelector:
    kubernetes.io/architecture: arm
    kubernetes.amazonaws.com/launchTemplateId: id-of-x86_64-based-ami

```

docs/aws/provisioner-launch-templates-options.md
JacobGabrielson: I think this might be now addressed (based on Ellis' suggestion to not allow both `architecture` and these labels at the same time?

prateekgogia: Are we going to make extra EC2 calls to get the architecture supported in Launch template ID provided in the label?

Also, is there a reason why we are not using this label key - `node.k8s.aws/launch-template-id/arm64:` ?

JacobGabrielson: No, the change would be to not try to validate the architecture at all, just simply not allow both `architecture` and launch template `labels` to be specified at the same time in a Provisioner spec.

docs/aws/provisioner-launch-templates-options.md
```diff
@@ -0,0 +1,238 @@
...
+`spec.labels`:
+
+- `node.k8s.aws/launch-template-id`: (optional) id of launch template
+  and cannot be specified if `architecture` 
+- `node.k8s.aws/launch-template-version`: version number or `$LATEST`
+  (optional, default `$LATEST`) (cannot be specified unless
+  `node.k8s.aws/launch-template-id` is present
+- `node.k8s.aws/capacity-type`: listed here for completeness
+
+While this will limit the provisioner to one architecture implicitly,
```
ellistarn, line 199: One default architecture. A pod can always select a different launch template to override the provisioner's default.

ellistarn:
LGTM, agree with recommendation.

docs/aws/provisioner-launch-templates-options.md
JacobGabrielson: fixing

docs/aws/provisioner-launch-templates-options.md
JacobGabrielson: Ok, hopefully addressed the above in latest revision @prateekgogia

prateekgogia (approved):
LGTM!

PR #333, opened 2021-04-02
add zone-ids to zone validation

bwagner5:
Issue #, if available:
https://github.com/awslabs/karpenter/issues/261

Description of changes:
Support for zone-ids was actually implemented as part of the instance type provider PR (https://github.com/awslabs/karpenter/pull/307), but this PR adds zone-ids to the cloud provider GetZones list so the provisioner validation webhook will succeed if a user puts a zone-id in the provisioner.

---
provisioner.yml
```
apiVersion: provisioning.karpenter.sh/v1alpha1
kind: Provisioner
metadata:
spec:
  cluster:
....
    name: karpenter-test
  ttlSeconds: 10
  zones:
  - use1-az2
```

My account maps use1-az2 -> us-east-1a.  Tested launching several instance which all went to the proper AZ. 

### DIFF
--- a/pkg/cloudprovider/aws/capacity.go
+++ b/pkg/cloudprovider/aws/capacity.go
@@ -17,6 +17,7 @@ package aws
 import (
 	"context"
 	"fmt"
+
 	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha1"
 	"github.com/awslabs/karpenter/pkg/cloudprovider"
 	"github.com/awslabs/karpenter/pkg/packing"
@@ -97,7 +98,15 @@ func (c *Capacity) GetInstanceTypes(ctx context.Context) ([]string, error) {
 }
 
 func (c *Capacity) GetZones(ctx context.Context) ([]string, error) {
-	return c.vpcProvider.GetZones(ctx, c.spec.Cluster.Name)
+	azs, err := c.vpcProvider.GetAllAvailabilityZones(ctx)
+	if err != nil {
+		return nil, err
+	}
+	zones := []string{}
+	for _, az := range azs {
+		zones = append(zones, []string{*az.ZoneName, *az.ZoneId}...)
+	}
+	return zones, nil
 }
 
 func (c *Capacity) GetArchitectures(ctx context.Context) ([]string, error) {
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/aws/capacity.go
```diff
@@ -97,7 +98,15 @@ func (c *Capacity) GetInstanceTypes(ctx context.Context) ([]string, error) {
...
 
 func (c *Capacity) GetZones(ctx context.Context) ([]string, error) {
-	return c.vpcProvider.GetZones(ctx, c.spec.Cluster.Name)
+	azs, err := c.vpcProvider.GetAllAvailabilityZones(ctx)
+	if err != nil {
+		return nil, err
+	}
+	zones := []string{}
+	for _, az := range azs {
+		zones = append(zones, []string{*az.ZoneName, *az.ZoneId}...)
```
JacobGabrielson, line 107: I think this can just be:

```go
zones = append(zones, *az.ZoneName, *az.ZoneId)
```

?

pkg/cloudprovider/aws/capacity.go
```diff
@@ -97,7 +98,15 @@ func (c *Capacity) GetInstanceTypes(ctx context.Context) ([]string, error) {
 }
 
 func (c *Capacity) GetZones(ctx context.Context) ([]string, error) {
-	return c.vpcProvider.GetZones(ctx, c.spec.Cluster.Name)
+	azs, err := c.vpcProvider.GetAllAvailabilityZones(ctx)
```
JacobGabrielson, line 101: Maybe call this `GetAllZones` (or even just `GetZones`) for consistency (since this method is called `GetZones`?)

pkg/cloudprovider/aws/capacity.go
bwagner5: oops, yeah... 👍  fixed

pkg/cloudprovider/aws/capacity.go
bwagner5: good point 👍  fixed

JacobGabrielson (approved):

PR #335, opened 2021-04-02
[WIP] Terminate nodes only if instance is terminated

njtran:
https://github.com/awslabs/karpenter/issues/275

Description of changes:
Added a check in the delete nodes call to make sure that it was an instance that was deleted in the cloud provider.

### DIFF
--- a/pkg/cloudprovider/fake/capacity.go
+++ b/pkg/cloudprovider/fake/capacity.go
@@ -54,8 +54,8 @@ func (c *Capacity) Create(ctx context.Context, constraints *cloudprovider.Constr
 	}}, nil
 }
 
-func (c *Capacity) Delete(ctx context.Context, nodes []*v1.Node) error {
-	return nil
+func (c *Capacity) Delete(ctx context.Context, nodes []*v1.Node) (map[string]bool, error) {
+	return nil, nil
 }
 
 func (c *Capacity) GetZones(ctx context.Context) ([]string, error) {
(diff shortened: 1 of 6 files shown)


ellistarn (commented):
How is it possible that delete can fail? I don't think we want to expose this possibility through the cloud provider API.

bwagner5:
> How is it possible that delete can fail? I don't think we want to expose this possibility through the cloud provider API.

The EC2 Terminate call could fail if the customer turns on Termination Protection on the instance.  

Another option rather than bubbling up the nodes that actually terminated is to just fail the request if any error returned and the reconcile the running instances.

ellistarn:
I think we definitely want to avoid leaking EC2's details through out cloud provider API. Brandon's suggestion seems reasonable to me.

pkg/cloudprovider/fake/capacity.go
```diff
@@ -54,8 +54,8 @@ func (c *Capacity) Create(ctx context.Context, constraints *cloudprovider.Constr
 	}}, nil
 }
 
-func (c *Capacity) Delete(ctx context.Context, nodes []*v1.Node) error {
-	return nil
+func (c *Capacity) Delete(ctx context.Context, nodes []*v1.Node) (map[string]bool, error) {
```
JacobGabrielson, line 57: I think we should avoid returning this `map` - I think it's too complicated to expect the calling code to really know what to do about the problem. Either it's a persistent problem and a human operator needs to get involved (some node is "stuck" being terminated), or else it's a transient problem and will go away on its own in the next few rounds of reconciliation. So really all the code calling `Delete` can do is log - but the cloud provider code could just as easily log as well (and then you don't need `GetInstanceId`.

njtran:
Got it. Thanks for the comments. Since the original code already did what we plan on doing, closing this PR.

PR #336, opened 2021-04-05
removed TODO per discussion

njtran:
https://github.com/awslabs/karpenter/issues/275

Discussed [here](https://github.com/awslabs/karpenter/pull/335). The TODO was placed to ensure that if the Terminate API command succeeded but only terminated partial instances, that the node wouldn't be deleted in the APIServer. After realizing that the command would not error only if all instances are deleted, this TODO becomes unneeded.

### DIFF
--- a/pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
+++ b/pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
@@ -130,8 +130,6 @@ func (t *Terminator) deleteNodes(ctx context.Context, nodes []*v1.Node, provisio
 		return fmt.Errorf("terminating cloudprovider instance, %w", err)
 	}
 	// 2. Delete node in APIServer
-	// TODO: Prevent leaked nodes: ensure a node is not deleted in apiserver if not deleted in cloudprovider
-	// Use the returned ids from the cloudprovider's Delete() function, and then only delete those ids in the apiserver
 	for _, node := range nodes {
 		if err := t.kubeClient.Delete(ctx, node); err != nil {
 			zap.S().Debugf("Continuing after failing to delete node %s, %s", node.Name, err.Error())


pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -130,8 +130,6 @@ func (t *Terminator) deleteNodes(ctx context.Context, nodes []*v1.Node, provisio
 		return fmt.Errorf("terminating cloudprovider instance, %w", err)
 	}
 	// 2. Delete node in APIServer
-	// TODO: Prevent leaked nodes: ensure a node is not deleted in apiserver if not deleted in cloudprovider
```
JacobGabrielson, line 133: We can probably remove this todo anyway, but before I forget, does the `Delete` implementation (for AWS) correctly handle the case where it tries to delete an already-deleted node? (It should recognize this case and not consider it an error). Otherwise you can end up in a situation where the nodes never get deleted on the K8s side (i.e., the `kubeClient.Delete` will never get called because of 1 (or more) seemingly un-deletable node)?

ellistarn: There's a safety valve in the Kubernetes cloud controller manager that checks EC2 if instances exist and deletes the node if it doesn't. a.k.a., kubernetes will reconcile.

JacobGabrielson: Do we need to do this at all then? I still think there would be a case where this code never gets past the error (unless we also check whether or not k8s thinks the node exists?)

njtran: Deleting the node in APIServer rather than waiting for garbage collection allows the node to get cleaned up immediately. Performance wise, the benefit is that it won't get called to delete in further reconcile loops (which if a reconcile loop is 5s and the garbage collection is 300s, that's 60 unnecessary calls). 
Functionality wise, I'm unsure if the deletion API for AWS correctly handles Jacob's aforementioned case. If it doesn't properly handle it, then we will be continually returning an error here until garbage collection cleans it up.

JacobGabrielson: Ok, I cut a ticket (https://github.com/awslabs/karpenter/issues/340) to investigate, and will approve this change.

JacobGabrielson (approved):

PR #338, opened 2021-04-05
don't require labels to match when terminating nodes

bwagner5:
Issue #, if available:
N/A

Description of changes:
Currently, only nodes that have the same labels as the provisioner spec Labels list is marked as underutilized and eventually terminated. We don't want to churn whole clusters because of a simple label change in the provisioner and if we don't churn, then nodes will be orphaned from the provisioners management.  So this PR removes the Labels requirement when looking for nodes to mark as underutilized.

### DIFF
--- a/pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
+++ b/pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
@@ -147,7 +147,7 @@ func (t *Terminator) getNodes(ctx context.Context, provisioner *v1alpha1.Provisi
 	if err := t.kubeClient.List(ctx, nodes, client.MatchingLabels(functional.UnionStringMaps(map[string]string{
 		v1alpha1.ProvisionerNameLabelKey:      provisioner.Name,
 		v1alpha1.ProvisionerNamespaceLabelKey: provisioner.Namespace,
-	}, provisioner.Spec.Labels, additionalLabels))); err != nil {
+	}, additionalLabels))); err != nil {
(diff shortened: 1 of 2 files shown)


ellistarn (approved):
Do we have a test case for this?

pkg/controllers/provisioning/v1alpha1/reallocation/terminate.go
```diff
@@ -147,7 +147,7 @@ func (t *Terminator) getNodes(ctx context.Context, provisioner *v1alpha1.Provisi
 	if err := t.kubeClient.List(ctx, nodes, client.MatchingLabels(functional.UnionStringMaps(map[string]string{
 		v1alpha1.ProvisionerNameLabelKey:      provisioner.Name,
 		v1alpha1.ProvisionerNamespaceLabelKey: provisioner.Namespace,
-	}, provisioner.Spec.Labels, additionalLabels))); err != nil {
+	}, additionalLabels))); err != nil {
```
JacobGabrielson, line 150: It does seem like we have the exact same code here as in *utilization.go* - could we eliminate the duplication?

njtran: It is the same code. I have a task to de-duplicate these functions. 

Brandon, for now, can you also change the code in `utilization.go`? I'll de-dup in a follow-up.

njtran (approved):
LGTM nice catch!

bwagner5:
> Do we have a test case for this?

Doesn't look like there is a test for this specific case, but will add an issue to add one.

PR #345, opened 2021-04-06
clarify node selector behavior

JacobGabrielson:
Issue #, if available:

Description of changes:

Update to doc to reflect potential clarification on behavior.

### DIFF
--- a/docs/aws/provisioner-launch-templates-options.md
+++ b/docs/aws/provisioner-launch-templates-options.md
@@ -192,7 +192,7 @@ then ignore pod specs that specify and incompatible
 `kubernetes.io/arch` in a [node
 selector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector).
 For example, imagine the pod spec of a pending pod contains:
-
+p
 ```yaml
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


docs/aws/provisioner-launch-templates-options.md
```diff
@@ -296,3 +328,10 @@ leverage and/or work smoothly with
 [CAPI](https://github.com/kubernetes-sigs/cluster-api). We will
 address this in a separate document.
 
+## Launch Template Names
+
+We could support something like:
+
+- `node.k8s.aws/launch-template-name`: (optional) name of launch template
```
bwagner5, line 335: Optional: Maybe call out the problems (or work needed to support this)?  For example if a provisioner used launch-template-id but the pod had an override with the launch-template-name, then two nodes would be provisioned. Maybe this is just a scenario that users shouldn't do and we shouldn't try to help too much.

bwagner5 (approved):
LGTM, left one comment but optional, and minor

docs/aws/provisioner-launch-templates-options.md
JacobGabrielson: I can address in a follow up

PR #347, opened 2021-04-07
update karpenter metrics clusterrole

razvan-moj:
Without this, I get 
`prometheus level=error ts=2021-04-07T13:21:13.199Z caller=klog.go:96 component=k8s_client_runtime func=ErrorDepth msg="pkg/mod/k8s.io/client-go@v0.20.5/tools/cache/reflector.go:167: Failed to watch *v1.Pod: failed to list *v1.Pod: pods is forbidden: User \"system:serviceaccount:karpenter:default\" cannot list resource \"pods\" in API group \"\" in the namespace \"karpenter\""`
in the `prometheus-karpenter-monitor-0` logs

Issue #, if available:

Description of changes:

### DIFF
--- a/charts/karpenter/templates/manifest.yaml
+++ b/charts/karpenter/templates/manifest.yaml
@@ -148,6 +148,7 @@ rules:
     resources:
       - services
       - endpoints
+      - pods
     verbs:
       - get
       - list


ellistarn:
Hey @razvan-moj, thanks for your contribution. This code is actually generated from https://github.com/awslabs/karpenter/tree/main/config as part of `make helm`.

If I understand correctly, your prometheus shouldn't need permissions to pods to scrape metrics (endpoints should do, instead). However, I could imagine this might be necessary on a different version of prometheus or with something like headless services. Have you been able to verify that once this change is made, your prometheus installation is able to read metrics from Karpenter?

razvan-moj:
> Have you been able to verify that once this change is made, your prometheus installation is able to read metrics from Karpenter?

Hi - my prometheus was able to scrape karpenter already, the error seems to be from a second prometheus that the karpenter chart installs:
```
$ kubectl get statefulset -A
NAMESPACE    NAME                                                   READY   AGE
karpenter    prometheus-karpenter-monitor                           1/1     6h57m
prometheus   alertmanager-prometheus-kube-prometheus-alertmanager   1/1     7h8m
prometheus   prometheus-prometheus-kube-prometheus-prometheus       1/1     7h8m
```

ellistarn:
Gotcha. We used to have a hard dependency on prometheus in a previous design iteration. I think it probably makes sense to simply remove these configurations, and rely on users to integrate with prometheus as necessary, rather than creating a special one for karp.

ellistarn:
WDYT? https://github.com/awslabs/karpenter/pull/348/files

razvan-moj:
not needed anymore

ellistarn:
Thanks for raising this to our attention!

PR #348, opened 2021-04-07
Removed prometheus

ellistarn:
Issue #, if available:

Description of changes:
No longer a hard dependency. We should instead rely on customers to integrate with prometheus themselves instead of creating a bad default (resource waste, bloat, etc).

See https://github.com/awslabs/karpenter/pull/347 


JacobGabrielson (approved):

razvan-moj (commented):
👍

PR #354, opened 2021-04-09
Fix bug in ec2 api mock

JacobGabrielson:
Issue #, if available: 273

Description of changes:

- wasn't resetting the list of fake ec2 instances when Reset was called

### DIFF
--- a/pkg/cloudprovider/aws/fake/ec2api.go
+++ b/pkg/cloudprovider/aws/fake/ec2api.go
@@ -62,7 +62,7 @@ func (e *EC2API) CreateFleetWithContext(ctx context.Context, input *ec2.CreateFl
 	}
 	instance := &ec2.Instance{
 		InstanceId:     aws.String(randomdata.SillyName()),
-		Placement:      &ec2.Placement{AvailabilityZone: aws.String("test-zone")},
+		Placement:      &ec2.Placement{AvailabilityZone: aws.String("test-zone-1a")},
 		PrivateDnsName: aws.String(fmt.Sprintf("test-instance-%d.example.com", len(e.Instances))),
 	}
 	e.Instances = append(e.Instances, instance)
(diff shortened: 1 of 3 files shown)


ellistarn (dismissed):

JacobGabrielson:
Turned out the problem was a little bigger than I thought, please re-review @ellistarn. Thanks!

pkg/cloudprovider/aws/fake/ec2api.go
```diff
@@ -36,82 +37,88 @@ type EC2API struct {
...
+	ec2iface.EC2API
+	EC2Behavior
 }
 
-func (a *EC2API) Reset() {
-	a.CalledWithCreateFleetInput = nil
+// Reset must be called between tests otherwise tests will pollute
+// each other.
+func (e *EC2API) Reset() {
+	e.EC2Behavior = EC2Behavior{}
```
ellistarn, line 52: Idk if this is crazy, but is it possible to keep this as 
```
func (a *EC2API) {
  a = &EC2API{}
}
```

JacobGabrielson: That is on the face of it some rather scary Go code, so I would be hesitant to go that route.

But also there's no guarantee that other code is not relying on the same pointer to the ec2api. In fact there is code like:

```go
	fakeEC2API = &fake.EC2API{}
	subnetProvider := &SubnetProvider{
		ec2api: fakeEC2API,
		cache:  subnetCache,
	}
```

ellistarn: I guess inline behavior is pretty clean :)

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -222,6 +222,9 @@ var _ = Describe("Allocation", func() {
 		})
 		It("should launch separate instances for pods with different node selectors", func() {
 			// Setup
+			fakeEC2API.DescribeSubnetsOutput = &ec2.DescribeSubnetsOutput{Subnets: []*ec2.Subnet{
```
ellistarn, line 225: Shouldn't there simply be a default subnet you can use in this test? I feel it shouldn't be necessary to specify this for this test.

JacobGabrielson: Fixed - good catch - turned out to be another bug in fakeEC2API

bwagner5 (approved):
I think the refactor looks good. I'm curious why we even need to reuse the FakeEC2API across tests and therefore need the Reset ability. Why can't we just use a new FakeEC2API on every test?

JacobGabrielson:
> I think the refactor looks good. I'm curious why we even need to reuse the FakeEC2API across tests and therefore need the Reset ability. Why can't we just use a new FakeEC2API on every test?

@bwagner5 I haven't looked at it that closely - I *think* it's to avoid re-creating the kubernetes environment for every individual test (which I think would be very slow?) - and the webhooks need to refer (indirectly) to the ec2 api.

ellistarn (approved):

PR #355, opened 2021-04-09
[WIP] Support launch template id and version tags on node selectors

JacobGabrielson:
Issue #, if available: 273

Description of changes:

This should result in pods that have node selectors specifying a launch template id/version being launched with the said launch template. 

Getting some early f'back on this ... unit tests pass but haven't finished the unit tests or tried IRL.


pkg/cloudprovider/aws/constraints.go
```diff
@@ -27,19 +27,39 @@ const (
...
 func (c *Constraints) GetCapacityType() string {
-	capacityType, ok := c.Labels[capacityTypeLabel]
+	capacityType, ok := c.Labels[CapacityTypeLabel]
 	if !ok {
 		capacityType = capacityTypeOnDemand
 	}
 	return capacityType
 }
+
+type LaunchTemplateDescriptor struct {
```
ellistarn, line 47: I like this, but consider renaming to just LaunchTemplate for more idiomatic code (it's a subset of the true object).

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -61,18 +61,28 @@ func launchTemplateName(clusterName string, arch string) string {
...
-func (p *LaunchTemplateProvider) Get(ctx context.Context, cluster *v1alpha1.ClusterSpec, constraints Constraints) (*ec2.LaunchTemplate, error) {
+func (p *LaunchTemplateProvider) Get(ctx context.Context, cluster *v1alpha1.ClusterSpec, constraints Constraints) (*LaunchTemplateDescriptor, error) {
+	// If the customer specified a launch template then just use it
+	if desc := constraints.GetLaunchTemplate(); desc != nil {
+		return desc, nil
+	}
+
 	arch := utils.NormalizeArchitecture(constraints.Architecture)
 	name := launchTemplateName(cluster.Name, *arch)
+	version := "$Default"
```
ellistarn, line 72: consider inlining with ptr.String("$Default")

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -61,18 +61,28 @@ func launchTemplateName(clusterName string, arch string) string {
...
+	result := &LaunchTemplateDescriptor{
+		Version: &version,
+	}
 	if launchTemplate, ok := p.cache.Get(name); ok {
```
ellistarn, line 76: Should we be caching Id+Version as the key?

ellistarn (commented):
I think we might have an additional gap around labels on the node. We should be populating all defaulted properties (i.e. instance type, launch template, capacity type, etc) into the node's labels. I think it just kinda magically happens right now.

pkg/cloudprovider/aws/launchtemplate.go
bwagner5: IIUC version is always `$Default` when querying using the EC2 API.  If a version were specified by the user, it would short circuit at the top of the func.  

Do you think there'd be a future feature where customers could specify to use `$Latest` instead of default by just providing the the launch template version and no launch template name?

bwagner5:
> I think we might have an additional gap around labels on the node. We should be populating all defaulted properties (i.e. instance type, launch template, capacity type, etc) into the node's labels. I think it just kinda magically happens right now when the node comes online, i.e., we're relying on the kubelet..

Ah you're right / maybe worse than what you described. It looks like `capacity-type` and `launch-template-id / version` won't get added to the node labels unless they are explicitly provided on the provisioner spec. 

Maybe there should be a `GetLabels()` func on the AWS constraints which would merge the cloudprovider constraints and add the proper default aws constraints. `GetLabels` could be used when passing to the nodefactory in `Create` rather than the `constraints.Labels`.  It might be a bit awkward to have aws constraints have a `.Labels` and a `.GetLabels()` that will return different things... but that can probably be solved by naming the func that merges something more explicit, like `GetMergedLabels`, idk, I'm not the best at naming :) 

What do you all think?

ellistarn:
> > I think we might have an additional gap around labels on the node. We should be populating all defaulted properties (i.e. instance type, launch template, capacity type, etc) into the node's labels. I think it just kinda magically happens right now when the node comes online, i.e., we're relying on the kubelet..
> 
> Ah you're right / maybe worse than what you described. It looks like `capacity-type` and `launch-template-id / version` won't get added to the node labels unless they are explicitly provided on the provisioner spec.
> 
> Maybe there should be a `GetLabels()` func on the AWS constraints which would merge the cloudprovider constraints and add the proper default aws constraints. `GetLabels` could be used when passing to the nodefactory in `Create` rather than the `constraints.Labels`. It might be a bit awkward to have aws constraints have a `.Labels` and a `.GetLabels()` that will return different things... but that can probably be solved by naming the func that merges something more explicit, like `GetMergedLabels`, idk, I'm not the best at naming :)
> 
> What do you all think?

I was wrong about all of this. If a pod has specific node selectors on it, those labels will be generated into the pod (thanks past Ellis). This means that for any of these special labels, they will only occur if necessary. Further, the kubelet will apply any labels that are specific to the node. I like this because it means that karpenter is doing the minimal possible behavior for everything to just work, and other labels are governed by the kubelet. 

Ref: https://github.com/awslabs/karpenter/blob/main/pkg/apis/provisioning/v1alpha1/provisioner.go#L155

Sorry for the thrash!

pkg/cloudprovider/aws/constraints.go
JacobGabrielson: Ok!

pkg/cloudprovider/aws/launchtemplate.go
JacobGabrielson: Ok!

JacobGabrielson:
> > > I think we might have an additional gap around labels on the node. We should be populating all defaulted properties (i.e. instance type, launch template, capacity type, etc) into the node's labels. I think it just kinda magically happens right now when the node comes online, i.e., we're relying on the kubelet..
> > 
> > 
> > Ah you're right / maybe worse than what you described. It looks like `capacity-type` and `launch-template-id / version` won't get added to the node labels unless they are explicitly provided on the provisioner spec.
> > Maybe there should be a `GetLabels()` func on the AWS constraints which would merge the cloudprovider constraints and add the proper default aws constraints. `GetLabels` could be used when passing to the nodefactory in `Create` rather than the `constraints.Labels`. It might be a bit awkward to have aws constraints have a `.Labels` and a `.GetLabels()` that will return different things... but that can probably be solved by naming the func that merges something more explicit, like `GetMergedLabels`, idk, I'm not the best at naming :)
> > What do you all think?
> 
> I was wrong about all of this. If a pod has specific node selectors on it, those labels will be generated into the pod (thanks past Ellis). This means that for any of these special labels, they will only occur if necessary. Further, the kubelet will apply any labels that are specific to the node. I like this because it means that karpenter is doing the minimal possible behavior for everything to just work, and other labels are governed by the kubelet.
> 
> Ref: https://github.com/awslabs/karpenter/blob/main/pkg/apis/provisioning/v1alpha1/provisioner.go#L155
> 
> Sorry for the thrash!

Yeah, I think I tested that out and noticed the magic was already happening.

JacobGabrielson:
> > I think we might have an additional gap around labels on the node. We should be populating all defaulted properties (i.e. instance type, launch template, capacity type, etc) into the node's labels. I think it just kinda magically happens right now when the node comes online, i.e., we're relying on the kubelet..
> 
> Ah you're right / maybe worse than what you described. It looks like `capacity-type` and `launch-template-id / version` won't get added to the node labels unless they are explicitly provided on the provisioner spec.
> 
> Maybe there should be a `GetLabels()` func on the AWS constraints which would merge the cloudprovider constraints and add the proper default aws constraints. `GetLabels` could be used when passing to the nodefactory in `Create` rather than the `constraints.Labels`. It might be a bit awkward to have aws constraints have a `.Labels` and a `.GetLabels()` that will return different things... but that can probably be solved by naming the func that merges something more explicit, like `GetMergedLabels`, idk, I'm not the best at naming :)
> 
> What do you all think?

So I think this isn't an issue per @ellistarn latest comments, but lmk.

pkg/cloudprovider/aws/launchtemplate.go
JacobGabrielson: I was thinking we would not support that - seems like too much magic and assumptions about what we're doing with the launch templates ourselves. WDYT?

JacobGabrielson: @ellistarn I'd prefer to leave it for now as it is always $Default, could add that in later if we ever change.

bwagner5: > I was thinking we would not support that - seems like too much magic and assumptions about what we're doing with the launch templates ourselves. WDYT?

👍  Supporting just default seems fine to me, was just thinking on why we'd need need to cache with the version key.

bwagner5:
> So I think this isn't an issue per @ellistarn latest comments, but lmk.

It still doesn't work, but @ellistarn and I discussed and we're going to leave it the way it is currently (requiring the user to specify the capacity-type or else you'll get an artificial scheduling constraint when capacityType is nil). Launch template doesn't seem to matter very much whether it's nil or not when using the default since the default doesn't include anything that useful.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -61,18 +61,32 @@ func launchTemplateName(clusterName string, arch string) string {
 	return fmt.Sprintf(launchTemplateNameFormat, clusterName, arch)
 }
 
-func (p *LaunchTemplateProvider) Get(ctx context.Context, cluster *v1alpha1.ClusterSpec, constraints Constraints) (*ec2.LaunchTemplate, error) {
+func (p *LaunchTemplateProvider) Get(ctx context.Context, cluster *v1alpha1.ClusterSpec, constraints Constraints) (*LaunchTemplate, error) {
+	// If the customer specified a launch template then just use it
+	if desc := constraints.GetLaunchTemplate(); desc != nil {
```
bwagner5, line 66: what is desc? Description?  Can we just call this launchTemplate or result ?

JacobGabrielson: will fix

bwagner5 (approved):
Lgtm

PR #368, opened 2021-04-15
Refactored packing logic to create an InstanceType abstraction to hide cloud provider details.

ellistarn:
Issue #, if available:

Description of changes:


pkg/packing/packing.go
```diff
@@ -70,77 +62,54 @@ func (p *packer) Pack(ctx context.Context, pods []*v1.Pod, instanceTypes []*Inst
...
-			zap.S().Infof("Excluding instance type %s because there are not enough resources for the kubelet overhead", nc.instanceType)
-			continue
-		}
-		nodeCapacities = append(nodeCapacities, nc)
-	}
-	return nodeCapacities
-}
-
 // packWithLargestPod will try to pack max number of pods with largest pod in
 // pods across all available node capacities. It returns Packing: max pod count
```
bwagner5, line 80: nit: "packWithLargestPod will try to pack the max number of pods starting with the largest pod across all available node capacities"

pkg/packing/instance.go
```diff
@@ -0,0 +1,89 @@
...
+
+package packing
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/awslabs/karpenter/pkg/utils/binpacking"
+	"github.com/awslabs/karpenter/pkg/utils/resources"
+	resourcesUtil "github.com/awslabs/karpenter/pkg/utils/resources"
```
bwagner5, line 23: delete dup resources

pkg/packing/instance.go
```diff
@@ -0,0 +1,89 @@
...
+	Zones []string
+}
+
+type PackableInstance struct {
+	*Instance
+	reserved     v1.ResourceList
+	total        v1.ResourceList
+}
+
+func PackableInstancesFrom(instanceTypes []*Instance, overhead v1.ResourceList) []*PackableInstance {
```
bwagner5, line 41: do these need to be exported? I don't think the cloud provider needs these funcs

pkg/packing/packing.go
```diff
@@ -70,77 +62,54 @@ func (p *packer) Pack(ctx context.Context, pods []*v1.Pod, instanceTypes []*Inst
...
 // that fit; with their node capacities and list of leftover pods
-func (p *packer) packWithLargestPod(unpackedPods []*v1.Pod, nodeCapacities []*nodeCapacity) (*Packing, []*v1.Pod) {
+func (p *packer) packWithLargestPod(unpackedPods []*v1.Pod, instanceTypes []*Instance, constraints *cloudprovider.Constraints) (*Packing, []*v1.Pod) {
 	bestPackedPods := []*v1.Pod{}
-	bestCapacities := []*nodeCapacity{}
+	bestInstances := []*Instance{}
 	remainingPods := unpackedPods
-	for _, nc := range nodeCapacities {
+	for _, packableInstance := range PackableInstancesFrom(instanceTypes, constraints.Overhead) {
```
bwagner5, line 86: nice!

pkg/packing/instance.go
```diff
@@ -0,0 +1,89 @@
...
+			// https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt#L20
+			v1.ResourcePods: resource.MustParse(fmt.Sprint(
+				*instance.NetworkInfo.MaximumNetworkInterfaces*(*instance.NetworkInfo.Ipv4AddressesPerInterface-1) + 2)),
+		},
+	}
+}
+
+func (p *PackableInstance) Copy() *PackableInstance {
+	return &PackableInstance{p.Instance, p.reserved.DeepCopy(), p.total.DeepCopy()}
+}
```
bwagner5, line 71: can we get rid of Copy?

bwagner5:
Since this did regress once, probably worth adding a test for scheduling beyond a single node's capacity

ellistarn:
> Since this did regress once, probably worth adding a test for scheduling beyond a single node's capacity

This is actually tough since we do eventually reconcile on a separate loop.

bwagner5:
> > Since this did regress once, probably worth adding a test for scheduling beyond a single node's capacity
> 
> This is actually tough since we do eventually reconcile on a separate loop. Let me think on this.

packing unit test could test this

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -47,25 +48,31 @@ func NewInstanceTypeProvider(ec2api ec2iface.EC2API) *InstanceTypeProvider {
...
 			return nil, err
 		}
-		p.cache.SetDefault(allInstanceTypesKey, supportedInstanceTypes)
-		zap.S().Debugf("Successfully discovered %d EC2 instance types", len(supportedInstanceTypes))
+		p.cache.SetDefault(allInstanceTypesKey, instanceTypes)
+		zap.S().Debugf("Successfully discovered %d EC2 instance types", len(instanceTypes))
 	}
-	return p.filterFrom(supportedInstanceTypes, constraints, zones), nil
+
+	// Filter by constraints and zones
```
JacobGabrielson, line 70: Could you change this comment to explain why a copy is being made in this fashion? (I'm not quite sure why not just return the result of `filterFrom` directly? (If it's just to convert from `*InstanceType` to `InstanceType`, why not just have `Get` return a pointer as it used to?) (Or change `filterFrom` not to return pointers?)

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -145,67 +152,123 @@ func (p *InstanceTypeProvider) getAllInstanceTypes(ctx context.Context) ([]*ec2.
...
+	return !instanceType.NvidiaGPUs().IsZero()
+}
+func (p *InstanceTypeProvider) isAWSNeuronSupported(requests v1.ResourceList, instanceType *InstanceType) bool {
+	if _, ok := requests[resources.AWSNeuron]; !ok {
+		return true
+	}
+	return !instanceType.AWSNeurons().IsZero()
+}
+
+func (p *InstanceTypeProvider) isZonesSupported(zones []string, instanceType *InstanceType) bool {
```
JacobGabrielson, line 220: nit is -> are? Also name is a bit confusing regardless; IIUC this is more like "any/some zones supported", "is" or "are" are kinda ambiguous as to whether you mean all of them or just at least one.

pkg/packing/packer.go
```diff
@@ -0,0 +1,151 @@
...
+func weightOf(instance cloudprovider.InstanceType) float64 {
+	return euclidean(
+		float64(instance.CPU().Value()),
+		float64(instance.Memory().ScaledValue(resource.Mega)), // 1 gb = 1 cpu
+		float64(instance.NvidiaGPUs().Value())*1000,           // Heavily weigh gpus x 1000
+		float64(instance.AWSNeurons().Value())*1000,           // Heavily weigh neurons x1000
+	)
+}
+
+func euclidean(values ...float64) float64 {
```
JacobGabrielson, line 137: add comment explaining why "euclidean" being used?

pkg/cloudprovider/aws/instancetypes.go
ellistarn: We're transforming from a map to a slice. In general, I'd like to make this simply not a map (I don't think this needs to be), but I'm addressing in a followon PR.

pkg/cloudprovider/aws/instancetypes.go
ellistarn: The naming of all of these is a bit complex, especially GPU (which is really semantically "has if needs"). Would love any high level naming recommendation you have for all of these.

pkg/packing/packer.go
```diff
@@ -0,0 +1,151 @@
...
+
+// sortByResources sorts instance type packings by vcpus and memory resources
+func sortByResources(instances []cloudprovider.InstanceType) {
+	sort.Slice(instances, func(i, j int) bool { return weightOf(instances[i]) < weightOf(instances[j]) })
+}
+
+func weightOf(instance cloudprovider.InstanceType) float64 {
+	return euclidean(
+		float64(instance.CPU().Value()),
+		float64(instance.Memory().ScaledValue(resource.Mega)), // 1 gb = 1 cpu
```
prateekgogia, line 131: With this we will always prefer instance with higher memory when CPU is the same, for example -`t2.large 2cpu-8gb` over `t2.medium 2cpu-4gb`?

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -47,25 +48,31 @@ func NewInstanceTypeProvider(ec2api ec2iface.EC2API) *InstanceTypeProvider {
 }
 
 // Get instance types that are availble per availability zone
-func (p *InstanceTypeProvider) Get(ctx context.Context, zonalSubnetOptions map[string][]*ec2.Subnet, constraints Constraints) ([]*packing.Instance, error) {
+func (p *InstanceTypeProvider) Get(ctx context.Context, zonalSubnetOptions map[string][]*ec2.Subnet, constraints Constraints) ([]cloudprovider.InstanceType, error) {
```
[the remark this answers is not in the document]
ellistarn: I'm looking to cover this in another refactor.

pkg/packing/packer.go
ellistarn: This doesn't change the current logic at all, it just refactors it. The sort uses a `<` operator, so IIUC it will prefer smaller.

pkg/cloudprovider/aws/instancetypes.go
JacobGabrielson: 👍🏻

pkg/packing/packer.go
JacobGabrielson: nit: Actually I meant more like *why* euclidean being used? This just explains what it is. (Maybe I should have mentioned the `weightOf` comment doesn't really explain why either?)

pkg/cloudprovider/aws/instancetypes.go
JacobGabrielson: I think the `validate` change looked good.

pkg/packing/packer.go
ellistarn: As I'm writing up the "why", I'm remembering that we should just be using price as a single dimension ordering construct. Euclidean is just a proxy for price...

pkg/cloudprovider/aws/instancetypes.go
bwagner5: Where are we transforming from a map to a slice?  `filterFrom` returns a slice

ellistarn: Ah -- sorry you're right. FilterFrom is returning an aws.InstanceType (struct) and we need to convert to cloudprovider.InstanceType (interface). Annoying go inheritance things =/. 

```
cannot use p.filterFrom(instanceTypes, constraints, zones) (value of type []*InstanceType) as []cloudprovider.InstanceType value in return statement
```

ellistarn: Added a comment here.

JacobGabrielson (approved):

bwagner5 (approved):
nice refactor!

PR #369, opened 2021-04-15
Properly implemented exponential backoff for reconciliation failures

njtran:
https://github.com/awslabs/karpenter/issues/358

If it fails, it exponentially backs off from `100ms` to `10s` detailed below.

```100ms -> 200ms -> 400ms -> 800 ms -> 1.6s -> 3.2s -> 6.4s -> 10s -> 10s -> 10s ....```

### DIFF
--- a/pkg/controllers/controller.go
+++ b/pkg/controllers/controller.go
@@ -50,10 +50,11 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 		resource.StatusConditions().MarkFalse(v1alpha1.Active, "", err.Error())
 		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
 			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
+		return reconcile.Result{Requeue: true}, nil
 	} else {
 		resource.StatusConditions().MarkTrue(v1alpha1.Active)
 	}
-	// 5. Update Status using a merge patch
+	// 4. Update Status using a merge patch
 	if err := c.Status().Patch(ctx, resource, client.MergeFrom(persisted)); err != nil {
 		return reconcile.Result{}, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
 	}
--- a/pkg/controllers/manager.go
+++ b/pkg/controllers/manager.go
@@ -16,15 +16,20 @@ package controllers
 
 import (
 	"context"
+	"time"
 
 	"github.com/awslabs/karpenter/pkg/apis"
 	"github.com/awslabs/karpenter/pkg/utils/log"
+
+	"golang.org/x/time/rate"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/workqueue"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: pkg/controllers/manager.go [hunk 1]
```diff
@@ -54,16 +59,22 @@ func NewManagerOrDie(config *rest.Config, options controllerruntime.Options) Man
 
 // RegisterControllers registers a set of controllers to the controller manager
 func (m *GenericControllerManager) RegisterControllers(controllers ...Controller) Manager {
-	for _, controller := range controllers {
-		controlledObject := controller.For()
-		var builder = controllerruntime.NewControllerManagedBy(m).For(controlledObject)
-		if namedController, ok := controller.(NamedController); ok {
+	for _, c := range controllers {
+		controlledObject := c.For()
+		var builder = controllerruntime.NewControllerManagedBy(m).For(controlledObject).WithOptions(controller.Options{
```
ellistarn, line 64: IIUC controllerruntime's default rate limiter is this one: https://github.com/kubernetes/client-go/blob/0c029ff945dd24c074f455220b9a4ed0506f5f4d/util/workqueue/default_rate_limiters.go#L39.

I think we never want 10QPS on the bucket limiter. We probably want a max of 1 QPS if we configure it at all. Agree that we want to slow down the max backoff to 10 seconds rather than 1 second.

pkg/controllers/controller.go
```diff
@@ -46,14 +46,17 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
...
-		resource.StatusConditions().MarkFalse(v1alpha1.Active, "", err.Error())
-		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
-			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
+	var reconcileError error
+	if reconcileError = c.Controller.Reconcile(ctx, resource); reconcileError != nil {
+		resource.StatusConditions().MarkFalse(v1alpha1.Active, "", reconcileError.Error())
+		reconcileError = fmt.Errorf("Controller failed to reconcile kind %s, %s",
+			resource.GetObjectKind().GroupVersionKind().Kind, reconcileError.Error())
+		zap.S().Errorf(reconcileError.Error())
+		return reconcile.Result{Requeue: true}, reconcileError
```
ellistarn, line 55: If we error here, we will never patch the status into the resource. This will prevent us from updating any object status during failure scenarios.

pkg/controllers/manager.go
```diff
@@ -54,16 +59,22 @@ func NewManagerOrDie(config *rest.Config, options controllerruntime.Options) Man
 
 // RegisterControllers registers a set of controllers to the controller manager
 func (m *GenericControllerManager) RegisterControllers(controllers ...Controller) Manager {
-	for _, controller := range controllers {
-		controlledObject := controller.For()
-		var builder = controllerruntime.NewControllerManagedBy(m).For(controlledObject)
-		if namedController, ok := controller.(NamedController); ok {
+	for _, c := range controllers {
```
bwagner5, line 62: can we go back to `controller` rather than `c`? We've used that style most other places

pkg/controllers/manager.go
bwagner5, line 64: this can be `builder := ...`

njtran: I'll put it back to the enumerated default. 👍

pkg/controllers/controller.go
njtran: Won't returning here prevent us from updating any object status? Not just if we return an error?

pkg/controllers/manager.go
njtran: It was originally `controller` but since I added in this package `"sigs.k8s.io/controller-runtime/pkg/controller"`, I had to rename the controller to c.

pkg/controllers/controller.go
ellistarn: We want to patch the object status in all cases.

pkg/controllers/manager.go
bwagner5: ahh i c

pkg/controllers/controller.go
```diff
@@ -46,14 +46,17 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
...
-		resource.StatusConditions().MarkFalse(v1alpha1.Active, "", err.Error())
-		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
-			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
+	var reconcileError error
+	if reconcileError = c.Controller.Reconcile(ctx, resource); reconcileError != nil {
+		resource.StatusConditions().MarkFalse(v1alpha1.Active, "", reconcileError.Error())
+		reconcileError = fmt.Errorf("Controller failed to reconcile kind %s, %s",
+			resource.GetObjectKind().GroupVersionKind().Kind, reconcileError.Error())
+		zap.S().Errorf(reconcileError.Error())
+		return reconcile.Result{Requeue: true}, nil
```
ellistarn, line 55: Oh this is kind of brilliant and way more elegant than returning an error. Nice!

pkg/controllers/controller.go
```diff
@@ -46,14 +46,17 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 	// 2. Copy object for merge patch base
 	persisted := resource.DeepCopyObject()
 	// 3. Reconcile
-	if err := c.Controller.Reconcile(ctx, resource); err != nil {
-		resource.StatusConditions().MarkFalse(v1alpha1.Active, "", err.Error())
-		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
-			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
+	var reconcileError error
```
ellistarn, line 49: let's just use the normal idiom if the error is scoped

if err := ...; err != nil

ellistarn (approved):

PR #372, opened 2021-04-19
Licenses

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - Cleanup Github Actions
 - Add License verification CI check
 - Add THIRD_PARTY_LICENSES file


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,34 @@
...
+        path: coverage.html
+
+    - name: Verify Licenses
+      if: github.event_name == 'push'
+      run: make verify-licenses
+      env: 
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    
+
```
ellistarn, line 34: nit extra lines

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,34 @@
+name: CI
+on: [push, pull_request, workflow_dispatch]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repo
```
ellistarn, line 7: nit: "Checkout" or "Checkout Repository".

Is it required to have a name? IIUC you can just have uses, which names things cleanly.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,34 @@
...
+      uses: actions/checkout@v2
+
```
ellistarn, line 9: Why empty lines here?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Setup Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.15.3
+     
+    - name: CI - Verifications and Tests
```
ellistarn, line 15: Consider just dropping the name and following 
```
    - run: make toolchain
    - run: echo "/usr/local/kubebuilder/bin" >> $GITHUB_PATH
    - run: make ci
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,34 @@
...
+      run: |
+        make toolchain 
+        echo "/usr/local/kubebuilder/bin" >> $GITHUB_PATH
+        make ci
+
+    - name: Upload code coverage
+      uses: actions/upload-artifact@v2
+      with:
+        name: coverage
```
ellistarn, line 24: Is name required here

.github/workflows/ci.yaml
bwagner5: It's not required, but I think it makes the file nicer to read

Makefile
```diff
@@ -36,6 +38,10 @@ verify: ## Verify code. Includes dependencies, linting, formatting, etc
 	go fmt ./...
 	golangci-lint run
 
+verify-licenses: ## Verifies dependency licenses and requires GITHUB_TOKEN to be set
```
ellistarn, line 41: consider just `make licenses`

.github/workflows/ci.yaml
bwagner5: just readability, I can remove if you really don't like those lines, but I  think it's easier to read blocks like this.

Makefile
```diff
@@ -36,6 +38,10 @@ verify: ## Verify code. Includes dependencies, linting, formatting, etc
...
+	go build $(GOFLAGS) -o $(BUILD_DIR)/karpenter cmd/controller/main.go
```
ellistarn, line 42: Consider just dropping the binary at root and adding a .gitignore

Makefile
```diff
@@ -46,7 +52,8 @@ codegen: ## Generate code. Must be run if changes are made to ./pkg/apis/...
 	./hack/codegen.sh
 
 publish: ## Generate release manifests and publish a versioned container image.
-	$(WITH_RELEASE_REPO) $(WITH_GOFLAGS) ko resolve -B -t $(RELEASE_VERSION) -f config > $(RELEASE_MANIFEST)
+	@aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin $(RELEASE_REPO)
```
ellistarn, line 55: nice +1

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,34 @@
...
+        path: coverage.html
+
+    - name: Verify Licenses
+      if: github.event_name == 'push'
```
ellistarn, line 28: Should we just always check licenses?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,34 @@
+name: CI
+on: [push, pull_request, workflow_dispatch]
```
ellistarn, line 2: When do these other events occur? Push is too late, right? What triggers workflow dispatch?

Makefile
bwagner5: 👍  I suppose you get more details in `make help` anyways

.github/workflows/ci.yaml
bwagner5: name isn't required but I think this makes it more readable in the Github actions summary view where it just says "Run actions/upload-artifact@v2" which is pretty unclear as to what it's doing :)

.github/workflows/ci.yaml
bwagner5: we can't, github rate limits pretty aggressively so we need GITHUB_TOKEN which is only available on push

.github/workflows/ci.yaml
ellistarn: Curious if this is still true if we drop names

.github/workflows/ci.yaml
bwagner5: push is too late for early feedback on licenses but it's basically all we can do, so better than nothing (still catches it before releasing which is the ultimate goal). Workflow dispatch is a manual trigger, which is just a nice to have here (I've found it useful for debugging flaky tests before).

.github/workflows/ci.yaml
bwagner5: I'm a fan of the explicit names, is there any reason you don't want them?  It makes the file a lot clearer if each step is named appropriately and the run summary becomes clearer too.

Makefile
bwagner5: We'll be adding some more files that will need to be temporarily held and gitignored. I generally like the build dir that acts as a temp dir for the repo.  Any specific reasons you're against that approach?

.github/workflows/ci.yaml
ellistarn: IMO it's just unnecessary aliasing. The names (as they were), were already quite explicit to me. If they weren't explicit to you, that's probably a good signal that naming them could reduce confusion :). I prefer the more brief form since it's already reasonably named, but it's not a hill that I'm willing to die on and I think your perspective on this is at least (if not more) valid than mine.

.github/workflows/ci.yaml
ellistarn: Agree that for this step (and other parameterized actions), adding a name can increase clarity.

Makefile
bwagner5: The generated release assets like helm chart, docs, and k8s yamls.  I suppose they don't have to be ignored if we do the PR and generation in one step.

.github/workflows/ci.yaml
bwagner5: ha we can, get call

Makefile
bwagner5: removed build dir

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,27 @@
...
+        echo "/usr/local/kubebuilder/bin" >> $GITHUB_PATH
+        make ci
+      env: 
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Upload Code Coverage
+      uses: actions/upload-artifact@v2
+      with:
+        name: coverage
+        path: coverage.html
```
ellistarn, line 27: nit missing newline

Makefile
```diff
@@ -1,19 +1,19 @@
 RELEASE_REPO ?= public.ecr.aws/karpenter
 RELEASE_VERSION ?= $(shell git describe --tags --always)
-RELEASE_MANIFEST = releases/${CLOUD_PROVIDER}/manifest.yaml
+RELEASE_MANIFEST = releases/$(CLOUD_PROVIDER)/manifest.yaml
```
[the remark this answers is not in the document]
bwagner5: no difference so standardized on the more widely used `$(`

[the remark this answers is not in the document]
bwagner5: https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html#SEC58

.github/workflows/ci.yaml
bwagner5: haha trigger happy delete from last comment

.github/workflows/ci.yaml
JacobGabrielson: fwiw I find the multiple runs harder to read than the pipe syntax

JacobGabrielson (approved):

ellistarn (approved):

PR #374, opened 2021-04-20
Update ci.yaml

ellistarn:
Removes pull from CI workflow due to test flakes. All code will instead be tested on pull_request.

Issue #, if available:

Description of changes:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,5 +1,5 @@
 name: CI
-on: [push, pull_request, workflow_dispatch]
+on: [pull_request, workflow_dispatch]
 jobs:
   build:
     runs-on: ubuntu-latest


.github/workflows/ci.yaml
```diff
@@ -1,5 +1,5 @@
 name: CI
-on: [push, pull_request, workflow_dispatch]
```
njtran, line 2: Are you removing this because our tests are racey? Without each checking on each push, we might not catch errors that come up mid pr-review.

ellistarn: We requires that tests are run after any code change on PR, and tests must be run on the latest changes before checkin.

bwagner5 (approved):
lgtm :(

PR #375, opened 2021-04-20
Refactored Binpacking to live outside of the AWS Cloud Provider

ellistarn:
Issue #, if available:

Description of changes:
- Removed the concept of packing from the cloud provider
- Simplified the cloudprovider APIs (i.e. removal of constraints, etc)
- Implemented Binpacking and InstanceType filtering as part of the Allocator
- No longer allows accelerators to be selected by non-accelerator workloads.

Followons:
- Move more testing logic to allocator (from aws cloud provider)
- Reimplement Zone IDs


pkg/cloudprovider/types.go
```diff
@@ -48,19 +49,15 @@ type Capacity interface {
...
 
-// Constraints for an efficient binpacking solution of pods onto nodes, given
-// overhead and node constraints.
-type Constraints struct {
-	v1alpha1.Constraints
-	// Pods is a list of equivalently schedulable pods to be binpacked.
-	Pods []*v1.Pod
-	// Overhead resources per node from daemonsets.
-	Overhead v1.ResourceList
+// Packing TODO
```
JacobGabrielson, line 52: What is the todo here?

pkg/cloudprovider/aws/subnets.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func NewSubnetProvider(ec2api ec2iface.EC2API) *SubnetProvider {
+	return &SubnetProvider{
+		ec2api: ec2api,
+		cache:  cache.New(CacheTTL, CacheCleanupInterval),
+	}
+}
+
+func (s *SubnetProvider) GetZonalSubnets(ctx context.Context, cluster *v1alpha1.ClusterSpec) (map[string][]*ec2.Subnet, error) {
```
JacobGabrielson, line 41: I asked this earlier but I think the comment got lost ... Wondering if this should be called `Get` to match other code like the instance provider?

pkg/cloudprovider/aws/utils/utils.go
```diff
@@ -15,19 +15,24 @@ limitations under the License.
...
-func NormalizeArchitecture(architecture *string) *string {
-	if architecture == nil {
-		return nil
-	}
-	switch *architecture {
-	case v1alpha1.ArchitectureAmd64:
-		return aws.String("x86_64")
-	default:
-		return architecture
+var KubeToAWSArchitectures = map[string]string{
```
JacobGabrielson, line 21: Wondering if this should be private given the accessor functions?

pkg/cloudprovider/types.go
ellistarn: A reminder for me to write a nice comment here :)

pkg/cloudprovider/aws/subnets.go
ellistarn: I played with this, but unlike the other accessors, this returns a map, which I think requires some clarity of what is actually in the map. In this case, it's zone -> subnets, or ZonalSubnets. Otherwise I might be tempted to write code like

subnetss = subnetProvider.Get(...)
for zone, subnets := range subnetss {
   ...
}

JacobGabrielson (dismissed):

pkg/cloudprovider/aws/utils/utils.go
```diff
@@ -15,19 +15,24 @@ limitations under the License.
...
-	default:
-		return architecture
+var kubeToAWSArchitectures = map[string]string{
+	v1alpha1.ArchitectureAmd64: "x86_64",
+	v1alpha1.ArchitectureArm64: v1alpha1.ArchitectureArm64,
+}
+
+// EC2Architecture translates architecture into an AWS-recognized architecture name
+func EC2Architecture(architecture string) string {
+	return kubeToAWSArchitectures[architecture]
```
njtran, line 28: are constraints validated and cleaned? If they're not, can run into an error here.

ellistarn: We use GetArchitectures in the cloud provider api to ensure this invariant.

njtran (commented):
One comment, LGTM otherwise

pkg/cloudprovider/aws/subnets.go
prateekgogia, line 41: Looks like we use only the cluster name here, can we only pass in the name as opposed to passing in the complete spec?

ellistarn: Yeah I went back and forth on this and I think I agree with you.

JacobGabrielson (approved):

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}
+	return resources.Quantity(fmt.Sprint(count))
+}
+
+func (i *InstanceType) AWSNeurons() *resource.Quantity {
+	count := int64(0)
+	if i.InferenceAcceleratorInfo != nil {
+		for _, accelerator := range i.InferenceAcceleratorInfo.Accelerators {
+			count += *accelerator.Count
```
eptiger, line 83: I don't follow what neurons means in this context - how are "accelerators" used in instance type selection? Is this  basically number of inferentia chips?

ellistarn: You got it
```
resources:
  requests:
     amazonaws.com/neuron: 3 # get me a machine w/ 3 chips
```

Issue #381, opened 2021-04-26
Make Karpenter pod QoS Guaranteed


njtran:
Currently the Karpenter pod has QoS Burstable. We should make this Guaranteed.

ellistarn:
Nice! Yes!

Issue #380, opened 2021-04-26
Karpenter's Controller Pod Priority should be very high


ellistarn:
We should allow karpenter to preempt non-karpenter pods, similar to other kube-system pods. The priority should be set as high as possible without interfering with kube-system pods. 

IIRC, kube-system pods run at 2,000,000 priority, and the max you can set outside kube-system is 1,000,000. Let's max it.

PR #397, opened 2021-05-04
Set Karpenter Pod to high priority and guaranteed

njtran:
Issue #, if available:
#381 #380 

Description of changes:

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -27,6 +27,14 @@ spec:
   selector:
     control-plane: karpenter
 ---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: karpenter-high-priority
+value: 1000000
+globalDefault: false
+
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -44,6 +52,7 @@ spec:
       labels:
         control-plane: karpenter
     spec:
+      priorityClassName: karpenter-high-priority
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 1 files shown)


config/controller.yaml
```diff
@@ -65,6 +74,9 @@ spec:
           requests:
             cpu: 100m
             memory: 20Mi
```
ellistarn, line 76: This seems way to low. Can we default to 1 cpu and 1gb?

njtran: Sure! I think these are pretty low as well, but didn't know if you had these for a reason. I want this to be able to fit on nodes with 1vCPU as well (assuming not that many daemonsets). WDYT about `800m` and `1Gi` ?

ellistarn: Nodes with 1vcpu (the t3.micros) have much more overhead than that 800 won't fit. Feel free to test it out ;)

njtran: For reference, `kube-proxy` requests `100m` and `aws-node` requests `10m`

prateekgogia: When I was doing some testing for the demo, I saw the Karpenter pod got OOM killed. IIRC, it was set to 20Mi limit. I think the real number can only be decided with scale testing. 
Can we skip setting limits for now, or do we need for something?

njtran: I'd like to at least increase it to make it a little more resistant to OOM killing until we get the scale testing done.

ellistarn: Agree with Nick.

ellistarn (approved):

ellistarn (approved):

PR #399, opened 2021-05-06
remove monitoring ns deletion

njtran:
Issue #, if available:

Description of changes:
We removed Prometheus, so no pods remain in the `monitoring` namespace. We don't want to delete namespaces that we do not own. 


ellistarn (approved):
:ty

Issue #105, opened 2020-11-03
[Termination Handler] Explore graceful termination for EC2 Instances


ellistarn:
As described in the [Design](https://github.com/ellistarn/karpenter/blob/master/docs/DESIGN.md#scale-down-edge-cases), termination handlers must be layered independently on top of Karpenter's autoscaler. By design, the node termination handler should have no knowledge of autoscaling behavior or configuration, or what even triggered the scale down (e.g. manual, preemption, autoscaling). 

Potential requirements/solutions include:
- Protect instances that are being deleted/scaled down to respect poddisruptionbudgets
- Build a Karpenter CRD to model lifecycle hooks?
- Use some sort of CloudProvider model to hook into EC2 lifecycle hooks to protect instances.

ellistarn:
My current stance on this is that we should recommend that users rely on https://github.com/aws/aws-node-termination-handler for at least v0.4. We should explore building a karpenter native interruption handler, but this use case (rebalance) should be scoped into the defragmentation design.

PR #405, opened 2021-05-17
Graceful Termination Design Doc

njtran:
#105 

Description of changes:
This introduces the Graceful Termination Design Doc. 


docs/designs/termination.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Parallel Node Termination
+
+The termination controller will be able to drain and terminate multiple nodes in parallel. We will expedite this process while rate limiting node termination to protect against throttling.
+
+We introduce an optional cluster-scoped CRD, the Node Disruption Budget (NDB), a [Pod Disruption Budget (PDB)](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets) for nodes. A user can scope an NDB to Karpenter provisioned nodes through a label selector, since nodes are created with their Provisioner’s labels. `Unavailable` nodes will be `NotReady` or have `metadata.DeletionTimestamp` set. `Available` nodes will be `Ready`.
+
+A termination is allowed if at least minAvailable nodes selected by selector will still be available after the termination. For example, you can prevent all terminations by specifying “100%”. A termination is also allowed if at most maxUnavailable nodes selected by selector are unavailable after the termination. For example, one can prevent all terminations by specifying 0. These two are mutually exclusive.
```
bwagner5, line 56: nit: "by a selector"

"maxUnavailable nodes selected by a selector"

Can you also say the parameter names "minAvailable" and "maxUnavailable" are the mutually exclusive parameters rather than just saying "These two"?

docs/designs/termination.md
```diff
@@ -0,0 +1,130 @@
...
+spec:
+  minAvailable: 80%
+  selector:
+    matchLabels:
+      karpenter.sh/name: default
+      karpenter.sh/namespace: default
+  maxUnavailable: 5%
+```
+
+Since most eviction failures come from a PDB violation, we will queue up evictions serially per PDB, where the PDB’s selector will act as the key. The pods will kick off eviction serially, run asynchronously, and exponentially back off and retry if they fail. This way, evicting pods will only be limited by other evicting pods managed by the same PDB, since each pod can only belong to one PDB. Pods that have a `do-not-evict` label will not be queued up for eviction, as we prevent the node with a `do-not-evict` pod from draining.
```
bwagner5, line 92: What does "The pods will kick off eviction serially"? Do you mean the controller will evict pods serially?

docs/designs/termination.md
```diff
@@ -0,0 +1,130 @@
+# Karpenter Graceful Node Termination
+
+## Overview
+Karpenter's scale down implementation is currently a proof of concept. The reallocation controller encodes two actions. Nodes are elected for termination when there aren't any pods scheduled to them. Node termination follows cordon and drain [best practices](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/).
```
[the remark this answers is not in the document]
njtran: Good ideas. Thanks!

docs/designs/termination.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Termination Workflow
+The current termination workflow finds nodes with the label `karpenter.sh/lifecycle-phase: terminable` then cordons, drains, and deletes them. This naively handles a few error cases, has no user safeguards, and makes no effort to rate limit. The new workflow will improve these but continue to monitor nodes with the same labels.
+
+The new termination process will begin with node that receives a delete request. After Karpenter validates the request, a Karpenter mutating webhook will add the Karpenter [finalizer](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers) to the node. Then, we cordon and begin draining the node. After no pods remain, we terminate the instance in the cloud provider, then remove the Karpenter finalizer. This will result in the APIServer deleting the node object.
```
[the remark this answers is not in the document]
njtran: Good catch!

docs/designs/termination.md
```diff
@@ -0,0 +1,130 @@
...
+
+After all pods evict, we will attempt to terminate the node. If the cloud-provider API returns an error, preventing instance termination, we will log the error in the Karpenter pod and node status then exponentially back off and retry.
```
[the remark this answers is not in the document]
njtran: Thanks for the first one. The second I think is correct, but it can be misleading. We will log the error in the Karpenter pod, and also change the node status. I'll make changes to reflect that more accurately.

docs/designs/termination.md
njtran: Thanks again. Good idea here, better to be explicit.

docs/designs/termination.md
njtran: I changed it to "The controller will call evictions serially". WDYT?

docs/designs/termination.md
```diff
@@ -0,0 +1,130 @@
+# Karpenter Graceful Node Termination
+
+## Overview
+Karpenter's scale down implementation is currently a proof of concept. The reallocation controller implements two actions. First, nodes are elected for termination when there aren't any pods scheduled to them. Second, Node are cordoned and drained and deleted. Node termination follows cordon and drain [best practices](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/).
```
bwagner5, line 4: nit: "Second, nodes are cordoned, drained, and then deleted."

docs/designs/termination.md
```diff
@@ -0,0 +1,130 @@
+# Karpenter Graceful Node Termination
+
+## Overview
+Karpenter's scale down implementation is currently a proof of concept. The reallocation controller implements two actions. First, nodes are elected for termination when there aren't any pods scheduled to them. Second, Node are cordoned and drained and deleted. Node termination follows cordon and drain [best practices](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/). 
+
+This design explores robustness improvements to the termination process and proposes the separation of this logic into a new termination controller.
```
bwagner5, line 6: You can probably remove "robustness" here, the design is just improving up the termination process.  Just want to make the doc as clear and concise as possible :)

bwagner5: The logic separation into a new termination controller will still live in the same binary right?  If so, can you make that explicit so it doesn't seem like this new thing is going to be a stand-alone component.

docs/designs/termination.md
```diff
@@ -0,0 +1,130 @@
...
+* Terminating a node with Karpenter will not leak the instance
+* If termination is requested, the node will eventually terminate
+* Users will be able to implement node deletion safeguards
+* Termination mechanisms will rate-limit at the pod evictions
+
+## Termination Controller
+The termination controller is responsible for gracefully terminating instances. The termination controller draws ideas from the [aws-node-termination-handler](https://github.com/aws/aws-node-termination-handler), [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node), and [best practices](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/) on drain given by the Kubernetes community.
+
+### Termination Workflow
+The current termination workflow finds nodes with the label `karpenter.sh/lifecycle-phase: terminable` then cordons, drains, and deletes them. This naively handles a few error cases, has no user safeguards, and makes no effort to rate limit. The new workflow will improve these but continue to monitor nodes with the same labels.
```
bwagner5, line 20: Will rate limiting be added to the labels, or are we cutting our losses here? It does seem like you could implement rate limiting w/ the current label approach. But again, I'm not sure it adds tremendous value for something that users probably shouldn't use anyways.

docs/designs/termination.md
```diff
@@ -0,0 +1,130 @@
...
+![](../images/termination-state-machine.png)
+### Triggering Termination
+
+The current termination process acts on a reconcile loop. We will change the termination controller to watch nodes and manage the Karpenter [finalizer](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers), making it responsible for all node termination and pod eviction logic. 
+
+Finalizers allow controllers to implement asynchronous pre-deletion hooks and are commonly used with CRDs like Karpenter’s Provisioners. Today, a user can call kubectl delete node to delete a node object, but will end up leaking the underlying instance by only deleting the node object in the cluster. We will use finalizers to gracefully terminate underlying instances before Karpenter provisioned nodes are deleted, preventing instance leaking. Relying on kubectl for terminations gives the user more control over their cluster and a Kubernetes-native way of deleting nodes - as opposed to the status quo  of doing it manually in a cloud provider console.
```
bwagner5, line 28: we had discussed previously about drawbacks for using finalizers, have those all been ruled out, or can you list some potential drawbacks?

bwagner5 (commented):
Just a few more questions and some nits. Overall, looks really good. I have no problems with the plan.

docs/designs/termination.md
njtran: Good callouts, improvements are improvements :)

The new termination controller will live in the same binary, but won't require a provisioner to be created for it to work. In that sense, it'll be a stand-alone component, but still a part of Karpenter. Thanks!

docs/designs/termination.md
njtran: Good point. I like rate limiting, but not if it's a necessity for the user to configure. For each label, we have innate rate-limiting. 
- A user can rate limit node deletions with a NodeDisruptionBudgets if they please. 
- While we drain a node, we mechanically limit the rate of pod evictions and exponentially back off and retry if they fail.
-For instance termination failures, we also exponentially back off and retry.

Since these are the places where labels could be specified, I think this is enough rate limiting, and it requires no user configuration. If we run into throttling issues in the future, we can explore more options to further rate limit each of these processes.

docs/designs/termination.md
njtran: We discuss the drawbacks in the final section before the Appendix. We've narrowed down the potential for a drawback here to a very small subset of cases by only creating the finalizer once the request is accepted as valid given the state of the cluster (and potential NDBs). Let me know if you think I should add more on this.

bwagner5: Good enough for me :) thanks

ellistarn (approved):

bwagner5 (approved):
Very nice doc, great job!

PR #406, opened 2021-05-18
[Bug Fix] Added Missing IAM Permission

ollypom:
**Description of changes:**
When working through the getting started docs I noticed an IAM permission was missing in the sample IAM cloudformation.

```
2021-05-18T09:08:12.349Z        ERROR   Continuing after failing to create capacity, getting launch template, getting instance profile, retrieving instance profile KarpenterNodeInstanceProfile-olly-karpenter-demo, AccessDenied: User: arn:aws:sts::<accountid>:assumed-role/KarpenterControllerRole-olly-karpenter-demo/1621328854438905270 is not authorized to perform: iam:GetInstanceProfile on resource: instance profile KarpenterNodeInstanceProfile-olly-karpenter-demo
```

### DIFF
--- a/docs/aws/karpenter.cloudformation.yaml
+++ b/docs/aws/karpenter.cloudformation.yaml
@@ -58,6 +58,7 @@ Resources:
               - "ec2:DescribeInstanceTypes"
               - "ec2:DescribeInstanceTypeOfferings"
               - "ec2:DescribeAvailabilityZones"
+              - "iam:GetInstanceProfile"
               - "ssm:GetParameter"
   KarpenterNodeInstanceProfile:
     Type: "AWS::IAM::InstanceProfile"


ellistarn:
Hey @ollypom. Thanks for reporting this issue. I think you must have a mismatch between your karpenter docs and the controller version. We actually removed our dependency on IAM entirely in v0.2.3 (to support airgapped VPCs) so you shouldn't need this permission.

I recommend redeploying the cloudformation stack and making sure that you're installing karpenter on v0.2.3. I'll also take a quick look at that helm chart and double check everything is working as expected.

ollypom:
Ah @ellistarn thank you for pointing me in the right direction 😄 

I haven't tested v0.2.3, but I can see how I ended up with Karpenter v0.2.1. It looks like the [install script](https://raw.githubusercontent.com/awslabs/karpenter/v0.2.3/hack/quick-install.sh) deploys version 0.2.1 of the Helm Chart.

ellistarn:
Nice catch! Happy to accept a PR removing the version on that install script so we can always be latest. Finders keepers if you want the PR. Otherwise I can take care of it.

ollypom:
Thanks, however it looks like the change has already been made to the [script](https://github.com/awslabs/karpenter/blob/main/hack/quick-install.sh.) in main. I don't think I can change the tagged / released version of that file (but I could be wrong), so you may need to cut another release or point the docs to the main branch.

PR #407, opened 2021-05-18
Explicitly set output of aws cli to json

pjlewisuk:
**Description of changes:**
 When working through the getting started docs I noticed an issue with `aws` commands that were piped to `jq`, and cause an error like:

```
parse error: Invalid numeric literal at line 1, column 8
```

This was because my default output was not set to `json` for the AWS CLI, so added the `--output json` option to the relevant commands.

### DIFF
--- a/docs/aws/README.md
+++ b/docs/aws/README.md
@@ -30,7 +30,7 @@ aws cloudformation deploy \
   --stack-name Karpenter-${CLUSTER_NAME} \
   --template-file ./docs/aws/karpenter.cloudformation.yaml \
   --capabilities CAPABILITY_NAMED_IAM \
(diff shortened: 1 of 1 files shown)


bwagner5:
Thanks for bringing this up! Rather than specifying the `--output json` in each command, would you mind adding the environment variable output config at the top of the file in the `Environment` section?  

Putting it there will reduce the amount of bloat in the commands and we won't have to remember to add the output flag on any new aws cli commands we may add to this page.

https://github.com/awslabs/karpenter/blob/c85f7e58ac8d43f6cb542948fd8e72d15cab13ca/docs/aws/README.md#environment 

```
...
export AWS_DEFAULT_OUTPUT="json"
```

bwagner5 (changes requested):
move aws cli output config to environment block

pjlewisuk:
@bwagner5 thanks for the feedback, updated as per your suggestion

bwagner5 (approved):
lgtm! Awesome, thanks for helping to keep our docs clean and correct!

PR #409, opened 2021-05-18
Allow Karpenter controller to run on Fargate

andskli:
Issue #, if available:

Description of changes:
Karpenter currently relies on EC2 instance metadata service (IMDS) to figure out which AWS region it is running in. IMDS is not available for pods running on EKS Fargate ([ref](https://docs.aws.amazon.com/eks/latest/userguide/fargate.html)). This change preserves region discovery via EC2 IMDS if the Go SDK has not picked up region configuration via the environment.

### DIFF
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -4,6 +4,10 @@ serviceAccount:
   # Annotations to add to the service account (like the ARN of the IRSA role)
   annotations: {}
 controller:
+  # List of environment items to add to the controller, for example
+  # - name: AWS_REGION
+  #   value: eu-west-1
+  env: []
   nodeSelector: {}
   tolerations: []
   affinity: {}
--- a/pkg/cloudprovider/aws/factory.go
+++ b/pkg/cloudprovider/aws/factory.go
@@ -29,6 +29,7 @@ import (
 	"github.com/awslabs/karpenter/pkg/utils/log"
 	"github.com/awslabs/karpenter/pkg/utils/project"
 	"github.com/patrickmn/go-cache"
+	"go.uber.org/zap"
 )
 
 const (
@@ -51,10 +52,15 @@ type Factory struct {
 }
 
 func NewFactory(options cloudprovider.Options) *Factory {
-	sess := withUserAgent(withRegion(session.Must(
+	sess := withUserAgent(session.Must(
 		session.NewSession(request.WithRetryer(
 			&aws.Config{STSRegionalEndpoint: endpoints.RegionalSTSEndpoint},
-			utils.NewRetryer())))))
+			utils.NewRetryer()))))
+	if *sess.Config.Region == "" {
+		zap.S().Debug("AWS region not configured, asking EC2 Instance Metadata Service")
+		*sess.Config.Region = getRegionFromIMDS(sess)
+	}
+	zap.S().Debugf("Using AWS region %s", *sess.Config.Region)
 	ec2api := ec2.New(sess)
 	launchTemplateProvider := &LaunchTemplateProvider{
 		ec2api: ec2api,
@@ -86,11 +92,11 @@ func (f *Factory) CapacityFor(provisioner *v1alpha1.Provisioner) cloudprovider.C
 	}
 }
 
-func withRegion(sess *session.Session) *session.Session {
+// get the current region from EC2 IMDS
+func getRegionFromIMDS(sess *session.Session) string {
 	region, err := ec2metadata.New(sess).Region()
 	log.PanicIfError(err, "failed to call the metadata server's region API")
-	sess.Config.Region = aws.String(region)
-	return sess
+	return region
 }
 
 // withUserAgent adds a karpenter specific user-agent string to AWS session
--- a/charts/karpenter/templates/controller.yaml
+++ b/charts/karpenter/templates/controller.yaml
@@ -79,6 +79,10 @@ spec:
             - mountPath: /tmp/k8s-webhook-server/serving-certs
               name: cert
               readOnly: true
+          env:
+          {{- with .Values.controller.env }}
+          {{- toYaml . | nindent 10 }}
+          {{- end }}
       # https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8#issuecomment-636888074
       securityContext:
         fsGroup: 1000


charts/karpenter/values.yaml
```diff
@@ -6,3 +6,6 @@ controller:
   tolerations: []
   affinity: {}
   image: "public.ecr.aws/karpenter/controller:v0.2.3@sha256:ad4332ddce1afb1080b269eeee70592c29e3e34ecd9f9f4dabe4102c415817e9"
+
+aws:
```
ellistarn, line 10: Do you mind implementing this using a env map? This will allow you to wire this in without AWS specific changes.
```
    {{- range .Values.controller.environment }}
    - name: {{ .name }}
      value: {{ .value }}
    {{- end }} 
```

pkg/cloudprovider/aws/factory.go
```diff
@@ -55,6 +56,7 @@ func NewFactory(options cloudprovider.Options) *Factory {
 		session.NewSession(request.WithRetryer(
 			&aws.Config{STSRegionalEndpoint: endpoints.RegionalSTSEndpoint},
 			utils.NewRetryer())))))
+	log.PrettyInfof("REGION: %v", *sess.Config.Region)
```
ellistarn, line 59: Friendly reminder to delete this.

pkg/cloudprovider/aws/factory.go
```diff
@@ -87,9 +89,15 @@ func (f *Factory) CapacityFor(provisioner *v1alpha1.Provisioner) cloudprovider.C
 }
 
 func withRegion(sess *session.Session) *session.Session {
-	region, err := ec2metadata.New(sess).Region()
-	log.PanicIfError(err, "failed to call the metadata server's region API")
-	sess.Config.Region = aws.String(region)
+	region, envIsSet := os.LookupEnv("AWS_REGION")
```
ellistarn, line 92: I think if you use AWS_DEFAULT_REGION, it will be overridden automatically. 

https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

ellistarn (changes requested):
Nice contribution! Thank you!

charts/karpenter/values.yaml
andskli: I agree it would be useful to allow passing arbitrary env vars to the controller. I'm thinking this could be done similarly to how it's dealt with in cluster autoscaler ([ref](https://github.com/helm/charts/blob/fd80db1fed08f4e4b9555d739a10f0242fec34bf/stable/cluster-autoscaler/templates/deployment.yaml#L145)), so a user could end up with something like:

```
aws:
  region: eu-west-1

controller:
  extraEnv:
    MY_LOGCONFIG: my_value
```

The upside of this approach is the clear way of specifying certain well-known options, while the downside is that a user could end up with duplicate definitions of env vars, which would be accepted by k8s, but not reflected properly inside the container.

What's your thoughts?

pkg/cloudprovider/aws/factory.go
andskli: Seems like the safe way would be to look for `AWS_DEFAULT_REGION` and use that if set, otherwise fall back to `AWS_REGION`. Adding this behavior.

charts/karpenter/values.yaml
ellistarn: I'd prefer simply the above:
```
    {{- range .Values.controller.environment }}
    - name: {{ .name }}
      value: {{ .value }}
    {{- end }} 
```
The cluster autoscaler suffers from complexity bloat, which is one of the reasons I'm trying to draw a stronger line on our helm chart. I don't think it results in significantly more compelx helm commands. WDYT?

pkg/cloudprovider/aws/factory.go
```diff
@@ -86,9 +87,24 @@ func (f *Factory) CapacityFor(provisioner *v1alpha1.Provisioner) cloudprovider.C
 	}
 }
 
+// Checks for env vars in the defined order and if none of those are set falls back to using EC2
+// IMDS for region lookup if neither AWS_DEFAULT_REGION or AWS_REGION is defined.
 func withRegion(sess *session.Session) *session.Session {
+	if region, envAwsDefaultRegion := os.LookupEnv("AWS_DEFAULT_REGION"); envAwsDefaultRegion {
+		log.PrettyInfof("Found AWS_DEFAULT_REGION env var set, using: %v", region)
```
ellistarn, line 94: Log.PrettyInfo is just for debugging. Do you mind using zap?

pkg/cloudprovider/aws/factory.go
```diff
@@ -86,9 +87,24 @@ func (f *Factory) CapacityFor(provisioner *v1alpha1.Provisioner) cloudprovider.C
 	}
 }
 
+// Checks for env vars in the defined order and if none of those are set falls back to using EC2
+// IMDS for region lookup if neither AWS_DEFAULT_REGION or AWS_REGION is defined.
 func withRegion(sess *session.Session) *session.Session {
+	if region, envAwsDefaultRegion := os.LookupEnv("AWS_DEFAULT_REGION"); envAwsDefaultRegion {
```
ellistarn, line 93: nit: typical go style is:
```
if region, ok := os.LookupEnv("AWS_DEFAULT_REGION"); ok {
   ...
}
```

pkg/cloudprovider/aws/factory.go
```diff
@@ -86,9 +87,24 @@ func (f *Factory) CapacityFor(provisioner *v1alpha1.Provisioner) cloudprovider.C
...
+		log.PrettyInfof("Found AWS_DEFAULT_REGION env var set, using: %v", region)
+		sess.Config.Region = aws.String(region)
+		return sess
+	}
+
+	if region, envAwsRegion := os.LookupEnv("AWS_REGION"); envAwsRegion {
```
ellistarn, line 99: I'm wary of supporting this legacy field, as it can confuse users and is supported inconsistently. Given that we made this explicit in the helm chart, I'd rather just support one flag.

pkg/cloudprovider/aws/factory.go
```diff
@@ -86,9 +87,24 @@ func (f *Factory) CapacityFor(provisioner *v1alpha1.Provisioner) cloudprovider.C
...
+		log.PrettyInfof("Found AWS_REGION env var set, using: %v", region)
+		sess.Config.Region = aws.String(region)
+		return sess
+	}
+
 	region, err := ec2metadata.New(sess).Region()
 	log.PanicIfError(err, "failed to call the metadata server's region API")
+	log.PrettyInfof("Retrieved region %v from EC2 Instance Metadata", region)
```
ellistarn, line 107: I'd love if we had parity for these log lines. We also tend to do debug statements for the cloud provider to avoid polluting the logs.
```
zap.S().Debug("Discovered region %s from env['AWS_DEFAULT_REGION']")
zap.S().Debug("Discovered region %s from IMDS")
```

charts/karpenter/values.yaml
andskli: Given that your suggested approach provides better flexibility, let's go with that. It is however a bit quirky to set those values from the helm command line given the [limitations of](https://helm.sh/docs/intro/using_helm/#the-format-and-limitations-of---set) `--set`. To achieve the desired result I end up adding `--set "controller.environment[0].name=AWS_REGION,controller.environment[0].value=eu-west-1"`, as opposed to `--set aws.region=eu-west-1` to my commands.

If the opinion changes, it's easier to add specific values rather than remove them.

charts/karpenter/values.yaml
```diff
@@ -4,7 +4,11 @@ serviceAccount:
   # Annotations to add to the service account (like the ARN of the IRSA role)
   annotations: {}
 controller:
+  # List of environment items to add to the controller, for example
+  # - name: AWS_DEFAULT_REGION
+  #   value: eu-west-1
+  environment: []
```
ellistarn, line 10: tiny nit: I'd just use the direct 1:1 naming w/ pod spec, in this case `env`

charts/karpenter/values.yaml
```diff
@@ -4,7 +4,11 @@ serviceAccount:
   # Annotations to add to the service account (like the ARN of the IRSA role)
   annotations: {}
 controller:
+  # List of environment items to add to the controller, for example
+  # - name: AWS_DEFAULT_REGION
```
ellistarn, line 8: I think this variable should be AWS_REGION from what we've learned, now.

ellistarn (dismissed):

ellistarn:
Nice! Needs a rebase though

charts/karpenter/values.yaml
andskli: Agreed.

andskli:
Messed up a bit with my rebase, merged all changes into a single commit and force pushed. Understandably GH would like another approval now

ellistarn (approved):
Nicely done!

PR #410, opened 2021-05-18
helm chart: namespace related fixes

andskli:
Issue #, if available:

Description of changes:
* namespaces was hard-coded into the templates, this is now templated via helm
* exposes the creation of service account in the helm chart, so that creation can be prevented if the SA already exist (for example, if created via other IaC tooling)

### DIFF
--- a/charts/karpenter/templates/manifests.yaml
+++ b/charts/karpenter/templates/manifests.yaml
@@ -4,7 +4,7 @@ metadata:
   creationTimestamp: null
   name: karpenter-mutating-webhook-configuration
   annotations:
-    cert-manager.io/inject-ca-from: karpenter/karpenter-serving-cert
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/karpenter-serving-cert
 webhooks:
   - admissionReviewVersions:
       - v1beta1
@@ -12,7 +12,7 @@ webhooks:
       caBundle: Cg==
       service:
         name: karpenter-webhook-service
-        namespace: karpenter
+        namespace: {{ .Release.Namespace }}
         path: /mutate-provisioning-karpenter-sh-v1alpha1-provisioner
     failurePolicy: Fail
     name: mutation.provisioning.karpenter.sh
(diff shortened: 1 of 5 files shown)


charts/karpenter/templates/manifests.yaml
```diff
@@ -4,15 +4,15 @@ metadata:
   creationTimestamp: null
   name: karpenter-mutating-webhook-configuration
   annotations:
-    cert-manager.io/inject-ca-from: karpenter/karpenter-serving-cert
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/karpenter-serving-cert
```
ellistarn, line 7: This YAML is generated by ./hack/codegen.sh. Can you inject those changes there?

charts/karpenter/templates/rbac.yaml
```diff
@@ -1,12 +1,14 @@
...
-  namespace: karpenter
+  namespace: {{ .Release.Namespace }}
   {{- with .Values.serviceAccount.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+{{- end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
```
ellistarn, line 14: I think we should include the clusterrolebinding inside the create flag, since the referenced ServiceAccount doesn't exist. WDYT?

ellistarn (changes requested):
This is great, thank you!

charts/karpenter/templates/rbac.yaml
andskli: I'm thinking that if the SA is configured via other mechanisms (TF, eksctl, etc), we should probably template the SA name to allow customization and keep the ClusterRoleBinding, similar to how it's done in cluster autoscaler ([ref](https://github.com/helm/charts/blob/fd80db1fed08f4e4b9555d739a10f0242fec34bf/stable/cluster-autoscaler/values.yaml#L121)).

Adding a commit for that change here and adjusting PR description

charts/karpenter/templates/manifests.yaml
andskli: Thanks, missed that. Updated codegen.sh, ran it and commited changes.

ellistarn (approved):
Nicely done!

Issue #361, opened 2021-04-14
Ignore pods that aren't healthy when calculating scale down


ellistarn:
I had a couple pods:
```
default        inflate-7bb86df7fb-sztk5                   0/1     Preempting   0          37m
default        inflate-7bb86df7fb-trmbb                   0/1     OutOfcpu     0          36m
```
Yet the instance never scaled down.

JacobGabrielson:
@ellistarn any idea why the `OutOfCpu` happened? Is that because the pod spec didn't ask for enough?

ellistarn:
I manually injected this error condition to simulate this case.

njtran:
Should we do this for any pod with this status? 

```
  - lastProbeTime: null
    lastTransitionTime: "2021-05-05T19:22:45Z"
    status: "False"
    type: Ready
```

njtran:
Additionally
```
    - lastProbeTime: null
      lastTransitionTime: "2021-05-05T19:22:39Z"
      status: "True"
      type: Initialized
```

ellistarn:
Report
```
➜  karpenter git:(outofcpu) ✗ k get pods
NAME                       READY   STATUS              RESTARTS   AGE
inflate-646745bfb8-29zqz   0/1     Evicted             0          3m55s
inflate-646745bfb8-2zkhb   0/1     Evicted             0          108s
inflate-646745bfb8-8cm7b   0/1     Evicted             0          6m59s
inflate-646745bfb8-d9fwn   0/1     Evicted             0          5m35s
inflate-646745bfb8-dgdfn   0/1     Evicted             0          56s
inflate-646745bfb8-fw9wr   0/1     Evicted             0          5m9s
inflate-646745bfb8-h2gwm   0/1     Evicted             0          48s
inflate-646745bfb8-r5h5k   0/1     Evicted             0          52s
inflate-646745bfb8-trg9t   0/1     Evicted             0          5m31s
inflate-646745bfb8-v99hq   0/1     Evicted             0          2m36s
inflate-646745bfb8-w9klx   0/1     ContainerCreating   0          36s
inflate-646745bfb8-zspwn   0/1     Evicted             0          50s
➜  karpenter git:(outofcpu) ✗ k get nodes
NAME                                            STATUS     ROLES    AGE     VERSION
ip-192-168-130-21.us-west-2.compute.internal    Ready      <none>   6m59s   v1.18.18
[...]

PR #414, opened 2021-05-19
Evicted pods are not counted in utilization

njtran:
#361 
Description of changes:

### DIFF
--- a/pkg/utils/node/predicates.go
+++ b/pkg/utils/node/predicates.go
@@ -16,7 +16,7 @@ package node
 
 import (
 	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha1"
-	"github.com/awslabs/karpenter/pkg/utils/scheduling"
+	"github.com/awslabs/karpenter/pkg/utils/pod"
 	v1 "k8s.io/api/core/v1"
 	"time"
 )
@@ -44,8 +44,11 @@ func IsPastTTL(node *v1.Node) bool {
 
 // IsUnderutilized returns if the node has 0 non-daemonset pods
 func IsUnderutilized(node *v1.Node, pods []*v1.Pod) bool {
-	for _, pod := range pods {
-		if !scheduling.IsOwnedByDaemonSet(pod) {
+	for _, p := range pods {
+		if pod.HasFailed(p) {
+			continue
+		}
+		if !pod.IsOwnedByDaemonSet(p) {
 			return false
 		}
 	}
--- a/pkg/controllers/provisioning/v1alpha1/allocation/constraints.go
+++ b/pkg/controllers/provisioning/v1alpha1/allocation/constraints.go
@@ -17,10 +17,10 @@ package allocation
 import (
 	"context"
 	"fmt"
+	"github.com/awslabs/karpenter/pkg/utils/pod"
 
 	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha1"
 	"github.com/awslabs/karpenter/pkg/packing"
-	"github.com/awslabs/karpenter/pkg/utils/scheduling"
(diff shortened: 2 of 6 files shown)


pkg/utils/scheduling/scheduling.go
```diff
@@ -60,6 +60,13 @@ func IsSchedulable(pod *v1.PodSpec, node *v1.Node) bool {
 	return true
 }
 
+func IsForcefullyEvicted(pod *v1.Pod) bool {
+	if pod.Status.Phase == "Failed" && pod.Status.Reason == "Evicted" {
```
ellistarn, line 64: Do we need to specify evicted? What about just failed?

pkg/utils/scheduling/scheduling.go
```diff
@@ -60,6 +60,13 @@ func IsSchedulable(pod *v1.PodSpec, node *v1.Node) bool {
...
+		return true
+	}
+	return false
```
ellistarn, line 67: Just return the predicate directly

njtran: 🤦

pkg/utils/scheduling/scheduling.go
```diff
@@ -60,6 +60,13 @@ func IsSchedulable(pod *v1.PodSpec, node *v1.Node) bool {
 	return true
 }
 
+func IsForcefullyEvicted(pod *v1.Pod) bool {
```
ellistarn, line 63: nit on Forcefully. Is there a difference?

pkg/utils/scheduling/scheduling.go
njtran: Failed makes more sense here. Good point. [Failed containers phase](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle)
> All containers in the Pod have terminated, and at least one container has terminated in failure. That is, the container either exited with non-zero status or was terminated by the system.

njtran: Since the pod won't be coming back, it makes sense to not count any pod that fails here.

pkg/utils/node/predicates.go
```diff
@@ -45,6 +45,9 @@ func IsPastTTL(node *v1.Node) bool {
 // IsUnderutilized returns if the node has 0 non-daemonset pods
 func IsUnderutilized(node *v1.Node, pods []*v1.Pod) bool {
 	for _, pod := range pods {
+		if scheduling.HasFailed(pod) {
```
ellistarn, line 48: This reads misleadingly. scheduling has failed? Can we move this library. wdyt `podanalysis.HasFailed` `podanalysis.IsOwnedbyDaemonSet`. Maybe `podutils`? and just name the package pod, and alias it where its used?

ellistarn (dismissed):
Ooo looks quite good.

pkg/utils/scheduling/scheduling.go
```diff
@@ -17,26 +17,8 @@ package scheduling
...
-	for _, ignoredOwner := range IgnoredOwners {
-		for _, owner := range pod.ObjectMeta.OwnerReferences {
-			if owner.APIVersion == ignoredOwner.GroupVersion().String() && owner.Kind == ignoredOwner.Kind {
-				return true
-			}
-		}
-	}
-	return false
-}
-
```
ellistarn, line 39: Can we move these other helpers, too?

njtran: I wanted to keep them separate since one was scheduling and one was just about pod states.

ellistarn (approved):

ellistarn:
Somehow you got a nil pointer. Not sure, but structure LGTM otherwise.

PR #436, opened 2021-06-04
Controller Watch Skeleton Code

njtran:
Issue #, if available:

Description of changes:
Adds in the function Watches() to the `GenericController` interface, and uses builder to register the controllers with the Watches.


charts/karpenter/templates/rbac.yaml
```diff
@@ -43,6 +43,13 @@ metadata:
   name: karpenter-leader-election
   namespace: {{ .Release.Namespace }}
 rules:
+- apiGroups:
+  - "coordination.k8s.io"
```
ellistarn, line 47: why is this change necessary?

pkg/controllers/controller.go
```diff
@@ -46,7 +49,7 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 	// 2. Copy object for merge patch base
 	persisted := resource.DeepCopyObject()
 	// 3. Reconcile
-	if err := c.Controller.Reconcile(ctx, resource); err != nil {
+	if _, err := c.Controller.Reconcile(ctx, resource); err != nil {
```
ellistarn, line 52: Assuming we're not doing anything with this result for now?

pkg/controllers/manager.go
```diff
@@ -75,6 +75,7 @@ func (m *GenericControllerManager) RegisterControllers(controllers ...Controller
 		for _, resource := range c.Owns() {
 			builder = builder.Owns(resource)
 		}
+		builder = builder.Watches(c.Watches(context.Background()))
```
ellistarn, line 78: inline this above? 

.For(...)
.Watches(...)
.WithOptions(...)

pkg/controllers/provisioning/v1alpha1/allocation/controller.go
```diff
@@ -110,5 +117,11 @@ func (c *Controller) Reconcile(ctx context.Context, object controllers.Object) e
...
 		}
 	}
-	return nil
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) Watches(context.Context) (source.Source, handler.EventHandler, builder.WatchesOption) {
+	return &source.Kind{Type: &v1.Pod{}},
+		&handler.EnqueueRequestForObject{},
+		builder.WithPredicates(predicate.NewPredicateFuncs(func(object client.Object) bool { return false }))
```
ellistarn, line 126: can you do builder.WithPredicates()?

njtran: With an empty `builder.WithPredicates()` we would enqueue a `reconcile.Request` for every event on a Pod here. Since we don't have watch properly implemented right now, this essentially says to not enqueue requests at all, where the `GenericController.Reconcile()` function requeues a request every 5 seconds as currently implemented.

charts/karpenter/templates/rbac.yaml
njtran: I added this as a response to an error locally where I accidentally installed Karpenter to the default namespace, my bad.

pkg/controllers/provisioning/v1alpha1/allocation/controller.go
ellistarn: Sounds good.

ellistarn (approved):

PR #439, opened 2021-06-08
Split termination

njtran:
Issue #, if available:

Description of changes:
- Refactored the Object interface we use
- Separated out termination logic into its own controller
- Removed dependency on the provisioner for the termination logic to be executed


cmd/controller/main.go
```diff
@@ -8,14 +8,14 @@ import (
 	"github.com/awslabs/karpenter/pkg/cloudprovider"
 	"github.com/awslabs/karpenter/pkg/cloudprovider/registry"
 	"github.com/awslabs/karpenter/pkg/controllers"
-	"github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation"
-
 	"github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/allocation"
+	"github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation"
+	termination "github.com/awslabs/karpenter/pkg/controllers/termination/v1alpha1"
```
JacobGabrielson, line 13: I'm wondering if we should call this `terminating` to match `provisioning`?

pkg/cloudprovider/aws/factory.go
```diff
@@ -92,6 +92,12 @@ func (f *Factory) CapacityFor(provisioner *v1alpha1.Provisioner) cloudprovider.C
 	}
 }
 
+func (f *Factory) TerminateFor() cloudprovider.Termination {
+	return &Termination{
```
JacobGabrielson, line 96: I think this should potentially be called `Terminator` instead of `Termination`? To match `Provisioner`, etc.

cmd/controller/main.go
ellistarn, line 13: Consider simply not using v1alpha1 as the controller's package name. We're not running multiple api versions (yet). I think it's unlikely we'll ever support different controllers simultaneously.

pkg/cloudprovider/aws/factory.go
```diff
@@ -92,6 +92,12 @@ func (f *Factory) CapacityFor(provisioner *v1alpha1.Provisioner) cloudprovider.C
 	}
 }
 
+func (f *Factory) TerminateFor() cloudprovider.Termination {
```
ellistarn, line 95: Why separate this to a different interface? In my mind, it's all the capacity interface on the cloud provider.

pkg/cloudprovider/aws/factory.go
ellistarn, line 96: Consider not even wrapping this struct. InstanceProvider already satisfies the terminate interface.

pkg/controllers/provisioning/v1alpha1/allocation/controller.go
```diff
@@ -75,8 +74,19 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 }
 
 // Reconcile executes an allocation control loop for the resource
-func (c *Controller) Reconcile(ctx context.Context, object controllers.Object) (reconcile.Result, error) {
+func (c *Controller) Reconcile(ctx context.Context, object client.Object) (reconcile.Result, error) {
 	provisioner := object.(*v1alpha1.Provisioner)
+
+	// If controller fails, reflect in ProvisionerStatus
```
ellistarn, line 80: comment nit: this isn't reflection.

pkg/cloudprovider/aws/factory.go
njtran: We don't need the provisioner to terminate nodes, so I separated it out into its own Factory implementation.

pkg/controllers/provisioning/v1alpha1/allocation/controller.go
```diff
@@ -46,13 +45,13 @@ type Controller struct {
 }
 
 // For returns the resource this controller is for.
-func (c *Controller) For() controllers.Object {
```
ellistarn, line 49: did we delete the definition of controllers.Object now that it's not used?

cmd/controller/main.go
njtran: I'd like to keep the pattern here the same until we have the ability to run multiple api versions like you mention.

pkg/cloudprovider/aws/factory.go
njtran: I think Termination fits here as the counterpart is Capacity, rather than Provisioner.

pkg/controllers/termination/v1alpha1/controller.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// For returns the resource this controller is for.
+func (c *Controller) For() client.Object {
+	return &v1.Node{}
+}
+
+// Owns returns the resources owned by this controller's resource.
+func (c *Controller) Owns() []client.Object {
+	return []client.Object{}
```
ellistarn, line 47: I think you should be able to return nil here?

pkg/controllers/termination/v1alpha1/controller.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (c *Controller) Interval() time.Duration {
```
ellistarn, line 50: Didn't we remove this interval concept?

pkg/cloudprovider/aws/factory.go
njtran: Cool! Didn't know you could do that.

pkg/controllers/provisioning/v1alpha1/allocation/controller.go
njtran: Yes.

pkg/controllers/termination/v1alpha1/controller.go
njtran: Not yet. We haven't implemented watches yet. This will be removed in the next PR.

pkg/controllers/termination/v1alpha1/terminate.go
```diff
@@ -139,13 +123,15 @@ func (t *Terminator) deleteNodes(ctx context.Context, nodes []*v1.Node, provisio
...
+func (t *Terminator) getLabeledNodes(ctx context.Context, phaseLabel string) ([]*v1.Node, error) {
 	nodes := &v1.NodeList{}
-	if err := t.kubeClient.List(ctx, nodes, client.MatchingLabels(functional.UnionStringMaps(map[string]string{
-		v1alpha1.ProvisionerNameLabelKey:      provisioner.Name,
-		v1alpha1.ProvisionerNamespaceLabelKey: provisioner.Namespace,
-	}, additionalLabels))); err != nil {
+	if err := t.kubeClient.List(ctx, nodes, client.HasLabels([]string{
+		provisioning.ProvisionerNameLabelKey,
+		provisioning.ProvisionerNamespaceLabelKey,
+	}), client.MatchingLabels(map[string]string{
```
ellistarn, line 132: I'm not sure I'm grokking client.HasLabels and client.MatchingLabels. Why did this change? What's the logic here?

pkg/controllers/termination/v1alpha1/suite_test.go
```diff
@@ -0,0 +1,97 @@
...
+		// Create Provisioner to give some time for the node to reconcile
+		provisioner = &v1alpha1.Provisioner{
+			ObjectMeta: metav1.ObjectMeta{Name: strings.ToLower(randomdata.SillyName()),
+				Namespace: "default",
+			},
+			Spec: v1alpha1.ProvisionerSpec{
+				Cluster: &v1alpha1.ClusterSpec{Name: "test-cluster", Endpoint: "http://test-cluster", CABundle: "dGVzdC1jbHVzdGVyCg=="},
+			},
+		}
+		ctx = context.Background()
```
ellistarn, line 70: nit: consider moving this above provisioner to reduce eye jumping when reading code (i.e. you can read provisioner and ctx right next to each other, and know that provisioner is the last variable).

pkg/cloudprovider/aws/factory.go
ellistarn: I see the challenge here, but we need a cleaner solution to this.

pkg/controllers/termination/v1alpha1/terminate.go
```diff
@@ -139,13 +123,15 @@ func (t *Terminator) deleteNodes(ctx context.Context, nodes []*v1.Node, provisio
 	return nil
 }
 
-// getNodes returns a list of nodes with the provisioner's labels and given labels
-func (t *Terminator) getNodes(ctx context.Context, provisioner *v1alpha1.Provisioner, additionalLabels map[string]string) ([]*v1.Node, error) {
+// getLabeledNodes returns a list of nodes with the provisioner's labels and given labels
+func (t *Terminator) getLabeledNodes(ctx context.Context, phaseLabel string) ([]*v1.Node, error) {
```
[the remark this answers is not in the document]
njtran: When we implement watches, we will reconcile nodes themselves. This means we won't need to query nodes, and this function will be deleted as part of that change.  I'll add the enum for this.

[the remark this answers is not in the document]
njtran: We define the phases in `provisioner.go` and use it to patch label maps of nodes. This would ultimately lead to a decent amount of string casting, which I think would be a redundant use of enum here. WDYT?

pkg/cloudprovider/types.go
```diff
@@ -28,14 +28,14 @@ import (
 type Factory interface {
 	// Capacity returns a provisioner for the provider to create instances
 	CapacityFor(provisioner *v1alpha1.Provisioner) Capacity
+	// Terminate allows the provider to terminate instances
+	TerminateFor() Termination
```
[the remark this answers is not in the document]
njtran: Got it. I totally agree. I think it makes sense to punt on this and clean it up later.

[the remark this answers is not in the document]
ellistarn: I'm happy to punt on this for now. Please cut an issue.

pkg/controllers/termination/v1alpha1/controller.go
ellistarn: Great.

cmd/controller/main.go
njtran: I'll cut an issue for that to track it.

pkg/controllers/termination/v1alpha1/terminate.go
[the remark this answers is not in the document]
ellistarn: What if the strings were themselves already the right type where they were defined :). I'm fine just using raw types though, probably cleaner overall -- I haven't played with it myself.

pkg/controllers/termination/v1alpha1/terminate.go
ellistarn: Ah gotcha. This is kind of a cool trick. It'll get deleted, but a good tool to have in your back pocket.

pkg/controllers/termination/v1alpha1/terminate.go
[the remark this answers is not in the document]
njtran: It's a little tricky when we use the labels in `utilization.go` as the label maps are map[string]string, and not passed in as a parameter. I like the idea, and think that either way would be the same net amount of code. I'll keep it the way as is, as we are changing it in the next PR.

pkg/controllers/provisioning/v1alpha1/reallocation/controller.go
```diff
@@ -62,19 +60,37 @@ func (c *Controller) Name() string {
...
 
 // Reconcile executes a reallocation control loop for the resource
-func (c *Controller) Reconcile(ctx context.Context, object controllers.Object) (reconcile.Result, error) {
+func (c *Controller) Reconcile(ctx context.Context, object client.Object) (reconcile.Result, error) {
 	provisioner := object.(*v1alpha1.Provisioner)
-	if err := c.utilization.Reconcile(ctx, provisioner); err != nil {
-		return reconcile.Result{}, fmt.Errorf("reconciling utilization sub-controller, %w", err)
+
+	// Mark ProvisionerStatus if Reconcile fails
+	var err error
```
ellistarn, line 72: I think we should remove this in this controller. Right now there's actually a race condition. If allocator writes true and reallocator writes false, it will flip flop.

njtran: I changed it so that the Provisioner would be set to StatusCondition true while we figure out what to do with this race condition. I'll cut an issue as well.

PR #446, opened 2021-06-10
Trigger termination logic on watch events

njtran:
Issue #, if available:

Description of changes:
- Termination controller reacts to all node events, and reconciles on a per-node basis, rather than querying `listNodes` for every reconcile.
- GenericController can work with controllers triggered on an interval or by watch events.
- Added status patching to properly work based on the result of the reconcile loop.

### DIFF
--- a/charts/karpenter/templates/controller/rbac.yaml
+++ b/charts/karpenter/templates/controller/rbac.yaml
@@ -91,7 +91,6 @@ rules:
   resources:
   - nodes
   - pods
-  - nodes/status
   verbs:
   - get
   - list
--- a/charts/karpenter/templates/controller/rbac.yaml
+++ b/charts/karpenter/templates/controller/rbac.yaml
@@ -91,6 +91,7 @@ rules:
   resources:
   - nodes
   - pods
+  - nodes/status
   verbs:
   - get
   - list
--- a/pkg/controllers/controller.go
+++ b/pkg/controllers/controller.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/awslabs/karpenter/pkg/utils/conditions"
 	"go.uber.org/zap"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"knative.dev/pkg/apis"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -52,8 +53,7 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 	// 4. Reconcile
 	result, err := c.Controller.Reconcile(ctx, resource)
 	if err != nil {
-		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
-			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
+		zap.S().Errorf("Controller failed to reconcile kind %s, %s", resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
 	}
 	// 5. Set status based on results of reconcile
 	if conditionsAccessor, ok := resource.(apis.ConditionsAccessor); ok {
@@ -65,8 +65,11 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 		}
 	}
 	// 6. Update Status using a merge patch
-	if err := c.Status().Patch(ctx, resource, client.MergeFrom(persisted)); err != nil {
-		return result, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
+	// If the controller is reconciling nodes, don't patch
+	if _, ok := resource.(*v1.Node); !ok {
+		if err := c.Status().Patch(ctx, resource, client.MergeFrom(persisted)); err != nil {
+			return result, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
+		}
 	}
-	return result, nil
+	return result, err
 }
--- a/pkg/controllers/controller.go
+++ b/pkg/controllers/controller.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 
 	"github.com/awslabs/karpenter/pkg/utils/conditions"
-
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"knative.dev/pkg/apis"
@@ -42,10 +41,10 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 	// 1. Read Spec
 	resource := c.For()
 	if err := c.Get(ctx, req.NamespacedName, resource); err != nil {
-		if errors.IsNotFound(err) {
+		if errors.IsNotFound(err) || errors.IsGone(err) {
 			return reconcile.Result{}, nil
 		}
-		return reconcile.Result{}, err
+		return reconcile.Result{Requeue: true}, err
 	}
 	// 2. Copy object for merge patch base
 	persisted := resource.DeepCopyObject()
@@ -58,23 +57,25 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 	if conditionsAccessor, ok := resource.(apis.ConditionsAccessor); ok {
 		apis.NewLivingConditionSet(conditions.Active).Manage(conditionsAccessor).MarkTrue(conditions.Active)
 	}
-	// 5. Reconcile
-	if _, err := c.Controller.Reconcile(ctx, resource); err != nil {
+
+	// 4. Reconcile
+	result, err := c.Controller.Reconcile(ctx, resource)
+	if err != nil {
 		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
 			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
-		return reconcile.Result{Requeue: true}, nil
+		return result, err
 	}
 	// 6. Update Status using a merge patch
 	if err := c.Status().Patch(ctx, resource, client.MergeFrom(persisted)); err != nil {
-		return reconcile.Result{}, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
+		return result, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
 	}
-	return reconcile.Result{RequeueAfter: c.Interval()}, nil
+	return result, nil
 }
 
 // WatchDescription returns the necessary information to create a watch
 //   a. source: the resource that is being watched
 //   b. eventHandler: which controller objects to be reconciled
 //   c. predicates: which events can be filtered out before processed
-func (c *GenericController) Watches(ctx context.Context) (source.Source, handler.EventHandler, builder.WatchesOption) {
-	return c.Controller.Watches(ctx)
+func (c *GenericController) Watches() (source.Source, handler.EventHandler, builder.WatchesOption) {
+	return c.Controller.Watches()
 }
--- a/pkg/controllers/controller.go
+++ b/pkg/controllers/controller.go
@@ -22,11 +22,8 @@ import (
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"knative.dev/pkg/apis"
-	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // GenericController implements controllerruntime.Reconciler and runs a
@@ -52,29 +49,24 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 	if _, ok := resource.(apis.Defaultable); ok {
 		resource.(apis.Defaultable).SetDefaults(ctx)
 	}
-	// 4. Set to true to remove race condition where multiple controllers set the status of the same object
-	// TODO: remove status conditions on provisioners
-	if conditionsAccessor, ok := resource.(apis.ConditionsAccessor); ok {
-		apis.NewLivingConditionSet(conditions.Active).Manage(conditionsAccessor).MarkTrue(conditions.Active)
-	}
-	// 5. Reconcile
+	// 4. Reconcile
 	result, err := c.Controller.Reconcile(ctx, resource)
 	if err != nil {
 		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
 			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
-		return result, err
+	}
+	// 5. Set status based on results of reconcile
+	if conditionsAccessor, ok := resource.(apis.ConditionsAccessor); ok {
+		m := apis.NewLivingConditionSet(conditions.Active).Manage(conditionsAccessor)
+		if err != nil {
+			m.MarkFalse(conditions.Active, err.Error(), "")
+		} else {
+			m.MarkTrue(conditions.Active)
+		}
 	}
 	// 6. Update Status using a merge patch
 	if err := c.Status().Patch(ctx, resource, client.MergeFrom(persisted)); err != nil {
 		return result, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
 	}
 	return result, nil
-}
-
-// WatchDescription returns the necessary information to create a watch
-//   a. source: the resource that is being watched
-//   b. eventHandler: which controller objects to be reconciled
-//   c. predicates: which events can be filtered out before processed
-func (c *GenericController) Watches() (source.Source, handler.EventHandler, builder.WatchesOption) {
-	return c.Controller.Watches()
 }
--- a/pkg/controllers/controller.go
+++ b/pkg/controllers/controller.go
@@ -41,10 +41,10 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 	// 1. Read Spec
 	resource := c.For()
 	if err := c.Get(ctx, req.NamespacedName, resource); err != nil {
-		if errors.IsNotFound(err) || errors.IsGone(err) {
+		if errors.IsNotFound(err) {
 			return reconcile.Result{}, nil
 		}
-		return reconcile.Result{Requeue: true}, err
+		return reconcile.Result{}, err
 	}
 	// 2. Copy object for merge patch base
 	persisted := resource.DeepCopyObject()
@@ -57,8 +57,7 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 	if conditionsAccessor, ok := resource.(apis.ConditionsAccessor); ok {
 		apis.NewLivingConditionSet(conditions.Active).Manage(conditionsAccessor).MarkTrue(conditions.Active)
 	}
-
-	// 4. Reconcile
+	// 5. Reconcile
 	result, err := c.Controller.Reconcile(ctx, resource)
 	if err != nil {
 		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
--- a/pkg/controllers/provisioning/v1alpha1/allocation/controller.go
+++ b/pkg/controllers/provisioning/v1alpha1/allocation/controller.go
@@ -25,14 +25,9 @@ import (
 	"github.com/awslabs/karpenter/pkg/utils/apiobject"
 
 	"go.uber.org/zap"
-	v1 "k8s.io/api/core/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
-	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // Controller for the resource
@@ -47,11 +42,6 @@ type Controller struct {
 // For returns the resource this controller is for.
 func (c *Controller) For() client.Object {
 	return &v1alpha1.Provisioner{}
-}
-
-// Owns returns the resources owned by this controller's resource.
-func (c *Controller) Owns() []client.Object {
-	return nil
 }
 
 func (c *Controller) Interval() time.Duration {
@@ -116,10 +106,4 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
 		}
 	}
 	return reconcile.Result{RequeueAfter: c.Interval()}, nil
-}
-
-func (c *Controller) Watches() (source.Source, handler.EventHandler, builder.WatchesOption) {
-	return &source.Kind{Type: &v1.Pod{}},
-		&handler.EnqueueRequestForObject{},
-		builder.WithPredicates(predicate.NewPredicateFuncs(func(object client.Object) bool { return false }))
 }
--- a/pkg/controllers/provisioning/v1alpha1/allocation/controller.go
+++ b/pkg/controllers/provisioning/v1alpha1/allocation/controller.go
@@ -82,7 +82,7 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
 		return reconcile.Result{}, fmt.Errorf("filtering pods, %w", err)
 	}
 	if len(pods) == 0 {
-		return reconcile.Result{}, nil
+		return reconcile.Result{RequeueAfter: c.Interval()}, nil
 	}
 	zap.S().Infof("Found %d provisionable pods", len(pods))
 
@@ -115,10 +115,10 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
 			zap.S().Errorf("Continuing after failing to bind, %s", err.Error())
 		}
 	}
-	return reconcile.Result{}, nil
+	return reconcile.Result{RequeueAfter: c.Interval()}, nil
 }
 
-func (c *Controller) Watches(context.Context) (source.Source, handler.EventHandler, builder.WatchesOption) {
+func (c *Controller) Watches() (source.Source, handler.EventHandler, builder.WatchesOption) {
 	return &source.Kind{Type: &v1.Pod{}},
 		&handler.EnqueueRequestForObject{},
 		builder.WithPredicates(predicate.NewPredicateFuncs(func(object client.Object) bool { return false }))
--- a/pkg/controllers/provisioning/v1alpha1/allocation/controller.go
+++ b/pkg/controllers/provisioning/v1alpha1/allocation/controller.go
@@ -108,7 +108,7 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
 		return reconcile.Result{}, fmt.Errorf("creating capacity, %w", err)
 	}
 
-	// 4. Bind pods to nodes
+	// 5. Bind pods to nodes
 	for _, packedNode := range packedNodes {
 		zap.S().Infof("Binding pods %v to node %s", apiobject.PodNamespacedNames(packedNode.Pods), packedNode.Node.Name)
 		if err := c.binder.Bind(ctx, packedNode.Node, packedNode.Pods); err != nil {
--- a/pkg/controllers/terminating/v1alpha1/controller.go
+++ b/pkg/controllers/terminating/v1alpha1/controller.go
@@ -25,13 +25,8 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
-	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/event"
-	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // Controller for the resource
@@ -43,11 +38,6 @@ type Controller struct {
 // For returns the resource this controller is for.
 func (c *Controller) For() client.Object {
 	return &v1.Node{}
-}
-
-// Owns returns the resources owned by this controller's resource.
-func (c *Controller) Owns() []client.Object {
-	return nil
 }
 
 func (c *Controller) Interval() time.Duration {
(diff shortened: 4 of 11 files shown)


pkg/cloudprovider/aws/factory.go
```diff
@@ -16,18 +16,19 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha1"
```
ellistarn, line 19: Why..?

pkg/cloudprovider/aws/instance.go
```diff
@@ -119,31 +119,26 @@ func (p *InstanceProvider) Create(ctx context.Context,
 	return createFleetOutput.Instances[0].InstanceIds[0], nil
 }
 
-func (p *InstanceProvider) Terminate(ctx context.Context, nodes []*v1.Node) error {
-	if len(nodes) == 0 {
-		return nil
+func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
```
ellistarn, line 122: What is EC2's delete instance default quota? We're moving from an n -> 1 model for deletion.

pkg/controllers/controller.go
```diff
@@ -44,7 +45,7 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 		if errors.IsNotFound(err) {
 			return reconcile.Result{}, nil
 		}
-		return reconcile.Result{}, err
+		return reconcile.Result{Requeue: true}, err
```
ellistarn, line 48: If it's not found, why requeue?

pkg/controllers/controller.go
```diff
@@ -54,22 +55,26 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
...
+	if result, err := c.Controller.Reconcile(ctx, resource); err != nil {
 		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
 			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
-		return reconcile.Result{Requeue: true}, nil
+		return result, err
 	}
 	// 5. Update Status using a merge patch
 	if err := c.Status().Patch(ctx, resource, client.MergeFrom(persisted)); err != nil {
-		return reconcile.Result{}, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
+		return reconcile.Result{Requeue: true}, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
```
ellistarn, line 65: nice

pkg/controllers/controller.go
```diff
@@ -54,22 +55,26 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
...
 	}
-	return reconcile.Result{RequeueAfter: c.Interval()}, nil
+	// 6. Requeue controllers on an interval
+	if controllerType := reflect.TypeOf(c.Controller).String(); controllerType != "*termination.Controller" {
```
ellistarn, line 68: Instead of this reflection, can we make the allocator controller simply return reconcile.Result{RequeueAfter: c.Interval()} and then this top level can just return the result directly?

pkg/controllers/controller.go
```diff
@@ -54,22 +55,26 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 		provisioner.StatusConditions().MarkTrue(v1alpha1.Active)
 	}
 	// 4. Reconcile
-	if _, err := c.Controller.Reconcile(ctx, resource); err != nil {
+	if result, err := c.Controller.Reconcile(ctx, resource); err != nil {
```
ellistarn, line 58: We need to return this result in this non error case, too

pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go
```diff
@@ -32,24 +32,6 @@ type Utilization struct {
 	kubeClient client.Client
 }
 
-func (u *Utilization) Reconcile(ctx context.Context, provisioner *v1alpha1.Provisioner) error {
```
ellistarn, line 35: Where'd this logic go? Is this implemented somewhere else not in this PR?

pkg/controllers/terminating/v1alpha1/controller.go
```diff
@@ -63,22 +65,68 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
...
+		}
 	}
-	return reconcile.Result{}, nil
+	// 4. Requeue if we still need to drain
+	return reconcile.Result{Requeue: !drained}, nil
 }
 
-func (c *Controller) Watches(context.Context) (source.Source, handler.EventHandler, builder.WatchesOption) {
+func (c *Controller) Watches() (source.Source, handler.EventHandler, builder.WatchesOption) {
 	return &source.Kind{Type: &v1.Node{}},
```
ellistarn, line 99: I'm confused, doesn't the termination controller need to trigger on pod events?

pkg/controllers/terminating/v1alpha1/controller.go
```diff
@@ -63,22 +65,68 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
...
 		&handler.EnqueueRequestForObject{},
-		builder.WithPredicates(predicate.NewPredicateFuncs(func(object client.Object) bool { return false }))
+		builder.WithPredicates(
+			predicate.Funcs{
+				CreateFunc: func(e event.CreateEvent) bool {
+					return false
+				},
+				// kubectl delete node is not supported yet
```
ellistarn, line 106: Will this comment ever be accurate? Does setting deletion timestamp trigger a delete event?

pkg/controllers/terminating/v1alpha1/controller.go
```diff
@@ -63,22 +65,68 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
...
+				DeleteFunc: func(e event.DeleteEvent) bool {
+					return false
+				},
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					nodeLabels := e.ObjectNew.GetLabels()
```
ellistarn, line 111: consider just `labels`

pkg/controllers/terminating/v1alpha1/controller.go
```diff
@@ -63,22 +65,68 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
...
+					// Don't watch nodes not created by provisioners
+					if _, ok := nodeLabels[provisioning.ProvisionerNameLabelKey]; !ok {
+						return false
+					}
+					if _, ok := nodeLabels[provisioning.ProvisionerNamespaceLabelKey]; !ok {
```
ellistarn, line 116: We can skip this check

pkg/controllers/terminating/v1alpha1/controller.go
```diff
@@ -63,22 +65,68 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
...
+						return false
+					}
+					if phase, ok := nodeLabels[provisioning.ProvisionerPhaseLabel]; ok {
+						return (phase == provisioning.ProvisionerTerminablePhase ||
+							phase == provisioning.ProvisionerDrainingPhase ||
+							phase == provisioning.ProvisionerDrainedPhase) &&
+							e.ObjectOld.GetResourceVersion() != e.ObjectNew.GetResourceVersion()
```
ellistarn, line 123: Why do we need to check resource versions?

pkg/controllers/terminating/v1alpha1/suite_test.go
```diff
@@ -75,21 +75,24 @@ var _ = Describe("Reallocation", func() {
...
 				Annotations: map[string]string{
 					v1alpha1.ProvisionerTTLKey: time.Now().Add(time.Duration(-100) * time.Second).Format(time.RFC3339),
 				},
 			})
 			ExpectCreatedWithStatus(env.Client, node)
 
-			ExpectCreated(env.Client, provisioner)
+			persisted := node.DeepCopy()
+			node.Labels[v1alpha1.ProvisionerPhaseLabel] = v1alpha1.ProvisionerDrainedPhase
+			Eventually(Expect(env.Client.Patch(ctx, node, client.MergeFrom(persisted))).To(Succeed()))
```
ellistarn, line 92: Why is this an eventually?

pkg/controllers/terminating/v1alpha1/suite_test.go
```diff
@@ -75,21 +75,24 @@ var _ = Describe("Reallocation", func() {
...
 
+			// Let some time go by
```
ellistarn, line 94: Is this comment accurate?

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,110 +37,72 @@ type Terminator struct {
...
 }
 
-// cordonNodes takes in a list of expired nodes as input and cordons them
-func (t *Terminator) cordonNodes(ctx context.Context) error {
-	// 1. Get terminable nodes
-	nodeList, err := t.getLabeledNodes(ctx, provisioning.ProvisionerTerminablePhase)
-	if err != nil {
-		return err
+// cordonNode takes in a list of expired nodes as input and cordons them
+func (t *Terminator) cordonNode(ctx context.Context, node *v1.Node) error {
```
ellistarn, line 41: Just musing, can we name this folder the "termination controller" instead of the "terminating controller".

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,110 +37,72 @@ type Terminator struct {
...
+	if node.Spec.Unschedulable {
```
ellistarn, line 42: Do we need this check? The idempotency should be handled by the kubectl patch logic (if its the same, it won't make the call.)

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,110 +37,72 @@ type Terminator struct {
...
-			node.Labels,
-			map[string]string{provisioning.ProvisionerPhaseLabel: provisioning.ProvisionerDrainingPhase},
-		)
-		if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
-			return fmt.Errorf("patching node %s, %w", node.Name, err)
-		}
-		zap.S().Debugf("Cordoned node %s", node.Name)
+	persisted := node.DeepCopy()
+	node.Spec.Unschedulable = true
+	node.Labels = functional.UnionStringMaps(
```
ellistarn, line 47: How about

`node.Labels[provisioning.ProvisionerPhaseLabel] = provisioning.ProvisionerDrainingPhase`

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,110 +37,72 @@ type Terminator struct {
...
-					zap.S().Debugf("Continuing after failing to evict pods from node %s, %s", node.Name, err.Error())
-				}
+	// non-daemonset pods
+	empty := true
+	// 2. Evict pods on node
+	for _, p := range pods {
+		if !pod.IsOwnedByDaemonSet(p) {
+			empty = false
+			if err := t.coreV1Client.Pods(p.Namespace).Evict(ctx, &v1beta1.Eviction{
+				ObjectMeta: metav1.ObjectMeta{
```
ellistarn, line 75: IMO you can cleanly inline all of this.

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,110 +37,72 @@ type Terminator struct {
...
-	}
-	return nil
+	return empty, nil
 }
 
 // deleteNode uses a cloudprovider-specific delete to delete a set of nodes
-func (t *Terminator) deleteNodes(ctx context.Context, nodes []*v1.Node) error {
-	// 1. Delete node in cloudprovider's instanceprovider
-	if err := t.cloudprovider.TerminateFor().Terminate(ctx, nodes); err != nil {
+func (t *Terminator) terminateNode(ctx context.Context, node *v1.Node) error {
```
ellistarn, line 87: all of these methods are scoped to the node concept, so consider

terminateNode(ctx,node) -> terminate(ctx,node)
cordonNode -> cordon
drainNode -> drain

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,110 +37,72 @@ type Terminator struct {
...
-	for _, node := range nodes {
-		if err := t.kubeClient.Delete(ctx, node); err != nil {
-			zap.S().Debugf("Continuing after failing to delete node %s, %s", node.Name, err.Error())
-		}
-		zap.S().Infof("Terminated node %s", node.Name)
+	if err := t.kubeClient.Delete(ctx, node); err != nil {
+		// Don't return error as Node Controller will clean this node up without underlying instance
+		zap.S().Debugf("Continuing after failing to delete node %s, %s", node.Name, err.Error())
 	}
+	zap.S().Infof("Terminated node %s", node.Name)
```
ellistarn, line 97: This logline will conflict with :95

pkg/cloudprovider/aws/factory.go
njtran: I wanted to separate 3rd party imports vs project imports. I'm fine with having it either way. I can change it back if you'd like.

pkg/controllers/controller.go
njtran: If it's not found, we won't requeue here. If we fail to get it due to something like an APIServer connection, we do want to requeue, hence putting this outside of the if statement.

pkg/controllers/controller.go
njtran: You're totally right. Good catch. If we want to continue draining here, it won't propagate properly. Thanks!

pkg/controllers/controller.go
njtran: Good call here. I like this better, and generalizes better.

pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go
njtran: I accidentally left this replicated in the previous PR. I removed this as it wasn't being used.

pkg/controllers/terminating/v1alpha1/controller.go
njtran: You're right. Doing `kubectl delete node` with a finalizer will end up with an update event, as the finalizer will set the deletion timestamp. I'll change this comment so it's a little easier to understand, as those who look at this for the first time will think deleteFunc should return true.

pkg/cloudprovider/aws/factory.go
ellistarn: We should be relying on a code formatter to make these decisions. Let's make sure we're not doing it manually, and that were using the same one.

pkg/controllers/controller.go
ellistarn: Oh I see. Nice catch!!!

ellistarn: That said, I think controller runtime automatically causes a requeue if you return an err

pkg/controllers/terminating/v1alpha1/controller.go
njtran: Why? Provisioners need a name and namespace. We only want to do work on nodes that were created by Provisioners, meaning they have to have both a name and namespace label.

pkg/controllers/terminating/v1alpha1/controller.go
njtran: We do this check to make sure we don't take action on a cache resync event. When we resync the watch client cache, we send an update, but the resource versions will be the same, as we are not changing the object itself.

https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/client-go/tools/cache/shared_informer.go#L152

pkg/controllers/terminating/v1alpha1/suite_test.go
njtran: I wanted to do Eventually here to make sure we don't time out if something goes wrong with the environment.

pkg/controllers/terminating/v1alpha1/suite_test.go
njtran: Creating the provisioner gives us a little more leeway to reduce this race condition. It's a little hacky, but will change when we have more control over the termination process, for instance `do-not-evict` pods.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: I named it terminating per @JacobGabrielson's comment in the last PR. The other package is named `provisioning` so I believe it's proper here. Do you feel strongly otherwise?

pkg/controllers/terminating/v1alpha1/suite_test.go
ellistarn: By that logic we should wrap the entire world in eventually until nothing ever happens :D

pkg/controllers/terminating/v1alpha1/suite_test.go
ellistarn: Let's talk.

pkg/controllers/terminating/v1alpha1/terminate.go
ellistarn: Ah I didn't notice this parity.

ellistarn: In that case, can we make it termination.go (since it's the "Termination Controller")

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: The Patch still makes a call here. When we make a patch, we need to make sure that the resource did not change while we were doing this logic. If the resource did change while we were doing this function, we can get a 409.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: That does the trick. Thanks.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: Good catch. I had this when we were deleting multiple nodes. Removing this.

pkg/controllers/controller.go
njtran: Yes. That's true, but I want to put it here so it's clearer when we re-queue or not to the reader.

pkg/controllers/terminating/v1alpha1/suite_test.go
njtran: Very fair. Thanks.

pkg/controllers/terminating/v1alpha1/controller.go
ellistarn: The point of the resync interval is to trigger the reconciliation logic in case something has accidentally gotten stale.

pkg/test/expectations/expectations.go
```diff
@@ -62,6 +63,15 @@ func ExpectCreatedWithStatus(c client.Client, objects ...client.Object) {
 	}
 }
 
+func ExpectEventuallyDeleted(c client.Client, object client.Object) {
+	Eventually(func() bool {
+		updated := &v1.Node{}
+		return errors.IsNotFound(c.Get(context.Background(), client.ObjectKey{Name: object.GetName(), Namespace: object.GetNamespace()}, updated))
+	}, APIServerPropagationTime, RequestInterval).Should(BeTrue(), func() string {
```
ellistarn, line 70: Use ReconcilerPropagationTime

pkg/cloudprovider/aws/instance.go
njtran: Since this is in line with how we're doing watch events, I'd like to re-evaluate this when we get to scale testing.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: Sure.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: Chatted offline. Removing this as the optimization here is pretty small if even present.

pkg/controllers/controller.go
```diff
@@ -44,7 +43,7 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 		if errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
njtran: Great call out. Thanks!
https://kubernetes.io/docs/reference/using-api/api-concepts/#410-gone-responses

pkg/cloudprovider/aws/instance.go
bwagner5: FYI It's pretty high for `TerminateInstances` so probably fine but a good thing to watch for when scale testing:

1000 Max and 20 Refill rate

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits

pkg/apis/provisioning/v1alpha1/provisioner.go
```diff
@@ -116,6 +116,7 @@ const (
 	ProvisionerUnderutilizedPhase = "underutilized"
 	ProvisionerTerminablePhase    = "terminable"
 	ProvisionerDrainingPhase      = "draining"
+	ProvisionerDrainedPhase       = "drained"
```
ellistarn, line 119: It doesn't look like this phase is ever set or used

ellistarn (changes requested):

pkg/controllers/controller.go
```diff
@@ -41,10 +40,10 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 	// 1. Read Spec
 	resource := c.For()
 	if err := c.Get(ctx, req.NamespacedName, resource); err != nil {
-		if errors.IsNotFound(err) {
+		if errors.IsNotFound(err) || errors.IsGone(err) {
```
ellistarn, line 43: What could cause the resource to be "Gone". I'm not aware of kubernetes using this. I also haven't seen it in other operator implementations.

pkg/controllers/terminating/v1alpha1/termination.go
```diff
@@ -0,0 +1,97 @@
...
+	if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
+		return fmt.Errorf("patching node %s, %w", node.Name, err)
+	}
+	zap.S().Debugf("Cordoned node %s", node.Name)
+	return nil
+}
+
+// terminateNodes takes in a list of expired non-drained nodes and calls drain on them
+func (t *Terminator) drain(ctx context.Context, node *v1.Node) (bool, error) {
+	if phase, ok := node.Labels[provisioning.ProvisionerPhaseLabel]; ok && phase == provisioning.ProvisionerDrainedPhase {
```
ellistarn, line 53: You don't need both of these checks, if not ok, then phase will be "". You can just do `node.Labels[provisioning.ProvisionerPhaseLabel] == provisioning.ProvisionerDrainedPhase`

pkg/controllers/terminating/v1alpha1/termination.go
```diff
@@ -0,0 +1,97 @@
...
+	kubeClient    client.Client
+	cloudprovider cloudprovider.Factory
+	coreV1Client  corev1.CoreV1Interface
+}
+
+// cordon takes in a list of expired nodes as input and cordons them
+func (t *Terminator) cordon(ctx context.Context, node *v1.Node) error {
+	persisted := node.DeepCopy()
+	node.Spec.Unschedulable = true
+	node.Labels[provisioning.ProvisionerPhaseLabel] = provisioning.ProvisionerDrainingPhase
```
ellistarn, line 43: Why do you even need the draining phase, can't you just use the fact that the node is cordoned?

pkg/controllers/controller.go
njtran: Jacob recently commented about this. Here's the doc that convinced me to include this.
https://kubernetes.io/docs/reference/using-api/api-concepts/#410-gone-responses

pkg/controllers/terminating/v1alpha1/controller.go
```diff
@@ -63,22 +65,68 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
...
 	return &source.Kind{Type: &v1.Node{}},
 		&handler.EnqueueRequestForObject{},
-		builder.WithPredicates(predicate.NewPredicateFuncs(func(object client.Object) bool { return false }))
+		builder.WithPredicates(
+			predicate.Funcs{
+				CreateFunc: func(e event.CreateEvent) bool {
+					return false
+				},
+				// Deleting nodes directly is not supported.
+				// Our termination design will add a webhook to validate delete requests, and then send an update call.
```
ellistarn, line 107: You can probably drop both of these comments. Consider inlining all of these funcs as well
```
CreateFunc: func(e event.CreateEvent) bool { return false }, // no-op
DeleteFunc: func(e event.DeleteEvent) bool { return false }, // no-op
...
```

pkg/controllers/terminating/v1alpha1/termination.go
ellistarn: Also -- looking at the code, how does this work? Is ProvisionerDrainedPhase ever set? IMO, we shouldn't even need `ProvisionerDrainedPhase`

pkg/controllers/terminating/v1alpha1/termination.go
```diff
@@ -0,0 +1,97 @@
...
+	if err := t.kubeClient.Delete(ctx, node); err != nil {
+		// Don't return error as Node Controller will clean this node up without underlying instance
+		zap.S().Debugf("Continuing after failing to delete node object %s, %s", node.Name, err.Error())
+	}
+	zap.S().Infof("Terminated instance %s", node.Name)
+	return nil
+}
+
+// getPods returns a list of pods scheduled to a node
+func (t *Terminator) getPods(ctx context.Context, nodeName string) ([]*v1.Pod, error) {
```
ellistarn, line 91: Consider just removing this function entirely. You're looping through podList.Items anyways, so it may not be worth it to convert to a []*v1.Pod. This function doesn't actually remove any lines of code in the implementation above, you could do:
```
podList := &v1.PodList{}
if err := t.kubeClient.List(ctx, pods, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {		
    return nil, fmt.Errorf("listing pods on node %s, %w", nodeName, err)	
}
empty := true
for _, p := range podList.Items {
   ... 
}
```

pkg/controllers/provisioning/v1alpha1/allocation/controller.go
```diff
@@ -82,7 +82,7 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
 		return reconcile.Result{}, fmt.Errorf("filtering pods, %w", err)
 	}
 	if len(pods) == 0 {
-		return reconcile.Result{}, nil
+		return reconcile.Result{RequeueAfter: c.Interval()}, nil
```
JacobGabrielson, line 85: What happens if you don't pass in the Interval? I just see that repeated a few times and wonder if in future people are going to forget to add it in the right cases?

njtran: With this PR, we have two different types of controllers now, operating on strictly watch events, and strictly intervals. With this, we rely on each of the controllers to return their own result. For controllers operating on an interval, they need to return this in a reconcile.Result{} struct as you see above. 

For those forget to add an interval on a new controller, they won't see their controller working properly, so testing should fix this. It may be useful to add a note to help those who are looking at the code to understand it as well.

pkg/controllers/controller.go
JacobGabrielson: BTW, I hadn't done all my homework so it's possibly unnecessary. According to the docs you found it looks like maybe not?


> When using resourceVersionMatch=NotOlderThan and limit is set, clients must handle HTTP 410 "Gone" responses. For example, the client might retry with a newer resourceVersion or fall back to resourceVersion="".

(I'm not sure if NotOlderThan and limit is actually being passed anywhere under the covers?)

ellistarn: Necessarily we're using latest here, so Gone is not possible.

JacobGabrielson: Ok, seems worth reverting - apologies for any churn I may have caused here.

pkg/controllers/types.go
```diff
@@ -46,7 +46,7 @@ type Controller interface {
 	//   a. Source: the resource that is being watched
```
bwagner5, line 46: This comment should start with the func and be `Watches`

pkg/controllers/controller.go
```diff
@@ -59,22 +58,23 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 		apis.NewLivingConditionSet(conditions.Active).Manage(conditionsAccessor).MarkTrue(conditions.Active)
 	}
 	// 5. Reconcile
-	if _, err := c.Controller.Reconcile(ctx, resource); err != nil {
+	result, err := c.Controller.Reconcile(ctx, resource)
+	if err != nil {
 		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
 			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
-		return reconcile.Result{Requeue: true}, nil
+		return result, err
```
ellistarn, line 65: Don't we want to patch if we get an error?

pkg/controllers/types.go
```diff
@@ -46,7 +46,7 @@ type Controller interface {
...
 	//   b. EventHandler: which controller objects to be reconciled
 	//   c. WatchesOption: which events can be filtered out before processed
-	Watches(context.Context) (source.Source, handler.EventHandler, builder.WatchesOption)
```
ellistarn, line 49: We can remove the Interval() API here now as well, right?

pkg/controllers/controller.go
```diff
@@ -59,22 +58,23 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
...
 	}
 	// 6. Update Status using a merge patch
 	if err := c.Status().Patch(ctx, resource, client.MergeFrom(persisted)); err != nil {
-		return reconcile.Result{}, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
+		return result, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
 	}
-	return reconcile.Result{RequeueAfter: c.Interval()}, nil
```
ellistarn, line 71: Now that we don't call Interval() here, we can remove it from the interface.

njtran: Not quite. We did remove it from the GenericController code in order to allow different controllers to return the reconcile.Result that they desire. This allows us to requeue some controllers every 5 seconds, and some not.

pkg/controllers/types.go
njtran: I'm keeping the Interval() here as the allocation an reallocation controllers still rely on it.

charts/karpenter/templates/controller/rbac.yaml
```diff
@@ -91,6 +91,7 @@ rules:
   resources:
   - nodes
   - pods
+  - nodes/status
```
ellistarn, line 94: We shouldn't be patching nodes.

pkg/controllers/controller.go
```diff
@@ -53,28 +49,24 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
...
-	// TODO: remove status conditions on provisioners
-	if conditionsAccessor, ok := resource.(apis.ConditionsAccessor); ok {
-		apis.NewLivingConditionSet(conditions.Active).Manage(conditionsAccessor).MarkTrue(conditions.Active)
-	}
-	// 5. Reconcile
-	if _, err := c.Controller.Reconcile(ctx, resource); err != nil {
+	// 4. Reconcile
+	result, err := c.Controller.Reconcile(ctx, resource)
+	if err != nil {
 		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
```
ellistarn, line 55: oneline!

pkg/controllers/controller.go
ellistarn: Discussed offline

ellistarn (approved):

PR #447, opened 2021-06-10
Removed certmanager dependency in favor of knative pkg.

ellistarn:
Issue #, if available:

Description of changes:
- Removes dependency on CertManager
- Removes code generation of webhooks (manually maintained, now) 
- Webhooks are now in knative/pkg style
- Webhook logic is now static and does not rely on network requests
- Webhooks are now a separate pod. This helps with scalability (webhooks can be HA and don't require many resources) and is necessary since controllerruntime and knativepkg both define the --kubeconfig flag and are incompatible in the same binary.


charts/karpenter/templates/webhook/deployment.yaml
```diff
@@ -0,0 +1,80 @@
...
+    spec:
+      priorityClassName: system-cluster-critical
+      serviceAccountName: {{ .Values.serviceAccount.name }}
+      containers:
+        - name: webhook
+          image: {{ .Values.webhook.image }}
+          resources:
+            limits:
+              cpu: 100m
+              memory: 50Mi
```
JacobGabrielson, line 38: Wondering where does 50Mi come from?

ellistarn: I played with this and got OOMs until ~25Mi. I doubled it, since this webhook shouldn't be under much load at all (i.e. only when provisioners are created/deleted)

JacobGabrielson: Great, thanks

JacobGabrielson (dismissed):

JacobGabrielson (approved):

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -98,9 +95,11 @@ var _ = AfterSuite(func() {
 })
 
 var _ = Describe("Allocation", func() {
+	var ctx context.Context
```
bwagner5, line 98: `ctx := context.Background()`   instead of declaring and then initializing in BeforeEach?   It's a nil context anyways so it doesn't matter, unless I'm missing something?

pkg/controllers/manager.go
```diff
@@ -52,9 +51,12 @@ type GenericControllerManager struct {
 func NewManagerOrDie(config *rest.Config, options controllerruntime.Options) Manager {
 	options.Scheme = scheme
 	manager, err := controllerruntime.NewManager(config, options)
-	log.PanicIfError(err, "Failed to create controller manager")
-	log.PanicIfError(manager.GetFieldIndexer().
-		IndexField(context.Background(), &v1.Pod{}, "spec.nodeName", podSchedulingIndex), "Failed to setup pod indexer")
+	if err != nil {
+		panic(fmt.Sprintf("Failed to create controller manager, %s", err.Error()))
```
bwagner5, line 55: might be cleaner to use `fmt.Fatalf` here?

bwagner5 (commented):
Looks quite good! Very nice! Just a few comments and questions

charts/karpenter/templates/logging.yaml
```diff
@@ -0,0 +1,31 @@
...
+        "nameKey": "name",
+        "callerKey": "caller",
+        "messageKey": "message",
+        "stacktraceKey": "stacktrace",
+        "levelEncoder": "capital",
+        "timeEncoder": "iso8601",
+      }
+    }
+  # Log level overrides
+  loglevel.controller: "debug"
```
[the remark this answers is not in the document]
ellistarn: We're not actually using this logger in our controller (yet), so this doesn't apply.

This override is mostly a nice mechanism to kubectl patch configmap without having to do inline parsing of the zap-logger-config. This is all copied from knative setup, so I left this in.

cmd/controller/main.go
```diff
@@ -58,21 +54,18 @@ func main() {
...
-		Port:                   options.WebhookPort,
 		MetricsBindAddress:     fmt.Sprintf(":%d", options.MetricsPort),
 		HealthProbeBindAddress: fmt.Sprintf(":%d", options.HealthProbePort),
 	})
 
 	clientSet := kubernetes.NewForConfigOrDie(manager.GetConfig())
-	cloudProviderFactory := registry.NewFactory(cloudprovider.Options{Client: manager.GetClient(), ClientSet: clientSet})
+	cloudProvider := registry.NewCloudProvider(cloudprovider.Options{ClientSet: clientSet})
 
-	err := manager.RegisterWebhooks(
```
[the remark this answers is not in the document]
ellistarn: No, actually. The new validation logic registers supported properties on startup (checkout the Register() function), rather than by API call. Even if we did it via API calls, this validation logic only happens on creation/modification of provisioners, which is very rare.

docs/aws/README.md
```diff
@@ -130,7 +118,7 @@ kubectl logs -f -n karpenter $(kubectl get pods -n karpenter -l control-plane=ka
 
 ### Cleanup
 ```bash
-./hack/quick-install.sh --delete
+helm delete karpenter
```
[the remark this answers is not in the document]
ellistarn: oops nice catch!

pkg/cloudprovider/aws/factory.go
```diff
@@ -99,7 +98,9 @@ func (f *Factory) TerminateFor() cloudprovider.Termination {
 // get the current region from EC2 IMDS
 func getRegionFromIMDS(sess *session.Session) string {
```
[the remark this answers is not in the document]
ellistarn: Someone contributed this support already :) https://github.com/awslabs/karpenter/pull/447/files/a98298935078da51f056d972a50ec34b0ac6164e#diff-be94596bba7f185f87293ee8510796a92a681761efc94db2e63dfad5f0610b25R59

pkg/cloudprovider/aws/suite_test.go
ellistarn: I consider it best practice to not re-use a context. If we ever stop using a nil context (e.g. a test context?!), I wouldn't want to accidentally be sharing it.

pkg/controllers/manager.go
ellistarn: Yeah good point. There are a number of these.

ellistarn: Uhhh fmt.Fatalf doesn't exist. I explicitly didn't want to rely on our logging solution in these OrDie startup functions, since in many cases the logger isn't bootstrapped yet.

charts/karpenter/templates/logging.yaml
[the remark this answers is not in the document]
bwagner5: Gotcha

pkg/cloudprovider/aws/factory.go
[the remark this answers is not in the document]
bwagner5: NICE!

pkg/cloudprovider/aws/suite_test.go
bwagner5: fair enough

pkg/controllers/manager.go
bwagner5: ah oops, `Fatalf` is in the go log pkg.  My bad.

bwagner5 (approved):
lgtm 💯

pkg/cloudprovider/aws/capacity.go
```diff
@@ -96,25 +86,5 @@ func (c *Capacity) Create(ctx context.Context, packings []*cloudprovider.Packing
 }
 
 func (c *Capacity) GetInstanceTypes(ctx context.Context) ([]cloudprovider.InstanceType, error) {
-	return c.instanceTypeProvider.Get(ctx, c.provisioner.Spec.Cluster)
-}
-
-func (c *Capacity) GetZones(ctx context.Context) ([]string, error) {
```
njtran, line 102: Why did you remove these following three functions? Do we not need them anymore?

charts/karpenter/values.yaml
```diff
@@ -11,4 +11,10 @@ controller:
   nodeSelector: {}
   tolerations: []
   affinity: {}
-  image: "public.ecr.aws/karpenter/controller:v0.2.5@sha256:44cf4a340637555ba46289c315d700283443533bb2b084488248cf955faf866a"
+  image: "public.ecr.aws/karpenter/controller:v0.2.6@sha256:1e18463a221d55b5944153b025ea3d948216e1cc994df3286922afcfd2ad7554"
```
prateekgogia, line 14: How do we generate this image tag, is it manually build and pushed?

cmd/webhook/main.go
```diff
@@ -0,0 +1,106 @@
+package main
```
prateekgogia, line 1: Missing License header for files

cmd/webhook/main.go
```diff
@@ -0,0 +1,106 @@
...
+	Port                  int
+	HealthProbePort       int
+	ServiceName           string
+	CertificateSecretName string
+}
+
+func main() {
+	flag.IntVar(&options.Port, "port", 8443, "The port the webhook endpoint binds to for validation and mutation of resources")
+	flag.IntVar(&options.HealthProbePort, "health-probe-port", 8081, "The port the health probe endpoint binds to for reporting controller health")
+	flag.StringVar(&options.ServiceName, "service-name", "karpenter-webhook", "The name of the webhook's service")
```
prateekgogia, line 43: Do we need a flag for this?

pkg/controllers/provisioning/v1alpha1/reallocation/suite_test.go
```diff
@@ -77,6 +72,7 @@ var _ = Describe("Reallocation", func() {
 			},
 			Spec: v1alpha1.ProvisionerSpec{
 				Cluster: &v1alpha1.ClusterSpec{Name: "test-cluster", Endpoint: "http://test-cluster", CABundle: "dGVzdC1jbHVzdGVyCg=="},
+				TTLSeconds: ptr.Int32(300),
```
njtran, line 75: This isn't necessary, right? We default to 300 seconds.

pkg/controllers/manager.go
```diff
@@ -32,16 +32,15 @@ import (
...
 )
 
 var (
 	scheme = runtime.NewScheme()
 )
 
 func init() {
-	log.PanicIfError(clientgoscheme.AddToScheme(scheme), "adding clientgo to scheme")
-	log.PanicIfError(apis.AddToScheme(scheme), "adding apis to scheme")
+	_ = clientgoscheme.AddToScheme(scheme)
```
prateekgogia, line 42: Do we not want to catch these errors?

prateekgogia (changes requested):

Makefile
```diff
@@ -47,20 +47,19 @@ apply: ## Deploy the controller into your ~/.kube/config cluster
...
 		--set controller.image=ko://github.com/awslabs/karpenter/cmd/controller \
+		--set webhook.image=ko://github.com/awslabs/karpenter/cmd/webhook \
 		| $(WITH_GOFLAGS) ko apply -B -f -
 
 delete: ## Delete the controller from your ~/.kube/config cluster
-	helm template karpenter charts/karpenter \
-		$(HELM_OPTS) \
-		--create-namespace --namespace karpenter \
-		| $(WITH_GOFLAGS) ko delete -f -
+	helm delete karpenter --namespace karpenter
```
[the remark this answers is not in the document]
ellistarn: Wow this is a great catch. If you install karpenter using helm, then make delete will work. If you make apply again, it's using helm template + ko, so the chart won't be installed, and make delete will fail.

charts/karpenter/templates/controller/deployment.yaml
```diff
@@ -1,60 +1,33 @@
...
-      control-plane: karpenter
+      karpenter: controller
   template:
     metadata:
       labels:
-        control-plane: karpenter
+        karpenter: controller
     spec:
-      priorityClassName: karpenter-high-priority
+      priorityClassName: system-cluster-critical
```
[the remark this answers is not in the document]
ellistarn: In general I agree, and thanks for calling this out. However, I think Karpenter is one of the rare addons (unlike prometheus, etc) which is actually cluster-critical (https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/), since users will rely on it for capacity management. In that sense, I think it shares a similar criticality to the kube-scheduler. 

There is still system-node-critical which is even higher, and is reserved for kube-proxy and aws-node pods.

pkg/cloudprovider/aws/capacity.go
ellistarn: I reoriented the logic such that everything is registered at process start time. Check out SupportedArchitecture and variants. This is all generated by a single GetInstanceTypes on startup and then traversing that list to compute the set of supported zones,arch,etc. These are registered when the cloudprovider is instantiated (see register.go)

Constraints will then be applied against those supported baselines. Not only does this cut down on API calls, it also was necessary to make the knative webhooks work. Knative webhooks do not allow dependency injection and highly discourage making dynamic decisions, which I agree with, as it can result in nondeterministic behavior. 

As a result, these functions are no longer necessary, since we can use the static Supported variables.

pkg/controllers/provisioning/v1alpha1/reallocation/suite_test.go
ellistarn: Good eye. This is calling out a deeper issue, which is that our defaulting webhooks no longer apply during our tests. In the future, I'd like to wire up knative webhooks for our test suites, but there's an incompatibility of knative/pkg and controller-runtime (both attempt to define --kubeconfig command line argument, resulting in panic). Before we wire up the new webhooks to our test infrastructure, we'd need to migrate to knative controllers or restore controller-runtime webhooks. 

While this is definitely a regression in our e2e testing scope, the only code change that resulted was this one. Now, our validation logic is tested directly in provisioner_validation_test.go. I hope to do full e2e testing by GA (with real EKS clusters) so we will capture the integration point in those tests.

bwagner5:
The CI is failing due to golangci-lint timing out. The default run duration is 1m so I guess we just need to raise that to maybe 5m? 

`golangci-lint run --timeout 5m`

charts/karpenter/templates/controller/deployment.yaml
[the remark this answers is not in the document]
njtran: Fair enough. I think that's a good argument.

pkg/cloudprovider/aws/capacity.go
njtran: Great. Thanks.

pkg/controllers/provisioning/v1alpha1/reallocation/suite_test.go
njtran: I see. Thanks for the in-depth answer!

cmd/webhook/main.go
ellistarn: Nice!

charts/karpenter/values.yaml
ellistarn: ```AWS_PROFILE=scalability make release```

pkg/controllers/manager.go
ellistarn: I don't believe this actually can error.

prateekgogia (approved):

JacobGabrielson (approved):

PR #454, opened 2021-06-15
Support for Subnets specification/override

ellistarn:
Issue #, if available:

Description of changes:
- Users may specify subnets using `node.k8s.aws/subnet-name: dns-compliant-name`
- Users may specify subnets using `node.k8s.aws/subnet-tag-key: kubernetes-label-compliant-tag-key-name`
- Added a bunch of validation and tests

### DIFF
--- a/pkg/apis/provisioning/v1alpha1/provisioner.go
+++ b/pkg/apis/provisioning/v1alpha1/provisioner.go
@@ -138,7 +138,7 @@ type ProvisionerList struct {
 }
 
 func (c *Constraints) WithLabel(key string, value string) *Constraints {
-	c.Labels[key] = value
+	c.Labels = functional.UnionStringMaps(c.Labels, map[string]string{key: value})
 	return c
 }
 
--- a/pkg/apis/provisioning/v1alpha1/provisioner.go
+++ b/pkg/apis/provisioning/v1alpha1/provisioner.go
@@ -138,7 +138,7 @@ type ProvisionerList struct {
 }
 
 func (c *Constraints) WithLabel(key string, value string) *Constraints {
-	c.Labels = functional.UnionStringMaps(c.Labels, map[string]string{key: value})
+	c.Labels[key] = value
 	return c
 }
 
--- a/pkg/apis/provisioning/v1alpha1/provisioner.go
+++ b/pkg/apis/provisioning/v1alpha1/provisioner.go
@@ -137,27 +137,20 @@ type ProvisionerList struct {
 	Items           []Provisioner `json:"items"`
 }
 
-func (p *Provisioner) ConstraintsWithOverrides(pod *v1.Pod) *Constraints {
-	return &Constraints{
-		Taints:          p.Spec.Taints,
-		Labels:          p.Spec.Constraints.getLabels(p.Name, p.Namespace, pod),
-		Zones:           p.Spec.Constraints.getZones(pod),
-		InstanceTypes:   p.Spec.Constraints.getInstanceTypes(pod),
-		Architecture:    p.Spec.Constraints.getArchitecture(pod),
-		OperatingSystem: p.Spec.Constraints.getOperatingSystem(pod),
-	}
+func (c *Constraints) WithLabel(key string, value string) *Constraints {
+	c.Labels = functional.UnionStringMaps(c.Labels, map[string]string{key: value})
+	return c
 }
 
-func (c *Constraints) getLabels(name string, namespace string, pod *v1.Pod) map[string]string {
-	// These keys are guaranteed to not collide due to validation logic
-	return functional.UnionStringMaps(
-		c.Labels,
-		pod.Spec.NodeSelector,
-		map[string]string{
-			ProvisionerNameLabelKey:      name,
-			ProvisionerNamespaceLabelKey: namespace,
-		},
-	)
+func (c *Constraints) WithOverrides(pod *v1.Pod) *Constraints {
+	return &Constraints{
+		Taints:          c.Taints,
+		Labels:          functional.UnionStringMaps(c.Labels, pod.Spec.NodeSelector),
+		Zones:           c.getZones(pod),
+		InstanceTypes:   c.getInstanceTypes(pod),
+		Architecture:    c.getArchitecture(pod),
+		OperatingSystem: c.getOperatingSystem(pod),
+	}
 }
 
 func (c *Constraints) getZones(pod *v1.Pod) []string {
--- a/pkg/apis/provisioning/v1alpha1/provisioner_validation.go
+++ b/pkg/apis/provisioning/v1alpha1/provisioner_validation.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 
 	"github.com/awslabs/karpenter/pkg/utils/functional"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"knative.dev/pkg/apis"
 )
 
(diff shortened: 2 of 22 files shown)


njtran (changes requested):
Looks like you have some merge/rebase conflicts you still need to resolve?

JacobGabrielson (approved):

pkg/apis/provisioning/v1alpha1/provisioner.go
```diff
@@ -137,27 +137,20 @@ type ProvisionerList struct {
...
-	return &Constraints{
-		Taints:          p.Spec.Taints,
-		Labels:          p.Spec.Constraints.getLabels(p.Name, p.Namespace, pod),
-		Zones:           p.Spec.Constraints.getZones(pod),
-		InstanceTypes:   p.Spec.Constraints.getInstanceTypes(pod),
-		Architecture:    p.Spec.Constraints.getArchitecture(pod),
-		OperatingSystem: p.Spec.Constraints.getOperatingSystem(pod),
-	}
+func (c *Constraints) WithLabel(key string, value string) *Constraints {
+	c.Labels = functional.UnionStringMaps(c.Labels, map[string]string{key: value})
```
njtran, line 141: Could this also be done with `c.Labels[key] = value`?  

`functional.UnionStringMaps()` works with "last write wins", so it seems like we overwrite an existing entry in `c.labels` with either versions, so it might be cleaner just to add it in.

pkg/apis/provisioning/v1alpha1/provisioner_validation.go
```diff
@@ -40,89 +41,118 @@ var (
...
-	ValidationHook            func(ctx context.Context, spec *ProvisionerSpec) *apis.FieldError
+	ConstraintsValidationHook func(ctx context.Context, constraints *Constraints) *apis.FieldError
 )
 
 func (p *Provisioner) Validate(ctx context.Context) (errs *apis.FieldError) {
 	return errs.Also(
-		apis.ValidateObjectMetadata(p),
-		p.Spec.Validate(ctx),
+		apis.ValidateObjectMetadata(p).ViaField("metadata"),
+		p.Spec.validate(ctx).ViaField("spec"),
```
njtran, line 50: Cool stuff.

pkg/apis/provisioning/v1alpha1/provisioner_validation_test.go
```diff
@@ -65,22 +65,31 @@ var _ = Describe("Validation", func() {
...
+			Expect(provisioner.Validate(ctx)).ToNot(Succeed())
+		})
+		It("should fail for restricted labels", func() {
+			for _, label := range []string{
+				ArchitectureLabelKey,
+				OperatingSystemLabelKey,
+				ProvisionerNameLabelKey,
+				ProvisionerNamespaceLabelKey,
+				ProvisionerPhaseLabel,
+				ProvisionerTTLKey,
```
njtran, line 84: This is an annotation, so you may not need to check this in labels.

pkg/controllers/provisioning/v1alpha1/allocation/filter.go
```diff
@@ -46,11 +46,11 @@ func (f *Filter) GetProvisionablePods(ctx context.Context, provisioner *v1alpha1
...
-			func() error { return f.isUnschedulable(&pod) },
-			func() error { return f.matchesProvisioner(&pod, provisioner) },
-			func() error { return f.hasSupportedSchedulingConstraints(&pod) },
-			func() error { return f.toleratesTaints(&pod, provisioner) },
-			func() error { return f.hasSupportedLabels(&pod) },
+			func() error { return f.isUnschedulable(ctx, &pod) },
+			func() error { return f.matchesProvisioner(ctx, &pod, provisioner) },
+			func() error { return f.hasSupportedSchedulingConstraints(ctx, &pod) },
+			func() error { return f.toleratesTaints(ctx, &pod, provisioner) },
+			func() error { return f.withValidConstraints(ctx, &pod, provisioner) },
```
njtran, line 53: Maybe I'm missing something, but it looks like none of these 5 validation functions and their sub-functions use contexts. Do we need these contexts here?

pkg/test/expectations/expectations.go
```diff
@@ -115,3 +117,10 @@ func ExpectCleanedUp(c client.Client) {
 		ExpectDeleted(c, &provisioner)
 	}
 }
+
+func AttemptProvisioning(c client.Client, provisioner *v1alpha1.Provisioner, pod *v1.Pod) *v1.Pod {
+	ExpectCreatedWithStatus(c, pod)
+	ExpectCreated(c, provisioner)
+	ExpectEventuallyReconciled(c, provisioner)
+	return ExpectPodExists(c, pod.GetName(), pod.GetNamespace())
```
njtran, line 125: I like this. It looks much cleaner. I have some quick sanity checks:

- Is it possible for us to `Get` the pod before a node has been provisioned for it? If so, it's possible for this to succeed, but have the following command in tests (`ExpectNodeExists`) to fail undesirably.
- Additionally, after using this in tests, we check to see if the node exists based on the pod's nodeName. Can we ensure the pod's nodeName will exist before we finish this function?

pkg/test/pods.go
```diff
@@ -37,7 +37,15 @@ type PodOptions struct {
 	Conditions           []v1.PodCondition
 }
 
-func defaults(options PodOptions) *v1.Pod {
+// Pod creates a test pod with defaults that can be overriden by PodOptions.
+// Overrides are applied in order, with a last write wins semantic.
+func Pod(optionss ...PodOptions) *v1.Pod {
```
njtran, line 42: nit: Is this normal? lol. Seems a little weird. Maybe `options` -> `opts` -> `o`?

pkg/controllers/provisioning/v1alpha1/allocation/filter.go
```diff
@@ -46,11 +46,11 @@ func (f *Filter) GetProvisionablePods(ctx context.Context, provisioner *v1alpha1
 	provisionable := []*v1.Pod{}
 	for _, pod := range pods.Items {
 		if err := functional.ValidateAll(
-			func() error { return f.isUnschedulable(&pod) },
-			func() error { return f.matchesProvisioner(&pod, provisioner) },
-			func() error { return f.hasSupportedSchedulingConstraints(&pod) },
-			func() error { return f.toleratesTaints(&pod, provisioner) },
-			func() error { return f.hasSupportedLabels(&pod) },
+			func() error { return f.isUnschedulable(ctx, &pod) },
```
bwagner5, line 49: Do you need to add ctx to all these? Looks like the only one that uses it is `withValidConstraints`

bwagner5 (commented):
Looks good just one question and a comment typo nit.

pkg/apis/provisioning/v1alpha1/provisioner.go
ellistarn: Oopsie, nice call!

pkg/apis/provisioning/v1alpha1/provisioner_validation_test.go
ellistarn: Good catch!

pkg/controllers/provisioning/v1alpha1/allocation/filter.go
ellistarn: Knative has a logging construct (logger.FromContext(ctx).Info ...) and other amazing utilities around the ctx object (i.e. reading global configuration values from a configmap). 

I had left this in as a best practice for passing contexts around. However, I think this is probably premature, and easy to add as necessary in the future, so I'll rip this out for now. Thanks!

pkg/test/expectations/expectations.go
ellistarn: As long as the reconciliation loop has executed, we're guaranteed that if a node/scheduling decisions would have been made, it's made by the time the reconcile loop is complete.

pkg/test/pods.go
ellistarn: How about overrides.

pkg/controllers/provisioning/v1alpha1/allocation/filter.go
ellistarn: Done

njtran (approved):

PR #456, opened 2021-06-15
move lowest-n instance type options to generic packing pkg

bwagner5:
Issue #, if available:
https://github.com/awslabs/karpenter/issues/438

Description of changes:
 - Move the truncation of instance type options from the cloud provider specific implementation to the generic packing pkg. 

### DIFF
--- a/pkg/packing/packer.go
+++ b/pkg/packing/packer.go
@@ -28,6 +28,11 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
+const (
+	// maxInstanceTypes defines the number of instance type options to return to the cloud provider
+	maxInstanceTypes = 20
+)
+
 // Constraints for an efficient binpacking solution of pods onto nodes, given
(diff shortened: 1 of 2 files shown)


pkg/packing/packer.go
```diff
@@ -28,6 +28,11 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
+const (
+	// maxInstanceTypes defines the number of instance type options to return to the cloud provider
+	maxInstanceTypes = 20
```
ellistarn, line 33: Consider making this value public so that cloud providers can override it as necessary.

ellistarn (approved):
What a short PR. What's it like having such an easy life :D

pkg/packing/packer.go
bwagner5: good idea

JacobGabrielson (approved):

PR #458, opened 2021-06-16
Added delete namespace to uninstall process

njtran:
Issue #, if available:

Description of changes:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,7 @@ apply: ## Deploy the controller into your ~/.kube/config cluster
 
 delete: ## Delete the controller from your ~/.kube/config cluster
 	helm template karpenter charts/karpenter --namespace karpenter | kubectl delete -f -
+	kubectl delete ns karpenter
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -42,16 +42,19 @@ licenses: ## Verifies dependency licenses and requires GITHUB_TOKEN to be set
 	go build $(GOFLAGS) -o karpenter cmd/controller/main.go
 	golicense hack/license-config.hcl karpenter
 
+## Assumes developer is working with AWS. Remove line 52 if you are not working with AWS.
```
ellistarn, line 45: I'd remove this comment. It's not really tenable to ask users to modify makefiles locally. We will need a better solution for other clouds, let's solve it when we get there. Further, this doesn't actually break anything if you're not on AWS, it just applies a bogus annotation that does nothing. 

In the long run, we might want to do something like check if CLOUD_PROVIDER is AWS and do something special, but let's hold off for now.

ellistarn (commented):
Minor nit. Otherwise very excited for this.

ellistarn (approved):

Issue #441, opened 2021-06-08
Remove dependence on factories for cloud provider APIs


njtran:
We currently use a factory implementation to retrieve cloud-provider specific APIs like the following:
- Node Factory
- Launch Template Provider
- Subnet Provider
- Instance Type Provider
- Instance Provider

We populate the factory with the Provisioner where needed. 

We should remove this and return a single cloud provider implementation with all necessary methods.

PR #460, opened 2021-06-16
Removed Factory from CloudProvider

njtran:
Issue #, if available:
#441 

Description of changes:
Removed concept of Factories from our codebase. This moves away from the programming concept of RAII that we used to make API calls to the cloud provider. We no longer need a factory to instantiate the API for us on a call.


cmd/controller/main.go
```diff
@@ -73,7 +73,7 @@ func main() {
 	})
 
 	clientSet := kubernetes.NewForConfigOrDie(manager.GetConfig())
-	cloudProvider := registry.NewCloudProvider(cloudprovider.Options{ClientSet: clientSet})
+	cloudProvider := registry.NewAPI(cloudprovider.Options{ClientSet: clientSet})
```
ellistarn, line 76: IMO keep this as register.NewCloudProvider. I actually renamed a bunch of factories to just cloudprovider in an earlier PR to make it easier for you.

pkg/cloudprovider/aws/api.go
```diff
@@ -0,0 +1,165 @@
...
+	SupportedOperatingSystems = []string{
+		v1alpha1.OperatingSystemLinux,
+	}
+	SupportedArchitectures = []string{
+		v1alpha1.ArchitectureAmd64,
+		v1alpha1.ArchitectureArm64,
+	}
+)
+
+type API struct {
```
ellistarn, line 58: IMO you should keep this concept named "cloud provider" since it's so common across the k8s ecosystem.

pkg/cloudprovider/aws/node.go
```diff
@@ -26,12 +26,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-type NodeFactory struct {
+type NodeAPI struct {
```
ellistarn, line 29: I'm not sold on this naming change. This isn't an API, it's a https://en.wikipedia.org/wiki/Factory_method_pattern. Any time you see .For() in my code, it's a factory. Factories are hugely useful when necessary. In this case, the NodeFactory encapsulates some dependencies (e.g. EC2) vends node objects.

pkg/cloudprovider/aws/validation.go
```diff
@@ -25,11 +25,11 @@ import (
 )
 
 // Validate cloud provider specific components of the cluster spec
-func (c *Capacity) Validate(ctx context.Context) (errs *apis.FieldError) {
+func (a *API) Validate(ctx context.Context, spec *v1alpha1.ProvisionerSpec) (errs *apis.FieldError) {
```
ellistarn, line 28: Can we move these functions aws/cloudprovider.go (name change recommended above). It's only a few lines, and I think the scope is easier to encapsulate now. It feels a bit awkward that this is standing alone now that the cloudprovider API surface is so much smaller (Create, Delete, Validate, GetInstanceTypes)

pkg/cloudprovider/fake/api.go
```diff
@@ -20,23 +20,35 @@ import (
...
 )
 
-type Capacity struct {
-}
-type Termination struct {
+var (
+	NotImplementedError = fmt.Errorf("provider is not implemented. Are you running the correct release for your cloud provider?")
+)
+
+type API struct {
```
ellistarn, line 36: I wouldn't make this name change for similar reasons. This looks like a bad merge to me. I removed the NotImplementedError (which isn't used anymore) in a previous PR.

pkg/cloudprovider/fake/api.go
```diff
@@ -20,23 +20,35 @@ import (
...
+	NodeGroupStable bool
 }
 
-func (t *Termination) Terminate(ctx context.Context, nodes []*v1.Node) error {
-	return nil
+func NewAPI() *API {
+	return &API{
+		WantErr:         NotImplementedError,
+		NodeReplicas:    make(map[string]*int32),
+		NodeGroupStable: true,
```
ellistarn, line 47: Similarly, I removed this. It seems like a bad merge.

pkg/cloudprovider/registry/README.md
```diff
@@ -8,8 +8,8 @@ import (
 	"github.com/awslabs/karpenter/pkg/cloudprovider/<YOUR_PROVIDER_NAME>"
 )
 
-func NewFactory() cloudprovider.Factory {
-	return <YOUR_PROVIDER_NAME>.NewFactory()
+func NewAPI() cloudprovider.API {
```
ellistarn, line 11: NewCloudProvider imo.

pkg/cloudprovider/types.go
```diff
@@ -24,26 +24,15 @@ import (
...
-	TerminateFor() Termination
-}
-
-// Capacity provisions a set of nodes that fulfill a set of constraints.
-type Capacity interface {
+// API holds contains the methods necessary in a cloud provider
+type API interface {
 	// Create a set of nodes for each of the given constraints.
-	Create(context.Context, []*Packing) ([]*PackedNode, error)
+	Create(context.Context, []*Packing, *v1alpha1.Provisioner) ([]*PackedNode, error)
```
ellistarn, line 30: Pass the provisioner first, imo.

pkg/controllers/provisioning/v1alpha1/allocation/suite_test.go
```diff
@@ -44,7 +44,7 @@ func TestAPIs(t *testing.T) {
 
 var controller *Controller
 var env = test.NewEnvironment(func(e *test.Environment) {
-	cloudProvider := &fake.Factory{}
+	cloudProvider := fake.NewAPI()
```
ellistarn, line 47: I this change is caused by the bad merge.

ellistarn (changes requested):
Overall looks great. A couple of naming nits and what looks like some stale code from a bad merge. But otherwise a huge simplification. I think you can go more red than this already is :D

pkg/cloudprovider/aws/node.go
njtran: I agree. Good point. I'll change this back.

pkg/cloudprovider/types.go
njtran: Before Packing, but not before context, right?

ellistarn (approved):
-150 lines! Nice!

Issue #463, opened 2021-06-18
Controller keeps crashing with a panic error (v0.2.6)


black-mirror-1:
I have encounter this error with v0.2.6, where the controller keeps crashing with an invalid memory address error. Please find the logs below.

Karpenter image used: public.ecr.aws/karpenter/controller:v0.2.6@sha256:e5e41d5dcb6597cb3cba3c09451a1c6de1c4c6ac7f6216c0c2a0ed788fa8c362

kubectl get pods -n karpenter
NAME                                    READY   STATUS             RESTARTS   AGE
karpenter-controller-649c856cbb-n97dp   0/1     CrashLoopBackOff   8          26m
karpenter-webhook-866cdb7865-kw7hm      1/1     Running   

E0618 20:55:19.602094       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 321 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1d1d5e0, 0x30ea700)
        k8s.io/apimachinery@v0.19.7/pkg/util/runtime/runtime.go:74 +0xa6
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
        k8s.io/apimachinery@v0.19.7/pkg/util/runtime/runtime.go:48 +0x89
panic(0x1d1d5e0, 0x30ea700)
        runtime/panic.go:969 +0x1b9
github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation.(*Utilization).markUnderutilized(0xc0001297c0, 0x23354c0, 0xc001c8fa40, 0xc000223c00, 0xc000477020, 0x1)
        github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go:84 +0x688
github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation.(*Controller).Reconcile(0xc0014323a0, 0x23354c0, 0xc001c8fa40, 0x235e640, 0xc000223c00, 0x200c786, 0x5, 0xc000477020, 0x1)
        github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation/controller.go:72 +0x65
[...]

PR #464, opened 2021-06-18
Added SetDefault() for each Reconcile function

njtran:
Issue #, if available:
#463 
Description of changes:

### DIFF
--- a/pkg/controllers/controller.go
+++ b/pkg/controllers/controller.go
@@ -17,12 +17,12 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"reflect"
 
-	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha1"
+	"github.com/awslabs/karpenter/pkg/utils/conditions"
 
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"knative.dev/pkg/apis"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -49,22 +49,24 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 	}
 	// 2. Copy object for merge patch base
 	persisted := resource.DeepCopyObject()
-	// 3. Set to true to remove race condition
-	// TODO: remove status conditions on provisioners
-	if provisioner, ok := resource.(*v1alpha1.Provisioner); ok {
-		provisioner.StatusConditions().MarkTrue(v1alpha1.Active)
+	// 3. Set defaults to enforce invariants on object being reconciled
+	if _, ok := resource.(apis.Defaultable); ok {
+		resource.(apis.Defaultable).SetDefaults(ctx)
 	}
-	// 4. Reconcile
+	// 4. Set to true to remove race condition where multiple controllers set the status of the same object
+	// TODO: remove status conditions on provisioners
+	if conditionsAccessor, ok := resource.(apis.ConditionsAccessor); ok {
+		apis.NewLivingConditionSet(conditions.Active).Manage(conditionsAccessor).MarkTrue(conditions.Active)
+	}
+	// 5. Reconcile
 	if _, err := c.Controller.Reconcile(ctx, resource); err != nil {
 		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
 			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
 		return reconcile.Result{Requeue: true}, nil
 	}
-	// 5. Update Status using a merge patch
-	if !reflect.DeepEqual(resource, persisted) {
-		if err := c.Status().Patch(ctx, resource, client.MergeFrom(persisted)); err != nil {
-			return reconcile.Result{}, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
-		}
+	// 6. Update Status using a merge patch
+	if err := c.Status().Patch(ctx, resource, client.MergeFrom(persisted)); err != nil {
+		return reconcile.Result{}, fmt.Errorf("Failed to persist changes to %s, %w", req.NamespacedName, err)
 	}
 	return reconcile.Result{RequeueAfter: c.Interval()}, nil
 }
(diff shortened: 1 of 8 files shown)


pkg/controllers/provisioning/v1alpha1/allocation/controller.go
```diff
@@ -76,6 +76,7 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 // Reconcile executes an allocation control loop for the resource
 func (c *Controller) Reconcile(ctx context.Context, object client.Object) (reconcile.Result, error) {
 	provisioner := object.(*v1alpha1.Provisioner)
+	provisioner.SetDefaults(ctx)
```
ellistarn, line 79: Can you use the defaultable interface and put this in the generic controller?

pkg/controllers/controller.go
```diff
@@ -49,18 +50,22 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
...
-	// 4. Reconcile
+	// 5. Reconcile
 	if _, err := c.Controller.Reconcile(ctx, resource); err != nil {
 		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
 			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
 		return reconcile.Result{Requeue: true}, nil
 	}
-	// 5. Update Status using a merge patch
+	// 6. Update Status using a merge patch
 	if !reflect.DeepEqual(resource, persisted) {
```
ellistarn, line 69: Unrelated to your PR, but I realized the other day that we should be comparing statuses and patching statuses, or comparing objects and patching objects. This is currently mismatched: we compare the objects, but only patch the statuses. If resource has a changed spec, we'll send an empty patch.

pkg/controllers/controller.go
```diff
@@ -49,18 +50,22 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
...
 	// 2. Copy object for merge patch base
 	persisted := resource.DeepCopyObject()
-	// 3. Set to true to remove race condition
+	// 3. Set Defaults
+	if _, ok := resource.(apis.Defaultable); ok {
+		resource.(apis.Defaultable).SetDefaults(ctx)
+	}
+	// 4. Set to true to remove race condition
 	// TODO: remove status conditions on provisioners
 	if provisioner, ok := resource.(*v1alpha1.Provisioner); ok {
```
ellistarn, line 59: Similarly, can you use the condition manager interface while you're in here? This would remove any special provisioner logic in this file.

pkg/controllers/provisioning/v1alpha1/reallocation/suite_test.go
```diff
@@ -72,7 +71,6 @@ var _ = Describe("Reallocation", func() {
 			},
 			Spec: v1alpha1.ProvisionerSpec{
 				Cluster:    &v1alpha1.ClusterSpec{Name: "test-cluster", Endpoint: "http://test-cluster", CABundle: "dGVzdC1jbHVzdGVyCg=="},
-				TTLSeconds: ptr.Int32(300),
```
ellistarn, line 75: Does your test rely on the TTL behavior at all? Will it change tests to have nodes instantly deleted after underutilized?

pkg/controllers/controller.go
njtran: I remember talking about that. I'll change it here.

njtran: Additionally, if we only update the status, this actually won't call the patch. I think until we have the statuses working, we should delete this line.

pkg/controllers/provisioning/v1alpha1/reallocation/suite_test.go
njtran: At the end of the first test in this suite, it checks to see if the TTL annotation is part of the node. Therefore, if our default logic fails here, the test will fail.

pkg/controllers/controller.go
ellistarn: SGTM -- can you own this?

pkg/controllers/controller.go
```diff
@@ -49,22 +50,27 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
...
+	// 3. Set Defaults
+	if _, ok := resource.(apis.Defaultable); ok {
+		resource.(apis.Defaultable).SetDefaults(ctx)
+	}
+	// 4. Set to true to remove race condition
 	// TODO: remove status conditions on provisioners
-	if provisioner, ok := resource.(*v1alpha1.Provisioner); ok {
-		provisioner.StatusConditions().MarkTrue(v1alpha1.Active)
+	if _, ok := resource.(apis.ConditionManager); ok {
+		resource.(apis.ConditionManager).SetCondition(apis.Condition{
```
ellistarn, line 60: I think you can do 
```
if conditionManager, ok := resource.(apis.ConditionManager); ok {
  conditionsManager.StatusConditions().MarkTrue(v1alpha1.Active)
}
```

njtran: the knative `apis.ConditionManager` interface does not have a `StatusConditions()` method.

ellistarn (approved):

pkg/controllers/controller.go
```diff
@@ -49,22 +49,24 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 	}
 	// 2. Copy object for merge patch base
 	persisted := resource.DeepCopyObject()
-	// 3. Set to true to remove race condition
+	// 3. Set Defaults
```
JacobGabrielson, line 52: Would be nice to explain why here - I could see eventually someone removes this code because they think it's redundant

pkg/controllers/controller.go
```diff
@@ -49,22 +49,24 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
...
+	if _, ok := resource.(apis.Defaultable); ok {
+		resource.(apis.Defaultable).SetDefaults(ctx)
+	}
+	// 4. Set to true to remove race condition
```
JacobGabrielson, line 56: This comment is confusing - what race condition?

pkg/controllers/controller.go
njtran: Good call. Will do

pkg/controllers/controller.go
njtran: If we have multiple controllers reconciling a provisioner, they can set the status of the provisioner to different statuses, where last write wins. We're setting this to true for now while we change what resources the controllers are reconciling. I'll add this in too to make it clearer.

ellistarn (approved):

PR #465, opened 2021-06-18
Resolved a race condition where provisioners could skip having defaults if applied before karpenter-webhook came online

ellistarn:
Issue #, if available: #463

Description of changes:
The main change is that we now statically populate webhooks.yaml instead of allowing the webhook controller to inject the resources into the webhook. This ensures that the webhook configuration prevents creation of provisioners even if the webhook controller fails to come online (e.g. auth issues)

```rules:  - apiGroups:    - provisioning.karpenter.sh    apiVersions:    - v1alpha1    resources:    - provisioners      provisioners/status    operations:    - CREATE    - UPDATE    - DELETE```

### DIFF
--- a/charts/karpenter/templates/webhook/deployment.yaml
+++ b/charts/karpenter/templates/webhook/deployment.yaml
@@ -42,8 +42,6 @@ spec:
           ports:
             - name: webhook
               containerPort: 8443
-            - name: health-probe
-              containerPort: 8081
           livenessProbe:
             httpGet:
               scheme: HTTPS
--- a/charts/karpenter/templates/webhook/deployment.yaml
+++ b/charts/karpenter/templates/webhook/deployment.yaml
@@ -46,8 +46,8 @@ spec:
               containerPort: 8081
           livenessProbe:
             httpGet:
-              path: /healthz
-              port: 8081
+              scheme: HTTPS
+              port: 8443
           env:
             - name: SYSTEM_NAMESPACE
               valueFrom:
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -17,9 +17,6 @@ package main
 import (
 	"context"
 	"flag"
-	"fmt"
-	"log"
-	"net/http"
 
 	"github.com/awslabs/karpenter/pkg/apis"
 	"github.com/awslabs/karpenter/pkg/cloudprovider"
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: charts/karpenter/templates/webhook/deployment.yaml [hunk 1]
```diff
@@ -46,8 +46,8 @@ spec:
               containerPort: 8081
           livenessProbe:
             httpGet:
-              path: /healthz
-              port: 8081
+              scheme: HTTPS
```
njtran, line 49: A couple of quick questions that may just be my lack of understanding of liveness probes:

- Do we still need a path here? 
- Does putting the port to the same port as the webhook container mean it only pings healthy when that container is healthy?
- Do we still need the health-probe port at 8081?

cmd/webhook/main.go
```diff
@@ -46,39 +40,29 @@ var (
...
-
 	// Controllers and webhook
 	sharedmain.MainWithConfig(
 		webhook.WithOptions(injection.WithNamespaceScope(signals.NewContext(), system.Namespace()), webhook.Options{
 			Port:        options.Port,
 			ServiceName: "karpenter-webhook",
 			SecretName:  "karpenter-webhook-cert",
 		}),
-		"Karpenter Webhooks",
+		"karpenter.webhooks",
```
njtran, line 61: Looks like this component is for logging purposes. Since we name the webhooks `karpenter-webhook`, maybe we could do `Karpenter-Webhooks`? WDYT?

charts/karpenter/templates/webhook/deployment.yaml
ellistarn: > Do we still need a path here?
Knative pkg sets up probe handlers here: https://github.com/knative/pkg/blob/b558677ab03404ed118ba3e179a7438fe2d8509a/network/handlers/drain.go#L83 that do not require a path and share the port with the webhook. It's actually pretty cool how it works. I just missed this the first time around and it was a good opportunity to simplify and remove our custom webhook.

> Does putting the port to the same port as the webhook container mean it only pings healthy when that container is healthy?
Yeah more or less. See the code above.

>Do we still need the health-probe port at 8081?
Good catch! No! Removed.

cmd/webhook/main.go
ellistarn: It's corresponds to a configmap, which means that the name must be dns compliant. This was causing noisy errors in the log and logging fallback behavior.

njtran (approved):
LGTM!

PR #466, opened 2021-06-21
Added finalizers addition and removal logic for nodes

njtran:
Issue #, if available:

Description of changes:
- Finalizers are added to nodes created by the allocation controller
- We no longer have a draining/drained phase for termination labels.
- Kubectl delete node works 
- Some testing and utils improvements

### DIFF
--- a/pkg/apis/provisioning/v1alpha1/provisioner.go
+++ b/pkg/apis/provisioning/v1alpha1/provisioner.go
@@ -102,7 +102,7 @@ var (
 	// Reserved labels
 	ProvisionerNameLabelKey      = SchemeGroupVersion.Group + "/name"
 	ProvisionerNamespaceLabelKey = SchemeGroupVersion.Group + "/namespace"
-	ProvisionerPhaseLabel        = SchemeGroupVersion.Group + "/lifecycle-phase"
+	ProvisionerUnderutilizedKey  = SchemeGroupVersion.Group + "/underutilized"
 
 	// Reserved annotations
 	ProvisionerTTLKey = SchemeGroupVersion.Group + "/ttl"
@@ -113,9 +113,7 @@ var (
 )
 
 const (
-	ProvisionerUnderutilizedPhase = "underutilized"
-	ProvisionerTerminablePhase    = "terminable"
-	ProvisionerDrainingPhase      = "draining"
+	KarpenterFinalizer = "karpenter.sh/termination"
 )
 
 // Provisioner is the Schema for the Provisioners API
--- a/pkg/apis/provisioning/v1alpha1/provisioner.go
+++ b/pkg/apis/provisioning/v1alpha1/provisioner.go
@@ -100,9 +100,9 @@ var (
 	OperatingSystemLabelKey = "kubernetes.io/os"
 
 	// Reserved labels
-	ProvisionerNameLabelKey      = SchemeGroupVersion.Group + "/name"
-	ProvisionerNamespaceLabelKey = SchemeGroupVersion.Group + "/namespace"
-	ProvisionerUnderutilizedKey  = SchemeGroupVersion.Group + "/underutilized"
+	ProvisionerNameLabelKey          = SchemeGroupVersion.Group + "/name"
+	ProvisionerNamespaceLabelKey     = SchemeGroupVersion.Group + "/namespace"
+	ProvisionerUnderutilizedLabelKey = SchemeGroupVersion.Group + "/underutilized"
 
 	// Reserved annotations
 	ProvisionerTTLKey = SchemeGroupVersion.Group + "/ttl"
--- a/pkg/apis/provisioning/v1alpha1/provisioner_validation.go
+++ b/pkg/apis/provisioning/v1alpha1/provisioner_validation.go
@@ -30,7 +30,7 @@ var (
 		OperatingSystemLabelKey,
 		ProvisionerNameLabelKey,
 		ProvisionerNamespaceLabelKey,
-		ProvisionerUnderutilizedKey,
+		ProvisionerUnderutilizedLabelKey,
 		ProvisionerTTLKey,
 		ZoneLabelKey,
 		InstanceTypeLabelKey,
--- a/pkg/apis/provisioning/v1alpha1/provisioner_validation.go
+++ b/pkg/apis/provisioning/v1alpha1/provisioner_validation.go
@@ -30,7 +30,7 @@ var (
 		OperatingSystemLabelKey,
 		ProvisionerNameLabelKey,
 		ProvisionerNamespaceLabelKey,
-		ProvisionerPhaseLabel,
+		ProvisionerUnderutilizedKey,
 		ProvisionerTTLKey,
 		ZoneLabelKey,
 		InstanceTypeLabelKey,
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -154,8 +154,8 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context) ([]cloudprovider.I
 	return c.instanceTypeProvider.Get(ctx)
 }
 
-func (a *CloudProvider) Terminate(ctx context.Context, node *v1.Node) error {
-	return a.instanceProvider.Terminate(ctx, node)
+func (c *CloudProvider) Terminate(ctx context.Context, node *v1.Node) error {
+	return c.instanceProvider.Terminate(ctx, node)
 }
 
 // Validate cloud provider specific components of the cluster spec
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -154,8 +154,8 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context) ([]cloudprovider.I
 	return c.instanceTypeProvider.Get(ctx)
 }
 
-func (c *CloudProvider) Terminate(ctx context.Context, nodes []*v1.Node) error {
-	return c.instanceProvider.Terminate(ctx, nodes)
+func (a *CloudProvider) Terminate(ctx context.Context, node *v1.Node) error {
+	return a.instanceProvider.Terminate(ctx, node)
 }
 
 // Validate cloud provider specific components of the cluster spec
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -108,31 +108,25 @@ func (p *InstanceProvider) Create(ctx context.Context,
 	return createFleetOutput.Instances[0].InstanceIds[0], nil
 }
 
-func (p *InstanceProvider) Terminate(ctx context.Context, nodes []*v1.Node) error {
-	if len(nodes) == 0 {
-		return nil
+func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
+	id, err := p.getInstanceID(node)
+	if err != nil {
+		return fmt.Errorf("getting instance ID for node %s, %w", node.Name, err)
 	}
-	ids := p.getInstanceIDs(nodes)
-
-	_, err := p.ec2api.TerminateInstancesWithContext(ctx, &ec2.TerminateInstancesInput{
-		InstanceIds: ids,
+	_, err = p.ec2api.TerminateInstancesWithContext(ctx, &ec2.TerminateInstancesInput{
+		InstanceIds: []*string{id},
 	})
 	if err != nil {
-		return fmt.Errorf("terminating %d instances, %w", len(ids), err)
+		return fmt.Errorf("terminating instance %s, %w", node.Name, err)
 	}
 
 	return nil
 }
 
-func (p *InstanceProvider) getInstanceIDs(nodes []*v1.Node) []*string {
-	ids := []*string{}
-	for _, node := range nodes {
-		id := strings.Split(node.Spec.ProviderID, "/")
-		if len(id) < 5 {
-			zap.S().Debugf("Continuing after failure to parse instance id, %s has invalid format", node.Name)
-			continue
-		}
-		ids = append(ids, aws.String(id[4]))
+func (p *InstanceProvider) getInstanceID(node *v1.Node) (*string, error) {
+	id := strings.Split(node.Spec.ProviderID, "/")
+	if len(id) < 5 {
+		return nil, fmt.Errorf("parsing instance id %s", node.Spec.ProviderID)
 	}
-	return ids
+	return aws.String(id[4]), nil
 }
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -26,6 +26,7 @@ import (
 
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 )
 
 type InstanceProvider struct {
@@ -113,13 +114,11 @@ func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
 	if err != nil {
 		return fmt.Errorf("getting instance ID for node %s, %w", node.Name, err)
 	}
-	_, err = p.ec2api.TerminateInstancesWithContext(ctx, &ec2.TerminateInstancesInput{
+	if _, err = p.ec2api.TerminateInstancesWithContext(ctx, &ec2.TerminateInstancesInput{
 		InstanceIds: []*string{id},
-	})
-	if err != nil {
+	}); err != nil && !errors.IsNotFound(err) {
 		return fmt.Errorf("terminating instance %s, %w", node.Name, err)
 	}
-
 	return nil
 }
 
--- a/pkg/controllers/provisioning/v1alpha1/allocation/bind.go
+++ b/pkg/controllers/provisioning/v1alpha1/allocation/bind.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha1"
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -32,7 +33,9 @@ type Binder struct {
 }
 
 func (b *Binder) Bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) error {
-	// 1. Mark NodeReady=Unknown
+	// 1. Add the Karpenter finalizer to the node to enable the termination workflow
+	node.Finalizers = append(node.Finalizers, v1alpha1.KarpenterFinalizer)
+	// 2. Mark NodeReady=Unknown
 	// Unfortunately, this detail is necessary to prevent kube-scheduler from
 	// scheduling pods to nodes before they're created. Node Lifecycle
 	// Controller will attach a Effect=NoSchedule taint in response to this
@@ -44,7 +47,7 @@ func (b *Binder) Bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) error 
 		Type:   v1.NodeReady,
 		Status: v1.ConditionUnknown,
 	}}
-	// 2. Idempotently create a node. In rare cases, nodes can come online and
+	// 3. Idempotently create a node. In rare cases, nodes can come online and
 	// self register before the controller is able to register a node object
 	// with the API server. In the common case, we create the node object
 	// ourselves to enforce the binding decision and enable images to be pulled
@@ -55,7 +58,7 @@ func (b *Binder) Bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) error 
 		}
 	}
 
-	// 3. Bind pods
+	// 4. Bind pods
 	for _, pod := range pods {
 		if err := b.bind(ctx, node, pod); err != nil {
 			zap.S().Errorf("Continuing after failing to bind, %s", err.Error())
--- a/pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go
+++ b/pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go
@@ -44,7 +44,7 @@ func (u *Utilization) Reconcile(ctx context.Context, provisioner *v1alpha1.Provi
 	}
 
 	// 3. Mark any Node past TTL as expired
-	if err := u.markTerminable(ctx, provisioner); err != nil {
+	if err := u.terminateExpired(ctx, provisioner); err != nil {
 		return fmt.Errorf("marking nodes terminable, %w", err)
 	}
 	return nil
@@ -77,7 +77,7 @@ func (u *Utilization) markUnderutilized(ctx context.Context, provisioner *v1alph
 		persisted := node.DeepCopy()
 		node.Labels = functional.UnionStringMaps(
 			node.Labels,
-			map[string]string{v1alpha1.ProvisionerPhaseLabel: v1alpha1.ProvisionerUnderutilizedPhase},
+			map[string]string{v1alpha1.ProvisionerUnderutilizedKey: "true"},
 		)
 		node.Annotations = functional.UnionStringMaps(
 			node.Annotations,
@@ -94,9 +94,7 @@ func (u *Utilization) markUnderutilized(ctx context.Context, provisioner *v1alph
 // clearUnderutilized removes the TTL on underutilized nodes if there is sufficient resource usage
 func (u *Utilization) clearUnderutilized(ctx context.Context, provisioner *v1alpha1.Provisioner) error {
 	// 1. Get underutilized nodes
-	nodes, err := u.getNodes(ctx, provisioner, map[string]string{
-		v1alpha1.ProvisionerPhaseLabel: v1alpha1.ProvisionerUnderutilizedPhase,
-	})
+	nodes, err := u.getNodes(ctx, provisioner, map[string]string{v1alpha1.ProvisionerUnderutilizedKey: "true"})
 	if err != nil {
 		return fmt.Errorf("listing labeled underutilized nodes, %w", err)
 	}
@@ -110,7 +108,7 @@ func (u *Utilization) clearUnderutilized(ctx context.Context, provisioner *v1alp
 
 		if !utilsnode.IsUnderutilized(node, pods) {
 			persisted := node.DeepCopy()
-			delete(node.Labels, v1alpha1.ProvisionerPhaseLabel)
+			delete(node.Labels, v1alpha1.ProvisionerUnderutilizedKey)
 			delete(node.Annotations, v1alpha1.ProvisionerTTLKey)
 			if err := u.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
 				zap.S().Debugf("Could not remove underutilized labels on node %s, %w", node.Name, err)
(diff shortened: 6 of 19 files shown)


pkg/apis/provisioning/v1alpha1/provisioner.go
```diff
@@ -114,8 +114,7 @@ var (
 
 const (
 	ProvisionerUnderutilizedPhase = "underutilized"
```
ellistarn, line 116: Is this a phase anymore, or is it ProvisionerUnderutilizedLabelKey?

pkg/cloudprovider/aws/instance.go
```diff
@@ -104,31 +104,28 @@ func (p *InstanceProvider) Create(ctx context.Context,
 	return createFleetOutput.Instances[0].InstanceIds[0], nil
 }
 
-func (p *InstanceProvider) Terminate(ctx context.Context, nodes []*v1.Node) error {
-	if len(nodes) == 0 {
+func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
+	if node == nil {
```
ellistarn, line 108: I don't think you even need this nil check. Just make sure the code never calls Terminate(ctx,nil). This is defensive programming, which is great for libraries, but perhaps a bit overly verbose when we own all the code.

pkg/apis/provisioning/v1alpha1/provisioner.go
njtran: That's more accurate. Good idea.

pkg/cloudprovider/aws/instance.go
```diff
@@ -104,31 +104,28 @@ func (p *InstanceProvider) Create(ctx context.Context,
...
-		id := strings.Split(node.Spec.ProviderID, "/")
-		if len(id) < 5 {
-			zap.S().Debugf("Continuing after failure to parse instance id, %s has invalid format", node.Name)
-			continue
-		}
-		ids = append(ids, aws.String(id[4]))
+func (p *InstanceProvider) getInstanceID(node *v1.Node) (*string, error) {
+	id := strings.Split(node.Spec.ProviderID, "/")
+	if len(id) < 5 {
+		return nil, fmt.Errorf("parsing instance id")
```
ellistarn, line 128: Can you include instanceId in this error message?

pkg/controllers/provisioning/v1alpha1/allocation/bind.go
```diff
@@ -32,6 +33,7 @@ type Binder struct {
 }
 
 func (b *Binder) Bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) error {
+	node.Finalizers = append(node.Finalizers, v1alpha1.KarpenterFinalizer)
```
ellistarn, line 36: Can you add a comment here?

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,105 +38,62 @@ type Terminator struct {
...
+	return empty, nil
 }
 
-// deleteNode uses a cloudprovider-specific delete to delete a set of nodes
-func (t *Terminator) deleteNodes(ctx context.Context, nodes []*v1.Node) error {
-	// 1. Delete node in cloudprovider's instanceprovider
-	if err := t.cloudProvider.Terminate(ctx, nodes); err != nil {
+// terminateNode terminates the node then removes the finalizer to delete the node
+func (t *Terminator) terminateNode(ctx context.Context, n *v1.Node) error {
+	if err := t.cloudProvider.Terminate(ctx, n); err != nil {
```
JacobGabrielson, line 78: I think we have to be super sure that `Terminate` won't return an error if the instance is truly gone, otherwise the finalizer might never be removed (that is, the code will never get to line 86 below)

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,105 +38,62 @@ type Terminator struct {
...
-		if err := t.kubeClient.Delete(ctx, node); err != nil {
-			zap.S().Debugf("Continuing after failing to delete node %s, %s", node.Name, err.Error())
+	zap.S().Infof("Terminated instance %s", n.Name)
+	// 2. Remove finalizer from node in APIServer
+	for i, finalizer := range n.Finalizers {
+		if finalizer == provisioning.KarpenterFinalizer {
+			persisted := n.DeepCopy()
+			node.RemoveFinalizer(n, i)
+			if err := t.kubeClient.Patch(ctx, n, client.MergeFrom(persisted)); err != nil {
+				return fmt.Errorf("removing finalizer from node %s, %w", n.Name, err)
```
JacobGabrielson, line 88: What actually happens in this case (later on)? Does it all get re-driven somehow? (Especially since the cloud instance probably is gone now, so (possibly?) some of the earlier calls will start failing (although, see other comments).

pkg/cloudprovider/aws/instance.go
njtran: You're right. This will only be triggered off a reconcile loop for a node, which means that we're guaranteed it's never nil. If it is, it will fail at the top level controller anyway. Good callout!

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,105 +38,62 @@ type Terminator struct {
...
 		return fmt.Errorf("terminating cloudprovider instance, %w", err)
 	}
-	// 2. Delete node in APIServer
-	for _, node := range nodes {
-		if err := t.kubeClient.Delete(ctx, node); err != nil {
-			zap.S().Debugf("Continuing after failing to delete node %s, %s", node.Name, err.Error())
+	zap.S().Infof("Terminated instance %s", n.Name)
+	// 2. Remove finalizer from node in APIServer
+	for i, finalizer := range n.Finalizers {
+		if finalizer == provisioning.KarpenterFinalizer {
```
JacobGabrielson, line 84: It seems a bit odd that this is checked twice (`node.RemoveFinalizer()` also checks this)

pkg/utils/node/finalizer.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha1"
+	v1 "k8s.io/api/core/v1"
+)
+
+func RemoveFinalizer(node *v1.Node, index int) {
+	if node.Finalizers == nil || index < 0 {
+		return
+	}
+	if node.Finalizers[index] == v1alpha1.KarpenterFinalizer {
```
JacobGabrielson, line 12: see earlier comment - seems a bit surprising to check this again

pkg/utils/node/finalizer.go
```diff
@@ -0,0 +1,17 @@
...
+		node.Finalizers[index] = node.Finalizers[len(node.Finalizers)-1]
```
JacobGabrielson, line 13: This doesn't maintain the order of the finalizers. Is that important? (I'm not sure but wanted to ask)

pkg/utils/node/predicates.go
```diff
@@ -54,3 +55,9 @@ func IsUnderutilized(node *v1.Node, pods []*v1.Pod) bool {
 	}
 	return true
 }
+
+func IsKarpenterProvisioned(node *v1.Node) bool {
+	_, okName := node.Labels[v1alpha1.ProvisionerNameLabelKey]
+	_, okNamespace := node.Labels[v1alpha1.ProvisionerNamespaceLabelKey]
+	return okName && okNamespace
```
JacobGabrielson, line 62: Could it ever happen that one of these is set but the other is not? If so, how would that happen?

pkg/controllers/provisioning/v1alpha1/allocation/bind.go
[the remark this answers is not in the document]
njtran: Done

pkg/controllers/terminating/v1alpha1/controller.go
```diff
@@ -65,16 +66,27 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 
 // Reconcile executes a reallocation control loop for the resource
 func (c *Controller) Reconcile(ctx context.Context, object client.Object) (reconcile.Result, error) {
+	n := object.(*v1.Node)
+	if n.DeletionTimestamp == nil || !node.IsKarpenterProvisioned(n) {
```
ellistarn, line 70: node.HasTerminationFinalizer?

pkg/utils/node/predicates.go
```diff
@@ -54,3 +55,9 @@ func IsUnderutilized(node *v1.Node, pods []*v1.Pod) bool {
 	}
 	return true
 }
+
+func IsKarpenterProvisioned(node *v1.Node) bool {
```
ellistarn, line 59: I don't think we want an awareness of the provisioner labels. More explicit would be to check the finalizer. At the end of the day, think of this controller as a node finalizer that is completely decoupled from provisioning concepts.

pkg/cloudprovider/aws/instance.go
```diff
@@ -104,31 +104,28 @@ func (p *InstanceProvider) Create(ctx context.Context,
...
+	if node == nil {
 		return nil
 	}
-	ids := p.getInstanceIDs(nodes)
-
-	_, err := p.ec2api.TerminateInstancesWithContext(ctx, &ec2.TerminateInstancesInput{
-		InstanceIds: ids,
+	id, err := p.getInstanceID(node)
+	if err != nil {
+		return fmt.Errorf("getting instance ID for node %s, %w", node.Name, err)
```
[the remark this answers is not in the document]
njtran: If the instance no longer exists, the node controller in CCM will eventually realize the node has no instance to link to, then attempt to terminate the node. Since this node will have a finalizer, we will see a `deletionTimestamp` happen, thus resulting in the termination workflow.

The only way this would happen forever is if the node wanted to be terminated, but had some improperly formed instanceID. In this case, the only viable way to fix this is to format the instanceID properly, or in the case where the instance is already terminated, manually removing the finalizer.

pkg/utils/ptr/ptr.go
```diff
@@ -39,3 +39,7 @@ func PodListToSlice(pods *v1.PodList) []*v1.Pod {
 	}
 	return podPointers
 }
+
+func Int64(i int64) *int64 {
```
ellistarn, line 43: Check out knative/pkg: ptr.Int64

pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go
```diff
@@ -125,20 +123,15 @@ func (u *Utilization) clearUnderutilized(ctx context.Context, provisioner *v1alp
 // markTerminable checks if a node is past its ttl and marks it
 func (u *Utilization) markTerminable(ctx context.Context, provisioner *v1alpha1.Provisioner) error {
```
bwagner5, line 124: Should this function name be changed since it now deletes rather than just marking the node? Or is the action the same with the finalizer?

pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go
```diff
@@ -125,20 +123,15 @@ func (u *Utilization) clearUnderutilized(ctx context.Context, provisioner *v1alp
...
 	for _, node := range nodes {
 		if utilsnode.IsPastTTL(node) {
-			persisted := node.DeepCopy()
-			node.Labels = functional.UnionStringMaps(
-				node.Labels,
-				map[string]string{v1alpha1.ProvisionerPhaseLabel: v1alpha1.ProvisionerTerminablePhase},
-			)
-			if err := u.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
+			if err := u.kubeClient.Delete(ctx, node); err != nil {
 				return fmt.Errorf("patching node %s, %w", node.Name, err)
```
bwagner5, line 135: s/patching/deleting/  ?

pkg/utils/node/finalizer.go
```diff
@@ -0,0 +1,17 @@
+package node
+
+import (
+	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha1"
+	v1 "k8s.io/api/core/v1"
+)
+
+func RemoveFinalizer(node *v1.Node, index int) {
```
ellistarn, line 8: This interface should be (node, string), so the caller doesn't need to be index aware.

ellistarn, line 8: Consider turning the slice into a map[string]bool, deleting the key you don't want, and then reconverting. This is the go idiomatic way to use a set.

pkg/test/expectations/expectations.go
```diff
@@ -109,7 +125,7 @@ func ExpectCleanedUp(c client.Client) {
 	nodes := v1.NodeList{}
 	Expect(c.List(ctx, &nodes)).To(Succeed())
 	for _, node := range nodes.Items {
-		ExpectDeleted(c, &node)
+		ExpectDeletedNode(c, &node)
```
ellistarn, line 128: ExpectNodeDeleted sounds better to me.

ellistarn: Is there a reason you can't use ExpectDeleted?

pkg/test/expectations/expectations.go
```diff
@@ -73,10 +75,24 @@ func ExpectCreatedWithStatus(c client.Client, objects ...client.Object) {
 	}
 }
 
-func ExpectDeleted(c client.Client, objects ...client.Object) {
-	for _, object := range objects {
-		Expect(c.Delete(context.Background(), object)).To(Succeed())
+func ExpectDeletedNode(c client.Client, n *v1.Node) {
+	persisted := n.DeepCopy()
+	for index, finalizer := range n.Finalizers {
```
ellistarn, line 80: IMO you should just set finalizers to []

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,105 +38,62 @@ type Terminator struct {
...
+			persisted := n.DeepCopy()
+			node.RemoveFinalizer(n, i)
+			if err := t.kubeClient.Patch(ctx, n, client.MergeFrom(persisted)); err != nil {
+				return fmt.Errorf("removing finalizer from node %s, %w", n.Name, err)
+			}
+			break
 		}
-		zap.S().Infof("Terminated node %s", node.Name)
 	}
+	zap.S().Debugf("Deleted node %s", n.Name)
```
ellistarn, line 93: Thoughts on making this info?

ellistarn: Why did we change the language from Terminated to Deleted?

pkg/cloudprovider/aws/instance.go
```diff
@@ -104,31 +104,28 @@ func (p *InstanceProvider) Create(ctx context.Context,
...
+	}
+	_, err = p.ec2api.TerminateInstancesWithContext(ctx, &ec2.TerminateInstancesInput{
+		InstanceIds: []*string{id},
 	})
 	if err != nil {
```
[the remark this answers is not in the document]
njtran: If the instance is not found (cleaned up after it's been terminated), then this will return an error. 

If the instance is terminated, but still present in the console, it will not return an error, but return that it has been "terminated". The same goes for "shutting-down".

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,105 +38,62 @@ type Terminator struct {
...
+	if err := t.cloudProvider.Terminate(ctx, n); err != nil {
 		return fmt.Errorf("terminating cloudprovider instance, %w", err)
 	}
-	// 2. Delete node in APIServer
-	for _, node := range nodes {
-		if err := t.kubeClient.Delete(ctx, node); err != nil {
-			zap.S().Debugf("Continuing after failing to delete node %s, %s", node.Name, err.Error())
+	zap.S().Infof("Terminated instance %s", n.Name)
+	// 2. Remove finalizer from node in APIServer
+	for i, finalizer := range n.Finalizers {
```
ellistarn, line 83: Let's encapsulate all of this finalizer indexing logic elsewhere (node.RemoveFinalizer()?)

ellistarn: The more I think of this, you could do this entirely generically with something like:
```
node.Finalizers = functional.StringsWithout(node.Finalizers, provisioning.KarpenterFinalizer)
```

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,105 +38,62 @@ type Terminator struct {
...
+	zap.S().Debugf("Cordoned node %s", node.Name)
 	return nil
 }
 
-// terminateNodes takes in a list of expired non-drained nodes and calls drain on them
-func (t *Terminator) terminateNodes(ctx context.Context) error {
-	// 1. Get draining nodes
-	draining, err := t.getLabeledNodes(ctx, provisioning.ProvisionerDrainingPhase)
+// drainNode evicts pods from the node and returns true when fully drained
+func (t *Terminator) drainNode(ctx context.Context, node *v1.Node) (bool, error) {
```
bwagner5, line 53: do you need the bool?  Couldn't you just define a successful drain as not returning an error?

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,105 +38,62 @@ type Terminator struct {
...
+	// 1. Get pods on node
+	pods, err := t.getPods(ctx, node)
```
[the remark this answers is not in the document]
njtran: We actually don't have to worry about it if our node doesn't exist. Since this function is called by the Termination Controller's `Reconcile()` function, it will attempt to reconcile only nodes that exist in the cluster. 

If a node were to not exist, or get removed while we are doing this work, it will return an error and attempt to re-queue up a reconcile request. Once that reconcile request for this node is processed, it will realize there is no node, and then cease work.

pkg/controllers/terminating/v1alpha1/controller.go
```diff
@@ -65,16 +66,27 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
...
+	}
 	// 1. Cordon terminable nodes
-	if err := c.terminator.cordonNodes(ctx); err != nil {
-		return reconcile.Result{}, fmt.Errorf("cordoning terminable nodes, %w", err)
+	if err := c.terminator.cordonNode(ctx, n); err != nil {
+		return reconcile.Result{}, fmt.Errorf("cordoning node %s, %w", n.Name, err)
+	}
+
+	// 2. Drain terminable nodes
+	drained, err := c.terminator.drainNode(ctx, n)
```
ellistarn, line 79: When we discussed this, did we decide we were going to hold off on draining until the next reconcile loop, otherwise new pods could schedule during draining.

pkg/utils/node/finalizer.go
bwagner5, line 8: Would it make more sense to pass in the string of the finalizer you want to delete here since it's just a string, right? Passing in an index seems a little too low level since you have to loop to find the index based on a string comparison anyways. You could get rid of the for loop in the termination logic too, I think.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: Good callout. I just thought about this. This cloudprovider's `Terminate()` can fail for a few reasons. The first two are acceptable, while the last one is known.
- The instance has termination protection
  - In this case, the termination API will return an error.
- EC2 Quota limits
  - In this case, we will exponentially back off and retry, as expected.

- The instance is not found
  - In this case, the node controller in CCM will attempt to clean up the resource (but be blocked by the finalizer), which will then call this function. This error will be a `InvalidInstanceID.NotFound` exception.
  - This is unlikely to happen, but could still happen, and was covered in the design doc. The next PR, which will only add the finalizer if the delete request is validated will lower the chances of this happening. This will make it so that only if Karpenter is uninstalled after the finalizer is added and instance is terminated, but the finalizer is not removed (which happens back-to-back, so is super super unlikely), and then remains uninstalled for ~1 hour (time for terminated ec2 instances to get cleaned up in console), then is reinstalled, that we will run into this problem.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: Got it. I'll be removing this function in favor of a more generalized `functional.StringsWithout()` which will remove the need to check this.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: This happens right after the instance is terminated. If this patch fails, it will have ~1 hour to succeed to remove the finalizer, since that's when EC2 will then return a failure here. Does this answer your question?

pkg/utils/node/finalizer.go
njtran: Nope. Order doesn't matter in finalizers since they are supposed to run asynchronously and be removed asynchronously be various controllers.

pkg/utils/node/predicates.go
njtran: As I understand it, this could happen if it's removed manually, or if someone who wants a node to become controlled by a Karpenter controller who only patched in one of the labels. 

In the case that we have only one of the labels, we won't be blocked from doing work, and we can still properly terminate the instance. I can see doing `&&` or `||` here in both cases. `&&` enforces that each node must have the proper labels, so that our other controller can discover the node and do work. `||` allows nodes that aren't perfect to be included, making it easier for work to happen.

pkg/controllers/terminating/v1alpha1/controller.go
njtran: Our discussion was with the assumption that we would patch the spec at the higher level `GenericController`, which we recently made the realization that we only want to patch the Status here. 

Since we are patching the node here, not the status, we want to cordon and patch this node before continuing on to drain.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: Cases:
- (true, nil)
 - Fully drained, good to terminate
- (false, nil)
  - Still draining, but nothing failing here (has pod with `do-not-evict` or just hasn't completed yet)
  - Need to include a bool here for this case
- (false, err)
  - still draining, but something failed

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: We have Terminated instance above, but deleted node right here. These two log lines assume that the user knows the difference between an instance and node. In most cases where we terminate the instance, the node should be able to be delete. 
In that sense, I don't want to print two log lines, but rather a Debug line for those who have the verbosity up.

pkg/test/expectations/expectations.go
njtran: Our nodes now have the finalizers added with this code, so ExpectDeleted will only call delete, but not patch the finalizer out. 

I was toying with `ExpectNodeDeleted` or `ExpectDeletedNode` and decided that the former sounds like we're not doing any deletion action, but just checking for existence. The latter sounds more like we're doing an action then checking to me.

pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go
njtran: Good call. Thanks!

pkg/controllers/terminating/v1alpha1/controller.go
njtran: Thanks. We want to continue this reconcile loop until the finalizer is removed. Replacing the line with 
`if !functional.ContainsString(n.Finalizers, provisioning.KarpenterFinalizer) {`

pkg/utils/node/predicates.go
njtran: Removed it.

pkg/test/expectations/expectations.go
njtran: That would make more sense, I'll do that.

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,105 +38,58 @@ type Terminator struct {
...
-		)
-		if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
-			return fmt.Errorf("patching node %s, %w", node.Name, err)
-		}
-		zap.S().Debugf("Cordoned node %s", node.Name)
+// cordonNode cordons a node
+func (t *Terminator) cordonNode(ctx context.Context, node *v1.Node) error {
+	persisted := node.DeepCopy()
+	node.Spec.Unschedulable = true
+	if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
```
JacobGabrielson, line 45: Is it worth checking on whether it's already Unschedulable in which case no need to talk to apiserver?

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -154,8 +154,8 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context) ([]cloudprovider.I
 	return c.instanceTypeProvider.Get(ctx)
 }
 
-func (c *CloudProvider) Terminate(ctx context.Context, nodes []*v1.Node) error {
-	return c.instanceProvider.Terminate(ctx, nodes)
+func (a *CloudProvider) Terminate(ctx context.Context, node *v1.Node) error {
```
prateekgogia, line 157: Why are changing the receiver from c -> a?

pkg/utils/functional/functional.go
```diff
@@ -31,6 +31,16 @@ func UnionStringMaps(maps ...map[string]string) map[string]string {
 	return result
 }
 
+func StringsWithout(vals []string, remove string) []string {
```
JacobGabrielson, line 34: Maybe call this StringSliceWithout to match the next fxn name in the file?

JacobGabrielson: Also it's a little surprising this doesn't return a (modified) copy of the string slice rather than the original.

pkg/controllers/terminating/v1alpha1/terminate.go
JacobGabrielson: I wonder if it would just be simplest to treat InvalidInstanceId.NotFound as a non-error?

pkg/apis/provisioning/v1alpha1/provisioner_validation.go
```diff
@@ -30,7 +30,7 @@ var (
 		OperatingSystemLabelKey,
 		ProvisionerNameLabelKey,
 		ProvisionerNamespaceLabelKey,
-		ProvisionerPhaseLabel,
+		ProvisionerUnderutilizedKey,
```
ellistarn, line 33: nit: LabelKey

pkg/controllers/terminating/v1alpha1/terminate.go
bwagner5: I see, so since this is async, we need the bool to determine if it's still working is the tldr.  So on the first termination run, we shouldn't ever finish if there is at least one non-daemonset pod on the node being drained.  I guess that's not an issue now since we only drain if the node is empty.  

But anyways, this seems fine, at lest for now. Maybe there's an optimization later to recheck after the eviction starts in the same reconcile call rather than waiting on the interval.

pkg/controllers/terminating/v1alpha1/terminate.go
JacobGabrielson: (or does `Patch()` already do that?)

pkg/cloudprovider/aws/cloudprovider.go
njtran: I don't know how this happened. Good catch.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: Ellis and I discussed this in the other PR for termination watch events (since this code was pretty similar).
https://github.com/awslabs/karpenter/pull/446#discussion_r650227408

In the controller-runtime patch code, it says that the MergeFrom computes the diff, so the Patch will check that for us.
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/patch.go#L33

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: You're right. I added it in.

pkg/utils/functional/functional.go
njtran: I changed it so that it creates a new slice, so that it won't mess with the original slice.

JacobGabrielson (approved):

bwagner5 (approved):

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -38,105 +38,61 @@ type Terminator struct {
...
-		}
-		zap.S().Infof("Terminated node %s", node.Name)
+	zap.S().Infof("Terminated instance %s", node.Name)
+	// 2. Remove finalizer from node in APIServer
+	persisted := node.DeepCopy()
+	node.Finalizers = functional.StringSliceWithout(node.Finalizers, provisioning.KarpenterFinalizer)
+	if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
+		return fmt.Errorf("removing finalizer from node %s, %w", node.Name, err)
 	}
+	zap.S().Debugf("Deleted node %s", node.Name)
```
ellistarn, line 92: Why was this changed to debug?

PR #468, opened 2021-06-22
add arch, os, and provisioner anti-affinities for the karpenter deployment

bwagner5:
Issue #, if available:
N/A

Description of changes:
There are a few ways that karpenter pods can get scheduled on bad compute nodes:

1. Karpenter pods can be scheduled to nodes that were provisioned by karpenter which is not a good idea
2. Karpenter pods can be scheduled to Windows nodes which we do not publish a docker image for yet
3. Karpenter pods can be scheduled to arm64 nodes which we do not publish an image for yet

This PR adds a default anti-affinity for the karpenter provisioner name label (`provisioner.karpenter.sh/name`) and affinity to linux and amd64 nodes. 

### DIFF
--- a/charts/karpenter/templates/controller/deployment.yaml
+++ b/charts/karpenter/templates/controller/deployment.yaml
@@ -67,8 +67,6 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
             - matchExpressions:
-              - key: provisioning.karpenter.sh/name
-                operator: DoesNotExist
               - key: kubernetes.io/arch
                 operator: In
                 values:
--- a/charts/karpenter/templates/controller/deployment.yaml
+++ b/charts/karpenter/templates/controller/deployment.yaml
@@ -67,10 +67,6 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
             - matchExpressions:
-              - key: kubernetes.io/arch
-                operator: In
-                values:
-                - amd64
               - key: kubernetes.io/os
                 operator: In
                 values:
--- a/charts/karpenter/templates/controller/deployment.yaml
+++ b/charts/karpenter/templates/controller/deployment.yaml
@@ -59,10 +59,25 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- with .Values.controller.affinity }}
       affinity:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
+        {{- if .Values.controller.affinity }}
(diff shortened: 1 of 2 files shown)


JacobGabrielson (approved):

charts/karpenter/templates/controller/deployment.yaml
```diff
@@ -59,10 +59,25 @@ spec:
...
       affinity:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
+        {{- if .Values.controller.affinity }}
+          {{- toYaml .Values.controller.affinity | nindent 8 }}
+        {{- else }}
+        nodeAffinity: 
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
```
ellistarn, line 69: I don't actually think we want this. Imagine if someone wants to run karpenter in a management cluster, who's nodes are themselves managed by a separate karpenter installation.

charts/karpenter/templates/controller/deployment.yaml
```diff
@@ -59,10 +59,25 @@ spec:
...
+              - key: provisioning.karpenter.sh/name
+                operator: DoesNotExist
+              - key: kubernetes.io/arch
+                operator: In
+                values:
+                - amd64
```
ellistarn, line 75: Don't we currently build arm images w/ ko?

charts/karpenter/templates/controller/deployment.yaml
```diff
@@ -59,10 +59,25 @@ spec:
...
+              - key: kubernetes.io/os
+                operator: In
+                values:
+                - linux
```
ellistarn, line 79: +1

charts/karpenter/templates/controller/deployment.yaml
bwagner5: The makefile has the flag to, but I only see the amd64 one in the public registry.

charts/karpenter/templates/controller/deployment.yaml
bwagner5: I went ahead and removed this one.

charts/karpenter/templates/controller/deployment.yaml
bwagner5: It might be that ecr public just doesn't show the other arch ones ... the manifest appears to have the hashes listed, let me do some digging real quick.

bwagner5: I tested on an arm64 instance and it does appear to pull down an arm64 variant, so I removed the arch affinity.

ellistarn (approved):

PR #469, opened 2021-06-22
Updates for example deployment

rothgar:
Description of changes:

Major differences I made throughout the doc

- Use [semantic line breaks](https://sembr.org/)
- By default assume someone creates a new cluster with 1.20
- Add commands to tag subnets as needed
- Add line breaks to long commands
- Use `--query` with aws cli instead of relying on `jq` where possible
- Add `helm repo update`
- Fix label match for karpenter pod
- Add cluster delete command to cleanup

### DIFF
--- a/docs/aws/README.md
+++ b/docs/aws/README.md
@@ -79,7 +79,7 @@ EOM
 helm repo add karpenter https://awslabs.github.io/karpenter/charts
 helm repo update
 # For additional values, see https://github.com/awslabs/karpenter/blob/main/charts/karpenter/values.yaml
-helm upgrade --install karpenter charts/karpenter --create-namespace --namespace karpenter \
+helm upgrade --install karpenter karpenter/karpenter --create-namespace --namespace karpenter \
   --set serviceAccount.annotations.'eks\.amazonaws\.com/role-arn'=arn:aws:iam::${AWS_ACCOUNT_ID}:role/KarpenterControllerRole-${CLUSTER_NAME}
 ```
 
--- a/docs/aws/README.md
+++ b/docs/aws/README.md
@@ -28,7 +28,7 @@ eksctl create cluster \
 
 Tag the cluster subnets with the required tags for Karpenter auto discovery.
 
-Note: If you already have a cluster with version 1.19 or below you can skip this step.
+Note: If you have a cluster with version 1.18 or below you can skip this step.
 More [detailed here](https://github.com/awslabs/karpenter/issues/404#issuecomment-845283904).
 
 ```bash
@@ -42,18 +42,11 @@ aws ec2 create-tags \
     --tags Key="kubernetes.io/cluster/${CLUSTER_NAME}",Value=
 ```
 
-For existing clusters you may need to add an OIDC provider for the cluster.
-We already added one with the `eksctl create cluster` command but it is safe to run again.
-
-```bash
-eksctl utils associate-iam-oidc-provider \
---cluster ${CLUSTER_NAME} \
---approve
-```
-
 ### Setup IRSA, Karpenter Controller Role, and Karpenter Node Role
 We recommend using [CloudFormation](https://aws.amazon.com/cloudformation/) and [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IRSA) to manage these permissions.
 For production use, please review and restrict these permissions for your use case.
+
+Note: For IRSA to work your [cluster needs an OIDC provider](https://docs.aws.amazon.com/eks/latest/userguide/enable-iam-roles-for-service-accounts.html)
 
 ```bash
 export OIDC_PROVIDER=$(aws eks describe-cluster \
--- a/docs/aws/README.md
+++ b/docs/aws/README.md
@@ -58,7 +58,7 @@ export OIDC_PROVIDER=$(aws eks describe-cluster \
 # Creates IAM resources used by Karpenter
 aws cloudformation deploy \
   --stack-name Karpenter-${CLUSTER_NAME} \
-  --template-file  $(git rev-parse --show-toplevel)/docs/aws/karpenter.cloudformation.yaml \
+  --template-file ./docs/aws/karpenter.cloudformation.yaml \
   --capabilities CAPABILITY_NAMED_IAM \
   --parameter-overrides ClusterName=${CLUSTER_NAME} OpenIDConnectIdentityProvider=${OIDC_PROVIDER}
 
--- a/docs/aws/README.md
+++ b/docs/aws/README.md
@@ -1,6 +1,8 @@
-
 # AWS
-This guide will provide a complete Karpenter installation for AWS. These steps are opinionated and may need to be adapted for your use case.
+
+This guide will provide a complete Karpenter installation for AWS.
+These steps are opinionated and may need to be adapted for your use case.
+
 ## Environment
 ```bash
 CLOUD_PROVIDER=aws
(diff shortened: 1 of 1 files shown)


docs/aws/README.md
```diff
@@ -11,36 +13,61 @@ export AWS_DEFAULT_OUTPUT=json
...
---managed
+--managed \
+--with-oidc
 ```
 
-### Setup IRSA, Karpenter Controller Role, and Karpenter Node Role
-We recommend using [CloudFormation](https://aws.amazon.com/cloudformation/) and [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IRSA) to manage these permissions. For production use, please review and restrict these permissions for your use case.
+Tag the cluster subnets with the required tags for Karpenter auto discovery.
+
+Note: If you already have a cluster with version 1.19 or below you can skip this step.
```
njtran, line 31: I believe version 1.19 changed the subnet discovery system. Correct me if I'm wrong, but that means that those with clusters 1.18 or below should be able to skip this step, right?

docs/aws/README.md
```diff
@@ -11,36 +13,61 @@ export AWS_DEFAULT_OUTPUT=json
...
 --name ${CLUSTER_NAME} \
---version 1.18 \
---region ${AWS_DEFAULT_REGION} \
 --node-type m5.large \
 --nodes 1 \
 --nodes-min 1 \
 --nodes-max 10 \
---managed
+--managed \
+--with-oidc
```
njtran, line 26: Do you know if there's a difference between using this flag in `eksctl create cluster` vs `eksctl utils associate-iam-oidc-provider` ?

docs/aws/README.md
```diff
@@ -113,13 +143,22 @@ spec:
...
 ```
 
 ### Cleanup
 ```bash
 helm delete karpenter -n karpenter
 aws cloudformation delete-stack --stack-name Karpenter-${CLUSTER_NAME}
-aws ec2 describe-launch-templates | jq -r ".LaunchTemplates[].LaunchTemplateName" | grep Karpenter | xargs -I{} aws ec2 delete-launch-template --launch-template-name {}
+aws ec2 describe-launch-templates \
+    | jq -r ".LaunchTemplates[].LaunchTemplateName" \
+    | grep -i karpenter \
```
njtran, line 155: I just realized that this is something that we should change, not necessarily for this PR. We delete any launch templates that have `karpenter` in them, meaning it's possible that a customer could install Karpenter in two clusters on the same account.  IIUC, this command would delete all Karpenter launch templates, which could have adverse affects on the another Karpenter installation on another cluster.

Would you be able to change this line to be the following? With this, I believe only the launch templates created by this installation would be deleted.
```
aws ec2 describe-launch-templates \
    | jq -r ".LaunchTemplates[].LaunchTemplateName" \
    | grep "Karpenter-${CLUSTER_NAME}" \
    | xargs -I{} aws ec2 delete-launch-template --launch-template-name {}
```

njtran (commented):
Great improvements to the doc! Thanks! Just a couple of questions and checks.

docs/aws/README.md
njtran: Actually, nevermind me here. We create launch templates at allocation time, so this should be fine.

docs/aws/README.md
ellistarn: If --with-oidc replaces it, that would be `sweet`

docs/aws/README.md
```diff
@@ -11,36 +13,61 @@ export AWS_DEFAULT_OUTPUT=json
...
+    --query 'Stacks[].Outputs[?OutputKey==`SubnetsPrivate`].OutputValue' \
+    --output text)
+
+aws ec2 create-tags \
+    --resources $(echo $SUBNET_IDS | tr ',' '\n') \
+    --tags Key="kubernetes.io/cluster/${CLUSTER_NAME}",Value=
+```
+
+For existing clusters you may need to add an OIDC provider for the cluster.
+We already added one with the `eksctl create cluster` command but it is safe to run again.
```
ellistarn, line 46: Is there any reason we'd need to be safe? I'm wary of unnecessary steps/complexity. The only case I can imagine here is if a customer brought their own cluster. In that case, we could just add a note by create cluster that irsa is a dependency.

docs/aws/README.md
```diff
@@ -11,36 +13,61 @@ export AWS_DEFAULT_OUTPUT=json
...
+    --name ${CLUSTER_NAME} \
+    --query 'cluster.identity.oidc.issuer' \
+    --output text \
+    | sed 's,https://,,')
 
 # Creates IAM resources used by Karpenter
 aws cloudformation deploy \
   --stack-name Karpenter-${CLUSTER_NAME} \
-  --template-file ./docs/aws/karpenter.cloudformation.yaml \
+  --template-file  $(git rev-parse --show-toplevel)/docs/aws/karpenter.cloudformation.yaml \
```
ellistarn, line 68: @njtran has an alternative fix for this one https://github.com/awslabs/karpenter/pull/470

docs/aws/README.md
```diff
@@ -11,36 +13,61 @@ export AWS_DEFAULT_OUTPUT=json
...
+### Setup IRSA, Karpenter Controller Role, and Karpenter Node Role
+We recommend using [CloudFormation](https://aws.amazon.com/cloudformation/) and [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IRSA) to manage these permissions.
+For production use, please review and restrict these permissions for your use case.
+
+```bash
+export OIDC_PROVIDER=$(aws eks describe-cluster \
+    --name ${CLUSTER_NAME} \
+    --query 'cluster.identity.oidc.issuer' \
+    --output text \
+    | sed 's,https://,,')
```
ellistarn, line 63: Sed is a bit sketchy, as it has different syntax depending on OSX vs Ubuntu. I appreciate the desire to remove the jq dependency -- I'll be impressed if we can do it. WDYT about using `cut -d'/' -f2` here?

docs/aws/README.md
```diff
@@ -80,8 +110,8 @@ metadata:
 spec:
   cluster:
     name: ${CLUSTER_NAME}
-    caBundle: $(aws eks describe-cluster --name ${CLUSTER_NAME} | jq ".cluster.certificateAuthority.data")
-    endpoint: $(aws eks describe-cluster --name ${CLUSTER_NAME} | jq ".cluster.endpoint")
+    caBundle: $(aws eks describe-cluster --name ${CLUSTER_NAME} --query "cluster.certificateAuthority.data")
+    endpoint: $(aws eks describe-cluster --name ${CLUSTER_NAME} --query "cluster.endpoint")
```
ellistarn, line 114: +1 to all of these

docs/aws/README.md
rothgar: I actually was worried about clobbering anything with karpenter in the name too. I originally had a more complex command that would look for the nodegroup stack launch template and delete it by ID. Really we only need to delete the karpenter-${CLUSTER_NAME} launch template we created and the rest will be deleted via `eksctl delete cluster` (if needed)

docs/aws/README.md
rothgar: `--with-oidc` does the same thing. One can just be performed post-cluster creation.

docs/aws/README.md
```diff
@@ -11,36 +13,61 @@ export AWS_DEFAULT_OUTPUT=json
...
 
 ### Create a Cluster
-Note: If you already have a cluster with version 1.19 or above, you may need to manually tag your subnets for Karpenter to work [detailed here](https://github.com/awslabs/karpenter/issues/404#issuecomment-845283904).
 
-If your cluster version is 1.18 or below, you can skip this step.
+Create an EKS cluster
 ```bash
 eksctl create cluster \
 --name ${CLUSTER_NAME} \
---version 1.18 \
```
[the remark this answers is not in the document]
rothgar: You're right. I specified 1.20 to make sure I could manually tag the subnets. 1.19 is still default in the console too. Let me check when that is changing to 1.20.

docs/aws/README.md
rothgar: I was wondering about that too. Using sed without any options seemed like a safe bet but can switch back to cut if you'd prefer. I don't have a mac to test on so I couldn't verify BSD `sed`.

docs/aws/README.md
rothgar: I wanted to add this step just in case, but I'm fine changing it to a note instead of a command.

docs/aws/README.md
njtran: My fix would curl a release, while this fix would use the local file. For those who are installing it without cloning the repo, I think using curl makes a little more sense. WDYT @rothgar ?

docs/aws/README.md
```diff
@@ -11,36 +13,54 @@ export AWS_DEFAULT_OUTPUT=json
...
---region ${AWS_DEFAULT_REGION} \
 --node-type m5.large \
 --nodes 1 \
```
rothgar, line 22: @ellistarn Is there a reason we don't make this cluster with `--fargate`? If karpenter is going to add all of our instances to the cluster what's the point of having a node group at all?

docs/aws/README.md
rothgar: I thought about this some more and like your solution better. With a curl option we never require them to clone the repo. Only question for your option is if there's automation to update the release version when we make a new tag. Otherwise we'd have to always remember to update doc versions whenever we publish a new tag.

njtran: Funny that you say that, Ellis and I talked about that yesterday, and I'll also be adding that to my PR :D

docs/aws/README.md
ellistarn: I'm not sure about this syntax in particular, but sed compatibility has been such a minefield for me in the past that I try to avoid it whenever I can.

docs/aws/README.md
ellistarn: This is an interesting idea. I kinda of like this idea since it creates a more "managed" karpenter experience. However, you do need *some* nodes to run core-dns, so we'd need to run those in fargate, too.

rothgar: using the `--fargate` option automatically schedules `kube-system` and `default` workloads onto fargate. So coredns will run. We can either add a config file to include the `karpenter` namespace or schedule karpenter into `default`. The example workloads would need to be schedule into a different namespace and karpenter can handle adding nodes for that.

docs/aws/README.md
rothgar: Switched back to cut and updated it to work for the url without `https://`

docs/aws/README.md
rothgar: I'm pretty sure I can get this to work with fargate using an eksctl config file. I'm going to make that a separate PR because it'll need to move around a few commands and I don't want this PR to get complicate.

docs/aws/README.md
```diff
@@ -11,36 +13,54 @@ export AWS_DEFAULT_OUTPUT=json
...
+
+Note: For IRSA to work your [cluster needs an OIDC provider](https://docs.aws.amazon.com/eks/latest/userguide/enable-iam-roles-for-service-accounts.html)
+
 ```bash
-# Enables IRSA for your cluster. This command is idempotent, but only needs to be executed once per cluster.
-eksctl utils associate-iam-oidc-provider \
---region ${AWS_DEFAULT_REGION} \
---cluster ${CLUSTER_NAME} \
---approve
+export OIDC_PROVIDER=$(aws eks describe-cluster \
```
ellistarn, line 52: I think we want to skip the export, since the variable only needs to exist in the local shell.

ellistarn (dismissed):

ellistarn (dismissed):

ellistarn (approved):
One more time, with feeling!

PR #470, opened 2021-06-22
Added curl command to docs to pull Karpenter cloud formation template

njtran:
Same PR as https://github.com/awslabs/karpenter/pull/420.

Description of changes:
Cutting this PR since I can't resolve the merge conflicts on the OP's branch.


bwagner5 (approved):
Lgtm

docs/aws/README.md
```diff
@@ -36,9 +36,11 @@ eksctl utils associate-iam-oidc-provider \
 --approve
 
 # Creates IAM resources used by Karpenter
-aws cloudformation deploy \
+TEMPOUT=$(mktemp)
+curl -fsSL https://raw.githubusercontent.com/awslabs/karpenter/v0.2.6/docs/aws/karpenter.cloudformation.yaml > $TEMPOUT \
+| aws cloudformation deploy \
```
prateekgogia, line 41: For my understanding, why do we need to pipe from `curl` to aws cli when we are passing this file as a `--template-file ${TEMPOUT}`?

prateekgogia (changes requested):

docs/aws/README.md
njtran: My understanding is that there's no difference here. If anything piping softly enforces that they run this command together rather than separate, which is the intended flow.

prateekgogia: Why do we need to run them together as same command?
If the first command fails its still going to try and execute the `aws cloudformation deploy` and specifically pipe is confusing because to me it seems like we are passing some output from first command as input to another command. 

We could do something like this - 
```
curl -fsSL https://raw.githubusercontent.com/awslabs/karpenter/v0.2.6/docs/aws/karpenter.cloudformation.yaml > $TEMPOUT && aws cloudformation deploy ...
```
This way if curl command fails we will not be executing the aws cli.

prateekgogia (approved):
LGTM!

PR #472, opened 2021-06-22
Constraint subnets based on zones

prateekgogia:
Issue #, if available:

Description of changes:
We allow pods to be constrained to a particular zone by using this tag `topology.kubernetes.io/zone: us-east-2a`. However, we don't constraint the subnets when calling fleet API and instance gets created in the incorrect zone.


njtran (approved):
LGTM. Good catch Prateek!

ellistarn (approved):

PR #473, opened 2021-06-22
Update toolchain script and make dep for docs target

rothgar:
Description of changes:

I added a check to see if `kubebuilder` is already installed before we install (and potentially override an existing binary).
I'm not sure if karpenter is tied to the 2.3.1 release of kubebuilder. I can update to the latest 3.1.0 (or latest) release if there are not specific requirements on 2.3.1.

If this breaks `make dev` we may not want to merge this but wanted to submit it as an option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ publish: ## Generate release manifests and publish a versioned container image.
 helm:  ## Generate Helm Chart
 	cd charts;helm lint karpenter;helm package karpenter;helm repo index .
 
-docs: ## Generate Docs
+docs: toolchain ## Generate Docs
 	gen-crd-api-reference-docs \
 		-api-dir ./pkg/apis/provisioning/v1alpha1 \
 		-config $(shell go env GOMODCACHE)/github.com/ahmetb/gen-crd-api-reference-docs@v0.2.0/example-config.json \
--- a/hack/toolchain.sh
+++ b/hack/toolchain.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -eu -o pipefail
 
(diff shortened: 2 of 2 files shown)


hack/toolchain.sh
```diff
@@ -7,14 +7,17 @@ trap "rm -rf $TEMP_DIR" EXIT
...
     tools
-    kubebuilder
+    if ! command -v kubebuilder &>/dev/null; then
+        install_kubebuilder
+    fi
 }
 
 tools() {
-    cd tools; GO111MODULE=on cat tools.go | grep _ | awk -F'"' '{print $2}' | xargs -tI % go install %
+    cd tools
```
ellistarn, line 16: Any reason to separate these lines?

Makefile
```diff
@@ -73,7 +73,7 @@ publish: ## Generate release manifests and publish a versioned container image.
 helm:  ## Generate Helm Chart
 	cd charts;helm lint karpenter;helm package karpenter;helm repo index .
 
-docs: ## Generate Docs
+docs: toolchain ## Generate Docs
```
ellistarn, line 76: I'm wary of adding toolchain into the critical path of any of these commands. Toolchain should be run once by developers when starting development on the project. I see that you've made the kubebuilder bit idempotent, but the go install is a fairly expensive call. 

My preference would be to leave the current dev flow as is.

hack/toolchain.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
ellistarn, line 1: +1 thanks!

hack/toolchain.sh
rothgar: no specific reason. I ran the script through `shfmt -i 4 -w hack/toolchain.sh` from https://github.com/mvdan/sh

It automatically split it based on some common formatting. I can move it back if you'd like

Makefile
rothgar: I can add a bit more information to the target so it's more clear why the target is failing if someone hasn't run the toolchain target yet..

ellistarn: Awesome thanks

hack/toolchain.sh
ellistarn: Nope seems fine :D

hack/toolchain.sh
```diff
@@ -7,14 +7,17 @@ trap "rm -rf $TEMP_DIR" EXIT
 
 main() {
     tools
-    kubebuilder
+    if ! command -v kubebuilder &>/dev/null; then
```
ellistarn, line 10: I'd probably leave this out -- just incase we bump the kb version.

JacobGabrielson: Should we change this to check if a new enough version of kubebuilder is installed?

ellistarn:
Closing, since we have a new way to install kb

PR #474, opened 2021-06-23
Support for Security Group specification/override

ellistarn:
Issue #, if available:

Description of changes:


pkg/cloudprovider/aws/ami.go
```diff
@@ -0,0 +1,51 @@
...
+		return "", fmt.Errorf("kube server version, %w", err)
+	}
+	name := fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, KubeToAWSArchitectures[*constraints.Architecture])
+	if id, ok := p.cache.Get(name); ok {
+		return id.(string), nil
+	}
+	output, err := p.ssm.GetParameterWithContext(ctx, &ssm.GetParameterInput{Name: aws.String(name)})
+	if err != nil {
+		return "", fmt.Errorf("getting ssm parameter, %w", err)
+	}
```
prateekgogia, line 41: Do we need to add to cache here?

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -19,138 +19,121 @@ import (
...
 type launchTemplateOptions struct {
-	Provisioner  types.NamespacedName
-	Cluster      v1alpha1.ClusterSpec
-	Architecture string
-	Labels       map[string]string
-	Taints       []v1.Taint
+	// Edge-triggered fields that will only change on kube events.
+	Cluster  v1alpha1.ClusterSpec
+	UserData string
+	// Constraints Constraints
```
prateekgogia, line 73: Missed removing Constraints field?

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -170,57 +153,38 @@ func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, optio
...
-
-func (p *LaunchTemplateProvider) kubeServerVersion() (string, error) {
-	version, err := p.clientSet.Discovery().ServerVersion()
-	if err != nil {
-		return "", err
+	if err := t.Execute(&userData, struct {
+		Constraints *Constraints
+		Cluster     v1alpha1.ClusterSpec
+	}{constraints, *provisioner.Spec.Cluster}); err != nil {
+		panic(fmt.Sprintf("Parsing user data from %v, %v, %s", provisioner, constraints, err.Error()))
```
prateekgogia, line 187: Seems like if missed the validation and there is something wrong user has added to provisioner, we want to panic here? This will impact others too, if there are multiple provisioners?

pkg/cloudprovider/aws/securitygroups.go
```diff
@@ -36,26 +39,47 @@ func NewSecurityGroupProvider(ec2api ec2iface.EC2API) *SecurityGroupProvider {
...
+	if err != nil {
+		return nil, err
 	}
-	return s.getSecurityGroups(ctx, clusterName)
+	// 2. Filter by subnet name if constrained
+	if name := constraints.GetSecurityGroupName(); name != nil {
+		securityGroups = filterSecurityGroups(utils.ByName(aws.StringValue(name)), securityGroups)
+	}
+	// 3. Filter by security group tag key if constrained
+	if name := constraints.GetSecurityGroupTagKey(); name != nil {
```
prateekgogia, line 53: nit: `name` -> `tag`

pkg/cloudprovider/aws/securitygroups.go
```diff
@@ -36,26 +39,47 @@ func NewSecurityGroupProvider(ec2api ec2iface.EC2API) *SecurityGroupProvider {
...
 		Filters: []*ec2.Filter{{
-			Name:   aws.String("tag-key"),
+			Name:   aws.String("tag-key"), // Security Groups must be tagged for the cluster
 			Values: []*string{aws.String(fmt.Sprintf(ClusterTagKeyFormat, clusterName))},
 		}},
 	})
 	if err != nil {
 		return nil, fmt.Errorf("describing security groups with tag key %s, %w", fmt.Sprintf(ClusterTagKeyFormat, clusterName), err)
 	}
+	s.cache.Set(clusterName, output.SecurityGroups, CacheTTL)
```
prateekgogia, line 72: Here is a scenario for a customer use case-

- Installed Karpenter, created a pending pods, Karpenter brings up an instance
- User checks the security group doesn't like it, reads the documentation, can create their own security group
- User creates a new security group, adds the SG name to pod and re-creates the pod

Will we not see the new security group until cacheTTL expires?

pkg/cloudprovider/aws/launchtemplate.go
ellistarn: We rely on cluster being set in so many other places that you'd seg fault well before this line. I want to address this holistically at some point, but for now, we assume some invariants.

pkg/cloudprovider/aws/securitygroups.go
ellistarn: Correct, this will fail (and loop) until the security group becomes visible and it eventually heals.

prateekgogia (changes requested):

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -19,139 +19,130 @@ import (
...
+	}, nil
 }
 
-// TODO, reconcile launch template if not equal to desired launch template (AMI upgrade, role changed, etc)
-func (p *LaunchTemplateProvider) getLaunchTemplate(ctx context.Context, options *launchTemplateOptions) (*ec2.LaunchTemplate, error) {
-	describelaunchTemplateOutput, err := p.ec2api.DescribeLaunchTemplatesWithContext(ctx, &ec2.DescribeLaunchTemplatesInput{
-		LaunchTemplateNames: []*string{aws.String(launchTemplateName(options))},
+func (p *LaunchTemplateProvider) ensureLaunchTemplate(ctx context.Context, options *launchTemplateOptions) (*ec2.LaunchTemplate, error) {
+	name := launchTemplateName(options)
+	output, err := p.ec2api.DescribeLaunchTemplatesWithContext(ctx, &ec2.DescribeLaunchTemplatesInput{
```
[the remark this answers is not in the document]
ellistarn: Nice catch!

ellistarn:
Hey @prateekgogia. Github won't let me respond for some reason, but I've made the changes you requested except for the cache invalidation on security groups. I've measured all of the AWS API calls, which are between 50 and 100ms. This makes me comfortable to simply reduce the cache ttl to 1 minute.

prateekgogia (approved):
LGTM!

prateekgogia (approved):

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -19,139 +19,137 @@ import (
...
 	"github.com/patrickmn/go-cache"
 	"go.uber.org/zap"
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/kubernetes"
 )
 
 const (
-	launchTemplateNameFormat = "Karpenter-%s/%s/%s-%s"
+	launchTemplateNameFormat = "Karpenter-%s-%s"
```
njtran, line 36: Is there a general launch template name format that we should follow here? Are forwards slashes not how they're normally created, or is it mainly hyphenated?

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -19,139 +19,137 @@ import (
...
+	securityGroups, err := p.getSecurityGroupIds(ctx, provisioner, constraints)
+	if err != nil {
+		return nil, err
 	}
 
-	// Call EC2 to get launch template, creating if necessary
-	launchTemplate, err := p.getLaunchTemplate(ctx, &options)
+	// 4. Ensure the launch template exists, or create it
+	launchTemplate, err := p.ensureLaunchTemplate(ctx, &launchTemplateOptions{
+		Cluster:        *provisioner.Spec.Cluster,
```
njtran, line 107: It looks like the launch templates are pulled from the spec of the provisioner. Just wanted to check that we can't create different Launch Templates for the same cluster from different provisioners? Seems like based on this PR's changes, we cannot.

pkg/cloudprovider/aws/node.go
```diff
@@ -52,6 +53,12 @@ func (n *NodeFactory) nodesFrom(reservations []*ec2.Reservation) map[string]*v1.
 	nodes := map[string]*v1.Node{}
 	for _, reservation := range reservations {
 		for _, instance := range reservation.Instances {
+			zap.S().Debugf("Launched instance: %s, type: %s, zone: %s, hostname: %s",
+				*instance.InstanceId,
```
njtran, line 57: Great to have in logs, nice!

pkg/cloudprovider/aws/securitygroups.go
njtran: We still need to filter by subnet name and group tag key after we call the AWS API though, right?

pkg/controllers/controller.go
```diff
@@ -59,16 +59,15 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 		apis.NewLivingConditionSet(conditions.Active).Manage(conditionsAccessor).MarkTrue(conditions.Active)
 	}
 	// 5. Reconcile
-	if _, err := c.Controller.Reconcile(ctx, resource); err != nil {
-		zap.S().Errorf("Controller failed to reconcile kind %s, %s",
-			resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
-		return reconcile.Result{Requeue: true}, nil
+	_, err := c.Controller.Reconcile(ctx, resource)
+	if err != nil {
+		zap.S().Errorf("Controller failed to reconcile kind %s, %s", resource.GetObjectKind().GroupVersionKind().Kind, err.Error())
```
prateekgogia, line 64: So now we won't requeue right away once we hit an error? I am not sure if this can cause issues, and we might have missed testing this?
Why do we need to make this change though?

pkg/cloudprovider/aws/launchtemplate.go
ellistarn: I haven't seen a consistent format or pattern for this. Luckily it's easy to change in a backwards compatible way.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -28,15 +28,19 @@ import (
...
-	CacheTTL = 5 * time.Minute
+	// CacheTTL restricts QPS to AWS APIs to this interval for verifying setup
+	// resources. This value represents the maximum eventual consistency between
+	// AWS actual state and the controller's ability to provision those
+	// resources. Cache hits enable faster provisioning and reduced API load on
+	// AWS APIs, which can have a serious import on performance and scalability.
+	// DO NOT CHANGE THIS VALUE WITHOUT DUE CONSIDERATION
+	CacheTTL = 60 * time.Second
 	// CacheCleanupInterval triggers cache cleanup (lazy eviction) at this interval.
 	CacheCleanupInterval = 10 * time.Minute
```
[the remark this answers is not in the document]
ellistarn: No reason. Just a shot in the dark.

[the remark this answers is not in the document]
ellistarn: This value was lowered to reduce @prateekgogia's concerns about 5 minutes being a long time to start using security groups and subnets after they're created.

pkg/cloudprovider/aws/ami.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ssm"
+	"github.com/aws/aws-sdk-go/service/ssm/ssmiface"
+	"github.com/patrickmn/go-cache"
+	"go.uber.org/zap"
+	"k8s.io/client-go/kubernetes"
+)
+
+const kubernetesVersionCacheKey = "kubernetesVersion"
```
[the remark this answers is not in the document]
ellistarn: This cache is specific to the aws cloud provider. It's not clear anywhere else in the codebase needs this version. I could imagine factoring it out if other cloud providers need it.

pkg/cloudprovider/aws/launchtemplate.go
ellistarn: The cache key is the name, which is a hash of launchTemplateOptions. You can have many launch templates per cluster.

pkg/controllers/controller.go
njtran: After we fail a reconcile loop, we need to patch the status to say that the reconciliation failed. Regardless of if that patch fails or not, we will end up requeue-ing because returning an error requeues.

njtran (dismissed):
Nice work! LGTM.

njtran (approved):

PR #479, opened 2021-06-25
Termination now only evicts pods that don't tolerate Unschedulable

njtran:
Issue #, if available:
When draining nodes, we did not gracefully evict any daemonset pods, but left them to forcefully terminate along with the node, when no other pods remained. 

Description of changes:
- This change makes it so that we evict all pods that won't be rescheduled even if the node is cordoned (tolerates unschedulable)
- Additionally, I remove the duplicated un-used Reconcile() function in the Reallocation controller.

### DIFF
--- a/pkg/utils/pod/scheduling.go
+++ b/pkg/utils/pod/scheduling.go
@@ -45,21 +45,24 @@ func IsSchedulable(pod *v1.PodSpec, node *v1.Node) bool {
 	return true
 }
 
-// ToleratesAllTaints returns an error if the pod does not tolerate the taints
+// ToleratesTaints returns an error if the pod does not tolerate the taints
 // https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#concepts
 func ToleratesTaints(spec *v1.PodSpec, taints ...v1.Taint) error {
 	var err error
 	for _, taint := range taints {
-		taintValid := false
-		for _, toleration := range spec.Tolerations {
-			if toleration.ToleratesTaint(&taint) {
-				taintValid = true
-				continue
-			}
-		}
-		if !taintValid {
+		if !Tolerates(spec.Tolerations, taint) {
 			err = multierr.Append(err, fmt.Errorf("did not tolerate %s=%s:%s", taint.Key, taint.Value, taint.Effect))
 		}
 	}
 	return err
+}
+
+// Tolerates returns true if one of the tolerations tolerate the taint
+func Tolerates(tolerations []v1.Toleration, taint v1.Taint) bool {
+	for _, t := range tolerations {
+		if t.ToleratesTaint(&taint) {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/utils/pod/scheduling.go
+++ b/pkg/utils/pod/scheduling.go
@@ -15,6 +15,9 @@ limitations under the License.
 package pod
 
 import (
+	"fmt"
+
+	"go.uber.org/multierr"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 )
@@ -31,7 +34,7 @@ func FailedToSchedule(pod *v1.Pod) bool {
 // IsSchedulable returns true if the pod can schedule to the node
 func IsSchedulable(pod *v1.PodSpec, node *v1.Node) bool {
 	// Tolerate Taints
-	if !ToleratesAllTaints(pod, node.Spec.Taints) {
+	if err := ToleratesTaints(pod, node.Spec.Taints...); err != nil {
 		return false
 	}
 	// Match Node Selector labels
@@ -42,32 +45,21 @@ func IsSchedulable(pod *v1.PodSpec, node *v1.Node) bool {
 	return true
 }
 
-// ToleratesAllTaints returns true if the pod tolerates all taints
-func ToleratesAllTaints(pod *v1.PodSpec, taints []v1.Taint) bool {
-	for _, taint := range taints {
-		if !ToleratesTaint(pod, taint) {
-			return false
-		}
-	}
-	return true
-}
-
-// ToleratesTaint returns true if the pod tolerates the taint
+// ToleratesAllTaints returns an error if the pod does not tolerate the taints
 // https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#concepts
-func ToleratesTaint(pod *v1.PodSpec, taint v1.Taint) bool {
-	// Soft constraints are consider to be always tolerated.
-	if taint.Effect == v1.TaintEffectPreferNoSchedule {
-		return true
-	}
-	for _, toleration := range pod.Tolerations {
-		if toleration.Key == taint.Key {
-			if toleration.Operator == v1.TolerationOpExists {
-				return true
-			}
-			if toleration.Operator == v1.TolerationOpEqual && toleration.Value == taint.Value {
-				return true
+func ToleratesTaints(spec *v1.PodSpec, taints ...v1.Taint) error {
+	var err error
+	for _, taint := range taints {
+		taintValid := false
+		for _, toleration := range spec.Tolerations {
+			if toleration.ToleratesTaint(&taint) {
+				taintValid = true
+				continue
 			}
 		}
+		if !taintValid {
+			err = multierr.Append(err, fmt.Errorf("did not tolerate %s=%s:%s", taint.Key, taint.Value, taint.Effect))
+		}
 	}
-	return false
+	return err
 }
--- a/pkg/utils/pod/scheduling.go
+++ b/pkg/utils/pod/scheduling.go
@@ -47,8 +47,7 @@ func IsSchedulable(pod *v1.PodSpec, node *v1.Node) bool {
 
 // ToleratesTaints returns an error if the pod does not tolerate the taints
 // https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#concepts
-func ToleratesTaints(spec *v1.PodSpec, taints ...v1.Taint) error {
-	var err error
+func ToleratesTaints(spec *v1.PodSpec, taints ...v1.Taint) (err error) {
 	for _, taint := range taints {
 		if !Tolerates(spec.Tolerations, taint) {
 			err = multierr.Append(err, fmt.Errorf("did not tolerate %s=%s:%s", taint.Key, taint.Value, taint.Effect))
--- a/pkg/controllers/allocation/filter.go
+++ b/pkg/controllers/allocation/filter.go
@@ -44,22 +44,22 @@ func (f *Filter) GetProvisionablePods(ctx context.Context, provisioner *v1alpha2
 
 	// 2. Filter pods that aren't provisionable
 	provisionable := []*v1.Pod{}
-	for _, pod := range pods.Items {
+	for _, p := range pods.Items {
 		if err := functional.ValidateAll(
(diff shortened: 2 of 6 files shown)


pkg/utils/pod/scheduling.go
```diff
@@ -71,3 +71,15 @@ func ToleratesTaint(pod *v1.PodSpec, taint v1.Taint) bool {
 	}
 	return false
 }
+
+// ToleratesUnschedulable checks if the pod tolerates the special Kubernetes Unschedulable taint
+func ToleratesUnschedulable(spec *v1.PodSpec) bool {
```
ellistarn, line 76: Can you re-use the ToleratesTaint function?

njtran: I created this as it declares that we're checking if it tolerates the Kubernetes built-in Unschedulable taint. Additionally, it reads nicely to those who are checking out the code, rather than creating our own Taint as input that has the same values as the Unschedulable taint.

What do you think?

JacobGabrielson:
Will any of this change with https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown ?

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -62,7 +62,8 @@ func (t *Terminator) drain(ctx context.Context, node *v1.Node) (bool, error) {
 	// 2. Evict pods on node
 	empty := true
 	for _, p := range pods {
-		if !pod.IsOwnedByDaemonSet(p) {
+		// Don't evict pods that will re-schedule after being evicted
+		if !pod.ToleratesUnschedulable(&p.Spec) {
```
JacobGabrielson, line 66: I'm wondering how this works? https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-nodes-by-condition suggests (I'm probably not understanding) that daemons usually do tolerate unschedulable - so will this really achieve the result of graceful shutdown of daemonsets?

ellistarn: The challenge here is that when you evict these pods, the kcm will recreate them, and the kube scheduler will put them right back on the node (since they tolerate the cordon). There's more or less no safe way to drain these pods gracefully without reschedule loops. Luckily https://kubernetes.io/blog/2021/04/21/graceful-node-shutdown-beta/#:~:text=With%20graceful%20node%20shutdown%2C%20the,any%20resources%20they%20are%20holding. should handle the cleanup logic as part of instance termination (assuming AWS EC2 isn't too aggressive).

njtran: With regards to daemonsets, all daemonsets will have the unschedulable taint [link](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#taints-and-tolerations).

My understanding is that daemonsets are commonly processes that are essential to the operation of the node, so when the instance/node terminates, the daemonsets will also be shut down.

In addition, I changed the code to drain non-critical pods first, then critical pods to go with https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -52,28 +54,30 @@ func (t *Terminator) cordon(ctx context.Context, node *v1.Node) error {
 	return nil
 }
 
-// drain evicts pods from the node and returns true when fully drained
+// drain evicts pods from the node and returns true when non-critical and critical pods are all evicted
```
ellistarn, line 57: This comment is confusing to me.

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -52,28 +54,30 @@ func (t *Terminator) cordon(ctx context.Context, node *v1.Node) error {
...
-				},
-			}); err != nil {
-				zap.S().Debugf("Continuing after failing to evict pods from node %s, %s", node.Name, err.Error())
-			}
+	// 2. Separate pods as non-critical and critical
+	// https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
+	nonCritical := []*v1.Pod{}
+	critical := []*v1.Pod{}
+	for _, pod := range pods {
+		if pod.Spec.PriorityClassName != "system-cluster-critical" && pod.Spec.PriorityClassName != "system-node-critical" {
```
ellistarn, line 69: Use https://en.wikipedia.org/wiki/De_Morgan%27s_laws to simplify here.

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -52,28 +54,30 @@ func (t *Terminator) cordon(ctx context.Context, node *v1.Node) error {
...
+			nonCritical = append(nonCritical, pod)
+		} else {
+			critical = append(critical, pod)
 		}
 	}
-	return empty, nil
+	// 3. Evict non-critical pods
+	if t.evictPods(ctx, nonCritical) {
```
ellistarn, line 76: if help() {
  if im() {
    if trapped() {
      if in() {
        if callback() {
          return hell()
        }
      }
    }
  }
}

ellistarn:
```
if !t.evictPods(ctx, noncritical) {
    return /requeue/ 
}
if !evictPods(ctx, critical) {
   return /requeue/
}
return /done/
```

pkg/controllers/terminating/v1alpha1/suite_test.go
```diff
@@ -78,5 +80,26 @@ var _ = Describe("Termination", func() {
...
+				},
+			})
+			pod := test.Pod(test.PodOptions{
+				NodeName:    node.Name,
+				Tolerations: []v1.Toleration{{Key: v1.TaintNodeUnschedulable, Operator: v1.TolerationOpExists, Effect: v1.TaintEffectNoSchedule}},
+			})
+			ExpectCreatedWithStatus(env.Client, node)
+			ExpectCreatedWithStatus(env.Client, pod)
+
+			newPods := &v1.PodList{}
```
ellistarn, line 98: I'd just call this pods

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -92,7 +99,27 @@ func (t *Terminator) terminate(ctx context.Context, node *v1.Node) error {
...
+	for _, p := range pods {
+		// If a pod tolerates the unschedulable taint, don't evict it as it could reschedule back onto the node
+		if err := pod.ToleratesTaints(&p.Spec, []v1.Taint{{Key: v1.TaintNodeUnschedulable, Effect: v1.TaintEffectNoSchedule}}); err != nil {
+			if err := t.coreV1Client.Pods(p.Namespace).Evict(ctx, &v1beta1.Eviction{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: p.Name,
+				},
+			}); err != nil {
+				// If an eviction fails, we need to eventually try again
+				zap.S().Debugf("Continuing after failing to evict pod %s from node %s, %s", p.Name, p.Spec.NodeName, err.Error())
```
ellistarn, line 114: I think we need to be more crisp with this log line (i.e. 429 or not)

ellistarn: We also need to figure out if we can differentiate the case where PDBs are misconfigured vs a server side error.

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -92,7 +99,27 @@ func (t *Terminator) terminate(ctx context.Context, node *v1.Node) error {
...
 
-// getPods returns a list of pods scheduled to a node
+// evictPods returns true if there are no evictable pods
+func (t *Terminator) evictPods(ctx context.Context, pods []*v1.Pod) bool {
+	empty := true
+	for _, p := range pods {
+		// If a pod tolerates the unschedulable taint, don't evict it as it could reschedule back onto the node
+		if err := pod.ToleratesTaints(&p.Spec, []v1.Taint{{Key: v1.TaintNodeUnschedulable, Effect: v1.TaintEffectNoSchedule}}); err != nil {
+			if err := t.coreV1Client.Pods(p.Namespace).Evict(ctx, &v1beta1.Eviction{
+				ObjectMeta: metav1.ObjectMeta{
```
ellistarn, line 109: nit: just inline this.

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -92,7 +99,27 @@ func (t *Terminator) terminate(ctx context.Context, node *v1.Node) error {
...
+					Name: p.Name,
```
ellistarn, line 110: Curious about why the eviction object needs an object meta. Can you link to some other code that does this so I can understand beter.

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -92,7 +99,27 @@ func (t *Terminator) terminate(ctx context.Context, node *v1.Node) error {
 	return nil
 }
 
-// getPods returns a list of pods scheduled to a node
+// evictPods returns true if there are no evictable pods
+func (t *Terminator) evictPods(ctx context.Context, pods []*v1.Pod) bool {
```
ellistarn, line 103: I think we need to make this function return (bool,error) and differentiate between a 429 (backoff) and other errors like 403.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: I want to punt on this for an upcoming PR on more robust pod evictions. Currently, both 429 and 500 error codes will need an exponential back off and retry, which is currently what we do.

pkg/controllers/terminating/v1alpha1/terminate.go
njtran: I don't know if that's quite necessary as both would require an exponential backoff and retry. Anytime we would be returning false here, we would need to retry anyways. I can see in the future as we implement more robust eviction logic that this might be necessary, but as we don't want to block all evictions on one failed eviction, this works right now.

pkg/controllers/terminating/v1alpha1/terminate.go
```diff
@@ -52,28 +54,33 @@ func (t *Terminator) cordon(ctx context.Context, node *v1.Node) error {
...
+	critical := []*v1.Pod{}
+	for _, pod := range pods {
+		if pod.Spec.PriorityClassName == "system-cluster-critical" || pod.Spec.PriorityClassName == "system-node-critical" {
+			critical = append(critical, pod)
+		} else {
+			nonCritical = append(nonCritical, pod)
 		}
 	}
-	return empty, nil
+	// 3. Evict non-critical0 pods
```
JacobGabrielson, line 75: typo: `critical0`

pkg/utils/pod/scheduling.go
```diff
@@ -42,32 +45,21 @@ func IsSchedulable(pod *v1.PodSpec, node *v1.Node) bool {
...
-	for _, taint := range taints {
-		if !ToleratesTaint(pod, taint) {
-			return false
-		}
-	}
-	return true
-}
-
-// ToleratesTaint returns true if the pod tolerates the taint
+// ToleratesAllTaints returns an error if the pod does not tolerate the taints
```
JacobGabrielson, line 48: `ToleratesAllTaints` -> `ToleratesTaints`? `the taints` -> `all of the taints`?

ellistarn (approved):

ellistarn (approved):

PR #489, opened 2021-06-29
Added Troubleshooting guide for commonly known problems

njtran:
Issue #, if available:

Description of changes:
Added a `TROUBLESHOOTING.md` file to list known issues in Karpenter and ways to solve those problems that are not included in the Github issues. This doc will see changes when an issue about Karpenter is raised and solved. We will host the ways to reconcile these solutions here.  

As Karpenter is in alpha, I thought it was best to host these solutions in one guide rather than making a user sift through issues to see if it's been solved.

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -1,0 +1,25 @@
+# Troubleshooting
+
+## Known Problems + Solutions
+
+### Unable to delete nodes after uninstalling Karpenter
+Karpenter adds a [finalizer](https://github.com/awslabs/karpenter/pull/466) to nodes that it provisions to support graceful node termination. If Karpenter is uninstalled, these finalizers will cause the API Server to block deletion until the finalizers are removed.
+
+You can fix this by patching the node objects:
+- `kubectl edit node <node_name>` and remove the line that says `karpenter.sh/termination` in the finalizers field.
+- Run the following script that gets all nodes with the finalizer and removes all the finalizers from those nodes.
+   - NOTE: this will remove ALL finalizers from nodes with the karpenter finalizer.
+```{bash}
+kubectl get nodes -ojsonpath='{range .items[*].metadata}{@.name}:{@.finalizers}{"\n"}' | grep "karpenter.sh/termination" | cut -d ':' -f 1 | xargs kubectl patch node --type='json' -p='[{"op": "remove", "path": "/metadata/finalizers"}]'
+```
+
+### Nil issues with Karpenter reallocation
+If you create a Karpenter Provisioner while the webhook to default it is unavailable, it's possible to get unintentionally nil fields. [Related Issue](https://github.com/awslabs/karpenter/issues/463).
+
+   You may see some logs like this.
+```{bash}
+github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go:84 +0x688
+github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation.(*Controller).Reconcile(0xc000b004c0, 0x23354c0, 0xc000e209f0, 0x235e640, 0xc002566c40, 0x200c786, 0x5, 0xc00259c1b0, 0x1)        github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation/controller.go:72 +0x65
+github.com/awslabs/karpenter/pkg/controllers.(*GenericController).Reconcile(0xc000b00720, 0x23354c0, 0xc000e209f0, 0xc001db9be0, 0x7, 0xc001db9bd0, 0x7, 0xc000e209f0, 0x7fc864172d20, 0xc0000be2a0, ...)
+```
+This is fixed in Karpenter v0.2.7+. Reinstall Karpenter on the latest version.


docs/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,26 @@
+# Troubleshooting
+
+## Known Problems + Solutions
+
+1) If you uninstall Karpenter after creating some nodes with Karpenter, you will be left with nodes that have finalizers that need to be patched out for the node object to be deleted.
```
ellistarn, line 5: Consider making this problem it's own `## Unable to delete Nodes after uninstalling Karpenter`

docs/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,26 @@
...
+
+To do this, you can do either of the following:
+- `kubectl edit node <node_name>` and remove the line that says `karpenter.sh/termination` in the finalizers field.
+- Run the following script that gets all nodes with the finalizer and removes all the finalizers from those nodes. NOTE: this will remove ALL finalizers from nodes with the karpenter finalizer.
+```{bash}
+kubectl get nodes -ojsonpath='{range .items[*].metadata}{@.name}:{@.finalizers}{"\n"}' | grep "karpenter.sh/termination" | cut -d ':' -f 1 | xargs kubectl patch node --type='json' -p='[{"op": "remove", "path": "/metadata/finalizers"}]'
```
ellistarn, line 11: Is json the default patch type?

ellistarn, line 11: Can you use the in operator to avoid an extra git call?
https://stackoverflow.com/questions/30030652/filter-the-json-according-to-string-in-an-array-in-jsonpath

ellistarn: Looks like https://kubernetes.io/docs/reference/kubectl/jsonpath/ recommends JQ if regex is necessary (though maybe direct match works?)

docs/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,26 @@
...
+2) If you create a Karpenter Provisioner while the webhook to default it is unavailable, it's possible to get unintentionally nil fields. [Related Issue](https://github.com/awslabs/karpenter/issues/463).
+
+   You may see some logs like this.
+```{bash}
+github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation/utilization.go:84 +0x688
+github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation.(*Controller).Reconcile(0xc000b004c0, 0x23354c0, 0xc000e209f0, 0x235e640, 0xc002566c40, 0x200c786, 0x5, 0xc00259c1b0, 0x1)        github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/reallocation/controller.go:72 +0x65
+github.com/awslabs/karpenter/pkg/controllers.(*GenericController).Reconcile(0xc000b00720, 0x23354c0, 0xc000e209f0, 0xc001db9be0, 0x7, 0xc001db9bd0, 0x7, 0xc000e209f0, 0x7fc864172d20, 0xc0000be2a0, ...)
+```
+There are multiple ways to solve this:
+- Update Karpenter to the most recent version and reinstall
```
ellistarn, line 23: "This is fixed in Karpenter v0.2.7+"

docs/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,26 @@
...
+  - The most recent version should fix this. If you encountered this issue on the most recent version, please cut an issue.
+- Re-apply the Provisioner .yaml you used after checking that `kubectl get pods -n karpenter` returns a Ready pod starting with `karpenter-webhook`
+- Manually edit the Provisioner file by using `kubectl edit provisioners` and add a `.spec.ttlSeconds` field
```
ellistarn, line 26: This has drifted from the api changes

docs/TROUBLESHOOTING.md
njtran: It seems like we're switching to use more jsonpath than jq in our docs, and I'd like to keep it similar. Just based off the links you posted, it seems like there are different opinions on whether to use regex in jq or jsonpath. I opted for grep here as we won't take any actions on nodes with finalizers that don't contain `karpenter.sh/termination`.

njtran: Nope. It looks the cleanest to read in my opinion though.

akestner:
@njtran Can you update the description here to be more descriptive? What kinds of common problems? Why is this needed, etc?

docs/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,25 @@
+# Troubleshooting
+
+## Known Problems + Solutions
+
```
ellistarn, line 4: nit newline

docs/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,25 @@
...
+
+### Unable to delete nodes after uninstalling Karpenter
+If you uninstall Karpenter after creating some nodes with Karpenter, you will be left with nodes that have finalizers that need to be patched out for the node object to be deleted.
```
ellistarn, line 7: I would invert these statements. The user isn't "left with" anything, and let's be more crisp than "some nodes".

Karpenter adds a [finalizer](link) to nodes that it provisions to support graceful node termination. If Karpenter is uninstalled, these finalizers will cause the API Server to block deletion until the finalizers are removed. 

You can do this manually by patching the node objects:
```


I'd remove any manual instructions for this.

ellistarn (approved):

Issue #382, opened 2021-04-26
Make GHA Suite Tests Deterministic


njtran:
Currently, our GHA tests run locally on computers with high success rates, but run into race conditions that fail tests on GHA. We need to investigate:
- What parts our tests are creating the race conditions?
- Is ginkgo creating these race conditions?
- Is there something in GHA we can change to fix decrease this error rate?

njtran:
Sample output from GHA tests that don't occur locally (at least for me).
```
E0623 00:49:32.460681   22179 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 988 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x2aff400, 0x4885580)
	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.19.7/pkg/util/runtime/runtime.go:74 +0xc9
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.19.7/pkg/util/runtime/runtime.go:48 +0xc2
panic(0x2aff400, 0x4885580)
	/opt/hostedtoolcache/go/1.16.3/x64/src/runtime/panic.go:971 +0x499
github.com/awslabs/karpenter/pkg/cloudprovider/aws.(*CloudProvider).Create(0xc0006854d0, 0x38dc858, 0xc000b00e70, 0xc000a2c380, 0xc0009de548, 0x1, 0x1, 0x1, 0x1, 0x1, ...)
	/home/runner/work/karpenter/karpenter/pkg/cloudprovider/aws/cloudprovider.go:143 +0xbd8
github.com/awslabs/karpenter/pkg/controllers/provisioning/v1alpha1/allocation.(*Controller).Reconcile(0xc0006b8140, 0x38dc858, 0xc000b00e70, 0x3903558, 0xc000a2c380, 0x38adb80, 0xc000a2c380, 0x2df5346, 0x6)
	/home/runner/work/karpenter/karpenter/pkg/controllers/provisioning/v1alpha1/allocation/controller.go:106 +0x97c
[...]

PR #495, opened 2021-06-30
Finally decided to prevent tests from flaking 🤷

ellistarn:
Issue #382, if available:

Description of changes:
- Tests now run in ~40 seconds instead of 1:20
- Now calls reconcilers directly to make tests more deterministic. Previous approach launched the controllers and waited, which frequently was subject to GHA noisy neighbors.

### DIFF
--- a/pkg/controllers/termination/suite_test.go
+++ b/pkg/controllers/termination/suite_test.go
@@ -41,11 +41,10 @@ var env = test.NewEnvironment(func(e *test.Environment) {
 	cloudProvider := &fake.CloudProvider{}
 	registry.RegisterOrDie(cloudProvider)
 	controller = NewController(
-		e.Manager.GetClient(),
-		corev1.NewForConfigOrDie(e.Manager.GetConfig()),
+		e.Client,
+		corev1.NewForConfigOrDie(e.Config),
 		cloudProvider,
 	)
-	e.Manager.RegisterControllers(controller)
 })
 
 var _ = BeforeSuite(func() {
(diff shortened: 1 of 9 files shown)


pkg/test/expectations/expectations.go
```diff
@@ -129,9 +104,36 @@ func ExpectCleanedUp(c client.Client) {
...
-	ExpectEventuallyReconciled(c, provisioner)
-	return ExpectPodExists(c, pod.GetName(), pod.GetNamespace())
+func ExpectProvisioningSucceeded(c client.Client, controller controllers.Controller, provisioner *v1alpha2.Provisioner, pods ...*v1.Pod) []*v1.Pod {
+	for _, pod := range pods {
+		ExpectCreatedWithStatus(c, pod)
+	}
+	ExpectReconcileSucceeded(controller, provisioner)
+	result := []*v1.Pod{}
+	for _, pod := range pods {
+		result = append(result, ExpectPodExists(c, pod.GetName(), pod.GetNamespace()))
```
njtran, line 114: Is it necessary to check if `len(result) == len(pods)`?  In other words, can reconciliation succeed, but result in not all pods being bin-packed?

pkg/controllers/termination/suite_test.go
```diff
@@ -76,8 +75,10 @@ var _ = Describe("Termination", func() {
 					v1alpha2.ProvisionerNamespaceLabelKey: "default",
 				},
 			})
-			ExpectCreatedWithStatus(env.Client, node)
+			ExpectCreated(env.Client, node)
 			Expect(env.Client.Delete(ctx, node)).To(Succeed())
+			node = ExpectNodeExists(env.Client, node.Name)
```
njtran, line 80: As a side note not for this PR, but I think we should add a check here to see if it has a `deletionTimestamp` set, for good measure.

njtran (approved):

Issue #451, opened 2021-06-13
AWS: Upgrade support for existing nodes


akestner:
_(Creating issues to track `v0.3` roadmap items in #435, will update this description with more details in the coming days)_

PR #498, opened 2021-07-01
Implemented support for TTLSecondsUntilExpired

ellistarn:
**Issue, if available:**
#451

**Description of changes:**
```
Triggering termination after 1m40s (+1.132343ms) for expired node ip-192-168-41-123.us-west-2.compute.internal
```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -76,7 +76,6 @@ func main() {
 	clientSet := kubernetes.NewForConfigOrDie(manager.GetConfig())
 	cloudProvider := registry.NewCloudProvider(cloudprovider.Options{ClientSet: clientSet})
 
-
 	if err := expiration.NewController(manager.GetClient()).Register(manager); err != nil {
 		panic(err)
 	}
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/awslabs/karpenter/pkg/cloudprovider/registry"
 	"github.com/awslabs/karpenter/pkg/controllers"
 	"github.com/awslabs/karpenter/pkg/controllers/allocation"
+	"github.com/awslabs/karpenter/pkg/controllers/expiration"
 	"github.com/awslabs/karpenter/pkg/controllers/reallocation"
 	"github.com/awslabs/karpenter/pkg/controllers/termination"
 	"github.com/awslabs/karpenter/pkg/utils/log"
@@ -74,6 +75,11 @@ func main() {
 
 	clientSet := kubernetes.NewForConfigOrDie(manager.GetConfig())
 	cloudProvider := registry.NewCloudProvider(cloudprovider.Options{ClientSet: clientSet})
+
+
+	if err := expiration.NewController(manager.GetClient()).Register(manager); err != nil {
+		panic(err)
+	}
 
 	if err := manager.RegisterControllers(
 		allocation.NewController(manager.GetClient(), clientSet.CoreV1(), cloudProvider),
--- a/pkg/apis/provisioning/v1alpha2/provisioner_validation.go
+++ b/pkg/apis/provisioning/v1alpha2/provisioner_validation.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 
 	"github.com/awslabs/karpenter/pkg/utils/functional"
+	"github.com/awslabs/karpenter/pkg/utils/ptr"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"knative.dev/pkg/apis"
 )
@@ -53,6 +54,8 @@ func (p *Provisioner) Validate(ctx context.Context) (errs *apis.FieldError) {
 
 func (s *ProvisionerSpec) validate(ctx context.Context) (errs *apis.FieldError) {
 	return errs.Also(
+		s.validateTTLSecondsUntilExpired(),
+		s.validateTTLSecondsAfterEmpty(),
 		s.Cluster.validate().ViaField("cluster"),
 		// This validation is on the ProvisionerSpec despite the fact that
 		// labels are a property of Constraints. This is necessary because
@@ -62,6 +65,19 @@ func (s *ProvisionerSpec) validate(ctx context.Context) (errs *apis.FieldError) 
 		s.validateRestrictedLabels(),
 		s.Constraints.Validate(ctx),
 	)
+}
+
+func (s *ProvisionerSpec) validateTTLSecondsUntilExpired() (errs *apis.FieldError) {
+	if ptr.Int64Value(s.TTLSecondsUntilExpired) < 0 {
+		return errs.Also(apis.ErrInvalidValue("cannot be negative", "ttlSecondsUntilExpired"))
+	}
+	return errs
+}
+func (s *ProvisionerSpec) validateTTLSecondsAfterEmpty() (errs *apis.FieldError) {
+	if ptr.Int64Value(s.TTLSecondsAfterEmpty) < 0 {
+		return errs.Also(apis.ErrInvalidValue("cannot be negative", "ttlSecondsAfterEmpty"))
+	}
+	return errs
 }
 
 func (s *ProvisionerSpec) validateRestrictedLabels() (errs *apis.FieldError) {
--- a/pkg/apis/provisioning/v1alpha2/provisioner_validation.go
+++ b/pkg/apis/provisioning/v1alpha2/provisioner_validation.go
@@ -31,7 +31,7 @@ var (
 		ProvisionerNameLabelKey,
 		ProvisionerNamespaceLabelKey,
 		ProvisionerUnderutilizedLabelKey,
-		ProvisionerTTLKey,
+		ProvisionerTTLAfterEmptyKey,
 		ZoneLabelKey,
 		InstanceTypeLabelKey,
 	}
--- a/pkg/controllers/expiration/controller.go
+++ b/pkg/controllers/expiration/controller.go
@@ -57,7 +57,7 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 	// 2. Ignore if node is already deleting
-	if node.DeletionTimestamp != nil {
+	if !node.DeletionTimestamp.IsZero() {
 		return reconcile.Result{}, nil
 	}
 	// 3. Ignore if provisioner doesn't exist
--- a/pkg/controllers/expiration/controller.go
+++ b/pkg/controllers/expiration/controller.go
@@ -1,0 +1,123 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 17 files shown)


cmd/controller/main.go
```diff
@@ -75,6 +76,11 @@ func main() {
 	clientSet := kubernetes.NewForConfigOrDie(manager.GetConfig())
 	cloudProvider := registry.NewCloudProvider(cloudprovider.Options{ClientSet: clientSet})
 
+
+	if err := expiration.NewController(manager.GetClient()).Register(manager); err != nil {
```
njtran, line 80: I understand that we're going to be moving this out eventually. Would you be able to put these with the rest of controllers until you change it?

pkg/controllers/expiration/controller.go
```diff
@@ -0,0 +1,123 @@
...
+	// 1. Get node
+	node := &v1.Node{}
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, node); err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+	// 2. Ignore if node is already deleting
+	if node.DeletionTimestamp != nil {
```
njtran, line 60: `if !node.DeletionTimestamp.IsZero() {`

pkg/controllers/expiration/controller.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	// 4. Ignore if TTLSecondsUntilExpired isn't defined
+	if provisioner.Spec.TTLSecondsUntilExpired == nil {
+		return reconcile.Result{}, nil
+	}
+	// 5. Trigger termination workflow if expired
+	expirationTTL := time.Duration(ptr.Int64Value(provisioner.Spec.TTLSecondsUntilExpired)) * time.Second
+	expirationTime := node.CreationTimestamp.Add(expirationTTL)
+	if time.Now().After(expirationTime) {
+		zap.S().Infof("Triggering termination after %s (+%s) for expired node %s", expirationTTL, time.Since(expirationTime), node.Name)
```
njtran, line 87: What do you think about referencing the name of the Expiration controller here to make it more explicit that it was the expiry controller? It could be confusing not knowing which type of expiration it was, considering we have this concept for utilization.

pkg/controllers/expiration/controller.go
```diff
@@ -0,0 +1,123 @@
...
+			return reconcile.Result{}, fmt.Errorf("expiring node %s, %w", node.Name, err)
+		}
+		return reconcile.Result{}, nil
+	}
+
+	// 6. Backoff until expired
+	return reconcile.Result{RequeueAfter: time.Until(expirationTime)}, nil
+}
+
+func (c *Controller) provisionerToNodes(o client.Object) (requests []reconcile.Request) {
```
njtran, line 98: Nice solution!

pkg/controllers/expiration/suite_test.go
```diff
@@ -0,0 +1,101 @@
...
+		ExpectCleanedUp(env.Client)
+	})
+
+	It("should ignore nodes without a provisioner", func() {
+		node := test.Node(test.NodeOptions{})
+		ExpectCreated(env.Client, provisioner, node)
+		ExpectReconcileSucceeded(controller, client.ObjectKeyFromObject(node))
+
+		node = ExpectNodeExists(env.Client, node.Name)
+		Expect(node.DeletionTimestamp).To(BeNil())
```
njtran, line 73: `Expect(node.DeletionTimestamp.IsZero()).To(BeTrue())`

pkg/controllers/expiration/suite_test.go
```diff
@@ -0,0 +1,101 @@
...
+			Labels: map[string]string{
+				v1alpha2.ProvisionerNameLabelKey:      provisioner.Name,
+				v1alpha2.ProvisionerNamespaceLabelKey: provisioner.Namespace,
+			},
+		})
+		ExpectCreated(env.Client, provisioner, node)
+		ExpectReconcileSucceeded(controller, client.ObjectKeyFromObject(node))
+
+		node = ExpectNodeExists(env.Client, node.Name)
+		Expect(node.DeletionTimestamp).To(BeNil())
```
njtran, line 86: Same as above, and same for below

pkg/controllers/expiration/suite_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+var _ = Describe("Reconciliation", func() {
+	var provisioner *v1alpha2.Provisioner
+
+	BeforeEach(func() {
+		provisioner = &v1alpha2.Provisioner{
+			ObjectMeta: metav1.ObjectMeta{Name: strings.ToLower(randomdata.SillyName()), Namespace: "default"},
+			Spec: v1alpha2.ProvisionerSpec{
+				Cluster:                &v1alpha2.Cluster{Name: "test-cluster", Endpoint: "http://test-cluster", CABundle: "dGVzdC1jbHVzdGVyCg=="},
+				TTLSecondsUntilExpired: ptr.Int64(30),
```
njtran, line 58: Could you add a test to check if it's a no-op if `TTLSecondsUntilExpired` is unset?

pkg/apis/provisioning/v1alpha2/provisioner_validation.go
```diff
@@ -31,7 +31,7 @@ var (
 		ProvisionerNameLabelKey,
 		ProvisionerNamespaceLabelKey,
 		ProvisionerUnderutilizedLabelKey,
-		ProvisionerTTLKey,
+		ProvisionerTTLAfterEmptyKey,
```
njtran, line 34: Can you also add in validation to make sure that these TTLs are not negative?

pkg/controllers/expiration/controller.go
ellistarn: This is the difference between "empty" and "expired". I'd prefer to leave as is, but I do want to do a followup PR (using kn/pkg) which prepends a [Expiration] or [Utilization] to all of the log lines here.

cmd/controller/main.go
ellistarn: Nope -- different interfaces.

pkg/apis/provisioning/v1alpha2/provisioner_validation.go
ellistarn: Sure!

pkg/controllers/expiration/controller.go
ellistarn: Thanks!

pkg/controllers/expiration/controller.go
```diff
@@ -0,0 +1,123 @@
...
+	// 1. Get node
+	node := &v1.Node{}
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, node); err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+	// 2. Ignore if node is already deleting
+	if !node.DeletionTimestamp.IsZero() {
```
JacobGabrielson, line 60: Could `DeletionTimestamp` be `nil`? Also wondering if it's a better practice to call `GetDeletionTimestamp()` instead? (I'm not sure - just askin').

pkg/controllers/expiration/controller.go
```diff
@@ -0,0 +1,123 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	// 6. Backoff until expired
+	return reconcile.Result{RequeueAfter: time.Until(expirationTime)}, nil
+}
+
+func (c *Controller) provisionerToNodes(o client.Object) (requests []reconcile.Request) {
+	nodes := &v1.NodeList{}
+	if err := c.kubeClient.List(context.Background(), nodes, client.MatchingLabels(map[string]string{
```
JacobGabrielson, line 100: Should a `context` be passed down through the call chain rather than assuming `Background` here?

pkg/controllers/expiration/controller.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (c *Controller) Register(m manager.Manager) error {
+	return controllerruntime.
+		NewControllerManagedBy(m).
+		Named("Expiration").
+		For(&v1.Node{}).
+		Watches(
+			// Reconcile all nodes related to a provisioner when it changes.
+			&source.Kind{Type: &v1alpha2.Provisioner{}},
```
JacobGabrielson, line 119: Just double-checking ... does this work (well) if there are multiple `Provisioner`s in the same cluster? (Say there are 3 Provisioners, I assume this means each one will do the fan-out any time one of them changes?) (Not necessarily a big deal, just making sure I understand).

pkg/controllers/expiration/controller.go
ellistarn: Kind of a neat trick of how go pointer receivers work, you can actually call member functions on nil objects.

In this case:
```
func (t *Time) IsZero() bool {
	if t == nil {
		return true
	}
	return t.Time.IsZero()
}
```
Even if this is nil, we won't get an NPE.

w.r.t using GetDeletionTimestamp() vs DeletionTimestamp, I'm against getters/setters in general, unless they're needed for some interface or generalization somewhere (which is why these functions even exists on metav1.ObjectMeta). In this case, we have a concrete object, instead of just a runtime.Object, so I prefer to use the concrete.

pkg/controllers/expiration/controller.go
ellistarn: I wish we had one, but we don't have the caller context. This is probably a good feature request for controller runtime. I'd much prefer the signature:
```func (ctx context.Context, o client.Obeject) ([]reconcile.Request, error)```

pkg/controllers/expiration/controller.go
ellistarn: It won't fully fan out, since we're constraining nodes that are "owned" by a provisioner on line 101/102. We're effectively mapping provisioners to nodes and will trigger a reconcile loop for all nodes whenever a provisioner changes. 

e.g. Provisioner 1 updates it's TTL, all nodes for that provisioner will reconcile to check that TTL.

JacobGabrielson: But won't all 3 Provisioners reconcile (because `For(&v1.Node{})` on line 116?) and do the fan-out?

ellistarn: For and Watches are two interfaces (at different levels of granularity) that establish watches. There is no coupling between the calls.

1. The For(Nodes) statement says that any changes to nodes will trigger an expiration reconcile loop for that node. This is *slightly* overkill. We could probably limit this to just update, and limit it to trigger an event if it has a provisioner w/ ttluntilexpired, but instead of putting those checks in the event handler, we just put them into the reconcile loop. The theory is that with ctrl batching, the performance is likely slightly better this way (less logic on the event batching side). 
2. The Watches(Provisioners, MapFunc(Provisioner -> Nodes)) says that any changes to provisioners will trigger an expiration reconcile loop for a set of nodes. This is again probably overkill -- could be applied on just create/update (not delete). The same logic applies for putting this logic on the event handling side vs the reconcile side. 

Probably easier to discuss in person if I'm still missing something.

JacobGabrielson: I don't think it's a blocking issue (probably just my misunderstanding), let's discuss offline.

JacobGabrielson (approved):

PR #499, opened 2021-07-01
Added do-not-evict option for pods

njtran:
**Issue, if available:**

**Description of changes:**
- Users can now specify annotations on their pods `karpenter.sh/do-not-evict` where requesting to terminate that node will cordon it, but wait to drain until those pods are terminated.

### DIFF
--- a/pkg/controllers/termination/controller.go
+++ b/pkg/controllers/termination/controller.go
@@ -55,7 +55,7 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 func (c *Controller) Reconcile(ctx context.Context, object client.Object) (reconcile.Result, error) {
 	node := object.(*v1.Node)
 	// 1. Check if node is terminable
-	if node.DeletionTimestamp == nil || !functional.ContainsString(node.Finalizers, provisioning.KarpenterFinalizer) {
+	if node.DeletionTimestamp.IsZero() || !functional.ContainsString(node.Finalizers, provisioning.KarpenterFinalizer) {
 		return reconcile.Result{}, nil
 	}
 	// 2. Cordon node
--- a/pkg/controllers/termination/suite_test.go
+++ b/pkg/controllers/termination/suite_test.go
@@ -22,13 +22,12 @@ import (
 	"github.com/awslabs/karpenter/pkg/cloudprovider/fake"
 	"github.com/awslabs/karpenter/pkg/cloudprovider/registry"
 	"github.com/awslabs/karpenter/pkg/test"
-	v1 "k8s.io/api/core/v1"
-	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 
 	. "github.com/awslabs/karpenter/pkg/test/expectations"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
 func TestAPIs(t *testing.T) {
@@ -57,9 +56,11 @@ var _ = AfterSuite(func() {
 
 var _ = Describe("Termination", func() {
 	var ctx context.Context
+	var node *v1.Node
 
 	BeforeEach(func() {
 		ctx = context.Background()
+		node = test.NodeWith(test.NodeOptions{Finalizers: []string{v1alpha2.KarpenterFinalizer}})
 	})
 
 	AfterEach(func() {
@@ -68,13 +69,6 @@ var _ = Describe("Termination", func() {
 
 	Context("Reconciliation", func() {
 		It("should terminate deleted nodes", func() {
-			node := test.NodeWith(test.NodeOptions{
-				Finalizers: []string{v1alpha2.KarpenterFinalizer},
-				Labels: map[string]string{
-					v1alpha2.ProvisionerNameLabelKey:      "default",
-					v1alpha2.ProvisionerNamespaceLabelKey: "default",
-				},
-			})
 			ExpectCreated(env.Client, node)
 			Expect(env.Client.Delete(ctx, node)).To(Succeed())
 			node = ExpectNodeExists(env.Client, node.Name)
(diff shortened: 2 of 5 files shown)


pkg/controllers/termination/controller.go
```diff
@@ -55,7 +55,7 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 func (c *Controller) Reconcile(ctx context.Context, object client.Object) (reconcile.Result, error) {
 	node := object.(*v1.Node)
 	// 1. Check if node is terminable
-	if node.DeletionTimestamp == nil || !functional.ContainsString(node.Finalizers, provisioning.KarpenterFinalizer) {
+	if node.DeletionTimestamp.IsZero() || !functional.ContainsString(node.Finalizers, provisioning.KarpenterFinalizer) {
```
ellistarn, line 58: How does IsZero work if timestamp is nil? Does the fn handle the case?

njtran: From the function documentation
```
// IsZero returns true if the value is nil or time is zero.
```

pkg/controllers/termination/suite_test.go
```diff
@@ -89,18 +88,83 @@ var _ = Describe("Termination", func() {
 					v1alpha2.ProvisionerNamespaceLabelKey: "default",
 				},
 			})
-			ExpectCreated(env.Client, node)
```
ellistarn, line 92: This test name should probably be generalized

pkg/controllers/termination/suite_test.go
```diff
@@ -89,18 +88,83 @@ var _ = Describe("Termination", func() {
...
-			ExpectCreated(env.Client, test.Pod(test.PodOptions{
+			podEvict := test.Pod(test.PodOptions{
```
ellistarn, line 91: consider inlining {podoptions{nodename}}

pkg/controllers/termination/suite_test.go
```diff
@@ -89,18 +88,83 @@ var _ = Describe("Termination", func() {
...
+				NodeName: node.Name,
+			})
+			podSkip := test.Pod(test.PodOptions{
 				NodeName:    node.Name,
 				Tolerations: []v1.Toleration{{Key: v1.TaintNodeUnschedulable, Operator: v1.TolerationOpExists, Effect: v1.TaintEffectNoSchedule}},
-			}))
+			})
+			ExpectCreated(env.Client, node)
+			ExpectCreated(env.Client, podEvict, podSkip)
```
ellistarn, line 99: This variadic is fancy and can take (env.Client, node, podevict, podskip)

njtran: True!

pkg/controllers/termination/suite_test.go
```diff
@@ -89,18 +88,83 @@ var _ = Describe("Termination", func() {
...
 			Expect(env.Client.Delete(ctx, node)).To(Succeed())
 			node = ExpectNodeExists(env.Client, node.Name)
 			ExpectReconcileSucceeded(controller, node)
-			pods := &v1.PodList{}
-			Expect(env.Client.List(ctx, pods, client.MatchingFields{"spec.nodeName": node.Name})).To(Succeed())
-			Expect(pods.Items).To(HaveLen(1))
+
+			// Expect podToEvict to be evicting, and delete it
+			podEvict = ExpectPodExists(env.Client, podEvict.Name, podEvict.Namespace)
+			Expect(podEvict.GetObjectMeta().GetDeletionTimestamp().IsZero()).To(Equal(false))
```
ellistarn, line 108: To(BeFalse())

pkg/controllers/termination/suite_test.go
```diff
@@ -89,18 +88,83 @@ var _ = Describe("Termination", func() {
...
+			podSkip = ExpectPodExists(env.Client, podSkip.Name, podSkip.Namespace)
+			Expect(podSkip.GetObjectMeta().GetDeletionTimestamp().IsZero()).To(Equal(true))
+
+			// Reconcile to delete node
+			node = ExpectNodeExists(env.Client, node.Name)
+			ExpectReconcileSucceeded(controller, node)
+			ExpectNotFound(env.Client, node)
+		})
+		It("should not terminate nodes that have a do-not-evict pod", func() {
+			node := test.NodeWith(test.NodeOptions{
```
ellistarn, line 120: Can you share this node between tests?

pkg/controllers/termination/suite_test.go
```diff
@@ -89,18 +88,83 @@ var _ = Describe("Termination", func() {
...
+				Finalizers: []string{v1alpha2.KarpenterFinalizer},
+				Labels: map[string]string{
```
ellistarn, line 122: Remind me why we care about these labels?

pkg/controllers/termination/suite_test.go
njtran: Good idea.

pkg/controllers/termination/suite_test.go
```diff
@@ -89,18 +88,83 @@ var _ = Describe("Termination", func() {
...
+				}),
+				test.Pod(test.PodOptions{
+					NodeName:    node.Name,
+					Annotations: map[string]string{v1alpha2.KarpenterDoNotEvictPodAnnotation: "true"},
+				}),
+			}
+
+			ExpectCreated(env.Client, node)
+			for _, pod := range pods {
+				ExpectCreated(env.Client, pod)
```
ellistarn, line 139: Does go type interference not allow ExpectCreated(env.Client, pods...)?

pkg/controllers/termination/suite_test.go
```diff
@@ -89,18 +88,83 @@ var _ = Describe("Termination", func() {
...
+		It("should not terminate nodes that have a do-not-evict pod", func() {
+			node := test.NodeWith(test.NodeOptions{
+				Finalizers: []string{v1alpha2.KarpenterFinalizer},
+				Labels: map[string]string{
+					v1alpha2.ProvisionerNameLabelKey:      "default",
+					v1alpha2.ProvisionerNamespaceLabelKey: "default",
+				},
+			})
+			pods := []*v1.Pod{
+				test.Pod(test.PodOptions{
```
ellistarn, line 128: consider naming these pods similar to the previous test, I think it will clean things up.

pkg/controllers/termination/suite_test.go
njtran: We don't, that was from a previous PR that I was copying. Good point.

pkg/controllers/termination/suite_test.go
njtran: Correct

pkg/controllers/termination/suite_test.go
```diff
@@ -89,18 +88,83 @@ var _ = Describe("Termination", func() {
...
+			Expect(node.Spec.Unschedulable).To(Equal(true))
+			for _, pod := range pods {
+				ExpectPodExists(env.Client, pod.Name, pod.Namespace)
+			}
+
+			// Delete do-not-evict pod
+			ExpectDeleted(env.Client, pods[1])
+
+			// Reconcile node to evict pod
+			node = ExpectNodeExists(env.Client, node.Name)
```
ellistarn, line 157: Do you need to check existence here?

pkg/controllers/termination/terminate.go
```diff
@@ -65,11 +65,19 @@ func (t *Terminator) drain(ctx context.Context, node *v1.Node) (bool, error) {
 	// https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
 	nonCritical := []*v1.Pod{}
 	critical := []*v1.Pod{}
-	for _, pod := range pods {
-		if pod.Spec.PriorityClassName == "system-cluster-critical" || pod.Spec.PriorityClassName == "system-node-critical" {
-			critical = append(critical, pod)
+	for _, p := range pods {
+		if val, ok := p.Annotations[provisioning.KarpenterDoNotEvictPodAnnotation]; ok && val == "true" {
+			zap.S().Debugf("Node %s has a do-not-evict pod %s. Watiing to drain until all do-not-evict pods are terminated.", node.Name, p.Name)
```
ellistarn, line 70: Remind me, we won't try to reconcile again until we get a pod event, right?

ellistarn, line 70: Watiing.

Perhaps
`Waiting to drain node %s,  pod %s has 'do-not-evict' annotation"`

pkg/controllers/termination/terminate.go
```diff
@@ -101,18 +109,13 @@ func (t *Terminator) terminate(ctx context.Context, node *v1.Node) error {
 
 // evictPods returns true if there are no evictable pods
 func (t *Terminator) evictPods(ctx context.Context, pods []*v1.Pod) bool {
-	empty := true
```
ellistarn, line 104: Does this function even need to return true anymore? Shouldn't it return err? We already know if the node is empty before we even call this function, since we can check len(pods) before we call this.

pkg/controllers/termination/suite_test.go
njtran: If termination unexpectedly terminates even with pods still existing, this test could pass. This is to make sure the node doesn't terminate prematurely.

pkg/controllers/termination/terminate.go
njtran: We will reconcile again. We will exponentially back off though.

pkg/controllers/termination/terminate.go
njtran: I think it's a matter of doing a len(check) before, or doing it as the return value. Either way, we still need to check if there was a pod that needs evicting. I added the return `len(pods)==0` so that if we call evictPods, it'll just return that immediately because the for loop would have nothing to iterate through.

With regards to erroring vs bool, I think we decided on an early comment to punt on this to another PR.

pkg/controllers/termination/suite_test.go
```diff
@@ -57,9 +56,13 @@ var _ = AfterSuite(func() {
 
 var _ = Describe("Termination", func() {
 	var ctx context.Context
+	var node *v1.Node
 
 	BeforeEach(func() {
 		ctx = context.Background()
+		node = test.NodeWith(test.NodeOptions{
```
ellistarn, line 63: Consider inlining.

ellistarn (approved):

PR #503, opened 2021-07-02
Implemented pod batching in the allocation/provisioner controller

bwagner5:
**Issue, if available:**
https://github.com/awslabs/karpenter/issues/467

**Description of changes:**
 * This PR removes the interval reconciliation in-favor for watching pod update events and using a batcher (included in this PR) to batch a reconciliation run per provisioner. 
 * The batcher uses a MaxTimeout and an IdleTimeout. I think this will create a good trade-off on reconciling small pod batches more quickly (for example, if someone scales up a deployment by 30, that will most like be seen quickly, within a second or two, the idle time will timeout the window and reconciliation will kick off in 4 seconds rather than waiting the max 10 seconds that is set for the MaxTimeOut in this PR).
 * Allocation controller now runs 4 reconciliation workers (but only once can run per provisioner at a time)
 * Removed Generic controller and converted the other controllers and tests to the new `Register()` method that was already done for the Expiration controller.


ellistarn:
> This PR also adds a new method on the GenericController to configure controllerruntime's concurrent reconciliations per controller. 

FYI, I'm planning on deprecating GenericController, since our controllers have drifted significantly.

pkg/cloudprovider/aws/instance.go
```diff
@@ -103,7 +103,7 @@ func (p *InstanceProvider) Create(ctx context.Context,
 		return nil, fmt.Errorf("expected 1 instance ids, but got %d due to errors %v", count, createFleetOutput.Errors)
 	}
 	if count := len(createFleetOutput.Errors); count > 0 {
-		zap.S().Warnf("CreateFleet encountered %d errors, but still launched instances, %v", count, createFleetOutput.Errors)
+		zap.S().Debugf("CreateFleet encountered %d errors, but still launched instances, %v", count, createFleetOutput.Errors)
```
ellistarn, line 106: Can we minimize the output here? It's super spammy. Maybe just the error codes?

pkg/controllers/allocation/filter.go
```diff
@@ -63,6 +57,35 @@ func (f *Filter) GetProvisionablePods(ctx context.Context, provisioner *v1alpha2
 	return provisionable, nil
 }
 
+// GetProvisionerFor retrieves the provisioner responsible for the pod
+func (f *Filter) GetProvisionerFor(ctx context.Context, p *v1.Pod) (*v1alpha2.Provisioner, error) {
+	provisionerKey := client.ObjectKey{
```
ellistarn, line 62: consider inlining this -- it won't be too long

pkg/controllers/allocation/filter.go
```diff
@@ -63,6 +57,35 @@ func (f *Filter) GetProvisionablePods(ctx context.Context, provisioner *v1alpha2
...
+	}
+	if name, ok := p.Spec.NodeSelector[v1alpha2.ProvisionerNameLabelKey]; ok {
+		provisionerKey.Name = name
+	}
+	if namespace, ok := p.Spec.NodeSelector[v1alpha2.ProvisionerNamespaceLabelKey]; ok {
+		provisionerKey.Namespace = namespace
+	}
+	provisioner := &v1alpha2.Provisioner{}
+	if err := f.kubeClient.Get(ctx, provisionerKey, provisioner); err != nil {
+		return nil, err
```
ellistarn, line 74: Can you make sure that this error message is as crisp as possible. We need to make it absolutely apparent that the default provisioner is required.

ellistarn: Perhaps w/ a IsNotFound check

pkg/controllers/allocation/controller.go
```diff
@@ -105,5 +127,50 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
...
+// Watches returns the necessary information to create a watch
+//   a. source: the resource that is being watched
+//   b. eventHandler: which controller objects to be reconciled
+//   c. predicates: which events can be filtered out before processed
+func (c *Controller) Watches(ctx context.Context) (source.Source, handler.EventHandler, builder.WatchesOption) {
+	return &source.Kind{Type: &v1.Pod{}},
+		handler.EnqueueRequestsFromMapFunc(handler.MapFunc(
+			func(obj client.Object) []reconcile.Request {
+				pod := obj.(*v1.Pod)
+				if pod.Spec.NodeName != "" || podutils.IsOwnedByDaemonSet(pod) {
```
ellistarn, line 142: Can you share these checks if the filter function somehow?

pkg/controllers/allocation/controller.go
```diff
@@ -66,13 +86,15 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 // Reconcile executes an allocation control loop for the resource
 func (c *Controller) Reconcile(ctx context.Context, object client.Object) (reconcile.Result, error) {
 	provisioner := object.(*v1alpha2.Provisioner)
+	c.batcher.Wait(fmt.Sprintf("%s/%s", provisioner.Name, provisioner.Namespace))
```
ellistarn, line 89: Do we need to wait inside the reconcile loop? Is it possible that this will trigger double reconciles? The first event will trigger this and then start waiting. All subsequent events will queue up. Once this wait finishes, the loop will execute and then ctrl will trigger this even due to the queued events.

pkg/controllers/allocation/controller.go
```diff
@@ -66,13 +86,15 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
...
+
 	// 1. Filter pods
 	pods, err := c.filter.GetProvisionablePods(ctx, provisioner)
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("filtering pods, %w", err)
 	}
 	if len(pods) == 0 {
-		return reconcile.Result{RequeueAfter: c.Interval()}, nil
+		return reconcile.Result{}, nil
```
ellistarn, line 97: If a pod is unschedulable for some reason, will we *ever* attempt to reschedule it? What if there's an ICE?

pkg/controllers/allocation/controller.go
```diff
@@ -105,5 +127,50 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
...
+				requests := []reconcile.Request{{
+					NamespacedName: types.NamespacedName{
+						Name:      provisioner.Name,
+						Namespace: provisioner.Namespace,
+					}}}
+				return requests
+			},
+		)),
+		builder.WithPredicates(
+			predicate.Funcs{
```
ellistarn, line 160: I wish there were a less verbose way to express this. Perhaps just inlining the functions? If you leave them null, presumably that will default to true?

pkg/controllers/allocation/controller.go
```diff
@@ -105,5 +127,50 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
...
+				pod := obj.(*v1.Pod)
+				if pod.Spec.NodeName != "" || podutils.IsOwnedByDaemonSet(pod) {
+					return nil
+				}
+				provisioner, err := c.filter.GetProvisionerFor(ctx, pod)
+				if err != nil {
+					zap.S().Errorf("Retrieving provisioner, %s", err.Error())
+					return nil
+				}
+				c.batcher.Add(fmt.Sprintf("%s/%s", provisioner.Name, provisioner.Namespace))
```
ellistarn, line 150: Consider client.ObjectKeyForResource

pkg/controllers/allocation/controller.go
```diff
@@ -105,5 +127,50 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
...
+				requests := []reconcile.Request{{
```
ellistarn, line 151: consider returning inline

pkg/controllers/allocation/controller.go
```diff
@@ -105,5 +127,50 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
...
+	return reconcile.Result{}, nil
+}
+
+// Watches returns the necessary information to create a watch
+//   a. source: the resource that is being watched
+//   b. eventHandler: which controller objects to be reconciled
+//   c. predicates: which events can be filtered out before processed
+func (c *Controller) Watches(ctx context.Context) (source.Source, handler.EventHandler, builder.WatchesOption) {
+	return &source.Kind{Type: &v1.Pod{}},
+		handler.EnqueueRequestsFromMapFunc(handler.MapFunc(
```
ellistarn, line 139: consider breaking apart this function into something like podToProvisioner(...)

pkg/controllers/allocation/controller.go
```diff
@@ -105,5 +127,50 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
...
+				c.batcher.Add(fmt.Sprintf("%s/%s", provisioner.Name, provisioner.Namespace))
+				requests := []reconcile.Request{{
+					NamespacedName: types.NamespacedName{
+						Name:      provisioner.Name,
+						Namespace: provisioner.Namespace,
+					}}}
+				return requests
+			},
+		)),
+		builder.WithPredicates(
```
ellistarn, line 159: I think you can omit these predicates by default if you want to.

pkg/controllers/allocation/controller.go
```diff
@@ -23,17 +23,31 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	MaxBatchWindow   = 10 * time.Second
```
ellistarn, line 42: Thoughts on putting these fields onto the provisioner?

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,155 @@
...
+	// If the channel is blocked, then it's already processing a window, so send an update signal
+	case b.updates <- true:
+	// If both are blocked, then no need to start or send an update signal
+	default:
+	}
+}
+
+// Wait blocks until a batching window completes
+// If the batch is empty, it will block until something is added or the window completes
+func (b *Batch) Wait() {
```
ellistarn, line 115: consider Complete()

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"github.com/patrickmn/go-cache"
+)
+
+const (
+	BatchCacheTTL             = 1 * time.Hour
+	BatchCacheCleanupInterval = 2 * time.Hour
+)
+
+type Batcher struct {
```
ellistarn, line 29: Thoughts on testing this?

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,155 @@
...
+	// It should be a smaller duration than MaxBatchPeriod
+	IdlePeriod time.Duration
+
+	// batches keeps a mapping of a key (like a provisioner name and namespace) to a specific Batch
+	batches *cache.Cache
+}
+
+// NewBatchManager creates a new batch manager to start multiple batch windows
+func NewBatcher(maxBatchPeriod time.Duration, idlePeriod time.Duration) *Batcher {
+	batchCache := cache.New(BatchCacheTTL, BatchCacheCleanupInterval)
```
ellistarn, line 42: consider name equality w/ your return variable (e.g. batches) or with your New function (cache)

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,155 @@
...
+	batchCache.OnEvicted(func(key string, val interface{}) {
+		batch := val.(*Batch)
+		batch.close()
```
ellistarn, line 45: Is it possible this could be called and trigger a npe (mentioned below)?

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,155 @@
...
+	})
+	return &Batcher{
+		MaxBatchPeriod: maxBatchPeriod,
+		IdlePeriod:     idlePeriod,
+		batches:        batchCache,
+	}
+}
+
+// Add starts a batching window or updates an existing one based on a batching key
+func (m *Batcher) Add(key string) {
```
ellistarn, line 55: Consider making these interfaces runtime.Object, and constructing the keys inside the implementation.

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,155 @@
...
+// NewBatchManager creates a new batch manager to start multiple batch windows
+func NewBatcher(maxBatchPeriod time.Duration, idlePeriod time.Duration) *Batcher {
+	batchCache := cache.New(BatchCacheTTL, BatchCacheCleanupInterval)
+	batchCache.OnEvicted(func(key string, val interface{}) {
+		batch := val.(*Batch)
+		batch.close()
+	})
+	return &Batcher{
+		MaxBatchPeriod: maxBatchPeriod,
+		IdlePeriod:     idlePeriod,
```
ellistarn, line 49: These should probably be passed in on NewBatch() so that it isn't globally configured

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,155 @@
...
+type Batch struct {
+	// start is a channel to coordinate window starts
+	start chan bool
+	// updates is a channel to communicate progress in filling a batch window
+	updates chan bool
+	// end is a channel to signal a window close
+	end chan bool
+
+	sync.Mutex
+	*Batcher
```
ellistarn, line 85: Is it possible to make this not rely on Batcher?

pkg/cloudprovider/aws/instance.go
bwagner5: I'm concerned that minimizing too much will cause clutter without any useful information.  The whole Errors response is quite confusing in general. It might actually be safe to check if the desired number of instances were launched and only print errors if there is a mismatch. That would reduce the logs here to only situations where the resulting errors had a meaningful impact. If we only show error codes, we might just see a bunch of 
"InsufficientCapacity", which doesn't tell you how to fix anything. Wdyt about starting with reducing and if we still find them noisy we can do some more firm reduction?

ellistarn: Sounds good!

pkg/controllers/allocation/controller.go
bwagner5: It shouldn't based on the docs (https://pkg.go.dev/k8s.io/client-go@v0.21.2/util/workqueue, since it's "Stingy") and from real world tests, it does not reprocess. I think I need to dig into this more to see exactly why this is the case though. I'm not entirely sure how it would know if a pod is queued up after we list pods in a reconcile.

pkg/controllers/allocation/batch.go
bwagner5: I have some basic unit tests for this that I can check in. We'll have some confidence that they work since this will be utilized in the suite tests.  Did you have any specific automated tests in mind?

pkg/controllers/allocation/batch.go
bwagner5: Yep, I'll move the config down to the Batch and that should break this unnecessary dependency.

pkg/controllers/allocation/controller.go
bwagner5: At first, I was thinking that would be a great idea to do. However, since we have two fields to adjust, I think we might be able to configure a pretty good default. Either way, I think I'd opt for "not yet" and see if it's needed.

pkg/controllers/allocation/controller.go
ellistarn: There are some nice benefits of putting this in the reconcile loop itself, namely simplicity/linearity of logic. In general, I much prefer to put filtering logic in here, rather than in the Watch()

bwagner5: Yeah, agreed, putting it in the Watch is basically a non-starter since it's completely synchronous.  We wouldn't be able to do provisioner based batching as efficiently.

pkg/controllers/allocation/filter.go
bwagner5: I assume you're referring to the `provisionerKey` default/default? I did it this way to make the override logic work more cleanly. I might not be seeing the alternative you're suggesting here, but if I in-line, then they'll need to be two `Get` calls and it'll get ugly IMO.

pkg/controllers/allocation/filter.go
bwagner5: The default provisioner is not required if every pod gets provisioner labels. Now I'm kind of thinking that's why you wanted the provisioner in-lined to make these two error messages distinct? Am I warm?

pkg/controllers/allocation/controller.go
bwagner5: actually, I think I can delete this all together since `isProvisionable` is already checked in the UpdateFunc predicate and that should cover the nodeName being nil since it checks if the pod failed to schedule.

pkg/controllers/allocation/batch.go
ellistarn: This won't be utilized by suite tests. We call reconcile directly now. Thinking on this, the batching window will massively slow our tests down :P

pkg/controllers/allocation/controller.go
ellistarn: Sounds good.

pkg/controllers/allocation/controller.go
ellistarn: I think we might be able to scrounge up the event queue somehow and pass it in. Then we could enqueue. Probably cleaner this way though.

pkg/controllers/allocation/controller.go
ellistarn: Thought on inverting this decision and putting it all in the map func, rather than the predicate func? 

I think controller-runtime's programming model is way too complex here.

pkg/controllers/allocation/filter.go
ellistarn:
```
provisionerKey := client.ObjectKey{ Name: "default", Namespace: "default" }
```

pkg/controllers/allocation/controller.go
bwagner5: In the case of capacity failing to come up for any reason, the Reconcile returns an error and the controller will retry based on the work-queue options `ExponentialFailureRateLimiter`. It looks like we're not returning an error in the aws cloudprovider `Create` if partial capacity is fulfilled.  I think that needs to be changed and looked at more closely.

pkg/controllers/allocation/filter.go
bwagner5: over thought that one 👍   will do

pkg/controllers/allocation/filter.go
bwagner5: not warm, but took a stab at a clear error message, lmk what you think

ellistarn: We're saying different things. I'm thinking specifically of the cases:

1. If a provisioner doesn't exist, we should state, loudly, "provisioner does not exist, please specify karpenter.sh/provisioner-name ... or create a default provisioner in the default namespace" 
2. Auth error, use normal error handling logic.

pkg/controllers/allocation/controller.go
bwagner5: yeah, that's probably best, I'll move it

pkg/controllers/allocation/controller.go
bwagner5: Since I moved the logic to check if we should add to a batch to the mapper, I just went back to processing all events to make it cleaner.

pkg/controllers/allocation/batch.go
bwagner5: unless we configure them to be microsecond Timeouts. That's what I did on my little unit tests

pkg/controllers/allocation/batch.go
bwagner5: Do we want these to not be globally configured?  A batch is created dynamically and the construct is really a `Batcher` on a resource type and Batches are for individual resources, I think it's simplest to share the config here. You can even have a single batch Batcher.  

If we decide later to configure batches differently per Provisioner, then I guess we could add the config as a parameter to `Add(key string, maxBatchPeriod time.Duration, idlePeriod time.Duration)` ?   

Right now, `Batch` would not be used standalone.  Wdut about holding off and keeping batches configures based on the parent batcher?

pkg/controllers/allocation/batch.go
bwagner5: I'm rethinking things now... see my above comment

pkg/controllers/allocation/batch.go
JacobGabrielson: Could you add docs here on whether this is concurrency-safe?

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// Add starts a batching window or adds to an existing in-progress window
+func (b *Batch) Add() {
+	b.Lock()
```
JacobGabrielson, line 95: Are there some higher-level abstractions you can use here instead of raw Lock? (WaitGroups or something else?) I haven't really thought it true but this kind of thing can be deceptively tricky.

pkg/controllers/allocation/batch.go
ellistarn: I think you can simply fake time in golang. Happy to discuss offline.

pkg/controllers/allocation/batch.go
ellistarn: IMO, we should still break the dependency and inject these values into the batch.

pkg/controllers/allocation/controller.go
```diff
@@ -66,13 +84,15 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 // Reconcile executes an allocation control loop for the resource
 func (c *Controller) Reconcile(ctx context.Context, object client.Object) (reconcile.Result, error) {
 	provisioner := object.(*v1alpha2.Provisioner)
+	c.batcher.Complete(fmt.Sprintf("%s/%s", provisioner.Name, provisioner.Namespace))
```
ellistarn, line 87: Ahh I had half written a comment to rename this to Complete and then thinking further on it, I deleted the comment because I liked the explicit nature of the "Wait()" semantic. Sorry for the thrash!

ellistarn: Similarly to the above comment on runtime.Object, batcher.Wait(provisioner)

pkg/controllers/allocation/controller.go
ellistarn: We currently have a semantic of "continue after" that needs to be address with this change. Alternatively, you can keep an interval in (in addition to batching) and address it in a separate PR.

pkg/controllers/types.go
```diff
@@ -31,6 +34,13 @@ type Controller interface {
 	// For returns a default instantiation of the resource and is injected by
 	// data from the API Server at the start of the reconciliation loop.
 	For() client.Object
+	// Watches returns the necessary information to create a watch
+	//   a. Source: the resource that is being watched
+	//   b. EventHandler: which controller objects to be reconciled
+	//   c. WatchesOption: which events can be filtered out before processed
+	Watches(context.Context) (source.Source, handler.EventHandler, builder.WatchesOption)
```
ellistarn, line 41: Reminder to rip these out.

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,150 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"github.com/patrickmn/go-cache"
+)
+
+const (
+	BatchCacheTTL             = 1 * time.Hour
```
JacobGabrielson, line 25: Why do we need to evict from the cache? Is it just because if the Provisioner goes away then you no longer need the batcher? (Seems like there might be a cleaner way to do that?)

pkg/controllers/allocation/controller.go
bwagner5: ha no worries, I'm not too attached to the names, so I'll change back. The inspiration was really WaitGroups

pkg/controllers/allocation/batch.go
bwagner5: Yes, I didn't want to keep them around forever if there was a long running installation with a lot of provisioners (created and delete), it could be storing a lot of batches. I don't think this would be a huge problem in reality, so I opted for the cache as a free-ish way to do the evictions based on time not used.   

An alternative would be to process Provisioner deletions in the reconcile and remove the batch for the provisioner. That would be really simple and explicit.

pkg/controllers/allocation/batch.go
bwagner5: I don't think that is possible... Batches are constructed lazily through the constructor which allocates the memory for the channel. If a batch is being evicted from the cache then it's guaranteed to have been initialized at some point.  Is there something I'm not seeing here?  I can add the nil checks to the `close` if you just don't want to take a chance of something odd happening.

pkg/controllers/allocation/batch.go
JacobGabrielson: After thinking about it some more, I would suggest something along these lines ... 

1. Have a single background goroutine that wakes up every N seconds (it will be used by all Provisioners)
2. Provisioners start new batches by sending messages to it on a channel
3. Every time the Batcher wakes up it goes thru all the currently active batches and does <<whatever>> with them (fire them off if they're ready, extend if not, deal with additions, yadda yadda)
4. Just punt on cleanup - a single goroutine in the bg isn't a big deal

pkg/controllers/allocation/batch.go
bwagner5: I could use a channel (first draft had a channel) or a WaitGroup, but honestly I think the lock in this case is the easiest to reason about. The channel seemed a bit convoluted to think and look at.

pkg/controllers/allocation/controller.go
bwagner5: Yeah I'll probably add back a requeue and address in a separate PR

pkg/controllers/allocation/controller.go
bwagner5: I'm not sure where this is? Did you mean `client.ObjectKeyFromObject`?  Either way, I changed the API to accept a client.Object and am pulling off name/namespace within the Batcher now.

pkg/controllers/allocation/controller.go
bwagner5: Do you think this is still needed after the simplifications?

pkg/controllers/allocation/batch.go
JacobGabrielson: I think it's more around calling `close` when someone's still using the channels (in the background?) https://go101.org/article/channel-closing.html type stuff ...

pkg/controllers/allocation/batch.go
JacobGabrielson: See my other comment around how to simplify this - I'm not sure any synchronization is really needed.

pkg/controllers/types.go
```diff
@@ -31,6 +34,13 @@ type Controller interface {
...
+	// ConcurrentReconciles controls the number of concurrent reconciles that can occur
```
JacobGabrielson, line 42: Could you add something around the Why here? I'm wondering what the intent of this is? (Maybe add a very simple representative example?)

pkg/controllers/allocation/controller.go
ellistarn: Is it possible to use a runtime.Object and avoid the controller-runtime dependency? It's an inner type of client.Object anyways.

bwagner5: okay, I can use `k8s.io/apimachinery/pkg/apis/meta/v1/Object` I don't think I can use runtime.Object though..

pkg/controllers/types.go
bwagner5: This is being removed

pkg/controllers/allocation/controller.go
ellistarn: I always confused those two.

pkg/controllers/allocation/batch.go
bwagner5: 👍  I refactored to get rid of the locks , let me know what you think of it.

pkg/controllers/allocation/controller.go
bwagner5: nvm, I see after look at #498

cmd/controller/main.go
```diff
@@ -37,6 +38,11 @@ import (
 	controllerruntimezap "sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
+const (
+	maxBatchWindow   = 10 * time.Second
```
ellistarn, line 42: Can we hide these values inside the allocation.NewController function and just manually instantiate the struct for our testing? I think main shouldn't have this awareness.

pkg/cloudprovider/aws/instance.go
```diff
@@ -102,9 +101,6 @@ func (p *InstanceProvider) Create(ctx context.Context,
 	if count := len(createFleetOutput.Instances[0].InstanceIds); count != 1 {
 		return nil, fmt.Errorf("expected 1 instance ids, but got %d due to errors %v", count, createFleetOutput.Errors)
 	}
-	if count := len(createFleetOutput.Errors); count > 0 {
-		zap.S().Warnf("CreateFleet encountered %d errors, but still launched instances, %v", count, createFleetOutput.Errors)
```
ellistarn, line 106: Thoughts on keeping this log line but just not printing the actual errors? Then we know it's happening, but it's not spammy.

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,178 @@
...
+
+const (
+	opAdd  = "add"
+	opWait = "wait"
+)
+
+// Batcher is a batch manager for multiple objects
+type Batcher struct {
+	// MaxBatchPeriod is the maximum amount of time to batch incoming pods before flushing
+	MaxBatchPeriod time.Duration
```
ellistarn, line 33: Consider just calling this MaxPeriod since it's already contextualized by Batcher

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,178 @@
...
+	windows map[string]*window
+	// ops is a stream of add and wait operations on a batch window
+	ops chan *batchOp
+	// isMonitorRunning indicates if the monitor go routine has been started
+	isMonitorRunning bool
+}
+
+type batchOp struct {
+	kind    string
+	key     string
```
ellistarn, line 48: Why do we need both kind and key?

pkg/controllers/allocation/controller.go
```diff
@@ -92,18 +106,60 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
...
+	return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: provisioner.Name, Namespace: provisioner.Namespace}}}
+}
+
+func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
+	err := controllerruntime.
+		NewControllerManagedBy(m).
+		Named("Allocation").
+		For(&v1alpha2.Provisioner{}).
+		Watches(
+			// Reconcile all pods related to a provisioner when it changes.
```
ellistarn, line 148: This comment is useless, imo.

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,178 @@
...
+// If the batch window has ended, then the batch closed channel will be notified and the window will be removed
+func (b *Batcher) checkForWindowEndAndNotify(key string, window *window) {
+	if time.Since(window.lastUpdated) < b.IdlePeriod && time.Since(window.started) < b.MaxBatchPeriod {
+		return
+	}
+	b.endWindow(key, window)
+}
+
+// endWindow signals the end of a window to all wait consumers and deletes the window
+func (b *Batcher) endWindow(key string, window *window) {
```
ellistarn, line 148: Thought's on calling this close?

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,178 @@
...
+			close(end)
+		default:
+		}
+	}
+	delete(b.windows, key)
+}
+
+// startOrUpdateWindow starts a new window for the object key if one does not already exist
+// if a window already exists for the object key, then the lastUpdate time is set
+func (b *Batcher) startOrUpdateWindow(key string) *window {
```
ellistarn, line 161: Consider `continue`

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,178 @@
...
+	batchWindow.lastUpdated = time.Now()
+	if batchWindow.started.IsZero() {
+		batchWindow.started = time.Now()
+	}
+	return batchWindow
+}
+
+// keyFrom takes an object and outputs a unique key
+func (b *Batcher) keyFrom(obj metav1.Object) string {
+	return fmt.Sprintf("%s/%s", obj.GetName(), obj.GetNamespace())
```
ellistarn, line 177: Thoughts on using selflink or id instead? This will avoid some niche collision scenarios.

pkg/controllers/allocation/controller.go
```diff
@@ -23,66 +23,80 @@ import (
...
-
-func (c *Controller) Name() string {
-	return "provisioner/allocator"
+	kubeClient    client.Client
 }
 
 // NewController constructs a controller instance
-func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.CloudProvider) *Controller {
+func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.CloudProvider,
+	MaxBatchWindow time.Duration, BatchIdleTimeout time.Duration) *Controller {
```
ellistarn, line 57: nit, function parameters shouldn't use capital letters `MaxBatchWindow`, but IMO you should keep them capital and just make them file constants.

pkg/controllers/allocation/controller.go
```diff
@@ -92,18 +106,60 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
...
 		return reconcile.Result{}, fmt.Errorf("creating capacity, %w", err)
 	}
 
-	// 5. Bind pods to nodes
+	// 7. Bind pods to nodes
+	var errs error
 	for _, packedNode := range packedNodes {
 		zap.S().Infof("Binding pods %v to node %s", apiobject.PodNamespacedNames(packedNode.Pods), packedNode.Node.Name)
 		if err := c.binder.Bind(ctx, packedNode.Node, packedNode.Pods); err != nil {
 			zap.S().Errorf("Continuing after failing to bind, %s", err.Error())
```
ellistarn, line 120: I think we can drop this log line now, as it will show up at the end.

pkg/controllers/allocation/controller.go
```diff
@@ -92,18 +106,60 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
...
+	err := controllerruntime.
+		NewControllerManagedBy(m).
+		Named("Allocation").
+		For(&v1alpha2.Provisioner{}).
+		Watches(
+			// Reconcile all pods related to a provisioner when it changes.
+			&source.Kind{Type: &v1.Pod{}},
+			handler.EnqueueRequestsFromMapFunc(c.podToProvisioner),
+		).
+		WithOptions(
```
ellistarn, line 152: Remind me why we can't use controller-runtime's defaults? Is it too fast?

pkg/controllers/allocation/filter.go
```diff
@@ -63,6 +58,36 @@ func (f *Filter) GetProvisionablePods(ctx context.Context, provisioner *v1alpha2
 	return provisionable, nil
 }
 
+// GetProvisionerFor retrieves the provisioner responsible for the pod
+func (f *Filter) GetProvisionerFor(ctx context.Context, p *v1.Pod) (*v1alpha2.Provisioner, error) {
```
ellistarn, line 62: I might move this up a level to controller, since that's the only place its used and the cohesion with the rest of the filtering logic feels off.

pkg/controllers/allocation/filter.go
```diff
@@ -63,6 +58,36 @@ func (f *Filter) GetProvisionablePods(ctx context.Context, provisioner *v1alpha2
...
+		provisionerKey.Name = name
+	}
+	if namespace, ok := p.Spec.NodeSelector[v1alpha2.ProvisionerNamespaceLabelKey]; ok {
+		provisionerKey.Namespace = namespace
+	}
+	provisioner := &v1alpha2.Provisioner{}
+	if err := f.kubeClient.Get(ctx, provisionerKey, provisioner); err != nil {
+		if errors.IsNotFound(err) {
+			return nil, fmt.Errorf("provisioner %s/%s does not exist, please specify a provisioner override or create the default provisioner",
+				provisionerKey.Name, provisionerKey.Namespace)
```
ellistarn, line 74: I think that %s for provisionerKey will format this the same way.

pkg/controllers/allocation/filter.go
```diff
@@ -63,6 +58,36 @@ func (f *Filter) GetProvisionablePods(ctx context.Context, provisioner *v1alpha2
...
+	if name, ok := p.Spec.NodeSelector[v1alpha2.ProvisionerNameLabelKey]; ok {
+		provisionerKey.Name = name
+	}
+	if namespace, ok := p.Spec.NodeSelector[v1alpha2.ProvisionerNamespaceLabelKey]; ok {
+		provisionerKey.Namespace = namespace
+	}
+	provisioner := &v1alpha2.Provisioner{}
+	if err := f.kubeClient.Get(ctx, provisionerKey, provisioner); err != nil {
+		if errors.IsNotFound(err) {
+			return nil, fmt.Errorf("provisioner %s/%s does not exist, please specify a provisioner override or create the default provisioner",
```
ellistarn, line 73: "create a default provisioner, or specify an alternative using the nodeSelector `karpenter.sh/provisioner-name`

pkg/controllers/allocation/suite_test.go
```diff
@@ -12,17 +12,19 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package allocation
+package allocation_test
```
ellistarn, line 15: Why change up the package here? I know it's a testing style, but it's outside of the convention of the project thus far. Should we be moving to adopt this pattern? I find it a bit of a surprising convention of golang and I haven't yet seen the value over just using the same package.

cmd/controller/main.go
bwagner5: Yeah, I think that can work.

pkg/test/expectations/expectations.go
```diff
@@ -105,42 +104,30 @@ func ExpectCleanedUp(c client.Client) {
...
+	ExpectReconcileFailed(reconciler, client.ObjectKeyFromObject(provisioner))
 	result := []*v1.Pod{}
 	for _, pod := range pods {
 		result = append(result, ExpectPodExists(c, pod.GetName(), pod.GetNamespace()))
 	}
 	return result
 }
 
-// ExpectControllerFailed should be deprecated in favor of ExpectReconcileFailed (TODO, once we move off of generic controller)
-func ExpectControllerFailed(controller controllers.Controller, object client.Object) {
```
ellistarn, line 133: :woot

ellistarn (commented):
Nicely done on many fronts. I'm a bit wary of the complexity of the batcher, but I think it's necessary complexity. I'd love to see a comprehensive suite of tests exercising the batcher, but feel free to implement in a followon PR. 

Otherwise just a couple naming nits here and there. Most of the batcher naming nits are just to help me wrap my mind around it.

pkg/cloudprovider/aws/instance.go
bwagner5: I don't think this is necessary especially if we're not printing Errors. We will print an error if the instance doesn't get launched which is what we really care about.   If we print something when an error occurs but we launch capacity, that may be a perfectly valid request w/ no issues but causes a red herring to the operator which is difficult to do anything about without the errors anyways.

pkg/controllers/allocation/batch.go
bwagner5: It used to be close, but it felt awkward to have a window and close and channels closing (since this is private to the Batcher object).

pkg/controllers/allocation/batch.go
bwagner5: At first, I thought you meant the continue keyword, and I got worried that your github account had been compromised ;)  I can change to `startOrContinueWindow` 👍

pkg/controllers/allocation/controller.go
bwagner5: Not sure, @njtran had tuned these recently, so I just grabbed those params. Do you know off-hand @njtran ?

cmd/controller/main.go
bwagner5: Oh actually, we'd have to make the Controller fields public for that to work, since the aws cloudprovider suite tests are not in the same package.

ellistarn: Public sounds fine to me. Better than forcing this awareness to this file.

pkg/controllers/allocation/batch.go
ellistarn: Let's revisit some of this naming at some point once I'm more familiar with this code.

pkg/controllers/allocation/batch.go
ellistarn: I was just thinking of a word that meant start or update. CreateOrUpdate is typically "Ensure". StartOrUpdate seemed a natural fit for Continue, or ContinueWindow. Maybe Refresh? As mentioned above, let's look at naming later.

pkg/controllers/allocation/suite_test.go
bwagner5: it's a protection against using non-public fields. For example, when I changed the AllocationController to have public fields so we can configure the batcher in the tests, it caught it in it's own suite test which was more noticeable since I was working in that package.  I actually thought about changing it back without the "_test", but the cloudprovider suite tests also needs to create a controller, so it really should behave the same. 

In general, I think we should absolutely stick with this convention.

pkg/controllers/allocation/batch.go
bwagner5: selflink is deprecated and is removed in 1.21 it looks like, but I think you're right that UID is probably a better choice here 👍

pkg/controllers/allocation/controller.go
```diff
@@ -23,97 +23,184 @@ import (
...
+			},
+		).
+		Complete(c)
+	c.Batcher.Start(ctx)
+	return err
+}
+
+// retrieveProvisionerFrom fetches the provisioner and returns a raw provisioner that was persisted in the api server
+// and a provisioner w/ default runtime values added that should not be persisted
+func (c *Controller) retrieveProvisionerFrom(ctx context.Context, req reconcile.Request) (*v1alpha2.Provisioner, *v1alpha2.Provisioner, error) {
```
ellistarn, line 154: I'm a fan of the convention `provisionerFor(ctx, req)`

ellistarn (approved):

pkg/controllers/allocation/batch.go
```diff
@@ -0,0 +1,173 @@
...
+	opAdd  = "add"
+	opWait = "wait"
+)
+
+// Batcher is a batch manager for multiple objects
+type Batcher struct {
+	// MaxPeriod is the maximum amount of time to batch incoming pods before flushing
+	MaxPeriod time.Duration
+	// IdlePeriod is the amount of time to wait to flush a batch when there are no incoming pods but the batch is not empty
+	// It should be a smaller duration than MaxBatchPeriod
```
njtran, line 35: MaxBatchPeriod -> MaxPeriod?

bwagner5: nice catch! I did a proper find-all and replace this time :)

ellistarn (approved):

PR #506, opened 2021-07-06
Update AWS Getting Started guide with additional explanations 

geoffcline:
**Description of changes:**
Revise https://github.com/awslabs/karpenter/tree/main/docs/aws for a more launch ready doc. 

### DIFF
--- a/docs/aws/README.md
+++ b/docs/aws/README.md
@@ -1,16 +1,23 @@
 # Getting Started with Karpenter on AWS
 
-Karpenter automatically provisions new nodes in response to unschedulable  pods. Karpenter does this by observing events within the kubernetes cluster, and then sending commands to the underlying cloud provider. 
+Karpenter automatically provisions new nodes in response to unschedulable
+pods. Karpenter does this by observing events within the kubernetes cluster,
+and then sending commands to the underlying cloud provider. 
 
-In this example, the cluster is running on Amazon Web Services (AWS) Elastic Kubernetes Service (EKS). Karpenter is designed to be cloud provider agnostic, but currently only supports AWS. Contributions are welcomed. 
+In this example, the cluster is running on Amazon Web Services (AWS) Elastic
+Kubernetes Service (EKS). Karpenter is designed to be cloud provider agnostic,
+but currently only supports AWS. Contributions are welcomed. 
 
-This guide should take less than 1 hour to complete, and cost less than $0.25. Follow the clean-up instructions to reduce any charges.
+This guide should take less than 1 hour to complete, and cost less than $0.25.
+Follow the clean-up instructions to reduce any charges.
 
 # Install
 
-Karpenter is installed in clusters with a simple helm chart.
+Karpenter is installed in clusters with a helm chart.
 
-Karpenter additionally requires IAM Roles for Service Accounts (IRSA). IRSA permits Karpenter (within the cluster) to make privileged requests to AWS (as the cloud provider). 
+Karpenter additionally requires IAM Roles for Service Accounts (IRSA). IRSA
+permits Karpenter (within the cluster) to make privileged requests to AWS (as
+the cloud provider). 
 
 ## Required Utilities
 
@@ -20,11 +27,13 @@ Install these tools before proceeding:
 2. `kubectl` - [the kubernetes CLI](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/)
 3. `eksctl` - [the CLI for AWS EKS](https://docs.aws.amazon.com/eks/latest/userguide/eksctl.html)
 
-Login to the AWS CLI with a user that has sufficient privileges to create a cluster. 
+Login to the AWS CLI with a user that has sufficient privileges to create a
+cluster. 
 
 ## Environment Variables
 
-After setting up the tools, set the following environment variables to store commonly used values. 
+After setting up the tools, set the following environment variables to store
+commonly used values. 
 
 ```bash
 export CLUSTER_NAME=$USER-karpenter-demo
@@ -37,19 +46,32 @@ KARPENTER_VERSION=$(curl -fsSL \
 
 ## Create a Cluster
 
-Create a cluster with `eksctl`. The below configuration file specifies a basic cluster (name, region), an IAM role for karpenter to use, and two fargate profiles. 
+Create a cluster with `eksctl`. The below configuration file specifies a basic
+cluster (name, region), an IAM role for Karpenter to use, and two fargate
+profiles. 
 
-The two fargate profiles host `kube-system` and the karpenter service itself. This permits karpenter to manage all nodes, without a circular dependency. 
+The two fargate profiles host `kube-system` and the Karpenter service itself.
+This permits Karpenter to manage all EC2 instances. 
+
+Karpenter itself can run anywhere, including on self-managed node groups,
+[managed node groups](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html), 
+or [AWS Fargate](https://aws.amazon.com/fargate/).
 
 Karpenter will provision traditional instances on EC2. 
 
-Additionally, the configuration file sets up an [OIDC provider](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens), necessary for IRSA (see below). Kubernetes supports OIDC as a standardized way of communicating with identity providers. 
+Additionally, the configuration file sets up an [OIDC
+provider](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens),
+necessary for IRSA (see below). Kubernetes supports OIDC as a standardized way
+of communicating with identity providers. 
 
 ## Setup Authentication from Kubernetes to AWS (IRSA)
 
-IAM Roles for Service Accounts (IRSA) maps kubernetes resources to roles (permission sets) on AWS. 
+IAM Roles for Service Accounts (IRSA) maps kubernetes resources to roles
+(permission sets) on AWS. 
 
-First, define a role using the template below. It provides full access to EC2, and limited access to other services such as EKS and Elastic Container Registry (ECR).
+First, define a role using the template below. It provides full access to EC2,
+and limited access to other services such as EKS and Elastic Container Registry
+(ECR).
 
 ```bash
 # Creates IAM resources used by Karpenter
@@ -65,7 +87,7 @@ curl -fsSL https://raw.githubusercontent.com/awslabs/karpenter/"${KARPENTER_VERS
 Second, create the mapping between kubernetes resources and the new IAM role. 
 
 ```bash
-# Add the karpenter node role to your aws-auth configmap, allowing nodes with this role to connect to the cluster.
+# Add the Karpenter node role to your aws-auth configmap, allowing nodes with this role to connect to the cluster.
 eksctl create iamidentitymapping \
   --username system:node:{{EC2PrivateDNSName}} \
   --cluster  ${CLUSTER_NAME} \
@@ -80,7 +102,8 @@ Now, Karpenter can send requests for new EC2 instances to AWS.
 
 Use helm to deploy Karpenter to the cluster. 
 
-We created a kubernetes service account when we created the cluster using eksctl. Thus, we don't need the helm chart to do that.
+We created a kubernetes service account when we created the cluster using
+eksctl. Thus, we don't need the helm chart to do that.
 
 ```bash
 helm repo add karpenter https://awslabs.github.io/karpenter/charts
@@ -91,11 +114,17 @@ helm upgrade --install karpenter karpenter/karpenter \
 
 ## Provisioner
 
-A single karpenter provisioner is capable of handling many different pod shapes. In other words, karpenter eliminates the need to manage many different node groups. Karpenter makes scheduling and provisioning decisions based on pod attributes such as labels and affinity. 
+A single Karpenter provisioner is capable of handling many different pod
+shapes. In other words, Karpenter eliminates the need to manage many different
+node groups. Karpenter makes scheduling and provisioning decisions based on pod
+attributes such as labels and affinity. 
 
-Create a simple default provisioner using the command below. This provisioner provides instances with the default certificate bundle, and the control plane endpoint url. 
+Create a simple default provisioner using the command below. This provisioner
+provides instances with the default certificate bundle, and the control plane
+endpoint url. 
 
-Importantly, the `ttlSecondsAfterEmpty` value configures karpenter to deprovision empty nodes. 
+Importantly, the `ttlSecondsAfterEmpty` value configures Karpenter to
+deprovision empty nodes. 
 
 ```bash
 cat <<EOF | kubectl apply -f -
@@ -115,11 +144,12 @@ kubectl get provisioner default -o yaml
 
 # First Use
 
-Karpenter is now active and ready to begin provisioning nodes. Create a workload (e.g., deployment) to see karpenter provision some nodes. 
+Karpenter is now active and ready to begin provisioning nodes. Create a
+workload (e.g., deployment) to see Karpenter provision some nodes. 
 
 ## Create Workload
 
-This deployment is based on the pause image and initially has zero replicas. 
+This deployment uses the [pause image](https://www.ianlewis.org/en/almighty-pause-container) and starts with zero replicas. 
 
 ```bash
 cat <<EOF | kubectl apply -f -
@@ -147,25 +177,36 @@ EOF
 ```
 ## Automatic Node Provisioning 
 
-Scale the previous deployment to 5 replicas, and begin watching the karpenter pod for log events. 
+Scale the previous deployment to 5 replicas, and begin watching the Karpenter
+pod for log events. 
 
 ```bash
 kubectl scale deployment inflate --replicas 5
 kubectl logs -f -n karpenter $(kubectl get pods -n karpenter -l karpenter=controller -o name)
 ```
 
-## Automatic Node Deprovisioning
+## Automatic Node Deprovisioning (timeout)
 
-Now, delete the deployment. After 30 seconds (`ttlSecondsAfterEmpty`), karpenter should deprovision the now empty nodes. 
+Now, delete the deployment. After 30 seconds (`ttlSecondsAfterEmpty`),
+Karpenter should deprovision the now empty nodes. 
 
 ```bash
 kubectl delete deployment inflate
 kubectl logs -f -n karpenter $(kubectl get pods -n karpenter -l karpenter=controller -o name)
 ```
 
+## Manual Node Deprovisioning
+
+If you delete a node with kubectl, Karpenter deprovisions the corresponding instance.
+
+```bash
+kubectl delete node $NODE_NAME
+```
+
 # Cleanup
 
-It's important to both delete the cluster instances and the cluster control plane. AWS charges for both. 
+It's important to both delete the cluster instances and the cluster control
+plane. AWS charges for both. 
 
 Delete cluster workloads and instances: 
 
--- a/docs/aws/README.md
+++ b/docs/aws/README.md
@@ -1,11 +1,31 @@
-# AWS
+# Getting Started with Karpenter on AWS
 
-This guide will provide a complete Karpenter installation for AWS.
-These steps are opinionated and may need to be adapted for your use case.
+Karpenter automatically provisions new nodes in response to unscheduleable pods. Karpenter does this by observing events within the kubernetes cluster, and then sending commands to the underlying cloud platform. 
 
-> This guide should take less than 1 hour to complete and cost less than $.25
+In this example, the cluster is running on Amazon Web Services (AWS) Elastic Kubernetes Service (EKS). Karpenter is designed to be cloud platform independent, but currently only supports AWS. Contributions are welcomed. 
 
-## Environment
+This guide should take less than 1 hour to complate, and cost less than $0.25. Follow the clean-up instructions to reduce any charges.
+
+# Install
+
+Karpenter is installed in clusters with a simple helm chart.
+
+Karpenter additionally requires IAM Roles for Service Accounts (IRSA). IRSA permits Karpenter (within the cluster) to make privlidged requests to AWS (as the cloud platform). 
+
+## Required Utilities
+
+Install these tools before proceeding:
+
+1. [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-linux.html)
+2. `kubectl` - [the kubernetes CLI](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/)
+3. `eksctl` - [the CLI for AWS EKS](https://docs.aws.amazon.com/eks/latest/userguide/eksctl.html)
+
+Login to the AWS CLI with a user that has sufficent privlidges to create a cluster. 
+
+## Environment Variables
+
+After setting up the tools, set the following environment variables to store commonly used values. 
+
 ```bash
 export CLUSTER_NAME=$USER-karpenter-demo
 export AWS_DEFAULT_REGION=us-west-2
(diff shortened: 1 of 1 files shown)


ellistarn:
Just a quick note on the PR name. These ultimately land in our patch note, so we typically try to make them informative and framed in a linguistic way that sounds good for notes, e.g. (Updated AWS Getting Started guide to include ...)

docs/aws/README.md
```diff
@@ -1,11 +1,31 @@
-# AWS
+# Getting Started with Karpenter on AWS
```
ellistarn, line 1: I might just call this `Getting Started`, as it's already scoped to the AWS page.

docs/aws/README.md
```diff
@@ -1,11 +1,31 @@
...
 
-This guide will provide a complete Karpenter installation for AWS.
-These steps are opinionated and may need to be adapted for your use case.
+Karpenter automatically provisions new nodes in response to unscheduleable pods. Karpenter does this by observing events within the kubernetes cluster, and then sending commands to the underlying cloud platform. 
 
-> This guide should take less than 1 hour to complete and cost less than $.25
+In this example, the cluster is running on Amazon Web Services (AWS) Elastic Kubernetes Service (EKS). Karpenter is designed to be cloud platform independent, but currently only supports AWS. Contributions are welcomed.
```
ellistarn, line 5: consider "cloud platform independent" -> "vendor agnostic".

docs/aws/README.md
```diff
@@ -1,11 +1,31 @@
-# AWS
+# Getting Started with Karpenter on AWS
 
-This guide will provide a complete Karpenter installation for AWS.
-These steps are opinionated and may need to be adapted for your use case.
+Karpenter automatically provisions new nodes in response to unscheduleable pods. Karpenter does this by observing events within the kubernetes cluster, and then sending commands to the underlying cloud platform.
```
ellistarn, line 3: unscheduleable > unschedulable

geoffcline:
ok thanks this is still very much a WIP :)

docs/aws/README.md
geoffcline: I'm planning on having this page linked prominently from the homepage.

docs/aws/README.md
geoffcline: can we do "cloud platform agnostic"? I'd rather be more specific than vendor.

docs/aws/README.md
```diff
@@ -1,11 +1,31 @@
-# AWS
+# Getting Started with Karpenter on AWS
 
-This guide will provide a complete Karpenter installation for AWS.
-These steps are opinionated and may need to be adapted for your use case.
+Karpenter automatically provisions new nodes in response to unschedulable  pods. Karpenter does this by observing events within the kubernetes cluster, and then sending commands to the underlying cloud provider.
```
rothgar, line 3: Can we try to stick with symantic line breaks in the markdown? It makes reviewing diffs and reorganizing content easier. https://sembr.org/
The main thing is to put each sentence on it's own line. Markdown will combine the lines back into a paragraph but diffs are easier to see what changed and it's easier to leave comments.

On that note the first sentence in this paragraph has an extra space after unschedulable.

geoffcline: ok I fixed the extra space. 

Kubernetes uses either "gqq" in vi (wrap long lines), or long lines. I'd rather stick with one of the two existing standards in that space. 

I will not use semantic line breaks. 

First, "A semantic line break should occur after an independent clause as punctuated by a comma (,), semicolon (;), colon (:), or em dash (—)."

This rule is difficult to implement (can't be automated), and doesn't work nicely with soft wrap in vscode. 

Second, it fails in the goal of making git diffs readable. Small changes a slightly more readable, but the benefits are lost when paragraphs are revised.

rothgar: I disagree it fails to make diffs readable, but I'm not going to nit over formatting if not everyone wants to use it. I'll keep writing with sembr on sentence breaks as it has saved me a lot of time for content management and PRs in the past.

ellistarn:
Is this still WIP?

docs/aws/README.md
```diff
@@ -106,8 +113,13 @@ EOF
...
 
-### Create some pods
-Create some dummy pods and observe logs.
+# First Use
+
+Karpenter is now active and ready to begin provisioning nodes. Create a workload (e.g., deployment) to see karpenter provision some nodes. 
+
+## Create Workload
+
+This deployment is based on the pause image and initially has zero replicas.
```
njtran, line 122: Consider: 
"This deployment uses the [pause image](https://www.ianlewis.org/en/almighty-pause-container) and starts with zero replicas."

docs/aws/README.md
```diff
@@ -132,29 +144,31 @@ spec:
...
 ```bash
-kubectl scale deployment inflate --replicas 0
+kubectl scale deployment inflate --replicas 5
+kubectl logs -f -n karpenter $(kubectl get pods -n karpenter -l karpenter=controller -o name)
 ```
 
-Or you can manually delete the node with
+## Automatic Node Deprovisioning
+
+Now, delete the deployment. After 30 seconds (`ttlSecondsAfterEmpty`), karpenter should deprovision the now empty nodes.
```
njtran, line 159: nit: karpenter -> Karpenter

docs/aws/README.md
```diff
@@ -81,15 +89,14 @@ helm upgrade --install karpenter karpenter/karpenter \
...
-    -n karpenter --type='json' \
-    -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/args", "value": ["--verbose"]}]'
-```
+## Provisioner
+
+A single karpenter provisioner is capable of handling many different pod shapes. In other words, karpenter eliminates the need to manage many different node groups. Karpenter makes scheduling and provisioning decisions based on pod attributes such as labels and affinity. 
+
+Create a simple default provisioner using the command below. This provisioner provides instances with the default certificate bundle, and the control plane endpoint url. 
+
+Importantly, the `ttlSecondsAfterEmpty` value configures karpenter to deprovision empty nodes.
```
njtran, line 98: nit: karpenter -> Karpenter

docs/aws/README.md
```diff
@@ -132,29 +144,31 @@ spec:
...
-kubectl delete node $NODE_NAME
+kubectl delete deployment inflate
+kubectl logs -f -n karpenter $(kubectl get pods -n karpenter -l karpenter=controller -o name)
 ```
 
-### Cleanup
-> To avoid additional costs make sure you delete all ec2 instances before deleting the other cluster resources.
+# Cleanup
+
+It's important to both delete the cluster instances and the cluster control plane. AWS charges for both.
```
njtran, line 168: Does AWS charge for `iamidentitymappings` (created on line 69)? If so, can we include this?

docs/aws/README.md
ellistarn, line 3: nit: there's a double space between `unschedulable pods`

docs/aws/README.md
```diff
@@ -132,29 +144,31 @@ spec:
...
+kubectl logs -f -n karpenter $(kubectl get pods -n karpenter -l karpenter=controller -o name)
 ```
 
-### Cleanup
-> To avoid additional costs make sure you delete all ec2 instances before deleting the other cluster resources.
+# Cleanup
+
+It's important to both delete the cluster instances and the cluster control plane. AWS charges for both. 
+
+Delete cluster workloads and instances:
```
ellistarn, line 170: This doesn't delete instances. This just uninstalls karpenter and deletes the IAM resources deployed by CFN.

docs/aws/README.md
```diff
@@ -132,29 +144,31 @@ spec:
...
 ```
 
-Or you can manually delete the node with
+## Automatic Node Deprovisioning
+
+Now, delete the deployment. After 30 seconds (`ttlSecondsAfterEmpty`), karpenter should deprovision the now empty nodes. 
 
-> Karpenter automatically adds a node finalizer to properly cordon and drain nodes before they are terminated.
 ```bash
-kubectl delete node $NODE_NAME
```
ellistarn, line 153: Why was this example flow deleted?

docs/aws/README.md
```diff
@@ -106,8 +113,13 @@ EOF
 kubectl get provisioner default -o yaml
 ```
 
-### Create some pods
-Create some dummy pods and observe logs.
+# First Use
+
+Karpenter is now active and ready to begin provisioning nodes. Create a workload (e.g., deployment) to see karpenter provision some nodes.
```
ellistarn, line 118: s/karpenter/Karpenter

docs/aws/README.md
```diff
@@ -81,15 +89,14 @@ helm upgrade --install karpenter karpenter/karpenter \
...
 
-### (Optional) Enable Verbose Logging
-```bash
-kubectl patch deployment karpenter-controller \
-    -n karpenter --type='json' \
-    -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/args", "value": ["--verbose"]}]'
-```
+## Provisioner
+
+A single karpenter provisioner is capable of handling many different pod shapes. In other words, karpenter eliminates the need to manage many different node groups. Karpenter makes scheduling and provisioning decisions based on pod attributes such as labels and affinity.
```
ellistarn, line 94: If clarify this to "node scheduling properties, such as instance type and labels" rather than pod shapes.

ellistarn: Also -- karpenter doesn't currently support affinity. This is coming in v0.4

docs/aws/README.md
```diff
@@ -81,15 +89,14 @@ helm upgrade --install karpenter karpenter/karpenter \
   --namespace karpenter --set serviceAccount.create=false
 ```
 
-### (Optional) Enable Verbose Logging
```
ellistarn, line 84: I think we want to keep this on for the demo.

docs/aws/README.md
```diff
@@ -67,12 +74,13 @@ eksctl create iamidentitymapping \
...
+Now, Karpenter can send requests for new EC2 instances to AWS. 
+
+## Install Karpenter Helm Chart
 
-Use [`helm`](https://helm.sh/) to deploy Karpenter to the cluster.
-For additional values, see [the helm chart values](https://github.com/awslabs/karpenter/blob/main/charts/karpenter/values.yaml)
+Use helm to deploy Karpenter to the cluster. 
 
-> We created a Kubernetes service account with our cluster so we don't need the helm chart to do that.
+We created a kubernetes service account when we created the cluster using eksctl. Thus, we don't need the helm chart to do that.
```
ellistarn, line 83: I think we can probably leave out the second sentence starting with "Thus", as it refers to something historical that isn't. A reader following this guide need not be aware that you could have done this with helm in an alternative version of this guide.

docs/aws/README.md
```diff
@@ -15,38 +35,21 @@ KARPENTER_VERSION=$(curl -fsSL \
...
-```
+## Setup Authentication from Kubernetes to AWS (IRSA)
 
-### Setup IRSA, Karpenter Controller Role, and Karpenter Node Role
-We recommend using [CloudFormation](https://aws.amazon.com/cloudformation/) and [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IRSA) to manage these permissions.
-For production use, please review and restrict these permissions for your use case.
+IAM Roles for Service Accounts (IRSA) maps kubernetes resources to roles (permission sets) on AWS. 
 
-> For IRSA to work your cluster needs an [OIDC provider](https://docs.aws.amazon.com/eks/latest/userguide/enable-iam-roles-for-service-accounts.html)
+First, define a role using the template below. It provides full access to EC2, and limited access to other services such as EKS and Elastic Container Registry (ECR).
```
ellistarn, line 52: I think this description of the policy is wrong. We definitely don't need full access to EC2 and it would be dangerous to do so.

docs/aws/README.md
```diff
@@ -15,38 +35,21 @@ KARPENTER_VERSION=$(curl -fsSL \
...
-Karpenter can run anywhere, including on self-managed node groups, [managed node groups](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html), or [AWS Fargate](https://aws.amazon.com/fargate/).
-This demo will run Karpenter on Fargate, which means all EC2 instances added to this cluster will be controlled by Karpenter.
+Create a cluster with `eksctl`. The below configuration file specifies a basic cluster (name, region), an IAM role for karpenter to use, and two fargate profiles. 
 
-```bash
-curl -fsSL https://raw.githubusercontent.com/awslabs/karpenter/"${KARPENTER_VERSION}"/docs/aws/eks-config.yaml \
-  | envsubst \
-  | eksctl create cluster -f -
-```
+The two fargate profiles host `kube-system` and the karpenter service itself. This permits karpenter to manage all nodes, without a circular dependency.
```
ellistarn, line 42: I don't think there's a circular dependency. There *is* a bootstrapping problem, but it's possible to run Karpenter on any node from any node provider (mng, self managed nodes, standalone ec2 instance, from your laptop, fargate, etc).

docs/aws/README.md
```diff
@@ -15,38 +35,21 @@ KARPENTER_VERSION=$(curl -fsSL \
   | jq -r '.tag_name')
 ```
 
-### Create a Cluster
+## Create a Cluster
 
-Karpenter can run anywhere, including on self-managed node groups, [managed node groups](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html), or [AWS Fargate](https://aws.amazon.com/fargate/).
```
ellistarn, line 20: I think it's really important that we don't remove this line. I don't want readers to misinterpret this as Karpenter relying on Fargate.

docs/aws/README.md
```diff
@@ -1,11 +1,31 @@
...
+Karpenter is installed in clusters with a simple helm chart.
+
+Karpenter additionally requires IAM Roles for Service Accounts (IRSA). IRSA permits Karpenter (within the cluster) to make privileged requests to AWS (as the cloud provider). 
+
+## Required Utilities
+
+Install these tools before proceeding:
+
+1. [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-linux.html)
+2. `kubectl` - [the kubernetes CLI](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/)
```
ellistarn, line 20: Any reason why we don't just say "[`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/)"

docs/aws/README.md
rothgar: `iamidentitymappings` is only adding to the aws-auth configmap inside the custer so there's no charge for it.

docs/aws/README.md
geoffcline: I didn't fully consider the impact of this example. I added a section calling out this behavior.

docs/aws/README.md
geoffcline: links inside code fmt can render inconsistently,

docs/aws/README.md
geoffcline: I updated this section generally

docs/aws/README.md
geoffcline:
```yaml
  KarpenterNodeRole:
    Type: "AWS::IAM::Role"
    Properties:
      RoleName: !Sub "KarpenterNodeRole-${ClusterName}"
      Path: /
      AssumeRolePolicyDocument:
        Version: "2012-10-17"
        Statement:
          - Effect: Allow
            Principal:
              Service:
                !Sub "ec2.${AWS::URLSuffix}"
            Action:
              - "sts:AssumeRole"
```

am I looking at the right thing? how would you describe this?

docs/aws/README.md
geoffcline: I think it's helpful to explain why we are setting certain options here, to connect the dots between the service account earlier and the option here. 

```
helm upgrade --install karpenter karpenter/karpenter \
  --namespace karpenter --set serviceAccount.create=false
```

docs/aws/README.md
geoffcline: can you send me a quick cmd for someone to view the output? what's an example of something reported only by verbose logging?

docs/aws/README.md
geoffcline: I currently think deleting the cluster deletes associated managed node groups. Does Karpenter modify this behavior? I hadn't considered that impact.

geoffcline:
Thanks for the feedback everyone! I pushed an update, and responded to comments. 

If I marked it as resolved, then I accepted the suggestion.

docs/aws/README.md
```diff
@@ -60,13 +60,13 @@ or [AWS Fargate](https://aws.amazon.com/fargate/).
 Karpenter will provision traditional instances on EC2. 
 
 Additionally, the configuration file sets up an [OIDC
-provider](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens),
+provider](https://Kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens),
```
ellistarn, line 63: overeager find/replace here, I think.

geoffcline: fixed :)

docs/aws/README.md
```diff
@@ -1,11 +1,40 @@
...
+Karpenter automatically provisions new nodes in response to unschedulable
+pods. Karpenter does this by observing events within the Kubernetes cluster,
+and then sending commands to the underlying cloud provider. 
 
-> This guide should take less than 1 hour to complete and cost less than $.25
+In this example, the cluster is running on Amazon Web Services (AWS) Elastic
+Kubernetes Service (EKS). Karpenter is designed to be cloud provider agnostic,
+but currently only supports AWS. Contributions are welcomed. 
+
+This guide should take less than 1 hour to complete, and cost less than $0.25.
```
ellistarn, line 11: Can we change 1 hour -> 30 minutes. I don't want to scare folks off and it just doesn't take that long.

docs/aws/README.md
```diff
@@ -1,11 +1,40 @@
...
+Follow the clean-up instructions to reduce any charges.
+
+# Install
+
+Karpenter is installed in clusters with a helm chart.
```
ellistarn, line 16: nit: `Karpenter can be installed to your cluster using helm.`

docs/aws/README.md
ellistarn: Sure, but can we at least make it consistent with the other list items.

docs/aws/README.md
```diff
@@ -1,11 +1,40 @@
...
+
+Karpenter additionally requires IAM Roles for Service Accounts (IRSA). IRSA
```
ellistarn, line 18: I'd be more crisp that Karpenter doesn't necessarily require IRSA, but it does require IAM roles. IRSA is a convenient way to manage IAM roles when using Kubernetes.

Also, IMO "as the cloud provider" is superflous.

docs/aws/README.md
```diff
@@ -1,11 +1,40 @@
...
+
+## Required Utilities
+
+Install these tools before proceeding:
+
+1. [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-linux.html)
+2. `kubectl` - [the kubernetes CLI](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/)
+3. `eksctl` - [the CLI for AWS EKS](https://docs.aws.amazon.com/eks/latest/userguide/eksctl.html)
+
+Login to the AWS CLI with a user that has sufficient privileges to create a
```
ellistarn, line 30: I'd be a little careful about this language. Users don't "login to the aws cli", rather, they ensure that their aws credentials are configured with sufficient permissions.

docs/aws/README.md
```diff
@@ -1,11 +1,40 @@
...
+cluster. 
+
+## Environment Variables
+
+After setting up the tools, set the following environment variables to store
```
ellistarn, line 35: I'd slice off "to score commonly used values." from this.

docs/aws/README.md
```diff
@@ -15,38 +44,34 @@ KARPENTER_VERSION=$(curl -fsSL \
...
+This permits Karpenter to manage all EC2 instances. 
 
-Tag the cluster subnets with the required tags for Karpenter auto discovery.
+Karpenter itself can run anywhere, including on self-managed node groups,
+[managed node groups](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html), 
+or [AWS Fargate](https://aws.amazon.com/fargate/).
 
-> If you are using a cluster with version 1.18 or below you can skip this step.
-More [detailed here](https://github.com/awslabs/karpenter/issues/404#issuecomment-845283904).
+Karpenter will provision traditional instances on EC2.
```
ellistarn, line 60: I'd be ultra crisp that these instances run in your account. "Karpenter provisions EC2 instances in your account"

docs/aws/README.md
```diff
@@ -15,38 +44,34 @@ KARPENTER_VERSION=$(curl -fsSL \
...
-More [detailed here](https://github.com/awslabs/karpenter/issues/404#issuecomment-845283904).
+Karpenter will provision traditional instances on EC2. 
 
-```bash
-SUBNET_IDS=$(aws cloudformation describe-stacks \
-    --stack-name eksctl-${CLUSTER_NAME}-cluster \
-    --query 'Stacks[].Outputs[?OutputKey==`SubnetsPrivate`].OutputValue' \
-    --output text)
-
-aws ec2 create-tags \
```
ellistarn, line 40: We absolutely need these or things will break.

docs/aws/README.md
```diff
@@ -15,38 +44,34 @@ KARPENTER_VERSION=$(curl -fsSL \
...
-### Setup IRSA, Karpenter Controller Role, and Karpenter Node Role
-We recommend using [CloudFormation](https://aws.amazon.com/cloudformation/) and [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IRSA) to manage these permissions.
-For production use, please review and restrict these permissions for your use case.
+## Setup Authentication from Kubernetes to AWS (IRSA)
 
-> For IRSA to work your cluster needs an [OIDC provider](https://docs.aws.amazon.com/eks/latest/userguide/enable-iam-roles-for-service-accounts.html)
+IAM Roles for Service Accounts (IRSA) maps Kubernetes resources to roles
+(permission sets) on AWS. 
+
+First, define a role using the template below. It provides full access to EC2,
```
ellistarn, line 72: I thought I left a comment here earlier about "full access to EC2". In case it got lost, I'll reiterate. We definitely don't do this, and it would be dangerous to do so.

docs/aws/README.md
```diff
@@ -67,12 +96,14 @@ eksctl create iamidentitymapping \
   --group system:nodes
 ```
 
-### Install Karpenter
+Now, Karpenter can send requests for new EC2 instances to AWS.
```
ellistarn, line 99: Now, Karpenter can send requests for new EC2 instances to AWS and those instances can connect to your cluster. 

These two pieces are core to our permission model.

docs/aws/README.md
```diff
@@ -81,15 +112,20 @@ helm upgrade --install karpenter karpenter/karpenter \
...
+
+Create a simple default provisioner using the command below. This provisioner
+provides instances with the default certificate bundle, and the control plane
+endpoint url. 
+
+Importantly, the `ttlSecondsAfterEmpty` value configures Karpenter to
+deprovision empty nodes. 
 
-### Create a Provisioner
-Create a default Provisioner that launches nodes configured with cluster name, endpoint, and caBundle.
```
ellistarn, line 92: We can now remove caBundle thanks to a recent PR!

docs/aws/README.md
```diff
@@ -132,29 +174,42 @@ spec:
...
 ```bash
-kubectl scale deployment inflate --replicas 0
+kubectl delete deployment inflate
+kubectl logs -f -n karpenter $(kubectl get pods -n karpenter -l karpenter=controller -o name)
 ```
 
-Or you can manually delete the node with
+## Manual Node Deprovisioning
+
+If you delete a node with kubectl, Karpenter deprovisions the corresponding instance.
```
ellistarn, line 200: We need to be a bit careful here. This is actually not supported in kubernetes without karpenter. It's one of our value-adds. This action is actually quite dangerous without Karpenter, since it can leak EC2 resources.

I'd keep the line "Karpenter automatically adds a node finalizer to properly cordon and drain nodes before they are terminated." which specifies why this is safe for folks who are experienced enough to be nervous about this.

docs/aws/README.md
ellistarn: Yeah this is actually quite interesting. Since we're not using node groups, these instances may leak.

docs/aws/README.md
geoffcline: https://github.com/awslabs/karpenter/pull/506#discussion_r668975648

geoffcline: can you help me understand the policy then? 

I'm only seeing this

```
          - Effect: Allow
            Principal:
              Service:
                !Sub "ec2.${AWS::URLSuffix}
```

docs/aws/README.md
geoffcline: good catch

docs/aws/README.md
geoffcline: I'll use that `delete nodes` cmd as an example, and call this out.

docs/aws/README.md
geoffcline: we talked and, for example, verbose logging includes more aws specific events

docs/aws/README.md
geoffcline: I started a new EKS cluster build and if it takes less than 20 minutes I'll update this.

docs/aws/README.md
geoffcline: I meant to rearrange this and dropped it. It will be after "create a cluster" and before IRSA.

docs/aws/README.md
geoffcline: it took only 11! 🥳 

it does *feel* like ages

docs/aws/README.md
geoffcline: 👍

docs/aws/README.md
geoffcline: ok, I will remove `caBundle` def from the provisioner def

docs/aws/README.md
geoffcline: done

Issue #501, opened 2021-07-02
Decouple `Cluster.Name` from `Cluster.Endpoint` and `Cluster.CABundle`


alex-berger:
**Tell us about your request**

The [`Cluster`](https://github.com/awslabs/karpenter/blob/v0.2.8/pkg/apis/provisioning/v1alpha2/provisioner.go#L36) field in [`ProvisionerSpec`](https://github.com/awslabs/karpenter/blob/v0.2.8/pkg/apis/provisioning/v1alpha2/provisioner.go#L33) is optional, it is a pointer `*Cluster` and thus can be `nil` and is documented with `+optional`. This is not consistent with the [validation logic](https://github.com/awslabs/karpenter/blob/v0.2.8/pkg/apis/provisioning/v1alpha2/provisioner_validation.go#L76) which requires the `Cluster` field [to be not `nil`](https://github.com/awslabs/karpenter/blob/v0.2.8/pkg/apis/provisioning/v1alpha2/provisioner_validation.go#L77). 

I guess the validation logic is biased towards the AWS provider which **requires** the [`Cluster.Name`](https://github.com/awslabs/karpenter/blob/v0.2.8/pkg/apis/provisioning/v1alpha2/provisioner.go) in several places:

- https://github.com/awslabs/karpenter/blob/v0.2.8/pkg/cloudprovider/aws/securitygroups.go
- https://github.com/awslabs/karpenter/blob/v0.2.8/pkg/cloudprovider/aws/subnets.go#L68
- https://github.com/awslabs/karpenter/blob/v0.2.8/pkg/cloudprovider/aws/launchtemplate.go#L72
- https://github.com/awslabs/karpenter/blob/v0.2.8/pkg/cloudprovider/aws/launchtemplate.go#L156
- ...

[...]

PR #508, opened 2021-07-08
Decouple Cluster.Name from Cluster.Endpoint and Cluster.CABundle

alex-berger:
Decouple Cluster.Name from Cluster.Endpoint and Cluster.CABundle and add
support to dynamically load CABundle from in-cluster configuration.

**Issue, if available:**

#501

**Description of changes:**


pkg/apis/provisioning/v1alpha2/provisioner.go
```diff
@@ -58,12 +57,14 @@ type ProvisionerSpec struct {
...
 type Cluster struct {
-	// Name is required to detect implementing cloud provider resources.
-	// +required
-	Name string `json:"name"`
-	// CABundle is required for nodes to verify API Server certificates.
-	// +required
-	CABundle string `json:"caBundle"`
+	// Name may be required to detect implementing cloud provider resources.
+	// +optional
+	Name string `json:"name,omitempty"`
```
ellistarn, line 62: I think we want this to be a *string now, to avoid the data serializing as "".

pkg/apis/provisioning/v1alpha2/provisioner.go
```diff
@@ -58,12 +57,14 @@ type ProvisionerSpec struct {
...
+	// CABundle used by nodes to verify API Server certificates. If omitted,
+	// it will be dynamically loaded at runtime from the in-cluster configuration
+	// file /var/run/secrets/kubernetes.io/serviceaccount/ca.crt.
+	// +optional
+	CABundle string `json:"caBundle,omitempty"`
```
ellistarn, line 67: same

pkg/apis/provisioning/v1alpha2/provisioner.go
```diff
@@ -58,12 +57,14 @@ type ProvisionerSpec struct {
...
 	// Endpoint is required for nodes to connect to the API Server.
 	// +required
 	Endpoint string `json:"endpoint"`
```
ellistarn, line 70: Consider moving endpoint to the top of these three fields, now that it's the only required one.

pkg/apis/provisioning/v1alpha2/provisioner_validation.go
```diff
@@ -95,15 +101,11 @@ func (c *Cluster) validate() (errs *apis.FieldError) {
 	if c == nil {
 		return errs.Also(apis.ErrMissingField())
 	}
-	if len(c.Name) == 0 {
-		errs = errs.Also(apis.ErrMissingField("name"))
-	}
 	if len(c.Endpoint) == 0 {
+		// Endpoint cannot be resolved from in-cluster configuration and
```
ellistarn, line 105: Consider skipping this extraneous comment. I think it's well covered in provisioner.go

pkg/apis/provisioning/v1alpha2/provisioner_validation_test.go
```diff
@@ -65,11 +65,13 @@ var _ = Describe("Validation", func() {
...
-			nil,
-			{Endpoint: "https://test-cluster", CABundle: "dGVzdC1jbHVzdGVyCg=="},
+		for _, cluster := range []Cluster{
+			{},
+			//{Endpoint: "https://test-cluster", CABundle: "dGVzdC1jbHVzdGVyCg=="},
 			{Name: "test-cluster", CABundle: "dGVzdC1jbHVzdGVyCg=="},
-			{Name: "test-cluster", Endpoint: "https://test-cluster"},
+			{Name: "test-cluster"},
+			{CABundle: "dGVzdC1jbHVzdGVyCg=="},
+			//{Name: "test-cluster", Endpoint: "https://test-cluster"},
```
ellistarn, line 74: Reminder to clean these up

pkg/cloudprovider/types.go
```diff
@@ -31,8 +31,12 @@ type CloudProvider interface {
 	// GetInstanceTypes returns the instance types supported by the cloud
 	// provider limited by the provided constraints and daemons.
 	GetInstanceTypes(context.Context) ([]InstanceType, error)
+	// ValidateSpec is a hook for additional spec validation logic specific to the cloud provider.
+	// Note, implementations should not validate constraints resp. call `ValidateConstraints`
+	// from whithin this method as constraints are validated separately.
+	ValidateSpec(context.Context, *v1alpha2.ProvisionerSpec) *apis.FieldError
 	// Validate is a hook for additional constraint validation logic specific to the cloud provider
-	Validate(context.Context, *v1alpha2.Constraints) *apis.FieldError
+	ValidateConstraints(context.Context, *v1alpha2.Constraints) *apis.FieldError
```
ellistarn, line 39: Can you add a note to the comment here about why we need this in addition to ValidateSpec, since constraints is a subfield. Something to the effect of,

"ValidateConstraints is used at provisioning time to ensure that pods are provisionable for the specified provisioner".

pkg/controllers/allocation/controller.go
```diff
@@ -92,13 +92,23 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
 		packings = append(packings, c.packer.Pack(ctx, constraintGroup, instanceTypes)...)
 	}
 
-	// 4. Create packedNodes for packings
-	packedNodes, err := c.cloudProvider.Create(ctx, provisioner, packings)
+	// 4. Complement provisioner with (dynamic) default values, which must not
+	//    be hoisted into the original CRD as they might change with each reconciliation
+	//    loop iteration.
+	provisionerWithDefaults, err := provisioner.WithDefaults()
```
ellistarn, line 98: It strikes me that it might be useful to include cluster.KubernetesVersion in this mechanism

ellistarn: Thoughts on running this as the very first step?

ellistarn: Also, consider calling this WithDynamicDefaults to differentiate between SetDefaults (which are static).

ellistarn (commented):
Nicely done!

pkg/controllers/allocation/controller.go
alex-berger: > It strikes me that it might be useful to include cluster.KubernetesVersion in this mechanism

What do we need it for and where would it be stored (in which struct)?

alex-berger: > Thoughts on running this as the very first step?

I am not sure what you have in mind, can your please elaborate a bit more.

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
```diff
@@ -16,10 +16,67 @@ package v1alpha2
...
+	if c.CABundle != nil {
+		// If CABundle is explicitly provided use that one. An empty string is
+		// a valid value here if no CABundle is required at all.
+		return c.CABundle, nil
+	} else {
+		// Otherwise, fallback to the in-cluster configuration.
+		if binary, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"); err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				zap.S().Infof("in-cluster CABundle file /var/run/secrets/kubernetes.io/serviceaccount/ca.crt not found, %w", err)
+				//TODO shall this be treated as an error or is it a valid use case that a cluster has no CABundle at all?
```
alex-berger, line 72: @ellistarn Maybe you can help me here. I am not sure whether we should raise an error if the in-cluster config does not contain a CABundle file or if it is fine to silently assume that in that case CABundle is not needed. 

I am inclined to the second option, just ignoring it and not treat this as an error, but would like to hear other opinions as well.

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
```diff
@@ -16,10 +16,67 @@ package v1alpha2
...
+	cluster.CABundle, err = cluster.getCABundle()
+	return cluster, err
+}
+
+func (c *Cluster) getCABundle() (*string, error) {
+	if c.CABundle != nil {
+		// If CABundle is explicitly provided use that one. An empty string is
+		// a valid value here if no CABundle is required at all.
+		return c.CABundle, nil
+	} else {
```
ellistarn, line 67: Since the first statement uses a shortcircuiting pattern, we can remove the else and unindent all of this logic.

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
```diff
@@ -16,10 +16,67 @@ package v1alpha2
...
+		// Otherwise, fallback to the in-cluster configuration.
+		if binary, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"); err != nil {
```
ellistarn, line 69: Consider extracting this constant to a variable above and referencing it in both places.

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
```diff
@@ -16,10 +16,67 @@ package v1alpha2
...
+			if errors.Is(err, os.ErrNotExist) {
+				zap.S().Infof("in-cluster CABundle file /var/run/secrets/kubernetes.io/serviceaccount/ca.crt not found, %w", err)
+				//TODO shall this be treated as an error or is it a valid use case that a cluster has no CABundle at all?
+				return nil, nil
+			}
+			return nil, err
+		} else {
+			zap.S().Debugf("using in-cluster CABundle from file /var/run/secrets/kubernetes.io/serviceaccount/ca.crt, %s", string(binary))
```
ellistarn, line 77: I'd avoid including the CABundle in the log line, since folks can always retrieve it via the service account of the pod. It also pollutes the logs as it's a fairly lengthy string.

ellistarn, line 77: Similarly, you can remove this else and the unnecessary indentation due to the short circuiting pattern.

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
```diff
@@ -16,10 +16,67 @@ package v1alpha2
...
+			encoded := base64.StdEncoding.EncodeToString(binary)
+			return &encoded, nil
```
ellistarn, line 79: Any reason we're returning this as a string pointer? Seems like it creates NPE opportunities that could be more safely handled by using an empty string for the CABundle if something goes wrong. It also allows you to simplify the code and use `return base64.StdEncoding.EncodeToString(binary), nil`

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
ellistarn, line 77: Do you mind following the project's logging convention of Capital letters. e.g.
`Using CABundle located at /var/run/secrets/kubernetes.io/serviceaccount/ca.crt`. I'm also not sure you're guaranteed that this is an in-cluster cabundle, since the user is able to override this path.

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
ellistarn: I agree with this. As long as we're explicit from debug lines which CABundle we're using (i.e, using spec configured CABundle, falling back to file base CABundle, or using default http client cabundle).

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
```diff
@@ -16,10 +16,67 @@ package v1alpha2
...
+// is not explicitly set.
+func (c *Cluster) withDynamicDefaults() (_ Cluster, err error) {
+	cluster := *c.DeepCopy()
+	cluster.CABundle, err = cluster.getCABundle()
+	return cluster, err
+}
+
+func (c *Cluster) getCABundle() (*string, error) {
+	if c.CABundle != nil {
+		// If CABundle is explicitly provided use that one. An empty string is
```
ellistarn, line 64: I'm not sure I understand this comment. My behavioral expectations are.

1. User configures CABundle = nil
  a. Fall back to file, if exists, then http default cabundle
2. User configures CABundle = ""
  a. Attempt to use the CABundle "", which will likely fail
3. User configures CABundle = "something real"
  a. Use the configured bundle 

I think your logic does this, but the comment is a bit confusing to me.

pkg/utils/ptr/ptr.go
```diff
@@ -18,6 +18,18 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
+func String(s string) *string {
+	return &s
+}
+
+func StringOrEmpty(s *string) string {
```
ellistarn, line 25: Check out [Knative's ptr.StringValue()](https://github.com/knative/pkg/blob/main/ptr/value.go#L68). I implemented all of these in Knative PKG to we should be able to deprecate all of these functions except for the pod specific ones.

pkg/controllers/allocation/controller.go
```diff
@@ -66,7 +66,16 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 // Reconcile executes an allocation control loop for the resource
 func (c *Controller) Reconcile(ctx context.Context, object client.Object) (reconcile.Result, error) {
 	provisioner := object.(*v1alpha2.Provisioner)
-	// 1. Filter pods
+
+	// 1. Complement provisioner with (dynamic) default values, which must not
```
ellistarn, line 70: I might say Hydrate instead of Complement. I'd also be explicit that they are not persisted, e.g. "must not be persisted into ..."

pkg/controllers/allocation/controller.go
```diff
@@ -66,7 +66,16 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
...
+	//    be hoisted into the original CRD as they might change with each reconciliation
+	//    loop iteration.
+	provisionerWithDefaults, err := provisioner.WithDynamicDefaults()
```
ellistarn, line 73: Consider naming this `provisioner` and naming the other one `persisted`, so the rest of the function doesn't need to be aware that anything has been defaulted.

pkg/controllers/allocation/controller.go
```diff
@@ -92,13 +101,15 @@ func (c *Controller) Reconcile(ctx context.Context, object client.Object) (recon
 		packings = append(packings, c.packer.Pack(ctx, constraintGroup, instanceTypes)...)
 	}
 
-	// 4. Create packedNodes for packings
-	packedNodes, err := c.cloudProvider.Create(ctx, provisioner, packings)
+	// 5. Create packedNodes for packings and also copy all Status changes made by the
+	//    cloud provider to the original provisioner instance.
+	packedNodes, err := c.cloudProvider.Create(ctx, &provisionerWithDefaults, packings)
+	provisioner.Status = provisionerWithDefaults.Status
```
ellistarn, line 107: I think we should skip this functionality for now. We started off with some status conditions which were carried forward from some legacy code in Karpenter's original POC. I want to revisit status and conditions more intentionally in the future.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -149,21 +150,22 @@ func (p *LaunchTemplateProvider) ensureLaunchTemplate(ctx context.Context, optio
 }
 
 func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, options *launchTemplateOptions) (*ec2.LaunchTemplate, error) {
+	clusterName := ptr.StringOrEmpty(options.Cluster.Name)
```
ellistarn, line 153: consider just inlining ptr.StringValue(options.Cluster.Name) to simplify the code (i.e. decrease the number of variables/concepts by 1)

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -37,8 +38,8 @@ const (
 	bottlerocketUserData     = `
 [settings.kubernetes]
 api-server = "{{.Cluster.Endpoint}}"
-cluster-certificate = "{{.Cluster.CABundle}}"
-cluster-name = "{{.Cluster.Name}}"
+cluster-certificate = "{{if .Cluster.CABundle}}{{.Cluster.CABundle}}{{end}}"
+cluster-name = "{{if .Cluster.Name}}{{.Cluster.Name}}{{end}}"
```
ellistarn, line 42: Just double checking that if Cluster.Name = ptr.String(""), then this if statement will return true and add this line.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -37,8 +38,8 @@ const (
 	bottlerocketUserData     = `
 [settings.kubernetes]
 api-server = "{{.Cluster.Endpoint}}"
-cluster-certificate = "{{.Cluster.CABundle}}"
-cluster-name = "{{.Cluster.Name}}"
+cluster-certificate = "{{if .Cluster.CABundle}}{{.Cluster.CABundle}}{{end}}"
```
ellistarn, line 41: If there any chance that this will cause the code `cluster-certificate = ` which might be misinterpreted by bottlerocket? Is that valid TOML?

Does it make more sense to wrap the entire line like we do with labels?

ellistarn (commented):
Couple comments and mostly questions. I think we're almost there. Thanks for the contribution!!

pkg/controllers/allocation/controller.go
ellistarn: You understood perfectly :)

ellistarn:
Can you remove the "Implement 501" from the PR Name? We use PR Names to generate release notes, and the issue # will be included if it's tagged to this issue.

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
alex-berger: This is because (as you suggested) I refactored `Clutser.CABundle` to type `*string`. Of course I can change the return type fo this function to `string` but that will just move the need for an intermediate variable (to get a pointer) to the calling function.

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
alex-berger: `""` (empty string) is an escape hatch for clients (users) which want to use the HTTP clients default trust store (CABundle).

pkg/controllers/allocation/controller.go
alex-berger: Actually, my motivation was to make sure the rest of the functions is aware of this and does not by accident persist (apply/update/patch/...) the `provisionerWithDefaults` object. Maybe a better approach to make intentions more clear and to prevent other code from accidentally persisting this `transient` object is to return `ProvisionerSpec` (instead of `Provisioner`) from `WithDynamicDefaults()` and to pass that spec to all the cloud provider methods.

ellistarn: We actually used to have this, but found it was useful to be able to use the provisioner name/namespace in caching logic (though I don't believe we do this anymore). You could also imagine that cloud providers may wish to customize behavior (beyond labels) by adding an annotation to the metadata on the provisioner. Perhaps it's better to discourage this though.

In the interest of gaining the value the PR brings already, I'm happy to punt on this question and leave as is. I'm also supportive of moving the cloud provider API to deal only with ProvisionerSpec to make the mutation question more clear.

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
ellistarn: I find this a bit awkward, but it doesn't look like it's possible to override `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`, which IMO would be the cleanest way to escape hatch this. However, at the end of the day, this bundle *must* be viable to communicate with the API Server, otherwise many other things would be broken. I'm willing to support this escape hatch edge case, but given how awkward it is, my preference would be to not support it until someone asks specifically for it.

pkg/apis/provisioning/v1alpha2/provisioner_defaults.go
ellistarn: Oopsie you're 100% correct.

pkg/controllers/allocation/controller.go
alex-berger: OK, I will keep that `ProvisionerSpec` refactoring out of this PR to keep is simple and focused. I guess that would first need a discussion/design on how to pass cloud provider specific configuration as you already mused about in https://github.com/awslabs/karpenter/issues/501#issuecomment-873117461.

ellistarn (approved):
Thank you!

PR #515, opened 2021-07-16
Added Per-pod eviction backoff and retry logic 

njtran:
**Issue, if available:**
https://github.com/awslabs/karpenter/issues/452

**Description of changes:**
- Pod evictions now have separate backoffs per pod
- Drain in termination controller queues up pods to a continually running queue processor as a goroutine in the background
- Accounts for different pod eviction errors 

### DIFF
--- a/pkg/controllers/termination/controller.go
+++ b/pkg/controllers/termination/controller.go
@@ -41,21 +41,16 @@ type Controller struct {
 	KubeClient client.Client
 }
 
-// For returns the resource this controller is for.
-func (c *Controller) For() client.Object {
-	return &v1.Node{}
-}
-
-func (c *Controller) Name() string {
-	return "terminator"
-}
-
 // NewController constructs a controller instance
 func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.CloudProvider) *Controller {
 	return &Controller{
 		KubeClient: kubeClient,
-		Terminator: NewTerminator(kubeClient, coreV1Client, cloudProvider,
-			workqueue.NewRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(evictionQueueBaseDelay, evictionQueueMaxDelay))),
+		Terminator: &Terminator{
+			KubeClient:    kubeClient,
+			CoreV1Client:  coreV1Client,
+			CloudProvider: cloudProvider,
+			EvictionQueue: NewEvictionQueue(coreV1Client),
+		},
 	}
 }
 
--- a/pkg/controllers/termination/controller.go
+++ b/pkg/controllers/termination/controller.go
@@ -17,11 +17,13 @@ package termination
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	provisioning "github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha3"
 	"github.com/awslabs/karpenter/pkg/cloudprovider"
 	"github.com/awslabs/karpenter/pkg/utils/functional"
+	set "github.com/deckarep/golang-set"
 	"golang.org/x/time/rate"
 
 	v1 "k8s.io/api/core/v1"
@@ -37,17 +39,31 @@ import (
 
 // Controller for the resource
 type Controller struct {
-	terminator    *Terminator
-	cloudProvider cloudprovider.CloudProvider
-	kubeClient    client.Client
+	terminator *Terminator
+	kubeClient client.Client
+}
+
+// For returns the resource this controller is for.
+func (c *Controller) For() client.Object {
+	return &v1.Node{}
+}
+
+func (c *Controller) Name() string {
+	return "terminator"
 }
 
 // NewController constructs a controller instance
 func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.CloudProvider) *Controller {
 	return &Controller{
-		terminator:    &Terminator{kubeClient: kubeClient, cloudProvider: cloudProvider, coreV1Client: coreV1Client},
-		cloudProvider: cloudProvider,
-		kubeClient:    kubeClient,
+		kubeClient: kubeClient,
+		terminator: &Terminator{kubeClient: kubeClient, cloudProvider: cloudProvider, coreV1Client: coreV1Client,
+			evictionQueue: EvictionQueue{
+				Queue:        workqueue.NewDelayingQueue(),
+				RateLimiter:  workqueue.NewItemExponentialFailureRateLimiter(100*time.Millisecond, 10*time.Second),
+				coreV1Client: coreV1Client,
+				enqueued:     set.NewSet(),
+				once:         sync.Once{},
+			}},
 	}
 }
 
@@ -75,13 +91,14 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("draining node %s, %w", node.Name, err)
 	}
-	// 5. If fully drained, terminate the node
-	if drained {
-		if err := c.terminator.terminate(ctx, node); err != nil {
-			return reconcile.Result{}, fmt.Errorf("terminating node %s, %w", node.Name, err)
-		}
+	if !drained {
+		return reconcile.Result{}, nil
 	}
-	return reconcile.Result{Requeue: !drained}, nil
+	// 4. If fully drained, terminate the node
+	if err := c.terminator.terminate(ctx, node); err != nil {
+		return reconcile.Result{}, fmt.Errorf("terminating node %s, %w", node.Name, err)
+	}
+	return reconcile.Result{}, nil
 }
 
 func (c *Controller) Register(_ context.Context, m manager.Manager) error {
--- a/pkg/controllers/termination/controller.go
+++ b/pkg/controllers/termination/controller.go
@@ -39,8 +39,8 @@ import (
 
 // Controller for the resource
 type Controller struct {
-	terminator *Terminator
-	kubeClient client.Client
+	Terminator *Terminator
+	KubeClient client.Client
 }
 
 // For returns the resource this controller is for.
@@ -55,11 +55,24 @@ func (c *Controller) Name() string {
 // NewController constructs a controller instance
 func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.CloudProvider) *Controller {
 	return &Controller{
-		kubeClient: kubeClient,
-		terminator: &Terminator{kubeClient: kubeClient, cloudProvider: cloudProvider, coreV1Client: coreV1Client,
+		KubeClient: kubeClient,
+		Terminator: &Terminator{kubeClient: kubeClient, cloudProvider: cloudProvider, coreV1Client: coreV1Client,
 			evictionQueue: EvictionQueue{
-				Queue:        workqueue.NewDelayingQueue(),
-				RateLimiter:  workqueue.NewItemExponentialFailureRateLimiter(100*time.Millisecond, 10*time.Second),
+				queue:        workqueue.NewRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(100*time.Millisecond, 10*time.Second)),
+				coreV1Client: coreV1Client,
+				enqueued:     set.NewSet(),
+				once:         sync.Once{},
+			}},
+	}
+}
+
+func NewControllerWithCustomQueue(kubeClient client.Client, coreV1Client corev1.CoreV1Interface,
+	cloudProvider cloudprovider.CloudProvider, rateLimitingQueue workqueue.RateLimitingInterface) *Controller {
+	return &Controller{
+		KubeClient: kubeClient,
+		Terminator: &Terminator{kubeClient: kubeClient, cloudProvider: cloudProvider, coreV1Client: coreV1Client,
+			evictionQueue: EvictionQueue{
+				queue:        rateLimitingQueue,
 				coreV1Client: coreV1Client,
 				enqueued:     set.NewSet(),
 				once:         sync.Once{},
@@ -71,7 +84,7 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	// 1. Retrieve node from reconcile request
 	node := &v1.Node{}
-	if err := c.kubeClient.Get(ctx, req.NamespacedName, node); err != nil {
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, node); err != nil {
 		if errors.IsNotFound(err) {
 			return reconcile.Result{}, nil
 		}
@@ -83,19 +96,19 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, nil
 	}
 	// 3. Cordon node
-	if err := c.terminator.cordon(ctx, node); err != nil {
+	if err := c.Terminator.cordon(ctx, node); err != nil {
 		return reconcile.Result{}, fmt.Errorf("cordoning node %s, %w", node.Name, err)
 	}
 	// 4. Drain node
-	drained, err := c.terminator.drain(ctx, node)
+	drained, err := c.Terminator.drain(ctx, node)
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("draining node %s, %w", node.Name, err)
 	}
 	if !drained {
-		return reconcile.Result{}, nil
+		return reconcile.Result{Requeue: true}, nil
 	}
 	// 4. If fully drained, terminate the node
-	if err := c.terminator.terminate(ctx, node); err != nil {
+	if err := c.Terminator.terminate(ctx, node); err != nil {
 		return reconcile.Result{}, fmt.Errorf("terminating node %s, %w", node.Name, err)
 	}
 	return reconcile.Result{}, nil
--- a/pkg/controllers/termination/controller.go
+++ b/pkg/controllers/termination/controller.go
@@ -17,13 +17,11 @@ package termination
 import (
 	"context"
 	"fmt"
-	"sync"
 	"time"
 
 	provisioning "github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha3"
 	"github.com/awslabs/karpenter/pkg/cloudprovider"
 	"github.com/awslabs/karpenter/pkg/utils/functional"
-	set "github.com/deckarep/golang-set"
 	"golang.org/x/time/rate"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


ellistarn (commented):
Nice! Overall looks good. I'll take a deeper look tomorrow. We need some robust tests though.

pkg/controllers/termination/suite_test.go
```diff
@@ -89,6 +89,8 @@ var _ = Describe("Termination", func() {
 			Expect(env.Client.Delete(ctx, node)).To(Succeed())
 			node = ExpectNodeExists(env.Client, node.Name)
 			ExpectReconcileSucceeded(controller, client.ObjectKeyFromObject(node))
+			// Sleep for one second to simulate queue processing work
+			time.Sleep(1 * time.Second)
```
bwagner5, line 93: why do we need to sleep here?  I would try hard to not introduce sleeps into tests, it generally makes them flakey. IIUC, the test should not evict the pod, so if we continue w/o a sleep, I'm guessing you're worried that it could pass before the queue processes?  

I wonder if there's a way to synchronize the ExpectReconcileSucceeded to make sure the work queue is empty before proceeding.

pkg/controllers/termination/controller.go
```diff
@@ -37,25 +39,52 @@ import (
...
+func NewControllerWithCustomQueue(kubeClient client.Client, coreV1Client corev1.CoreV1Interface,
+	cloudProvider cloudprovider.CloudProvider, rateLimitingQueue workqueue.RateLimitingInterface) *Controller {
+	return &Controller{
+		KubeClient: kubeClient,
+		Terminator: &Terminator{kubeClient: kubeClient, cloudProvider: cloudProvider, coreV1Client: coreV1Client,
+			evictionQueue: EvictionQueue{
+				queue:        rateLimitingQueue,
+				coreV1Client: coreV1Client,
+				enqueued:     set.NewSet(),
+				once:         sync.Once{},
```
ellistarn, line 78: You shouldn't need to initialize this, since this is already the default value.

pkg/controllers/termination/eviction.go
```diff
@@ -0,0 +1,87 @@
...
+type EvictionQueue struct {
+	queue        workqueue.RateLimitingInterface
+	coreV1Client corev1.CoreV1Interface
+
+	enqueued set.Set
+	once     sync.Once
+}
+
+// Evict adds a pod to the EvictionQueue
+func (e *EvictionQueue) Add(pods []*v1.Pod) {
```
ellistarn, line 27: A variadic might be a bit cleaner here, since you could imagine evicting a single pod.

pkg/controllers/termination/eviction.go
```diff
@@ -0,0 +1,87 @@
...
+	// Start processing eviction queue if it hasn't started already
+	e.once.Do(func() { go e.run() })
+
+	for _, pod := range pods {
+		nn := types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}
+		if !e.enqueued.Contains(nn) {
```
ellistarn, line 33: define this inline
```
if nn := ...; e.enqueued.Contains(nn) {
...
}
```

pkg/controllers/termination/controller.go
```diff
@@ -37,25 +39,52 @@ import (
...
+			evictionQueue: EvictionQueue{
+				queue:        workqueue.NewRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(100*time.Millisecond, 10*time.Second)),
+				coreV1Client: coreV1Client,
+				enqueued:     set.NewSet(),
+				once:         sync.Once{},
+			}},
+	}
+}
+
+func NewControllerWithCustomQueue(kubeClient client.Client, coreV1Client corev1.CoreV1Interface,
```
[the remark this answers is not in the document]
njtran: I understand where you're coming from. I created this function precisely for the reason you're talking about, but also because I didn't want to make the EvictionQueue struct variables public.

Instead, I decided to make a terminator constructor. LMK what you think.

pkg/controllers/termination/eviction.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+	}
+}
+
+func (e *EvictionQueue) run() {
+	for {
+		// Get pod from queue. This waits until queue is non-empty.
+		item, shutdown := e.queue.Get()
+		if shutdown {
+			zap.S().Infof("Shutdown queue is broken")
```
ellistarn, line 45: I think we want to break here.

pkg/controllers/termination/eviction.go
```diff
@@ -0,0 +1,87 @@
...
+		e.queue.Done(nn)
+		if evicted {
+			zap.S().Debugf("Successfully evicted pod %s", nn.String())
+			e.queue.Forget(nn)
+			e.enqueued.Remove(nn)
+			continue
+		}
+		// Requeue pod if eviction failed
+		e.queue.AddRateLimited(nn)
+	}
```
ellistarn, line 60: I'd add a log line here that the queue is stopping for some reason

pkg/controllers/termination/eviction.go
```diff
@@ -0,0 +1,87 @@
...
+		return false
+	}
+	// 429
+	if errors.IsTooManyRequests(err) {
+		zap.S().Debugf("Failed to evict pod %s due to PDB violation. Backing off then trying again.", nn.String())
+		return false
+	}
+	// 404
+	if errors.IsNotFound(err) {
+		zap.S().Debugf("Continuing after failing to evict pod %s since it was not found.", nn.String())
```
ellistarn, line 80: I don't think we want this log line. I think this is just an idempotent success case.

pkg/controllers/termination/eviction.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// returns true if successful eviction call, error is returned if not eviction-related error
+func (e *EvictionQueue) evict(nn types.NamespacedName) bool {
+	err := e.coreV1Client.Pods(nn.Namespace).Evict(context.Background(), &v1beta1.Eviction{
+		ObjectMeta: metav1.ObjectMeta{Name: nn.Name, Namespace: nn.Namespace},
+	})
+	// 500
+	if errors.IsInternalError(err) {
+		zap.S().Debugf("Failed to evict pod %s due to PDB misconfiguration error. Backing off then trying again.", nn.String())
```
ellistarn, line 70: I'd remove the second clause "backing off", since you don't actually know that the code will back off in this context.

pkg/controllers/termination/suite_test.go
njtran: The EvictionQueue actually processes evictions independent of the reconcile loop. Because of this, I was trying to figure out how to make the program wait. Sometimes these tests would fail simply because the Reconcile loop succeeded so quickly that the EvictionQueue couldn't process the one pod in the queue and update the pod before we called GetPod in the next line.  

I played with the idea of using Eventually here, but since the way we see if a pod has successfully received an evict call is checking if the DeletionTimestamp is nonZero, propagation delay means that our Eventually would have to rely on a timeout for this to work, IIUC, which is the same as a sleep.

pkg/controllers/termination/controller.go
```diff
@@ -37,25 +39,52 @@ import (
...
 // NewController constructs a controller instance
 func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.CloudProvider) *Controller {
 	return &Controller{
-		terminator:    &Terminator{kubeClient: kubeClient, cloudProvider: cloudProvider, coreV1Client: coreV1Client},
-		cloudProvider: cloudProvider,
-		kubeClient:    kubeClient,
+		KubeClient: kubeClient,
+		Terminator: &Terminator{kubeClient: kubeClient, cloudProvider: cloudProvider, coreV1Client: coreV1Client,
+			evictionQueue: EvictionQueue{
+				queue:        workqueue.NewRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(100*time.Millisecond, 10*time.Second)),
```
ellistarn, line 61: Can you extract and name these constants?

pkg/controllers/termination/eviction.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (e *EvictionQueue) run() {
+	for {
+		// Get pod from queue. This waits until queue is non-empty.
+		item, shutdown := e.queue.Get()
+		if shutdown {
+			zap.S().Infof("Shutdown queue is broken")
+		}
+		// Panic if not a NamespacedName
```
ellistarn, line 47: I'd remove this comment.

pkg/controllers/termination/eviction.go
```diff
@@ -0,0 +1,87 @@
...
+		nn := item.(types.NamespacedName)
+		// Evict pod
+		evicted := e.evict(nn)
+		e.queue.Done(nn)
+		if evicted {
+			zap.S().Debugf("Successfully evicted pod %s", nn.String())
```
ellistarn, line 53: I'm trying to get away from the "Successfully" log lines, since it's typically already captured by the log type (i.e., we can assume success if its not errorf). I might simply say "Evicted pod %s"

pkg/controllers/termination/eviction.go
```diff
@@ -0,0 +1,87 @@
...
+	for {
+		// Get pod from queue. This waits until queue is non-empty.
+		item, shutdown := e.queue.Get()
+		if shutdown {
+			zap.S().Infof("Shutdown queue is broken")
+		}
+		// Panic if not a NamespacedName
+		nn := item.(types.NamespacedName)
+		// Evict pod
+		evicted := e.evict(nn)
```
ellistarn, line 50: Thoughts on

```
defer e.queue.Done(nn)
if e.evict(nn) {
  ...
  continue
}
e.queue.AddRateLimited(nn)
```

pkg/controllers/termination/eviction.go
```diff
@@ -0,0 +1,87 @@
...
+	if errors.IsInternalError(err) {
+		zap.S().Debugf("Failed to evict pod %s due to PDB misconfiguration error. Backing off then trying again.", nn.String())
+		return false
+	}
+	// 429
+	if errors.IsTooManyRequests(err) {
+		zap.S().Debugf("Failed to evict pod %s due to PDB violation. Backing off then trying again.", nn.String())
+		return false
+	}
+	// 404
```
ellistarn, line 78: consider inlining these commends
```
if errors.IsNotFound(err) { // 404
}
```

pkg/controllers/termination/suite_test.go
```diff
@@ -41,11 +43,9 @@ var controller *termination.Controller
 var env = test.NewEnvironment(func(e *test.Environment) {
 	cloudProvider := &fake.CloudProvider{}
 	registry.RegisterOrDie(cloudProvider)
-	controller = termination.NewController(
-		e.Client,
-		corev1.NewForConfigOrDie(e.Config),
-		cloudProvider,
-	)
+	coreV1Client := corev1.NewForConfigOrDie(e.Config)
```
ellistarn, line 46: inline?

pkg/controllers/termination/eviction.go
njtran: Thanks. Fun fact, since `types.NamespacedName{}` is init a new struct, we actually need to surround it by parentheses or the compiler will think it's the end of the if statement condition!

```
if nn := (types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}); !e.enqueued.Contains(nn) {
...
}
```

pkg/controllers/termination/controller.go
njtran: Good idea. Will do.

pkg/controllers/termination/eviction.go
njtran: Since the only code that is adding a pod is our code, I thought it would make more sense just to put an array here, since that's how we form the list of pods anyways. I have no strong feelings either way though.

pkg/controllers/termination/eviction.go
njtran: Since this is running in a continual forloop as a goroutine in the background, defer won't ever actually complete here, right? We would need to call it right after.

pkg/controllers/termination/terminate.go
```diff
@@ -17,26 +17,48 @@ package termination
...
+	set "github.com/deckarep/golang-set"
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	evictionQueueMaxDelay  = 10 * time.Second
```
ellistarn, line 37: Thoughts on pushing this to 60 seconds? I'm a bit wary of too much eviction QPS. 1,000 pods at 10 seconds is 100qps.

njtran: I'm a little wary of pushing it to 60 seconds. Imagine I have 5 pods, with a PDB that allows me to evict one at a time. If we assume that our nodes take 50 seconds to be fully provisioned, we have enough time for the eviction delay to get to reach a minute for any one of these pods. At the worst, we could wait a whole minute after the PDB would allow an eviction to evict another pod. Let's say these are all on the same node, this node would then take much much longer to terminate and drain.  

I realize this is a very specific case, but since our allocation logic tends to group pods that are created together (since scaling up deployments is the most basic happenstance of pending pods), our logic may end up running into something like this more than not.

I think we could fine tune these numbers when we get to performance and scale testing. What do you think?

pkg/controllers/termination/suite_test.go
bwagner5: I still haven't thought of a cleaner way to synchronize the queue for deterministic tests (without creating functionality in the queue that we only need for the tests), so I suppose this is okay for now...  We should definitely keep thinking on a minimalist way to get rid of this sleep though.

pkg/test/expectations/expectations.go
```diff
@@ -137,3 +145,37 @@ func ExpectReconcileSucceeded(reconciler reconcile.Reconciler, key client.Object
 	_, err := reconciler.Reconcile(context.Background(), reconcile.Request{NamespacedName: key})
 	Expect(err).ToNot(HaveOccurred())
 }
+
+func ExpectPodsEvictingSucceeded(c client.Client, pods ...*v1.Pod) {
```
ellistarn, line 149: Thoughts on calling this 
```ExpectEvicted(c client.Client, pods ...*v1.Pod)```
and then actually doing the deletion after we verify the deletion timestamp is set?

File: pkg/test/expectations/expectations.go [hunk 18]
```diff
@@ -137,3 +145,37 @@ func ExpectReconcileSucceeded(reconciler reconcile.Reconciler, key client.Object
...
+	}
+}
+
+func ExpectPodsEnqueuedForEviction(c client.Client, enqueued set.Set, pods ...*v1.Pod) {
+	for _, pod := range pods {
+		Expect(enqueued.Contains(types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace})).To(BeTrue())
+	}
+}
+
+func ExpectPodsNotEnqueuedForEviction(c client.Client, enqueued set.Set, pods ...*v1.Pod) {
```
ellistarn, line 177: IMO these methods are way to specific to the eviction logic to live in this package. I don't want to dilute the cohesion of this package, so can we put them at the bottom of suite_test.go in the eviction package?

ellistarn, line 177: ExpectEnqueuedForEviction

pkg/test/expectations/expectations.go
```diff
@@ -137,3 +145,37 @@ func ExpectReconcileSucceeded(reconciler reconcile.Reconciler, key client.Object
...
+		Eventually(func() bool {
+			return ExpectPodExists(c, pod.Name, pod.Namespace).GetDeletionTimestamp().IsZero() &&
+				queue.NumRequeues(types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}) > 0
+		}, ReconcilerPropagationTime, RequestInterval).Should(BeTrue(), func() string {
+			return fmt.Sprintf("expected %s/%s to not be evicting, but it is", pod.Namespace, pod.Name)
+		})
+	}
+}
+
+func ExpectPodsEnqueuedForEviction(c client.Client, enqueued set.Set, pods ...*v1.Pod) {
```
ellistarn, line 171: ExpectNotEnqueuedForEviction

pkg/controllers/termination/suite_test.go
```diff
@@ -116,29 +127,68 @@ var _ = Describe("Termination", func() {
...
 
-			// Terminate Node
+			// Reconcile to delete node
+			node = ExpectNodeExists(env.Client, node.Name)
+			ExpectReconcileSucceeded(controller, client.ObjectKeyFromObject(node))
+			ExpectNotFound(env.Client, node)
+		})
+		It("should fail to evict pods that violate a PDB", func() {
+			key, value := randomdata.SillyName(), randomdata.SillyName()
+			pdb := test.PodDisruptionBudget(test.PDBOptions{
```
ellistarn, line 159: What do you think about simplifying the options struct to avoid leaking some of this type conversion, e.g.
```
test.PDBOptions{
  LabelSelector: map[string]string{key: value},
  MinAvailable: 1,
}
```
Do the type conversion inside your helper.

njtran: It's a little different in that PDB options should be able to specify intstr.IntorString, but the default value of the non-pointer value is 0, which is still a valid PDB option. Since you can't specify both maxUnavailable and minAvailable, we then run into a PDB configuration issue.

pkg/controllers/termination/eviction.go
```diff
@@ -0,0 +1,82 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/util/workqueue"
+)
+
+type EvictionQueue struct {
+	queue        workqueue.RateLimitingInterface
+	coreV1Client corev1.CoreV1Interface
+
+	enqueued set.Set
```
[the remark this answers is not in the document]
njtran: Ellis and I talked about this, and we think there is a possibility to merge them here. We decided to punt on this merging effort later in favor of getting this feature in.

pkg/test/expectations/expectations.go
njtran: I would rather have this check just check it, then separately delete the pod, just in case there is a time where we don't immediately want to delete it, and check some other things first.

pkg/controllers/termination/suite_test.go
```diff
@@ -38,14 +43,21 @@ func TestAPIs(t *testing.T) {
...
+
 var env = test.NewEnvironment(func(e *test.Environment) {
 	cloudProvider := &fake.CloudProvider{}
 	registry.RegisterOrDie(cloudProvider)
-	controller = termination.NewController(
-		e.Client,
-		corev1.NewForConfigOrDie(e.Config),
-		cloudProvider,
-	)
+	evictionQueue = termination.NewEvictionQueue(corev1.NewForConfigOrDie(e.Config))
```
ellistarn, line 51: Don't duplicate the corev1 client.

pkg/test/pods.go
```diff
@@ -36,6 +38,15 @@ type PodOptions struct {
...
 	Annotations          map[string]string
+	Labels               map[string]string
+}
+
+type PDBOptions struct {
+	Name           string
+	Namespace      string
+	MinAvailable   *intstr.IntOrString
+	Selector       *metav1.LabelSelector
+	MaxUnavailable *intstr.IntOrString
```
ellistarn, line 49: I can't find the comment where I said this, but let's make this just an int.

pkg/test/pods.go
```diff
@@ -36,6 +38,15 @@ type PodOptions struct {
...
 	Conditions           []v1.PodCondition
 	Annotations          map[string]string
+	Labels               map[string]string
+}
+
+type PDBOptions struct {
+	Name           string
+	Namespace      string
+	MinAvailable   *intstr.IntOrString
+	Selector       *metav1.LabelSelector
```
ellistarn, line 48: I can't find the comment where I said this, but let's make this just a map[string]string

pkg/controllers/termination/suite_test.go
```diff
@@ -116,32 +131,106 @@ var _ = Describe("Termination", func() {
...
+	for _, pod := range pods {
+		Eventually(func() bool {
+			return ExpectPodExists(c, pod.Name, pod.Namespace).GetDeletionTimestamp().IsZero() && e.NumRequeues(podToNN(pod)) > 0
+		}, ReconcilerPropagationTime, RequestInterval).Should(BeTrue(), func() string {
+			return fmt.Sprintf("expected %s/%s to not be evicting, but it is", pod.Namespace, pod.Name)
+		})
+	}
+}
+
+func podToNN(pod *v1.Pod) types.NamespacedName {
```
ellistarn, line 234: There's already a function for this in client go. It's used elsewhere.

pkg/controllers/termination/suite_test.go
```diff
@@ -116,32 +131,106 @@ var _ = Describe("Termination", func() {
...
+	for _, pod := range pods {
+		Eventually(func() bool {
+			return ExpectPodExists(c, pod.Name, pod.Namespace).GetDeletionTimestamp().IsZero()
+		}, ReconcilerPropagationTime, RequestInterval).Should(BeFalse(), func() string {
+			return fmt.Sprintf("expected %s/%s to be evicting, but it isn't", pod.Namespace, pod.Name)
+		})
+	}
+}
+
+func ExpectEvictingFailed(e *termination.EvictionQueue, c client.Client, pods ...*v1.Pod) {
```
ellistarn, line 224: nit: I'd pass client first.

pkg/controllers/termination/suite_test.go
```diff
@@ -116,32 +131,106 @@ var _ = Describe("Termination", func() {
...
+
+			// Reconcile to delete node
 			node = ExpectNodeExists(env.Client, node.Name)
 			ExpectReconcileSucceeded(controller, client.ObjectKeyFromObject(node))
 			ExpectNotFound(env.Client, node)
 		})
 	})
 })
+
+func ExpectEnqueuedForEviction(e *termination.EvictionQueue, pods ...*v1.Pod) {
```
ellistarn, line 202: ExpectEvicting, ExpectNotEvicting

pkg/controllers/termination/suite_test.go
njtran: The sleep is gone now!

ellistarn (approved):

PR #518, opened 2021-07-19
Parallelizes capacity creation

ellistarn:
**Description of changes:**
Previously, the capacity API took multiple instances to allow cloud providers to batch create/get/delete API calls. However, this had a linearizing side effect such that all instances must be launched before karpenter was able to bind pods. This resulted in increased cases where the scheduler would schedule to capacity (launch by the linear loop) before Karpenter could execute the binds. Parallelizing also has the obvious side effect of making things launch a lot faster (unmeasured).

There are some long term considerations to this change:
1. Cloud providers will no longer (easily) be able to batch capacity API calls
2. Cloud providers will no longer be able to see multiple packings at once, which could allow for re-packing.

The primary benefits are:
1. The Capacity API is now simpler and more natural. 
2. Launching capacity for constraint groups is now parallelized in vendor neutral code.
3. A slow cloud provider implementation can no longer hang on linear capacity creation.
3. More code (node object construction, parallelization) is moved out of the AWS Cloud Provider and into Karpenter's core.

Additionally, I swapped the verbosity of a few log statements.

### DIFF
--- a/pkg/cloudprovider/aws/fake/ec2api.go
+++ b/pkg/cloudprovider/aws/fake/ec2api.go
@@ -40,10 +40,10 @@ type EC2Behavior struct {
 	DescribeInstanceTypesOutput         *ec2.DescribeInstanceTypesOutput
 	DescribeInstanceTypeOfferingsOutput *ec2.DescribeInstanceTypeOfferingsOutput
 	DescribeAvailabilityZonesOutput     *ec2.DescribeAvailabilityZonesOutput
-	CalledWithCreateFleetInput          set.Set //[]*ec2.CreateFleetInput
-	CalledWithCreateLaunchTemplateInput set.Set // []*ec2.CreateLaunchTemplateInput
+	CalledWithCreateFleetInput          set.Set
+	CalledWithCreateLaunchTemplateInput set.Set
 	Instances                           sync.Map
-	LaunchTemplates                     sync.Map //[]*ec2.LaunchTemplate
+	LaunchTemplates                     sync.Map
 }
 
 type EC2API struct {
@@ -317,13 +317,3 @@ func (e *EC2API) DescribeInstanceTypeOfferingsPagesWithContext(ctx context.Conte
 	}, false)
 	return nil
 }
-
-// func (e *EC2API) ExpectCalledWithFleetLaunchTemplateOverridesRequest(overrides ...*ec2.FleetLaunchTemplateOverridesRequest) {
-// 	results := []*ec2.FleetLaunchTemplateOverridesRequest{}
-// 	e.calledWithCreateFleetInput.Range(func(key, value interface{}) bool {
-// 		for _, lt := range value.(*ec2.CreateFleetInput).LaunchTemplateConfigs {
-// 			results = append(results, lt.Overrides...)
-// 		}
-// 		return true
-// 	})
-// }
--- a/pkg/cloudprovider/aws/fake/ec2api.go
+++ b/pkg/cloudprovider/aws/fake/ec2api.go
@@ -39,7 +39,7 @@ type EC2Behavior struct {
 	DescribeAvailabilityZonesOutput     *ec2.DescribeAvailabilityZonesOutput
 	CalledWithCreateFleetInput          []*ec2.CreateFleetInput
 	CalledWithCreateLaunchTemplateInput []*ec2.CreateLaunchTemplateInput
-	CreatedInstance                     *ec2.Instance
+	Instances                           map[string]*ec2.Instance
 	LaunchTemplates                     []*ec2.LaunchTemplate
 }
 
@@ -66,7 +66,10 @@ func (e *EC2API) CreateFleetWithContext(ctx context.Context, input *ec2.CreateFl
 		PrivateDnsName: aws.String(randomdata.IpV4Address()),
 		InstanceType:   input.LaunchTemplateConfigs[0].Overrides[0].InstanceType,
 	}
-	e.CreatedInstance = instance
+	if e.Instances == nil  {
+		e.Instances = map[string]*ec2.Instance{}
+	}
+	e.Instances[*instance.InstanceId] = instance
 	return &ec2.CreateFleetOutput{Instances: []*ec2.CreateFleetInstance{{InstanceIds: []*string{instance.InstanceId}}}}, nil
 }
 
@@ -77,12 +80,12 @@ func (e *EC2API) CreateLaunchTemplateWithContext(ctx context.Context, input *ec2
 	return &ec2.CreateLaunchTemplateOutput{LaunchTemplate: launchTemplate}, nil
 }
 
-func (e *EC2API) DescribeInstancesWithContext(context.Context, *ec2.DescribeInstancesInput, ...request.Option) (*ec2.DescribeInstancesOutput, error) {
+func (e *EC2API) DescribeInstancesWithContext(ctx context.Context, input *ec2.DescribeInstancesInput, options ...request.Option) (*ec2.DescribeInstancesOutput, error) {
 	if e.DescribeInstancesOutput != nil {
 		return e.DescribeInstancesOutput, nil
 	}
 	return &ec2.DescribeInstancesOutput{
-		Reservations: []*ec2.Reservation{{Instances: []*ec2.Instance{e.CreatedInstance}}},
+		Reservations: []*ec2.Reservation{{Instances: []*ec2.Instance{e.Instances[*input.InstanceIds[0]]}}},
 	}, nil
 }
 
--- a/pkg/cloudprovider/aws/fake/ec2api.go
+++ b/pkg/cloudprovider/aws/fake/ec2api.go
@@ -17,6 +17,7 @@ package fake
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/Pallinder/go-randomdata"
 	"github.com/aws/aws-sdk-go/aws"
@@ -39,7 +40,7 @@ type EC2Behavior struct {
 	DescribeAvailabilityZonesOutput     *ec2.DescribeAvailabilityZonesOutput
 	CalledWithCreateFleetInput          []*ec2.CreateFleetInput
 	CalledWithCreateLaunchTemplateInput []*ec2.CreateLaunchTemplateInput
-	Instances                           map[string]*ec2.Instance
+	Instances                           sync.Map
 	LaunchTemplates                     []*ec2.LaunchTemplate
 }
 
@@ -66,10 +67,7 @@ func (e *EC2API) CreateFleetWithContext(ctx context.Context, input *ec2.CreateFl
 		PrivateDnsName: aws.String(randomdata.IpV4Address()),
 		InstanceType:   input.LaunchTemplateConfigs[0].Overrides[0].InstanceType,
 	}
-	if e.Instances == nil  {
-		e.Instances = map[string]*ec2.Instance{}
-	}
-	e.Instances[*instance.InstanceId] = instance
+	e.Instances.Store(*instance.InstanceId, instance)
 	return &ec2.CreateFleetOutput{Instances: []*ec2.CreateFleetInstance{{InstanceIds: []*string{instance.InstanceId}}}}, nil
 }
 
@@ -84,8 +82,9 @@ func (e *EC2API) DescribeInstancesWithContext(ctx context.Context, input *ec2.De
 	if e.DescribeInstancesOutput != nil {
 		return e.DescribeInstancesOutput, nil
 	}
+	instance, _ := e.Instances.Load(*input.InstanceIds[0])
 	return &ec2.DescribeInstancesOutput{
-		Reservations: []*ec2.Reservation{{Instances: []*ec2.Instance{e.Instances[*input.InstanceIds[0]]}}},
+		Reservations: []*ec2.Reservation{{Instances: []*ec2.Instance{instance.(*ec2.Instance)}}},
 	}, nil
 }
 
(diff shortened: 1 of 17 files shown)


pkg/cloudprovider/aws/fake/ec2api.go
```diff
@@ -91,13 +98,13 @@ func (e *EC2API) DescribeLaunchTemplatesWithContext(ctx context.Context, input *
...
-			if launchTemplate.LaunchTemplateName == wanted {
-				output.LaunchTemplates = append(output.LaunchTemplates, launchTemplate)
-			}
+	e.LaunchTemplates.Range(func(key, value interface{}) bool {
+		launchTemplate := value.(*ec2.LaunchTemplate)
+		if functional.ContainsString(aws.StringValueSlice(input.LaunchTemplateNames), aws.StringValue(launchTemplate.LaunchTemplateName)) {
+			output.LaunchTemplates = append(output.LaunchTemplates, launchTemplate)
 		}
-	}
+		return true
```
njtran, line 106: Why would this function always be returning true? Wouldn't we want it to only return true if it does contain the string?

pkg/cloudprovider/aws/fake/ec2api.go
```diff
@@ -310,3 +317,13 @@ func (e *EC2API) DescribeInstanceTypeOfferingsPagesWithContext(ctx context.Conte
 	}, false)
 	return nil
 }
+
+// func (e *EC2API) ExpectCalledWithFleetLaunchTemplateOverridesRequest(overrides ...*ec2.FleetLaunchTemplateOverridesRequest) {
```
njtran, line 321: Did you mean to leave this in?

pkg/cloudprovider/aws/node.go
```diff
@@ -34,38 +34,27 @@ type NodeFactory struct {
...
-	// EC2 will return all instances if unspecified, so we must short circuit
-	if len(instanceIDs) == 0 {
-		return nil, nil
-	}
-	describeInstancesOutput, err := n.ec2api.DescribeInstancesWithContext(ctx, &ec2.DescribeInstancesInput{InstanceIds: instanceIDs})
-	if err == nil {
-		return n.nodesFrom(describeInstancesOutput.Reservations), nil
-	}
+func (n *NodeFactory) For(ctx context.Context, instanceId *string) (*v1.Node, error) {
+	describeInstancesOutput, err := n.ec2api.DescribeInstancesWithContext(ctx, &ec2.DescribeInstancesInput{InstanceIds: []*string{instanceId}})
```
njtran, line 38: What happens if the instanceID is nil, does this command fail, and is that okay?

pkg/cloudprovider/aws/node.go
```diff
@@ -34,38 +34,27 @@ type NodeFactory struct {
...
-
-func (n *NodeFactory) nodeFrom(instance *ec2.Instance) *v1.Node {
+	if len(describeInstancesOutput.Reservations) != 1 {
+		return nil, fmt.Errorf("expected a single instance reservation, got %d", len(describeInstancesOutput.Reservations))
+	}
+	if len(describeInstancesOutput.Reservations[0].Instances) != 1 {
+		return nil, fmt.Errorf("expected a single instance, got %d", len(describeInstancesOutput.Reservations[0].Instances))
+	}
+	instance := *describeInstancesOutput.Reservations[0].Instances[0]
+	zap.S().Infof("Launched instance: %s, type: %s, zone: %s, hostname: %s",
```
njtran, line 52: Is it possible to include this outside of the cloudprovider, so that the logging logic remains the same across cloud providers?

pkg/cloudprovider/types.go
```diff
@@ -27,7 +27,7 @@ import (
 // CloudProvider holds contains the methods necessary in a cloud provider
```
bwagner5, line 27: can you fix this comment while you're here? :)

pkg/controllers/allocation/controller.go
```diff
@@ -105,32 +104,38 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, fmt.Errorf("building constraint groups, %w", err)
 	}
 
+	// 5. Get Instance Types Options
+	instanceTypes, err := c.CloudProvider.GetInstanceTypes(ctx)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting instance types, %w", err)
+	}
+
 	// 5. Binpack each group
```
bwagner5, line 113: 6

pkg/controllers/allocation/controller.go
```diff
@@ -105,32 +104,38 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
-		if err != nil {
-			return reconcile.Result{}, fmt.Errorf("getting instance types, %w", err)
-		}
 		packings = append(packings, c.Packer.Pack(ctx, constraintGroup, instanceTypes)...)
 	}
 
-	// 6. Create packedNodes for packings and also copy all Status changes made by the
-	//    cloud provider to the original provisioner instance.
-	packedNodes, err := c.CloudProvider.Create(ctx, provisioner, packings)
+	// 6. Create capacity
```
bwagner5, line 119: 7

pkg/controllers/allocation/controller.go
```diff
@@ -105,32 +104,38 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	errs := make([]error, len(packings))
+	workqueue.ParallelizeUntil(ctx, len(packings), len(packings), func(index int) {
```
bwagner5, line 121: maybe we should limit the workers so that we don't get throttled immediately, 5 might be a good starting point based on https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits

pkg/controllers/allocation/bind.go
```diff
@@ -59,12 +60,11 @@ func (b *Binder) Bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) error
...
-		if err := b.bind(ctx, node, pod); err != nil {
-			zap.S().Errorf("Continuing after failing to bind, %s", err.Error())
-		}
-	}
-	return nil
+	errs := make([]error, len(pods))
+	workqueue.ParallelizeUntil(ctx, len(pods), len(pods), func(index int) {
+		errs[index] = b.bind(ctx, node, pods[index])
+	})
+	return multierr.Combine(errs...)
```
njtran, line 67: fancy code block

pkg/packing/packer.go
```diff
@@ -78,7 +78,7 @@ func (p *packer) Pack(ctx context.Context, constraints *Constraints, instances [
 			continue
 		}
 		packings = append(packings, packing)
-		zap.S().Debugf("Computed packing for pod(s) %v with instance type option(s) %s", apiobject.PodNamespacedNames(packing.Pods), instanceTypeNames(packing.InstanceTypeOptions))
+		zap.S().Infof("Computed packing for %d pod(s) with instance type option(s) %s", len(packing.Pods), instanceTypeNames(packing.InstanceTypeOptions))
```
njtran, line 81: In my experience, this tends to run very verbose. Do these changes reduce how much is being printed here? If not, do you have an example, because imo I think this would be a lot to print in logs.

pkg/controllers/reallocation/utilization.go
```diff
@@ -68,7 +68,7 @@ func (u *Utilization) markUnderutilized(ctx context.Context, provisioner *v1alph
 		if err := u.kubeClient.Patch(ctx, node, client.MergeFrom(persisted)); err != nil {
 			return fmt.Errorf("patching node %s, %w", node.Name, err)
 		}
-		zap.S().Debugf("Added TTL and label to underutilized node %s", node.Name)
+		zap.S().Infof("Added TTL and label to underutilized node %s", node.Name)
```
njtran, line 71: I think the two log lines in this file should be Debug. I think that adding TTL and removing TTL could happen a lot, and don't think it's as important as creating and terminating an instance.

pkg/cloudprovider/types.go
```diff
@@ -27,7 +27,7 @@ import (
...
 type CloudProvider interface {
 	// Create a set of nodes for each of the given constraints.
-	Create(context.Context, *v1alpha3.Provisioner, []*Packing) ([]*PackedNode, error)
+	Create(context.Context, *v1alpha3.Provisioner, *Packing) (*v1.Node, error)
```
bwagner5, line 30: I'm concerned this limits the cloud providers' ability to do any optimization. We're kind of brute-forcing our way out of it at the controller level.  At the expense of making the Create interface more complicated, would it be worth making it asynchronous with a richer response than just a v1.Node. The controller could track the actual provisioning status and drive the packings down until the desired state is reached.

bwagner5: One optimization that can probably happen pretty easily in the cloud provider currently is on the usual scale up case where Create will be launching a few or a bunch of the largest nodes possible. We can probably get significantly better scaling speeds for this case if we do one request for multiple instances. We should even be able to maintain a diverse capacity request since the pod density should be the same in this situation.

bwagner5: I don't want us to limit that capability as we find more cases or more capabilities are introduced to improve the cloud provider.

pkg/cloudprovider/aws/fake/ec2api.go
```diff
@@ -37,10 +40,10 @@ type EC2Behavior struct {
 	DescribeInstanceTypesOutput         *ec2.DescribeInstanceTypesOutput
 	DescribeInstanceTypeOfferingsOutput *ec2.DescribeInstanceTypeOfferingsOutput
 	DescribeAvailabilityZonesOutput     *ec2.DescribeAvailabilityZonesOutput
-	CalledWithCreateFleetInput          []*ec2.CreateFleetInput
-	CalledWithCreateLaunchTemplateInput []*ec2.CreateLaunchTemplateInput
-	Instances                           []*ec2.Instance
-	LaunchTemplates                     []*ec2.LaunchTemplate
+	CalledWithCreateFleetInput          set.Set //[]*ec2.CreateFleetInput
+	CalledWithCreateLaunchTemplateInput set.Set // []*ec2.CreateLaunchTemplateInput
+	Instances                           sync.Map
```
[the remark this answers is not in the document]
ellistarn: Oh this is actually a mistake -- I didn't mean to leave those comments in, lol.

pkg/cloudprovider/aws/fake/ec2api.go
ellistarn: The Range function continues if you return true. We want to process the entire list, so we always return true. The check for adding the LT to the list is on line 103.

pkg/cloudprovider/aws/node.go
ellistarn: Since it's not longer a list of instanceIDs and just a single one, our logical flow can never call this with []. The previous check was an awkward side effect of how the logic looped, without ever checking that we actually got an instance. Now that it's a single instance, we don't have this case.

pkg/cloudprovider/aws/node.go
ellistarn: Note that this code was just moved, but unfortunately, it's a little difficult to wire this up with our current return type being a node, which doesn't hold this specific information until it actually comes online.

pkg/cloudprovider/types.go
ellistarn: I completely agree with you here and played with the idea of using a callback model.

`Create(ctx, provisioner, packing, func(node v1.Node) error {})`

In the callback, we'd execute all of the bind logic.

Can we punt on this until we're ready to make more than one request at a time? I can also play with it here.

pkg/packing/packer.go
ellistarn: I'm now printing just the number of pods, rather than all of the pods. Since we limit instance-types to 20, it significantly simplifies this.

pkg/controllers/reallocation/utilization.go
ellistarn: Happy to discuss offline. I'm not sure this should be happening a lot, and if it is, I think it makes sense to be visible.

pkg/controllers/allocation/controller.go
ellistarn: I think this limiting should happen on the cloud provider side. I don't want to artificially limit things on the karp/core side now that we have the async interface. WDYT?

pkg/cloudprovider/types.go
ellistarn: Implemented this.

pkg/controllers/allocation/controller.go
bwagner5: yeah definitely on CP side 👍

pkg/cloudprovider/types.go
bwagner5: ha love it!

pkg/utils/parallel/workqueue.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// WorkQueue is a thread safe task runner
+type WorkQueue struct {
+	workqueue.RateLimitingInterface
+	once    sync.Once
+}
+
+// NewWorkQueue instantiates a new WorkQueue
+func NewWorkQueue(qps int, burst int) *WorkQueue {
```
bwagner5, line 17: nice!

bwagner5 (dismissed):
Very nice work on this!

bwagner5 (approved):
lgtm

PR #522, opened 2021-07-22
Cut v0.2.9 release

ellistarn:
**Issue, if available:**

**Description of changes:**


Makefile
```diff
@@ -17,7 +17,7 @@ dev: verify test ## Run all steps in the developer loop
 
 ci: verify licenses battletest ## Run all steps used by continuous integration
 
-release: publish helm docs ## Run all steps in release workflow
+release: codegen publish helm docs ## Run all steps in release workflow
```
akestner, line 20: What changed to necessitate including this here?

pkg/utils/parallel/workqueue.go
```diff
@@ -1,3 +1,17 @@
+/*
```
akestner, line 1: Are these license headers manually added?

akestner (approved):
Couple quick questions, but nothing that would require any changes

Makefile
ellistarn: I realized it was missing. It's just an extra safe check that all of our files have the correct headers, that apidocs are up to date, etc.

pkg/utils/parallel/workqueue.go
ellistarn: Added with `make codegen`

PR #529, opened 2021-07-22
add Name tag to default launch template

bwagner5:
**Issue, if available:**
N/A

**Description of changes:**
 - Adds a Name tag to the default Karpenter Launch Template (this is the launch template hardcoded in Karpenter that is used when a user does not override the launch template).  The Name tag in EC2 is a special tag that is used to fill the Name column of the EC2 Web Console. 
 - If a user overrides the launch template, they can specify any Name tag they'd like on the EC2 instance Karpenter will create.

The name format is `Karpenter/<Cluster-Name>` I can't add the K8s Node Name since the name doesn't exist until after an instance is launched. 

### DIFF
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -160,6 +160,10 @@ func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, optio
 				ResourceType: aws.String(ec2.ResourceTypeInstance),
 				Tags: []*ec2.Tag{
 					{
+						Key:   aws.String("Name"),
+						Value: aws.String(fmt.Sprintf("Karpenter/%s", ptr.StringValue(options.Cluster.Name))),
+					},
+					{
 						Key:   aws.String(fmt.Sprintf(ClusterTagKeyFormat, ptr.StringValue(options.Cluster.Name))),
 						Value: aws.String("owned"),
 					},


pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -159,6 +159,10 @@ func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, optio
 			TagSpecifications: []*ec2.LaunchTemplateTagSpecificationRequest{{
 				ResourceType: aws.String(ec2.ResourceTypeInstance),
 				Tags: []*ec2.Tag{
+					{
+						Key:   aws.String("Name"),
+						Value: aws.String(fmt.Sprintf("Karpenter/%s", ptr.StringValue(options.Cluster.Name))),
```
ellistarn, line 164: Won't this collide, since we have many LTs per cluster?

bwagner5: I don't think we can avoid collisions, maybe this should also have the provisioner as well, but it will still be the same name per instance per provisioner.

ellistarn: It's also not per node. It's per-unique-hash-of-scheduling-constraints.

ellistarn: I can see a reasonable argument that they should just all have the same name, but the philosophy behind AWS's naming concept is completely opaque to me.

ellistarn: Nevermind I was totally confused about what entity this was tagging. All good.

ellistarn (approved):

PR #534, opened 2021-07-23
Readjusted reallocation controller for nodes that fail to come up

njtran:
**Description of changes:**
This PR changes how reallocation controls the underutilization of nodes depending on their status. 

Previously we used to set nodes as underutilized and set a TTL (as specified in the Provisioner) for nodes, regardless of their status. This results in a race condition where nodes are brought up, and we could set a TTL before pods bind to them. This change now only allows us to do so for Ready nodes. 

In addition, this PR adds the feature to delete nodes that fail to become ready after 15 minutes, designed [here](https://github.com/awslabs/karpenter/blob/main/designs/termination.md#karpenter-availability-and-termination-cases). Since there are many reasons why nodes could become NotReady after becoming ready, we only consider nodes that were unable to come up, but not nodes that were able to come up but had connectivity issues, like KubeletNotReady. 

### DIFF
--- a/pkg/controllers/reallocation/controller.go
+++ b/pkg/controllers/reallocation/controller.go
@@ -24,6 +24,7 @@ import (
 	"golang.org/x/time/rate"
 	"knative.dev/pkg/logging"
 
+	"github.com/benbjohnson/clock"
 	"k8s.io/apimachinery/pkg/api/errors"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/util/workqueue"
@@ -36,17 +37,17 @@ import (
 
 // Controller for the resource
 type Controller struct {
-	utilization   *Utilization
-	cloudProvider cloudprovider.CloudProvider
-	kubeClient    client.Client
+	Utilization   *Utilization
+	CloudProvider cloudprovider.CloudProvider
+	KubeClient    client.Client
 }
 
 // NewController constructs a controller instance
 func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.CloudProvider) *Controller {
 	return &Controller{
-		utilization:   &Utilization{kubeClient: kubeClient},
-		cloudProvider: cloudProvider,
-		kubeClient:    kubeClient,
+		Utilization:   &Utilization{KubeClient: kubeClient, Clock: clock.New()},
+		CloudProvider: cloudProvider,
+		KubeClient:    kubeClient,
 	}
 }
 
@@ -56,7 +57,7 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	// 1. Retrieve provisioner from reconcile request
 	provisioner := &v1alpha3.Provisioner{}
-	if err := c.kubeClient.Get(ctx, req.NamespacedName, provisioner); err != nil {
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, provisioner); err != nil {
 		if errors.IsNotFound(err) {
 			return reconcile.Result{}, nil
 		}
@@ -64,7 +65,7 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	}
 
 	// 2. Delete any node that has been unable to come up for 5 minutes.
-	if err := c.utilization.terminateFailedToJoin(ctx, provisioner); err != nil {
+	if err := c.Utilization.terminateFailedToJoin(ctx, provisioner); err != nil {
 		return reconcile.Result{}, fmt.Errorf("terminating nodes that failed to join, %w", err)
 	}
 
@@ -72,18 +73,19 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	if provisioner.Spec.TTLSecondsAfterEmpty == nil {
 		return reconcile.Result{}, nil
 	}
+
 	// 3. Set TTL on TTLable Nodes
-	if err := c.utilization.markUnderutilized(ctx, provisioner); err != nil {
+	if err := c.Utilization.markUnderutilized(ctx, provisioner); err != nil {
 		return reconcile.Result{}, fmt.Errorf("adding ttl and underutilized label, %w", err)
 	}
 
 	// 4. Remove TTL from Utilized Nodes
-	if err := c.utilization.clearUnderutilized(ctx, provisioner); err != nil {
+	if err := c.Utilization.clearUnderutilized(ctx, provisioner); err != nil {
 		return reconcile.Result{}, fmt.Errorf("removing ttl from node, %w", err)
 	}
 
 	// 5. Delete any node past its TTL
-	if err := c.utilization.terminateExpired(ctx, provisioner); err != nil {
+	if err := c.Utilization.terminateExpired(ctx, provisioner); err != nil {
 		return reconcile.Result{}, fmt.Errorf("marking nodes terminable, %w", err)
 	}
 
--- a/pkg/controllers/reallocation/controller.go
+++ b/pkg/controllers/reallocation/controller.go
@@ -26,7 +26,6 @@ import (
 
 	"github.com/benbjohnson/clock"
 	"k8s.io/apimachinery/pkg/api/errors"
-	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/util/workqueue"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -43,7 +42,7 @@ type Controller struct {
 }
 
 // NewController constructs a controller instance
-func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.CloudProvider) *Controller {
+func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider) *Controller {
 	return &Controller{
 		Utilization:   &Utilization{KubeClient: kubeClient, Clock: clock.New()},
 		CloudProvider: cloudProvider,
--- a/pkg/controllers/reallocation/controller.go
+++ b/pkg/controllers/reallocation/controller.go
@@ -63,24 +63,30 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
+	// 2. Delete any node that has been unable to come up for 5 minutes.
+	if err := c.utilization.terminateFailedToBecomeReady(ctx, provisioner); err != nil {
+		return reconcile.Result{}, fmt.Errorf("terminating nodes that failed to become ready, %w", err)
+	}
+
 	// Skip reconciliation if utilization ttl is not defined.
 	if provisioner.Spec.TTLSecondsAfterEmpty == nil {
 		return reconcile.Result{}, nil
 	}
-	// 2. Set TTL on TTLable Nodes
+	// 3. Set TTL on TTLable Nodes
 	if err := c.utilization.markUnderutilized(ctx, provisioner); err != nil {
 		return reconcile.Result{}, fmt.Errorf("adding ttl and underutilized label, %w", err)
 	}
 
-	// 3. Remove TTL from Utilized Nodes
+	// 4. Remove TTL from Utilized Nodes
 	if err := c.utilization.clearUnderutilized(ctx, provisioner); err != nil {
 		return reconcile.Result{}, fmt.Errorf("removing ttl from node, %w", err)
 	}
 
-	// 4. Delete any node past its TTL
+	// 5. Delete any node past its TTL
 	if err := c.utilization.terminateExpired(ctx, provisioner); err != nil {
 		return reconcile.Result{}, fmt.Errorf("marking nodes terminable, %w", err)
 	}
+
 	return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
 }
 
--- a/pkg/controllers/reallocation/controller.go
+++ b/pkg/controllers/reallocation/controller.go
@@ -64,8 +64,8 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	}
 
 	// 2. Delete any node that has been unable to come up for 5 minutes.
-	if err := c.utilization.terminateFailedToBecomeReady(ctx, provisioner); err != nil {
-		return reconcile.Result{}, fmt.Errorf("terminating nodes that failed to become ready, %w", err)
+	if err := c.utilization.terminateFailedToJoin(ctx, provisioner); err != nil {
+		return reconcile.Result{}, fmt.Errorf("terminating nodes that failed to join, %w", err)
 	}
 
 	// Skip reconciliation if utilization ttl is not defined.
--- a/pkg/controllers/reallocation/controller.go
+++ b/pkg/controllers/reallocation/controller.go
@@ -24,7 +24,6 @@ import (
 	"golang.org/x/time/rate"
 	"knative.dev/pkg/logging"
 
-	"github.com/benbjohnson/clock"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/util/workqueue"
 	controllerruntime "sigs.k8s.io/controller-runtime"
@@ -44,7 +43,7 @@ type Controller struct {
 // NewController constructs a controller instance
 func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider) *Controller {
 	return &Controller{
-		Utilization:   &Utilization{KubeClient: kubeClient, Clock: clock.New()},
+		Utilization:   &Utilization{KubeClient: kubeClient},
 		CloudProvider: cloudProvider,
 		KubeClient:    kubeClient,
 	}
@@ -63,7 +62,7 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
-	// 2. Delete any node that has been unable to come up for 5 minutes.
+	// 2. Delete any node that has been unable to join.
 	if err := c.Utilization.terminateFailedToJoin(ctx, provisioner); err != nil {
 		return reconcile.Result{}, fmt.Errorf("terminating nodes that failed to join, %w", err)
 	}
--- a/pkg/controllers/reallocation/suite_test.go
+++ b/pkg/controllers/reallocation/suite_test.go
@@ -96,7 +96,7 @@ var _ = Describe("Reallocation", func() {
 			Expect(updatedNode.Labels).ToNot(HaveKey(v1alpha3.ProvisionerUnderutilizedLabelKey))
 			Expect(updatedNode.Annotations).ToNot(HaveKey(v1alpha3.ProvisionerTTLAfterEmptyKey))
 		})
-		It("should not TTL nodes that have ready status unknown", func() {
+		It("should not TTL nodes that have ready status false", func() {
 			node := test.Node(test.NodeOptions{
 				ReadyStatus: v1.ConditionFalse,
 			})
--- a/pkg/controllers/reallocation/suite_test.go
+++ b/pkg/controllers/reallocation/suite_test.go
@@ -26,26 +26,42 @@ import (
 	"github.com/awslabs/karpenter/pkg/cloudprovider/registry"
 	"github.com/awslabs/karpenter/pkg/controllers/reallocation"
 	"github.com/awslabs/karpenter/pkg/test"
-	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
-	"knative.dev/pkg/ptr"
 
 	. "github.com/awslabs/karpenter/pkg/test/expectations"
+	"github.com/benbjohnson/clock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	. "knative.dev/pkg/logging/testing"
+	"knative.dev/pkg/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var ctx context.Context
 var controller *reallocation.Controller
 var env *test.Environment
+
 func TestAPIs(t *testing.T) {
 	ctx = TestContextWithLogger(t)
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Provisioner/Reallocator")
 }
+
+var controller *reallocation.Controller
+var mock *clock.Mock
+var env = test.NewEnvironment(func(e *test.Environment) {
+	cloudProvider := &fake.CloudProvider{}
+	registry.RegisterOrDie(cloudProvider)
+	// Create a Mock Clock and set it to the current time
+	mock = clock.NewMock()
+	mock.Add(time.Duration(time.Now().Unix()) * time.Second)
+	controller = &reallocation.Controller{
+		Utilization:   &reallocation.Utilization{KubeClient: e.Client, Clock: mock},
+		CloudProvider: cloudProvider,
+		KubeClient:    e.Client,
+	}
+})
 
 var _ = BeforeSuite(func() {
 	env = test.NewEnvironment(ctx, func(e *test.Environment) {
@@ -132,7 +148,7 @@ var _ = Describe("Reallocation", func() {
 					v1alpha3.ProvisionerUnderutilizedLabelKey: "true",
 				},
 				Annotations: map[string]string{
-					v1alpha3.ProvisionerTTLAfterEmptyKey: time.Now().Add(time.Duration(100) * time.Second).Format(time.RFC3339),
+					v1alpha3.ProvisionerTTLAfterEmptyKey: time.Now().Add(100 * time.Second).Format(time.RFC3339),
 				},
 			})
 			ExpectCreated(env.Client, provisioner)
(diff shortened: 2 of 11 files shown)


pkg/utils/node/predicates.go
```diff
@@ -31,7 +31,28 @@ func IsReadyAndSchedulable(node v1.Node) bool {
...
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == v1.NodeReady {
+			return condition.Status == v1.ConditionTrue
+		}
+	}
+	return false
+}
+
+func FailsToBecomeReady(node v1.Node) bool {
+	if time.Since(node.GetCreationTimestamp().Time) < (15 * time.Minute) {
```
bwagner5, line 44: Can we pass this 15 min duration thru from maybe the reallocation controller (or if you don't want to pass thru that deep, at least in utilization.go)? The comment in the reallocation controller says 15 minutes but you have to dig down 2 more levels to actually see how to change the duration.  So I think the comment could get stale at some point or it'll just be awkward to configure this value.

pkg/controllers/reallocation/utilization.go
```diff
@@ -43,11 +43,15 @@ func (u *Utilization) markUnderutilized(ctx context.Context, provisioner *v1alph
 
 	// 2. Get underutilized nodes
 	for _, node := range nodes {
+		// Don't consider nodes that are not ready
```
bwagner5, line 46: can probably get rid of this comment, I think the next line is quite readable.

pkg/controllers/reallocation/utilization.go
```diff
@@ -122,6 +126,24 @@ func (u *Utilization) terminateExpired(ctx context.Context, provisioner *v1alpha
 	return nil
 }
 
+func (u *Utilization) terminateFailedToBecomeReady(ctx context.Context, provisioner *v1alpha3.Provisioner) error {
+	// 1. Get underutilized nodes
```
bwagner5, line 130: This is all provisioner nodes. This comment and the error message on line 133 is inaccurate.

pkg/utils/node/predicates.go
```diff
@@ -31,7 +31,28 @@ func IsReadyAndSchedulable(node v1.Node) bool {
...
+		return false
+	}
+	for _, condition := range node.Status.Conditions {
```
bwagner5, line 47: maybe we should make a helper func that puts these condition in a map or something so we don't have to do this ugly for-loop thing through here.

pkg/controllers/reallocation/controller.go
```diff
@@ -78,6 +78,11 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	if err := c.utilization.terminateExpired(ctx, provisioner); err != nil {
 		return reconcile.Result{}, fmt.Errorf("marking nodes terminable, %w", err)
 	}
+
+	// 5. Delete any node that has been unable to come up for 15 minutes.
+	if err := c.utilization.terminateFailedToBecomeReady(ctx, provisioner); err != nil {
```
bwagner5, line 83: This will be skipped if `provisioner.Spec.TTLSecondsAfterEmpty` is not specified.  I think we'd want to do this no matter what?

bwagner5 (commented):
A few nits and questions. Overall, looks great!

pkg/utils/node/predicates.go
ellistarn, line 44: Didn't we test this to be 5 minutes before pods are rescheduled?

ellistarn: It would be great if we could find the line of code upstream that does this (maybe in replicaset?)

ellistarn: I'm noticing 
```
  tolerations:
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
```
on my pods.

pkg/controllers/reallocation/controller.go
```diff
@@ -78,6 +78,11 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		return reconcile.Result{}, fmt.Errorf("terminating hanging NotReady nodes, %w", err)
```
ellistarn, line 84: "terminating nodes that failed to become ready"

pkg/controllers/reallocation/utilization.go
```diff
@@ -43,11 +43,15 @@ func (u *Utilization) markUnderutilized(ctx context.Context, provisioner *v1alph
...
+		// Don't consider nodes that are not ready
+		if !utilsnode.IsReady(*node) {
+			continue
+		}
 		pods, err := u.getPods(ctx, node)
 		if err != nil {
 			return fmt.Errorf("getting pods for node %s, %w", node.Name, err)
 		}
-		if utilsnode.IsEmpty(node, pods) {
+		if utilsnode.IsEmpty(pods) {
```
ellistarn, line 54: It's a bit odd that utilsnode wouldn't take a node object. This helper maybe makes sense in scheduling or pods?

pkg/utils/node/predicates.go
```diff
@@ -31,7 +31,28 @@ func IsReadyAndSchedulable(node v1.Node) bool {
...
+		if condition.Type == v1.NodeReady {
+			return condition.Status == v1.ConditionUnknown && condition.LastTransitionTime.IsZero() && condition.LastHeartbeatTime.IsZero()
```
ellistarn, line 49: I think checking ConditionUnknown should be sufficient.

pkg/utils/node/predicates.go
```diff
@@ -31,7 +31,28 @@ func IsReadyAndSchedulable(node v1.Node) bool {
...
+func IsReady(node v1.Node) bool {
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == v1.NodeReady {
+			return condition.Status == v1.ConditionTrue
+		}
+	}
+	return false
+}
+
+func FailsToBecomeReady(node v1.Node) bool {
```
ellistarn, line 43: This function name isn't correct, because a node could connect to the cluster, post Ready=False, and this function wouldn't return true. I think more accurately, this function is something like node.FailedToJoin().

pkg/utils/node/predicates.go
```diff
@@ -31,7 +31,28 @@ func IsReadyAndSchedulable(node v1.Node) bool {
 	return false
 }
 
-func IsPastEmptyTTL(node *v1.Node) bool {
+func IsReady(node v1.Node) bool {
```
ellistarn, line 34: *v1.Node

pkg/utils/node/predicates.go
njtran: You make a good point. While I don't think we want to make this super configurable, I would not like to lose the ability in levels, so I'll make this a parameter to pass in. Thanks!

pkg/utils/node/predicates.go
njtran: I agree. I added a helper to reduce some lines, lemme know what you think.

pkg/controllers/reallocation/controller.go
njtran: You're right. Although I don't know if we want to make a whole new controller for this. I could make it so that this function runs regardless of the specification of TTLSecondsAfterEmpty, but that might confuse those reading the code base. WDYT?

ellistarn: I think we need to step back and try to come up with the core logical concepts behind this, and how the behaviors relate to each other.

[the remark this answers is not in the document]
njtran: Will do.

pkg/controllers/reallocation/utilization.go
njtran: Although the logic is related to the node itself, I agree that it's weird. I'll move it.

pkg/utils/node/predicates.go
njtran: It looks like even if we set this value, these tolerations are configurable [here](
https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions), meaning that we could have severely different logic depending on user configuration. 

While 15 minutes doesn't capture all cases, not leaving any room for leeway after the default timeout value could lead to some race conditions in terms of propagating pod state. It could be that setting a default value like this might not be the answer for a lot of users.

pkg/utils/node/predicates.go
njtran: If it goes from Unknown -> False -> True -> Unknown for some reason ([example[(https://kubernetes.io/docs/concepts/architecture/nodes/#node-controller)), then these times would be nonZero. If we check that they're zero, then we know it has never advanced from Unknown to False.

pkg/utils/node/predicates.go
ellistarn: If only https://github.com/knative/pkg/blob/main/apis/condition_set.go

pkg/utils/node/predicates.go
ellistarn: How about we just check heartbeat? If we've never heard a heartbeat, then we know it was never alive.

pkg/utils/node/predicates.go
njtran: Talked offline, will also check to see if all the pods on the node are terminating as well.

pkg/utils/node/predicates.go
njtran: I wonder if it's possible for that to get erased. Can we ensure that status will be never be nulled? What if there's some issue with data retention?

pkg/controllers/reallocation/controller.go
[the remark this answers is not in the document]
njtran: Made an issue here. https://github.com/awslabs/karpenter/issues/555

pkg/utils/node/predicates.go
```diff
@@ -43,15 +49,11 @@ func IsPastEmptyTTL(node *v1.Node) bool {
...
 
-// IsEmpty returns if the node has 0 non-daemonset pods
-func IsEmpty(node *v1.Node, pods []*v1.Pod) bool {
-	for _, p := range pods {
-		if pod.HasFailed(p) {
-			continue
-		}
-		if !pod.IsOwnedByDaemonSet(p) {
-			return false
+func getNodeCondition(conditions []v1.NodeCondition, match v1.NodeConditionType) v1.NodeCondition {
```
ellistarn, line 52: nit: s/match/type would read more cleanly.

njtran: type is a keyword, so we can't use it.

pkg/controllers/reallocation/controller.go
```diff
@@ -24,8 +24,8 @@ import (
 	"golang.org/x/time/rate"
 	"knative.dev/pkg/logging"
 
+	"github.com/benbjohnson/clock"
```
ellistarn, line 27: You absolutely don't want a fake clock in our real code. Make sure this is only included in the test imports. This will ensure that golang's treeshaking doesn't build this package into our binary.

pkg/controllers/reallocation/controller.go
```diff
@@ -56,31 +56,38 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 	provisioner := &v1alpha3.Provisioner{}
-	if err := c.kubeClient.Get(ctx, req.NamespacedName, provisioner); err != nil {
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, provisioner); err != nil {
 		if errors.IsNotFound(err) {
 			return reconcile.Result{}, nil
 		}
 		return reconcile.Result{}, err
 	}
 
+	// 2. Delete any node that has been unable to come up for 5 minutes.
```
ellistarn, line 66: I'd avoid specifying 5 minutes here. This will become stale.

pkg/controllers/reallocation/suite_test.go
```diff
@@ -26,21 +26,23 @@ import (
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	. "knative.dev/pkg/logging/testing"
+	"knative.dev/pkg/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var ctx context.Context
 var controller *reallocation.Controller
+var mock *clock.Mock
```
ellistarn, line 43: I'd name this clock, not mock.

pkg/controllers/reallocation/suite_test.go
```diff
@@ -51,11 +53,14 @@ var _ = BeforeSuite(func() {
...
 		cloudProvider := &fake.CloudProvider{}
 		registry.RegisterOrDie(cloudProvider)
-		controller = reallocation.NewController(
-			e.Client,
-			corev1.NewForConfigOrDie(e.Config),
-			cloudProvider,
-		)
+		// Create a Mock Clock and set it to the current time
+		mock = clock.NewMock()
+		mock.Add(time.Duration(time.Now().Unix()) * time.Second)
```
ellistarn, line 58: Why does it need to be now?

pkg/controllers/reallocation/suite_test.go
```diff
@@ -122,5 +155,61 @@ var _ = Describe("Reallocation", func() {
...
+			// Set pod DeletionTimestamp and do another reconcile
+			Expect(env.Client.Delete(ctx, pod)).To(Succeed())
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(provisioner))
+
+			// Expect node not deleted
+			updatedNode = ExpectNodeExists(env.Client, node.Name)
+			Expect(updatedNode.DeletionTimestamp.IsZero()).To(BeTrue())
+
+			// Simulate 500 seconds passing and do another reconcile
+			mock.Add(500 * time.Second)
```
ellistarn, line 207: How does this help us test our implementation? Isn't this just a fancy sleep?

njtran: With a sleep we would wait 500 seconds, this just sets the mock clock's time to +500 seconds.

pkg/controllers/reallocation/suite_test.go
```diff
@@ -122,5 +155,61 @@ var _ = Describe("Reallocation", func() {
...
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(provisioner))
+
+			// Expect node deleting
```
ellistarn, line 210: I think you can remove this comment and the line above.

pkg/controllers/reallocation/suite_test.go
```diff
@@ -122,5 +155,61 @@ var _ = Describe("Reallocation", func() {
...
+
+			// Set pod DeletionTimestamp and do another reconcile
+			Expect(env.Client.Delete(ctx, pod)).To(Succeed())
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(provisioner))
+
+			// Expect node not deleted
+			updatedNode = ExpectNodeExists(env.Client, node.Name)
+			Expect(updatedNode.DeletionTimestamp.IsZero()).To(BeTrue())
+
+			// Simulate 500 seconds passing and do another reconcile
```
ellistarn, line 206: I'd say something like, "Simulate time passing and a node failing to join"

pkg/utils/node/predicates.go
```diff
@@ -18,17 +18,24 @@ import (
 	"time"
 
 	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha3"
-	"github.com/awslabs/karpenter/pkg/utils/pod"
+	"github.com/benbjohnson/clock"
```
ellistarn, line 21: Likewise.

pkg/utils/pod/scheduling.go
```diff
@@ -45,6 +45,29 @@ func IsSchedulable(pod *v1.PodSpec, node *v1.Node) bool {
...
+		if !IsOwnedByDaemonSet(p) {
+			return false
+		}
+	}
+	return true
+}
+
+// AreAllTerminating returns true if the set of pods were all unable to join
+func AreAllTerminating(pods []*v1.Pod) bool {
+	for _, p := range pods {
```
ellistarn, line 63: I think you want to reuse your `IgnoredForUnderutilization` helper, since this code should also take into account `HasFailed`. It does make the naming get a bit weird, so maybe we should just check HasFailed here.

I think there must be a way to factor this to share the concept. It is the same concept after all, we just need a name for it.

pkg/utils/pod/scheduling.go
```diff
@@ -45,6 +45,29 @@ func IsSchedulable(pod *v1.PodSpec, node *v1.Node) bool {
...
+		}
+		if !IsOwnedByDaemonSet(p) {
+			return false
+		}
+	}
+	return true
+}
+
+// AreAllTerminating returns true if the set of pods were all unable to join
+func AreAllTerminating(pods []*v1.Pod) bool {
```
ellistarn, line 62: Is this function the same as IgnoredForTermination or something different?

njtran: IgnoredForTermination feels like we're saying we want to terminate these pods, while in reality they are already terminating. I like this name more.

pkg/utils/pod/scheduling.go
njtran: These are a little different. We decided that we only wanted to terminate the node if all of the pods were terminating. This means we need to check the deletionTimestamp, whereas the ignoredForUnderutilization also checks if they failed, which doesn't necessarily mean they were requested to be deleted yet.

pkg/controllers/reallocation/suite_test.go
```diff
@@ -122,5 +151,64 @@ var _ = Describe("Reallocation", func() {
...
+			Expect(env.Client.Delete(ctx, pod)).To(Succeed())
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(provisioner))
+
+			// Expect node not deleted
+			updatedNode = ExpectNodeExists(env.Client, node.Name)
+			Expect(updatedNode.DeletionTimestamp.IsZero()).To(BeTrue())
+
+			future := time.Now().Add(500 * time.Second)
+			// Monkey patch time.Now() so it returns 500 seconds in the future
+			monkey.Patch(time.Now, func() time.Time {
```
ellistarn, line 204: consider `monkey.Patch(time.Now, func() time.Time { time.Now().Add(500 * time.Second) })`

pkg/controllers/reallocation/utilization.go
```diff
@@ -109,12 +112,34 @@ func (u *Utilization) terminateExpired(ctx context.Context, provisioner *v1alpha
...
+}
+
+func (u *Utilization) terminateFailedToJoin(ctx context.Context, provisioner *v1alpha3.Provisioner) error {
+	// 1. Get nodes
+	nodes, err := u.getNodes(ctx, provisioner, map[string]string{})
+	if err != nil {
+		return fmt.Errorf("listing nodes, %w", err)
+	}
+	// 2. Trigger termination workflow if node has failed to become ready for 5 minutes
+	for _, node := range nodes {
```
ellistarn, line 134: if you did `_, n : = range nodes` then later you could s/nodeutils/node

pkg/controllers/reallocation/utilization.go
```diff
@@ -109,12 +112,34 @@ func (u *Utilization) terminateExpired(ctx context.Context, provisioner *v1alpha
...
+		pods, err := u.getPods(ctx, node)
+		if err != nil {
+			logging.FromContext(ctx).Debugf("Continuing after failing to get pods for node %s", node.Name)
+		}
+		// If the node failed to join and all pods are terminating, delete the pod
+		if utilsnode.FailedToJoin(node, FailedToJoinTimeout) && pod.AreAllTerminating(pods) {
+			logging.FromContext(ctx).Infof("Deleting node %s after it failed to join in 5 minutes", node.Name)
```
ellistarn, line 141: Let's create parity with
`logging.FromContext(ctx).Infof("Triggering termination for empty node %s", node.Name)`

e.g.
`logging.FromContext(ctx).Infof("Triggering termination for node that failed to join %s", node.Name)`

pkg/controllers/reallocation/suite_test.go
njtran: Can't reference `time.Now` while we patch `time.Now`'s function address. This results in a buffer overflow.

pkg/controllers/reallocation/utilization.go
njtran: I don't see it as an issue here. Node reads cleaner to me. Fine to change it as well.

pkg/controllers/reallocation/suite_test.go
ellistarn: Interesting! Still, you could inline 204-206.

pkg/controllers/reallocation/utilization.go
```diff
@@ -109,12 +112,34 @@ func (u *Utilization) terminateExpired(ctx context.Context, provisioner *v1alpha
...
+	// 1. Get nodes
+	nodes, err := u.getNodes(ctx, provisioner, map[string]string{})
+	if err != nil {
+		return fmt.Errorf("listing nodes, %w", err)
+	}
+	// 2. Trigger termination workflow if node has failed to become ready for 5 minutes
+	for _, node := range nodes {
+		pods, err := u.getPods(ctx, node)
+		if err != nil {
+			logging.FromContext(ctx).Debugf("Continuing after failing to get pods for node %s", node.Name)
```
ellistarn, line 137: I don't think we want to continue here.

pkg/test/expectations/expectations.go
```diff
@@ -79,7 +79,9 @@ func ExpectDeleted(c client.Client, objects ...client.Object) {
 		persisted := object.DeepCopyObject()
 		object.SetFinalizers([]string{})
 		Expect(c.Patch(context.Background(), object, client.MergeFrom(persisted))).To(Succeed())
-		Expect(c.Delete(context.Background(), object, &client.DeleteOptions{GracePeriodSeconds: ptr.Int64(0)})).To(Succeed())
+		if err := c.Delete(context.Background(), object, &client.DeleteOptions{GracePeriodSeconds: ptr.Int64(0)}); !errors.IsNotFound(err) {
```
ellistarn, line 82: Which type of error were you getting that made you need to change this line?

pkg/utils/node/predicates.go
```diff
@@ -18,17 +18,23 @@ import (
...
 	v1 "k8s.io/api/core/v1"
 )
 
-func IsReadyAndSchedulable(node v1.Node) bool {
-	for _, condition := range node.Status.Conditions {
-		if condition.Type == v1.NodeReady {
-			return condition.Status == v1.ConditionTrue && !node.Spec.Unschedulable
-		}
+func IsReadyAndSchedulable(node *v1.Node) bool {
+	return getNodeCondition(node.Status.Conditions, v1.NodeReady).Status == v1.ConditionTrue && !node.Spec.Unschedulable
```
ellistarn, line 25: return IsReady(node) && !node.Spec.Unschedulable

pkg/utils/node/predicates.go
```diff
@@ -18,17 +18,23 @@ import (
...
+}
+
+func IsReady(node *v1.Node) bool {
+	return getNodeCondition(node.Status.Conditions, v1.NodeReady).Status == v1.ConditionTrue
+}
+
+func FailedToJoin(node *v1.Node, duration time.Duration) bool {
```
ellistarn, line 32: I might call this "grace period"

Now that I think about it, what do you think about using the ObjectMeta graceperiodseconds field?
https://github.com/kubernetes/apimachinery/blob/v0.21.3/pkg/apis/meta/v1/meta.go#L159

pkg/test/expectations/expectations.go
njtran: I was getting a notFound error

pkg/utils/pod/scheduling.go
ellistarn: Sorry my typo was misleading. I meant:
>Is this function the same as IgnoredForUnderutilizationor something different?

What I mean is, the logic looks very similar to me. The only difference is that they behave differently w/ deletion timestamp. They both should do the same thing for daemonsets and failed pods.

pkg/utils/pod/scheduling.go
ellistarn: What if a pod is failed (i.e. scheduling misbehavior)? I think we should still terminate the node in this case. This is for the exact same reason you want to consider failed pods in the other function. 

This similarity leads me to believe that there is common behavior that should be factored out.

pkg/controllers/reallocation/utilization.go
njtran: I added a continue

pkg/utils/node/predicates.go
```diff
@@ -18,17 +18,23 @@ import (
...
+	return getNodeCondition(node.Status.Conditions, v1.NodeReady).Status == v1.ConditionTrue
+}
+
+func FailedToJoin(node *v1.Node, gracePeriod time.Duration) bool {
+	if time.Since(node.GetCreationTimestamp().Time) < gracePeriod {
+		return false
 	}
-	return false
+	condition := getNodeCondition(node.Status.Conditions, v1.NodeReady)
+	return condition.Status == v1.ConditionUnknown && condition.LastHeartbeatTime.IsZero()
```
ellistarn, line 37: thoughts on `getNodeCondition(node.Status.Conditions, v1.NodeReady).LastHeartbeatTime.IsZero()`

pkg/controllers/reallocation/utilization.go
```diff
@@ -109,12 +112,29 @@ func (u *Utilization) terminateExpired(ctx context.Context, provisioner *v1alpha
...
 		return fmt.Errorf("listing underutilized nodes, %w", err)
 	}
-
 	// 2. Trigger termination workflow if past TTLAfterEmpty
 	for _, node := range nodes {
 		if utilsnode.IsPastEmptyTTL(node) {
 			logging.FromContext(ctx).Infof("Triggering termination for empty node %s", node.Name)
-			if err := u.kubeClient.Delete(ctx, node); err != nil {
+			if err := u.KubeClient.Delete(ctx, node); err != nil {
+				return fmt.Errorf("sending delete for node %s, %w", node.Name, err)
```
ellistarn, line 120: consider `deleting node`. Sending delete for is a bit of a nod and wink to finalizers, which we know may or may not exist for all kubernetes objects. I'd hate to be forced into "sending" language everywhere because of this. Easier to just use the rest concept `delete`, regardless of finalization hooks.

ellistarn (approved):

PR #535, opened 2021-07-23
Comprehensive logging improvements

ellistarn:
**Issue, if available:**

**Description of changes:**
- Migrates to knative style logging
- Logging contexts are now propagated via ctx
- Logging configuration is now controlled via configmap
- Logging level can be dynamically adjusted by updating the configmap.


cmd/controller/main.go
```diff
@@ -49,34 +53,36 @@ func init() {
...
+	config := controllerruntime.GetConfigOrDie()
+	clientSet := kubernetes.NewForConfigOrDie(config)
+
+	// Setup knative/pkg style logging
+	ctx, startinformers := injection.EnableInjectionOrDie(signals.NewContext(), config)
+	go startinformers()
+	logger, atomicLevel := sharedmain.SetupLoggerOrDie(ctx, component)
+	ctx = logging.WithLogger(ctx, logger)
+	cmw := informer.NewInformedWatcher(clientSet, system.Namespace())
+	cmw.Watch(logging.ConfigMapName(), logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
```
njtran, line 74: This code block looks honestly pretty confusing at first, and clutters up the main() function. I see the comment, but is it possible to separate this out into a helper function that summarizes what is happening here, like `SetUpLogging()`? 

I'd love to understand what is being used to set up logging in Karpenter without having to dive deep into another package.

charts/karpenter/templates/webhook/webhooks.yaml
```diff
@@ -50,3 +50,19 @@ webhooks:
     - CREATE
     - UPDATE
     - DELETE
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: validation.webhook.config.karpenter.sh
```
njtran, line 56: Maybe this is the magic of knative, but why do we now need to validate configmaps? Is it possible for any user to change the configmap that wasn't already done before, or did we just not have this as needed?

pkg/apis/provisioning/v1alpha3/provisioner_defaults.go
```diff
@@ -39,48 +39,48 @@ func (s *ProvisionerSpec) SetDefaults(ctx context.Context) {}
...
-func (c *Cluster) withDynamicDefaults() (_ Cluster, err error) {
+func (c *Cluster) withDynamicDefaults(ctx context.Context) (_ Cluster, err error) {
 	cluster := *c.DeepCopy()
-	cluster.CABundle, err = cluster.getCABundle()
+	cluster.CABundle, err = cluster.getCABundle(ctx)
 	return cluster, err
 }
 
-func (c *Cluster) getCABundle() (*string, error) {
+func (c *Cluster) getCABundle(ctx context.Context) (*string, error) {
```
njtran, line 66: WDYT about removing the debug lines in this function? Regardless of where we get the CABundle from, this usually floods the logs with repetitive messages, even in the Debug mode. Is it possible to only log once from this? Maybe just not log at all?

njtran (commented):
Great work. This is some really nice improvements to our logging. Could you include a change to show how to adjust the logging in the configmaps, considering that's two of the changes? (forgive me if you did and I missed it)

charts/karpenter/templates/webhook/rbac.yaml
```diff
@@ -31,14 +31,9 @@ metadata:
...
 rules:
-- apiGroups:
-  - ""
-  resources:
-  - configmaps
-  verbs:
-  - get
-  - list
-  - watch
+- apiGroups: [""]
```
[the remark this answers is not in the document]
ellistarn: I think it's a lot easier to read when condensed like this. It's following the tekton/knative projects' example.

charts/karpenter/templates/webhook/webhooks.yaml
ellistarn: We're validating the logging configuration.

pkg/apis/provisioning/v1alpha3/provisioner_defaults.go
ellistarn: I want to cache these guys. I'll remove the log lines now and deal with the cache in another PR.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -29,15 +29,15 @@ import (
...
 	v1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/apis"
+	"knative.dev/pkg/logging"
 )
 
 const (
 	// CreationQPS limits the number of requests per second to CreateFleet
 	// https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html#throttling-limits
-	CreationQPS = 2
+	CreationQPS = 10
```
bwagner5, line 40: why are we raising this to 10?  I think raising burst makes sense, but I don't think raising this to 10 is a good idea, maybe 5 ?

Makefile
```diff
@@ -43,7 +43,6 @@ licenses: ## Verifies dependency licenses and requires GITHUB_TOKEN to be set
 	golicense hack/license-config.hcl karpenter
 
 apply: ## Deploy the controller into your ~/.kube/config cluster
-	kubectl create ns karpenter || true
 	helm template karpenter charts/karpenter --namespace karpenter \
```
bwagner5, line 46: don't we need `--create-namespace` for a fresh install to work?

ellistarn: This workflow assumes that the namespace isn't created or deleted. I think it's much safer to just assume that the developer manages this namespace since there are so many ways it could be created (helm, eksctl, etc) that have bitten us. Worst case, a new dev needs to run `kubectl create namespace karpenter`

charts/karpenter/templates/webhook/rbac.yaml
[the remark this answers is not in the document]
njtran: Could you change the rest of the file to match this format?

charts/karpenter/templates/webhook/webhooks.yaml
njtran: After reading it more I understand that now. Thanks.

njtran (approved):
Quite a lengthy one. Good work!

charts/karpenter/templates/webhook/rbac.yaml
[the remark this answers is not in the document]
ellistarn: https://github.com/awslabs/karpenter/issues/554

PR #541, opened 2021-07-23
Add Favicon and Brand Images

rothgar:
**Description of changes:**
Adds svg and ico favicons for main site.


rothgar:
Before merging can someone with a mac test it in safari? The favicon _should_ switch color with light/dark mode. Right now it switches between #5c62b0 and #8890e2 which are the banner colors but we can make it switch between black/white if we like that batter.

website/layouts/partials/favicons.html
```diff
@@ -0,0 +1,4 @@
+<head>
```
akestner, line 1: Did it not cooperate out of the box once the favicons were in the right directory? The [Docsy docs](https://www.docsy.dev/docs/adding-content/iconsimages/#add-your-favicons suggest that you only need the partial if you want to customize

akestner (commented):
I have a better overview diagram than the one here that I was planning on adding to the landing page. Let's drop that one, but other wise looks good

geoffcline:
Attaching screenshots of light/dark mode in safari on macos. 

Screen Shot 2021-07-23 at 4 44 02 PM
Screen Shot 2021-07-23 at 4 44 10 PM

website/layouts/partials/favicons.html
rothgar: I tried dropping the favicon.ico and favicon.svg in static/favicons/ and it didn't work for me. Looking at the theme source (docsy/layouts/partials/favicons.html) it appears they only support .ico and .png files which means we couldn't do dark mode/light mode switching or svgs.

akestner (approved):

akestner (changes requested):
Actually, I have to rescind my approval since I think deleting the `images/**` files will break various non-website markdown files, e.g. `README.md`. 

Can you please make sure any references are updated @rothgar?

rothgar:
Good catch. Just removed the overview image (deleted) and updated the other two links.

website/layouts/partials/favicons.html
akestner: Good to know, I'm fine including the partial, just wanted to understand the rationale.

akestner (approved):

Issue #549, opened 2021-07-25
EC2 Instances leaked after 1,000 node scale down.


ellistarn:
## Version

**Karpenter**: v0.2.9

**Kubernetes**: v1.0.0

## Expected Behavior
I expected all of the instances to be deleted.

## Actual Behavior
Around 10% of the instances were leaked. I believe this may be related to the bug where nodename == "" and we fail to register the node with the API Server. 

## Steps to Reproduce the Problem
- 30 second underutilization ttl
- 1000 pod scale up
- limit provisioner to m5.large

njtran:
If I'm understanding this correctly, then this should be unrelated to scale down. We only delete a node if a termination request is successful, where we wouldn't be able to leak the instance. 

It seems like this issue is related to allocation. When Karpenter creates too many instances/nodes at the same time, it results in a successful instance creation but not a successful node creation, resulting in a leaked instance on start up time. 

Can you post the logs that you saw here to help narrow down where this issue is originating from?

ellistarn:
```karpenter-controller-79f7787f57-rlkjd manager 2021-07-26T21:22:37.147Z	ERROR	controller.controller.Reallocation	Reconciler error	{"reconciler group": "karpenter.sh", "reconciler kind": "Provisioner", "name": "default", "namespace": "", "error": "marking nodes terminable, sending delete for node ip-192-168-10-226.us-west-2.compute.internal, nodes \"ip-192-168-10-226.us-west-2.compute.internal\" not found"}

```

ellistarn:
I'm noticing that when nodes ~1000, node names start to be reused, depending on the size of the subnet CIDR.

ellistarn:
Current hypothesis.

If nodes come online before we're able to create them in the API Server, they may not get finalizers added to them. This will result in kubectl delete node leaking the actual instance. 

Looking at logs, I don't see termination lines for these instances.

PR #553, opened 2021-07-26
Fixed bug in aws cloudprovider where API machinery IsNotFound was incorrectly used on ec2 Terminate resp

bwagner5:
**Issue, if available:**
N/A

**Description of changes:**
 - Fixes a bug where API Machinery errors.IsNotFound() was being used to evaluate the error response from an AWS sdk call to EC2. The specific call is checking if the TerminateInstances API returns an "Instance ID does not exist" error which would indicate that the instance was already terminated or for whatever other reason, no longer exists. In this case, Karpenter should move along and delete the nodes from the k8s API server.  

TLDR: 
The logic was correct, but the wrong error check was used which caused nodes to hang when the ec2 instance had already been termianted.

### DIFF
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -20,13 +20,17 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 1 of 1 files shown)


ellistarn:
Related to https://github.com/awslabs/karpenter/issues/549?

ellistarn (approved):
Nice!

bwagner5:
> Related to #549?

I think that is a different issue. I suspect you're correct about the NodeName == ""

njtran (approved):
lgtm

PR #557, opened 2021-07-27
Add AWS Cloud Provider Docs

geoffcline:
**Description of changes:**
Add AWS cloud provider page.
Add description of well known labels and corresponding behavior on AWS. For example, launch nodes in a particular availability zone.  


geoffcline:
Changed page https://deploy-preview-557--karpenter-docs-prod.netlify.app/docs/cloud-providers/aws/

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+    spec:
+      nodeSelector:
+        node.k8s.aws/instance-type: m5.large
+```
+
+## Node Lables in Provisioner Spec
+
+### Launch Template
+
+Karpenter uses [AWS Bottlerocket](https://aws.amazon.com/bottlerocket/) by default.
```
rothgar, line 60: Maybe say AWS Bottlerocket OS by default. I'm not even sure everyone will know Bottlerocket is an OS type.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+
+You can specify a different [launch
+template](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-launch-templates.html),
+such as a customized version of Bottlerocket or a different platform
+altogether.
+
+**Template ID**
+- key: `node.k8s.aws/launch-template-id`
+- value example: `bottlerocket` (default)
+- value list: `aws ec2 describe-launch-templates`
```
rothgar, line 70: We should specify that Karpenter automatically creates a launch template with the name `Karpenter-<cluster name>-uuid` in the account/region where it is provisioned.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+### Architecture (e.g., ARM) 
+
+- key: `kubernetes.io/arch`
+- values
+  - `amd64` (default)
+  - `arm64`
+
+Karpenter supports `amd64` (e.g., intel) nodes, and `arm64` nodes. The default
```
rothgar, line 118: Instead of "intel" I would either leave off the example or say x86_64 since technically Karpenter may use a *a instance type which uses AMD processors.

"The default is `amd64`" is redundant since you say that above.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+        kubernetes.io/arch: amd64
+```
+
+## Operating System
+
+- key: `kubernetes.io/os`
+- values
+  - `linux` (default)
+
+At this time, Karpenter only supports linux nodes.
```
rothgar, line 147: nit: "linux os"

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+
+### AWS Region
+
+`topology.kubernetes.io/region=us-east-1`
```
rothgar, line 151: Should this be `=` or `:`?

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Operating System
+
+- key: `kubernetes.io/os`
+- values
+  - `linux` (default)
+
+At this time, Karpenter only supports linux nodes.
+
+### AWS Region
```
rothgar, line 149: Do we support karpenter making nodes in a different region than where it is running? That seems like a bad idea.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+
+- key: `topology.kubernetes.io/region`
+- value example: `us-east-1`
+- value list: [AWS regional endpoints](https://docs.aws.amazon.com/general/latest/gr/rande.html)
+
+Karpenter can be configured to create nodes in a particular region.
+
+### Availability Zones
+
+`topology.kubernetes.io/zone=us-east-1c`
```
rothgar, line 161: Same question as above. Is this `=` or `:`?

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+
+*Set Default with provisioner.yaml*
+
+```yaml
+spec:
+  instanceTypes:
+    - m5.large
+```
+
+*Override with pod.yaml*
```
rothgar, line 46: I would say "Override with workload manifest" because that can be set in a deployment, statefulset, pod, job, etc. This is in the document a few times and I would replace all of them.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+and GPU instances. View the full list with `aws ec2 describe-instance-types`.
+
+**Example**
+
+*Set Default with provisioner.yaml*
+
+```yaml
+spec:
+  instanceTypes:
+    - m5.large
```
ellistarn, line 43: This field is actually fully vendor neutral, so it should probably exist in vendor neutral docs. It just happens that for AWS, the values accepted are aws specific.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+template](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-launch-templates.html),
+such as a customized version of Bottlerocket or a different platform
+altogether.
+
+**Template ID**
+- key: `node.k8s.aws/launch-template-id`
+- value example: `bottlerocket` (default)
+- value list: `aws ec2 describe-launch-templates`
+
+**Template Version**
```
ellistarn, line 72: We're actually removing this field for release and instead will always choose $DEFAULT.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+### Capacity Type (e.g., spot)
+
+- key: `node.k8s.aws/capacity-type`
+- values
+  - `on-demand` (default)
+  - `spot`
+
+Karpenter supports specifying capacity type and defaults to on-demand.
+
+Specify this value on the provisioner to enable spot instance pricing. [Spot
```
ellistarn, line 87: Spot instances are more than just pricing.

website/content/en/docs/cloud-providers/AWS.md
ellistarn: We don't support a region label.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+## Pod Labels
+
+### Accelerators, GPU 
+
+Accelerator (e.g., GPU) values include
+- `nvidia.com/gpu`
+- `amd.com/gpu`
+- `aws.amazon.com/neuron`
+
+Karpenter supports GPUs. To specify a specific GPU type, use the instance type
```
ellistarn, line 178: Accelerators are a superset of gpus, so I'd just use that term.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,193 @@
...
+- value list: [AWS regional endpoints](https://docs.aws.amazon.com/general/latest/gr/rande.html)
+
+Karpenter can be configured to create nodes in a particular region.
+
+### Availability Zones
+
+`topology.kubernetes.io/zone=us-east-1c`
+
+- key: `topology.kubernetes.io/zone`
+- value example: `us-east-1c`
```
bwagner5, line 164: We also support Zone IDs on this field in addition to zone names, since zone names are not necessarily aligned across accounts. https://docs.aws.amazon.com/ram/latest/userguide/working-with-az-ids.html

website/content/en/docs/cloud-providers/AWS.md
geoffcline: The kubernetes docs use `=`, and I merely assumed they had a good reason

https://kubernetes.io/docs/reference/labels-annotations-taints/

website/content/en/docs/cloud-providers/AWS.md
geoffcline: oh i didn't know you could actually resolve that between accounts, good to know 👍

website/content/en/docs/cloud-providers/AWS.md
geoffcline: I would like to keep this here for now. The info in this section is largely AWS specific.

geoffcline:
thanks for the reviews all! I pushed an update and responded to some comments.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+nodes accordingly. For example, in response to a label of
+`topology.kubernetes.io/zone=us-east-1c`, Karpenter will provision nodes in
+that availability zone.
+
+Regarding AWS, 3 types of provisioning constraints are recognized: 
+- [Instance Type](#instance-type-allowlist)
+- [Node Labels](#node-lables-in-provisioner-spec) (e.g., Architecture, Capacity Type)
+- [Pod Labels](#pod-labels) (e.g., GPU)
+
+## Instance Type Allowlist
```
ellistarn, line 22: I think allowlist is the wrong frame. Karpenter is built on a concepts of defaults (provisioner) + overrides (pod). A pod can specify a node selector for an instance type that is not in the provisioner's defaults.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+
+The [Provisioner CRD]({{< ref "provisioner-crd.md" >}}) supports defining
+instance types, node labels, and node taints.
+
+For certain well-known labels (documented below), Karpenter will provision
+nodes accordingly. For example, in response to a label of
+`topology.kubernetes.io/zone=us-east-1c`, Karpenter will provision nodes in
+that availability zone.
+
+Regarding AWS, 3 types of provisioning constraints are recognized:
```
ellistarn, line 17: "Regarding AWS" seems like an awkward way to start a sentence to me. Also -- specifying the number in the sentence will easily become stale as we add more things to this list.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+The [Provisioner CRD]({{< ref "provisioner-crd.md" >}}) supports defining
+instance types, node labels, and node taints.
+
+For certain well-known labels (documented below), Karpenter will provision
+nodes accordingly. For example, in response to a label of
+`topology.kubernetes.io/zone=us-east-1c`, Karpenter will provision nodes in
+that availability zone.
+
+Regarding AWS, 3 types of provisioning constraints are recognized: 
+- [Instance Type](#instance-type-allowlist)
```
ellistarn, line 18: What about Architecture and Operating System and Zone?

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+- [Node Labels](#node-lables-in-provisioner-spec) (e.g., Architecture, Capacity Type)
+- [Pod Labels](#pod-labels) (e.g., GPU)
```
ellistarn, line 20: I think this pod label's piece isn't quite accurate. Specifically, these are pod.spec.nodeSelector.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Instance Type Allowlist
+
+Karpenter supports specifying [AWS instance type](https://aws.amazon.com/ec2/instance-types/). 
+
+If one instance type is listed, Karpenter will always provision that type.
```
ellistarn, line 26: This isn't true -- see overrides above.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+
+If more than one type is listed, Karpenter will intelligently select the
```
ellistarn, line 28: I'd avoid use of the word "intelligently", as it implies something that it's not. It's just logic.

ellistarn, line 28: We should recommend that customers leave instance types unconstrained, unless their use case requires it.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+instance type to maximize node utilization.
```
ellistarn, line 29: It will binpack pending pods into the smallest number of nodes allowed by scheduling constraints. This is to minimize node overhead.

website/content/en/docs/cloud-providers/AWS.md
ellistarn: Any label starting with `node.k8s.aws` is AWS specific. IIUC, we're using this one https://kubernetes.io/docs/reference/labels-annotations-taints/#nodekubernetesioinstance-type

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+      nodeSelector:
+        node.k8s.aws/instance-type: m5.large
+```
+
+## Node Lables in Provisioner Spec
+
+### Launch Template
+
+Karpenter uses [AWS Bottlerocket OS](https://aws.amazon.com/bottlerocket/) by
+default. More specifically, Karpenter automatically creates a launch template
```
ellistarn, line 61: I'd avoid language like "More specifically". Just state what it is.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+*Override with workload manifest (e.g., pod)*
+
+```yaml
+spec:
+  template:
+    spec:
+      nodeSelector:
+        node.k8s.aws/instance-type: m5.large
```
ellistarn, line 53: https://github.com/awslabs/karpenter/blob/main/pkg/apis/provisioning/v1alpha3/provisioner.go#L130

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+## Node Lables in Provisioner Spec
+
+### Launch Template
+
+Karpenter uses [AWS Bottlerocket OS](https://aws.amazon.com/bottlerocket/) by
+default. More specifically, Karpenter automatically creates a launch template
+with the name `Karpenter-<cluster name>-uuid` for each region where a node is
+provisioned.
+
+You can specify a different [launch
```
ellistarn, line 65: We should state that customers should rely on Karpenter's default launch templates unless their use case requires something else.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+spec:
+  template:
+    spec:
+      nodeSelector:
+        kubernetes.io/arch: amd64
+```
+
+## Operating System
+
+- key: `kubernetes.io/os`
```
ellistarn, line 139: This is a first class field in the provisioner (since it's an upstream k8s concept), and a nodeSelector at the pod level (since everything is a nodeSelector for pods).

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+### Accelerators, GPU 
+
+Accelerator (e.g., GPU) values include
+- `nvidia.com/gpu`
+- `amd.com/gpu`
+- `aws.amazon.com/neuron`
+
+Karpenter supports accelerators, such as GPUs. 
+
+To specify a specific GPU type, use the [instance type
```
ellistarn, line 176: Can you clarify that this should be done in addition to `resources`. Otherwise the kubelet won't allocate the GPU to the pod.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+`topology.kubernetes.io/zone=us-east-1c`
+
+- key: `topology.kubernetes.io/zone`
+- value example: `us-east-1c` or `use1-az1`
+- value list: `aws ec2 describe-availability-zones --region <region-name>`
+
+Karpenter can be configured to create nodes in a particular zone. Karpenter
+supports (1) availability zone IDs, and (2) availability zone names. 
+
+Availability zone IDs, such as `use1-az1`, are consitent between AWS accounts.
```
ellistarn, line 156: I'm not actually sure that this works.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -0,0 +1,191 @@
...
+  - `linux` (default)
+
+At this time, Karpenter only supports Linux OS nodes.
+
+### Availability Zones
+
+`topology.kubernetes.io/zone=us-east-1c`
+
+- key: `topology.kubernetes.io/zone`
+- value example: `us-east-1c` or `use1-az1`
```
ellistarn, line 150: I'm not sure that az ids work. I think there was a technical blocker.

geoffcline: according to @bwagner5 they do?

website/content/en/docs/cloud-providers/AWS.md
geoffcline: I'm confused - if a pod specifies an instance type not in karpenter's instance type list, karpenter will still provision it?

website/content/en/docs/cloud-providers/AWS.md
geoffcline: Instance type is specified as it's own list in the crd yaml, so I kept it separate. 

"Node labels" are things that go under "labels:" in the provisioner crd.

I also separated the pod GPU stuff because it only makes sense to include a podspec and not the provisioner crd.

I'm very open to alternate ways to structure some page. I did struggle with trying to make it more than a random list.

website/content/en/docs/cloud-providers/AWS.md
geoffcline: what is a good more general name? pod requirements?

website/content/en/docs/cloud-providers/AWS.md
geoffcline: what is the difference between minimizing node overhead and maximizing node utilization?

website/content/en/docs/cloud-providers/AWS.md
geoffcline: I'm leaning towards keeping it.

What does it mean to use bottlerocket by default? It means we create a launch template for you with that OS.

website/content/en/docs/cloud-providers/AWS.md
geoffcline: I'm hesitant to add this. I think the reader would understand when a different launch template is needed.

Is there an unexpected consequence of using your own launch template?

website/content/en/docs/cloud-providers/AWS.md
geoffcline: I'm reluctant to fragment the docs further at this point. For this release (where we only support AWS), I would like to have a single page where users can ctrl-f for the different tags/values.

website/content/en/docs/cloud-providers/AWS.md
bwagner5: They definitely did at one point, unless something regressed. I can check tomorrow, but they need to be supported, so if it has regressed I’ll open an issue to fix.

website/content/en/docs/cloud-providers/AWS.md
geoffcline: I clarified this, but the new content needs further review. I misunderstood this previously.

website/content/en/docs/cloud-providers/AWS.md
geoffcline: I changed it to "...karpenter will determine the instance type..."

geoffcline:
Note to self: revert desc of AZ IDs

geoffcline:
pushed a revision removing description of AZ IDs

website/content/en/docs/cloud-providers/AWS.md
ellistarn: Correct. The provisioner controls a set of defaults, but if the pod has a specific opinion, it will override the instance type and we'll use that.

website/content/en/docs/cloud-providers/AWS.md
ellistarn: Here's how I think about it:

Vendor neutral constraints (provisioner top level fields, pod.spec.nodeSelector)
- instance type
- zone
- architecture
- operating system

Vendor specific constraints (provisioner.spec.labels, pod.spec.nodeSelector)
- [aws] capacity type
- [aws] launch template
- [aws] security groups
- [aws] subnets

website/content/en/docs/cloud-providers/AWS.md
ellistarn: It's a classical NP-Hard computer science problem to minimize node overhead. Minimizing the number of nodes (which is what we do currently) minimizes the node overhead, which tends to do better w.r.t overall utilization. 

e.g.
Given 2 pods, 1 core + 35 cores = 36 cores total. It's NP-Hard to calculate that this should be two instances (c5.9xlarge + c5.large), and instead our logic will provision a c5.12xlarge, which is 8 additional "wasted" cores. However, pods that get created in the future will be able to schedule here, so in the limit of the cluster, it's likely this capacity will get used, but the node-level overhead (daemonsets, etc) will be reduced.

website/content/en/docs/cloud-providers/AWS.md
geoffcline: This has been reverted.

website/content/en/docs/cloud-providers/AWS.md
geoffcline: This definitely needs to be documented.

website/content/en/docs/cloud-providers/AWS.md
geoffcline: this will be resolved after we refactor the categories

akestner (approved):
Worked on this together

PR #563, opened 2021-07-28
Update getting started eksctl.yaml

rothgar:
Switching from self managed node group to managed node group. This could help customers in the future if they keep a cluster around a while and want to update the EKS version.

Also specified the EKS version because depending on which version of `eksctl` the user has it may deploy a different version. I would use 1.21 but the official `eksctl` with 1.21 support isn't out yet. 1.20 has been covered for a while and gets us a semi-new version of a cluster to avoid unknown issues.

**Issue, if available:**

**Description of changes:**

### DIFF
--- a/website/content/en/docs/getting-started/eksctl.yaml
+++ b/website/content/en/docs/getting-started/eksctl.yaml
@@ -4,7 +4,8 @@ kind: ClusterConfig
 metadata:
   name: ${CLUSTER_NAME}
   region: ${AWS_DEFAULT_REGION}
-nodeGroups:
+  version: "1.20"
+managedNodeGroups:
   - instanceType: m5.large
     name: ${CLUSTER_NAME}-ng
     desiredCapacity: 1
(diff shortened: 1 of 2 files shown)


website/content/en/docs/getting-started/eksctl.yaml
```diff
@@ -4,7 +4,8 @@ kind: ClusterConfig
 metadata:
   name: ${CLUSTER_NAME}
   region: ${AWS_DEFAULT_REGION}
-nodeGroups:
+  version: "1.20"
```
ellistarn, line 7: I was going to set this version, but I realized that it will cause our doc to go stale quickly. It's a bit tricky -- perhaps we could add a note/step upgrade eksctl on mac/linux to ensure the latest version?

ellistarn: Unless you think we should just maintain this?

rothgar: The docs will need to be versioned at some point so we're going to have to maintain versions either way. Some companies will have really old versions of EKS and we'll want to keep karpenter docs for them and maybe in the future we need to stop karpenter users from using the latest EKS (for whatever reason).

ellistarn (approved):
Nice!

rothgar:
If you want to merge this in with .3 release we'll have to update the helm chart release too.

ellistarn:
W/ current docs, we will need to hold off on updating until the release is cut.

Issue #569, opened 2021-07-28
Karpenter webhook fails to start on EKS 1.18


prateekgogia:
## Version

**Karpenter**: v0.2.9

**Kubernetes**: v1.18.16-eks-7737de

## Expected Behavior

Karpenter fresh install to work on 1.18 version

## Actual Behavior

Webhook pod keeps crashing 
```
➜  ~ k get pods -A -w
NAMESPACE     NAME                                    READY   STATUS             RESTARTS   AGE
karpenter     karpenter-controller-769d64bdd5-dbh4v   1/1     Running            0          5m19s
karpenter     karpenter-webhook-764fdf4f5-sdsl6       0/1     CrashLoopBackOff   5          5m19s
```

With the following log message -
```
k logs -f karpenter-webhook-764fdf4f5-sdsl6 -n karpenter
2021/07/28 22:14:13 Registering 1 clients
2021/07/28 22:14:13 Registering 2 informer factories
2021/07/28 22:14:13 Registering 3 informers
2021/07/28 22:14:13 Registering 3 controllers
{"severity":"INFO","timestamp":"2021-07-28T22:14:13.968745002Z","logger":"fallback-logger","caller":"logging/config.go:79","message":"Fetch GitHub commit ID from kodata failed","error":"open /var/run/ko/HEAD: no such file or directory"}
{"severity":"WARNING","timestamp":"2021-07-28T22:14:13.968790559Z","logger":"fallback-logger","caller":"logging/config.go:72","message":"Failed to parse logging config - using default zap production config"}
{"severity":"INFO","timestamp":"2021-07-28T22:14:13.968815035Z","logger":"fallback-logger.karpenter.webhooks","caller":"profiling/server.go:64","message":"Profiling enabled: false"}
[...]

PR #570, opened 2021-07-28
Set Kubernetes minimum version for karpenter webhook

prateekgogia:
**Issue, if available:**
#569 

**Description of changes:**

### DIFF
--- a/charts/karpenter/templates/webhook/rbac.yaml
+++ b/charts/karpenter/templates/webhook/rbac.yaml
@@ -68,4 +68,10 @@ rules:
   - watch
   - list
   - update
+- apiGroups:
+  - ""
+  resources:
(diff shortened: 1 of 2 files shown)


charts/karpenter/templates/webhook/rbac.yaml
```diff
@@ -68,4 +68,10 @@ rules:
   - watch
   - list
   - update
+- apiGroups:
```
ellistarn, line 71: This role is already applied to the service account via the controller's RBAC. Are you at HEAD? That said, it may not be a bad idea to duplicate here.

ellistarn:
I think this has already been resolved at HEAD, and we're requiring v1.19 now.

PR #572, opened 2021-07-29
AWS Cloud Provider now uses accurate pod/memory values when constructing node objects

ellistarn:
**Issue, if available:**

**Description of changes:**
```
v1.ResourcePods:   *instanceType.Pods(),
v1.ResourceCPU:    *instanceType.CPU(),
v1.ResourceMemory: *instanceType.Memory(),
```

1. Resolves an issue where the scheduler would pod additional pods onto nodes before they come online since the capacity values were way too high.
2. Deleted the NodeAPI abstraction and folded it into the InstanceProvider 


pkg/cloudprovider/aws/instance.go
```diff
@@ -35,19 +40,69 @@ const (
...
+	id, err := p.launchInstance(ctx, launchTemplate, instanceTypes, subnets, capacityType)
+	if err != nil {
+		return nil, err
+	}
+	// 2. Get Instance with backoff retry since EC2 is eventually consistent
+	instance := &ec2.Instance{}
+	if err := retry.Do(
+		func() (err error) { return p.getInstance(ctx, id, instance) },
+		retry.Delay(1*time.Second),
+		retry.Attempts(3),
```
njtran, line 67: Do you have a reason why you picked 3 attempts of 1 second each? It seems pretty likely for EC2 to throttle for 3 seconds if you're making a lot of requests. 

Can we make this a larger interval with more retries? If an instance creation fails, we'll have to do binpacking all over again anyways and attempt to create the instance.

pkg/cloudprovider/aws/instance.go
```diff
@@ -107,22 +162,54 @@ func (p *InstanceProvider) Create(ctx context.Context,
 	return createFleetOutput.Instances[0].InstanceIds[0], nil
 }
 
-func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
-	id, err := getInstanceID(node)
+func (p *InstanceProvider) getInstance(ctx context.Context, id *string, instance *ec2.Instance) error {
+	describeInstancesOutput, err := p.ec2api.DescribeInstancesWithContext(ctx, &ec2.DescribeInstancesInput{InstanceIds: []*string{id}})
+	if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "InvalidInstanceID.NotFound" {
```
njtran, line 167: I think this is the same constant as in Line 93. Can you change this accordingly?

pkg/controllers/allocation/bind.go
```diff
@@ -61,11 +61,11 @@ func (b *Binder) Bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) error
 	}
 
 	// 4. Bind pods
-	logging.FromContext(ctx).Infof("Binding %d pod(s) to node %s", len(pods), node.Name)
 	errs := make([]error, len(pods))
 	workqueue.ParallelizeUntil(ctx, len(pods), len(pods), func(index int) {
 		errs[index] = b.bind(ctx, node, pods[index])
 	})
+	logging.FromContext(ctx).Infof("Bound %d pod(s) to node %s", len(pods), node.Name)
```
njtran, line 68: Let's say that we had to bind 10 pods, and only 5 succeeded. This would log that we binded 10 pods, right? Can we change this to be more accurate?

njtran (commented):
Nice work! Just a couple of comments.

pkg/cloudprovider/aws/instance.go
ellistarn: I haven't seen any throttling failures here, even with some preliminary 1k node scale tests. I've never seen this one fail due to rate limiting, as the describe limits are actually quite high. I'm open to increasing these, but I'm a bit wary of the retries themselves causing cascading effects on other threads. Regardless, we will recover gracefully w/ the reallocator if these retries do fail.

pkg/cloudprovider/aws/instance.go
ellistarn: Nice catch!

pkg/cloudprovider/aws/instance.go
njtran: Cool. We should definitely keep track of this somewhere, especially when we get to larger cluster sizes.

pkg/controllers/allocation/bind.go
```diff
@@ -61,12 +61,13 @@ func (b *Binder) Bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) error
...
 
 	// 4. Bind pods
-	logging.FromContext(ctx).Infof("Binding %d pod(s) to node %s", len(pods), node.Name)
 	errs := make([]error, len(pods))
 	workqueue.ParallelizeUntil(ctx, len(pods), len(pods), func(index int) {
 		errs[index] = b.bind(ctx, node, pods[index])
 	})
-	return multierr.Combine(errs...)
+	err := multierr.Combine(errs...)
+	logging.FromContext(ctx).Infof("Bound %d/%d pod(s) to node %s", len(pods) - len(multierr.Errors(err)),  len(pods), node.Name)
```
njtran, line 69: I don't know if we need fractions here. I think it'd be sufficient to just say how many did bind.

njtran (approved):

PR #594, opened 2021-08-02
Refactored expiration/utilization controller into node controller

ellistarn:
**Issue, if available:**

**Description of changes:**


pkg/controllers/node/controller.go
```diff
@@ -45,50 +51,84 @@ func NewController(kubeClient client.Client) *Controller {
...
+	if !stored.DeletionTimestamp.IsZero() {
+		return reconcile.Result{}, nil
+	}
 
-	// 2. Execute node reconcilers
+	// 2. Retrieve Provisioner
+	provisioner := &v1alpha3.Provisioner{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: stored.Labels[v1alpha3.ProvisionerNameLabelKey]}, provisioner); err != nil {
+		return reconcile.Result{}, err
+	}
```
njtran, line 80: Do we want to do the same error handling logic here as with the node? If it does exist but failed another way, do we want to return `result.RetryIfError(ctx, err)`?

pkg/controllers/node/controller.go
```diff
@@ -45,50 +51,84 @@ func NewController(kubeClient client.Client) *Controller {
...
-	// 3. Patch any changes, regardless of errors
+	// 4. Patch any changes, regardless of errors
 	if !reflect.DeepEqual(node, stored) {
 		if err := c.kubeClient.Patch(ctx, node, client.MergeFrom(stored)); err != nil {
 			return reconcile.Result{}, fmt.Errorf("patching node %s, %w", node.Name, err)
 		}
 	}
+
+	// 5. Requeue if error or if retryAfter is set
+	if requeueAfter := functional.MaxDuration(backoffs...); requeueAfter != 0 {
```
njtran, line 106: I don't think you need this check here. From the docs:
```
// RequeueAfter if greater than 0, tells the Controller to requeue the reconcile key after the Duration.
// Implies that Requeue is true, there is no need to set Requeue to true at the same time as RequeueAfter.
```

pkg/controllers/node/controller.go
```diff
@@ -45,50 +51,84 @@ func NewController(kubeClient client.Client) *Controller {
...
+					logging.FromContext(ctx).Errorf("Failed to list nodes when mapping expiration watch events, %s", err.Error())
+					return requests
+				}
+				for _, node := range nodes.Items {
+					requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Name: node.Name}})
+				}
+				return requests
+			}),
+		).
 		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
```
njtran, line 132: I realize this was included in a prior PR, but why did we pick 10 here? Are there any detriments to bumping this number up?

pkg/controllers/node/expiration.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha3"
+	"github.com/awslabs/karpenter/pkg/utils/ptr"
+	v1 "k8s.io/api/core/v1"
+	"knative.dev/pkg/logging"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// Expiration is a tiny reconciler that ensures nodes have the termination
```
njtran, line 30: Could you replace tiny with something like "child/leaf/sub-reconciler"?

pkg/controllers/node/expiration.go
```diff
@@ -0,0 +1,58 @@
...
+// Expiration. This protects against instances that launch when Karpenter fails
+// to create the node object. In this case, the node will come online without
+// the termination Expiration. This controller will update the node accordingly.
+type Expiration struct {
+	kubeClient client.Client
+}
+
+// Reconcile adds the termination Expiration if the node is not deleting
```
njtran, line 38: This comment doesn't make much sense. Could you say "Reconcile deletes the node if it is past its expiration date"?

pkg/controllers/node/expiration.go
```diff
@@ -0,0 +1,58 @@
...
+	expirationTTL := time.Duration(ptr.Int64Value(provisioner.Spec.TTLSecondsUntilExpired)) * time.Second
+	expirationTime := node.CreationTimestamp.Add(expirationTTL)
+	if now.After(expirationTime) {
+		logging.FromContext(ctx).Infof("Triggering termination for expired node %s after %s (+%s)", node.Name, expirationTTL, time.Since(expirationTime))
+		if err := r.kubeClient.Delete(ctx, node); err != nil {
+			return reconcile.Result{}, fmt.Errorf("deleting node, %w", err)
+		}
+	}
+
+	// 3. Backoff until expired, if already expired, negative values do not requeue
```
njtran, line 56: I don't think you need to include the details about negative values here. You could say positive values do requeue, since a value of 0 doesn't requeue.

njtran (commented):
nice work. some comments on your comments

pkg/controllers/node/suite_test.go
```diff
@@ -53,57 +55,117 @@ var _ = AfterSuite(func() {
...
+					v1alpha3.ProvisionerNameLabelKey: provisioner.Name,
+				},
+			})
+			ExpectCreated(env.Client, provisioner, node)
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(node))
+
+			node = ExpectNodeExists(env.Client, node.Name)
+			Expect(node.DeletionTimestamp.IsZero()).To(BeTrue())
+		})
+		It("should terminate nodes after expiry", func() {
```
JacobGabrielson, line 106: Do we have the ability to simulate time going by in these tests? I worry a bit that there might be some code paths not quite exercised here ... compared to having a TTL set (to a non-zero value like 30), prove that (say) 15 seconds in the node is not marked for termination, then (say) 30+ seconds in it actually is so marked?

pkg/controllers/node/controller.go
ellistarn: We could consider setting this to kubeclient qps (default 20). I found that 10 was reasonably high parallelism without exhausting default qps. Perhaps we can evaluate this number as we do more scale testing.

pkg/controllers/node/controller.go
ellistarn: +1

pkg/controllers/node/suite_test.go
ellistarn: These tests are just copied over since it's a refactor, but I'll look into mocking time for these tests.

ellistarn: done

pkg/controllers/node/emptiness.go
```diff
@@ -0,0 +1,91 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// Emptiness is a subreconciler that terminates nodes that are empty after a ttl
+type Emptiness struct {
+	kubeClient client.Client
+}
+
+// Reconcile reconciles the node
+func (r *Emptiness) Reconcile(ctx context.Context, provisioner *v1alpha3.Provisioner, n *v1.Node) (reconcile.Result, error) {
```
njtran, line 39: Can we change this from r *Emptiness to e *Emptiness to fall in line with how we've been doing it?

pkg/controllers/node/controller.go
```diff
@@ -45,50 +53,97 @@ func NewController(kubeClient client.Client) *Controller {
...
+	var backoffs []time.Duration
 	var errs error
 	for _, reconciler := range []interface {
-		Reconcile(*v1.Node) error
+		Reconcile(context.Context, *v1alpha3.Provisioner, *v1.Node) (reconcile.Result, error)
 	}{
 		c.readiness,
+		c.liveness,
+		c.expiration,
+		c.emptiness,
```
njtran, line 96: If a node doesn't have a finalizer but will be expired for emptiness/expiration, we could choose to delete this node before we add the finalizer since the expiration and emptiness controllers send a delete request before we patch the node.

In addition, since we're running these in parallel, is it possible that the order in which we execute these controllers change?

pkg/controllers/node/emptiness.go
```diff
@@ -0,0 +1,91 @@
...
+			logging.FromContext(ctx).Infof("Removed emptiness TTL from node %s", n.Name)
+		}
+		return reconcile.Result{}, nil
+	}
+	// 3. Set TTL if not set
+	n.Annotations = functional.UnionStringMaps(n.Annotations)
+	if _, ok := n.Annotations[v1alpha3.ProvisionerTTLAfterEmptyKey]; !ok {
+		ttl := time.Duration(ptr.Int64Value(provisioner.Spec.TTLSecondsAfterEmpty)) * time.Second
+		n.Annotations[v1alpha3.ProvisionerTTLAfterEmptyKey] = time.Now().Add(ttl).Format(time.RFC3339)
+		logging.FromContext(ctx).Infof("Added TTL (+%s) to empty node %s", ttl, n.Name)
```
njtran, line 64: I remember discussing the potential for us to change the TTL to a "TTL timestamp" indicating where it was added. We would then grab the TTL value from the provisioner to see if it's been that long.  This would help us with cases where the provisioner TTL value changes, and change up this log line. WDYT?

pkg/controllers/node/emptiness.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+// Emptiness is a subreconciler that terminates nodes that are empty after a ttl
+type Emptiness struct {
+	kubeClient client.Client
+}
+
+// Reconcile reconciles the node
+func (r *Emptiness) Reconcile(ctx context.Context, provisioner *v1alpha3.Provisioner, n *v1.Node) (reconcile.Result, error) {
+	// 1. Ignore node if not applicable
```
njtran, line 40: Can you change this line to the following to match the expiration controller?
```
	// 1. Ignore if TTLSecondsAfterEmpty isn't defined 
```

pkg/controllers/node/expiration.go
```diff
@@ -0,0 +1,53 @@
...
+	// 1. Ignore if TTLSecondsUntilExpired isn't defined
+	if provisioner.Spec.TTLSecondsUntilExpired == nil {
+		return reconcile.Result{}, nil
+	}
+	// 2. Trigger termination workflow if expired
+	now := time.Now()
+	expirationTTL := time.Duration(ptr.Int64Value(provisioner.Spec.TTLSecondsUntilExpired)) * time.Second
+	expirationTime := node.CreationTimestamp.Add(expirationTTL)
+	if now.After(expirationTime) {
+		logging.FromContext(ctx).Infof("Triggering termination for expired node %s after %s (+%s)", node.Name, expirationTTL, time.Since(expirationTime))
```
njtran, line 46: doesn't this need to be %s, %d, %d?

pkg/controllers/node/liveness.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/awslabs/karpenter/pkg/utils/node"
+	v1 "k8s.io/api/core/v1"
+	"knative.dev/pkg/logging"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const LivenessTimeout = 5 * time.Minute
+
+// Liveness is a subreconciler that deletes nodes if its determined to be unrecoverable
```
njtran, line 32: I think this comment might be a little misleading. From reading this, I would assume that a node that has joined, but then has become in an unrecoverable state would then be subject to the workings of this controller.

pkg/utils/functional/functional.go
```diff
@@ -109,3 +112,26 @@ func InvertStringMap(stringMap map[string]string) map[string]string {
...
+	var max time.Duration
+	for _, duration := range durations {
+		if duration > max {
+			max = duration
+		}
+	}
+	return max
+}
+
+func JsonEquals(a, b interface{}) bool {
```
njtran, line 127: Do we need this anymore?

pkg/controllers/node/emptiness.go
ellistarn: I named these all r since they're all reconcilers.

pkg/controllers/node/controller.go
ellistarn: Semantically, I want the order of these to not matter. I know that in theory, there are some edge cases (i.e, node comes online, instantly ready, we can't reconcile, and emptyttl is 0), but I think they're contrived enough that I'd much rather define an invariant where the order of these reconcilers don't matter.

pkg/controllers/node/emptiness.go
ellistarn: I'd love to run this by @akestner, as I think it's a product decision. I'm not certain which is cleanest conceptually.

pkg/controllers/node/emptiness.go
ellistarn: I'd rather do the reverse and use "not applicable" since it's less likely to get stale.

pkg/controllers/node/expiration.go
ellistarn: I want to print the time String() not int64.

pkg/utils/functional/functional.go
ellistarn: Nice catch

pkg/controllers/node/controller.go
```diff
@@ -45,50 +53,97 @@ func NewController(kubeClient client.Client) *Controller {
...
-		errs = multierr.Append(errs, reconciler.Reconcile(node))
+		res, err := reconciler.Reconcile(ctx, provisioner, node)
+		errs = multierr.Append(errs, err)
+		backoffs = append(backoffs, res.RequeueAfter)
 	}
 
-	// 3. Patch any changes, regardless of errors
-	if !reflect.DeepEqual(node, stored) {
+	// 4. Patch any changes, regardless of errors
+	if !equality.Semantic.DeepEqual(node, stored) {
```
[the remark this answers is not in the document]
ellistarn: Interestingly, this doesn't exist anywhere else ;)

pkg/controllers/node/emptiness.go
ellistarn: Tentatively, I think this is the right call.

pkg/controllers/node/liveness.go
ellistarn: I think in theory, that's what this controller should do. I think we do want to implement something like "auto repair" in here.

pkg/controllers/node/emptiness.go
njtran: That sounds good to me. I'd like to keep the comments roughly the same irrespective of the content.

pkg/controllers/node/expiration.go
njtran: Can you use the `.String()` function then to make sure it formats properly?

pkg/controllers/node/controller.go
njtran: I agree that these cases probably won't happen naturally. I also think our logging is in a good enough state that we would be able to debug these cases if they do happen naturally. From then, I'd love to reconsider this approach if necessary.

pkg/controllers/node/emptiness.go
```diff
@@ -0,0 +1,96 @@
...
+		if _, ok := n.Annotations[v1alpha3.EmptinessTimestampAnnotationKey]; ok {
+			delete(n.Annotations, v1alpha3.EmptinessTimestampAnnotationKey)
+			logging.FromContext(ctx).Infof("Removed emptiness TTL from node %s", n.Name)
+		}
+		return reconcile.Result{}, nil
+	}
+	// 3. Set TTL if not set
+	n.Annotations = functional.UnionStringMaps(n.Annotations)
+	ttl := time.Duration(ptr.Int64Value(provisioner.Spec.TTLSecondsAfterEmpty)) * time.Second
+	emptinessTimestamp, ok := n.Annotations[v1alpha3.EmptinessTimestampAnnotationKey]
```
njtran, line 62: WDYT?

```suggestion
	emptinessTimestamp, ok := n.Annotations[v1alpha3.EmptinessTimestampAnnotationKey]
	if !empty {
		if ok {
			delete(n.Annotations, v1alpha3.EmptinessTimestampAnnotationKey)
			logging.FromContext(ctx).Infof("Removed emptiness TTL from node %s", n.Name)
		}
		return reconcile.Result{}, nil
	}
	// 3. Set TTL if not set
	n.Annotations = functional.UnionStringMaps(n.Annotations)
	ttl := time.Duration(ptr.Int64Value(provisioner.Spec.TTLSecondsAfterEmpty)) * time.Second
```

pkg/controllers/node/expiration.go
ellistarn: This is what the %s directive does.

pkg/controllers/node/liveness.go
ellistarn: >That sounds good to me. I'd like to keep the comments roughly the same irrespective of the content.

njtran (approved):
LGTM! Super nice work. Excited to see this work.

PR #604, opened 2021-08-10
[fix] back slash for iam service account creation

marco-aws:
Quick fix to avoid line break after `eksctl create iamserviceaccount`

Tested on MacOS/zsh.


ellistarn:
I think I might've just beaten you to this change yesterday.

ellistarn:
Done! https://github.com/awslabs/karpenter/releases/tag/v0.3.1

PR #611, opened 2021-08-12
Reenabled time mocking for node tests

ellistarn:
**Issue, if available:**

**Description of changes:**

### DIFF
--- a/pkg/controllers/node/liveness.go
+++ b/pkg/controllers/node/liveness.go
@@ -36,7 +36,10 @@ type Liveness struct {
 
 // Reconcile reconciles the node
 func (r *Liveness) Reconcile(ctx context.Context, provisioner *v1alpha3.Provisioner, n *v1.Node) (reconcile.Result, error) {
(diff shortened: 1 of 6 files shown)


pkg/controllers/node/liveness.go
```diff
@@ -36,7 +36,10 @@ type Liveness struct {
 
 // Reconcile reconciles the node
 func (r *Liveness) Reconcile(ctx context.Context, provisioner *v1alpha3.Provisioner, n *v1.Node) (reconcile.Result, error) {
-	if !node.FailedToJoin(n, LivenessTimeout) {
+	if Now().Sub(n.GetCreationTimestamp().Time) < LivenessTimeout {
```
njtran, line 39: This seems a little convoluted compared to the `.Since()` function in the time library. It’s essentially the same functionality, but I think Since() looks nicer.

ellistarn: I'd need to mock since as well. Tough tradeoff.

njtran (approved):
LGTM

Issue #613, opened 2021-08-12
Spot as Default Capacity Type 


rr-paras-patel:
We should start with default capacity type as spot instances. 
I recently encountered issue due to some reason karpenter was keep launching nodes and node were in "not ready" state.
and previously launched nodes were not being terminated by karpenter result into 500 or More On Demand node was launched in few hours and it was costly mistake for someone who is trying to contribute in open source project. 
I am happy to raise PR for this change. 
Current Hidden documentation about this default is here : https://karpenter.sh/docs/provisioner-crd/

ellistarn:
Hey @rr-paras-patel this is a known bug right now. We thought that the [liveness controller](https://github.com/awslabs/karpenter/blob/main/pkg/controllers/node/liveness.go#L38) should take care of it, but it looks like there's a bug for some kubernetes version. The behavior *should* be to terminate nodes when they cannot connect for 5 minutes. In this case, Karpenter will attempt to provision nodes in response to pending pods, and the pods have a toleration for 5 minutes of a node not being ready, so you'll see a new node every 5 minutes or so. I'm going to look into a fix for this. 

I think it's a great idea to put `node.k8s.aws/capacity-type: "spot"` into the getting started guide. If you're suggesting that we should launch spot by default (without labels), I think this is a longer discussion. Feel free to reach out on the Kubernetes slack if you want to chat synchronously. I can also help debug this NotReady issue.

Issue #601, opened 2021-08-10
Karpenter Provisioned EC2 Node Fails to Join Cluster


akestner:
## Version

**Karpenter**: v0.3.0

**Kubernetes**: v1.20

## Expected Behavior
New nodes spun up by Karpenter will join the cluster successfully and start receiving workloads.

## Actual Behavior
Nodes (EC2 instances) provisioned by Karpenter never leave `NotReady` status.

## Steps to Reproduce the Problem
1. Create a cluster using the following `eksctl` cluster config:
```yaml
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig
metadata:
  name: cluster-name
  region: us-west-2
  version: "1.20"
managedNodeGroups:
  - instanceType: m5.xlarge
    name: cluster-name-nodegroup
    desiredCapacity: 2
    minSize: 2
    maxSize: 10
    ssh:
      enableSsm: true
secretsEncryption:
  keyARN: <ARN>
iam:
  withOIDC: true
  serviceAccounts:
    - metadata:
        name: karpenter
        namespace: karpenter
      attachPolicy:
        Version: "2012-10-17"
        Statement:
          - Effect: Allow
            Resource: "*"
            Action:
              # Write Operations
              - "ec2:CreateLaunchTemplate"
              - "ec2:CreateFleet"
              - "ec2:RunInstances"
              - "ec2:CreateTags"
              - "iam:PassRole"
              - "ec2:TerminateInstances"
              # Read Operations
              - "ec2:DescribeLaunchTemplates"
              - "ec2:DescribeInstances"
              - "ec2:DescribeSecurityGroups"
              - "ec2:DescribeSubnets"
              - "ec2:DescribeInstanceTypes"
              - "ec2:DescribeInstanceTypeOfferings"
              - "ec2:DescribeAvailabilityZones"
              - "ssm:GetParameter"
addons:
- name: vpc-cni
  version: 1.9.0
```
2. Follow Getting Started Guide from ["Tag Subnets"](https://karpenter.sh/docs/getting-started/#tag-subnets).
3. Scale the `inflate` deployment up enough to require new nodes.
4. Observe that some `inflate` pods are not scheduled and that the new nodes don't join the cluster.

## Additional Info
Here's a dump of `kubectl get nodes <node-name> -o yaml` from one of the nodes that failed to join:
```yaml
apiVersion: v1
kind: Node
metadata:
  annotations:
    node.alpha.kubernetes.io/ttl: "0"
[...]

PR #614, opened 2021-08-13
Fixed a bug with liveness logic

ellistarn:
**Issue, if available:**
#613, #601

**Description of changes:**
Fixes an issue where nodes don't terminate if they can't connect to the API Server, resulting in runaway scaling. 

The liveness controller will terminate nodes that don't ready up. This is critical, because pods will only survive on not ready nodes for 5 minutes by default (tolerations). The controller must differentiate between a node that never connected (cleanup) vs a node that connected but lost connection (e.g. network partition, don't cleanup). 

The previous code assumed that the timestamp would be nil. There is a note on this here: https://github.com/kubernetes/kubernetes/blob/release-1.17/pkg/controller/nodelifecycle/node_lifecycle_controller.go#L1012. The nodelifecycle controller writes status here: https://github.com/kubernetes/kubernetes/blob/release-1.17/pkg/controller/nodelifecycle/node_lifecycle_controller.go#L1131. 

This change should catch both the case where the node controller never writes anything, as well as the case where it writes `NodeStatusNeverUpdated` which is the expected behavior.

We will continue to only take any action after the liveness timeout (5 mins). 

### DIFF
--- a/pkg/controllers/node/suite_test.go
+++ b/pkg/controllers/node/suite_test.go
@@ -215,7 +215,6 @@ var _ = Describe("Controller", func() {
 
 			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(provisioner))
 
-			// Expect node not be deleted
 			n = ExpectNodeExists(env.Client, n.Name)
 			Expect(n.DeletionTimestamp.IsZero()).To(BeTrue())
 
--- a/pkg/controllers/node/suite_test.go
+++ b/pkg/controllers/node/suite_test.go
@@ -187,21 +187,12 @@ var _ = Describe("Controller", func() {
 				ReadyStatus: v1.ConditionUnknown,
 				ReadyReason: "NodeStatusNeverUpdated",
 			})
-			pod := test.Pod(test.PodOptions{NodeName: n.Name})
-			ExpectCreated(env.Client, provisioner, pod)
+			ExpectCreated(env.Client, provisioner)
 			ExpectCreatedWithStatus(env.Client, n)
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/node/suite_test.go
```diff
@@ -180,7 +180,39 @@ var _ = Describe("Controller", func() {
...
+				ReadyStatus: v1.ConditionUnknown,
+				ReadyReason: "NodeStatusNeverUpdated",
+			})
+			pod := test.Pod(test.PodOptions{NodeName: n.Name})
+			ExpectCreated(env.Client, provisioner, pod)
+			ExpectCreatedWithStatus(env.Client, n)
+
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(provisioner))
+
+			// Expect n not deleted
```
njtran, line 196: Can you make n -> node here

pkg/controllers/node/suite_test.go
```diff
@@ -180,7 +180,39 @@ var _ = Describe("Controller", func() {
...
+			n = ExpectNodeExists(env.Client, n.Name)
+			Expect(n.DeletionTimestamp.IsZero()).To(BeTrue())
+
+			// Delete pod and do another reconcile
+			Expect(env.Client.Delete(ctx, pod)).To(Succeed())
```
njtran, line 201: We shouldn’t need to make and delete a pod for this test to work, it’s sufficient to have it be ready unknown, which won’t trigger underutilization.

ellistarn: FWIW, I ported these tests from your liveness logic tests, which at some point I think relied on pods existing.

JacobGabrielson (approved):

PR #619, opened 2021-08-15
Implemented topology spread constraints for zone and hostname

ellistarn:
**Issue, if available:**
https://github.com/awslabs/karpenter/issues/481

**Description of changes:**


pkg/apis/provisioning/v1alpha3/provisioner_validation_test.go
```diff
@@ -87,13 +87,7 @@ var _ = Describe("Validation", func() {
...
 		})
 		It("should fail for restricted labels", func() {
-			for _, label := range []string{
-				ArchitectureLabelKey,
-				OperatingSystemLabelKey,
-				ProvisionerNameLabelKey,
-				ZoneLabelKey,
-				InstanceTypeLabelKey,
-			} {
+			for _, label := range RestrictedLabels {
```
JacobGabrielson, line 90: I realize it already was this way, so this is just a nit, but looking at this unit test, it looks almost exactly like `validateRestrictedLabels()`, which seems like a bad sign? (I.e., is it really testing anything?). I wonder if it would make more sense to make this a bit beefier, like every time through add in a bunch of valid labels and then also add on the bad one, or something? And also make sure it works if there are two bad labels or something?

bwagner5 (approved):
Had a small comment on the leakiness of the label injection to make the topology scheduling simpler, but overall, feel free to merge, looks great!

Issue #628, opened 2021-08-20
Support extra taints not considered when scheduling Pods


rustrial:
**Tell us about your request**

As of today, if taints are defined in [`spec.Constraints.taints`](https://karpenter.sh/docs/reference/#karpenter.sh/v1alpha3.Constraints), the provisioner will not provision nodes for pods that do not have matching tolerations.

> Taints will be applied to every node launched by the Provisioner. If specified, the provisioner will not provision nodes for pods that do not have matching tolerations.

However, some cluster add-ons (usually CNI plugins) use temporary taints on new nodes to make sure no Pods are scheduled onto that nodes unless the CNI `DaemonSet` is ready on that node. Usually, those *temporary node startup taints* are then removed by the `DaemonSet` or by a corresponding controller. A prominent example is [Cilium](https://docs.cilium.io/en/stable/gettingstarted/k8s-install-default/) which uses the `node.cilium.io/agent-not-ready` taint. 

Adding toleration to all Pods for such *temporary node startup taints* would of course void their purpose and might lead to disfunctional Pods as they are scheduled and started before CNI was properly setup on that node. With its current capabilities karpenter is not powerful enough to support such a setup.

**What do you want us to build?**

Add something like `spec.Constraints.extraTaints`, which contains taints that must be applied to newly created nodes, but might later be removed by workloads and are ignored by the provisioner for scheduling purposes. Specifically, karpenter should support the following behaviour:

- provisioner adds all `extraTaints` to newly created nodes and then forgets about them (does not re-add them if they are later removed by a CNI controller).
- provisioner ignores `extraTaints` when making scheduling decisions

Of course, the proposed solution is just a (viable) placeholder and maybe somebody has a better ideas to resolve this problem, so feel free to propose other solutions.

**Are you currently working around this issue?**

Not using karpenter at all :-(  

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ellistarn:
[...]

PR #631, opened 2021-08-23
[WIP] Add support for node readiness taints

alex-berger:
Implement #628 adding support for node readiness taints.

**Issue, if available:**

- #628

**Description of changes:**

This PR adds support for node readiness taints as described in #628.


pkg/controllers/allocation/bind.go
```diff
@@ -59,10 +65,41 @@ func (b *Binder) Bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) error
 		if !errors.IsAlreadyExists(err) {
 			return fmt.Errorf("creating node %s, %w", node.Name, err)
 		}
+		// If the node object already exists, make sure finalizer and taint are in place.
+		if err := b.KubeClient.Patch(ctx, node, client.StrategicMergeFrom(stored)); err != nil {
```
ellistarn, line 69: We have a controller (node/finalizer.go) that enforces this already. Further, if we move forward with this change, we should make sure that ```v1.Taint{
		Key:    v1alpha3.NotReadyTaintKey,
		Effect: v1.TaintEffectNoSchedule,
	}``` is applied as part of the user data.

pkg/controllers/allocation/bind.go
```diff
@@ -59,10 +65,41 @@ func (b *Binder) Bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) error
...
+			return fmt.Errorf("patching node %s, %w", node.Name, err)
+		}
 	}
-
-	// 4. Bind pods
 	errs := make([]error, len(pods))
+	// 4. Wait for node readiness.
```
ellistarn, line 74: We can't wait here. This will block the entire provisioning process on nodes coming online, which will massively reduce provisioning throughput. If we move forward with this path, we will need some async mechanism for this.

ellistarn (commented):
I'll admit I'm not a huge fan of this approach, but I don't immediately have a better alternative.
 
1. We need to be careful with any sort of delayed/async binding behavior, as it may cause kube-scheduler race conditions.
2. I don't like adding a new API concept for folks to think about.

I want to figure out a way for this to "just work", but it may take some more time to think through. Perhaps we can discuss this at the Karpenter Working group and go through the use cases?

pkg/controllers/allocation/bind.go
rustrial: OK, I removed the `Patch` call and `v1alpha3.NotReadyTaintKey` is now appended to the `ReadinessTaints` to make sure it gets added in the user data (launch template).

pkg/controllers/allocation/bind.go
rustrial: I pushed a commit which uses asynchronous bind, but only if it is required.

ellistarn:
Closing as stale. Happy to reopen this discussion in the future.

BryanStenson-okta:
any more thoughts/motivation on this approach?  i'm stuck on the original problem while trying to implement Cilium in my cluster via taints.

ellistarn:
Which version of Cilium? I was using Cilium and Karpenter in my cluster up until a few weeks ago.

BryanStenson-okta:
1.11.3 -- without using the node taint method, I end up with pods which are do not have matching `CiliumEndpoints` (which basically means I cannot ensure any of the NetworkPolicy goodness is applied to these nodes).

ellistarn:
Makes sense, thanks for clarifying!

PR #636, opened 2021-08-24
Use controllerruntime to discover CABundle

JacobGabrielson:
**Issue, if available:**

N/A

**Description of changes:**

Apologies for the churn folks, after talking to @ellistarn offline, it seems like the best path forward, given some other changes coming soon, is to explicitly call GetCABundle() from the cloudprovider, rather than dynamically default it in.

Also, this change now uses `client-go` to discover the CABundle contents (effectively it's still doing the same thing
as before, just using their library code now).


pkg/cloudprovider/aws/suite_test.go
```diff
@@ -27,12 +27,15 @@ import (
...
 	"github.com/awslabs/karpenter/pkg/test"
 	. "github.com/awslabs/karpenter/pkg/test/expectations"
+	"github.com/awslabs/karpenter/pkg/utils/filesys"
 	"github.com/awslabs/karpenter/pkg/utils/parallel"
 	"github.com/awslabs/karpenter/pkg/utils/resources"
 	"github.com/patrickmn/go-cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"io/fs"
```
bwagner5, line 37: nit: this should be below the non-std lib imports

Not sure why there are spaces separating these non-std lib imports, they can be pushed together.

bwagner5:
looks like the toolchain install is broken because this link does not work to download kubebuilder 2.3.1 anymore:

https://go.kubebuilder.io/dl/2.3.1/linux/amd64

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -163,5 +163,8 @@ func (c *CloudProvider) ValidateSpec(ctx context.Context, spec *v1alpha3.Provisi
 	if spec.Cluster.Name == nil || len(*spec.Cluster.Name) == 0 {
 		errs = errs.Also(apis.ErrMissingField("name")).ViaField("cluster")
 	}
+	if spec.Cluster.CABundle == nil || len(*spec.Cluster.CABundle) == 0 {
```
ellistarn, line 166: check out `ptr.StringValue(spec.Cluster.CaBundle) == ""`, which handles both of these cases.

Does this mean this can no longer be nil?

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -51,8 +52,27 @@ var launchTemplateCache *cache.Cache
 var fakeEC2API *fake.EC2API
 var controller reconcile.Reconciler
 
+type singletonFS struct {
```
ellistarn, line 55: consider making a fake.Filesystem like the fake.EC2API

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -51,8 +52,27 @@ var launchTemplateCache *cache.Cache
...
+func (s *singletonFS) ReadFile(name string) ([]byte, error) {
+	if name == s.filename {
+		return s.contents, nil
+	}
+	return nil, fs.ErrNotExist
+}
+
 func TestAPIs(t *testing.T) {
-	ctx = TestContextWithLogger(t)
+	ctx = filesys.Inject(TestContextWithLogger(t), &singletonFS{
```
ellistarn, line 72: Thoughts on just injecting this for the test where you need it?

pkg/utils/filesys/filesys.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+	"io/fs"
+
+	"github.com/spf13/afero"
+)
+
+type contextKey string
+
+const (
+	KarpenterFS contextKey = "karpenter.sh/fs"
```
ellistarn, line 25: consider making this a private variable. I also like to avoid referencing Karpenter anywhere in the project, since names can always change. Consider making this its own package inside utils like `filesystem`
```
filesystem.For(ctx)
filesystem.Inject(ctx)
```

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -98,18 +118,19 @@ var _ = Describe("Allocation", func() {
...
 			},
 			Spec: v1alpha3.ProvisionerSpec{
 				Cluster: v1alpha3.Cluster{
 					Name:     ptr.String("test-cluster"),
 					Endpoint: "https://test-cluster",
-					CABundle: ptr.String("dGVzdC1jbHVzdGVyCg=="),
 				},
 			},
-		}
+		}).WithDynamicDefaults(ctx)
```
ellistarn, line 132: thoughts on just isolating this to the validation?

pkg/cloudprovider/aws/cloudprovider.go
JacobGabrielson: Ok!

> Does this mean this can no longer be nil?

For AWS, it has to either be specified or else in `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`.

pkg/cloudprovider/aws/suite_test.go
JacobGabrielson: Why would that be better? In practice all this code under test is going to happen after defaulting has occurred. But I could be missing something.

pkg/cloudprovider/aws/suite_test.go
JacobGabrielson: See my other response.

pkg/utils/filesys/filesys.go
JacobGabrielson: Ok on the first point.

> Consider making this its own package inside utils like filesystem [...]

I'm confused here - this already is a package under `utils` and usage reads `filesys.For(ctx)`?

pkg/cloudprovider/aws/cloudprovider.go
ellistarn: Doesn't this validation prevent the case where I set the CABundle to nil (i.e., so that it falls back to the ca.cert file).

pkg/cloudprovider/aws/suite_test.go
ellistarn: You're right I was confused. 

In general -- I don't like this dynamic defaulting behavior, since it results in an extra file read per provisioner per reconciliation and adds a nontrivial amount of complexity. Perhaps not a huge deal. 

I wonder if we could get away with statically defaulting it into the provisioner spec (in the Default() webhook), since AFAICT clusters don't typically rotate their root CA bundle. Even if one did, you'd simply need to apply the provisioner again and it would regenerate the new bundle.

pkg/utils/filesys/filesys.go
ellistarn: Oh wild -- I was misreading. You're totally correct XD. Any reason you chose filesys over filesystem for naming?

ellistarn (dismissed):
Feel free to merge. Comments are optional and nonblocking.

pkg/cloudprovider/aws/cloudprovider.go
JacobGabrielson: I added a test for that - see line 591 in `suite_test.go`. AFAICT it's better for most of the code to run w/ defaulting and then explicitly turn it off as needed.

pkg/cloudprovider/aws/suite_test.go
JacobGabrielson: I suspect the OS level performance impact would be minimal, presumably that small of a file would stick around in memory if it were being accessed frequently and I think the overhead of a system call won't be too much given the other stuff we're already doing. 

I'm a little leery of the putting it into the static defaulter. While I agree that it seems mostly better/cleaner, I'm worrying about the case you're outlining where one would have to reapply the Provisioner (that seems slightly yucky), even if we don't know of a case where that happens today (I assume people will want to rotate those creds at some point, and it might be kind of a bummer to diagnose Karpenter issues as a result of a seemingly unrelated action). Also it just seems like the spirit of:

https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#directly-accessing-the-rest-api

is to honor these files under `/var` even if it is a bit fiddly?

pkg/utils/filesys/filesys.go
JacobGabrielson: No real reason, `filesystem` sounds better to me.

pkg/cloudprovider/aws/cloudprovider.go
JacobGabrielson: FYI, latest code has the `ptr.StringValue` usage

pkg/cloudprovider/aws/suite_test.go
JacobGabrielson: I was thinking of holding off until we see if there are more uses, lmk if that seems like a bad idea.

JacobGabrielson: Actually, I take that back - will do.

pkg/cloudprovider/aws/suite_test.go
ellistarn: Ack

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -160,8 +161,11 @@ func (c *CloudProvider) ValidateConstraints(ctx context.Context, constraints *v1
 
 // Validate cloud provider specific components of the cluster spec.
 func (c *CloudProvider) ValidateSpec(ctx context.Context, spec *v1alpha3.ProvisionerSpec) (errs *apis.FieldError) {
-	if spec.Cluster.Name == nil || len(*spec.Cluster.Name) == 0 {
+	if ptr.StringValue(spec.Cluster.Name) == "" {
 		errs = errs.Also(apis.ErrMissingField("name")).ViaField("cluster")
 	}
+	if ptr.StringValue(spec.Cluster.CABundle) == "" {
```
rustrial, line 167: FYI: I deliberately skipped this check in my initial version of CABundle handling, to support the use-case where there is no CABundle (e.g. if the k8s API endpoint is served in plain-text without TLS). I know this might not be possible with EKS (which is always TLS), but it can still happen with non EKS clusters hosted on EC2.

Personally, I don't care to much, as I only use TLS protected k8s API endpoints and plain-text endpoints are a bad practice anyway. However, I still prefer to have secure defaults while still allowing others to opt-out (for whatever reasons they might have).

JacobGabrielson: @rustrial thanks! That makes sense to me, I'll make the change and see what everyone thinks.

pkg/cloudprovider/aws/suite_test.go
JacobGabrielson: Done

pkg/utils/filesys/filesys.go
JacobGabrielson: Done

pkg/cloudprovider/aws/fake/filesystem.go
```diff
@@ -0,0 +1,20 @@
+package fake
```
ellistarn, line 1: Thoughts on moving this to pkg/test/fake since it's not aws specific?

JacobGabrielson: Will do

pkg/apis/provisioning/v1alpha3/provisioner_validation.go
```diff
@@ -103,6 +106,23 @@ func (c *Cluster) validate() (errs *apis.FieldError) {
...
+// explicitly configured with a non-TLS scheme, like 'http'.
+func (c *Cluster) HasTLSEndpoint() bool {
+	endpoint, err := c.parseEndpoint()
+	return err != nil || endpoint.Scheme != "http"
+}
+
+// validate performs cloud-provider-agnostic validations (only).
+func (c *Cluster) validate() (errs *apis.FieldError) {
+	if c == nil {
+		return errs.Also(apis.ErrMissingField())
```
ellistarn, line 121: Should we specify the field here?

ellistarn: I see we have `.ViaField("cluster"),` above. Does this print cleanly as is?

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -53,12 +54,13 @@ var fakeEC2API *fake.EC2API
...
 
 func TestAPIs(t *testing.T) {
-	ctx = TestContextWithLogger(t)
+	ctx = filesystem.Inject(TestContextWithLogger(t), testfake.Filesystem())
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "CloudProvider/AWS")
 }
 
 var _ = BeforeSuite(func() {
+
```
ellistarn, line 63: remove

ellistarn (approved):

pkg/apis/provisioning/v1alpha3/provisioner_validation.go
JacobGabrielson: > Should we specify the field here?

That seems already to work:

```
error: error validating "STDIN": error validating data: ValidationError(Provisioner.spec): missing required field "cluster" in sh.karpenter.v1alpha3.Provisioner.spec; if you choose to ignore these errors, turn validation off with --validate=false
```

Regarding the ViaField question, all that code is going to change, per our offline conversation.

pkg/cloudprovider/aws/cloudprovider.go
JacobGabrielson: Ok, for now I've changed the code back to the way it was before - it won't complain if CABundle is missing. I'll add the http/https nuance back in in a future change.

bwagner5 (approved):
lgtm

ellistarn (approved):
Curious why we move away from gotemplates for userdata. I find it a bit more difficult to parse.

JacobGabrielson:
> Curious why we move away from gotemplates for userdata. I find it a bit more difficult to parse.

It started getting complicated trying to call a function from the templates, so it seemed easier overall to switch to the more clunky old school method.

PR #640, opened 2021-08-27
Pod Isolation Support: Implemented taint generation for pods with tolerations

ellistarn:
**Issue, if available:**
https://github.com/awslabs/karpenter/issues/634

**Description of changes:**
Note: These changes are built on top of topology code which must be merged first

1. Generate matching taints on nodes for pods that contain tolerations
2. Only OpEquals is supported
3. If OpExists, pod still schedules, but taints aren't generated
4. If Effect isn't specified, generates taints for both NoExecute and NoSchedule
5. Refactored constraints logic to be separated from v1alpha3


pkg/controllers/allocation/scheduling/constraints.go
```diff
@@ -0,0 +1,101 @@
+package scheduling
```
JacobGabrielson, line 1: missing license banner?

JacobGabrielson (approved):

PR #652, opened 2021-08-31
improve bin-packing performance

bwagner5:
**Issue, if available:**
https://github.com/awslabs/karpenter/issues/625

**Description of changes:**
 - Increased performance of bin-packing. This specific optimization found a loop where a terminable case was not being checked which allowed a significant number of loop iterations to be skipped. 
   - The loop was responsible for packing pods on to a specific packable node. Once the node was full, the loop would continue unnecessarily. This change also adds another small optimization to the check to determine if the node is full by passing through the smallest pod from the list (it's already sorted when passed thru) and adds it to the currently reserved resources in order to see if anymore progress could possibly be made by continuing through the loop. 
   
- The terminable case optimization in the packing loop had a 30x improvement
- The smaller "min-pod" optimization resulted in a 50% improvement

**Benchmarks:**
   - 20,000 pods previously took 91 seconds and now takes about 3 seconds and drastically reduces memory usage. NOTE: the benchmark graph "post-optimization" was ran before the minPod optimization which reduced the 20,000 pod case from around 6 seconds to 3 seconds. The benchmarks were executed on 2 m5.large instance types. 

graphs

**Testing:**
 - Performed A/B testing w/ 1:1 input and both gave the same output.  The specific test walked up cpu, memory, and pods to pack in increments of 100 milli-cores, 64mb, and 1 respectively.  The upper limits were 10,000 mc, 8GB, and 90 pods respectively. 
 - A/B tests on multiple node packings were also performed ad-hoc.

### DIFF
--- a/pkg/controllers/allocation/binpacking/packable.go
+++ b/pkg/controllers/allocation/binpacking/packable.go
@@ -87,7 +87,7 @@ func PackableFor(i cloudprovider.InstanceType) *Packable {
 
 // Pack attempts to pack the pods into capacity, keeping track of previously
 // packed pods. If the capacity cannot fit the pod, they are set aside.
-// pods should be sorted in descending order.
+// Pods must be sorted in descending order.
 func (p *Packable) Pack(pods []*v1.Pod) *Result {
 	result := &Result{}
 	for i, pod := range pods {
--- a/pkg/controllers/allocation/binpacking/packable.go
+++ b/pkg/controllers/allocation/binpacking/packable.go
@@ -87,9 +87,10 @@ func PackableFor(i cloudprovider.InstanceType) *Packable {
 
 // Pack attempts to pack the pods into capacity, keeping track of previously
 // packed pods. If the capacity cannot fit the pod, they are set aside.
+// pods should be sorted in descending order.
 func (p *Packable) Pack(pods []*v1.Pod) *Result {
 	result := &Result{}
-	for _, pod := range pods {
+	for i, pod := range pods {
 		if ok := p.reservePod(pod); ok {
 			result.packed = append(result.packed, pod)
 			continue
@@ -100,8 +101,27 @@ func (p *Packable) Pack(pods []*v1.Pod) *Result {
 			return result
 		}
 		result.unpacked = append(result.unpacked, pod)
+
+		if p.isFull(pods[len(pods)-1]) {
+			result.unpacked = append(result.unpacked, pods[i+1:]...)
+			return result
+		}
 	}
 	return result
+}
+
+// isFull checks if the reserved resources + the minimum resourced pod overflows the total resources available.
+func (p *Packable) isFull(minPod *v1.Pod) bool {
+	minResourceList := resources.RequestsForPods(minPod)
+	for resourceName, totalQuantity := range p.total {
+		reservedQuantity := p.reserved[resourceName].DeepCopy()
+		// Add the minimum resourced pod to reserved
+		reservedQuantity.Add(minResourceList[resourceName])
+		if !totalQuantity.IsZero() && reservedQuantity.Cmp(totalQuantity) >= 0 {
+			return true
+		}
+	}
+	return false
 }
 
 func (p *Packable) reserve(requests v1.ResourceList) bool {
--- a/pkg/controllers/allocation/binpacking/packable.go
+++ b/pkg/controllers/allocation/binpacking/packable.go
@@ -95,7 +95,7 @@ func (p *Packable) Pack(pods []*v1.Pod) *Result {
 			result.packed = append(result.packed, pod)
 			continue
 		}
-		if p.isFull(pods[len(pods)-1]) {
+		if p.fits(pods[len(pods)-1]) {
 			result.unpacked = append(result.unpacked, pods[i:]...)
 			return result
 		}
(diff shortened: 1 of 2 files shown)


pkg/packing/packable.go
```diff
@@ -87,9 +87,10 @@ func PackableFor(i cloudprovider.InstanceType) *Packable {
 
 // Pack attempts to pack the pods into capacity, keeping track of previously
 // packed pods. If the capacity cannot fit the pod, they are set aside.
+// pods should be sorted in descending order.
```
ellistarn, line 90: ascending? Assuming you mean smallest -> largest.

pkg/packing/packer.go
```diff
@@ -142,8 +142,9 @@ func sortByResources(instanceTypes []cloudprovider.InstanceType) {
 func weightOf(instanceType cloudprovider.InstanceType) float64 {
 	return euclidean(
 		float64(instanceType.CPU().Value()),
-		float64(instanceType.Memory().ScaledValue(resource.Mega)), // 1 gb = 1 cpu
+		float64(instanceType.Memory().ScaledValue(resource.Giga)), // 1 gb = 1 cpu
```
ellistarn, line 145: Wow nice find!

pkg/packing/packable.go
```diff
@@ -100,10 +101,29 @@ func (p *Packable) Pack(pods []*v1.Pod) *Result {
...
+		if p.isFull(pods[len(pods)-1]) {
+			result.unpacked = append(result.unpacked, pods[i+1:]...)
+			return result
+		}
 	}
 	return result
 }
 
+// isFull checks if the reserved resources + the minimum resourced pod overflows the total resources available.
+func (p *Packable) isFull(minPod *v1.Pod) bool {
```
ellistarn, line 114: Any reason you can't use 
```
func (p *Packable) reservePod(pod *v1.Pod) bool {
	requests := resources.RequestsForPods(pod)
	requests[v1.ResourcePods] = *resource.NewQuantity(1, resource.BinarySI)
	return p.reserve(requests)
}
func (p *Packable) reserve(requests v1.ResourceList) bool {
	candidate := resources.Merge(p.reserved, requests)
	// If any candidate resource exceeds total, fail to reserve
	for resourceName, quantity := range candidate {
		if quantity.Cmp(p.total[resourceName]) > 0 {
			return false
		}
	}
	p.reserved = candidate
	return true
}
```

As far as I understand it, this function does exactly what you need.

pkg/packing/packable.go
```diff
@@ -100,10 +101,29 @@ func (p *Packable) Pack(pods []*v1.Pod) *Result {
 			return result
 		}
 		result.unpacked = append(result.unpacked, pod)
+
+		if p.isFull(pods[len(pods)-1]) {
```
ellistarn, line 105: Is it possible to move this check to the beginning of the loop? Unless I'm misunderstanding.

pkg/packing/packable.go
JacobGabrielson, line 90: `should` -> `must` (RFC style)?

pkg/packing/packable.go
bwagner5: Yeah, you're probably right that this can be reduced. Let me try to merge these. I think the dry-run would be the isFull and the reserve would still exist but do less and just call isFull and the reserve.

JacobGabrielson:
Nice!

pkg/packing/packable.go
bwagner5: actually, after digging into trying to merge these two or use each other, I think it's actually making the interface of reserve significantly worse off. The reserve function is doing something slightly different and returning a different result. 

The reserve function will take a resource list and reserve on the node if it doesn't overflow the node's resource. The return value signifies that the reservation succeeded (did not overflow) or failed (overflowed).

The isFull function is only concerned about the node totals and seeing they are all the way full, but not overflowed OR would overflow with the smallest pod that needs to be reserved in the list. 

I think merging these functions makes it less readable, adds complexity to reserve, and the return value gets muddied since we'd need two bools returns, one indicating if the reservation succeeded and one noting if the node is full.

pkg/packing/packable.go
bwagner5: it can be moved to the top. I think it might make more sense to move after the reserve block though. 

Here's how I reason in english about the func:

Try to reserve the space for the pod, if that fails, then check if the node is full and potentially exit the loop. If the node isn't full (next block) and nothing has been packed (meaning the largest pod was unable to be packed), exit the loop because we need to find a bigger node. If we've made progress, but the current node is just getting full, add the items to the unpacked list and keep trying smaller pods.

pkg/controllers/allocation/binpacking/packable.go
```diff
@@ -104,6 +109,20 @@ func (p *Packable) Pack(pods []*v1.Pod) *Result {
 	return result
 }
 
+// isFull checks if the reserved resources + the minimum resourced pod overflows the total resources available.
+func (p *Packable) isFull(minPod *v1.Pod) bool {
```
ellistarn, line 113: optional: s/minPod/pod : you could imagine that we'd check pods that we're min if we decided to not sort for some reason.
optional: s/isFull/fits : isFull seems a bit odd since the node is only full in relation to the pod. The node isn't necessarily full, since another pod might fit if it had no resource reqs.

pkg/controllers/allocation/binpacking/packable.go
```diff
@@ -104,6 +109,20 @@ func (p *Packable) Pack(pods []*v1.Pod) *Result {
...
+	minResourceList := resources.RequestsForPods(minPod)
+	for resourceName, totalQuantity := range p.total {
+		reservedQuantity := p.reserved[resourceName].DeepCopy()
+		// Add the minimum resourced pod to reserved
```
ellistarn, line 117: nit: remove this comment, as the code is self explanatory.

ellistarn (dismissed):

ellistarn:
Amazing!

ellistarn (approved):

PR #654, opened 2021-09-02
update development guide

cjerad:
**Issue, if available:**

**Description of problem(s):**
1. `make codegen` fails, even after running `make toolchain`, because of missing executables
2. `make apply` fails because cluster namespace, "karpenter", does not exist

**Description of changes:**
1. Add instruction to update `PATH` with Go workspace's `bin` after running `make toolchain`
2. Add instruction to create namespace, "karpenter"

### DIFF
--- a/hack/toolchain.sh
+++ b/hack/toolchain.sh
@@ -8,18 +8,12 @@ main() {
 }
 
 tools() {
-    # Before installing tools check that the environment has been configured.
-    if [[ -z "$GOPATH" ]]; then
-        echo "error: GOPATH is not set."
-        exit 1
-    fi
-
     cd tools
     go mod tidy
     GO111MODULE=on cat tools.go | grep _ | awk -F'"' '{print $2}' | xargs -tI % go install %
 
-    if ! echo "$PATH" | grep -q "$GOPATH/bin"; then
-        echo "Go workspace's \"bin\" directory is not in PATH. Run 'export PATH=\"\$PATH:\$GOPATH/bin\"'."
+    if ! echo "$PATH" | grep -q "${GOPATH:-undefined}/bin\|$HOME/go/bin"; then
+        echo "Go workspace's \"bin\" directory is not in PATH. Run 'export PATH=\"\$PATH:\${GOPATH:-\$HOME/go}/bin\"'."
     fi
 }
 
--- a/hack/toolchain.sh
+++ b/hack/toolchain.sh
@@ -8,9 +8,19 @@ main() {
 }
 
 tools() {
+    # Before installing tools check that the environment has been configured.
+    if [[ -z "$GOPATH" ]]; then
+        echo "error: GOPATH is not set."
+        exit 1
+    fi
+
     cd tools
     go mod tidy
     GO111MODULE=on cat tools.go | grep _ | awk -F'"' '{print $2}' | xargs -tI % go install %
+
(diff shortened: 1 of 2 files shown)


website/content/en/docs/development-guide.md
```diff
@@ -13,7 +13,7 @@ The following tools are required for contributing to the Karpenter project.
 | [go](https://golang.org/dl/)                                       | v1.15.3+ | [Instructions](https://golang.org/doc/install)   |
 | [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) |          | `brew install kubectl` |
 | [helm](https://helm.sh/docs/intro/install/)                        |          | `brew install helm`    |
-| Other tools                                                        |          | `make toolchain`       |
+| Other tools                                                        |          | `make toolchain` then ensure `PATH` contains Go workspace's `bin` |
```
ellistarn, line 16: Instead of putting this here, what do you think about adding some output in the toolchain.sh script to echo something like: `PATH=$PATH:$GOPATH/bin`

website/content/en/docs/development-guide.md
```diff
@@ -39,6 +39,7 @@ make delete # Uninstall Karpenter
 ### Build and Deploy
 ```
 make dev                                  # build and test code
+kubectl create namespace karpenter        # create target namespace for deployment
```
ellistarn, line 42: Nice catch. We typically have this namespace created as part of the getting-started guide (helm install --create-namespace).

website/content/en/docs/development-guide.md
cjerad: 🆗

cjerad:
Apparently GOPATH is not set in the CI environment.  Is the preferred solution to add GOPATH to the CI environment (and update the check in toolchain.sh to avoid the "unbound variable" error) or just remove the check in toolchain.sh?

bwagner5:
> Apparently GOPATH is not set in the CI environment. Is the preferred solution to add GOPATH to the CI environment (and update the check in toolchain.sh to avoid the "unbound variable" error) or just remove the check in toolchain.sh?

GOPATH is technically optional in go now since modules have taken over most of its use. We could change the check in toolchain.sh to check if `$HOME/go/bin` is in the path or if GOPATH is set and in the PATH. I'd guess the GHA has $HOME/go/bin in the PATH as part of the go install (famous last words).

hack/toolchain.sh
```diff
@@ -11,6 +11,10 @@ tools() {
     cd tools
     go mod tidy
     GO111MODULE=on cat tools.go | grep _ | awk -F'"' '{print $2}' | xargs -tI % go install %
+
+    if ! echo "$PATH" | grep -q "${GOPATH:-undefined}/bin\|$HOME/go/bin"; then
```
ellistarn, line 15: My path is actually at $HOME/workspaces/go. Thoughts on removing this check entirely and just printing a friendly message that isn't exact, but can at least provide users a hint.

ellistarn (dismissed):

hack/toolchain.sh
bwagner5: I'm guessing `$HOME/workspaces/go` is what your `$GOPATH` is set to?  But if that weren't set for you, then `go` would default to `$HOME/go`

bwagner5 (dismissed):
🎉

bwagner5 (approved):
lgtm

PR #655, opened 2021-09-02
Auto-discover caBundle when missing (take 2)

JacobGabrielson:
**Issue, if available:**

This is a prerequisite for https://github.com/awslabs/karpenter/issues/626 but does not fix it.

**Description of changes:**

fix problems with https://github.com/awslabs/karpenter/pull/636/files

1. conflict between knative and controllerruntime was causing webhook to crash, and, 
2. CAData was empty so launch template created was invalid


pkg/apis/provisioning/v1alpha3/provisioner_defaults.go
```diff
@@ -17,65 +17,35 @@ package v1alpha3
...
-		if errors.Is(err, os.ErrNotExist) {
-			return nil, nil
-		}
+		logging.FromContext(ctx).Infof("Unable to discover caBundle, loading transport config, %v", err)
 		return nil, err
 	}
-	encoded := base64.StdEncoding.EncodeToString(binary)
-	return &encoded, nil
+	_, err = transport.TLSConfigFor(transportConfig) // fills in CAData!
+	logging.FromContext(ctx).Infof("Discovered caBundle, length %d", len(transportConfig.TLS.CAData))
```
ellistarn, line 49: This should minimally be a debug statement like the other discovery bits. How often does this log line print?

pkg/utils/restconfig/restconfig.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+
+	"k8s.io/client-go/rest"
+)
+
+type contextKey string
+
+const (
+	restKey contextKey = "restKey"
```
ellistarn, line 12: consider a struct and this style: https://github.com/knative/pkg/blob/b558677ab03404ed118ba3e179a7438fe2d8509a/apis/contexts.go#L28

ellistarn (commented):
Looks good! Thanks for driving this through

pkg/apis/provisioning/v1alpha3/provisioner_defaults.go
JacobGabrielson: Only when creating a launch template

pkg/utils/restconfig/restconfig.go
JacobGabrielson: Ok

pkg/apis/provisioning/v1alpha3/provisioner_defaults.go
JacobGabrielson: done

pkg/utils/restconfig/restconfig.go
JacobGabrielson: done

ellistarn (dismissed):

ellistarn (approved):

PR #664, opened 2021-09-08
Emptiness reconcile now properly backs off

njtran:
**1. Issue, if available:**
https://github.com/awslabs/karpenter/issues/663

- [x] No


pkg/utils/functional/functional.go
```diff
@@ -112,12 +112,12 @@ func InvertStringMap(stringMap map[string]string) map[string]string {
 }
 
 // MaxDuration returns the largest duration
```
bwagner5, line 114: update comment

pkg/utils/functional/functional.go
```diff
@@ -112,12 +112,12 @@ func InvertStringMap(stringMap map[string]string) map[string]string {
...
-func MaxDuration(durations ...time.Duration) time.Duration {
-	var max time.Duration
+func MinDuration(durations ...time.Duration) time.Duration {
+	var min time.Duration
 	for _, duration := range durations {
-		if duration > max {
-			max = duration
+		if (duration < min && duration != 0) || (duration != 0 && min == 0) {
```
bwagner5, line 118: if you initialize min to the first duration, you can simplify this check by removing all the 0s

JacobGabrielson, line 118: Just for safety, clarity, can you use https://pkg.go.dev/time#Time.IsZero ?

njtran: IIUC, if the first duration is 0, we still run into the problem.

njtran: Yes! Thank you.

njtran: Since we're comparing `time.Duration{}` and not `time.Time{}`, this kind of comparison should be okay. 

From their code 
`type Duration int64`

JacobGabrielson: Yeah, I realize it works either way, but I think IsZero is preferable from a readability standpoint.

njtran: So since there's not a `time.Duration.IsZero()` function, are you suggesting I add it?

bwagner5: oh you want to return the smallest non-zero, ok.  I'm guessing it's fine to return 0 if all the durations are 0 then?

bwagner5 (approved):
lgtm

ellistarn (approved):

PR #665, opened 2021-09-08
Changed Logging configmap to not be overridden in dev workflow

njtran:
**1. Issue, if available:**
Possibly related to this issue https://github.com/awslabs/karpenter/issues/649

When doing `make apply` to test local changes, we apply the configmap yaml, where the overrides will set the log level back to the info level regardless of if the developer wants it to be at the debug level.

- [x] No


charts/karpenter/templates/100-config-logging.yaml
```diff
@@ -31,6 +31,3 @@ data:
         "timeEncoder": "iso8601"
       }
     }
-  # Log level overrides
-  loglevel.controller: info # debug
```
JacobGabrielson, line 35: I think it would be good to just comment these two lines out (completely) - that way people can see how to change the log level if they wish.

ellistarn (approved):

PR #667, opened 2021-09-08
Termination Controller waits to terminate nodes until all evicting pods have terminated

njtran:
**1. Issue, if available:**
https://github.com/awslabs/karpenter/issues/633

- Only allowed drain function to succeed when all pods have finished terminating
- Added a test function to test if termination is being done at all stages

- [x] No


pkg/controllers/termination/terminate.go
```diff
@@ -66,6 +69,7 @@ func (t *Terminator) drain(ctx context.Context, node *v1.Node) (bool, error) {
 	// https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
 	nonCritical := []*v1.Pod{}
 	critical := []*v1.Pod{}
+	readyToTerminate := true
```
ellistarn, line 72: terminable?

pkg/controllers/termination/suite_test.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/awslabs/karpenter/pkg/cloudprovider/registry"
 	"github.com/awslabs/karpenter/pkg/controllers/termination"
 	"github.com/awslabs/karpenter/pkg/test"
+	f "github.com/awslabs/karpenter/pkg/utils/functional"
```
ellistarn, line 28: why not functional?

pkg/controllers/node/liveness.go
```diff
@@ -27,13 +27,13 @@ import (
...
 )
 
-const LivenessTimeout = 5 * time.Minute
-
 // Liveness is a subreconciler that deletes nodes if its determined to be unrecoverable
 type Liveness struct {
 	kubeClient client.Client
 }
 
+const LivenessTimeout = 5 * time.Minute
```
ellistarn, line 35: Why move this?

pkg/controllers/termination/suite_test.go
```diff
@@ -195,6 +193,37 @@ var _ = Describe("Termination", func() {
...
+			ExpectDeleted(env.Client, pods[1])
+
+			// Expect node to exist and be draining, but not deleted
+			node = ExpectNodeExists(env.Client, node.Name)
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(node))
+			ExpectNodeDraining(env.Client, node.Name)
+
+			ExpectDeleted(env.Client, pods[0])
+
+			// Expect node to exist and be draining, but not deleted
```
ellistarn, line 222: This comment is stale.

pkg/controllers/termination/terminate.go
```diff
@@ -77,7 +81,15 @@ func (t *Terminator) drain(ctx context.Context, node *v1.Node) (bool, error) {
 		}
 		// Don't attempt to evict a pod that's already evicting
 		if !pod.DeletionTimestamp.IsZero() {
-			continue
+			// If the kubelet cannot register a pod as deleted, we need to force delete the pod in order to terminate the node
+			if time.Since(pod.DeletionTimestamp.Time) > time.Duration(*pod.DeletionGracePeriodSeconds) {
```
ellistarn, line 85: To avoid NPEs, we should use something like `ptr.Int64Value(pod.DeletionGracePeriodSeconds)`

pkg/controllers/termination/terminate.go
```diff
@@ -77,7 +81,15 @@ func (t *Terminator) drain(ctx context.Context, node *v1.Node) (bool, error) {
...
+				if err := t.KubeClient.Delete(ctx, pod, &client.DeleteOptions{GracePeriodSeconds: kptr.Int64(0)}); err != nil {
```
ellistarn, line 86: Imagine the pods have some sort of finalizer on them. These delete calls can be asynchronous (even if we force). 

Instead, I think we should rely simply on len(pods) > 0. This will ensure that there are actually 0 pods when we delete the node. It has a side benefit of also deduping the boolean and continue statement.

pkg/controllers/termination/terminate.go
ellistarn, line 85: Duration defaults to nanoseconds, so this comparison is going to be super wrong IIUC.

If DeletionTimestamp is zero, this will never be greater than the deletiongraceperiodseconds (even if they're both zero). Can we collapse this into a single check like:

```
if time.Since(pod.DeletionTimestamp.Time) > time.Duration(ptr.Int64Value(pod.DeletionGracePeriodSeconds) {
   // kubectl delete
}
```

pkg/controllers/node/liveness.go
njtran: Bad merge, gotta fix this

pkg/controllers/termination/terminate.go
njtran: Good callout on the default for nanoseconds. 

This line: `if !pod.DeletionTimestamp.IsZero() {` refers to a pod evicting. It's possible for it to be nonZero (meaning it's evicting), but for it to not be past the podDeletionGracePeriodSeconds yet, so we shouldn't collapse this into one line.

pkg/controllers/termination/terminate.go
njtran: Since we wait for the gracePeriodSeconds here before actually force terminating, the pods will have been evicting + in the terminating state, and would have triggered finalizer workflows. Since we don't know how long finalizers will take, I think it's fair to say that after the gracePeriodSeconds we can attempt to force terminate. 

We need some force delete pod logic. If we don't, we will never have `len(pods) == 0` if the Kubelet is down and can't register the pod as notReady. Are you suggesting that these delete calls lie somewhere else?

pkg/controllers/termination/terminate.go
```diff
@@ -122,3 +115,32 @@ func (t *Terminator) getPods(ctx context.Context, node *v1.Node) ([]*v1.Pod, err
...
+		evictable = append(evictable, p)
+	}
+	return evictable
+}
+
+func (t *Terminator) evict(ctx context.Context, pods []*v1.Pod) {
+	// 1. Prioritize noncritical pods https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
+	prioritized := []*v1.Pod{}
+	for _, p := range pods {
+		if !p.DeletionTimestamp.IsZero() {
```
ellistarn, line 134: IIUC, these pods will be included on line 143. Filtering these before they even enter this context might save you. Maybe most easily done on line 131.

pkg/utils/pod/terminating.go
```diff
@@ -0,0 +1,15 @@
+package pod
+
+import (
+	"time"
+
+	"github.com/awslabs/karpenter/pkg/utils/ptr"
+	v1 "k8s.io/api/core/v1"
+)
+
+func IsStuckTerminating(pod *v1.Pod) bool {
```
ellistarn, line 10: I'd leave this scoped to the termination logic since I'd hope we wouldn't use this anywhere else.

ellistarn (approved):
Optional comment

ellistarn (approved):

PR #668, opened 2021-09-09
Amazon Linux 2 as default AMI

JacobGabrielson:
**1. Issue, if available:**

https://github.com/awslabs/karpenter/issues/626

Changing the default Launch Template to use alinux2, for now. Will re-add support for Bottlerocket as part of further enhancements to the Launch template. 

Tested by launching nodes that only had taints, only had labels, had both taints and labels, and also ones that had neither.

- [ ] No

### DIFF
--- a/pkg/cloudprovider/aws/ami.go
+++ b/pkg/cloudprovider/aws/ami.go
@@ -53,8 +53,7 @@ func (p *AMIProvider) Get(ctx context.Context, constraints *Constraints) (string
 	if *constraints.Architecture == v1alpha3.ArchitectureArm64 {
 		architectureSuffix = "-arm64"
 	}
-	// TODO: support for the "amazon-linux-2-gpu" AMI for nvidia use
-	// cases
+	// TODO: support for the "amazon-linux-2-gpu" AMI for nvidia use cases
 	name := fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, architectureSuffix)
 	if id, ok := p.cache.Get(name); ok {
 		return id.(string), nil
--- a/pkg/cloudprovider/aws/ami.go
+++ b/pkg/cloudprovider/aws/ami.go
@@ -15,6 +15,7 @@ limitations under the License.
 package aws
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"strings"
@@ -22,6 +23,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ssm"
 	"github.com/aws/aws-sdk-go/service/ssm/ssmiface"
+	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha3"
 	"github.com/patrickmn/go-cache"
 	"k8s.io/client-go/kubernetes"
 	"knative.dev/pkg/logging"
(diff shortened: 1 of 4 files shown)


geoffcline:
This is a situation where I'd welcome using the new "Yes, issue opened" option.

pkg/cloudprovider/aws/ami.go
```diff
@@ -48,7 +50,13 @@ func (p *AMIProvider) Get(ctx context.Context, constraints *Constraints) (string
 	if err != nil {
 		return "", fmt.Errorf("kube server version, %w", err)
 	}
-	name := fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, KubeToAWSArchitectures[*constraints.Architecture])
+	var architectureSuffix bytes.Buffer
+	if *constraints.Architecture == v1alpha3.ArchitectureArm64 {
+		architectureSuffix.WriteString("-arm64")
```
bwagner5, line 55: why not just use a string here rather than a byte buffer?

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -192,26 +193,51 @@ func (p *LaunchTemplateProvider) getSecurityGroupIds(ctx context.Context, provis
 
 func (p *LaunchTemplateProvider) getUserData(ctx context.Context, provisioner *v1alpha3.Provisioner, constraints *Constraints) (string, error) {
 	var userData bytes.Buffer
-	userData.WriteString(fmt.Sprintf("[settings.kubernetes]\napi-server = \"%s\"\n", provisioner.Spec.Cluster.Endpoint))
-	userData.WriteString(fmt.Sprintf("cluster-name = \"%s\"\n", *provisioner.Spec.Cluster.Name))
+	userData.WriteString(fmt.Sprintf(`#!/bin/bash
+yum install -y https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm
```
bwagner5, line 197: `ec2-downloads-windows`? Weird, I tested this and it seems to work, just odd bucket name I guess

JacobGabrielson: Yes, I believe it is due to the fact that the ssm-agent was originally developed for Windows, and later was extended to support Linux as well.

pkg/cloudprovider/aws/ami.go
JacobGabrielson: fixed

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -192,26 +193,51 @@ func (p *LaunchTemplateProvider) getSecurityGroupIds(ctx context.Context, provis
...
+/etc/eks/bootstrap.sh '%s' \
```
bwagner5, line 198: does this bootstrap script mean that the userdata only works with EKS or will this work fine with kops and other control-plane deployments?

JacobGabrielson: I had the same question - according to @ellistarn the script, despite the name, is not EKS-specific.

JacobGabrielson:
> This is a situation where I'd welcome using the new "Yes, issue opened" option. :)

After looking into it more, there don't seem to be any docs to update based on this change (other than the minor edits I added)

pkg/cloudprovider/aws/launchtemplate.go
bwagner5: ok that's good, I guess we'll see. I'm about to set this up in kops

bwagner5 (approved):
lgtm!

PR #669, opened 2021-09-09
Remove envsubst dependency

bellkev:
**1. Issue, if available:**

n/a, but discussed with @ellistarn 

Moved `eksctl` config file inline in heredoc instead of in a separate file. The main motivator for this is to remove the dependency on envsubst (which is not installed by default on macOS). Also made a small change to clarify that the cluster creation does not actually create a new IAM role, but rather sets up an IAM OIDC provider to allow subsequent association of IAM roles with Service Accounts.

- [ x ] Yes, PR includes docs updates
- [ ] No


geoffcline:
I will have to think on this. I had some schemes around automated testing of guides and formatting of yaml I was working on personally. Please don't merge in the next 24 hours.

website/content/en/docs/getting-started/_index.md
```diff
@@ -52,12 +52,28 @@ AWS_ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
...
 ### Create a Cluster
 
-Create a cluster with `eksctl`. The [example configuration](eksctl.yaml) file specifies a basic cluster (name, region), and an IAM role for Karpenter to use.
+Create a cluster with `eksctl`. This example configuration file specifies a basic cluster with one initial node and sets up an IAM OIDC provider for the cluster to enable IAM roles for pods:
 
 ```bash
-curl -fsSL https://karpenter.sh/docs/getting-started/eksctl.yaml \
-  | envsubst \
-  | eksctl create cluster -f -
+cat <<EOF > cluster.yaml
```
ellistarn, line 58: Iirc it's possible to pass the file directly to eksctl inline to avoid dropping a file on the local filesystem.

ellistarn (commented):
Thank you!!

website/content/en/docs/getting-started/_index.md
bellkev: It is and would work here. Personally I just always want to take a look at the config file before launching, because it takes like 20min to launch cluster + ng and some things are annoying to change later. I wouldn’t object to going back to the `eksctl -f -` option though.

bellkev:
> I will have to think on this. I had some schemes around automated testing of guides and formatting of yaml I was working on personally. Please don't merge in the next 24 hours.
> 
> Edit: also, welcome to the repo, fellow newcomer :)

I thought about that and figured it was definitely better not to have a separate copy of the cluster config inline and in another file. I was thinking there might be some kind of “include” tag in hugo that could help, but didn’t know enough about the website build process to attempt it now.

website/content/en/docs/getting-started/_index.md
ellistarn: For consistency with the rest of the guide, I'd probably lean doing it all inline. 

> Personally I just always want to take a look at the config file before launching

I often edit them inline after pasting but before entering them.

> Also, sometimes multi-line pasting can subtly break depending on the terminal emulator, etc.

This is a decent point, but at some point, there's only so much we can do to make this smooth with all user environments

geoffcline:
Started a discussion in slack regarding the merits of inline yaml vs hosting a file. Removing blocked label.

ellistarn (approved):
Approved. Happy to reapprove if you accept my last comment.

ellistarn:
Thank you!

PR #675, opened 2021-09-13
Set docsy to use main branch rather than master

bwagner5:
Set docsy to use main branch rather than master.

Reported by @haugenj


ellistarn:
What was broken by this? Previews?

haugenj:
> What was broken by this? Previews?

image

these led to 404s

ellistarn:
Nice!

ellistarn (approved):

PR #676, opened 2021-09-14
v1alpha4 API Proposal

ellistarn:
**1. Issue, if available:**
https://github.com/awslabs/karpenter/issues/661

- [X] No

### DIFF
--- a/designs/v1alpha4-api.md
+++ b/designs/v1alpha4-api.md
@@ -1,0 +1,114 @@
+# v1alpha4 API Proposal
+This document proposes comprehensive Provisioner API improvements prior to the v0.4 release. Due to minor backwards incompatible changes, this will result in a API version bump (v1alpha3 -> v1alpha4). This is in accordance with [Kubernetes API Deprecation Policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/). This API is not considered to be the final state of the Provisioner API, and is expected to evolve further as we explore more advanced scale down (defragmentation).
+
+These changes recommend:
+1. Cloud Provider specific extensions under `spec.provider`.
+2. Removal of `spec.cluster`.
+3. Pluralization of `spec.architecture` and `spec.operatingSystem`.
+4. Provisoning limits for maximum cpu, memory, etc, under `spec.limits`.
+
+## Strongly Typed Vendor Specific Fields
+Cloud Providers are currently limited to using well known `spec.labels` for configuration of vendor specific parameters. For example:
+- `spec.labels['node.k8s.aws/launch-template-name']`
+- `spec.labels['node.k8s.aws/subnet-name']`
+
+One benefit of this approach is that pods may use corresponding `spec.nodeSelector[...]` to request additional constraints on provisioned nodes. This information must be communicated through Kubernetes label values, which is awkward for the following use cases:
+1. Parameters are a list (e.g. `subnets: ["subneta", "subnetb", "subnetc"]`)
+2. Parameters are a struct (e.g. `tags: { "foo" : "bar" }`)
+3. Parameters do not comply with [the label value character set](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set). See: [#646](https://github.com/awslabs/karpenter/issues/646).
+
+We introduce a new `provider` field that enables strongly typed vendor specific parameters without violating vendor neutrality principles in the Karpenter codebase. We leverage Kubernetes `runtime.RawExtensions` to encapsulate these fields as raw bytes, which are then unmarshaled in vendor specific code. Vendors may implement arbitrary validation, defaulting, and provisioning behavior over the entire structure of these extensions. These structures are versioned separately from the Provisioner GVK to enable Cloud Providers to make backwards incompatible changes to provider specific configuration without requiring a version bump to the Provisioner CRD. If versioning is not specified by the user, it will be inferred and defaulted.
+
+```yaml
+apiVersion: karpenter.sh/v1alpha4
+kind: Provisioner
+spec:
+  provider:
+    apiVersion: extensions.karpenter.sh/v1alpha1
+    kind: AWS
+    securityGroups: ["abc", "def"]
+    subnets: ["123", "456"]
+    launchTemplateName: "foo"
+```
+
+The Cloud Provider *may* continue to support corresponding well known labels at the pod level (e.g. `node.k8s.aws/subnet-name`).
+
+```yaml
+# PodSpec: Simple key value constraints
+spec:
+  nodeSelector:
+    node.k8s.aws/subnet-name: "123" # Vendor Specific Field
+    kubernetes.io/instance-type: "m5.large" # Vendor Neutral Field
+```
+
+```yaml
+# PodSpec: More expressive than node selectors, can specify multiple or preferences
+spec:
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key: node.k8s.aws/subnet-name
+            operator: In
+            values: ["123", "456"]
+```
+
+*Note: Not all `spec.provider` fields must have a corresponding well known label.*
+
+### Alternative: Inline vs Nested Parameters
+It's possible to leverage inline json capabilities to collapse these parameters to the top level. For example
+```yaml
+spec:
+  zones: ["us-west-2a", "us-west-2b"] # Vendor neutral
+  securityGroups: ["abc", "def"] # Vendor specific, but sibling of zones
+```
+
+1. (+) Parameters that are natural siblings (e.g. `zones`, `securityGroups`) will be no longer be separated by `provider`.
+2. (-) It's not immediately clear to readers which parameters are vendor specific.
+3. (-) Key name conflicts may arise between provider and vendor fields.
+4. (-) `kind` and `apiVersion` are awkwardly placed at the top level of spec.
+
+### Alternative: Duck Typing
+
+It's possible to follow [Knative's Duck Typing approach](https://www.youtube.com/watch?v=kldVg63Utuw) and build vendor specific CRDs that contain vendor neutral snippets (or ducks) that Karpenter can recognize. Cloud Providers would consume Karpenter generic controllers as libaries, which would behave against the generic API snippets. For example:
+
+```yaml
+apiVersion: karpenter.k8s.aws/v1alpha1 # Vendor Specific
+kind: Provisioner
+spec:
+  limits: {} # Vendor neutral, recognized by Karpenter generic controllers
+  labels: {} # Vendor neutral, recognized by Karpenter generic controllers
+  taints: [] # Vendor netural, recognized by Karpenter generic controllers
+  subnets: [] # Vendor specific, only recognized by AWS Cloud Provider code
+```
+
+1. (+) Versioning is only defined in one place
+2. (+) Vendors may change versions decoupled from each other.
+3. (+) Parameters that are natural siblings (e.g. `zones`, `securityGroups`) will be no longer be separated by `provider`.
+4. (-) Cloud Provider must do more than simply implementing an interface (they must define an CR, etc).
+5. (-) Potential for name collision as duck types evolve or confusion about what fields providers own.
+6. (-) Knative duck typing APIs are alpha, missing documentation, and not widely adopted.
+
+## Pluralizion for all Constraints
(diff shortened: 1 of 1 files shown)


designs/v1alpha4-api.md
```diff
@@ -0,0 +1,130 @@
...
+spec:
+  provider:
+    apiVersion: extensions.karpenter.sh/v1alpha1
+    kind: AWS
+    securityGroups: ["abc", "def"]
+    subnets: ["123", "456"]
+    launchTemplateName: "foo"
+```
+
+The Cloud Provider *may* continue to support corresponding well known labels at the pod level (e.g. `node.k8s.aws/subnet-name`).
```
JacobGabrielson, line 34: It is not clear whether this will hold true for the alternatives listed in later sections.

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,130 @@
...
+        - matchExpressions:
+          - key: node.k8s.aws/subnet-name
+            operator: In
+            values: ["123", "456"]
+```
+
+*Note: Not all `spec.provider` fields must have a corresponding well known label.*
+
+### Alternative: Inline vs Nested Parameters
+It's possible to leverage inline json capabilities to collapse these parameters to the top level. For example
```
JacobGabrielson, line 60: It is not clear what 'inline json capabilities' means - capabilities of what/where?

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,130 @@
...
+1. Parameters are a list (e.g. `subnets: ["subneta", "subnetb", "subnetc"]`)
+2. Parameters are a struct (e.g. `tags: { "foo" : "bar" }`)
+3. Parameters do not comply with [the label value character set](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set). See: [#646](https://github.com/awslabs/karpenter/issues/646).
+
+We introduce a new `provider` field that enables strongly typed vendor specific parameters without violating vendor neutrality principles in the Karpenter codebase. We leverage Kubernetes `runtime.RawExtensions` to encapsulate these fields as raw bytes, which are then unmarshaled in vendor specific code. Vendors may implement arbitrary validation, defaulting, and provisioning behavior over the entire structure of these extensions. These structures are versioned separately from the Provisioner GVK to enable Cloud Providers to make backwards incompatible changes to provider specific configuration without requiring a version bump to the Provisioner CRD. If versioning is not specified by the user, it will be inferred and defaulted.
+
+```bash
```
JacobGabrielson, line 22: why are these `bash` rather than `yaml`?

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,130 @@
...
+It's possible to follow [Knative's Duck Typing approach](https://www.youtube.com/watch?v=kldVg63Utuw) and build vendor specific CRDs that contain vendor neutral snippets (or ducks) that Karpenter can recognize. Cloud Providers would consume Karpenter generic controllers as libaries, which would behave against the generic API snippets. For example:
+
+```bash
+apiVersion: karpenter.k8s.aws/v1alpha1 # Vendor Specific
+kind: Provisioner
+spec:
+  limits: {} # Vendor neutral, recognized by Karpenter generic controllers
+  labels: {} # Vendor neutral, recognized by Karpenter generic controllers
+  taints: [] # Vendor netural, recognized by Karpenter generic controllers
+  subnets: [] # Vendor specific, only recognized by AWS Cloud Provider code
```
JacobGabrielson, line 83: Aren't you back to the name collision (potential) problem here?

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+1. (+) Versioning is only defined in one place
+2. (+) Vendors may change versions decoupled from each other.
+3. (+) Parameters that are natural siblings (e.g. `zones`, `securityGroups`) will be no longer be separated by `provider`.
+4. (-) Cloud Provider must do more than simply implementing an interface (they must define an CR, etc).
+
```
JacobGabrielson, line 90: I think there's a con here that customers can get confused about whether a field is vendor-specific or not. And also that when "porting" someone else's `yaml` code from one vendor to another, they may not even realize certain fields need to change (e.g., the `subnets` example above).

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,130 @@
...
+This approach is quite attractive, but requires significant effort and changes to the project.
```
JacobGabrielson, line 91: I'd probably re-word this a bit, as "quite attractive" is somewhat difficult to quantify. I'd suggest more focusing on how much effort it would be (perhaps quantify that a bit) as a reason not to do it. (I'm also not 100% sure all of the potential downsides were listed?)

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,130 @@
+# v1alpha4 API Proposal
+This document proposes comprehensive Provisioner API improvements prior to the v0.4 release. Due to minor backwards incompatible changes, this will result in a API version bump (v1alpha3 -> v1alpha4). This is in accordance with [Kubernetes API Deprecation Policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/).
+
+These changes recommend:
+1. Cloud Provider specific extensions under `spec.provider`.
+2. Removal of `spec.cluster`.
+3. Pluralization of `spec.architecture` and `spec.operatingSystem`.
```
JacobGabrielson, line 7: This was pretty confusing, I thought you meant you were going to change the names to (eg) `operatingSystems`.

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,130 @@
...
+    ready: 500 # Flat or Percentage. Karpenter will not launch additional capacity if ready nodes exceeds this value
+    unready: 20% # Flat or Percentage. Karpenter will not launch additional capacity if current unready nodes exceeds this value
+    resources: # Karpenter will not launch additional capacity if current capacity exceeds this value
+      cpu: 1000
+      memory: 1000Gi
+```
+
+## Reparenting top level fields
+
+With the introduction of additional top level fields, top level parameters no longer carry the same conceptual weight. To address this, we will add parent fields for `constraints` and `scale down`.
```
JacobGabrielson, line 119: I'm having trouble parsing this... why won't they carry the same "conceptual weight?"

designs/v1alpha4-api.md
ellistarn: In my mind, we'd be resetting most of these and just adding them at the provisioner level (for now).

designs/v1alpha4-api.md
ellistarn: https://play.golang.org/p/NnwTh9KI5r

designs/v1alpha4-api.md
ellistarn: good call

designs/v1alpha4-api.md
ellistarn: Yep!

designs/v1alpha4-api.md
ellistarn: I do!

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,115 @@
...
+2. Parameters are a struct (e.g. `tags: { "foo" : "bar" }`)
+3. Parameters do not comply with [the label value character set](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set). See: [#646](https://github.com/awslabs/karpenter/issues/646).
+
+We introduce a new `provider` field that enables strongly typed vendor specific parameters without violating vendor neutrality principles in the Karpenter codebase. We leverage Kubernetes `runtime.RawExtensions` to encapsulate these fields as raw bytes, which are then unmarshaled in vendor specific code. Vendors may implement arbitrary validation, defaulting, and provisioning behavior over the entire structure of these extensions. These structures are versioned separately from the Provisioner GVK to enable Cloud Providers to make backwards incompatible changes to provider specific configuration without requiring a version bump to the Provisioner CRD. If versioning is not specified by the user, it will be inferred and defaulted.
+
+```yaml
+apiVersion: karpenter.sh/v1alpha4
+kind: Provisioner
+spec:
+  provider:
```
bwagner5, line 26: ++ provider :)

JacobGabrielson (dismissed):

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,115 @@
...
+3. Pluralization of `spec.architecture` and `spec.operatingSystem`.
+4. Provisoning limits for maximum cpu, memory, etc, under `spec.limits`.
+
+## Strongly Typed Vendor Specific Fields
+Cloud Providers are currently limited to using well known `spec.labels` for configuration of vendor specific parameters. For example:
+- `spec.labels['node.k8s.aws/launch-template-name']`
+- `spec.labels['node.k8s.aws/subnet-name']`
+
+One benefit of this approach is that pods may use corresponding `spec.nodeSelector[...]` to request additional constraints on provisioned nodes. This information must be communicated through Kubernetes label values, which is awkward for the following use cases:
+1. Parameters are a list (e.g. `subnets: ["subneta", "subnetb", "subnetc"]`)
```
geoffcline, line 16: I'd change to "may be a list" unless all parameters are a list.

ellistarn: This is in reference to the previous "in the following use cases" -> "parameters are a list", "parameters are a struct"

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+The Cloud Provider *may* continue to support corresponding well known labels at the pod level (e.g. `node.k8s.aws/subnet-name`).
+
+```yaml
+# PodSpec: Simple key value constraints
+spec:
+  nodeSelector:
+    node.k8s.aws/subnet-name: "123" # Vendor Specific Field
+    kubernetes.io/instance-type: "m5.large" # Vendor Neutral Field
```
geoffcline, line 41: oh so it's vendor neutral based on the key and not the value?

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Limits
+
+We introduce a new field `spec.limits` that contains configuration parameters to limit scaling and control costs.
+
+### Example
+```yaml
+spec:
+  limits:
+    ready: 500 # Flat or Percentage. Karpenter will not launch additional capacity if ready nodes exceeds this value
```
anguslees, line 110: What does "ready=90%" mean?  (percentage of what total?)

designs/v1alpha4-api.md
anguslees: Structural vs nominal types.  Discuss.  :stuck_out_tongue: 

Unless we _want_ to generalise over abstract providers for some reason, I recommend avoiding this and sticking to composition of nominal types (the RawExtension recommendation above).

designs/v1alpha4-api.md
ellistarn: This is a good point. It slipped my mind when I was thinking "IntOrString all the things" and "unready should have parity, so let's make ready". I suppose it could mean, this provisioner can deploy 50% of the nodes in this cluster. Though I'd much prefer that the provisioners are their own universes and are not aware of other provisioners. 

I keep going back and forth on a node count as a limit (as opposed to resource limits below). I'm leaning simply leaving this out entirely until I have a real use case for it.

bwagner5 (approved):
nice doc!

designs/v1alpha4-api.md
ellistarn: Correct -- the value is just a string

designs/v1alpha4-api.md
elmiko: +1, this looks like a great improvement. from our discussion on Sep 16, would this be an appropriate place for a provider to inject a secret location for kubeconfigs that it might need?

designs/v1alpha4-api.md
```diff
@@ -0,0 +1,115 @@
...
+3. (+) Parameters that are natural siblings (e.g. `zones`, `securityGroups`) will be no longer be separated by `provider`.
+4. (-) Cloud Provider must do more than simply implementing an interface (they must define an CR, etc).
+5. (-) Potential for name collision as duck types evolve or confusion about what fields providers own.
+6. (-) Knative duck typing APIs are alpha, missing documentation, and not widely adopted.
+
+## Pluralizion for all Constraints
+We expand the constraints of a Provisioner from a scalar to a vector in all cases. This will apply to both `operatingSystem` and `architecture` enabling the operator to specify greater flexibility. This supports use cases such as heterogenous architectures within a single provisioner, selected dynamically at runtime by the cloud provider. More importantly, this change creates a consistent and predictable semantic for all vendor neutral constraints. Similar to other constraints, the semantic of this change allows cloud providers to choose any value in the constraint slice, e.g. prioritizing arm64 for cost reasons.
+### Example
+```yaml
+spec:
```
bwagner5, line 97: How would these be expanded?  We'd need to inspect the container manifest's support for the architecture and operating system before we could expand the constraints which is possible.  

Architecture could be considered optimistic where Karpenter assumes that a container for both amd64 and arm64 exist. But that same strategy couldn't be used for OS, I don't think (at least from a practical use-case).

designs/v1alpha4-api.md
ellistarn: Correct!

bwagner5 (approved):
lgtm, nice doc

Issue #612, opened 2021-08-12
Emit Metrics for Karpenter


vinayan3:
**Tell us about your request**
What do you want us to build?
Emit metrics in prometheus format for Karpenter. The follow metrics would be great and are mainly suggestions on what operationally I've found useful to monitor the open source clusuter autoscaler. Therefore some of the metrics below may be less applicable to Karpenter but this gives a sense of what is useful to monitor

Metric Name | Metric Type | Labels | Description
------------ | -------------  | ------------ | ------------ |
reconcile_duration_ms | Histogram | N/A| How long the reconcile loop takes where it's looking at the unscheduled pods
nodes_added | Gauge | node_type | Number of nodes that were created for a certain type
node_creation_duration_ms | Histogram | provisioner_name | How long it takes from creating a fleet request to having a node being created
node_creation_failures | Gauge | provisioner_name, reason | Number of failures to get capacity and an indication of reason
nodes_marked_for_deletion | Gauge | provisioner_name | Number of nodes that are marked for deletion
nodes_deleted | Gauge | provisioner_name | Number of nodes that were deleted
spot_fleet_calls | Counter | path, method | Number of calls made; there maybe times when no response every comes so knowing a request was sent is very useful
spot_fleet_calls_duration_ms | Histogram | path, method, status_code | create/delete/get calls
ec2_calls | Counter | method | Number of calls made
ec2_calls_response_duration_ms | Histogram | path, method, status_code | create/delete/get calls
kube_api_calls | Counter | path, method | Calls made to the Kube API
kube_api_response_duration_ms | path, method, status_code  | create/delete/get calls

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
Having metrics emitted by Karpenter will allow someone to observe what Karpenter is doing. It's important to understand what actions is taking or external resources Karpenter is accessing. 

**Are you currently working around this issue?**
There isn't really a work around.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

tsunamishaun:
[...]

PR #677, opened 2021-09-14
add allocation bind duration metric

cjerad:
**1. Issue, if available:**
#612  "Emit Metrics for Karpenter"

This PR does _not_ fully resolve the issue.  More changes will be needed.

Add a histogram metric to record the duration of bind operations.

- [x] No


pkg/controllers/allocation/bind.go
```diff
@@ -27,14 +29,23 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"knative.dev/pkg/logging"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
-type Binder struct {
-	KubeClient   client.Client
-	CoreV1Client corev1.CoreV1Interface
+type Binder interface {
```
ellistarn, line 35: Curious why you promoted this to an interface? Do we expect multiple implementations?

pkg/controllers/allocation/bind.go
```diff
@@ -82,3 +93,55 @@ func (b *Binder) bind(ctx context.Context, node *v1.Node, pod *v1.Pod) error {
...
+	binder               Binder
+	bindTimeHistogramVec *prometheus.HistogramVec
+}
+
+const metricLabelResult = "result"
+
+func DecorateBinderMetrics(binder Binder) Binder {
+	bindTimeHistogramVec := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "karpenter",
```
ellistarn, line 107: Thoughts on building a /pkg/metrics component with helpers to encapsulate the buckets, Name, Subsystem, etc?

```
package metrics

type HistogramOptions {
   Namespace: string // defaults to karpenter
   Buckets []float64
   ...
}

func NewHistogram(options HistogramOptions) {
   // default options if some not set (e.g. buckets)
}
```

pkg/controllers/allocation/bind.go
```diff
@@ -82,3 +93,55 @@ func (b *Binder) bind(ctx context.Context, node *v1.Node, pod *v1.Pod) error {
...
+	startTime := time.Now()
+	bindErr := b.binder.Bind(ctx, node, pods)
+	durationSeconds := time.Since(startTime).Seconds()
+
+	result := "success"
+	if bindErr != nil {
+		result = "error"
+	}
+
+	observer, promErr := b.bindTimeHistogramVec.GetMetricWith(prometheus.Labels{metricLabelResult: result})
```
ellistarn, line 133: Under what scenarios can we get an error? My understanding is that this is all static memory allocation.

pkg/controllers/allocation/bind.go
```diff
@@ -82,3 +93,55 @@ func (b *Binder) bind(ctx context.Context, node *v1.Node, pod *v1.Pod) error {
...
 }
+
+type binderMetricsDecorator struct {
+	binder               Binder
+	bindTimeHistogramVec *prometheus.HistogramVec
+}
+
+const metricLabelResult = "result"
+
+func DecorateBinderMetrics(binder Binder) Binder {
```
ellistarn, line 104: I wonder if (in the future) we could abstract this decoration even further and minimally wrap the code as we need it with something generic.

I'm imagining an interface like
```
// inside binder

var err error
metrics.recordLatency(metrics.LatencyOptions{ Name: "bind" }, func() {
  err = b.Bind()
})
```

pkg/controllers/allocation/bind.go
```diff
@@ -82,3 +93,55 @@ func (b *Binder) bind(ctx context.Context, node *v1.Node, pod *v1.Pod) error {
...
+	if bindErr != nil {
+		result = "error"
+	}
+
+	observer, promErr := b.bindTimeHistogramVec.GetMetricWith(prometheus.Labels{metricLabelResult: result})
+	if promErr != nil {
+		logging.FromContext(ctx).Warnf(
+			"Failed to record bind duration metric [%s=%s, duration=%f]: error=%w",
+			metricLabelResult,
+			result,
```
ellistarn, line 138: I'm a bit wary of the code complexity of tracking everything this error is going to log. Especially since we will already bubble up the errors in the normal karpenter controller logic to print something along the lines of (failed to X).

pkg/controllers/allocation/bind.go
```diff
@@ -82,3 +93,55 @@ func (b *Binder) bind(ctx context.Context, node *v1.Node, pod *v1.Pod) error {
 	}
 	return nil
 }
+
+type binderMetricsDecorator struct {
+	binder               Binder
+	bindTimeHistogramVec *prometheus.HistogramVec
+}
+
+const metricLabelResult = "result"
```
ellistarn, line 102: Any reason this isn't inlined?

pkg/controllers/allocation/bind.go
```diff
@@ -82,3 +93,55 @@ func (b *Binder) bind(ctx context.Context, node *v1.Node, pod *v1.Pod) error {
...
+
+func DecorateBinderMetrics(binder Binder) Binder {
+	bindTimeHistogramVec := prometheus.NewHistogramVec(
```
ellistarn, line 105: Thoughts on putting this whole thing in a Var that can be accessed at package scope?

e.g. https://github.com/prometheus/prometheus/blob/main/discovery/kubernetes/client_metrics.go

ellistarn (commented):
I'm curious how many metrics we get from client-go for free: https://github.com/kubernetes/client-go/blob/master/tools/metrics/metrics.go. Maybe there's a way to wire this up?

pkg/controllers/allocation/bind.go
cjerad: It facilitates using the decorator pattern, with the metric logic being in a separate decorate type.

pkg/controllers/allocation/bind.go
cjerad: An old habit of avoiding bare keys in map lookups; misspellings cause runtime problems.  When using a constant, misspellings cause compile-time errors.

pkg/controllers/allocation/bind.go
cjerad: Perhaps after more metrics have been implemented and patterns have been identified.  Defining constants for *Namespace* and *Subsystem* seems likely, but not necessary in this commit as each appears only once in the code.

pkg/controllers/allocation/bind.go
cjerad: Underlying implementations may change over time.  Since the API *may* return an `error` it seems advisable to treat it as though it *could* fail, even if the current implementation makes that very unlikely.

Another question is whether a failure in the metrics reporting is considered a recoverable error or whether karpenter should `panic`?  This implementation assumes such a failure is not a cause to `panic`.

pkg/controllers/allocation/bind.go
cjerad: Should the karpenter controller logic treat a failure to record the metric as a failure of the bind operation?

bwagner5: I don't think metric emission errors should bubble up to the controller interface on bind. I think the warn provides enough visibility if this were to fail, but it's not necessarily critical to the actual bind operation (and it's unlikely to occur).

pkg/controllers/allocation/bind.go
bwagner5: I would agree, that metric emission is not a cause to panic.

pkg/controllers/allocation/bind.go
ellistarn: If it was errorable, I agree, but I'm unsure whether or not recording a metric should be able to fail. 

Taking a quick peek at the implementation, it looks like the only way it can fail is if the labels can't be hashed. Of course, this implementation could change, but I expect from how I've seen prometheus metrics used across k8s projects is that the intent of these metrics is to be used as statics. 
For example we could do something like. If it does change in an unexpected way, we'll fail fast at process startup.
```
var (
  bindLatencyHistogramVec = prometheus.NewHistogramVec( ... )
  successfulBindLatencyHistogram = runtimeutil.Must(bindLatencyHistogramVec.GetMetricWith( ... )
  failedBindLatencyHistogram = runtimeutil.Must(bindLatencyHistogramVec.GetMetricWith( ... )
)

func bind(... ) {
   ...
   successfulBindLatencyHistogram.Observe(...)
}
```

pkg/controllers/allocation/bind.go
ellistarn: See above

ellistarn (commented):
Client-go instruments requests for all API calls made, including `/pods/bind` (https://github.com/kubernetes/client-go/blob/master/tools/metrics/metrics.go). This means we should have a bunch of metrics for bind latency (on individual bind pod requests).

That name conflicts with our "bind" concept which means "create the node, bind all the pods". Now that we're exposing the name of this class to customers as a metric, I think it might make sense to rename the concept slightly to avoid confusing customers. I'm struggling a bit with exactly what this name should be. Perhaps something like `forge_packing`, `apply_packing`.

pkg/controllers/allocation/bind.go
ellistarn: Thinking further on this, I wonder if we should be including Provisioner labels (name/namespace) into these metrics, which would make my above points moot (since the labels are dynamic).

pkg/controllers/allocation/bind.go
ellistarn: I worry a bit about this pattern for structs that have more methods. You could use struct embedding to achieve a similar mechanism without requiring full method overloading. 
```
type DecoratedBinder {
   Binder // embedded struct, all methods are promoted to DecoratedBinder
}

func (b *Binder) foo() {} 
func (b *Binder) bar() {}

func (d *DecoratedBinder) foo() { 
  // decorate before
  d.Binder.foo()
  // decorate after
}
```
In this case, bar is not overloaded, so anyone calling `decoratedBinder.bar()` will get the undecorated implementation.

bwagner5 (approved):
lgtm

ellistarn (approved):

PR #678, opened 2021-09-14
Ignores stuck pods rather than deleting them to avoid stateful set edge cases

ellistarn:
**1. Issue, if available:**

@anguslees pointed out that we must allow the kubelet to delete pods that are terminating to avoid violating stateful set guarantees if a kubelet is partitioned. Instead, we simply ignore pods that are past their grace window and delete the node. This ensures the guarantee is met, since the pod will be deleted by KCM once the node no longer exists.

### Testing
1. Modified my provisioner to have a bad endpoint.
2. Created pod
3. Node came online, didn't connect
4. Pod tolerations ran out after 5 minute, pod evicted (w/ grace period 30 sec default)
5. Node deleted by liveness controller (hanging)
6. Node waits for pod to terminate (grace period)
7. Node deleted by termination controller
8. Pod cleaned up by KCM (2 minutes later)

- [X] No

### DIFF
--- a/pkg/controllers/node/suite_test.go
+++ b/pkg/controllers/node/suite_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha3"
 	"github.com/awslabs/karpenter/pkg/controllers/node"
 	"github.com/awslabs/karpenter/pkg/test"
+	"github.com/awslabs/karpenter/pkg/utils/injectabletime"
 
 	. "github.com/awslabs/karpenter/pkg/test/expectations"
 	. "github.com/onsi/ginkgo"
@@ -66,7 +67,7 @@ var _ = Describe("Controller", func() {
 	})
 
 	AfterEach(func() {
-		node.Now = time.Now
+		injectabletime.Now = time.Now
 		ExpectCleanedUp(env.Client)
 	})
 
@@ -108,7 +109,7 @@ var _ = Describe("Controller", func() {
 			Expect(n.DeletionTimestamp.IsZero()).To(BeTrue())
 
 			// Simulate time passing
-			node.Now = func() time.Time {
+			injectabletime.Now = func() time.Time {
 				return time.Now().Add(time.Duration(*provisioner.Spec.TTLSecondsUntilExpired) * time.Second)
 			}
 			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(n))
@@ -197,7 +198,7 @@ var _ = Describe("Controller", func() {
 			Expect(n.DeletionTimestamp.IsZero()).To(BeTrue())
 
 			// Simulate time passing and a n failing to join
-			node.Now = func() time.Time { return time.Now().Add(node.LivenessTimeout) }
+			injectabletime.Now = func() time.Time { return time.Now().Add(node.LivenessTimeout) }
 			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(n))
 
 			n = ExpectNodeExists(env.Client, n.Name)
@@ -219,7 +220,7 @@ var _ = Describe("Controller", func() {
 			Expect(n.DeletionTimestamp.IsZero()).To(BeTrue())
 
 			// Simulate time passing and a n failing to join
-			node.Now = func() time.Time { return time.Now().Add(node.LivenessTimeout) }
+			injectabletime.Now = func() time.Time { return time.Now().Add(node.LivenessTimeout) }
 			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(n))
 
 			n = ExpectNodeExists(env.Client, n.Name)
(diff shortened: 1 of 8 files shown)


pkg/controllers/termination/terminate.go
```diff
@@ -135,7 +131,7 @@ func (t *Terminator) evict(ctx context.Context, pods []*v1.Pod) {
 	critical := []*v1.Pod{}
 	nonCritical := []*v1.Pod{}
 	for _, pod := range pods {
-		if !pod.DeletionTimestamp.IsZero() {
+		if pod.DeletionTimestamp != nil {
```
njtran, line 134: Any reason to not use the IsZero() function here?

pkg/controllers/termination/terminate.go
```diff
@@ -153,8 +149,8 @@ func (t *Terminator) evict(ctx context.Context, pods []*v1.Pod) {
 }
 
 func IsStuckTerminating(pod *v1.Pod) bool {
-	if pod.DeletionTimestamp.IsZero() {
+	if pod.DeletionTimestamp == nil {
 		return false
 	}
-	return time.Since(pod.DeletionTimestamp.Time) > time.Duration(ptr.Int64Value(pod.DeletionGracePeriodSeconds))*time.Second
+	return injectabletime.Now().After(pod.DeletionTimestamp.Time)
```
njtran, line 155: I think we want to check the `DeletionGracePeriodSeconds` here. Evictions don't always take some `ms` amount of time, and we want to make sure we don't consider it a pod that `IsStuckTerminating` if it's just taking longer to evict.

Additionally, this would essentially always be true whenever the `DeletionTimestamp` is non-nil, boiling it down to a `return true`. `DeletionTimestamp` is always nil if it hasn't been told to delete, and we would only check if it `IsStuckTerminating()` after we have told it to delete. Thus, since this logically comes after deleting a pod, this will always be true.

pkg/controllers/termination/controller.go
```diff
@@ -36,6 +36,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+// Now is a time.Now() that may be mocked by tests.
+var Now = time.Now
```
njtran, line 40: Do we need this if we have `utils/injectabletime`?

ellistarn: Thanks for this!

pkg/controllers/termination/terminate.go
ellistarn: Yep! It's actually the wrong semantic. If the DeletionTimestamp is truly 0 (e.g. 1969 or w/e), then we want to behave differently. Zero and Nil are different for our use case.

njtran: So weird. Thanks!

pkg/controllers/termination/terminate.go
ellistarn: Discussed offline.

njtran (dismissed):
Nice work and nice niche catches!

ellistarn:
> Nice work and nice niche catches!

niche caches!

pkg/controllers/node/suite_test.go
```diff
@@ -66,7 +67,7 @@ var _ = Describe("Controller", func() {
 	})
 
 	AfterEach(func() {
-		node.Now = time.Now
```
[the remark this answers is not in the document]
ellistarn: IMO time should work normally unless we explicitly need to control it.

pkg/controllers/node/suite_test.go
```diff
@@ -108,7 +109,7 @@ var _ = Describe("Controller", func() {
 			Expect(n.DeletionTimestamp.IsZero()).To(BeTrue())
 
 			// Simulate time passing
-			node.Now = func() time.Time {
+			injectabletime.Now = func() time.Time {
 				return time.Now().Add(time.Duration(*provisioner.Spec.TTLSecondsUntilExpired) * time.Second)
```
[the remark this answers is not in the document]
ellistarn: I'm not sure I understand. This effectively forces time to be 30 seconds into the future (which is the explicit behavior we're trying to test).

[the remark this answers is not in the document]
JacobGabrielson: what if the previous steps take a few seconds to execute, won't it now be (say) 32 seconds in the future, since we are calling "live" `time.Now`?

pkg/controllers/node/suite_test.go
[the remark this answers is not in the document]
JacobGabrielson: I doubt it's causing a problem right now, so not a blocking issue, but in general this will add variability - say the unit test process itself takes longer to schedule (on a heavily loaded machine running a lot of github actions for example), then some of these times might stretch longer. Probably not a big deal here since the code should generally do the right thing if things take "at least N seconds", but could lead to flakiness.

JacobGabrielson (approved):

PR #680, opened 2021-09-16
add allocation scheduling duration metric

cjerad:
**1. Issue, if available:**

Add a histogram metric to record the duration of scheduling operations.

- [x] No

### DIFF
--- a/pkg/controllers/allocation/controller.go
+++ b/pkg/controllers/allocation/controller.go
@@ -56,7 +56,7 @@ type Controller struct {
 	Batcher       *Batcher
 	Filter        *Filter
 	Binder        Binder
-	Scheduler     *scheduling.Scheduler
+	Scheduler     scheduling.Scheduler
 	Packer        binpacking.Packer
 	CloudProvider cloudprovider.CloudProvider
 	KubeClient    client.Client
@@ -68,7 +68,7 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 		Filter:        &Filter{KubeClient: kubeClient},
 		Binder:        DecorateBinderMetrics(NewBinder(kubeClient, coreV1Client)),
 		Batcher:       NewBatcher(maxBatchWindow, batchIdleTimeout),
-		Scheduler:     scheduling.NewScheduler(cloudProvider, kubeClient),
+		Scheduler:     scheduling.DecorateSchedulerMetrics(scheduling.NewScheduler(cloudProvider, kubeClient)),
 		Packer:        binpacking.NewPacker(),
 		CloudProvider: cloudProvider,
 		KubeClient:    kubeClient,
--- a/pkg/controllers/allocation/controller.go
+++ b/pkg/controllers/allocation/controller.go
@@ -56,7 +56,7 @@ type Controller struct {
 	Batcher       *Batcher
 	Filter        *Filter
 	Binder        Binder
-	Scheduler     scheduling.Scheduler
+	Scheduler     *scheduling.Scheduler
 	Packer        binpacking.Packer
 	CloudProvider cloudprovider.CloudProvider
 	KubeClient    client.Client
@@ -68,7 +68,7 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 		Filter:        &Filter{KubeClient: kubeClient},
 		Binder:        DecorateBinderMetrics(NewBinder(kubeClient, coreV1Client)),
 		Batcher:       NewBatcher(maxBatchWindow, batchIdleTimeout),
-		Scheduler:     scheduling.DecorateSchedulerMetrics(scheduling.NewScheduler(cloudProvider, kubeClient)),
+		Scheduler:     scheduling.NewScheduler(cloudProvider, kubeClient),
 		Packer:        binpacking.NewPacker(),
 		CloudProvider: cloudProvider,
 		KubeClient:    kubeClient,
--- a/pkg/controllers/allocation/scheduling/scheduler.go
+++ b/pkg/controllers/allocation/scheduling/scheduler.go
@@ -17,30 +17,29 @@ package scheduling
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha3"
 	"github.com/awslabs/karpenter/pkg/cloudprovider"
+	"github.com/awslabs/karpenter/pkg/metrics"
 	"github.com/mitchellh/hashstructure/v2"
+	"github.com/prometheus/client_golang/prometheus"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"knative.dev/pkg/logging"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
-func NewScheduler(cloudProvider cloudprovider.CloudProvider, kubeClient client.Client) *Scheduler {
-	return &Scheduler{
-		KubeClient: kubeClient,
-		Topology: &Topology{
-			cloudProvider: cloudProvider,
-			kubeClient:    kubeClient,
-		},
-	}
+type Scheduler interface {
+	Solve(context.Context, *v1alpha3.Provisioner, []*v1.Pod) ([]*Schedule, error)
 }
 
-type Scheduler struct {
-	KubeClient client.Client
-	Topology   *Topology
+type scheduler struct {
+	kubeClient client.Client
+	topology   *Topology
 }
 
 type Schedule struct {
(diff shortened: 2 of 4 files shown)


pkg/controllers/allocation/scheduling/scheduler.go
```diff
@@ -147,3 +156,56 @@ func IsSchedulable(pod *v1.Pod, node *v1.Node) bool {
...
+	startTime := time.Now()
+	schedules, scheduleErr := s.scheduler.Solve(ctx, provisioner, pods)
+	durationSeconds := time.Since(startTime).Seconds()
+
+	result := "success"
+	if scheduleErr != nil {
+		result = "error"
+	}
+
+	provisionerName := provisioner.ObjectMeta.Name
```
ellistarn, line 191: minor: I'd probably lean inlining this to reduce lines

pkg/controllers/allocation/controller.go
```diff
@@ -68,7 +68,7 @@ func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface
 		Filter:        &Filter{KubeClient: kubeClient},
 		Binder:        DecorateBinderMetrics(NewBinder(kubeClient, coreV1Client)),
 		Batcher:       NewBatcher(maxBatchWindow, batchIdleTimeout),
-		Scheduler:     scheduling.NewScheduler(cloudProvider, kubeClient),
+		Scheduler:     scheduling.DecorateSchedulerMetrics(scheduling.NewScheduler(cloudProvider, kubeClient)),
```
ellistarn, line 71: thoughts on encapsulating this inside NewScheduler so the test doesn't need to be aware it's decorated?

pkg/controllers/allocation/scheduling/scheduler.go
```diff
@@ -147,3 +156,56 @@ func IsSchedulable(pod *v1.Pod, node *v1.Node) bool {
 	// TODO, support node affinity
 	return true
 }
+
+type schedulerMetricsDecorator struct {
+	scheduler                Scheduler
```
ellistarn, line 161: Optional: 

type metricsDecoratedScheduler struct {
  Scheduler // embedded
  ...
}

pkg/controllers/allocation/scheduling/scheduler.go
```diff
@@ -147,3 +156,56 @@ func IsSchedulable(pod *v1.Pod, node *v1.Node) bool {
...
+
+	return &schedulerMetricsDecorator{scheduler: scheduler, scheduleTimeHistogramVec: scheduleTimeHistogramVec}
+}
+
+func (s *schedulerMetricsDecorator) Solve(ctx context.Context, provisioner *v1alpha3.Provisioner, pods []*v1.Pod) ([]*Schedule, error) {
+	startTime := time.Now()
+	schedules, scheduleErr := s.scheduler.Solve(ctx, provisioner, pods)
+	durationSeconds := time.Since(startTime).Seconds()
+
+	result := "success"
```
ellistarn, line 186: I'm noticing that controller runtime separate success and error cases into separate metrics https://github.com/kubernetes-sigs/controller-runtime/blob/v0.10.0/pkg/internal/controller/metrics/metrics.go#L37. Thoughts on following the same pattern?

pkg/controllers/allocation/scheduling/scheduler.go
```diff
@@ -147,3 +156,56 @@ func IsSchedulable(pod *v1.Pod, node *v1.Node) bool {
...
+			"Failed to record scheduling duration metric [%s=%s, %s=%s, duration=%f]: error=%w",
+			metrics.ProvisionerLabel,
+			provisionerName,
+			metrics.ResultLabel,
+			result,
+			durationSeconds,
+			promErr,
+		)
+	} else {
+		observer.Observe(durationSeconds)
```
ellistarn, line 207: Thoughts on adding a count metric, too? Or can we get this by counting the number of latency metrics?

pkg/controllers/allocation/scheduling/scheduler.go
```diff
@@ -147,3 +156,56 @@ func IsSchedulable(pod *v1.Pod, node *v1.Node) bool {
...
+			Buckets:   metrics.DurationBuckets(),
+		},
+		[]string{metrics.ProvisionerLabel, metrics.ResultLabel},
+	)
+	crmetrics.Registry.MustRegister(scheduleTimeHistogramVec)
+
+	return &schedulerMetricsDecorator{scheduler: scheduler, scheduleTimeHistogramVec: scheduleTimeHistogramVec}
+}
+
+func (s *schedulerMetricsDecorator) Solve(ctx context.Context, provisioner *v1alpha3.Provisioner, pods []*v1.Pod) ([]*Schedule, error) {
```
ellistarn, line 181: Given the similarity of all of these histograms, I think we could likely achieve a lot of code-deduplication. Feel free to punt for a future issue.

pkg/controllers/allocation/scheduling/scheduler.go
cjerad: A separate reference is used to ensure the same value used in the query (line 193) will also be logged (line 200) if the query fails.

pkg/controllers/allocation/controller.go
cjerad: That's a good point, I agree that this may not be the best place to apply the metrics decoration.  However, I would prefer pushing the metrics decoration up to the caller of `allocation.NewController()` rather than down into `scheduling.NewScheduler()`.

pkg/controllers/allocation/scheduling/scheduler.go
cjerad: Tried it.  The go compiler won't allow embedding an interface within a struct.

The compiler will allow embedding the `scheduler` struct but that had the effect of coupling `metricsDecoratedScheduler` to `scheduler`, rather than allowing `metricsDecoratedScheduler` to work with any `Scheduler`.

pkg/controllers/allocation/scheduling/scheduler.go
cjerad: I think using labels better encapsulates the metric information.  An analogy in code would be passing around a `struct` vs passing around multiple related but individual variables.

pkg/controllers/allocation/scheduling/scheduler.go
cjerad: The histogram also emits "<metric_name>_count" and "<metric_name>_total" metrics with the count of observations and the sum of all observations, respectively.

pkg/controllers/allocation/scheduling/scheduler.go
ellistarn: You can assume that provisioner.ObjectMeta.Name is immutable. In kubernetes, spec/metadata are considered immutable at reconcile time, while status is mutable.

pkg/controllers/allocation/scheduling/scheduler.go
ellistarn: awesome!

pkg/controllers/allocation/scheduling/scheduler.go
ellistarn: I think of the metricsdecorator as a subclass of scheduler (with is ~= to embedding in golang, though not quite the same). I tend to see the go philosophy as leaning away from things like interfaces, setters, getters, and instead towards duck typing and direct variable access. I also expect that we won't have multiple scheduler implementations, but I'm fine either way.

pkg/controllers/allocation/controller.go
ellistarn: It depends whether or not you think the scheduler should innately emit metrics (i.e., the decorator is a convenience impl detail) or whether something should be decorating it. I'd lean the former, as I don't see multiple scheduler implementations in our future. This does feel somewhat like go/java philosophies colliding.

Nonblocking, though.

pkg/controllers/allocation/scheduling/scheduler.go
ellistarn: Do you have some examples from similar kubernetes projects?

pkg/controllers/allocation/scheduling/scheduler.go
cjerad: The scenario of concern is that the provisioner name might be fetch from elsewhere.  Hypothetically, if a new field is added, e.g. `provisioner.spec.metricDisplayName`; then, if the lookup is inlined, whoever updates this code will need to be sure to update *both* locations to keep the log message in sync.

pkg/controllers/allocation/scheduling/scheduler.go
cjerad: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.10.0/pkg/internal/controller/metrics/metrics.go#L30-L33 brakes out the metric by result classification.

ellistarn: Great thanks!

pkg/controllers/allocation/scheduling/scheduler.go
```diff
@@ -51,7 +61,49 @@ type Schedule struct {
...
+		result = "error"
+	}
+
+	provisionerName := provisioner.ObjectMeta.Name
+	observer, promErr := scheduleTimeHistogramVec.GetMetricWith(prometheus.Labels{
+		metrics.ProvisionerLabel: provisionerName,
+		metrics.ResultLabel:      result,
+	})
+	if promErr != nil {
+		logging.FromContext(ctx).Warnf(
```
ellistarn, line 90: Minor: It just struck me that our ctx already decorates the log line with the provisioner name `controller.allocation.provisioner/default`, so you could get away with leaving it out here.

pkg/controllers/allocation/scheduling/scheduler.go
```diff
@@ -51,7 +61,49 @@ type Schedule struct {
...
+			"Failed to record scheduling duration metric [%s=%s, %s=%s, duration=%f]: error=%w",
```
ellistarn, line 91: %w is intended for error chaining. When logging an error, you should use something like fmt.Errorf("%s", error.Error())

pkg/controllers/allocation/scheduling/scheduler.go
```diff
@@ -51,7 +61,49 @@ type Schedule struct {
...
+	startTime := time.Now()
+	schedules, scheduleErr := s.solve(ctx, provisioner, pods)
+	durationSeconds := time.Since(startTime).Seconds()
+
+	result := "success"
+	if scheduleErr != nil {
+		result = "error"
+	}
+
+	provisionerName := provisioner.ObjectMeta.Name
```
ellistarn, line 84: Not to overly focus on this, but I just realized that you could print out the labels in the log line to avoid the case you were talking about. It's a map, so it should print [key1: value, key2: value2]
```
labels = prometheus.Labels{...}
logging(...)Warnf("%s", labels)
```

ellistarn (approved):

PR #687, opened 2021-09-17
add allocation binpacking duration metric

cjerad:
**1. Issue, if available:**

Add a histogram metric to record the duration of binpacking operations.

- [x] No

### DIFF
--- a/pkg/controllers/allocation/binpacking/packer.go
+++ b/pkg/controllers/allocation/binpacking/packer.go
@@ -82,13 +82,10 @@ type Packing struct {
 // https://en.wikipedia.org/wiki/Bin_packing_problem#First_Fit_Decreasing_(FFD)
 func (p *packer) Pack(ctx context.Context, schedule *scheduling.Schedule, instances []cloudprovider.InstanceType) []*Packing {
 	startTime := time.Now()
-	packings := p.pack(ctx, schedule, instances)
-	packTimeHistogram.Observe(time.Since(startTime).Seconds())
+	defer func() {
+		packTimeHistogram.Observe(time.Since(startTime).Seconds())
+	}()
 
-	return packings
-}
-
-func (p *packer) pack(ctx context.Context, schedule *scheduling.Schedule, instances []cloudprovider.InstanceType) []*Packing {
 	// Sort pods in decreasing order by the amount of CPU requested, if
 	// CPU requested is equal compare memory requested.
 	sort.Sort(sort.Reverse(ByResourcesRequested{SortablePods: schedule.Pods}))
--- a/pkg/controllers/allocation/binpacking/packer.go
+++ b/pkg/controllers/allocation/binpacking/packer.go
@@ -18,22 +18,40 @@ import (
 	"context"
 	"math"
 	"sort"
+	"time"
 
 	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha3"
 	"github.com/awslabs/karpenter/pkg/cloudprovider"
 	"github.com/awslabs/karpenter/pkg/controllers/allocation/scheduling"
+	"github.com/awslabs/karpenter/pkg/metrics"
 	"github.com/awslabs/karpenter/pkg/utils/apiobject"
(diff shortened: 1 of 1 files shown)


pkg/controllers/allocation/binpacking/packer.go
```diff
@@ -63,6 +81,14 @@ type Packing struct {
...
 // https://en.wikipedia.org/wiki/Bin_packing_problem#First_Fit_Decreasing_(FFD)
 func (p *packer) Pack(ctx context.Context, schedule *scheduling.Schedule, instances []cloudprovider.InstanceType) []*Packing {
+	startTime := time.Now()
+	packings := p.pack(ctx, schedule, instances)
+	packTimeHistogram.Observe(time.Since(startTime).Seconds())
+
+	return packings
+}
+
+func (p *packer) pack(ctx context.Context, schedule *scheduling.Schedule, instances []cloudprovider.InstanceType) []*Packing {
```
bwagner5, line 91: To remove the need for a private function that just calls the public, you could place this at the top of original and still make minimal code additions to the core packing logic.  

Pack:
```
startTime := time.Now() 
defer packTimeHistogram.Observe(time.Since(startTime).Seconds())
```

The `defer` is advantageous on the Observe call anyways since you'll get observed durations even if a panic or some runtime errors wreak havoc.

cjerad: > To remove the need for a private function that just calls the public [...]

It's the other way around: the public method calls the private method.  The wrapper allows separating the metric logic from the binpacking logic.  Is there an advantage to breaking the separation of concerns?

bwagner5: yeah, sorry mistyped: Public calling Private. 

Since the metric logic is tiny (two lines), seems a bit heavy to do the function proxying. The metric logic can be placed at the top with the defer to keep it together to separate it a bit for the reader.  The reason to keep the private without metrics is if we have a use-case (maybe in testing) to not emit metrics and use just the `pack` func. But that doesn't seem like a great thing to do anyways. I would guess it would be configurable elsewhere to make the Observe func a noop, if we don't want metrics.

pkg/controllers/allocation/binpacking/packer.go
```diff
@@ -63,6 +81,11 @@ type Packing struct {
 // It follows the First Fit Decreasing bin packing technique, reference-
 // https://en.wikipedia.org/wiki/Bin_packing_problem#First_Fit_Decreasing_(FFD)
 func (p *packer) Pack(ctx context.Context, schedule *scheduling.Schedule, instances []cloudprovider.InstanceType) []*Packing {
+	startTime := time.Now()
+	defer func() {
+		packTimeHistogram.Observe(time.Since(startTime).Seconds())
```
ellistarn, line 86: Cool pattern!

ellistarn (approved):

pkg/controllers/allocation/binpacking/packer.go
```diff
@@ -63,6 +81,11 @@ type Packing struct {
 // It follows the First Fit Decreasing bin packing technique, reference-
 // https://en.wikipedia.org/wiki/Bin_packing_problem#First_Fit_Decreasing_(FFD)
 func (p *packer) Pack(ctx context.Context, schedule *scheduling.Schedule, instances []cloudprovider.InstanceType) []*Packing {
+	startTime := time.Now()
+	defer func() {
```
bwagner5, line 85: is there a reason to wrap this in a func rather than just doing: 

```
startTime := time.Now()
defer packTimeHistogram.Observe(time.Since(startTime).Seconds())
```

cjerad: According to [Defer, Panic, and Recover](https://go.dev/blog/defer-panic-and-recover) from the Go blog: 
> A deferred function’s arguments are evaluated when the defer statement is evaluated.

So the proposed code will not record the elapsed time because only the `.Observe()` call is defered; `time.Since(startTime).Seconds()` will be evaluated immediately.

PR #701, opened 2021-09-23
[WIP] add node metrics controller

cjerad:
**1. Issue, if available:**

Add a controller that watches node events and updates counters across multiple dimensions, e.g. provisioner, readiness, and instance type

- [x] No


pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,162 @@
...
+	currLabels := getLabels(node)
+	pastLabels, isKnown := prometheusLabelsFor[name]
+	switch {
+	case !isKnown && node != nil:
+		handleCreatedNode(ctx, name, currLabels)
+	case isKnown && node == nil:
+		handleDeletedNode(ctx, name, pastLabels)
+	case isKnown && node != nil:
+		handleUpdatedNode(ctx, name, currLabels, pastLabels)
+	default: // An unknow node was deleted.
```
bwagner5, line 91: typo: unknow

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,162 @@
...
+	if node != nil {
+		labels[metrics.ProvisionerLabel] = node.Labels["karpenter.sh/provisioner-name"]
+		labels[metricArchLabel] = node.Labels["kubernetes.io/arch"]
+		labels[metricInstanceTypeLabel] = node.Labels["node.kubernetes.io/instance-type"]
+		labels[metricTopologyRegionLabel] = node.Labels["topology.kubernetes.io/region"]
+		labels[metricTopologyZoneLabel] = node.Labels["topology.kubernetes.io/zone"]
+		labels[metricOsLabel] = node.Labels["kubernetes.io/os"]
+
+		for _, c := range node.Status.Conditions {
+			labelName, found := conditionTypeToMetricLabel[c.Type]
```
bwagner5, line 138: optional: could condense to one line and tighten the scope:

```
if label_name, found := conditionTypeToMetricLabel[c.Type]; found {
  labels[labelName] = strings.ToLower(string(c.Status))
}
```

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,162 @@
...
+func incrementNodeCount(labels prometheus.Labels) error {
+	return updateNodeCount(labels, prometheus.Gauge.Inc)
+}
+
+func decrementNodeCount(labels prometheus.Labels) error {
+	return updateNodeCount(labels, prometheus.Gauge.Dec)
+}
+
+func updateNodeCount(labels prometheus.Labels, update func(prometheus.Gauge)) error {
+	nodeCount, err := nodeCountGaugeVec.GetMetricWith(labels)
```
bwagner5, line 157: optional:  Seems kind of backwards from the normal go idioms of failing early if err and happy case is not indented.

cmd/controller/main.go
```diff
@@ -89,6 +90,7 @@ func main() {
 		allocation.NewController(manager.GetClient(), clientSet.CoreV1(), cloudProvider),
 		termination.NewController(ctx, manager.GetClient(), clientSet.CoreV1(), cloudProvider),
 		node.NewController(manager.GetClient()),
+		&nodemetrics.Controller{KubeClient: manager.GetClient()},
```
bwagner5, line 93: optional: Might want to make a constructor just to stay uniform with the others controllers.

bwagner5 (approved):
Looks really good, and very readable!

ellistarn:
Would love to review this. Can you give me until Friday?

cjerad:
> Would love to review this. Can you give me until Friday?

Sure

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func getLabels(node *v1.Node) prometheus.Labels {
+	labels := make(prometheus.Labels)
+	if node == nil {
+		return labels
+	}
+
+	labels[metrics.ProvisionerLabel] = node.Labels["karpenter.sh/provisioner-name"]
+	labels[metricArchLabel] = node.Labels["kubernetes.io/arch"]
```
bwagner5, line 134: you could switch out these hardcoded labels to these: https://github.com/kubernetes/api/blob/master/core/v1/well_known_labels.go

cjerad: Nice 👍

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,79 @@
...
+	return &Controller{KubeClient: kubeClient}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("NodeMetrics"))
+
+	node := &v1.Node{}
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, node); err != nil {
+		if errors.IsNotFound(err) {
+			// Pass `nil` to UpdateCount rather than a "zeroed" node.
```
ellistarn, line 48: The typical reconciler pattern is to return reconcile.Result{} (terminate), since it only occurs in a race condition where the object is deleted after the watch event arrives but before the loop is executed. 

I can potentially see some value in tracking this case, but doing so with `node = nil` seems awkward and confusing to folks who might be looking at prometheus.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,79 @@
...
+			node = nil
+		} else {
+			return reconcile.Result{}, err
+		}
+	}
+	counter.UpdateCount(
+		logging.WithLogger(ctx, logging.FromContext(ctx).Named("Count")),
```
ellistarn, line 55: Why redecorate the logger with a new name? Simpler to just `counter.Update(ctx, node)`?

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (c *Controller) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.
+		NewControllerManagedBy(m).
+		Named("NodeMetrics").
+		For(&v1.Node{}).
+		WithOptions(
+			controller.Options{
+				RateLimiter: workqueue.NewMaxOfRateLimiter(
```
ellistarn, line 70: I know Karpenter defines this in other controllers, but I'd just leave this nil and rely on the default backoff retry settings. Not a bad idea to keep the concurrency though `controller.Options{MaxConcurrentReconciles: 4}`.

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,166 @@
...
+		prometheus.GaugeOpts{
+			Namespace: metrics.KarpenterNamespace,
+			Subsystem: "cluster",
+			Name:      "node_count",
+			Help:      "Count of cluster nodes. Broken out by topology and status.",
+		},
+		[]string{
+			metrics.ProvisionerLabel,
+			metricArchLabel,
+			metricConditionDiskPressureLabel,
```
ellistarn, line 54: Consider leaving out DiskPressure, MemoryPressure, and PIDPressure, since they're out of scope of something Karpenter can help with. I'd keep these metrics focused on Karpenter's domain.

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,166 @@
...
+	metricOsLabel                      = "os"
+)
+
+var (
+	nodeCountGaugeVec = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: metrics.KarpenterNamespace,
+			Subsystem: "cluster",
+			Name:      "node_count",
+			Help:      "Count of cluster nodes. Broken out by topology and status.",
```
ellistarn, line 49: Nit: `cluster nodes` is a bit redundant. Consider something like `Nodes by topology and status`

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,166 @@
...
+	metricTopologyRegionLabel          = "region"
+	metricTopologyZoneLabel            = "zone"
+	metricOsLabel                      = "os"
+)
+
+var (
+	nodeCountGaugeVec = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: metrics.KarpenterNamespace,
+			Subsystem: "cluster",
```
ellistarn, line 47: Thoughts on calling this capacity? Conceptually
- "Karpenter Capacity Metrics" = "How good of a job is Karpenter doing managing my capacity? 
- "Karpenter Controller Metrics" = "How healthy is the karpenter process itself?"

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,166 @@
...
+
+import (
+	"context"
+	"reflect"
+	"strings"
+
+	karpenterapi "github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha4"
+	"github.com/awslabs/karpenter/pkg/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+	coreapi "k8s.io/api/core/v1"
```
ellistarn, line 25: Kubernetes golang conventions recommend calling this `v1` or `corev1` if there's a conflict.

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,166 @@
...
+*/
+
+package counter
+
+import (
+	"context"
+	"reflect"
+	"strings"
+
+	karpenterapi "github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha4"
```
ellistarn, line 22: Import are recommended to be of the form `v1alpha4` or `provisioningv1alpha4` (walking up the chain, if there's a conflict). Typically you'd only alias an import if there was some sort of conflict. This give developers a common convention to avoid alias sprawl of the same concept. It does mean that package names (and heirarchy) is critical, but it's a good sign of project organization if this convention yields english readable names. FWIW, I think you've done excellently re: package naming in this PR.

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+func init() {
+	crmetrics.Registry.MustRegister(nodeCountGaugeVec)
+}
+
+// UpdateCount updates the emitted metric based on the node's current status relative to the
+// past status. If the data for `node` cannot be populated then `nil` should be passed as the
+// argument.
+func UpdateCount(ctx context.Context, name types.NamespacedName, node *coreapi.Node) {
```
ellistarn, line 82: Consider counter.Update() instead of counter.UpdateCount(), which stutters. You can also get the entire namespaced name from the node object `node.Name` (every kubernetes object has it). In this case, nodes are a cluster global resource -- namespace for nodes will never make sense.

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,166 @@
...
+	currLabels := getLabels(node)
+	pastLabels, isKnown := prometheusLabelsFor[name]
+	switch {
```
ellistarn, line 85: I see now why you're passing node as nil if not found. 

This is a bit tricky since there's no guarantee that your controller will catch the deleted event (events are not guaranteed to arrive -- it could be delivered five times or never). The controller pattern is best effort and employs a combination of intervals (e.g. resync interval) and watches.

It feels a bit dirty, but I might just statelessly recompute everything on some interval. Even at 10k node scale, this should be pretty quick. If you're unfamiliar with the client.Client, all objects are kept in a cache and synced with incoming events (a.k.a ListWatch), so it's cheap to call client.List(nodes).

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,166 @@
...
+func handleUpdatedNode(ctx context.Context, name types.NamespacedName, currLabels prometheus.Labels, pastLabels prometheus.Labels) {
+	// Only report node updates that affect tracked dimensions.
+	if reflect.DeepEqual(currLabels, pastLabels) {
+		return
+	}
+
+	prometheusLabelsFor[name] = currLabels
+
+	if err := decrementNodeCount(pastLabels); err != nil {
+		logging.FromContext(ctx).Warnf("Failed to decrement previous count for updated node [labels=%s]: error=%s", pastLabels, err.Error())
```
ellistarn, line 121: I'd prefer we continue to follow the convention of ErrorF instead of WarnF.

pkg/controllers/metrics/node/counter/counter.go
```diff
@@ -0,0 +1,166 @@
...
+		logging.FromContext(ctx).Warnf("Failed to increment current count for updated node [labels=%s]: error=%s", currLabels, err.Error())
+	}
+}
+
+func getLabels(node *coreapi.Node) prometheus.Labels {
+	labels := make(prometheus.Labels)
+	if node == nil {
+		return labels
+	}
+
```
ellistarn, line 133: consider just inlining all of this

labels := prometheus.Labels{
   key: value,
   ...
}

pkg/controllers/metrics/node/counter/counter.go
bwagner5, line 121: I was testing this code out and I think there's a false assumption in here with the node properties being set.  It could be that node properties like arch, os, instance type, etc are not set.  

I'm getting floods of this message:

```
karpenter-controller-9f4878f7b-8rt9b manager 2021-09-27T21:54:04.263Z	WARN	controller.NodeMetrics.Count	Failed to decrement previous count for updated node [labels=map[arch: instancetype: os: provisioner:default region: zone:]]: error=inconsistent label cardinality: expected 10 label values but got 6 in prometheus.Labels{"arch":"", "instancetype":"", "os":"", "provisioner":"default", "region":"", "zone":""}	{"commit": "5655c94"}
```

cjerad:
There are some gaps in this approach.  Closing this PR and will open a new PR with the revised approach.

PR #721, opened 2021-10-01
ignore static pods owned by a node in node emptiness check

bwagner5:
**1. Issue, if available:**
N/A

I ran into a few problems running Karpenter on kOps because kube-proxy is deployed as a static pod. Static pods have owner references to the node resource they are running on. This change checks if a pod is a static pod when performing karpenter's node emptiness check. Static pods shouldn't hold up node eviction since they are effectively an agent similar to a daemonset which karpenter already ignores.

- [x] No


ellistarn (dismissed):
Should it be checked here, too? 
```func (f *Filter) isUnschedulable(p *v1.Pod) error {```

ellistarn (approved):

PR #734, opened 2021-10-05
add node metrics controller

cjerad:
**1. Issue, if available:**

Add a controller that runs periodically, and in response to some provisioner events, and publishes different counts of nodes broken out by zone, instance-type, etc.

- [x] No

### DIFF
--- a/pkg/controllers/metrics/node/controller.go
+++ b/pkg/controllers/metrics/node/controller.go
@@ -58,7 +58,7 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		}
 
 		// The provisioner has been deleted. Reset all the associated counts to zero.
-		if err := publishNodeCountsForProvisioner(provisionerName, consumeNoNodes); err != nil {
+		if err := publishNodeCountsForProvisioner(provisionerName, consumeZeroNodes); err != nil {
 			// One or more metrics were not zeroed. Try again later.
 			return reconcile.Result{Requeue: true}, err
 		}
@@ -115,7 +115,7 @@ func (c *Controller) consumeNodesWith(ctx context.Context) consumeNodesWithFunc 
 	}
 }
 
-// consumeNoNodes calls `consume` with an empty slice and returns any resulting error.
-func consumeNoNodes(_ client.MatchingLabels, consume nodeListConsumerFunc) error {
+// consumeZeroNodes calls `consume` with an empty slice and returns any resulting error.
+func consumeZeroNodes(_ client.MatchingLabels, consume nodeListConsumerFunc) error {
 	return consume([]v1.Node{})
 }
--- a/pkg/controllers/metrics/node/controller.go
+++ b/pkg/controllers/metrics/node/controller.go
@@ -1,0 +1,122 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package node
+
+import (
+	"context"
+	"time"
+
+	karpenterv1 "github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha4"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"knative.dev/pkg/logging"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	controllerName = "NodeMetrics"
+
+	requeueInterval = 10 * time.Second
+)
+
+type Controller struct {
+	KubeClient client.Client
+}
+
+func NewController(kubeClient client.Client) *Controller {
+	return &Controller{KubeClient: kubeClient}
+}
+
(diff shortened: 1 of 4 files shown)


pkg/controllers/metrics/node/counter.go
```diff
@@ -0,0 +1,243 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package node
+
+import (
+	"strings"
+
+	karpenterv1 "github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha4"
```
JacobGabrielson, line 20: Could you rename this `provisioning`? (Or just use `v1alpha4` which the rest of the codebase does)?

pkg/controllers/metrics/node/counter.go
```diff
@@ -0,0 +1,243 @@
...
+	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const (
+	metricNamespace = metrics.KarpenterNamespace
+	metricSubsystem = "capacity"
+
+	metricLabelArch         = "arch"
+	metricLabelInstanceType = "instancetype"
+	metricLabelOs           = "os"
```
JacobGabrielson, line 35: Can you rename to `metricLabelOS`? (Per our coding conventions). Thanks!

pkg/controllers/metrics/node/counter.go
```diff
@@ -0,0 +1,243 @@
...
+	metricLabelProvisioner  = metrics.ProvisionerLabel
+	metricLabelZone         = "zone"
+
+	nodeLabelArch         = corev1.LabelArchStable
+	nodeLabelInstanceType = corev1.LabelInstanceTypeStable
+	nodeLabelOs           = corev1.LabelOSStable
```
JacobGabrielson, line 41: Can you rename to `nodeLabelOS`?

pkg/controllers/metrics/node/counter.go
```diff
@@ -0,0 +1,243 @@
...
+	crmetrics.Registry.MustRegister(readyNodeCountByOsProvisionerZone)
+}
+
+func publishNodeCountsForProvisioner(provisioner string, consumeNodesWith consumeNodesWithFunc) error {
+	archValues := knownValuesForNodeLabels[nodeLabelArch]
+	instanceTypeValues := knownValuesForNodeLabels[nodeLabelInstanceType]
+	osValues := knownValuesForNodeLabels[nodeLabelOs]
+	zoneValues := knownValuesForNodeLabels[nodeLabelZone]
+
+	errors := make([]error, 0, len(archValues)*len(instanceTypeValues)*len(osValues)*len(zoneValues))
```
JacobGabrielson, line 139: I'm wondering if this optimization is necessary? Do we anticipate this slice to get very large?

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,122 @@
...
+		nodes := corev1.NodeList{}
+		if err := c.KubeClient.List(ctx, &nodes, nodeLabels); err != nil {
+			return err
+		}
+		return consume(nodes.Items)
+	}
+}
+
+// consumeNoNodes calls `consume` with an empty slice and returns any resulting error.
+func consumeNoNodes(_ client.MatchingLabels, consume nodeListConsumerFunc) error {
```
JacobGabrielson, line 120: I'd suggest a name like `resetNodeCounts` or something like that? (I.e., more of an idea of what it's used for, rather than how it's implemented).

pkg/controllers/metrics/node/counter.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func publishCount(gaugeVec *prometheus.GaugeVec, labels prometheus.Labels, count int) (err error) {
+	var gauge prometheus.Gauge
+	gauge, err = gaugeVec.GetMetricWith(labels)
+	if err != nil {
+		return
+	}
+	gauge.Set(float64(count))
+	return
```
JacobGabrielson, line 242: This can be simplified to:

```go
	gauge, err := gaugeVec.GetMetricWith(labels)
	if err == nil {
	    gauge.Set(float64(count))
        }
	return err
```

Also, the return value would go from `(err error)` -> `error`.

pkg/controllers/metrics/node/counter.go
cjerad: > I'm wondering if this optimization is necessary?

In a test cluster, the total length was >1500 (`len(instanceTypeValues) > 500` and `len(zoneValues) == 3`).

> Do we anticipate this slice to get very large?

The current implementation appends an `error` or `nil` for each metric published.

pkg/controllers/metrics/node/controller.go
cjerad: The function itself does not reset the counts.  How about `consumeZeroNodes`?

JacobGabrielson: Sounds good, thanks

pkg/controllers/metrics/node/counter.go
```diff
@@ -0,0 +1,241 @@
...
+	// 1. Publish the count of all nodes associated with `provisioner`.
+	nodeLabels := client.MatchingLabels{nodeLabelProvisioner: provisioner}
+	errors = append(errors, consumeNodesWith(nodeLabels, func(nodes []v1.Node) error {
+		return publishCount(nodeCountByProvisioner, metricLabelsFrom(nodeLabels), len(nodes))
+	}))
+
+	// 2. Publish the count of all nodes associated with `provisioner`, in `zone`, and reported as "ready".
+	// 3. Publish the count of all nodes with `arch`, associated with `provisioner`, in `zone`, and reported as "ready".
+	// 4. Publish the count of all nodes with `instanceType`, associated with `provisioner`, in `zone`, and reported as "ready"
+	// 5. Publish the count of all nodes with `os`, associated with `provisioner`, in `zone`, and reported as "ready".
```
JacobGabrielson, line 150: Honestly I don't think these comments are really necessary, they pretty much just restate the code - you could just say Publish counts of nodes along different dimensions, or something to that effect (otherwise, every time a new count is added, the comment will have to be updated).

pkg/controllers/metrics/node/counter.go
```diff
@@ -0,0 +1,241 @@
...
+				nodeLabelProvisioner: provisioner,
+				nodeLabelZone:        zone,
+			}
+			errors = append(errors, consumeNodesWith(nodeLabels, filterReadyNodes(func(readyNodes []v1.Node) error {
+				return publishCount(readyNodeCountByOsProvisionerZone, metricLabelsFrom(nodeLabels), len(readyNodes))
+			})))
+		}
+	}
+
+	// Combine will filter out `nil` values; if no errors remain then it will return `nil`.
```
JacobGabrielson, line 194: nit: I'd just remove this comment (if people are confused they can look up what `multierr.Combine` does, but it's a bit overkill to document it here.)

JacobGabrielson (approved):
I have a few minor optional comments, lgtm

Thanks!!!

PR #740, opened 2021-10-06
[AWS] Batch CreateFleet calls when possible

bwagner5:
**1. Issue, if available:**
https://github.com/awslabs/karpenter/issues/732

The AWS CloudProvider uses the CreateFleet API to provision EC2 Instances as k8s nodes. Before this PR, we were requesting 1 instance per CreateFleet call because we are passing in up to 20 possible instance types that EC2 Fleet could use to fulfill the request.  

This PR batches packings that have the same constraints and same instance type options so that we can more efficiently make remote calls to EC2 Fleet and stay within API rate limits.

- [x] No


pkg/cloudprovider/aws/instance.go
```diff
@@ -51,34 +51,40 @@ type InstanceProvider struct {
...
+			aws.StringValue(instance.PrivateDnsName),
+			aws.StringValue(instance.InstanceType),
+			aws.StringValue(instance.Placement.AvailabilityZone),
+			getCapacityType(instance),
+		)
+
+		// Convert Instance to Node
+		node, err := p.instanceToNode(ctx, instance, instanceTypes)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
bwagner5: ah you are right 👍

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -119,23 +120,29 @@ func withUserAgent(sess *session.Session) *session.Session {
...
 	if err != nil {
 		return err
 	}
-	// 3. Create instance
-	node, err := c.instanceProvider.Create(ctx, vendorConstraints, instanceTypes)
+	// Create instance
+	nodes, err := c.instanceProvider.Create(ctx, vendorConstraints, instanceTypes, quantity)
 	if err != nil {
-		return fmt.Errorf("launching instance, %w", err)
+		return fmt.Errorf("launching %d instance(s), %w", quantity, err)
```
JacobGabrielson, line 137: can't you now have the case where `nodes` has some nodes in it _and_ you have an `err` that isn't `nil`? (if so I don't think you want to just return here without doing something with the nodes you did get?)

pkg/controllers/allocation/binpacking/packer.go
```diff
@@ -89,20 +91,32 @@ func (p *packer) Pack(ctx context.Context, schedule *scheduling.Schedule, instan
...
+	packs := map[uint64]*Packing{}
 	var packings []*Packing
 	var packing *Packing
 	remainingPods := schedule.Pods
 	for len(remainingPods) > 0 {
 		packables := PackablesFor(ctx, instances, schedule)
 		packing, remainingPods = p.packWithLargestPod(ctx, schedule.Constraints, remainingPods, packables)
 		// checked all instance types and found no packing option
-		if len(packing.Pods) == 0 {
+		if len(flatten(packing.Pods...)) == 0 {
```
JacobGabrielson, line 102: this might be over thinking things, but seems like flatten might be getting called quite a bit here (up to three times per iteration thru the loop?) Seems like it's only needed to get the length - maybe just use a function that walks thru and counts the length instead?

pkg/controllers/allocation/binpacking/packer.go
```diff
@@ -199,6 +213,14 @@ func instanceTypeNames(instanceTypes []cloudprovider.InstanceType) []string {
 	return names
 }
 
+func flatten(pods ...[]*v1.Pod) []*v1.Pod {
```
JacobGabrielson, line 216: It seems like this is mostly (only?) used to get a count, the actually flattened list isn't needed? Maybe just create a function that just figures out the count (and doesn't allocate?)

JacobGabrielson (approved):

pkg/cloudprovider/fake/cloudprovider.go
```diff
@@ -32,43 +32,45 @@ import (
...
-	go func() {
-		err <- bind(&v1.Node{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: name,
-				Labels: map[string]string{
-					v1.LabelTopologyZone:       zone,
-					v1.LabelInstanceTypeStable: instance.Name(),
+	for i := 0; i < quantity; i++ {
+		name := strings.ToLower(randomdata.SillyName())
+		// Pick first instance type option
```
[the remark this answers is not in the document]
bwagner5: I can remove the comments where it says pick first {instance, zone}. I'm not entirely sure the reason on picking 1 since this was existing code that I just wrapped with the for-loop.  I don't think doing anything more complicated would make a difference to tests that depend on this code. If the real provider were used, it would be semi-random (not really, but allocation strategy would pick one on the Fleet backend).

PR #744, opened 2021-10-13
add pod metrics controller

cjerad:
**1. Issue, if available:**

Add a controller that updates pod count metrics across multiple dimensions, e.g. provisioner, phase, and zone.

- [x] No


pkg/controllers/metrics/pod/counter.go
```diff
@@ -0,0 +1,131 @@
...
+			metricLabelProvisioner: provisioner,
+		}
+		errors = append(errors, publishCount(podCountByPhaseProvisioner, metricLabels, countByPhase[phase]))
+	}
+	for _, zone := range zoneValues {
+		metricLabels := prometheus.Labels{
+			metricLabelProvisioner: provisioner,
+			metricLabelZone:        zone,
+		}
+		errors = append(errors, publishCount(runningPodCountByProvisionerZone, metricLabels, countByZone[zone]))
```
ellistarn, line 119: check out multierr.Append instead of managing the slice of errors yourself.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,146 @@
...
+		zone := node.Labels[v1.LabelTopologyZone]
+		if zone == "" {
+			return map[string][]v1.Pod{}, fmt.Errorf(
+				"Node %q provisioned by %q is missing label %s",
+				node.Name, provisioner.Name, v1.LabelTopologyZone,
+			)
+		}
+
+		podList := v1.PodList{}
+		withNodeName := client.MatchingFields{"spec.nodeName": node.Name}
```
ellistarn, line 137: any reason to not inline this?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,146 @@
...
+		return map[string][]v1.Pod{}, err
+	}
+
+	// 2. Get all the pods scheduled to each node and append them to the slice
+	//    associated with the node's zone.
+	for _, node := range nodeList.Items {
+		zone := node.Labels[v1.LabelTopologyZone]
+		if zone == "" {
+			return map[string][]v1.Pod{}, fmt.Errorf(
+				"Node %q provisioned by %q is missing label %s",
```
ellistarn, line 131: TIL %q, cool trick. We currently have a pattern of simply naming the logger after the resource being reconciled, so you can deduplicate including the provisioner's name here.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// 3. Schedule the next run.
+	return reconcile.Result{RequeueAfter: requeueInterval}, nil
+}
+
+func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
+	return controllerruntime.
+		NewControllerManagedBy(m).
+		Named(controllerName).
+		For(&v1alpha4.Provisioner{}, builder.WithPredicates(
```
ellistarn, line 97: I'd drop all of these predicates and just use the default. I don't see the harm in running this on an update.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,146 @@
...
+			predicate.Funcs{
+				CreateFunc:  func(_ event.CreateEvent) bool { return true },
+				DeleteFunc:  func(_ event.DeleteEvent) bool { return true },
+				UpdateFunc:  func(_ event.UpdateEvent) bool { return false },
+				GenericFunc: func(_ event.GenericEvent) bool { return false },
+			},
+		)).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: 1,
```
ellistarn, line 106: I'd keep this at ~10, since there may be a decent number of provisioners in some clusters

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,146 @@
...
+		// Since the provisioner is gone, do not requeue.
+		return reconcile.Result{}, nil
+	}
+
+	// 2. Update pod counts associated with the provisioner.
+	podsByZone, err := c.podsByZone(ctx, &provisioner)
+	if err != nil {
+		return reconcile.Result{Requeue: true}, err
+	}
+	provisionablePods, err := c.Filter.GetProvisionablePods(ctx, &provisioner)
```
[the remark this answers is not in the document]
ellistarn: +1. I think we shouldn't try to share this library, since it does fundamentally different things (and isn't very complicated)

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return reconcile.Result{Requeue: true}, err
+	}
+	if err := publishPodCounts(provisioner.Name, podsByZone, provisionablePods); err != nil {
+		// An updated value for one or more metrics was not published. Try again later.
```
ellistarn, line 85: nit: unnecessary comment

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,146 @@
...
+		return reconcile.Result{Requeue: true}, err
```
ellistarn, line 86: You don't actually need to signal requeue, simply `reconcile.Result{}, err` will do the trick. If I were a controller-runtime author, I'd probably make this a pointer so you could return nil instead of an empty struct. It's a go anti-pattern to return both an error and a concrete value -- the convention is that the value may not have been safely initialized if there is an error.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,146 @@
...
+
+func NewController(kubeClient client.Client) *Controller {
+	return &Controller{
+		Filter:     &podutils.Filter{KubeClient: kubeClient},
+		KubeClient: kubeClient,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named(controllerName))
```
ellistarn, line 55: Consider annotating with req.Name, as well.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,146 @@
...
+			// One or more metrics were not zeroed. Try again later.
+			return reconcile.Result{Requeue: true}, err
+		}
+
+		// Since the provisioner is gone, do not requeue.
+		return reconcile.Result{}, nil
+	}
+
+	// 2. Update pod counts associated with the provisioner.
+	podsByZone, err := c.podsByZone(ctx, &provisioner)
```
ellistarn, line 76: Curious what the design options are here:
- pods w/ zone data
- pods w/o zone data
- other instance attributes (arch/os/etc)
I wonder if we should follow https://en.wikipedia.org/wiki/KISS_principle and just include phase.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// 1. Has the provisioner been deleted?
+	provisioner := v1alpha4.Provisioner{}
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, &provisioner); err != nil {
+		if !errors.IsNotFound(err) {
+			// Unable to determine existence of the provisioner, try again later.
+			return reconcile.Result{Requeue: true}, err
+		}
+
+		// The provisioner has been deleted. Reset all the associated counts to zero.
```
ellistarn, line 65: Isn't this the same issue we ran into w/ the previous PR? Thoughts on just letting the data expire rather than trying to zero it out?

pkg/controllers/metrics/pod/counter.go
```diff
@@ -0,0 +1,131 @@
...
+
+	metricLabelPhase       = "phase"
+	metricLabelProvisioner = metrics.ProvisionerLabel
+	metricLabelZone        = "zone"
+
+	nodeLabelZone = v1.LabelTopologyZone
+)
+
+var (
+	knownValuesForNodeLabels = v1alpha4.WellKnownLabels
```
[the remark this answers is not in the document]
cjerad: To improve clarity and avoid package name "noise" -- in this case it doesn't provide helpful context.

```go
zoneValues := knownValuesForNodeLabels[nodeLabelZone]
// vs
zoneValues := v1alpha4.WellKnownLabels[nodeLabelZone]
```

pkg/controllers/metrics/pod/controller.go
ellistarn: If we did want some sort of deletion assurance, we could use a finalizer. Maybe overkill though.

pkg/controllers/metrics/pod/counter.go
ellistarn: Hopefully they wouldn't all error. The typical pattern I've seen used across the k8s space (and in this project) is:
```
if err := publicCount(...); err != nil {
  errs = multierr.Append(errs, err)
}
```

pkg/controllers/metrics/constants.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"time"
+
+	"github.com/awslabs/karpenter/pkg/apis/provisioning/v1alpha4"
+	"github.com/awslabs/karpenter/pkg/metrics"
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
```
bwagner5, line 25: can we just put these constants in common.go ?

pkg/controllers/metrics/common.go
```diff
@@ -0,0 +1,25 @@
...
+limitations under the License.
+*/
+
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+func publishCount(gaugeVec *prometheus.GaugeVec, labels prometheus.Labels, count int) error {
+	gauge, err := gaugeVec.GetMetricWith(labels)
+	if err == nil {
```
bwagner5, line 21: Generally in go, the happy-path is left indented. So the err would be returned in a branch statement rather than at the end of the func.  I'm guessing you did this to save the extra `return nil` but I think it's better to stick with the convention here.

pkg/controllers/metrics/controller.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (c *Controller) updatePodCounts(ctx context.Context, provisioner *v1alpha4.Provisioner) error {
+	podsByZone, err := c.podsByZone(ctx, provisioner)
+	if err != nil {
+		return err
+	}
+
+	provisionablePodList := &v1.PodList{}
+	if err := c.KubeClient.List(ctx, provisionablePodList, client.MatchingFields{"spec.nodeName": ""}); err != nil {
+		return nil
```
bwagner5, line 135: do we not care about this list err?

pkg/controllers/metrics/constants.go
```diff
@@ -0,0 +1,52 @@
...
+	controllerName  = "Metrics"
+	requeueInterval = 10 * time.Second
```
ellistarn, line 27: If this is only used once, I'd just leave it inline.

pkg/controllers/metrics/constants.go
```diff
@@ -0,0 +1,52 @@
...
+
+	metricNamespace         = metrics.KarpenterNamespace
+	metricSubsystemCapacity = "capacity"
```
ellistarn, line 30: I'd call this variable `subsystem` to avoid stuttering both the package name and the constant value.

pkg/controllers/metrics/controller.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func NewController(kubeClient client.Client) *Controller {
+	return &Controller{
+		KubeClient: kubeClient,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	loggerName := fmt.Sprintf("%s.provisioner/%s", strings.ToLower(controllerName), req.Name)
```
ellistarn, line 46: consider inlining to avoid creation of an extra variable/concept/line.

pkg/controllers/metrics/controller.go
```diff
@@ -0,0 +1,173 @@
...
+			return reconcile.Result{}, err
+		}
+
+		// The provisioner has been deleted. Reset all associated counts to zero.
+		if err := c.zeroCounts(provisioner); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		// Since the provisioner is gone, do not requeue.
+		return reconcile.Result{}, nil
```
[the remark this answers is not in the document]
ellistarn: Further,

```
return reconcile.Result{}, multierr.Combine(
  c.zeroNodeCounts(provisioner),
  c.zeroPodCounts(provisioner),
)
```

pkg/controllers/metrics/pods.go
```diff
@@ -0,0 +1,109 @@
...
+var (
+	phaseValues = []v1.PodPhase{
+		v1.PodFailed,
+		v1.PodPending,
+		v1.PodRunning,
+		v1.PodSucceeded,
+		v1.PodUnknown,
+	}
+
+	provisionablePodCount = prometheus.NewGauge(
```
ellistarn, line 35: thoughts on calling this unscheduled? The check is making sure `spec.nodeName =""`, but that doesn't guarantee that it's provisonable.

pkg/controllers/metrics/pods.go
```diff
@@ -0,0 +1,72 @@
...
+
+package metrics
+
+import (
+	"strings"
+
+	"github.com/awslabs/karpenter/pkg/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/multierr"
+	v1 "k8s.io/api/core/v1"
```
bwagner5, line 23: unneeded alias

pkg/controllers/metrics/nodes.go
```diff
@@ -174,7 +152,7 @@ func publishNodeCountsForProvisioner(provisioner string, consumeNodesWith consum
 			})))
 		}
 
-		for _, os := range osValues {
+		for os := range osValues {
```
bwagner5, line 155: https://github.com/awslabs/karpenter/pull/780 removes most things related to OS since we currently only support linux. You can remove OS metrics here for now and we'll defer until we implement multiple OS support.

https://github.com/awslabs/karpenter/pull/780/files#diff-1a039ebbb2e08a2c87ccc7228041496010992b28a47f2e7e13a9840eee5fc0f4L59

bwagner5 (changes requested):
Looks good, just one comment to remove OS metrics for now.

pkg/controllers/metrics/pods.go
cjerad: `goimports` re-adds the alias if I try to remove it.  Found this [issue](https://github.com/golang/go/issues/29041) pertaining to `goimports` and packages named "v1".

bwagner5 (approved):
/lgtm

pkg/controllers/metrics/common.go
```diff
@@ -0,0 +1,52 @@
...
+	nodeLabelArch         = v1.LabelArchStable
+	nodeLabelInstanceType = v1.LabelInstanceTypeStable
+	nodeLabelZone         = v1.LabelTopologyZone
+
+	nodeConditionTypeReady = v1.NodeReady
+)
+
+var nodeLabelProvisioner = v1alpha5.ProvisionerNameLabelKey
+
+func publishCount(gaugeVec *prometheus.GaugeVec, labels prometheus.Labels, count int) error {
```
ellistarn, line 45: Given that the labels are deterministic, we could use https://github.com/prometheus/client_golang/blob/v1.11.0/prometheus/counter.go#L256

```gaugeVec.With(labels).Set(float64(count))```
which would collapse this helper.

ellistarn (approved):

PR #746, opened 2021-10-15
golangci added to verify target, fixes

JacobGabrielson:
**1. Issue, if available:**

N/A

Per @ellistarn add golangci config; fix errors found by same

- [x] No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ verify: ## Verify code. Includes dependencies, linting, formatting, etc
 	go mod download
 	go vet ./...
 	go fmt ./...
-	golangci-lint run --timeout 5m
+	golangci-lint run
 
 licenses: ## Verifies dependency licenses and requires GITHUB_TOKEN to be set
 	go build $(GOFLAGS) -o karpenter cmd/controller/main.go


pkg/cloudprovider/aws/errors.go
```diff
@@ -0,0 +1,39 @@
...
+	notFoundErrorCodes = []string{
+		"InvalidInstanceID.NotFound",
+		"InvalidLaunchTemplateName.NotFoundException",
+	}
+)
+
+// isNotFound returns true if the err is an AWS error (even if it's
+// wrapped) and is a known to mean "not found" (as opposed to a more
+// serious or unexpected error)
+func isNotFound(err error) bool {
```
bwagner5, line 33: nice!

pkg/controllers/allocation/filter.go
```diff
@@ -39,7 +39,8 @@ func (f *Filter) GetProvisionablePods(ctx context.Context, provisioner *v1alpha4
 	}
 	// 2. Filter pods that aren't provisionable
 	provisionable := []*v1.Pod{}
-	for _, p := range pods.Items {
+	for i := range pods.Items {
```
bwagner5, line 42: curious why we'd switch to using `i` here rather than just  ranging over the actual pod?  We do the range over the object ignoring the index in other loops.

pkg/controllers/node/emptiness.go
```diff
@@ -86,7 +86,8 @@ func (r *Emptiness) isEmpty(ctx context.Context, n *v1.Node) (bool, error) {
 	if err := r.kubeClient.List(ctx, pods, client.MatchingFields{"spec.nodeName": n.Name}); err != nil {
 		return false, fmt.Errorf("listing pods for node %s, %w", n.Name, err)
 	}
-	for _, p := range pods.Items {
+	for i := range pods.Items {
```
bwagner5, line 89: I'm seeing this pattern in a lot of places now. Is this a lint recommendation? I'm not sure I agree that this is better. Is this supposed to protect against the pointer location changing out from under the local loop var?

pkg/cloudprovider/aws/errors.go
ellistarn, line 33: consider making this public

pkg/controllers/allocation/filter.go
ellistarn: Probably to avoid variations of https://dev.to/kkentzo/the-golang-for-loop-gotcha-1n35

.golangci.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - errorlint
+    - gosec
+    - revive
+    - stylecheck
+    - tparallel
+    - unconvert
+    - unparam
+  disable:
+    - errcheck
+    - prealloc
```
ellistarn, line 29: This is the only field that diverges from kn IIUC

.golangci.yaml
```diff
@@ -0,0 +1,33 @@
+# See https://github.com/golangci/golangci-lint/blob/master/.golangci.example.yml
```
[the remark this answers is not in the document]
JacobGabrielson: I started with those, but they had chosen a few that weren't a good idea. For example, one of them exhorted one to use `make` to pre-allocate slices. Which, while sometimes a good idea, can be a pretty dangerous change to make without a lot of testing as the semantics do change subtly.

.golangci.yaml
JacobGabrielson: See my other comment about why `prealloc` isn't a good idea.

.golangci.yaml
```diff
@@ -0,0 +1,33 @@
...
+run:
+  tests: false
+  
+  timeout: 5m
```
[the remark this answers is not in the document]
JacobGabrielson: Will do!

pkg/controllers/allocation/filter.go
JacobGabrielson: @bwagner5 the difference is that in other loops, the address of the range object isn't being taken (and there's no closure going on either), so the linter isn't warning about those.

pkg/controllers/node/emptiness.go
JacobGabrielson: Only when it's closed over or its address is taken.

bwagner5 (approved):
/lgtm

ellistarn (approved):

PR #752, opened 2021-10-20
Draft: Adding CapacityTypes to instancetype in anticipation of ICE caching

eptiger:
**1. Issue, if available:**
https://github.com/awslabs/karpenter/issues/371

We need to incorporate CapacityTypes into the instance type struct so that the packer can account for this and re-pack if it sees that it can't create a packing with any available instance types for some capacity type.

- [X] No


pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -125,7 +125,7 @@ func (c *CloudProvider) create(ctx context.Context, constraints *v1alpha4.Constr
 	// Partial fulfillment will be logged
 	nodes, err := c.instanceProvider.Create(ctx, vendorConstraints, instanceTypes, quantity)
 	if err != nil {
-		return fmt.Errorf("launching %d instance(s), %w", quantity, err)
+		return fmt.Errorf("launching %d instance(s) with constraints %+v and instance types %+v, %w", quantity, vendorConstraints, instanceTypes, err)
```
ellistarn, line 128: We try to avoid explosively complex log lines, at least for commonly logged things. Now that I see this, maybe we should revise this idea that our Info/Debug should avoid data dumps like this, but it's fine for Error().

Check out the pretty.Concise() helper to jsonify these. I used to have pretty.Verbose() that included newlines, but this somehow got removed when I was on vacation :(. It's here if you need a reference: https://github.com/ellistarn/slang/blob/main/pkg/pretty/pretty.go

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -37,10 +37,16 @@ type InstanceType struct {
 func (i *InstanceType) Name() string {
 	return aws.StringValue(i.InstanceType)
 }
+
 func (i *InstanceType) Zones() []string {
 	return i.ZoneOptions
 }
 
+func (i *InstanceType) CapacityTypes() []string {
+	// values would be "spot" and/or "on-demand": https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceTypeInfo.html
```
ellistarn, line 46: This comment is a bit redundant IMO.

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -37,10 +37,16 @@ type InstanceType struct {
...
+	return aws.StringValueSlice(i.SupportedUsageClasses)
```
ellistarn, line 47: very cool!

pkg/cloudprovider/registry/aws.go
```diff
@@ -1,3 +1,4 @@
+//go:build aws
```
ellistarn, line 1: What's going on here?

pkg/cloudprovider/aws/instancetype.go
eptiger: As you can tell, I'm pretty big on documentation :) But where else is this documented? I feel like you'd have to trace this back if I didn't call it out. If you don't like this, can I put a link to the EC2 API docs up top, at least?

pkg/cloudprovider/aws/instance.go
```diff
@@ -42,6 +42,12 @@ type InstanceProvider struct {
 	subnetProvider         *SubnetProvider
 }
 
+type InstancePool struct {
+	InstanceType     *string
```
eptiger, line 46: Sorry this is a leak from the next PR. I'll remove this for now.

pkg/cloudprovider/types.go
```diff
@@ -54,7 +54,10 @@ type Options struct {
 // InstanceType describes the properties of a potential node
 type InstanceType interface {
 	Name() string
+	// these are the zones that support this instance type
```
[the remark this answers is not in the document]
eptiger: I'll remove the AWS cloud provider comment and keep this one since it's universal.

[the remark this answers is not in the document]
eptiger: Wait, this comment doesn't appear elsewhere AFAICT. Do you just think it's self-evident? The other things are actual attributes of this type, whereas this and capacity types are more about what it supports, which is why I specially called them out.

pkg/cloudprovider/aws/instancetype.go
ellistarn: >can I put a link to the EC2 API docs up top, at least?
SGTM

pkg/cloudprovider/registry/aws.go
eptiger: Looks like it's to do with Go 1.17.2 - it's getting added by `make dev`

API.md
```diff
@@ -347,5 +347,5 @@ its target, and indicates whether or not those conditions are met.</p>
 <hr/>
 <p><em>
 Generated with <code>gen-crd-api-reference-docs</code>
-on git commit <code>8dd6b65</code>.
+on git commit <code>3872084</code>.
```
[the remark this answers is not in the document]
eptiger: Nvm, I can't revert the v1alpha4 changes: `make dev` in Go 1.17.2 adds those lines in, not codegen.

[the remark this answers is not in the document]
bwagner5: ah ok, so it's go fmt probably for 1.17. So if you update all the required go versions to 1.17, then we can keep the new formatting. If not, downgrade your go to 1.16 and rerun `make dev`.

pkg/cloudprovider/registry/fake.go
```diff
@@ -1,3 +1,4 @@
+//go:build !aws
 // +build !aws
```
bwagner5, line 2: can you go ahead and remove the `// +build`

pkg/cloudprovider/registry/aws.go
```diff
@@ -1,3 +1,4 @@
...
 // +build aws
```
bwagner5, line 2: here too

PR #769, opened 2021-10-23
Adds support for tags in aws provider spec

suket22:
**1. Issue, if available:**
https://github.com/awslabs/karpenter/issues/642

* Adding tags to the provider which will be applied to all EC2 instances and launch templates being created by Karpenter.

- [x] Yes, issue opened: TODO
- [ ] No

Wasn't sure but I think this should be documented here? - https://github.com/awslabs/karpenter/blob/main/website/content/en/docs/cloud-providers/AWS.md ? I can edit this PR to include the doc changes if everything else looks good. 

**Testing done**
* Set up tags in the ProvisionerSpec and saw the tags get applied on both the EC2 instance and the launch templates. 
* Passed in an empty tag key and saw that validation work.

### DIFF
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -308,14 +308,10 @@ func (p *LaunchTemplateProvider) GetCABundle(ctx context.Context) (*string, erro
 
 func tagsFor(options *launchTemplateOptions) []*ec2.Tag {
 	var tags []*ec2.Tag
-	// Combine user defined and karpenter tags. If keys conflict, choose the latter.
 	for key, value := range functional.UnionStringMaps(
 		options.Tags,
-		map[string]string{
-			"Name": fmt.Sprintf("karpenter.sh/%s", options.ClusterName),
-			fmt.Sprintf(ClusterTagKeyFormat, options.ClusterName):   "owned",
-			fmt.Sprintf(KarpenterTagKeyFormat, options.ClusterName): "owned",
-		}) {
+		v1alpha1.ManagedTagsFor(options.ClusterName),
+	) {
 		tags = append(tags, &ec2.Tag{
 			Key:   aws.String(key),
 			Value: aws.String(value),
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/aws/apis/v1alpha1/aws_validation.go
```diff
@@ -72,6 +73,18 @@ func (a *AWS) validateSecurityGroups() (errs *apis.FieldError) {
 	return errs
 }
 
+func (a *AWS) validateTags() (errs *apis.FieldError) {
+	// Only checking for empty tag keys. Avoiding a check on number of tags (hard limit of 50) since that limit is shared by user
+	// defined and Karpenter tags, and the latter could change over time.
+	for tagKey, tagValue := range a.Tags {
+		if tagKey == "" {
+			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf(
+				"your tag with key : '' and value : '%s' is invalid because empty tag keys aren't supported", tagValue), "tags"))
```
ellistarn: style nit: I try to avoid referencing the user (e.g. "your"). I'd say something like `'' is not a valid key`

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -311,3 +305,21 @@ func (p *LaunchTemplateProvider) GetCABundle(ctx context.Context) (*string, erro
 	logging.FromContext(ctx).Debugf("Discovered caBundle, length %d", len(transportConfig.TLS.CAData))
 	return ptr.String(base64.StdEncoding.EncodeToString(transportConfig.TLS.CAData)), nil
 }
+
+func tagsFor(options *launchTemplateOptions) []*ec2.Tag {
+	var tags []*ec2.Tag
+	// Combine user defined and karpenter tags. If keys conflict, choose the latter.
```
ellistarn, line 311: This is interesting. Should we validate user tags to ensure they can't collide? Seems like perhaps surprising behavior to have the user defined tags overridden.

PR #779, opened 2021-11-02
Set RequeueAfter so liveness subreconciler fires when no more node events

suket22:
**1. Issue, if available:**
N/A

This sets a RequeueAfter on the liveness reconciler result, if it's been less than `LivenessTimeout` since the node was created. This ensures that even if there's no more subsequent node events we still consider the worker node for termination later. 

If you don't do so, you may enter into a runaway scaling problem where Karpenter will keep launching new nodes and pods indefinitely. The pods get stuck in `Terminating` state and the liveness reconciler never terminates the old `NotReady` worker nodes because no new node events are being generated and therefore the reconciler is never triggered. 

How to reproduce the issue?
Configure Karpenter to launch worker nodes in a subnet that has no network connectivity. For example my subnet had a bad route table (no gateways). 

I verified after this code change, that after 15 minutes the nodes in NotReady status were terminated.

- [x] No

### DIFF
--- a/pkg/controllers/node/liveness.go
+++ b/pkg/controllers/node/liveness.go
@@ -37,8 +37,8 @@ type Liveness struct {
 
 // Reconcile reconciles the node
 func (r *Liveness) Reconcile(ctx context.Context, _ *v1alpha5.Provisioner, n *v1.Node) (reconcile.Result, error) {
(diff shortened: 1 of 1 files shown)


ellistarn (approved):
Thanks for fixing my buggy snippit I messaged to you :D

PR #780, opened 2021-11-04
Add Offerings to InstanceType to hold zone and capacity type, remove OS from InstanceType, promote capacity-type label

eptiger:
**1. Issue, if available:**
Part of https://github.com/awslabs/karpenter/issues/371

* OS is not an instance type level concept - this seems to have been shoehorned in and causes unnecessary confusion. Removing it until we figure out proper support for OS constraints.
* When we do ICE caching we need to be able to filter out combinations of zone and capacity type when we see ICEs. Since packing otherwise does not care about them it seemed preferable to put a slice of these tuples ("Offerings") inside InstanceType rather than having to permute all the possible combinations in bin packing (i.e. if there are 3 zones and 2 capacity types then we can still consider just a single InstanceType for an actual instance type with this Offering re-factoring rather than having to create 5 InstanceType instances in the status quo to cover the permutations of zone and capacity type).
* Added a note on tailing logs to the dev guide. Maybe obvious to those of you very familiar with K8s, but was not obvious to me =P
* In talking to Ellis, decided to bite the bullet and re-name the capacity-type label to make it a first class citizen rather than being relegated to the AWS cloud provider. This simplifies some of the scheduler code a little and honestly just makes more sense.

I'm not entirely sure if I need to add additional unit tests at this time - I think we're covered with what's there but open to feedback.

- [ ] No 


pkg/controllers/allocation/binpacking/packable.go
```diff
@@ -164,16 +166,28 @@ func (p *Packable) validateArchitecture(schedule *scheduling.Schedule) error {
...
 }
 
-func (p *Packable) validateZones(schedule *scheduling.Schedule) error {
-	if schedule.Requirements.Zones().Intersection(p.Zones()).Len() == 0 {
-		return fmt.Errorf("zones %v are not in %v", p.Zones(), schedule.Requirements.Zones().List())
+func (p *Packable) validateCapacityTypes(schedule *scheduling.Schedule) error {
+	scheduledCapacityTypes := schedule.Requirements.Requirement(v1alpha1.CapacityTypeLabel)
+	if scheduledCapacityTypes.Len() == 0 {
+		return nil
+	}
```
eptiger, line 184: TBH I'm not entirely sure if this is correct, but the tests fail without it. Need to get more sleep and read more code to think through this.

pkg/cloudprovider/aws/instance.go
```diff
@@ -186,12 +193,16 @@ func (p *InstanceProvider) getOverrides(instanceTypeOptions []cloudprovider.Inst
...
+					zonesUsed.Insert(zone)
 					// FleetAPI cannot span subnets from the same AZ, so break after the first one.
 					break
 				}
 			}
 		}
 	}
+	if len(overrides) == 0 {
+		fmt.Printf("Overrides was empty. Capacity type was %v and instance types were %v", capacityType, instanceTypeOptions)
+	}
```
eptiger, line 205: Whoopsie, this was for debugging. I will remove in my next revision.

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -129,14 +129,6 @@ spec:
         kubernetes.io/arch: amd64
 ```
 
-### Operating System
```
bwagner5, line 132: we still need this

pkg/controllers/allocation/scheduling/scheduler.go
```diff
@@ -102,13 +102,15 @@ func globalRequirements(instanceTypes []cloudprovider.InstanceType) (requirement
...
 	for _, instanceType := range instanceTypes {
+		scheduledZoneSet := sets.String{}
+		for _, offering := range instanceType.Offerings() {
+			scheduledZoneSet.Insert(offering.Zone)
+		}
 		supported[v1.LabelInstanceTypeStable].Insert(instanceType.Name())
-		supported[v1.LabelTopologyZone].Insert(instanceType.Zones().UnsortedList()...)
+		supported[v1.LabelTopologyZone].Insert(scheduledZoneSet.UnsortedList()...)
 		supported[v1.LabelArchStable].Insert(instanceType.Architecture())
-		supported[v1.LabelOSStable].Insert(instanceType.OperatingSystems().UnsortedList()...)
```
bwagner5, line 111: We might still need this if other cloud providers don't support Windows on all instance types.

pkg/apis/provisioning/v1alpha5/register.go
```diff
@@ -57,7 +57,6 @@ var (
 		v1.LabelTopologyZone,
 		v1.LabelInstanceTypeStable,
 		v1.LabelArchStable,
-		v1.LabelOSStable,
```
bwagner5, line 60: This is still a well known label that we will support for Windows eventually.

ellistarn: This currently will break scheduling assumptions since the options are [] and it will be overly constrained. Let's add it back in when we support it.

website/content/en/docs/cloud-providers/AWS.md
eptiger: What value does it add? We don't really support OS, why advertise it?

pkg/controllers/allocation/scheduling/scheduler.go
eptiger: We can add it back then. I don't even know what it means for an instance type to not support a certain OS, that would be such an artificial restriction for the cloud provider.

eptiger: (then meaning at that time)

pkg/cloudprovider/aws/instance.go
```diff
@@ -171,8 +171,15 @@ func (p *InstanceProvider) getLaunchTemplateConfigs(ctx context.Context, constra
 
 func (p *InstanceProvider) getOverrides(instanceTypeOptions []cloudprovider.InstanceType, subnets []*ec2.Subnet, capacityType string) []*ec2.FleetLaunchTemplateOverridesRequest {
 	var overrides []*ec2.FleetLaunchTemplateOverridesRequest
+	zonesUsed := sets.String{}
 	for i, instanceType := range instanceTypeOptions {
-		for zone := range instanceType.Zones() {
+		for _, offering := range instanceType.Offerings() {
+			zone := offering.Zone
+			// we can't assume that all zones will be available for all capacity types, hence this check
+			// (we shouldn't ever have duplicate zones, just being paranoid to create a well-formed set of overrides)
```
ellistarn, line 179: This type of check is like wearing kevlar in your own living room in case you accidentally shoot yourself. There's no customer exposed surface here and Offerings() is a set (right?). IMO we should skip the zone uniqueness check.

pkg/cloudprovider/aws/instance.go
```diff
@@ -240,8 +248,9 @@ func (p *InstanceProvider) instanceToNode(instance *ec2.Instance, instanceTypes
 						v1.ResourceMemory: *instanceType.Memory(),
 					},
 					NodeInfo: v1.NodeSystemInfo{
-						Architecture:    v1alpha1.AWSToKubeArchitectures[aws.StringValue(instance.Architecture)],
-						OSImage:         aws.StringValue(instance.ImageId),
+						Architecture: v1alpha1.AWSToKubeArchitectures[aws.StringValue(instance.Architecture)],
+						OSImage:      aws.StringValue(instance.ImageId),
+						//TODO: stop hard-coding this
```
ellistarn, line 253: I think you can drop this TODO.

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -72,7 +72,15 @@ func (p *InstanceTypeProvider) Get(ctx context.Context, constraints *v1alpha1.Co
 	// Convert to cloudprovider.InstanceType
 	result := []cloudprovider.InstanceType{}
 	for _, instanceType := range instanceTypes {
-		instanceType.ZoneOptions = subnetZones.Intersection(instanceTypeZones[instanceType.Name()])
+		//TODO filter out possible zones and capacity types using an ICE cache
+		possibleZones := subnetZones.Intersection(instanceTypeZones[instanceType.Name()])
```
ellistarn, line 76: any reason to not inline this?
```
for zone := range subnetZones.Intersection(instanceTypeZones[instanceType.Name()]) {
...
}
```

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -72,7 +72,15 @@ func (p *InstanceTypeProvider) Get(ctx context.Context, constraints *v1alpha1.Co
...
+		offerings := []cloudprovider.Offering{}
+		for zone := range possibleZones {
+			for _, capacityType := range instanceType.SupportedUsageClasses {
+				offerings = append(offerings, cloudprovider.Offering{Zone: zone, CapacityType: *capacityType})
```
ellistarn, line 80: aws.StringValue(capacityType)

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -419,8 +419,8 @@ var _ = Describe("Allocation", func() {
 			})
 			It("should support a capacity type label", func() {
 				for _, value := range []string{v1alpha1.CapacityTypeOnDemand, v1alpha1.CapacityTypeSpot} {
-					provisioner.Spec.Labels = map[string]string{v1alpha1.CapacityTypeLabel: value}
-					Expect(provisioner.Validate(ctx)).ToNot(Succeed())
+					provisioner.Spec.Labels = map[string]string{v1alpha5.LabelCapacityType: value}
+					Expect(provisioner.Validate(ctx)).To(Succeed())
```
ellistarn, line 423: Why was this false before?

pkg/apis/provisioning/v1alpha5/provisioner.go
```diff
@@ -114,6 +114,10 @@ func (r Requirements) OperatingSystems() sets.String {
 	return r.Requirement(v1.LabelOSStable)
 }
 
+func (r Requirements) CapacityTypes() sets.String {
```
JacobGabrielson, line 117: To be consistent I'd add the same comment on the line above as per line 112 (eg)

pkg/cloudprovider/fake/cloudprovider.go
```diff
@@ -37,8 +38,25 @@ func (c *CloudProvider) Create(_ context.Context, constraints *v1alpha5.Constrai
 	for i := 0; i < quantity; i++ {
 		name := strings.ToLower(randomdata.SillyName())
 		instance := instanceTypes[0]
-		zone := instance.Zones().Intersection(constraints.Requirements.Zones()).UnsortedList()[0]
-		operatingSystem := instance.OperatingSystems().UnsortedList()[0]
+		zoneSet := sets.String{}
```
ellistarn, line 41: I'm having a hard time understanding why this is so complicated to fake a zone. Can we make some helpers?

pkg/controllers/allocation/binpacking/packable.go
```diff
@@ -164,16 +165,24 @@ func (p *Packable) validateArchitecture(schedule *scheduling.Schedule) error {
 	return nil
 }
 
-func (p *Packable) validateOperatingSystem(schedule *scheduling.Schedule) error {
-	if schedule.Requirements.OperatingSystems().Intersection(p.OperatingSystems()).Len() == 0 {
-		return fmt.Errorf("operating system %s is not in %v", p.OperatingSystems(), schedule.Requirements.OperatingSystems().List())
+func (p *Packable) validateZones(schedule *scheduling.Schedule) error {
+	packableZoneSet := sets.String{}
```
JacobGabrielson, line 169: nit: I'd suggest just calling this `packableZones` - typically it's easier to read that way (no need to put the type into the variable name). It's also more consistent with the convention used on line 180.

pkg/controllers/allocation/scheduling/scheduler.go
```diff
@@ -102,13 +102,19 @@ func globalRequirements(instanceTypes []cloudprovider.InstanceType) (requirement
 		v1.LabelInstanceTypeStable: sets.NewString(),
 		v1.LabelTopologyZone:       sets.NewString(),
 		v1.LabelArchStable:         sets.NewString(),
-		v1.LabelOSStable:           sets.NewString(),
+		v1alpha5.LabelCapacityType: sets.NewString(),
 	}
 	for _, instanceType := range instanceTypes {
+		scheduledZoneSet := sets.String{}
+		scheduledCapacityTypeSet := sets.String{}
```
JacobGabrielson, line 109: I'd suggest removing `Set` from the variable names.

pkg/controllers/allocation/suite_test.go
```diff
@@ -138,15 +140,18 @@ var _ = Describe("Allocation", func() {
...
-			Expect(len(nodes.Items)).To(Equal(6)) // 5 schedulable -> 5 node, 2 coschedulable -> 1 node
 			for _, pod := range schedulable {
 				scheduled := ExpectPodExists(env.Client, pod.GetName(), pod.GetNamespace())
 				ExpectNodeExists(env.Client, scheduled.Spec.NodeName)
 			}
 			for _, pod := range unschedulable {
 				unscheduled := ExpectPodExists(env.Client, pod.GetName(), pod.GetNamespace())
+				if unscheduled.Spec.NodeName != "" {
+					fmt.Printf("This pod shouldn't have been scheduled: %s", pretty.Concise(pod))
+				}
```
JacobGabrielson, line 151: This might just be stray debugging code left in?

pkg/controllers/allocation/binpacking/packable.go
ellistarn, line 169: consider s/packableZoneSet/zones. I try to avoid naming by variables after their type (`set`).

pkg/controllers/allocation/scheduling/scheduler.go
```diff
@@ -102,13 +102,19 @@ func globalRequirements(instanceTypes []cloudprovider.InstanceType) (requirement
...
+		for _, offering := range instanceType.Offerings() {
+			scheduledZoneSet.Insert(offering.Zone)
+			scheduledCapacityTypeSet.Insert(offering.CapacityType)
+		}
```
ellistarn, line 113:
```suggestion
		for _, offering := range instanceType.Offerings() {
			supported[v1.LabelTopologyZone].Insert(offering.Zone)
			supported[v1alpha5.LabelCapacityType].Insert(offering.CapacityType)
		}
```

pkg/controllers/allocation/suite_test.go
```diff
@@ -138,15 +140,18 @@ var _ = Describe("Allocation", func() {
...
 			Expect(env.Client.List(ctx, nodes)).To(Succeed())
-			Expect(len(nodes.Items)).To(Equal(6)) // 5 schedulable -> 5 node, 2 coschedulable -> 1 node
 			for _, pod := range schedulable {
 				scheduled := ExpectPodExists(env.Client, pod.GetName(), pod.GetNamespace())
 				ExpectNodeExists(env.Client, scheduled.Spec.NodeName)
 			}
 			for _, pod := range unschedulable {
 				unscheduled := ExpectPodExists(env.Client, pod.GetName(), pod.GetNamespace())
+				if unscheduled.Spec.NodeName != "" {
+					fmt.Printf("This pod shouldn't have been scheduled: %s", pretty.Concise(pod))
```
ellistarn, line 150: Is the Expect() error message not good enough?

website/content/en/docs/cloud-providers/AWS.md
```diff
@@ -65,16 +65,15 @@ IDs.](https://docs.aws.amazon.com/ram/latest/userguide/working-with-az-ids.html)
 
 ### Capacity Type
 
-- key: `node.k8s.aws/capacity-type`
+- key: `karpenter.sh/capacity-type`
```
ellistarn, line 68: We need to be a lil' careful on updating docs before releasing.

pkg/apis/provisioning/v1alpha5/provisioner.go
eptiger: Would it offend you if I just remove all the other comments? I don't mean to be defensive, but I feel like I'm getting the theme of being too verbose in the feedback on this PR, so I feel like I'm gonna add that comment and then someone is going to tell me that it's a useless comment, and I'm going to agree with them =P

ellistarn: The style is definitely to write self documenting code where possible and add comments where things are actually clear. I agree with you that these methods are basically just getters and could get by without comments.

pkg/cloudprovider/aws/instance.go
eptiger: Offerings() is _not_ a set, it's a slice, so I fundamentally/philosophically disagree with you here. A duplicate is very much _possible_. A bug could easily be introduced where we have duplicate offerings. Because there's no constructor like Java, it's not like I can validate the initialization of the InstanceType struct (correct me if I'm wrong here), so there's nothing to maintain the invariant that the Offerings are all unique, unless I invent a set type specifically for Offering, which seems like way more work than adding 2 lines of code here.

The downside if I'm wrong and we never have a duplicate is we have an extra set that has a handful of elements in it and an extra conditional check. The downside if you're wrong is that users will fail to scale out because of invalid LT overrides.

pkg/cloudprovider/aws/instance.go
eptiger: I can drop it, I just wanted to understand why. Is this a stylistic thing with TODOs? Do you think we'll never change this? If so, why?

pkg/cloudprovider/aws/instancetypes.go
eptiger: Right, thanks

pkg/cloudprovider/aws/instancetypes.go
eptiger: No, this was an artifact of me re-working this code I think. I will fix it.

pkg/cloudprovider/aws/suite_test.go
eptiger: Honestly, i was hoping you would know since you had done this XD https://github.com/awslabs/karpenter/commit/67d637a3c3a2db5eb6c23801f3c12e79b8948f08#diff-d9acf1c0f1f6b83a9cecc0e967d90f0a2fd5bc6e7886d0607c05b533ad905828R415

Which was a re-factor of code that you also wrote: https://github.com/awslabs/karpenter/commit/69894662390c0fc3c4c7f07b5896055ebcfe9e43#diff-d9acf1c0f1f6b83a9cecc0e967d90f0a2fd5bc6e7886d0607c05b533ad905828R452

If I had to guess, you made a copy/paste mistake and so there was a legitimate bug in the code that was being missed because of this test. 

It didn't fail until my changes this morning so I expect is has something to do with promoting capacity type to be a well known label. But when I un-do that promotion other tests fail as well so I can't easily pinpoint what I did that fixed this.

pkg/cloudprovider/fake/cloudprovider.go
eptiger: It's possible that I'm just over-complicating things by mis-understanding how the fakes should work. I need the instance I fake create to match the requirements, right? So I first need to filter the offerings by what capacity types are allowed by the requirements and then by what zones are allowed by the requirements.

I _do_ think that I can pretty easily eliminate lines 58 and 59 by just picking the first offering that matches. And I think that lines 44-46 are unnecessary entirely now so I can cut those.

Not sure if I'm following your helper method ask, are you just wanting me to de-compose this code chunk so that it's shorter?

pkg/cloudprovider/aws/instance.go
JacobGabrielson: Generally we try to avoid them, they just tend to pile up and aren't really a great way to track future work (compared to filing an issue).

pkg/controllers/allocation/scheduling/scheduler.go
eptiger: It's a personal habit of mine from Java, I like having the type in the variable names (especially in code chunks where I use the same variable prefix because it's different forms of a singular entity), but I can stop doing that here.

pkg/controllers/allocation/scheduling/scheduler.go
eptiger: I like it!

pkg/controllers/allocation/suite_test.go
eptiger: I'm not entirely sure if you're being serious or sarcastic, I'll assume you're being serious though. Yes, the Expect() error message is basically worthless. 

```
Allocation Reconcilation [It] should provision nodes for pods with supported node selectors 
/local/workplace2/go/src/karpenter/pkg/controllers/allocation/suite_test.go:109

  Expected
      <string>: thiefcurly
  to equal
      <string>: 

  /local/workplace2/go/src/karpenter/pkg/controllers/allocation/suite_test.go:152
------------------------------
••••
```

That doesn't help me in the slightest other than to tell me that one of the "unschedulable" pods (that term is so confusing, btw, because it means two different things in this test) was scheduled. Once I printed out the pod labels it was trivially easy to debug:

> [1636000414] Controllers/Allocation - 6/6 specs •This pod shouldn't have been scheduled: {"metadata":{"name":"legweak","namespace":"default","selfLink":"/api/v1/namespaces/default/pods/legweak","uid":"013203f3-cf1b-42f1-8580-62112f5c0e0e","resourceVersion":"86","creationTimestamp":"2021-11-04T04:33:55Z","managedFields":[{"manager":"allocation.test","operation":"Update","apiVersion":"v1","time":"2021-11-04T04:33:55Z","fieldsType":"FieldsV1","fieldsV1":{"f:spec":{"f:containers":{"k:{\"name\":\"legweak\"}":{".":{},"f:image":{},"f:imagePullPolicy":{},"f:name":{},"f:resources":{},"f:terminationMessagePath":{},"f:terminationMessagePolicy":{}}},"f:dnsPolicy":{},"f:enableServiceLinks":{},"f:nodeSelector":{".":{},"f:kubernetes.io/os":{}},"f:restartPolicy":{},"f:schedulerName":{},"f:securityContext":{},"f:terminationGracePeriodSeconds":{}},"f:status":{"f:conditions":{".":{},"k:{\"type\":\"PodScheduled\"}":{".":{},"f:lastProbeTime":{},"f:lastTransitionTime":{},"f:reason":{},"f:status":{},"f:type":{}}}}}}]},"spec":{"containers":[{"name":"legweak","image":"k8s.gcr.io/pause","resources":{},"terminationMessagePath":"/dev/termination-log","terminationMessagePolicy":"File","imagePullPolicy":"Always"}],"restartPolicy":"Always","terminationGracePeriodSeconds":30,"dnsPolicy":"ClusterFirst","nodeSelector":{"kubernetes.io/os":"unknown"},"securityContext":{},"schedulerName":"default-scheduler","tolerations":[{"key":"node.kubernetes.io/not-ready","operator":"Exists","effect":"NoExecute","tolerationSeconds":300},{"key":"node.kubernetes.io/unreachable","operator":"Exists","effect":"NoExecute","tolerationSeconds":300}],"priority":0,"enableServiceLinks":true,"preemptionPolicy":"PreemptLowerPriority"},"status":{"phase":"Pending","qosClass":"BestEffort"}}

pkg/controllers/allocation/suite_test.go
eptiger: Do you have a problem with it? It's a test, this helps people figure out which test failed. I have no idea how someone is supposed to debug this test failure without this.

website/content/en/docs/cloud-providers/AWS.md
eptiger: I'm not following your comment here. Are you saying I shouldn't update the docs? Or is there some other issue with my doc update?

This change is essentially backwards incompatible so it's important IMO that the docs match the functionality.

pkg/cloudprovider/aws/instance.go
eptiger: Again: totally ok to abide by your conventions, but just wondering: why not just add a TODO and link to the issue? I offered to create an issue but Ellis said it wasn't necessary, which is why I didn't do that here. I generally tell people _to_ add TODOs and link to the issues created so that the spot in the code is easier to find by whoever is working the issue and also so that passerbys in the code base understand that it's a known issue.

website/content/en/docs/cloud-providers/AWS.md
geoffcline: for docs, it is better to answer in the negative.

for pr, it is better to be silent.

geoffcline:
I appreciate you including docs changes, it's helping me keep the thread.

website/content/en/docs/cloud-providers/AWS.md
eptiger: Sorry, I don't think I'm used to the communication style of this group XD Can you translate this for me, Geoff?

bwagner5: If a customer comes looking at the FAQs to see if we support Windows, they should find this section that says we only support Linux currently.

website/content/en/docs/cloud-providers/AWS.md
bwagner5: The docs are published immediately when the PR is merged, but we won't release a new version immediately when merged. So the docs will be updated with the new `karpenter.sh/capacity-type` but our latest published container image will not include it.  Anyone following the docs between that time will get errors.  

It might be better to open a second PR w/ the docs changes and label it with the release to coordinate the docs release at the same time as the artifact releases?  We don't have a great way of handling it today :(

pkg/apis/provisioning/v1alpha5/provisioner.go
eptiger: I'm gonna delete these comments then in my next commit then where I also (very sadly) revert all my doc updates =P

website/content/en/docs/cloud-providers/AWS.md
eptiger: Fair enough. Will revert.

website/content/en/docs/cloud-providers/AWS.md
eptiger: Ah, that's a PITA. Should we have a doc branch then? I don't know what to do with this second PR exactly - what does approval of it mean, because we can't pull it until we're ready to cut a release, right? And then, if there are multiple doc changes from multiple PRs someone has to go through the fun exercise of resolving conflicts. Am I missing something?

Is there some special syntax with labeling the commit? Do you mean just putting v0.5 in the commit message? Or do you mean adding the label in the PR in the GitHib UI? If so, I also have no idea how to do this - on this very PR on the right side the labels say "None yet" and there's no edit button or anything.

I'll back this change out, but I don't know how we make sure everyone currently contributing to Karpenter knows about this in the interim because it is a backwards incompatible change.

eptiger: Re: labeling - am I supposed to do something like this?

`git tag v0.4.2 <commit_sha>`

bwagner5: I meant in the github UI, you should have labeling permissions. Maybe we should just cut a v0.4.3 release when this gets merged since we are in a state of rapid iteration here.

eptiger: Pretty sure I don't have those permissions, how can I get them? I should already be in awslabs.

pkg/cloudprovider/aws/instance.go
JacobGabrielson: I hear ya'. I don't think there's going to be a great solution to this type of thing until Go adds generics to the language, at which point it would be pretty easy to make Offerings() return a set. FWIW, I think that the "house style" of a lot of Go code (in Kubernetes land) is to assume validation has occurred elsewhere and not keep checking invariants. Personally, I would suggest you just make the call here whether to keep or not, although I would probably remove the check myself.

pkg/cloudprovider/aws/suite_test.go
eptiger: Re-reading this I feel like I just threw this back in your face, not my intention - sorry! My point was to say: I couldn't figure it out, and didn't get a chance to ping you. I had figured out it was a test you made but didn't understand the test itself. My point was: I think you made an honest mistake in your original test and I happened to fix that bug in my PR. If you think that your original code was correct, would love help figuring out what I'm getting lost on here.

pkg/controllers/allocation/suite_test.go
eptiger: This comment also feels inflammatory as I re-read it. What I really wanted to say was: is there some conventional reason that it's wrong to print additional debugging information in this way for a test failure? Is there a preferred convention in Go? Without this change, I honestly don't know how to debug this test failure.

pkg/controllers/allocation/suite_test.go
eptiger: Haha again, as I re-read this it feels inflammatory but I promise I'm being earnest: I truly didn't know if you were being honest or sarcastic, it was not an aggressive message :) To me, the test failures I faced generally gave me back useless messages, I had made an assumption that this was a known issue.

pkg/cloudprovider/aws/instance.go
ellistarn: I don't think we really know what we want to do with OS at this point. In general, the custom LT story isn't fleshed out. Instead of guessing where we might need lines of code, I think it's just as good to just figure that out when we're fleshing out the OS story. It helps avoid leaving spurious TODOs around the codebase that may or may not ever get addressed.

pkg/controllers/allocation/suite_test.go
ellistarn: The tests are designed around Expect() calls. If any of the expectations fail, then the test will fail loudly with that message.

Expect(unscheduled.Spec.NodeName).ToNot(BeEmpty())
// logline is: "expected unscheduled.Spec.NodeName to not be empty, but it was _____"

pkg/controllers/allocation/suite_test.go
ellistarn: I think the error message will be more clear if we say `Expect(unscheduled.Spec.NodeName).To(BeEmpty())`. If the builtin expect matchers are spitting out crappy error messages, there are ways to customize them:
https://github.com/awslabs/karpenter/blob/main/pkg/test/expectations/expectations.go#L57

website/content/en/docs/cloud-providers/AWS.md
JacobGabrielson: @eptiger I just made you an Admin on this repo, so hopefully that'll work?

eptiger: Oh Brandon gave me the permissions I needed, but I don't see a label for v0.4.3 yet - should I create one?

pkg/controllers/allocation/suite_test.go
eptiger: I don't know if I'd consider that alone to be an improvement:

```
Allocation Reconcilation [It] should provision nodes for pods with supported node selectors 
/local/workplace2/go/src/karpenter/pkg/controllers/allocation/suite_test.go:109

  Expected
      <string>: thunderrust
  to be empty

  /local/workplace2/go/src/karpenter/pkg/controllers/allocation/suite_test.go:152
```

I think I'll just make a custom one instead.

pkg/test/expectations/expectations.go
```diff
@@ -59,6 +60,15 @@ func ExpectNotFound(c client.Client, objects ...client.Object) {
 	}
 }
 
+func ExpectPodNotScheduled(c client.Client, name string, namespace string) {
+	pod := &v1.Pod{}
+	Expect(c.Get(context.Background(), client.ObjectKey{Name: name, Namespace: namespace}, pod)).To(Succeed())
+
```
ellistarn, line 66: nit: empty line

ellistarn (commented):
I haven't had a chance to fully review this. Feel free to push forward w/ Jacob's review. I agree with the high level direction.

bwagner5 (approved):
/lgtm

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -72,7 +72,14 @@ func (p *InstanceTypeProvider) Get(ctx context.Context, constraints *v1alpha1.Co
 	// Convert to cloudprovider.InstanceType
 	result := []cloudprovider.InstanceType{}
 	for _, instanceType := range instanceTypes {
-		instanceType.ZoneOptions = subnetZones.Intersection(instanceTypeZones[instanceType.Name()])
+		//TODO filter out possible zones and capacity types using an ICE cache https://github.com/awslabs/karpenter/issues/371
+		offerings := []cloudprovider.Offering{}
+		for zone := range subnetZones.Intersection(instanceTypeZones[instanceType.Name()]) {
+			for _, capacityType := range instanceType.SupportedUsageClasses {
```
ellistarn, line 78: ```for capacityType := range sets.NewString(aws.StringValue(instanceType.SupportedUsageClasses)...)```

ellistarn (commented):
one final comment

pkg/cloudprovider/aws/instance.go
eptiger: Ellis and I came up with a good compromise - I'm going to move the validation to the instance type provider. Update coming momentarily.

Team work makes the dream work :rainbow:

pkg/cloudprovider/aws/instance.go
```diff
@@ -172,7 +172,12 @@ func (p *InstanceProvider) getLaunchTemplateConfigs(ctx context.Context, constra
 func (p *InstanceProvider) getOverrides(instanceTypeOptions []cloudprovider.InstanceType, subnets []*ec2.Subnet, capacityType string) []*ec2.FleetLaunchTemplateOverridesRequest {
 	var overrides []*ec2.FleetLaunchTemplateOverridesRequest
 	for i, instanceType := range instanceTypeOptions {
-		for zone := range instanceType.Zones() {
+		for _, offering := range instanceType.Offerings() {
+			zone := offering.Zone
```
ellistarn, line 176: optional: I might just inline this where its used. I don't think there's much of a perf angle, and it adds an extra concept to think about as you read.

pkg/cloudprovider/fake/cloudprovider.go
```diff
@@ -37,8 +36,15 @@ func (c *CloudProvider) Create(_ context.Context, constraints *v1alpha5.Constrai
...
 		name := strings.ToLower(randomdata.SillyName())
 		instance := instanceTypes[0]
-		zone := instance.Zones().Intersection(constraints.Requirements.Zones()).UnsortedList()[0]
-		operatingSystem := instance.OperatingSystems().UnsortedList()[0]
+		var zone, capacityType string
+		for _, o := range instance.Offerings() {
+			if constraints.Requirements.CapacityTypes().Has(o.CapacityType) {
+				if constraints.Requirements.Zones().Has(o.Zone) {
+					zone = o.Zone
+					capacityType = o.CapacityType
```
ellistarn, line 44: Do you want to break here? Does it matter?

pkg/cloudprovider/fake/instancetype.go
```diff
@@ -44,33 +51,33 @@ func NewInstanceType(options InstanceTypeOptions) *InstanceType {
...
-	architecture     string
-	operatingSystems sets.String
-	cpu              resource.Quantity
-	memory           resource.Quantity
-	pods             resource.Quantity
-	nvidiaGPUs       resource.Quantity
-	amdGPUs          resource.Quantity
-	awsNeurons       resource.Quantity
+	name          string
+	zones         sets.String
```
ellistarn, line 71: possible to remove zone/capacityTypes?

pkg/controllers/allocation/binpacking/packable.go
```diff
@@ -164,16 +165,24 @@ func (p *Packable) validateArchitecture(schedule *scheduling.Schedule) error {
 	return nil
 }
 
-func (p *Packable) validateOperatingSystem(schedule *scheduling.Schedule) error {
-	if schedule.Requirements.OperatingSystems().Intersection(p.OperatingSystems()).Len() == 0 {
-		return fmt.Errorf("operating system %s is not in %v", p.OperatingSystems(), schedule.Requirements.OperatingSystems().List())
+func (p *Packable) validateZones(schedule *scheduling.Schedule) error {
+	packableZones := sets.String{}
```
ellistarn, line 169: Optional:

Not sure if I mentioned earlier, but go style would be to let the "packable" piece of this variable be inferred by context of it being within a Packable struct. This is known as "stuttering". Sometimes if you have multiple concepts that collide (i.e. two different sources of zone), you need to get a bit clever about naming.

ellistarn (approved):
Nice work! I noticed a couple of optional things. You should feel totally free to merge as is. Happy to approve again in case you want to tweak.

pkg/cloudprovider/aws/instance.go
eptiger: Yeah, this was an artifact of the uniqueness check

pkg/cloudprovider/fake/cloudprovider.go
eptiger: Oh shoot yeah, good catch. I mean it ultimately doesn't matter, but yeah I was supposed to. I don't know how I missed that =P

pkg/cloudprovider/fake/instancetype.go
eptiger: Darn it, in the midst of test debugging I missed this - thanks!

pkg/controllers/allocation/binpacking/packable.go
eptiger: This is I think actually an artifact of me doing something stupidly complicated here where I had two sets of zones and had to differentiate them. Fixed.

ellistarn (dismissed):

ellistarn (approved):

PR #785, opened 2021-11-09
fix aws cp caching empty instance types set

bwagner5:
**1. Issue, if available:**
N/A

1. The AWS CP InstanceTypes provider was using a defer func to cache results. If the DescribeInstanceTypes call err'd for any reason, the defer block still runs and caches an empty value. On subsequent reconciliations, the scheduler would not fill in well known labels causing this logging:

```
2021-11-08T22:39:28.831Z	INFO	controller.allocation.provisioner/default	Ignored pod test1-basic-0-774f48fcd-hkf85/koi-test3, kubernetes.io/arch is too constrained: 
kubernetes.io/os is too constrained: 
node.kubernetes.io/instance-type is too constrained: 
```

2. The test I was running received a RequestLimitExceeded on the DescribeInstanceTypes call and resulted in an empty set of instance types cached. 

```
2021-11-08T22:39:23.117Z	ERROR	controller.controller.Allocation	Reconciler error	{"commit": "82226ca", "reconciler group": "karpenter.sh", "reconciler kind": "Provisioner", "name": "default", "namespace": "", "error": "getting instance types, retrieving all instance types, fetching instance types using ec2.DescribeInstanceTypes, RequestLimitExceeded: Request limit exceeded.\n\tstatus code: 503
```

Since DescribeInstanceTypes should change very rarely, I set a custom cache TTL for the instance types and zones cache. The default it was using was set to 1 minute. I set the new value to 5 minutes. 

3. I cleaned up some comment typos

4. Cleaned up some redundant error messages since we were wrapping and propagating errors in the InstanceTypes provider. 

- [x] No

### DIFF
--- a/pkg/cloudprovider/aws/instancetypes.go
+++ b/pkg/cloudprovider/aws/instancetypes.go
@@ -88,7 +88,7 @@ func (p *InstanceTypeProvider) Get(ctx context.Context, constraints *v1alpha1.Co
 	return result, nil
 }
 
-func (p *InstanceTypeProvider) getInstanceTypeZones(ctx context.Context) (result map[string]sets.String, err error) {
+func (p *InstanceTypeProvider) getInstanceTypeZones(ctx context.Context) (map[string]sets.String, error) {
 	if cached, ok := p.cache.Get(instanceTypeZonesCacheKey); ok {
 		return cached.(map[string]sets.String), nil
 	}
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -82,7 +84,6 @@ func (p *InstanceTypeProvider) getInstanceTypeZones(ctx context.Context) (result
 	if cached, ok := p.cache.Get(instanceTypeZonesCacheKey); ok {
 		return cached.(map[string]sets.String), nil
 	}
-	defer func() { p.cache.SetDefault(instanceTypeZonesCacheKey, result) }()
```
ellistarn, line 85: I think this is bugged because I created a new map on line 87. If you use the named return type it should work. Otherwise remove the name from the named return type.

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -105,7 +107,6 @@ func (p *InstanceTypeProvider) getInstanceTypes(ctx context.Context) (result map
 	if cached, ok := p.cache.Get(instanceTypesCacheKey); ok {
 		return cached.(map[string]*InstanceType), nil
 	}
-	defer func() { p.cache.SetDefault(instanceTypesCacheKey, result) }()
 	instanceTypes := map[string]*InstanceType{}
```
ellistarn, line 110: same stupid bug 🤦‍♂️

ellistarn (approved):

PR #786, opened 2021-11-09
add default provisioner to helm chart

bwagner5:
**1. Issue, if available:**
https://github.com/awslabs/karpenter/issues/781
https://github.com/awslabs/karpenter/issues/784

 - Add default provisioner to helm chart
 - Move CRD to `crds` dir per helm3 guidance so that CRD is sequenced before the provisioner

- [x] No


bwagner5:
There seems to be an uninstall problem with helm and using CRDs. I'll have to dig into that more, but wanted to note this issue in the PR.

charts/karpenter/values.yaml
```diff
@@ -18,3 +18,12 @@ webhook:
   tolerations: []
   affinity: {}
   image: "public.ecr.aws/karpenter/webhook:v0.4.2@sha256:2d004cd277209242d7e886308be419f86107811b4d0f1ed21c908b77e284202b"
+defaultProvisioner:
+  create: true
```
ellistarn, line 22: Can you use the non-nil-ness of the default provisioner field to decide to create this?

ellistarn:
Check out https://github.com/awslabs/karpenter/issues/781

charts/karpenter/values.yaml
bwagner5: Generally, helm charts use an explicit `create` variable, so I'd prefer to keep this. It would be awkward on the CLI using `--set` with the default values.yaml if we just use the truthiness of defaultProvisioner.

charts/karpenter/values.yaml
```diff
@@ -18,3 +18,12 @@ webhook:
...
+  ttlSecondsAfterEmpty: 300
+  #ttlSecondsUntilExpired: 86400
```
JacobGabrielson, line 24: Just wondering - do we want to leave this commented-out value here?

bwagner5: I was going to leave it here to show that it is configurable since we don't have great helm docs now.

ellistarn (approved):

krisdock:
did this ever make it into a release? the template file and values block for this don't seem to exist in any tag.

bwagner5:
It did and then it was removed due to issues with Helm managing custom resources well.

PR #788, opened 2021-11-09
Adds design doc to discuss Karpenter limits

suket22:
Discusses limits to the amount of resources that Karpenter provisions. I've left implementation details out of here but wanted to document and align on our thinking about this before I move forward with code changes. 

**1. Issue, if available:**
N/A

Just a doc.

- [x] No

Docs will come in once we've implemented the thing.

### DIFF
--- a/designs/limits.md
+++ b/designs/limits.md
@@ -1,30 +1,28 @@
 # Karpenter Limits
 
-
 This document proposes an approach to limit the scaling that Karpenter will perform, in order to control customer costs and to prevent underutilized capacity in cases where pending pods cannot be scheduled on new capacity.
-
 
 ## Problem
 
 Karpenter responds to unschedulable pods stuck in the pending status by launching new worker nodes that will satisfy its constraints.
 
-However, in certain situations new worker nodes may not be able to satisfy these pending pods. Consider the situation where the AMI being used to create the worker nodes is buggy, which prevents the kubelet from ever spinning up healthy. As a result new worker nodes continue to remain in the `NotReady` state and the node lifecycle controller will eventually taint the worker node after a default wait of 5 minutes leading to the already bound pods getting stuck in `Terminating`. Karpenter won't terminate these nodes since they're not empty. When using a deployment, new replacement pods will be spun up which become stuck in `Pending` and the cycle repeats. Similar scenarios can be replicated by forcing worker nodes to be created in a subnet with a bad route table. For the remainder of this document, I'll refer to this as the `runaway-scaling` problem.
+However, in certain situations new worker nodes may not be able to satisfy these pending pods. Consider the situation where the AMI being used to create the worker nodes is buggy, which prevents the kubelet from ever spinning up healthy. As a result new worker nodes continue to remain in the `NotReady` state and the node lifecycle controller will eventually taint the worker node after a default wait of 5 minutes leading to the already bound pods getting stuck in `Terminating`. Karpenter won't terminate these nodes since they're not empty. When using a deployment, new replacement pods will be spun up which become stuck in `Pending` and the cycle repeats leading to new worker nodes being launched every 5 minutes indefinitely. Similar scenarios can be replicated by forcing worker nodes to be created in a subnet with a bad route table. For the remainder of this document, I'll refer to this as the `runaway-scaling` problem.
 
 The next large problem is the inability to define a hard ceiling on cluster costs. When using AWS, the common pattern for setting up your infrastructure-level scaling is using EC2 AutoScaling groups with Cluster Autoscaler (CA). Typically, by limiting the maximum size of an autoscaling group we can enforce a ceiling on how expensive your cluster can get since CA will honor the bounds specified by the auto scaling group.
 
 We need to provide similar functionality via Karpenter as well wherein there's a hard limit a customer can configure.
 
 
-## Current Implementation
+## Current State
 
 To address the runaway-scaling problem the current fix in place is to detect if the kubelet for a worker node has never reported its status to the K8s control plane. If it's been longer than 15 minutes, Karpenter assumes that there's a hard failure mode due to which this worker node will never become healthy and terminates the worker node. If the condition map of the node object in the API Server says `NodeStatusNeverUpdated` then we use that as an indicator of the node having never come up.
 
-This ensures that if there are other scenarios where a worker node has become unhealthy due to some network partition or power outage in a availability zone, we don't terminate those worker nodes. It's important we don't make the static stability of a cluster worse during such an event. On the other hand, if there is an edge case where worker nodes come online and soon go offline, it will lead to runaway scaling again. This edge case should be unlikely to happen in the near term, so this document focuses on just the ability to limit costs within Karpenter. That way even if runaway scaling does occur, there's a way to bound it.
+This fix ensures that if there are other scenarios where a worker node has become unhealthy due to a network partition or power outage in a availability zone, we don't terminate those worker nodes. It's important we don't make the static stability of a cluster worse during such an event. On the other hand, if there is an edge case where worker nodes come online and soon go offline, it will lead to runaway scaling again. This edge case should be unlikely to happen in the near term, so this document focuses on just the ability to limit costs within Karpenter. That way even if runaway scaling does occur there's a way to bound it. A longer-term solution to handle the runaway problem will be discussed separately.
 
 
 ## Proposed Solution for Limits
 
-There are two broad forms of limiting we could apply. The first is that we could introduce a limit to the number of in-flight worker node being provisioned at a point in time. A worker node that's in the `NotReady` state could be considered to be in-flight.
+There are two broad forms of limiting we could apply. The first is that we could introduce a limit to the number of in-flight worker node being provisioned at a point in time. A worker node that's in the `NotReady` state could be considered to be in-flight. The second form is an absolute limit of the number of resources Karpenter can provision.
 
 
 ### **In-flight limit**
@@ -35,7 +33,7 @@ spec:
     unready: 20% # StringOrInt
 ```
 
-The good bit about this approach is that we don't artificially constrain how many worker nodes can be spun up by Karpenter. Karpenter doesn't need to be aware of how limits are represented by each cloudprovider, while also making sure that if we're launching worker nodes that aren't healthy we stop the scaling and save costs.
+The good bit about this approach is that we don't artificially constrain how many total worker nodes can be spun up by Karpenter, while also making sure that if we're launching worker nodes that aren't healthy we stop the scaling and save costs.
 
 The two main problems with this approach though are -
 1. This limit while meant to constrain the number of unhealthy worker nodes in a cluster, will also inihibit the rate at which Karpenter can respond to pods that aren't schedulable. This somewhat goes against the goal of minimizing launch times of workers.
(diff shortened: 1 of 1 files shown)


designs/limits.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Problem
+
+Karpenter responds to unschedulable pods stuck in the pending status by launching new worker nodes that will satisfy its constraints.
+
+However, in certain situations new worker nodes may not be able to satisfy these pending pods. Consider the situation where the AMI being used to create the worker nodes is buggy, which prevents the kubelet from ever spinning up healthy. As a result new worker nodes continue to remain in the `NotReady` state and the node lifecycle controller will eventually taint the worker node after a default wait of 5 minutes leading to the already bound pods getting stuck in `Terminating`. Karpenter won't terminate these nodes since they're not empty. When using a deployment, new replacement pods will be spun up which become stuck in `Pending` and the cycle repeats leading to new worker nodes being launched every 5 minutes indefinitely. Similar scenarios can be replicated by forcing worker nodes to be created in a subnet with a bad route table. For the remainder of this document, I'll refer to this as the `runaway-scaling` problem.
```
JacobGabrielson, line 9: Typo: `healthy` -> `healthy nodes`?

designs/limits.md
```diff
@@ -0,0 +1,60 @@
...
+This is a much simpler semantic to follow, where you provide an upper cap on the amount of resources being provisioned irrespective of their health. This is more in-line with alternative solutions like CA and should be easier to reason with.
+
+The actual limit of the number of resources being spun up can be defined either through something similar to resource requests or just a flat count of instances.
+
+```yaml
+spec:
+  limits:
+    resources:
+      cpu: 1000
+      memory: 1000Gi
```
JacobGabrielson, line 53: I'm guessing you already had this in mind, but could you add GPU here as well?

designs/limits.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+Modeling your limit via CPU and Memory constraints aligns well with customers that use resource requests with their pods. This should give everyone a consistent interface to define the capacity requirements of their applications as well as their underlying infrastructure. There shouldn't be a need to think of the actual number and type of instances that are being chosen by Karpenter.
+
+As a cost control mechanism, this requires a little more work from our users if they're looking for higher precision. Since some cloud providers like AWS charge differently per instance type rather than the actual resource request and consumption, a little more translation will be needed to estimate costs. The alternative of a flat count of instance types - say `spec.limits.resources.count`, doesn't help solve that problem either since the instance types are chosen by Karpenter by default. In case overrides are defined via `kubernetes.io/instance-type`, you should be able to calculate the CPU and Memory bounds fairly trivially through some cloud provider API calls.
```
JacobGabrielson, line 58: I'd also mention that things like reserved instances (of varying durations/up-front, etc), also make calculating the price tricky in any given situation, so that's yet another reason to keep things simple and just go with aggregate cpu/mem/gpu counts. I guess one question is, will people want to set limits on total (aggregate) volume size too?

designs/limits.md
```diff
@@ -0,0 +1,60 @@
...
+
+This fix ensures that if there are other scenarios where a worker node has become unhealthy due to a network partition or power outage in a availability zone, we don't terminate those worker nodes. It's important we don't make the static stability of a cluster worse during such an event. On the other hand, if there is an edge case where worker nodes come online and soon go offline, it will lead to runaway scaling again. This edge case should be unlikely to happen in the near term, so this document focuses on just the ability to limit costs within Karpenter. That way even if runaway scaling does occur there's a way to bound it. A longer-term solution to handle the runaway problem will be discussed separately.
+
+
+## Proposed Solution for Limits
+
+There are two broad forms of limiting we could apply. The first is that we could introduce a limit to the number of in-flight worker node being provisioned at a point in time. A worker node that's in the `NotReady` state could be considered to be in-flight. The second form is an absolute limit of the number of resources Karpenter can provision.
```
JacobGabrielson, line 25: Maybe this is a premature question, but I'm not sure if you're actually recommending one of the two approaches below? Also, were there other approaches we considered but rejected already?

designs/limits.md
bwagner5, line 58: IMO this is a good enough limits mechanism for the specific case. I don't think we really want this to be a general cost control mechanism, but rather a "something is very wrong, let's stop" type of control.

A general cost mechanism would be difficult w/ all the purchase types for an individual instance type, like spot vs on-demand, reserved instances, savings plans, etc. 

I'm just agreeing with your point that it's hard and not worth calculating at a higher precision.

designs/limits.md
```diff
@@ -0,0 +1,60 @@
...
+
+### **Absolute limit**
+
+This is a much simpler semantic to follow, where you provide an upper cap on the amount of resources being provisioned irrespective of their health. This is more in-line with alternative solutions like CA and should be easier to reason with.
+
+The actual limit of the number of resources being spun up can be defined either through something similar to resource requests or just a flat count of instances.
+
+```yaml
+spec:
+  limits:
```
bwagner5, line 50: This will get hard to manage if users have a lot of provisioners especially if we end up having pods that could schedule to multiple provisioners.  I wonder if we should also have an optional limit passed directly to the karpenter binary that would limit the total amount of resources karpenter is managing across all provisioners.

designs/limits.md
```diff
@@ -0,0 +1,60 @@
+# Karpenter Limits
+
+This document proposes an approach to limit the scaling that Karpenter will perform, in order to control customer costs and to prevent underutilized capacity in cases where pending pods cannot be scheduled on new capacity.
```
bwagner5, line 3: How does this doc address "underutilized capacity" ?

bwagner5 (commented):
Nice design! I have a few questions and comments.

designs/limits.md
suket22: Yes will do

designs/limits.md
suket22: Right yeah, I'll add that callout too!

That's a good point on volume sizes. I was playing with the [aws pricing calculator](https://calculator.aws/#/createCalculator/EC2) and to me it seemed that instance costs were far larger than storage so maybe not all customers would need to specify this constraint. 

I think we could start with just start with cpu, mem and GPU and introduce additional knobs in the future if we hear differently from customers? It'd be backwards compatible for us to do so.

designs/limits.md
suket22: I'll make that more clear, but I recommend the second approach. I wanted to keep things as simple as possible so didn't consider any other approaches

designs/limits.md
suket22: It does not, let me fix this. I originally meant to discuss more about the runaway-scaling problem where your instances are effectively underutilized since none of your pods are running but ended up not doing that

designs/limits.md
```diff
@@ -0,0 +1,60 @@
...
+
+There are two broad forms of limiting we could apply. The first is that we could introduce a limit to the number of in-flight worker node being provisioned at a point in time. A worker node that's in the `NotReady` state could be considered to be in-flight. The second form is an absolute limit of the number of resources Karpenter can provision.
+
+
+### **In-flight limit**
+
+```yaml
+spec:
+  limits:
+    unready: 20% # StringOrInt
```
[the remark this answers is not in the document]
suket22: Right it's a percentage of currently running nodes and we'd round up

designs/limits.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+The good bit about this approach is that we don't artificially constrain how many total worker nodes can be spun up by Karpenter, while also making sure that if we're launching worker nodes that aren't healthy we stop the scaling and save costs.
```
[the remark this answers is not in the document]
suket22: What I meant is that there's no upper cap on how big your cluster can get. We cap how many worker nodes can be unhealthy at any point in time, but not the total size to which your cluster can grow. 

This is helpful in cases where all new worker nodes coming up are unhealthy and scale ups need to be stopped. However, if all nodes we provision continue to be healthy then there's no bound to the scaling.

designs/limits.md
```diff
@@ -0,0 +1,60 @@
...
+
+The two main problems with this approach though are -
+1. This limit while meant to constrain the number of unhealthy worker nodes in a cluster, will also inihibit the rate at which Karpenter can respond to pods that aren't schedulable. This somewhat goes against the goal of minimizing launch times of workers.
+2. While this helps ensure that costs don't increase due to runaway scaling, it won't help those who want a stricter cap on the amount of resources that's being provisioned even when nodes are otherwise healthy.
```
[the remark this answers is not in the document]
suket22: Yes - I tried to explain that a little more in my previous comment. I'll expand on this section a bit so it's clearer

designs/limits.md
```diff
@@ -0,0 +1,60 @@
...
+
+### **Absolute limit**
+
+This is a much simpler semantic to follow, where you provide an upper cap on the amount of resources being provisioned irrespective of their health. This is more in-line with alternative solutions like CA and should be easier to reason with.
+
+The actual limit of the number of resources being spun up can be defined either through something similar to resource requests or just a flat count of instances.
```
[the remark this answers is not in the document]
suket22: > Should we consider supporting number of pods as an upper limit?

I'm not sure how I'd be able to set that value reasonably. In a cluster I'd probably have multiple deployments running with HPA behind the scenes so it'll be difficult for me to keep this number up to date.  I also think since Karpenter provisions your infrastructure, it's limits being modeled as infra characteristics like cpu / mem / gpu would be a little more intuitive too. 

> Also, if we're going to do absolute limits should we break it out by capacity type?

Ooh this would be nice. I know we just made capacity-type a well known concept too so I wouldn't be against it. Although I suppose we could be even more generic where you could also break it down by other well known constraints like availability-zone? With AZs, we could kind of enforce Karpenter does some form of balancing across zones for higher availability. I wonder if this is too much complexity to start with?

designs/limits.md
suket22: Oh right, that makes a lot of sense. I don't think I considered what limits would mean with multiple provisioners. 
I like the idea of having a single global limit we pass directly to the controller binary. I'll add it in.

designs/limits.md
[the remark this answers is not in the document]
eptiger: I think the balancing is probably a bit advanced and maybe overloads this limits concept a bit, but I think factoring in having the limit be per capacity type or across all capacity types is still reasonable. Especially since a big part of why this limit will exist is to control costs and different capacity types can (and in AWS's case certainly do) have drastic cost implications.

designs/limits.md
suket22: We discussed this a little more in the working group and there were some other concerns with establishing a cluster-wide limit from @ellistarn -  like what if there's multiple karpenter installations etc. 

It sounds like we have a pretty compelling use case for having limits per provisioner but not so sure about the cluster-wide case just yet. Should we punt on that for now?

designs/limits.md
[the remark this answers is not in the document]
suket22: Yeah that's fair on the balancing bit. Following up on the conversation in the working group meeting though, this might be a little difficult to represent and use if we keep adding new dimensions to split the limits. Even with say capacity type, it might end up looking something like - 

```
  limits:
    resources:
        cpu: 1000 # irrespective of capacity type?
      capacity-type: on-demand
        cpu: 1000
      capacity-type: spot
        cpu: 500

or 

  limits:
    resources:
        cpu:
          capacity-types: 
             spot: 500
             on-demand: 1000
         mem:
  ....
```

And it'd start to get more complicated from there. I think to the other feedback that these limits are more to bound any scaling when something seems really off and isn't so much about precise cost control, I think we could probably leave this optimization out. Does that sound reasonable to you?

ellistarn (approved):

PR #803, opened 2021-11-13
Implemented support for selecting multiple provisioners

ellistarn:
TODO:
- Add tests for multiple provisioners


pkg/controllers/provisioning/provisioner.go
```diff
@@ -0,0 +1,120 @@
...
+	close(p.results)
+}
+
+func (p *Provisioner) provision(ctx context.Context) (err error) {
+	// Wait for a batch of pods
+	pods := p.Batch(ctx)
+	// Send results
+	defer func() {
+		for i := 0; i < len(pods); i++ {
+			p.results <- err
```
JacobGabrielson, line 78: Could you add a comment clarifying here? It's a bit hard to get my head around why we'd send the same `err` back along the channel for each pod. Also how are these errors "matched up" with the Pod(s) by whatever is pulling results off of this channel?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+		}
+	}
+}
+
+func (p *Provisioner) stop(ctx context.Context) {
+	logging.FromContext(ctx).Info("Stopping provisioner")
+	p.cancelFunc()
+	close(p.pods)
+	close(p.results)
```
JacobGabrielson, line 69: Is there any chance there could be a race condition here? I can't tell if (for example) `Enqueue` could be called after this `close` (on a different goroutine) and cause a panic?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -0,0 +1,120 @@
...
+	return nil
+}
+
+func (p *Provisioner) Enqueue(ctx context.Context, pod *v1.Pod) error {
+	p.pods <- pod
+	return <-p.results
+}
+
+func (p *Provisioner) Batch(ctx context.Context) (pods []*v1.Pod) {
+	logging.FromContext(ctx).Infof("Waiting for unschedulable pods")
```
JacobGabrielson, line 99: I'm wondering if there's enough info in this log message, in the case where there are multiple provisioners?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -0,0 +1,120 @@
...
+	pods = append(pods, <-p.pods)
```
JacobGabrielson, line 100: Just wondering ... do you really need this line? (seems like the `case pod...` line later on should cover it, unless there isn't at least one pod queued up on the channel, but in that case something else would be wrong?)

pkg/controllers/scheduling/controller.go
```diff
@@ -0,0 +1,178 @@
...
+	if provisioner == nil {
+		err := fmt.Errorf("matched 0/%d provisioners", len(multierr.Errors(errs)))
+		if errs != nil {
+			err = fmt.Errorf("%s, %w", err, errs)
+		}
+		return err
+	}
+	logging.FromContext(ctx).Debugf("Matched provisioner %q", provisioner.Name)
+	if err := provisioner.Enqueue(ctx, pod); err != nil {
+		return fmt.Errorf("provisioner %q failed", provisioner.Name)
```
JacobGabrielson, line 97: I think the error is needed here too?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	// Launch capacity and bind pods
+	if err := p.launcher.Launch(ctx, schedules, p.instanceTypes); err != nil {
+		return fmt.Errorf("launching capacity, %w", err)
+	}
+	return nil
+}
+
+func (p *Provisioner) Enqueue(ctx context.Context, pod *v1.Pod) error {
+	p.pods <- pod
```
JacobGabrielson, line 94: IIUC, this line will block until someone pulls the pod off of the channel. I'm wondering if this is really necessary, since on the very next line we're going to block on the `results` channel anyway? (Seems like it'll just slow things down?) But maybe I'm missing something.

pkg/controllers/allocation/binpacking/packer.go
```diff
@@ -97,14 +97,13 @@ func (p *Packer) Pack(ctx context.Context, schedule *scheduling.Schedule, instan
...
 				mainPack.Pods = append(mainPack.Pods, packing.Pods...)
-				logging.FromContext(ctx).Debugf("Incremented node count to %d on packing for %d pod(s) with instance type option(s) %v", mainPack.NodeQuantity, flattenedLen(packing.Pods...), instanceTypeNames(mainPack.InstanceTypeOptions))
 				continue
 			} else {
 				packs[key] = packing
 			}
 		}
 		packings = append(packings, packing)
-		logging.FromContext(ctx).Infof("Computed packing for %d pod(s) with instance type option(s) %s", flattenedLen(packing.Pods...), instanceTypeNames(packing.InstanceTypeOptions))
+		logging.FromContext(ctx).Infof("Computed packing of %d nodes for %d pod(s) with instance type option(s) %s", packing.NodeQuantity, flattenedLen(packing.Pods...), instanceTypeNames(packing.InstanceTypeOptions))
```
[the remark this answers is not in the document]
ellistarn: I've developed a habit of not being very surgical over the last year. There are some velocity benefits, but it does make it harder to review. Have I swung too far?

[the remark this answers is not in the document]
JacobGabrielson: Not a big deal, can just leave it in

pkg/controllers/provisioning/scheduler.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	if provisioner == nil {
+		err := fmt.Errorf("matched 0/%d provisioners", len(multierr.Errors(errs)))
+		if errs != nil {
+			err = fmt.Errorf("%s, %w", err, errs)
+		}
+		return err
+	}
+	logging.FromContext(ctx).Debugf("Scheduling pod to provisioner %q", provisioner.Name)
+	if err := provisioner.Enqueue(ctx, pod); err != nil {
```
JacobGabrielson, line 92: I think this line might still be dangerous (even after the change to use the `cancelFunc` as `Stop`) because it could write to the `pods` channel after `close()` was called successful (from a different goroutine) on the said channel.

IIUC, this call to Enqueue is happening on a different goroutine from the one the closes the channel. Specifically:

```go
// Enqueue a pod to the provisioner and block until it's processed
func (p *Provisioner) Enqueue(ctx context.Context, pod *v1.Pod) error {
	p.pods <- pod // <<<===scehduler goroutine?===<<<
	return <-p.results
}
```

 could could happen after this one:

```go
func (p *Provisioner) Start(ctx context.Context) {
	go func() {
		logging.FromContext(ctx).Info("Starting provisioner")
		for {
			select {
			case <-ctx.Done():
				logging.FromContext(ctx).Info("Stopping provisioner")
				close(p.pods) // <<<=== different goroutine from scheduler ===<<<
				close(p.results)
				return
			default:
				if err := p.Provision(ctx); err != nil {
					logging.FromContext(ctx).Errorf("Provisioning failed, %s", err.Error())
				}
			}
		}
	}()
}
```

According to https://go101.org/article/channel-closing.html one should normally refrain from closing a channel from the receiving side, so one possible solution would be to have the Provisioner *not* be the thing that closes the `pods` channel. But I haven't thought it through that much.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -49,14 +49,14 @@ func (c *Constraints) Supports(pod *v1.Pod) error {
 	podRequirements := PodRequirements(pod)
 	for _, key := range podRequirements.Keys() {
 		if c.Requirements.Requirement(key).Len() == 0 {
-			return fmt.Errorf("%s is too constrained", key)
+			return fmt.Errorf("invalid constraint %q, %v not in %v", key, podRequirements.Requirement(key).UnsortedList(), c.Requirements.Requirement(key).UnsortedList())
```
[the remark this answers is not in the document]
ellistarn: https://github.com/awslabs/karpenter/issues/812

pkg/controllers/provisioning/provisioner.go
JacobGabrielson: Nevermind, wasn't thinking that through :-)

pkg/controllers/provisioning/provisioner.go
```diff
@@ -0,0 +1,138 @@
...
+		return fmt.Errorf("solving scheduling constraints, %w", err)
+	}
+	// Launch capacity and bind pods
+	if err := p.launcher.Launch(ctx, schedules, p.instanceTypes); err != nil {
+		return fmt.Errorf("launching capacity, %w", err)
+	}
+	return nil
+}
+
+// Enqueue a pod to the provisioner and block until it's processed
```
JacobGabrielson, line 109: I'd probably rename this - `Enqueue` kind of a suggests a "fire and forget" type action (add something to a queue), whereas this is actually adding to a "queue" (channel), then waiting for it to be fully processed (i.e., gone from the channel).

pkg/controllers/provisioning/provisioner.go
ellistarn: This logger has already been decorated With("name", provisioner.Name)

pkg/controllers/provisioning/provisioner.go
ellistarn: I don't want to start the timer until the first pod kicks it off.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -0,0 +1,120 @@
...
+		default:
+			if err := p.provision(ctx); err != nil {
+				logging.FromContext(ctx).Errorf("Provisioning failed, %s", err.Error())
+			}
+		}
+	}
+}
+
+func (p *Provisioner) stop(ctx context.Context) {
+	logging.FromContext(ctx).Info("Stopping provisioner")
```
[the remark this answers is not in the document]
ellistarn: decorated

pkg/controllers/provisioning/controller.go
```diff
@@ -0,0 +1,123 @@
...
+	)
+	provisioner.Spec.Requirements = provisioner.Spec.Requirements.
+		With(scheduling.GlobalRequirements(instanceTypes)). // TODO(etarn) move GlobalRequirements to this file
+		With(v1alpha5.LabelRequirements(provisioner.Spec.Labels))
+
+	// Stop the existing provisioner if exists. This will drain the current
+	// workflow and replace it with an updated provisioner configuration.
+	// Requeue in order to discover any changes from GetInstanceTypes.
+	c.DeleteProvisioner(ctx, req)
+	c.CreateProvisioner(ctx, provisioner, instanceTypes)
```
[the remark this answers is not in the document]
ellistarn: We need *some* way to refresh the provisioner. I could use mutexes/channels to get updates from the provisioner controller, but it seemed a bit harder to reason about. Maybe we can discuss over voice?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -0,0 +1,120 @@
...
+	cancelFunc    context.CancelFunc
+
+	// Dependencies
+	scheduler *scheduling.Scheduler
+	launcher  *allocation.Launcher
+}
+
+// Start the provisioner's loop
+func (p *Provisioner) start(ctx context.Context) {
+	logging.FromContext(ctx).Info("Starting provisioner")
```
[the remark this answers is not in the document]
ellistarn: decorated

pkg/controllers/provisioning/provisioner.go
ellistarn: Good catch!

pkg/controllers/scheduling/controller.go
ellistarn: I trimmed it since its hugely verbose. The behavior is that the top level provisioner will fail, and then each of the pods that relied on it will say that it also failed.

pkg/controllers/provisioning/scheduler.go
ellistarn: Trying out a solution w/ context.Done() and select statements.

pkg/controllers/provisioning/provisioner.go
ellistarn: Agree.

JacobGabrielson (approved):

PR #810, opened 2021-11-17
move default provisioner to helm post-install hook

bwagner5:
**1. Issue, if available:**
N/A

 - When adding the default provisioner to the helm chart (https://github.com/awslabs/karpenter/pull/786), there is a race condition if you rely on defaulting from the webhook deployment since the provisioner resource is applied at the same time the webhook pod is coming online.  
 - This change moves the default provisioner to a helm `post-install-hook` resource. Helm resources applied in a post-install-hook are applied after the regular release resources are applied and are technically not part of the release. If you specify the `--wait` flag when helm installing, the post-install-hook will wait until release resources are in a Ready state. So, I have added readiness probes to both the controller and webhook deployments so that helm is signaled properly to begin the post-install-hook.  

- [ ] No

### DIFF
--- a/charts/karpenter/templates/controller/deployment.yaml
+++ b/charts/karpenter/templates/controller/deployment.yaml
@@ -48,6 +48,10 @@ spec:
             httpGet:
               path: /healthz
               port: 8081
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: 8081
           env:
             - name: CLUSTER_NAME
               value: {{ .Values.controller.clusterName }}
--- a/charts/karpenter/templates/controller/deployment.yaml
+++ b/charts/karpenter/templates/controller/deployment.yaml
@@ -50,7 +50,7 @@ spec:
               port: 8081
           readinessProbe:
             httpGet:
-              path: /healthz
+              path: /readyz
               port: 8081
           env:
             - name: CLUSTER_NAME
--- a/charts/karpenter/templates/default_provisioner.yaml
+++ b/charts/karpenter/templates/default_provisioner.yaml
@@ -3,6 +3,9 @@ apiVersion: karpenter.sh/v1alpha5
 kind: Provisioner
 metadata:
   name: default
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/resource-policy": keep
 spec:
   {{ if .Values.defaultProvisioner.ttlSecondsAfterEmpty }}
     ttlSecondsAfterEmpty: {{ .Values.defaultProvisioner.ttlSecondsAfterEmpty }}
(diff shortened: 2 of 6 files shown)


website/content/en/docs/getting-started/_index.md
```diff
@@ -156,10 +156,11 @@ eksctl. Thus, we don't need the helm chart to do that.
 ```bash
 helm repo add karpenter https://charts.karpenter.sh
 helm repo update
-helm upgrade --install --skip-crds karpenter karpenter/karpenter --namespace karpenter \
-  --create-namespace --set serviceAccount.create=false --version 0.4.1 \
+helm upgrade --install karpenter karpenter/karpenter --namespace karpenter \
+  --create-namespace --set serviceAccount.create=false --version 0.4.2 \
```
njtran, line 160: Can we do 0.4.3 here since that's the latest release?

charts/karpenter/templates/controller/deployment.yaml
```diff
@@ -48,6 +48,10 @@ spec:
             httpGet:
               path: /healthz
               port: 8081
+          readinessProbe:
```
ellistarn, line 51: Can we put this on the webhook, too?

charts/karpenter/templates/default_provisioner.yaml
```diff
@@ -3,6 +3,9 @@ apiVersion: karpenter.sh/v1alpha5
 kind: Provisioner
 metadata:
   name: default
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/resource-policy": keep
```
ellistarn, line 8: What do these do?

pkg/controllers/manager.go
```diff
@@ -50,7 +50,10 @@ func (m *GenericControllerManager) RegisterControllers(ctx context.Context, cont
 		}
 	}
 	if err := m.AddHealthzCheck("healthz", healthz.Ping); err != nil {
```
ellistarn, line 52: Can you call this livez (liveness)

ellistarn (commented):
hoooraaay

website/content/en/docs/getting-started/_index.md
bwagner5: oh man, where have I been. Yep

charts/karpenter/templates/default_provisioner.yaml
bwagner5: this tells helm that it's a post-install hook and to keep the resource rather than deleting it right after applying.

pkg/controllers/manager.go
bwagner5: I can't tell if this is a joke. If it's not, the convention is healthz for liveness check

ellistarn (approved):

PR #814, opened 2021-11-19
Fixed a bug where daemonsets weren't correctly included in binpacking calculations

ellistarn:
**1. Issue, if available:**

- [ ] No

### DIFF
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -113,7 +113,7 @@ func (s *Scheduler) getSchedules(ctx context.Context, constraints *v1alpha5.Cons
 	// schedule uniqueness is tracked by hash(Constraints)
 	schedules := map[uint64]*Schedule{}
 	for _, pod := range pods {
-		if err := constraints.Supports(pod); err != nil {
+		if err := constraints.ValidatePod(pod); err != nil {
 			logging.FromContext(ctx).Infof("Unable to schedule pod %s/%s, %s", pod.Name, pod.Namespace, err.Error())
 			continue
 		}
@@ -155,7 +155,7 @@ func (s *Scheduler) getDaemons(ctx context.Context, constraints *v1alpha5.Constr
 	pods := []*v1.Pod{}
 	for _, daemonSet := range daemonSetList.Items {
 		pod := &v1.Pod{Spec: daemonSet.Spec.Template.Spec}
-		if constraints.Supports(pod) == nil {
+		if constraints.ValidatePod(pod) == nil {
 			pods = append(pods, pod)
 		}
 	}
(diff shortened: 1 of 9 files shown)


pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -147,17 +147,15 @@ func (s *Scheduler) getSchedules(ctx context.Context, constraints *v1alpha5.Cons
...
 		return nil, fmt.Errorf("listing daemonsets, %w", err)
 	}
-
-	// 2. filter DaemonSets to include those that will schedule on this node
+	// Include daemonsets that will schedule on this node
 	pods := []*v1.Pod{}
 	for _, daemonSet := range daemonSetList.Items {
 		pod := &v1.Pod{Spec: daemonSet.Spec.Template.Spec}
-		if constraints.Supports(pod) != nil {
+		if constraints.Supports(pod) == nil {
```
JacobGabrielson, line 158: I think part of the issue here is that the name `Supports` (which I think reads most naturally as "supports?") suggests a boolean (true/false)? Is there some reason this has to return a nullable value?

pkg/controllers/provisioning/suite_test.go
```diff
@@ -134,33 +133,81 @@ var _ = Describe("Provisioning", func() {
...
+					test.PodOptions{
+						Tolerations:          []v1.Toleration{{Operator: v1.TolerationOperator(v1.NodeSelectorOpExists)}},
+						ResourceRequirements: v1.ResourceRequirements{Requests: v1.ResourceList{v1.ResourceCPU: resource.MustParse("1"), v1.ResourceMemory: resource.MustParse("1Gi")}},
+					},
+				))[0]
+				node := ExpectScheduled(ctx, env.Client, pod)
+				Expect(*node.Status.Allocatable.Cpu()).To(Equal(resource.MustParse("2")))
+				Expect(*node.Status.Allocatable.Memory()).To(Equal(resource.MustParse("2Gi")))
+			})
+			It("should ignore daemonsets an invalid selector", func() {
```
JacobGabrielson, line 178: missing word before `an`

pkg/controllers/provisioning/suite_test.go
```diff
@@ -134,33 +133,81 @@ var _ = Describe("Provisioning", func() {
...
+					test.PodOptions{
+						ResourceRequirements: v1.ResourceRequirements{Requests: v1.ResourceList{v1.ResourceCPU: resource.MustParse("1"), v1.ResourceMemory: resource.MustParse("1Gi")}},
+					},
+				))[0]
 				node := ExpectScheduled(ctx, env.Client, pod)
 				Expect(*node.Status.Allocatable.Cpu()).To(Equal(resource.MustParse("4")))
 				Expect(*node.Status.Allocatable.Memory()).To(Equal(resource.MustParse("4Gi")))
-			}
+			})
+			It("should not schedule if overhead is too large", func() {
```
JacobGabrielson, line 152: Is the reason why this test works that none of the instance types in the fake EC2 implementation have 10Gi of memory? I'd probably at least add a comment to that effect, especially considering this is the main test that tests out that functionality. Also, why not make the daemonset require way more memory and 10Gi just to make it more clear?

ellistarn: Good call.

pkg/controllers/provisioning/scheduling/scheduler.go
ellistarn: This is going to change a bit soon. I want to cover this in a refactor, but for now, I've renamed to ValidatePod

JacobGabrielson (approved):
🚢

PR #816, opened 2021-11-19
Implement ICE caching for AWS cloud provider and fix bug in filtering for SubnetProvider

eptiger:
**1. Issue, if available:**
https://github.com/awslabs/karpenter/issues/371

Finally - the exhilirating conclusion to my set of code changes on ice caching! All the changes I've made fit nicely within the AWS cloud provider
* I added a cache in the InstanceProvider. I debated a bunch with myself for where this should live and decided that it was cleaner to do that than to have the InstanceProvider pass the ICEs back to the cloudprovider to manage it. I liked the idea that the InstanceProvider would encapsulate the interactions with Fleet. 
* After a discussion with Ellis regarding my confusion on the constraints being passed into GetInstanceTypes, I removed the erroneous filtering in the SubnetProvider that was constraining AvailabilityZones (the only place the filtering was supposed to really happen was in the InstanceProvider when creating LT overrides for launch, so I moved the filtering into the InstanceProvider exclusively). I added brief comments to help others understand this better.
* I hope it's not too controversial that I added AvailabilityZone to our LT overrides. It's redundant but it makes our lives easier for processing ICEs and given that we're not near our Fleet request size limits it feels pretty safe.

I haven't done real cluster testing yet, only unit tests (but I believe Brandon has). Will do a sanity test on a real cluster before pushing. I also didn't add a unit test yet where we fall back to another zone - I may still try to add this when addressing the next round of comments.

- [x] No

### DIFF
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -132,11 +132,7 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context, constraints *v1alp
 	if err != nil {
 		return nil, apis.ErrGeneric(err.Error())
 	}
-	instanceTypes, err := c.instanceTypeProvider.Get(ctx, vendorConstraints)
-	if err != nil {
-		return nil, err
-	}
-	return c.instanceProvider.WithoutUnavailableOfferings(ctx, instanceTypes), nil
+	return c.instanceTypeProvider.Get(ctx, vendorConstraints)
 }
 
 func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
@@ -156,11 +152,11 @@ func (c *CloudProvider) Validate(ctx context.Context, constraints *v1alpha5.Cons
 func (c *CloudProvider) Default(ctx context.Context, constraints *v1alpha5.Constraints) {
 	vendorConstraints, err := v1alpha1.Deserialize(constraints)
 	if err != nil {
-		logging.FromContext(ctx).Fatalf("Failed to deserialize provider, %s", err.Error())
+		logging.FromContext(ctx).Errorf("Failed to deserialize provider, %s", err.Error())
 		return
 	}
 	vendorConstraints.Default(ctx)
 	if err := vendorConstraints.Serialize(constraints); err != nil {
-		logging.FromContext(ctx).Fatalf("Failed to serialize provider, %s", err.Error())
+		logging.FromContext(ctx).Errorf("Failed to serialize provider, %s", err.Error())
 	}
 }
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -24,7 +24,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
-	"github.com/aws/aws-sdk-go/service/ssm"
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
@@ -80,14 +79,8 @@ func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *Cloud
 	return &CloudProvider{
 		instanceTypeProvider: instanceTypeProvider,
 		subnetProvider:       subnetProvider,
-		instanceProvider: &InstanceProvider{ec2api, instanceTypeProvider, subnetProvider,
-			NewLaunchTemplateProvider(
-				ec2api,
-				NewAMIProvider(ssm.New(sess), options.ClientSet),
-				NewSecurityGroupProvider(ec2api),
-			),
-		},
-		creationQueue: parallel.NewWorkQueue(CreationQPS, CreationBurst),
+		instanceProvider:     NewInstanceProvider(ec2api, instanceTypeProvider, subnetProvider, sess, options.ClientSet),
+		creationQueue:        parallel.NewWorkQueue(CreationQPS, CreationBurst),
 	}
 }
 
@@ -132,12 +125,18 @@ func (c *CloudProvider) create(ctx context.Context, constraints *v1alpha5.Constr
 	return errs
 }
 
+// Despite accepting a Constraints struct, note that it does not utilize Requirements at all and instead
+// returns all available InstanceTypes.
 func (c *CloudProvider) GetInstanceTypes(ctx context.Context, constraints *v1alpha5.Constraints) ([]cloudprovider.InstanceType, error) {
 	vendorConstraints, err := v1alpha1.Deserialize(constraints)
 	if err != nil {
 		return nil, apis.ErrGeneric(err.Error())
 	}
-	return c.instanceTypeProvider.Get(ctx, vendorConstraints)
+	allInstanceTypes, err := c.instanceTypeProvider.Get(ctx, vendorConstraints)
+	if err != nil {
+		return nil, err
+	}
+	return c.instanceProvider.DiscardICEdInstanceTypes(ctx, allInstanceTypes), nil
 }
 
 func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -80,8 +80,14 @@ func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *Cloud
 	return &CloudProvider{
 		instanceTypeProvider: instanceTypeProvider,
 		subnetProvider:       subnetProvider,
-		instanceProvider:     NewInstanceProvider(ec2api, instanceTypeProvider, subnetProvider, ssm.New(sess), options.ClientSet),
-		creationQueue:        parallel.NewWorkQueue(CreationQPS, CreationBurst),
+		instanceProvider: &InstanceProvider{ec2api, instanceTypeProvider, subnetProvider,
+			NewLaunchTemplateProvider(
+				ec2api,
+				NewAMIProvider(ssm.New(sess), options.ClientSet),
+				NewSecurityGroupProvider(ec2api),
+			),
+		},
+		creationQueue: parallel.NewWorkQueue(CreationQPS, CreationBurst),
 	}
 }
 
@@ -132,7 +138,7 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context, constraints *v1alp
 	if err != nil {
 		return nil, apis.ErrGeneric(err.Error())
 	}
-	return c.instanceTypeProvider.Get(ctx, vendorConstraints)
+	return c.instanceTypeProvider.Get(ctx, vendorConstraints.AWS)
 }
 
 func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -24,6 +24,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ssm"
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
@@ -79,7 +80,7 @@ func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *Cloud
 	return &CloudProvider{
 		instanceTypeProvider: instanceTypeProvider,
 		subnetProvider:       subnetProvider,
-		instanceProvider:     NewInstanceProvider(ec2api, instanceTypeProvider, subnetProvider, sess, options.ClientSet),
+		instanceProvider:     NewInstanceProvider(ec2api, instanceTypeProvider, subnetProvider, ssm.New(sess), options.ClientSet),
 		creationQueue:        parallel.NewWorkQueue(CreationQPS, CreationBurst),
 	}
 }
@@ -125,18 +126,17 @@ func (c *CloudProvider) create(ctx context.Context, constraints *v1alpha5.Constr
 	return errs
 }
 
-// Despite accepting a Constraints struct, note that it does not utilize Requirements at all and instead
-// returns all available InstanceTypes.
+// GetInstanceTypes returns all available InstanceTypes despite accepting a Constraints struct (note that it does not utilize Requirements)
 func (c *CloudProvider) GetInstanceTypes(ctx context.Context, constraints *v1alpha5.Constraints) ([]cloudprovider.InstanceType, error) {
 	vendorConstraints, err := v1alpha1.Deserialize(constraints)
 	if err != nil {
 		return nil, apis.ErrGeneric(err.Error())
 	}
-	allInstanceTypes, err := c.instanceTypeProvider.Get(ctx, vendorConstraints)
+	instanceTypes, err := c.instanceTypeProvider.Get(ctx, vendorConstraints)
 	if err != nil {
 		return nil, err
 	}
-	return c.instanceProvider.DiscardICEdInstanceTypes(ctx, allInstanceTypes), nil
+	return c.instanceProvider.WithoutUnavailableOfferings(ctx, instanceTypes), nil
 }
 
 func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
@@ -156,11 +156,11 @@ func (c *CloudProvider) Validate(ctx context.Context, constraints *v1alpha5.Cons
 func (c *CloudProvider) Default(ctx context.Context, constraints *v1alpha5.Constraints) {
 	vendorConstraints, err := v1alpha1.Deserialize(constraints)
 	if err != nil {
-		logging.FromContext(ctx).Errorf("Failed to deserialize provider, %s", err.Error())
+		logging.FromContext(ctx).Fatalf("Failed to deserialize provider, %s", err.Error())
 		return
 	}
 	vendorConstraints.Default(ctx)
 	if err := vendorConstraints.Serialize(constraints); err != nil {
-		logging.FromContext(ctx).Errorf("Failed to serialize provider, %s", err.Error())
+		logging.FromContext(ctx).Fatalf("Failed to serialize provider, %s", err.Error())
 	}
 }
--- a/pkg/cloudprovider/aws/errors.go
+++ b/pkg/cloudprovider/aws/errors.go
@@ -25,8 +25,11 @@ var (
 		"InvalidInstanceID.NotFound",
 		"InvalidLaunchTemplateName.NotFoundException",
 	}
-	iceErrorCode = "InsufficientInstanceCapacity"
 )
+
+// InsufficientCapacityErrorCode indicates that EC2 is temporarily lacking capacity for this
+// instance type and availability zone combination
+const InsufficientCapacityErrorCode = "InsufficientInstanceCapacity"
 
 // isNotFound returns true if the err is an AWS error (even if it's
 // wrapped) and is a known to mean "not found" (as opposed to a more
@@ -37,8 +40,4 @@ func isNotFound(err error) bool {
 		return functional.ContainsString(notFoundErrorCodes, awsError.Code())
 	}
 	return false
-}
-
-func isInsufficientCapacityCode(errorCode string) bool {
-	return errorCode == iceErrorCode
 }
--- a/pkg/cloudprovider/aws/errors.go
+++ b/pkg/cloudprovider/aws/errors.go
@@ -25,6 +25,7 @@ var (
 		"InvalidInstanceID.NotFound",
 		"InvalidLaunchTemplateName.NotFoundException",
 	}
+	iceErrorCode = "InsufficientInstanceCapacity"
 )
 
 // isNotFound returns true if the err is an AWS error (even if it's
@@ -36,4 +37,8 @@ func isNotFound(err error) bool {
 		return functional.ContainsString(notFoundErrorCodes, awsError.Code())
 	}
 	return false
+}
+
+func isInsufficientCapacityCode(errorCode string) bool {
+	return errorCode == iceErrorCode
 }
--- a/pkg/cloudprovider/aws/fake/ec2api.go
+++ b/pkg/cloudprovider/aws/fake/ec2api.go
@@ -44,6 +44,7 @@ type EC2Behavior struct {
 	CalledWithCreateLaunchTemplateInput set.Set
 	Instances                           sync.Map
 	LaunchTemplates                     sync.Map
+	ShouldTriggerInsufficientCapacity   bool
 }
 
 type EC2API struct {
@@ -51,12 +52,20 @@ type EC2API struct {
 	EC2Behavior
 }
 
+// DefaultSupportedUsageClasses is a var because []*string can't be a const
+var DefaultSupportedUsageClasses = aws.StringSlice([]string{"on-demand", "spot"})
+
+const InsufficientCapacityInstanceType = "inf1.6xlarge"
+
 // Reset must be called between tests otherwise tests will pollute
 // each other.
 func (e *EC2API) Reset() {
 	e.EC2Behavior = EC2Behavior{
 		CalledWithCreateFleetInput:          set.NewSet(),
 		CalledWithCreateLaunchTemplateInput: set.NewSet(),
+		Instances:                           sync.Map{},
+		LaunchTemplates:                     sync.Map{},
+		ShouldTriggerInsufficientCapacity:   false,
 	}
 }
 
@@ -68,6 +77,9 @@ func (e *EC2API) CreateFleetWithContext(_ context.Context, input *ec2.CreateFlee
 	instances := []*ec2.Instance{}
 	instanceIds := []*string{}
 	for i := 0; i < int(*input.TargetCapacitySpecification.TotalTargetCapacity); i++ {
+		if e.ShouldTriggerInsufficientCapacity && aws.StringValue(input.LaunchTemplateConfigs[0].Overrides[0].InstanceType) == InsufficientCapacityInstanceType {
+			continue
+		}
 		instances = append(instances, &ec2.Instance{
 			InstanceId:     aws.String(randomdata.SillyName()),
 			Placement:      &ec2.Placement{AvailabilityZone: aws.String("test-zone-1a")},
@@ -78,7 +90,22 @@ func (e *EC2API) CreateFleetWithContext(_ context.Context, input *ec2.CreateFlee
 		instanceIds = append(instanceIds, instances[i].InstanceId)
 	}
 
-	return &ec2.CreateFleetOutput{Instances: []*ec2.CreateFleetInstance{{InstanceIds: instanceIds}}}, nil
+	result := &ec2.CreateFleetOutput{
+		Instances: []*ec2.CreateFleetInstance{{InstanceIds: instanceIds}}}
+	if e.ShouldTriggerInsufficientCapacity {
+		result.Errors = []*ec2.CreateFleetError{{
+			ErrorCode: aws.String("InsufficientInstanceCapacity"),
+			LaunchTemplateAndOverrides: &ec2.LaunchTemplateAndOverridesResponse{
+				LaunchTemplateSpecification: &ec2.FleetLaunchTemplateSpecification{
+					LaunchTemplateId:   input.LaunchTemplateConfigs[0].LaunchTemplateSpecification.LaunchTemplateId,
+					LaunchTemplateName: input.LaunchTemplateConfigs[0].LaunchTemplateSpecification.LaunchTemplateName},
+				Overrides: &ec2.FleetLaunchTemplateOverrides{
+					InstanceType:     aws.String(InsufficientCapacityInstanceType),
+					AvailabilityZone: aws.String("test-zone-1a"),
+					SubnetId:         aws.String("test-subnet-1")},
+			}}}
+	}
+	return result, nil
 }
 
 func (e *EC2API) CreateLaunchTemplateWithContext(_ context.Context, input *ec2.CreateLaunchTemplateInput, _ ...request.Option) (*ec2.CreateLaunchTemplateOutput, error) {
@@ -166,7 +193,7 @@ func (e *EC2API) DescribeInstanceTypesPagesWithContext(_ context.Context, _ *ec2
 		InstanceTypes: []*ec2.InstanceTypeInfo{
 			{
 				InstanceType:                  aws.String("m5.large"),
-				SupportedUsageClasses:         []*string{aws.String("on-demand"), aws.String("spot")},
+				SupportedUsageClasses:         DefaultSupportedUsageClasses,
 				SupportedVirtualizationTypes:  []*string{aws.String("hvm")},
 				BurstablePerformanceSupported: aws.Bool(false),
 				BareMetal:                     aws.Bool(false),
@@ -186,7 +213,7 @@ func (e *EC2API) DescribeInstanceTypesPagesWithContext(_ context.Context, _ *ec2
 			},
 			{
 				InstanceType:                  aws.String("m5.xlarge"),
-				SupportedUsageClasses:         []*string{aws.String("on-demand"), aws.String("spot")},
+				SupportedUsageClasses:         DefaultSupportedUsageClasses,
 				SupportedVirtualizationTypes:  []*string{aws.String("hvm")},
 				BurstablePerformanceSupported: aws.Bool(false),
 				BareMetal:                     aws.Bool(false),
@@ -206,7 +233,7 @@ func (e *EC2API) DescribeInstanceTypesPagesWithContext(_ context.Context, _ *ec2
 			},
 			{
 				InstanceType:                  aws.String("p3.8xlarge"),
-				SupportedUsageClasses:         []*string{aws.String("on-demand"), aws.String("spot")},
+				SupportedUsageClasses:         DefaultSupportedUsageClasses,
 				SupportedVirtualizationTypes:  []*string{aws.String("hvm")},
 				BurstablePerformanceSupported: aws.Bool(false),
 				BareMetal:                     aws.Bool(false),
@@ -232,7 +259,7 @@ func (e *EC2API) DescribeInstanceTypesPagesWithContext(_ context.Context, _ *ec2
 			},
 			{
 				InstanceType:                  aws.String("c6g.large"),
-				SupportedUsageClasses:         []*string{aws.String("on-demand"), aws.String("spot")},
+				SupportedUsageClasses:         DefaultSupportedUsageClasses,
 				SupportedVirtualizationTypes:  []*string{aws.String("hvm")},
 				BurstablePerformanceSupported: aws.Bool(false),
 				BareMetal:                     aws.Bool(false),
@@ -251,8 +278,33 @@ func (e *EC2API) DescribeInstanceTypesPagesWithContext(_ context.Context, _ *ec2
 				},
 			},
 			{
+				InstanceType:                  aws.String("inf1.2xlarge"),
+				SupportedUsageClasses:         DefaultSupportedUsageClasses,
+				SupportedVirtualizationTypes:  []*string{aws.String("hvm")},
+				BurstablePerformanceSupported: aws.Bool(false),
+				BareMetal:                     aws.Bool(false),
+				ProcessorInfo: &ec2.ProcessorInfo{
+					SupportedArchitectures: aws.StringSlice([]string{"x86_64"}),
+				},
+				VCpuInfo: &ec2.VCpuInfo{
+					DefaultVCpus: aws.Int64(8),
+				},
+				MemoryInfo: &ec2.MemoryInfo{
+					SizeInMiB: aws.Int64(16384),
+				},
+				InferenceAcceleratorInfo: &ec2.InferenceAcceleratorInfo{
+					Accelerators: []*ec2.InferenceDeviceInfo{{
+						Manufacturer: aws.String("AWS"),
+						Count:        aws.Int64(1),
+					}}},
+				NetworkInfo: &ec2.NetworkInfo{
+					MaximumNetworkInterfaces:  aws.Int64(4),
+					Ipv4AddressesPerInterface: aws.Int64(60),
+				},
+			},
+			{
 				InstanceType:                  aws.String("inf1.6xlarge"),
-				SupportedUsageClasses:         []*string{aws.String("on-demand"), aws.String("spot")},
+				SupportedUsageClasses:         DefaultSupportedUsageClasses,
 				SupportedVirtualizationTypes:  []*string{aws.String("hvm")},
 				BurstablePerformanceSupported: aws.Bool(false),
 				BareMetal:                     aws.Bool(false),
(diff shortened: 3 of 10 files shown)


pkg/cloudprovider/aws/instance.go
```diff
@@ -105,6 +128,37 @@ func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
...
+				}
+			} else if hasSpotIces {
+				if !spotIceZones.(sets.String).Has(offering.Zone) {
+					cleanedOfferings = append(cleanedOfferings, offering)
+				}
+			}
+		}
+		if len(cleanedOfferings) > 0 {
+			instanceType.AvailableOfferings = cleanedOfferings
+			cleanedInstanceTypes = append(cleanedInstanceTypes, &instanceTypes[index])
```
eptiger, line 156: This syntax seemed funky to me but doing I get a linting error if I just append `*instanceType` - there's certainly some Go idiosyncrasy I don't follow here.

pkg/cloudprovider/aws/instance.go
```diff
@@ -105,6 +128,37 @@ func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
 	return nil
 }
 
+// The intention is to remove Offerings from the provided possible InstanceTypes based on recently observed ICEs, which will
```
bwagner5, line 131: same here w/ go docs style, start with func name. 

```
// DiscardICEdInstanceTypes
```

Also, you could probably shorten to `DiscardICEd(ctx, instanceTypes)` since instanceTypes will be passed in.  The ICEd is awkward looking, so maybe we could do `DiscardInsufficientCapacityFor(ctx, instanceTypes)` ?  I'm not sure, naming is hard :p

pkg/cloudprovider/aws/instance.go
```diff
@@ -105,6 +128,37 @@ func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
...
+// indirectly impact the packer to keep it from spinning its wheels on packings that are doomed to fail due to EC2 capacity constraints.
+func (p *InstanceProvider) DiscardICEdInstanceTypes(ctx context.Context, instanceTypes []InstanceType) []cloudprovider.InstanceType {
+	var cleanedInstanceTypes = []cloudprovider.InstanceType{}
```
bwagner5, line 134: can remove `var` and just do 

```
cleanedInstanceTypes := []cloudprovider.InstanceType{}
```

pkg/cloudprovider/aws/instance.go
```diff
@@ -105,6 +128,37 @@ func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
...
+	for index, instanceType := range instanceTypes {
+		odIceZones, hasODIces := p.iceCache.Get(createIceCacheKey(instanceType.Name(), v1alpha1.CapacityTypeOnDemand))
+		spotIceZones, hasSpotIces := p.iceCache.Get(createIceCacheKey(instanceType.Name(), v1alpha1.CapacityTypeSpot))
+		if !hasODIces && !hasSpotIces {
+			cleanedInstanceTypes = append(cleanedInstanceTypes, &instanceTypes[index])
```
bwagner5, line 139: why are you grabbing the instanceType from the slice using an index here but below you are using the iterator variable below for offering?  I assumed it was to get around the iterator loop variable from having the value overridden but it would still point to the same memory address?  Does that not apply below in the offering loop?

pkg/cloudprovider/aws/instance.go
bwagner5: I believe this is the issue I mentioned above where the range variable in a for loop uses one memory address and overrides it for each iteration. So you'd need to set another temp variable in the loop, and take the address of it when appending.

pkg/cloudprovider/aws/instance.go
```diff
@@ -105,6 +128,37 @@ func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
...
+			continue
+		}
+		cleanedOfferings := []cloudprovider.Offering{}
+		for _, offering := range instanceType.Offerings() {
+			if hasODIces && offering.CapacityType == v1alpha1.CapacityTypeOnDemand {
```
bwagner5, line 144: if this short circuits (hasODIces returns false), won't this go to the `else If` and add an OD offering?

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -80,14 +79,8 @@ func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *Cloud
...
 		subnetProvider:       subnetProvider,
-		instanceProvider: &InstanceProvider{ec2api, instanceTypeProvider, subnetProvider,
-			NewLaunchTemplateProvider(
-				ec2api,
-				NewAMIProvider(ssm.New(sess), options.ClientSet),
-				NewSecurityGroupProvider(ec2api),
-			),
-		},
-		creationQueue: parallel.NewWorkQueue(CreationQPS, CreationBurst),
+		instanceProvider:     NewInstanceProvider(ec2api, instanceTypeProvider, subnetProvider, sess, options.ClientSet),
```
ellistarn, line 82: I love that you've encapsulated the LT provider inside the instance provider. Do you mind instantiating an SSM client out here and passing it into NewInstanceProvider instead of sess? This reduces the scope of what lower level constructors are able to instantiate and makes it more clear what services the cloud provider relies on by just looking at this construct.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -132,12 +125,18 @@ func (c *CloudProvider) create(ctx context.Context, constraints *v1alpha5.Constr
 	return errs
 }
 
+// Despite accepting a Constraints struct, note that it does not utilize Requirements at all and instead
```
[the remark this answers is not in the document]
ellistarn: The key bit being that exported methods should typically have comments that start with the name of the exported thing

```
// Foo is definitely not a bar
type Foo struct

// GetBar for the foo
func (f *Foo) GetBar()
```

Though as we've discussed, sometimes for very simple/self-documented things, you can omit the comment entirely.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -132,12 +125,18 @@ func (c *CloudProvider) create(ctx context.Context, constraints *v1alpha5.Constr
...
 	vendorConstraints, err := v1alpha1.Deserialize(constraints)
 	if err != nil {
 		return nil, apis.ErrGeneric(err.Error())
 	}
-	return c.instanceTypeProvider.Get(ctx, vendorConstraints)
+	allInstanceTypes, err := c.instanceTypeProvider.Get(ctx, vendorConstraints)
+	if err != nil {
+		return nil, err
+	}
+	return c.instanceProvider.DiscardICEdInstanceTypes(ctx, allInstanceTypes), nil
```
ellistarn, line 139: Given that this is the only place that this is used, does it make sense to hide this behavior inside the instance provider? That is to say, instanceTypeProvider.Get() inherently factors in unavailable offerings. Is there ever a case where we want to include unavailable offerings on this call?

pkg/cloudprovider/aws/errors.go
```diff
@@ -25,6 +25,7 @@ var (
 		"InvalidInstanceID.NotFound",
 		"InvalidLaunchTemplateName.NotFoundException",
 	}
+	iceErrorCode = "InsufficientInstanceCapacity"
```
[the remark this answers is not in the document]
ellistarn: Can we spell out ICE to remove jargon? `InsufficientCapacityErrorCode`

pkg/cloudprovider/aws/instance.go
```diff
@@ -143,6 +197,13 @@ func (p *InstanceProvider) launchInstances(ctx context.Context, constraints *v1a
 	if err != nil {
 		return nil, fmt.Errorf("creating fleet %w", err)
 	}
+	if len(createFleetOutput.Errors) > 0 {
+		for _, fleetError := range createFleetOutput.Errors {
+			if isInsufficientCapacityCode(*fleetError.ErrorCode) {
```
ellistarn, line 202: use aws.StringValue(fleetError.ErrorCode) to avoid NPE if AWS ever serves us a bad response.

pkg/cloudprovider/aws/instance.go
```diff
@@ -36,11 +40,30 @@ import (
 	"github.com/awslabs/karpenter/pkg/utils/options"
 )
 
+const (
+	iceCacheODTTL           = 15 * time.Second
```
ellistarn, line 44:
```suggestion
	InsufficientCapacityExceptionCacheOnDemandTTL           = 15 * time.Second
```

pkg/cloudprovider/aws/instance.go
```diff
@@ -36,11 +40,30 @@ import (
...
+
+func NewInstanceProvider(ec2api ec2iface.EC2API, instanceTypeProvider *InstanceTypeProvider, subnetProvider *SubnetProvider, sess *session.Session,
+	clientSet *kubernetes.Clientset) *InstanceProvider {
+	return &InstanceProvider{ec2api, instanceTypeProvider, subnetProvider,
+		NewLaunchTemplateProvider(
+			ec2api,
+			NewAMIProvider(ssm.New(sess), clientSet),
+			NewSecurityGroupProvider(ec2api),
+		),
+		cache.New(iceCacheSpotTTL, iceCacheCleanupInterval),
```
ellistarn, line 65: Seems a bit odd to me to use the spot TTL since it happens to be lower. WDYT about setting to -1 to disable eviction? It's a finite set, so I'm not concerned about cache size.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -132,12 +125,18 @@ func (c *CloudProvider) create(ctx context.Context, constraints *v1alpha5.Constr
...
 
+// Despite accepting a Constraints struct, note that it does not utilize Requirements at all and instead
+// returns all available InstanceTypes.
 func (c *CloudProvider) GetInstanceTypes(ctx context.Context, constraints *v1alpha5.Constraints) ([]cloudprovider.InstanceType, error) {
 	vendorConstraints, err := v1alpha1.Deserialize(constraints)
 	if err != nil {
 		return nil, apis.ErrGeneric(err.Error())
 	}
-	return c.instanceTypeProvider.Get(ctx, vendorConstraints)
+	allInstanceTypes, err := c.instanceTypeProvider.Get(ctx, vendorConstraints)
```
ellistarn, line 135: consider simplifying to:
```suggestion
	instanceTypes, err := c.instanceTypeProvider.Get(ctx, vendorConstraints)
```
"all" can be inferred from context.

pkg/cloudprovider/aws/instance.go
```diff
@@ -105,6 +128,37 @@ func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
 	return nil
 }
 
+// The intention is to remove Offerings from the provided possible InstanceTypes based on recently observed ICEs, which will
+// indirectly impact the packer to keep it from spinning its wheels on packings that are doomed to fail due to EC2 capacity constraints.
+func (p *InstanceProvider) DiscardICEdInstanceTypes(ctx context.Context, instanceTypes []InstanceType) []cloudprovider.InstanceType {
```
ellistarn, line 133: naming suggestion:
```suggestion
func (p *InstanceProvider) WithoutUnavailableOfferings(ctx context.Context, instanceTypes []InstanceType) []InstanceType {
```
It also might be a bit more idiomatic to do type conversion at the cloud provider interface and just think in terms of aws.InstanceTypes here.

pkg/cloudprovider/aws/instance.go
```diff
@@ -143,6 +197,13 @@ func (p *InstanceProvider) launchInstances(ctx context.Context, constraints *v1a
 	if err != nil {
 		return nil, fmt.Errorf("creating fleet %w", err)
 	}
+	if len(createFleetOutput.Errors) > 0 {
```
ellistarn, line 200: You can skip this check, since the for loop will not execute if the length == 0.

pkg/cloudprovider/aws/instance.go
```diff
@@ -143,6 +197,13 @@ func (p *InstanceProvider) launchInstances(ctx context.Context, constraints *v1a
...
+		for _, fleetError := range createFleetOutput.Errors {
```
ellistarn, line 201:
```suggestion
		for _, err := range createFleetOutput.Errors {
```
In go, it's common to shadow `err` in nested scopes.

pkg/cloudprovider/aws/instance.go
```diff
@@ -262,6 +327,22 @@ func (p *InstanceProvider) instanceToNode(instance *ec2.Instance, instanceTypes
 	return nil, fmt.Errorf("unrecognized instance type %s", aws.StringValue(instance.InstanceType))
 }
 
+func (p *InstanceProvider) addToIceCache(overrides *ec2.FleetLaunchTemplateOverrides, capacityType string) {
+	instanceType := aws.StringValue(overrides.InstanceType)
```
ellistarn, line 331: consider inlining if only used once.

pkg/cloudprovider/aws/instance.go
```diff
@@ -262,6 +327,22 @@ func (p *InstanceProvider) instanceToNode(instance *ec2.Instance, instanceTypes
...
+	zone := aws.StringValue(overrides.AvailabilityZone)
```
ellistarn, line 332: likewise

pkg/cloudprovider/aws/instance.go
```diff
@@ -304,3 +385,7 @@ func combineReservations(reservations []*ec2.Reservation) []*ec2.Instance {
 	}
 	return instances
 }
+
+func createIceCacheKey(instanceType string, capacityType string) string {
```
ellistarn, line 389:
```suggestion
func iceCacheKey(instanceType string, capacityType string) string {
```

pkg/cloudprovider/aws/instance.go
```diff
@@ -304,3 +385,7 @@ func combineReservations(reservations []*ec2.Reservation) []*ec2.Instance {
...
+	return fmt.Sprintf("%s-%s", instanceType, capacityType)
```
ellistarn, line 390: I think od == on-demand, so dash feels a bit odd.
```suggestion
	return fmt.Sprintf("%s/%s", instanceType, capacityType)
```

pkg/cloudprovider/aws/instance.go
```diff
@@ -36,11 +40,30 @@ import (
...
+	iceCacheSpotTTL         = 45 * time.Second
+	iceCacheCleanupInterval = 5 * time.Minute
+)
+
 type InstanceProvider struct {
 	ec2api                 ec2iface.EC2API
 	instanceTypeProvider   *InstanceTypeProvider
 	subnetProvider         *SubnetProvider
 	launchTemplateProvider *LaunchTemplateProvider
+	iceCache               *cache.Cache
```
ellistarn, line 54:
```suggestion
	offeringAvailability               *cache.Cache
```

pkg/cloudprovider/aws/instance.go
```diff
@@ -176,7 +237,7 @@ func (p *InstanceProvider) getLaunchTemplateConfigs(ctx context.Context, constra
 	return launchTemplateConfigs, nil
 }
 
-func (p *InstanceProvider) getOverrides(instanceTypeOptions []cloudprovider.InstanceType, subnets []*ec2.Subnet, capacityType string) []*ec2.FleetLaunchTemplateOverridesRequest {
+func (p *InstanceProvider) getOverrides(instanceTypeOptions []cloudprovider.InstanceType, subnets []*ec2.Subnet, constrainedZones sets.String, capacityType string) []*ec2.FleetLaunchTemplateOverridesRequest {
```
ellistarn, line 240:
```suggestion
func (p *InstanceProvider) getOverrides(instanceTypeOptions []cloudprovider.InstanceType, subnets []*ec2.Subnet, zones sets.String, capacityType string) []*ec2.FleetLaunchTemplateOverridesRequest {
```

pkg/cloudprovider/aws/instance.go
```diff
@@ -185,10 +246,14 @@ func (p *InstanceProvider) getOverrides(instanceTypeOptions []cloudprovider.Inst
 				continue
 			}
 			for _, subnet := range subnets {
-				if aws.StringValue(subnet.AvailabilityZone) == offering.Zone {
+				subnetZone := aws.StringValue(subnet.AvailabilityZone)
+				if subnetZone == offering.Zone && (constrainedZones == nil || constrainedZones.Has(subnetZone)) {
```
ellistarn, line 250: I think you're guaranteed that constrainedZones won't be nil, since Karpenter has to produce the full set of viable zones for the requirements `NotIn` operator to work. e.g. `us-west-2a` not in `nil` is impossible to compute.
```suggestion
				if zone := aws.StringValue(subnet.AvailabilityZone); zone == offering.Zone && zones.has(zone) {
```

pkg/cloudprovider/aws/instance.go
```diff
@@ -185,10 +246,14 @@ func (p *InstanceProvider) getOverrides(instanceTypeOptions []cloudprovider.Inst
...
 					override := &ec2.FleetLaunchTemplateOverridesRequest{
 						InstanceType: aws.String(instanceType.Name()),
 						SubnetId:     subnet.SubnetId,
+						// This is technically redundant, but is useful if we have to parse ICEs from CreateFleet
```
ellistarn, line 254: Nice comment!

pkg/cloudprovider/aws/instance.go
```diff
@@ -262,6 +327,22 @@ func (p *InstanceProvider) instanceToNode(instance *ec2.Instance, instanceTypes
...
+	cacheTTL := iceCacheSpotTTL
+	if capacityType == v1alpha1.CapacityTypeOnDemand {
+		cacheTTL = iceCacheODTTL
+	}
+	cacheKey := createIceCacheKey(instanceType, capacityType)
+	if zones, exists := p.iceCache.Get(cacheKey); exists {
+		zones.(sets.String).Insert(zone)
+		p.iceCache.Set(cacheKey, zones, cacheTTL)
+	} else {
+		p.iceCache.Set(cacheKey, sets.String{zone: sets.Empty{}}, cacheTTL)
```
ellistarn, line 342: Don't you need to include the zone in the cache? 
```suggestion
		p.iceCache.Set(cacheKey, sets.String{zone: sets.NewString(zones)}, cacheTTL)
```

pkg/cloudprovider/aws/instance.go
```diff
@@ -262,6 +327,22 @@ func (p *InstanceProvider) instanceToNode(instance *ec2.Instance, instanceTypes
...
+func (p *InstanceProvider) addToIceCache(overrides *ec2.FleetLaunchTemplateOverrides, capacityType string) {
+	instanceType := aws.StringValue(overrides.InstanceType)
+	zone := aws.StringValue(overrides.AvailabilityZone)
+	cacheTTL := iceCacheSpotTTL
+	if capacityType == v1alpha1.CapacityTypeOnDemand {
+		cacheTTL = iceCacheODTTL
+	}
+	cacheKey := createIceCacheKey(instanceType, capacityType)
+	if zones, exists := p.iceCache.Get(cacheKey); exists {
+		zones.(sets.String).Insert(zone)
```
ellistarn, line 339: Because sets.String is a pointer (the underlying type is a map, which is a pointer), you should be able to just read and inject the zone and not call cache.Set() again

pkg/cloudprovider/aws/instance.go
```diff
@@ -262,6 +327,22 @@ func (p *InstanceProvider) instanceToNode(instance *ec2.Instance, instanceTypes
...
+		p.iceCache.Set(cacheKey, zones, cacheTTL)
```
ellistarn, line 340: Thinking about this, I think there might be a weird case
- zone-a is added to the cache, 5m
- 4 minutes pass 
- zone-b is added to the cache, 5m
- both zone-a and zone-b are now cached for 5m
- 1 minute passes
- zone-a might be available again.

You could potentially store a timestamp in the cache to keep track of this, but it feels a bit complicated to me. It's probably fine in the grand scheme. Happy to discuss this more.

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -75,6 +76,7 @@ var _ = BeforeSuite(func() {
 					securityGroupProvider: NewSecurityGroupProvider(fakeEC2API),
 					cache:                 launchTemplateCache,
 				},
+				cache.New(5*time.Second, 30*time.Second),
```
ellistarn, line 79: If you export these values you can reference them directly in the tests. Assuming they're not different.

ellistarn: You'll also likely need to use injectable.Time to test this, though.

pkg/cloudprovider/aws/instance.go
eptiger: You are right, I need to reverse the order here because the else if should only happen if we know it's not od. However, you helped me realize another important issue. If there's no spice at all for this instance type and yet this is a spot offering I need to add the offering to the cleaned slice (yet I don't if the cache has an od ice for this instance type). In trying to be too clever with the conditionals I've created this problem.

pkg/cloudprovider/aws/instance.go
bwagner5: I think the spot and OD cache TTLs should be the same 45 seconds.

eptiger (commented):
About to post the next update, just give me a few minutes to run through the checks.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -157,11 +156,11 @@ func (c *CloudProvider) Validate(ctx context.Context, constraints *v1alpha5.Cons
 func (c *CloudProvider) Default(ctx context.Context, constraints *v1alpha5.Constraints) {
 	vendorConstraints, err := v1alpha1.Deserialize(constraints)
 	if err != nil {
-		logging.FromContext(ctx).Errorf("Failed to deserialize provider, %s", err.Error())
+		logging.FromContext(ctx).Fatalf("Failed to deserialize provider, %s", err.Error())
```
ellistarn, line 159: I think we want to keep this as Errorf

pkg/cloudprovider/aws/instance.go
```diff
@@ -24,23 +24,51 @@ import (
...
 
+const (
+	InsufficientCapacityErrorCacheOnDemandTTL     = 15 * time.Second
+	InsufficientCapacityErrorCacheSpotTTL         = 45 * time.Second
+	InsufficientCapacityErrorCacheCleanupInterval = 5 * time.Minute
+)
+
+type CachedZones struct {
+	zones            sets.String
+	initialCacheTime time.Time
```
ellistarn, line 51: Optional: simplify to time. I think it's obvious in context.

pkg/cloudprovider/aws/instance.go
```diff
@@ -143,6 +205,13 @@ func (p *InstanceProvider) launchInstances(ctx context.Context, constraints *v1a
 	if err != nil {
 		return nil, fmt.Errorf("creating fleet %w", err)
 	}
+	insufficientCapacityOfferings := map[string]sets.String{}
+	for _, err := range createFleetOutput.Errors {
+		if InsufficientCapacityErrorCode == aws.StringValue(err.ErrorCode) {
+			createOrAppendToMapValue(insufficientCapacityOfferings, aws.StringValue(err.LaunchTemplateAndOverrides.Overrides.InstanceType), aws.StringValue(err.LaunchTemplateAndOverrides.Overrides.AvailabilityZone))
+		}
+	}
```
ellistarn, line 213: Thoughts on encapsulating this inside `updateUnavailableOfferingsCache`. This method is already pretty beefy.
```
for _, err := range createFleetOutput.Errors {
   p.cacheUnavailableOfferings(ctx, err, capacityType)
}
```
or
``` 
if createFleetOutput.Errors != nil {
  p.cacheUnavailableOfferings(ctx, createFleetOutput.Errors, capacityType)
}
```

pkg/cloudprovider/aws/instance.go
```diff
@@ -185,10 +254,14 @@ func (p *InstanceProvider) getOverrides(instanceTypeOptions []cloudprovider.Inst
 				continue
 			}
 			for _, subnet := range subnets {
-				if aws.StringValue(subnet.AvailabilityZone) == offering.Zone {
+				subnetZone := aws.StringValue(subnet.AvailabilityZone)
+				if subnetZone == offering.Zone && constrainedZones.Has(subnetZone) {
```
ellistarn, line 258: Thoughts on shortcircuiting zonal matches similar to how we do w/ capacity type on 253

```
if offering.CapacityType != capacityType {
  continue
}
if !constrainedZones.Has(offering.Zone) {
  continue
}
for _, subnet := range subnets {
  if  offering.Zone == aws.StringValue(subnet.AvailabilityZone) {
    override = ...
  }
}
```

pkg/cloudprovider/aws/instance.go
ellistarn: There's a simplification below that *might* help with this, but I'm happy either way.

pkg/cloudprovider/aws/cloudprovider.go
eptiger: Yes, I think a bad rebase as I didn't touch this code.

pkg/cloudprovider/aws/instance.go
```diff
@@ -43,6 +45,17 @@ type InstanceProvider struct {
 	launchTemplateProvider *LaunchTemplateProvider
 }
 
+func NewInstanceProvider(ec2api ec2iface.EC2API, instanceTypeProvider *InstanceTypeProvider, subnetProvider *SubnetProvider, ssm *ssm.SSM,
```
eptiger, line 48: I left this in since i had already created it - lmk if you'd rather me revert this.

pkg/cloudprovider/aws/instance.go
```diff
@@ -185,21 +201,26 @@ func (p *InstanceProvider) getOverrides(instanceTypeOptions []cloudprovider.Inst
...
-					// to reduce the likelihood of getting an excessively large instance type.
-					// instanceTypeOptions are sorted by vcpus and memory so this prioritizes smaller instance types.
-					if capacityType == v1alpha1.CapacityTypeSpot {
-						override.Priority = aws.Float64(float64(i))
-					}
-					overrides = append(overrides, override)
-					// FleetAPI cannot span subnets from the same AZ, so break after the first one.
-					break
+				subnetZone := aws.StringValue(subnet.AvailabilityZone)
+				if subnetZone != offering.Zone || !zones.Has(offering.Zone) {
```
bwagner5, line 205: you can in-line `subnetZone` 

```
if aws.StringValue(subnet.AvailabilityZone) != offering.Zone || !zone.Has(offering.Zone) {
...
```

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -165,16 +170,67 @@ var _ = Describe("Allocation", func() {
...
+				// it should've tried to pack them on a single inf1.6xlarge then hit an insufficient capacity error
+				for _, pod := range pods {
+					ExpectNotScheduled(ctx, env.Client, pod)
 				}
+				// capacity shortage is over - wait for expiry (N.B. the Karpenter logging will not show the overridden cache expiry in this test context)
+				fakeEC2API.ShouldTriggerInsufficientCapacity = false
+				Eventually(func(g Gomega) int {
+					nodeNames := sets.NewString()
+					for _, pod := range ExpectProvisioned(ctx, env.Client, scheduler, provisioners, provisioner, pods...) {
+						nodeNames = nodeNames.Insert(ExpectScheduledWithInstanceTypeAndGomega(ctx, env.Client, pod, "inf1.6xlarge", g).Name)
```
eptiger, line 230: Just preempting questions here - this is a brand new feature of Gomega, which makes this `Eventually` function a lot less complicated than it could've been: https://onsi.github.io/gomega/#eventually

pkg/cloudprovider/aws/fake/ec2api.go
```diff
@@ -44,19 +44,28 @@ type EC2Behavior struct {
 	CalledWithCreateLaunchTemplateInput set.Set
 	Instances                           sync.Map
 	LaunchTemplates                     sync.Map
+	ShouldTriggerInsufficientCapacity   bool
```
ellistarn, line 47: Thoughts on collapsing this and InsufficientCapacityInstanceType into a single configurable concept
`InsufficientCapacityExceptionInstanceTypes sets.Set`

pkg/cloudprovider/aws/fake/ec2api.go
```diff
@@ -68,6 +77,9 @@ func (e *EC2API) CreateFleetWithContext(_ context.Context, input *ec2.CreateFlee
 	instances := []*ec2.Instance{}
 	instanceIds := []*string{}
 	for i := 0; i < int(*input.TargetCapacitySpecification.TotalTargetCapacity); i++ {
+		if e.ShouldTriggerInsufficientCapacity && aws.StringValue(input.LaunchTemplateConfigs[0].Overrides[0].InstanceType) == InsufficientCapacityInstanceType {
```
ellistarn, line 80: Then this would be if `e.InsufficientCapacityExceptionInstanceTypes.Has(aws.StringValue(input.LaunchTemplateConfigs[0].Overrides[0].InstanceType))`

pkg/cloudprovider/aws/instance.go
ellistarn: Either way. I hate constructors that span many lines -- in a sense, explicit object creation in go is very similar to the builder pattern.

pkg/cloudprovider/aws/instance.go
ellistarn, line 205: Does it make sense to put the check
```
if !zones.Has(offering.Zone) {
  continue
}
```
 on line 203?

ellistarn: Ofc we need to still do the check  `if subnetZone != offering.Zone { continue }` here

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -31,34 +31,41 @@ import (
...
+		ec2api:               ec2api,
+		subnetProvider:       subnetProvider,
+		cache:                cache.New(InstanceTypesAndZonesCacheTTL, CacheCleanupInterval),
+		unavailableOfferings: cache.New(InsufficientCapacityErrorCacheTTL, InsufficientCapacityErrorCacheCleanupInterval),
 	}
 }
 
-// Get instance type options given the constraints
+// Get all instance type options (the constraints are only used for tag filtering on subnets, not for Requirements filtering)
 func (p *InstanceTypeProvider) Get(ctx context.Context, constraints *v1alpha1.Constraints) ([]cloudprovider.InstanceType, error) {
```
ellistarn, line 61: Is it possible to do
```suggestion
func (p *InstanceTypeProvider) Get(ctx context.Context, provider *v1alpha1.AWS) ([]cloudprovider.InstanceType, error) {
```

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -153,3 +168,25 @@ func (p *InstanceTypeProvider) filter(instanceType *ec2.InstanceTypeInfo) bool {
 		"p", "inf", "g", // Accelerators
 	)
 }
+
+// TrackUnavailableOfferings allows the InstanceProvider to communicate recently observed temporary capacity shortages in
+// the provided offerings
+func (p *InstanceTypeProvider) TrackUnavailableOfferings(ctx context.Context, offerings map[string]sets.String, capacityType string) {
```
ellistarn, line 174:
```suggestion
func (p *InstanceTypeProvider) MarkUnavailable(ctx context.Context, instanceType string, zone string, capacityType) {
   p.unavailableOfferings.SetDefault(fmt.Sprintf("%s:%s:%s", instanceType, zone, capacityType)
}
```

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -153,3 +168,25 @@ func (p *InstanceTypeProvider) filter(instanceType *ec2.InstanceTypeInfo) bool {
...
+				capacityType,
+				InsufficientCapacityErrorCacheTTL)
+			// even if the key is already in the cache, we still need to call Set to extend the cached entry's TTL
+			p.unavailableOfferings.SetDefault(cacheKey, struct{}{})
+		}
+	}
+}
+
+func unavailableOfferingsCacheKey(capacityType string, instanceType string, zone string) string {
+	return fmt.Sprintf("%s:%s:%s", capacityType, instanceType, zone)
```
ellistarn, line 191: var UnavailableOfferingsCacheKeyFormat = "%s:%s:%s"

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -153,3 +168,21 @@ func (p *InstanceTypeProvider) filter(instanceType *ec2.InstanceTypeInfo) bool {
 		"p", "inf", "g", // Accelerators
 	)
 }
+
+// CacheUnavailable allows the InstanceProvider to communicate recently observed temporary capacity shortages in
+// the provided offerings
+func (p *InstanceTypeProvider) CacheUnavailable(ctx context.Context, instanceType string, zone string, capacityType string) {
+	cacheKey := UnavailableOfferingsCacheKey(capacityType, instanceType, zone)
```
ellistarn, line 175: Optional: consider inlining since it's used once.

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -153,3 +168,21 @@ func (p *InstanceTypeProvider) filter(instanceType *ec2.InstanceTypeInfo) bool {
...
+	logging.FromContext(ctx).Debugf("Saw %s for offering { instanceType: %s, zone: %s, capacityType: %s }, avoiding for %s",
+		InsufficientCapacityErrorCode,
+		instanceType,
+		zone,
+		capacityType,
+		InsufficientCapacityErrorCacheTTL)
+	// even if the key is already in the cache, we still need to call Set to extend the cached entry's TTL
+	p.unavailableOfferings.SetDefault(cacheKey, struct{}{})
```
ellistarn, line 183: Consider aliasing something like 

var EmptyValue = struct{}{}

pkg/cloudprovider/aws/subnets.go
```diff
@@ -61,15 +61,9 @@ func (s *SubnetProvider) Get(ctx context.Context, constraints *v1alpha1.Constrai
...
+func getFilters(constraints *v1alpha1.AWS) []*ec2.Filter {
 	filters := []*ec2.Filter{}
-	// Filter by zone
-	if zones := constraints.Requirements.Zones(); zones != nil {
-		filters = append(filters, &ec2.Filter{
-			Name:   aws.String("availability-zone"),
-			Values: aws.StringSlice(zones.UnsortedList()),
-		})
-	}
+
```
ellistarn, line 66:
```suggestion
```

pkg/test/expectations/expectations.go
```diff
@@ -65,10 +65,19 @@ func ExpectNotFound(c client.Client, objects ...client.Object) {
 
 func ExpectScheduled(ctx context.Context, c client.Client, pod *v1.Pod) *v1.Node {
 	p := ExpectPodExists(c, pod.Name, pod.Namespace)
-	Expect(p.Spec.NodeName).ToNot(BeEmpty(), fmt.Sprintf("expected %s/%s to scheduled", pod.Namespace, pod.Name))
+	Expect(p.Spec.NodeName).ToNot(BeEmpty(), fmt.Sprintf("expected %s/%s to be scheduled", pod.Namespace, pod.Name))
 	return ExpectNodeExists(c, p.Spec.NodeName)
 }
 
+func ExpectScheduledWithInstanceType(ctx context.Context, c client.Client, pod *v1.Pod, instanceType string) *v1.Node {
```
ellistarn, line 72: I think this is probably a bit too specific for the general expectations library. I'd do something like this in your test

```
Expect(node.Labels).To(HaveKeyWithValue(v1.LabelInstanceTypeStable, "expected-instance-type"))
```

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -165,16 +169,66 @@ var _ = Describe("Allocation", func() {
...
-				for _, override := range overrides {
-					Expect(*override.InstanceType).To(Equal("inf1.6xlarge"))
+				// capacity shortage is over - expire the item from the cache and try again
+				fakeEC2API.InsufficientCapacityPools = []fake.CapacityPool{}
+				unavailableOfferingsCache.Delete(UnavailableOfferingsCacheKey(v1alpha1.CapacityTypeOnDemand, "inf1.6xlarge", "test-zone-1a"))
+				nodeNames := sets.NewString()
+				for _, pod := range ExpectProvisioned(ctx, env.Client, scheduler, provisioners, provisioner, pods...) {
+					nodeNames = nodeNames.Insert(ExpectScheduledWithInstanceType(ctx, env.Client, pod, "inf1.6xlarge").Name)
 				}
+				Expect(len(nodeNames)).To(Equal(1))
```
ellistarn, line 231: We test binpacking elsewhere, so you could reduce the scope of this test to not test len(nodes).

eptiger: If you don't feel strongly about it, I'd rather keep these checks. I don't think it hurts to sanity test that the test is working as the author expected.

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -165,16 +173,111 @@ var _ = Describe("Allocation", func() {
...
+				for _, pod := range ExpectProvisioned(ctx, env.Client, scheduler, provisioners, provisioner, pods...) {
+					node := ExpectScheduled(ctx, env.Client, pod)
+					Expect(node.Labels).To(SatisfyAll(
+						HaveKeyWithValue(InstanceTypeLabel, "p3.8xlarge"),
+						HaveKeyWithValue(ZoneLabel, "test-zone-1b")))
+					nodeNames.Insert(node.Name)
+				}
+				Expect(nodeNames.Len()).To(Equal(2))
+			})
+			It("should launch instances on later recon attempt with Insufficient Capacity Error Cache expiry", func() {
```
ellistarn, line 251:
```suggestion
			It("should launch instances on later reconciliation attempt with Insufficient Capacity Error Cache expiry", func() {
```

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -165,16 +173,111 @@ var _ = Describe("Allocation", func() {
...
+				fakeEC2API.InsufficientCapacityPools = []fake.CapacityPool{{InstanceType: "inf1.6xlarge", Zone: "test-zone-1a"}}
+				pods := ExpectProvisioned(ctx, env.Client, scheduler, provisioners, provisioner,
+					test.UnschedulablePod(test.PodOptions{
+						ResourceRequirements: v1.ResourceRequirements{
+							Requests: v1.ResourceList{resources.AWSNeuron: resource.MustParse("2")},
```
ellistarn, line 256: It might be cleaner to do 
```
test.PodOptions{
  NodeSelector: map[string]string{v1.LabelInstanceTypeStable: "inf1.6xlarge"}
}
```
And explicitly test an iced instance type. I worry that if someone adds a new instance type for testing, then this test will break. I might also simplify to a single pod instead of two. In general, test testing style in karpenter is to be surgical and focus on the specific behavior being tested, but also exercise the e2e happy code path outside of the specific behavior..

ellistarn (approved):

pkg/cloudprovider/aws/fake/ec2api.go
```diff
@@ -67,18 +79,50 @@ func (e *EC2API) CreateFleetWithContext(_ context.Context, input *ec2.CreateFlee
...
 			InstanceType:   input.LaunchTemplateConfigs[0].Overrides[0].InstanceType,
 		})
 		e.Instances.Store(*instances[i].InstanceId, instances[i])
 		instanceIds = append(instanceIds, instances[i].InstanceId)
 	}
 
-	return &ec2.CreateFleetOutput{Instances: []*ec2.CreateFleetInstance{{InstanceIds: instanceIds}}}, nil
+	result := &ec2.CreateFleetOutput{
+		Instances: []*ec2.CreateFleetInstance{{InstanceIds: instanceIds}}}
+	if len(skippedPools) > 0 {
```
eptiger, line 108: Whoops, meant to remove this. Sorry :)

PR #817, opened 2021-11-19
Adds provisioner limits based on cpu and mem

suket22:
**1. Issue, if available:**
N/A

Adds limits to the Karpenter provisioner based on CPU and Memory. With the current implementation these are *soft* limits and aren't strictly enforced. 

The way the limiting works is that we keep launching worker nodes until we have **any amount** of capacity yet. Only once we've actually hit / crossed our limits do we stop the provisioning. It's a pretty naive implementation, and can be problematic sometimes. 

Scenario 1 - 
You've got 2 CPU left, and a 10 new pods come in each asking for 1 CPU. In an ideal world, we should only provision a tiny instance with 2CPU that can host 2 pods and the other 8 remain stuck in pending. However, what this implementation will do is provision whatever instance type Karpenter would do anyway based on our bin packing logic and fit all 10 pods. The next batch of pending pods however, will be entirely stuck since we'd detect we've crossed our limits.

Scenario 2 - 
If you set the CPU or Memory to 0 in the provisioner, Karpenter won't be able to launch *any* worker nodes at all. This is by design and can help when you're looking to block instance creation entirely. Will be more useful once I add GPUs in.

**What is to come next?**
* Adding GPU support. That should come quick - I just need to play around with it a little more. 
* Adding documentation on how this limiting works. 
* Adding metrics for the resource consumption in the cluster per provisioner. 

**Testing done so far**
* It's backwards compatible so if you haven't used limits before, your provisioning will continue to remain unlimited. 
* Specifying just cpu or just memory and a combination etc. 
* Done small scale ups to verify that we allow provisioning until we detect that we've *hit/crossed* our limits and then it blocks all provisioning.

- [x] No


pkg/controllers/provisioning/resourcecounter.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, err
+	}
+	resourceCount := ResourceCount{
+		CPU:    resource.Zero.AsDec(),
+		Memory: resource.Zero.AsDec(),
+	}
+	for _, node := range nodeList.Items {
+		resourceCount.CPU.Add(resourceCount.CPU, node.Status.Capacity.Cpu().AsDec())
+		resourceCount.Memory.Add(resourceCount.Memory, node.Status.Capacity.Memory().AsDec())
+	}
```
suket22, line 59: Custom resources like GPUs aren't going to show up in the nodeSpec. So it's difficult to determine how many GPUs have been provisioned by the cluster like this. 

I'm likely going to do a translation that goes nodeObject -> instanceTypeLabel -> retrieveInstanceTypeInfo. Figuring out this bit.

pkg/apis/provisioning/v1alpha5/limits.go
```diff
@@ -0,0 +1,37 @@
...
+
+// Limits define bounds on the resources being provisioned by Karpenter
+type Limits struct {
+	// Resources contains all the allocatable resources that Karpenter supports for limiting.
+	Resources Resources `json:"resources,omitempty"`
+}
+
+func (l *Limits) Default(ctx context.Context) {
+	if l.Resources.CPU == nil {
+		l.Resources.CPU = resource.NewScaledQuantity(math.MaxInt64, 0)
```
ellistarn, line 32: Seems odd to me that we'd limit karpenter at 0 by default. I wonder what a safe value is here. We could potentially leave it unbounded, but set it to 100/100Gi in the docs so it's highly visible to users.

pkg/apis/provisioning/v1alpha5/provisioner.go
```diff
@@ -46,6 +46,8 @@ type ProvisionerSpec struct {
 	// Termination due to expiration is disabled if this field is not set.
 	// +optional
 	TTLSecondsUntilExpired *int64 `json:"ttlSecondsUntilExpired,omitempty"`
+	// Defines a set of bounds that Karpenter obeys when provisioning capacity.
```
ellistarn, line 49: Go comment style starts with the thing you've exported
```suggestion
	// Limits define a set of bounds that Karpenter obeys when provisioning capacity.
```

pkg/apis/provisioning/v1alpha5/resources.go
```diff
@@ -0,0 +1,28 @@
...
+
+package v1alpha5
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// Resources contains a list of all the allocatable resources that can be used to define a bound on penter's
+// scaling actions.
+type Resources struct {
```
ellistarn, line 23: I'm not convinced we need to define a new type. Can we simply use a resource.ResourceList inline?

pkg/controllers/provisioning/resourcecounter.go
```diff
@@ -0,0 +1,85 @@
...
+	"knative.dev/pkg/logging"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ResourceCount struct {
+	CPU    *inf.Dec
+	Memory *inf.Dec
+}
+
+type ResourceCounter struct {
```
ellistarn, line 33:
```suggestion
type Limiter struct {
```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -66,6 +66,7 @@ var _ = BeforeEach(func() {
 		ObjectMeta: metav1.ObjectMeta{Name: v1alpha5.DefaultProvisioner.Name},
 		Spec:       v1alpha5.ProvisionerSpec{},
 	}
+	provisioner.SetDefaults(ctx)
```
ellistarn, line 69: We should include limits in tests that exercise limits.

pkg/controllers/provisioning/launcher.go
```diff
@@ -38,24 +38,36 @@ import (
...
 	errs := make([]error, len(schedules))
+	remainingResources, err := l.ResourceCounter.remainingResourceCounts(ctx, provisioner.Spec.Limits, provisioner.Name)
+	if err != nil {
+		return fmt.Errorf("cannot determine remaining resources, not launching instances: %w", err)
+	}
 	workqueue.ParallelizeUntil(ctx, len(schedules), len(schedules), func(index int) {
 		for _, packing := range l.Packer.Pack(ctx, schedules[index], instanceTypes) {
+			if remainingResources.isInsufficient(ctx) {
+				errs[index] = multierr.Append(errs[index], fmt.Errorf("provisioner limits exceeded"))
+				return
```
ellistarn, line 61: Should we continue here? What if a pod needs CPU but not GPU and GPUs are limited.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -85,7 +85,7 @@ func (p *Provisioner) provision(ctx context.Context) (err error) {
 		return fmt.Errorf("getting instance types")
 	}
 	// Launch capacity and bind pods
-	if err := p.launcher.Launch(ctx, schedules, instanceTypes); err != nil {
+	if err := p.launcher.Launch(ctx, schedules, instanceTypes, *p.Provisioner); err != nil {
```
ellistarn, line 88: Can we pass limits instead of the full provisioner?

pkg/controllers/provisioning/launcher.go
```diff
@@ -38,24 +38,36 @@ import (
...
+	ResourceCounter *ResourceCounter
 }
 
-func (l *Launcher) Launch(ctx context.Context, schedules []*scheduling.Schedule, instanceTypes []cloudprovider.InstanceType) error {
+func (l *Launcher) Launch(ctx context.Context, schedules []*scheduling.Schedule,
+	instanceTypes []cloudprovider.InstanceType, provisioner v1alpha5.Provisioner) error {
+
 	// Pack and bind pods
 	errs := make([]error, len(schedules))
+	remainingResources, err := l.ResourceCounter.remainingResourceCounts(ctx, provisioner.Spec.Limits, provisioner.Name)
```
ellistarn, line 53: I'm not sure how good of an idea this is, but WDYT about keeping track of launched resources in a separate controller that updates `provisioner.status.resources` and then we can check that without doing the math in the critical path.

We'd need to read the status of the provisioner on each loop, but it should hit the cache and be pretty quick. It opens us up to eventual consistency issues, but we already have them as you've described. 

It has a nice side effect of customers being able to see the data with `kubectl get provisioner`

pkg/apis/provisioning/v1alpha5/limits.go
suket22: It's not limited to 0 - it's actually set to the max value of a 64 bit signed int. I'd be okay changing this to like 100/100Gi though

pkg/apis/provisioning/v1alpha5/resources.go
suket22: Good point, let me fix.

pkg/controllers/provisioning/scheduling/suite_test.go
suket22: Yeah I think I need to add specific tests to verify the limits behavior. I had to add this in because otherwise you end up dereferencing nil pointers in the launcher code which broke all unrelated tests. I think defaults should be setup regardless

pkg/controllers/provisioning/launcher.go
suket22: Ah, won't it be a part of a different schedule in that case? My understanding was that everything a part of the same packings will have identical constraints

pkg/controllers/provisioning/launcher.go
suket22: > I'm not sure how good of an idea this is, but WDYT about keeping track of launched resources in a separate controller that updates provisioner.status.resources and then we can check that without doing the math in the critical path.

That's a good idea. I tried caching things within the provisioning controller because I'm worried about listing all nodes in the cluster each provisioning loop (I've seen this call take 3-4 seconds for large clusters). I realized really quick that it breaks because the termination's happening in a different controller and the cache goes out of date but I should've broken out the control logic entirely.

pkg/controllers/provisioning/launcher.go
```diff
@@ -38,24 +38,36 @@ import (
...
+	if err != nil {
+		return fmt.Errorf("cannot determine remaining resources, not launching instances: %w", err)
+	}
 	workqueue.ParallelizeUntil(ctx, len(schedules), len(schedules), func(index int) {
 		for _, packing := range l.Packer.Pack(ctx, schedules[index], instanceTypes) {
+			if remainingResources.isInsufficient(ctx) {
+				errs[index] = multierr.Append(errs[index], fmt.Errorf("provisioner limits exceeded"))
```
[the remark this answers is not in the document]
suket22: Agreed. I do have a log line within `isInsufficient()` which logs which limit is being breached, but I'll modify this so it's bubbled up here and part of the error itself.

cmd/controller/main.go
```diff
@@ -91,6 +92,7 @@ func main() {
 		terminator,
 		node.NewController(manager.GetClient()),
 		metrics.NewController(manager.GetClient(), cloudProvider),
+		counter.NewController(ctx, manager.GetClient()),
```
ellistarn, line 95: Can you create an issue to fold this into the metrics controller?m

charts/karpenter/templates/controller/rbac.yaml
```diff
@@ -74,7 +74,7 @@ rules:
   - patch
   - get
   - list
-  - patch
```
ellistarn, line 77: This seems wrong if you're patching below

pkg/controllers/counter/controller.go
```diff
@@ -0,0 +1,127 @@
...
+func NewController(ctx context.Context, kubeClient client.Client) *Controller {
+	return &Controller{
+		ctx:        ctx,
+		kubeClient: kubeClient,
+	}
+}
+
+// Reconcile a control loop for the resource
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	logging.FromContext(ctx).Infof("Updating resource counts on provisioner")
```
ellistarn, line 56: I'd keep this controller quiet. Do you think we need this?

pkg/controllers/counter/controller.go
```diff
@@ -0,0 +1,127 @@
...
+		return reconcile.Result{}, err
+	}
+
+	provisioner.Status.Resources = resourceCounts
+	if err := c.kubeClient.Status().Patch(ctx, provisioner, client.MergeFrom(persisted)); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to persist changes to %s, %w", req.NamespacedName, err)
+	}
+
+	// Refresh the reconciler state values every 5 minutes irrespective of node events
+	return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
```
ellistarn, line 79: Do we need this? Watch should be enough.

pkg/controllers/counter/controller.go
```diff
@@ -0,0 +1,127 @@
...
+				// No need to reconcile the provisioner when nodes are updated since those don't affect the status of the provisioner.
+				return false
+			},
+		}).
+		Watches(
+			// Reconcile provisioner state when a node managed by it is created or deleted.
+			&source.Kind{Type: &v1.Node{}},
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) (requests []reconcile.Request) {
+				if provisionerName, ok := o.GetLabels()[v1alpha5.ProvisionerNameLabelKey]; ok {
+					requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Name: provisionerName}})
```
ellistarn, line 119: consider just returning directly and not naming the return type
```
return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: provisionerName}}
```

pkg/controllers/counter/controller.go
```diff
@@ -0,0 +1,127 @@
...
+func (c *Controller) resourceCountsFor(ctx context.Context, provisionerName string) (v1.ResourceList, error) {
+	nodes := v1.NodeList{}
+	if err := c.kubeClient.List(ctx, &nodes, client.MatchingLabels{v1alpha5.ProvisionerNameLabelKey: provisionerName}); err != nil {
+		return nil, err
+	}
+
+	var cpu = resource.NewScaledQuantity(0, 0)
+	var memory = resource.NewScaledQuantity(0, resource.Giga)
+
+	for _, node := range nodes.Items {
```
ellistarn, line 91: Can we factor in GPUs here?

pkg/controllers/provisioning/controller.go
```diff
@@ -129,6 +131,19 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 		NewControllerManagedBy(m).
 		Named("provisioning").
 		For(&v1alpha5.Provisioner{}).
+		WithEventFilter(predicate.Funcs{
```
ellistarn, line 134: Instead of this, what do you think about doing an `object.DeepEquals(provisoner.Spec, active.Spec)` and only restarting the provisioner if something has changed? We'd need to check the instance types as well.

pkg/controllers/provisioning/controller.go
```diff
@@ -129,6 +131,19 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
...
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				// If the only change to the provisioner resource has been in the status field, then we shouldn't be reconciling.
+				// This is important because otherwise you can end up in an odd loop. Consider this -
+				// 1. The launcher adds new nodes -> which leads to provisioner.status.resources being updated.
+				// 2. We then think a new provisioner is needed and cancel the context for the old provisioner.
+				// 3. If the old provisioner was doing things (say binding pods after the node was spun up) - it might then fail.
+				//
+				// When there's a change in just the status field (and not the spec), the generation values will remain the same so use that to
+				// determine if we need to reconcile.
+				return e.ObjectNew.GetGeneration() != e.ObjectOld.GetGeneration()
```
ellistarn, line 144: Cool trick!

pkg/apis/provisioning/v1alpha5/limits.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	if _, ok := l.Resources[ResourceLimitsCPU]; !ok {
+		l.Resources[ResourceLimitsCPU] = *DefaultCPULimits
+	}
+	if _, ok := l.Resources[ResourceLimitsMemory]; !ok {
+		l.Resources[ResourceLimitsMemory] = *DefaultMemoryLimits
+	}
+}
+
+func (p *Provisioner) HasExceededResources() (bool, error) {
```
ellistarn, line 57: Thoughts on just returning an error instead of bool, error. Currently, the bool will only be true if the error is also not nil, so we can collapse the concept.

ellistarn (commented):
Nice! I really like this approach. Some quick comments.

charts/karpenter/templates/controller/rbac.yaml
suket22: We already have patch on #L74 so I changed this to update so we've always got both. I can just remove this line

pkg/controllers/counter/controller.go
suket22: Agreed, I'll remove.

pkg/controllers/counter/controller.go
suket22: Fair enough, I'll remove

pkg/controllers/counter/controller.go
suket22: GPUs is a little more work. Node objects don't have any notion of GPUs as far as I can tell so I've to go the long route of fetching the instanceType from the node labels and then asking the cloudprovider about the GPUs that instanceType comes with. 

I was thinking of building a separate cache for all instanceTypes (I think there was a cloudprovider interface that gets us that data already) in this controller.

suket22: Basically, I'm happy to do it but was thinking I'll do it either in my next PR or after the CPU memory bit looks good :)

pkg/apis/provisioning/v1alpha5/limits.go
suket22: Yup yup, I'll just return the err

pkg/apis/provisioning/v1alpha5/limits.go
```diff
@@ -0,0 +1,71 @@
...
+	"context"
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+const (
+	// CPU limit, in cores. (500m = .5 cores)
+	ResourceLimitsCPU v1.ResourceName = "cpu"
```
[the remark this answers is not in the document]
suket22: Oooh don't know how I missed that - fixed.

pkg/apis/provisioning/v1alpha5/provisioner_status.go
```diff
@@ -29,6 +30,9 @@ type ProvisionerStatus struct {
 	// its target, and indicates whether or not those conditions are met.
 	// +optional
 	Conditions apis.Conditions `json:"conditions,omitempty"`
+
+	// Resources is the list of resources that have been provisioned.
+	Resources v1.ResourceList `json:"resources,omitempty"`
```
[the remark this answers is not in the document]
suket22: Nice call. I think there's definite value in having that in ProvisionerSpec. Might even help show how much better the instance utilization is with Karpenter. I think in the API though, it should could maybe be Provisioner.Status.Requests? 

Agreed though, that it might mean we update the provisioner on each pod event so I'm a little concerned from a scaling front.

pkg/controllers/provisioning/controller.go
suket22: I'd be fine with either checks - the generation value is probably a little cheaper?
Not sure I follow the instance types - do you mean the fact that we restart the provisioner every 5 minutes in any case to refresh our knowledge of ICEs?

cmd/controller/main.go
suket22: https://github.com/awslabs/karpenter/issues/840

pkg/apis/provisioning/v1alpha5/limits.go
```diff
@@ -0,0 +1,71 @@
...
+	// Memory limit, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
+	ResourceLimitsMemory v1.ResourceName = "memory"
+)
+
+var DefaultCPULimits *resource.Quantity = resource.NewScaledQuantity(100, 0)
```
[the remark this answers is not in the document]
suket22: Discussed this offline and decided we're not going to enforce any defaults. We will enforce limits to be specified as part of the provisionerSpec though.

pkg/controllers/provisioning/launcher.go
```diff
@@ -44,11 +45,21 @@ type Launcher struct {
...
-func (l *Launcher) Launch(ctx context.Context, schedules []*scheduling.Schedule, instanceTypes []cloudprovider.InstanceType) error {
+func (l *Launcher) Launch(ctx context.Context, provisioner *v1alpha5.Provisioner, schedules []*scheduling.Schedule,
+	instanceTypes []cloudprovider.InstanceType) error {
+
 	// Pack and bind pods
 	errs := make([]error, len(schedules))
+	provisioner, err := l.updateState(ctx, provisioner)
+	if err != nil {
+		return fmt.Errorf("unable to determine status of provisioner")
+	}
```
suket22, line 56: I'm not sure why I didn't need this in my earlier revisions. Maybe I was doing something else that was forcing the provisioner to get updated or I wasn't testing my scale ups fast enough, but without this the state in this provisioner object is really off. Since it's hitting a cache, I don't mind moving this into the for loop below either so we reduce the chance of staleness

pkg/apis/provisioning/v1alpha5/limits.go
```diff
@@ -0,0 +1,43 @@
...
+	// Resources contains all the allocatable resources that Karpenter supports for limiting.
+	Resources v1.ResourceList `json:"resources,omitempty"`
+}
+
+func (p *Provisioner) HasExceededResources() error {
+	var currentResource = p.Status.Resources
+	var currentLimits = p.Spec.Limits.Resources
+
+	for resourceName, usage := range currentResource {
+		fmt.Printf("resourceName %v resourceUsage %v\n", resourceName, usage)
```
ellistarn, line 34: Reminder to remove these before submiting.

pkg/apis/provisioning/v1alpha5/limits.go
```diff
@@ -0,0 +1,43 @@
...
+		if limit, ok := currentLimits[resourceName]; ok {
```
ellistarn, line 35: consider inlining p.Spec.Limits.Resources, since it's only used here.

pkg/apis/provisioning/v1alpha5/limits.go
```diff
@@ -0,0 +1,43 @@
...
+type Limits struct {
+	// Resources contains all the allocatable resources that Karpenter supports for limiting.
+	Resources v1.ResourceList `json:"resources,omitempty"`
+}
+
+func (p *Provisioner) HasExceededResources() error {
+	var currentResource = p.Status.Resources
+	var currentLimits = p.Spec.Limits.Resources
+
+	for resourceName, usage := range currentResource {
```
ellistarn, line 33: consider lining p.Status.Resources since it's only used here.

pkg/apis/provisioning/v1alpha5/limits.go
```diff
@@ -0,0 +1,43 @@
...
+		fmt.Printf("resourceName %v resourceUsage %v\n", resourceName, usage)
+		if limit, ok := currentLimits[resourceName]; ok {
+			fmt.Printf("limitName %v limitUsage %v\n", resourceName, limit)
+			if usage.Cmp(limit) >= 0 {
+				return fmt.Errorf("%v limits exceeded", resourceName)
```
ellistarn, line 38: We typically explicitly use "%s" instead of "%v". I'm open minded to migrating to %v for everything though: https://pkg.go.dev/fmt.

Can we make this error a little more expressive?

"resource limit %q exceeded, %v > %v"

pkg/apis/provisioning/v1alpha5/provisioner.go
```diff
@@ -46,6 +46,8 @@ type ProvisionerSpec struct {
 	// Termination due to expiration is disabled if this field is not set.
 	// +optional
 	TTLSecondsUntilExpired *int64 `json:"ttlSecondsUntilExpired,omitempty"`
+	// Limits define a set of bounds that Karpenter obeys when provisioning capacity.
+	Limits `json:"limits,omitempty"`
```
ellistarn, line 50: Did you mean to use an inline struct? I'd expect
```
Limits Limits `json:"limits,omitempty"`
```

pkg/apis/provisioning/v1alpha5/provisioner_status.go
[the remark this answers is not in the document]
ellistarn: Controller-runtime will batch the reconciles and has a serial execution guarantee. I think this will roughly translate into a linear stream of writes to the API Server as it detects new resources. AFAICT, it shouldn't be enough to overwhelm anything.

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -52,13 +53,21 @@ func (s *ProvisionerSpec) validateTTLSecondsUntilExpired() (errs *apis.FieldErro
...
+
 func (s *ProvisionerSpec) validateTTLSecondsAfterEmpty() (errs *apis.FieldError) {
 	if ptr.Int64Value(s.TTLSecondsAfterEmpty) < 0 {
 		return errs.Also(apis.ErrInvalidValue("cannot be negative", "ttlSecondsAfterEmpty"))
 	}
 	return errs
 }
 
+func (s *ProvisionerSpec) validateResourceLimits() (errs *apis.FieldError) {
+	if s.Limits.Resources == nil || len(s.Limits.Resources) == 0 {
```
ellistarn, line 65: Zero is allowed, right? Nil is not.

pkg/controllers/counter/controller.go
```diff
@@ -0,0 +1,122 @@
...
+// Controller for the resource
+type Controller struct {
+	ctx        context.Context
+	kubeClient client.Client
+}
+
+// NewController is a constructor
+func NewController(ctx context.Context, kubeClient client.Client) *Controller {
+	return &Controller{
+		ctx:        ctx,
```
ellistarn, line 47: I don't think you want to hold onto ctx for this controller. This is typically used to signal cleanup logic, which we don't have here.

pkg/controllers/counter/controller.go
```diff
@@ -0,0 +1,122 @@
...
+		WithEventFilter(predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				// No need to reconcile the provisioner when nodes are updated since those don't affect the status of the provisioner.
+				return false
+			},
+		}).
+		Watches(
+			// Reconcile provisioner state when a node managed by it is created or deleted.
+			&source.Kind{Type: &v1.Node{}},
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) (requests []reconcile.Request) {
```
ellistarn, line 113: Dont forget to remove this unused named return type.

pkg/controllers/counter/controller.go
```diff
@@ -0,0 +1,122 @@
...
+				if provisionerName, ok := o.GetLabels()[v1alpha5.ProvisionerNameLabelKey]; ok {
```
ellistarn, line 114: optional brevity:
```suggestion
				if name, ok := o.GetLabels()[v1alpha5.ProvisionerNameLabelKey]; ok {
```

pkg/apis/provisioning/v1alpha5/limits.go
suket22: YIKES

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
suket22: Yup zero is allowed - I just set memory to 0 and made sure it works.

pkg/apis/provisioning/v1alpha5/limits.go
```diff
@@ -0,0 +1,38 @@
...
+type Limits struct {
+	// Resources contains all the allocatable resources that Karpenter supports for limiting.
+	Resources v1.ResourceList `json:"resources,omitempty"`
+}
+
+func (p *Provisioner) HasExceededResources() error {
+	for resourceName, usage := range p.Status.Resources {
+		if limit, ok := p.Spec.Limits.Resources[resourceName]; ok {
+			if usage.Cmp(limit) >= 0 {
+				return fmt.Errorf("resource limit %s exceeded", resourceName)
```
ellistarn, line 33: Can we include the values in the log line?

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
ellistarn: Oooooops I read len(s.Limits.Resources) == 0 as checking for zero. Curious how it's possible to hit this case though.

pkg/controllers/counter/controller.go
```diff
@@ -0,0 +1,120 @@
...
+	// Retrieve the provisioner
+	provisioner := &v1alpha5.Provisioner{}
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, provisioner); err != nil {
+		if !errors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
+	persisted := provisioner.DeepCopy()
+
```
ellistarn, line 61: consider cutting these extra new lines. I typically separate sections of code with a comment explaining the section instead of with spaces, which are a bit more ambiguous.

pkg/controllers/provisioning/controller.go
```diff
@@ -117,6 +122,12 @@ func (c *Controller) Apply(ctx context.Context, provisioner *v1alpha5.Provisione
 	return nil
 }
 
+func (c *Controller) areEqual(provisionerSpecOld v1alpha5.ProvisionerSpec, provisionerSpecNew v1alpha5.ProvisionerSpec) bool {
+	hashKeyOld, _ := hashstructure.Hash(provisionerSpecOld, hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: true})
```
ellistarn, line 126: We should log.Fatal if the hash fails. If they both fail, the hash keys may silently be equal.

pkg/controllers/provisioning/controller.go
```diff
@@ -97,6 +98,10 @@ func (c *Controller) Apply(ctx context.Context, provisioner *v1alpha5.Provisione
 	provisioner.Spec.Requirements = provisioner.Spec.Requirements.
 		With(requirements(instanceTypes)).
 		With(v1alpha5.LabelRequirements(provisioner.Spec.Labels))
+	if currentProvisioner, ok := c.provisioners.Load(provisioner.Name); ok && c.areEqual(currentProvisioner.(*Provisioner).Spec, provisioner.Spec) {
```
ellistarn, line 101: Consider moving this complex casting into your helper.

if !c.hasChanged(provisioner) {
  return nil
}

pkg/controllers/provisioning/launcher.go
```diff
@@ -111,6 +122,14 @@ func (l *Launcher) bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) (err
 	return nil
 }
 
+func (l *Launcher) updateState(ctx context.Context, provisioner *v1alpha5.Provisioner) (*v1alpha5.Provisioner, error) {
```
ellistarn, line 125: I think that this might be dangerous. It's possible that the provisioners configuration has changed in a way that would violate current launch parameters. We're currently very careful about refreshing provisioner config, using a threadsafe cache and ensuring that pods are properly drained on the old configuration.

It looks like the provisioner in-memory isn't updated w/ the latest status because we're shortcircuitting on spec differences in the provisioners controller. Perhaps we could do something like update the status of the provisioner, but I worry a bit about how threadsafe that might be. 

I haven't thought too deeply about it, but we might need to pivot to using an in memory store. If we go this approach, we could update the store after launch and check it before launch. I know we've thrashed a bit on this -- happy to discuss offline.

suket22: Discussed this offline and changed this logic. We didn't really need to re-assign the provisioner, but rather only fetch the latest provisioner.Status so doing that instead. This should be fine since provisioner.Status is updated fairly quick (off a node watch)

pkg/controllers/provisioning/launcher.go
```diff
@@ -111,6 +118,21 @@ func (l *Launcher) bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) (err
 	return nil
 }
 
+func (l *Launcher) verifyResourceLimits(ctx context.Context, provisioner *v1alpha5.Provisioner) error {
+	provisionerLatest := &v1alpha5.Provisioner{}
+	if err := l.KubeClient.Get(ctx, types.NamespacedName{Name: provisioner.Name, Namespace: provisioner.Namespace}, provisionerLatest); err != nil {
+		return err
+	}
+	for resourceName, usage := range provisionerLatest.Status.Resources {
+		if limit, ok := provisionerLatest.Spec.Limits.Resources[resourceName]; ok {
```
suket22, line 127: I thought of using provisioner.Spec.Limits here but figured that if we hot update the limits it should be fine

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -52,13 +53,21 @@ func (s *ProvisionerSpec) validateTTLSecondsUntilExpired() (errs *apis.FieldErro
...
 }
+
 func (s *ProvisionerSpec) validateTTLSecondsAfterEmpty() (errs *apis.FieldError) {
 	if ptr.Int64Value(s.TTLSecondsAfterEmpty) < 0 {
 		return errs.Also(apis.ErrInvalidValue("cannot be negative", "ttlSecondsAfterEmpty"))
 	}
 	return errs
 }
 
+func (s *ProvisionerSpec) validateResourceLimits() (errs *apis.FieldError) {
```
ellistarn, line 64: optional: you could put this on limits, e.g.

`func (l *Limits) Validate()`

pkg/controllers/provisioning/launcher.go
ellistarn: I'm a little nervous of this, since it violates assumptions about which policy is applied. I'd just used provisioner.Spec.Limits to reduce the complexity surface.

pkg/controllers/provisioning/suite_test.go
```diff
@@ -69,8 +69,15 @@ var _ = Describe("Provisioning", func() {
 			ObjectMeta: metav1.ObjectMeta{
 				Name: v1alpha5.DefaultProvisioner.Name,
 			},
-			Spec: v1alpha5.ProvisionerSpec{},
+			Spec: v1alpha5.ProvisionerSpec{
+				Limits: v1alpha5.Limits{
+					Resources: v1.ResourceList{
+						v1.ResourceCPU: *resource.NewScaledQuantity(10, 0),
```
ellistarn, line 75: I'm surprised that we don't need to include memory here.

ellistarn (commented):
Nice! We just need a unit test now.

pkg/controllers/provisioning/suite_test.go
suket22: So the way I've written this, you can limit via any combination of resources. So just CPU, just Memory, both CPU and Memory are all valid. The only invalid input is if no resource limits are specified at all.

When there's no memory limits specified, we'll still calculate the memory usage but won't limit on it

pkg/controllers/provisioning/launcher.go
```diff
@@ -124,7 +124,7 @@ func (l *Launcher) verifyResourceLimits(ctx context.Context, provisioner *v1alph
 		return err
 	}
 	for resourceName, usage := range provisionerLatest.Status.Resources {
-		if limit, ok := provisionerLatest.Spec.Limits.Resources[resourceName]; ok {
+		if limit, ok := provisioner.Spec.Limits.Resources[resourceName]; ok {
```
ellistarn, line 127: Sorry to keep layering on comments. One more optional idiomatic improvement:

```
latest := &v1alpha5.Provisioner{}
if err := l.KubeClient.Get(ctx, client.ObjectKeyFromObject(provisioner), latest); err != nil {
	return err
}
if err := provisioner.Spec.Limits.ExceededBy(latest.Status.Resources); err != nil {
  return err
}
return nil
```

pkg/apis/provisioning/v1alpha5/provisioner.go
```diff
@@ -46,6 +46,8 @@ type ProvisionerSpec struct {
 	// Termination due to expiration is disabled if this field is not set.
 	// +optional
 	TTLSecondsUntilExpired *int64 `json:"ttlSecondsUntilExpired,omitempty"`
+	// Limits define a set of bounds that Karpenter obeys when provisioning capacity.
```
ellistarn, line 49: As a general practice, try to avoid referencing the project name from within the project
```suggestion
	// Limits define a set of bounds for provisioning capacity.
```

pkg/controllers/counter/controller.go
```diff
@@ -0,0 +1,114 @@
...
+// NewController is a constructor
+func NewController(kubeClient client.Client) *Controller {
+	return &Controller{
+		kubeClient: kubeClient,
+	}
+}
+
+// Reconcile a control loop for the resource
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	// Retrieve the provisioner
```
ellistarn, line 52:
```suggestion
```

pkg/controllers/counter/controller.go
```diff
@@ -0,0 +1,114 @@
...
+	return controllerruntime.
+		NewControllerManagedBy(m).
+		Named("counter").
+		For(&v1alpha5.Provisioner{}).
+		WithEventFilter(predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				// No need to reconcile the provisioner when nodes are updated since those don't affect the status of the provisioner.
+				return false
+			},
+		}).
```
ellistarn, line 101: We can simplify this now that we have the additional checks. This will catch the case where the node comes online with slightly different values than Karpenter predicted.
```suggestion
```

pkg/controllers/counter/controller.go
```diff
@@ -0,0 +1,114 @@
...
+		Watches(
+			// Reconcile provisioner state when a node managed by it is created or deleted.
```
ellistarn, line 103: no longer accurate, obvious in code
```suggestion
```

pkg/controllers/counter/controller.go
suket22: > This will catch the case where the node comes online with slightly different values than Karpenter predicted.

This filters out a *lot* of node events - I think the conditionList is updated every 10 seconds or so by the kubelet so t I don't know if we should remove this entirely. 

Also how likely is it that nodes come up with values other than what Karpenter predicted? Isn't it all quite deterministic? If it weren't, wouldn't there also be a chance that the packing logic also misfires

ellistarn: ok SGTM 👍

pkg/controllers/provisioning/launcher.go
suket22: Nice, thanks! Done.

pkg/controllers/provisioning/suite_test.go
```diff
@@ -134,6 +141,20 @@ var _ = Describe("Provisioning", func() {
...
+			It("should not schedule when limits are exceeded", func() {
+				provisioner.Status = v1alpha5.ProvisionerStatus{
+					Resources: v1.ResourceList{
+						v1.ResourceCPU: resource.MustParse("100"),
+					},
+				}
+				provisioner.Spec.Limits.Resources[v1.ResourceCPU] = resource.MustParse("20")
+				pod := ExpectProvisioned(ctx, env.Client, scheduler, controller, provisioner, test.UnschedulablePod(
+					test.PodOptions{},
+				))[0]
```
ellistarn, line 154: Pod options are optional.
```suggestion
				pod := ExpectProvisioned(ctx, env.Client, scheduler, controller, provisioner, test.UnschedulablePod())[0]
```

pkg/controllers/provisioning/launcher.go
```diff
@@ -111,6 +117,14 @@ func (l *Launcher) bind(ctx context.Context, node *v1.Node, pods []*v1.Pod) (err
 	return nil
 }
 
+func (l *Launcher) verifyResourceLimits(ctx context.Context, provisioner *v1alpha5.Provisioner) error {
+	latest := &v1alpha5.Provisioner{}
+	if err := l.KubeClient.Get(ctx, client.ObjectKeyFromObject(provisioner), latest); err != nil {
+		return err
```
ellistarn, line 123: Are you familiar with the error wrapping/annotation pattern?

```suggestion
		return fmt.Errorf("getting current resource usage, %w", err)
```

pkg/controllers/provisioning/launcher.go
```diff
@@ -45,11 +45,17 @@ type Launcher struct {
...
-func (l *Launcher) Launch(ctx context.Context, schedules []*scheduling.Schedule, instanceTypes []cloudprovider.InstanceType) error {
+func (l *Launcher) Launch(ctx context.Context, provisioner *v1alpha5.Provisioner, schedules []*scheduling.Schedule,
+	instanceTypes []cloudprovider.InstanceType) error {
+
 	// Pack and bind pods
 	errs := make([]error, len(schedules))
 	workqueue.ParallelizeUntil(ctx, len(schedules), len(schedules), func(index int) {
 		for _, packing := range l.Packer.Pack(ctx, schedules[index], instanceTypes) {
+			if err := l.verifyResourceLimits(ctx, provisioner); err != nil {
+				errs[index] = multierr.Append(errs[index], err)
```
ellistarn, line 56: Here too
```suggestion
				errs[index] = multierr.Append(errs[index], fmt.Errorf("verifying limits, %w", err))
```

pkg/controllers/provisioning/launcher.go
```diff
@@ -45,11 +45,17 @@ type Launcher struct {
 	CloudProvider cloudprovider.CloudProvider
 }
 
-func (l *Launcher) Launch(ctx context.Context, schedules []*scheduling.Schedule, instanceTypes []cloudprovider.InstanceType) error {
+func (l *Launcher) Launch(ctx context.Context, provisioner *v1alpha5.Provisioner, schedules []*scheduling.Schedule,
+	instanceTypes []cloudprovider.InstanceType) error {
+
```
ellistarn, line 50: Extra line
```suggestion
```

pkg/controllers/provisioning/controller.go
```diff
@@ -118,6 +123,23 @@ func (c *Controller) Apply(ctx context.Context, provisioner *v1alpha5.Provisione
 	return nil
 }
 
+// Returns true if the new candidate provisioner is different than the provisioner in memory.
+func (c *Controller) hasChanged(ctx context.Context, provisionerNew *v1alpha5.Provisioner) bool {
+	oldProvisioner, _ := c.provisioners.Load(provisionerNew.Name)
+	if oldProvisioner == nil {
```
ellistarn, line 129: I don't think it's safe to rely on the oldProvisioner being nil here. In go, the foo-or-error pattern has the semantic guarantee that the object will be set if the error is nil, but the reverse is not always true. I'd probably do something like 
```suggestion
	oldProvisioner, ok := c.provisioners.Load(provisionerNew.Name)
	if !ok {
```

ellistarn (dismissed):

ellistarn (approved):

PR #823, opened 2021-11-19
add a max pod batch to prevent large build ups

bwagner5:
**1. Issue, if available:**
N/A

 - During scale tests, I ran into Karpenter running out of memory and restarting, with a large batch of pods that built up which causes Karpenter to continuously fall over.  The same thing can happen if Karpenter is started when there is already a large build up of pods. 
 - This change adds a MaxPodsBatch set at 2,000 initially. This limit is very high for average cluster configurations and is adequate for large scale-ups as well. 
 - If Karpenter does come back online w/ a huge batch of pods, it can progress a little more quickly as well since it doesn't need to wait for the 1 sec idle timeout:
```
karpenter-controller-7b69b7ff89-rpcbg manager 2021-11-19T19:13:06.789Z	INFO	controller.provisioning	Batched 2000 pods in 32.78526ms	{"commit": "19d476f", "provisioner": "default"}
```

 - It may be a good idea to make this configurable in CLI parameters at some point. If the reviewers feel strongly about doing that now, I can do that before we merge this. 

- [x] No

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -29,7 +29,8 @@ import (
 var (
 	MaxBatchDuration = time.Second * 10
 	MinBatchDuration = time.Second * 1
-	MaxPodsPerBatch  = 2_000
+	// MaxPodsPerBatch limits the number of pods we process at one time to avoid using too much memory
+	MaxPodsPerBatch = 2_000
 )
 
 // Provisioner waits for enqueued pods, batches them, creates capacity and binds the pods to the capacity.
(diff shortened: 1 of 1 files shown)


pkg/controllers/provisioning/provisioner.go
```diff
@@ -29,6 +29,7 @@ import (
 var (
 	MaxBatchDuration = time.Second * 10
 	MinBatchDuration = time.Second * 1
+	MaxPodsPerBatch  = 2_000
```
JacobGabrielson, line 32: Could you add a comment explaining why we're doing this (eg to limit memory consumption)?

JacobGabrielson (approved):

PR #830, opened 2021-11-20
Moved faq and provisioner to new reference section

chrisnegus:
As per our new docs organization, I moved the FAQ and provisioner CRD files to a new Reference section. I believe @ellistarn suggested some of the FAQ info is out of date and we might want to enhance the provisioner section. So I'm making this a WIP to work on those sections.

### DIFF
--- a/website/content/en/pre-docs/cloud-providers/AWS/aws-spec-fields.md
+++ b/website/content/en/pre-docs/cloud-providers/AWS/aws-spec-fields.md
@@ -4,7 +4,7 @@ linkTitle: "Spec Fields"
 weight: 10
 ---
 
-The [Provisioner CRD]({{< ref "pre-docs/provisioner-crd.md" >}}) supports defining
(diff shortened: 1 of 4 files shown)


geoffcline:
I previously got feedback from Justin G. and Jesse B. that the provisioned CRD should be featured prominently, and it would be the 2nd most used page for them. I'm worried moving it a level deeper hides it a bit. Any thoughts?

chrisnegus:
> I previously got feedback from Justin G. and Jesse B. that the provisioned CRD should be featured prominently, and it would be the 2nd most used page for them. I'm worried moving it a level deeper hides it a bit. Any thoughts?

That's a good thought. I was thinking that the examples of two provisioners would be in a file under Tasks and the provisioner full reference would be under Reference. That would make each only one click deep. I'm inclined to put them in those sections for the moment, then decide later if we want either of the to move up to the top level of the left nav.

geoffcline (approved):

geoffcline:
I fixed a broken link. This looks ready to merge.

ellistarn (approved):

PR #832, opened 2021-11-21
fix duplicate bind issues

bwagner5:
**1. Issue, if available:**
N/A

 - Fix pod binding issue where the provisioner would try to bind the pod twice resulting in bind failure logging. 
 - Renamed helper func `IsUnschedulable` to `isSchedulable` 
   - `IsUnschedulable` doesn't make sense to describe what the function does. Although the func does do one check to determine if kube-scheduler is failed to schedule the pod, it also does a number of other checks such as if the pod is not already scheduled to a node.  I think it makes more sense to name the func in the context of Karpenter's view of schedulable, hence, `isSchedulable`. 

- [x] No

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -146,16 +146,17 @@ func (p *Provisioner) Batch(ctx context.Context) (pods []*v1.Pod) {
 // between the time it was ingested into the scheduler and the time it is included
 // in a provisioner batch.
 func (p *Provisioner) FilterProvisionable(ctx context.Context, pods []*v1.Pod) []*v1.Pod {
-	unscheduled := []*v1.Pod{}
+	provisionable := []*v1.Pod{}
 	for _, pod := range pods {
-		candidate := pod
+		// the original pod should be returned rather than the newly fetched pod in case the scheduler relaxed constraints
+		original := pod
+		candidate := &v1.Pod{}
 		if err := p.kubeClient.Get(ctx, types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}, candidate); err != nil {
 			logging.FromContext(ctx).Errorf("Unexpected error retrieving pod \"%s/%s\" while checking if it is provisionable", pod.Namespace, pod.Name)
 		}
-		if !isProvisionable(candidate) {
-			continue
+		if candidate.Spec.NodeName == "" {
+			provisionable = append(provisionable, original)
 		}
-		unscheduled = append(unscheduled, candidate)
 	}
-	return unscheduled
+	return provisionable
 }
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -23,8 +23,9 @@ import (
 	"github.com/awslabs/karpenter/pkg/cloudprovider"
 	"github.com/awslabs/karpenter/pkg/controllers/provisioning/scheduling"
 	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/pkg/logging"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var (
@@ -44,6 +45,7 @@ type Provisioner struct {
 	Stop    context.CancelFunc
 	// Dependencies
 	cloudProvider cloudprovider.CloudProvider
+	kubeClient    client.Client
 	scheduler     *scheduling.Scheduler
 	launcher      *Launcher
 }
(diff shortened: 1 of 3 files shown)


pkg/controllers/provisioning/provisioner.go
```diff
@@ -136,3 +138,17 @@ func (p *Provisioner) Batch(ctx context.Context) (pods []*v1.Pod) {
 		}
 	}
 }
+
+func (p *Provisioner) RemoveScheduled(ctx context.Context, pods []*v1.Pod) []*v1.Pod {
```
ellistarn, line 142: Can you add a comment here about why this check is useful?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -136,3 +138,17 @@ func (p *Provisioner) Batch(ctx context.Context) (pods []*v1.Pod) {
...
+	unscheduledPods := []*v1.Pod{}
+	for _, pod := range pods {
+		candidatePod, err := p.launcher.CoreV1Client.Pods(pod.Namespace).Get(ctx, pod.Name, metav1.GetOptions{})
```
ellistarn, line 145: optional:
s/unscheduledPods/unscheduled
s/candidatePod/candidate

nonoptional:
This looks like dependency theft to me. Can you inject in a client.Client instead of relying on the launcher's CoreV1Client?

pkg/controllers/provisioning/scheduler.go
```diff
@@ -102,7 +102,7 @@ func (s *Scheduler) Schedule(ctx context.Context, pod *v1.Pod) error {
 	return nil
 }
 
-func isUnschedulable(p *v1.Pod) error {
+func isSchedulable(p *v1.Pod) error {
```
ellistarn, line 105: @JacobGabrielson and I went back and forth on this naming, since there was a similar thing that bit us recently. I think the weird bit is that since the function returns an error, but it seems like it should return a boolean, and that if the boolean is true, then it's the happy case. Instead, you need to write isSchedulable != nil, which is really confusing.

We could make the functions all oriented around the validate concept, which semantically returns an error.

e.g.

```
validateSchedulable(pod) // potentially makes more sense as provisionable, since it technically is not schedulable by the scheduler.
```

ellistarn: Alternatively, it doesn't look like we're using these error messages anymore, so you could just make this a good old fashioned boolean. That maybe results in the simplest outcome `if isProvisionable(pod) { }`.

pkg/controllers/provisioning/provisioner.go
bwagner5: Yeah, this was a WIP, forgot to prepend. Definitely going to clean it up there :D

bwagner5:
Not ready yet, need to look into test failures.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -136,3 +140,22 @@ func (p *Provisioner) Batch(ctx context.Context) (pods []*v1.Pod) {
 		}
 	}
 }
+
+// RemoveScheduled removes pods that have been assigned a node.
+// This check is needed to prevent duplicate binds when a pod is scheduled to a node
+// between the time it was ingested into the scheduler and the time it is included
+// in a provisioner batch.
+func (p *Provisioner) RemoveScheduled(ctx context.Context, pods []*v1.Pod) []*v1.Pod {
```
ellistarn, line 148: The hIgHeR ORdeR fUNcTioNS naming for this would be FilterProvisionable(ctx, pods)

pkg/controllers/provisioning/provisioner.go
```diff
@@ -136,3 +140,22 @@ func (p *Provisioner) Batch(ctx context.Context) (pods []*v1.Pod) {
...
+	unscheduled := []*v1.Pod{}
+	for _, pod := range pods {
+		candidate := pod
+		if err := p.kubeClient.Get(ctx, types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}, candidate); err != nil {
+			logging.FromContext(ctx).Errorf("Checking if pod \"%s\" is provisionable", pod)
```
ellistarn, line 153: This error log line seems a bit odd to me. Also, what if the pod is deleted at this point? Perhaps we should continue if errors.IsNotFound, but log an error otherwise? "Unexpected error filtering provisionable pods, %s"

pkg/controllers/provisioning/scheduler.go
```diff
@@ -102,20 +102,11 @@ func (s *Scheduler) Schedule(ctx context.Context, pod *v1.Pod) error {
 	return nil
 }
 
-func isUnschedulable(p *v1.Pod) error {
-	if p.Spec.NodeName != "" {
-		return fmt.Errorf("already scheduled")
+func isProvisionable(p *v1.Pod) bool {
+	if p.Spec.NodeName != "" || !pod.FailedToSchedule(p) || pod.IsOwnedByDaemonSet(p) || pod.IsOwnedByNode(p) {
```
ellistarn, line 106: Thoughts on
```
return p.spec.NodeName == "" && pod.FailedToSchedule(p) && !pod.OwnedByDaemonsSet(p) && !pod.OwnedByNode(p)
```

pkg/controllers/provisioning/provisioner.go
bwagner5: tbh I think a lot of our error logging is really weird like this and doesn't read well in the actual log. I've tried to keep with our convention for the most part, but I definitely like explicitly saying "Unexpected error" rather than relying on the log level to convey it's an error. Maybe we should clean up the other error logging.

ellistarn: I love this comment! No need to say failed if it's an error!

ellistarn (dismissed):

ellistarn (approved):

PR #842, opened 2021-11-24
[Experimental][WIP] Refactored all scheduling logic to scheduling controller

ellistarn:
**1. Issue, if available:**

- [ ] No


pkg/controllers/provisioning/provisioner.go
```diff
@@ -17,122 +17,142 @@ package provisioning
...
+	key, err := hashstructure.Hash(constraints, hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: true})
 	if err != nil {
-		return fmt.Errorf("solving scheduling constraints, %w", err)
+		logging.FromContext(ctx).Fatalf("Failed add pod to provisioner, hashing constraints, %s, %w", pretty.Concise(constraints), err.Error())
 	}
-	// Get Instance Types, offering availability may vary over time
-	instanceTypes, err := p.cloudProvider.GetInstanceTypes(ctx, &p.Spec.Constraints)
-	if err != nil {
-		return fmt.Errorf("getting instance types")
+	batch, ok := p.Batches.LoadOrStore(key, NewBatch(ctx, fmt.Sprint(key), constraints))
```
JacobGabrielson, line 62: It could just be me, but I think this code is a bit confusing ... since `key` is derived from `constraints`, the fact that it's repeated twice (passed to both `NewBatch` and `LoadOrStore`) comes across as unusual. It feels like since the `Batch` struct already knows about `Constraints`, it should maybe deal with this problem (of not starting more than one provisioner per batch?) rather than force the caller to do so?

pkg/controllers/provisioning/batch.go
```diff
@@ -0,0 +1,109 @@
...
+var (
+	MaxBatchDuration = time.Second * 10
+	MinBatchDuration = time.Second * 1
+	// MaxPodsPerBatch limits the number of pods we process at one time to avoid using too much memory
+	MaxPodsPerBatch = 2_000
+)
+
+type Batch struct {
+	Key         string
+	Constraints *v1alpha5.Constraints
```
JacobGabrielson, line 34: Is there any way for the `Batch` struct not to "know" about `Constraints`? AFAICT it's just acting as a sort of pass-through (while I'm not suggesting this change, as a thought experiment, it seems like `Batch` could just have an inscrutable `interface{}` here just as well, since it doesn't care about the actual contents? Another way to look at it is that having `Constraints` "in" the `Batch` struct makes it seems like the methods in this file are going to do more than just passively add pods to a channel, but instead be more involved in determining which pods should be added to which channels?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -17,128 +17,148 @@ package provisioning
...
 	if err != nil {
-		return fmt.Errorf("getting instance types")
+		return err
 	}
-	// Launch capacity and bind pods
-	if err := p.launcher.Launch(ctx, p.Provisioner, schedules, instanceTypes); err != nil {
-		return fmt.Errorf("launching capacity, %w", err)
+
+	// Atomically create or use an existing batch.
+	value, ok := p.Batches.LoadOrStore(constraints.MustHash(), NewBatch(ctx, defaultBatchOptions))
```
felix-zhe-huang, line 75: Why do we create multiple batches here? When we batch pods according to constraints hash, doesn't it limited the number of pods that can be batched together? It seems to me this is a bit redundant given that the packer.Pack also groups pods according to constraints hash.

felix-zhe-huang: I am also a little bit worry that this constraints-hash grouping mechanism will further limit ourselves for later if we want to try grouping pods with compatible constraints.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -17,128 +17,148 @@ package provisioning
...
+	batch := value.(*Batch)
+	if !ok {
+		go func() {
+			batch.Start()
+			p.provision(p.Ctx, constraints, batch)
```
felix-zhe-huang, line 80: My understanding is that context should be passed down through the call path. Can we use ctx from the Add function arguments?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -17,128 +17,148 @@ package provisioning
...
 
 // Provisioner waits for enqueued pods, batches them, creates capacity and binds the pods to the capacity.
 type Provisioner struct {
 	// State
 	*v1alpha5.Provisioner
-	pods    chan *v1.Pod
-	results chan error
-	done    <-chan struct{}
+	Batches sync.Map
+	Ctx     context.Context
```
felix-zhe-huang, line 56: Is there any way that we can avoid storing the context here?

pkg/controllers/provisioning/provisioner.go
ellistarn: >It seems to me this is a bit redundant given that the packer.Pack also groups pods according to constraints hash.

Binpacking and scheduling are doing different things. In binpacking it (currently) computes a set of alternatives (OR). In scheduling, it's tightening a set of rules (AND).

> I am also a little bit worry that this constraints-hash grouping mechanism will further limit ourselves for later if we want to try grouping pods with compatible constraints.

This is a very good point -- as long as this mechanism is async we're going to have race conditions that will require retry. The atomicity here likely isn't helping as much as I think it is.

ellistarn: > Why do we create multiple batches here?
We need multiple batches because the constraints may not be the same for (podB + provisionerA) and (podB + provisionerA).

pkg/controllers/provisioning/provisioner.go
ellistarn: Those contexts are owned by the pod reconcile loop.  There's a many : one relationship between pod : batch. The batch needs to collapse many pods into a single provisioning method. The batch itself can be interrupted by the provisioner being cancelled.

ellistarn: Perhaps it would be simpler to just create a new context.

pkg/controllers/provisioning/provisioner.go
ellistarn: I've been searching for a way. I know that the best practice is to wire through functions, but that pattern hasn't fit well with this use case. Perhaps we can discuss w/ screen share?

pkg/controllers/provisioning/batch.go
ellistarn: I like this idea. Check out the new version that's type agnostic and supports range.

pkg/controllers/provisioning/provisioner.go
felix-zhe-huang: > 
So now the multiple batches is responsible of grouping pods according to their constraints. I like that. Much cleaner.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -17,128 +17,148 @@ package provisioning
...
+	value, ok := p.Batches.LoadOrStore(constraints.MustHash(), NewBatch(ctx, defaultBatchOptions))
+	batch := value.(*Batch)
+	if !ok {
+		go func() {
+			batch.Start()
+			p.provision(p.Ctx, constraints, batch)
+			batch.Stop()
+		}()
 	}
+	batch.Add(ctx, pod)
```
felix-zhe-huang, line 84: I am a bit confused here. I think there is something obvious that I am missing. Won't this block the reconcile loop until the batch is completed? Since the reconcile loop schedule one pod in each iteration, how are other pods being added to the batch?

pkg/controllers/provisioning/provisioner.go
felix-zhe-huang: Hm... the relationship is indeed quite complicated. In fact we have multiple contexts here and should all of them being able to interrupted the provisioning method?

pkg/controllers/provisioning/provisioner.go
felix-zhe-huang, line 84: If I understand this correctly, the reconcile loop will be blocked until the provisioning is finished. The controller specifics how many concurrent reconciles call (now it is 10).  Giving that we take about a minute to launch a node, does it mean that the maximum number of pods that it can process is around 10/minute? what will happen when we have 150 pods/second churn rate?

ellistarn: We don't block the reconcile loop node creation. It ends as soon as the node is launched and the pod bound. The pod concurrency we use is 10,000.

PR #849, opened 2021-11-24
Priorize provisioners alphabetically if multiple match

ellistarn:
**1. Issue, if available:**

@felix-zhe-huang brought up an odd behavior where unconstrained pods would schedule randomly if multiple provisioners were defined. Eventually I'd like to define a `provisioner.spec.priority` feature, but in the short term, we will just alphabetize to make this deterministic. 

- [ ] No

### DIFF
--- a/pkg/controllers/provisioning/suite_test.go
+++ b/pkg/controllers/provisioning/suite_test.go
@@ -38,7 +38,7 @@ import (
 )
 
 var ctx context.Context
-var controller *provisioning.Controller
+var provisioners *provisioning.Controller
 var scheduler *scheduling.Controller
 var env *test.Environment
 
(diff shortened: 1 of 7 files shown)


pkg/test/expectations/expectations.go
```diff
@@ -105,7 +105,7 @@ func ExpectCreatedWithStatus(c client.Client, objects ...client.Object) {
 	}
 }
 
-func ExpectDeleted(c client.Client, objects ...client.Object) {
+func ExpectDeleted(ctx context.Context, c client.Client, objects ...client.Object) {
```
eptiger, line 108: I must be missing something here - why did you add ctx here? You don't seem to use it AFAICT

pkg/controllers/provisioning/suite_test.go
```diff
@@ -52,8 +52,8 @@ var _ = BeforeSuite(func() {
 	env = test.NewEnvironment(ctx, func(e *test.Environment) {
 		cloudProvider := &fake.CloudProvider{}
 		registry.RegisterOrDie(ctx, cloudProvider)
-		controller = provisioning.NewController(ctx, e.Client, corev1.NewForConfigOrDie(e.Config), cloudProvider)
-		scheduler = scheduling.NewController(e.Client, controller)
+		provisioners = provisioning.NewController(ctx, e.Client, corev1.NewForConfigOrDie(e.Config), cloudProvider)
```
eptiger, line 55: Why the plural variable name if you're only creating one provisioner? I noticed this in expectations.go, too, and was curious.

ellistarn: provisioners is the name I've been using for the controller that manages provisioners. I could use something like provisionerController, but 🤷

bwagner5 (approved):
lgtm

dnutels:
Hi. Is that still how the selection is implemented?

I am asking (and didn't want to open an issue_, because in the [docs](https://karpenter.sh/v0.6.0/tasks/provisioning/):

> Karpenter attempts to provision capacity only when there are unschedulable pods

PR #861, opened 2021-11-26
Minor logging touch ups

ellistarn:
**1. Issue, if available:**

- [ ] No

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -40,7 +40,7 @@ type Provisioner struct {
 	// State
 	*v1alpha5.Provisioner
 	pods    chan *v1.Pod
-	results chan error
+	results chan struct{}
 	done    <-chan struct{}
 	Stop    context.CancelFunc
 	// Dependencies
@@ -74,7 +74,7 @@ func (p *Provisioner) provision(ctx context.Context) (err error) {
 	defer func() {
 		for i := 0; i < len(pods); i++ {
 			select {
-			case p.results <- err: // Block until result is communicated
+			case p.results <- struct{}{}: // Block until result is communicated
 			case <-p.done: // Leave if closed
 			}
 		}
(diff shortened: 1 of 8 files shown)


pkg/controllers/provisioning/provisioner.go
```diff
@@ -40,7 +40,7 @@ type Provisioner struct {
 	// State
 	*v1alpha5.Provisioner
 	pods    chan *v1.Pod
-	results chan error
+	results chan struct{}
```
JacobGabrielson, line 43: After reading up on this topic, I think we might want to use a different strategy here. AFAICT `results` (the way it's being done here) is not a common idiom. Also, although I don't know of any problem specifically, it does seem a bit fragile to rely on "counting" (i.e., the number of writes to the `results` channel needs to match the number of goroutines blocked waiting to read something). I think rather than re-use the `results` channel in the `Provisioner` struct, callers of `Provisioner.Add` should pass in a channel that they will wait on (or pass in a callback that will do something arbitrary, like close a channel, which the caller will just happen to block on). Thoughts?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -74,7 +74,7 @@ func (p *Provisioner) provision(ctx context.Context) (err error) {
 	defer func() {
 		for i := 0; i < len(pods); i++ {
 			select {
-			case p.results <- err: // Block until result is communicated
+			case p.results <- struct{}{}: // Block until result is communicated
```
JacobGabrielson, line 77: `results` seems like a slightly confusing name (especially since there's also a `done` channel right next to it), since no "results" are actually being communicated (e.g., about success/failure/status)?

pkg/controllers/scheduling/controller.go
```diff
@@ -166,5 +166,6 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 		Named("scheduling").
 		For(&v1.Pod{}).
 		WithOptions(controller.Options{MaxConcurrentReconciles: 10_000}).
+		WithLogger(zapr.NewLogger(zap.NewNop())).
```
JacobGabrielson, line 169: Just for my edification, why `NewNop` here?

pkg/controllers/provisioning/provisioner.go
ellistarn: Agree. I need to completely re-write most of this to get the scheduling logic out of this controller. Think of the current impl as an interim step to get multiple provisioners working. Can we address this during the refactor? You can see a preview here: https://github.com/aws/karpenter/pull/842

pkg/controllers/scheduling/controller.go
ellistarn: Controller runtime has a really annoying logging format and logs whenever you return an error. I wanted to use our kpkg logger but also leverage the backoff retry logic that controller-runtime implements on error.

ellistarn: In the longer run, I think we need to figure out how to be in harmony with our controller framework.

JacobGabrielson (approved):
/lgtm

PR #864, opened 2021-11-27
Some chan cleanups

JacobGabrielson:
**1. Issue, if available:**

N/A

Some minor cleanups to `chan` usage per Concurrency in Go book. The main thing is it is good to close when possible, and also it should be clear who owns channel (usually owner should only be able to write/close and reader should only read), using either `<-chan` or `chan<-` types.

Also, I don't understand why provisioner limits are required, so reverted that. But I am probably missing something and someone will point out why :-)

- [X] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/provisioner_validation.go
+++ b/pkg/apis/provisioning/v1alpha5/provisioner_validation.go
@@ -42,7 +42,6 @@ func (s *ProvisionerSpec) validate(ctx context.Context) (errs *apis.FieldError) 
 	return errs.Also(
 		s.validateTTLSecondsUntilExpired(),
 		s.validateTTLSecondsAfterEmpty(),
-		s.Limits.validateResourceLimits(),
 		s.Constraints.Validate(ctx),
 	)
 }
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -108,7 +108,7 @@ func withUserAgent(sess *session.Session) *session.Session {
 }
 
 // Create a node given the constraints.
-func (c *CloudProvider) Create(ctx context.Context, constraints *v1alpha5.Constraints, instanceTypes []cloudprovider.InstanceType, quantity int, callback func(*v1.Node) error) chan error {
+func (c *CloudProvider) Create(ctx context.Context, constraints *v1alpha5.Constraints, instanceTypes []cloudprovider.InstanceType, quantity int, callback func(*v1.Node) error) <-chan error {
 	return c.creationQueue.Add(func() error {
 		return c.create(ctx, constraints, instanceTypes, quantity, callback)
 	})
--- a/pkg/controllers/provisioning/launcher.go
+++ b/pkg/controllers/provisioning/launcher.go
@@ -45,6 +45,16 @@ type Launcher struct {
 	CloudProvider cloudprovider.CloudProvider
 }
 
+// Thread safe channel to pop off packed pod slices
+func queueFor(pods [][]*v1.Pod) <-chan []*v1.Pod {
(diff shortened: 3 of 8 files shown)


pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -108,7 +108,7 @@ func withUserAgent(sess *session.Session) *session.Session {
 }
 
 // Create a node given the constraints.
-func (c *CloudProvider) Create(ctx context.Context, constraints *v1alpha5.Constraints, instanceTypes []cloudprovider.InstanceType, quantity int, callback func(*v1.Node) error) chan error {
+func (c *CloudProvider) Create(ctx context.Context, constraints *v1alpha5.Constraints, instanceTypes []cloudprovider.InstanceType, quantity int, callback func(*v1.Node) error) <-chan error {
```
ellistarn, line 111: The original purpose of making this a channel (as opposed to a callback) was to allow cloud providers to implement arbitrary batching behavior. Instead, we implemented batching in vendor neutral binpacking code. I think we should sync on @bwagner5 on what the desired implementation strategy is.

I don't think this should necessarily block, but I wanted to bring it up since we're touching it.

pkg/controllers/provisioning/launcher.go
```diff
@@ -45,6 +45,16 @@ type Launcher struct {
 	CloudProvider cloudprovider.CloudProvider
 }
 
+// Thread safe channel to pop off packed pod slices
+func queueFor(pods [][]*v1.Pod) <-chan []*v1.Pod {
```
ellistarn, line 49: What does this code do? Is it trying to avoid the latency of a for loop and start creating immediately? It's not actually clear how this works at all -- isn't the queue closed immediately?

pkg/utils/parallel/workqueue.go
```diff
@@ -36,12 +36,11 @@ func NewWorkQueue(qps int, burst int) *WorkQueue {
...
 type task struct {
 	do   func() error
-	done chan error
-	err  error
+	done chan<- error
 }
 
 // Add work to the queue, returns a channel that signals when the work is done.
-func (c *WorkQueue) Add(do func() error) chan error {
+func (c *WorkQueue) Add(do func() error) <-chan error {
```
ellistarn, line 43: I get this one -- this is a good cleanup.

pkg/controllers/provisioning/launcher.go
bwagner5: The channel is loaded and then closed since all items have been placed on it. After it is closed consumers can still dequeue pod batches, the channel just can't receive anymore. So this is just a thread-safe queue so that the CP can consume batches concurrently.

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -42,7 +42,6 @@ func (s *ProvisionerSpec) validate(ctx context.Context) (errs *apis.FieldError)
 	return errs.Also(
 		s.validateTTLSecondsUntilExpired(),
 		s.validateTTLSecondsAfterEmpty(),
-		s.Limits.validateResourceLimits(),
```
bwagner5, line 45: 🤩

pkg/cloudprovider/aws/cloudprovider.go
bwagner5: I think we actually want to keep this channel and refactor the way it's used currently. Right now, we receive from it right after calling which blocks per packing. Practically, it's not much overhead since the packing is batched now and we do have a concurrent work queue on schedules. But we could receive from the channels after we've called all the creates and actually utilize the concurrency this provides. 

TLDR: I think this is good :)

bwagner5 (approved):
lgtm

PR #866, opened 2021-11-27
Fixed a bug where hostname topology launched extra nodes

ellistarn:
**1. Issue, if available:**

This bug is caused by the provisioners controller comparing the state of the current provisioner to the persisted provisioner state and the mutation happening in hostname topology. The scheduling logic mutates the provisioners' requirements, which causes the comparison to inadvertently force refresh the provisioner while nodes are being launched. This causes the provisioner to be drained, which can happen after launching capacity but before binding the pods. This results in multiple scale out (and eventual self-healing / scale-in).

The bug is due to me violating my own principles about not mutating state: https://github.com/aws/karpenter/blob/5d5798b5fefc757ef353889204c56138d8042066/pkg/controllers/provisioning/scheduling/topology.go#L99. In the long term, this mutation will be removed as part of a scheduling refactor.

- [ ] No


JacobGabrielson (approved):
/lgtm

PR #893, opened 2021-12-02
Fixed a bug where spot capacity was not prioritized if flexible to both spot and on demand capacity

ellistarn:
**1. Issue, if available:**

- AWS Cloud Provider defaults to [ on-demand ] if nothing is specified
- if [ on-demand, spot ] is specified, spot will be prioritized if offerings are available
- if spot offerings become unavailable, on demand will be selected

- [ ] No

### DIFF
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -283,7 +283,7 @@ func (p *InstanceProvider) getCapacityType(constraints *v1alpha1.Constraints, in
 	if constraints.Requirements.CapacityTypes().Has(v1alpha1.CapacityTypeSpot) {
 		for _, instanceType := range instanceTypes {
 			for _, offering := range instanceType.Offerings() {
-				if offering.CapacityType == v1alpha1.CapacityTypeSpot {
+				if constraints.Requirements.Zones().Has(offering.Zone) && offering.CapacityType == v1alpha1.CapacityTypeSpot {
 					return v1alpha1.CapacityTypeSpot
 				}
 			}
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -276,10 +276,10 @@ func (p *InstanceProvider) updateUnavailableOfferingsCache(ctx context.Context, 
 	}
 }
 
+// getCapacityType selects spot if both constraints are flexible and there is an
+// available offering. The AWS Cloud Provider defaults to [ on-demand ], so spot
+// must be explicitly included in capacity type requirements.
 func (p *InstanceProvider) getCapacityType(constraints *v1alpha1.Constraints, instanceTypes []cloudprovider.InstanceType) string {
-	// Use spot if constraints are flexible to it and there is an offering for
-	// it. The AWS Cloud Provider defaults to [ on-demand ], so spot must be
-	// explicitly included in capacity type requirements.
 	if constraints.Requirements.CapacityTypes().Has(v1alpha1.CapacityTypeSpot) {
 		for _, instanceType := range instanceTypes {
 			for _, offering := range instanceType.Offerings() {
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -106,16 +106,8 @@ func (p *InstanceProvider) Terminate(ctx context.Context, node *v1.Node) error {
 }
 
 func (p *InstanceProvider) launchInstances(ctx context.Context, constraints *v1alpha1.Constraints, instanceTypes []cloudprovider.InstanceType, quantity int) ([]*string, error) {
-	// Default to on-demand unless constrained otherwise or if flexible to spot and
-	// on-demand. This code assumes two options: {spot, on-demand}, which is enforced
-	// by constraints.Constrain(). Spot may be selected by constraining the provisioner,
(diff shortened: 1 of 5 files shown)


pkg/cloudprovider/aws/instance.go
```diff
@@ -285,6 +276,22 @@ func (p *InstanceProvider) updateUnavailableOfferingsCache(ctx context.Context,
 	}
 }
 
+func (p *InstanceProvider) getCapacityType(constraints *v1alpha1.Constraints, instanceTypes []cloudprovider.InstanceType) string {
+	// Use spot if constraints are flexible to it and there is an offering for
+	// it. The AWS Cloud Provider defaults to [ on-demand ], so spot must be
+	// explicitly included in capacity type requirements.
```
eptiger, line 282: Nit: Should this go above the function instead of inside of it?

Anyway, good catch with this fix! I had thought of this case when working on my capacity type PR and then it totally slipped my mind when actually implementing ICE caching so I'm glad you thought of it.

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -179,7 +179,7 @@ func (p *InstanceTypeProvider) CacheUnavailable(ctx context.Context, instanceTyp
 		capacityType,
 		InsufficientCapacityErrorCacheTTL)
 	// even if the key is already in the cache, we still need to call Set to extend the cached entry's TTL
-	p.unavailableOfferings.SetDefault(UnavailableOfferingsCacheKey(capacityType, instanceType, zone), sets.Empty{})
+	p.unavailableOfferings.SetDefault(UnavailableOfferingsCacheKey(capacityType, instanceType, zone), struct{}{})
```
eptiger, line 182: I did the `sets.Empty{}` thing because it was _you_ who asked I use a variable instead of `struct{}{}` XD

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -260,23 +260,38 @@ var _ = Describe("Allocation", func() {
...
+				pod := ExpectProvisioned(ctx, env.Client, scheduler, provisioners, provisioner, test.UnschedulablePod())[0]
+				node := ExpectScheduled(ctx, env.Client, pod)
+				Expect(node.Labels).To(HaveKeyWithValue(v1alpha5.LabelCapacityType, v1alpha1.CapacityTypeSpot))
 				Expect(fakeEC2API.CalledWithCreateFleetInput.Cardinality()).To(Equal(1))
 				input := fakeEC2API.CalledWithCreateFleetInput.Pop().(*ec2.CreateFleetInput)
 				Expect(input.LaunchTemplateConfigs).To(HaveLen(1))
 				Expect(*input.TargetCapacitySpecification.DefaultTargetCapacityType).To(Equal(v1alpha1.CapacityTypeSpot))
 			})
+			It("should launch on-demand capacity if flexible to both spot and on demand, but only on demand available", func() {
+				unavailableOfferingsCache.SetDefault(UnavailableOfferingsCacheKey(v1alpha1.CapacityTypeSpot, "m5.large", "test-zone-1a"), struct{}{})
```
eptiger, line 281: I assume this would work the same if you used `InsufficientCapacityPools` like in my test cases and you just wrote it this way for stylistic reasons?

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -260,23 +260,38 @@ var _ = Describe("Allocation", func() {
...
 			})
 			It("should launch spot capacity if flexible to both spot and on demand", func() {
 				provisioner.Spec.Requirements = v1alpha5.Requirements{{Key: v1alpha5.LabelCapacityType, Operator: v1.NodeSelectorOpIn, Values: []string{v1alpha1.CapacityTypeSpot, v1alpha1.CapacityTypeOnDemand}}}
-				pod := ExpectProvisioned(ctx, env.Client, scheduler, provisioners, provisioner,
-					test.UnschedulablePod(test.PodOptions{NodeSelector: map[string]string{v1alpha5.LabelCapacityType: v1alpha1.CapacityTypeSpot}}),
-				)[0]
-				ExpectScheduled(ctx, env.Client, pod)
+				pod := ExpectProvisioned(ctx, env.Client, scheduler, provisioners, provisioner, test.UnschedulablePod())[0]
+				node := ExpectScheduled(ctx, env.Client, pod)
+				Expect(node.Labels).To(HaveKeyWithValue(v1alpha5.LabelCapacityType, v1alpha1.CapacityTypeSpot))
```
eptiger, line 274: Nice! I like when we validate the node that got created

eptiger (approved):
Thanks for fixing this! Sorry I missed this for the v0.5 release, but glad you caught it.

pkg/cloudprovider/aws/instancetypes.go
ellistarn: Yeah. However, I realized that we were abusing the sets package. While we do happen to depend on it here, I didn't want our cache to by the only depender on this set package if things got refactored. I looked around for someone else who aliased it, and decided to use the raw guts.

pkg/cloudprovider/aws/suite_test.go
ellistarn: I would've had to reconcile twice. I decided that this is an ice caching test and have updated accordingly. Nice catch!!

eptiger (approved):

PR #904, opened 2021-12-03
add robots.txt 

geoffcline:
**1. Issue, if available:**
lack of /robots.txt file reduces SEO and increases 404 errors

adds robots.txt based on kubernetes/website. 

- [ ] No


website/layouts/robots.txt
```diff
@@ -0,0 +1,7 @@
+User-agent: *
+
+Disallow: /v0.4.3-docs/
+
+Allow: /docs/
```
akestner, line 5: Should this also include the homepage?

website/config.toml
```diff
@@ -6,6 +6,7 @@ baseURL = "/"
 disableKinds = ["taxonomy", "taxonomyTerm"]
 
 # Language settings
+enableRobotsTXT = true
```
akestner, line 9: Does this mean the robots.txt that's committed was generated by Hugo?

akestner (commented):
Couple questions

website/config.toml
geoffcline: no. it just means that the committed robots.txt makes it into the live site.

website/layouts/robots.txt
geoffcline: sure

geoffcline: i've read https://developers.google.com/search/docs/advanced/robots/robots_txt and don't really understand `allow`. it just seems to say google *may* index it, but it's also seems that everything not explictly disallowed is indexed.

website/layouts/robots.txt
```diff
@@ -0,0 +1,8 @@
+User-agent: *
+
+Disallow: /v0.4.3-docs/
```
ellistarn, line 3: This is going to get stale. We need a longer term solution to avoid updating this on every release. Perhaps we can simply allow `/docs`.

ellistarn (approved):

PR #908, opened 2021-12-03
Allow override of webhook port in Helm chart

mikesir87:
**1. Issue, if available:**
No issue

This change allows the override of the port used by the webhook deployment when using the Helm chart. This is useful in clusters where the AWS CNI is not used (host networking is required) and we want to ensure pods don't accidentally step on each other's toes if we have other webhook admission controllers. We're specifically mapping the ports used by these controllers to prevent these types of collisions, so having the ability to override the port is quite useful.

- [X] No


ellistarn (approved):
Nice!

ellistarn:
@tuananh your helm chart CI checks are already coming in handy :)

ellistarn (approved):

mikesir87:
Haha... yeah! I'll have to look at how you're doing that because I know we run into similar problems. Certainly a nice check to have in place.

On another note... I didn't do any bumping of the chart version. I assumed you'd take care of that whenever you do your next release (which looks to be on the same lifecycle as karpenter itself)?

ellistarn:
Yes -- for now. Planning on cutting a release this afternoon.

ellistarn:
> Haha... yeah! I'll have to look at how you're doing that because I know we run into similar problems. Certainly a nice check to have in place.

Just a simple dirty check that's run on `make dev` and `make ci`: https://github.com/aws/karpenter/blob/main/Makefile#L42

PR #911, opened 2021-12-03
v0.5.1

ellistarn:
**1. Issue, if available:**

- [ ] No


njtran (changes requested):
Can you make the changes to the Getting Started guide as well?

ellistarn:
Hey @tuananh, I found a way to simplify the helm README to just generate `helm-docs` whenever we run `make helm`. Since chart is published alongside release versions, we only want to update the README when we release a new version of the chart. This is important to avoid version skew. 

I just wanted to check that I'm not missing anything from you change and get your blessing on these changes.

ellistarn:
> Can you make the changes to the Getting Started guide as well?

Let's cut docs in a followon -- it's a bit complicated.

njtran (approved):

tuananh:
@ellistarn this works for the current manual release i guess. but later on, if you want to automate the release process of helm chart, it'll get a bit complicated because we then need CI to comit back the changes of helm-docs back which i don't think is a good idea.

what i have in mind is
- check helm-docs for every pr
- make sure the chart version bump is getting check too.

then we can automate the chart release progress, without waiting for karpenter. since i figure chart release is usually a lot more frequent than karpenter itself.

ellistarn:
> @ellistarn this works for the current manual release i guess. but later on, if you want to automate the release process of helm chart, it'll get a bit complicated because we then need CI to comit back the changes of helm-docs back which i don't think is a good idea.
> 
> what i have in mind is
> 
> * check helm-docs for every pr
> * make sure the chart version bump is getting check too.
> 
> then we can automate the chart release progress, without waiting for karpenter. since i figure chart release is usually a lot more frequent than karpenter itself.

Good points. Let's cover this in working group.

tuananh:
Sure can :)

Issue #909, opened 2021-12-03
Set default instance profile


mikesir87:
**Tell us about your request**
I would like to have the ability to set a default instance profile, making it optional when defining Provisioner objects.

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
When we deploy Karpenter using our pipeline, we use TF to create the dependencies (IAM roles, instance profile, etc.) and deploy Karpenter using a `helm_release` resource. It would be nice if I could pass in the created instance profile as an argument to make it the default profile. This will prevent the sharing of a magic/well-known name with our separate Helm chart that configures all of our tenant resources (has no access to this pipeline and its outputs).

**Are you currently working around this issue?**
Simply sharing the name of the profile and hoping for the best! 😆 

**Additional context**
I do recognize that this is a provider specific field, but do see other env vars that are being passed in that are provider specific. So, I don't feel it'll muddy the waters too much asking for this in that regard.

**Attachments**
None

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ellistarn:
I'm completely supportive of this idea. We recently moved `clusterName` and `clusterEndpoint` from a CRD field to a binary flag. It should be straightforward to wire it up here: https://github.com/aws/karpenter/blob/a859cad1e3b3c5cc916f69bb2435198a3cd0af13/pkg/utils/options/options.go#L29. 

We still need to have the ability to override InstanceProfile at the CRD level as there are use cases for running nodes with different roles in different provisioners. However, we can make this an optional field.

Is this something you'd be able to contribute?

mikesir87:
Awesome! And yeah... makes sense to allow overrides at the CRD level as well.

I'll gladly contribute it. Just wanted to get the go-ahead before jumping in. I'll see if I can send something your way here soon!

PR #914, opened 2021-12-03
Allowing settings of a default instance profile

mikesir87:
**1. Issue, if available:**
#909 

- Adds the ability to set `-default-instance-profile` on the controller, which does as it suggests
- Update default constraints setting to use the default if one is provided
- Updated the Helm chart to allow setting of this flag and the website to provide a little bit of documentation about it
- Updated unit tests to test for both default setting and for overriding through a Provisioner

- [ ] No

I'll fully admit I'm not the best Go programmer, so feel free to tell me how it should be made better. I tried to follow along with what was already there. But, might have missed an easier way to do something.

### DIFF
--- a/charts/karpenter/README.md
+++ b/charts/karpenter/README.md
@@ -25,6 +25,7 @@ You can follow the detailed installation instruction [here](https://karpenter.sh
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | additionalLabels | object | `{}` | Additional labels to add into metadata |
+| aws.defaultInstanceProfile | string | `""` | The default instance profile to use when launching nodes on AWS |
 | controller.affinity | object | `{}` | Affinity rules for scheduling |
 | controller.clusterEndpoint | string | `""` | Cluster endpoint |
 | controller.clusterName | string | `""` | Cluster name |
--- a/charts/karpenter/templates/controller/deployment.yaml
+++ b/charts/karpenter/templates/controller/deployment.yaml
@@ -63,6 +63,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            {{- if .Values.aws.defaultInstanceProfile }}
+            - name: AWS_DEFAULT_INSTANCE_PROFILE
+              value: {{ .Values.aws.defaultInstanceProfile }}
+            {{- end }}
             {{- with .Values.controller.env }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -55,3 +55,6 @@ webhook:
       cpu: 100m
       memory: 50Mi
   replicas: 1
+aws:
+  # -- The default instance profile to use when launching nodes on AWS
+  defaultInstanceProfile: ""
--- a/pkg/cloudprovider/aws/apis/v1alpha1/provider.go
+++ b/pkg/cloudprovider/aws/apis/v1alpha1/provider.go
@@ -43,8 +43,8 @@ type AWS struct {
 	// TypeMeta includes version and kind of the extensions, inferred if not provided.
 	// +optional
 	metav1.TypeMeta `json:",inline"`
-	// InstanceProfile is the AWS identity that instances use.
-	// +required
+	// InstanceProfile to use for provisioned nodes, overriding the default profile.
+	// +optional
 	InstanceProfile string `json:"instanceProfile"`
 	// LaunchTemplate for the node. If not specified, a launch template will be generated.
 	// +optional
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -89,6 +90,10 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, constraints *v1alpha1.
 	if constraints.LaunchTemplate != nil {
 		return map[string][]cloudprovider.InstanceType{ptr.StringValue(constraints.LaunchTemplate): instanceTypes}, nil
 	}
+	instanceProfile, err := p.getInstanceProfile(ctx, constraints)
+	if err != nil {
+		return nil, err
+	}
 	// Get constrained security groups
 	securityGroupsIds, err := p.securityGroupProvider.Get(ctx, constraints)
 	if err != nil {
@@ -111,7 +116,7 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, constraints *v1alpha1.
 		launchTemplate, err := p.ensureLaunchTemplate(ctx, &launchTemplateOptions{
 			UserData:          userData,
 			ClusterName:       injection.GetOptions(ctx).ClusterName,
-			InstanceProfile:   constraints.InstanceProfile,
+			InstanceProfile:   instanceProfile,
 			AMIID:             amiID,
 			SecurityGroupsIds: securityGroupsIds,
 			Tags:              constraints.Tags,
@@ -310,6 +315,17 @@ func (p *LaunchTemplateProvider) getNodeTaintArgs(constraints *v1alpha1.Constrai
 		}
 	}
 	return nodeTaintsArgs
+}
+
+func (p *LaunchTemplateProvider) getInstanceProfile(ctx context.Context, constraints *v1alpha1.Constraints) (string, error) {
+	if constraints.InstanceProfile != "" {
+		return constraints.InstanceProfile, nil
+	}
+	defaultProfile := injection.GetOptions(ctx).AWSDefaultInstanceProfile
+	if defaultProfile == "" {
+		return "", errors.New("neither spec.provider.instanceProfile nor --aws-default-instance-profile is specified")
+	}
+	return defaultProfile, nil
 }
 
 func (p *LaunchTemplateProvider) GetCABundle(ctx context.Context) (*string, error) {
--- a/pkg/cloudprovider/aws/suite_test.go
+++ b/pkg/cloudprovider/aws/suite_test.go
@@ -69,10 +69,11 @@ func TestAPIs(t *testing.T) {
 var _ = BeforeSuite(func() {
 	env = test.NewEnvironment(ctx, func(e *test.Environment) {
 		opts = options.Options{
-			ClusterName:             "test-cluster",
-			ClusterEndpoint:         "https://test-cluster",
-			AWSNodeNameConvention:   string(options.IPName),
-			AWSENILimitedPodDensity: true,
+			ClusterName:               "test-cluster",
+			ClusterEndpoint:           "https://test-cluster",
+			AWSNodeNameConvention:     string(options.IPName),
+			AWSENILimitedPodDensity:   true,
+			AWSDefaultInstanceProfile: "test-instance-profile",
 		}
 		Expect(opts.Validate()).To(Succeed(), "Failed to validate options")
 		ctx = injection.WithOptions(ctx, opts)
@@ -122,7 +123,6 @@ var _ = Describe("Allocation", func() {
 
 	BeforeEach(func() {
 		provider = &v1alpha1.AWS{
-			InstanceProfile:       "test-instance-profile",
 			SubnetSelector:        map[string]string{"foo": "bar"},
 			SecurityGroupSelector: map[string]string{"foo": "bar"},
 		}
@@ -506,6 +506,27 @@ var _ = Describe("Allocation", func() {
 					Expect(string(userData)).To(ContainSubstring("--dns-cluster-ip '10.0.10.100'"))
 				})
 			})
+			Context("Instance Profile", func() {
+				It("should use the default instance profile if none specified on the Provisioner", func() {
+					provisioner.Spec.KubeletConfiguration.ClusterDNS = []string{"10.0.10.100"}
+					pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioners, ProvisionerWithProvider(provisioner, provider), test.UnschedulablePod())[0]
+					ExpectScheduled(ctx, env.Client, pod)
+					Expect(fakeEC2API.CalledWithCreateLaunchTemplateInput.Cardinality()).To(Equal(1))
+					input := fakeEC2API.CalledWithCreateLaunchTemplateInput.Pop().(*ec2.CreateLaunchTemplateInput)
+					Expect(*input.LaunchTemplateData.IamInstanceProfile.Name).To(Equal("test-instance-profile"))
+				})
+				It("should use the instance profile on the Provisioner when specified", func() {
+					provider = &v1alpha1.AWS{InstanceProfile: "overridden-profile"}
+					ProvisionerWithProvider(&v1alpha5.Provisioner{ObjectMeta: metav1.ObjectMeta{Name: strings.ToLower(randomdata.SillyName())}}, provider)
+					provisioner.SetDefaults(ctx)
+
+					pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioners, ProvisionerWithProvider(provisioner, provider), test.UnschedulablePod())[0]
+					ExpectScheduled(ctx, env.Client, pod)
+					Expect(fakeEC2API.CalledWithCreateLaunchTemplateInput.Cardinality()).To(Equal(1))
+					input := fakeEC2API.CalledWithCreateLaunchTemplateInput.Pop().(*ec2.CreateLaunchTemplateInput)
+					Expect(*input.LaunchTemplateData.IamInstanceProfile.Name).To(Equal("overridden-profile"))
+				})
+			})
 		})
 		Context("Metadata Options", func() {
 			It("should default metadata options on generated launch template", func() {
@@ -539,6 +560,14 @@ var _ = Describe("Allocation", func() {
 		})
 	})
 	Context("Defaulting", func() {
+		// Intent here is that if updates occur on the controller, the Provisioner doesn't need to be recreated
+		It("should not set the InstanceProfile with the default if none provided in Provisioner", func() {
+			provisioner.SetDefaults(ctx)
+			constraints, err := v1alpha1.Deserialize(&provisioner.Spec.Constraints)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(constraints.InstanceProfile).To(Equal(""))
+		})
+
 		It("should default requirements", func() {
 			provisioner.SetDefaults(ctx)
 			Expect(provisioner.Spec.Requirements.CapacityTypes().UnsortedList()).To(ConsistOf(v1alpha1.CapacityTypeOnDemand))
--- a/pkg/utils/options/options.go
+++ b/pkg/utils/options/options.go
@@ -41,6 +41,7 @@ func MustParse() Options {
 	flag.IntVar(&opts.KubeClientBurst, "kube-client-burst", env.WithDefaultInt("KUBE_CLIENT_BURST", 300), "The maximum allowed burst of queries to the kube-apiserver")
 	flag.StringVar(&opts.AWSNodeNameConvention, "aws-node-name-convention", env.WithDefaultString("AWS_NODE_NAME_CONVENTION", string(IPName)), "The node naming convention used by the AWS cloud provider. DEPRECATION WARNING: this field may be deprecated at any time")
 	flag.BoolVar(&opts.AWSENILimitedPodDensity, "aws-eni-limited-pod-density", env.WithDefaultBool("AWS_ENI_LIMITED_POD_DENSITY", true), "Indicates whether new nodes should use ENI-based pod density")
+	flag.StringVar(&opts.AWSDefaultInstanceProfile, "aws-default-instance-profile", env.WithDefaultString("AWS_DEFAULT_INSTANCE_PROFILE", ""), "The default instance profile to use when provisioning nodes in AWS")
 	flag.Parse()
 	if err := opts.Validate(); err != nil {
 		panic(err)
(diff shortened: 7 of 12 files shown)


charts/karpenter/templates/controller/deployment.yaml
```diff
@@ -32,6 +32,10 @@ spec:
       containers:
         - name: manager
           image: {{ .Values.controller.image }}
+        {{- if .Values.controller.defaultInstanceProfile }}
+          args:
+            - -default-instance-profile={{ .Values.controller.defaultInstanceProfile }}
```
ellistarn, line 37:
```suggestion
            - --default-instance-profile={{ .Values.controller.defaultInstanceProfile }}
```

charts/karpenter/values.yaml
```diff
@@ -22,6 +22,8 @@ controller:
   clusterName: ""
   # -- Cluster endpoint
   clusterEndpoint: ""
+  # -- Default instance profile
+  defaultInstanceProfile: ""
```
ellistarn, line 26: Thoughts on putting a note that clusterEndpoint, clusterName, and defaultInstanceProfile are all required?

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
```diff
@@ -34,8 +34,8 @@ type AWS struct {
 	// TypeMeta includes version and kind of the extensions, inferred if not provided.
 	// +optional
 	metav1.TypeMeta `json:",inline"`
-	// InstanceProfile is the AWS identity that instances use.
-	// +required
+	// InstanceProfile is the AWS identity that instances use. Required if no default is configured on the controller.
```
ellistarn, line 37: Thoughts on requiring the defaultInstanceProfile to be set. I think that semantic is a little more easily consumable.

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -389,6 +389,21 @@ var _ = Describe("Allocation", func() {
...
+
+			constraints, err := v1alpha1.Deserialize(&provisioner.Spec.Constraints)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(constraints.InstanceProfile).To(Equal("default-profile"))
+		})
+		It("should use overridden instanceProfile if provided in Provider", func() {
+			provisioner.SetDefaults(ctx)
+			constraints, err := v1alpha1.Deserialize(&provisioner.Spec.Constraints)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(constraints.InstanceProfile).To(Equal("test-instance-profile"))
```
ellistarn, line 405: Curious how this test passes. Don't you need to set this in the provisioner before defaulting?

pkg/utils/options/options.go
```diff
@@ -27,6 +27,7 @@ func MustParse() Options {
 	opts := Options{}
 	flag.StringVar(&opts.ClusterName, "cluster-name", env.WithDefaultString("CLUSTER_NAME", ""), "The kubernetes cluster name for resource discovery")
 	flag.StringVar(&opts.ClusterEndpoint, "cluster-endpoint", env.WithDefaultString("CLUSTER_ENDPOINT", ""), "The external kubernetes cluster endpoint for new nodes to connect with")
+	flag.StringVar(&opts.DefaultInstanceProfile, "default-instance-profile", env.WithDefaultString("DEFAULT_INSTANCE_PROFILE", ""), "The default instance profile during provisioning if none provided in a Provisioner")
```
ellistarn, line 30:
```suggestion
	flag.StringVar(&opts.DefaultInstanceProfile, "default-instance-profile", env.WithDefaultString("DEFAULT_INSTANCE_PROFILE", ""), "The default instance profile which nodes run as")
```

pkg/utils/options/options.go
```diff
@@ -41,13 +42,14 @@ func MustParse() Options {
...
-	KubeClientQPS   int
-	KubeClientBurst int
+	ClusterName            string
+	ClusterEndpoint        string
+	MetricsPort            int
+	HealthProbePort        int
+	WebhookPort            int
+	KubeClientQPS          int
+	KubeClientBurst        int
+	DefaultInstanceProfile string
```
ellistarn, line 52: This makes me think about neutrality as we add more vendor specific flags. I wonder if there's a way for us to register cloud provider flags into the binary somehow. We can leave this for another day.

website/content/en/docs/AWS/constraints.md
```diff
@@ -11,9 +11,9 @@ This section covers parameters of the AWS Cloud Provider.
 [Review these fields in the code.](https://github.com/awslabs/karpenter/blob/main/pkg/cloudprovider/aws/apis/v1alpha1/provider.go#L33)
 
 ### InstanceProfile
-An `InstanceProfile` is a way to pass a single IAM role to an EC2 instance.
+An `InstanceProfile` is a way to pass a single IAM role to an EC2 instance. A default instance profile can be set on the controller (by specifying the `-default-instance-profile` argument) and overridden through Provisioner settings.
 
-It is required, and specified by name. A suitable `KarpenterNodeRole` is created in the getting started guide.
+It is required, if no default was specified with the controller, and specified by name. A suitable `KarpenterNodeRole` is created in the getting started guide.
```
ellistarn, line 16: See above re: the semantic flip.

ellistarn (commented):
Wow that was fast 💪! A couple minor comments.

mikesir87:
Right on! I like the idea of making the default profile a required attribute and make the Provisioner's overriding completely optional. That's obviously a breaking change going forward, but I guess it's early enough to make those kind of changes.

If we want to do that, I can also update the "Getting Started" tutorial to set that flag as well. I can do a run-through of the docs to see what else might need to be tweaked as well.

pkg/cloudprovider/aws/suite_test.go
mikesir87: That's because this test is using the Provisioner set up earlier, which was specifying the InstanceProfile. I'll probably flip this test if we flip the logic and make the default required.

pkg/utils/options/options.go
mikesir87: I thought about that too. But... will leave that one up to you all to figure out! :laughing:

ellistarn:
> Right on! I like the idea of making the default profile a required attribute and make the Provisioner's overriding completely optional. That's obviously a breaking change going forward, but I guess it's early enough to make those kind of changes.

Technically, it's only breaking to the helm chart, and not to the API. We're relaxing the requirement that an instance profile must be specified in the API. We're introducing a new requirement that defaultinstanceprofile must be specified as a flag. 

If we do need to make breaking changes, that will result in an API bump, which as you mention, we have some wiggle room on.

> If we want to do that, I can also update the "Getting Started" tutorial to set that flag as well. I can do a run-through of the docs to see what else might need to be tweaked as well.

Just as a quick primer, all doc edits are committed to the `predocs` folder. Upon release, we copy them into a release specific folder, and change the pointer on the "latest version" dropdown. Because we don't have automation in place for this (yet) v0.5.1 aren't cut yet. I'll do that right now.

mikesir87:
Alrighty! Just pushed changes that make the `--default-instance-profile` required on the controller (validations and unit tests) and in the Helm chart (I also ensured the `clusterName` and `clusterEndpoint` are specified too). I also updated the Getting Started guide for the docs to specify the profile generated by the CF template.

Feel free to take a look and let me know what else might need to be adjusted!

And... feel free to wait until after the weekend to do any more work on it. I just knew I wouldn't be able to fully relax until I finished up my end 😆 Thanks and enjoy your weekend!

website/content/en/docs/AWS/constraints.md
```diff
@@ -11,9 +11,7 @@ This section covers parameters of the AWS Cloud Provider.
 [Review these fields in the code.](https://github.com/awslabs/karpenter/blob/main/pkg/cloudprovider/aws/apis/v1alpha1/provider.go#L33)
 
 ### InstanceProfile
-An `InstanceProfile` is a way to pass a single IAM role to an EC2 instance.
-
-It is required, and specified by name. A suitable `KarpenterNodeRole` is created in the getting started guide.
+An `InstanceProfile` is a way to pass a single IAM role to an EC2 instance. This can be specified to override the default profile specified during installation.
```
geoffcline, line 14: note to self: add link or example giving some indication of what "specified during installation" means -- perhaps helm chart readme?

charts/karpenter/README.md
```diff
@@ -15,7 +15,8 @@ $ helm upgrade --install karpenter karpenter/karpenter --namespace karpenter \
   --create-namespace --set serviceAccount.create=false --version 0.5.1 \
   --set controller.clusterName=${CLUSTER_NAME} \
   --set controller.clusterEndpoint=$(aws eks describe-cluster --name ${CLUSTER_NAME} --query "cluster.endpoint" --output json) \
-  --wait # for the defaulting webhook to install before creating a Provisioner 
+  --set controller.defaultInstanceProfile=KarpenterNodeInstanceProfile-${CLUSTER_NAME} \
```
ellistarn, line 18: FYI, this is generated by `make helm`. You might already know this, since it looks correctly autogenerated.

charts/karpenter/templates/controller/deployment.yaml
```diff
@@ -32,6 +32,8 @@ spec:
       containers:
         - name: manager
           image: {{ .Values.controller.image }}
+          args:
+            - --default-instance-profile={{ required "controller.defaultInstanceProfile is required to be set" .Values.controller.defaultInstanceProfile }}
```
ellistarn, line 36: I *think* the webhook needs all of this as well, since the defaulting logic happens there.

ellistarn (commented):
Great work! One minor comment.

ellistarn:
> And... feel free to wait until after the weekend to do any more work on it. I just knew I wouldn't be able to fully relax until I finished up my end 😆 Thanks and enjoy your weekend!

I think I may have an addiction.

charts/karpenter/templates/controller/deployment.yaml
mikesir87: Yes... it does look like the webhook needs this. Since we're already making a breaking change in the Helm chart, would it make sense to move these options out of the controller tree in the values? I'm thinking something like `options.defaultInstanceProfile`, `options.ClusterName`, and `options.clusterEndpoint`. Thoughts?

mikesir87:
> I think I may have an addiction.

Haha... I _totally_ know the feeling! Especially for a new project like this! 😆

pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
```diff
@@ -75,3 +76,10 @@ func (c *Constraints) defaultSecurityGroups(clusterName string) {
 	}
 	c.SecurityGroupSelector = map[string]string{fmt.Sprintf(ClusterDiscoveryTagKeyFormat, clusterName): "*"}
 }
+
+func (c *Constraints) defaultInstanceProfile(defaultInstanceProfile string) {
```
ellistarn, line 80: This will cause some drift if the karpenter controller is restarted w/ a new default-instance-profile. Any provisioner created on the previous install will stay on the previous value, which may be surprising. 

I'd probably just leave any defaulting out and just use the value if it happens to be set, or fallback to GetOptions(ctx). WDYT?

mikesir87: Ah! Good catch there. I'll admit I hadn't thought of that as I'm still fairly new to controller development. I'll send something your way in just a few mins 👍

ellistarn:
Another thing I just realized is that the controller may not play nice with
`spec.provider = nil`. Can you verify this behavior and make sure this doesn't crash? It wasn't possible to be nil before because the instanceProfile was validated.

mikesir87:
> Another thing I just realized is that the controller may not play nice with `spec.provider = nil`. Can you verify this behavior and make sure this doesn't crash? It wasn't possible to be nil before because the instanceProfile was validated.

Good point. I'll try that out and let ya know. I'm going to send another PR too that adds a `make local-build` so it's easier to build and test without needing release credentials.

ellistarn:
> Good point. I'll try that out and let ya know. I'm going to send another PR too that adds a `make local-build` so it's easier to build and test without needing release credentials.

I use `make apply` to push my changes and verify things, and `make dev` to run the tests.

mikesir87:
> I use `make apply` to push my changes and verify things against a real cluster, and `make dev` to run the tests.

Ah! I'll give that a try! And another spot that I need to change the Helm value keys.

pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
mikesir87: So yeah... there is a problem if the `spec.provider` is empty. While I can obviously skip the decoding, I'm not sure how we should handle the skipping of the `SetGroupVersionKind`. Thoughts?

https://github.com/aws/karpenter/blob/5392413df2fc6963ea1f80a6cc88970c73abcda2/pkg/cloudprovider/aws/apis/v1alpha1/provider.go#L54-L67

ellistarn: If it's nil, I think we can just leave it fully empty. We can just return an empty constraints object for now. I haven't fully thought it through yet, but this is one of my least favorite pieces of our API. I think we will likely separate it out into a new CRD and make `provider` into `providerRef` or something.

pkg/utils/options/options.go
```diff
@@ -27,6 +27,7 @@ func MustParse() Options {
 	opts := Options{}
 	flag.StringVar(&opts.ClusterName, "cluster-name", env.WithDefaultString("CLUSTER_NAME", ""), "The kubernetes cluster name for resource discovery")
 	flag.StringVar(&opts.ClusterEndpoint, "cluster-endpoint", env.WithDefaultString("CLUSTER_ENDPOINT", ""), "The external kubernetes cluster endpoint for new nodes to connect with")
+	flag.StringVar(&opts.DefaultInstanceProfile, "default-instance-profile", env.WithDefaultString("DEFAULT_INSTANCE_PROFILE", ""), "The default instance profile to use when provisioning nodes")
```
bwagner5, line 30: thoughts on prepending this w/ `AWS`? I would think when we do have cloud providers register flags, we'll want to namespace them within their cloud provider, so prepending with the cloud provider now may save us a backwards incompatible change later :)

pkg/utils/options/options.go
```diff
@@ -41,20 +42,24 @@ func MustParse() Options {
...
+	KubeClientBurst        int
+	DefaultInstanceProfile string
 }
 
 func (o Options) Validate() (err error) {
 	err = multierr.Append(err, o.validateEndpoint())
 	if o.ClusterName == "" {
 		err = multierr.Append(err, fmt.Errorf("CLUSTER_NAME is required"))
 	}
+	if o.DefaultInstanceProfile == "" {
```
bwagner5, line 60: This is not required right? IIUC, It's required to be passed as a CLI arg OR provided in the provisioner. 

Or am I mistaken and we're requiring a default one passed in the CLI arg?

mikesir87: That was something we talked about earlier in the review. I had it originally as you mentioned, with the default at the controller being optional. We changed it to an override. But yeah... that does affect it's coupling to a specific cloud provider. 

We don't have any mechanism to validate the flags based on provider, but maybe that's an idea too? For AWS-based deploys, it's required?

pkg/utils/options/options.go
mikesir87: I'm good with this idea and can make the adjustment

mikesir87:
Just pushed a rebase onto latest `main`, as well as adding the `aws` prefix to the CLI flag and other locations (Helm chart). PTAL and let me know what you'd like adjusted 👍

pkg/utils/options/options.go
bwagner5: I could see a use-case of provisioners where they are handed out per team but a central team manages the Karpenter install. As the central team, I don't know if I'd want to put a default InstanceProfile in the controller for all provisioners to inherit as a default value. Seems too easy to make a mistake and have a provisioner inherit a broader set of permissions than desired. You could make cases for using IRSA and applying best practices of only doing a least privileges profile in the controller options, but it won't always happen. 

Maybe a default instanceProfile doesn't exist for a use-case. Maybe a platform team will craft individual roles for each provisioner and if the argument is required on the controller, they now have to specify a dummy profile that wouldn't work with a node. 

If DefaultInstanceProfile is optional on the controller, the webhook defaulting logic could validate the provisioner based on both values where the provisioner one wins since it's more specific. The webhook should have all of that knowledge at validation time.

mikesir87: I agree with the sentiment and had started with that idea, having the default as optional. The tricky part is that there's no guarantee the webhook and controller are configured the same way, since they are deployed separately (granted the Helm chart will make sure they are configured similarly). But, there's a chance the webhook doesn't know a default that the controller might be using. 

I'm good with the idea of reverting it back (as it makes it less backwards-disruptive for those wanting to update). Just want to make sure everyone is good with this approach then. Cc/ @ellistarn

charts/karpenter/values.yaml
```diff
@@ -5,6 +5,13 @@ serviceAccount:
   name: karpenter
   # -- Annotations to add to the service account (like the ARN of the IRSA role)
   annotations: {}
+options:
+  # -- Required. Cluster endpoint
+  clusterEndpoint: ""
+  # -- Required. Cluster name
+  clusterName: ""
+  # -- Required. The default instance profile to use when launching nodes on AWS
```
ellistarn, line 13: If required, do we need to include the required directive in the chart?

charts/karpenter/values.yaml
```diff
@@ -18,10 +25,11 @@ controller:
   affinity: {}
   # -- Image to use for the Karpenter controller
   image: "public.ecr.aws/karpenter/controller:v0.5.2@sha256:6b08af0fbeab6da8d9b97ebfb0750bc72a3ac4a97dd0337a1607048292425c48"
-  # -- Cluster name
   clusterName: ""
-  # -- Cluster endpoint
+  # -- Required. Cluster endpoint
   clusterEndpoint: ""
+  # -- Required. The default instance profile to use when launching nodes
+  defaultInstanceProfile: ""
```
ellistarn, line 32: stray?

charts/karpenter/values.yaml
mikesir87: That's kinda the question right now... should it be required (see previous thread). There's obviously a mismatch here that would need to be fixed either way. Good catch on that

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -102,14 +102,15 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, constraints *v1alpha1.
...
+		instanceProfile := p.getInstanceProfile(ctx, constraints)
 		if err != nil {
 			return nil, err
 		}
 		// Ensure the launch template exists, or create it
 		launchTemplate, err := p.ensureLaunchTemplate(ctx, &launchTemplateOptions{
 			UserData:          userData,
 			ClusterName:       injection.GetOptions(ctx).ClusterName,
-			InstanceProfile:   constraints.InstanceProfile,
+			InstanceProfile:   instanceProfile,
```
ellistarn, line 113: consider just inlining 
```
InstanceProfile: p.getInstanceProfile(ctx, constraints)
```

ellistarn (commented):
A few minor changes. Otherwise LGTM

pkg/utils/options/options.go
mikesir87: To resolve and document the resolution to this discussion, all three of us chatted on Slack and determined the best approach going forward is to make the setting of the instance profile optional at both the API and CLI. We will then try and fail early in the runtime reconciliation. With this approach, we don't have to worry about defaults coming and going, but also have the flexibility for various use cases.

charts/karpenter/values.yaml
mikesir87: Fixed.

pkg/cloudprovider/aws/launchtemplate.go
mikesir87: Adjusted with latest commit, as the function now returns an error if no usable profile was found.

charts/karpenter/values.yaml
mikesir87: Yup and removed.

mikesir87:
Ok. Made all of the changes we discussed. The only thing we might want to include in any release notes is the change of the Helm chart values from `controller.clusterName`/`controller.clusterEndpoint` to `options.clusterName` and `options.clusterEndpoint`. The idea there was that the values are being used on both the webhook and controller, so letting the naming clear that up a little bit.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -284,6 +289,17 @@ exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
 	return base64.StdEncoding.EncodeToString(userData.Bytes()), nil
 }
 
+func (p *LaunchTemplateProvider) getInstanceProfile(ctx context.Context, constraints *v1alpha1.Constraints) (string, error) {
+	if constraints.InstanceProfile != "" {
+		return constraints.InstanceProfile, nil
+	}
+	defaultProfile := injection.GetOptions(ctx).AwsDefaultInstanceProfile
+	if defaultProfile == "" {
+		return "", errors.New("no instance profile found on provisioner and no default profile set")
```
ellistarn, line 298:
```suggestion
		return "", errors.New("neither spec.provider.instanceProfile nor aws-default-instance-profile is specified.")
```

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -284,6 +289,17 @@ exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
 	return base64.StdEncoding.EncodeToString(userData.Bytes()), nil
 }
 
+func (p *LaunchTemplateProvider) getInstanceProfile(ctx context.Context, constraints *v1alpha1.Constraints) (string, error) {
+	if constraints.InstanceProfile != "" {
+		return constraints.InstanceProfile, nil
+	}
+	defaultProfile := injection.GetOptions(ctx).AwsDefaultInstanceProfile
+	if defaultProfile == "" {
+		return "", errors.New("neither spec.provider.instanceProfile nor aws-default-instance-profile is specified")
```
ellistarn, line 298:
```suggestion
		return "", errors.New("neither spec.provider.instanceProfile nor --aws-default-instance-profile is specified")
```

ellistarn (dismissed):

ellistarn (approved):

mikesir87:
@ellistarn - just pushed a rebase, but also made a few adjustments due to other changes going on in the project. Quick summary:

- No longer handling a `null` provider since subnetSelector and securityGroupSelectors are required and the provider will no longer be null
- Renamed the option from `AwsDefaultInstanceProfile` to `AWSDefaultInstanceProfile` to align with other names
- I had previously moved the shared clusterEndpoint/Name fields in the Helm chart under an `options` key so they would be shared between the controller and webhook. That felt funky, so reverted that to no change. But, did create a new `aws.defaultInstanceProfile` value. We can use to hang other AWS-specific options, such as the new ENI pod density option.

ellistarn (approved):
Nice work!

bwagner5 (approved):
lgtm, thanks for this addition!

Issue #905, opened 2021-12-03
Satisfied Labels


bdwyertech:
Allow setting "satisfied" labels in the provisioner config.  Some labels are set automatically or by another process, e.g. `kubernetes.io/os`.  I have some pods which use node selectors such as this, and I cannot set these labels as these are reserved/well-known domains, so Karpenter does not know my provisioner will satisfy the condition.

![image](https://user-images.githubusercontent.com/2973273/144548628-c9aaa791-ec9e-434b-b4fc-96b79f150e1a.png)

Maybe we could have a `labels_satisfied` section?

```yaml
labels_satisified:
  kubernetes.io/os: linux
```

**Are you currently working around this issue?**
Nope.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ellistarn:
Hmm -- is this necessary because you don't have control over these labels somehow or because you're trying to colocate on a mixed capacity cluster (non-karpenter). 

We used to support the kubernetes.io/os as a well known label, but since linux was the only supported value, we removed it. At the end of the day, we only prevent you from using well known labels -- you have full control over custom labels. If the label is well known, and useful from a scheduling perspective, we should probably just support the label.

Are there other labels other than os that you need? I'm a bit wary of supporting a new concept like "label satisfaction", and instead would lean towards just supporting OS. If you have examples of labels that would be harder for us to support as well known labels, I'd be very interested to hear what they are.

bdwyertech:
Yeah, its because I don't have control of this specific label.  For now it's just that one, however I am curious about some others.  For example, `vpc.amazonaws.com/has-trunk-attached=true`.  Only certain instance types support Branch ENI.

With cluster-autoscaler, you would tag your ASG with labels such as this to tell the cluster-autoscaler which labels it would satisfy.  How can we achieve same with Karpenter?

ellistarn:
`vpc.amazonaws.com/has-trunk-attached=true` is very interesting. We do have a concept of "Well Known Labels". These sound like well known labels to me, and could be used to filter instances. A couple examples:

- gpu-family
- instance-type-family
- has-nvme
- has-trunk-attached

[...]

PR #924, opened 2021-12-06
Pod ENI

bdwyertech:
**1. Issue, if available:**
#905 

This improves instance type selection to factor in [Security Groups for Pods](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html) by ensuring only instance types capable of Pod ENI are launched if set in a pods resource request.

This leverages the lookup table in [vpc-resource-controller](https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/05e89ff9300a5cc0ebea705834cf27f0a7f3b509/pkg/aws/vpc/limits.go#L27) to determine Pod ENI support.

- [ ] No


ellistarn:
This is a pretty cool WIP! I'm assuming that `aws.amazon.com/pod-eni` is already written onto nodes by the CNI controller and works w/ the kube scheduler?

bdwyertech:
> This is a pretty cool WIP! I'm assuming that `aws.amazon.com/pod-eni` is already written onto nodes by the CNI controller and works w/ the kube scheduler?

So the namespace is up for debate.  When the CNI first initializes, it sets `vpc.amazonaws.com/has-trunk-attached=false` and stays this way until a pod runs on it which requires a branch interface, then it flips over to `true`.  This label cannot really be used to make a determination of Pod ENI support.

If we want to stay out of the `aws.amazon.com` namespace, maybe `aws.karpenter.sh/pod-eni` or `aws.karpenter.sh/pod-eni-capable` would be more appropriate.

I was leveraging the initial node template to set the value initially and then reverted it back.

https://github.com/aws/karpenter/blob/e8537a3622dac212de4bb7a68f31c07dbcdbe8fd/pkg/cloudprovider/aws/instance.go#L240-L250

ellistarn:
Awesome! This is shaping up really nicely!

bdwyertech:
I was kind of surprised by the hard-coded lookup table in the CNI and that the new m6i and m6a instances don't support this!

ellistarn:
Ready for review?

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -121,6 +121,20 @@ var _ = Describe("Allocation", func() {
 
 	Context("Reconciliation", func() {
 		Context("Specialized Hardware", func() {
+			It("should not launch AWS Pod ENI on a t3", func() {
```
ellistarn, line 124: Can you have a positive test that shows that it does schedule w/ this requirement?

ellistarn (commented):
One minor comment. Otherwise great job!

pkg/controllers/provisioning/binpacking/packable.go
```diff
@@ -242,7 +242,7 @@ func (p *Packable) validateAWSNeurons(pods []*v1.Pod) error {
 func (p *Packable) validateAWSPodENI(pods []*v1.Pod) error {
 	for _, pod := range pods {
 		for _, container := range pod.Spec.Containers {
-			if _, ok := container.Resources.Requests[resources.AWSPodENI]; ok {
+			if _, ok := container.Resources.Limits[resources.AWSPodENI]; ok {
```
ellistarn, line 245: I think you want to use Requests here. Kubernetes will default requests to the value of limits if only limits are set.

pkg/cloudprovider/aws/suite_test.go
bdwyertech: Done!

pkg/controllers/provisioning/binpacking/packable.go
bdwyertech: Yup, I tested and you're right.  swapped it back.  Saw an issue when i set requests but not limit on the client side.

ellistarn (approved):
Nicely done!

PR #931, opened 2021-12-06
First phase of scheduler refactor

ellistarn:
**1. Issue, if available:**

- [ ] No


pkg/controllers/provisioning/provisioner.go
```diff
@@ -161,3 +190,91 @@ func (p *Provisioner) FilterProvisionable(ctx context.Context, pods []*v1.Pod) [
 	}
 	return provisionable
 }
+
+func (p *Provisioner) launch(ctx context.Context, constraints *v1alpha5.Constraints, packing *binpacking.Packing) error {
```
suket22, line 194: Why did we collapse the launcher into this file btw? Don't feel too strongly about it either way

suket22 (approved):

pkg/controllers/provisioning/provisioner.go
ellistarn: It will become more apparent as I continue to refactor this, but essentially the logic got small enough that it didn't really make sense to have all of the overhead ~70 lines of code. Additionally, I'm going to be moving other provisioner behavior (e.g. batching/scheduling) elsewhere. You can think about the provisioner as becoming a renamed launcher.

Issue #942, opened 2021-12-08
podAntiAffinity


joebowbeer:
**Tell us about your request**

Please implement `podAntiAffinity`.

I don't see anything about this on the roadmap. Is this planned?

(In fact, I see almost nothing on the [roadmap](https://github.com/aws/karpenter/projects/3) currently. Issue?)

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**

I want to deploy [k8ssandra](https://k8ssandra.io/) but cannot because it is using pod anti-affinity to prevent multiple pods from being scheduled on the same worker

> Cass Operator uses pod anti-affinity by default to ensure that Cassandra pods are isolated from one another.

and Karpenter refuses to schedule.

I suspect there are a lot of high-availability operators that will currently _prefer_ or _require_ podAntiAffinity.

Prometheus: https://github.com/prometheus-operator/kube-prometheus/issues/1537

**Are you currently working around this issue?**

`allowMultipleNodesPerWorker: true`

From: https://docs.k8ssandra.io/tasks/scale/

**Additional context**

What should these operators be using instead?

I see in the FAQ that this is not a priority for Karpenter but I think podAntiAffinity is fairly common among some popular operators, and it is not clear what they should be doing instead.

Some discussion here: https://github.com/aws/karpenter/pull/695#discussion_r715817587

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

akestner:
Apologies for the lack of detail on the roadmap, we're working to get that updated to reflect the next steps for Karpenter after the launch at re:Invent last week. 

Regarding `podAntiAffinity`, this is something we've discussed in the past, but it presents a number of challenges to Karpenter's scheduling capabilities. Often, as seems to be the case with k8ssandra and the other projects mentioned, topology spread could be a suitable alternative. I'll let @ellistarn weigh in as well since I know he's thought quite a bit about this too.

joebowbeer:
@akestner Thanks for the background.

I suspect several CNCF projects are configured with _soft_ podAntiAffinity now.

[...]

PR #948, opened 2021-12-08
docs(concepts): Clarify podAffinity/podAntiAffinity note

maxbrunet:
**1. Issue, if available:**

Related to #942

Hello 👋 

This notes has been confusing me since I read  it. At first, I thought `podAntiAffinity` would cause scalability issues for the deployments using them, but reading further around `topologySpreadConstraints` (especially its KEP), it seems to be an issue for scalability of the Kubernetes Scheduler.

Do they plan to deprecate `podAffinity`/`podAntiAffinity`? It has been mentioned, but no concrete action toward it.
Or do they plan to improve them? They seem to be some activity in this direction (e.g. kubernetes/enhancements#2249), thus should Karpenter plan to actually support them? (And update the doc to something like "Karpenter doesn't support them for the moment (You can follow their consideration by subscribing to #942).")

TL;DR For this recommendation, imho the doc should have:

- a short reason
- its source (e.g. official Kubernetes documentation, KEP)

I may still be wrong about the why, but I hope to use this PR to get clarification for myself  and everyone else.

- [ ] No

### DIFF
--- a/website/content/en/docs/concepts/_index.md
+++ b/website/content/en/docs/concepts/_index.md
@@ -157,7 +157,7 @@ Those that are implemented in Karpenter include:
 
 {{% alert title="Note" color="primary" %}}
 Don't use `podAffinity` and `podAntiAffinity` to schedule pods on the same or different nodes as other pods.
-Kubernetes SIG scalability recommends against these features due to their negative performance impact on the Kubernetes Scheduler (see [Motivation](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/895-pod-topology-spread#motivation)) and Karpenter doesn't support them.
+Kubernetes SIG scalability recommends against these features due to their negative performance impact on the Kubernetes Scheduler (see [KEP 895](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/895-pod-topology-spread#impact-to-other-features)) and Karpenter doesn't support them.
 Instead, the Karpenter project recommends `topologySpreadConstraints` to reduce blast radius and `nodeSelectors` and `taints` to implement colocation.
 {{% /alert %}}
 
--- a/website/content/en/docs/concepts/_index.md
+++ b/website/content/en/docs/concepts/_index.md
@@ -157,7 +157,7 @@ Those that are implemented in Karpenter include:
 
 {{% alert title="Note" color="primary" %}}
 Don't use `podAffinity` and `podAntiAffinity` to schedule pods on the same or different nodes as other pods.
-Kubernetes SIG scalability recommends against these features and Karpenter doesn't support them.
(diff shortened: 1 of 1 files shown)


ellistarn:
This looks like a bad link: https://github.com/kubernetes/enhancements/pull/249. Can you update?

maxbrunet:
Oops I meant kubernetes/enhancements#2249

ellistarn:
We didn't initially implement affinity due to:
- scalability recommendations
- topology as an alternative
- complexity/effort of implementation

I'm open minded to implementing pod antiaffinity (or at least a subset of the spec). However, pod affinity seems to be a bit of a nonstarter. When provisioning capacity, it's impossible to know whether or not all the pods that need to coschedule have been created yet. It's extremely easy to get wedged this way. This is more viable in static clusters, but still relatively challenging.

I think your current docs change is a good recommendation, and we can continue the discussion in https://github.com/aws/karpenter/issues/942. tl;dr, given the number of charts that use antiaffinity, I think we may be causing undue pain by not supporting it. It's probably worth the investment, at least insofar as antiaffinity is translatable to topology.

maxbrunet:
Thank you @ellistarn for the reply, I have updated the KEP link to point to the section describing the problems, please do not hesitate to suggest a better reference if you have one.

website/content/en/docs/concepts/_index.md
```diff
@@ -157,7 +157,7 @@ Those that are implemented in Karpenter include:
 
 {{% alert title="Note" color="primary" %}}
 Don't use `podAffinity` and `podAntiAffinity` to schedule pods on the same or different nodes as other pods.
-Kubernetes SIG scalability recommends against these features and Karpenter doesn't support them.
+Kubernetes SIG scalability recommends against these features due to their negative performance impact on the Kubernetes Scheduler (see [KEP 895](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/895-pod-topology-spread#impact-to-other-features)) and Karpenter doesn't support them.
```
maxbrunet, line 160: Should we add a link to the issue?

```suggestion
Kubernetes SIG scalability recommends against these features due to their negative performance impact on the Kubernetes Scheduler (see [KEP 895](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/895-pod-topology-spread#impact-to-other-features)) and Karpenter doesn't support them for the moment (you can follow their consideration by subscribing to the [issue](https://github.com/aws/karpenter/issues/942)).".
```

ellistarn (approved):
Thanks!

maxbrunet:
I have finally found the literal recommendation

> **Note:** Inter-pod affinity and anti-affinity require substantial amount of processing which can slow down scheduling in large clusters significantly. We do not recommend using them in clusters larger than several hundred nodes. 

https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity

PR #957, opened 2021-12-09
add cloudprovider metrics

cjerad:
**1. Issue, if available:**

Add a decorator for CloudProvider objects that publishes latency metrics.

- [x] No


pkg/cloudprovider/metrics/cloudprovider.go
```diff
@@ -0,0 +1,161 @@
...
+	v1 "k8s.io/api/core/v1"
+	"knative.dev/pkg/apis"
+	"knative.dev/pkg/logging"
+	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const (
+	metricLabelComponent = "component"
+	metricLabelMethod    = "method"
+	metricLabelProvider  = "provider"
```
felix-zhe-huang, line 34: When do we use component name and when do we use provider name?

cjerad: *Component* specifies which part of karpenter initiated the call to the cloud provider, e.g. metrics controller or termination controller.

*Provider* indicates the cloud service implementation, e.g. "aws".

felix-zhe-huang:
I like the new metrics shim layer design. It will make coding for cloud provider easier since now metrics measurement is automatically handled for the interface functions. At the same time, I am a bit worried that it will muddy the cloud provider abstraction. Now the cloud provider behavior is actually defined by the cloud provider metrics package instead of the interface. Since this is not explicit (the WithComponentName function hides the fact that a decorator is returned instead of a cloud provider), it may confuse people. 

The only real alternative to this design is to let each cloud provider implements their own way to emit the prometheus metrics. A huge advantage for this alternative design is that each cloud provider can decide how the metrics are measured and in what granularity. For example, in AWS create I can totally image breaking down Create function runtime to the time that local alternative instancetype (instancetype expansion) and instance launch time. This may not be sustainable for a large number of cloud providers. But given that we only have one real cloud provider now and in the future we may not have too many, i think this is a problem that we can deal with later.

In my opinion my vote will be to the alternative solution for now, but I would like to hear others' opinions.

cjerad:
>  Now the cloud provider behavior is actually defined by the cloud provider metrics package and the cloud provider.

Not so, the`CloudProvider`  interface defines the methods and signatures that must be implemented by `decorator` in order to be returned as a `CloudProvider`.  The Go compiler will enforce that `decorator`'s method signatures satisfy the `CloudProvider` interface.

> Since this is not explicit (the WithComponentName function hides the fact that a decorator is returned instead of a cloud provider)

This is intentional.  `CloudProvider` is an interface so any implementation that satisfies the interface may be returned, and the caller may use the returned `CloudProvider` without having to be concerned about the underlying implementation.

> The only real alternative to this design is to let each cloud provider implements their own way to emit the prometheus metrics.

The decorator approach does not prevent an underlying implementation from recording its own metrics.  The `decorator` simply records latency metrics which are presumed to be common across underlying implementations.

felix-zhe-huang:
> The Go compiler will enforce that `decorator`'s method signatures satisfy the `CloudProvider` interface.

Make sense. So metrics is really implementing cloud provider. 

> The decorator approach does not prevent an underlying implementation from recording its own metrics.

Agree. There is no issue here.

What do you think about calling WithComponentName in the NewCloudProvider function of each cloud provider instead?
In that way we don't need to change the other controllers' codes.  I am trying to think of a way to hide the metrics layer from outside of cloud provider package.  Maybe put the component and provider name into context?

cjerad:
> What do you think about calling WithComponentName in the NewCloudProvider function of each cloud provider instead?

Meaning [here](https://github.com/aws/karpenter/blob/b94747aed409412531ff610cb8346892486ca70f/pkg/cloudprovider/registry/aws.go#L28) and [here](https://github.com/aws/karpenter/blob/b94747aed409412531ff610cb8346892486ca70f/pkg/cloudprovider/registry/fake.go#L28)?  How would the code know the correct value for the component name?  The `NewCloudProvider` function is only called once during karpenter startup (see [here](https://github.com/aws/karpenter/blob/b94747aed409412531ff610cb8346892486ca70f/cmd/controller/main.go#L76)) then that single `CloudProvider` instance is passed to each controller during initialization.  Some controllers also pass along the `CloudProvider` instance when initializing helper objects.

>  I am trying to think of a way to hide the metrics layer from outside of cloud provider package.

It's possible but would likely require extensive refactoring of the controllers. Up until now, we have avoided jumping through too many hoops to separate metrics code from the code being measured.

> Maybe put the component and provider name into context?

Each controller's `Reconcile` method is called directly by controller-runtime. Do you know if there is a means to intercept the call so that we could inject the component and provider name into the `Context` object?

felix-zhe-huang:
> Meaning [here](https://github.com/aws/karpenter/blob/b94747aed409412531ff610cb8346892486ca70f/pkg/cloudprovider/registry/aws.go#L28) and [here](https://github.com/aws/karpenter/blob/b94747aed409412531ff610cb8346892486ca70f/pkg/cloudprovider/registry/fake.go#L28)? How would the code know the correct value for the component name? The `NewCloudProvider` function is only called once during karpenter startup (see [here](https://github.com/aws/karpenter/blob/b94747aed409412531ff610cb8346892486ca70f/cmd/controller/main.go#L76)) then that single `CloudProvider` instance is passed to each controller during initialization. Some controllers also pass along the `CloudProvider` instance when initializing helper objects.

This is where I think context can be helpful. So in each reconcile call, the first step we do is to inject the component and provider name into the context (e.g., [here](https://github.com/aws/karpenter/blob/b405f79d397c60e0cde0ffc80a491293ac8dc68a/pkg/controllers/provisioning/controller.go#L66), we need to create new injection functions for component name and provider name [here](https://github.com/aws/karpenter/blob/main/pkg/utils/injection/injection.go)) Then the context will be passed down through the call path. In the observe function, instead of using d.componentName, we get it from context (e.g., [here](https://github.com/aws/karpenter/blob/b405f79d397c60e0cde0ffc80a491293ac8dc68a/pkg/controllers/provisioning/scheduling/scheduler.go#L68)). This way, the same cloud provider instance will have different component name and provider name based on where the interface functions are called. We can also remove WithComponentName and WithProviderName and expose NewDecorator. This way we can call NewDecorator(AWS) in AWS's NewCloudProvider function.

felix-zhe-huang:
Another thought is that if we don't want to create the decorator in the cloudprovider's NewCloudProvider function, we can also do it in main.go. Something like
`// Set up controller runtime controller
	cloudProvider := metrics.DecorateCloudProvider(registry.NewCloudProvider(ctx, cloudprovider.Options{ClientSet: clientSet}))`
We still need context injection in this case.

felix-zhe-huang:
Somehow the CI got stuck?

felix-zhe-huang (dismissed):
Looks good to me. :shipit:

cmd/controller/main.go
```diff
@@ -75,6 +76,7 @@ func main() {
 
 	// Set up controller runtime controller
 	cloudProvider := registry.NewCloudProvider(ctx, cloudprovider.Options{ClientSet: clientSet})
+	cloudProvider = cpmetrics.PublishLatency(cloudProvider)
```
ellistarn, line 79: Semantically, I think this is decorating a cloud provider with cloud provider metrics. 
```suggestion
	cloudProvider = cpmetrics.Decorate(cloudProvider)
```

pkg/cloudprovider/metrics/cloudprovider.go
```diff
@@ -0,0 +1,143 @@
...
+type decorator struct {
+	cloudprovider.CloudProvider
+}
+
+// PublishLatency returns a new `CloudProvider` instance that will delegate all method
+// calls to the argument, `cloudProvider`, and publish aggregated latency metrics. The
+// value used for the metric label, "component", is taken from the `Context` object
+// passed to the methods of `CloudProvider`.
+func PublishLatency(cloudProvider cloudprovider.CloudProvider) cloudprovider.CloudProvider {
+	switch c := cloudProvider.(type) {
```
ellistarn, line 67: Why bother checking whether or not you're passing in a decorator? Easier to just do the default in all cases and then never call this twice in a row.

pkg/cloudprovider/metrics/cloudprovider.go
```diff
@@ -0,0 +1,143 @@
...
+		Namespace: metrics.Namespace,
+		Subsystem: "cloudprovider",
+		Name:      "duration_seconds",
+		Help:      "Duration of cloud provider method calls.",
+	},
+	[]string{
+		metricLabelComponent,
+		metricLabelMethod,
+		metricLabelProvider,
+		metricLabelResult,
```
ellistarn, line 50: I'm not sure we want to track the result of these calls. I'd delegate error tracking elsewhere. This will allow you to simplify and use `metrics.Measure()`

pkg/cloudprovider/metrics/cloudprovider.go
```diff
@@ -0,0 +1,143 @@
...
+			d.observe(ctx, "Create", time.Since(startTime), err)
+			out <- err
+		case <-ctx.Done():
+		}
+		close(out)
+	}(time.Now(), d.CloudProvider.Create(ctx, constraints, instanceTypes, quantity, callback))
+	return out
+}
+
+func (d *decorator) Delete(ctx context.Context, node *v1.Node) error {
```
ellistarn, line 89: I'd love if all of these methods were as simple as 
```
defer metrics.Measure(...)()
return d.CloudProvider.Delete(ctx, node)
```

pkg/utils/injection/injection.go
```diff
@@ -63,3 +63,19 @@ func GetConfig(ctx context.Context) *rest.Config {
 	}
 	return retval.(*rest.Config)
 }
+
+type componentNameKeyType struct{}
+
+var componentNameKey = componentNameKeyType{}
+
+func WithComponentName(ctx context.Context, name string) context.Context {
```
ellistarn, line 71: Thoughts on 
```suggestion
func WithControllerName(ctx context.Context, name string) context.Context {
```

cmd/controller/main.go
cjerad: `Decorate` is very generic.  For something more descriptive, how about `DecorateWithLatency`?

pkg/cloudprovider/metrics/cloudprovider.go
cjerad: Double-decoration would be a mistake (each method call would be counted twice) but would not produce a compile-time or runtime error.  Karpenter would simply publish incorrect metric values.

Detecting and fixing this situation is easy, so it was added to the code.  The alternative would be to return an `error`.

pkg/utils/injection/injection.go
cjerad: The term "controller" was too course, e.g. the *provisioning* controller creates `Provisioner` objects which also create a `Packer` object, all use the `CloudProvider` instance.  Applying the label "controller" to calls attributed to `Provisioner` and `Packer` objects would be misleading, but it did  still seem preferable to attribute the calls to the responsible objects rather than rolling them all up to the *provisioning* controller.  So the more general term "component" was used.

pkg/cloudprovider/metrics/cloudprovider.go
ellistarn: There are an infinite number of programming mistakes that could be made that would result in undesirable behavior. Attempting to protect against them with more code increases the surface area of the project and needs to be understood, tested. 

This code base follows a pattern where our external interfaces (e.g. provisioner) are validated and protected, but for internal interfaces, we assume that developers are doing the right thing. This allows the code to accurately reflect its purpose, without the additional complexity of whether or not a developer decided to protect against a specific failure mode.

cmd/controller/main.go
ellistarn: This is of course, not blocking, but...

When I think about semantically what this, I think of a "Cloud Provider Metrics Decorator". Following the standard go package name resolution, this becomes cloudprovidermetrics.Decorate(cloudprovider), which returns a "Cloud Provider Metrics Decorator". 

I'm not sure there's anything fundamentally related to Latency. e.g. we could start recording counts, and I'd hope we wouldn't need to change the name of this method.

pkg/utils/injection/injection.go
ellistarn: I think we *do* want to associate the calls with the controller, not the component. The individual implementations of packer/scheduler, etc, are non-fundamental to the code base. It's simply how the code happens to have been implemented and will (and has) change frequently. The controllers however, are fundamental in that they are essential to how to the system is architected. 

If we did simply want to know the caller, we could use `runtime.Caller(0)`. Of course, I think we should be controller-aware, instead.

pkg/cloudprovider/metrics/cloudprovider.go
```diff
@@ -0,0 +1,109 @@
...
+// value used for the metric label, "controller", is taken from the `Context` object
+// passed to the methods of `CloudProvider`.
+//
+// Do not decorate a `CloudProvider` multiple times or published metrics will contain
+// duplicated method call counts and latencies.
+func Decorate(cloudProvider cloudprovider.CloudProvider) cloudprovider.CloudProvider {
+	return &decorator{cloudProvider}
+}
+
+func (d *decorator) Create(ctx context.Context, constraints *v1alpha5.Constraints, instanceTypes []cloudprovider.InstanceType, quantity int, callback func(*v1.Node) error) <-chan error {
```
ellistarn, line 69: Dang this is fancy :D

ellistarn (approved):
Thank you!

PR #962, opened 2021-12-09
perform min instance bin-packing to short circuit unneeded packings

bwagner5:
**1. Issue, if available:**
N/A

 - This PR changes the bin-packing logic to find the maximum packing possible and then iterate through instance types until it finds a smaller instance type that supports the maximum packing. This algorithm simplifies the packing logic a bit as well.
 - This PR also adds GPU resource validation to the packables instance type filtering logic. Previously, GPU instance types were only excluded when the pods did not request a GPU. If a pod did request a GPU, the instance types without GPUs were excluded at packing time which increases the packing iterations substantially.  
 - A general performance improvement was made by moving the `PackablesFor(..` outside of the `remainingPods` loop and instead DeepCopy the packables on each iteration to short circuit validation that has already occurred.

## Some Numbers:

Worst case (meaning min-packing didn't optimize anything):
 - Before this change a 5,000 pod packing took 1,260ms
 - After this change a 5,000 pod packing took 426ms

Best Case (min-packing engaged on smallest instance-type):
 - Before this change a 5,000 pod packing took 1,658ms
 - After this change a 5,000 pod packing took 99ms

- [x] No


pkg/cloudprovider/fake/instancetype.go
```diff
@@ -23,106 +25,124 @@ import (
...
+	Pods             resource.Quantity
+	NvidiaGPUs       resource.Quantity
+	AMDGPUs          resource.Quantity
+	AWSNeurons       resource.Quantity
+	AWSPodENI        resource.Quantity
 }
 
 type InstanceType struct {
-	InstanceTypeOptions
+	options InstanceTypeOptions
```
ellistarn, line 101: Too fancy?

pkg/controllers/provisioning/binpacking/packable.go
```diff
@@ -39,10 +41,36 @@ type Result struct {
 	unpacked []*v1.Pod
 }
 
+type Packables []*Packable
```
ellistarn, line 44: I find this sort implementation pattern to be vastly inferior to https://pkg.go.dev/sort#Slice. You can separate out the comparator as necessary.

pkg/controllers/provisioning/binpacking/packable.go
```diff
@@ -56,13 +84,11 @@ func PackablesFor(ctx context.Context, instanceTypes []cloudprovider.InstanceTyp
...
 			packable.validateCapacityTypes(constraints),
 			packable.validateAWSPodENI(pods),
-			// Although this will remove instances that have GPUs when
-			// not required, removal of instance types that *lack*
-			// GPUs will be done later.
 			packable.validateNvidiaGpus(pods),
 			packable.validateAMDGpus(pods),
 			packable.validateAWSNeurons(pods),
 		); err != nil {
+			logging.FromContext(ctx).Debugf("Excluding instance type %s because %v", packable.Name(), err.Error())
```
ellistarn, line 91: The standard pattern in the codebase tends to be:
```suggestion
			logging.FromContext(ctx).Debugf("Excluding instance type %s, %v", packable.Name(), err.Error())
```

pkg/controllers/provisioning/binpacking/packable.go
```diff
@@ -198,45 +234,44 @@ func (p *Packable) validateCapacityTypes(constraints *v1alpha5.Constraints) erro
...
-	if p.InstanceType.NvidiaGPUs().IsZero() {
-		return nil
-	}
-	for _, pod := range pods {
-		for _, container := range pod.Spec.Containers {
-			if _, ok := container.Resources.Requests[resources.NvidiaGPU]; ok {
-				return nil
-			}
-		}
+	needsGpu := p.needsResource(pods, resources.NvidiaGPU)
```
ellistarn, line 237: Can you collapse this into 

```
if p.requiresResource(pods, resources.NvidiaGPU) == p.InstanceType.NvidiaGPUs().IsZero() {
  return fmt.Errorf("%s is only used if required", resources.NvidiaGPU)
}
```

pkg/controllers/provisioning/binpacking/packable.go
```diff
@@ -198,45 +234,44 @@ func (p *Packable) validateCapacityTypes(constraints *v1alpha5.Constraints) erro
...
-		return nil
-	}
-	for _, pod := range pods {
-		for _, container := range pod.Spec.Containers {
-			if _, ok := container.Resources.Requests[resources.AMDGPU]; ok {
-				return nil
-			}
-		}
+	needsGpu := p.needsResource(pods, resources.AMDGPU)
+	if needsGpu && p.InstanceType.AMDGPUs().IsZero() {
```
ellistarn, line 248: I can almost taste the upcoming abstraction ;)

pkg/controllers/provisioning/binpacking/packer.go
```diff
@@ -150,83 +156,30 @@ func (p *Packer) packWithLargestPod(unpackedPods []*v1.Pod, packables []*Packabl
...
-		// considered as a backup option in case we get ICE
-		if p.podsMatch(bestPackedPods, result.packed) {
-			bestInstances = append(bestInstances, packable.InstanceType)
-		} else if len(result.packed) > len(bestPackedPods) {
-			// If pods packed are more than compared to what we got in last
-			// iteration, consider using this instance type
+		if result := packable.Pack(unpackedPods); len(result.packed) == maxPodsPacked {
+			// Add all packable nodes that have more resources than this one
+			// Trim the bestInstances so that provisioning APIs in cloud providers are not overwhelmed by the number of instance type options
+			// For example, the AWS EC2 Fleet API only allows the request to be 145kb which equates to about 130 instance type options.
```
ellistarn, line 171:
```suggestion
```

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -51,11 +52,18 @@ type Scheduler struct {
 }
 
 type Schedule struct {
-	*v1alpha5.Constraints
+	*RuntimeConstraints
 	// Pods is a set of pods that may schedule to the node; used for binpacking.
 	Pods []*v1.Pod
 }
 
+type RuntimeConstraints struct {
```
ellistarn, line 60: Thoughts on calling this object scheduling.Constraints for parity with v1alpha5.Constraints. I don't think that "Runtime" differentiates more meaningfully than the local context.

ellistarn: Also -- I'm not understanding why this object is useful.

pkg/test/pods.go
```diff
@@ -96,6 +96,19 @@ func Pod(overrides ...PodOptions) *v1.Pod {
 	}
 }
 
+// Pods creates homogeneous groups of pods based on the passed in options, evenly divided by the total pods requested
+func Pods(total int, options ...PodOptions) []*v1.Pod {
+	pods := []*v1.Pod{}
+	count := 0
+	for _, opts := range options {
+		for i := 0; i < total/len(options); i++ {
+			opts.Name = fmt.Sprintf("fake-test-pod-%d", count)
```
ellistarn, line 105: Were you getting conflicts with randomdata.Sillyname? Consider using that as the prefix if you need to suffix -%d

pkg/utils/pod/sortable.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"github.com/aws/karpenter/pkg/utils/resources"
+	v1 "k8s.io/api/core/v1"
+)
+
+type SortablePods []*v1.Pod
+type ByResourcesRequested struct{ SortablePods }
+
+func (pods SortablePods) Len() int {
```
ellistarn, line 25: Thoughts on switching to the sort.Slice() approach mentioned above?

pkg/controllers/provisioning/binpacking/packable.go
```diff
@@ -39,10 +41,36 @@ type Result struct {
...
+		packables[a].AWSNeurons().Equal(*packables[b].AWSNeurons()) {
+		if packables[a].CPU().Equal(*packables[b].CPU()) {
+			// check for memory
+			return packables[a].Memory().Cmp(*packables[b].Memory()) == -1
+		}
+		return packables[a].CPU().Cmp(*packables[b].CPU()) == -1
+	}
+	return packables[a].AMDGPUs().Cmp(*packables[b].AMDGPUs()) == -1 ||
+		packables[a].NvidiaGPUs().Cmp(*packables[b].NvidiaGPUs()) == -1 ||
+		packables[a].AWSNeurons().Cmp(*packables[b].AWSNeurons()) == -1
```
njtran, line 67: Do the number of AWSNeurons, AMDGPUs, and NvidiaGPUs on each instance increase with instance size at the same rate? If not, wouldn't it be possible for this to have non-deterministic behavior?

pkg/controllers/provisioning/binpacking/packable.go
njtran: Maybe this error message: 
```
return errors.New("pod requirements do not match Nvidia GPUs")
```
"not required" could be understood as "is allowed" which is confusing to get in an error message.

pkg/controllers/provisioning/binpacking/packable.go
```diff
@@ -198,45 +234,44 @@ func (p *Packable) validateCapacityTypes(constraints *v1alpha5.Constraints) erro
...
 }
 
 func (p *Packable) validateAWSNeurons(pods []*v1.Pod) error {
-	if p.InstanceType.AWSNeurons().IsZero() {
-		return nil
+	needsGpu := p.needsResource(pods, resources.AWSNeuron)
+	if needsGpu && p.InstanceType.AWSNeurons().IsZero() {
+		return errors.New("aws neuron is required")
+	} else if !needsGpu && !p.InstanceType.AWSNeurons().IsZero() {
+		return errors.New("aws neuron is not required")
```
njtran, line 261: "not required" could be understood as "is allowed"

pkg/cloudprovider/fake/instancetype.go
bwagner5: This was necessary because the funcs defined in the Interface now match the struct fields of InstanceTypeOptions, since they are now exported.

pkg/controllers/provisioning/binpacking/packable.go
bwagner5: ah you're right, I'll refactor those sorts.

pkg/controllers/provisioning/binpacking/packable.go
bwagner5: They increase, not necessarily at the same rate by gpu but by vcpu and memory.  I'm not sure how that would introduce non-deterministic behavior though? 

For example:

```
$ ec2-instance-selector  -o table-wide --allow-list g4ad
Instance Type  VCPUs   Mem (GiB)  Hypervisor  Current Gen  Hibernation Support  CPU Arch  Network Performance  ENIs    GPUs    GPU Mem (GiB)  GPU Info             On-Demand Price/Hr
-------------  -----   ---------  ----------  -----------  -------------------  --------  -------------------  ----    ----    -------------  --------             ------------------
g4ad.xlarge    4       16         nitro       true         false                x86_64    Up to 10 Gigabit     2       1       8              AMD Radeon Pro V520  -No Price Filter Specified-
g4ad.2xlarge   8       32         nitro       true         false                x86_64    Up to 10 Gigabit     2       1       8              AMD Radeon Pro V520  -No Price Filter Specified-
g4ad.4xlarge   16      64         nitro       true         false                x86_64    Up to 10 Gigabit     3       1       8              AMD Radeon Pro V520  -No Price Filter Specified-
g4ad.8xlarge   32      128        nitro       true         false                x86_64    15 Gigabit           4       2       16             AMD Radeon Pro V520  -No Price Filter Specified-
g4ad.16xlarge  64      256        nitro       true         false                x86_64    25 Gigabit           8       4       32             AMD Radeon Pro V520  -No Price Filter Specified-
```

pkg/controllers/provisioning/binpacking/packable.go
ellistarn: To be clear my suggestion was to not explicitly reference anything in the string and instead use the resource type. This will allow us to abstract :)

pkg/cloudprovider/fake/instancetype.go
ellistarn: Hmm -- I actually kind of did this intentionally to encourage not making them public. Why did you need to make them public?

pkg/controllers/provisioning/binpacking/packable.go
ellistarn: Did you forget to push?

pkg/cloudprovider/fake/instancetype.go
bwagner5: Making them public allows for more isolated tests with cloud provider specific data. For example, in my testing of the binpacking pkg, I loaded all the real AWS InstanceTypes into fake instance types and ran that thru the packing logic for lightweight,  real-world testing.  

I'm more in favor of keeping them public since it makes loading arbitrary instance types easier.

pkg/controllers/provisioning/binpacking/packable.go
bwagner5: no? I might have done an intermediate push when you looked, should be okay now.

ellistarn (approved):

PR #968, opened 2021-12-10
Adds pod and node metrics controllers

felix-zhe-huang:
**1. Issue, if available:**
N/A

Add a generic pod metrics controller that provide pod status metrics to Prometheus.
This will allow the dashboard to perform metrics aggregation at Promeheus/Grafana level.

- [X] No

This is a sample dashboard created using the pod metric controller.
Screen Shot 2021-12-10 at 5 57 50 PM

This is a sample dashboard created using the node metric controller.
Screen Shot 2021-12-16 at 5 21 16 PM


pkg/controllers/metrics/podmetrics/podmetrics.go
```diff
@@ -0,0 +1,185 @@
...
+	podname             = "name"
+	podnamespace        = "namespace"
+	ownerselflink       = "owner"
+	podhostname         = "node"
+	podprovisioner      = "provisioner"
+	podhostzone         = "zone"
+	podhostarchitecture = "arch"
+	podhostcapacitype   = "capacitytype"
+	podhostinstancetype = "instancetype"
+	podphase            = "phase"
```
njtran, line 48: Can we make these CamelCaseVariableNames and more descriptive of how they're used?
e.g.
`PodNameLabelKey`
`PodNamespaceLabelKey`

cmd/controller/main.go
```diff
@@ -23,7 +23,7 @@ import (
 	"github.com/aws/karpenter/pkg/cloudprovider/registry"
 	"github.com/aws/karpenter/pkg/controllers"
 	"github.com/aws/karpenter/pkg/controllers/counter"
-	"github.com/aws/karpenter/pkg/controllers/metrics"
+	"github.com/aws/karpenter/pkg/controllers/metrics/podmetrics"
```
ellistarn, line 26: Don't let the name collision force your package naming elsewhere. Consider something like 
```suggestion
	metricspod "github.com/aws/karpenter/pkg/controllers/metrics/pod"
```

pkg/controllers/metrics/podmetrics/podmetrics.go
felix-zhe-huang: good point. should have caught this before.

felix-zhe-huang: done

cmd/controller/main.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,338 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	nodeName         = "name"
+	nodeProvisioner  = "provisioner"
+	nodeZone         = "zone"
+	nodeArchitecture = "arch"
+	nodeCapacityType = "capacitytype"
```
ellistarn, line 47: prometheus naming best practice is snake_case https://prometheus.io/docs/practices/naming/

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+			Help:      "Pod state.",
+		},
+		getLabelNames(),
+	)
+)
+
+// Controller for the resource
+type Controller struct {
+	KubeClient   client.Client
+	CoreV1Client corev1.CoreV1Interface
```
ellistarn, line 68: Looks like you're using the kubeclient instead. Typically we only use the corev1client for resource specific verbs like pods/evict pods/bind

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+	podPhase            = "phase"
+	podLabels           = "podLabels"
+)
+
+var (
+	podGaugeVec = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "karpenter",
+			Subsystem: "pods",
+			Name:      "podstate",
```
ellistarn, line 58: pod_state. Maybe simply to just name this pod.

ellistarn: Does this stutter with pods? pods_podstate? Maybe we can do karpenter_pod_state_X

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// Retrieve pod from reconcile request
+	pod := &v1.Pod{}
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, pod); err != nil {
+		if errors.IsNotFound(err) {
+			// Remove gauge due to pod deletion
+			if labels, ok := c.LabelsMap[req.NamespacedName]; ok {
+				podGaugeVec.Delete(*labels)
+			} else {
+				logging.FromContext(ctx).Errorf("Failed to delete gauge: failed to locate labels")
```
ellistarn, line 115: I'm not sure it's useful to log here. It's possible that we're simply not logging labels. I wouldn't want to concern a customer w/ an error.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+	// Remove the previous gauge after pod labels are updated
+	if labels, ok := c.LabelsMap[req.NamespacedName]; ok {
```
ellistarn, line 122: I'm not grokking this. Can't we just update it?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
+	err := controllerruntime.
+		NewControllerManagedBy(m).
+		Named("podmetrics").
+		For(&v1.Pod{}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10000}).
```
ellistarn, line 148: I might just leave this as the unspecified default (1). It should be wicked fast to execute, but I don't want to spend too many cycles computing this.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// generateLabels creates the labels using the current state of the pod
+func (c *Controller) generateLabels(ctx context.Context, pod *v1.Pod) (*prometheus.Labels, error) {
+	metricLabels := prometheus.Labels{}
+	metricLabels[podName] = pod.GetName()
+	metricLabels[podNameSpace] = pod.GetNamespace()
+	// Selflink has been deprecated after v.1.20
+	// Manually generate the selflink for the first owner reference
+	// Currently we do not support multiple over references
```
ellistarn, line 160:
```suggestion
	// Currently we do not support multiple owner references
```

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+	selflink := ""
+	if len(pod.GetOwnerReferences()) != 0 {
```
ellistarn, line 162: more idiomatic as:
```suggestion
	if len(pod.GetOwnerReferences()) > 0 {
```

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+		ownerreference := pod.GetOwnerReferences()[0]
+		selflink = fmt.Sprintf("/apis/%s/namespaces/%s/%ss/%s", ownerreference.APIVersion, pod.Namespace, strings.ToLower(ownerreference.Kind), ownerreference.Name)
+	}
+	metricLabels[ownerSelfLink] = selflink
```
ellistarn, line 166: Why not include this in the if statement? Do we want it to be ""?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+	metricLabels[podHostName] = pod.Spec.NodeName
+	metricLabels[podPhase] = string(pod.Status.Phase)
+	provisioner := v1alpha5.DefaultProvisioner
```
ellistarn, line 169: I don't think this is right.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+	if name, ok := pod.Spec.NodeSelector[v1alpha5.ProvisionerNameLabelKey]; ok {
+		provisioner.Name = name
+	}
+	metricLabels[podProvisioner] = provisioner.Name
+	nodename := types.NamespacedName{Name: pod.Spec.NodeName}
```
ellistarn, line 174: Consider inlining, since it's only used once.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+	node := &v1.Node{}
+	if err := c.KubeClient.Get(ctx, nodename, node); err != nil {
+		metricLabels[podHostZone] = "N/A"
```
ellistarn, line 177: Do we need to include this as "N/A", or can it simply not exist or be ""

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		logging.FromContext(ctx).Errorf("Failed to generate new labels: %s", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	gauge, err := podGaugeVec.GetMetricWith(*newlabels)
+	if err != nil {
+		logging.FromContext(ctx).Errorf("Failed to generate new gauge: %s", err.Error())
+		return reconcile.Result{}, err
+	}
```
ellistarn, line 135: Consider wrapping this in another function to avoid repeating yourself. E.g.

```
if err := c.Record(pod); err != nil {
  logging.FromContext(ctx).Errorf("Failed to record metrics, %s", err.Error())
}
```
You may also want to use the noop logger to avoid controller runtime logging this twice.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+
+}
+
+// NewController constructs a controller instance
+func NewController(kubeClient client.Client, coreV1Client corev1.CoreV1Interface) *Controller {
+	newcontroller := Controller{
+		KubeClient:   kubeClient,
+		CoreV1Client: coreV1Client,
+		LabelsMap:    make(map[types.NamespacedName]*prometheus.Labels),
```
ellistarn, line 98: Is prometheus.labels a slice or map under the hood? If so, it's already a pointer so we don't need to point to that.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	return &newcontroller
```
ellistarn, line 100: consider returning inline 
`return &Controller{...}`

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,194 @@
...
+		metricLabels[podHostCapacityType] = "N/A"
+		metricLabels[podHostInstanceType] = "N/A"
+	} else {
+		metricLabels[podHostZone] = node.Labels[v1.LabelTopologyZone]
+		metricLabels[podHostArchitecture] = node.Labels[v1.LabelArchStable]
+		metricLabels[podHostCapacityType] = node.Labels[v1alpha5.LabelCapacityType]
+		metricLabels[podHostInstanceType] = node.Labels[v1.LabelInstanceTypeStable]
+	}
+	// Add pod labels
+	labels, err := json.Marshal(pod.GetLabels())
```
ellistarn, line 188: Why do we want to jsonify this? Does this always result in the same string (i.e. map keys are not ordered).

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,338 @@
...
+
+func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
+	err := controllerruntime.
+		NewControllerManagedBy(m).
+		Named("nodemetrics").
+		For(&v1.Node{}).
+		Watches(
+			// Reconcile all nodes related to a provisioner when it changes.
+			&source.Kind{Type: &v1alpha5.Provisioner{}},
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) (requests []reconcile.Request) {
```
ellistarn, line 152: We should abstract this somewhere -- it's used multiple times. `MapProvisionerToNode`

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,338 @@
...
+			// Reconcile node when a pod assigned to it changes.
+			&source.Kind{Type: &v1.Pod{}},
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) (requests []reconcile.Request) {
+				if name := o.(*v1.Pod).Spec.NodeName; name != "" {
+					requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Name: name}})
+				}
+				return requests
+			}),
+		).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10000}).
```
ellistarn, line 174: same

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,338 @@
...
+		Complete(c)
+	return err
+}
+
+func (c *Controller) deleteGaguges(labels *prometheus.Labels) {
```
ellistarn, line 179:
```suggestion
func (c *Controller) deleteGauges(labels *prometheus.Labels) {
```

ellistarn: Why does this work differently than the pod controller?

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,338 @@
...
+	for _, gaugeVec := range c.GaugeVecMap {
+		gaugeVec.Delete(*labels)
+	}
+}
+
+// generateLabels creates the labels using the current state of the pod
+func (c *Controller) generateLabels(node *v1.Node) (*prometheus.Labels, error) {
+	metricLabels := prometheus.Labels{}
```
ellistarn, line 187: naming nit: 
we're already contextualized with the "metric" concept by nature of being in the metrics controller. You could just say `labels := prometheus.Labels{}` and it would read just as well.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,338 @@
...
+
+	metricLabels[nodeName] = node.GetName()
+	metricLabels[nodeProvisioner] = node.Labels[v1alpha5.ProvisionerNameLabelKey]
```
ellistarn, line 190: Is it crazy to include *every* label on the node in a `for`?

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,338 @@
...
+	// Add node labels
+	labels, err := json.Marshal(node.GetLabels())
+	if err != nil {
+		return nil, fmt.Errorf("marshal pod labels: %w", err)
+	}
+	metricLabels[nodeLabels] = string(labels)
+	return &metricLabels, nil
+}
+
+func (c *Controller) updateGaguges(ctx context.Context, node *v1.Node, labels *prometheus.Labels) error {
```
ellistarn, line 205:
```suggestion
func (c *Controller) updateGauges(ctx context.Context, node *v1.Node, labels *prometheus.Labels) error {
```

ellistarn, line 205: It's a bit odd to me that you've structured the code differently for these two controllers, yet they do almost exactly the same thing over different resources.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,338 @@
...
+					value.Add(quantity)
+					limits[resourceName] = value
+				}
+			}
+		}
+	}
+	return
+}
+
+func addResourceQuantity(resources v1.ResourceList, resourceMap map[v1.ResourceName]resource.Quantity) {
```
ellistarn, line 328: Do we even need to record this at the node level? If we record it at the pod level, can't we just sum the metric over pods with a specific name?

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,338 @@
+/*
```
ellistarn, line 1: If we just have one file per controller, thoughts on putting these in the same package?

`pkg/controllers/metrics/node.go`  `metrics.NodeController{}`
`pkg/controllers/metrics/pod.go` `metrics.PodController{}`

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: I modified it to be karpenter_pods_state.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Good suggestion. Was confused how the golang prometheus package handles naming. Did some experiments and find out how to use the snake_case here. 

Done!

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: Removed CoreV1Client since it is never used.

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: Agree. what do you think about downgrading this to debug?

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: Not really. Prometheus treats metric with different label values as different active time series. So if we have new label values (representing the state of the pod), it will create a new gauge in gaugeVec. We need to delete the old one so that Prometheus will not create wrong counts.

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: Yeah I have no good understanding how much we can spend on calculating metrics. Ok I will tune it down.

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: Done

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: Yes we want "" for those that doesn't have any ownerreference.

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: Agree. Changing these to "".

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: prometheus.labels is a map.  Good catch.

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: We discussed  and want to also generate metrics according to pod's k8s labels. We talked about adding them to the prometheus labels dynamically. However, after doing some research I am convinced that it is an extrem anti-pattern  to add labels dynamically.

The solution is to put the k8s labels as the value of one prometheus label (i.e., podLabels). This allow us to do metric filtering using the podLabels prometheus label. Therefore, we need to convert he k8s labels into one single string. Json is just one way to do it. we can also just do simple key value pairs. But this works pretty well.

BTW, we don't really care about the order of these k8s lables.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Lets change this in the next PR. I don't want to modify other controllers in this PR.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: I was thinking about distinguish the k8s labels and the prometheus labels. Maybe I should name them prometheusLabels?

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: The labels need to be declared when creating the gaugeVec. It is anti-pattern to change the labels (not the values, but the definition) on the fly. So it is impossible to just add new prometheus labels from node's k8s labels without some serious hacking. As a result, I add all the node k8s labels as the value of a prometheus label called nodeLabel.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

felix-zhe-huang: One of the major differences is that pod controller has only one gaugeVec (the pod state), but node controller needs to generates one gaugeVec for each resource type. Some resource type is pretty standard like CPU, memory but there are cloud specific type like attachable-volumes-aws-ebs. I don't want to hardcode the resource types so we dynamically generate gaugeVec on the flight instead.

Another major difference is that the pod controller doesn't really calculate a metric value (it always return 1 to indicate that there is a pod in that state). But the node controller needs to find out resource levels from nodes.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: In my opinion resource level should be a node level metric instead of a pod level. We can definitely record the pod resource level but that will introduce huge loads to Prometheus for both scraping and when query hits. Aggregating data at the query time is gonna make the dashboard unusable.

The only advantage to record the pod level resource consumption is that we can break the node's workload down and identify which pods are causing problem. But I would rather the customer to figure that out using kubectl.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Will do once we figure out how to integrate these new metric controllers with the existing pkg/controllers/metrics/node.go and pkg/controllers/metrics/pod.go controller.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,338 @@
...
+		nodeCapacityType,
+		nodeInstanceType,
+		nodePhase,
+		nodeLabels,
+	}
+}
+
+type Controller struct {
+	KubeClient   client.Client
+	CoreV1Client corev1client.CoreV1Interface
```
ellistarn, line 68: Can you remove this guy?

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,338 @@
...
+	LabelsMap    map[types.NamespacedName]*prometheus.Labels
+	GaugeVecMap  map[string]*prometheus.GaugeVec
```
ellistarn, line 70: I typically see prometheus metrics stored as static variables, rather than member variables. It makes it easier to access metrics without a reference to an encapsulating context.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
ellistarn: This is fine.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+	labels, err := json.Marshal(node.GetLabels())
+	if err != nil {
+		return nil, fmt.Errorf("marshal pod labels: %w", err)
+	}
+	metricLabels[nodeLabels] = string(labels)
+	return metricLabels, nil
+}
+
+func (c *Controller) updateGauges(ctx context.Context, node *v1.Node) error {
+
```
ellistarn, line 220:
```suggestion
```

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+	daemonSetRequests, _ := getPodsTotalRequestsAndLimits(daemonSetPods)
+	addResourceQuantity(daemonSetRequests, overheads)
+	return overheads
+}
+
+// GetPodsTotalRequestsAndLimits calculates the total resource requests and limits for the pods.
+// If pod overhead is non-nil, the pod overhead is added to the
+// total container resource requests and to the total container limits which have a non-zero quantity.
+func getPodsTotalRequestsAndLimits(pods []v1.Pod) (reqs map[v1.ResourceName]resource.Quantity, limits map[v1.ResourceName]resource.Quantity) {
```
ellistarn, line 300: We have helpers for this: 
https://github.com/aws/karpenter/blob/0113a91d7bb2630aaffef2842ac3dc17b4a55517/pkg/utils/resources/resources.go#L30

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+		c.LabelsMap[nodeNamespacedName] = append(c.LabelsMap[nodeNamespacedName], labels)
+		if err != nil {
+			return fmt.Errorf("generate new labels: %w", err)
+		}
+		gauge, err := gaugeVec.GetMetricWith(labels)
+		if err != nil {
+			return fmt.Errorf("generate new gauge: %w", err)
+		}
+		if resourceName == v1.ResourceCPU {
+			gauge.Set(float64(quantity.MilliValue()))
```
ellistarn, line 275: Reminder from offline discussion, let's just make this `Value()` in all cases.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+	if err := c.insertGaugeValues(allocatable, node, allocatableGaugeVec); err != nil {
+		logging.FromContext(ctx).Debugf("Failed to generate gauge: %w", err)
+	}
+
+	return nil
+}
+
+func (c *Controller) insertGaugeValues(resources map[v1.ResourceName]resource.Quantity, node *v1.Node, gaugeVec *prometheus.GaugeVec) error {
+
+	for resourceName, quantity := range resources {
```
ellistarn, line 261: Minor style nit, but I typically find that functions whose entire body is a for loop should probably instead have the for loop outside. It forces all of the logic to be indented, which hampers readability. It also tends to be simpler to reason about shortcircuiting, etc.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+		allocatable = node.Status.Allocatable
+		// calculating system daemons overhead
+		for resourceName, quantity := range allocatable {
+			overhead := node.Status.Capacity[resourceName]
+			overhead.Sub(quantity)
+			overheads[resourceName] = overhead
+		}
+	}
+
+	// Populate overhead metrics
```
ellistarn, line 239: Minor: consider wrapping all of these in a for, like
```
for vec, resources := range map[*prometheus.GaugeVec]map[v1.ResourceName]resource.Quantity{
   overheadGaugeVec: overhead,
   requestsGaugeVec: requests,
   ...
}{
  c.insert...
}
```

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+// Reconcile executes a termination control loop for the resource
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("nodemetrics").With("node", req.Name))
+	// Retrieve node from reconcile request
+	node := &v1.Node{}
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, node); err != nil {
+		if errors.IsNotFound(err) {
+			// Remove gauge due to node deletion
+			if labels, ok := c.LabelsMap[req.NamespacedName]; ok {
+				c.deleteGauges(labels)
```
ellistarn, line 134: It looks like we're deleting in all cases except wen GetNode returns an error. I'd argue that we might actually want to clean up if we're unable to reconcile for some reason, since we don't know what the data is. It may be simpler to just reset the data as the first step of the reconcile loop.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+	return nil
+}
+
+func calculateDaemonSetOverhead(pods []v1.Pod) v1.ResourceList {
+	overheads := v1.ResourceList{}
+	// calculating daemonset overhead
+	daemonSetPods := []v1.Pod{}
+	for _, pod := range pods {
+		if pod.GetOwnerReferences()[0].Kind == "DaemonSet" {
```
ellistarn, line 288: Check out utils/pod.IsOwnedByDaemonSet

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+		if resourceName == v1.ResourceCPU {
+			gauge.Set(float64(quantity.MilliValue()))
+		} else {
+			gauge.Set(float64(quantity.Value()))
+		}
+	}
+	return nil
+}
+
+func calculateDaemonSetOverhead(pods []v1.Pod) v1.ResourceList {
```
ellistarn, line 283: This misses out on kube reserved 
image

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+func (c *Controller) updateGauges(ctx context.Context, node *v1.Node) error {
+
+	// Calculate total requests and limits from all pods
+	podlist := &v1.PodList{}
+	if err := c.KubeClient.List(ctx, podlist, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
+		return fmt.Errorf("listing pods on node %s, %w", node.Name, err)
+	}
+	reqs, limits := getPodsTotalRequestsAndLimits(podlist.Items)
```
ellistarn, line 226: Nit:
I'd use `requests` to avoid introducing a new noun. I'm all for simplifying/aliasing when things are overly verbose, but I think it muddies things rather than clarifies them.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+type Controller struct {
+	KubeClient client.Client
+	LabelsMap  map[types.NamespacedName][]prometheus.Labels
+}
+
+// NewController constructs a controller instance
+func NewController(kubeClient client.Client) *Controller {
+	return &Controller{
+		KubeClient: kubeClient,
+		LabelsMap:  make(map[types.NamespacedName][]prometheus.Labels),
```
ellistarn, line 121: IIUC, the set of labels is deterministic. Can't we just pass LabelNames() to the delete function and avoid tracking any of this per node.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+)
+
+func init() {
+	crmetrics.Registry.MustRegister(allocatableGaugeVec)
+	crmetrics.Registry.MustRegister(requestsGaugeVec)
+	crmetrics.Registry.MustRegister(limitsGaugeVec)
+	crmetrics.Registry.MustRegister(overheadGaugeVec)
+}
+
+func getLabelNames() []string {
```
ellistarn, line 98: For static accessors, I often avoid the "get" and just name like "labelNames()". To me, get implies that you have to do something (e.g. network call, cache lookup) to get the result. This is very much a go-ism rather than a java-ism, at least as far as I've seen.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+// Reconcile executes a termination control loop for the resource
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("nodemetrics").With("node", req.Name))
+	// Retrieve node from reconcile request
+	node := &v1.Node{}
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, node); err != nil {
+		if errors.IsNotFound(err) {
+			// Remove gauge due to node deletion
```
njtran, line 132: I might not be as onboarded to how GaugeVecs work, but does this mean that we won't retain data on nodes after they're deleted? I think it might be worth distinguishing metrics between long term for user-analysis and short term for capacity optimization. 
 
For example: I can imagine a use-case where someone has a cluster with high churn (e.g. a low value of `ttlSecondsUntilExpired`) on one provisioner, and wants to see how that affects their cost in vs in a provisioner with a higher value of `ttlSecondsUntilExpired`

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+					return requests
+				}
+				for _, node := range nodes.Items {
+					requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Name: node.Name}})
+				}
+				return requests
+			}),
+		).
+		Watches(
+			// Reconcile node when a pod assigned to it changes.
```
njtran, line 177: How about the following? I want to disambiguate how this works a little more in the codebase.
```suggestion
			// Reconcile all nodes where pods have changed
```

pkg/controllers/metrics/node/controller.go
njtran: Does this mean that it isn't possible for users to add their own label filters when looking at dashboard, or are these functions independent of each other?

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+	metricLabels[nodeProvisioner] = node.Labels[v1alpha5.ProvisionerNameLabelKey]
+	metricLabels[nodeZone] = node.Labels[v1.LabelTopologyZone]
+	metricLabels[nodeArchitecture] = node.Labels[v1.LabelArchStable]
+	metricLabels[nodeCapacityType] = node.Labels[v1alpha5.LabelCapacityType]
+	metricLabels[nodeInstanceType] = node.Labels[v1.LabelInstanceTypeStable]
+	metricLabels[nodePhase] = string(node.Status.Phase)
+	// Add node labels
+	labels, err := json.Marshal(node.GetLabels())
+	if err != nil {
+		return nil, fmt.Errorf("marshal pod labels: %w", err)
```
njtran, line 213:
```suggestion
		return nil, fmt.Errorf("marshalling pod metric labels: %w", err)
```

pkg/controllers/metrics/node/controller.go
njtran: Additionally, you could use a `multi.err` and log them together to reduce log noise.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+// GetPodsTotalRequestsAndLimits calculates the total resource requests and limits for the pods.
+// If pod overhead is non-nil, the pod overhead is added to the
+// total container resource requests and to the total container limits which have a non-zero quantity.
+func getPodsTotalRequestsAndLimits(pods []v1.Pod) (reqs map[v1.ResourceName]resource.Quantity, limits map[v1.ResourceName]resource.Quantity) {
+	reqs, limits = map[v1.ResourceName]resource.Quantity{}, map[v1.ResourceName]resource.Quantity{}
+	for _, pod := range pods {
+		// Excluding pods that are completed or failed
+		if pod.Status.Phase == v1.PodFailed || pod.Status.Phase == v1.PodSucceeded {
```
njtran, line 304: Can you use the pod util function `IsTerminal` here? https://github.com/aws/karpenter/blob/main/pkg/utils/pod/scheduling.go#L31-L33

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,188 @@
...
+	metricLabels := prometheus.Labels{}
+	metricLabels[podName] = pod.GetName()
+	metricLabels[podNameSpace] = pod.GetNamespace()
+	// Selflink has been deprecated after v.1.20
+	// Manually generate the selflink for the first owner reference
+	// Currently we do not support multiple owner references
+	selflink := ""
+	if len(pod.GetOwnerReferences()) > 0 {
+		ownerreference := pod.GetOwnerReferences()[0]
+		selflink = fmt.Sprintf("/apis/%s/namespaces/%s/%ss/%s", ownerreference.APIVersion, pod.Namespace, strings.ToLower(ownerreference.Kind), ownerreference.Name)
```
njtran, line 158: Was the extra `s` here intentional?
```suggestion
		selflink = fmt.Sprintf("/apis/%s/namespaces/%s/%s/%s", ownerreference.APIVersion, pod.Namespace, strings.ToLower(ownerreference.Kind), ownerreference.Name)
```

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,188 @@
...
+		metricLabels[podHostZone] = node.Labels[v1.LabelTopologyZone]
+		metricLabels[podHostArchitecture] = node.Labels[v1.LabelArchStable]
+		metricLabels[podHostCapacityType] = node.Labels[v1alpha5.LabelCapacityType]
+		metricLabels[podHostInstanceType] = node.Labels[v1.LabelInstanceTypeStable]
+		metricLabels[podProvisioner] = node.Labels[v1alpha5.ProvisionerNameLabelKey]
+	}
+	// Add pod labels
+	labels, err := json.Marshal(pod.GetLabels())
+	if err != nil {
+		return nil, fmt.Errorf("marshal pod labels: %w", err)
```
njtran, line 184: Same suggestion as above.
```suggestion
		return nil, fmt.Errorf("marshalling pod metric labels: %w", err)
```

pkg/controllers/metrics/pod/controller.go
njtran: Additionally, what do you think about creating this as a constant? We may want to use it elsewhere.

```
selfLinkFormatString := "/apis/%s/namespaces/%s/%ss/%s"
```

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,188 @@
...
+
+// Reconcile executes a termination control loop for the resource
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("podmetrics").With("pod", req.Name))
+
+	// Retrieve pod from reconcile request
+	pod := &v1.Pod{}
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, pod); err != nil {
+		if errors.IsNotFound(err) {
+			// Remove gauge due to pod deletion
```
njtran, line 106: I have the same question here as above as with the node metrics.

njtran (commented):
Nice work!

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+	if err := c.updateGauges(ctx, node); err != nil {
+		logging.FromContext(ctx).Debugf("Failed to update gauges: %s", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
+	err := controllerruntime.
```
ellistarn, line 157:
```suggestion
	return controllerruntime.
```

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+		return reconcile.Result{}, err
+	}
+	// Remove the previous gauge after node labels are updated
+	if labelSlice, ok := c.LabelsMap[req.NamespacedName]; ok {
+		c.deleteGauges(labelSlice)
+	}
+	c.LabelsMap[req.NamespacedName] = []prometheus.Labels{}
+
+	if err := c.updateGauges(ctx, node); err != nil {
+		logging.FromContext(ctx).Debugf("Failed to update gauges: %s", err.Error())
```
ellistarn, line 149:
```suggestion
		logging.FromContext(ctx).Errorf("Failed to update gauges: %s", err.Error())
```

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,340 @@
...
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("nodemetrics").With("node", req.Name))
+	// Retrieve node from reconcile request
+	node := &v1.Node{}
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, node); err != nil {
+		if errors.IsNotFound(err) {
+			// Remove gauge due to node deletion
+			if labels, ok := c.LabelsMap[req.NamespacedName]; ok {
+				c.deleteGauges(labels)
+			} else {
+				logging.FromContext(ctx).Debugf("Failed to delete gauge: failed to locate labels")
```
ellistarn, line 136:
```suggestion
				logging.FromContext(ctx).Errorf("Failed to delete gauge, failed to locate labels")
```

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: good suggestion. Done.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: User can add prometheus label filters based on values. But they can't use prometheus label that are not defined in the controller.  Therefore, if user want to add a K8S label and want to filter according to that, we will have to use a workaround. The hack is to put all the K8S label key value pairs under one prometheus label (i called it pod_labels) as a json file. Then we do string parsing and filtering regarding to this.

After some discussion we conclude that there is no good use case at this time. Maybe we can revisit this idea later.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: After thinking about this for a while I think we should keep the one that I have here.
First is that the helper function does not return limits. Also I want to calculate the request and limits in one for loop.
Also about the pod.spec.overhead calculation. It is not doing anything now so I think we can keep it. 

Later I think we should update the helper function to also include limits calculation.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: I agree with you. But here is my counter argument for this case.
Resources in K8S libraries are mainly described using v1.ResourceList. Most functions return this type. I think it will be more consistent to also use v1.ResourceList as input and output variables.

Also taking the for loop outside will creates quite a few very repetitive for loops that looks like what I have written in CS101.

ellistarn: Totally optional. see my other comment though:
```
for vec, resources := range map[*prometheus.GaugeVec]map[v1.ResourceName]resource.Quantity{
   overheadGaugeVec: overhead,
   requestsGaugeVec: requests,
   ...
}{
  for ... {
      c.insert...
  }
}
```

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Good suggestions! done!

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Done.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Yes now refactored to return system overhead (the three gray boxes) and daemonset pod separatly.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: yes agree! done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: The set of labels are determinstic. But the values are not.
Both delete functions requires matching of the label values. I dig very deep into the source code and try to find a clever way to delete gauge. Unfortunately this doesn't seem possible since the library does not perform label filtering and joining for gauge indexing.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Good point. I agree. Done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Good question. deleting the GaugeVec does not delete the past data/metrics in Prometheus database. It just tell Prometheus to stop tracking this by closing the active time series. So Prometheus will have the perfect memory of everything that happened in the past.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: In the current design the controller reconcile one node at a time. Do we need to reconcile all nodes when there are pods that have changed?

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Yeah user won't be able to define prometheus labels. Because when someone defines a label, he/she need to teach the controller how to calculate the values for node/pods.

felix-zhe-huang: We are no longer returning the node labels since there is no good use case for now.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Yup. Good suggestion. Using the helper function now.

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: > Was the extra `s` here intentional?

Yes. very hacky way to generate the selflink. In the selflink they want sometime like "replicasets" but the `ownerreference.Kind = "replicaset"`. I am not 100% happy about this. 

> Additionally, what do you think about creating this as a constant? We may want to use it elsewhere.
I think this is really cool idea. And we should really have a helper function to do that.
But this can be potentially controversial since K8S deprecated selflink. 
Can somebody educate me why they decided to deprecate selflink?

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: not including this anymore

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: Please refer to my comments under the previous question.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

felix-zhe-huang:
I am removing the reference dashboard. The dashboard is changing too frequently. Let's fine tune it carefully and submit a new PR for that.

pkg/controllers/metrics/node/controller.go
njtran: Got it. Does this mean that for some visualization, nodes that don't have datapoints within that time frame won't show up in the legend? If they do, that may be too much clutter.

pkg/controllers/metrics/node/controller.go
njtran: IIUC, since the condition below does `name != ""` the code will functionally queue up requests to all nodes that have pod changes where the nodeName is not defined. This is intended and may be functionally the same, but is a slightly different process from enqueueing a request to one node when any pod on that node changes. 
```
				if name := o.(*v1.Pod).Spec.NodeName; name != "" {
```
You could also make it
`// Reconcile nodes where pods have changed`

pkg/controllers/metrics/node/controller.go
njtran: I see. Can you cut an issue here for tracking in the future so that we don't forget about it? 

If possible, it would be good to have a non-hacky way to do this. When we have a better understanding of what metrics we'll need and what use-cases are needed, I think we could have another discussion.

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: Turns out that Prometheus does not like "". It will basically create a null value for the label. It confused the concept of not having a label value and an empty value. I am using N/A so that it will be more readable to the users.

pkg/controllers/metrics/node/controller.go
njtran: Was this comment in the right place? I wanted to follow convention of using -ing verbs in our `fmt.Errorf()` messages.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: You are exactly right. If there is no data, it will not show up in the graph.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Great suggestion! Done.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Yeah user won't be able to define prometheus labels. Because when someone defines a label, he/she need to teach the controller how to calculate the values for node/pods.

njtran (approved):
LGTM. Nice work.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,345 @@
...
+		nodeArchitecture,
+		nodeCapacityType,
+		nodeInstanceType,
+		nodePhase,
+	}
+}
+
+type Controller struct {
+	KubeClient    client.Client
+	LabelSliceMap map[types.NamespacedName][]prometheus.Labels
```
ellistarn, line 131: IIUC, go considers naming variables with their type to be a bit of an anti-style. I'd just call this labels.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,345 @@
...
+			allocatableGaugeVec.Delete(labels)
+			podRequestsGaugeVec.Delete(labels)
+			podLimitsGaugeVec.Delete(labels)
+			daemonRequestsGaugeVec.Delete(labels)
+			daemonLimitsGaugeVec.Delete(labels)
+			overheadGaugeVec.Delete(labels)
+		}
+	}
+	c.LabelSliceMap[nodeNamespacedName] = []prometheus.Labels{}
+
```
ellistarn, line 207: unnecessary newline
```suggestion
```

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,345 @@
...
+					requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Name: name}})
+				}
+				return requests
+			}),
+		).
+		Complete(c)
+}
+
+func (c *Controller) deleteGauges(nodeNamespacedName types.NamespacedName) {
+	if labelSlice, ok := c.LabelSliceMap[nodeNamespacedName]; ok {
```
ellistarn, line 196: similarly
```suggestion
	if label, ok := c.Labels[nodeNamespacedName]; ok {
```

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,345 @@
...
+		KubeClient:    kubeClient,
+		LabelSliceMap: make(map[types.NamespacedName][]prometheus.Labels),
+	}
+}
+
+// Reconcile executes a termination control loop for the resource
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("nodemetrics").With("node", req.Name))
+	// Remove the previous gauge after node labels are updated
+	c.deleteGauges(req.NamespacedName)
```
ellistarn, line 146: Naming nit:
```suggestion
	c.reset(req.NamespacedName)
```

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,345 @@
...
+	// Retrieve node from reconcile request
+	node := &v1.Node{}
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, node); err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+	if err := c.updateGauges(ctx, node); err != nil {
```
ellistarn, line 155: Thoughts on using Prometheuses names "Delete()" and "Record()":

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,345 @@
...
+
+}
+
+// generateLabels creates the labels using the current state of the pod
+func (c *Controller) generateLabels(node *v1.Node, resourceTypeName string) prometheus.Labels {
+	metricLabels := prometheus.Labels{}
+	metricLabels[resourceType] = resourceTypeName
+	metricLabels[nodeName] = node.GetName()
+	if provisionerName, ok := node.Labels[v1alpha5.ProvisionerNameLabelKey]; !ok {
+		metricLabels[nodeProvisioner] = "N/A"
```
ellistarn, line 216: Why does this have to be N/A? Can't we just use "" and simplify to 

metricsLabels[nodeProvisioner] = node.Labels[v1alpha5.ProvisionerNameLabelKey]

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,345 @@
...
+	} else {
+		metricLabels[nodeProvisioner] = provisionerName
+	}
+	metricLabels[nodeZone] = node.Labels[v1.LabelTopologyZone]
+	metricLabels[nodeArchitecture] = node.Labels[v1.LabelArchStable]
+	if capacityType, ok := node.Labels[v1alpha5.LabelCapacityType]; !ok {
+		metricLabels[nodeCapacityType] = "N/A"
```
ellistarn, line 223: Similarly "N/A" -> ""

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,345 @@
...
+	metricLabels[nodePhase] = string(node.Status.Phase)
+	return metricLabels
+}
+
+func (c *Controller) updateGauges(ctx context.Context, node *v1.Node) error {
+	podlist := &v1.PodList{}
+	if err := c.KubeClient.List(ctx, podlist, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
+		return fmt.Errorf("listing pods on node %s, %w", node.Name, err)
+	}
+	var daemonSetPods, pods []*v1.Pod
```
ellistarn, line 237:
```suggestion
	var daemons, pods []*v1.Pod
```

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,345 @@
...
+			continue
+		}
+		for _, container := range pod.Spec.Containers {
+			// Calculate Resource Requests
+			addResourceQuantity(container.Resources.Requests, reqs)
+			// Calculate Resource Limits
+			addResourceQuantity(container.Resources.Limits, limits)
+		}
+		// Add overhead for running a pod to the sum of requests and to non-zero limits:
+		if pod.Spec.Overhead != nil {
```
ellistarn, line 319: IIRC we were going to skip Overhead and just use the existing `resources.RequestsForPods(pods), resources.LimitsForPods(pods)` helpers.

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,345 @@
...
+			overhead := node.Status.Capacity[resourceName]
+			overhead.Sub(quantity)
+			systemOverheads[resourceName] = overhead
+		}
+	}
+	return systemOverheads
+}
+
+func (c *Controller) insertGaugeValues(resourceList v1.ResourceList, node *v1.Node, gaugeVec *prometheus.GaugeVec) error {
+
```
ellistarn, line 282:
```suggestion
```

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,186 @@
...
+		ownerSelfLink,
+		podHostName,
+		podProvisioner,
+		podHostZone,
+		podHostArchitecture,
+		podHostCapacityType,
+		podHostInstanceType,
+		podPhase,
+	}
+
```
ellistarn, line 84:
```suggestion
```

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// NewController constructs a controller instance
+func NewController(kubeClient client.Client) *Controller {
+	return &Controller{
+		KubeClient: kubeClient,
+		LabelsMap:  make(map[types.NamespacedName]prometheus.Labels),
```
ellistarn, line 91: Why is naming styled differently on this one?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,186 @@
...
+// Reconcile executes a termination control loop for the resource
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("podmetrics").With("pod", req.Name))
+	// Retrieve pod from reconcile request
+	pod := &v1.Pod{}
+	if err := c.KubeClient.Get(ctx, req.NamespacedName, pod); err != nil {
+		if errors.IsNotFound(err) {
+			// Remove gauge due to pod deletion
+			if labels, ok := c.LabelsMap[req.NamespacedName]; ok {
+				podGaugeVec.Delete(labels)
```
ellistarn, line 104: Can we follow the same pattern as the other controller where we delete first in all cases?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,186 @@
...
+		podGaugeVec.Delete(labels)
+	}
+	if err := c.updateGauges(ctx, pod); err != nil {
+		logging.FromContext(ctx).Errorf("Failed to record metrics, %s", err.Error())
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) updateGauges(ctx context.Context, pod *v1.Pod) error {
```
ellistarn, line 123: Feels odd to me that the two controllers follow different patterns. This controller relies on a helper to generate labels, while the node controller does it all inline. Is there any reason for this? 

Some of these comments may seem like nit picks, and they are, but they are valuable for project health and for the development of your own style. Do your best to be intentional about every line you write. It may feel like it slows you down in the short term, but it will accelerate you in the long term. Happy to discuss more of these ideas in a less public setting.

pkg/controllers/metrics/pod/controller.go
ellistarn: This pluralization is one of the awkward bits of the kubernetes API. It's an english semantic that only holds true in some cases, e.g. "sheep" -> "sheeps" 😭 . It maybe one of the reasons that Kubernetes deprecated selflink, since it is one of the rare places where this pluralization is used. 

Instead of relying on ownerreferences or jsonified-labels (which both seems like it's a bit awkward here for different reasons), we could instead support https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels like `app.kubernetes.io/name`, `app.kubernetes.io/part-of`. I'm not sure we need to necessarily be in the business of perfectly tying together all possible connections of concepts.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,186 @@
...
+	if len(pod.GetOwnerReferences()) > 0 {
+		ownerreference := pod.GetOwnerReferences()[0]
+		selflink = fmt.Sprintf("/apis/%s/namespaces/%s/%ss/%s", ownerreference.APIVersion, pod.Namespace, strings.ToLower(ownerreference.Kind), ownerreference.Name)
+	}
+	metricLabels[ownerSelfLink] = selflink
+	metricLabels[podHostName] = pod.Spec.NodeName
+	metricLabels[podPhase] = string(pod.Status.Phase)
+	node := &v1.Node{}
+	if err := c.KubeClient.Get(ctx, types.NamespacedName{Name: pod.Spec.NodeName}, node); err != nil {
+		metricLabels[podHostZone] = "N/A"
```
ellistarn, line 161: Same for N/A vs "".

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,186 @@
...
+	return nil
+}
+
+func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
+	err := controllerruntime.
+		NewControllerManagedBy(m).
+		Named("podmetrics").
+		For(&v1.Pod{}).
+		Complete(c)
+	return err
```
ellistarn, line 140:
```suggestion
	return controllerruntime.
		NewControllerManagedBy(m).
		Named("podmetrics").
		For(&v1.Pod{}).
		Complete(c)
```

pkg/controllers/metrics/pod/controller.go
```diff
@@ -0,0 +1,186 @@
...
+	if err := c.updateGauges(ctx, pod); err != nil {
+		logging.FromContext(ctx).Errorf("Failed to record metrics, %s", err.Error())
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) updateGauges(ctx context.Context, pod *v1.Pod) error {
+	newlabels := c.generateLabels(ctx, pod)
+	gauge, err := podGaugeVec.GetMetricWith(newlabels)
```
ellistarn, line 125: We control all the inputs and outputs here, so we can guarantee it won't panic if you use `With` instead:
```
// With works as GetMetricWith, but panics where GetMetricWithLabels would have
// returned an error. Not returning an error allows shortcuts like
//     myVec.With(prometheus.Labels{"code": "404", "method": "GET"}).Add(42)
func (v *GaugeVec) With(labels Labels) Gauge {
	g, err := v.GetMetricWith(labels)
	if err != nil {
		panic(err)
	}
	return g
}
```

ellistarn: This simplifies your implementation to
```
podGaugeVec.With(labels).Set(float64(1))
c.LabelsMap[client.ObjectKeyFromObject(pod)] labels
```

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Good point. Done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: I tried this. Unfortunately Prometheus automatically ignores the label value. It creates confusion since we can no longer distinguish if the label value doesn't exist or it is just empty.

Also this N/A is more human readable so that the customer will know that we don't have a value for this label.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: Please refer to the above comment.

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
ellistarn: It would be nice to just let them be empty string, and then we can postprocess later. This isolates prometheus's quirk and allows us to follow go conventions elsewhere.
```
for key, value := range metricLabels {
  if value == "" {
    metricLabels[key] = "N/A"
  }
}
```

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/node/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: The pod controller only keeps a map of labels, however, the node controller keeps a map of slice of labels. I don't want to call the one in node controller just labels since that will confuse the layer of structure here.

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: The node controller also calls the generateLabels helper function. But instead of calling it in updateGauges, it calls the helper function in insertGaugeValues. The updateGauges function in the node controller is too long and break the gocyclo test without breaking it down into subroutine (i.e., insertGaugeValues).

I totally agree with you. These suggestions make great sense and can only be learned through experience and careful thinking. I really appreciate them!

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: Please refer to the above comment.

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: done

pkg/controllers/metrics/pod/controller.go
felix-zhe-huang: Great suggestion! done!

pkg/controllers/metrics/node/controller.go
```diff
@@ -0,0 +1,301 @@
...
+	nodeProvisioner  = "provisioner"
+	nodeZone         = "zone"
+	nodeArchitecture = "arch"
+	nodeCapacityType = "capacity_type"
+	nodeInstanceType = "instance_type"
+	nodePhase        = "phase"
+)
+
+var (
+	allocatableGaugeVec = prometheus.NewGaugeVec(
```
ellistarn, line 52: More naming nits. Consider omitting the type from these variable names. It adds a fair amount of bloat, and any modern IDE can show you the type by mousing over.

ellistarn (approved):
Nice work Felix!

PR #970, opened 2021-12-11
(aws) Expose instance metadata options in the provisioner spec

### LINKED ISSUE #969: Support configuring instance metadata options for AWS provisioner
What do you want us to build?

AWS currently documents EKS security best practices, including restricting access to Instance Metadata Service (IMDS) from pods (https://aws.github.io/aws-eks-best-practices/security/docs/iam/#restrict-access-to-the-instance-profile-assigned-to-the-worker-node). Exposure of IMDS is controlled via the metadata options on the Launch Template of provisioned nodes.

EKS Managed Node Groups currently defaults the metadata options to expose both IMDSv1 and IMDSv2 with a maximum hop limit of 2 (https://aws.amazon.com/about-aws/whats-new/2020/08/amazon-eks-supports-ec2-instance-metadata-service-v2/). This favors compatibility for workloads rather than the recommended security hardening.

For Karpenter, it is currently possible to harden security for IMDS by separately creating a Launch Template and referencing it from the Provisioner resource. It would be nice if metadata options were exposed directly in the Provisioner API so that the security best practices could be followed without needing a separate tool like CloudFormation or Terraform for the Launch Templates.

Something like:
```
kind: Provisioner
metadata:
  name: example
spec:
  ...
  provider:
    ...
    metadataOptions:
      httpEndpoint: enabled
      httpPutResponseHopLimit: 1
      httpTokens: required
```

The goal here is to ensure that Provisioner API can be a one-stop shop for users following EKS best practices, without the need to manage EC2 API resources directly.

The current approach for anyone following best practices would be to manage a custom Launch Template themselves.

**Additional context**
Anything else we should know?

**Attachments**

### Community Note

orirawlings:
**1. Issue, if available:**

Fixes #969 

This PR extends the Provisioner API for AWS to allow customizing the metadata options on the Launch Templates created by Karpenter. This enables users to follow EKS Security Best Practices and configure hardened settings to limit exposure of Instance Metadata Service (IMDS) to pods (https://aws.github.io/aws-eks-best-practices/security/docs/iam/#restrict-access-to-the-instance-profile-assigned-to-the-worker-node).

The default configuration, if `metadataOptions` is omitted, enables IMDS on IPv4 only, requires authentication, and is accessible from both the hostNetwork and pods.
```
spec:
  provider:
    metadataOptions:
      httpEndpoint: enabled
      httpProtocolIpv6: disabled
      httpPutResponseHopLimit: 2
      httpTokens: required
```

Note, this diverges from the precedent established by EKS Managed Node Groups, which creates a configuration that favors backward compatibility (https://aws.amazon.com/about-aws/whats-new/2020/08/amazon-eks-supports-ec2-instance-metadata-service-v2/), i.e. HTTP Tokens are optional and a hop limit of 2.

- [x] Yes, issue opened: #969 
- [ ] No


orirawlings:
I'm just opening this as a draft for the moment. I'm still working on getting more testing and docs in place. If anyone has time, this is my first contribution so I'm receptive to any feedback on the API, approach, design, etc.

I'm not sure if this warrants bumping the Provisioner API to v1alpha6 or not, so if anyone can clarify guidance on that it would be helpful.

ellistarn:
>I'm just opening this as a draft for the moment. I'm still working on getting more testing and docs in place. If anyone has time, this is my first contribution so I'm receptive to any feedback on the API, approach, design, etc.

I'm very excited to see this work! Thank you!

> I'm not sure if this warrants bumping the Provisioner API to v1alpha6 or not, so if anyone can clarify guidance on that it would be helpful. I suspect it might be warranted because it changes the defaulting and validation behavior which would conflict with any existing resources already present in the API at v1alpha5.

As long as it's backwards compatible, you don't need to change the API. Since it only adds fields and does not remove, I think you're in the clear. It depends how you implement validation/defaults. As long as we leave this field optional in the api, and then default to values in memory, you're good to go.

ellistarn:
Check out the latest discussion here: #969

orirawlings:
> As long as it's backwards compatible, you don't need to change the API. Since it only adds fields and does not remove, I think you're in the clear. It depends how you implement validation/defaults. As long as we leave this field optional in the api, and then default to values in memory, you're good to go.

I think the way I'm currently doing things, the validation and defaulting aren't happening strictly in memory. Can you point me to another example in the code that is defaulting strictly in memory? I'm having a little trouble thinking through the implications for existing Provisioner CRs in the cluster that might not pass through the defaulting webhook again (if they are preexisting). Also, I think there would be implications to the existing cache keys for launch templates?

> Regarding the approach, I think we want to be careful about which LT features we support -- many of them do not make sense. For a sense of how big the problem is, search issues with the tag launch-template. I can see the spec.provider eventually growing to a subset of LTs, but I want to make sure that it doesn't explode into too many parameters.

Agreed. I'd hate to see the Provisioner API converging to a superset of Launch Template.

> I can also see this as an aws-specific CLI flag. This would allow karpenter to simply use the imdsv2 best practices with an escape hatch for customers that haven't migrated yet. I'm not sure I see the use case for mixed imdsv1/2 in a single cluster.

My initial need is actually a mixed cluster scenario and I suspect many other users will require the same in the future. Basically, disabling access to IMDSv1 or to IMDS altogether from pods is a breaking change for pod workloads. So I have to configure multiple node pools with labels and taints to allow workloads to gradually migrate to "secure" nodes, while I keep separate nodes around for backward compatibility. I imagine other organizations that want to gradually repair their workloads to function without IMDS access will want to pursue a similar path.

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
```diff
@@ -49,6 +49,55 @@ type AWS struct {
...
+// MetadataOptions contains parameters for specifying the exposure of the
+// Instance Metadata Service to provisioned EC2 nodes.
+type MetadataOptions struct {
+	// HttpEndpoint enables or disables the HTTP metadata endpoint on provisioned
+	// nodes. If the parameter is not specified, the default state is "enabled".
+	//
+	// If you specify a value of "disabled", instance metadata will not be accessible
+	// on the node.
+	// +optional
+	HttpEndpoint string `json:"httpEndpoint,omitempty"`
```
bwagner5, line 76: generally, acronyms and initialisms in go should have the same case i.e. `HTTPEndpoint`. Although I realize the aws-sdk-go doesn't follow that :(    

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

bwagner5 (commented):
nice work on this PR!

bwagner5:
> Also, I think there would be implications to the existing cache keys for launch templates?

The LaunchTemplateOptions hash will be changed adding this, but that's what will have to happen to generate new LTs for the new IMDS options, so I think that's fine.

ellistarn:
Thoughts on what we discussed on slack?
`imds: v1 | v2 | v1v2`

bwagner5:
> imds: v1 | v2 | v1v2

We'd need another property to cover hop count since it seems desirable to adjust hops so that only hostNetworking pods can access IMDS. 

```
imds: 
  version: v1 | v2 | v1v2
  hops: [0-64]
```

ellistarn:
At this point, we should probably expose the full spec.

orirawlings:
Sorry, I've been busy swatting other problems the last few days (log4j and AWS outages 😢).

@ellistarn I did consider a simplified API some more. I was thinking that maybe we could collapse this into a single field with a few enumerated options to cover a handful of common/best-practice configurations (maybe 3 or 4 max). Any users that need more granular control could still fall back to referencing an external launch template.

I'm thinking of some options like the following:

- `imds: Disabled`
  - IMDS is not exposed to nodes or pods at all
  - This is equivalent to the following launch template settings:
     ```
     metadataOptions:
       httpEndpoint: disabled
    ```
- `imds: HostNetworkOnly`
  - IMDSv2 is exposed with a hop limit of 1, IMDSv1 isn't possible
  - This is equivalent to the following launch template settings:
     ```
     metadataOptions:
       httpEndpoint: enabled
       httpPutResponseHopLimit: 1
       httpTokens: required
    ```
- `imds: HostAndPodNetwork`
  - IMDSv2 is exposed with a hop limit of 2, IMDSv1 isn't possible
  - This would be the default setting if `imds` field is unspecified. Note this is still a departure from the current default Launch Template settings that Karpenter provisions.
  - This is equivalent to the following launch template settings:
     ```
     metadataOptions:
       httpEndpoint: enabled
       httpPutResponseHopLimit: 2
       httpTokens: required
    ```

Karpenter might be on the hook to add support for additional enum options in the future if new best-practice configurations emerge (for ex. if IMDSv3 is ever a thing).

Let me know your thoughts on this. It will probably make the implementation a bit easier as the defaulting behavior will be a little bit better contained to a single field and could be handled in memory more easily.

orirawlings:
With the suggested options I mentioned above note that none of them are backward compatible with things that require IMDSv1 to be exposed (i.e. `httpTokens: optional`).

Also, the naming of the options I suggested might imply certain behaviors that are AWS VPC CNI specific (regarding how it does NAT between pods and IMDS API)? My understanding here is a little fuzzy. I'm not an expert on either how VPC CNI handles NAT or how IMDSv2 enforces the hop limit. I think my only concern is that a term like `HostNetworkOnly` might be a misnomer for clusters with a different setup than my own.

ellistarn:
I like this approach, but agree about the naming. I think we also need an option where IMDS v1 is supported. In fact, I thought that this was something you needed.

ellistarn:
@suket22 may have some interesting ideas here.

orirawlings:
> I like this approach, but agree about the naming. I think we also need an option where IMDS v1 is supported. In fact, I thought that this was something you needed.

Yeah, I'd be ok with an option where IMDSv1 is supported. I'm open to good naming suggestions though. :)

orirawlings:
> > I like this approach, but agree about the naming. I think we also need an option where IMDS v1 is supported. In fact, I thought that this was something you needed.
> 
> Yeah, I'd be ok with an option where IMDSv1 is supported. I'm open to good naming suggestions though. :)

Maybe something that suggests the insecurity of the setup?
- `imds: NoLoginRequired` or `imds: AnonymousAccessAllowed`
  - Both IMDSv1 and IMDSv2 are exposed to pods and host network
  - This is equivalent to the following launch template settings:
     ```
     metadataOptions:
       httpEndpoint: enabled
       httpPutResponseHopLimit: 2
       httpTokens: optional
    ```

bwagner5:
I'm a little torn on using shorthand enums that may not convey a configuration very clearly. 

The users configuring this are probably already familiar w/ configuring IMDS or else they wouldn't be touching the defaults we set. Rather than creating our own names and mapping them to reduce the configuration, I think using the AWS configuration names are going to be clearer for users trying to configure this and finding relevant docs within AWS for what they are actually setting up.

If we come up with a clearer way of doing this later when we address other LT configurations we want to expose in the Provisioner, we could revise at that time. 

```
metadataOptions:
  httpEndpoint: enabled
  httpPutResponseHopLimit: 2
  httpTokens: optional
```

pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
```diff
@@ -75,3 +77,24 @@ func (c *Constraints) defaultSecurityGroups(clusterName string) {
...
+)
+
+func (c *Constraints) defaultMetadataOptions() {
+	if c.MetadataOptions == nil {
+		c.MetadataOptions = &MetadataOptions{}
+	}
+	if c.MetadataOptions.HTTPEndpoint == "" {
+		c.MetadataOptions.HTTPEndpoint = DefaultMetadataOptionsHTTPEndpoint
+	}
+	if c.MetadataOptions.HTTPPutResponseHopLimit == 0 {
```
ellistarn, line 94: Maybe a dumb question. Is it possible that a user would want to configure the hop to 0? Does this mean anything? If so, we'd want to make this a pointer so it can be left nil.

pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
```diff
@@ -75,3 +77,24 @@ func (c *Constraints) defaultSecurityGroups(clusterName string) {
...
+		c.MetadataOptions.HTTPPutResponseHopLimit = DefaultMetadataOptionsHTTPPutResponseHopLimit
+	}
+	if c.MetadataOptions.HTTPTokens == "" {
```
ellistarn, line 97: If it makes sense for this to potentially be nil, let's make it a ptr

pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
```diff
@@ -75,3 +77,24 @@ func (c *Constraints) defaultSecurityGroups(clusterName string) {
...
+	DefaultMetadataOptionsHTTPEndpoint            = ec2.LaunchTemplateInstanceMetadataEndpointStateEnabled
+	DefaultMetadataOptionsHTTPPutResponseHopLimit = 2
+	DefaultMetadataOptionsHTTPTokens              = ec2.LaunchTemplateHttpTokensStateOptional
+)
+
+func (c *Constraints) defaultMetadataOptions() {
+	if c.MetadataOptions == nil {
+		c.MetadataOptions = &MetadataOptions{}
+	}
+	if c.MetadataOptions.HTTPEndpoint == "" {
```
ellistarn, line 91: If it makes sense for this to potentially be nil, let's make it a ptr

pkg/cloudprovider/aws/apis/v1alpha1/provider_validation.go
```diff
@@ -81,3 +84,44 @@ func (a *AWS) validateTags() (errs *apis.FieldError) {
...
+	return errs.Also(
+		a.validateHTTPEndpoint(),
+		a.validateHTTPPutResponseHopLimit(),
+		a.validateHTTPTokens(),
+	).ViaField("metadataOptions")
+}
+
+func (a *AWS) validateHTTPEndpoint() (errs *apis.FieldError) {
+	errs = a.validateStringEnum(a.MetadataOptions.HTTPEndpoint, "httpEndpoint", ec2.LaunchTemplateInstanceMetadataEndpointState_Values())
+	return errs
```
ellistarn, line 102:
```suggestion
	return a.validateStringEnum(a.MetadataOptions.HTTPEndpoint, "httpEndpoint", ec2.LaunchTemplateInstanceMetadataEndpointState_Values())
```

pkg/controllers/provisioning/provisioner.go
```diff
@@ -47,6 +47,8 @@ var (
 )
 
 func NewProvisioner(ctx context.Context, provisioner *v1alpha5.Provisioner, kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.CloudProvider) *Provisioner {
+	provisioner = provisioner.DeepCopy()
+	provisioner.SetDefaults(ctx)
```
ellistarn, line 51: I'm not sure how comfortable I am with runtime defaults. What's the advantage of this vs just defaulting this where it's used?

pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
orirawlings: The AWS API only allows values from 1 to 64. I don't think there is a valid case for 0.

pkg/controllers/provisioning/provisioner.go
orirawlings: I think my main concern is DRYing defaulting logic. If we don't need to support defaulting metadata option fields during mutating webhook then the defaulting for metadata options fields can probably go elsewhere.

Of course, if we don't default the new fields with a mutating webhook then we also have to loosen some of the validation that will run during the validating webhook (which currently requires that the `metadataOptions` field is non-nil). Looser validation might be acceptable though.

orirawlings: My thinking was that new fields are all optional, but they are defaulted and validated at the API server (via webhooks). That gives the advantage of rejecting bad objects synchronous with the user. It also more clearly communicates to the user what the values will be when defaulted (as they become visible in the API upon creation).

The main disadvantage is when updating the controller to a version that includes this new feature. Any objects previously created in the API were not defaulted yet. So the controller would essentially be reconciling invalid objects. By defaulting again here we could fix all previously created objects in the API in-memory as part of reconciliation.

I suppose that by defaulting only where the metadataOption fields are consumed we can side-step the issue of invalid objects previously stored in the API. I agree that might be a little cleaner and we'd just have to lean more on documentation to accurately communicate default values to the user. Let me give that approach a go.

pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
orirawlings: I stuggled a bit with this actually. I'm not sure what would work best. It seems like since an empty string would never be a valid value either, it wouldn't matter much if we added the indirection of a pointer. As such, either `""` or `nil` could represent that the user hasn't specified a value and we should pick for them.

I can change this to a pointer though.

pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
orirawlings: I'll update this too.

alex-berger:
Can we please also add `httpProtocolIpv6` to `MetadataOptions`, to support all currently available metadata options.

orirawlings:
I have to do some more testing in real clusters yet, but moving this Ready for review.

orirawlings:
I went ahead and stuck to exposing a similar API to the underlying launch template. I also added support for `httpProtocolIpv6`. All defaulting behavior is done in-memory during pod reconciliation rather than in the defaulting webhook. Defaulting only takes effect if the entire `metadataOptions` field is omitted. If `metadataOptions` is partially specified, defaulting of individual fields is deferred to the underlying createLaunchTemplate API of EC2.

orirawlings:
Alright, I had finally had time to test this out in some real clusters and everything is looking good from my side.

If anyone has a chance to review/integrate that'd be awesome! Thanks!

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
```diff
@@ -49,6 +49,66 @@ type AWS struct {
...
+	// If you specify a value of "disabled", instance metadata will not be accessible
+	// on the node.
+	// +optional
+	HTTPEndpoint *string `json:"httpEndpoint,omitempty"`
+
+	// HTTPProtocolIpv6 enables or disables the IPv6 endpoint for the instance metadata
+	// service on provisioned nodes. If metadata options is non-nil, but this parameter
+	// is not specified, the default state is "disabled".
+	// +optional
+	HTTPProtocolIpv6 *string `json:"httpProtocolIpv6,omitempty"`
```
ellistarn, line 86: Can you try to find an example of a K8s API that uses IPV6 vs Ipv6 vs IPv6 as a YAML **key**?

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -81,6 +81,14 @@ type launchTemplateOptions struct {
 	SecurityGroupsIds []string
 	AMIID             string
 	Tags              map[string]string
+	MetadataOptions   launchTemplateMetadataOptions
+}
+
+type launchTemplateMetadataOptions struct {
```
ellistarn, line 87: Why do we need a new type for this? Can't we use the v1alpha1.MetadataOptions type?

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -123,6 +133,30 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, constraints *v1alpha1.
...
 }
 
+const (
+	defaultMetadataOptionsHTTPEndpoint            = ec2.LaunchTemplateInstanceMetadataEndpointStateEnabled
+	defaultMetadataOptionsHTTPProtocolIpv6        = ec2.LaunchTemplateInstanceMetadataProtocolIpv6Disabled
+	defaultMetadataOptionsHTTPPutResponseHopLimit = 2
+	defaultMetadataOptionsHTTPTokens              = ec2.LaunchTemplateHttpTokensStateRequired
+)
+
+func (p *LaunchTemplateProvider) getMetadataOptions(options *v1alpha1.MetadataOptions) launchTemplateMetadataOptions {
```
ellistarn, line 143: Consider attaching this behavior to the constraints object itself. 

constraints.GetMetadataOptions()

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
orirawlings: On Node [`status.address.addressType`](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/node-v1/#NodeStatus), `ExternalIP` and `InternalIP` are possible values. 

On Service, we have [`spec`](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec) fields like `clusterIP`, `clusterIPs`, `externalIPs`, etc. [Valid values for `ipFamilies`](https://kubernetes.io/docs/tasks/network/validate-dual-stack/) seems to include `IPv6`.

I'll go ahead and change this field to `httpProtocolIPv6`.

ellistarn: I think I come to the same confusion. For further nuance, values follow UpperCamel where keys follow lowerCamel. I *think* because the `spec.clusterIP` uses caps, it means we should use `IPv6` instead of `Ipv6`

orirawlings:
@ellistarn I went ahead and addressed your review feedback. Feel free to take another look. Thanks!

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -538,6 +579,112 @@ var _ = Describe("Allocation", func() {
 				}
 			})
 		})
+		Context("MetadataOptions", func() {
+			It("should allow missing values", func() {
+				provider, err := ProviderFromProvisioner(provisioner)
```
ellistarn, line 584: Instead of this, I would just use the scoped variable `provider`, mutate it, and then call ProvisionerWithProvider. Other tests follow this pattern.

orirawlings: I think there is actually a bug with the other tests. Currently the defaulting of some of the provider fields are handled in the `BeforeEach`. Those defaults are only persisted into a copy of the provider that is deserialized into the provisioned. As a result when we do `ProviderFromProvisioner` again some of the defaulted values are reverted back to being unset. This results in validation errors that are unrelated to the specific test case. I think this is unnoticed because all existing test cases that mutate the provider expect an unqualified validation error.

For example, if I have the following tests, the second one fails due to `missing field(s): spec.provider.securityGroupSelector, spec.provider.subnetSelector`:
```
		FIt("should validate", func() {
			Expect(provisioner.Validate(ctx)).To(Succeed())
		})
		FIt("should validate", func() {
			Expect(ProvisionerWithProvider(provisioner, provider).Validate(ctx)).To(Succeed())
		})
```

Because defaulting the provider fields is encapsulated inside the `(p *Provisioner) SetDefaults(ctx context.Context)`, I'm not sure of a cleaner way to default the provider fields. I'm open to other ideas though.

ellistarn: Should we call SetDefaults() inside of ProvisionerWithProvider to enforce this? That will ensure the defaults are set for the test, but won't override anything if we want to set something else. WDYT?

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
```diff
@@ -18,10 +18,19 @@ import (
...
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	DefaultMetadataOptionsHTTPEndpoint            = ec2.LaunchTemplateInstanceMetadataEndpointStateEnabled
+	DefaultMetadataOptionsHTTPProtocolIPv6        = ec2.LaunchTemplateInstanceMetadataProtocolIpv6Disabled
+	DefaultMetadataOptionsHTTPPutResponseHopLimit = 2
+	DefaultMetadataOptionsHTTPTokens              = ec2.LaunchTemplateHttpTokensStateRequired
```
suket22, line 31: I think we should default to Optional. Having IMDSv2 be required is the right thing to be doing long-term but I don't think every application in the wild is there yet and forcing them to always specify metadata in the providerSpec might be too heavy-handed

suket22 (commented):
Most of it LGTM. Thanks for making the changes and testing it  👍

pkg/cloudprovider/aws/suite_test.go
orirawlings: I had considered that originally, but some of the existing tests are trying to `nil` out fields to ensure they'd be considered invalid. If we try to default them again we'll override the intention of the test and cause the test to fail.

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
suket22: Alright let's wait for @bwagner5  to ping here before you change it again :)

suket22: Sounds good, let's make sure to call that out in the release notes since we've got the launch templates are currently configured with v1v2. Thanks for chiming in!

pkg/cloudprovider/aws/suite_test.go
orirawlings: @ellistarn I was just circling back on this. Let me know if you want me to do anything further here. I think that without expanding the API of the `v1alpha1.AWS` struct the only way to obtain a provider that's been defaulted to a valid configuration is to pull it out of a provisioner that's been defaulted. I can certainly adjust the approach here and bind new functions to the `v1alpha1.AWS` struct but I'm not sure if the intent was to keep that API surface-area minimal or not.

If the current approach is suitable given existing constraints, it would be cool if we can get this integrated. 👍

ellistarn: I'll bias for merging since it's test code. I'll clean it up later if it bothers me. Thanks for your contribution!

ellistarn (approved):

PR #991, opened 2021-12-14
Set karpenter as user agent for the rest client

olemarkus:
**1. Issue, if available:**

The default user agent for rest clients is the binary name, in this case "controller".
This PR will instead make karpenter show as "karpenter" in access/audit logs, as field owner etc. 

- [x] No

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -65,6 +65,7 @@ func main() {
 
 	config := controllerruntime.GetConfigOrDie()
 	config.RateLimiter = flowcontrol.NewTokenBucketRateLimiter(float32(opts.KubeClientQPS), opts.KubeClientBurst)
+	config.UserAgent = "karpenter"
 	clientSet := kubernetes.NewForConfigOrDie(config)
 
 	// Set up logger and watch for changes to log level


cmd/controller/main.go
```diff
@@ -65,6 +65,7 @@ func main() {
 
 	config := controllerruntime.GetConfigOrDie()
 	config.RateLimiter = flowcontrol.NewTokenBucketRateLimiter(float32(opts.KubeClientQPS), opts.KubeClientBurst)
+	config.UserAgent = "karpenter"
```
bwagner5, line 68: thoughts on including the version as well? 

```
config.UserAgent = fmt.Sprintf("karpenter-%s", project.Version)
```

olemarkus: If karpenter-A sets some field and then version change to B, karpenter-B will have a conflict with itself.

See https://kubernetes.io/docs/reference/using-api/server-side-apply/

bwagner5: o wow, I did not realize server side apply used user-agent... https://kubernetes.io/docs/reference/using-api/server-side-apply/#:~:text=For%20other%20updates%2C%20its%20default%20is%20computed%20from%20the%20user%2Dagent.

bwagner5 (approved):
lgtm

PR #999, opened 2021-12-15
Allow kOps to set labels required for cluster operations

olemarkus:
**1. Issue, if available:**

**2. Description of changes:*
kOps require two labels set on Nodes: `kops.k8s.io/instancegroup` and `kops.k8s.io/gpu`.
The latter is especially important as any pod using the `nvidia` runtimeclass will get that label as nodeSelector.

To make this more future proof, I allowed the entire `kops.k8s.io` domain as there will not be a case where these labels are restricted from Karpenters point of view.

- [x] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/provisioner_validation.go
+++ b/pkg/apis/provisioning/v1alpha5/provisioner_validation.go
@@ -90,6 +90,9 @@ func (c *Constraints) validateLabels() (errs *apis.FieldError) {
 
 func IsRestrictedLabelDomain(key string) bool {
 	labelDomain := getLabelDomain(key)
+	if AllowedLabelDomains.Has(labelDomain) {
+		return false
+	}
 	for restrictedLabelDomain := range RestrictedLabelDomains {
 		if strings.HasSuffix(labelDomain, restrictedLabelDomain) {
 			return true
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -26,6 +26,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
 	"github.com/aws/karpenter/pkg/utils/functional"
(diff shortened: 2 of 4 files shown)


pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -90,6 +90,9 @@ func (c *Constraints) validateLabels() (errs *apis.FieldError) {
 
 func IsRestrictedLabelDomain(key string) bool {
 	labelDomain := getLabelDomain(key)
+	if labelDomain == "kops.k8s.io" {
```
ellistarn, line 93: Can you implement something similar to RestrictedLabelDomains (e.g. AllowedLabelDomains) and generalize this bit of code?

olemarkus: That I can.

olemarkus:
No. As far as I know, the entire k8s.io domain is banned, including subdomains. But these labels make no sense to set on kubelet, and kOps does not support setting labels as kubelet arguments. So for these we should be fine.

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
olemarkus: Take a look if this is better now.

ellistarn:
> No. As far as I know, the entire k8s.io domain is banned, including subdomains. But these labels make no sense to set on kubelet, and kOps does not support setting labels as kubelet arguments. So for these we should be fine.

Currently we pass all labels to the kubelet. This is to avoid an issue where if karpenter crashes right after launching the instance, but doesn't get a chance to create the node, the labels will still be applied. 

I can potentially see moving away from applying *all* labels in the kubelet args and instead moving towards a model where
1. The karpenter.sh/provisioner-name is set in kubelet args (for safety)
2. All labels are created/patched onto the node by Karpenter

Currently, karpenter only does this with a create call, but during periods of high congestion (e.g. fast node boot and kubeclient throttling), we could support this with a mechanism like CreateOrPatch.

ellistarn (commented):
Change LGTM other than the LT generation bit I pointed out.

olemarkus:
As mentioned on Slack, restricted labels should never be set by components other than those who own the respective keys. That is why they are restricted.

For now, I suggest filtering out allowed labels in the launch template. It is a very curious edge case where EKS users would set flags that are owned by kOps, so I think the risk here is very low.

ellistarn (approved):

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -250,16 +251,16 @@ exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
...
-			if !first {
-				nodeLabelArgs.WriteString(",")
+			if v1alpha5.AllowedLabelDomains.Has(k) {
+				continue
 			}
-			first = false
-			nodeLabelArgs.WriteString(fmt.Sprintf("%s=%v", k, nodeLabels[k]))
+			labelStrings = append(labelStrings, fmt.Sprintf("%s=%v", k, nodeLabels[k]))
 		}
+		nodeLabelArgs.WriteString(strings.Join(labelStrings, ","))
```
ellistarn, line 263: Thoughts on simplifying to:
```
fmt.Sprintf("--node-labels=%s", strings.Join(labelStrings, ","))
```

ellistarn (dismissed):

ellistarn (approved):
Nice work!

Tyler887:
> No. As far as I know, the entire k8s.io domain is banned, including subdomains. But these labels make no sense to set on kubelet, and kOps does not support setting labels as kubelet arguments. So for these we should be fine.

Wrong. It moved to **another domain**, as I checked the thing and it took me somewhere else. Try going into it and you should be redirected to another domain.

olemarkus:
@Tyler887 I suspect you are not talking about label domains/prefixes here

PR #1001, opened 2021-12-15
Refactoring Helm labels

nlamirault:
**1. Issue, if available:**

Refactoring Helm labels like in [Helm documentation](https://helm.sh/docs/chart_best_practices/labels/)

- [x] No

### DIFF
--- a/charts/karpenter/Chart.yaml
+++ b/charts/karpenter/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 name: karpenter
 description: A Helm chart for https://github.com/aws/karpenter/.
 type: application
-version: 0.5.2
+version: 0.5.3
--- a/charts/karpenter/Chart.yaml
+++ b/charts/karpenter/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 name: karpenter
 description: A Helm chart for https://github.com/aws/karpenter/.
 type: application
-version: 0.5.3
+version: 0.5.2
--- a/charts/karpenter/templates/100-config-logging.yaml
+++ b/charts/karpenter/templates/100-config-logging.yaml
@@ -4,7 +4,7 @@ metadata:
   name: config-logging
   namespace: {{ .Release.Namespace }}
   labels:
-    app.kubernetes.io/part-of: karpenter
+    {{- include "karpenter.labels" . | indent 4 }}
 data:
   # https://github.com/uber-go/zap/blob/aa3e73ec0896f8b066ddf668597a02f89628ee50/config.go
   zap-logger-config: |
--- a/charts/karpenter/templates/_helpers.tpl
+++ b/charts/karpenter/templates/_helpers.tpl
@@ -1,4 +1,3 @@
-{{/* vim: set filetype=mustache: */}}
 {{/*
 Expand the name of the chart.
 */}}
--- a/charts/karpenter/templates/_helpers.tpl
+++ b/charts/karpenter/templates/_helpers.tpl
@@ -1,0 +1,68 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "karpenter.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "karpenter.fullname" -}}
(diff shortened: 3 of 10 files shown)


agrrh:
I've just canceled my PR with Deployment/Service labels, your work seems much more flexible :+1:

charts/karpenter/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v2
 name: karpenter
 description: A Helm chart for https://github.com/aws/karpenter/.
 type: application
-version: 0.5.2
+version: 0.5.3
```
bwagner5, line 5: can you revert this back to 0.5.2? We'll increment the version during our release process.

charts/karpenter/templates/_helpers.tpl
```diff
@@ -0,0 +1,68 @@
...
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Generate basic labels
+*/}}
+{{- define "karpenter.labels" }}
+helm.sh/chart: {{ include "karpenter.chart" . }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/component: node-provisionning-controller
```
bwagner5, line 40: can we just call this `karpenter`?

charts/karpenter/templates/_helpers.tpl
```diff
@@ -0,0 +1,68 @@
...
+app.kubernetes.io/part-of: {{ template "karpenter.name" . }}
+{{- include "karpenter.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
```
bwagner5, line 44: We'd need to keep chart and the karpenter application versions in-sync for this to be accurate. We are currently doing that, but can you use the image version instead just in case that deviates for any reason?

charts/karpenter/templates/controller/deployment.yaml
```diff
@@ -15,20 +17,24 @@ kind: Deployment
...
     type: Recreate
   selector:
     matchLabels:
       karpenter: controller
+      {{- include "karpenter.selectorLabels" . | indent 6 }}
   template:
     metadata:
       labels:
         karpenter: controller
+        {{- include "karpenter.labels" . | indent 8 }}
```
ellistarn, line 34: Shouldn't this match line 29?

nlamirault: @ellistarn line 34 include all labels, instead the line 29 only labels used for selection. I made a mistake ?

charts/karpenter/templates/100-config-logging.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: config-logging
   namespace: {{ .Release.Namespace }}
   labels:
-    app.kubernetes.io/part-of: karpenter
```
ellistarn, line 7: IIUC this is required by our runtime framework to discover the logging configuration.

charts/karpenter/templates/_helpers.tpl
```diff
@@ -0,0 +1,68 @@
+{{/* vim: set filetype=mustache: */}}
```
ellistarn, line 1: Is it necessary to have this implementation? Can you briefly outline the features that this gives us? I worry that it will be challenging to maintain and perhaps unnecessary complexity.

charts/karpenter/templates/100-config-logging.yaml
nlamirault: helper have : `app.kubernetes.io/part-of: {{ template "karpenter.name" . }}`
Is it ok for your runtime framework ? or i must change to `app.kubernetes.io/part-of: karpenter` @ellistarn

ellistarn: I would use:
> app.kubernetes.io/part-of: karpenter

We shouldn't be parameterizing this value.

bwagner5 (changes requested):
other than the serviceAccountName change, I think this looks good.  

I did helm template diff between the main branch and the PR branch and it shows only added labels, so all seems good there!

charts/karpenter/templates/serviceaccount.yaml
```diff
@@ -2,8 +2,10 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ .Values.serviceAccount.name }}
+  name: {{ template "karpenter.fullname" . }}
```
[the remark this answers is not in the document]
nlamirault: Done. Thanks

bwagner5 (approved):
lgtm! Thanks!

PR #1006, opened 2021-12-16
Make it possible to set node naming convention

### LINKED ISSUE #927: Karpenter creates Node objects with the wrong name
## Version

**Karpenter**: v0.51.1

**Kubernetes**: v1.23.0-rc.1

## Expected Behavior
Karpenter creates node objects with the correct node name (e.g i-0df38da50804be2c0 )

## Actual Behavior
Karpeneter always create objects using the private DNS name of instances (e.g ip-172-20-122-63.eu-central-1.compute.internal)

* Launch a cluster using external AWS Cloud Controller Manager master branch
* Ensure Resource Based Names is used
* Do not set `--hostname-override` on the kubelet, letting it default to `os.hostname`

Karpenter will then assume the node name is ip-x-x-x-x.<region>.compute.internal, but the actual node name will be i-xxxxxxx

Note that the node names should not be FQDN, only the hostname.
This change was implemented in cloud-provider-AWS to support things like IPv6-native subnets and will be a part of CCM 1.23.

I am not sure how Karpenter should know the correct Node name up front. The easiest implementation would probably be a CLI flag on the controller.

olemarkus:
As of k8s 1.23, CCM supports both instance ID (resource name) and private dns (ip name) based naming.
It is not possible for karpenter to determine when which convention is used, and this may also depend on the installer used.
Therefore this is added as a CLI argument.

**1. Issue, if available:**
Fixes #927

This adds a CLI flag for the node naming convention.
This is AWS specifc, so I added an "AWS" prefix. I think this is the first "global" cloud provider setting, so I was not sure how you wanted to name this.

I think a CLI arg for this is the only way of fixing this issue. Trying to parse k8s version and whether or not the external CCM is used is a bit fragile. It is also not given that the external CCM is the "official" one.

- [X] No

### DIFF
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -33,6 +33,7 @@ import (
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
+	"github.com/aws/karpenter/pkg/utils/injection"
 )
 
 type InstanceProvider struct {
@@ -73,9 +74,8 @@ func (p *InstanceProvider) Create(ctx context.Context, constraints *v1alpha1.Con
 			aws.StringValue(instance.Placement.AvailabilityZone),
 			getCapacityType(instance),
 		)
-
 		// Convert Instance to Node
-		node, err := p.instanceToNode(instance, instanceTypes)
+		node, err := p.instanceToNode(ctx, instance, instanceTypes)
 		if err != nil {
 			logging.FromContext(ctx).Errorf("creating Node from an EC2 Instance: %s", err.Error())
 			continue
@@ -229,12 +229,16 @@ func (p *InstanceProvider) getInstances(ctx context.Context, ids []*string) ([]*
 	return instances, err
 }
 
-func (p *InstanceProvider) instanceToNode(instance *ec2.Instance, instanceTypes []cloudprovider.InstanceType) (*v1.Node, error) {
+func (p *InstanceProvider) instanceToNode(ctx context.Context, instance *ec2.Instance, instanceTypes []cloudprovider.InstanceType) (*v1.Node, error) {
 	for _, instanceType := range instanceTypes {
 		if instanceType.Name() == aws.StringValue(instance.InstanceType) {
+			nodeName := strings.ToLower(aws.StringValue(instance.PrivateDnsName))
+			if injection.GetOptions(ctx).AWSNodeNameConvention == "resource-name" {
+				nodeName = aws.StringValue(instance.InstanceId)
+			}
 			return &v1.Node{
 				ObjectMeta: metav1.ObjectMeta{
-					Name: aws.StringValue(instance.PrivateDnsName),
+					Name: nodeName,
 					Labels: map[string]string{
 						v1.LabelTopologyZone:       aws.StringValue(instance.Placement.AvailabilityZone),
 						v1.LabelInstanceTypeStable: aws.StringValue(instance.InstanceType),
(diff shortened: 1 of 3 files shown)


ellistarn (commented):
This is only supported by the external CCM, right? I'd love to get to a point where Karpenter only runs in `ip-name` mode.

pkg/cloudprovider/aws/instance.go
```diff
@@ -229,12 +229,18 @@ func (p *InstanceProvider) getInstances(ctx context.Context, ids []*string) ([]*
...
 
-func (p *InstanceProvider) instanceToNode(instance *ec2.Instance, instanceTypes []cloudprovider.InstanceType) (*v1.Node, error) {
+func (p *InstanceProvider) instanceToNode(ctx context.Context, instance *ec2.Instance, instanceTypes []cloudprovider.InstanceType) (*v1.Node, error) {
 	for _, instanceType := range instanceTypes {
 		if instanceType.Name() == aws.StringValue(instance.InstanceType) {
+			var nodeName string
+			if injection.GetOptions(ctx).AWSNodeNameConvention == "ip-name" {
+				nodeName = aws.StringValue(instance.PrivateDnsName)
+			} else {
+				nodeName = *instance.InstanceId
```
ellistarn, line 239: nit: aws.StringValue() helps us avoid NPEs.

pkg/cloudprovider/aws/instance.go
```diff
@@ -229,12 +229,18 @@ func (p *InstanceProvider) getInstances(ctx context.Context, ids []*string) ([]*
 	return instances, err
 }
 
-func (p *InstanceProvider) instanceToNode(instance *ec2.Instance, instanceTypes []cloudprovider.InstanceType) (*v1.Node, error) {
+func (p *InstanceProvider) instanceToNode(ctx context.Context, instance *ec2.Instance, instanceTypes []cloudprovider.InstanceType) (*v1.Node, error) {
 	for _, instanceType := range instanceTypes {
 		if instanceType.Name() == aws.StringValue(instance.InstanceType) {
+			var nodeName string
```
ellistarn, line 235: Consider collapsing to 
```
nodeName := ...
if ... {
  nodeName = ...
}
```

pkg/utils/options/options.go
```diff
@@ -41,20 +42,24 @@ func MustParse() Options {
...
+	KubeClientBurst       int
+	AWSNodeNameConvention string
 }
 
 func (o Options) Validate() (err error) {
 	err = multierr.Append(err, o.validateEndpoint())
 	if o.ClusterName == "" {
 		err = multierr.Append(err, fmt.Errorf("CLUSTER_NAME is required"))
 	}
+	if o.AWSNodeNameConvention != "ip-name" && o.AWSNodeNameConvention != "resource-name" {
```
ellistarn, line 60: Can we default to ip-name, for now? I'd hate for folks to have to pass this flag in all cases.

olemarkus: `ip-name` is the default when parsing the flags. There shouldn't be a need to set this explicitly when running through `main()`. At least that was my intention.

pkg/utils/options/options.go
```diff
@@ -32,6 +32,7 @@ func MustParse() Options {
 	flag.IntVar(&opts.WebhookPort, "port", 8443, "The port the webhook endpoint binds to for validation and mutation of resources")
 	flag.IntVar(&opts.KubeClientQPS, "kube-client-qps", env.WithDefaultInt("KUBE_CLIENT_QPS", 200), "The smoothed rate of qps to kube-apiserver")
 	flag.IntVar(&opts.KubeClientBurst, "kube-client-burst", env.WithDefaultInt("KUBE_CLIENT_BURST", 300), "The maximum allowed burst of queries to the kube-apiserver")
+	flag.StringVar(&opts.AWSNodeNameConvention, "aws-node-name-convention", env.WithDefaultString("AWs_NODE_NAME_CONVENTION", "ip-name"), "The node naming convention used by the AWS cloud provider")
```
ellistarn, line 35:
```suggestion
	flag.StringVar(&opts.AWSNodeNameConvention, "aws-node-name-convention", env.WithDefaultString("AWS_NODE_NAME_CONVENTION", "ip-name"), "The node naming convention used by the AWS cloud provider")
```

olemarkus: Good catch.

ellistarn (dismissed):

ellistarn (approved):

PR #1013, opened 2021-12-16
Add Ability to Specify Kubelet Configuration, specifically clusterDNS

### LINKED ISSUE #928: Possibility to set dns-cluster-ip for Karpenter provider
It could be very nice to add a way to overload the dns-cluster-ip on the LaunchTemplate when you use a custom svc IP.

I'm actually using 10.10.0.0 as my svc IP and i can't have DNS working because my instance is launched with 172.20.0.10 as DNS IP.
So every of my pods launched in instance provided by Karpenter are on the wrong DNS IP.

Maybe you already have something that i haven't found yet in the documentation .. 

Thanks.
BR Guillaume.

mikesir87:
**1. Issue, if available:**
Resolves #928

Adds the ability to specify kubeletConfiguration, specifically the `clusterDNS` setting.

- [ ] No


pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -34,6 +34,9 @@ type Constraints struct {
 	Taints Taints `json:"taints,omitempty"`
 	// Requirements are layered with Labels and applied to every node.
 	Requirements Requirements `json:"requirements,omitempty"`
+	// KubeletArgs define args to be used when configuring kubelet on provisioned nodes
```
ellistarn, line 37:
```suggestion
	// KubeletArgs are passed to the kubelet when provisioning nodes
```

pkg/apis/provisioning/v1alpha5/kubelet_args.go
```diff
@@ -0,0 +1,22 @@
...
+limitations under the License.
+*/
+
+package v1alpha5
+
+// KubeletArgs define args to be used when configuring kubelet on provisioned nodes
+type KubeletArgs struct {
+	// The IP address for in-cluster DNS resolution
+	//+optional
+	ClusterDNSIP string `json:"clusterDnsIp"`
```
ellistarn, line 21: A couple approaches here

1. `type KubeletArgs []string`
2. `type KubeletArgs map[string]string`
3. `type KubeletArgs map[string]interface{}` <-- not sure if possible with crd generation
4. `type KubeletArgs struct { ...expandedovertime... }`
5.  Use this directly https://github.com/kubernetes/kubelet/blob/master/config/v1beta1/types.go#L88

I wonder how far we would get with option 1. It follows from ContainerSpec, which I like. It does mean that we have to do a bit of parsing for bottlerocket, but I'm not sure that this will be too hard.

I also like option 5, though it relies on regularly updating the kubelet dependency. It's the most future proof, but may expand our CRD significantly. We may be able to hack the CRD generation so that it's an "empty" to avoid too much bloat. It doesn't seem crazy to me that a node controller would be aware of a kubelet config though.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -34,6 +34,9 @@ type Constraints struct {
...
+	//+optional
+	KubeletArgs KubeletArgs `json:"kubeletArgs,omitempty"`
```
ellistarn, line 39: I might just call this `kubelet`, since args is implicit to the fact that you're configuring it. `spec.kubelet.clusterDnsIP` reads quite cleanly.

pkg/apis/provisioning/v1alpha5/kubelet_args.go
mikesir87: I talked with @suket22 about this and we both felt it was important to keep this strongly typed. Otherwise, we're going to be doing lots of specific key checks (like this DNS cluster IP setting) because it doesn't go directly into the normal arg string. And because some require specific handling, we wanted to introduce only the ones we're currently supporting at the time (why we didn't want to go with option 5).

pkg/apis/provisioning/v1alpha5/constraints.go
mikesir87: As a consumer of the API spec, I feel the name `spec.kubelet` hints there might be other config beyond just the args. So, that would make me think I'd see something like `spec.kubelet.args`.

pkg/apis/provisioning/v1alpha5/kubelet_args.go
ellistarn: If we're not going with option 5, can we at least go with an explicit subset of it? i.e., same key names, json annotations, etc. It should be as if we're using it, but we've disallowed most of the keys.

mikesir87: And done 👍

pkg/apis/provisioning/v1alpha5/constraints.go
ellistarn: I guess if we want to make this explicit, should we make it `kubeletConfiguration`: https://github.com/kubernetes/kubelet/blob/master/config/v1beta1/types.go#L88

mikesir87: I like that. Just pushed it 👍

ellistarn (commented):
LGTM. I'm a bit tired -- can you get an approval from Suket just in case I'm missing anything?

ellistarn:
Nice work, btw!

mikesir87:
@suket22 Can you take a look? Thanks!

mikesir87:
Rebased on top of latest to resolve conflict

ellistarn (dismissed):

mikesir87:
Just noticed that there was a name mismatch between the JSON object and the Go type. So, fixing that (any chance that could be caught in a lint check?).

And @akestner - I saw the title change. I'm going to tweak it just a tiny bit because the names ended up being adjusted as we did the work and I neglected to go back and update the original description of changes. I've rectified my transgression :smile:

ellistarn (approved):

justinsnair:
@mikesir87 

I have installed helm-chart v 0.5.5 which should have these changes however when trying to create a provisioner I am getting this error:

error: error validating "provisioner.yaml": error validating data: ValidationError(Provisioner.spec): unknown field "kubeletConfiguration" in sh.karpenter.v1alpha5.Provisioner.spec; if you choose to ignore these errors, turn validation off with --validate=false

Here is my provisioner:
`apiVersion: karpenter.sh/v1alpha5
kind: Provisioner
metadata:
  name: default-provisioner
spec:
  kubeletConfiguration: 
    clusterDNS: ["redacted"]
  requirements:
      operator: In
      values: ["spot"]
  limits:
    resources:
      cpu: 500m
  provider:
    instanceProfile: "redacted"
  ttlSecondsAfterEmpty: 30`

Am i missing something?

mikesir87:
@justinsnair Did you install the chart as an upgrade to a previous version? If so, you'll need to update the CRD as Helm doesn't do CRD updates.

justinsnair:
This was it.  I knew this but totally slipped my mind.. Thanks for the reminder!

PR #1031, opened 2021-12-20
fix CloudProvider metric

cjerad:
**1. Issue, if available:**
None

Previously, the latency recorded for the `CloudProvider.Create()` method may have missed some latency.  Now the latency "start time" is set correctly.

- [x] No

### DIFF
--- a/pkg/cloudprovider/metrics/cloudprovider.go
+++ b/pkg/cloudprovider/metrics/cloudprovider.go
@@ -66,19 +66,9 @@ func Decorate(cloudProvider cloudprovider.CloudProvider) cloudprovider.CloudProv
 	return &decorator{cloudProvider}
 }
 
-func (d *decorator) Create(ctx context.Context, constraints *v1alpha5.Constraints, instanceTypes []cloudprovider.InstanceType, quantity int, callback func(*v1.Node) error) <-chan error {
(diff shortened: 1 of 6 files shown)


pkg/cloudprovider/metrics/cloudprovider.go
```diff
@@ -67,9 +67,10 @@ func Decorate(cloudProvider cloudprovider.CloudProvider) cloudprovider.CloudProv
 }
 
 func (d *decorator) Create(ctx context.Context, constraints *v1alpha5.Constraints, instanceTypes []cloudprovider.InstanceType, quantity int, callback func(*v1.Node) error) <-chan error {
+	recordLatency := metrics.Measure(methodDurationHistogramVec.WithLabelValues(getControllerName(ctx), "Create", d.Name()))
```
ellistarn, line 70: The cloud provider was originally written to be async (i.e. return a chan) in order to support batching on the cloud provider side. Instead, batching was built into the arguments (i.e. quantity). 

I don't think it would be crazy to change the cloud provider to return an error instead of a chan(error), which would simplify the complexity in the metrics implementation.

felix-zhe-huang: I have a noob question. Does removing the async design prohibit us from performing time consuming callbacks when a node is launched? For example, can i still wait until the node is ready in the callback function?

ellistarn (approved):

Issue #954, opened 2021-12-09
Karpenter makes wrong assumptions about CNI and kubelet max pods


olemarkus:
## Expected Behavior

Karpenter either takes in max pods as a configurable constraint or does not take it into consideration when calculating the amount of pods needed.

## Actual Behavior
When small instances are provisioned, it only assigns it a handful of pods, way below what the node can actually handle.
In other cases, it creates really huge instances and assigns it hundreds of points, but the kubelet only accepts 110 (the default maxPods kubelet setting) of them.

The calculations assume that the AWS VPC CNI is used and that prefix mode is not enabled, which will be an increasingly rare configuration.

## Steps to Reproduce the Problem

Spin up 500-600 pods with cpu.request=100m on any cluster not running AWS VPC CNI.

ellistarn:
This is on our radar and we're working on finding a solution to this as part of cohesive improvements to launch template flexibility.

bdwyertech:
We could probably tap into the same lookup table I used for Pod ENI (might even be same value)

ellistarn:
I'd really like to get to a world where we limit max pods by cpu/memory, and choose acceptable density/overhead values. The fact that <1.8 CNI is used widely is an inhibitor to this. I can see potentially supporting higher density by default, and including a flag that forces a more restrictive value. 

@suket22 would love your opinion on this.

olemarkus:
In kops, most people use other CNIs. MaxPods is a per instance group configuration one can set if one wants to. Default is 110 and from my own experience one rarely has that high density. I am also not sure I ever want such high density as a significant part of capacity would be gone if the instance fails. 

Should someone use vpc cni, then same logic as Karpenter is used.

ellistarn:
One idea to explore, use "IntOrString" to support things like `spec.density: 30`, `spec.density: vpc-cni`, `spec.density: default` (some strategy like cpu/memory)

olemarkus:
Topology spread constraints won't do anything to avoid say 1/3 of capacity to go away if a single instance dies. 1/3 capacity going away during AZ failure may be something one gets away with, but not on a regular basis. I think a lot of cluster admins aim to have a certain amount of Nodes in the cluster and adjusts instance type thereafter. 

If karpenter intends to control max pods, it must also expose what it intends it to be so installers can configure accordingly. E.g an annotation on the Node object.

mikesir87:
[...]

PR #1032, opened 2021-12-21
Add flexibility in pod density for non-VPC CNI nodes

### LINKED ISSUE #445: Specify max pods on provisioner resource
The max pod node constraint uses the default setting for the Bottlerocket AMI. This value is not accurate when configuring aws-vpc [custom CNI](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html) where the max pods is reduced because one network interface is no longer available for pods.

As well when using other [CNI plugins](https://docs.aws.amazon.com/eks/latest/userguide/alternate-cni-plugins.html), the default max pods can be increased as max pods is no longer constrained by the available network interfaces.

Another interesting note about max pods per node is the 110 default that is often used when not limited by CNI is a bit arbitrary, and this value can be safely 250+ on larger instance types ([reference](https://github.com/kubernetes/kubernetes/issues/23349))

mikesir87:
**1. Issue, if available:**
Resolves #445 and associated with #954

This adds the ability to specify `spec.kubeletConfiguration.maxPods`. 

While the kubelet type only supports non-negative values, there are cases in which other providers (such as the AL2 EKS-optimized AMI) use other flags. In the AL2 case, it uses `--use-max-pods=false` to indicate it shouldn't refer to a max ENI listing for the number of pods (which means it's making the assumption you're using the AWS VPC CNI). So, if a negative number is provided, it's expected that the provider will not use the negative number, but indicate no value should be specified and defaults used. In the case of the AL2 provider, we'd add the `--use-max-pods=false` flag.

As we only have the AL2 provider right now, I updated the user data generation to specify either the `--use-max-pods=false` flag or add a `--max-pods` flag to the `--extra-kubelet-args`.

- [ ] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
+++ b/pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
@@ -22,10 +22,4 @@ type KubeletConfiguration struct {
 	// Note that not all providers may use all addresses.
 	//+optional
 	ClusterDNS []string `json:"clusterDNS,omitempty"`
-	// MaxPods is the maximum number of Pods that can run on the Kubelet. While the value must be
-	// a non-negative integer, some providers might use separate flags to indicate whether a value
-	// will be specified. Therefore, if a negative value is provided, it indicates this field should
-	// not be specified on the underlying kubelet and defaults should be used.
-	//+optional
-	MaxPods int32 `json:"maxPods,omitempty"`
 }
--- a/pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
+++ b/pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
@@ -22,4 +22,10 @@ type KubeletConfiguration struct {
 	// Note that not all providers may use all addresses.
 	//+optional
 	ClusterDNS []string `json:"clusterDNS,omitempty"`
+	// MaxPods is the maximum number of Pods that can run on the Kubelet. While the value must be
+	// a non-negative integer, some providers might use separate flags to indicate whether a value
+	// will be specified. Therefore, if a negative value is provided, it indicates this field should
+	// not be specified on the underlying kubelet and defaults should be used.
+	//+optional
+	MaxPods int32 `json:"maxPods,omitempty"`
 }
--- a/pkg/cloudprovider/aws/instancetype.go
+++ b/pkg/cloudprovider/aws/instancetype.go
@@ -16,6 +16,7 @@ package aws
 
 import (
 	"fmt"
+	"github.com/aws/karpenter/pkg/utils/options"
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/vpc"
 	"github.com/aws/aws-sdk-go/aws"
@@ -34,6 +35,7 @@ const EC2VMAvailableMemoryFactor = .925
 type InstanceType struct {
 	ec2.InstanceTypeInfo
 	AvailableOfferings []cloudprovider.Offering
+	Options            options.Options
 }
 
 func (i *InstanceType) Name() string {
@@ -70,6 +72,10 @@ func (i *InstanceType) Memory() *resource.Quantity {
 }
 
 func (i *InstanceType) Pods() *resource.Quantity {
+	if !i.Options.AWSENILimitedPodDensity {
+		return resources.Quantity("110")
+	}
+
 	// The number of pods per node is calculated using the formula:
 	// max number of ENIs * (IPv4 Addresses per ENI -1) + 2
 	// https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt#L20
--- a/pkg/cloudprovider/aws/instancetype.go
+++ b/pkg/cloudprovider/aws/instancetype.go
@@ -25,6 +25,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"knative.dev/pkg/ptr"
 )
 
 // EC2VMAvailableMemoryFactor assumes the EC2 VM will consume <7.25% of the memory of a given machine
@@ -71,7 +72,7 @@ func (i *InstanceType) Memory() *resource.Quantity {
 
 func (i *InstanceType) Pods() *resource.Quantity {
 	if i.MaxPods != nil {
-		return resources.Quantity(fmt.Sprint(*i.MaxPods))
+		return resources.Quantity(fmt.Sprint(ptr.Int32Value(i.MaxPods)))
 	}
 	return resources.Quantity(fmt.Sprint(i.eniLimitedPods()))
 }
--- a/pkg/cloudprovider/aws/instancetype.go
+++ b/pkg/cloudprovider/aws/instancetype.go
@@ -16,8 +16,6 @@ package aws
 
 import (
 	"fmt"
-	"github.com/aws/karpenter/pkg/utils/options"
-
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/vpc"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -35,7 +33,7 @@ const EC2VMAvailableMemoryFactor = .925
 type InstanceType struct {
 	ec2.InstanceTypeInfo
 	AvailableOfferings []cloudprovider.Offering
-	Options            options.Options
+	MaxPods            *int32
 }
 
 func (i *InstanceType) Name() string {
@@ -72,14 +70,10 @@ func (i *InstanceType) Memory() *resource.Quantity {
 }
 
 func (i *InstanceType) Pods() *resource.Quantity {
-	if !i.Options.AWSENILimitedPodDensity {
-		return resources.Quantity("110")
+	if i.MaxPods != nil {
+		return resources.Quantity(fmt.Sprint(*i.MaxPods))
 	}
-
-	// The number of pods per node is calculated using the formula:
-	// max number of ENIs * (IPv4 Addresses per ENI -1) + 2
-	// https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt#L20
-	return resources.Quantity(fmt.Sprint(*i.NetworkInfo.MaximumNetworkInterfaces*(*i.NetworkInfo.Ipv4AddressesPerInterface-1) + 2))
+	return resources.Quantity(fmt.Sprint(i.getNumberOfEniBasedPods()))
 }
 
 func (i *InstanceType) AWSPodENI() *resource.Quantity {
@@ -126,7 +120,9 @@ func (i *InstanceType) AWSNeurons() *resource.Quantity {
 }
 
 // Overhead computes overhead for https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#node-allocatable
-// using calculations copied from https://github.com/bottlerocket-os/bottlerocket#kubernetes-settings
+// using calculations copied from https://github.com/bottlerocket-os/bottlerocket#kubernetes-settings.
+// While this doesn't calculate the correct overhead for non-ENI-limited nodes, we're using this approach until further
+// analysis can be performed
 func (i *InstanceType) Overhead() v1.ResourceList {
 	overhead := v1.ResourceList{
 		v1.ResourceCPU: *resource.NewMilliQuantity(
@@ -134,7 +130,7 @@ func (i *InstanceType) Overhead() v1.ResourceList {
 			resource.DecimalSI),
 		v1.ResourceMemory: resource.MustParse(fmt.Sprintf("%dMi",
 			// kube-reserved
-			((11*i.Pods().Value())+255)+
+			((11*i.getNumberOfEniBasedPods())+255)+
 				// system-reserved
 				100+
 				// eviction threshold https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/kubelet/apis/config/v1beta1/defaults_linux.go#L23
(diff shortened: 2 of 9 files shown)


olemarkus:
This doesn't really solve your problem. If you set maxPod on kubelet, Karpenter will still try to over/under allocate.
Look at https://github.com/aws/karpenter/blob/main/pkg/cloudprovider/aws/instance.go#L253 (and the similar call some lines below) and the algorithm it uses here https://github.com/aws/karpenter/blob/12cbb2d3b2a32e5f467749750da779a0164ca12d/pkg/cloudprovider/aws/instancetype.go#L72-L77

This is the bit I mentioned in https://github.com/aws/karpenter/issues/954#issuecomment-998919327 that is actually causing problem.

Using your approach, you need to get the maxPod value into the code above, overriding the logic in `InstanceType.Pods()`

mikesir87:
But doesn't Karpenter only do scheduling if a pod is unscheduled? It's not trying to take the place of the kube-scheduler, right? I'm currently running into the problem that we have underutilized nodes due to this setting being in place. I guess I'm not understanding where/when this is being used.

olemarkus:
It replaces kube-scheduler in the case of Pending pods.
So e.g if you have 20 Pods unscheduled, it will disregard t3.medium because it thinks the Pods cannot fit.  Similarly, if the kubelet is configured with default maxPods of 110, it will think it can fit 700 Pods onto a t3.16xlarge.

mikesir87:
Ok. That's what I thought. The problem I'm having is I'm getting into that Pending state earlier that expected because I can't pack more pods onto a node (because I can't set this flag).

So, to me, it sounds like we need to have both the setting, but then make it available down in this calculation you're referring to. If the maxPod value is set, we should use that. Otherwise, we assume (possibly wrongly) that the AWS VPC CNI is being used and fallback to what value makes sense for that instance type. That make sense?

olemarkus:
Many are not using AMIs that assume VPC CNI (e.g I just use Ubuntu default), so they don't run into your specific part of the problem. But yes, to solve your case, you need both.

olemarkus:
I amended your commit here: https://github.com/aws/karpenter/compare/main...olemarkus:445-specify-max-pods-setting
* The settings propagates also into `Pods`
* Also added support for `--pods-per-core` should anyone want the more dynamic limitation.

mikesir87:
I like your updates @olemarkus and I'll fold them into my PR (if that's cool with you).

olemarkus:
Please do. That was what I hoped for 😊

pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
```diff
@@ -22,4 +22,10 @@ type KubeletConfiguration struct {
 	// Note that not all providers may use all addresses.
 	//+optional
 	ClusterDNS []string `json:"clusterDNS,omitempty"`
+	// MaxPods is the maximum number of Pods that can run on the Kubelet. While the value must be
+	// a non-negative integer, some providers might use separate flags to indicate whether a value
+	// will be specified. Therefore, if a negative value is provided, it indicates this field should
+	// not be specified on the underlying kubelet and defaults should be used.
+	//+optional
+	MaxPods int32 `json:"maxPods,omitempty"`
```
ellistarn, line 30: Can we make this a *int32 so it doesn't default to zero?

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -258,10 +256,34 @@ exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
...
+	}
+	if constraints.KubeletConfiguration.MaxPods < 0 {
+		userData.WriteString(` \
+    --use-max-pods=false`)
 	}
 	return base64.StdEncoding.EncodeToString(userData.Bytes()), nil
 }
 
+func (p *LaunchTemplateProvider) getKubeletExtraArgs(constraints *v1alpha1.Constraints, additionalLabels map[string]string) string {
+	var kubeletArgs bytes.Buffer
```
ellistarn, line 268: Thoughts on just concatting to a string? Not sure I understand the point of using a byte buffer instead.

pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
```diff
@@ -22,4 +22,10 @@ type KubeletConfiguration struct {
 	// Note that not all providers may use all addresses.
 	//+optional
 	ClusterDNS []string `json:"clusterDNS,omitempty"`
+	// MaxPods is the maximum number of Pods that can run on the Kubelet. While the value must be
```
ellistarn, line 25: Thoughts on defaulting this to 110?

ellistarn: Seems like there's a tradeoff for small nodes, since the overhead will be so large. You have to reserve headroom for 110 even if you'll only ever use 1 pod.

Another approach would be to use some function e.g. Max(5%, 110pods). I'm wary of exposing too low level control over max pods, since the kubelet flag assumptions were not built with karpenter in mind. Ideally karpenter would build a simple strategy(s) that fits all use cases.

pkg/cloudprovider/aws/launchtemplate.go
mikesir87: No idea either. It was just the pattern being used before and continued it. 😆

pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
mikesir87: I thought about it, but that is a change from how things are working right now where the default is whatever the EKS optimized image is doing (defaulting to the VPC CNI settings). So, we'd want to have a way to know whether we use that or not. I figured having a default would mean we would need yet another setting to have it fallback to the ENI count approach currently used. That make sense?

pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
mikesir87: Absolutely! Blame that on my Go ignorance 😉

pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
olemarkus: My idea was that `nil` meant "do nothing", i.e use whatever EKS AMI does, which at the moment is assuming non-prefix VPC CNI. EKS Managed node groups using prefix VPC CNI set 110 as default on all instance types with < 30 cores, and a rather curious default of 250 for > 30 cores (lots of very, very small pods on the same node). In other words, outside of a "legacy" VPC CNI configuration, kubelet almost always runs with this set to 110. I don't see why this would be a problem for Karpenter.

pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
olemarkus: One thing to keep in mind is that 0 cannot be a valid value. Setting 0 would be the same as omitting the flag, leaving Kubelet to use its own default. This won't be supported by Karpenter, as Karpenter has now knowledge of what that default may be. So technically it should be fine to say that 0 is the same as EKS AMI behavior.

mikesir87:
Updating this PR comment thread based on conversations on Slack and in the working group... 

We've decided not to introduce any API changes at this point, but add a CLI flag to indicate we are not running in "AWS ENI mode". The flag, `--aws-eni-limited-pod-density` will be a boolean flag and default to `true`. When set to true, the current behavior will run. Otherwise, we will set `maxPod` to 110. We all recognize that this still isn't configurable, but isn't a long-term answer while we perform additional analysis on overhead and how this can/should be solved longer term.

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -34,6 +35,7 @@ const EC2VMAvailableMemoryFactor = .925
 type InstanceType struct {
 	ec2.InstanceTypeInfo
 	AvailableOfferings []cloudprovider.Offering
+	Options            options.Options
```
ellistarn, line 38: Can we unwrap this into MaxPods? Passing options seems like it violates cohesion principles.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -259,6 +259,12 @@ exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
 	nodeTaintsArgs := p.getNodeTaintArgs(constraints)
 	kubeletExtraArgs := strings.Trim(strings.Join([]string{nodeLabelArgs, nodeTaintsArgs.String()}, " "), " ")
 
+	if !injection.GetOptions(ctx).AWSENILimitedPodDensity {
+		userData.WriteString(` \
```
ellistarn, line 263: Do you mind moving towards string concatenation, now that you're in here?

pkg/cloudprovider/aws/instancetype.go
mikesir87: Shall do. I knew it felt bad, but figured I'd get input on how you might want to see it. Shall do.

pkg/cloudprovider/aws/launchtemplate.go
mikesir87: Can do. 👍

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -70,6 +72,10 @@ func (i *InstanceType) Memory() *resource.Quantity {
 }
 
 func (i *InstanceType) Pods() *resource.Quantity {
```
ellistarn, line 74: I think we need to update the Overhead logic to always use the ENILimited MaxPods value, even if not using max pods. We're gonna need a nice verbose comment on this including the fact that we want to make it more accurate XD.

mikesir87: Ok. So... are you proposing we don't make any changes (minus a comment update) in this func then since that's what it's currently doing/using?

ellistarn: Sorry if this is a bit confusing. 

The Pods() function should return 
- 110 if not eni limited
- the limit if eni limited

The Overhead() function should return
- overhead calculated by the eni limited pod count, regardless of the result of Pods().

I think you're going to minimally need a help to extract it out. I think this was outlined in slack, but it's pretty confusing. The idea was to just follow the eksoptimizedami convention. @suket22 may be the best person for this.

mikesir87: Ah! Gotcha... I haven't spent enough time in the scheduling part, so had totally missed the `Overhead()` function. I can take a stab at it and get feedback on it.

mikesir87: Alrighty... ready for another look. I added some comments, but nothing too crazy. The blame can take you back to this PR with more history too. Feel free to add more if you'd like 👍

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -157,3 +159,10 @@ func (i *InstanceType) Overhead() v1.ResourceList {
 	}
 	return overhead
 }
+
+// The number of pods per node is calculated using the formula:
+// max number of ENIs * (IPv4 Addresses per ENI -1) + 2
+// https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt#L20
+func (i *InstanceType) getNumberOfEniBasedPods() int64 {
```
ellistarn, line 166:
```suggestion
func (i *InstanceType) eniLimitedPods() int64 {
```

ellistarn:
LGTM. Nice work. Make sure you get an approval from @suket22

mikesir87:
Thanks @ellistarn for the input and help! Not sure about the failing netlify deploy tasks. But, looks like we're good to go now that the CI builds are passing. Will pass it on the @suket22 👍

ellistarn:
I think you might need a rebase to fix netlify.

mikesir87:
Ok. Thought I had rebased. Anywho... did it again and it looks like we're good to go now 👍

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -70,10 +70,10 @@ func (i *InstanceType) Memory() *resource.Quantity {
 }
 
 func (i *InstanceType) Pods() *resource.Quantity {
-	// The number of pods per node is calculated using the formula:
-	// max number of ENIs * (IPv4 Addresses per ENI -1) + 2
-	// https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt#L20
-	return resources.Quantity(fmt.Sprint(*i.NetworkInfo.MaximumNetworkInterfaces*(*i.NetworkInfo.Ipv4AddressesPerInterface-1) + 2))
+	if i.MaxPods != nil {
+		return resources.Quantity(fmt.Sprint(*i.MaxPods))
```
ellistarn, line 74: consider ptr.Int32Value, which we frequently use as an insurance policy against crashing.

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -85,6 +86,10 @@ func (p *InstanceTypeProvider) Get(ctx context.Context, provider *v1alpha1.AWS)
 			instanceType.AvailableOfferings = offerings
 			result = append(result, instanceType)
 		}
+		if !injection.GetOptions(ctx).AWSENILimitedPodDensity {
+			instanceType.MaxPods = new(int32)
```
ellistarn, line 90: instanceType.MaxPods = ptr.Int32(110)

pkg/cloudprovider/aws/launchtemplate.go
ellistarn: Reminder

pkg/cloudprovider/aws/instancetype.go
mikesir87: I went this route because Quantity needs a string and I saw it used in several of the other places Quantity was being used. May be a chance to loop back and clean all of them up in the future?

pkg/cloudprovider/aws/instancetypes.go
mikesir87: I knew there had to be a better way to do that. Thanks 👍

pkg/cloudprovider/aws/launchtemplate.go
mikesir87: Ah... now I see why it was a bytes.Buffer instead of string concatenation. The userData has to be base64 encoded, which needs a byte[].

ellistarn: Gotcha

pkg/cloudprovider/aws/instancetype.go
mikesir87: Ah... I get what you mean now. Fixing it 👍

ellistarn (approved):
nice work!

Issue #1067, opened 2022-01-01
Karpenter fails to create node object if an instance does not have PrivateDnsName


olemarkus:
## Version

**Karpenter**: master

**Kubernetes**: v1.23.0

## Expected Behavior

Karpenter creates node object.

## Actual Behavior

Karpenter does launch instance, but no node object. Then it let Pods hanging.

The instance eventually does join the cluster just fine.

I am mostly concerned with the Pod being left hanging. 

## Steps to Reproduce the Problem

It just happens randomly on scaleup. I suspect the EC2 API simply doesn't set PrivateDNSName sometimes.

## Resource Specs and Logs

```
2022-01-01T11:03:09.387Z        ERROR   controller.provisioning Could not launch node, launching instances, All attempts fail:
#1: got instance i-0759df6abd68fe02c but PrivateDnsName was not set
#2: got instance i-0759df6abd68fe02c but PrivateDnsName was not set
#3: got instance i-0759df6abd68fe02c but PrivateDnsName was not set     {"commit": "75afd60", "provisioner": "nodes"}
```

ellistarn:
> I suspect the EC2 API simply doesn't set PrivateDNSName sometimes.

- This can happen in rare cases, when it does
- It's imperative that the node has the provisioner label in its kubelet config so karpenter can manage it
- The reconcile loop should fail, resulting in the pods triggering a new reconcile loop
- The node should come online, pods may schedule to it or to another node
- The extra node should be scaled down

> Then it let Pods hanging. 

This doesn't sound right to me. What do you mean by this? For how long?

olemarkus:
There is no reason not to create the node object. In this case, PrivateDNSName is not used for anything interesting. I think I should be able to fix that part.

The reconcile loop doesn't seem to be failing. Karpenter just stops caring about that Pod. The pod is hanging until the Node object is created by Kubelet and eventually schedules by kube-scheduler. This bit worries me a lot more. I have seen pods being left behind in a number of occasions now.

ellistarn:
It's currently used for the node's name.

ellistarn:
I think this should be fixed with https://github.com/aws/karpenter/pull/1070. Let's reopen if we see it after v0.5.4.

davidt-gh:
I have this issue

PR #1068, opened 2022-01-02
Don't fail on missing privatednsname if using RBN

olemarkus:
**1. Issue, if available:**
#1067 

If we are using resource-based naming, we do not need to bail on private DNS name missing.
We still need the describeInstances call for emulating CCM.

- [X] No

### DIFF
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -34,6 +34,7 @@ import (
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
 	"github.com/aws/karpenter/pkg/utils/injection"
+	"github.com/aws/karpenter/pkg/utils/options"
 )
 
 type InstanceProvider struct {
@@ -218,6 +219,10 @@ func (p *InstanceProvider) getInstances(ctx context.Context, ids []*string) ([]*
 	if len(describedInstances) != len(ids) {
 		return nil, fmt.Errorf("expected %d instance(s), but got %d", len(ids), len(describedInstances))
 	}
+	if injection.GetOptions(ctx).GetAWSNodeNameConvention() == options.ResourceName {
+		return describedInstances, nil
+	}
+
 	instances := []*ec2.Instance{}
 	for _, instance := range describedInstances {
 		if len(aws.StringValue(instance.PrivateDnsName)) == 0 {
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/aws/instance.go
```diff
@@ -219,12 +219,16 @@ func (p *InstanceProvider) getInstances(ctx context.Context, ids []*string) ([]*
...
-	for _, instance := range describedInstances {
-		if len(aws.StringValue(instance.PrivateDnsName)) == 0 {
-			err = multierr.Append(err, fmt.Errorf("got instance %s but PrivateDnsName was not set", aws.StringValue(instance.InstanceId)))
-			continue
+	if injection.GetOptions(ctx).AWSNodeNameConvention == "resource-name" {
+		instances = describedInstances
+	} else {
+		for _, instance := range describedInstances {
+			if len(aws.StringValue(instance.PrivateDnsName)) == 0 {
+				err = multierr.Append(err, fmt.Errorf("got instance %s but PrivateDnsName was not set", aws.StringValue(instance.InstanceId)))
```
ellistarn, line 227: We should shortcircuit rather than checking all instances. I'm not sure the multierr is actually useful here:
```
if len(aws.StringValue(instance.PrivateDnsName)) == 0 {
  return nil, fmt.Errorf("got instance %s but PrivateDnsName was not set", aws.StringValue(instance.InstanceId)
}
```

pkg/cloudprovider/aws/instance.go
```diff
@@ -219,12 +219,16 @@ func (p *InstanceProvider) getInstances(ctx context.Context, ids []*string) ([]*
 		return nil, fmt.Errorf("expected %d instance(s), but got %d", len(ids), len(describedInstances))
 	}
 	instances := []*ec2.Instance{}
-	for _, instance := range describedInstances {
-		if len(aws.StringValue(instance.PrivateDnsName)) == 0 {
-			err = multierr.Append(err, fmt.Errorf("got instance %s but PrivateDnsName was not set", aws.StringValue(instance.InstanceId)))
-			continue
+	if injection.GetOptions(ctx).AWSNodeNameConvention == "resource-name" {
```
ellistarn, line 222: Cleaner to shortcircuit instead of if/else.
```
if injection.GetOptions(ctx).AWSNodeNameConvention == "resource-name" {
  return instances, nil
}
```

olemarkus: You wouldn't want the constants in the options package instead?

ellistarn: You're all over it :). Wasn't looking at the code, but you're 100% correct.

ellistarn (approved):

PR #1069, opened 2022-01-02
check-spelling: add dictionaries

jsoref:
**1. Issue, if available:**

This pulls in the [aws](https://github.com/check-spelling/cspell-dicts/blob/master/dictionaries/aws/aws.txt) and [go](https://github.com/check-spelling/cspell-dicts/blob/master/dictionaries/golang/go.txt) dictionaries.

Essentially, if a word is in one of these dictionaries, it will be treated as if it's in the dictionary and thus not flagged. You can see how this impacts expect to get a sense of how much/little coverage changes.

- [x] No


jsoref:
The current version of check-spelling does not like the changes to expect -- because it can't see the changes to the dictionary, so if check-spelling was mandatory, the changes to expect would have to be dropped and then they could be removed after merging.

In the somewhat near future, I hope to move the most of the config out of the workflow and into a normal file. Amongst the many benefits, it'd make updates like this mostly possible w/o the old engine screaming at me.

ellistarn:
> In the somewhat near future, I hope to move the most of the config out of the workflow and into a normal file. Amongst the many benefits, it'd make updates like this mostly possible w/o the old engine screaming at me.

Doesn't Github actions use the workflow file in the given PR?

ellistarn (approved):

jsoref:
For a `push` it would, for a `pull_request` / `pull_request_target`, it does not. I know, it's confusing.

I've started working on migrating the config to JSON this weekend, which should smooth that problem out.

PR #1070, opened 2022-01-02
Rewrote batching logic to no longer rely on a results channel

ellistarn:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/1065

The aim of this PR is to reduce the potential failure modes of the batching logic after reports of "pod stuck". We haven't been able to successfully reproduce this -- however, I have a theory as to how this could happen

- The current logic is count-based, and returns n messages to unblock n pods.
- If somehow the count is too low, one of the pods will remain stuck
- By moving to a channel (context) close mechanism, we remove this possibility.

- [ ] No

### DIFF
--- a/pkg/controllers/provisioning/batcher.go
+++ b/pkg/controllers/provisioning/batcher.go
@@ -21,7 +21,7 @@ import (
 )
 
 var (
-	MaxBatchDuration = time.Second * 10
+	MaxBatchDuration  = time.Second * 10
 	BatchIdleDuration = time.Second * 1
 	// MaxItemsPerBatch limits the number of items we process at one time to avoid using too much memory
 	MaxItemsPerBatch = 2_000
@@ -52,17 +52,20 @@ func NewBatcher(running context.Context) *Batcher {
 // Add an item to the batch, returning the next gate which the caller may block
 // on. The gate is protected by a read-write mutex, and may be modified by
 // Flush(), which makes a new gate.
-func (b *Batcher) Add(ctx context.Context, item interface{}) <-chan struct{} {
-	b.RLock()
-	defer b.RUnlock()
+//
+// In rare scenarios, if a goroutine hangs after enqueueing but before acquiring
+// the gate lock, the batch could be flushed, resulting in the pod waiting on
+// the next gate. This will be flushed on the next batch, and may result in
+// delayed retries for the individual pod if the provisioning loop fails. In
+// practice, this won't be encountered because this time window is O(seconds).
+func (b *Batcher) Add(item interface{}) <-chan struct{} {
 	select {
 	case b.queue <- item:
-		return b.gate.Done()
-	case <-b.gate.Done():
-		return b.gate.Done()
-	case <-ctx.Done():
-		return ctx.Done()
+	case <-b.running.Done():
 	}
+	b.RLock()
+	defer b.RUnlock()
+	return b.gate.Done()
 }
 
 // Flush all goroutines blocking on the current gate and create a new gate.
@@ -73,8 +76,8 @@ func (b *Batcher) Flush() {
 	b.gate, b.flush = context.WithCancel(b.running)
 }
 
-// Wait returns a slice of enqueued items after idle or timeout
-func (b *Batcher) Wait(ctx context.Context) (items []interface{}, window time.Duration) {
+// Wait starts a batching window and returns a slice of items when closed.
+func (b *Batcher) Wait() (items []interface{}, window time.Duration) {
 	// Start the batching window after the first item is received
 	items = append(items, <-b.queue)
 	start := time.Now()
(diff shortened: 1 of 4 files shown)


pkg/controllers/provisioning/provisioner.go
```diff
@@ -89,28 +71,44 @@ type Provisioner struct {
...
-func (p *Provisioner) Add(ctx context.Context, pod *v1.Pod) {
-	select {
-	case p.pods <- pod: // Block until pod is enqueued
-		<-p.wait
-	case <-p.running.Done(): // Leave if closed
+// is responsible for verifying that the pod was scheduled correctly.
+func (p *Provisioner) Add(ctx context.Context, pod *v1.Pod) <-chan struct{} {
+	p.RLock()
+	defer p.RUnlock()
+	p.batch.Add(pod)
```
bwagner5, line 78: it might make sense to have batch.Add return the Done chan. It would save a line and you probably want to monitor the batch completion if you're adding something.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -89,28 +71,44 @@ type Provisioner struct {
...
+
+func (p *Provisioner) run(ctx context.Context) {
+	if ctx.Err() != nil {
+		logging.FromContext(ctx).Info("Stopping provisioner")
+		return
+	}
+	if err := p.provision(ctx); err != nil {
+		logging.FromContext(ctx).Errorf("Provisioning failed, %s", err.Error())
 	}
+	p.run(ctx)
```
bwagner5, line 90: what's happening here? Why not put this in a for-loop?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -89,28 +71,44 @@ type Provisioner struct {
...
 func (p *Provisioner) provision(ctx context.Context) (err error) {
-	// Wait for a batch of pods, release when done
-	pods := p.batch(ctx)
-	defer func() {
-		for i := 0; i < len(pods); i++ {
-			p.wait <- struct{}{}
+	// Batch pods
+	logging.FromContext(ctx).Infof("Waiting for unschedulable pods")
+	items, window := p.batch.Wait(ctx)
+	defer p.batch.Complete()
```
bwagner5, line 97: this is kind of weird, in that you're completing a batch after you've created a NewBatch on line 99.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -89,28 +71,44 @@ type Provisioner struct {
...
+	p.Lock()
+	p.batch = NewBatch(ctx)
```
bwagner5, line 99: I don't quite get why the locks are here?  

It seems like this would wait until nothing is being Added, and then create a new batch, but something could have been added before the Lock was acquired.    Wouldn't it be easier to control this if you used the same batch so that it could be continuous rather than having to move between batches?

pkg/controllers/provisioning/provisioner.go
ellistarn: I had it in one but moved to tail recursion. I can move back.

pkg/controllers/provisioning/provisioner.go
ellistarn: This is an interesting idea. I can move the locking mechanism inside the batch and reuse it. I have mutation either way. Let me play with this.

pkg/controllers/provisioning/batcher.go
```diff
@@ -0,0 +1,102 @@
...
+		window = time.Since(start)
+	}()
+	timeout := time.NewTimer(MaxBatchDuration)
+	idle := time.NewTimer(BatchIdleDuration)
+	for {
+		if len(items) >= MaxItemsPerBatch {
+			return
+		}
+		select {
+		case task := <-b.queue:
```
bwagner5, line 91: is this a task? I mean I guess it's generic-y but maybe just `item`?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -89,28 +76,27 @@ type Provisioner struct {
...
 func (p *Provisioner) provision(ctx context.Context) (err error) {
-	// Wait for a batch of pods, release when done
-	pods := p.batch(ctx)
-	defer func() {
-		for i := 0; i < len(pods); i++ {
-			p.wait <- struct{}{}
+	// Batch pods
+	logging.FromContext(ctx).Infof("Waiting for unschedulable pods")
+	items, window := p.batcher.Wait(ctx)
+	defer p.batcher.Flush()
```
bwagner5, line 88: I think this defer Flush is still incorrect. 

When `Wait(ctx)` returns the items for the current batch, items can still be added to the batch before the function exits (so before the defer runs `Flush`).  The items added in-between the `Wait()` and the `Flush()` will be released when the function returns even though they were not in the items returned from this iteration of `Wait()`. 

Subsequent calls of `Wait` would return the items added in-between, but the chan returned from Add() would have already released the producer (pod controller).

bwagner5: nvm, since the chan is unbuffered, Wait() has to be called before anything can be added.

bwagner5 (dismissed):
nice, much cleaner! LGTM 🚀

bwagner5 (approved):
lgtm

PR #1074, opened 2022-01-04
Fix the makefile bug when the cluster name includes region [WIP]

felix-zhe-huang:
**1. Issue, if available:**
When creating cluster using eksctl, the cluster name in kube config file is ${CLUSTER_NAME}.${AWS_DEFAULT_REGION}.eksctl.io.
It causes the helm options in the makefile to use the wrong ${CLUSTER_NAME} value.

Fix the command that compute the CLUSTER_NAME environment value in makefile.

- [X] No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ WITH_GOFLAGS = GOFLAGS=$(GOFLAGS)
 WITH_RELEASE_REPO = KO_DOCKER_REPO=$(RELEASE_REPO)
 
 ## Extra helm options
-CLUSTER_NAME ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].name}' | rev | cut -d"/" -f1 | rev)
+CLUSTER_NAME ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].name}' | rev | cut -d"/" -f1 | rev | cut -d"." -f1)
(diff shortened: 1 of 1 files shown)


ellistarn (approved):

felix-zhe-huang:
Just found that this may be a bigger problem. It. seems that it may cause node failing to join the cluster.

Makefile
```diff
@@ -12,7 +12,7 @@ WITH_GOFLAGS = GOFLAGS=$(GOFLAGS)
 WITH_RELEASE_REPO = KO_DOCKER_REPO=$(RELEASE_REPO)
 
 ## Extra helm options
-CLUSTER_NAME ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].name}' | rev | cut -d"/" -f1 | rev)
+CLUSTER_NAME ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].name}' | rev | cut -d"/" -f1 | rev | cut -d"." -f1)
```
bwagner5, line 15: I think this will work for EKS but breaks kops since Kops uses the FQDN

felix-zhe-huang: Yes I realized that we need a more consistent and inclusive way to config the Karpenter deployment process instead of hacking the environment variable.  Created an issue #1075 to track this.

olemarkus: For kOps the true way is `kubectl config current-context`

felix-zhe-huang:
This is a wrong solution. Closing for good.

Issue #1034, opened 2021-12-22
Binpacking algorithm incorrectly selects instance type with insufficient allocatable memory


jalawala:
## Version

**Karpenter**: v0.0.0

**Kubernetes**: v1.0.0
0.5.3

## Expected Behavior

 This Issue is observed with smaller instance say t4g.micro which takes longer time to get into ready state
T0 =  pod in pending
Line 4: arm64-6bcdd8f45-95j9v   0/1     Pending   0          5s    <none>   ip-192-168-77-153.ec2.internal   <none>           <none>
T1 = pod is terminated by Karpenter assuming that there is some problem with node but node is not ready yet
Line 5: arm64-6bcdd8f45-95j9v   0/1     Terminating   0          52s   <none>   ip-192-168-77-153.ec2.internal   <none>           <none>
T2 = node is actually ready
Line 161: ip-192-168-77-153.ec2.internal   Ready                      <none>   67s    v1.21.5-eks-bc4871b   t4g.micro       arm64   spot
Since pod is terminated and node is empty, TTL will be applied and removed. Karpenter will trigger a new instance but same Issue repets forerver and pod NEVER gets scheduled.
Karpenter should consider slow start up of small instance types before considering it as un healthy and then removing. Attacched file captures this beahivour in detail including logs.
## Actual Behavior

Karpenter should wait long enough for nodes to be ready before considering them to unhealthy

## Steps to Reproduce the Problem

I used below config for arm64
      nodeSelector:
        kubernetes.io/arch: arm64
      containers:
      - image: public.ecr.aws/nginx/nginx:latest
        imagePullPolicy: Always
        name: arm64
        resources:
          limits:
            cpu:  400m
            memory: 1000Mi
          requests:
            cpu: 400m
            memory: 1000Mi   

Karpenter selected t4g.micro where this issue is observed. detailed logs are attached below

## Resource Specs and Logs

[karpenter.txt](https://github.com/aws/karpenter/files/7759471/karpenter.txt)

mikesir87:
I’ve run into this as well and I think narrowed it down to the number of DaemonSets I had running. For a t4g.micro instance, the max pod limit is only 4. So, there actually wasn’t enough room on the nodes to run the workloads. Are you running any DaemonSets?

ellistarn:
It looks like our maxpods calculations are not factoring in daemonsets. We need to fix this.

mikesir87:
[...]

PR #1080, opened 2022-01-04
Fix the insufficient memory instance type bug

felix-zhe-huang:
**1. Issue, if available:**
resolve issue #1034 

- [X] No

### DIFF
--- a/pkg/cloudprovider/fake/cloudprovider.go
+++ b/pkg/cloudprovider/fake/cloudprovider.go
@@ -22,6 +22,7 @@ import (
 	"github.com/Pallinder/go-randomdata"
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
+	"github.com/aws/karpenter/pkg/utils/injection"
 	"go.uber.org/multierr"
 	"knative.dev/pkg/apis"
 
@@ -39,6 +40,17 @@ func (c *CloudProvider) Create(_ context.Context, constraints *v1alpha5.Constrai
 	for i := 0; i < quantity; i++ {
 		name := strings.ToLower(randomdata.SillyName())
 		instance := instanceTypes[0]
+		// To create error test cases, pick the zero instance types when they are included
+		for _, instanceType := range instanceTypes {
+			if instanceType.Name() == "invalid-pod-instance-type" {
+				instance = instanceType
+				break
+			}
+			if instanceType.Name() == "invalid-memory-instance-type" {
+				instance = instanceType
+				break
+			}
+		}
 		var zone, capacityType string
 		for _, o := range instance.Offerings() {
 			if constraints.Requirements.CapacityTypes().Has(o.CapacityType) {
@@ -78,11 +90,12 @@ func (c *CloudProvider) Create(_ context.Context, constraints *v1alpha5.Constrai
 	return err
 }
 
-func (c *CloudProvider) GetInstanceTypes(_ context.Context, _ *v1alpha5.Constraints) ([]cloudprovider.InstanceType, error) {
+func (c *CloudProvider) GetInstanceTypes(ctx context.Context, _ *v1alpha5.Constraints) ([]cloudprovider.InstanceType, error) {
 	if c.InstanceTypes != nil {
 		return c.InstanceTypes, nil
 	}
-	return []cloudprovider.InstanceType{
+
+	instanceTypes := []cloudprovider.InstanceType{
 		NewInstanceType(InstanceTypeOptions{
 			Name: "default-instance-type",
 		}),
@@ -111,7 +124,29 @@ func (c *CloudProvider) GetInstanceTypes(_ context.Context, _ *v1alpha5.Constrai
 			Name:         "arm-instance-type",
 			Architecture: "arm64",
 		}),
-	}, nil
+	}
+
+	if injection.GetTestInstance(ctx) {
+		invalidTypes := []cloudprovider.InstanceType{
+			NewInstanceType(InstanceTypeOptions{
+				Name:   "invalid-memory-instance-type",
+				CPU:    resource.MustParse("6"),
+				Memory: resource.MustParse("500Mi"),
+			}),
+			NewInstanceType(InstanceTypeOptions{
+				Name: "invalid-pod-instance-type",
+				CPU:  resource.MustParse("6"),
+				Pods: resource.MustParse("1"),
+			}),
+			NewInstanceType(InstanceTypeOptions{
+				Name:   "large-pod-instance-type",
+				CPU:    resource.MustParse("8"),
+				Memory: resource.MustParse("8Gi"),
+			}),
+		}
+		instanceTypes = append(instanceTypes, invalidTypes...)
+	}
+	return instanceTypes, nil
 }
 
 func (c *CloudProvider) Delete(context.Context, *v1.Node) error {
--- a/pkg/cloudprovider/fake/cloudprovider.go
+++ b/pkg/cloudprovider/fake/cloudprovider.go
@@ -22,7 +22,6 @@ import (
 	"github.com/Pallinder/go-randomdata"
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
-	"github.com/aws/karpenter/pkg/utils/injection"
 	"go.uber.org/multierr"
 	"knative.dev/pkg/apis"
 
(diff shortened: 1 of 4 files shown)


ellistarn (commented):
Can you include a test to prevent this from regressing?

pkg/controllers/provisioning/binpacking/packable.go
```diff
@@ -74,6 +74,9 @@ func PackablesFor(ctx context.Context, instanceTypes []cloudprovider.InstanceTyp
 		packables = append(packables, packable)
 	}
 	// Sort in ascending order so that the packer can short circuit bin-packing for larger instance types
+	// Sort according to the lexicographical order of [CPU, memory]
+	// It may result in cases where an instance type may have larger index when it has larger CPU but smaller memory
```
ellistarn, line 78: This comment is better placed next to the if check where it applies

pkg/controllers/provisioning/binpacking/packer.go
```diff
@@ -182,7 +182,10 @@ func (p *Packer) packWithLargestPod(unpackedPods []*v1.Pod, packables []*Packabl
 			// Trim the bestInstances so that provisioning APIs in cloud providers are not overwhelmed by the number of instance type options
 			// For example, the AWS EC2 Fleet API only allows the request to be 145kb which equates to about 130 instance type options.
 			for j := i; j < len(packables) && j-i < MaxInstanceTypes; j++ {
-				bestInstances = append(bestInstances, packables[j])
+				// Excluding instance type with smaller memory
+				if packables[i].Memory().Cmp(*packables[j].Memory()) != 1 {
```
ellistarn, line 186: I typically see this written in the form:
```
<= 0  // a <= b
>= 0 // a >= b
== 0 // a == b
< 0  // a < b 
> 0 //  a > b
```

pkg/controllers/provisioning/binpacking/packable.go
felix-zhe-huang: done

pkg/controllers/provisioning/binpacking/packer.go
felix-zhe-huang: done

felix-zhe-huang:
> Can you include a test to prevent this from regressing?

Happy to do so.

pkg/controllers/provisioning/suite_test.go
```diff
@@ -145,6 +147,26 @@ var _ = Describe("Provisioning", func() {
...
+		Context("Instance Type Selection", func() {
+			BeforeEach(func() {
+				ctx = injection.WithTestInstance(ctx, true)
+			})
+
+			AfterEach(func() {
+				ctx = injection.WithTestInstance(ctx, false)
+			})
+
+			It("should not select invalid instance types", func() {
```
ellistarn, line 159: I would explicitly use a specific instance type (using a node selector) and then verify that the pod can't be scheduled.

pkg/controllers/provisioning/suite_test.go
```diff
@@ -145,6 +147,26 @@ var _ = Describe("Provisioning", func() {
...
+				pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioningController, provisioner, test.UnschedulablePod(
+					test.PodOptions{
+						ResourceRequirements: v1.ResourceRequirements{Requests: v1.ResourceList{v1.ResourceCPU: resource.MustParse("3"), v1.ResourceMemory: resource.MustParse("3Gi")}},
+					},
+				))[0]
+				node := ExpectScheduled(ctx, env.Client, pod)
```
ellistarn, line 165: I'd check the node.Labels[instancetype] for a positive scheduling test.

pkg/utils/injection/injection.go
```diff
@@ -79,3 +79,17 @@ func GetControllerName(ctx context.Context) string {
 	}
 	return name.(string)
 }
+
+type testInstanceKey struct{}
```
ellistarn, line 83: This injection pattern is useful to keep interfaces in production code clean. I'm wary of coupling test implementation details to this injection package. There are some other examples of injecting data into the EC2 fake. Happy to help offline if you need.

pkg/controllers/provisioning/suite_test.go
felix-zhe-huang: So here is the minor catch. The illy created instance types that I add will also fails the other tests in all the suites. Unless we also specify node selector for other tests to exclude these illy created instance types, it may create confusions what the problem is.

pkg/controllers/provisioning/suite_test.go
felix-zhe-huang: By positive test you meant it will select the illy constructed instance type?

pkg/utils/injection/injection.go
felix-zhe-huang: Yeah this may not be ideal. I am removing it and implements a function to switch on and off those special instance types instead.

pkg/controllers/provisioning/suite_test.go
```diff
@@ -145,6 +145,18 @@ var _ = Describe("Provisioning", func() {
 				ExpectScheduled(ctx, env.Client, pod)
 			}
 		})
+		Context("Instance Type Selection", func() {
+			It("should not select invalid instance types", func() {
+				pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioningController, provisioner, test.UnschedulablePod(
+					test.PodOptions{
+						ResourceRequirements: v1.ResourceRequirements{Requests: v1.ResourceList{v1.ResourceCPU: resource.MustParse("3"), v1.ResourceMemory: resource.MustParse("3Gi")}},
+					},
+				))[0]
```
bwagner5, line 154: I'm not sure what the output is if ExpectedProvisioned returns 0 pods, but I'm guessing it's a not very graceful crash. Would it be more informative to the user to add an explicit:

```
Expect(len(pods)).To(Equal(1))
node := ExpectScheduled(ctx, env.Client, pod[0])
```

pkg/cloudprovider/fake/cloudprovider.go
```diff
@@ -39,6 +39,17 @@ func (c *CloudProvider) Create(_ context.Context, constraints *v1alpha5.Constrai
 	for i := 0; i < quantity; i++ {
 		name := strings.ToLower(randomdata.SillyName())
 		instance := instanceTypes[0]
+		// To create error test cases, give preferences to the illy constructed instance types
```
bwagner5, line 42: I'd be confused looking at just the tests on how the bad instance types are guaranteed to cause a test failure. I think it would be more apparent if the fake cloudprovider was more generic and the setup of a bad set of instance types is set in the test. 

You could use the exported InstanceTypes field in the CloudProvider struct to set a specific set of bad instance types that would be apparent in the actual test that it is guaranteed to fail if the binpacking short circuit logic is not working. Wdyt?

pkg/controllers/provisioning/suite_test.go
```diff
@@ -145,6 +145,18 @@ var _ = Describe("Provisioning", func() {
...
+				node := ExpectScheduled(ctx, env.Client, pod)
+				Expect(*node.Status.Allocatable.Cpu()).To(Equal(resource.MustParse("4")))
```
bwagner5, line 156: This seems a bit fragile if we change the fake instance type data, the functionality of not choosing an instance type may still work properly but the test could fail if the fake instance type doesn't have 4 vcpus and 4 GBs of allocatable RAM.

pkg/controllers/provisioning/binpacking/packer.go
```diff
@@ -182,7 +182,12 @@ func (p *Packer) packWithLargestPod(unpackedPods []*v1.Pod, packables []*Packabl
 			// Trim the bestInstances so that provisioning APIs in cloud providers are not overwhelmed by the number of instance type options
 			// For example, the AWS EC2 Fleet API only allows the request to be 145kb which equates to about 130 instance type options.
 			for j := i; j < len(packables) && j-i < MaxInstanceTypes; j++ {
-				bestInstances = append(bestInstances, packables[j])
+				// packable nodes are sorted lexicographically according to the order of [CPU, memory]
+				// It may result in cases where an instance type may have larger index value when it has more CPU but fewer memory
+				// Need to exclude instance type with smaller memory and fewer pods
+				if packables[i].Memory().Cmp(*packables[j].Memory()) <= 0 && packables[i].Pods().Cmp(*packables[j].Pods()) <= 0 {
```
bwagner5, line 188: nice!

ellistarn (commented):
Discussed offline. Testing in this way is fairly misleading, and is better done in real CI w/ chaos testing. Let's cut an issue for chaos testing of pod resource requests in a real CI environment and merge this with just the code change.

bwagner5 (approved):
lgtm

ellistarn (approved):

PR #1083, opened 2022-01-04
Fail if multiple aws security groups tagged for cluster

suket22:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/1039

*Previous Behavior* - 
If no SecurityGroupSelector is specified, we find all securityGroups that have a tag key of `kubernetes.io/cluster/$clusterName` and randomly select a single one of them. We select just one of them because the Load Balancer controller forces us to. See [this](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2367) for more information

* New Behavior* - 
* If no SecurityGroupSelector is specified, we will find a securityGroup that has a tag key of `kubernetes.io/cluster/$clusterName` and use that. If we find more than one such securityGroup with that tag key, we will fail launching worker nodes. 

This will likely force customers using kOps or eksctl to provide a securityGroupSelector in their provisionerSpec to ensure worker nodes are launched successfully. 

(WIP - will add)
- [ ] No

**Testing**

I added that tag to an additional SG in my VPC and saw the provisioner complain - 
```
2022-01-04T21:35:28.621Z	INFO	controller.provisioning	Computed packing of 1 node(s) for 7 pod(s) with instance type option(s) [c1.xlarge c4.2xlarge c3.2xlarge c5ad.2xlarge c5a.2xlarge c6i.2xlarge c5d.2xlarge c5.2xlarge c5n.2xlarge m3.2xlarge t3.2xlarge m6a.2xlarge m5ad.2xlarge m5.2xlarge m5dn.2xlarge m4.2xlarge t3a.2xlarge m6i.2xlarge m5zn.2xlarge m5d.2xlarge]	{"commit": "32dbe02", "provisioner": "default"}
2022-01-04T21:35:28.627Z	ERROR	controller.provisioning	Could not launch node, launching instances, getting launch template configs, getting launch templates, only one group with tag kubernetes.io/cluster/pdx-cluster is allowed	{"commit": "32dbe02", "provisioner": "default"}
```

I then made my securityGroupSelector a little more specific and saw nodes were being launched successfully.

### DIFF
--- a/pkg/cloudprovider/aws/fake/ec2api.go
+++ b/pkg/cloudprovider/aws/fake/ec2api.go
@@ -68,6 +68,7 @@ func (e *EC2API) Reset() {
 	e.EC2Behavior = EC2Behavior{
 		CalledWithCreateFleetInput:          set.NewSet(),
 		CalledWithCreateLaunchTemplateInput: set.NewSet(),
+		DescribeSecurityGroupsOutput:        nil,
 		Instances:                           sync.Map{},
 		LaunchTemplates:                     sync.Map{},
 		InsufficientCapacityPools:           []CapacityPool{},
--- a/pkg/cloudprovider/aws/securitygroups.go
+++ b/pkg/cloudprovider/aws/securitygroups.go
@@ -46,11 +46,14 @@ func (s *SecurityGroupProvider) Get(ctx context.Context, constraints *v1alpha1.C
 	if err != nil {
 		return nil, err
 	}
-	// This hack works around
-	// https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2367
 	// The LoadBalancer Controller expects a single security group with the
-	// cluster tag, but provisioning tools like eksctl and kops create multiple.
-	securityGroups = s.filterClusterTaggedGroups(ctx, securityGroups)
+	// cluster tag, but provisioning tools like eksctl and kops create multiple. Until
+	// https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2367 is addressed, we will
+	// enforce that only a single security group with the cluster tag can be chosen by the selector.
+	err = s.verifyClusterTaggedGroups(ctx, securityGroups)
+	if err != nil {
+		return nil, err
+	}
 	// Fail if no security groups found
 	if len(securityGroups) == 0 {
 		return nil, fmt.Errorf("no security groups exist given constraints")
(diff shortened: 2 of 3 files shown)


pkg/cloudprovider/aws/suite_test.go
```diff
@@ -411,6 +418,20 @@ var _ = Describe("Allocation", func() {
 					aws.String("test-security-group-3"),
 				))
 			})
+			It("allow multiple security groups with cluster tag", func() {
```
suket22, line 421: This unit test made more sense in an earlier code revision when I wanted to verify that multiple security groups with the cluster tag would lead to a runtime controller failure. I'm still leaving it in though because if we ever do decide to regress on this by re-adding that functionality, it should break.

pkg/cloudprovider/aws/fake/ec2api.go
```diff
@@ -68,6 +68,7 @@ func (e *EC2API) Reset() {
 	e.EC2Behavior = EC2Behavior{
 		CalledWithCreateFleetInput:          set.NewSet(),
 		CalledWithCreateLaunchTemplateInput: set.NewSet(),
+		DescribeSecurityGroupsOutput:        nil,
```
ellistarn, line 71: You don't need to include this, since by default it's nil. There are other fields that do the same thing.

pkg/cloudprovider/aws/securitygroups.go
```diff
@@ -46,11 +45,6 @@ func (s *SecurityGroupProvider) Get(ctx context.Context, constraints *v1alpha1.C
 	if err != nil {
 		return nil, err
 	}
-	// This hack works around
-	// https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2367
-	// The LoadBalancer Controller expects a single security group with the
-	// cluster tag, but provisioning tools like eksctl and kops create multiple.
-	securityGroups = s.filterClusterTaggedGroups(ctx, securityGroups)
```
ellistarn, line 53: Looking at line 63:
`if value == "*" {`
I think this is the default behavior of the API, since it supports wildcards. We may be able to simplify.

ellistarn: Make sure we point to  https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2367 in the docs

pkg/cloudprovider/aws/suite_test.go
ellistarn: This is in the class of "inversion testing" or testing that things don't happen. I tend to lean away from it, since there are infinitely many things that we can't test in this class. 

It does look like we were missing a test for using the security group selector, so perhaps you can just rename the `It` block to something like `should discover security groups by tag`?

suket22: Agreed, so just renamed the It block.

pkg/cloudprovider/aws/securitygroups.go
suket22: Nice!

I didn't find any mention of it in the documentation but it worked when I tried it out.

```
aws ec2 describe-security-groups --filters Name=vpc-id,Values=vpc-123 Name=tag-key,Values=kubernetes.io/cluster/myCluster

and 

aws ec2 describe-security-groups --filters Name=vpc-id,Values=vpc-123 Name=tag:kubernetes.io/cluster/myCluster,Values="*"
```

gave me identical results, so using just the latter.

ellistarn (approved):
Nice!

PR #1092, opened 2022-01-06
fix selector labels on helm chart

bwagner5:
**1. Issue, if available:**
N/A

 - #1001 added standard k8s labels across all resources and the ability for users to provide labels that are propagated to all resources. Part of that PR also updated the selectorLabels with these additional labels, however, selectorLabels are immutable so when upgrading the chart from the previous, those resources fail to Patch.  This PR removes the additional selectors and sticks with the `karpenter: controller` and `karpenter: webhook` to maintain backwards compatibility.  

tested in my local cluster by updating a previous installation with helm.

- [x] No

### DIFF
--- a/charts/karpenter/templates/controller/deployment.yaml
+++ b/charts/karpenter/templates/controller/deployment.yaml
@@ -26,7 +26,6 @@ spec:
   selector:
     matchLabels:
       karpenter: controller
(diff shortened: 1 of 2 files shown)


felix-zhe-huang:
Verified that the helm chart renders. LGTM

felix-zhe-huang (approved):

Issue #923, opened 2021-12-06
Ability to specify amiFamily


aldredb:
**Tell us about your request**
In `eksctl` **managedNodeGroup** definition, we can specify the **amiFamily** as **Bottlerocket** or **AmazonLinux2** . For Bottlerocket, there is even a section to in `eksctl` to specify customizations. Request to add similar feature in Karpenter

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
WIthout specifying any launch template, Karpenter will default to Amazon Linux 2. Customers are increasingly adopting Bottlerocket. Having to maintain custom launch templates just to run another AMI family increases barrier to entry.

**Are you currently working around this issue?**
Custom launch templates

**Additional context**
N/A

**Attachments**
If you think you might have additional information that you'd like to include via an attachment, please do - we'll take a look. (Remember to remove any personally-identifiable information.)

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

inductor:
Now that Bottlerocket does not necessarily require LaunchTemplate to bootstrap as part of EKS cluster, so I think this should be possible.

lucasreed:
👍  Would love to see this as an option instead of maintaining a separate LaunchConfiguration.

rayterrill:
We'd like to see this as well - our default is Bottlerocket today. Would love to just specify the AMI family and let Karpenter grab the latest AMI if possible.

PR #1110, opened 2022-01-09
Adding initial support for bottlerocket without custom launch template

rayterrill:
**1. Issue, if available:**
#923 

Adds initial support for amiFamily, allowing Karpenter to choose bottlerocket AMIs in additional to EKS Optimized, without a custom launch template.

Tested manually in POC cluster.

- [ ] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/requirements.go
+++ b/pkg/apis/provisioning/v1alpha5/requirements.go
@@ -30,7 +30,6 @@ var (
 	ArchitectureAmd64    = "amd64"
 	ArchitectureArm64    = "arm64"
 	OperatingSystemLinux = "linux"
-	BottleRocket         = "bottlerocket"
 
 	// RestrictedLabels are injected by Cloud Providers
 	RestrictedLabels = stringsets.NewString(
--- a/pkg/apis/provisioning/v1alpha5/requirements.go
+++ b/pkg/apis/provisioning/v1alpha5/requirements.go
@@ -30,6 +30,7 @@ var (
 	ArchitectureAmd64    = "amd64"
 	ArchitectureArm64    = "arm64"
 	OperatingSystemLinux = "linux"
+	BottleRocket         = "bottlerocket"
 
 	// RestrictedLabels are injected by Cloud Providers
 	RestrictedLabels = stringsets.NewString(
--- a/pkg/cloudprovider/aws/ami.go
+++ b/pkg/cloudprovider/aws/ami.go
@@ -24,6 +24,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ssm/ssmiface"
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
+	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
 	"github.com/patrickmn/go-cache"
 	"k8s.io/client-go/kubernetes"
 	"knative.dev/pkg/logging"
@@ -46,7 +47,7 @@ func NewAMIProvider(ssm ssmiface.SSMAPI, clientSet *kubernetes.Clientset) *AMIPr
 }
 
 // Get returns a set of AMIIDs and corresponding instance types. AMI may vary due to architecture, accelerator, etc
-func (p *AMIProvider) Get(ctx context.Context, instanceTypes []cloudprovider.InstanceType) (map[string][]cloudprovider.InstanceType, error) {
+func (p *AMIProvider) Get(ctx context.Context, instanceTypes []cloudprovider.InstanceType, constraints *v1alpha1.Constraints) (map[string][]cloudprovider.InstanceType, error) {
 	version, err := p.kubeServerVersion(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("kube server version, %w", err)
@@ -54,7 +55,7 @@ func (p *AMIProvider) Get(ctx context.Context, instanceTypes []cloudprovider.Ins
 	// Separate instance types by unique queries
 	amiQueries := map[string][]cloudprovider.InstanceType{}
 	for _, instanceType := range instanceTypes {
-		query := p.getSSMQuery(instanceType, version)
+		query := p.getSSMQuery(ctx, constraints.AMIFamily, instanceType, version)
 		amiQueries[query] = append(amiQueries[query], instanceType)
 	}
 	// Separate instance types by unique AMIIDs
@@ -83,12 +84,21 @@ func (p *AMIProvider) getAMIID(ctx context.Context, query string) (string, error
 	return ami, nil
 }
 
-func (p *AMIProvider) getSSMQuery(instanceType cloudprovider.InstanceType, version string) string {
+func (p *AMIProvider) getSSMQuery(ctx context.Context, amiFamily string, instanceType cloudprovider.InstanceType, version string) string {
 	var amiSuffix string
+	var arch string
 	if !instanceType.NvidiaGPUs().IsZero() || !instanceType.AWSNeurons().IsZero() {
 		amiSuffix = "-gpu"
 	} else if instanceType.Architecture() == v1alpha5.ArchitectureArm64 {
 		amiSuffix = "-arm64"
+		arch = "arm64"
+	} else if instanceType.Architecture() == v1alpha5.ArchitectureAmd64 {
+		arch = "arm64"
+	}
+
+	if amiFamily == v1alpha5.BottleRocket {
+		logging.FromContext(ctx).Debugf("AMIFamily was: %s", amiFamily)
+		return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, arch)
 	}
 	return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
 }
--- a/pkg/cloudprovider/aws/ami.go
+++ b/pkg/cloudprovider/aws/ami.go
@@ -47,7 +47,7 @@ func NewAMIProvider(ssm ssmiface.SSMAPI, clientSet *kubernetes.Clientset) *AMIPr
 }
 
 // Get returns a set of AMIIDs and corresponding instance types. AMI may vary due to architecture, accelerator, etc
-func (p *AMIProvider) Get(ctx context.Context, instanceTypes []cloudprovider.InstanceType, constraints *v1alpha1.Constraints) (map[string][]cloudprovider.InstanceType, error) {
+func (p *AMIProvider) Get(ctx context.Context, constraints *v1alpha1.Constraints, instanceTypes []cloudprovider.InstanceType) (map[string][]cloudprovider.InstanceType, error) {
 	version, err := p.kubeServerVersion(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("kube server version, %w", err)
@@ -55,7 +55,7 @@ func (p *AMIProvider) Get(ctx context.Context, instanceTypes []cloudprovider.Ins
 	// Separate instance types by unique queries
 	amiQueries := map[string][]cloudprovider.InstanceType{}
 	for _, instanceType := range instanceTypes {
-		query := p.getSSMQuery(ctx, constraints.AWS.AMIFamily, instanceType, version)
+		query := p.getSSMQuery(ctx, constraints, instanceType, version)
 		amiQueries[query] = append(amiQueries[query], instanceType)
 	}
 	// Separate instance types by unique AMIIDs
@@ -84,7 +84,7 @@ func (p *AMIProvider) getAMIID(ctx context.Context, query string) (string, error
 	return ami, nil
 }
 
-func (p *AMIProvider) getSSMQuery(ctx context.Context, amiFamily string, instanceType cloudprovider.InstanceType, version string) string {
+func (p *AMIProvider) getSSMQuery(ctx context.Context, constraints *v1alpha1.Constraints, instanceType cloudprovider.InstanceType, version string) string {
 	var amiSuffix string
 	var arch string
 	if !instanceType.NvidiaGPUs().IsZero() || !instanceType.AWSNeurons().IsZero() {
@@ -96,10 +96,13 @@ func (p *AMIProvider) getSSMQuery(ctx context.Context, amiFamily string, instanc
 		arch = "arm64"
 	}
 
-	if amiFamily == v1alpha5.BottleRocket {
-		logging.FromContext(ctx).Debugf("AMIFamily was: %s", amiFamily)
-		return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, arch)
+	if constraints.AMIFamily != nil {
+		switch *constraints.AMIFamily {
+		case v1alpha1.OperatingSystemBottleRocket:
+			return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, arch)
+		}
 	}
+
 	return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
 }
 
--- a/pkg/cloudprovider/aws/ami.go
+++ b/pkg/cloudprovider/aws/ami.go
@@ -93,7 +93,7 @@ func (p *AMIProvider) getSSMQuery(ctx context.Context, constraints *v1alpha1.Con
 		amiSuffix = "-arm64"
 		arch = "arm64"
 	} else if instanceType.Architecture() == v1alpha5.ArchitectureAmd64 {
-		arch = "arm64"
+		arch = "x86_64"
 	}
 
 	if constraints.AMIFamily != nil {
--- a/pkg/cloudprovider/aws/ami.go
+++ b/pkg/cloudprovider/aws/ami.go
@@ -84,29 +84,40 @@ func (p *AMIProvider) getAMIID(ctx context.Context, query string) (string, error
 	return ami, nil
 }
 
-func (p *AMIProvider) getSSMQuery(ctx context.Context, constraints *v1alpha1.Constraints, instanceType cloudprovider.InstanceType, version string) string {
-	var amiSuffix string
-	var arch string
+// getAL2Alias returns a properly-formatted alias for an Amazon Linux AMI from SSM
+func (p *AMIProvider) getAL2Alias(version string, instanceType cloudprovider.InstanceType) string {
+	amiSuffix := ""
 	if !instanceType.NvidiaGPUs().IsZero() || !instanceType.AWSNeurons().IsZero() {
 		amiSuffix = "-gpu"
 	} else if instanceType.Architecture() == v1alpha5.ArchitectureArm64 {
-		amiSuffix = "-arm64"
-		arch = "arm64"
-	} else if instanceType.Architecture() == v1alpha5.ArchitectureAmd64 {
-		arch = "x86_64"
+		amiSuffix = fmt.Sprintf("-%s", instanceType.Architecture())
 	}
+	return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
+}
 
+// getBottlerocketAlias returns a properly-formatted alias for a Bottlerocket AMI from SSM
+func (p *AMIProvider) getBottlerocketAlias(version string, instanceType cloudprovider.InstanceType) string {
+	arch := "x86_64"
+	if instanceType.Architecture() == v1alpha5.ArchitectureArm64 {
+		arch = instanceType.Architecture()
+	}
+	return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, arch)
+}
+
+func (p *AMIProvider) getSSMQuery(ctx context.Context, constraints *v1alpha1.Constraints, instanceType cloudprovider.InstanceType, version string) string {
+	ssmQuery := p.getAL2Alias(version, instanceType)
 	if constraints.AMIFamily != nil {
 		if *constraints.AMIFamily == v1alpha1.OperatingSystemBottleRocket {
-			return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, arch)
+			ssmQuery = p.getBottlerocketAlias(version, instanceType)
 		} else if *constraints.AMIFamily == v1alpha1.OperatingSystemEKSOptimized {
-			return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
+			ssmQuery = p.getAL2Alias(version, instanceType)
 		} else {
-			return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
+			logging.FromContext(ctx).Warnf("AMIFamily was set, but was not one of %s or %s. Setting to %s as the default.", v1alpha1.OperatingSystemEKSOptimized, v1alpha1.OperatingSystemBottleRocket, v1alpha1.OperatingSystemEKSOptimized)
+			ssmQuery = p.getAL2Alias(version, instanceType)
 		}
-	} else {
-		return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
 	}
+
+	return ssmQuery
 }
 
 func (p *AMIProvider) kubeServerVersion(ctx context.Context) (string, error) {
--- a/pkg/cloudprovider/aws/ami.go
+++ b/pkg/cloudprovider/aws/ami.go
@@ -97,13 +97,16 @@ func (p *AMIProvider) getSSMQuery(ctx context.Context, constraints *v1alpha1.Con
 	}
 
 	if constraints.AMIFamily != nil {
-		switch *constraints.AMIFamily {
-		case v1alpha1.OperatingSystemBottleRocket:
+		if *constraints.AMIFamily == v1alpha1.OperatingSystemBottleRocket {
 			return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, arch)
+		} else if *constraints.AMIFamily == v1alpha1.OperatingSystemEKSOptimized {
+			return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
+		} else {
+			return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
 		}
+	} else {
+		return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
 	}
-
-	return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
 }
 
 func (p *AMIProvider) kubeServerVersion(ctx context.Context) (string, error) {
--- a/pkg/cloudprovider/aws/ami.go
+++ b/pkg/cloudprovider/aws/ami.go
@@ -55,7 +55,7 @@ func (p *AMIProvider) Get(ctx context.Context, instanceTypes []cloudprovider.Ins
 	// Separate instance types by unique queries
 	amiQueries := map[string][]cloudprovider.InstanceType{}
 	for _, instanceType := range instanceTypes {
-		query := p.getSSMQuery(ctx, constraints.AMIFamily, instanceType, version)
+		query := p.getSSMQuery(ctx, constraints.AWS.AMIFamily, instanceType, version)
 		amiQueries[query] = append(amiQueries[query], instanceType)
 	}
 	// Separate instance types by unique AMIIDs
--- a/pkg/cloudprovider/aws/apis/v1alpha1/provider.go
+++ b/pkg/cloudprovider/aws/apis/v1alpha1/provider.go
@@ -45,7 +45,7 @@ type AWS struct {
 	metav1.TypeMeta `json:",inline"`
 	// AMIFamily is the AMI family that instances use.
 	// +optional
-	AMIFamily string `json:"amiFamily"`
+	AMIFamily *string `json:"amiFamily,omitempty"`
 	// InstanceProfile is the AWS identity that instances use.
 	// +required
 	InstanceProfile string `json:"instanceProfile"`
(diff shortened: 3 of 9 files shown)


rayterrill:
Go easy on me - new to Karpenter and still a novice at Go itself - but eager to learn + grow! :)

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -25,6 +25,7 @@ var (
 	ArchitectureAmd64    = "amd64"
 	ArchitectureArm64    = "arm64"
 	OperatingSystemLinux = "linux"
+	BottleRocket         = "bottlerocket"
```
ellistarn, line 28: We try to keep AWS specific concepts isolated to the cloudprovider/aws package, check out the APIs.

pkg/cloudprovider/aws/ami.go
```diff
@@ -46,15 +47,15 @@ func NewAMIProvider(ssm ssmiface.SSMAPI, clientSet *kubernetes.Clientset) *AMIPr
 }
 
 // Get returns a set of AMIIDs and corresponding instance types. AMI may vary due to architecture, accelerator, etc
-func (p *AMIProvider) Get(ctx context.Context, instanceTypes []cloudprovider.InstanceType) (map[string][]cloudprovider.InstanceType, error) {
+func (p *AMIProvider) Get(ctx context.Context, instanceTypes []cloudprovider.InstanceType, constraints *v1alpha1.Constraints) (map[string][]cloudprovider.InstanceType, error) {
```
ellistarn, line 50: Do you mind reordering to (ctx, constraints, instanceTypes) to match other functions. Context is always first (go style), constraints is second in this codebase, as it's one of the most important concepts.

pkg/cloudprovider/aws/ami.go
```diff
@@ -46,15 +47,15 @@ func NewAMIProvider(ssm ssmiface.SSMAPI, clientSet *kubernetes.Clientset) *AMIPr
...
 	version, err := p.kubeServerVersion(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("kube server version, %w", err)
 	}
 	// Separate instance types by unique queries
 	amiQueries := map[string][]cloudprovider.InstanceType{}
 	for _, instanceType := range instanceTypes {
-		query := p.getSSMQuery(instanceType, version)
+		query := p.getSSMQuery(ctx, constraints.AMIFamily, instanceType, version)
```
ellistarn, line 58: consider passing the full constraints object through.

pkg/cloudprovider/aws/ami.go
```diff
@@ -83,12 +84,21 @@ func (p *AMIProvider) getAMIID(ctx context.Context, query string) (string, error
...
 		amiSuffix = "-gpu"
 	} else if instanceType.Architecture() == v1alpha5.ArchitectureArm64 {
 		amiSuffix = "-arm64"
+		arch = "arm64"
+	} else if instanceType.Architecture() == v1alpha5.ArchitectureAmd64 {
+		arch = "arm64"
+	}
+
+	if amiFamily == v1alpha5.BottleRocket {
+		logging.FromContext(ctx).Debugf("AMIFamily was: %s", amiFamily)
```
ellistarn, line 100: Don't forget to remove this before checking in.

pkg/cloudprovider/aws/ami.go
```diff
@@ -83,12 +84,21 @@ func (p *AMIProvider) getAMIID(ctx context.Context, query string) (string, error
...
 	if !instanceType.NvidiaGPUs().IsZero() || !instanceType.AWSNeurons().IsZero() {
 		amiSuffix = "-gpu"
 	} else if instanceType.Architecture() == v1alpha5.ArchitectureArm64 {
 		amiSuffix = "-arm64"
+		arch = "arm64"
+	} else if instanceType.Architecture() == v1alpha5.ArchitectureAmd64 {
+		arch = "arm64"
+	}
+
+	if amiFamily == v1alpha5.BottleRocket {
```
ellistarn, line 99: Thoughts on doing a switch statement here for all ami families? We may need this for flatcar, ubuntu, etc.

pkg/cloudprovider/aws/ami.go
```diff
@@ -83,12 +84,21 @@ func (p *AMIProvider) getAMIID(ctx context.Context, query string) (string, error
 	return ami, nil
 }
 
-func (p *AMIProvider) getSSMQuery(instanceType cloudprovider.InstanceType, version string) string {
+func (p *AMIProvider) getSSMQuery(ctx context.Context, amiFamily string, instanceType cloudprovider.InstanceType, version string) string {
 	var amiSuffix string
+	var arch string
 	if !instanceType.NvidiaGPUs().IsZero() || !instanceType.AWSNeurons().IsZero() {
```
ellistarn, line 90: @suket22 , what about arm+gpus? https://aws.amazon.com/blogs/machine-learning/aws-and-nvidia-to-bring-arm-based-instances-with-gpus-to-the-cloud/. I don't see this here: https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id.html.

pkg/cloudprovider/aws/ami.go
```diff
@@ -83,12 +84,21 @@ func (p *AMIProvider) getAMIID(ctx context.Context, query string) (string, error
...
 	} else if instanceType.Architecture() == v1alpha5.ArchitectureArm64 {
 		amiSuffix = "-arm64"
+		arch = "arm64"
+	} else if instanceType.Architecture() == v1alpha5.ArchitectureAmd64 {
+		arch = "arm64"
+	}
+
+	if amiFamily == v1alpha5.BottleRocket {
+		logging.FromContext(ctx).Debugf("AMIFamily was: %s", amiFamily)
+		return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, arch)
```
ellistarn, line 101: @suket22 It would be really nice if bottlerocket and eksoptimizedami were aligned to allow for simple substitution.

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
```diff
@@ -34,6 +34,9 @@ type AWS struct {
 	// TypeMeta includes version and kind of the extensions, inferred if not provided.
 	// +optional
 	metav1.TypeMeta `json:",inline"`
+	// AMIFamily is the AMI family that instances use.
+	// +optional
+	AMIFamily string `json:"amiFamily"`
```
ellistarn, line 39: Since this is optional, let's make it a `*string` and `json:"amiFamily,omitempty"`

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -40,6 +40,10 @@ func (i *InstanceType) Name() string {
 	return aws.StringValue(i.InstanceType)
 }
 
+func (i *InstanceType) AMIFamily() string {
```
ellistarn, line 43: This type currently implements cloudprovider.InstanceType. I don't see where this is even used -- perhaps from a previous implementation?

pkg/cloudprovider/types.go
```diff
@@ -57,6 +57,7 @@ type InstanceType interface {
 	// Note that though this is an array it is expected that all the Offerings are unique from one another
 	Offerings() []Offering
 	Architecture() string
+	AMIFamily() string
```
ellistarn, line 60: This is a vendor neutral API. We do our best to keep AWS specific concepts in our cloud provider code.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -220,6 +225,39 @@ func sortedKeys(m map[string]string) []string {
 	return keys
 }
 
+func (p *LaunchTemplateProvider) getBottleRocketUserData(ctx context.Context, constraints *v1alpha1.Constraints, additionalLabels map[string]string) (string, error) {
+	var userData bytes.Buffer
```
ellistarn, line 229: Perhaps cleaner to use standard strings? I'm not sure why we're using a buffer in the other generator.
```
userdata := fmt.Sprintf(`
  cluster-name ....
`)
if labels := functional.UnionStringMaps(additionalLabels, constraints.Labels); len(labels) > 0 {
  userdata += "[settings..."
  for label, value := range labels {
    userdata += ...
  }
}
```

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -94,15 +94,20 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, constraints *v1alpha1.
...
 	// Construct launch templates
 	launchTemplates := map[string][]cloudprovider.InstanceType{}
 	for amiID, instanceTypes := range amis {
 		// Get userData for Node
-		userData, err := p.getUserData(ctx, constraints, instanceTypes, additionalLabels)
+		var userData string
+		if constraints.AMIFamily == "bottlerocket" {
+			userData, err = p.getBottleRocketUserData(ctx, constraints, additionalLabels)
+		} else {
+			userData, err = p.getUserData(ctx, constraints, instanceTypes, additionalLabels)
```
ellistarn, line 109:
```suggestion
			userData, err = p.getEKSOptimizedUserData(ctx, constraints, instanceTypes, additionalLabels)
```

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -94,15 +94,20 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, constraints *v1alpha1.
...
 	if err != nil {
 		return nil, err
 	}
 	// Construct launch templates
 	launchTemplates := map[string][]cloudprovider.InstanceType{}
 	for amiID, instanceTypes := range amis {
 		// Get userData for Node
-		userData, err := p.getUserData(ctx, constraints, instanceTypes, additionalLabels)
+		var userData string
+		if constraints.AMIFamily == "bottlerocket" {
```
ellistarn, line 106: Thoughts on a switch statement?

ellistarn (commented):
Nice work! Some minor code nits, but I think you nailed it in concept. I want to make sure @akestner and @suket22 are feeling good about this API change. We can always refine it since this is an alpha API, but it's good to be careful regardless.

I think we might want use defaulting logic to set eksoptimized in the API when the provisioner is created. This has minor repercussions to existing provisioners, so perhaps we can do in memory defaulting (something we've talked about it the past) to maintain backwards compatibility. Happy to chat more on this on slack.

pkg/cloudprovider/aws/ami.go
suket22: EKS optimized AMIs for this combination doesn't exist right now

pkg/cloudprovider/aws/ami.go
suket22: I believe these SSM parameters are managed and owned by the [BottleRocket](https://github.com/bottlerocket-os/bottlerocket) folks - we could ask them if they'd like to restructure these but I don't think it's likely.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -220,6 +225,39 @@ func sortedKeys(m map[string]string) []string {
...
 }
 
+func (p *LaunchTemplateProvider) getBottleRocketUserData(ctx context.Context, constraints *v1alpha1.Constraints, additionalLabels map[string]string) (string, error) {
+	var userData bytes.Buffer
+	userData.WriteString(fmt.Sprintf(`[settings.kubernetes]
+cluster-name = "%s"
+api-server = "%s"
+`,
+		injection.GetOptions(ctx).ClusterName,
+		injection.GetOptions(ctx).ClusterEndpoint))
```
suket22, line 235: We might need to also handle `cluster-dns-ip` now that we support it in the [provisioner](https://karpenter.sh/docs/provisioner/#speckubeletconfiguration)

pkg/apis/provisioning/v1alpha5/requirements.go
suket22: nit: Should this be `OperatingSystemBottleRocket` to match the linux var

pkg/cloudprovider/aws/ami.go
rayterrill: @ellistarn we actually talked about this today at work.

Is there a concept of "release channels" for these OS versions (bottlerocket and amazon linux)? I see AL2's is called "recommended" - does that imply there's a latest that's newer/less tested? Similarly for bottlerocket - does latest imply there's a recommended?

We were also wondering if it might be beneficial to allow users to specify their own SSM query, so they can control their own "release cadence" outside of AWS' as needed (we might want to do something like let our DEV stuff pull latest, but "pin" PROD to something like "stable" that's more battle tested.

pkg/cloudprovider/aws/instancetype.go
rayterrill: This is where my go knowledge is lacking. If I remove this, I get this error from line 86 in cloudprovider/aws/instancetypes.go:
```
cannot use instanceType (variable of type *InstanceType) as cloudprovider.InstanceType value in argument to append: missing method AMIFamilycompilerInvalidIfaceAssign
```

rayterrill: Removed.

pkg/cloudprovider/types.go
rayterrill: Removed.

pkg/apis/provisioning/v1alpha5/requirements.go
rayterrill: moved this to cloudprovider/aws/apis/v1alpha1/register.go, renamed to OperatingSystemBottleRocket.

rayterrill:
I'll push another commit today with these changes.

rayterrill:
Just pushed a big commit that hopefully addresses all the feedback. Please let me know what else I can do - really enjoying working on this! :) Cheers!

pkg/cloudprovider/aws/ami.go
```diff
@@ -83,13 +84,25 @@ func (p *AMIProvider) getAMIID(ctx context.Context, query string) (string, error
...
 	if !instanceType.NvidiaGPUs().IsZero() || !instanceType.AWSNeurons().IsZero() {
 		amiSuffix = "-gpu"
 	} else if instanceType.Architecture() == v1alpha5.ArchitectureArm64 {
 		amiSuffix = "-arm64"
+		arch = "arm64"
+	} else if instanceType.Architecture() == v1alpha5.ArchitectureAmd64 {
+		arch = "arm64"
 	}
+
+	if constraints.AMIFamily != nil {
```
ellistarn, line 99: Switch statements don't work particularly well with pointers. I might just use an if series
```
if constraints.AMIFamily == v1alpha1.OperatingSystemBottleRocket {
  ...
} else if constraints.AMIFamily == v1alpha1.OperatingSystemEKSOptimized {
  ...
} else {
  // default behavior
}
```

Also -- we should have a value you can specify for `eks-optimized`

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -40,6 +40,12 @@ func (i *InstanceType) Name() string {
 	return aws.StringValue(i.InstanceType)
 }
 
+/*
```
ellistarn, line 43: reminder to clean up.

pkg/cloudprovider/aws/apis/v1alpha1/register.go
```diff
@@ -32,6 +32,7 @@ var (
 	AWSRestrictedLabelDomains = []string{
 		"k8s.aws",
 	}
+	OperatingSystemBottleRocket = "bottlerocket"
```
ellistarn, line 35: Kubernetes convention for yaml values is UpperCamelCase or `BottleRocket` `EKSOptimized`

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -94,15 +94,23 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, constraints *v1alpha1.
...
 		return nil, err
 	}
 	// Construct launch templates
 	launchTemplates := map[string][]cloudprovider.InstanceType{}
 	for amiID, instanceTypes := range amis {
 		// Get userData for Node
-		userData, err := p.getUserData(ctx, constraints, instanceTypes, additionalLabels)
+		var userData string
+		if constraints.AMIFamily != nil {
+			switch *constraints.AMIFamily {
```
ellistarn, line 107: similarly, I'd use if here.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -220,10 +228,43 @@ func sortedKeys(m map[string]string) []string {
...
+api-server = "%s"
+`,
+		injection.GetOptions(ctx).ClusterName,
+		injection.GetOptions(ctx).ClusterEndpoint)
+
+	if constraints.KubeletConfiguration.ClusterDNS != nil {
+		userData += fmt.Sprintf("cluster-dns-ip = \"%s\"", constraints.KubeletConfiguration.ClusterDNS)
+	}
+
+	caBundle, err := p.GetCABundle(ctx)
```
ellistarn, line 244: optional:

You might be able to collapse a fair amount of complexity by getting the CABundle outside of both userdata generation methods and passing it in. This means these functions will no longer error and no longer need error handling logic.

ellistarn: optional:

Looking deeper -- it really shouldnt be possible for GetCABundle to error either, since we're just reading a static file. We could consider panicing on error and collapsing all of the error handling.

pkg/cloudprovider/fake/instancetype.go
```diff
@@ -105,6 +106,10 @@ func (i *InstanceType) Name() string {
 	return i.options.Name
 }
 
+func (i *InstanceType) AMIFamily() string {
```
ellistarn, line 109: This can be removed.

pkg/cloudprovider/fake/instancetype.go
```diff
@@ -87,6 +87,7 @@ type InstanceTypeOptions struct {
 	Name             string
 	Offerings        []cloudprovider.Offering
 	Architecture     string
+	AMIFamily        string
```
ellistarn, line 90: This can be removed.

ellistarn:
Really nice work so far @rayterrill . A couple of comments and optional suggestions.

pkg/cloudprovider/aws/launchtemplate.go
rayterrill: I left this for another day - Didn't wanna bite off too much at once. :)

pkg/cloudprovider/aws/ami.go
rayterrill: I modified this - but still not sure it's correct. Feel free to correct if I'm still misunderstanding how this should be done - still learning. :)

pkg/cloudprovider/aws/ami.go
```diff
@@ -83,14 +84,29 @@ func (p *AMIProvider) getAMIID(ctx context.Context, query string) (string, error
...
+func (p *AMIProvider) getSSMQuery(ctx context.Context, constraints *v1alpha1.Constraints, instanceType cloudprovider.InstanceType, version string) string {
 	var amiSuffix string
+	var arch string
 	if !instanceType.NvidiaGPUs().IsZero() || !instanceType.AWSNeurons().IsZero() {
 		amiSuffix = "-gpu"
 	} else if instanceType.Architecture() == v1alpha5.ArchitectureArm64 {
 		amiSuffix = "-arm64"
+		arch = "arm64"
+	} else if instanceType.Architecture() == v1alpha5.ArchitectureAmd64 {
+		arch = "arm64"
```
bwagner5, line 96: this should be `x86_64` (https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id-bottlerocket.html#:~:text=aws%2Dk8s%2D1.21/-,x86_64,-/latest/image_id%20%2D%2Dregion)

pkg/cloudprovider/aws/ami.go
```diff
@@ -83,14 +84,29 @@ func (p *AMIProvider) getAMIID(ctx context.Context, query string) (string, error
...
+	}
+
+	if constraints.AMIFamily != nil {
+		if *constraints.AMIFamily == v1alpha1.OperatingSystemBottleRocket {
+			return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, arch)
+		} else if *constraints.AMIFamily == v1alpha1.OperatingSystemEKSOptimized {
+			return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
```
bwagner5, line 103: You could put this format string into a const since it's duplicated 3 times. Wouldn't want to accidentally typo one of those :D

pkg/cloudprovider/aws/apis/v1alpha1/register.go
```diff
@@ -32,6 +32,8 @@ var (
 	AWSRestrictedLabelDomains = []string{
 		"k8s.aws",
 	}
+	OperatingSystemBottleRocket = "BottleRocket"
```
bwagner5, line 35: per the project, it should be one word: `bottlerocket` https://github.com/bottlerocket-os/bottlerocket#bottlerocket-os

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -94,15 +94,26 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, constraints *v1alpha1.
...
 		return nil, err
 	}
 	// Construct launch templates
 	launchTemplates := map[string][]cloudprovider.InstanceType{}
 	for amiID, instanceTypes := range amis {
 		// Get userData for Node
-		userData, err := p.getUserData(ctx, constraints, instanceTypes, additionalLabels)
+		var userData string
+		if constraints.AMIFamily != nil {
+			if *constraints.AMIFamily == v1alpha1.OperatingSystemBottleRocket {
```
bwagner5, line 107: can you use `if strings.EqualFold(*constraints.AMIFaily, v1alpha1.OperatingSystemBottlerocket)` on both of these conditionals for a case-insensitive equals?

bwagner5 (commented):
Awesome work! Just a few comments and hopefully we can get this merged very soon!! :)

pkg/cloudprovider/aws/apis/v1alpha1/register.go
ellistarn: I had some feedback here. It should be upper case since it's a yaml value. `Bottlerocket`.

pkg/cloudprovider/aws/ami.go
ellistarn: This might be useful `v1alpha5.AWSToKubeArchitectures`

bwagner5: ^ `v1alpha1.AWSToKubeArchitectures`

rayterrill: I don't totally understand this - isn't the default from SSM x86-64? What's the purpose of the new arch variable in the suggestion (vs the existing amiSuffix)?

rayterrill: Fixed.

pkg/cloudprovider/aws/apis/v1alpha1/register.go
rayterrill: Fixed - my bad.

pkg/cloudprovider/aws/ami.go
rayterrill: I'm not sure I'm following @bwagner5 - can you give me an example? Sorry mate still learning Go so maybe I'm just missing something simple - not sure how a const would work with interpolated values.

pkg/cloudprovider/aws/launchtemplate.go
rayterrill: TIL!

rayterrill:
Just pushed another commit with I believe the requested changes, minus the one I still have a question about.

pkg/cloudprovider/aws/ami.go
bwagner5: No worries!  

I had meant something like this:

```
const (
  AL2AliasFormat = "/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id"
)

...

fmt.Sprintf(AL2AliasFormat, version, amiSuffix)
```

However, after looking at this again, a better idea is to use a function. The constant alias format string may be confusing since it takes two parameters. Breaking them into distinct functions will hide the complexity of building up the alias, make it easier to test separately, and there's potentially more validation we could do in these funcs in the future.

Can you do something like this?

```
func (p *AMIProvider) getAL2Alias(version string, instanceType cloudprovider.InstanceType) string {
	amiSuffix := ""
	if !instanceType.NvidiaGPUs().IsZero() || !instanceType.AWSNeurons().IsZero() {
		amiSuffix = "-gpu"
	} else if instanceType.Architecture() == v1alpha5.ArchitectureArm64 {
		amiSuffix = fmt.Sprintf("-%s", instanceType.Architecture())
	}
	return fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2%s/recommended/image_id", version, amiSuffix)
}

func (p *AMIProvider) getBottlerocketAlias(version string, instanceType cloudprovider.InstanceType) string {
	arch := "x86_64"
	if instanceType.Architecture() == v1alpha5.ArchitectureArm64 {
		arch = instanceType.Architecture()
	}
	return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, arch)
}
```

This will encapsulate the aliasing logic (like knowing the arch determines the amiSuffix) into a function that returns the properly formatted alias.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -220,10 +231,43 @@ func sortedKeys(m map[string]string) []string {
...
+
+	nodeLabelArgs := functional.UnionStringMaps(additionalLabels, constraints.Labels)
+	if len(nodeLabelArgs) > 0 {
+		userData += `[settings.kubernetes.node-labels]
+`
+		for key, val := range nodeLabelArgs {
+			userData += fmt.Sprintf("\"%s\" = \"%s\"", key, val)
+		}
+	}
+
```
alex-berger, line 263: What about seeting [taints](https://github.com/bottlerocket-os/bottlerocket#kubernetes-settings) here? To make sure we are  on par with what `getEKSOptimizedUserData` offers.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -94,15 +94,26 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, constraints *v1alpha1.
...
 		// Get userData for Node
-		userData, err := p.getUserData(ctx, constraints, instanceTypes, additionalLabels)
+		var userData string
+		if constraints.AMIFamily != nil {
+			if strings.EqualFold(*constraints.AMIFamily, v1alpha1.OperatingSystemBottleRocket) {
+				userData, err = p.getBottleRocketUserData(ctx, constraints, additionalLabels)
+			} else if strings.EqualFold(*constraints.AMIFamily, v1alpha1.OperatingSystemEKSOptimized) {
+				userData, err = p.getEKSOptimizedUserData(ctx, constraints, instanceTypes, additionalLabels)
+			} else {
+				userData, err = p.getEKSOptimizedUserData(ctx, constraints, instanceTypes, additionalLabels)
```
alex-berger, line 112: If `AMIFamily` is non-empty, but contains an invalid value, shouldn't we at least log something here or maybe even return an error? Or maybe even better add proper validation to [provider_validation.go](https://github.com/aws/karpenter/blob/e6bc0e795412573e3fc1057a07f4680c43a84cca/pkg/cloudprovider/aws/apis/v1alpha1/provider_validation.go)?

pkg/cloudprovider/aws/launchtemplate.go
rayterrill: Added a section for this.

pkg/cloudprovider/aws/launchtemplate.go
rayterrill: Added logging here, as well as to the getSSMQuery section in ami.go.

rayterrill:
Just pushed another commit with additional requested changes.

pkg/cloudprovider/aws/ami.go
```diff
@@ -83,16 +84,42 @@ func (p *AMIProvider) getAMIID(ctx context.Context, query string) (string, error
...
+func (p *AMIProvider) getBottlerocketAlias(version string, instanceType cloudprovider.InstanceType) string {
+	arch := "x86_64"
+	if instanceType.Architecture() == v1alpha5.ArchitectureArm64 {
+		arch = instanceType.Architecture()
+	}
+	return fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, arch)
+}
+
+func (p *AMIProvider) getSSMQuery(ctx context.Context, constraints *v1alpha1.Constraints, instanceType cloudprovider.InstanceType, version string) string {
+	ssmQuery := p.getAL2Alias(version, instanceType)
```
suket22, line 108: Could you add documentation changes as part of this PR? We should call out that you will be defaulted to the EKS-optimized AMI.

pkg/cloudprovider/aws/apis/v1alpha1/register.go
```diff
@@ -32,6 +32,8 @@ var (
 	AWSRestrictedLabelDomains = []string{
 		"k8s.aws",
 	}
+	OperatingSystemBottleRocket = "Bottlerocket"
+	OperatingSystemEKSOptimized = "EKSOptimized"
```
suket22, line 36: Sorry, should've mentioned this earlier but can we change this to EKSOptimizedLinux? There's also [EKS Optimized Windows AMIs](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html) so we'd need to disambiguate.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -220,10 +232,52 @@ func sortedKeys(m map[string]string) []string {
 	return keys
 }
 
-// getUserData returns the exact same string for equivalent input,
+func (p *LaunchTemplateProvider) getBottleRocketUserData(ctx context.Context, constraints *v1alpha1.Constraints, additionalLabels map[string]string) (string, error) {
```
suket22, line 235: Could you also support maxPods as that was functionality added in [this PR](https://github.com/aws/karpenter/pull/1032/files). 

For BR, it should be `settings.kubernetes.max-pods`

suket22 (commented):
Most of it looks good :+1:

pkg/cloudprovider/aws/ami.go
suket22: This is optional - we can have this be a separate PR too. Just want to make sure we include it before we do another release :)

pkg/cloudprovider/aws/launchtemplate.go
suket22: This could be a follow-up change too :)

rayterrill:
Rebased.

bwagner5:
Ah looks like CI failed due to codegen.  Should be fixed if you run `make codegen` and commit the aws cloudprovider's `zz_generated_deepcopy.go` which will include `AMIFamily` now.

rayterrill:
Just pushed another commit with that - sorry about that.

bwagner5 (approved):
lgtm! Thanks for working on this! 🚀

akestner:
What's the plan to get this documented?

PR #1123, opened 2022-01-11
Helm chart: use AppVersion in tags

mbevc1:
**1. Issue, if available:**
n/a

Use `AppVersion` for calculating container image tags. Still allows override via `tag` attribute.

Local chart generation.

- [x] No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,8 +76,12 @@ codegen: ## Generate code. Must be run if changes are made to ./pkg/apis/...
 
 publish: ## Generate release manifests and publish a versioned container image.
 	@aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin $(RELEASE_REPO)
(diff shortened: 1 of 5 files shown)


ellistarn:
Take a look at our release process `make helm`

mbevc1:
I'm assuming you are refering to `publish` target? We could probably drop https://github.com/aws/karpenter/blob/main/Makefile#L79-L80 now, unless I'm missing some context? :thinking:

ellistarn:
> I'm assuming you are refering to `publish` target? We could probably drop https://github.com/aws/karpenter/blob/main/Makefile#L79-L80 now, unless I'm missing some context? 🤔

The helm values are autogenerated by `make publish`. Perhaps we can just take the SHA and inject it into the values file.

ellistarn:
Apologies if I'm being dense, but what's the benefit of using AppVersion over the current implementation?

mbevc1:
No need to apologise! Initially seemed lie a good approach to use `appVersion` attribute in image versions, but in retrospective looking into your release process and auto-generating and overriding the values probably doesn't make sense to do that as everything is overridden as part of release.

Thanks for feedback and closing this PR.

ellistarn:
Appreciate the attempt at simplifying! Will keep chart version in mind as this evolves.

PR #1127, opened 2022-01-11
ICE cache bin-packing validation fix

bwagner5:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/1126

Fixes an issue where if an instance type offering is found to be unavailable and therefore placed in the ICE cache, the binpacking logic could still use the instance type in the packing. 

The bug is occurring in the bin packing's packable validation logic. Packables (Instance Type and Offering) are validated by zone and capacity type separately even though an offering is the tuple of (zone, capacity type). This change validates an offering and removes the independent zone and capacity-type validation.

The issue was originally found while doing load tests where I ran into an Insufficient Capacity Error (ICE). After the ICE, Karpenter behaved very strangely, by throwing "Security Group does not belong to the same network as Subnet". This error is because of the overrides bug described in the issue linked. 

- Reproduced in a cluster by injecting ICEs directly into the cache (was able to reproduce and verified the fix)
- Added a unit-test that fails without the patch and passes with the patch.

- [x] No

### DIFF
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -172,7 +172,7 @@ func (p *InstanceProvider) getLaunchTemplateConfigs(ctx context.Context, constra
 		}
 	}
 	if len(launchTemplateConfigs) == 0 {
-		return nil, fmt.Errorf("no capacity offerings are currently available for the instance types and capacity type requested")
+		return nil, fmt.Errorf("no capacity offerings are currently available given the constraints")
 	}
 	return launchTemplateConfigs, nil
 }
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -160,13 +160,19 @@ func (p *InstanceProvider) getLaunchTemplateConfigs(ctx context.Context, constra
 		return nil, fmt.Errorf("getting launch templates, %w", err)
 	}
 	for launchTemplateName, instanceTypes := range launchTemplates {
-		launchTemplateConfigs = append(launchTemplateConfigs, &ec2.FleetLaunchTemplateConfigRequest{
+		launchTemplateConfig := &ec2.FleetLaunchTemplateConfigRequest{
 			Overrides: p.getOverrides(instanceTypes, subnets, constraints.Requirements.Zones(), capacityType),
 			LaunchTemplateSpecification: &ec2.FleetLaunchTemplateSpecificationRequest{
 				LaunchTemplateName: aws.String(launchTemplateName),
 				Version:            aws.String("$Default"),
 			},
-		})
+		}
+		if len(launchTemplateConfig.Overrides) > 0 {
+			launchTemplateConfigs = append(launchTemplateConfigs, launchTemplateConfig)
+		}
+	}
+	if len(launchTemplateConfigs) == 0 {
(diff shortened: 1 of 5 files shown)


pkg/cloudprovider/aws/instance.go
```diff
@@ -160,13 +160,19 @@ func (p *InstanceProvider) getLaunchTemplateConfigs(ctx context.Context, constra
...
-		launchTemplateConfigs = append(launchTemplateConfigs, &ec2.FleetLaunchTemplateConfigRequest{
+		launchTemplateConfig := &ec2.FleetLaunchTemplateConfigRequest{
 			Overrides: p.getOverrides(instanceTypes, subnets, constraints.Requirements.Zones(), capacityType),
 			LaunchTemplateSpecification: &ec2.FleetLaunchTemplateSpecificationRequest{
 				LaunchTemplateName: aws.String(launchTemplateName),
 				Version:            aws.String("$Default"),
 			},
-		})
+		}
+		if len(launchTemplateConfig.Overrides) > 0 {
```
ellistarn, line 170: Yowza what a bug! Makes me realize how brittle this overrides mechanism is.

pkg/cloudprovider/aws/instance.go
```diff
@@ -160,13 +160,19 @@ func (p *InstanceProvider) getLaunchTemplateConfigs(ctx context.Context, constra
...
+			launchTemplateConfigs = append(launchTemplateConfigs, launchTemplateConfig)
+		}
+	}
+	if len(launchTemplateConfigs) == 0 {
+		return nil, fmt.Errorf("no capacity offerings are currently available for the instance types and capacity type requested")
```
ellistarn, line 175:
```suggestion
		return nil, fmt.Errorf("no capacity offerings are currently available given the constraints")
```

pkg/controllers/provisioning/binpacking/packable.go
```diff
@@ -195,26 +193,13 @@ func (p *Packable) validateOperatingSystems(constraints *v1alpha5.Constraints) e
...
-		capacityTypes.Insert(offering.CapacityType)
-	}
-	if constraints.Requirements.CapacityTypes().Intersection(capacityTypes).Len() == 0 {
-		return fmt.Errorf("capacity types %v are not in %v", capacityTypes, constraints.Requirements.CapacityTypes().List())
+		if constraints.Requirements.CapacityTypes().Has(offering.CapacityType) && constraints.Requirements.Zones().Has(offering.Zone) {
+			return nil
+		}
 	}
-	return nil
+	return fmt.Errorf("offerings %v are not available for capacity types %v and zones %v", p.Offerings(), constraints.Requirements.CapacityTypes(), constraints.Requirements.Zones())
```
ellistarn, line 202: Can you show me an example of this string? Seems like it could be really verbose, especially printing offerings.

bwagner5:
```
offerings [{spot test-zone-1a} {on-demand test-zone-1b} {spot test-zone-1b}] are not available for capacity types [on-demand] and zones [test-zone-1a]
```

ellistarn: Nice!

ellistarn (dismissed):

ellistarn (approved):

PR #1151, opened 2022-01-14
Remove subnet and security group selector defaulting.

ellistarn:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/1017

Removed discovery via cluster name to avoid tag limitations for multi-cluster discovery. Instead, users must define their own discovery mechanisms for their use case. Moved cluster-based discovery to the getting started guide.

- [ ] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/register.go
+++ b/pkg/apis/provisioning/v1alpha5/register.go
@@ -40,11 +40,11 @@ var (
 		metav1.AddToGroupVersion(scheme, SchemeGroupVersion)
 		return nil
 	})
-	ProvisionerNameLabelKey         = SchemeGroupVersion.Group + "/provisioner-name"
-	NotReadyTaintKey                = SchemeGroupVersion.Group + "/not-ready"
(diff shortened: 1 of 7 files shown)


website/content/en/docs/getting-started/_index.md
```diff
@@ -236,6 +223,10 @@ spec:
     resources:
       cpu: 1000
   provider:
+    subnetSelector:
```
suket22, line 226: Now that we don't autodiscover, we should be telling customers how to choose a subnetSelector and a securityGroupSelector in our documentation. Maybe a line or two. 
 
We could use EKS as an example in our docs since this is AWS specific so should make sense to most customers. We could reference the clusterWorkerSG and cluster's subnets as good examples to select here 

Also do we [need to edit this para](https://github.com/aws/karpenter/blob/main/website/content/en/docs/getting-started/_index.md?plain=1#L212)?

ellistarn: done

suket22 (approved):

PR #1155, opened 2022-01-15
Refactor apis.provisioning requirements

### LINKED ISSUE #1084: Constraint validation incorrectly excluding daemonset pods in provisioning
## Version

**Karpenter**: v0.5.3

**Kubernetes**: v1.21

## Expected Behavior

Daemonset pods should be included in provisioning process.

## Actual Behavior

Daemonset pods with NotIn node selector were excluded when node constraints do not specify the value of the corresponding key. Further investigation suggests the constraint validation logic is incorrect.

This error also applies to application pods.

felix-zhe-huang:
**1. Issue, if available:**
Karpenter relies on the provisioner API to calculate the requirements and constraints for pod scheduling and resource provisioning. The current implementation logic only handles specify use cases and causes wrong scheduling and resource estimation errors for general use cases (Noticeably issue #1084 that daemonSet pods are not considered during resource provisioning). This patch corrects the requirement logics and relaxes the inflexibility of the current design. It will enable the development of the following features.

- Improve resource utilization by grouping pods according to requirement compatibility
- Support arbitrary requirement labels for topology and affinity features
- Enable more use cases with Exists and DoesNotExist operators

Resolve #1084 

- Refactor the requirement implementation.
- Support Exists and DoesNotExists operator
- Enable constraints compatibility testing
- Correct constraint validation bugs
 
Included unit tests that test the new features.
Manually tested:
- Provisioner requirement with NodeSelector works
- Label with NodeSelector works
- Scaling to 100 pods works
- Provisioner zone requirement with Pod Affinity rules works
- Provisioner selecting instance type works
- Pods with In, NotIn, Exists and DoesNotExist Affinity rules works
 
- [X] No


pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -33,7 +33,7 @@ type Constraints struct {
 	// +optional
 	Taints Taints `json:"taints,omitempty"`
 	// Requirements are layered with Labels and applied to every node.
-	Requirements Requirements `json:"requirements,omitempty"`
+	Requirements *Requirements `json:"requirements,inline,omitempty"`
```
ellistarn, line 36:
```suggestion
	Requirements *Requirements `json:",inline"`
```

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -48,27 +48,22 @@ func (c *Constraints) ValidatePod(pod *v1.Pod) error {
...
+		return fmt.Errorf("pod requirements not feasible, %v", errs)
 	}
-	// The combined requirements are not compatible
-	combined := c.Requirements.Add(podRequirements...)
-	for _, key := range podRequirements.Keys() {
-		if combined.Requirement(key).Len() == 0 {
-			return fmt.Errorf("invalid nodeSelector %q, %v not in %v", key, podRequirements.Requirement(key).UnsortedList(), c.Requirements.Requirement(key).UnsortedList())
-		}
+	// Test if pod requirements are compatible
+	if errs := c.Requirements.Compatible(podRequirements); errs != nil {
```
ellistarn, line 57: Love how clean this is

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -48,27 +48,22 @@ func (c *Constraints) ValidatePod(pod *v1.Pod) error {
...
+		return fmt.Errorf("incompatible  nodeSelectors %v", errs)
```
ellistarn, line 58:
```suggestion
		return fmt.Errorf("incompatible requirements %w", err)
```

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -133,29 +132,18 @@ func (c *Constraints) validateTaints() (errs *apis.FieldError) {
 	return errs
 }
 
+// This function is used by the provisioner validation webhook to verify the provisioner requirements.
+// When this function is called, the provisioner's requirments do not include the requirements from labels.
+// Provisioner requirement section only support wellknown labels for now.
```
ellistarn, line 137:
```suggestion
// Provisioner requirements only support well known labels.
```

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -36,21 +42,21 @@ var (
 	// AllowedLabelDomains are domains that may be restricted, but that is allowed because
 	// they are not used in a context where they may be passed as argument to kubelet.
 	// AllowedLabelDomains are evaluated before RestrictedLabelDomains
-	AllowedLabelDomains = sets.NewString(
+	AllowedLabelDomains = stringsets.NewString(
```
ellistarn, line 45: Why not use the set object you built here? Does it have the same interface?

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,244 @@ var (
...
-	return r.Requirement(v1.LabelTopologyZone)
-}
-
-func (r Requirements) InstanceTypes() sets.String {
-	return r.Requirement(v1.LabelInstanceTypeStable)
+type Requirements struct {
+	// Requirements are layered with Labels and applied to every node.
+	Requirements []v1.NodeSelectorRequirement `json:"requirements,omitempty"`
+	// Ground truth for record keeping
+	Records map[string][]v1.NodeSelectorRequirement `json:"-"`
```
ellistarn, line 83: It's not clear to me why we need this instead of just Requirements?

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,244 @@ var (
...
+		}
+	}
+	return rd
 }
 
-func (r Requirements) CapacityTypes() sets.String {
-	return r.Requirement(LabelCapacityType)
+func (r *Requirements) MarshalJSON() ([]byte, error) {
+	var result []v1.NodeSelectorRequirement
+	for _, requirements := range r.Records {
```
ellistarn, line 119: Can't we do this via r.Requirements?

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,244 @@ var (
...
+			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("no feasible value for requirement, %s", key), "values"))
+		}
+		i := 0
+		for value := range values.Members {
+			for _, err := range validation.IsValidLabelValue(value) {
+				errs = errs.Also(apis.ErrInvalidArrayValue(fmt.Sprintf("%s, %s", value, err), "values", i))
+			}
+			i++
+		}
+		for _, requirement := range r.Records[key] {
```
ellistarn, line 167: Can't we use r.Requirements (here and everywhere else)

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,244 @@ var (
...
+		case v1.NodeSelectorOpNotIn:
+			rd.Allows[requirement.Key] = rd.Allow(requirement.Key).Intersection(sets.NewComplementSet(requirement.Values...))
+		}
+	}
+	return rd
 }
 
-func (r Requirements) CapacityTypes() sets.String {
-	return r.Requirement(LabelCapacityType)
+func (r *Requirements) MarshalJSON() ([]byte, error) {
```
ellistarn, line 117: I think you only needed to do this because of the bug I pointed out here: https://github.com/aws/karpenter/pull/1155/files#r791073256

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,244 @@ var (
...
+	return result.Add(requirements...)
 }
 
-func (r Requirements) OperatingSystems() sets.String {
-	return r.Requirement(v1.LabelOSStable)
+func (r *Requirements) Add(requirements ...v1.NodeSelectorRequirement) *Requirements {
+	if r == nil {
+		return NewRequirements(requirements...)
+	}
+	rd := r.DeepCopy()
```
ellistarn, line 99: I like this because it's immutable (similar to go slice style)

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,244 @@ var (
...
+	for _, requirement := range requirements {
+		// Normalizing the key
```
ellistarn, line 101:
```suggestion
```

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,244 @@ var (
...
+		if newKey, ok := NormalizedLabels[requirement.Key]; ok {
+			requirement.Key = newKey
+		}
+		//rd.Requirements = append(rd.Requirements, requirement)
```
ellistarn, line 105: Don't forget about this guy.

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,244 @@ var (
...
-		})
+		for _, err := range validation.IsQualifiedName(key) {
+			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("%s, %s", key, err), "key"))
+		}
+		values := r.Allow(key)
+		if values.Len() == 0 {
+			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("no feasible value for requirement, %s", key), "values"))
+		}
+		i := 0
+		for value := range values.Members {
```
ellistarn, line 161: consider using something like `i, value := range values.Members.UnsortedList()`

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: Tried and it doesn't work without those two tags.
Without requirements, there is an extra `Requirements: []v1.NodeSelectorRequirement` and any future changes to the requirement objects appears under key `Requirements`.
The omitempty is needed in case for Provisioner.Spec without requirements. Without it, there will be nil pointer errors.

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: :)

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: done

felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: Actually after our discussion I decided to remove most of the unused interface functions from the set object I created. Limiting the scope of the set to only requirement calculation allows us to avoid dealing with quite a few corner cases. It simplify the codes by quite a lot.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: In this commit `Requirements []v1.NodeSelectorRequirement` is not used at all other than API codegen.
I chose to use `Records map[string][]v1.NodeSelectorRequirement` to simplify the codes by geting ride of all indexing for loops. It is also faster.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: currently r.Requirements is just a nil pointer. I never populate it so it contains no data. This way we save some memory.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: please refer to my comments for the previous two questions.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: I was hoping there is a simple explanation. But the way I discovered this special combinations of tags and MarshalJSON functions that finally works can confirm that MarshalJSON is really needed.

So with just a simple `json:",inline"`, I noticed that data sent to the API server is actually an object instead of a slice. That's when I think for some reasons the build-in Marshaling function failed to understand the `json:",inline"` tag structure. Once I added the MarhsalJSON and UnmarsalJSON functions, I noticed that in the API call there are two slices, one under `Requirements` and one under `requirements`. Strange, but then I tried to force it by adding the `json:"requirements,inline". At least now in most of the test cases where requirements with none-nil contents works.

Then I encounter nil pointer error. That's when I found that for empty requirements, the API call still include `requirements` key. That's when I added the `json:"requirements,inline,omitempty"`.

Very strange case. I spent two days stretching my head trying to figure out why the behavior and settle with this abomination.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: This is actually critical. Otherwise you will see all sorts of hashing strangeness. 
Good suggestion. Done.

pkg/apis/provisioning/v1alpha5/constraints.go
ellistarn: Can we make it a non-pointer? Let's pair program it.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: I am removing that line. I used it for debugging.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: So we should index the errors using the requirement index instead?
Since the requirement index will keep changing for UnsortedList(), doesn't that make the index inconsistent?

Also are we allowed to have a gap between error indexes?

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: Yeah I thought about it. Then I realize initialization will become painful.

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,245 @@ var (
...
-	// OpIn
-	for _, requirement := range r {
-		if requirement.Key == key && requirement.Operator == v1.NodeSelectorOpIn {
-			if result == nil {
-				result = sets.NewString(requirement.Values...)
-			} else {
-				result = result.Intersection(sets.NewString(requirement.Values...))
+// isKeyDefined returns errors when the given requirements requires a key that is not defined by this requirements
+// This is a non-commutative function designed to catch the following cases:
+// * Pod Spec has a label selector but provisioner doesn't have the label
```
bwagner5, line 236: This case is permitted right?  For example, if my provisioner does not specify a `karpenter.sh/capacity-type` requirement, then it is unconstrained, meaning that if a pod has a nodeSelector for `spot`, then we should successfully validate the Pod's requirements with the provisioner.

charts/karpenter/crds/karpenter.sh_provisioners.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.7.0
+    controller-gen.kubebuilder.io/version: v0.4.1
```
ellistarn, line 7: Is your controller gen version out of date?

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -79,15 +79,13 @@ var (
 type Requirements struct {
 	// Requirements are layered with Labels and applied to every node.
 	Requirements []v1.NodeSelectorRequirement `json:"requirements,omitempty"`
-	// Ground truth for record keeping
-	Records map[string][]v1.NodeSelectorRequirement `json:"-"`
-	Allows  map[string]sets.Set                     `json:"-"`
+	Allows       map[string]sets.Set          `json:"-"`
```
ellistarn, line 82: One trick you could use is to call this 
```suggestion
	requirements map[string]sets.Set          `json:"-"`
```
which implies that it's the same thing in a more compressed form.

pkg/utils/sets/sets.go
```diff
@@ -56,11 +56,11 @@ func (s Set) DeepCopy() Set {
 
 // Values returns the members of the set.
 // If the set has an infinite size, returns an error
```
ellistarn, line 58: This is in accurate, now

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: I actually had the same confusion a while a go and finally it hit me that provisioner labels will be bypassed in this situation since a pod with a node-selector can be validated to any provisioner with or without a proper label value.

Logically, we want the provisioner requirements to be tight (i.e., if it is not defined, it is not allowed) but the pod requirements to be as broad as possible (e.g., if not defined, pod doesn't care). This is to make sure that the provisioners behave according to what is defined, but not to what is not defined. This also allow us to always pick the tightest fix provisioner in selection.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -48,27 +48,22 @@ func (c *Constraints) ValidatePod(pod *v1.Pod) error {
...
 		return err
 	}
-	// The constraints do not support this requirement
-	podRequirements := PodRequirements(pod)
-	for _, key := range podRequirements.Keys() {
-		if c.Requirements.Requirement(key).Len() == 0 {
-			return fmt.Errorf("invalid nodeSelector %q, %v not in %v", key, podRequirements.Requirement(key).UnsortedList(), c.Requirements.Requirement(key).UnsortedList())
-		}
+	// Test if pod requirements are valid
+	podRequirements := NewRequirements(PodRequirements(pod)...)
```
ellistarn, line 52: 🤔 how could the pod requirements be invalid? We should have validated them before we even get to provisioning.

ellistarn: I think we probably want to do this out in the selection controller.

ellistarn: What does it look like for the pod's requirements to be invalid, anyways? Is it possible for them to be invalid, but the compatibility check to succeed?

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -133,29 +132,18 @@ func (c *Constraints) validateTaints() (errs *apis.FieldError) {
 	return errs
 }
 
+// This function is used by the provisioner validation webhook to verify the provisioner requirements.
+// When this function is called, the provisioner's requirments do not include the requirements from labels.
+// Provisioner requirements only support well known labels.
```
ellistarn, line 137: It's possible to include well known labels in provisioner.labels. I think we need to combine these for validation.

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,125 +76,245 @@ var (
...
-func (r Requirements) Architectures() sets.String {
-	return r.Requirement(v1.LabelArchStable)
+// Add function returns a new Requirements object with new requirements inserted.
+// It is critical to keep the original Requirements object immutable to keep the
+// requirement hashing behavior consistent.
+func (r *Requirements) Add(requirements ...v1.NodeSelectorRequirement) *Requirements {
+	if r == nil {
+		return NewRequirements(requirements...)
+	}
+	rd := r.DeepCopy()
```
ellistarn, line 100: optional: use r = r.DeepCopy() to keep the name the same.

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,125 +76,245 @@ var (
...
+	return result.Add(requirements...)
 }
 
-func (r Requirements) Architectures() sets.String {
-	return r.Requirement(v1.LabelArchStable)
+// Add function returns a new Requirements object with new requirements inserted.
+// It is critical to keep the original Requirements object immutable to keep the
+// requirement hashing behavior consistent.
+func (r *Requirements) Add(requirements ...v1.NodeSelectorRequirement) *Requirements {
+	if r == nil {
```
ellistarn, line 97: when are we calling add on nil requirements?

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,125 +76,245 @@ var (
...
+func (r *Requirements) MarshalJSON() ([]byte, error) {
+	return json.Marshal(r.Requirements)
 }
 
-func (r Requirements) CapacityTypes() sets.String {
-	return r.Requirement(LabelCapacityType)
+func (r *Requirements) UnmarshalJSON(b []byte) error {
+	var requirements []v1.NodeSelectorRequirement
+	json.Unmarshal(b, &requirements)
+	result := NewRequirements(requirements...)
```
ellistarn, line 123: r = NewRequirements?

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,213 @@ var (
...
+	for i, requirement := range r.Requirements {
+		for _, err := range validation.IsQualifiedName(requirement.Key) {
+			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("%s, %s", requirement.Key, err), "key"))
+		}
+		for _, value := range requirement.Values {
+			for _, err := range validation.IsValidLabelValue(value) {
+				errs = errs.Also(apis.ErrInvalidArrayValue(fmt.Sprintf("%s, %s", value, err), "values", i))
+			}
+		}
+		if !functional.ContainsString(SupportedNodeSelectorOps, string(requirement.Operator)) {
```
ellistarn, line 206: Can we make this set.String{}.Has()

ellistarn: We can also deprecate functional.IntersectStringSlice

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,213 @@ var (
...
+		// Excludes cases when DoesNotExists appears together with In, NotIn, Exists
+		if requirement.Operator == v1.NodeSelectorOpDoesNotExist && (r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpIn)) ||
+			r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpNotIn)) ||
+			r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpExists))) {
+			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("operator %s and %s conflict", v1.NodeSelectorOpDoesNotExist, v1.NodeSelectorOpDoesNotExist), "operator"))
+		}
+	}
+	for key := range r.Keys() {
+		values := r.Values(key)
+		if values.Len() == 0 {
```
ellistarn, line 218: Can we inline r.Values(key)

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,213 @@ var (
...
+			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("no feasible value for requirement, %s", key), "values"))
```
ellistarn, line 219:
```suggestion
			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("no feasible value for %s", key), "values"))
```

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,213 @@ var (
...
+			}
+		}
+		if !functional.ContainsString(SupportedNodeSelectorOps, string(requirement.Operator)) {
+			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("%s not in %s", requirement.Operator, SupportedNodeSelectorOps), "operator"))
+		}
+		// Excludes cases when DoesNotExists appears together with In, NotIn, Exists
+		if requirement.Operator == v1.NodeSelectorOpDoesNotExist && (r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpIn)) ||
+			r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpNotIn)) ||
+			r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpExists))) {
+			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("operator %s and %s conflict", v1.NodeSelectorOpDoesNotExist, v1.NodeSelectorOpDoesNotExist), "operator"))
```
ellistarn, line 213:
```suggestion
			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("operator %s and %s conflict for %s", v1.NodeSelectorOpDoesNotExist, v1.NodeSelectorOpDoesNotExist, requirement.Key), "operator"))
```

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,213 @@ var (
...
+		// Key must be defined if required
+		if values := requirements.Values(key); values.Len() != 0 && !values.Complement && !r.hasRequirement(withKey(key)) {
+			errs = errs.Also(apis.ErrInvalidValue("is not defined", "key")).ViaFieldIndex("requirements", i)
+		}
+		// Values must overlap
+		if values := r.Values(key); values.Intersection(requirements.Values(key)).Len() == 0 {
+			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("%s not in %s", values, requirements.Values(key)), "values")).ViaFieldIndex("requirements", i)
+		}
+
+		for r1, r2 := range map[*Requirements]*Requirements{
```
ellistarn, line 238: consider unrolling

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,126 +76,213 @@ var (
...
+					errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("%s prohibits %s", v1.NodeSelectorOpExists, v1.NodeSelectorOpDoesNotExist), "operator")).ViaFieldIndex("requirements", i)
+				}
+			}
+			// DoesNotExist requires DoesNotExist or undefined
+			if r1.hasRequirement(withKeyAndOperator(key, v1.NodeSelectorOpDoesNotExist)) {
+				if !(r2.hasRequirement(withKeyAndOperator(key, v1.NodeSelectorOpDoesNotExist)) || !r2.hasRequirement(withKey(key))) {
+					errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("%s requires %s", v1.NodeSelectorOpDoesNotExist, v1.NodeSelectorOpDoesNotExist), "operator")).ViaFieldIndex("requirements", i)
+				}
 			}
+
```
ellistarn, line 254:
```suggestion
```

pkg/apis/provisioning/v1alpha5/suite_test.go
```diff
@@ -140,29 +131,206 @@ var _ = Describe("Validation", func() {
...
 		It("should fail for unknown labels", func() {
 			for label := range sets.NewString("unknown", "invalid", "rejected") {
-				provisioner.Spec.Requirements = Requirements{{Key: label, Operator: v1.NodeSelectorOpIn, Values: []string{"test"}}}
+				provisioner.Spec.Requirements = NewRequirements(
+					v1.NodeSelectorRequirement{Key: label, Operator: v1.NodeSelectorOpIn, Values: []string{"test"}},
+				)
 				Expect(provisioner.Validate(ctx)).ToNot(Succeed())
 			}
 		})
+		It("should fail because no feasible value", func() {
```
ellistarn, line 166: Let's separate validation and compatibility tests

pkg/apis/provisioning/v1alpha5/suite_test.go
```diff
@@ -140,29 +131,206 @@ var _ = Describe("Validation", func() {
...
+			A := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpNotIn, Values: []string{"test", "foo"}})
+			B := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpExists})
+			Expect(A.Compatible(B)).To(Succeed())
+		})
+		It("A should fail to be compatible to B, <NotIn, DoesNotExist> operator, conflicting", func() {
+			A := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpNotIn, Values: []string{"test", "foo"}})
+			B := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpDoesNotExist})
+			Expect(A.Compatible(B)).ToNot(Succeed())
+		})
+
```
ellistarn, line 254:
```suggestion
```

pkg/controllers/provisioning/binpacking/packable.go
```diff
@@ -174,21 +174,21 @@ func (p *Packable) reservePod(pod *v1.Pod) bool {
...
-		return fmt.Errorf("instance type %s not in %v", p.Name(), constraints.Requirements.InstanceTypes().List())
+		return fmt.Errorf("instance type %s not in %s", p.Name(), constraints.Requirements.InstanceTypes())
 	}
 	return nil
 }
 
 func (p *Packable) validateArchitecture(constraints *v1alpha5.Constraints) error {
 	if !constraints.Requirements.Architectures().Has(p.Architecture()) {
-		return fmt.Errorf("architecture %s is not in %v", p.Architecture(), constraints.Requirements.Architectures().List())
+		return fmt.Errorf("architecture %s not in %s", p.Name(), constraints.Requirements.Architectures())
```
ellistarn, line 184: consider using %v in all these cases

pkg/controllers/provisioning/controller.go
```diff
@@ -100,8 +101,10 @@ func (c *Controller) Apply(ctx context.Context, provisioner *v1alpha5.Provisione
 	provisioner.Spec.Labels = functional.UnionStringMaps(provisioner.Spec.Labels, map[string]string{v1alpha5.ProvisionerNameLabelKey: provisioner.Name})
 	provisioner.Spec.Requirements = provisioner.Spec.Requirements.
 		Add(requirements(instanceTypes)...).
-		Add(v1alpha5.LabelRequirements(provisioner.Spec.Labels)...).
-		Consolidate()
+		Add(v1alpha5.NewLabelRequirements(provisioner.Spec.Labels).Requirements...)
+	if err := provisioner.Spec.Requirements.Validate(); err != nil {
+		return fmt.Errorf("provisioner requirements validation failed, %v", err)
```
ellistarn, line 106:
```suggestion
		return fmt.Errorf("invalid requirements, %w", err)
```

pkg/controllers/provisioning/controller.go
```diff
@@ -155,10 +158,12 @@ func requirements(instanceTypes []cloudprovider.InstanceType) (requirements v1al
 		supported[v1.LabelArchStable].Insert(instanceType.Architecture())
 		supported[v1.LabelOSStable].Insert(instanceType.OperatingSystems().List()...)
 	}
+	requirements := []v1.NodeSelectorRequirement{}
 	for key, values := range supported {
 		requirements = append(requirements, v1.NodeSelectorRequirement{Key: key, Operator: v1.NodeSelectorOpIn, Values: values.UnsortedList()})
 	}
 	return requirements
+
```
ellistarn, line 166:
```suggestion
```

pkg/controllers/selection/controller.go
```diff
@@ -95,6 +97,7 @@ func (c *Controller) selectProvisioner(ctx context.Context, pod *v1.Pod) (errs e
 		return nil
 	}
 	for _, candidate := range c.provisioners.List(ctx) {
+
```
ellistarn, line 100:
```suggestion
```

pkg/controllers/selection/controller.go
```diff
@@ -165,7 +168,7 @@ func validateNodeSelectorTerm(term v1.NodeSelectorTerm) (errs error) {
 	}
 	if term.MatchExpressions != nil {
 		for _, requirement := range term.MatchExpressions {
-			if !sets.NewString(string(v1.NodeSelectorOpIn), string(v1.NodeSelectorOpNotIn)).Has(string(requirement.Operator)) {
+			if !functional.ContainsString(v1alpha5.SupportedNodeSelectorOps, string(requirement.Operator)) {
```
ellistarn, line 171: Can we use a set here?

pkg/utils/sets/suite_test.go
```diff
@@ -0,0 +1,81 @@
...
+		It("A intersect with full set should return itself", func() {
+			Expect(setA.Intersection(fullSet)).To(Equal(setA))
+		})
+		It("A intersect with empty set should return empty", func() {
+			Expect(setA.Intersection(emptySet)).To(Equal(emptySet))
+		})
+	})
+
+	Context("Functional Correctness", func() {
+
```
ellistarn, line 61:
```suggestion
```

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/suite_test.go
felix-zhe-huang: done

pkg/controllers/provisioning/controller.go
felix-zhe-huang: done

pkg/controllers/selection/controller.go
felix-zhe-huang: done

ellistarn:
Interesting CI errors:
>  Message: "Provisioner.karpenter.sh \"condorpaint\" is invalid: spec.requirements: Invalid value: \"null\": spec.requirements in body must be of type array: \"null\"",

felix-zhe-huang:
> Interesting CI errors:

This is indeed a very interesting case caused by the init problem we discussed about. Turns out that when a constraints object is created without properly init the requirements, since requirements is not null (no longer a pointer), it will not be ignored (tag ignoreempty no longer work). So when the constraints object is serialized, the requirements.Requirements is a null. That's causing the problem. I managed to patch this by init the requirments.Requirments in the MarshalJSON function. I am hesitate to call this a fix because it may happen if other functions are called.

ellistarn (commented):
LGTM! Worth getting some eyes from @bwagner5. I'd also like to hear about some manual testing. Can you do thorough manual testing against a real cluster and update the description.

felix-zhe-huang:
> Can you do thorough manual testing against a real cluster and update the description.

Did a brunch of manual tests and it works fabulously. Please see the PR descriptions about the manual tests performed.

ellistarn (approved):

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -51,30 +51,22 @@ func (c *Constraints) ValidatePod(pod *v1.Pod) error {
 	if err := c.Taints.Tolerates(pod); err != nil {
 		return err
 	}
-	// The constraints do not support this requirement
-	podRequirements := PodRequirements(pod)
-	for _, key := range podRequirements.Keys() {
-		if IgnoredLabels.Has(key) {
```
ellistarn, line 57: Did you reimplement this for the new requirements

ellistarn (changes requested):

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: This ignore label may change how we implement the validation. I need to get a better understanding the reason before I can implement it.

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -133,29 +133,22 @@ func (c *Constraints) validateTaints() (errs *apis.FieldError) {
...
 func (c *Constraints) validateRequirements() (errs *apis.FieldError) {
-	for i, requirement := range c.Requirements {
-		if err := validateRequirement(requirement); err != nil {
-			errs = errs.Also(apis.ErrInvalidArrayValue(err, "requirements", i))
+	for _, requirement := range c.Requirements.Requirements {
+		// Ensure requirements are well known
+		if !WellKnownLabels.Has(requirement.Key) {
```
bwagner5, line 142: To support `IgnoredLabels` I think you just need another conditional here to allow IgnoredLables that are not "well-known".  IgnoredLabels are just labels that should not affect scheduling. For example, the in-tree EBS driver attached a region label to pods that request an EBS volume. However, Karpenter doesn't support cross-region provisioning, so we can safely ignore the region label.

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -68,133 +73,222 @@ var (
 		"beta.kubernetes.io/os":       v1.LabelOSStable,
 		v1.LabelInstanceType:          v1.LabelInstanceTypeStable,
 	}
-	// IgnoredLables are not considered in scheduling decisions
-	// and prevent validation errors when specified
-	IgnoredLabels = sets.NewString(
```
bwagner5, line 73: see other comment about why this IgnoredLabels are required

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -68,133 +73,222 @@ var (
...
-		if requirement.Key == key && requirement.Operator == v1.NodeSelectorOpIn {
-			if result == nil {
-				result = sets.NewString(requirement.Values...)
-			} else {
-				result = result.Intersection(sets.NewString(requirement.Values...))
+// Compatible ensures the provided requirements can be met. It is
+// non-commutative (i.e., A.Compatible(B) != B.Compatible(A))
+//gocyclo:ignore
+func (r Requirements) Compatible(requirements Requirements) (errs *apis.FieldError) {
+	for i, key := range r.Keys().Union(requirements.Keys()).UnsortedList() {
```
bwagner5, line 225: I guess this will also need to be updated to support IgnoredLabels as well?

pkg/utils/sets/sets.go
```diff
@@ -0,0 +1,105 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// Set is a logical set of string values for the requirements.
+// It supports representations using complement operator.
+// e.g., if C={"A", "B"}, setting complement = true means
+// C' contains every possible string values other than "A" and "B"
+type Set struct {
+	values     sets.String
+	Complement bool
```
bwagner5, line 30: does this need to be exported? Would you ever create a set and then configure it to be a complement set?

pkg/utils/sets/sets.go
```diff
@@ -0,0 +1,105 @@
...
+			s.values = s.values.Intersection(set.values)
+		}
+	}
+	return s
+}
+
+// Len returns the size of the set.
+func (s Set) Len() int {
+	if s.Complement {
+		return math.MaxInt64 - s.values.Len()
```
bwagner5, line 102: I'm not sure how you'd reason about a complement's set size other than infinity? There's no case where a complement set would not be infinity right?  Should this return `-1` or something instead so at least it could be checked...    With this implementation, you could have two sets, a complement set and a regular set, that have the same Len() but one is infinite and one is not.

bwagner5 (changes requested):
Nice work! Huge change!  The only critical thing to address is the `IgnoredLabels` since this will break the EBS in-tree driver.

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: Had an offline discussion with Brandon. Now the igore labels are excluded completely when we create the requriements.

pkg/utils/sets/sets.go
felix-zhe-huang: Good catch! Done!

pkg/utils/sets/sets.go
felix-zhe-huang: After thinking about this for a while I decided to keep this implement because this allows us to compare size of the logical set. Even all the complement sets have infinite size, the can still be ordered by the number of the excluded elements. This will be useful when we need to compare the tightness of requirement fits in the future.

I was also thinking about using negative value to represents the excluded number but then the ordering logic will be flipped for the complement set which will also increase the complexity.

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -70,131 +75,228 @@ var (
...
+	requirements := []v1.NodeSelectorRequirement{}
 	for key, value := range pod.Spec.NodeSelector {
-		r = r.Add(v1.NodeSelectorRequirement{Key: key, Operator: v1.NodeSelectorOpIn, Values: []string{value}})
+		requirements = append(requirements, v1.NodeSelectorRequirement{Key: key, Operator: v1.NodeSelectorOpIn, Values: []string{value}})
 	}
 	if pod.Spec.Affinity == nil || pod.Spec.Affinity.NodeAffinity == nil {
-		return r
+		return NewRequirements(requirements...)
 	}
+	// The legal operators for pod affinity and anti-affinity are In, NotIn, Exists, DoesNotExist.
```
ellistarn, line 115: This comment is inaccurate. This is not pod affinity. This is node affinity. There is no node antiaffinity.  I think you can get away with not having any comment at all.
```suggestion
```

pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
```diff
@@ -32,10 +31,10 @@ func (c *Constraints) defaultCapacityTypes() {
 	if _, ok := c.Labels[v1alpha5.LabelCapacityType]; ok {
 		return
 	}
-	if functional.ContainsString(c.Requirements.Keys(), v1alpha5.LabelCapacityType) {
+	if c.Requirements.Keys().Has(v1alpha5.LabelCapacityType) {
```
ellistarn, line 34: Can we use c.Requirements.Get(v1alpha5.LabelCapacityType).Len()?

ellistarn (approved):

bwagner5 (approved):
lgtm! nice job!

cdenneen:
Can't wait to try it!

Issue #1183, opened 2022-01-19
Official documentation for fallback from spot to on-demand configuration.


Miszcz-AndY:
**What karpenter features are relevant?**
Spot, on-demand.

**How should the docs be improved?**

I know there is no native karpenter solution to fallback to on-demand instances if spot is not available yet and you are working on [that](https://github.com/aws/karpenter/issues/714). **If it's true, that we can configure karpenter in this way** (within fallback to on-demand if spot is not available (no spot capacity / quota is reached / other)) **will be super to see how to do that in the official karpenter information (as a 'fallback to on-demand' section').** 

I can understand that it's not a karpenter feature but rather pod configuration but if you can describe it (e.g. within some example of provisioner configuration and pod configuration (within nodeAffinity / preference section ?)) it can be really helpful.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

Miszcz-AndY:
🤔 I can see documentation is changing dynamically (so cool and tx a lot!) 😄 I can see there is such section now: https://karpenter.sh/v0.5.3/tasks/running-pods/#node-affinity-nodeaffinity and I think it's mostly what I was asking for. 

As there is no information directly about fallback to on-demand I don't close this gentle request and want to ask you to decide if you want to close it (as actual documentation is good enough for spot vs ondemand case) or you want to back to that later and add such direct example.

prateekgogia:
We can revisit this once we have the fallback to OD support available in Karpenter. 
I guess at the moment you found what you were looking for :)
I will close this issue for now :)

ellistarn:
We need to document this more explicitly. If your provisioner supports capacitytype=[spot, on-demand], karpenter will prioritize spot, and launch on demand if every instance type is out of capacity. Leaving this open as a docs issue.

bitshop:
The link above is no longer works, however this link currently has the documentation: https://karpenter.sh/docs/concepts/nodepools/#capacity-type

jonathan-innis:
Thanks @bitshop for helping point anyone who comes here to the right place 🎉

PR #1204, opened 2022-01-22
add desc of capacity fallback

geoffcline:
**1. Issue, if available:**
#1183 

add description of what happens if spot is unavailable 

netlify

- [ ] No


website/content/en/preview/faq.md
```diff
@@ -95,6 +95,12 @@ If the instance type is unavailable for some reason, then fleet will move on to
 If you are using the spot capacity type, Karpenter uses the capacity-optimized-prioritized allocation strategy which tells fleet to find the instance type that EC2 has the most capacity of which will decrease the probability of a spot interruption happening in the near term.
 See [Choose the appropriate allocation strategy](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-fleet-allocation-strategy.html#ec2-fleet-allocation-use-cases) for information on fleet optimization.
 
+### What if there is no spot capacity? Will Karpenter fallback to on-demand?
+
+Karpenter will fallback to on-demand, if your provisioner specifies both spot and on-demand. 
+
+More specifically, Karpenter maintains a concept of "offerings" which is a combination of instance type, zone, and purchase type. Spot offerings are prioritized, if they're available. Whenever the fleet API returns an insufficient capacity for spot, that offering is removed from the list for 5 minutes. Since fleet gives expressive responses about which offerings are unavilable, it results in at worst 2 API calls. The retry will happen immediately within milliseconds.
```
eptiger, line 102:
```suggestion
More specifically, Karpenter maintains a concept of "offerings" for each instance type, which is a combination of zone and capacity type (equivalent in the AWS cloud provider to an EC2 purchase option). Spot offerings are prioritized, if they're available. Whenever the Fleet API returns an insufficient capacity error for Spot instances, those particular offerings are temporarily removed from consideration (across the entire provisioner) for 45 seconds so that Karpenter can make forward progress through fallback. The retry will happen immediately within milliseconds.
```

> Since Fleet gives expressive responses about which offerings are unavailable, it results in at worst 2 API calls for that initial failure. 

I think this is confusing language because I'm not sure that it's true. First of all, you're talking about the Fleet API but there are other EC2 APIs called by Fleet. More importantly, I don't really know what this means because if we retry and still get ICEd on OD we'll re-compute the bin packing and try again. Hence, it could be several Fleet API calls. Karpenter does an admirable job though of trying to keep the the number of round trips to Fleet very small though and should often succeed in this effort.

website/content/en/preview/provisioner.md
```diff
@@ -139,11 +139,13 @@ Karpenter supports `amd64` nodes, and `arm64` nodes.
...
 
 - values
-  - `spot` (default)
-  - `on-demand` 
+  - `spot` 
+  - `on-demand` (default)
 
 Karpenter supports specifying capacity type, which is analogous to [EC2 purchase options](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-purchasing-options.html).
 
+Karpenter prioritizes spot offerings. If the provider API (e.g., EC2 fleet API) indicates spot capacity is unavailable, Karpenter caches that result for five minutes. Next, Karpenter will provision on-demand instances, generally within milliseconds.
```
eptiger, line 147:
```suggestion
Karpenter prioritizes Spot offerings if the provisioner allows Spot and on-demand instances. If the provider API (e.g. EC2 Fleet's API) indicates Spot capacity is unavailable, Karpenter caches that result across all attempts to provision EC2 capacity for that instance type and zone for the next 45 seconds. If there are no other possible offerings available for Spot, Karpenter will attempt to provision on-demand instances, generally within milliseconds. 
```

website/content/en/preview/faq.md
bwagner5: I'd remove the specific time we cache for since that's an implementation detail that may be changed I'd also remove the last sentence on the retry happening immediately within milliseconds. 

Since pods are separated by scheduling constraints, if the first group of scheduling fails due to Insufficient Capacity, then the next group will still attempt to provision nodes and will likely take longer than a few milliseconds. Even if there is no other scheduling group, since pods are batched at the start of the provisioning loop, the next run will take at least 1 second since that is the minimum pod batching window duration.

website/content/en/preview/provisioner.md
bwagner5: see the comment above on removing timings

ellistarn (approved):

PR #1221, opened 2022-01-25
[Experimental] Implemented Hostname AntiAffinity

ellistarn:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/942

Implements hostname antiaffinity by translating these rules into topology rules.

Requested feedback:
- Are the limitations going to work for users? Do we need a fully fledged implementation?
- Am I missing edge cases where my assumptions don't hold?
- Is anyone able to build this from HEAD and test it out?

- [ ] No

### DIFF
--- a/pkg/controllers/selection/antiaffinity.go
+++ b/pkg/controllers/selection/antiaffinity.go
@@ -34,17 +34,19 @@ type AntiAffinity struct{}
 // Validate that the affinity terms are supported
 func (a *AntiAffinity) Validate(pod *v1.Pod) (errs *apis.FieldError) {
 	for i, term := range a.termsFor(pod) {
-		errs = errs.Also(a.validateTerm(term).ViaIndex(i))
+		errs = errs.Also(a.validateTerm(pod, term).ViaIndex(i))
 	}
 	return errs
 }
 
-func (a *AntiAffinity) validateTerm(term v1.PodAffinityTerm) (errs *apis.FieldError) {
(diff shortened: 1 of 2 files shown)


pkg/controllers/selection/antiaffinity.go
```diff
@@ -0,0 +1,79 @@
...
+	if len(term.Namespaces) != 0 {
+		errs = errs.Also(apis.ErrDisallowedFields("namespaces"))
+	}
+	if !AllowedAntiAffinityKeys.Has(term.TopologyKey) {
+		errs = errs.Also(apis.ErrInvalidKeyName(fmt.Sprintf("%s not in %v", term.TopologyKey, AllowedAntiAffinityKeys.UnsortedList()), "topologyKey"))
+	}
+	return errs
+}
+
+// Transform pod anti affinity into topology rules
```
akestner, line 55: Oh this is interesting. Is this how this works under the hood in K8s as well or a new approach that gets us there quicker in Karpenter?

ellistarn: This is absolutely not how it works under the hood. It's a bit of a hack that takes advantage of simplifying antiaffinity enough until it's translatable to topology.

felix-zhe-huang (commented):
From https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/ I saw this.
>"Note: Pod anti-affinity requires nodes to be consistently labelled, in other words every node in the cluster must have an appropriate label matching topologyKey. If some or all nodes are missing the specified topologyKey label, it can lead to unintended behavior."

So it seems that we also need to create a mechanism to label the existing nodes with the topologKey in order for this to work.

PR #1227, opened 2022-01-26
Assume AWS CNI v1.9+ by default

ellistarn:
**1. Issue, if available:**

- [ ] No

### DIFF
--- a/pkg/utils/options/options.go
+++ b/pkg/utils/options/options.go
@@ -40,7 +40,7 @@ func MustParse() Options {
 	flag.IntVar(&opts.KubeClientQPS, "kube-client-qps", env.WithDefaultInt("KUBE_CLIENT_QPS", 200), "The smoothed rate of qps to kube-apiserver")
 	flag.IntVar(&opts.KubeClientBurst, "kube-client-burst", env.WithDefaultInt("KUBE_CLIENT_BURST", 300), "The maximum allowed burst of queries to the kube-apiserver")
 	flag.StringVar(&opts.AWSNodeNameConvention, "aws-node-name-convention", env.WithDefaultString("AWS_NODE_NAME_CONVENTION", string(IPName)), "The node naming convention used by the AWS cloud provider. DEPRECATION WARNING: this field may be deprecated at any time")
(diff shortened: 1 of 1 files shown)


pkg/utils/options/options.go
```diff
@@ -40,7 +40,7 @@ func MustParse() Options {
 	flag.IntVar(&opts.KubeClientQPS, "kube-client-qps", env.WithDefaultInt("KUBE_CLIENT_QPS", 200), "The smoothed rate of qps to kube-apiserver")
 	flag.IntVar(&opts.KubeClientBurst, "kube-client-burst", env.WithDefaultInt("KUBE_CLIENT_BURST", 300), "The maximum allowed burst of queries to the kube-apiserver")
 	flag.StringVar(&opts.AWSNodeNameConvention, "aws-node-name-convention", env.WithDefaultString("AWS_NODE_NAME_CONVENTION", string(IPName)), "The node naming convention used by the AWS cloud provider. DEPRECATION WARNING: this field may be deprecated at any time")
-	flag.BoolVar(&opts.AWSENILimitedPodDensity, "aws-eni-limited-pod-density", env.WithDefaultBool("AWS_ENI_LIMITED_POD_DENSITY", true), "Indicates whether new nodes should use ENI-based pod density")
+	flag.BoolVar(&opts.AWSENILimitedPodDensity, "aws-eni-limited-pod-density", env.WithDefaultBool("AWS_ENI_LIMITED_POD_DENSITY", false), "Indicates whether new nodes should use ENI-based pod density")
```
suket22, line 43: I think the problem though is that even if someone does have CNI v1.9+, you still have to opt into prefix delegation by toggling a bunch of environment variables because those are *still* disabled by default.

mikesir87: I used `true` as the default as that's the current behavior and we're only changing things if the setting is flipped around. As @suket22 mentioned, I think it might be better for folks to use CNI v1.9+, but there is more setup we should include in the getting started guides.

ellistarn:
Closing for now. Thanks! Will revisit later.

PR #1228, opened 2022-01-26
support region wellknown label

bwagner5:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/1109
https://github.com/aws/karpenter/issues/1218

Implement label ignoring w/ region as the first ignored label

 - scaled a node that got the region label 

- [x] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/requirements.go
+++ b/pkg/apis/provisioning/v1alpha5/requirements.go
@@ -52,6 +52,7 @@ var (
 	// WellKnownLabels supported by karpenter
 	WellKnownLabels = sets.NewString(
 		v1.LabelTopologyZone,
+		v1.LabelTopologyRegion,
 		v1.LabelInstanceTypeStable,
 		v1.LabelArchStable,
 		v1.LabelOSStable,
@@ -63,10 +64,11 @@ var (
 	// however, Provisioner labels are still restricted to WellKnownLabels.
 	// Additional labels may be injected by cloud providers.
 	NormalizedLabels = map[string]string{
-		v1.LabelFailureDomainBetaZone: v1.LabelTopologyZone,
-		"beta.kubernetes.io/arch":     v1.LabelArchStable,
-		"beta.kubernetes.io/os":       v1.LabelOSStable,
-		v1.LabelInstanceType:          v1.LabelInstanceTypeStable,
+		v1.LabelFailureDomainBetaZone:   v1.LabelTopologyZone,
+		v1.LabelFailureDomainBetaRegion: v1.LabelTopologyRegion,
+		"beta.kubernetes.io/arch":       v1.LabelArchStable,
+		"beta.kubernetes.io/os":         v1.LabelOSStable,
+		v1.LabelInstanceType:            v1.LabelInstanceTypeStable,
 	}
 )
 
@@ -75,6 +77,10 @@ type Requirements []v1.NodeSelectorRequirement
 
 func (r Requirements) Zones() sets.String {
 	return r.Requirement(v1.LabelTopologyZone)
+}
+
+func (r Requirements) Regions() sets.String {
+	return r.Requirement(v1.LabelTopologyRegion)
 }
 
 func (r Requirements) InstanceTypes() sets.String {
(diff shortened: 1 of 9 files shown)


pkg/cloudprovider/aws/instance.go
```diff
@@ -249,17 +249,19 @@ func (p *InstanceProvider) instanceToNode(ctx context.Context, instance *ec2.Ins
...
 				nodeName = aws.StringValue(instance.InstanceId)
 			}
+			zone := aws.StringValue(instance.Placement.AvailabilityZone)
 			return &v1.Node{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: nodeName,
 					Labels: map[string]string{
-						v1.LabelTopologyZone:       aws.StringValue(instance.Placement.AvailabilityZone),
+						v1.LabelTopologyZone:       zone,
+						v1.LabelTopologyRegion:     zone[:len(zone)-1],
```
suket22, line 258: Does `zone[:len(zone)-1]` work for outposts / localzones etc?  Should we get region from an env var instead - I'm guessing an env var is set since the AWS SDK is sending requests to the right region

ellistarn:
Can you try this with the in-tree EBS volume controller to make sure that this isn't the only blocker?

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -77,6 +79,10 @@ func (r Requirements) Zones() sets.String {
 	return r.Requirement(v1.LabelTopologyZone)
 }
 
+func (r Requirements) Regions() sets.String {
+	return r.Requirement(v1.LabelTopologyRegion)
```
ellistarn, line 83: IIUC, since this is nil, it means that region is unconstrained? What if you specify an incorrect region. Does it behave correctly?

pkg/cloudprovider/aws/instance.go
ellistarn: This complexity is exactly the reason I didn't want to Karpenter to be region aware.

pkg/apis/provisioning/v1alpha5/requirements.go
bwagner5: It does now by failing in the CreateInstance func at runtime.  Before my last push it was a noop.  I don't know a usecase where a region would be specified that isn't where Karpenter is deployed, but I suppose we still shouldn't provision if they are conflicting.

bwagner5:
> Can you try this with the in-tree EBS volume controller to make sure that this isn't the only blocker?

Yeah, I tested with the in-tree and reproduced with the current main HEAD. I deployed this branch and a node gets provisioned successfully and the pod goes to Running.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -111,6 +111,9 @@ func withUserAgent(sess *session.Session) *session.Session {
 
 // Create a node given the constraints.
 func (c *CloudProvider) Create(ctx context.Context, constraints *v1alpha5.Constraints, instanceTypes []cloudprovider.InstanceType, quantity int, callback func(*v1.Node) error) error {
+	if constraints.Requirements.Regions().Len() != 0 && !constraints.Requirements.Regions().Has(c.instanceTypeProvider.region) {
```
ellistarn, line 114: We don't need this logic anymore, since the requirements guarantee it.

pkg/cloudprovider/aws/instance.go
```diff
@@ -249,17 +249,26 @@ func (p *InstanceProvider) instanceToNode(ctx context.Context, instance *ec2.Ins
...
+					break
+				}
+			}
 			return &v1.Node{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: nodeName,
 					Labels: map[string]string{
-						v1.LabelTopologyZone:       aws.StringValue(instance.Placement.AvailabilityZone),
+						v1.LabelTopologyZone:       zone,
+						v1.LabelTopologyRegion:     region,
```
ellistarn, line 265: I'd happily skip applying this label. Not sure I see the value -- the kubelet will put it on, anyways.

bwagner5:
Super simple now! It just ignores it :)

ellistarn (approved):
har har har har har

PR #1231, opened 2022-01-27
Update Provisioner Docs Example

DWSR:
The `spec.limits.resources.cpu` field should be a string, not an int.

**1. Issue, if available:**

An example in the docs is incorrect

Updated example to be correct

N/A

- [ ] No

### DIFF
--- a/website/content/en/preview/provisioner.md
+++ b/website/content/en/preview/provisioner.md
@@ -162,7 +162,7 @@ The provisioner spec includes a limits section (`spec.limits.resources`), which 
 
 Presently, Karpenter supports `memory` and `cpu` limits. 
 
-CPU limits are described with a `DecimalSI` value, usually a natural integer. Note that the Kubernetes API will return a string value here, so when using e.g. ArgoCD, it's better to specify the CPU limit as a string.
+CPU limits are described with a `DecimalSI` value. Note that the Kubernetes API will coerce this into a string, so we recommend against using integers to avoid GitOps skew.
 
 Memory limits are described with a [`BinarySI` value, such as 1000Gi.](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory)
 
--- a/website/content/en/preview/provisioner.md
+++ b/website/content/en/preview/provisioner.md
@@ -52,7 +52,7 @@ spec:
   # Limits prevent Karpenter from creating new instances once the limit is exceeded.
   limits:
     resources:
-      cpu: 1000 
+      cpu: "1000"
       memory: 1000Gi
 
   # These fields vary per cloud provider, see your cloud provider specific documentation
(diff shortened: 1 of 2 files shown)


bwagner5:
Did you experience an error with that yaml?  I just applied it to my cluster and it worked for me:

```
$ cat <<EOF | kubectl apply -f -
kind: Provisioner
metadata:
  name: test-1
spec:
  limits:
    resources:
      cpu: 1000
      memory: 1000Gi
  requirements:
....
```

```
$ k get provisioner test-1 -o yaml
kind: Provisioner
metadata:
  ...
  name: test-1
spec:
  kubeletConfiguration: {}
  limits:
    resources:
      cpu: 1k
      memory: 1000Gi```

File: website/content/en/v0.5.6/provisioner.md [hunk 1]
```diff
@@ -52,7 +52,7 @@ spec:
   # Limits prevent Karpenter from creating new instances once the limit is exceeded.
   limits:
     resources:
-      cpu: 1000 
+      cpu: "1000"
```
ellistarn, line 55: Is this necessary? IIUC, the integer works just fine.

bwagner5:
I believe quotes are only required if the field is a string and the input is a number. In that case, you'd need to use quotes to tell yaml that you're actually passing a string rather than a number. 

https://github.com/aws/karpenter/blob/main/charts/karpenter/crds/karpenter.sh_provisioners.yaml#L65-L67

DWSR:
@bwagner5 @ellistarn The kube API seems to return a string regardless of the type submitted to the API server, based on my Very Scientific Testing on my 1.20.11 EKS cluster.

Applying a provisioner that uses integer limits via ArgoCD will cause ArgoCD to always mark the resource as Out of Sync.

ellistarn (commented):
Nice catch!

Can you include this in /preview docs? We copy preview docs into the next release version when we release. Optionally, you could backport this to older versions of docs, but I'm happy with it just being included in the next release.

website/content/en/preview/provisioner.md
```diff
@@ -162,7 +162,7 @@ The provisioner spec includes a limits section (`spec.limits.resources`), which
 
 Presently, Karpenter supports `memory` and `cpu` limits. 
 
-CPU limits are described with a `DecimalSI` value, usually a natural integer. 
+CPU limits are described with a `DecimalSI` value, usually a natural integer. Note that the Kubernetes API will return a string value here, so when using e.g. ArgoCD, it's better to specify the CPU limit as a string.
```
ellistarn, line 165:
```suggestion
CPU limits are described with a `DecimalSI` value. Note that the Kubernetes API will coerce this into a string, so we recommend against using integers to avoid GitOps skew.
```

ellistarn (approved):

ellistarn:
Thank you!

PR #1232, opened 2022-01-28
Support IPv6 cluster for generated LaunchTemplate

literalice:
**1. Issue, if available:**
When using generated LaunchTemplate and IPv6 cluster it cannot provision nodes normally.

Adds IP Family options

* Created a IPv6 cluster
* Deployed Karpenter built from the branch
* Confirmed the node launched and host the scaled pods

- [x] No


pkg/utils/options/options.go
```diff
@@ -39,6 +39,8 @@ func MustParse() Options {
 	flag.IntVar(&opts.WebhookPort, "port", 8443, "The port the webhook endpoint binds to for validation and mutation of resources")
 	flag.IntVar(&opts.KubeClientQPS, "kube-client-qps", env.WithDefaultInt("KUBE_CLIENT_QPS", 200), "The smoothed rate of qps to kube-apiserver")
 	flag.IntVar(&opts.KubeClientBurst, "kube-client-burst", env.WithDefaultInt("KUBE_CLIENT_BURST", 300), "The maximum allowed burst of queries to the kube-apiserver")
+	flag.StringVar(&opts.IpFamily, "ip-family", env.WithDefaultString("IP_FAMILY", ""), "The kubernetes ip family (ipv4/ipv6)")
+	flag.StringVar(&opts.ServiceIpv6Cidr, "service-ipv6-cidr", env.WithDefaultString("SERVICE_IPV6_CIDR", ""), "The kubernetes service CIDR for IPv6 cluster")
```
ellistarn, line 43: I wonder if we can do something like https://pkg.go.dev/net#ParseCIDR to reduce our API Surface. I could imagine that this configuration lives in our provisioner API itself, rather than the helm chart: https://github.com/aws/karpenter/blob/main/pkg/apis/provisioning/v1alpha5/kubelet_configuration.go

pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
```diff
@@ -22,4 +22,8 @@ type KubeletConfiguration struct {
 	// Note that not all providers may use all addresses.
 	//+optional
 	ClusterDNS []string `json:"clusterDNS,omitempty"`
+
+	// serviceClusterIPRange is a IP CIDR that will be used for assigning IP to services created on Kubernetes.
+	//+optional
+	ServiceClusterIPRange string `json:"serviceClusterIPRange,omitempty"`
```
literalice, line 28: I add a Provisioner API's attribute for specifying API server's `--service-cluster-ip-range` parameter.
EKS optimized AMI uses the parameter to infer Cluster DNS Service IP.

pkg/utils/options/options.go
literalice: I agree that we can resolve IP Family from a service cidr:

https://github.com/aws/karpenter/blob/f1f9e51fd024fd138df32b87ccb1199051331a35/pkg/cloudprovider/aws/launchtemplate.go#L262-L277

pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
anguslees: Why do we need the service-cluster-ip-range, if we can just specify the cluster-dns ip directly?

literalice: We can specify only `service-cluster-ip-range` parameter without `cluster-dns`, however current implementation need `service-cluster-ip-range` even if `cluster-dns` is specified when using IPv6 cluster.
It's caused the EKS Optimized AMI's implementation in which `SERVICE_IPV6_CIDR` is mandatory:
https://github.com/awslabs/amazon-eks-ami/blob/v20220123/files/bootstrap.sh#L303

ellistarn: If this is the case, we should fix the AMI, rather than polluting karpenter with additional complexity.

literalice: Thank you for the feedbacks. I changed the PR to be able to use IPv6 provisioning node only when users specify the IPv6 DNS Cluster IP.
It may not so straightforward however we can use IPv6 with Karpenter without changing the Karpenter API.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -349,9 +350,20 @@ exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
 		userData.WriteString(fmt.Sprintf(` \
     --kubelet-extra-args '%s'`, kubeletExtraArgs))
 	}
-	if len(constraints.KubeletConfiguration.ClusterDNS) > 0 {
+	if dnsClusterIp := constraints.KubeletConfiguration.ClusterDNS; len(dnsClusterIp) > 0 {
 		userData.WriteString(fmt.Sprintf(` \
-    --dns-cluster-ip '%s'`, constraints.KubeletConfiguration.ClusterDNS[0]))
+    --dns-cluster-ip '%s'`, dnsClusterIp[0]))
+
+		ipAddr, ipNet, _ := net.ParseCIDR(dnsClusterIp[0] + "/108")
```
ellistarn, line 357: We should panic if we fail to parse instead of squashing the error with `_`.

ellistarn (commented):
Awesome feature thank you! 

Can we wire this up for bottlerocket, too? https://github.com/bottlerocket-os/bottlerocket/#kubernetes-settings

As you suggest, I wonder if we should be parsing the Kubernetes Service object and discover the ClusterDNS from that. This will enable a zero-config solution. Thoughts on including this in the same change?

literalice:
> Can we wire this up for bottlerocket, too? https://github.com/bottlerocket-os/bottlerocket/#kubernetes-settings
> 
> As you suggest, I wonder if we should be parsing the Kubernetes Service object and discover the ClusterDNS from that. This will enable a zero-config solution. Thoughts on including this in the same change?

Hi, Thank you for the feedback. Parsing the k8s Service was implemented like:

https://github.com/aws/karpenter/blob/c21cded3da61969697f408783c852dab9d965f01/pkg/cloudprovider/aws/launchtemplate.go#L77-L85

Currently bottlerocket can derive their cluster CIDR from EKS API, and set up IPv6 node normally in EKS env at least.
https://github.com/bottlerocket-os/bottlerocket/pull/1547

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -54,21 +57,33 @@ type LaunchTemplateProvider struct {
...
 	}
 	l.cache.OnEvicted(l.onCacheEvicted)
 	l.hydrateCache(ctx)
 	return l
 }
 
+func discoverClusterDNS(ctx context.Context, clientSet *kubernetes.Clientset) net.IP {
+	dnsService, err := clientSet.CoreV1().Services("default").Get(ctx, "kubernetes", metav1.GetOptions{})
+	if err != nil {
+		logging.FromContext(ctx).Debugf("Unable to discover clusterDNS, %v", err)
```
ellistarn, line 80: I think this is a FatalF

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -357,27 +372,49 @@ exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
...
     --kubelet-extra-args '%s'`, kubeletExtraArgs))
-	}
-	if len(constraints.KubeletConfiguration.ClusterDNS) > 0 {
-		userData.WriteString(fmt.Sprintf(` \
-    --dns-cluster-ip '%s'`, constraints.KubeletConfiguration.ClusterDNS[0]))
 	}
 	return base64.StdEncoding.EncodeToString(userData.Bytes())
 }
 
+func isIPv6(ip net.IP) bool {
```
ellistarn, line 411: It would be great if we were somehow guaranteed that this isn't nil. Then we could just use 
```
len(ip.To16()) == net.IPv6len
```
without a helper. 
Why do we even need to check To4?

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -54,21 +57,33 @@ type LaunchTemplateProvider struct {
...
 
-func NewLaunchTemplateProvider(ctx context.Context, ec2api ec2iface.EC2API, amiProvider *AMIProvider, securityGroupProvider *SecurityGroupProvider) *LaunchTemplateProvider {
+func NewLaunchTemplateProvider(ctx context.Context, ec2api ec2iface.EC2API, clientSet *kubernetes.Clientset, amiProvider *AMIProvider, securityGroupProvider *SecurityGroupProvider) *LaunchTemplateProvider {
 	l := &LaunchTemplateProvider{
 		ec2api:                ec2api,
 		logger:                logging.FromContext(ctx).Named("launchtemplate"),
 		amiProvider:           amiProvider,
 		securityGroupProvider: securityGroupProvider,
 		cache:                 cache.New(CacheTTL, CacheCleanupInterval),
+		clusterDNSIP:          discoverClusterDNS(ctx, clientSet),
```
ellistarn, line 70: @bwagner5, I love that this happens at init time. WDYT about discovering cabundle here in a followon PR?

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -54,21 +57,33 @@ type LaunchTemplateProvider struct {
...
 	}
 	l.cache.OnEvicted(l.onCacheEvicted)
 	l.hydrateCache(ctx)
 	return l
 }
 
+func discoverClusterDNS(ctx context.Context, clientSet *kubernetes.Clientset) net.IP {
+	dnsService, err := clientSet.CoreV1().Services("default").Get(ctx, "kubernetes", metav1.GetOptions{})
```
ellistarn, line 78: I think this is looking at the wrong service. It should be `kube-system/kube-dns`: https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/#is-dns-service-up

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -54,21 +57,33 @@ type LaunchTemplateProvider struct {
...
+	if err != nil {
+		logging.FromContext(ctx).Debugf("Unable to discover clusterDNS, %v", err)
+		return nil
+	}
+	logging.FromContext(ctx).Debugf("Discovered clusterDNS, %s", dnsService.Spec.ClusterIP)
+	return net.ParseIP(dnsService.Spec.ClusterIP)
```
bwagner5, line 84: This wouldn't support dual stack right? https://kubernetes.io/docs/concepts/services-networking/dual-stack/

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -357,27 +372,49 @@ exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
 			*caBundle))
 	}
 
+	clusterDNSIP := p.clusterDNSIP
```
bwagner5, line 375: I think we'd also need to constrain the instance types that we can use when doing ipv6. Only nitro instance types support IPv6 Prefix Delegation.

bwagner5:
The IMDS IPv6 endpoint is also disabled by default because it only works on Nitro instance types. https://aws.amazon.com/about-aws/whats-new/2021/08/Ipv6-amazon-ec2-metadata-time-sync-vpc-dns/

pkg/cloudprovider/aws/launchtemplate.go
literalice: You're right, I'd fixed it. Thank you.

pkg/cloudprovider/aws/launchtemplate.go
literalice: Yes we need to launch Nitro instances to use IPv6. However I think it's very useful if Karpenter has a function to be able to filter non-Nitro instances by using provisioner's constraints.
Currently we need to specify instance types which are on Nitro by provisioner settings.

pkg/cloudprovider/aws/launchtemplate.go
literalice: When I ran a test the (ipv4).To16() returns IPv6 formatted IPv4 address which length is IPv6len. So I added a check To4 which returns nil for IPv6 addresses.

pkg/cloudprovider/aws/amifamily/bootstrap/eksbootstrap.go
```diff
@@ -40,16 +41,30 @@ func (e EKS) Script() string {
...
+
+	if e.KubeletConfig != nil && len(e.KubeletConfig.ClusterDNS) > 0 {
+		clusterDNSIP = net.ParseIP(e.KubeletConfig.ClusterDNS[0])
+		userData.WriteString(fmt.Sprintf(" \\\n--dns-cluster-ip '%s'", clusterDNSIP))
+	}
+
+	ipv6Cluster := isIPv6(clusterDNSIP)
+	if ipv6Cluster {
+		userData.WriteString(" \\\n--ip-family ipv6")
+		_, ipNet, err := net.ParseCIDR(clusterDNSIP.String() + "/108")
```
anguslees, line 54: Rather than build a string, then parse it - just construct the result directly:
```suggestion
		ipNet := net.IPNet{IP: clusterDNSIP, Mask: net.CIDRMask(108, 128)}
```

pkg/cloudprovider/aws/amifamily/bootstrap/eksbootstrap.go
```diff
@@ -77,3 +92,10 @@ func (e EKS) nodeLabelArg() string {
 	}
 	return fmt.Sprintf("%s%s", nodeLabelArg, strings.Join(labelStrings, ","))
 }
+
+func isIPv6(ip net.IP) bool {
+	if ip != nil {
+		return len(ip.To4()) != net.IPv4len && len(ip.To16()) == net.IPv6len
+	}
+	return false
```
anguslees, line 100: It's a trap!  😱

You can't check length, since golang's `net.IP` will often encode IPv4 addresses embedded in IPv6.  ([Example](https://go.dev/play/p/hSoy0H9mLvL))

```suggestion
	return ip.To4() == nil
```

pkg/test/environment.go
```diff
@@ -67,9 +67,17 @@ type EnvironmentOption func(env *Environment)
 
 func NewEnvironment(ctx context.Context, options ...EnvironmentOption) *Environment {
 	ctx, stop := context.WithCancel(ctx)
+	customApiServerFlags := []string{
```
anguslees, line 70: This creates a dual-stack cluster.  Which is fine, but I should point out that it is not the same as an EKS IPv6 cluster.  (EKS is an _IPv6-only_ cluster, with dual-stack pods)

pkg/cloudprovider/aws/amifamily/bootstrap/eksbootstrap.go
```diff
@@ -40,16 +41,30 @@ func (e EKS) Script() string {
...
+		_, ipNet, err := net.ParseCIDR(clusterDNSIP.String() + "/108")
+		if err != nil {
+			panic(fmt.Sprintf("Parsing IPv6 cluster DNS, %s", err))
+		}
+		userData.WriteString(fmt.Sprintf(" \\\n--service-ipv6-cidr '%s'", ipNet))
+	}
+
+	if !e.AWSENILimitedPodDensity || ipv6Cluster {
+		userData.WriteString(" \\\n--use-max-pods false")
+		kubeletExtraArgs += " --max-pods 110"
```
anguslees, line 63: Comment: `--max-pods=110` is the default, so it's fine to leave it here for clarity - but we could also omit it and it would have the same result.

anguslees (commented):
(otherwise lgtm)

pkg/cloudprovider/aws/launchtemplate.go
anguslees: `net.IP` is fun and often encodes IPv4 addresses embedded in IPv6 - see my comment elsewhere for a play.golang example.

You have to check `To4() != nil`

pkg/cloudprovider/aws/amifamily/bootstrap/bootstrap.go
```diff
@@ -29,6 +31,7 @@ type Options struct {
 	Labels                  map[string]string `hash:"set"`
 	CABundle                *string
 	AWSENILimitedPodDensity bool
+	KubeDnsIP               net.IP
```
ellistarn, line 34: Is it possible to just use the value in KubeletConfig?

bwagner5: Using the value in the kubelet config would depend on the user passing that optional parameter. So if we relied on this, we'd need to default it based on our discovery or require the user to pass in cluster dns ip to engage the ipv6 functionality. Requiring the user to specify the dns ip would be very surprising. 

I do agree that adding another `kubeDNSIP` to the bootstrap struct is probably not the right path here though since you'd need to decide which dns ip to use in a bootstrap implementation.

pkg/cloudprovider/aws/launchtemplate.go
literalice: In dual stack mode we can get IPv6 address from the field if the user prefers the address family. So it seems to be able to support dual stack.

pkg/cloudprovider/aws/amifamily/bootstrap/bootstrap.go
literalice: I appreciate your feedback. I agree that the additional valuables in the struct is not appropriate. I moved the DNS discovery to `eksbootstrap` because the PR is only related to `bootstrap.sh` of AMI.

pkg/cloudprovider/aws/amifamily/bootstrap/eksbootstrap.go
literalice: Thank you for the suggestion. I fixed it.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -265,3 +268,14 @@ func (p *LaunchTemplateProvider) kubeServerVersion(ctx context.Context) (string,
 	logging.FromContext(ctx).Debugf("Discovered kubernetes version %s", version)
 	return version, nil
 }
+
+func (p *LaunchTemplateProvider) kubeDnsIP(ctx context.Context) net.IP {
+	dnsService, err := p.clientSet.CoreV1().Services("kube-system").Get(ctx, "kube-dns", metav1.GetOptions{})
+	if err != nil {
+		logging.FromContext(ctx).Debug("Failed to DNS in the cluster. Maybe kube-dns serivce doesn't exist.")
+		return nil
```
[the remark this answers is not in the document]
literalice: I'm not sure it's so critical if kube-dns cannot be found in the PR so in the fix I made the loglevel Warnf and consider the cluster IPv4.

pkg/cloudprovider/aws/amifamily/bootstrap/eksbootstrap.go
```diff
@@ -16,15 +16,23 @@ package bootstrap
...
 
+	"knative.dev/pkg/logging"
+
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8s "k8s.io/client-go/kubernetes"
 )
 
 type EKS struct {
+	context.Context
```
bwagner5, line 34: I don't think this is the right place to move these to.  I think the way you had it before was okay, but the parameter that the bootstrap script cares about is something denoting if the node should be an ipv6 node or not. The bootstrap script doesn't need to care about how that was decided (it could be a user setting it explicitly to be ipv6 or the mechanism to discover the kube-dns IP and check if that is ipv4 or ipv6). 

EKSBootstrap doesn't bootstrap bottlerocket which should also support IPv6. 

I would suggest going back to the way you had it before, but instead of passing a second DNS IP, to add an `IPFamily` field to the base Bootstrap Options. That way the eks bootstrapper and the bottlerocket bootstrapper can work with that parameter directly.  You can also add a parameter for the `IPv6CIDR`

bwagner5 (changes requested):
forgot to set this to request changes, see the comment above

pkg/cloudprovider/aws/amifamily/bootstrap/eksbootstrap.go
literalice: Thanks, I added the IPv6CIDR params in the struct.

On the other hand, I found the users need to pass the cluster dns ip explicitly to launch bottlerocket nodes correctly when their cluster cannot use EKS API . 
https://github.com/bottlerocket-os/bottlerocket/blob/5d6cc099d247ecc12e8250898c48d9ee762f8c42/sources/api/pluto/src/main.rs#L256

I think we have to pass cluster dns IP to the bottlerocket bootstrap code in karpenter which generates the toml for bottlerocket even if the user doesn't specify their DNS IP.

Could I post another issue for the bottlerocket?

bwagner5:
Hey @literalice ! If you don't mind, I've opened a PR that includes your commits for the ipv6 support. Hopefully this can help to get this merged sooner and take the burden off of you to constantly rebase :D 

https://github.com/aws/karpenter/pull/1605

The new PR builds off your commits here but also uses the `--dns-cluster-ip` `--ip-family` semantic from your eks-optimized AMI PR https://github.com/awslabs/amazon-eks-ami/pull/860 . Thanks for all these contributions!!

literalice:
Hi @bwagner5 ,
I'm very glad to contribute this issue and see the new your PR submitted! 
I applied your suggestion on the related issue: https://github.com/awslabs/amazon-eks-ami/pull/860
Thank you very much for your feedbacks!

Issue #1141, opened 2022-01-13
Validation does not catch invalid capacity-type values


Tasmana-banana:
## Version

**Karpenter**: v0.5.3

**Kubernetes**: v1.20.0

Hi all!
When apply provisioner and scale test deployment i get some error in logs controller:

```
2022-01-13T21:43:17.290Z	INFO	controller.provisioning	Waiting for unschedulable pods	{"commit": "5047f3c", "provisioner": "default"}
2022-01-13T21:43:19.291Z	INFO	controller.provisioning	Batched 4 pods in 1.001107034s	{"commit": "5047f3c", "provisioner": "default"}
2022-01-13T21:43:19.333Z	ERROR	controller.provisioning	Failed to find instance type option(s) for [default/inflate-6b88c9fb68-8vvs5 default/inflate-6b88c9fb68-22hlt default/inflate-6b88c9fb68-dfzgm default/inflate-6b88c9fb68-ttrhh]	{"commit": "5047f3c", "provisioner": "default"}
```
Use Provisioner:

```
apiVersion: karpenter.sh/v1alpha5
kind: Provisioner
metadata:
  name: default
spec:
  requirements:
    - key: "topology.kubernetes.io/zone"
      operator: In
      values: ["eu-west-2a", "eu-west-2b", "eu-west-2c"]
    - key: "kubernetes.io/arch"
      operator: In
      values: ["arm64", "amd64", "x86"]
    - key: node.kubernetes.io/instance-type
      operator: In
      values: [ "t2.medium", "t3a.medium", "t3.large"]
    - key: karpenter.sh/capacity-type
      operator: In
      values: ["spot", "on-demand"]
  limits:
    resources:
      cpu: 1000
  provider:
    instanceProfile: dev-eks-20211224135100984100000004
  ttlSecondsAfterEmpty: 30
```
Deploy karpenter chart by ansible:

```
- name: Deploy karpenter chart
  kubernetes.core.helm:
    name: "{{ karpenter_namespace }}"
    create_namespace: true
    release_namespace: "{{ karpenter_namespace }}"
    chart_ref: karpenter/karpenter
    chart_version: 0.5.3
    release_values:
      serviceAccount:
        create: true
        name: karpenter
        annotations:
          eks.amazonaws.com/role-arn: "arn:aws:iam::{{ account_id }}:role/EKS-Karpenter-Role"
      controller:
        clusterName: "{{ cluster_name }}"
[...]

PR #1254, opened 2022-02-01
Add capacity type value validation

felix-zhe-huang:
**1. Issue, if available:**
#1141 

Add label value validation for capacity type.

Unit tests, e2e tests

- [X] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/provisioner_validation.go
+++ b/pkg/apis/provisioning/v1alpha5/provisioner_validation.go
@@ -29,6 +29,7 @@ import (
 
 var (
 	SupportedNodeSelectorOps = []string{string(v1.NodeSelectorOpIn), string(v1.NodeSelectorOpNotIn)}
+	SupportedCapacityType    = []string{CapacityTypesOnDemand, CapacityTypesSpot}
 )
 
 func (p *Provisioner) Validate(ctx context.Context) (errs *apis.FieldError) {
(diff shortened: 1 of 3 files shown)


pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -29,6 +29,7 @@ import (
 
 var (
 	SupportedNodeSelectorOps = []string{string(v1.NodeSelectorOpIn), string(v1.NodeSelectorOpNotIn)}
+	SupportedCapacityType    = []string{CapacityTypesOnDemand, CapacityTypesSpot}
```
ellistarn, line 32: Whatever our validation strategy is, it should be applied equally to all well known labels (zone, arch, os, capacity type, instance type). The current validation logic happens at runtime as part of requirements math: https://github.com/aws/karpenter/blob/main/pkg/controllers/provisioning/controller.go#L141

I'm not sure we want to do api time validation, as the zones, etc may change over time given subnet configuration, etc. Instead, runtime validation (requirements math) is working pretty well. 

The key issue from the referenced bug is that 
```
2022-01-13T21:43:19.333Z	ERROR	controller.provisioning	Failed to find instance type option(s) for [default/inflate-6b88c9fb68-8vvs5 default/inflate-6b88c9fb68-22hlt default/inflate-6b88c9fb68-dfzgm default/inflate-6b88c9fb68-ttrhh]	{"commit": "5047f3c", "provisioner": "default"}
```
is an unintuitive error message. I'd focus on making sure that we error with something related to capacity type (or zone, etc if relevant), rather than trying to do api time valiation..

WDYT?

PR #1260, opened 2022-02-02
Added hugo to go toolchain

ellistarn:
**1. Issue, if available:**

Dependency in a language other than go? 🙅 

- [ ] No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,3 @@ coverage.out
 coverage.html
 *.test
 *.csv
-website/node_modules
-website/resources
-website/public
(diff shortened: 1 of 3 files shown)


njtran:
AFAIK, Hugo is only used for creating releases. While small, I'm don't think we should add in tools that aren't necessary for the everyone into the toolchain, since it adds another method of tool installation failure. WDYT?

geoffcline:
I think this is the incorrect version of Hugo. we need Hugo-extended for the docsy theme. 

I am unfamiliar with how go installs Hugo, so maybe it defaults to extended?

appreciate the gitingore updates

ellistarn:
> I think this is the incorrect version of Hugo. we need Hugo-extended for the docsy theme.
> 
> I am unfamiliar with how go installs Hugo, so maybe it defaults to extended?
> 
> appreciate the gitingore updates

I found it was nice to use the same packaging mechanism for all our dependencies. e.g. `brew` only works on macs, and not always for the new M1s. I'm using the latest version here: https://github.com/gohugoio/hugo

geoffcline:
excellent!

geoffcline (approved):

Issue #993, opened 2021-12-14
Why Karpenter waits more than specified in ttlSecondsAfterEmpty?


Noksa:
Hello.
Could you please explain why Karpenter waited 4 minutes instead of 2?

```
│ 2021-12-14T21:31:26.470Z    INFO    controller.node    Added TTL to empty node    {"commit": "870e2f6", "node": "ip-192-168-12-22.eu-central-1.compute.internal"}              │
│ 2021-12-14T21:31:47.047Z    INFO    controller.node    Removed emptiness TTL from node    {"commit": "870e2f6", "node": "ip-192-168-12-22.eu-central-1.compute.internal"}      │
│ 2021-12-14T21:32:02.114Z    INFO    controller.node    Added TTL to empty node    {"commit": "870e2f6", "node": "ip-192-168-12-22.eu-central-1.compute.internal"}              │
│ 2021-12-14T21:36:33.041Z    INFO    controller.node    Triggering termination after 2m0s for empty node    {"commit": "870e2f6", "node": "ip-192-168-12-22.eu-central-1.comput │
│ 2021-12-14T21:36:33.065Z    INFO    controller.termination    Cordoned node    {"commit": "870e2f6", "node": "ip-192-168-12-22.eu-central-1.compute.internal"}                 │
│ 2021-12-14T21:36:33.219Z    INFO    controller.termination    Deleted node    {"commit": "870e2f6", "node": "ip-192-168-12-22.eu-central-1.compute.internal"}
```

njtran:
Can you share your Provisioner Spec(s)?

ellistarn:
Hmm -- I wonder if something is wrong with the Node reconcile trigger. 2m0s looks like the controller thought it was up to date at least. Perhaps there's a time skew between the API Server's clock and your Karpenter's clock?

Noksa:
Thanks guys for quick reaction! :)

Let me describe a bit.

I can reproduce it with the following manifests (I reduced TTL time).

Provisioner:

```
apiVersion: karpenter.sh/v1alpha5
kind: Provisioner
metadata:
  name: default
spec:
  ttlSecondsAfterEmpty: 30
  taints:
    - key: example.com/special-taint
      effect: NoSchedule
  labels:
    test: test

  requirements:
    - key: "node.kubernetes.io/instance-type"
      operator: In
      values: ["m5.large"]
    - key: "topology.kubernetes.io/zone"
      operator: In
      values: ["eu-central-1a"]
[...]

Issue #1103, opened 2022-01-07
TTLSecondsAfterEmpty not applied retroactively to nodes.


matti:
## Version
**Karpenter**: v0.5.3

**Kubernetes**: v1.0.0
```
kubectl version | grep Server                
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.5-eks-bc4871b", GitCommit:"5236faf39f1b7a7dabea8df12726f25608131aa9", GitTreeState:"clean", BuildDate:"2021-10-29T23:32:16Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}
```

## Expected Behavior

Updating `ttlSecondsAfterEmpty:` should be enough for nodes to start deleting

## Actual Behavior
Nothing happens (sometimes) and I need to delete the karpeter controller pod. When the a new pod starts, it marks the nodes for deletion.

## Steps to Reproduce the Problem

Set `ttlSecondsAfterEmpty: 3600`, have some nodes over 10min and update to `ttlSecondsAfterEmpty: 600`

felix-zhe-huang:
May I ask if you are referring to updating `ttlSecondsAfterEmpty` value on the fly? Can you share what command you use to update that value? I would like to recreate it step by step.

ellistarn:
This is all about the semantic of whether or not these are properties of nodes (ttlafterempty set on node creation) or whether or not it's a policy that applies retroactively to all nodes (ttlafterempty value is always current provisioner value).

I believe that currently we set a timestamp label on nodes of when it should be terminated, so if the provisioner is modified after the TTL is set, then it will use the previous value. I agree that this is a bit confusing, and I'd lean towards the design option that you're asking for.

@njtran designed this, so perhaps he can chime in.

matti:
> May I ask if you are referring to updating `ttlSecondsAfterEmpty` value on the fly? Can you share what command you use to update that value? I would like to recreate it step by step.

just kubectl apply -f - it

suket22:
> Set ttlSecondsAfterEmpty: 3600, have some nodes over 10min and update to ttlSecondsAfterEmpty: 600

To clarify, the nodes have to be empty over 10 minutes. If they were empty for just 5 minutes and you changed the `ttlSecondsAfterEmpty` to something, the reconcile event was getting dropped [due to this bug](https://github.com/aws/karpenter/pull/1262).

suket22:
[...]

PR #1262, opened 2022-02-02
Fix emptiness reconciler when ttl not expired

suket22:
**1. Issue, if available:**
#993 

* If ttl has not expired, requeueAfter at the right time. 

* Added a unit test. 
* Tested this manually as well. 

Before it would take a long time to clear up (4 minutes)- 
```
2022-02-02T00:47:51.832Z	INFO	controller.node	Added TTL to empty node	{"commit": "4829931", "node": "ip-192-168-191-121.us-west-2.compute.internal"}
2022-02-02T00:47:54.791Z	INFO	controller.node	Removed emptiness TTL from node	{"commit": "4829931", "node": "ip-192-168-191-121.us-west-2.compute.internal"}
2022-02-02T00:48:11.846Z	INFO	controller.node	Added TTL to empty node	{"commit": "4829931", "node": "ip-192-168-191-121.us-west-2.compute.internal"}
2022-02-02T00:52:13.964Z	INFO	controller.node	Triggering termination after 30s for empty node	{"commit": "4829931", "node": "ip-192-168-191-121.us-west-2.compute.internal"}
2022-02-02T00:52:13.996Z	INFO	controller.termination	Cordoned node	{"commit": "4829931", "node": "ip-192-168-191-121.us-west-2.compute.internal"}
2022-02-02T00:52:14.219Z	INFO	controller.termination	Deleted node	{"commit": "4829931", "node": "ip-192-168-191-121.us-west-2.compute.internal"}
```

Now it's as expected on a quick cycle of empty -> full -> empty and it drains as expected. 
```
2022-02-02T16:51:16.166Z	INFO	controller.provisioning	Waiting for unschedulable pods	{"commit": "69bb318", "provisioner": "default"}
2022-02-02T16:52:52.765Z	INFO	controller.node	Added TTL to empty node	{"commit": "69bb318", "node": "ip-192-168-179-90.us-west-2.compute.internal"}
2022-02-02T16:52:56.132Z	INFO	controller.node	Removed emptiness TTL from node	{"commit": "69bb318", "node": "ip-192-168-179-90.us-west-2.compute.internal"}
2022-02-02T16:53:09.093Z	INFO	controller.node	Added TTL to empty node	{"commit": "69bb318", "node": "ip-192-168-179-90.us-west-2.compute.internal"}
2022-02-02T16:53:39.001Z	INFO	controller.node	Triggering termination after 30s for empty node	{"commit": "69bb318", "node": "ip-192-168-179-90.us-west-2.compute.internal"}
2022-02-02T16:53:39.040Z	INFO	controller.termination	Cordoned node	{"commit": "69bb318", "node": "ip-192-168-179-90.us-west-2.compute.internal"}
2022-02-02T16:53:39.272Z	INFO	controller.termination	Deleted node	{"commit": "69bb318", "node": "ip-192-168-179-90.us-west-2.compute.internal"}
```

I even tried out a basic emptiness check to verify no regressions - 
```
2022-02-02T16:50:11.094Z	INFO	controller.node	Added TTL to empty node	{"commit": "69bb318", "node": "ip-192-168-171-248.us-west-2.compute.internal"}
2022-02-02T16:50:41.001Z	INFO	controller.node	Triggering termination after 30s for empty node	{"commit": "69bb318", "node": "ip-192-168-171-248.us-west-2.compute.internal"}
2022-02-02T16:50:41.033Z	INFO	controller.termination	Cordoned node	{"commit": "69bb318", "node": "ip-192-168-171-248.us-west-2.compute.internal"}
2022-02-02T16:50:41.254Z	INFO	controller.termination	Deleted node	{"commit": "69bb318", "node": "ip-192-168-171-248.us-west-2.compute.internal"}
```

- [x] No

### DIFF
--- a/pkg/controllers/node/suite_test.go
+++ b/pkg/controllers/node/suite_test.go
@@ -34,6 +34,7 @@ import (
 	. "knative.dev/pkg/logging/testing"
 	"knative.dev/pkg/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 var ctx context.Context
@@ -284,6 +285,27 @@ var _ = Describe("Controller", func() {
 
 			node = ExpectNodeExists(ctx, env.Client, node.Name)
 			Expect(node.DeletionTimestamp.IsZero()).To(BeFalse())
+		})
+		It("should requeue reconcile if node is empty, but not past emptiness TTL", func() {
+			provisioner.Spec.TTLSecondsAfterEmpty = ptr.Int64(30)
+			currentTime := time.Now()
+			injectabletime.Now = func() time.Time { return currentTime } // injectabletime.Now() is called multiple times in function being tested.
+			emptinessTime := injectabletime.Now().Add(-10 * time.Second)
+			// Emptiness timestamps are first formatted to a string friendly (time.RFC3339) (to put it in the node object)
+			// and then eventually parsed back into time.Time when comparing ttls. Repeating that logic in the test.
+			emptinessTimestamp, _ := time.Parse(time.RFC3339, emptinessTime.Format(time.RFC3339))
+			expectedrequeueTime := emptinessTimestamp.Add(time.Duration(30 * time.Second)).Sub(injectabletime.Now()) // we should requeue in ~20 seconds.
+			node := test.Node(test.NodeOptions{ObjectMeta: metav1.ObjectMeta{
+				Finalizers: []string{v1alpha5.TerminationFinalizer},
+				Labels:     map[string]string{v1alpha5.ProvisionerNameLabelKey: provisioner.Name},
+				Annotations: map[string]string{
+					v1alpha5.EmptinessTimestampAnnotationKey: emptinessTime.Format(time.RFC3339),
+				}},
+			})
+			ExpectCreated(ctx, env.Client, provisioner, node)
+			ExpectReconcileSucceededWithResult(ctx, controller, client.ObjectKeyFromObject(node), reconcile.Result{Requeue: true, RequeueAfter: expectedrequeueTime})
+			node = ExpectNodeExists(ctx, env.Client, node.Name)
+			Expect(node.DeletionTimestamp.IsZero()).To(BeTrue())
 		})
 	})
 	Context("Finalizer", func() {
--- a/pkg/controllers/node/suite_test.go
+++ b/pkg/controllers/node/suite_test.go
@@ -303,7 +303,8 @@ var _ = Describe("Controller", func() {
 				}},
 			})
 			ExpectCreated(ctx, env.Client, provisioner, node)
-			ExpectReconcileSucceededWithResult(ctx, controller, client.ObjectKeyFromObject(node), reconcile.Result{Requeue: true, RequeueAfter: expectedrequeueTime})
+			reconcileResult := ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(node))
+			Expect(reconcileResult).To(Equal(reconcile.Result{Requeue: true, RequeueAfter: expectedrequeueTime}))
 			node = ExpectNodeExists(ctx, env.Client, node.Name)
 			Expect(node.DeletionTimestamp.IsZero()).To(BeTrue())
 		})
--- a/pkg/controllers/node/suite_test.go
+++ b/pkg/controllers/node/suite_test.go
@@ -288,13 +288,13 @@ var _ = Describe("Controller", func() {
 		})
 		It("should requeue reconcile if node is empty, but not past emptiness TTL", func() {
 			provisioner.Spec.TTLSecondsAfterEmpty = ptr.Int64(30)
-			currentTime := time.Now()
-			injectabletime.Now = func() time.Time { return currentTime } // injectabletime.Now() is called multiple times in function being tested.
+			now := time.Now()
+			injectabletime.Now = func() time.Time { return now } // injectabletime.Now() is called multiple times in function being tested.
 			emptinessTime := injectabletime.Now().Add(-10 * time.Second)
 			// Emptiness timestamps are first formatted to a string friendly (time.RFC3339) (to put it in the node object)
 			// and then eventually parsed back into time.Time when comparing ttls. Repeating that logic in the test.
 			emptinessTimestamp, _ := time.Parse(time.RFC3339, emptinessTime.Format(time.RFC3339))
-			expectedrequeueTime := emptinessTimestamp.Add(time.Duration(30 * time.Second)).Sub(injectabletime.Now()) // we should requeue in ~20 seconds.
+			expectedRequeueTime := emptinessTimestamp.Add(time.Duration(30 * time.Second)).Sub(injectabletime.Now()) // we should requeue in ~20 seconds.
 			node := test.Node(test.NodeOptions{ObjectMeta: metav1.ObjectMeta{
 				Finalizers: []string{v1alpha5.TerminationFinalizer},
 				Labels:     map[string]string{v1alpha5.ProvisionerNameLabelKey: provisioner.Name},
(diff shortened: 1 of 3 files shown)


ellistarn:
Thoughts on solving https://github.com/aws/karpenter/issues/1103 at the same time, since I think it touches the same code.

suket22:
> Thoughts on solving #1103 at the same time, since I think it touches the same code.

That one is a little complicated and might need more thought on whether it needs to be fixed.

So our node controller watches for changes in the provisioner and if say `TTLSecondsAfterEmpty` was modified, a reconcile event for each node [should've been fired as per this](https://github.com/aws/karpenter/blob/main/pkg/controllers/node/controller.go#L133). However, this is where we run into expected `controller-runtime` behavior - since there's already a reconcile request for this node waiting (due to the original TTL) we drop our reconcile event. 

Essentially, we're only use the TTL to set the right `RequeueAfter` on our reconcile event. Once we tell controller-runtime "tell me about this node in an hour", I'm not sure how we can then tell controller-runtime "process this node right away" because it'll just dedupe any reconcile until the hour ends.

pkg/controllers/node/suite_test.go
```diff
@@ -285,6 +286,27 @@ var _ = Describe("Controller", func() {
...
 		})
+		It("should requeue reconcile if node is empty, but not past emptiness TTL", func() {
+			provisioner.Spec.TTLSecondsAfterEmpty = ptr.Int64(30)
+			currentTime := time.Now()
+			injectabletime.Now = func() time.Time { return currentTime } // injectabletime.Now() is called multiple times in function being tested.
+			emptinessTime := injectabletime.Now().Add(-10 * time.Second)
+			// Emptiness timestamps are first formatted to a string friendly (time.RFC3339) (to put it in the node object)
+			// and then eventually parsed back into time.Time when comparing ttls. Repeating that logic in the test.
+			emptinessTimestamp, _ := time.Parse(time.RFC3339, emptinessTime.Format(time.RFC3339))
+			expectedrequeueTime := emptinessTimestamp.Add(time.Duration(30 * time.Second)).Sub(injectabletime.Now()) // we should requeue in ~20 seconds.
```
njtran, line 297:
```suggestion
			emptinessTime := time.Now().Add(-10 * time.Second)
			// Emptiness timestamps are first formatted to a string friendly (time.RFC3339) (to put it in the node object)
			// and then eventually parsed back into time.Time when comparing ttls. Repeating that logic in the test.
			emptinessTimestamp, _ := time.Parse(time.RFC3339, emptinessTime.Format(time.RFC3339))
			expectedrequeueTime := 20
```

pkg/controllers/node/suite_test.go
```diff
@@ -285,6 +286,27 @@ var _ = Describe("Controller", func() {
...
+				Labels:     map[string]string{v1alpha5.ProvisionerNameLabelKey: provisioner.Name},
+				Annotations: map[string]string{
+					v1alpha5.EmptinessTimestampAnnotationKey: emptinessTime.Format(time.RFC3339),
+				}},
+			})
+			ExpectCreated(ctx, env.Client, provisioner, node)
+			ExpectReconcileSucceededWithResult(ctx, controller, client.ObjectKeyFromObject(node), reconcile.Result{Requeue: true, RequeueAfter: expectedrequeueTime})
+			node = ExpectNodeExists(ctx, env.Client, node.Name)
+			Expect(node.DeletionTimestamp.IsZero()).To(BeTrue())
+		})
```
njtran, line 309: Would you be able to add another unit test that replicates and targets the issue this was opened for?

The steps would be the following:
1) Create node with pod on it
2) Set Provisioner.TTLAfterEmpty to 30
3) Delete pod from node
4) Check Node.TTLAfterEmpty is set
5) Add another pod
6) Check Node.TTLAfterEmpty unset
7) remove second pod
8) Check Node.TTLAfterEmpty is set
9) Simulate 30 seconds of time passing
10) check that node deletion timestamp is nonZero

njtran (commented):
Looks good. Thanks for the change, just a couple comments.

pkg/test/expectations/expectations.go
```diff
@@ -190,3 +190,9 @@ func ExpectReconcileSucceeded(ctx context.Context, reconciler reconcile.Reconcil
 	_, err := reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: key})
 	Expect(err).ToNot(HaveOccurred())
 }
+
+func ExpectReconcileSucceededWithResult(ctx context.Context, reconciler reconcile.Reconciler, key client.ObjectKey, result reconcile.Result) {
```
ellistarn, line 194: I'd rather keep this logic inline until we use it broadly. The concern is about bloating the expectations library to have so many methods that it results in tests written many ways using similar helpers for the same effect.

pkg/controllers/node/suite_test.go
```diff
@@ -285,6 +286,27 @@ var _ = Describe("Controller", func() {
 			node = ExpectNodeExists(ctx, env.Client, node.Name)
 			Expect(node.DeletionTimestamp.IsZero()).To(BeFalse())
 		})
+		It("should requeue reconcile if node is empty, but not past emptiness TTL", func() {
+			provisioner.Spec.TTLSecondsAfterEmpty = ptr.Int64(30)
+			currentTime := time.Now()
+			injectabletime.Now = func() time.Time { return currentTime } // injectabletime.Now() is called multiple times in function being tested.
```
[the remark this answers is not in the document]
suket22: Kinda related to the below comment - I needed to do this because I had to repeat the math in the test since we lose precision in the timestamp conversions.

pkg/controllers/node/suite_test.go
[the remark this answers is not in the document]
suket22: You can't. The problem is when you do `time.Parse(time.RFC3339, emptinessTime.Format(time.RFC3339))` you lose precision and therefore it's not `20` but something close which is why I have the comment stating `~20`. 

```
[Expected
        <reconcile.Result>: {
            Requeue: true,
            RequeueAfter: 19960566000,
        }
    to equal
        <reconcile.Result>: {
            Requeue: true,
            RequeueAfter: 20000000000,
        }[
        ```

It's why I kept the test agnostic to specific numbers and just repeated the controller logic.

File: pkg/controllers/node/suite_test.go (hunk 2 above)
Comment (suket22) [reply pkg/controllers/node/suite_test.go:309]: I don't think that specific scenario is suitable for a unit test IMO. I think we should be testing what our code expects to do (i.e set a RequeueAfter with value foo) rather than test every possible permutation.

File: pkg/test/expectations/expectations.go (hunk 3 above)
Comment (suket22) [reply pkg/test/expectations/expectations.go:194]: Given the back and forth and the time it took to get this RequeueAfter logic right, I fully expect (pun intended) us to verify this in other controllers too which is why I added it here. 

Don't feel too strongly about it though

Comment (ellistarn) [reply pkg/test/expectations/expectations.go:194]: What about returning a result that can be optionally used for future assertions? Some other expectations work this way:
```
func ExpectReconcileSucceeded(ctx context.Context, reconciler reconcile.Reconciler, key client.ObjectKey) reconcile.Result {
	result, err := reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: key})
        return result
}
```
This won't force any existing code to change, either.

File: pkg/controllers/node/suite_test.go [hunk 5]
```diff
@@ -285,6 +286,28 @@ var _ = Describe("Controller", func() {
...
+			expectedrequeueTime := emptinessTimestamp.Add(time.Duration(30 * time.Second)).Sub(injectabletime.Now()) // we should requeue in ~20 seconds.
+			node := test.Node(test.NodeOptions{ObjectMeta: metav1.ObjectMeta{
+				Finalizers: []string{v1alpha5.TerminationFinalizer},
+				Labels:     map[string]string{v1alpha5.ProvisionerNameLabelKey: provisioner.Name},
+				Annotations: map[string]string{
+					v1alpha5.EmptinessTimestampAnnotationKey: emptinessTime.Format(time.RFC3339),
+				}},
+			})
+			ExpectCreated(ctx, env.Client, provisioner, node)
+			reconcileResult := ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(node))
```
ellistarn, line 306: nit: go style recommends to use single words unless additional qualification is necessary given the scope, e.g. `result`

pkg/controllers/node/suite_test.go
```diff
@@ -285,6 +286,28 @@ var _ = Describe("Controller", func() {
...
+			Expect(reconcileResult).To(Equal(reconcile.Result{Requeue: true, RequeueAfter: expectedrequeueTime}))
```
ellistarn, line 307: nit: expectedrequeueTime has weird casing. Further, the variable is only used once, so I might just do it inline (optional).

pkg/controllers/node/suite_test.go
```diff
@@ -285,6 +286,28 @@ var _ = Describe("Controller", func() {
 			node = ExpectNodeExists(ctx, env.Client, node.Name)
 			Expect(node.DeletionTimestamp.IsZero()).To(BeFalse())
 		})
+		It("should requeue reconcile if node is empty, but not past emptiness TTL", func() {
+			provisioner.Spec.TTLSecondsAfterEmpty = ptr.Int64(30)
+			currentTime := time.Now()
```
ellistarn, line 291: optional:`now :=`

ellistarn (approved):

PR #1270, opened 2022-02-03
Design Doc on AWS Provider LaunchTemplate evolution

suket22:
**1. Issue, if available:**
N/A

We've had a lot of requests around launch template functionality. This document goes over some options we can consider for the AWS Provider. 

N/A

- [x] No


designs/aws/aws-launch-templates-v2.md
```diff
@@ -0,0 +1,146 @@
...
+* The Provider API starts to have too many configuration knobs and becomes difficult to document and use.
+* Difficult messaging for users - how do we define what is and isn't supported on the provisioner? Forcing a user to define an entire launch template as an alternative is a lot of friction.
+
+
+**Approach 2 - Accept partial launch templates and fill in the rest.**
+
+This approach is inspired by Managed Nodegroups  - we let a customer only specify a certain set of properties in their launch template and let Karpenter automatically figure out the rest. That way a customer needing to only update the sizes of their root volumes needs to specify a much smaller and easier to maintain launch template like -
+
+```
+aws ec2 describe-launch-template-versions --launch-template-id lt-123
```
bwagner5, line 56: is this a valid launch template?  We'd need to copy this data into our own launch template which would be difficult if we're using `$LATEST` or `$DEFAULT` to track the LT since it could change while we're copying.

suket22: It is a valid launch template - but yes we'd need to copy that data into our own. 

> if we're using $LATEST or $DEFAULT to track the LT since it could change while we're copying.

I think that semantic isn't too surprising. If the alias they've given us to use changes which version it's pointing to, we'd change which LT we copy from too. If we're very concerned about this, we shouldn't let folks provide a version alias to Karpenter at all.

Fwiw, the pattern I've seen most folks adopt in the past is to let their infra tool resolve the alias and have an actual versionNumber passed to ASG / EC2 etc. So if you were using CloudFormation you do a `Fn::GetAtt MyLaunchTemplate.LatestVersionNumber` or something like that.

designs/aws/aws-launch-templates-v2.md
```diff
@@ -0,0 +1,146 @@
...
+The rationale we’ve stuck to so far is
+
+1. We only support basic use cases that should apply to most users. Categorizing a use case as basic has been somewhat arbitrary, but we’ve used upvotes on open issues and PRs to get that signal.
+2. At the minimum, functionality that other AWS provisioners like ManagedNodegroups give you without having to use a launch template should be available through Karpenter as well. Most of the parameters we’ve recently added fit this tenet.
+
+### How do we evolve LT usage?
+
+We’ve considered a few different approaches for this so far.
+
+**Approach 1 - Continue as is. Inline fields as necessary and enforce complete launch templates otherwise.**
```
olemarkus, line 36: There is a point worth mentioning here is the relationship between kubelet settings, assumptions made by karpenter, and launch templates.
With karpenter-managed LTs, karpenter dictates kubelet settings through LTs matching its own logic. But with provided LTs, there is often a mismatch between karpenter assumptions and actuals. MaxPods is one example, ephemeral disk size, pod overhead, cgroup limits are others.

So for approach 1, I think it is quite important that karpenter is also configurable so that they match relevant kubelet settings and LT parameters.

designs/aws/aws-launch-templates-v2.md
```diff
@@ -0,0 +1,146 @@
...
+2. Partially filled out launch templates are difficult to interpret. Anyone specifying a custom AMI to use will also need to explicitly specify how it needs to be bootstrapped via UserData. This is important because Karpenter cannot make guesses on how to start the kubelet on an unknown AMI.
+
+The main, critical downside of this approach has been that a user needs to maintain a fully-formed launch template even when they’re only looking to customize a single property of their worker nodes. Consider these examples -
+
+1. In order to override [the root volume](https://github.com/aws/karpenter/issues/885), you’d need to author an entire launch template containing security groups, userData etc.
+2. In order to just have all statically provisioned [volumes be encrypted](https://github.com/aws/karpenter/issues/933), you need to author an entire launch template.
+
+While some users do like complete control over the launch template since they customize everything (kOps), most only need to override a few settings and would therefore like those to be present as knobs on the provisioner itself, and for all other settings to be figured out by Karpenter.
```
olemarkus, line 20: The challenge here is really that karpenter assumes EKS and anything that is not EKS need different launch templates. Rancher, OKD etc all will have this requirement. Had Karpenter been made to work firstly with kOps, EKS would have this requirement.

I don't think it is feasible that Karpenter truly can manage LTs that will work across all these installers, and appreciate that for EKS this is needed since there is no tooling for managing LTs. But other installers do provide that tooling. So certainly no kOps _user_ will _want_ to manage their own LTs. In fact, kOps do not allow custom LTs, and I haven't seen a single request for such a feature (our approach is somewhat similar to approach 2 below). End goal being sort of the same as this aspect of Karpenter: hide the complexity of LTs.

olemarkus (dismissed):

designs/aws/aws-launch-templates-v2.md
ellistarn: This suggests to me that we probably need to `provider: Kops` or `provider: AWSKops` implementation, since they vary so much.

designs/aws/aws-launch-templates-v2.md
```diff
@@ -0,0 +1,146 @@
...
+* You get the full ability of EC2 launch template right within your Provider
+* You don't need to maintain a launch template at all. Any new EC2 feature will work out of the box as long as the deserialization logic uses the latest AWS SDK i.e keep the controller up to date.
+* It most directly addresses what every customer is asking - let them configure EC2 properties from within the provisioner.
+
+*Cons*
+* APIs should really never accept weakly typed input. It's going to be very easy for a customer to get something wrong what went wrong break the deserialization. When that happens, it might be equally difficult to debug YAML.
+* If customers have to fill out the whole launch templates, it's going to make the provisionerSpec verbose. If they only fill it out partially, then you have the same cons as approach 2.
+
+
+
```
bwagner5, line 139: The 4th approach we talked about at the community meeting is specifying more parameters in the podspec as `karpenter.sh` namespaced labels. Can we explore this more?

designs/aws/aws-launch-templates-v2.md
suket22: Right that makes sense, I'll add it to the cons list. The time lag between the 'source' launch template being updated and us copying the updates over to the launch template we manage is likely going to be unpredictable - although I'm hoping it's in the order of ~seconds. We update our instance cache every 10 seconds, so maybe a LT cache in < 60 seconds wouldn't be too unreasonable.

designs/aws/aws-launch-templates-v2.md
suket22: I'll add that in. @ellistarn that's pretty much what we discussed at the last working group. It's nice to have a separation of responsibilities and have the application developer only define what their application requires and not have entire control over the actual infrastructure. It'd make it much harder to enforce things like "all our nodes should be using AMIs XX with Security Tooling YY" if every app developer needs to do per podSpec.

designs/aws/aws-launch-templates-v2.md
suket22: Right, good call let me add that in.  One of the first fields we do intend to make configurable inline is the UserData so the kubelet can always be fully customized without needing a provided LT. 

This is going to be tricky to implement but I expect Karpenter to ignore any setting that would compete with an inline UserData override like specProvider.kubeletConfiguration or even maxPods

designs/aws/aws-launch-templates-v2.md
```diff
@@ -0,0 +1,163 @@
...
+  * Along the same lines, should the application developer always have complete control over their infrastructure? Some configuration like ImageIds might need to be enforced by the cluster administrator, so it would be beneficial to never expose ImageIds as a knob to the app developer at all.
+* By moving instance configuration from the provider label to the provisioner label, each separate cloud provider implementation will need to support that label which introduces unneeded complexity. The feature set across all cloud providers is unlikely to ever be homogenous.
+
+
+### Recommendation
+
+I recommend we continue with Approach 1, and give everyone the ability to customize a few select, commonly used fields in the provisionerSpec and ask them to configure all others via a complete LaunchTemplate. I think this keeps our API easy to use. The fields I expect us to add to the Provider in the short term are _BlockDeviceMappings, ImageId / AMIFamily and UserData._ The use cases for other fields either don’t apply to Karpenter, or they’re quite niche where it probably makes sense for the customer to manage all other aspects of the launch template in any case.
```
alex-berger, line 160: Personally, I am more in favor of Approach 3 as it is more flexible (allows for more use-cases) without having to change karpenter's code base again and again. I am looking at this from an operator's (not developer's) point-of-view, in which it is often frustrating and annoying that developers do not understand that operator's (customer's) need for flexible tools (and not just opinionated and limited tools that don't get the job done).

suket22: I try and call this out early in the doc, but the core reason I'm recommending the first approach is because we often see a need to only override *one* or *two* of the AWS Provider's opinionated defaults, so defining an entire launch template is too heavy handed. 

But if you've got more advanced needs that frequently change, maintaining an entire launch template shouldn't be too cumbersome and likely similar to embedding all of that information within the provisioner.

ellistarn:
I'd like to see some discussion around the complexity of the bootstrap.sh script in eksoptimized ami, as well as partial user data. e.g., bottlerocket userdata is pretty easy to share, since it's a toml file that can be merged or appended easily. EKSOptimizedAMI is much more challenging since the bootstrap script may have user args (iprefix) and karpenter args (labels/taints) applied.

suket22:
> I'd like to see some discussion around the complexity of the bootstrap.sh script in eksoptimized ami, as well as partial user data. e.g., bottlerocket userdata is pretty easy to share, since it's a toml file that can be merged or appended easily. EKSOptimizedAMI is much more challenging since the bootstrap script may have user args (iprefix) and karpenter args (labels/taints) in the same arg string.

Agreed. I'm going to add a separate doc to flesh out the different options on what the support of AMI and UserData is going to look like. Sorry about the delay on adding that in here.

bwagner5 (approved):
lgtm!

PR #1278, opened 2022-02-04
AWS launch template deletion on cache eviction

bwagner5:
**1. Issue, if available:**
N/A

 - The LaunchTemplate provider in the AWS Cloud Provider will not delete launch templates when they are evicted from the provider's local cache. This helps to keep the number of dynamically generated launch templates under control. 

 - Tested via a manual cluster deployment. I set the Cache TTL and Cache Cleanup Interval to 2 seconds and did a series of launches that would result in different launch templates generated. All launch templates were cleaned up w/ no launch issues. 

- [ ] No

### DIFF
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -66,6 +66,7 @@ type CloudProvider struct {
 }
 
 func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *CloudProvider {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("aws"))
 	sess := withUserAgent(session.Must(session.NewSession(
 		request.WithRetryer(
 			&aws.Config{STSRegionalEndpoint: endpoints.RegionalSTSEndpoint},
@@ -85,6 +86,7 @@ func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *Cloud
 		subnetProvider:       subnetProvider,
 		instanceProvider: &InstanceProvider{ec2api, instanceTypeProvider, subnetProvider,
 			NewLaunchTemplateProvider(
+				ctx,
 				ec2api,
 				NewAMIProvider(ssm.New(sess), options.ClientSet),
 				NewSecurityGroupProvider(ec2api),
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -65,6 +65,7 @@ func NewLaunchTemplateProvider(ctx context.Context, ec2api ec2iface.EC2API, amiP
 		cache:                 cache.New(CacheTTL, CacheCleanupInterval),
 	}
 	l.cache.OnEvicted(l.onCacheEvicted)
+	l.hydrateCache(ctx)
 	return l
 }
 
@@ -215,6 +216,24 @@ func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, optio
 	}
 	logging.FromContext(ctx).Debugf("Created launch template, %s", *output.LaunchTemplate.LaunchTemplateName)
 	return output.LaunchTemplate, nil
+}
+
+// hydrateCache queries for existing Launch Templates created by Karpenter for the current cluster and adds to the LT cache.
+// Any error during hydration will result in a panic
+func (p *LaunchTemplateProvider) hydrateCache(ctx context.Context) {
+	queryKey := fmt.Sprintf(launchTemplateNameFormat, injection.GetOptions(ctx).ClusterName, "*")
+	p.logger.Debugf("Hydrating the launch template cache with names matching \"%s\"", queryKey)
+	if err := p.ec2api.DescribeLaunchTemplatesPagesWithContext(ctx, &ec2.DescribeLaunchTemplatesInput{
+		Filters: []*ec2.Filter{{Name: aws.String("launch-template-name"), Values: []*string{aws.String(queryKey)}}},
+	}, func(output *ec2.DescribeLaunchTemplatesOutput, _ bool) bool {
+		for _, lt := range output.LaunchTemplates {
+			p.cache.SetDefault(*lt.LaunchTemplateName, lt)
+		}
+		return true
+	}); err != nil {
+		panic(fmt.Sprintf("Unable to hydrate the AWS launch template cache, %s", err.Error()))
+	}
+	p.logger.Debugf("Finished hydrating the launch template cache with %d items", p.cache.ItemCount())
 }
 
 func (p *LaunchTemplateProvider) onCacheEvicted(key string, lt interface{}) {
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -57,15 +57,15 @@ type LaunchTemplateProvider struct {
 }
 
 func NewLaunchTemplateProvider(ctx context.Context, ec2api ec2iface.EC2API, amiProvider *AMIProvider, securityGroupProvider *SecurityGroupProvider) *LaunchTemplateProvider {
-	ltProvider := &LaunchTemplateProvider{
+	l := &LaunchTemplateProvider{
 		ec2api:                ec2api,
 		logger:                logging.FromContext(ctx).Named("launchtemplate"),
 		amiProvider:           amiProvider,
 		securityGroupProvider: securityGroupProvider,
 		cache:                 cache.New(CacheTTL, CacheCleanupInterval),
 	}
-	ltProvider.cache.OnEvicted(ltProvider.onCacheEvicted)
-	return ltProvider
+	l.cache.OnEvicted(l.onCacheEvicted)
+	return l
 }
 
 func launchTemplateName(options *launchTemplateOptions) string {
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -59,7 +59,7 @@ type LaunchTemplateProvider struct {
 func NewLaunchTemplateProvider(ctx context.Context, ec2api ec2iface.EC2API, amiProvider *AMIProvider, securityGroupProvider *SecurityGroupProvider) *LaunchTemplateProvider {
 	ltProvider := &LaunchTemplateProvider{
 		ec2api:                ec2api,
-		logger:                logging.FromContext(ctx).Named("launchTemplateProvider"),
+		logger:                logging.FromContext(ctx).Named("launchtemplate"),
 		amiProvider:           amiProvider,
 		securityGroupProvider: securityGroupProvider,
 		cache:                 cache.New(CacheTTL, CacheCleanupInterval),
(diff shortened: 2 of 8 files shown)


pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -47,19 +49,23 @@ const (
...
 	ec2api                ec2iface.EC2API
 	amiProvider           *AMIProvider
 	securityGroupProvider *SecurityGroupProvider
 	cache                 *cache.Cache
 }
 
-func NewLaunchTemplateProvider(ec2api ec2iface.EC2API, amiProvider *AMIProvider, securityGroupProvider *SecurityGroupProvider) *LaunchTemplateProvider {
-	return &LaunchTemplateProvider{
+func NewLaunchTemplateProvider(ctx context.Context, ec2api ec2iface.EC2API, amiProvider *AMIProvider, securityGroupProvider *SecurityGroupProvider) *LaunchTemplateProvider {
+	ltProvider := &LaunchTemplateProvider{
```
ellistarn, line 60:
```suggestion
	l := &LaunchTemplateProvider{
```

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -47,19 +49,23 @@ const (
...
 		ec2api:                ec2api,
+		logger:                logging.FromContext(ctx).Named("launchTemplateProvider"),
```
ellistarn, line 62: consider `launchtemplate`. camelCase logger names feel icky to me.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -211,6 +218,20 @@ func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, optio
...
 
+func (p *LaunchTemplateProvider) onCacheEvicted(key string, lt interface{}) {
+	p.Lock()
+	defer p.Unlock()
+	if _, expiration, _ := p.cache.GetWithExpiration(key); expiration.After(time.Now()) {
+		return
+	}
+	launchTemplate := lt.(*ec2.LaunchTemplate)
+	if _, err := p.ec2api.DeleteLaunchTemplate(&ec2.DeleteLaunchTemplateInput{LaunchTemplateId: launchTemplate.LaunchTemplateId}); err != nil {
+		p.logger.Debugf("unable to delete launch template, %v", err)
```
ellistarn, line 229: Should this be an error?

ellistarn: All of our other log lines start with a capital letter.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -211,6 +218,20 @@ func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, optio
...
+		return
+	}
+	p.logger.Debugf("deleted launch template %v", *launchTemplate.LaunchTemplateId)
```
ellistarn, line 232: optional: aws.StringValue

website/content/en/preview/getting-started/cloudformation.yaml
```diff
@@ -48,6 +48,7 @@ Resources:
               - ec2:CreateTags
               - iam:PassRole
               - ec2:TerminateInstances
+              - ec2:DeleteLaunchTemplate
```
ellistarn, line 51: We should make some upgrade instructions to help users migrate.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -211,6 +218,20 @@ func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, optio
 	return output.LaunchTemplate, nil
 }
 
+func (p *LaunchTemplateProvider) onCacheEvicted(key string, lt interface{}) {
```
ellistarn, line 221: I guess we leak if we miss evictions?

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -211,6 +218,20 @@ func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, optio
...
+	p.Lock()
```
ellistarn, line 222: I don't think we need to lock since the cache eviction is already threadsafe.

bwagner5: cache eviction is threadsafe, however onCacheEvicted is not threadsafe. For example, if an LT is evicted from the cache, ensureLaunchTemplate can be executed that will receive a cache miss, and then query LTs and find the LT that was evicted but has not been deleted yet. After it finds that LT, onCacheEvicted can run (if the lock is removed) and delete the LT before it is used which will propagate as a launch failure in Fleet.

pkg/cloudprovider/aws/launchtemplate.go
bwagner5: I went back-and-forth on making this Error or not. I'm fine making it an Error but the reason I didn't is because this is not strictly required to operate correctly and we are adding a permission. But it's probably better to alert users sooner to update permissions if they miss it in the release notes.  I'm convinced again on making it error after writing this :) Updated and updated.

pkg/cloudprovider/aws/launchtemplate.go
bwagner5: Yes, and this can happen when Karpenter shuts down...  I thought about adding a shutdown hook and draining the cache... but I'm not sure it's worth it to do that.  

Another option I just thought of is loading all LTs into the cache that Karpenter has created by querying the Karpenter tag so that when Karpenter starts back up, it will load in all LTs and then they will naturally be evicted if they are not used.

website/content/en/preview/getting-started/cloudformation.yaml
bwagner5: we definitely should... not sure how we'd structure them. I was thinking release notes, but another way would be adding an upgrade section in the versioned docs.

pkg/cloudprovider/aws/launchtemplate.go
ellistarn: I quite like this, actually. Hydrate the LT cache on startup. We can panic if it fails, since it's initialization logic.

pkg/cloudprovider/aws/launchtemplate.go
ellistarn: The lock isn't saving us in this case, since the LT creation has already happened. The only thing saving us from this race is the expiration timeout -- I guess this is fine (necessary?).

website/content/en/preview/getting-started/cloudformation.yaml
ellistarn: I just want a single command to run which updates my IAM. Maybe cfn deploy works out of the box.

bwagner5: yeah cfn should work fine, but I suspect most users are integrating our template (cfn or terraform or cdk) into their own infrastructure-as-code so there won't be a one-size fits all solution.

pkg/cloudprovider/aws/launchtemplate.go
bwagner5: the lock does save us in the scenario I mentioned. The LT creation I mentioned would occur in the `ensureLaunchTemplate` func which also takes a lock. Locking in both of these funcs ensures that the cache is always consistent with the state of EC2.

pkg/cloudprovider/aws/launchtemplate.go
bwagner5: 👍  Cool, I'll do the cache hydration. Since we don't currently tag the LT with anything specific, this will only work for LTs created after this update is installed :(

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -369,6 +354,21 @@ var _ = Describe("Allocation", func() {
 			})
 		})
 		Context("LaunchTemplates", func() {
+			It("should hydrate the launch template cache on instantiation of the LaunchTemplateProvider", func() {
```
bwagner5, line 357: these honestly aren't very good tests... let me know if you want me to keep these or not. They're in a separate commit, so I can remove all the test mods I made :)

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -66,6 +66,7 @@ type CloudProvider struct {
 }
 
 func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *CloudProvider {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("aws"))
```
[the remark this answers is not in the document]
ellistarn: This is just a bunch, but do the logs indent more cleanly if if name this `cloudprovider.aws` since the width is more similar to `provisioning` `selection`

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -47,19 +49,24 @@ const (
...
+func NewLaunchTemplateProvider(ctx context.Context, ec2api ec2iface.EC2API, amiProvider *AMIProvider, securityGroupProvider *SecurityGroupProvider) *LaunchTemplateProvider {
+	l := &LaunchTemplateProvider{
 		ec2api:                ec2api,
+		logger:                logging.FromContext(ctx).Named("launchtemplate"),
 		amiProvider:           amiProvider,
 		securityGroupProvider: securityGroupProvider,
 		cache:                 cache.New(CacheTTL, CacheCleanupInterval),
 	}
+	l.cache.OnEvicted(l.onCacheEvicted)
+	l.hydrateCache(ctx)
```
ellistarn, line 68: This ctx won't benefit from the annotated logger. You need to do something like `ctx = ...` on line 60.

ellistarn: Oh perhaps this doesn't matter.

bwagner5: yeah, it's not async so I was just going to use the ctx passed for initialization

pkg/cloudprovider/aws/cloudprovider.go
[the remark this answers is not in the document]
bwagner5: I'd rather keep it AWS since it's just startup logic so the logging is actually nice. When it actually matters in the async process the indentation is nice since it's `aws.launchtemplate` which lines up with `provisioning` etc

njtran (approved):
Nice!

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -210,6 +218,38 @@ func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, optio
...
+
+func (p *LaunchTemplateProvider) onCacheEvicted(key string, lt interface{}) {
+	p.Lock()
+	defer p.Unlock()
+	if _, expiration, _ := p.cache.GetWithExpiration(key); expiration.After(time.Now()) {
+		return
+	}
+	launchTemplate := lt.(*ec2.LaunchTemplate)
+	if _, err := p.ec2api.DeleteLaunchTemplate(&ec2.DeleteLaunchTemplateInput{LaunchTemplateId: launchTemplate.LaunchTemplateId}); err != nil {
+		p.logger.Errorf("Unable to delete launch template, %v", err)
```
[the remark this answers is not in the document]
bwagner5: It's not retried (doesn't reconcile) since this is only executed on cache eviction, BUT it does rehydrate on startup, so if something did happen, a restart of Karpenter would clean them up.

Issue #1282, opened 2022-02-07
Karpenter fails to create node object if an instance does not have PrivateDnsName still an issue


bwmetcalf:
## Version
**Karpenter**: v0.5.4

**Kubernetes**: v1.21.5

## Actual Behavior
We have seen this issue twice in the past week running 0.5.4.  According to #1067, the issue is fixed in 0.5.4.

```2022-02-07T15:40:39.115Z        ERROR   controller.provisioning Could not launch node, launching instances, All attempts fail:
#1: got instance i-0eaa445d1c9aa01bb but PrivateDnsName was not set
#2: got instance i-0eaa445d1c9aa01bb but PrivateDnsName was not set
#3: got instance i-0eaa445d1c9aa01bb but PrivateDnsName was not set     {"commit": "7e79a67", "provisioner": "gitlab-runner-provisioner-karpenter-provisioner"}
```

## Resource Specs and Logs
Same issue as #1067

ellistarn:
To be clear from that issue, this is an unfortunate race condition that EC2 sometimes takes a long time to propagate the private dns name. If you're using karpenter owned launch templates, it should still register and work since karpenter injects all the expected labels into the userdata. If you're using a custom launch template, make sure you include `karpenter.sh/provisioner-name` as a label passed to the kubelet. 

I think we could explore increasing the retries/timeout on EC2, as well.

ellistarn:
To be clear from that issue, this is an unfortunate race condition that EC2 sometimes takes a long time to propagate the private dns name. If you're using karpenter owned launch templates, it should still register and work since karpenter injects all the expected labels into the userdata. If you're using a custom launch template, make sure you include `karpenter.sh/provisioner-name` as a label passed to the kubelet. 

I think we could explore increasing the retries/timeout on EC2, as well.

felix-zhe-huang:
Is there a known way to reproduce this bug constantly?

PR #1286, opened 2022-02-07
Double the node creation retry

felix-zhe-huang:
**1. Issue, if available:**
When EC2 takes a longer time to assign the private DNS name of the newly created instance, Karpenter may failed all attends to create the node object due to missing private DNS name.
#1282 

Double the get instance retry to 6 times, one second apart.
This will double the time that a private DNS name can be assigned before Karpenter raises an error.

Unit tests, manually tested.

- [X] No

### DIFF
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -59,7 +59,7 @@ func (p *InstanceProvider) Create(ctx context.Context, constraints *v1alpha1.Con
 	if err := retry.Do(
 		func() (err error) { instances, err = p.getInstances(ctx, ids); return err },
 		retry.Delay(1*time.Second),
-		retry.Attempts(3),
+		retry.Attempts(6),
 	); err != nil && len(instances) == 0 {
 		return nil, err
 	} else if err != nil {


pkg/cloudprovider/aws/instance.go
```diff
@@ -59,7 +59,7 @@ func (p *InstanceProvider) Create(ctx context.Context, constraints *v1alpha1.Con
 	if err := retry.Do(
 		func() (err error) { instances, err = p.getInstances(ctx, ids); return err },
 		retry.Delay(1*time.Second),
-		retry.Attempts(3),
+		retry.Attempts(6),
```
ellistarn, line 62: Does this grow exponentially? It would also be nice to reduce the delay to something like p50 propagation time.

ellistarn: Hmm -- maybe we don't want to increase because the first call is 0 seconds of delay.

felix-zhe-huang: So if I understand this correctly, we need to increase the overall time before raising an error so that AWS can have enough time to assign the private DNS name. So if we also shorten the delay, we do not increase the overall time window but just increase the sampling number, that doesn't necessary help. 

As for growing exponentially can you elaborate a bit more? I think it will just try 3 more times with constant runtime in each iteration. And for most of the cases we shouldn't even see retry. So the worst case scenario is that we will see a 6 second delay. But this should be extremely rare.

ellistarn: I don't have good data on tail latencies for this propagation, but it's not uncommon for tail latencies to be quite a bit longer than p50 latency. e.g., I could imagine that p50 = 100ms, but p99 = 30 seconds. Since tails latencies tend to be logarithmic, an exponential is a decent model to minimize the number of retries in all cases.

ellistarn (approved):

pkg/cloudprovider/aws/instance.go
tzneal: I looked at avast/retry and it has exponential backoff built in by [default](https://github.com/avast/retry-go/blob/65a9fd47eb133d84be62342e1c778ccd966c116b/retry.go#L158).

A quick test shows that is works, and using a delay of 1 second with 6 attempts yields the desired 30 second delay:

```
try count 0 current delay 0s cumulative delay 600ns
try count 1 current delay 1.048829547s cumulative delay 1.048830347s
try count 2 current delay 2.08424261s cumulative delay 3.133073157s
try count 3 current delay 4.06698648s cumulative delay 7.200060237s
try count 4 current delay 8.04234009s cumulative delay 15.242400427s
try count 5 current delay 16.090066559s cumulative delay 31.332467186s
```

PR #1287, opened 2022-02-07
Improve the error message readability

felix-zhe-huang:
**1. Issue, if available:**
Currently the error messages for requirements validation is not very meaningful.

Provide more detail errors.

- [X] No


felix-zhe-huang:
Example of API error messages:
Screen Shot 2022-02-07 at 6 16 33 PM

Example of runtime error messages:
Screen Shot 2022-02-07 at 6 17 33 PM

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -54,7 +55,11 @@ func (c *Constraints) ValidatePod(pod *v1.Pod) error {
 	// Test if pod requirements are valid
 	requirements := NewPodRequirements(pod)
 	if errs := requirements.Validate(); errs != nil {
-		return fmt.Errorf("pod requirements not feasible, %v", errs)
+		var multiErrors error
+		for _, err := range errs {
+			multiErrors = multierr.Append(multiErrors, fmt.Errorf(err))
```
ellistarn, line 60: IMO this function shouldn't be aware of the multierr. Encapsulate this inside requirements.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -54,7 +55,11 @@ func (c *Constraints) ValidatePod(pod *v1.Pod) error {
...
+		}
+		return fmt.Errorf("pod requirements not feasible, %w", multiErrors)
```
ellistarn, line 62: This reads poorly. `are not feasible` is more correct, but that's pretty verbose. I might just say "invalid requirements, %w".

With the caller of this, it would read as, `Unable to schedule pod foo, invalid requirements, reasons...`

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -137,18 +138,25 @@ func (c *Constraints) validateTaints() (errs *apis.FieldError) {
 // When this function is called, the provisioner's requirments do not include the requirements from labels.
 // Provisioner requirements only support well known labels.
 func (c *Constraints) validateRequirements() (errs *apis.FieldError) {
+	var multiErrors error
```
ellistarn, line 141:
```suggestion
	var errs error
```

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -137,18 +138,25 @@ func (c *Constraints) validateTaints() (errs *apis.FieldError) {
...
 		}
 		// Ensure requirements operator is allowed
 		if !SupportedProvisionerOps.Has(string(requirement.Operator)) {
-			errs = errs.Also(apis.ErrInvalidKeyName(fmt.Sprintf("%s not in %v", requirement.Operator, SupportedProvisionerOps.UnsortedList()), "key"))
+			multiErrors = multierr.Append(multiErrors, fmt.Errorf("key %s has an unsupported operator %s, provisioner only supports %s", requirement.Key, requirement.Operator, SupportedProvisionerOps.UnsortedList()))
 		}
 	}
-	if err := c.Requirements.Validate(); err != nil {
-		errs = errs.Also(err)
+	if errors := c.Requirements.Validate(); errors != nil {
```
ellistarn, line 152: The beauty of multierr append is that it collapses nil errors. You can squash all of this into 


return multierr.Append(errs, c.Requirements.Validate())

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -202,27 +202,30 @@ func (r Requirements) CapacityTypes() stringsets.String {
 }
 
 // Validate validates the feasibility of the requirements.
-func (r Requirements) Validate() (errs *apis.FieldError) {
-	for i, requirement := range r.Requirements {
+//gocyclo:ignore
+func (r Requirements) Validate() (errs []string) {
```
ellistarn, line 206: I feel pretty strongly that we should model these as true errors. We can model this plainly using multierr as well.
```suggestion
func (r Requirements) Validate() (error) {
```

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -202,27 +202,30 @@ func (r Requirements) CapacityTypes() stringsets.String {
...
+	for _, requirement := range r.Requirements {
 		for _, err := range validation.IsQualifiedName(requirement.Key) {
-			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("key %s, %s", requirement.Key, err), "key"))
+			errs = append(errs, fmt.Sprintf("key %s is not a qualified name, %s", requirement.Key, err))
```
ellistarn, line 209:
```suggestion
                         errs = multierr.Append(errs, fmt.Errorf("key %s is not a qualified name, %s", requirement.Key, err))) 
```

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -202,27 +202,30 @@ func (r Requirements) CapacityTypes() stringsets.String {
...
 		}
 		for _, value := range requirement.Values {
 			for _, err := range validation.IsValidLabelValue(value) {
-				errs = errs.Also(apis.ErrInvalidArrayValue(fmt.Sprintf("key %s, value %s, %s", requirement.Key, value, err), "values", i))
+				errs = append(errs, fmt.Sprintf("key %s has an invalid value %s, %s", requirement.Key, value, err))
 			}
 		}
 		if !SupportedNodeSelectorOps.Has(string(requirement.Operator)) {
-			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("%s not in %s for %s", requirement.Operator, SupportedNodeSelectorOps.UnsortedList(), requirement.Key), "operator"))
+			errs = append(errs, fmt.Sprintf("key %s has an unsupported operator %s, supported operators are %s", requirement.Key, requirement.Operator, SupportedNodeSelectorOps.UnsortedList()))
```
ellistarn, line 217: Why are we moving away from %s not in %s.

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -202,27 +202,30 @@ func (r Requirements) CapacityTypes() stringsets.String {
...
 			r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpExists))) {
-			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("operator %s and %s conflict for %s", v1.NodeSelectorOpDoesNotExist, v1.NodeSelectorOpDoesNotExist, requirement.Key), "operator"))
+			errs = append(errs, fmt.Sprintf("operator %s and %s conflict for key %s", v1.NodeSelectorOpDoesNotExist, v1.NodeSelectorOpDoesNotExist, requirement.Key))
 		}
-		if r.Get(requirement.Key).Len() == 0 {
-			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("no feasible value for %s", requirement.Key), "values"))
+	}
+	for key := range r.Keys() {
+		if r.Get(key).Len() == 0 {
+			errs = append(errs, fmt.Sprintf("requirements render no feasible value for key %s", key))
```
ellistarn, line 228: The caller already states `requirements`, right?

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: IMO we need to give users a hint on what's failing. (In this case, the constraints from the pod.Spec is illy constructed.)
This is mainly catching the provisioner matching errors. So it would read as.
`Could not schedule pod, matched 0/3 provisioners, tried provisioner/default: pod requirements not feasible, some reason`

If we just say "invalid requirements" in the error message, we won't be able to tell if it is caused by provisioner requirements or pod requirements.

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: I was thinking it may be nice to allow functions that raise errors to determine which message among all the error messages should be reported at runtime.
This is the same model that the runtime validation API is using (see `validation.IsQualifiedName(key)`).

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
felix-zhe-huang: Since c.Requirements.Validate() returns []string, we still need to parse them to error in a for loop.
If we later decide to encapsulate the multierror in requirements, we can change this together.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: What's your opinion about keeping it consistent with
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go

I am open to either way.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: No, the upper layer is either "Validation" or "provisioner".
For example, `Could not schedule pod, matched 0/3 provisioners, tried provisioner/default: pod requirements not feasible, requirements render no feasible value for key XXX`

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -137,18 +138,25 @@ func (c *Constraints) validateTaints() (errs *apis.FieldError) {
...
+		for _, err := range errors {
+			fmt.Printf("\n=============\n%s\n", err)
```
felix-zhe-huang, line 154: LOL, will be removed.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: I was trying `"%s not in %s for key %s ` and it reads `Gt not in [In, NotIn Exists, DoesNotExist] for key XXX`. I feel that the customers may not immediately realize that we are talking about operator violation.  I think it is more readable to say "Key $s has an unsupported operator". 
I am open to change it back.

ellistarn: What about `operator %s not in %s for key ...`

pkg/apis/provisioning/v1alpha5/constraints.go
ellistarn: Oh I see -- I think this problem is caused by not all consumers logging in an intelligible way. As is, your change would repeat the word pod: `unable to schedule pod foo, pod requirements...`. 

In your error `Could not schedule pod`, I think it's already pretty well contextualized given this prefix.

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
ellistarn: The multierror library is designed to encapsulate this complexity for us.

pkg/apis/provisioning/v1alpha5/requirements.go
ellistarn: apimachinery's requirements are different, as they need to be careful about their dependency closure. They expect many packages to rely on them, so adding something like multierr results in huge dependency exposure. This problem doesn't exist for us, and all other validators in the codebase return `error`.

pkg/apis/provisioning/v1alpha5/requirements.go
ellistarn: Let's find a solution that doesn't stutter in any case.

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: Just checked, the API error does put the error category under spec.validateRequirements. I think we are good to remove the redundancy here.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: done

felix-zhe-huang: done

pkg/controllers/provisioning/controller.go
```diff
@@ -102,8 +102,8 @@ func (c *Controller) Apply(ctx context.Context, provisioner *v1alpha5.Provisione
 	provisioner.Spec.Requirements = provisioner.Spec.Requirements.
 		Add(requirements(instanceTypes)...).
 		Add(v1alpha5.NewLabelRequirements(provisioner.Spec.Labels).Requirements...)
-	if err := provisioner.Spec.Requirements.Validate(); err != nil {
-		return fmt.Errorf("validating requirements, %w", err)
+	if errs := provisioner.Spec.Requirements.Validate(); errs != nil {
```
bwagner5, line 105: is this only instance types it is validating?

felix-zhe-huang: This place we are validating if the instance types have any legit offers based on the provisioner's requirements.
Provisioners are validated at the webhook.

felix-zhe-huang: one example that this will caught is that if provisioner has a typo in the requirement value (e.g., ondemand instead of on-demand), it will raise an error here.

bwagner5: wouldn't it be clearer to keep it `validating requirements` then?  Since InstanceTypes is something that can be specified in requirements, if the capacity-type is wrong, saying that instance types failed validation seems confusing.

felix-zhe-huang: In this error, technically the requirements from both provisioner and instance types are legit. So there is really nothing wrong with those. The problem is that we can't validate what the values of the requirements are at the time when the API is called. So we have to wait until runtime to see if those values actually match up to the values used by the instance types.

If we just say validating requirements, it will say something like
`validating requirements, no feasible value for karpenter.sh/capacity-type`. This is very confusing because the customers have defined a valid value in their provisioner spec. They will have hard time understanding why it suddenly become not feasible.

felix-zhe-huang: This is indeed a tricky one. There are many possible reasons that may cause the instance type values to be different from the provisioner requirements values. We can't try to guess the reason for the customers. So i think the best we can do is to just let them know that their provisioner is incompatible to the instance types due to the encapsulated error.

bwagner5 (approved):
lgtm

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -136,19 +137,21 @@ func (c *Constraints) validateTaints() (errs *apis.FieldError) {
...
 		if !SupportedProvisionerOps.Has(string(requirement.Operator)) {
-			errs = errs.Also(apis.ErrInvalidKeyName(fmt.Sprintf("%s not in %v", requirement.Operator, SupportedProvisionerOps.UnsortedList()), "key"))
+			errs = multierr.Append(errs, fmt.Errorf("key %s has an unsupported operator %s, provisioner only supports %s", requirement.Key, requirement.Operator, SupportedProvisionerOps.UnsortedList()))
 		}
 	}
-	if err := c.Requirements.Validate(); err != nil {
-		errs = errs.Also(err)
+	errs = multierr.Append(errs, c.Requirements.Validate())
+	if errs != nil {
+		errors = errors.Also(apis.ErrInvalidValue(errs, "validateRequirements"))
```
ellistarn, line 154: `validateRequirements` is not a field, so I expect this will read strangely. See above about using `error` instead of `FieldError`

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -136,19 +137,21 @@ func (c *Constraints) validateTaints() (errs *apis.FieldError) {
...
 // When this function is called, the provisioner's requirments do not include the requirements from labels.
 // Provisioner requirements only support well known labels.
-func (c *Constraints) validateRequirements() (errs *apis.FieldError) {
+func (c *Constraints) validateRequirements() (errors *apis.FieldError) {
+	var errs error
 	for _, requirement := range c.Requirements.Requirements {
 		// Ensure requirements are well known
 		if !WellKnownLabels.Has(requirement.Key) {
-			errs = errs.Also(apis.ErrInvalidKeyName(fmt.Sprintf("%s not in %v", requirement.Key, WellKnownLabels.UnsortedList()), "key"))
+			errs = multierr.Append(errs, fmt.Errorf("key %s is not in wellknown labels %s", requirement.Key, WellKnownLabels.UnsortedList()))
```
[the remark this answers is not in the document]
felix-zhe-huang: done

pkg/controllers/provisioning/controller.go
```diff
@@ -102,8 +102,8 @@ func (c *Controller) Apply(ctx context.Context, provisioner *v1alpha5.Provisione
...
+		return fmt.Errorf("incompatible with the instance types, %w", errs)
```
ellistarn, line 106: `requirements are not compatible with cloud provider, %w, err`

PR #1293, opened 2022-02-08
Handle unhandled error when unmarshaling JSON

spring1843:
**1. Issue, if available:**
Unhandled error here can cause unexpected outcomes

Adds error handling when JSON is being unmarshaled

- [x] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/requirements.go
+++ b/pkg/apis/provisioning/v1alpha5/requirements.go
@@ -298,7 +298,9 @@ func (r *Requirements) MarshalJSON() ([]byte, error) {
 
 func (r *Requirements) UnmarshalJSON(b []byte) error {
 	var requirements []v1.NodeSelectorRequirement
-	json.Unmarshal(b, &requirements)
+	if err := json.Unmarshal(b, &requirements); err != nil {
+		return err
+	}
 	*r = NewRequirements(requirements...)
 	return nil
 }


ellistarn (approved):
Wow nice catch!!

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -298,7 +298,9 @@ func (r *Requirements) MarshalJSON() ([]byte, error) {
 
 func (r *Requirements) UnmarshalJSON(b []byte) error {
 	var requirements []v1.NodeSelectorRequirement
-	json.Unmarshal(b, &requirements)
+	if err := json.Unmarshal(b, &requirements); err != nil {
+		return err
```
njtran, line 302: nit: could you change the erroring to fit how we do it elsewhere in the repo?
```suggestion
		return fmt.Errorf("unmarshalling json, %w", err)
```

ellistarn: I'd actually argue against this, since it's a wrapper function. We want our errors to model the business logic flow, but not necessarily the exact implementation. We have two levels of unmarshalling json, so we probably want to return the naked error to avoid repeating ourselves.

spring1843: I see both styles in the repo I agree with @ellistarn on this.

PR #1294, opened 2022-02-08
Check DoNotEvict after filtering evictable pods to ensure termination can complete.

ellistarn:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/1166

Previously, the`do-not-evict` annotation could block termination logic, even if the node was unreachable. This change modifies the `do-not-evict` logic to only apply to evictable pods. Currently, evictable pods exclude

- Pods owned by the node (a.k.a. static pods)
- Pods with deletion time stamp set past their graceful termination period (a.k.a. stuck termination)
- Pods that tolerate the nodes taints (i.e. would reschedule immediately) 

This means that the `do-not-evict` annotation will no longer work for pods in the above categories. Given that we already did not attempt to evict these pods, the semantic holds. However, previously, pods in these categories with the `do-not-evict` annotation could block node termination.  

Reproduced 
1. Added do not evict annotation to pod
2. Manually terminated a node in EC2
3. Observe Node and Pod stuck Terminating

Fixed
1. Start with step 3 of above scenario
2. Apply PR changes to local environment
3. Node is gracefully terminated

- [ ] No

### DIFF
--- a/pkg/controllers/termination/eviction.go
+++ b/pkg/controllers/termination/eviction.go
@@ -54,7 +54,7 @@ func NewEvictionQueue(ctx context.Context, coreV1Client corev1.CoreV1Interface) 
 }
 
 // Add adds pods to the EvictionQueue
-func (e *EvictionQueue) Add(ctx context.Context, pods []*v1.Pod) {
+func (e *EvictionQueue) Add(pods []*v1.Pod) {
 	for _, pod := range pods {
 		if nn := client.ObjectKeyFromObject(pod); !e.Set.Contains(nn) {
 			e.Set.Add(nn)
--- a/pkg/controllers/termination/eviction.go
+++ b/pkg/controllers/termination/eviction.go
@@ -49,12 +49,12 @@ func NewEvictionQueue(ctx context.Context, coreV1Client corev1.CoreV1Interface) 
 
 		coreV1Client: coreV1Client,
 	}
-	go queue.Start(ctx)
+	go queue.Start(logging.WithLogger(ctx, logging.FromContext(ctx).Named("eviction")))
 	return queue
 }
 
 // Add adds pods to the EvictionQueue
-func (e *EvictionQueue) Add(pods []*v1.Pod) {
+func (e *EvictionQueue) Add(ctx context.Context, pods []*v1.Pod) {
 	for _, pod := range pods {
 		if nn := client.ObjectKeyFromObject(pod); !e.Set.Contains(nn) {
 			e.Set.Add(nn)
@@ -92,11 +92,11 @@ func (e *EvictionQueue) evict(ctx context.Context, nn types.NamespacedName) bool
 		ObjectMeta: metav1.ObjectMeta{Name: nn.Name, Namespace: nn.Namespace},
 	})
 	if errors.IsInternalError(err) { // 500
-		logging.FromContext(ctx).Debugf("Failed to evict pod %s due to PDB misconfiguration error.", nn.String())
+		logging.FromContext(ctx).Errorf("Could not evict pod %s due to PDB misconfiguration error.", nn.String())
 		return false
 	}
 	if errors.IsTooManyRequests(err) { // 429
-		logging.FromContext(ctx).Debugf("Failed to evict pod %s due to PDB violation.", nn.String())
+		logging.FromContext(ctx).Debugf("Did not to evict pod %s due to PDB violation.", nn.String())
 		return false
 	}
 	if errors.IsNotFound(err) { // 404
--- a/pkg/controllers/termination/suite_test.go
+++ b/pkg/controllers/termination/suite_test.go
@@ -162,25 +162,30 @@ var _ = Describe("Termination", func() {
 			ExpectNotFound(ctx, env.Client, node)
 		})
 		It("should delete nodes that have do-not-evict on pods for which it does not apply", func() {
-			doNotEvict := test.Pod(test.PodOptions{
-				NodeName:   node.Name,
-				ObjectMeta: metav1.ObjectMeta{Annotations: map[string]string{v1alpha5.DoNotEvictPodAnnotationKey: "true"}},
-			})
-			tolerate := test.Pod(test.PodOptions{
-				NodeName:   node.Name,
-				Tolerations: []v1.Toleration{{Operator: v1.TolerationOpExists}},
-			})
-			static := test.Pod(test.PodOptions{
-				NodeName:   node.Name,
-				ObjectMeta: metav1.ObjectMeta{OwnerReferences: []metav1.OwnerReference{{ Kind: "Node", APIVersion: "v1", Name: node.Name}}},
-			})
-			ExpectCreated(ctx, env.Client, node, doNotEvict, tolerate)
+			ExpectCreated(ctx, env.Client, node)
+			node = ExpectNodeExists(ctx, env.Client, node.Name)
+			pods := []*v1.Pod{
+				test.Pod(test.PodOptions{
+					NodeName:   node.Name,
+					ObjectMeta: metav1.ObjectMeta{Annotations: map[string]string{v1alpha5.DoNotEvictPodAnnotationKey: "true"}},
+				}),
+				test.Pod(test.PodOptions{
+					NodeName:    node.Name,
+					Tolerations: []v1.Toleration{{Operator: v1.TolerationOpExists}},
+				}),
+				test.Pod(test.PodOptions{
+					NodeName:   node.Name,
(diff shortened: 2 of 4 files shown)


pkg/controllers/termination/eviction.go
```diff
@@ -49,12 +49,12 @@ func NewEvictionQueue(ctx context.Context, coreV1Client corev1.CoreV1Interface)
 
 		coreV1Client: coreV1Client,
 	}
-	go queue.Start(ctx)
+	go queue.Start(logging.WithLogger(ctx, logging.FromContext(ctx).Named("eviction")))
```
njtran, line 52: Can we name this `termination.eviction` or `termination/eviction`? In log lines we see `controller.termination` for other termination related behaviors, but since this lies only touches termination behavior, I think it'd make sense to notate it as a subcontroller of termination.

pkg/controllers/termination/terminate.go
```diff
@@ -102,34 +95,30 @@ func (t *Terminator) terminate(ctx context.Context, node *v1.Node) error {
...
 		return nil, fmt.Errorf("listing pods on node, %w", err)
 	}
-	return ptr.PodListToSlice(pods), nil
-}
-
-func (t *Terminator) getEvictablePods(pods []*v1.Pod) []*v1.Pod {
-	evictable := []*v1.Pod{}
-	for _, p := range pods {
+	pods := []*v1.Pod{}
+	for i := range podList.Items {
```
njtran, line 103: Could you do the following? It's not a problem here, but [has been in the past](https://medium.com/swlh/golang-tips-why-pointers-to-slices-are-useful-and-how-ignoring-them-can-lead-to-tricky-bugs-cac90f72e77b) so I'd like to reduce the chance of that happening. You can change the later references to be just `i` in this case.
```suggestion
	for _, i := range ptr.PodListToSlice(podList) {
```

website/content/en/preview/tasks/deprovisioning.md
```diff
@@ -78,11 +78,9 @@ Review what [disruptions are](https://kubernetes.io/docs/concepts/workloads/pods
...
-In either case, the node will be cordoned to prevent additional work from scheduling.
+If a pod exists with the annotation `karpenter.sh/do-not-evict` on a node, and a request is made to delete the node, Karpenter will not drain any pods from that node or otherwise try to delete the node. This annotation does not apply to static pods, pods that tolerate `NoSchedule`, or pods terminating past their graceful termination period.
 
-That annotation is used for pods that you want to run on one node from start to finish without interruption.
+This is useful for pods that you want to run from start to finish without interruption.
 Examples might include a real-time, interactive game that you don't want to interrupt or a long batch job (such as you might have with machine learning) that would need to start over if it were interrupted.
 
-If you want to terminate a `do-not-evict` pod, you can simply remove the annotation and the finalizer will delete the pod and continue the node deprovisioning process.
+If you want to terminate node with a `do-not-evict` pod, you can simply remove the annotation and the deprovisioning process will continue.
```
njtran, line 86: s/node with/a node with or s/node with/nodes with

pkg/controllers/termination/terminate.go
```diff
@@ -102,34 +95,30 @@ func (t *Terminator) terminate(ctx context.Context, node *v1.Node) error {
...
 		}
-		evictable = append(evictable, p)
+		pods = append(pods, &podList.Items[i])
 	}
-	return evictable
+	return pods, nil
 }
 
-func (t *Terminator) evict(pods []*v1.Pod) {
+func (t *Terminator) evict(ctx context.Context, pods []*v1.Pod) {
```
njtran, line 121: I understand why it's good to have a context here, but if we're not actually using the context that's passed in here and used in `EvictionQueue.Add()`, I don't think there's any difference in code here.

pkg/controllers/termination/eviction.go
ellistarn: I played with this and it changes the indentation, which makes the logs harder to read. I think it's probably 6-of-1-half-a-dozen-of-the-other. We can always change this too, so keeping as is if users get confused.

pkg/controllers/termination/terminate.go
ellistarn: You have it backwards. `i` is the index, or `for i, pod := range ...`. I'm avoiding the bug by using the index.

website/content/en/preview/tasks/deprovisioning.md
ellistarn: Oh nice catch!

pkg/controllers/termination/terminate.go
ellistarn: Total oversight on my part. I had this wired up for a log line I eventually removed.

pkg/controllers/termination/terminate.go
njtran: Resolved offline. Using `_, i` where i represents the pod.

pkg/controllers/termination/terminate.go
```diff
@@ -57,28 +57,22 @@ func (t *Terminator) cordon(ctx context.Context, node *v1.Node) error {
...
 	pods, err := t.getPods(ctx, node)
 	if err != nil {
 		return false, fmt.Errorf("listing pods for node, %w", err)
 	}
-
-	// 2. Separate pods as non-critical and critical
-	// https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
+	// Skip node due to do-not-evict
 	for _, pod := range pods {
+		// https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
```
njtran, line 67: I feel like this link is in a weird spot now without the mention of non-critical + critical pods, even though it was in a weird spot in the first place. Fits better above the drain function imo.

pkg/controllers/termination/terminate.go
```diff
@@ -102,31 +96,27 @@ func (t *Terminator) terminate(ctx context.Context, node *v1.Node) error {
 
 // getPods returns a list of pods scheduled to a node based on some filters
```
njtran, line 97: Can you change the comment to be more accurate since it's merged the `getEvictablePods()` call into here? 

```suggestion
// getPods returns a list of evictable pods scheduled to a node based on some filters
```

website/content/en/preview/tasks/deprovisioning.md
```diff
@@ -78,11 +78,9 @@ Review what [disruptions are](https://kubernetes.io/docs/concepts/workloads/pods
 
 ### Pod set to do-not-evict
 
-If a pod exists with the annotation `karpenter.sh/do-not-evict` on a node, and a request is made to delete the node, Karpenter will not drain any pods from that node or otherwise try to delete the node.
-However, if a `do-not-evict` pod is added to a node while the node is draining, the remaining pods will still evict, but that pod will block termination until it is removed.
-In either case, the node will be cordoned to prevent additional work from scheduling.
+If a pod exists with the annotation `karpenter.sh/do-not-evict` on a node, and a request is made to delete the node, Karpenter will not drain any pods from that node or otherwise try to delete the node. This annotation does not apply to static pods, pods that tolerate `NoSchedule`, or pods terminating past their graceful termination period.
```
njtran, line 81: nit in wording: 

```suggestion
If a pod exists with the annotation `karpenter.sh/do-not-evict` on a node, and a request is made to delete the node, Karpenter will not drain any pods from that node or otherwise try to delete the node. This annotation will have no effect for static pods, pods that tolerate `NoSchedule`, or pods terminating past their graceful termination period.
```

pkg/controllers/termination/suite_test.go
```diff
@@ -160,6 +161,23 @@ var _ = Describe("Termination", func() {
 			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(node))
 			ExpectNotFound(ctx, env.Client, node)
 		})
+		It("should delete nodes that have a do-not-evict pod that's ignored", func() {
+			pod := test.Pod(test.PodOptions{
+				NodeName:   node.Name,
+				ObjectMeta: metav1.ObjectMeta{Annotations: map[string]string{v1alpha5.DoNotEvictPodAnnotationKey: "true"}},
+			})
```
njtran, line 168: Consider testing static pods and a pod that tolerates `NoSchedule`, should just be two more pods creation lines, and adding to the respective lines below.

pkg/controllers/termination/terminate.go
ellistarn: Good point, moving above.

njtran (approved):
Thanks for the fix! Looks good!

njtran (approved):

Issue #1306, opened 2022-02-09
Karpenter cannot fit workload on instance type where it should fit


nonoswz:
## Version

**Karpenter**: v0.6.1

**Kubernetes**: v1.20+

## Expected Behavior

I expect Karpenter to be able to schedule a deployment on an instance type where the workload (resources) fits

## Actual Behavior

I am trying to switch from ASG managed nodes to Karpenter. Currently it fails to fit one of our deployment (prometheus) on the same instance type as it was before in one of the ASG node (r5.12xlarge).

Our Prometheus deployment requests around 350GiB memory and 40 CPU, and a r5.12xlarge has 48 vCPU and 384 GiB as per [AWS docs](https://aws.amazon.com/ec2/instance-types/r5/). 
Extract of prometheus pod spec
````
    resources:
      limits:
        memory: 350Gi
      requests:
        cpu: "40"
        memory: 350Gi
````

Karpenter fails to run it on this specific instance type saying it won't fit.
````
2022-02-09T20:11:34.749Z	ERROR	controller.provisioning	Failed to compute packing, pod(s) [monitoring/prometheus-infrastructure-0] did not fit in instance type option(s) [r5.12xlarge]	{"commit": "df57892", "provisioner": "prometheus"}
````

Notes: 
- we also have multiple damonsets running on each nodes (they are shown below in the node description), in case that comes in the calculation to see if the workload can fit.
- I tried to use a bigger instance type (r5.16xlarge) and it works fine. We can use bigger nodes as a workaround, but ultimately it would be good to be able to get the smaller instance type working.

## Steps to Reproduce the Problem

1. Create a deployment requesting 40vCPU and 350Gi memory
2. Create a provisioners with only one instance type : r5.12xlarge
3. Scale up replicas of this deployment to 1 and see if Karpenter is able to fit the deployment into a r5.12xlarge instance type.

## Resource Specs and Logs

Pod spec (prometheus, I included relevant part only)
````
spec:
  containers:
   # container 1
    ....
    resources:
      limits:
        memory: 350Gi
      requests:
        cpu: "40"
        memory: 350Gi
   # container 2
    .....
    resources:
      limits:
        cpu: 100m
        memory: 25Mi
      requests:
        cpu: 100m
        memory: 25Mi
   # container 3
   ......
    resources:
      limits:
        cpu: 100m
        memory: 25Mi
      requests:
        cpu: 100m
        memory: 25Mi
  nodeSelector:
    group: prometheus
[...]

PR #1317, opened 2022-02-10
Fix overhead calculation to correctly apply kube-reserved cpu

bwagner5:
**1. Issue, if available:**
N/A

 - The kube-reserved overhead calculation was not being applied correctly on the incremental core percentages.

 - Printed the overhead calc when investigating the #1306 (this change does not fix that issue) and found that the overhead didn't include the incremental core calc. It only used the initial 100m cpu overhead. 
 - After this change, I printed the overhead for r5.12xl and it correctly showed `cpu 290m` of overhead

- [x] No

### DIFF
--- a/pkg/cloudprovider/aws/instancetype.go
+++ b/pkg/cloudprovider/aws/instancetype.go
@@ -16,6 +16,7 @@ package aws
 
 import (
 	"fmt"
+
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/vpc"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/aws/instancetype.go
```diff
@@ -155,7 +156,9 @@ func (i *InstanceType) Overhead() v1.ResourceList {
 			if cpu < cpuRange.end {
 				r = float64(cpu - cpuRange.start)
 			}
-			overhead.Cpu().Add(*resource.NewMilliQuantity(int64(r*cpuRange.percentage), resource.DecimalSI))
+			cpuOverhead := overhead[v1.ResourceCPU]
+			cpuOverhead.Add(*resource.NewMilliQuantity(int64(r*cpuRange.percentage), resource.DecimalSI))
+			overhead[v1.ResourceCPU] = cpuOverhead
```
ellistarn, line 161: This resource library has bitten us before.

Possible to one-line? 
```
overhead[v1.ResourceCPU] = overhead.CPU().Add(*resource.NewMilliQuantity(int64(r*cpuRange.percentage), resource.DecimalSI))
```

bwagner5: we can one-line it using our `resources.Merge` but then it's longer... so idk but the example you put wouldn't work because Add() doesn't return a Quantity, it just updates the pointer :(

ellistarn (approved):

PR #1323, opened 2022-02-11
support ubuntu amiFamily

bwagner5:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/923

 - Changed EKSOptimized to AL2 because EKS Optimized is a class of AMIs that can be different AMI Families which are optimized for EKS. 
 - Lowercased values expected in the provisioner
 - Add support for `ubuntu` as an `amiFamily`
 - Adds docs for `amiFamily`

 - Tested on an EKS cluster with 2 deployments `inflate-amd64` and `inflate-arm64`. 
   - I varied the provisioner `amiFamily` between `al2`, `bottlerocket`, and `ubuntu` scaling each to verify the node becomes ready and the pods go to running.

- [ ] No

### DIFF
--- a/pkg/cloudprovider/aws/apis/v1alpha1/provider.go
+++ b/pkg/cloudprovider/aws/apis/v1alpha1/provider.go
@@ -48,7 +48,7 @@ type AWS struct {
 	AMIFamily *string `json:"amiFamily,omitempty"`
 	// InstanceProfile is the AWS identity that instances use.
 	// +optional
-	InstanceProfile *string `json:"instanceProfile"`
+	InstanceProfile *string `json:"instanceProfile,omitempty"`
 	// LaunchTemplate for the node. If not specified, a launch template will be generated.
 	// +optional
 	LaunchTemplate *string `json:"launchTemplate,omitempty"`
--- a/pkg/cloudprovider/aws/apis/v1alpha1/provider.go
+++ b/pkg/cloudprovider/aws/apis/v1alpha1/provider.go
@@ -47,8 +47,8 @@ type AWS struct {
 	// +optional
 	AMIFamily *string `json:"amiFamily,omitempty"`
 	// InstanceProfile is the AWS identity that instances use.
-	// +required
-	InstanceProfile string `json:"instanceProfile"`
+	// +optional
+	InstanceProfile *string `json:"instanceProfile"`
 	// LaunchTemplate for the node. If not specified, a launch template will be generated.
 	// +optional
 	LaunchTemplate *string `json:"launchTemplate,omitempty"`
--- a/pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
+++ b/pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
@@ -60,5 +60,5 @@ func (c *Constraints) defaultAMIFamily() {
 	if c.AMIFamily != nil {
 		return
 	}
-	c.AMIFamily = &AMIFamilyEKSOptimized
+	c.AMIFamily = &AMIFamilyAL2
 }
--- a/pkg/cloudprovider/aws/apis/v1alpha1/provider_validation.go
+++ b/pkg/cloudprovider/aws/apis/v1alpha1/provider_validation.go
@@ -22,6 +22,15 @@ import (
 	"knative.dev/pkg/apis"
 )
 
+const (
+	launchTemplatePath          = "launchTemplate"
+	securityGroupSelectorPath   = "securityGroupSelector"
+	fieldPathSubnetSelectorPath = "subnetSelector"
+	amiFamilyPath               = "amiFamily"
+	metadataOptionsPath         = "metadataOptions"
+	instanceProfilePath         = "instanceProfile"
+)
+
 func (a *AWS) Validate() (errs *apis.FieldError) {
 	return a.validate().ViaField("provider")
 }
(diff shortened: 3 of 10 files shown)


website/content/en/preview/AWS/provisioning.md
```diff
@@ -150,6 +150,21 @@ spec:
       httpTokens: required
 ```
 
+### Amazon Machine Image (AMI) Family
+
+The AMI used when provisioning nodes can be controlled by the `amiFamily` field. Based on the value set for `amiFamily`, Karpenter will automatically query for the appropriate [EKS optimized AMI](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-amis.html) via AWS Systems Manager (SSM). 
+
+Currently, Karpenter supports `amiFamily` values `al2`, `bottlerocket`, and `ubuntu`. GPUs are only supported with `al2`.
+
+Note: If a custom launch template is specified, then the AMI value in the launch template is used rather than the `amiFamily` value.
```
ellistarn, line 159: Should we include this in validation?

website/content/en/preview/AWS/provisioning.md
```diff
@@ -150,6 +150,21 @@ spec:
       httpTokens: required
 ```
 
+### Amazon Machine Image (AMI) Family
+
+The AMI used when provisioning nodes can be controlled by the `amiFamily` field. Based on the value set for `amiFamily`, Karpenter will automatically query for the appropriate [EKS optimized AMI](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-amis.html) via AWS Systems Manager (SSM).
```
akestner, line 155: We should be clear about the AMI that we'll be resolving here. For AL2 it's the EKS Optimized AMI, but this reads as though there's Bottlerocket and Ubuntu variants of the optimized AMI as well, which isn't the case.

website/content/en/preview/AWS/provisioning.md
bwagner5: That's a good idea. We should include the other fields like this as well (securityGroupSelector, InstanceProfileName, MetadataOptipns, etc)

akestner, line 159: I expect some confusion over this, just like there is for other LT parameters that take precedence over the Provisioner-specified values. Can we add a warning or something making this more explicit?

akestner: Ditto, see my comment below along the same lines

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
```diff
@@ -47,8 +47,8 @@ type AWS struct {
 	// +optional
 	AMIFamily *string `json:"amiFamily,omitempty"`
 	// InstanceProfile is the AWS identity that instances use.
-	// +required
-	InstanceProfile string `json:"instanceProfile"`
+	// +optional
+	InstanceProfile *string `json:"instanceProfile"`
```
ellistarn, line 51:
```suggestion
	InstanceProfile *string `json:"instanceProfile,omitempty"`
```

pkg/cloudprovider/aws/apis/v1alpha1/provider_validation.go
```diff
@@ -22,6 +22,15 @@ import (
 	"knative.dev/pkg/apis"
 )
 
+const (
+	launchTemplatePath          = "launchTemplate"
+	securityGroupSelectorPath   = "securityGroupSelector"
+	fieldPathSubnetSelectorPath = "subnetSelector"
+	amiFamilyPath               = "amiFamily"
+	metadataOptionsPath         = "metadataOptions"
+	instanceProfilePath         = "instanceProfile"
```
ellistarn, line 31: I'm not a fan of this pattern. If any of these values change, you'd need to change the variable name as well. The only benefit I can see compared to using the strings inline is compile time checks. I can see this happening due to an unsuccessful find+replace, but this feels like a stretch.

pkg/cloudprovider/aws/apis/v1alpha1/provider_validation.go
```diff
@@ -38,29 +47,43 @@ func (a *AWS) validate() (errs *apis.FieldError) {
...
 	for key, value := range a.SubnetSelector {
 		if key == "" || value == "" {
-			errs = errs.Also(apis.ErrInvalidValue("\"\"", fmt.Sprintf("subnetSelector['%s']", key)))
+			errs = errs.Also(apis.ErrInvalidValue("\"\"", fmt.Sprintf("%s['%s']", fieldPathSubnetSelectorPath, key)))
 		}
 	}
 	return errs
 }
 
 func (a *AWS) validateSecurityGroups() (errs *apis.FieldError) {
```
ellistarn, line 80: Consider shortcircuitting if launchtemplate is not nil instead.
```
if l.LaunchTemplate != nil {
  return nil
}
```

pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
```diff
@@ -60,5 +60,5 @@ func (c *Constraints) defaultAMIFamily() {
 	if c.AMIFamily != nil {
 		return
 	}
-	c.AMIFamily = &AMIFamilyEKSOptimized
+	c.AMIFamily = &AMIFamilyAL2
```
olemarkus, line 63: These AMIs aren't the plain AL2 images, but EKS specific AL2 images. Perhaps the naming should reflect that?

pkg/cloudprovider/aws/apis/v1alpha1/register.go
```diff
@@ -32,11 +32,13 @@ var (
 	AWSRestrictedLabelDomains = []string{
 		"k8s.aws",
 	}
-	AMIFamilyBottlerocket = "Bottlerocket"
-	AMIFamilyEKSOptimized = "EKSOptimized"
+	AMIFamilyBottlerocket = "bottlerocket"
```
ellistarn, line 35: In Kubernetes, YAML values should be capital letters.
Bottlerocket, AL2, Ubunutu

pkg/cloudprovider/aws/apis/v1alpha1/provider_defaults.go
ellistarn: The way I see it, we only want to support one AL2 family. The implementation happens to be EKSOptimized at the moment.

pkg/cloudprovider/aws/apis/v1alpha1/provider_validation.go
bwagner5: I was more worried about typos in the hardcoded strings since they were hardcoded in multiple spots.

ellistarn: I can live with it, as long as we don't move towards turning every string in the codebase into a constant.

pkg/cloudprovider/aws/apis/v1alpha1/register.go
bwagner5: Should capacityTypes be uppercase too then?   `[Spot, On-Demand]`?

ellistarn: Yes. `OnDemand`, I think.

bwagner5: that would be a breaking change right now... so maybe we should just leave it for capacityType? It seems to not be a hard rule values of labels since architecture and operating system label values are lowercase...

ellistarn: Hmm this is a good point -- perhaps the difference is whether or not it's a finite enum? Maybe worth asking in sig-api-machinery, or similar.

ellistarn (approved):
💯

PR #1324, opened 2022-02-11
Simplify conversion of errors to string in formatted function calls

spring1843:
**1. Issue, if available:**

When using functions that format strings (they  ends with an f suffix e.g. `fmt.Errorf`, or `fmt.Sprintf`)  there is no need to  use `err.Error()` in order to get the error message in string. It is sufficient to use `err` because errors in Go implement the `fmt.Stringer` interface allowing an automatic conversion to string when needed.

- [x] No


ellistarn:
Does this print out the exact same message or are there slight differences ? I remember adopting this pattern for some reason, but perhaps it was because I was using "%v" instead of "%s".

ellistarn (approved):

spring1843:
> 

It is exactly the same, here's an example:
https://go.dev/play/p/yuEado-6TvX

Using `%v` would make it different i.e. it would show the type of the variable (error in this case).

PR #1328, opened 2022-02-11
Lower the minimum supported k8s version

felix-zhe-huang:
**1. Issue, if available:**
Lower the minimum supported k8s version

Add an environment variable to webhook container to define the minimum supported k8s version.

- [X] No


charts/karpenter/values.yaml
```diff
@@ -134,6 +134,9 @@ logLevel: info
 # -- Cluster name.
 clusterName: ""
 
+# -- Minimum kubernetes version that will be supported
+k8sMinVersion: "1.19.0-0"
```
ellistarn, line 138: Is the -0 suffix required? 
```suggestion
kubernetesMinVersion: "1.19.0-0"
```

felix-zhe-huang: I am not sure. but that's the format used by knative.

felix-zhe-huang:
This patch is tested with a EKS cluster with k8s v1.19.

charts/karpenter/values.yaml
```diff
@@ -134,6 +134,9 @@ logLevel: info
 # -- Cluster name.
 clusterName: ""
 
+# -- Minimum kubernetes version that will be supported
+kubernetesMinVersion: "1.19.0-0"
```
bwagner5, line 138: should we include this in values.yaml?  I don't think anyone is going to be touching this, ideally we should know what min version is supported with Karpenter and hardcode into deployment.yaml.  values.yaml should be things that a user would need to change.  Maybe the thought is that they test with a lower version of k8s than we have and could configure it here?

felix-zhe-huang: good catch. Done.

ellistarn (approved):
Nice!

Issue #638, opened 2021-08-25
Reenable PDB Eviction Test


ellistarn:
This test was removed in https://github.com/awslabs/karpenter/pull/637

PR #1334, opened 2022-02-14
re-enable the PDB eviction test

tzneal:
**1. Issue, if available:**
#638

- the pod must be in a non-pending state for eviction  to consider the PDB
- the Status.ObservedGeneration must be >= the pdb.Generation  or the eviction results in a 429 retry with a 10 second delay

Unit testing

- [X] No

### DIFF
--- a/pkg/controllers/termination/suite_test.go
+++ b/pkg/controllers/termination/suite_test.go
@@ -204,9 +204,11 @@ var _ = Describe("Termination", func() {
 			podNoEvict := test.Pod(test.PodOptions{
 				NodeName:   node.Name,
 				ObjectMeta: metav1.ObjectMeta{Labels: labelSelector},
+				Phase:      v1.PodRunning,
 			})
 
-			ExpectCreated(ctx, env.Client, node, podNoEvict, pdb)
+			ExpectCreated(ctx, env.Client, node)
+			ExpectCreatedWithStatus(ctx, env.Client, podNoEvict, pdb)
 
 			// Trigger Termination Controller
 			Expect(env.Client.Delete(ctx, node)).To(Succeed())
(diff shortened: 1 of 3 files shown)


pkg/controllers/termination/suite_test.go
```diff
@@ -204,9 +204,17 @@ var _ = Describe("Termination", func() {
...
 			})
 
-			ExpectCreated(ctx, env.Client, node, podNoEvict, pdb)
+			ExpectCreated(ctx, env.Client, node, pdb)
+			ExpectCreatedWithStatus(ctx, env.Client, podNoEvict)
+
+			// kube-controller-manager normally does this, but we don't have one. If this isn't modified
+			// the eviction controller assumes that the PDB hasn't been processed by the disruption controller
+			// yet and adds a 10 second retry to our evict() call
+			pdb.Status.ObservedGeneration = 1
```
ellistarn, line 216: Can we include this in the `test.PodDisruptionBudget` library? We will then need to use ExpectCreatedWithStatus on it.

tzneal: Yup, done, the update call required the `ResourceVersion` to match, I thought that was with everything so I'm not sure why it only showed up for the PDB.

pkg/test/pods.go
```diff
@@ -126,6 +128,13 @@ func PodDisruptionBudget(overrides ...PDBOptions) *v1beta1.PodDisruptionBudget {
 			},
 			MaxUnavailable: options.MaxUnavailable,
 		},
+		Status: v1beta1.PodDisruptionBudgetStatus{
+			// To be considererd for application by eviction, the Status.ObservedGeneration must be >= the PDB generation.
```
ellistarn, line 132:
```suggestion
			// To be considered for application by eviction, the Status.ObservedGeneration must be >= the PDB generation.
```

ellistarn (dismissed):

pkg/test/pods.go
tzneal: Good catch, done.

ellistarn (approved):

Issue #706, opened 2021-09-24
TestFlake: should not delete nodes that have a do-not-evict pod


ellistarn:
https://pipelines.actions.githubusercontent.com/baKYZvxShjtYEIWbzcki60CC3Adx3gE55cEMzCllkpRTV8tu3Y/_apis/pipelines/1/runs/2026/signedlogcontent/3?urlExpires=2021-09-24T19%3A15%3A50.1092818Z&urlSigningMethod=HMACV1&urlSignature=lATu%2Fyb6DPSNuoeCGUFOcXa6UGEWV7Ff45OE1ht00Ek%3D

PR #1335, opened 2022-02-14
add a test/EvictionQueue to eliminate test timing issues

tzneal:
**1. Issue, if available:**

Maybe #706 ?

Depending on when the eviction queue was acted upon, items
could be queued and removed before the Expect() calls that
checked for them in the queue.  This change adds a test
eviction queue that maintains a set of all items ever
queued.

Unit tests only

- [X] No

### DIFF
--- a/pkg/controllers/termination/eviction.go
+++ b/pkg/controllers/termination/eviction.go
@@ -35,15 +35,19 @@ const (
 	evictionQueueMaxDelay  = 10 * time.Second
 )
 
-type EvictionQueue struct {
+type EvictionQueue interface {
+	Add(pods []*v1.Pod)
+	Contains(i ...interface{}) bool
+}
+type evictionQueue struct {
 	workqueue.RateLimitingInterface
 	set.Set
 
 	coreV1Client corev1.CoreV1Interface
 }
 
-func NewEvictionQueue(ctx context.Context, coreV1Client corev1.CoreV1Interface) *EvictionQueue {
-	queue := &EvictionQueue{
+func NewEvictionQueue(ctx context.Context, coreV1Client corev1.CoreV1Interface) EvictionQueue {
+	queue := &evictionQueue{
 		RateLimitingInterface: workqueue.NewRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(evictionQueueBaseDelay, evictionQueueMaxDelay)),
 		Set:                   set.NewSet(),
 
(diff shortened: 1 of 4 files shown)


tzneal:
I'm not sure if this is the same issue as #706, since the log is gone, but it's probably likely. If you add a `time.Sleep(10 * time.Millisecond)` to the beginning of `ExpectEnqueuedForEviction`, you should see six tests fail without this change, including "should not delete nodes that have a do-not-evict pod".

I ran into this while fixing the PDB test.

pkg/controllers/termination/eviction.go
```diff
@@ -35,15 +35,19 @@ const (
 	evictionQueueMaxDelay  = 10 * time.Second
 )
 
-type EvictionQueue struct {
+type EvictionQueue interface {
+	Add(pods []*v1.Pod)
+	Contains(i ...interface{}) bool
+}
```
tzneal, line 41: This interface looks weird, but it's what was used previously.

ellistarn:
I'm hesitant about this change. I'm concerned that were adding additional test complexity and reducing the surface of tested code. 

You might've noticed that Karpenter doesn't use traditional unit tests. Instead, we use "functional" tests, which cover the integration between multiple go packages and 3p libraries. When absolutely necessary (e.g. aws/kube apis), we mock behavior. There are multiple goals of this approach. Firstly, it covers a broader surface, increasing our confidence that not only does the business logic of each package work, but the integration between the packages is what we expect. Additionally, it enables us to perform fairly substantial refactors without modifying tests, increasing our confidence that our changes don't break anything.

What do you think about using `Eventually` to solve timing issues? We use this broadly in the test code. As long as the timeout is high enough that we are guaranteed to win the race in any reasonable scenario, we get the benefit of testing the same code that runs in production while also avoiding test flakes.

tzneal:
> 
Sorry, I'm not following 100%.  This change didn't add a mock, it just a wrapper around the actual EvicitonQueue to track the pods that were added:

```go
func NewEvictionQueue(ctx context.Context, coreV1Client corev1.CoreV1Interface) *EvictionQueue {
	return &EvictionQueue{
		hasContained: set.NewSet(),
		q:            termination.NewEvictionQueue(ctx, coreV1Client),
	}
}
```


The race condition I saw were pods entering and exiting the queue *before* we can check the queue, so Eventually() wouldn't help here either.

tzneal:
Will rework to modify existing test to be unaware of the queue.

PR #1339, opened 2022-02-14
Add codeql security analysis workflow to the repo

spring1843:
**1. Issue, if available:**
N/A

Adds Github's security scanner to the workflows 

https://github.blog/2020-09-30-code-scanning-is-now-available/

Should be able to see it running on this PR

- [ X] No

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,14 +1,3 @@
-# For most projects, this workflow file will not need changing; you simply need
-# to commit it to your repository.
-#
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
-#
-# ******** NOTE ********
-# We have attempted to detect the languages in your repository. Please check
-# the `language` matrix defined below to confirm you have the correct set of
-# supported CodeQL languages.
-#
 name: "CodeQL"
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,68 @@
...
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go' ]
+
+    steps:
+    - name: Checkout repository
```
ellistarn, line 38: Not sure how you feel about this style, but I find that it can be a bit more clear to drop the name rather than defining some bespoke name, e.g.g
```
- uses: actions/checkout@v2
- uses: nextstep-foo@v-bar
```

The exception for this is if the step is not apparent in the name of the `uses` directive (e.g. executing a fully fledged script)

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,68 @@
...
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main, gh-pages ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
```
ellistarn, line 20: Worth analyzing on PR instead of cron? How long does this take?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,68 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main, gh-pages ]
```
ellistarn, line 16: I'd drop gh-pages from this. That's just a deployment branch.

.github/workflows/codeql-analysis.yml
spring1843: It took between 15 and 17 minutes each time I tried. That is pretty long, I removed the PR check so that it doesn't block contributors for that long. Instead now it will run once a PR is merged.

I'm open to removing the Cron it was set by default. Running it on a schedule has the benefit that it will proactively scan all the dependencies regularly so theoretically if any of them have a new security (either recently introduced to their code base or a new vulnerability added to the scanner) problem we will know earlier.

ellistarn: Ah -- if it's that long, I'm totally fine with doing it periodically. On merged might be ideal, since our go mod closure shouldn't be changing outside of a merge.

ellistarn (approved):

PR #1348, opened 2022-02-15
[WIP]Enable provisioning bare metal instances

felix-zhe-huang:
**1. Issue, if available:**

Enable provisioning bare metal instances.

Add corresponding unit tests.
Manually tested in EKS envionment.

- [X] No

### DIFF
--- a/pkg/cloudprovider/aws/instancetype.go
+++ b/pkg/cloudprovider/aws/instancetype.go
@@ -170,3 +170,7 @@ func (i *InstanceType) Overhead() v1.ResourceList {
 func (i *InstanceType) eniLimitedPods() int64 {
 	return *i.NetworkInfo.MaximumNetworkInterfaces*(*i.NetworkInfo.Ipv4AddressesPerInterface-1) + 2
 }
+
+func (i *InstanceType) IsBareMetal() bool {
+	return aws.BoolValue(i.BareMetal)
+}
--- a/pkg/cloudprovider/aws/instancetype.go
+++ b/pkg/cloudprovider/aws/instancetype.go
@@ -56,7 +56,7 @@ func (i *InstanceType) Architecture() string {
 			return value
 		}
 	}
-	return fmt.Sprint(aws.StringValueSlice(i.ProcessorInfo.SupportedArchitectures)) // Unrecognized, but used for error printing
+	return aws.StringValueSlice(i.ProcessorInfo.SupportedArchitectures)[0] // Unrecognized, return the first architecture value for error printing
 }
 
 func (i *InstanceType) CPU() *resource.Quantity {
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -66,6 +66,7 @@ type InstanceType interface {
 	AWSNeurons() *resource.Quantity
 	AWSPodENI() *resource.Quantity
 	Overhead() v1.ResourceList
+	IsBareMetal() bool
 }
 
 // An Offering describes where an InstanceType is available to be used, with the expectation that its properties
--- a/pkg/controllers/provisioning/controller.go
+++ b/pkg/controllers/provisioning/controller.go
@@ -100,7 +100,7 @@ func (c *Controller) Apply(ctx context.Context, provisioner *v1alpha5.Provisione
 	}
 	provisioner.Spec.Labels = functional.UnionStringMaps(provisioner.Spec.Labels, map[string]string{v1alpha5.ProvisionerNameLabelKey: provisioner.Name})
 	provisioner.Spec.Requirements = provisioner.Spec.Requirements.
-		Add(requirements(instanceTypes)...).
+		Add(requirements(provisioner.Spec.Requirements, instanceTypes)...).
 		Add(v1alpha5.NewLabelRequirements(provisioner.Spec.Labels).Requirements...)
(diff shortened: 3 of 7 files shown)


pkg/cloudprovider/aws/instancetype.go
```diff
@@ -56,7 +56,7 @@ func (i *InstanceType) Architecture() string {
 			return value
 		}
 	}
-	return fmt.Sprint(aws.StringValueSlice(i.ProcessorInfo.SupportedArchitectures)) // Unrecognized, but used for error printing
+	return aws.StringValueSlice(i.ProcessorInfo.SupportedArchitectures)[0] // Unrecognized, return the first architecture value for error printing
```
suket22, line 59: Not sure I understand why this was done?

suket22 (commented):
1. Can you [update the documentation](https://github.com/aws/karpenter/blob/main/website/content/en/preview/provisioner.md?plain=1#L82). 
2. Can you add your testing to the PR overview

pkg/cloudprovider/aws/instancetype.go
felix-zhe-huang: Previously this function returns the string representation of an array that contains the architecture values if those values cannot be translate. It was meant for outputting error messages.

Instead of outputting an array, now the function returns the first architecture value in that case. This will be a valid AWS specify architecture value. It won't change anything because the provisioner architecture requirement does not overlap with this AWS specify one. It is automatically ignored.

bwagner5:
I'm not sure we should do this since `metal` instance types are resource equivalent to the largest instance type in the instance family. 

```
$ ec2-instance-selector --allow-list '^m5\.(24xlarge|metal)$' -o table-wide
Instance Type  VCPUs   Mem (GiB)  Hypervisor  Current Gen  Hibernation Support  CPU Arch  Network Performance  ENIs    GPUs    GPU Mem (GiB)  GPU Info  On-Demand Price/Hr  Spot Price/Hr (30d avg)
-------------  -----   ---------  ----------  -----------  -------------------  --------  -------------------  ----    ----    -------------  --------  ------------------  -----------------------
m5.24xlarge    96      384        nitro       true         false                x86_64    25 Gigabit           15      0       0                        $4.608              $0.97123
m5.metal       96      384        none        true         false                x86_64    25 Gigabit           15      0       0                        $4.608              $0.9578
``` 

So it's possible we bin-pack on `metal` and `24xlarge` instance types where ec2 fleet could pick the `metal` type since it's not differentiating on anything except price. Most users will not want to use `metal` instance types, so this could force them to add a list of `NotIn` for instance-type in the provisioner.  We need some sort of parameter to switch on metal instance types at the pod level or requirements.

suket22:
> Most users will not want to use `metal` instance types, so this could force them to add a list of `NotIn` for instance-type in the provisioner. 

Why is that? My understanding of metal instances was that you get guaranteed single tenancy but didn't think there were other functional differences with other `.24xlarge`s, that would make users not inclined to use them.

> And maybe more importantly, we need to remove our opinionated filters when a user specifies an In on instance-types. This would allow a customer to specify metal instance types in requirements and we would respect what they'd like to choose from, but keep our defaults for the normal case.

Strongly agree 💯 . I thought this was already the case tbh, but if not let's do this regardless.

felix-zhe-huang:
So we should rethink how preferences are given to the instance types in bin packing. I feel that the intelligence should be part of the bin packing logic. Adding opinionated filters should be just a work around solution.

felix-zhe-huang:
In this implementation, bare metal instance types are not enabled by default. Bare metal instances will be provision-able when provisioner spec includes 
```

    operator: In
    values:
    - INSTANCE_TYPE_NAME
```

pkg/controllers/provisioning/controller.go
```diff
@@ -150,19 +150,25 @@ func requirements(instanceTypes []cloudprovider.InstanceType) []v1.NodeSelectorR
 		v1alpha5.LabelCapacityType: sets.NewString(),
 	}
 	for _, instanceType := range instanceTypes {
-		for _, offering := range instanceType.Offerings() {
-			supported[v1.LabelTopologyZone].Insert(offering.Zone)
-			supported[v1alpha5.LabelCapacityType].Insert(offering.CapacityType)
+		// filter out bare metal instance type unless required by provisioner requirements
+		if instanceType.IsBareMetal() &&
```
bwagner5, line 154: rather than making this specific for baremetal, why not just allow any instance type if explicitly added to InstanceType requirements?  

I think if we change the AWS cloudprovider `GetInstanceTypes` to key off the passed in constraints (which it currently ignores), we can return InstanceTypes for the specific constraints. If the instance-type requirements are empty in the constrains, then we'll use the logic we use now. But if explicit InstanceTypes are specified with `In` we'd just return those instance types.

pkg/cloudprovider/types.go
```diff
@@ -66,6 +66,7 @@ type InstanceType interface {
 	AWSNeurons() *resource.Quantity
 	AWSPodENI() *resource.Quantity
 	Overhead() v1.ResourceList
+	IsBareMetal() bool
```
ellistarn, line 69: I'm not sure this is the right approach -- lets discuss offline.

bwagner5:
closing in favor of https://github.com/aws/karpenter/pull/1623

PR #1361, opened 2022-02-16
Draft: Disable CGO so that C code is not used in Go code of the package

spring1843:
Need to do some more testing please do not merge.

Disable CGO so that C code is not used in Go code of the package
and its dependencies thereby disallowing introduction C code with security and compatibility issues
For more info see: https://groups.google.com/g/golang-nuts/c/Jd9tlNc6jUE/m/fJodJUqZV-YJ
With this environment variable in place if any of our dependencies start using CGO we
will know and will either have to avoid the dependency or decide to allow it.

**1. Issue, if available:**
https://github.com/aws/karpenter/issues/1322#issuecomment-1042326861

This will disable CGO, as explained in the issue comment above this will add to our future protection against a class of bugs that can exist in C libraries and can be exposed in a Go binary that uses them.

This will make the build process fail if any of the Go libraries including dependencies use CGO, and show a message like `build constraints exclude all Go files in`

- [X] No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ export KUBEBUILDER_ASSETS ?= ${HOME}/.kubebuilder/bin
 LDFLAGS ?= "-ldflags=-X=github.com/aws/karpenter/pkg/utils/project.Version=$(shell git describe --tags --always)"
 GOFLAGS ?= "-tags=$(CLOUD_PROVIDER) $(LDFLAGS)"
 WITH_GOFLAGS = GOFLAGS=$(GOFLAGS)
+export CGO_ENABLED=0 # Disallows inclusion of CGO in the code and all dependencies
 
 ## Extra helm options
 CLUSTER_NAME ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].name}' | rev | cut -d"/" -f1 | rev)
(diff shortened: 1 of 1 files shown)


tzneal:
This also changes the way Go works slightly (e.g. uses a native Go DNS resolver instead of the system one). Just something to be aware of in case there are any weird DNS resolution issues after this is merged.

Makefile
```diff
@@ -16,6 +16,8 @@ HELM_OPTS ?= --set serviceAccount.annotations.eks\\.amazonaws\\.com/role-arn=${K
 			--set clusterEndpoint=${CLUSTER_ENDPOINT} \
 			--set aws.defaultInstanceProfile=KarpenterNodeInstanceProfile-${CLUSTER_NAME}
 
+CGO_ENABLED=0 # Disallows inclusion of CGO in the code and all dependencies
```
bwagner5, line 19: I believe you'll need to export to set this flag:

```suggestion
export CGO_ENABLED=0 # Disallows inclusion of CGO in the code and all dependencies
```

Can you move this up to the other compile related flag section of the Makefile as well?

bwagner5 (changes requested):

spring1843:
> This also changes the way Go works slightly (e.g. uses a native Go DNS resolver instead of the system one). Just something to be aware of in case there are any weird DNS resolution issues after this is merged.

Thanks for pointing that out, I'll do some more testing before merging this, do you happen to have a link with more details on this or other similar issues?

https://pkg.go.dev/net#hdr-Name_Resolution
Mentions that by by default the pure Go way is used, and a C library is used as a fallback under certain conditions e.g. on Windows.

tzneal:
> > This also changes the way Go works slightly (e.g. uses a native Go DNS resolver instead of the system one). Just something to be aware of in case there are any weird DNS resolution issues after this is merged.
> 
> Thanks for pointing that out, I'll do some more testing before merging this, do you happen to have a link with more details on this or other similar issues?
> 
> https://pkg.go.dev/net#hdr-Name_Resolution Mentions that by by default the pure Go way is used, and a C library is used as a fallback under certain conditions e.g. on Windows.

Sorry, I had forgotten exactly how it worked.  The other comment on that page indicates that the libc getaddrinfo is used "when /etc/resolv.conf or /etc/nsswitch.conf specify the use of features that the Go resolver does not implement".  I don't think it's a reason to avoid CGO_ENABLED=0,  just to be aware that in some odd environment you won't have the ability to fallback to the C resolver.

bwagner5 (commented):
Will this affect `make toolchain` for `go install <some-module>` ? 

If so, it might be better to make a `WITH_BUILD_ENV` that includes `WITH_GOFLAGS`... That does nest things quite a bit, but if we're planning on adding any other env vars that affect builds, it might be worth it. 

If this only affects `go build` (or in our case `ko build`) then I think the export is fine.

spring1843:
I'm going to close this for now because
.It is a big change and needs much more testing
.We may have to build for Windows in future so the one known incompatibility in the standard Go library may be a deal breaker

PR #1372, opened 2022-02-17
Fix GHA cache key for go modules

spring1843:
…o.sum as key

**1. Issue, if available:**

Attempt to make the GHA for CI faster by properly caching the go modules and build. Currently it seems like the cache is not as effective since the dependencies are still being deployed 

- [x] No

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,5 +1,5 @@
 name: CI
-on: [push, workflow_dispatch]
+on: [push, pull_request, workflow_dispatch]
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,17 +13,14 @@ jobs:
     - uses: actions/setup-go@v2
       with:
         go-version: 1.17
-    - run: |
-      echo "::set-output name=go-build::$(go env GOCACHE)"
-      echo "::set-output name=go-mod::$(go env GOMODCACHE)"
     - uses: actions/cache@v2
       with:
         path: |
-          ${{ steps.go-cache-paths.outputs.go-build }}
-          ${{ steps.go-cache-paths.outputs.go-mod }}
+          ~/.cache/go-build
+          ~/go/pkg/mod
           ~/go/bin/
           ~/.kubebuilder/bin/k8s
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
           ~/go/pkg/mod
           ~/go/bin/
           ~/.kubebuilder/bin/k8s
-        key: gocache
+        key: go-cache-${{ hashFiles('**/go.sum') }}
```
bwagner5, line 23: how does this make it faster?  Wouldn't the `gocache` get the same amount of hits (or maybe more) than hashing the deps checksums?

spring1843:
I tried a few different things here it doesn't look like the only improvement we can make and hope for the best is changing this key, and it will become _better_.

The problem is in very few cases there is a hit to the cache, in the other cases when it tries to save the cache it fails with this error: `It doesn't look like there are problems with the caching settings`.  The recommendation [here](https://github.com/actions/toolkit/issues/658#issuecomment-1019398065) and [here](https://github.com/actions/toolkit/issues/658#issuecomment-1019398065) is to change the key.

This will cause the cache creation to happen only when there's a real change in dependencies, and the unique key helps in not trying to overwrite an existing cache. This key is recommended [here](https://markphelps.me/posts/speed-up-your-go-builds-with-actions-cache/) and I've seen it work for another Go project so I believe replacing this key will give us more consistent results.

ellistarn:
> By "each suite" do you mean each of the tests we run for different kubernetes versions? or something else?

Each suite_test.go in our packages stands up a new kubernetes API server and ETCD, which takes about 10 seconds. IIRC ginkgo doesn't have a `BeforeAll` `AfterAll` directive, but `testenv` does allow you to reattach to an existing cluster. This can totally be a separate change though.

bwagner5 (approved):
lgtm, we can follow-up w/ more improvements 🚀

spring1843:
>  IIRC ginkgo doesn't have a BeforeAll AfterAll directive, but testenv does allow you to reattach to an existing cluster. This can totally be a separate change though.

It looks like the Ginkgo v2 DSL supports BeforeAll and AfterAll: https://github.com/onsi/ginkgo/blame/master/dsl/core/core_dsl.go#L56

ellistarn:
Super cool!

PR #1375, opened 2022-02-17
re-add on pull_request

spring1843:
**1. Issue, if available:**

It looks like we still need the on pull_request here.
I removed it in https://github.com/aws/karpenter/pull/1373 because it was double running the tests. It looks like when you add both it will each test twice and when you don't do pull_request it will not be triggered on pull request. 

 This seems to be the best of two words where it 1) Does run tests after merge 2) It runs on pull requests

- [x] No

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,7 +1,7 @@
 name: CI
 on:
   push:
-    branches: [master]
+    branches: [main]
   pull_request:
   workflow_dispatch:
 jobs:


.github/workflows/ci.yaml
```diff
@@ -1,5 +1,9 @@
 name: CI
-on: [push, workflow_dispatch]
+on:
+  push:
+    branches: [master]
```
bwagner5, line 4: should be `main`

bwagner5 (changes requested):

bwagner5 (approved):
lgtm

PR #1379, opened 2022-02-18
Fix metrics controller races

tzneal:
**1. Issue, if available:**
N/A

add some race testing and fix two identified issues

ginkgo -r --race

- [X] No

### DIFF
--- a/pkg/controllers/metrics/node/controller.go
+++ b/pkg/controllers/metrics/node/controller.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 
 	"knative.dev/pkg/logging"
 
@@ -128,15 +129,14 @@ func labelNames() []string {
 }
 
 type Controller struct {
-	KubeClient      client.Client
-	LabelCollection map[types.NamespacedName][]prometheus.Labels
+	kubeClient      client.Client
+	labelCollection sync.Map
 }
 
 // NewController constructs a controller instance
 func NewController(kubeClient client.Client) *Controller {
 	return &Controller{
-		KubeClient:      kubeClient,
-		LabelCollection: make(map[types.NamespacedName][]prometheus.Labels),
+		kubeClient: kubeClient,
 	}
 }
 
@@ -147,7 +147,7 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	c.cleanup(req.NamespacedName)
 	// Retrieve node from reconcile request
 	node := &v1.Node{}
-	if err := c.KubeClient.Get(ctx, req.NamespacedName, node); err != nil {
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, node); err != nil {
 		if errors.IsNotFound(err) {
 			return reconcile.Result{}, nil
 		}
@@ -170,7 +170,7 @@ func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
 			&source.Kind{Type: &v1alpha5.Provisioner{}},
 			handler.EnqueueRequestsFromMapFunc(func(o client.Object) (requests []reconcile.Request) {
 				nodes := &v1.NodeList{}
-				if err := c.KubeClient.List(ctx, nodes, client.MatchingLabels(map[string]string{v1alpha5.ProvisionerNameLabelKey: o.GetName()})); err != nil {
+				if err := c.kubeClient.List(ctx, nodes, client.MatchingLabels(map[string]string{v1alpha5.ProvisionerNameLabelKey: o.GetName()})); err != nil {
 					logging.FromContext(ctx).Errorf("Failed to list nodes when mapping expiration watch events, %s", err)
 					return requests
 				}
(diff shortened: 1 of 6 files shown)


pkg/controllers/metrics/node/controller.go
```diff
@@ -128,15 +128,15 @@ func labelNames() []string {
 }
 
 type Controller struct {
-	KubeClient      client.Client
-	LabelCollection map[types.NamespacedName][]prometheus.Labels
+	kubeClient      client.Client
+	labelCollection *labelsMap
```
ellistarn, line 132: Is it possible to just use a sync.map here?

tzneal: Sounds good to me, but I would change labelsMap to use a sync.Map internally to preserve the type safety in the controller.

ellistarn:
I'm a bit wary of `ExpectRaceFree`, though I'll admit it's pretty cool.

- I'd prefer to catch this sort of thing in e2e tests, where the raceable scope is much larger. I recognize that we won't be using `-race` in this environment, though
- I worry that this might give us a false sense of security that our tests won't race
- In general, testing that something doesn't happen I really hard, since there are an infinite number of things that won't happen. I could see an argument for specific cases like paralleism/racing, but I'd prefer to limit our package testing to behavior checking to catch regressions.
- I'm a bit wary of the time added to CI tests (how fast is this?)

pkg/controllers/metrics/pod/labelsmap.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type labelsMap struct {
+	mu     sync.Mutex
```
ellistarn, line 25: I really like this idiom
```suggestion
	sync.Mutex
```
It's effectively makes your struct "lockable" `m.lock`

pkg/controllers/metrics/node/controller.go
ellistarn: I think you could probably get away without the type safety. Usage of these maps are quite limited in their scope -- especially if we make them private like you've done here.
 
- pod.LabelsMap is used twice, once to get and another to set a value
- node.LabelCollection is used three times, once to get and twice to set a value

It would be a different story if these maps were exposed or used broadly. You might notice across your PRs so far that I'm really focused on minimizing complexity and the number of concepts in the codebase. Each new abstraction introduces new concepts that must be understood and used correctly by a huge community of engineers who will read/write for the future of the codebase.

tzneal:
> I'm a bit wary of `ExpectRaceFree`, though I'll admit it's pretty cool.
> 
>    * I'd prefer to catch this sort of thing in e2e tests, where the raceable scope is much larger. I recognize that we won't be using `-race` in this environment, though

You can build with -race, though  it will run much slower.  That's probably the best solution as a -race e2e test should negate the need for this entirely.

>    * I worry that this might give us a false sense of security that our tests won't race
>
>    * In general, testing that something doesn't happen I really hard, since there are an infinite number of things that won't happen. I could see an argument for specific cases like paralleism/racing, but I'd prefer to limit our package testing to behavior checking to catch regressions.

On the other hand, the test does expose two bugs :) 

>     * I'm a bit wary of the time added to CI tests (how fast is this?)

The 50 objects * 5 runs was to *hopefully* trip an issue like a concurrent map write during normal unit tests.  You can just create two objects and run them once to trigger detection with the race detector, so time should be negligible in that case.

Possible solution is to just fix the two bugs and create an issue to ensure that at least some e2e test runs on a -race build?

ellistarn:
> Possible solution is to just fix the two bugs and create an issue to ensure that at least some e2e test runs on a -race build?

Perfect.

pkg/test/expectations/expectations.go
```diff
@@ -196,3 +198,16 @@ func ExpectReconcileSucceeded(ctx context.Context, reconciler reconcile.Reconcil
 	Expect(err).ToNot(HaveOccurred())
 	return result
 }
+
+func ExpectMetricFamily(name string) *dto.MetricFamily {
```
ellistarn, line 202:
```suggestion
func ExpectMetric(prefix string) *dto.MetricFamily {
```

pkg/test/expectations/expectations.go
```diff
@@ -22,6 +22,8 @@ import (
 
 	//nolint:revive,stylecheck
 	. "github.com/onsi/gomega"
+	dto "github.com/prometheus/client_model/go"
```
ellistarn, line 25: what is `dto`? I might just alias this as `prometheus`

pkg/test/expectations/expectations.go
```diff
@@ -22,6 +22,8 @@ import (
...
+	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
```
ellistarn, line 26: Any reason to not import this as `metrics`?

pkg/test/expectations/expectations.go
tzneal: `dto` is what that package is import as by the prometheus package itself:

```
package prometheus

import (
        "errors"
        "math"
        "sync/atomic"
        "time"

        dto "github.com/prometheus/client_model/go"
)
```

I'll rename it.

pkg/test/expectations/expectations.go
tzneal: It's imported as `crmetrics` in other parts of Karpenter. I tend to keep import names the same for all uses of a package.

ellistarn: We aim to import package names without aliasing. There are exceptions:
- The package is really poorly named (e.g. github.com/prometheus/client_model/go)
- There is a conflict with two packages name the same thing.

Generally, we aim to name our packages such that naming conflicts simply don't exist between a dependency and its dependent. This is occasionally violated by libraries we rely on. When conflicts do exist, we typically move up the path, e.g. `cloudprovidermetrics "github.com/aws/karpenter/pkg/cloudprovider/metrics"`.

This pattern isn't perfect around the codebase, but we have pretty high conformance if you look around.

tzneal: SGTM

ellistarn (approved):

PR #1389, opened 2022-02-21
Fix cluster DNS IP format in Bottlerocket userdata

mKeRix:
**1. Issue, if available:**
Using Bottlerocket with `spec.kubeletConfiguration.clusterDNS` in the provider will fail to generate valid user data (`"[ip]"` instead of `"ip"`), which leads to the nodes not becoming ready.

This change matches the Bottlerocket behavior to the AL2 one, where only the first IP will be picked into the option. Since Bottlerocket OS also only supports a single IP for that option this seems to be the only plausible way atm.

Built a custom controller image and used it in my own EKS cluster w/ IPv6 (where setting the cluster DNS is required). Used `spec.provider.amiFamily: Bottlerocket`.

- [x] No

### DIFF
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -289,7 +289,7 @@ func (p *LaunchTemplateProvider) getUserData(ctx context.Context, constraints *v
 func (p *LaunchTemplateProvider) getBottlerocketUserData(ctx context.Context, constraints *v1alpha1.Constraints, additionalLabels map[string]string, caBundle *string) string {
 	userData := fmt.Sprintf("[settings.kubernetes]\ncluster-name = \"%s\"\napi-server = \"%s\"\n", injection.GetOptions(ctx).ClusterName, injection.GetOptions(ctx).ClusterEndpoint)
 	if constraints.KubeletConfiguration.ClusterDNS != nil {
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -289,7 +289,7 @@ func (p *LaunchTemplateProvider) getUserData(ctx context.Context, constraints *v
 func (p *LaunchTemplateProvider) getBottlerocketUserData(ctx context.Context, constraints *v1alpha1.Constraints, additionalLabels map[string]string, caBundle *string) string {
 	userData := fmt.Sprintf("[settings.kubernetes]\ncluster-name = \"%s\"\napi-server = \"%s\"\n", injection.GetOptions(ctx).ClusterName, injection.GetOptions(ctx).ClusterEndpoint)
 	if constraints.KubeletConfiguration.ClusterDNS != nil {
```
bwagner5, line 291: to make this more apparent, can this be:

```suggestion
	if len(constraints.KubeletConfiguration.ClusterDNS) > 0 { 
```

mKeRix: Great point. I updated the PR with your suggestion.

bwagner5:
Ah looks like there's a tiny formatting issue in that change, can you run:
```
make verify
```
and commit the change that produces. Sorry about that!

mKeRix:
Hopefully this does the trick :)

ellistarn (approved):

PR #1392, opened 2022-02-21
[WIP]Refactor scheduling and provisioner logics

felix-zhe-huang:
**1. Issue, if available:**
Will fill in soon.

- [X] No


felix-zhe-huang:
First draft. Need to clean up.

pkg/controllers/provisioning/batcher.go
```diff
@@ -22,7 +22,7 @@ import (
 
 var (
 	MaxBatchDuration  = time.Second * 10
-	BatchIdleDuration = time.Second * 1
+	BatchIdleDuration = time.Second * 8
```
felix-zhe-huang, line 25: What should the batch duration be? Currently 1 second is too short for grouping pods.

pkg/apis/provisioning/v1alpha5/labels.go
```diff
@@ -0,0 +1,133 @@
+/*
```
ellistarn, line 1: I like this separation of concerns. +1

pkg/controllers/provisioning/batcher.go
felix-zhe-huang: set it back to one. We can revisit this later. Maybe allow the user to adjust it will provide some flexibility for different use cases.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -47,26 +49,62 @@ type Provider = runtime.RawExtension
 
 // ValidatePod returns an error if the pod's requirements are not met by the constraints
 func (c *Constraints) ValidatePod(pod *v1.Pod) error {
+	p := pod.DeepCopy()
+	// The soft preference may conflict with the requirements.
+	// Remove soft constraints/ preferences
+	if p.Spec.Affinity != nil && p.Spec.Affinity.NodeAffinity != nil {
```
ellistarn, line 55: We shouldn't need to do this at this level. It should happen at the preferences.go level.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -47,26 +49,62 @@ type Provider = runtime.RawExtension
...
+	}
 	// Tolerate Taints
-	if err := c.Taints.Tolerates(pod); err != nil {
+	if err := c.Taints.Tolerates(p); err != nil {
 		return err
 	}
+	requirements := NewPodRequirements(p)
+	// Test if labels are allowed
+	for key := range requirements.Keys() {
+		if err := IsRestrictedLabel(key); err != nil && !PodLabelExceptions.Has(key) {
```
ellistarn, line 65: Do we even need to validate this? These labels are combined with the provisioner. If the pod requests labels that the provisioner doesn't support, scheduling can't happen.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -47,26 +49,62 @@ type Provider = runtime.RawExtension
...
+			// NotIn, Exists and DoesNotExist will have a complement value set.
+			// Only write a random value if the requirements operator is Exists.
+			// NotIn operator and DoesNotExist operator will schedule without the label
+			if values.IsFull() {
+				label := make([]byte, 32)
+				_, err := rand.Read(label)
+				if err != nil {
+					panic(err)
+				}
+				labels[key] = base32.StdEncoding.EncodeToString(label)[:10]
```
ellistarn, line 101: Let's extract this logic into a helper in `util`

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -47,26 +49,62 @@ type Provider = runtime.RawExtension
...
+			}
+		}
+	}
 	return &Constraints{
-		Labels:               c.Labels,
-		Requirements:         c.Requirements.Add(NewPodRequirements(pod).Requirements...).WellKnown(),
+		Labels:               labels,
```
ellistarn, line 106: Do we still need labels?

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -203,6 +140,7 @@ func (r Requirements) CapacityTypes() stringsets.String {
 }
 
 // Validate validates the feasibility of the requirements.
+// Do not apply validation to requiremnts after merging with other requirements.
```
ellistarn, line 143: Why not?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -157,6 +157,7 @@ func (p *Provisioner) launch(ctx context.Context, constraints *v1alpha5.Constrai
 		pods <- ps
 	}
 	return p.cloudProvider.Create(ctx, constraints, packing.InstanceTypeOptions, packing.NodeQuantity, func(node *v1.Node) error {
+		// Inserting labels from constraints
```
ellistarn, line 160: I'm not sure this comment adds much value. 
```suggestion
```

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -75,52 +72,30 @@ func (s *Scheduler) Solve(ctx context.Context, provisioner *v1alpha5.Provisioner
 		return nil, fmt.Errorf("injecting topology, %w", err)
 	}
 	// Separate pods into schedules of isomorphic scheduling constraints.
-	schedules, err := s.getSchedules(ctx, constraints, pods)
-	if err != nil {
-		return nil, fmt.Errorf("getting schedules, %w", err)
-	}
+	schedules := s.getSchedules(constraints, pods)
```
ellistarn, line 75: return s.getSchedules(constraints, pods), nil

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -75,52 +72,30 @@ func (s *Scheduler) Solve(ctx context.Context, provisioner *v1alpha5.Provisioner
...
-		schedulingConstraints := struct {
-			*v1alpha5.Constraints
-			GPURequests v1.ResourceList
-		}{
-			Constraints: tightened,
-			GPURequests: resources.GPULimitsFor(pod),
+		isCompatible := false
+		for index, schedule := range schedules {
+			if err := schedule.Requirements.Compatible(v1alpha5.NewPodRequirements(pod)); err == nil {
+				//TODO: The cross product of the constraints values from cloud provider may not be supported by any instance type.
```
ellistarn, line 88: Can you cut an issue for this?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -75,52 +72,30 @@ func (s *Scheduler) Solve(ctx context.Context, provisioner *v1alpha5.Provisioner
...
+				// Need to test if there is any instance type that can support the combined constraints.
+				schedules[index].Constraints = schedules[index].Tighten(pod)
+				schedules[index].Pods = append(schedules[index].Pods, pod)
+				isCompatible = true
```
ellistarn, line 92: Instead of using a boolean and break, can we simply return once we've found a match? 

This is a pretty common idiom

```
for ... {
    if found {
       return ...
    }
}
// not found case
return ...

```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -159,17 +122,18 @@ var _ = Describe("Constraints", func() {
...
 					{Key: "test-key", Operator: v1.NodeSelectorOpNotIn, Values: []string{"test-value"}},
 				}},
 			))[0]
-			ExpectNotScheduled(ctx, env.Client, pod)
+			node := ExpectScheduled(ctx, env.Client, pod)
+			Expect(node.Labels).To(HaveKeyWithValue("test-key", "test-value"))
 		})
 	})
-	Context("Well Known Labels", func() {
+	Context("CloudProvider Injected Constraints", func() {
```
ellistarn, line 136: Let's discuss these context names. This seems off to me.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -47,26 +49,62 @@ type Provider = runtime.RawExtension
...
 func (c *Constraints) Tighten(pod *v1.Pod) *Constraints {
+	requirements := c.Requirements.Add(NewPodRequirements(pod).Requirements...)
+	labels := map[string]string{}
+	for key, value := range c.Labels {
+		labels[key] = value
+	}
+	for key := range requirements.Keys() {
+		if !IsRestrictedNodeLabel(key) {
+			values := requirements.Get(key)
+			if !values.IsComplement() && !values.IsEmpty() {
```
ellistarn, line 89: Can we abstract this inside a requirements helper? 
```
requirements.PickAny(key)
```

pkg/apis/provisioning/v1alpha5/labels.go
```diff
@@ -0,0 +1,126 @@
...
+	// LabelDomainException are sub-domains of the RestrictedLabelDomains but allowed because
+	// they are not used in a context where they may be passed as argument to kubelet.
+	LabelDomainExceptions = stringsets.NewString(
+		"kops.k8s.io",
+	)
+
+	// LabelExceptions are labels that belong to the RestrictedLabelDomains but allowed.
+	// Karpenter is aware of these labels, and they can be used to further narrow down
+	// the range of the corresponding values by either provisioner or pods.
+	LabelExceptions = stringsets.NewString(
```
ellistarn, line 50: Why did you rename this from WellKnownLabels? This is a core kubernetes concept: https://kubernetes.io/docs/reference/labels-annotations-taints/

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -223,26 +161,30 @@ func (r Requirements) Validate() (errs error) {
...
 func (r Requirements) Compatible(requirements Requirements) (errs error) {
 	for _, key := range r.Keys().Union(requirements.Keys()).UnsortedList() {
 		// Key must be defined if required
 		if values := requirements.Get(key); values.Len() != 0 && !values.IsComplement() && !r.hasRequirement(withKey(key)) {
 			errs = multierr.Append(errs, fmt.Errorf("require values for key %s but is not defined", key))
 		}
-		// Values must overlap
-		if values := r.Get(key); values.Intersection(requirements.Get(key)).Len() == 0 {
+		// Values must overlap except DoesNotExist operator
+		// Excluding DoesNotExist cases to avoid generating mutiple error mesasges for the same error.
```
ellistarn, line 186: Can you show a case of what this error message looks like? Line 187 makes my brain hurt.
```suggestion
		// Excluding DoesNotExist cases to avoid generating multiple error messages for the same error.
```

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -223,26 +161,30 @@ func (r Requirements) Validate() (errs error) {
 			r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpExists))) {
 			errs = multierr.Append(errs, fmt.Errorf("operator %s cannot coexist with other operators for key %s", v1.NodeSelectorOpDoesNotExist, requirement.Key))
 		}
+		// Excludes cases when In and NotIn have overlaps.
```
ellistarn, line 164: How were we missing this case before? Is this caught somewhere else? Isn't it acceptable to have overlap of In and NotIn, we just take the intersection?

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -15,6 +15,8 @@ limitations under the License.
 package v1alpha5
 
 import (
+	"crypto/rand"
```
tzneal, line 18: math/rand read

felix-zhe-huang: Good point. Done.

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: Without doing it, pod validation will fail if preferences is conflicting with requirements because at the beginning all preferences are expressed as hard requirements.

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: This validation function is also used to validate provisioner. Now we allow the provisioner to use arbitrary labels. We need to make sure that reserved labels like those of k8s.io/XXX are not used.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -47,26 +49,62 @@ type Provider = runtime.RawExtension
...
 }
 
 func (c *Constraints) Tighten(pod *v1.Pod) *Constraints {
+	requirements := c.Requirements.Add(NewPodRequirements(pod).Requirements...)
+	labels := map[string]string{}
+	for key, value := range c.Labels {
+		labels[key] = value
+	}
+	for key := range requirements.Keys() {
+		if !IsRestrictedNodeLabel(key) {
```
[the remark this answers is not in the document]
felix-zhe-huang: In this case it won't help much I think. It will look like
```
	for key := range requirements.Keys() {
		if IsRestrictedNodeLabel(key) {
			continue
		}
		values := requirements.Get(key)
		if !values.IsComplement() && !values.IsEmpty() {
			labels[key] = values.Values().UnsortedList()[0]
		}
		if values.IsFull() {
			label := make([]byte, 32)
			_, err := rand.Read(label) //nolint
			if err != nil {
				panic(err)
			}
			labels[key] = base32.StdEncoding.EncodeToString(label)[:10]
		}
	}
```
The first `if` after the for loop is filtering our the edgeCases.

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: Since we only pick a value for complement set or Exists operator, when we abstract this inside the requirement helper, we still need all these conditions in Constraints.go. It looks to me like unnecessary duplications.

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/labels.go
felix-zhe-huang: Currently Karpenter's WellKnownLabel set is a super set of the kuberetes WellKnownLabels (e.g., karpenter.sh domain is in it).
```
	WellKnownLabels = stringsets.NewString(
		v1.LabelTopologyZone,
		v1.LabelInstanceTypeStable,
		v1.LabelArchStable,
		v1.LabelOSStable,
		LabelCapacityType,
		v1.LabelHostname, // Used internally for hostname topology spread
	)
```
It is really trying to capture the concept of restricted labels.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: After merging requirements, values may got canceled out and validation will return error due to 
```
if r.Get(key).Len() == 0 && !r.hasRequirement(withKeyAndOperator(key, v1.NodeSelectorOpDoesNotExist)) {
	errs = multierr.Append(errs, fmt.Errorf("no feasible value for key %s", key))
}
```
People should really use compatible in this case.

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: DoesNotExist requires DoesNotExist, key topology.kubernetes.io/zone; Exists prohibits DoesNotExist, key topology.kubernetes.io/zone

pkg/controllers/provisioning/provisioner.go
felix-zhe-huang: done

pkg/controllers/provisioning/scheduling/scheduler.go
felix-zhe-huang: This multi-dimentional matching problem is actually the core of the scheduling problem that make it NP-complete. We are avoiding it because of some strong assumptions and relaxations that allows us to decouple pod grouping and node selection. The virtual node solution will allow us to properly model this process.

Optimally, the solution workflow will be like:
1. Given the current instance types selection for the virtual nodes, find pod groupings.
2. Adjust the  instance type selections (e.g., number of nodes, switch to a compatible type) to optimize the objective
3. Recalculate step 1 to see the new  instance type selections in step 2 is feasible.
Repeat until converge.

The above validation in line 87 is actually a simplified version of step 2 + 3.
The above iterative approach can be extremely expensive. Luckily cloud providers like AWS has quite a collection of homogenous instance type with different size in each instance type groups. We should make use of that property and do some heuristic design to short cut step 3. 

The virtual node approach will be the core of the future consolidation solution as well. Let's spend some time to go over the design.

pkg/controllers/provisioning/scheduling/scheduler.go
felix-zhe-huang: Unfortunately it is not easy to use that idiom in this case because we still need to loop through other pods after the `if found` case. The actual common idiom that we need is more like
```
for ... {
NewIteration:
    for ... {
        if found{
            continue NewIteration
        } 
        // Do something if not found
    }
}
```
I don't like it because it closely resemble `goto`. But I am open to suggestion here.

pkg/controllers/provisioning/scheduling/scheduler.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/constraints.go
ellistarn: If they conflict -- it should fail. I don't think we want this to maintain the additional complexity to help dodge user misconfigurations. We intentionally want the logs to complain about the conflict and then relax.

pkg/apis/provisioning/v1alpha5/constraints.go
ellistarn: We're using ValidatePod to validate the provisioner?

pkg/apis/provisioning/v1alpha5/labels.go
ellistarn: Given that this is now the set of known labels (without hostname), I really think that's what we should call it.

pkg/apis/provisioning/v1alpha5/labels.go
```diff
@@ -0,0 +1,126 @@
...
+		v1.LabelTopologyRegion,
+	)
+)
+
+// IsRestrictedLabel returns an error if the label is restricted.
+func IsRestrictedLabel(key string) error {
+	if RestrictedLabels.Has(key) {
+		return fmt.Errorf("label is restricted, %s", key)
+	}
+	if LabelExceptions.Has(key) {
```
ellistarn, line 93: minor/optional nit. Can we move this before line 90 so that the flow is

1. key exceptions
2. key restrictions
3. domain exceptions
4. domain restrictions

pkg/apis/provisioning/v1alpha5/constraints.go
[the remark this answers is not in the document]
ellistarn: The benefit is squashing an additional layer of indentation. It's optional -- happy to leave as is if you feel strongly.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -47,26 +49,57 @@ type Provider = runtime.RawExtension
...
+			values := requirements.Get(key)
+			if !values.IsComplement() && !values.IsEmpty() {
+				labels[key] = values.Values().UnsortedList()[0]
+			}
+			// NotIn, Exists and DoesNotExist will have a complement value set.
```
ellistarn, line 92: From this comment, I realize that we only want to generate a value of we have a complement set, can we do something like
```
if !values.IsComplement() {
  continue
}
if values.Len() == 0 {
  labels[key] = rand.String(10)
} else {
  labels[key] = values.Values().UnsortedList()[0]
}
```
Then we can get rid of the helpers, who's names are a bit confusing.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -47,26 +49,57 @@ type Provider = runtime.RawExtension
...
+			// NotIn operator and DoesNotExist operator will schedule without the label
+			if values.IsFull() {
+				labels[key] = rand.String(10)
+			}
+		}
+	}
 	return &Constraints{
-		Labels:               c.Labels,
-		Requirements:         c.Requirements.Add(NewPodRequirements(pod).Requirements...).WellKnown(),
+		Labels:               labels,
```
ellistarn, line 101: From our chat, we can remove this.

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -225,24 +163,24 @@ func (r Requirements) Validate() (errs error) {
...
 func (r Requirements) Compatible(requirements Requirements) (errs error) {
 	for _, key := range r.Keys().Union(requirements.Keys()).UnsortedList() {
 		// Key must be defined if required
 		if values := requirements.Get(key); values.Len() != 0 && !values.IsComplement() && !r.hasRequirement(withKey(key)) {
 			errs = multierr.Append(errs, fmt.Errorf("require values for key %s but is not defined", key))
 		}
-		// Values must overlap
-		if values := r.Get(key); values.Intersection(requirements.Get(key)).Len() == 0 {
+		// Values must overlap except DoesNotExist operator
+		// Excluding DoesNotExist cases to avoid generating multiple error mesasges for the same error.
```
ellistarn, line 182:
```suggestion
		// Both DoesNotExist and conflicting { In, NotIn } rules are represented by the empty set. DoesNotExist is a valid configuration, but conflicting { In, NotIn } is not. 
```

pkg/controllers/provisioning/provisioner.go
```diff
@@ -100,7 +100,7 @@ func (p *Provisioner) provision(ctx context.Context) error {
 		}
 	}
 	// Separate pods by scheduling constraints
-	schedules, err := p.scheduler.Solve(ctx, p.Provisioner, pods)
+	schedules, err := p.scheduler.Solve(ctx, p.Provisioner, p.cloudProvider, pods)
```
ellistarn, line 103: Let's inject the cloud provider using dependency injection.

pkg/utils/rand/rand.go
```diff
@@ -0,0 +1,29 @@
...
+*/
+
+package rand
+
+import (
+	"encoding/base32"
+	"math/rand"
+)
+
+func String(lenght int) string {
```
ellistarn, line 22: length

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: Agree. It has been removed.

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: I mis-spoken. Given that all keys will be defined in the provisioner, we are fine to not check during pod validation.

pkg/controllers/provisioning/scheduling/suite_test.go
felix-zhe-huang: Done.

pkg/apis/provisioning/v1alpha5/labels.go
felix-zhe-huang: Agree. Done.

pkg/apis/provisioning/v1alpha5/labels.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: done.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -47,26 +49,57 @@ type Provider = runtime.RawExtension
...
 	return nil
 }
 
 func (c *Constraints) Tighten(pod *v1.Pod) *Constraints {
+	requirements := c.Requirements.Add(NewPodRequirements(pod).Requirements...)
+	labels := map[string]string{}
+	for key, value := range c.Labels {
+		labels[key] = value
+	}
+	for key := range requirements.Keys() {
```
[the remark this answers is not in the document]
felix-zhe-huang: Done.

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: done

pkg/controllers/provisioning/provisioner.go
felix-zhe-huang: done.

pkg/utils/rand/rand.go
felix-zhe-huang: done

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -51,22 +51,48 @@ func (c *Constraints) ValidatePod(pod *v1.Pod) error {
...
 	}
-	// Test if pod requirements are compatible
+	// Test if pod requirements are compatible to the provisioner
 	if errs := c.Requirements.Compatible(requirements); errs != nil {
 		return fmt.Errorf("incompatible requirements, %w", errs)
 	}
 	return nil
 }
 
+func (c *Constraints) GenerateLabels() *Constraints {
```
ellistarn, line 66: Rather than mutating constraints, can we just return the map and allow the caller to use them as useful? This is approach is more stateless/functional.
```suggestion
func (c *Constraints) GenerateLabels() map[string]string {
```

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -203,6 +158,7 @@ func (r Requirements) CapacityTypes() stringsets.String {
 }
 
 // Validate validates the feasibility of the requirements.
+// Do not apply validation to requiremnts after merging with other requirements.
```
ellistarn, line 161:
```suggestion
// Do not apply validation to requirements after merging with other requirements.
```

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -74,53 +74,67 @@ func (s *Scheduler) Solve(ctx context.Context, provisioner *v1alpha5.Provisioner
...
-			*v1alpha5.Constraints
-			GPURequests v1.ResourceList
-		}{
-			Constraints: tightened,
-			GPURequests: resources.GPULimitsFor(pod),
+		if !isCompatible {
+			schedules = append(schedules, &Schedule{Constraints: constraints.Tighten(pod), Pods: []*v1.Pod{pod}})
 		}
+	}
+	for _, schedule := range schedules {
```
ellistarn, line 111: I'd rather do this at launch time:
```
	return p.cloudProvider.Create(ctx, constraints, packing.InstanceTypeOptions, packing.NodeQuantity, func(node *v1.Node) error {
		node.Labels = functional.UnionStringMaps(node.Labels, constraints.GenerateLabels())
		node.Spec.Taints = append(node.Spec.Taints, constraints.Taints...)
		return p.bind(ctx, node, <-pods)
})
```

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -74,53 +74,67 @@ func (s *Scheduler) Solve(ctx context.Context, provisioner *v1alpha5.Provisioner
...
 	for _, pod := range pods {
-		if err := constraints.ValidatePod(pod); err != nil {
-			logging.FromContext(ctx).Infof("Unable to schedule pod %s/%s, %s", pod.Namespace, pod.Name, err)
-			continue
+		isCompatible := false
+		for index, schedule := range schedules {
+			if err := schedule.Requirements.Compatible(v1alpha5.NewPodRequirements(pod)); err == nil {
+				// Test if there is any instance type that can support the combined constraints
+				// TODO: Implement a virtual node approach solution that combine scheduling and node selection to solve this problem.
+				c := schedules[index].Tighten(pod)
```
ellistarn, line 96:
```suggestion
				c := schedule.Tighten(pod)
```

ellistarn: You could even shadow the variable in this inner scope
```suggestion
				constraints := schedule.Tighten(pod)
```

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: Agree. Done

pkg/apis/provisioning/v1alpha5/requirements.go
felix-zhe-huang: Done

pkg/controllers/provisioning/scheduling/scheduler.go
felix-zhe-huang: done

pkg/controllers/provisioning/scheduling/scheduler.go
felix-zhe-huang: golangci-lint doesn't like the shadowing idea. Throw out an error for `make verify`.

felix-zhe-huang: done

ellistarn (approved):
Amazing work! Really excited to see this unblock KIT. I'd like to get an approval from @tzneal, too.

felix-zhe-huang:
Meanwhile let me put down more description for the PR.

ellistarn:
Perhaps docs for Exists and DoesNotExist. Perhaps you can work with @chrisnegus

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -79,36 +78,13 @@ func (c *Constraints) validateLabels() (errs *apis.FieldError) {
 		for _, err := range validation.IsValidLabelValue(value) {
 			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("%s, %s", value, err), fmt.Sprintf("labels[%s]", key)))
 		}
-		if RestrictedLabels.Has(key) {
-			errs = errs.Also(apis.ErrInvalidKeyName(key, "labels", "label is restricted"))
-		}
-		if _, ok := WellKnownLabels[key]; !ok && IsRestrictedLabelDomain(key) {
-			errs = errs.Also(apis.ErrInvalidKeyName(key, "labels", "label domain not allowed"))
+		if err := IsRestrictedLabel(key); err != nil {
+			errs = errs.Also(apis.ErrInvalidKeyName(key, "labels", fmt.Sprintf("%s", err)))
```
tzneal, line 82: Small nit, could just be:
`errs = errs.Also(apis.ErrInvalidKeyName(key, "labels", err.Error()))`

pkg/utils/rand/rand.go
```diff
@@ -0,0 +1,29 @@
...
+package rand
+
+import (
+	"encoding/base32"
+	"math/rand"
+)
+
+func String(length int) string {
+	label := make([]byte, 32)
+	_, err := rand.Read(label) //nolint
```
tzneal, line 24: Crashes for length >= 57, `make([]byte,length)` will generate more than enough data to get `length` characters out of it. Not a big deal at the moment since it's not called with large lengths.

tzneal (approved):

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
felix-zhe-huang: Good catch. Done.

tzneal (approved):
LGTM

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -74,53 +74,64 @@ func (s *Scheduler) Solve(ctx context.Context, provisioner *v1alpha5.Provisioner
...
 	for _, pod := range pods {
-		if err := constraints.ValidatePod(pod); err != nil {
-			logging.FromContext(ctx).Infof("Unable to schedule pod %s/%s, %s", pod.Namespace, pod.Name, err)
-			continue
+		isCompatible := false
+		for _, schedule := range schedules {
+			if err := schedule.Requirements.Compatible(v1alpha5.NewPodRequirements(pod)); err == nil {
+				// Test if there is any instance type that can support the combined constraints
+				// TODO: Implement a virtual node approach solution that combine scheduling and node selection to solve this problem.
+				c := schedule.Tighten(pod)
```
tzneal, line 96: This test would fail currently since the instance type checking isn't done.  Previously they would get different schedules due to hashing, so it wasn't an issue. 

```go
It("should launch instances for Nvidia and AMD GPU resources on different instances", func() {
				nodeNames := sets.NewString()
				for _, pod := range ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner,
					test.UnschedulablePod(test.PodOptions{
						ResourceRequirements: v1.ResourceRequirements{
							Requests: v1.ResourceList{resources.NvidiaGPU: resource.MustParse("1")},
							Limits:   v1.ResourceList{resources.NvidiaGPU: resource.MustParse("1")},
						},
					}),
					// Should pack onto a different instance
					test.UnschedulablePod(test.PodOptions{
						ResourceRequirements: v1.ResourceRequirements{
							Requests: v1.ResourceList{resources.AMDGPU: resource.MustParse("1")},
							Limits:   v1.ResourceList{resources.AMDGPU: resource.MustParse("1")},
						},
					})) {
					nodeNames.Insert(node.Name)
				}
			})
```

tzneal (changes requested):
I wanted to discuss the instance type checking issue before merging.

tzneal (dismissed):
LGTM

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -51,22 +51,42 @@ func (c *Constraints) ValidatePod(pod *v1.Pod) error {
...
+	labels := map[string]string{}
+	for key, value := range c.Labels {
+		labels[key] = value
+	}
+	for key := range c.Requirements.Keys() {
+		if !IsRestrictedNodeLabel(key) {
+			values := c.Requirements.Get(key)
+			// Ignore cases when values set is empty (i.e., DoesNotExist or <In, NotIn> cancling out)
+			// and cases with Complement set that is not a full set (i.e., only NotIn)
+			if values.IsEmpty() || (values.IsComplement() && values.Len() != 0) {
```
ellistarn, line 76: I think this breaks if you have `NotIn` and `Exists`
```suggestion
			if values.IsEmpty() {
```

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -51,22 +51,42 @@ func (c *Constraints) ValidatePod(pod *v1.Pod) error {
...
 }
 
+func (c *Constraints) GenerateLabels() map[string]string {
+	labels := map[string]string{}
+	for key, value := range c.Labels {
+		labels[key] = value
+	}
+	for key := range c.Requirements.Keys() {
+		if !IsRestrictedNodeLabel(key) {
+			values := c.Requirements.Get(key)
```
ellistarn, line 73: can you inline values, since it's only used in one place.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -74,53 +74,64 @@ func (s *Scheduler) Solve(ctx context.Context, provisioner *v1alpha5.Provisioner
 	if err := s.Topology.Inject(ctx, constraints, pods); err != nil {
 		return nil, fmt.Errorf("injecting topology, %w", err)
 	}
-	// Separate pods into schedules of isomorphic scheduling constraints.
-	schedules, err := s.getSchedules(ctx, constraints, pods)
+	instanceTypes, err := s.CloudProvider.GetInstanceTypes(ctx, constraints.Provider)
```
ellistarn, line 77: We're already getting instance types in provisioner.go. Can we pass them in rather than wiring up the cloudprovider in here?

pkg/apis/provisioning/v1alpha5/constraints.go
felix-zhe-huang: done

pkg/controllers/provisioning/scheduling/scheduler.go
felix-zhe-huang: Yeah we definitely can. Done.

ellistarn (approved):

Issue #1394, opened 2022-02-21
Support optional Context for EC2 Fleet


gjmveloso:
**Tell us about your request**
Support optional `Context` parameter when creating instances using `CreateFleet` operation

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
`Context` parameter is an optional and reserved parameter that should be used by customers when advised by EC2 team at AWS. By adding an optional parameter to the AWS cloudprovider spec customers that need to use `Context` would be able to use Karpenter

**Are you currently working around this issue?**
There is no current work around and customers that needed to use `Context` when issuing `CreateFleet` API/SDK can't use Karpenter at this time

**Additional context**
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html

**Attachments**
If you think you might have additional information that you'd like to include via an attachment, please do - we'll take a look. (Remember to remove any personally-identifiable information.)

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ellistarn:
Can you give an example of how this might be used?

gjmveloso:
For those specific customers that were advised by AWS to use this parameter they would like to have all their EC2 Fleet `CreateFleet` and `ModifyFleet` using it, including for those launched by Karpenter.

dkheyman:
Currently EKS managed node groups call the ASG APIs behind the scenes to then provision the assigned instances. The `Context` parameter in `create-auto-scaling-group` API call is provided, but `create-nodegroup` it is not. We need support for the `Context` parameter in the EKS API in order to allow customers to take advantage of potential custom pricing agreements made via this parameter.

PR #1395, opened 2022-02-21
Initial work on supporting 'Context' on ec2.CreateFleet API call

gjmveloso:
**1. Issue, if available:**
#1394 

When the AWS Cloud Provider issues a `CreateFleet` API call as part of `instance.launchInstances()` a `Context` parameter as string will be added if `fleetContext` was set

N/A

- [ ] No


bwagner5:
I think making the Fleet Context a CLI parameter might be better since not many users will need to touch it. 

Here's where the CLI args are defined:

https://github.com/aws/karpenter/blob/main/pkg/utils/options/options.go#L43-L45

The options struct is injected into the context so that the cloud provider can use them. Here's an example:
https://github.com/aws/karpenter/blob/main/pkg/cloudprovider/aws/launchtemplate.go#L124

pkg/cloudprovider/aws/instance.go
```diff
@@ -128,11 +128,19 @@ func (p *InstanceProvider) launchInstances(ctx context.Context, constraints *v1a
 			},
 		},
 	}
+
+	fleetContext := injection.GetOptions(ctx).AWSEC2FleetContext
```
bwagner5, line 132: I think you can just add `Context` to the CreateFleetInput struct above. The default should be nil anyways, so doing this should be equivalent

```
...
Context: aws.String(injection.GetOptions(ctx).AWSEC2FleetContext)
```

pkg/cloudprovider/aws/instance.go
```diff
@@ -128,11 +128,19 @@ func (p *InstanceProvider) launchInstances(ctx context.Context, constraints *v1a
...
+	if fleetContext != "" {
+		createFleetInput.Context = aws.String(fleetContext)
+	}
+
 	if capacityType == v1alpha1.CapacityTypeSpot {
 		createFleetInput.SpotOptions = &ec2.SpotOptionsRequest{AllocationStrategy: aws.String(ec2.SpotAllocationStrategyCapacityOptimizedPrioritized)}
 	} else {
 		createFleetInput.OnDemandOptions = &ec2.OnDemandOptionsRequest{AllocationStrategy: aws.String(ec2.FleetOnDemandAllocationStrategyLowestPrice)}
 	}
+
```
bwagner5, line 143: remove this extra line break :)

bwagner5 (changes requested):

ellistarn:
I'd like to hold on this feature, since it tightly couples Karpenter to the fleet API and leaks Fleet details through the API. This will break if Karpenter ever decides to use ASG for RunInstances.

bwagner5:
> I'd like to hold on this feature, since it tightly couples Karpenter to the fleet API and leaks Fleet details through the API. This will break if Karpenter ever decides to use ASG for RunInstances.

FYI ASG supports Context as well :) 

https://docs.aws.amazon.com/cli/latest/reference/autoscaling/create-auto-scaling-group.html#:~:text=instance%2Dlifetime%20%3Cvalue%3E%5D-,%5B%2D%2Dcontext%20%3Cvalue%3E%5D,-%5B%2D%2Ddesired%2Dcapacity%2Dtype

gjmveloso:
> > I'd like to hold on this feature, since it tightly couples Karpenter to the fleet API and leaks Fleet details through the API. This will break if Karpenter ever decides to use ASG for RunInstances.
> 
> FYI ASG supports Context as well :)
> 
> https://docs.aws.amazon.com/cli/latest/reference/autoscaling/create-auto-scaling-group.html#:~:text=instance%2Dlifetime%20%3Cvalue%3E%5D-,%5B%2D%2Dcontext%20%3Cvalue%3E%5D,-%5B%2D%2Ddesired%2Dcapacity%2Dtype

Probably we would then change the options var from `AWSEC2FleetContext` to `AWSCapacityLaunchContext`

ellistarn:
Given that this is also included in ASG, I'm much more comfortable with the idea of exposing it through the API. 
I can potentially see
```
spec:
  provider:
    context:
```
or
```
spec:
  provider:
    launchContext:
```

pkg/utils/options/options.go
```diff
@@ -63,6 +64,7 @@ type Options struct {
 	AWSNodeNameConvention     string
 	AWSENILimitedPodDensity   bool
 	AWSDefaultInstanceProfile string
+	AWSEC2FleetContext        string
```
ellistarn, line 67: If we're serious about exposing this, we should absolutely do so at the provisioner level. The only reason to make this a CLI arg is to enable this without promoting to the full API. There are compatibility concerns with CLI args too, so it doesn't get it fully out of the woods. 

Is it interesting to use multiple fleet contexts in a single cluster?

gjmveloso:
> Given that this is also included in ASG, I'm much more comfortable with the idea of exposing it through the API. I can potentially see
> 
> ```
> spec:
>   provider:
>     context:
> ```
> 
> or
> 
> ```
> spec:
>   provider:
>     launchContext:
> ```

Reverted to the original code changes and renamed `fleetContext` to `context`, more aligned with your last suggestion

pkg/cloudprovider/aws/apis/v1alpha1/provider_validation.go
```diff
@@ -115,6 +117,16 @@ func (a *AWS) validateTags() (errs *apis.FieldError) {
 	return errs
 }
 
+func (a *AWS) validateContext() (errs *apis.FieldError) {
```
bwagner5, line 120: I think we can remove this validation and assume the user will provide a string if they're using it. If not, I think the logs will sufficiently convey an error.

pkg/cloudprovider/aws/instance.go
```diff
@@ -155,6 +157,9 @@ func (p *InstanceProvider) launchInstances(ctx context.Context, constraints *v1a
 	} else {
 		createFleetInput.OnDemandOptions = &ec2.OnDemandOptionsRequest{AllocationStrategy: aws.String(ec2.FleetOnDemandAllocationStrategyLowestPrice)}
 	}
+	if fleetContext != "" {
+		createFleetInput.Context = &fleetContext
```
bwagner5, line 161: can you remove the `getFleetContext` func and pass the `constraints.Context` directly to the Context field in the CreateFleetInput struct? 

```
createFleetInput := &ec2.CreateFleetInput{
    Type: aws.STring(ec2.FleetTypeInstant),
    Context: constraints.Context,
....
```

website/content/en/v0.7.0/AWS/provisioning.md
```diff
@@ -188,6 +188,18 @@ spec:
           throughput: 125
 ```
 
+### Context
```
bwagner5, line 191: can you remove this and update the `preview` docs ? We store snapshot docs for the versions that are meant to be immutable so that users can view the docs at a specific version snapshot.  When a new release is made, the preview docs will become that version's snapshot.

bwagner5 (changes requested):
a few comments to simplify the code a bit.

gjmveloso:
> a few comments to simplify the code a bit.

Done ;)

pkg/cloudprovider/aws/instance.go
```diff
@@ -140,6 +140,7 @@ func (p *InstanceProvider) launchInstances(ctx context.Context, constraints *v1a
 	tags := v1alpha1.MergeTags(ctx, constraints.Tags, map[string]string{fmt.Sprintf("kubernetes.io/cluster/%s", injection.GetOptions(ctx).ClusterName): "owned"})
 	createFleetInput := &ec2.CreateFleetInput{
 		Type:                  aws.String(ec2.FleetTypeInstant),
+		Context:               aws.String(*constraints.Context),
```
bwagner5, line 143: this will actually result in a nil pointer fault.  

This should work:
```suggestion
		Context:             constraints.Context,
```

pkg/cloudprovider/aws/apis/v1alpha1/provider_validation.go
```diff
@@ -31,6 +31,7 @@ const (
 	metadataOptionsPath         = "metadataOptions"
 	instanceProfilePath         = "instanceProfile"
 	blockDeviceMappingsPath     = "blockDeviceMappings"
+	contextPath                 = "context"
```
bwagner5, line 34: this isn't needed anymore

bwagner5 (changes requested):
can you run `make verify` as well

gjmveloso:
> can you run `make verify` as well

receiving the following error:

```
./hack/toolchain.sh
go: downloading github.com/norwoodj/helm-docs v1.7.0
go install: github.com/norwoodj/helm-docs/cmd/helm-docs@v1.7.0: github.com/norwoodj/helm-docs@v1.7.0: verifying module: checksum mismatch
        downloaded: h1:ibDij3KZY3nDKN62bYqxD8GlSvr89FagIfl9tEd5QA4=
        sum.golang.org: h1:m+IHHdNs8QEo2y5ooLFr7NuhLTpTuyLTOhGZfa1DgMU=

SECURITY ERROR
This download does NOT match the one reported by the checksum server.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
make: *** [toolchain] Error 1
```

tzneal:
It looks like the author of that package re-tagged it.  Try the following:

`GOSUMDB=off ./hack/toolchain.sh`

ellistarn:
> > can you run `make verify` as well
> 
> receiving the following error:
> 
> ```
> ./hack/toolchain.sh
> go: downloading github.com/norwoodj/helm-docs v1.7.0
> go install: github.com/norwoodj/helm-docs/cmd/helm-docs@v1.7.0: github.com/norwoodj/helm-docs@v1.7.0: verifying module: checksum mismatch
>         downloaded: h1:ibDij3KZY3nDKN62bYqxD8GlSvr89FagIfl9tEd5QA4=
>         sum.golang.org: h1:m+IHHdNs8QEo2y5ooLFr7NuhLTpTuyLTOhGZfa1DgMU=
> 
> SECURITY ERROR
> This download does NOT match the one reported by the checksum server.
> The bits may have been replaced on the origin server, or an attacker may
> have intercepted the download attempt.
> 
> For more information, see 'go help module-auth'.
> make: *** [toolchain] Error 1
> ```

This is a bug upstream with helm-docs. You can just comment it out from toolchain.sh

PR #1403, opened 2022-02-22
create eni-max-pods.txt file in generated LT

bwagner5:
**1. Issue, if available:**
N/A

 - This change uses Karpenterr's knowledge of max-pods to generate the eni-max-pods.txt file which the EKS bootstrap script reads when starting the k8s components. There have been several issues when new instance types are released, the eks-optimized AMIs do not have the new instance types listed in the eni-max-pods.txt file. Karpenter can pick up new instance types very early which makes this a problem. Further, the Ubuntu eks-optimized AMIs are further behind than the AL2 AMIs. 
 - NOTE: there is a 16KB limit to user-data. The current user-data, with this change, is 5.2KB so there is quite a bit of headroom. 

 - Tested on a manual cluster with AL2 & Ubuntu on instance types that previously were not listed (r6i.32xlarge on Ubuntu and c6a.4x.large.

- [x] No


pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -324,8 +326,10 @@ func (p *LaunchTemplateProvider) getAL2UserData(ctx context.Context, constraints
 	var userData bytes.Buffer
 	userData.WriteString(fmt.Sprintf(`#!/bin/bash -xe
 exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
+%s
 /etc/eks/bootstrap.sh '%s' %s \
     --apiserver-endpoint '%s'`,
+		p.ReplaceENIMaxPodsFile(ctx, constraints),
```
ellistarn, line 332: I worry that we're gonna run headlong into userdata limits with this approach.

bwagner5: We could double the size of our user-data today and still be within limits. 

We could also only populate the instance type options that we are going to launch. I'm doing all of them here for cache efficiency for the LT. Since we clean the LTs up now, it might be less of a big deal, other than the small bit of latency creating the LT.

stevehipwell:
@bwagner5 have you checked out the various issues with the current AL2 max pods implementation? My specific issue is https://github.com/awslabs/amazon-eks-ami/issues/782 which might shed a bit more light on this? It also covers the reserved values which might also be something Karpenter should look at? I'm not sure what BottleRocket uses for these values but it's probably also worth considering?

PR #1405, opened 2022-02-23
lower the number of max reconciles to 100 for the selection controller

tzneal:
**1. Issue, if available:**

Each 'max concurrent reconcile' is turned into a launched goroutine
by the controller-runtime that watches the queue.  I found this as
the race detector errors out when you have more than 8k goroutines
running simultaneously.

Lower concurrent reconciliations to 100

unit testing

- [X] No

### DIFF
--- a/pkg/controllers/selection/controller.go
+++ b/pkg/controllers/selection/controller.go
@@ -179,7 +179,7 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 		NewControllerManagedBy(m).
 		Named(controllerName).
 		For(&v1.Pod{}).
-		WithOptions(controller.Options{MaxConcurrentReconciles: 10_000}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 100}).
 		WithLogger(zapr.NewLogger(zap.NewNop())).
 		Complete(c)
 }


tzneal:
10k goroutines watching the same queue seems excessive, 100 probably still is.  Any reason it was 10k?

bwagner5:
> 10k goroutines watching the same queue seems excessive, 100 probably still is.  Any reason it was 10k?

Since the selection controller responds on individual pod events, karpenter may need a lot of concurrency to prevent scheduling delays. Each pod event seen in the selection controller is sent to a batch in the provisioning controller which returns a channel  that will close when the pod has successfully been bound to a node. The selection controller blocks on this channel to verify the pod was scheduled, and if not it can run  another loop to relax constraints. 

The Provisioners batching window is dynamic and can range from 1 to a few seconds so that we are efficiently packing groups of pods on to the largest nodes. The actual time to provision a node can also range from a few seconds to 10s of seconds. 

So although 10k may be slightly high, I don't think it's an outrageous value with the processing times building up over selection, batching, and provisioning. 100 would be too low though since this would effectively limit our pod batches to 100 pods. 

Is the problem just that race detection falls over after 8k go routines? That seems like a pretty small number to fall over on. We could go to 6k or so since each provisioner batch window maxes out at 2000 pods before it's flushed. So 6k would support 3 Provisioners doing 2000 pod scale ups at the same time which is a pretty big scale up. Scale ups could obviously be bigger over a period of 10-30 seconds. 

Id like to understand the race detector limit more before we make any changes though.

tzneal:
> The selection controller blocks on this channel to verify the pod was scheduled, and if not it can run another loop to relax constraints.

Ah thanks, I didn't notice the blocking before.  Closing this PR for now, but having 10k goroutines idling as a "plan A" doesn't seem ideal.

> Is the problem just that race detection falls over after 8k go routines? That seems like a pretty small number to fall over on. We could go to 6k or so since each provisioner batch window maxes out at 2000 pods before it's flushed. So 6k would support 3 Provisioners doing 2000 pod scale ups at the same time which is a pretty big scale up. Scale ups could obviously be bigger over a period of 10-30 seconds.

That was how I noticed what was occurring.  There is a change to remove the limit, but so far it didn't make it to 1.18 (https://go-review.googlesource.com/c/go/+/333529/)

PR #1410, opened 2022-02-23
encrypt bottlerocket second volume

bwagner5:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/1408

 - Bottlerocket mounts two volumes, xvda is for bottlerocket system components (like the API, OS, bootloader, etc) and xvdb is for containers (container images, logs, etc). 

 - Launched Bottlerocket node before the change to reproduce the second volume being unencrypted
 - Launched Bottlerocket node w/ change and observed both volumes are now encrypted 

- [x] No

### DIFF
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -119,7 +119,7 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, constraints *v1alpha1.
 	}
 	for amiID, instanceTypes := range amis {
 		// Ensure the launch template exists, or create it
-		launchTemplate, err := p.ensureLaunchTemplate(ctx, &launchTemplateOptions{
+		launchTemplate, err := p.ensureLaunchTemplate(ctx, constraints, &launchTemplateOptions{
 			UserData:          p.getUserData(ctx, constraints, instanceTypes, additionalLabels, caBundle),
 			ClusterName:       injection.GetOptions(ctx).ClusterName,
 			InstanceProfile:   instanceProfile,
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -182,17 +182,39 @@ func needsDocker(is []cloudprovider.InstanceType) bool {
 	return false
 }
 
-func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, options *launchTemplateOptions) (*ec2.LaunchTemplate, error) {
-	output, err := p.ec2api.CreateLaunchTemplateWithContext(ctx, &ec2.CreateLaunchTemplateInput{
-		LaunchTemplateName: aws.String(launchTemplateName(options)),
-		LaunchTemplateData: &ec2.RequestLaunchTemplateData{
-			BlockDeviceMappings: []*ec2.LaunchTemplateBlockDeviceMappingRequest{{
+func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, constraints *v1alpha1.Constraints, options *launchTemplateOptions) (*ec2.LaunchTemplate, error) {
+	blockDeviceMappings := []*ec2.LaunchTemplateBlockDeviceMappingRequest{{
```
ellistarn, line 186: Maybe worth factoring out into a method, which will make things easier when we expose this to users.

ellistarn (approved):
Nice fix! Needs docs?

pkg/cloudprovider/aws/launchtemplate.go
bwagner5: planning on refactoring when we support the provisioner root volume configuration

bwagner5:
> Nice fix! Needs docs?

We probably do need docs on the general feature of Karpenter encrypting volumes by default. I'll put in an issue for those docs. I guess they'd live in AWS/Provisioning.

PR #1417, opened 2022-02-25
Enforce error checking in CI

spring1843:
**1. Issue, if available:**

Enables a check that makes sure all errors are checked. This is good idea specially because we are already checking all errors except only in one test which this PR fixes.

There is really no good reason to leave errors unchecked. Beginners to Go are often surprised by how much error handling they have to do. In my perspective this is one of the beauties of Go. It makes the programmer think and react everywhere things can go wrong and we know at scale things that can go wrong will go wrong. Handling all errors therefore is  part of what it takes to produce high quality and reliable software.

- [x] No


ellistarn (approved):
Great change to linting. I like how you handled the sometimes case.

PR #1427, opened 2022-02-28
Helm topologySpreadConstraints

sylr:
**1. Issue, if available:**

https://github.com/aws/karpenter/issues/1354

Add possibility to specify topology spread constraints in the Helm Chart.

Manually on my setup.

- [ ] No

### DIFF
--- a/charts/karpenter/README.md
+++ b/charts/karpenter/README.md
@@ -66,6 +66,7 @@ helm upgrade --install --namespace karpenter --create-namespace \
 | strategy | object | `{"type":"Recreate"}` | Strategy for updating the pod. |
 | terminationGracePeriodSeconds | string | `nil` | Override the default termination grace period for the pod. |
 | tolerations | list | `[]` | Tolerations to allow the pod to be scheduled to nodes with taints. |
+| topologySpreadConstraints | list | `[]` | Constraints to spread replicas across the cluster. |
 | webhook.env | list | `[]` | Additional environment variables for the webhook pod. |
 | webhook.image | string | `"public.ecr.aws/karpenter/webhook:v0.6.3@sha256:cddaa650e52e62bc0520697d8662a1dab99d009e4ea3ffa176367dfe6cf84d2d"` | Webhook image. |
 | webhook.logLevel | string | `""` | Webhook log level, defaults to the global log level |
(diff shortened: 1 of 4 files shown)


charts/karpenter/values.yaml
```diff
@@ -7,7 +7,9 @@ additionalLabels: {}
 # app: karpenter
 
 # -- Additional annotations to add into metadata.
-additionalAnnotations: {}
+additionalAnnotations:
+  kubectl.kubernetes.io/default-container: controller
+  kubectl.kubernetes.io/default-logs-container: controller
```
ellistarn, line 12: TIL. This is really cool.

ellistarn:
Can you please elaborate on

> 3. How was this change tested?

charts/karpenter/README.md
```diff
@@ -45,6 +45,7 @@ helm upgrade --install --namespace karpenter --create-namespace \
 | controller.logLevel | string | `""` | Controller log level, defaults to the global log level |
 | controller.resources | object | `{"limits":{"cpu":1,"memory":"1Gi"},"requests":{"cpu":1,"memory":"1Gi"}}` | Resources for the controller pod. |
 | controller.securityContext | object | `{}` | SecurityContext for the controller container. |
+| deploymentAdditionalAnnotations | object | `{"kubectl.kubernetes.io/default-container":"controller","kubectl.kubernetes.io/default-logs-container":"controller"}` | Additional annotations to add into deployment's metadata. |
```
ellistarn, line 48: We should just bake this directly into to deployment.yaml instead of passing it through.

sylr: Yeah I thought about this but I remember that, as a Helm user, there is nothing that I hate more than when there is hardcoded properties in the templates that I can't modify without forking the chart.

bwagner5: I think this annotation is a little different since you shouldn't ever modify it. I would also advocate for this to be hardcoded in the `deployment.yaml`.

For other annotations you'd like to add, you still have the option to add to the `additionalAnnotations` field, even if it adds the annotations to more than just the deployment resource :)

stevehipwell: I agree, there are the `additionalAnnotations` or `podAnnotations` value to pass any custom annotations into the chart.

charts/karpenter/README.md
```diff
@@ -66,6 +67,7 @@ helm upgrade --install --namespace karpenter --create-namespace \
 | strategy | object | `{"type":"Recreate"}` | Strategy for updating the pod. |
 | terminationGracePeriodSeconds | string | `nil` | Override the default termination grace period for the pod. |
 | tolerations | list | `[]` | Tolerations to allow the pod to be scheduled to nodes with taints. |
+| topologySpreadConstraints | list | `[]` | Constraints to spread replicas across the cluster. |
```
stevehipwell, line 70: Why can't you use `affinity`? Unless you're running an unbounded number of replicas this would be better practice.

Issue #1414, opened 2022-02-24
Expired nodes are stuck in `Ready,SchedulingDisabled` without any error


nandiheath:
## Version
**Karpenter**: v0.6.3
**Kubernetes**: v1.21.5

## Expected Behavior
Nodes provisioned by Karpenter with expiry should be deleted after they are expired.

## Actual Behavior
The nodes are stuck in `Ready,SchedulingDisabled` state and there was no error from Karpenter controller.
All the workloads on the node were running and healthy (no single pod was terminating).
I couldn't even delete the node manually probably because of Karpenter's finalizer.

## Steps to Reproduce the Problem
deploy an istio ingressgateway deployment with nodeAffinity: (currently we are running only on amd64 nodes)
```
spec:
  affinity:
    nodeAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
      - preference:
          matchExpressions:
          - key: kubernetes.io/arch
            operator: In
            values:
            - amd64
        weight: 2
      - preference:
          matchExpressions:
          - key: kubernetes.io/arch
            operator: In
            values:
            - arm64
        weight: 2
      - preference:
          matchExpressions:
          - key: kubernetes.io/arch
            operator: In
            values:
            - ppc64le
        weight: 2
      - preference:
          matchExpressions:
          - key: kubernetes.io/arch
            operator: In
            values:
            - s390x
        weight: 2
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: kubernetes.io/arch
            operator: In
            values:
            - amd64
            - arm64
            - ppc64le
            - s390x
```

- multiple nodes expire at the same time (or manually delete them via `kubectl`)
- new nodes are created and some of the nodes are being removed
- but the nodes that has `istio-ingressgateway` pod running on it will be stuck in `Ready,SchedulingDisabled` forever
- once you manually delete the `istio-ingressgateway` pod the node will be deleted by Karpenter shortly

## Resource Specs and Logs
This is the log for the Karpenter controller.

```
2022-02-24T18:29:51.573Z	INFO	Successfully created the logger.
2022-02-24T18:29:51.573Z	INFO	Logging level set to: debug
{"level":"info","ts":1645727391.6794684,"logger":"fallback","caller":"injection/injection.go:61","msg":"Starting informers..."}
[...]

PR #1432, opened 2022-02-28
Add logging for unexpected pod eviction errors

dewjam:
**1. Issue, if available:**
#1414

The eviction API is supposed to only return 429 or 500 status codes when an error is encountered while evicting a pod (per [k8s documentation](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#eviction-api)).  Today, Karpenter only writes a debug log when a 500 or 429 status code is returned from the eviction API, but doesn't account for the case where an unexpected error/status code is received.  This change adds a catch-all so Karpenter will also log unexpected errors from the eviction API.  

Edit: This change also introduces structured logging for log emitted by eviction attempts.  Also, logs the returned error messages reported by the eviction API for consistency, instead of using custom error messages.

Related to this change, we identified what looks like a [bug in the eviction API](https://github.com/kubernetes/kubernetes/blob/v1.21.5/pkg/registry/core/pod/storage/eviction.go#L194-L198).  In the case where multiple PDBs are found for a pod, the error which is returned correlates to an InternalServerError (500), but the "reason" field excluded.  We believe this is the root cause of the problem identified in #1414 .  We will be pursuing an upstream fix, but this "catch-all" is also worth implementing to account for unexpected errors.  

* Deployed Karpenter including the change to a test EKS cluster
* Enabled debug logging
* Introduced a conflicting PDB

(edit) Result:
```
2022-03-03T15:37:19.267Z	DEBUG	controller.eviction	Cannot evict pod as it would violate the pod's disruption budget.	{"commit": "cd8a26f", "pod": "istio-system/istiod-d8576dfdf-hrsnk"}
2022-03-03T15:37:29.274Z	ERROR	controller.eviction	This pod has more than one PodDisruptionBudget, which the eviction subresource does not support.	{"commit": "cd8a26f", "pod": "istio-system/istio-ingressgateway-6559b68fb5-gpm29"}
2022-03-03T15:37:29.327Z	DEBUG	controller.eviction	Evicted pod	{"commit": "cd8a26f", "pod": "istio-system/istiod-d8576dfdf-hrsnk"}
```

- [x] No

### DIFF
--- a/pkg/controllers/termination/eviction.go
+++ b/pkg/controllers/termination/eviction.go
@@ -73,7 +73,6 @@ func (e *EvictionQueue) Start(ctx context.Context) {
 		nn := item.(types.NamespacedName)
 		// Evict pod
 		if e.evict(ctx, nn) {
-			logging.FromContext(ctx).Debugf("Evicted pod %s", nn.String())
 			e.RateLimitingInterface.Forget(nn)
 			e.Set.Remove(nn)
 			e.RateLimitingInterface.Done(nn)
@@ -83,27 +82,26 @@ func (e *EvictionQueue) Start(ctx context.Context) {
 		// Requeue pod if eviction failed
 		e.RateLimitingInterface.AddRateLimited(nn)
 	}
-	logging.FromContext(ctx).Errorf("EvictionQueue is broken and has shutdown.")
+	logging.FromContext(ctx).Errorf("EvictionQueue is broken and has shutdown")
 }
 
 // evict returns true if successful eviction call, error is returned if not eviction-related error
 func (e *EvictionQueue) evict(ctx context.Context, nn types.NamespacedName) bool {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("pod", nn))
 	err := e.coreV1Client.Pods(nn.Namespace).Evict(ctx, &v1beta1.Eviction{
 		ObjectMeta: metav1.ObjectMeta{Name: nn.Name, Namespace: nn.Namespace},
 	})
-	if errors.IsInternalError(err) { // 500
-		logging.FromContext(ctx).Errorf("Could not evict pod %s due to PDB misconfiguration error.", nn.String())
-		return false
-	}
-	if errors.IsTooManyRequests(err) { // 429
-		logging.FromContext(ctx).Debugf("Did not evict pod %s due to PDB violation.", nn.String())
-		return false
-	}
 	if errors.IsNotFound(err) { // 404
 		return true
 	}
-	if err != nil {
+	if errors.IsTooManyRequests(err) { // 429
+		logging.FromContext(ctx).Debug(err)
 		return false
 	}
+	if err != nil {
+		logging.FromContext(ctx).Error(err)
+		return false
+	}
+	logging.FromContext(ctx).Debug("Evicted pod")
 	return true
 }


njtran (approved):
LGTM! Good work

pkg/controllers/termination/eviction.go
```diff
@@ -103,6 +103,7 @@ func (e *EvictionQueue) evict(ctx context.Context, nn types.NamespacedName) bool
 		return true
 	}
 	if err != nil {
+		logging.FromContext(ctx).Debugf("Could not evict pod %s due to an unexpected error: %s", nn.String(), err)
```
ellistarn, line 106: This won't print the error string cleanly. We can also simplify this message.
```suggestion
		logging.FromContext(ctx).Debugf("Could not evict pod %s, %s", nn.String(), err.Error())
```

While we're in here, can we move the `.` from the other log lines?

dewjam: Sure thing!

ellistarn, line 106: I think this should be ERROR level.

ellistarn: Unless this is massively noisy. It only happens during misconfiguration, right?

ellistarn: See the difference between lines 95 and 99

spring1843: both `nn.String(), err.Error()` can be replaced with `nn, err` since they have .String they will both turn into strings.

spring1843 (changes requested):

pkg/controllers/termination/eviction.go
tzneal: Should the 429 violation of PDB be turned up to an info log as well?  It may save a round of "turn on debug logs" when troubleshooting why a node won't drain.

dewjam: Seems to me they all should be Error level.  "Error" logs will show up with the default log config, I believe.  

One note, we do retry eviction failures, so we will see multiple messages in the log file for each pod.

tzneal: The 429 seems like a different sort of transient condition to me. It may resolve itself if another pod becomes ready. It doesn't seem right for a log to have repeated error messages when everything is possibly working as expected. 

I agree that the multiple PDBs definitely seems like an error level condition. The user has to do something to resolve it.

ellistarn: > The 429 seems like a different sort of transient condition to me. It may resolve itself if another pod becomes ready. It doesn't seem right for a log to have repeated error messages when everything is possibly working as expected.

This is exactly the reason. We want to avoid spamming INFO logs for normal operation. We're pretty careful about what meets the bar for info. I expect most customers will want to run with DEBUG on and then filter with different tooling. Perhaps we should make DEBUG the default log level in our helm chart.

njtran: A 429 should just be debug log IMO. If you run into this error message once, you'll run into it at least 10 times per pod due to the fast exponential backoff, which would make the logs pretty noisy. 

A 500 I would think should be info level as well. An unexpected error would make sense to be info level as well.

dewjam: Updated commit with this change.

spring1843 (dismissed):

pkg/controllers/termination/eviction.go
```diff
@@ -92,17 +92,18 @@ func (e *EvictionQueue) evict(ctx context.Context, nn types.NamespacedName) bool
 		ObjectMeta: metav1.ObjectMeta{Name: nn.Name, Namespace: nn.Namespace},
 	})
 	if errors.IsInternalError(err) { // 500
-		logging.FromContext(ctx).Errorf("Could not evict pod %s due to PDB misconfiguration error.", nn.String())
+		logging.FromContext(ctx).Errorf("Could not evict pod %s due to PDB misconfiguration error", nn)
 		return false
 	}
 	if errors.IsTooManyRequests(err) { // 429
-		logging.FromContext(ctx).Debugf("Did not evict pod %s due to PDB violation.", nn.String())
+		logging.FromContext(ctx).Errorf("Did not evict pod %s due to PDB violation", nn)
```
njtran, line 99: I don't think we should be considering this an `Errorf` since it's an expected function of PDBs. Since it's fairly noisy, I'd like to keep this at the `Debug` level.

dewjam: Reverted this back to a 429 in the most recent push.

pkg/controllers/termination/eviction.go
```diff
@@ -73,7 +73,7 @@ func (e *EvictionQueue) Start(ctx context.Context) {
 		nn := item.(types.NamespacedName)
 		// Evict pod
 		if e.evict(ctx, nn) {
-			logging.FromContext(ctx).Debugf("Evicted pod %s", nn.String())
+			logging.FromContext(ctx).Debug("Evicted pod %s", nn)
```
ellistarn, line 76: This is gonna print poorly w/o Debugf

ellistarn: Does it make sense to move this log line into the final line inside `e.evict(...)`. That way we can take advantage of the `With()` call on the logger.

dewjam: Took care of that in the latest push.  Also removed logs for the 404 condition.

ellistarn (approved):
nice!

PR #1442, opened 2022-03-01
deeper podAffinity and anti-affinity checks

bwagner5:
**1. Issue, if available:**
N/A

 - This adds a deeper check for podAffinity and podAntiAffinity. Before we were only checking if it was nil. There are some helm charts will will expose the podAffinity and anti-affinity in the values.yaml and will therefore populate them as empty instead of not specified. Karpenter would ignore these pods unnecessarily. 

 - Tested a pod w/ anti-affinity -> Karpenter ignores pod
 - Tested a pod w/ empty anti-affinity -> Karpenter schedules
 - Tested a pod w/ no anti-affinity -> Karpenter schedules

- [x] No

### DIFF
--- a/pkg/controllers/selection/controller.go
+++ b/pkg/controllers/selection/controller.go
@@ -137,22 +137,22 @@ func validateTopology(pod *v1.Pod) (errs error) {
 	return errs
 }
 
-func validateAffinity(pod *v1.Pod) (errs error) {
-	if pod.Spec.Affinity == nil {
+func validateAffinity(p *v1.Pod) (errs error) {
+	if p.Spec.Affinity == nil {
 		return nil
 	}
-	if pod.Spec.Affinity.PodAffinity != nil {
+	if pod.HasPodAffinity(p) {
 		errs = multierr.Append(errs, fmt.Errorf("pod affinity is not supported"))
 	}
-	if pod.Spec.Affinity.PodAntiAffinity != nil {
+	if pod.HasPodAntiAffinity(p) {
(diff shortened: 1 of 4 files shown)


pkg/controllers/selection/controller.go
```diff
@@ -137,22 +137,22 @@ func validateTopology(pod *v1.Pod) (errs error) {
...
 }
 
-func validateAffinity(pod *v1.Pod) (errs error) {
-	if pod.Spec.Affinity == nil {
+func validateAffinity(p *v1.Pod) (errs error) {
+	if p.Spec.Affinity == nil {
 		return nil
 	}
-	if pod.Spec.Affinity.PodAffinity != nil {
+	if pod.HasPodAffinity(p) {
```
ellistarn, line 144: At one point I had a PR that attached validation behavior to code modules that handled such pieces of scheduling logic. For example

topology.Validate(pod)
antiaffinity.Validate(pod)
affinity.Validate(pod)

What do you think about moving that logic instead of adding more to the `pod` library, which is a bit of a grab bag.

bwagner5: Longer term I think that makes sense for when we implement support for affinity and anti-affinity. But since we don't support it currently, those modules wouldn't have anything except the validation logic. 

Can we defer on co-locating these until we implement those scheduling features? And at that time, these checks will actually be removed anyways.

ellistarn: SGTM.

ellistarn (commented):
Do you mind writing a test for this?

ellistarn (approved):

PR #1451, opened 2022-03-02
raise the minimum TTL seconds until expired to 300

tzneal:
**1. Issue, if available:**
N/A

Zero was a valid value which caused nodes to immediately be
terminated after creation. Update docs to replace 'nil' with
'omitted' for clarity when referring to intentionally missing
provisioner spec values.

Unit tests

- [ ] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/provisioner_validation.go
+++ b/pkg/apis/provisioning/v1alpha5/provisioner_validation.go
@@ -28,9 +28,13 @@ import (
 )
 
 var (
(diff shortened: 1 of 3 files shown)


ellistarn:
To clarify this PR, it's still possible to disable the feature by leaving the value as nil, right? Can we make sure we have a test case for this?

tzneal:
In light of today's standup discussion and the desire to:

- keep validation/behavior for both expire times identical
- not break existing provisioner specs

I think it may make sense to just make this a doc+tests change only.   Any thoughts?

tzneal:
Putting up a clearer PR with tests & doc changes only

PR #1458, opened 2022-03-03
update tags doc

tzneal:
**1. Issue, if available:**

- Document that tags are applied to created volumes as well
- Reword to indicate that we both merge & override

Unit test & doc changes only.

- [ ] No

### DIFF
--- a/website/content/en/preview/AWS/provisioning.md
+++ b/website/content/en/preview/AWS/provisioning.md
@@ -114,8 +114,15 @@ Select security groups by name using a wildcard:
 
 ### Tags
 
-Tags will be added to every EC2 Instance launched by this provisioner.
+Karpenter adds tags to all resources it creates, including EC2 Instances, EBS volumes, and Launch Templates. The default set of AWS tags are listed below.
 
+```
+Name: karpenter.sh/cluster/<cluster-name>/provisioner/<provisioner-name>
+karpenter.sh/cluster/<cluster-name>: owned
+kubernetes.io/cluster/<cluster-name>: owned
+```
+
+Additional tags can be added in the provider tags section which are merged with and can override the default tag values.
 ```
 spec:
   provider:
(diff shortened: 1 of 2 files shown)


website/content/en/preview/AWS/provisioning.md
```diff
@@ -114,8 +114,15 @@ Select security groups by name using a wildcard:
 
 ### Tags
 
-Tags will be added to every EC2 Instance launched by this provisioner.
+Karpenter adds tags to both the EC2 Instances and attached volumes launched by provisioners. The default set of AWS tags are listed below.
```
ellistarn, line 117:
```suggestion
Karpenter adds tags to all resources it creates, including EC2 Instances, EBS volumes, and Launch Templates. The default set of AWS tags are listed below.

```

website/content/en/preview/AWS/provisioning.md
```diff
@@ -114,8 +114,15 @@ Select security groups by name using a wildcard:
 
 ### Tags
 
-Tags will be added to every EC2 Instance launched by this provisioner.
+Karpenter adds tags to both the EC2 Instances and attached volumes launched by provisioners. The default set of AWS tags are listed below.
 
+```
+Name: karpenter.sh/cluster/<cluster-name>/provisioner/<provisioner-name>
+karpenter.sh/cluster/<cluster-name>: owned
+kubernetes.io/cluster/<cluster-name>: owned
```
njtran, line 122: Do we tag both instances and volumes with all three of these? Or is this the union-ed set of the two? I assume that the volume is also tagged with the instance-id as well.

tzneal: As of this week the volumes are tagged with these along with any user specified tags from the provisioner.  Unless I'm misunderstanding you, there is no instance-id tag on the volume itself. I just checked a volume attached to a node and it only has these three default tags.

ellistarn (approved):

Issue #1452, opened 2022-03-02
Error request context canceled


amitdrana:
Hi Team, 
 I am facing this issue while trying to deploy Karpenter via helm.
 Any idea..
```
ubuntu@ip-10-40-3-188:~/devops-preprod/karpenter$ kubectl logs karpenter-669bb94648-fb6ts -n karpenter controller
2022-03-02T13:44:37.416Z	INFO	Successfully created the logger.
2022-03-02T13:44:37.416Z	INFO	Logging level set to: debug
{"level":"info","ts":1646228677.5214024,"logger":"fallback","caller":"injection/injection.go:61","msg":"Starting informers..."}
2022-03-02T13:44:37.521Z	DEBUG	controller.aws	Using AWS region eu-west-1	{"commit": "fd19ba2"}
2022-03-02T13:44:37.521Z	DEBUG	controller.aws.launchtemplate	Hydrating the launch template cache with names matching "Karpenter-hyke-eks-cluster-preprod-*"	{"commit": "fd19ba2"}
panic: Unable to hydrate the AWS launch template cache, RequestCanceled: request context canceled
caused by: context canceled

goroutine 1 [running]:
github.com/aws/karpenter/pkg/cloudprovider/aws.(*LaunchTemplateProvider).hydrateCache(0xc00086d580, {0x27c0938, 0xc0003810b0})
	github.com/aws/karpenter/pkg/cloudprovider/aws/launchtemplate.go:234 +0x4f4
github.com/aws/karpenter/pkg/cloudprovider/aws.NewLaunchTemplateProvider({0x27c0938, 0xc0003810b0}, {0x28952a0, 0xc0003222d8}, 0xc00019caa0, 0xc0006dff80)
	github.com/aws/karpenter/pkg/cloudprovider/aws/launchtemplate.go:68 +0x1d1
github.com/aws/karpenter/pkg/cloudprovider/aws.NewCloudProvider({0x27c0938, 0xc000381080}, {0x1faaee0})
	github.com/aws/karpenter/pkg/cloudprovider/aws/cloudprovider.go:88 +0x4fb
github.com/aws/karpenter/pkg/cloudprovider/registry.newCloudProvider(...)
	github.com/aws/karpenter/pkg/cloudprovider/registry/aws.go:27
github.com/aws/karpenter/pkg/cloudprovider/registry.NewCloudProvider({0x27c0938, 0xc000381080}, {0x1faaee0})
[...]

Issue #1425, opened 2022-02-28
incompatible requirements, [] not in [], key kubernetes.io/hostname


sylr:
## Version

**Karpenter**: v0.6.3

**Kubernetes**: v1.21.5-eks-bc4871b

## Expected Behavior

New nodes are scheduled when I delete a node and pods on that nodes can not be scheduled on remaining existing nodes.

## Actual Behavior

No new node is being created.

## Steps to Reproduce the Problem

Delete a node hosting pods that can not be recreated on existing nodes due to tolerations.

Removing the `kubernetes.io/hostname` topology spread constraint fixes the problem.

## Resource Specs and Logs

```
2022-02-28T08:03:58.067Z	INFO	controller.termination	Cordoned node	{"commit": "fd19ba2", "node": "ip-10-128-30-125.eu-central-1.compute.internal"}
2022-02-28T08:03:58.324Z	INFO	controller.termination	Deleted node	{"commit": "fd19ba2", "node": "ip-10-128-30-125.eu-central-1.compute.internal"}
2022-02-28T08:04:16.174Z	INFO	controller.termination	Cordoned node	{"commit": "fd19ba2", "node": "ip-10-128-24-167.eu-central-1.compute.internal"}
2022-02-28T08:04:17.445Z	INFO	controller.provisioning	Batched 2 pods in 1.024899682s	{"commit": "fd19ba2", "provisioner": "dedicated-ingress-arm64"}
2022-02-28T08:04:17.445Z	INFO	controller.provisioning	Unable to schedule pod ingress/traefik-5fcc45756f-qdxv2, incompatible requirements, [] not in [rhv6mc5e], key kubernetes.io/hostname	{"commit": "fd19ba2", "provisioner": "dedicated-ingress-arm64"}
2022-02-28T08:04:17.445Z	INFO	controller.provisioning	Unable to schedule pod sso/pomerium-proxy-54f9d5b996-w8lxq, incompatible requirements, [] not in [], key kubernetes.io/hostname	{"commit": "fd19ba2", "provisioner": "dedicated-ingress-arm64"}
2022-02-28T08:04:17.447Z	INFO	controller.provisioning	Waiting for unschedulable pods	{"commit": "fd19ba2", "provisioner": "dedicated-ingress-arm64"}
2022-02-28T08:04:23.448Z	INFO	controller.provisioning	Batched 2 pods in 1.000604413s	{"commit": "fd19ba2", "provisioner": "dedicated-ingress-arm64"}
[...]

PR #1460, opened 2022-03-03
fix empty requirement matching

tzneal:
An empty list is always a valid subset of any other list.

**1. Issue, if available:**
#1452

Ensure an empty list can match any other list.

I was able to reproduce this on my cluster and tested the change live in addition to unit testing.

- [X] No


pkg/apis/provisioning/v1alpha5/suite_test.go
```diff
@@ -211,7 +211,22 @@ var _ = Describe("Validation", func() {
 			B := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpIn, Values: []string{"foo"}})
 			Expect(A.Compatible(B)).To(Succeed())
 		})
-		It("A should fail to be compatible to B, <In, In> operaton, no overlap", func() {
+		It("A should be compatible to B, <In, In> operator, A is empty list", func() {
+			// We're asking here ifs zone [] compatible with zone ["foo"].  K8s itself won't allow this to be configured on a pod
+			// as it requires values if you specify operator In or NotIn and won't allow an empty list. Logically we consider an empty set
+			// to be a subset of any other set.
+			A := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpIn, Values: []string{}})
```
ellistarn, line 218: I'm not sure I agree with this. There's a difference with a rule not being `nil` and a rule being defined as `[]`.

tzneal: Users can't create the empty rule since k8s prevents it.  The code a few lines above was similar in that it prevents an empty rule from preventing a match. 

```go
if values := requirements.Get(key); values.Len() != 0 && !values.IsComplement() && !r.hasRequirement(withKey(key)) {
			errs = multierr.Append(errs, fmt.Errorf("require values for key %s but is not defined", key))
		}
```

 Since users can't create these types of rules, another option is to go modify/fix our code so it stops putting empty rules in.

ellistarn: > another option is to go modify/fix our code so it stops putting empty rules in.

Sounds good. This sounds like an invariant is being violated.

sylr:
I build this and tested it and it seems to fix the problem I reported 👍

sylr:
Is there a reason for this to not have been merged yet ?

tzneal:
> Is there a reason for this to not have been merged yet ?

Hey @sylr , there have been several scheduler improvements this week and this PR is no longer relevant.  If you checkout v0.7.1, the specific issue of empty requirements matching should no longer happen.  I've left #1425 open as there is still an improvement or two to be made to the scheduler to handle that case better.  It should schedule now without issues.

PR #1468, opened 2022-03-03
Add Make target to serve the hugo docs site locally

bwagner5:
**1. Issue, if available:**
N/A

 - Add a make target that serves the hugo docs site locally

`make website` 

- [x] No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,4 +76,7 @@ issues: ## Run GitHub issue analysis scripts
 	./hack/github/feature_request_reactions.py > "karpenter-feature-requests-$(shell date +"%Y-%m-%d").csv"
 	./hack/github/label_issue_count.py > "karpenter-labels-$(shell date +"%Y-%m-%d").csv"
 
-.PHONY: help dev ci release test battletest verify codegen apply delete toolchain release licenses issues
+docs: ## Serve the docs site locally
+	cd website && npm install && git submodule update --init --recursive && hugo server -l layouts/ --themesDir themes/
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -76,4 +76,7 @@ issues: ## Run GitHub issue analysis scripts
 	./hack/github/feature_request_reactions.py > "karpenter-feature-requests-$(shell date +"%Y-%m-%d").csv"
 	./hack/github/label_issue_count.py > "karpenter-labels-$(shell date +"%Y-%m-%d").csv"
 
-.PHONY: help dev ci release test battletest verify codegen apply delete toolchain release licenses issues
+docs: ## Serve the docs site locally
```
spring1843, line 79: WDYT about calling this `website` to be consistent with the directory name? Specially since the website is a bit more than just docs.

spring1843 (dismissed):

Makefile
ellistarn: +1

Makefile
```diff
@@ -76,4 +76,7 @@ issues: ## Run GitHub issue analysis scripts
...
+	cd website && npm install && git submodule update --init --recursive && hugo server -l layouts/ --themesDir themes/
```
ellistarn, line 80: Hmm -- `hugo server` without the flags works for me.

bwagner5: ah you're right, only needed those when I was trying not to `cd website`, but turns out that adds a lot of complexity :D

Makefile
bwagner5: good idea, done!

spring1843 (approved):

PR #1472, opened 2022-03-04
Make script path dynamic and related to the page path

spring1843:
…either release version or preview

**1. Issue, if available:**

This will change the behavior where the scripts are being read from the `preview` directory, and instead use the release directory that comes from the path of the page, for example for the preview version it will read from `./website/content/en/preview/...` and for v0.6.4 it will read from `./website/content/en/v0.6.4/...`.

This allows us to manage the scripts of each version independently should we need to.

- [ ] No

### DIFF
--- a/website/layouts/shortcodes/script.html
+++ b/website/layouts/shortcodes/script.html
@@ -1,3 +1,5 @@
-{{ $file := .Get "file" | readFile }}
+{{ $file := .Get "file" }}
+{{ $version := index ( split .Page.Path `/` ) 0 }}
+{{ $fileContents := replace $file `{VERSION}` $version | readFile }}
 {{ $lang := .Get "language" }}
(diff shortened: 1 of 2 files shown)


website/layouts/shortcodes/script.html
```diff
@@ -1,3 +1,5 @@
-{{ $file := .Get "file" | readFile }}
+{{ $file := .Get "file" }}
```
ellistarn, line 1: Cool! Can we call this file version.html?

spring1843: The name of the files corresponds to the shortcode tag, e.g. `{{% script file="...` so if we change the name of the file we will have to use `{{% version` which would be a bit confusing.

ellistarn: ahhh gotcha.

ellistarn (approved):

PR #1477, opened 2022-03-07
Windows support

khoa-hatch:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/1131

- add windows operating system support
- recognize aws vpc private ipv4 address resources requests/limits
- add windows amiFamily

- Functional Tests
- Manual tests on EKS:
   - Enable IAM permissions so [vpc-resource-controller](https://github.com/aws/amazon-vpc-resource-controller-k8s) run and configure vpc-cni to enable windows ipam
   - Create windows example deployment:
```
    spec:
      containers:
      - name: windows-server-iis
        image: mcr.microsoft.com/windows/servercore:ltsc2019
        resources:
          requests:
            cpu: 1000m
            memory: 2Gi
        command:
        - powershell.exe
        - -command
        - "Add-WindowsFeature Web-Server; Invoke-WebRequest -UseBasicParsing -Uri 'https://dotnetbinaries.blob.core.windows.net/servicemonitor/2.0.1.6/ServiceMonitor.exe' -OutFile 'C:\\ServiceMonitor.exe'; echo '<html><body><br/><br/><marquee><H1>Hello EKS!!!<H1><marquee></body><html>' > C:\\inetpub\\wwwroot\\default.html; C:\\ServiceMonitor.exe 'w3svc'; "
      nodeSelector:
        kubernetes.io/os: windows
        kubernetes.io/arch: amd64
```
   - Tested a provisioner with custom launch template using windows ami -> windows instances can join cluster, pod is scheduled
   - Tested a provisioner with amiFamily: Windows -> launchtemplate for windows instances created, can join cluster, pod is scheduled

- [x] No

### DIFF
--- a/pkg/cloudprovider/aws/instancetype.go
+++ b/pkg/cloudprovider/aws/instancetype.go
@@ -16,6 +16,7 @@ package aws
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/aws/amazon-vpc-resource-controller-k8s/pkg/aws/vpc"
 	"github.com/aws/aws-sdk-go/aws"
@@ -27,6 +28,7 @@ import (
 
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
+	"github.com/aws/karpenter/pkg/utils/functional"
 	"github.com/aws/karpenter/pkg/utils/resources"
 )
 
@@ -48,7 +50,16 @@ func (i *InstanceType) Offerings() []cloudprovider.Offering {
 }
 
 func (i *InstanceType) OperatingSystems() sets.String {
-	return sets.NewString("linux")
+	// https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html#windows-support-prerequisites
+	if functional.HasAnyPrefix(i.Name(), "c3", "c4", "d2", "i2", "m6a", "r3") {
+		return sets.NewString("linux")
+	}
+
+	if strings.HasPrefix(i.Name(), "m4") && i.Name() != "m4.16xlarge" {
+		return sets.NewString("linux")
+	}
+
+	return sets.NewString("linux", "windows")
 }
 
 func (i *InstanceType) Architecture() string {
@@ -84,6 +95,15 @@ func (i *InstanceType) AWSPodENI() *resource.Quantity {
 	limits, ok := vpc.Limits[aws.StringValue(i.InstanceType)]
 	if ok && limits.IsTrunkingCompatible {
 		return resources.Quantity(fmt.Sprint(limits.BranchInterface))
+	}
+	return resources.Quantity("0")
+}
+
+func (i *InstanceType) AWSPodPrivateIPv4() *resource.Quantity {
+	// https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html#windows-support-prerequisites
+	limits, ok := vpc.Limits[aws.StringValue(i.InstanceType)]
+	if ok {
+		return resources.Quantity(fmt.Sprint(limits.IPv4PerInterface - 1))
 	}
 	return resources.Quantity("0")
 }
--- a/pkg/apis/provisioning/v1alpha5/labels.go
+++ b/pkg/apis/provisioning/v1alpha5/labels.go
@@ -23,9 +23,10 @@ import (
 )
 
 var (
-	ArchitectureAmd64    = "amd64"
-	ArchitectureArm64    = "arm64"
-	OperatingSystemLinux = "linux"
+	ArchitectureAmd64      = "amd64"
+	ArchitectureArm64      = "arm64"
+	OperatingSystemLinux   = "linux"
+	OperatingSystemWindows = "windows"
 
 	// Karpenter specific domains and labels
 	KarpenterLabelDomain = "karpenter.sh"
--- a/pkg/cloudprovider/aws/amifamily/bootstrap/bootstrap.go
+++ b/pkg/cloudprovider/aws/amifamily/bootstrap/bootstrap.go
@@ -15,6 +15,10 @@ limitations under the License.
 package bootstrap
 
 import (
+	"fmt"
+	"strings"
+	"sync"
+
 	core "k8s.io/api/core/v1"
 
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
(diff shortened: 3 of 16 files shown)


halovanic:
Fellow Windows container user here: I noticed you set the AMI alias to Windows Server Core 20H2.  Per [Container Version Compatibility](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility), Windows users need to match a container base to their OS version in a way Linux users generally don't have to.  The [Windows Server release information](https://docs.microsoft.com/en-us/windows-server/get-started/windows-server-release-info) indicates that Microsoft has consolidated on the LTSC releases only, so 20H2 seems effectively deprecated at this point. I'd suggest making this "Windows-2019" as the most viable release currently available on EKS (matching your manual test) and to avoid future unintentionally inflicted upgrades that break Windows workloads.

ellistarn:
Apologies for the slow response to this. We're gonna get to it soon.

MrBones757:
@ellistarn We were just about to get around to looking at this based on previous discussion, if there is any validation / extension needed we are happy to help here. Will eagerly await this release and get into testing in our env!

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -48,7 +50,16 @@ func (i *InstanceType) Offerings() []cloudprovider.Offering {
 }
 
 func (i *InstanceType) OperatingSystems() sets.String {
-	return sets.NewString("linux")
+	// https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html#windows-support-prerequisites
+	if functional.HasAnyPrefix(i.Name(), "c3", "c4", "d2", "i2", "m6a", "r3") {
```
MrBones757, line 54: I think there may need to be some additional filtering done here...
1 ) mac* should probably cause errors in general if this is not already catered for
2) any of the graviton series instances (arm) should probably be hard-coded to linux in the same way until ARM support is around (https://eksctl.io/usage/arm-support/) unless there's something i'm unaware of, so this would be t4g, m6g, a1, c7g, c6g, c6gn, r6g, x2gd, im4gn, is4gen and others if i missed them

halovanic:
@MrBones757, The [Service Channels Comparison](https://docs.microsoft.com/en-us/windows-server/get-started/servicing-channels-comparison#semi-annual-channel) makes it clear there are no future SAC Server releases after 20H2.

As far as Windows Server container choices, it would seem to be:

- 2019 Core
- 20H2 Core (ends support August 2022)
- 2022 Core (not supported yet for EKS)

I do agree that a choice of Windows version should be accounted for here; the transition between versions is not seamless for workloads so generic "Windows" as an alias could be problematic.

khoa-hatch:
> @khoa-hatch I also just noticed you changed the AMI lookup to the "Full" variant from "Core". IMO, Core was correct for both 2019 and 20H2.

I will make 2 windows amiFamily 
- WindowsFull
- WindowsCore
both use version 2019 for now since

halovanic:
I would have to defer to Karpenter owners on how many AMI families feels like too many for Windows on if supporting Core and Full feels warranted.  But I would prefer to include "2019" in those two AMI family names; the user has to be aware of which major version they are provisioning if we are to support any kind of cutover planning as 2019 is rolled off to 2022 and beyond.

bwagner5:
> I would have to defer to Karpenter owners on how many AMI families feels like too many for Windows on if supporting Core and Full feels warranted. But I would prefer to include "2019" in those two AMI family names; the user has to be aware of which major version they are provisioning if we are to support any kind of cutover planning as 2019 is rolled off to 2022 and beyond.

We need to address the AMI Family versioning problem anyways, like brought up in https://github.com/aws/karpenter/issues/1495 so I'm not sure the `amiFamily` field should have the version appended. Maybe it's in a different field that we can default to the latest. 

We need to think about the user experience in the Provisioner Spec on AMIs more generally. We'd like to support AMI Family w/ version which are opinionated defaults on how to setup a node, AMI Aliases (SSM query parameter to lookup an AMI ID specific to an account), and AMI IDs.  


Also, I did a large refactor of the AMI Family logic in the AWS Cloud Provider https://github.com/aws/karpenter/pull/1420 which has caused some conflicts in your branch. Apologies on the timing of that merge, but I think it will be simpler to implement Windows support with that change and I'm happy to help with any questions you have on the new structure.

halovanic:
Thanks for that insight into how this will evolve. With choices like versioning being considered elsewhere, it does sound like a single "Windows" family targeting Server Core 2019 is best for now:
https://cloudblogs.microsoft.com/windowsserver/2018/07/05/server-core-and-server-with-desktop-which-one-is-best-for-you/

khoa-hatch:
@bwagner5 updated to new structure

yi-fan-song:
@khoa-hatch @bwagner5 any updates on this? 

I'm noticing a lot of weird merge conflicts, if we can get this merged sometime soon, I'd be happy to go fix the merge conflicts.

bwagner5:
@yi-fan-song Apologies on the delay in reviewing this PR! I will try to get to it as soon as I can!  No need to worry about the merge conflicts, when I review I can resolve them and push up some commits. The codebase is changing fast, so I know it's hard to keep up on the conflicts.

marciogmorales:
Hello,
We should support both Server Core and Full. 

I see this return:

```
// SSMAlias returns the AMI Alias to query SSM
func (w Windows) SSMAlias(version string, _ cloudprovider.InstanceType) string {
	return fmt.Sprintf("/aws/service/ami-windows-latest/Windows_Server-2019-English-Core-EKS_Optimized-%s/image_id", version)
}
```

Lets add Full as well:

`/aws/service/ami-windows-latest/Windows_Server-2019-English-Full-EKS_Optimized-%s/image_id`

bwagner5:
@marciogmorales what's the use-cases for why a user would want to use full vs core?  I guess my question boils down to whether full and core should be modeled as an AMI family or a version and keep amiFamily as `windows`

bwagner5:
Hey! I carried over your commits to my branch, rebased, fixed conflicts, and fixed a few odds and ends. I've pushed a new PR https://github.com/aws/karpenter/pull/1774 to push this forward!  I need to do some more testing on that branch, but it seems to be in a good state currently.  I think next steps is to figure out AMIFamily versioning since that will be particularly important for Windows.  I haven't add Full support yet, I think we'll want to just do Core first, and then potentially add Full later.  I appreciate everyones patience on this PR :D

PR #1483, opened 2022-03-08
set a grace period of zero when cleaning up between tests

tzneal:
**1. Issue, if available:**
N/A

Remove grace period when cleaning up between tests.

Hard-coded a pod name in two different tests, without this change the second test fails due to 'object is being deleted'.  With this change, both tests pass.

- [X] No


bwagner5 (approved):
lgtm!

PR #1484, opened 2022-03-08
add pod affinity/anti-affinity tests that are skip()'d

tzneal:
**1. Issue, if available:**
N/A

This sets up some groundwork for beginning work on pod affinity/anti-affinity.

Unit tests

- [X] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/suite_test.go
+++ b/pkg/apis/provisioning/v1alpha5/suite_test.go
@@ -191,7 +191,7 @@ var _ = Describe("Validation", func() {
 		It("should fail because DoesNotExists conflicting", func() {
 			for _, op := range []v1.NodeSelectorOperator{v1.NodeSelectorOpIn, v1.NodeSelectorOpNotIn, v1.NodeSelectorOpExists} {
 				provisioner.Spec.Requirements = NewRequirements(
-					v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOperator(op), Values: []string{"test"}},
+					v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: op, Values: []string{"test"}},
 					v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpDoesNotExist},
 				)
 			}
--- a/pkg/cloudprovider/aws/suite_test.go
+++ b/pkg/cloudprovider/aws/suite_test.go
@@ -207,6 +207,48 @@ var _ = Describe("Allocation", func() {
 				}
 				Expect(nodeNames.Len()).To(Equal(2))
 			})
+			It("should not schedule a non-GPU workload on a node w/GPU", func() {
+				nodeNames := sets.NewString()
+				for _, pod := range ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner,
+					test.UnschedulablePod(test.PodOptions{
+						ResourceRequirements: v1.ResourceRequirements{
+							Requests: v1.ResourceList{
+								resources.NvidiaGPU: resource.MustParse("1"),
+								"cpu":               resource.MustParse("31"),
+							},
+							Limits: v1.ResourceList{
+								resources.NvidiaGPU: resource.MustParse("1"),
+								"cpu":               resource.MustParse("31"),
+							},
+						},
+					}),
+					// Can't pack onto the same instance due to consuming too much CPU
+					test.UnschedulablePod(test.PodOptions{
+						ResourceRequirements: v1.ResourceRequirements{
+							Requests: v1.ResourceList{
+								"cpu": resource.MustParse("3"),
+							},
+							Limits: v1.ResourceList{
+								"cpu": resource.MustParse("3"),
+							},
+						},
+					}),
+				) {
+					node := ExpectScheduled(ctx, env.Client, pod)
+
+					// This test has a GPU workload that nearly maxes out the test instance type.  It's intended to ensure
+					// that the second pod won't get a GPU node since it doesn't require one, even though it's compatible
+					// with the first pod that does require a GPU.
+					if _, isGpuPod := pod.Spec.Containers[0].Resources.Requests[resources.NvidiaGPU]; isGpuPod {
+						Expect(node.Labels).To(HaveKeyWithValue(v1.LabelInstanceTypeStable, "p3.8xlarge"))
+					} else {
+						Expect(node.Labels).ToNot(HaveKeyWithValue(v1.LabelInstanceTypeStable, "p3.8xlarge"))
+					}
+					nodeNames.Insert(node.Name)
+				}
+				Expect(nodeNames.Len()).To(Equal(2))
+			})
+
 			It("should launch instances for AWS Neuron resource requests", func() {
 				nodeNames := sets.NewString()
 				for _, pod := range ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner,
@@ -280,8 +322,8 @@ var _ = Describe("Allocation", func() {
 				pod.Spec.Affinity = &v1.Affinity{NodeAffinity: &v1.NodeAffinity{PreferredDuringSchedulingIgnoredDuringExecution: []v1.PreferredSchedulingTerm{
 					{
 						Weight: 1, Preference: v1.NodeSelectorTerm{MatchExpressions: []v1.NodeSelectorRequirement{
-							{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpIn, Values: []string{"test-zone-1a"}},
-						}},
+						{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpIn, Values: []string{"test-zone-1a"}},
+					}},
 					},
 				}}}
 				pod = ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner, pod)[0]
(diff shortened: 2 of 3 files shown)


pkg/apis/provisioning/v1alpha5/suite_test.go
```diff
@@ -191,7 +191,7 @@ var _ = Describe("Validation", func() {
 		It("should fail because DoesNotExists conflicting", func() {
 			for _, op := range []v1.NodeSelectorOperator{v1.NodeSelectorOpIn, v1.NodeSelectorOpNotIn, v1.NodeSelectorOpExists} {
 				provisioner.Spec.Requirements = NewRequirements(
-					v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOperator(op), Values: []string{"test"}},
+					v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: op, Values: []string{"test"}},
```
ellistarn, line 194: wow good catch

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -207,6 +207,29 @@ var _ = Describe("Allocation", func() {
 				}
 				Expect(nodeNames.Len()).To(Equal(2))
 			})
+			It("should not combine GPU and non-GPU workloads", func() {
```
ellistarn, line 210: I don't think I agree with this test. We definitely want to binpack pods into spare CPU/Mem on GPU nodes. Further, we don't even want to be aware of what a GPU is, as there are many unknown custom resource providers (vendor specific asic, etc) If customers want to prevent this, they can configured taints and tolerations accordingly.

tzneal: I don't like the test either, but my concern was:

Pod A - needs a GPU
Pod B - doesn't need a GPU

The binpacker can't fit Pod A and B onto the same node, so it launches two nodes with GPUs to support them.  The end result is a node with a GPU when a potentially cheaper node would have sufficed.

ellistarn: Agree that we want to catch that binpacking case, but I that's not what were testing here. We want to test that two pods, one which requires a GPU, doesn't result in two GPUs. If they can fit together on a single gpu node's resources, they should.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -502,6 +507,53 @@ var _ = Describe("Topology", func() {
...
+				test.UnschedulablePod(test.PodOptions{ObjectMeta: metav1.ObjectMeta{Labels: labels}, TopologySpreadConstraints: topology}))
+			ExpectSkew(ctx, env.Client, "default", &topology[0]).To(ConsistOf(1))
+
+			provisioner.Spec.Requirements = v1alpha5.NewRequirements(
+				v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpIn, Values: []string{"test-zone-2", "test-zone-3"}})
+			ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner,
+				MakePods(10, test.PodOptions{ObjectMeta: metav1.ObjectMeta{Labels: labels}, TopologySpreadConstraints: topology})...,
+			)
+
+			// max skew of 1, so test-zone-2/3 will have 2 nodes each and the rest of the pods will fail to schedule
```
ellistarn, line 554:
```suggestion
			// max skew of 1, so test-zone-2/3 will have 2 nodes each and the rest of the pods will schedule anyways
```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -543,6 +595,30 @@ var _ = Describe("Topology", func() {
 			)
 			ExpectSkew(ctx, env.Client, "default", &topology[0]).To(ConsistOf(1))
 		})
+		It("should handle a non-self selecting topology", func() {
```
ellistarn, line 598:
```suggestion
		It("should handle a interdependent selectors", func() {
```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -543,6 +595,30 @@ var _ = Describe("Topology", func() {
...
+				MakePods(5, test.PodOptions{TopologySpreadConstraints: topology})...,
+			)
+			// This is weird, but the topology label selector is used for determining domain counts. The pod that
+			// owns the topology is what the spread actually applies to.  In this test case, there are no pods matching
+			// the label selector, so the max skew is zero.  This means we can pack all the pods onto the same node since
+			// it doesn't violate the topology spread constraint (i.e. adding new pods doesn't increase skew since the
+			// pods we are adding don't count toward skew). This behavior is called out at
+			// https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/ , though it's not
+			// recommended for users.
+			nodeNames := map[string]struct{}{}
```
ellistarn, line 616: consider using a set.String

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -675,6 +799,515 @@ var _ = Describe("Topology", func() {
...
+			Skip("enable after pod-affinity is finished")
+			ExpectCreated(ctx, env.Client)
+			pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner, test.UnschedulablePod(test.PodOptions{
+				PodRequirements:     []v1.PodAffinityTerm{},
+				PodAntiRequirements: []v1.PodAffinityTerm{},
+			}))[0]
+			ExpectScheduled(ctx, env.Client, pod)
+		})
+		It("should respect pod affinity", func() {
+			Skip("enable after pod-affinity is finished")
```
ellistarn, line 814: Can we put the skip directive at the context level?

pkg/cloudprovider/aws/suite_test.go
tzneal: Reworked to test the intent

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -207,6 +207,48 @@ var _ = Describe("Allocation", func() {
 				}
 				Expect(nodeNames.Len()).To(Equal(2))
 			})
+			It("should not schedule a non-GPU workload on a node w/GPU", func() {
```
ellistarn, line 210: Is there any way we can make this a vendor neutral test?

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -761,6 +1394,63 @@ var _ = Describe("Taints", func() {
 	})
 })
 
+var _ = Describe("Incompatible pod networking", func() {
```
ellistarn, line 1397:
```suggestion
var _ = Describe("Networking constraints", func() {
```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -761,6 +1394,63 @@ var _ = Describe("Taints", func() {
...
+	Context("container HostPort usage", func() {
```
ellistarn, line 1398:
```suggestion
	Context("HostPort", func() {
```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -761,6 +1394,63 @@ var _ = Describe("Taints", func() {
...
+		It("shouldn't co-locate pods that use the same HostPort and protocol", func() {
+			Skip("enable after scheduler is aware of hostport usage")
```
ellistarn, line 1400: similar re the higher level skip.

pkg/controllers/provisioning/scheduling/suite_test.go
tzneal: No, there was a feature request for that feature with ginkgo but it wasn't implemented.

pkg/cloudprovider/aws/suite_test.go
tzneal: Once GPU support itself is vendor neutral, I think this will come along with that.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -675,6 +799,515 @@ var _ = Describe("Topology", func() {
...
+			}
+
+			// should have one node
+			Expect(nodeNames).To(HaveLen(1))
+			// with two pods
+			for _, podCount := range nodeNames {
+				Expect(podCount).To(Equal(2))
+			}
+		})
+		It("should handle multiple chained affinities", func() {
```
ellistarn, line 1169:
```suggestion
		It("should handle multiple dependent affinities", func() {
```

pkg/controllers/provisioning/scheduling/suite_test.go
tzneal: Done

pkg/controllers/provisioning/scheduling/suite_test.go
tzneal: Not supported in ginkgo

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -675,6 +799,515 @@ var _ = Describe("Topology", func() {
...
+					TopologyKey: v1.LabelTopologyZone,
+				}}})
+
+			affPods = append(affPods, affPod1)
+
+			ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner, affPods...)
+			top := &v1.TopologySpreadConstraint{TopologyKey: v1.LabelTopologyZone}
+			ExpectSkew(ctx, env.Client, "default", top).To(ConsistOf(11))
+		})
+		It("should support pod affinity with hostname topology and hostname topology spread", func() {
```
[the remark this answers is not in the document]
tzneal: sgtm, will need some future tests where it's treated as a soft preference instead

ellistarn (approved):

Issue #1433, opened 2022-03-01
Karpenter will not spin up node unless pod tolerates PreferNoSchedule taint


bwmetcalf:
## Version

**Karpenter**: v0.6.3

**Kubernetes**: v1.21

## Expected Behavior
We have a karpenter provisioner defined to attach a `PreferNoSchedule` taint to each node.  This provisioner should create a new node for pending pods even if they do not tolerate this taint as the pod can still be scheduled on these nodes as a last resort.

## Actual Behavior
The provisioner does not respond by creating a new node unless the pending pod tolerates the taint described above.

## Steps to Reproduce the Problem
Create a provisioner that adds a `PreferNoSchedule` taint.  Attempt to schedule a pod that does not tolerate this taint.

dewjam:
Hey @bwmetcalf ,
Thanks for reaching out.  Let me try to reproduce this in my test environment.  I'll post an update shortly.

dewjam:
@bwmetcalf I was able to reproduce this issue and have a rudimentary fix which will allow pods to schedule on nodes which have the PreferNoSchedule effect.  But I need to put some more thought into how to properly implement this capability.  

Is this issue causing any operational pain for you at the moment?  Is it preventing you from accommodating any use cases that can't be worked around by using a second Provisioner?

bwmetcalf:
It is currently not blocking us due to some other issues we are also encountering with https://github.com/kubernetes-sigs/descheduler.  Here is our use case; hopefully this helps.

We use karpenter to spin up additional capacity in response to pending generic pods that do not have a node selector or taints.  We do have provisioners specific to certain workloads like gitlab runners.  We also have the above descheduler installed using a `HighNodeUtilization` strategy with the goal of evicting pods from the generic karpenter nodes so that karpenter will see them as empty and delete them based on the `ttlSecondsAfterEmpty` value.  The reason we are adding a `PreferNoSchedule` taint to these generic karpenter nodes is so the default k8s scheduler will not schedule a pod on the same node from which it was evicted (or other generic karpenter nodes for that matter) unless that is a last resort.

The descheduler documentation indicates that the default k8s scheduler `LeastAllocated` strategy should be replaced with a `MostAllocated` strategy, but we want to avoid installing and maintaining a custom scheduler as there may be unintended consequences to doing so.

[...]

PR #1485, opened 2022-03-08
Relax PreferNoSchedule Taint

dewjam:
**1. Issue, if available:**
#1433

When taints with the effect of `PreferNoSchedule` are applied to nodes, the behavior of kube-scheduler is to only schedule pods to said nodes when no other scheduling options exist.  Currently, Karpenter treats taints with the effect of `PreferNoSchedule` as `NoSchedule` and will not provision a node with the `PreferNoSchedule` taint unless the pod tolerates it explicitly.  

This change adds a blanket `PreferNoSchedule` toleration to pods when no other Provisioners are selectable and after attempting to relax Affinity terms.  In short, Provisioners with `PreferNoSchedule` taints which would otherwise meet the pods requirements will be used, but only as a last resort.  

A test was implemented in the test suite.  Additionally, manual tests were executed with a combination of provisioners configured.

**Example using a matching provisioner (PreferNoSchedule toleration is not added to pod):**
```
2022-03-08T17:23:29.999Z	INFO	controller.provisioning	Batched 5 pods in 1.074480675s	{"commit": "774d5ed", "provisioner": "default"}
````

**Example where Affinity terms exist and are relaxed which then match a provisioner (PreferNoSchedule toleration is not added to pod):**
```
2022-03-08T18:06:52.002Z	DEBUG	controller.selection	Could not schedule pod, matched 0/2 provisioners, tried provisioner/restricted: did not tolerate fake=fake:NoSchedule; tried provisioner/zonal: incompatible requirements, [us-west-2d] not in [us-west-2a], key topology.kubernetes.io/zone	{"commit": "774d5ed", "pod": "default/inflate-7c98f5b68d-kd78b"}
2022-03-08T18:06:52.007Z	DEBUG	controller.selection	Relaxing soft constraints for pod since it previously failed to schedule, removing: spec.affinity.nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution[0]={"weight":1,"preference":{"matchExpressions":[{"key":"topology.kubernetes.io/zone","operator":"In","values":["us-west-2a"]}]}}	{"commit": "774d5ed", "pod": "default/inflate-7c98f5b68d-kd78b"}
2022-03-08T18:06:53.008Z	INFO	controller.provisioning	Batched 1 pods in 1.000311298s	{"commit": "774d5ed", "provisioner": "zonal"}
```

**Example where only matching provisioner has PreferNoSchedule taint (PreferNoSchedule toleration is added to pod):**
```
2022-03-08T17:31:32.796Z	DEBUG	controller.selection	Could not schedule pod, matched 0/2 provisioners, tried provisioner/prefer-no-schedule: did not tolerate fake=fake:PreferNoSchedule; tried provisioner/restricted: did not tolerate fake=fake:NoSchedule	{"commit": "774d5ed", "pod": "default/inflate-6b88c9fb68-t7cvm"}
2022-03-08T17:31:32.801Z	DEBUG	controller.selection	Relaxing soft constraints for pod since it previously failed to schedule, adding: toleration for PreferNoSchedule taints	{"commit": "774d5ed", "pod": "default/inflate-6b88c9fb68-t7cvm"}
2022-03-08T17:31:33.802Z	INFO	controller.provisioning	Batched 1 pods in 1.000636795s	{"commit": "774d5ed", "provisioner": "prefer-no-schedule"}
```

**Example where no provisioners match (PreferNoSchedule toleration is added to pod, but still no provisioners match):**
```
2022-03-08T17:36:28.079Z	DEBUG	controller.selection	Could not schedule pod, matched 0/1 provisioners, tried provisioner/restricted: did not tolerate fake=fake:NoSchedule	{"commit": "774d5ed", "pod": "default/inflate-6b88c9fb68-5jj8f"}
2022-03-08T17:36:28.085Z	DEBUG	controller.selection	Relaxing soft constraints for pod since it previously failed to schedule, adding: toleration for PreferNoSchedule taints	{"commit": "774d5ed", "pod": "default/inflate-6b88c9fb68-5jj8f"}
2022-03-08T17:36:28.085Z	DEBUG	controller.selection	Could not schedule pod, matched 0/1 provisioners, tried provisioner/restricted: did not tolerate fake=fake:NoSchedule	{"commit": "774d5ed", "pod": "default/inflate-6b88c9fb68-5jj8f"}
```

- [x] No

### DIFF
--- a/pkg/controllers/selection/preferences.go
+++ b/pkg/controllers/selection/preferences.go
@@ -49,16 +49,23 @@ func NewPreferences() *Preferences {
 // scheduled to that zone. Preferences are removed iteratively until only hard
 // constraints remain. Pods relaxation is reset (forgotten) after 5 minutes.
 func (p *Preferences) Relax(ctx context.Context, pod *v1.Pod) {
-	affinity, ok := p.cache.Get(string(pod.UID))
+	spec, ok := p.cache.Get(string(pod.UID))
 	// Add to cache if we've never seen it before
 	if !ok {
-		p.cache.SetDefault(string(pod.UID), pod.Spec.Affinity)
+		// Limit cached PodSpec to only required data
+		cachedSpec := v1.PodSpec{
+			Affinity:    pod.Spec.Affinity,
+			Tolerations: pod.Spec.Tolerations,
+		}
+		p.cache.SetDefault(string(pod.UID), cachedSpec)
 		return
 	}
 	// Attempt to relax the pod and update the cache
-	pod.Spec.Affinity = affinity.(*v1.Affinity)
+	cachedSpec := spec.(v1.PodSpec)
+	pod.Spec.Affinity = cachedSpec.Affinity
+	pod.Spec.Tolerations = cachedSpec.Tolerations
 	if relaxed := p.relax(ctx, pod); relaxed {
-		p.cache.SetDefault(string(pod.UID), pod.Spec.Affinity)
+		p.cache.SetDefault(string(pod.UID), pod.Spec)
 	}
 }
 
@@ -66,9 +73,10 @@ func (p *Preferences) relax(ctx context.Context, pod *v1.Pod) bool {
 	for _, relaxFunc := range []func(*v1.Pod) *string{
 		func(pod *v1.Pod) *string { return p.removePreferredNodeAffinityTerm(pod) },
 		func(pod *v1.Pod) *string { return p.removeRequiredNodeAffinityTerm(pod) },
+		func(pod *v1.Pod) *string { return p.toleratePreferNoScheduleTaints(pod) },
 	} {
 		if reason := relaxFunc(pod); reason != nil {
-			logging.FromContext(ctx).Debugf("Relaxing soft constraints for %s/%s since it previously failed to schedule, removing: %s", pod.Namespace, pod.Name, ptr.StringValue(reason))
+			logging.FromContext(ctx).Debugf("Relaxing soft constraints for pod since it previously failed to schedule, %s", ptr.StringValue(reason))
 			return true
 		}
 	}
@@ -85,7 +93,7 @@ func (p *Preferences) removePreferredNodeAffinityTerm(pod *v1.Pod) *string {
 		// Sort descending by weight to remove heaviest preferences to try lighter ones
 		sort.SliceStable(terms, func(i, j int) bool { return terms[i].Weight > terms[j].Weight })
 		pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = terms[1:]
-		return ptr.String(fmt.Sprintf("spec.affinity.nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution[0]=%s", pretty.Concise(terms[0])))
+		return ptr.String(fmt.Sprintf("removing: spec.affinity.nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution[0]=%s", pretty.Concise(terms[0])))
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/controllers/selection/preferences.go [hunk 1]
```diff
@@ -49,26 +49,34 @@ func NewPreferences() *Preferences {
...
-		p.cache.SetDefault(string(pod.UID), pod.Spec.Affinity)
+		p.cache.SetDefault(string(pod.UID), pod.Spec)
 	}
 }
 
 func (p *Preferences) relax(ctx context.Context, pod *v1.Pod) bool {
 	for _, relaxFunc := range []func(*v1.Pod) *string{
 		func(pod *v1.Pod) *string { return p.removePreferredNodeAffinityTerm(pod) },
 		func(pod *v1.Pod) *string { return p.removeRequiredNodeAffinityTerm(pod) },
+		func(pod *v1.Pod) *string { return p.toleratePreferNoScheduleTaints(pod) },
```
dewjam, line 76: Just thought of a case where this approach may produce a less than ideal outcome.  A Provisioner is able to satisfy the pods Node Affinity terms, but the pod does not tolerate the `PreferNoSchedule` taint.  In this case we would strip the node affinity terms unnecessarily.

pkg/controllers/selection/preferences.go
```diff
@@ -49,26 +49,34 @@ func NewPreferences() *Preferences {
...
+	spec, ok := p.cache.Get(string(pod.UID))
 	// Add to cache if we've never seen it before
 	if !ok {
-		p.cache.SetDefault(string(pod.UID), pod.Spec.Affinity)
+		// Limit cached PodSpec to only required data
+		cachedSpec := v1.PodSpec{
+			Affinity:    pod.Spec.Affinity,
+			Tolerations: pod.Spec.Tolerations,
+		}
+		p.cache.SetDefault(string(pod.UID), cachedSpec)
```
ellistarn, line 60: There are lots of pointers in this nested struct. How does equality checking work?

ellistarn: What if two pods have the same spec? I'm not sure we want to relax both of them.

pkg/controllers/selection/preferences.go
ellistarn: This is a necessary bit of complexity with how relaxation works. I'm open minded about randomizing, or doing a well known order. In practice, it's rare to have so many constraints.

pkg/controllers/selection/preferences.go
```diff
@@ -49,26 +49,34 @@ func NewPreferences() *Preferences {
...
 
 func (p *Preferences) relax(ctx context.Context, pod *v1.Pod) bool {
 	for _, relaxFunc := range []func(*v1.Pod) *string{
 		func(pod *v1.Pod) *string { return p.removePreferredNodeAffinityTerm(pod) },
 		func(pod *v1.Pod) *string { return p.removeRequiredNodeAffinityTerm(pod) },
+		func(pod *v1.Pod) *string { return p.toleratePreferNoScheduleTaints(pod) },
 	} {
 		if reason := relaxFunc(pod); reason != nil {
-			logging.FromContext(ctx).Debugf("Relaxing soft constraints for %s/%s since it previously failed to schedule, removing: %s", pod.Namespace, pod.Name, ptr.StringValue(reason))
+			logging.FromContext(ctx).Debugf("Relaxing soft constraints for pod since it previously failed to schedule, %s", ptr.StringValue(reason))
```
ellistarn, line 79:
```suggestion
			logging.FromContext(ctx).Debugf("Relaxing soft constraints since pod previously failed to schedule, %s", ptr.StringValue(reason))
```

pkg/controllers/selection/preferences.go
```diff
@@ -101,7 +109,23 @@ func (p *Preferences) removeRequiredNodeAffinityTerm(pod *v1.Pod) *string {
 	// Remove the first term if there's more than one (terms are an OR semantic), Unlike preferred affinity, we cannot remove all terms
 	if len(terms) > 1 {
 		pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms = terms[1:]
-		return ptr.String(fmt.Sprintf("spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution[0]=%s", pretty.Concise(terms[0])))
+		return ptr.String(fmt.Sprintf("removing: spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution[0]=%s", pretty.Concise(terms[0])))
 	}
 	return nil
 }
+
+func (p *Preferences) toleratePreferNoScheduleTaints(pod *v1.Pod) *string {
```
ellistarn, line 117: Thoughts on making these all return error? `error` is more or less a nillable string, and I think it expresses the semantic more cleanly.

ellistarn (commented):
Nice fix! A couple comments. 

Also -- does this require docs updates?

pkg/controllers/selection/preferences.go
dewjam: > There are lots of pointers in this nested struct. How does equality checking work?

I agree using an empty PodSpec is probably not ideal.  Lots of empty fields there.  I'll define a custom type instead.

> What if two pods have the same spec? I'm not sure we want to relax both of them. This could potentially cause over-relaxation if we see 10 identical pods all at once.

I'm not totally sure I follow.  The pod.UID should always be unique, but maybe that's not what you're getting at.

pkg/controllers/selection/preferences.go
dewjam: Thinking about this further, are you suggesting that we return `error` instead of `*string`?  Or in addition to `*string`?

ellistarn: Instead of *string.

ellistarn (approved):

PR #1489, opened 2022-03-09
remove the specialness from GPU requests

tzneal:
**1. Issue, if available:**

N/A

This modifies Karpenter to treat resource requests identically. It
removes the special logic for some AWS specific GPU types and
allows CloudProvider implementers to provide their own custom
resource types along with an ordering to be used as a hint for
binpacking regarding which instance types to prefer.

Unit tests & deploying GPU/non-GPU workloads.  Quota limits prevented GPU instances from being
created, but I could see the requests.

- [X] No


pkg/controllers/provisioning/binpacking/packable.go
```diff
@@ -203,52 +186,41 @@ func (p *Packable) validateOfferings(constraints *v1alpha5.Constraints) error {
...
-func (p *Packable) validateGPUs(pods []*v1.Pod) error {
-	gpuResources := map[v1.ResourceName]*resource.Quantity{
-		resources.NvidiaGPU: p.InstanceType.NvidiaGPUs(),
-		resources.AMDGPU:    p.InstanceType.AMDGPUs(),
-		resources.AWSNeuron: p.InstanceType.AWSNeurons(),
-	}
-	for resourceName, instanceTypeResourceQuantity := range gpuResources {
-		if p.requiresResource(pods, resourceName) && instanceTypeResourceQuantity.IsZero() {
-			return fmt.Errorf("%s is required", resourceName)
-		} else if !p.requiresResource(pods, resourceName) && !instanceTypeResourceQuantity.IsZero() {
```
tzneal, line 215: This is a change from previous behavior.  For the three known GPU types, it would mark instance types that had those GPUs as incompatible with pods that didn't require them.  

The change I've made leaves them as compatible since you can run a pod that doesn't require resource XYZ on an instance type that has resource XYZ.

pkg/controllers/provisioning/binpacking/packer.go
```diff
@@ -180,7 +180,16 @@ func (p *Packer) packWithLargestPod(unpackedPods []*v1.Pod, packables []*Packabl
...
 				// Need to exclude instance type with smaller memory and fewer pods
-				if packables[i].Memory().Cmp(*packables[j].Memory()) <= 0 && packables[i].Pods().Cmp(*packables[j].Pods()) <= 0 {
+				lv := packables[i]
+				rv := packables[j]
+
+				lvMem := lv.Resources()[v1.ResourceMemory]
+				rvMem := rv.Resources()[v1.ResourceMemory]
+
+				lvPods := lv.Resources()[v1.ResourcePods]
+				rvPods := rv.Resources()[v1.ResourcePods]
```
tzneal, line 190: Some of this awkwardness about assigning to a temp value comes from Cmp() requiring a pointer receiver, so you can't use it on the item directly from a map (e.g. lv.Resources()[v1.ResourceMemory].Cmp(...) )

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -458,18 +497,29 @@ var _ = Describe("Allocation", func() {
...
+				ltNames.Insert(*fakeEC2API.CalledWithCreateLaunchTemplateInput.Pop().(*ec2.CreateLaunchTemplateInput).LaunchTemplateName)
+
 				Expect(fakeEC2API.CalledWithCreateFleetInput.Cardinality()).To(Equal(1))
+
 				createFleetInput := fakeEC2API.CalledWithCreateFleetInput.Pop().(*ec2.CreateFleetInput)
-				Expect(createFleetInput.LaunchTemplateConfigs).To(HaveLen(1))
 				launchTemplate := createFleetInput.LaunchTemplateConfigs[0].LaunchTemplateSpecification
-				Expect(*launchTemplate.LaunchTemplateName).To(Equal(*createLaunchTemplateInput.LaunchTemplateName))
-				Expect(createLaunchTemplateInput.LaunchTemplateData.BlockDeviceMappings[0].Ebs.Encrypted).To(Equal(aws.Bool(true)))
+				Expect(createFleetInput.LaunchTemplateConfigs).To(HaveLen(2))
```
tzneal, line 515: The test changes come since it creates two launch templates, one for non-GPU and one for GPU instances when the workload doesn't require a GPU.  I *think* this works with CreateFleet since we tell it to only launch one instance, but I'm very interested in feedback on this behavior.

pkg/cloudprovider/types.go
```diff
@@ -44,6 +43,11 @@ type CloudProvider interface {
 	Validate(context.Context, *v1alpha5.Constraints) *apis.FieldError
 	// Name returns the CloudProvider implementation name.
 	Name() string
+	// ResourcePriorities provides information to bin-packing regarding the sort priority of resources. If CPU is before memory
+	// in this list, the bin-packer will prefer to schedule on nodes with less CPU than memory.  This should always include at
+	// least v1.ResourceMemory and v1.ResourceCPU, however it also is suggested to include any specific resource types
+	// (e.g GPUs) ordered before CPU and memory.
+	ResourcePriorities() []v1.ResourceName
```
tzneal, line 50: This is used to provide an orderings for:

-  {cpu: 2, mem: 8GB, nvidia: 1} <  {cpu: 2, mem: 8GB, amd: 1} # so you can define a consistent ordering for disparate resource types
- {cpu: 8, mem: 32Gb} < { cpu:1 mem: 1GB, nvidia: 1} # prefer instance types with more cpu/memory over those with special resources

pkg/cloudprovider/aws/suite_test.go
bwagner5: the LTs are merged and the allocation strategy is used. This will probably work 99% of the time. There could be cases for spot where a GPU instance could be provisioned since we are using the capacity-optimized allocation strategy. It would be extremely rare though.

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -60,35 +60,47 @@ func (i *InstanceType) Architecture() string {
 	return fmt.Sprint(aws.StringValueSlice(i.ProcessorInfo.SupportedArchitectures)) // Unrecognized, but used for error printing
 }
 
-func (i *InstanceType) CPU() *resource.Quantity {
-	return resources.Quantity(fmt.Sprint(*i.VCpuInfo.DefaultVCpus))
+func (i *InstanceType) Resources() v1.ResourceList {
+	rl := v1.ResourceList{}
```
bwagner5, line 64: you can do these assignment within this block:

```
return v1.ResourceList{
  v1.ResourceCPU: i.cpu(),
  ...
}
```

pkg/cloudprovider/types.go
bwagner5: Could we expose this as a `Compare` func?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -103,7 +103,7 @@ func (s *Scheduler) getSchedules(ctx context.Context, constraints *v1alpha5.Cons
 			GPURequests v1.ResourceList
 		}{
 			Constraints: tightened,
-			GPURequests: resources.GPULimitsFor(pod),
+			GPURequests: resources.NonGenericResources(pod),
```
bwagner5, line 106: so this could be more than just `GPURequests` right?  I guess the thought is that it doesn't matter with the current scheduler implementation?  Can we change the name of this variable?

pkg/utils/resources/resources.go
```diff
@@ -20,10 +20,10 @@ import (
 )
 
 const (
-	NvidiaGPU = "nvidia.com/gpu"
-	AMDGPU    = "amd.com/gpu"
-	AWSNeuron = "aws.amazon.com/neuron"
-	AWSPodENI = "vpc.amazonaws.com/pod-eni"
+	NvidiaGPU v1.ResourceName = "nvidia.com/gpu"
+	AMDGPU    v1.ResourceName = "amd.com/gpu"
+	AWSNeuron v1.ResourceName = "aws.amazon.com/neuron"
```
bwagner5, line 25: I don't think the AWS specific resource names should be in this core utils pkg.  Can we move to an AWS Cloud Provider utils pkg? Or maybe just on the instanceType in the AWS CP?

bwagner5 (commented):
Nice! It is nice to get the AWS stuff out of the core code!

pkg/cloudprovider/aws/ami.go
```diff
@@ -98,7 +100,9 @@ func (p *AMIProvider) getSSMQuery(constraints *v1alpha1.Constraints, instanceTyp
 // getAL2Alias returns a properly-formatted alias for an Amazon Linux AMI from SSM
 func (p *AMIProvider) getAL2Alias(version string, instanceType cloudprovider.InstanceType) string {
 	amiSuffix := ""
-	if !instanceType.NvidiaGPUs().IsZero() || !instanceType.AWSNeurons().IsZero() {
+	nvidia := instanceType.Resources()[resources.NvidiaGPU]
```
ellistarn, line 103: consider inlining.

pkg/cloudprovider/types.go
ellistarn, line 50: I'm always nervous about expanding the cloud provider API, especially since i'm not sure this is the right semantic. For example, the resource `pod-eni` doesn't matter in terms of cost, we only need to make sure there is "enough". For GPU, you basically never want a GPU instance if your workload doesn't need one. Even if every other instance type is unavailable.

tzneal: I think there's got to be something to support cloud providers that may have new resource types that we aren't aware of.  Consider {cpu: 4, mem: 8, xyz: 1} and {cpu: 4, mem: 8, abc: 2}  , which instance type should we prefer?

Brandon suggested a comparison func, I was looking at something like this to allow the cloud provider to directly provide the comparison:
```
	// LessThan is used by bin-packing to give control to the CloudProvider over how instance types are
	// compared.  If two instance types can both meet the resource requests for a given demand, bin-packing will
	// prefer the 'lesser' one.
	LessThan(lhs, rhs InstanceType) bool
```

There probably does need to be some method for the cloud provider to let bin packing know about those hard preferences (e.g. don't consider GPU instance types for non-GPU workloads).  The old code was hard-coded for the three supported GPU types that were known.  With this PR, we would request both non-GPU and GPU instance types for any non-GPU workload and let `CreateFleet` give us whatever it prefers.

Maybe a compare with a broader return type to allow expressing the incompatibilities?
```
type InstanceCompareResult byte
const (
InstanceCompareResultIncompatible InstanceCompareResult = iota
InstanceCompareResultLessThan
InstanceCompareResultEqual
InstanceCompareResultGreaterThan
``

File: pkg/utils/resources/resources.go (hunk 7 above)
Comment (tzneal) [reply pkg/utils/resources/resources.go:25]: Good call, moving.

File: pkg/controllers/provisioning/scheduling/scheduler.go (hunk 6 above)
Comment (tzneal) [reply pkg/controllers/provisioning/scheduling/scheduler.go:106]: Fixed the name, I do like providing a comparison function better.  I suggested to Ellis below about a compare that provides both ordering & incompatibility information.

File: pkg/cloudprovider/types.go (hunk 4 above)
Comment (tzneal) [reply pkg/cloudprovider/types.go:50]: Actually that doesn't work if you compare instance types, but I think it may if you compare v1.ResourceList. You need to both be able to:
- order instance types
- identify if there is some reason why you would prefer not to use an instance type that can satisfy a pod's resource requests

File: pkg/cloudprovider/aws/ami.go (hunk 8 above)
Comment (ellistarn) [reply pkg/cloudprovider/aws/ami.go:103]: Bummer. Ok!

File: pkg/cloudprovider/types.go (hunk 4 above)
Comment (tzneal) [reply pkg/cloudprovider/types.go:50]: Or maybe something like?

```go
	// LessThan is used by bin-packing to give control to the CloudProvider over how instance types are
	// compared.  If two instance types can both meet the resource requests for a given demand, bin-packing will
	// prefer the 'lesser' one.
	LessThan(lhs, rhs InstanceType) bool

	// IsCompatible allows the cloud provider to prevent scheduling a pod against a specific instance type
        // even if it would otherwise normally be possible.  For example, this can be used to prevent pods that
        // don't require GPUs from scheduling against instances that have GPUs.
        IsCompatible(podResources v1.ResourceList, it InstanceType) bool

```

ellistarn: This is getting into sketchy territory. It's really increasing the onus of the cloud provider to make decisions about this. In some sense, this really does come down to cost. I wonder if we could solve this via a Price() parameter in offering. 
This gives the cloud provider control without building terse api surface.

I'd lean away from looking up the real cost, but we could simply use a heuristic like (1cpu=1, 1memory=1, 1gpu=1000). We could also use some sort of price threshold, where we don't launch an instance if the threshold isn't met.

tzneal: A single `Price(i InstanceType)` method on cloud provider, or a `Price()` method on the instance provides the ordering the bin packer needs, but doesn't solve the "Don't schedule a non-GPU Workload on a GPU instance type" problem. 

We'll have the non-GPU instance types sorted first in our list provided to the cloud provider `Create()` call.  There's still a chance that with `CreateFleet` and `ec2.SpotAllocationStrategyCapacityOptimizedPrioritized`, you'll get the GPU instance type. 

I suppose the CloudProvider itself could be responsible for further filtering (e.g. if we ask for non-GPU nodes and GPU nodes, the CloudProvider filters out the GPU types).

ellistarn: That's where the threshold comes in (e.g. 2-3x). You could design it such that GPUs are simply too expensive.

ellistarn: s/weary/wary :)

Agree about the threshold. Perhaps we could make the compare function be able to say "no this is not supported", rather than "prefer this over that".

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -58,6 +60,9 @@ const (
 
 func init() {
 	v1alpha5.NormalizedLabels = functional.UnionStringMaps(v1alpha5.NormalizedLabels, map[string]string{"topology.ebs.csi.aws.com/zone": v1.LabelTopologyZone})
```
ellistarn, line 62: Man it would be sweet to register NormalizedLabels in the same way!

pkg/cloudprovider/registry.go
```diff
@@ -0,0 +1,71 @@
...
+	// ResourceFlagScaleOut
+	// TODO(todd): implement once bin-packing and scheduling are combined
+
+	// ResourceFlagExotic indicates that instances of this resource should not be considered unless a pod specifically
+	// requests this resource type.  This is useful for preventing non-GPU workloads from possibly scheduling to more
+	// expensive GPU instance type, or from causing a GPU instance type to scale up to the next larger type.
+	ResourceFlagExotic
+)
+
+var mu sync.RWMutex
```
ellistarn, line 42: Does this need to be threadsafe?

pkg/cloudprovider/registry.go
```diff
@@ -0,0 +1,71 @@
...
+var resourceInfo = map[v1.ResourceName]ResourceFlags{}
```
ellistarn, line 43: Thoughts on making this public?

pkg/controllers/provisioning/binpacking/packer.go
```diff
@@ -180,7 +180,16 @@ func (p *Packer) packWithLargestPod(unpackedPods []*v1.Pod, packables []*Packabl
...
+
+				if lvMem.Cmp(rvMem) <= 0 && lvPods.Cmp(rvPods) <= 0 {
```
ellistarn, line 192: 🤔 This makes me realize that `.Price()` implicitly implements compare. You have full control over weights, so you can line up the instance types in whatever order.

pkg/cloudprovider/registry.go
tzneal: Not really, I was being paranoid.  Just need to document that cloud provider implementations should call it in their init() functions and that should prevent any issues.

pkg/cloudprovider/registry.go
tzneal: There may be a reason to change it some day? I don't think making it public provides much benefit, but it's a toss up either way.

pkg/cloudprovider/registry.go
ellistarn: More risky to be threadsafe than not, IMO, since programmers will assume there's a reason to use it in a multithreaded way.

pkg/cloudprovider/registry.go
ellistarn: It enables us to remove the setters/getters in favor of direct access. In my understanding, go style typically leans towards direct access rather than setters/getters unless they're doing something specific. In this case, you were wrapping with a mutex, but if we remove the thread safety, I don't see the value anymore.

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -60,66 +60,99 @@ func (i *InstanceType) Architecture() string {
...
+			if gpu.Count != nil {
+				gpuCount += float64(*gpu.Count)
+			}
+		}
+	}
+
+	// These are meant to give some relative weighting
+	return float64(*i.VCpuInfo.DefaultVCpus) +
+		float64(*i.MemoryInfo.SizeInMiB)/MbToGb +
+		5*gpuCount + 10*gpuMemoryMiB/MbToGb
```
ellistarn, line 94: Consider extracting these values to constants

var (
  GPUCostWeight
  CPUCostWeight
  MemoryGBCostWeight
)

pkg/cloudprovider/registry.go
tzneal: sgtm

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -60,66 +60,100 @@ func (i *InstanceType) Architecture() string {
...
+		v1alpha1.ResourceAWSNeuron: i.awsNeurons(),
+	}
+}
+
+func (i *InstanceType) Price() float64 {
+	const (
+		GPUCostWeight      = 5
+		CPUCostWeight      = 1
+		MemoryMBCostWeight = 1
+		MbToGb             = 1 / 1024.0
```
ellistarn, line 80: Thoughts on modeling this in MB?
```suggestion
		MemoryMBCostWeight = 1024
```

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -60,66 +60,100 @@ func (i *InstanceType) Architecture() string {
...
+	gpuCount := 0.0
+	if i.GpuInfo != nil {
+		for _, gpu := range i.GpuInfo.Gpus {
+			if gpu.Count != nil {
+				gpuCount += float64(*gpu.Count)
+			}
+		}
+	}
+
+	// These are meant to give some relative weighting
```
ellistarn, line 92: This comment makes me uneasy as a reader. They're meant to do X, but do they? Is it accurate to be more concrete?

pkg/cloudprovider/registry.go
```diff
@@ -0,0 +1,35 @@
...
+const (
+	// ResourceFlagNone is the default settings for resources where we normally scale up to larger nodes with no
+	// restrictions.
+	ResourceFlagNone ResourceFlags = iota
+
+	// ResourceFlagMinimizeUsage indicates that instances which have this resource should not be considered for node
+	// creation unless a pod specifically requests this resource type.  This is useful for preventing non-GPU workloads
+	// from possibly scheduling to more expensive GPU instance type, or from causing a GPU instance type to scale up to
+	// the next larger type due to a non-GPU workload
+	ResourceFlagMinimizeUsage
```
ellistarn, line 32: I keep thinking about this complexity. I'm wondering if it's enough to purely delineate on WellKnown vs NotWellKnown resource types? If we know about it (e.g., CPU, Memory, EphemeralStorage) then we can use the `ResourceFlagNone` behavior. If we don't know about it, we can use `ResourceFlagMinimizeUsage` behavior. This will enable us to simply not configure any of this (for now).

My guess is that we will need to be more sophisticated with binpacking in the future, but I don't think we have enough use cases to have clarity on what the parameters are. Given that with currently known resource types, the above suggestion is equivalent in implementation.

Happy to move forward in either direction.

pkg/cloudprovider/types.go
```diff
@@ -59,14 +58,14 @@ type InstanceType interface {
...
-	NvidiaGPUs() *resource.Quantity
-	AMDGPUs() *resource.Quantity
-	AWSNeurons() *resource.Quantity
-	AWSPodENI() *resource.Quantity
+	// Resources are the full allocatable resource capacities for this instance type
+	Resources() v1.ResourceList
+	// Overhead is the amount of resource overhead expected to be used by kubelet and any other system daemons outside
+	// of Kubernetes.
 	Overhead() v1.ResourceList
+	// Price is a metric that is used by Karpenter's scheduler to prefer some instances over others.  This can be an
```
ellistarn, line 66: I try to avoid using the name of a project inside a project. Not that we have any plans to change the name -- just as a best practice ;). 
```suggestion
	// Price is used to optimize pod placement onto nodes. This can be an
```

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -161,26 +168,15 @@ func (n Node) isPodCompatible(pod *v1.Pod, it cloudprovider.InstanceType) bool {
...
-			if it.AMDGPUs().Cmp(quantity) < 0 {
-				return false
-			}
-		case resources.AWSPodENI:
-			if it.AWSPodENI().Cmp(quantity) < 0 {
-				return false
-			}
-		default:
+		// we don't care if the pod is requesting zero quantity of some resource
+		if quantity.IsZero() {
```
ellistarn, line 172: Is this a real case we need to check? Does the API Server even allow this?

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -161,26 +168,15 @@ func (n Node) isPodCompatible(pod *v1.Pod, it cloudprovider.InstanceType) bool {
...
 			continue
 		}
+		// instance must have a non-zero quantity
+		instanceQuantity := it.Resources()[name]
```
ellistarn, line 176: consider inlining in the if statement

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -74,18 +76,103 @@ func (s *Scheduler) Solve(ctx context.Context, provisioner *v1alpha5.Provisioner
...
+				node.unschedulable = true
+			}
 		}
 	}
 	return nodes
 }
+
+type podSet struct {
+	pods     []*v1.Pod
+	isExotic bool
```
ellistarn, line 125: IIUC this abstraction is only necessary to get us over the hump until binpacking and scheduling are combined. FWIW, I think we could probably get away without it, given that the current behavior has this bug. I worry a bit about the complexity or debt that may be left behind from too many code transformations.

pkg/cloudprovider/aws/instancetype.go
tzneal: Maybe I should just remove that comment entirely.  Trying to indicate that it's not a real "price", but people should be able to figure that out easily enough.

pkg/controllers/provisioning/scheduling/node.go
tzneal: You're right, I tested and the API server just drops the zero quantity resource.

tzneal: Spoke too soon, API server does pass along an empty resource request.  I was running the wrong test. Maybe we still don't care though, I'm fine leaving or removing the check.

pkg/controllers/provisioning/scheduling/scheduler.go
tzneal: The issue we seem to keep bumping into with making resource handling more generic is that some resource types  are much more expensive relative to any other types which causes us to a) not want to use instance with those resources if possible, and b) pack pods that do use those resources onto the minimum number of nodes possible.

This code is support "b" where we don't want to create a GPU node for a pod, fill it up with non-GPU workloads, then need to create another GPU node in the same batch.  This problem would also exist when we do bin-packing and scheduling together as far as I can tell.

It's somewhat of a weird thing to worry about though since if the nodes/workloads aren't using tolerations/taints, kube-scheduler will happily schedule non-GPU workloads to the GPU nodes (or similarly to instance types with any other expensive resource in the future).  We're adding complexity to work around a problem that we can't solve outside the single batch of pods we're about to schedule. If workloads are using tolerations/taints to segregate GPU/non-GPU workloads, then part b doesn't exist for us any longer and we can easily handle part a in our AWS cloud provider implementation.

I'm almost convinced we can remove the special resource type registration and just improve the docs to more explicitly state that for expensive resource types you need to use taints/tolerations to prevent co-mingling workloads for both Karpenter and kube-scheduler.  We can then add in a bit of code to the AWS provider to filter out GPU instance types if they aren't required so that capacity optimized selection won't use them, and the problem seems solved.

pkg/cloudprovider/registry.go
tzneal: I left a longer comment below, but I'm starting to think this complexity isn't worth it since kube-scheduler won't follow this logic and we're left with using taints/tolerations again which are already built-in.

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -184,138 +183,63 @@ var _ = Describe("Allocation", func() {
...
 					})) {
 					node := ExpectScheduled(ctx, env.Client, pod)
 					Expect(node.Labels).To(HaveKeyWithValue(v1.LabelInstanceTypeStable, "p3.8xlarge"))
-					Expect(node.Status.Capacity).To(HaveKeyWithValue(nvidiaGPUResourceName, resource.MustParse("4")))
+					Expect(node.Status.Capacity).To(HaveKeyWithValue(v1alpha1.ResourceNVIDIAGPU, resource.MustParse("4")))
 					nodeNames.Insert(node.Name)
 				}
 				Expect(nodeNames.Len()).To(Equal(2))
 			})
-			It("should launch pods with Nvidia and Neuron resources on different instances", func() {
```
tzneal, line 218: This test was made generic and turned into "should launch pods with resources that aren't on any single instance type on different instances"

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -184,138 +183,63 @@ var _ = Describe("Allocation", func() {
...
-							Requests: v1.ResourceList{resources.AWSNeuron: resource.MustParse("1")},
-							Limits:   v1.ResourceList{resources.AWSNeuron: resource.MustParse("1")},
-						},
-					})) {
-					node := ExpectScheduled(ctx, env.Client, pod)
-					nodeNames.Insert(node.Name)
-				}
-				Expect(nodeNames.Len()).To(Equal(2))
-			})
-			It("should fail to schedule a pod with both Nvidia and Neuron resources requests", func() {
```
tzneal, line 239: Made generic into "should fail to schedule a pod with resources requests that aren't on a single instance type"

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -184,138 +183,63 @@ var _ = Describe("Allocation", func() {
...
-				pods := ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner,
-					test.UnschedulablePod(test.PodOptions{
-						ResourceRequirements: v1.ResourceRequirements{
-							Requests: v1.ResourceList{resources.NvidiaGPU: resource.MustParse("1"), resources.AWSNeuron: resource.MustParse("1")},
-							Limits:   v1.ResourceList{resources.NvidiaGPU: resource.MustParse("1"), resources.AWSNeuron: resource.MustParse("1")},
-						},
-					}))
-				ExpectNotScheduled(ctx, env.Client, pods[0])
-			})
-			It("should not schedule a non-GPU workload on a node w/GPU", func() {
```
tzneal, line 249: Removed this test entirely

pkg/controllers/provisioning/scheduling/scheduler.go
tzneal: I put a new commit on top that removes all the special registration, leaving special resource handling for taints/tolerations so that kube-scheduler will be aware as well.  This simplifies the code quite a bit and now the AWS provider just has a bit of special code to handle the non-GPU on a GPU instance type issue.

bwagner5 (dismissed):
nice! lgtm

pkg/utils/resources/resources.go
```diff
@@ -66,3 +59,15 @@ func Quantity(value string) *resource.Quantity {
 	r := resource.MustParse(value)
 	return &r
 }
+
+// Cmp implements lhs.Cmp(rhs). This method is provided to make some code a bit cleaner as the Quantity.Cmp takes
+// a pointer receiver and map index expressions aren't addressable, so it can't be called directly.
+func Cmp(lhs, rhs resource.Quantity) int {
```
ellistarn, line 65: This is no longer used.

ellistarn (dismissed):

PR #1499, opened 2022-03-11
Add nightly flaky test finder

spring1843:
**1. Issue, if available:**

This will run everyday to find flaky tests, those that only fail sometimes. See it running here: https://github.com/aws/karpenter/runs/5504952681?check_suite_focus=true

- [ ] No


.github/workflows/flaky-test.yml
```diff
@@ -0,0 +1,28 @@
+name: Flaky Test Finder
```
ellistarn, line 1: If you wanted to simplify this into a verb, I've seen this called `Deflake` or `--deflake` before.

ellistarn (approved):

.github/workflows/deflake.yml
```diff
@@ -0,0 +1,28 @@
+name: Deflake
+on:
+  schedule:
+    - cron: '22 18 * * 0'
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      K8S_VERSION: "1.21.x"
```
bwagner5, line 9: you can remove this env block and just take the default `K8S_VERSION` from the Makefile

Makefile
```diff
@@ -26,11 +26,17 @@ ci: toolchain verify licenses battletest ## Run all steps used by continuous int
...
-		-cover -coverprofile=coverage.out -outputdir=. -coverpkg=./pkg/... \
-		--randomizeAllSpecs --randomizeSuites -race
+			-cover -coverprofile=coverage.out -outputdir=. -coverpkg=./pkg/... \
+			--randomizeAllSpecs --randomizeSuites -race
+
+deflake:
+	for i in {1..10}; do make strongertests; done
+
+battletest: strongertests
 	go tool cover -html coverage.out -o coverage.html
```
bwagner5, line 40: why did this fail when it wasn't broken out into a separate target?

ellistarn (dismissed):

Makefile
spring1843: I think when you run it multiple times it corrupts the file or something like that you can see the outcome here https://github.com/aws/karpenter/runs/5504735463?check_suite_focus=true

```
cover: line "" doesn't match expected format: couldn't find a   before Count
make: *** [Makefile:32: battletest] Error 1
Error: Process completed with exit code 2.
```

Regardless we don't need the coverage conversion at every step because it will just add to the total time taken.

Makefile
```diff
@@ -26,11 +26,16 @@ ci: toolchain verify licenses battletest ## Run all steps used by continuous int
...
+strongertests:
 	# Run randomized, parallelized, racing, code coveraged, tests
 	ginkgo -r \
-		-cover -coverprofile=coverage.out -outputdir=. -coverpkg=./pkg/... \
-		--randomizeAllSpecs --randomizeSuites -race
+			-cover -coverprofile=coverage.out -outputdir=. -coverpkg=./pkg/... \
+			--randomizeAllSpecs --randomizeSuites -race
+
+deflake:
+	for i in {1..10}; do make strongertests; done
```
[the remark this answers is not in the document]
spring1843: Thanks for pointing that out it now [fails](https://github.com/aws/karpenter/runs/5518734002?check_suite_focus=true). 

> Since it's unlikely we'd run this locally,
We are likely to run it locally if it finds problems while running in the CI pipeline. Also a good idea to run it locally after changing parts that do concurrency. Which is why I moved it to the Makefile.

ellistarn (approved):

PR #1515, opened 2022-03-15
introduce theoretical node

tzneal:
**1. Issue, if available:**
N/A

Resolve issue related to pod compatibility leading to an impossible instance type construct

Unit tested + tested deployment scale-up/scale-down on EKS

- [X] No

### DIFF
--- a/pkg/cloudprovider/aws/suite_test.go
+++ b/pkg/cloudprovider/aws/suite_test.go
@@ -184,6 +184,7 @@ var _ = Describe("Allocation", func() {
 					Expect(supportsPodENI()).To(Equal(true))
 				}
 			})
+			// TODO(todd): this set of tests should move to scheduler once resource handling is made more generic
 			It("should launch instances for Nvidia GPU resource requests", func() {
 				nodeNames := sets.NewString()
 				for _, pod := range ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner,
@@ -213,6 +214,37 @@ var _ = Describe("Allocation", func() {
 					nodeNames.Insert(node.Name)
 				}
 				Expect(nodeNames.Len()).To(Equal(2))
+			})
+			It("should launch pods with Nvidia and Neuron resources on different instances", func() {
+				nodeNames := sets.NewString()
+				for _, pod := range ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner,
+					test.UnschedulablePod(test.PodOptions{
+						ResourceRequirements: v1.ResourceRequirements{
+							Requests: v1.ResourceList{resources.NvidiaGPU: resource.MustParse("1")},
+							Limits:   v1.ResourceList{resources.NvidiaGPU: resource.MustParse("1")},
+						},
+					}),
+					// Should pack onto a different instance since no type has both Nvidia and Neuron
+					test.UnschedulablePod(test.PodOptions{
+						ResourceRequirements: v1.ResourceRequirements{
+							Requests: v1.ResourceList{resources.AWSNeuron: resource.MustParse("1")},
+							Limits:   v1.ResourceList{resources.AWSNeuron: resource.MustParse("1")},
+						},
+					})) {
+					node := ExpectScheduled(ctx, env.Client, pod)
+					nodeNames.Insert(node.Name)
+				}
+				Expect(nodeNames.Len()).To(Equal(2))
+			})
+			It("should fail to schedule a pod with both Nvidia and Neuron resources requests", func() {
+				pods := ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner,
+					test.UnschedulablePod(test.PodOptions{
+						ResourceRequirements: v1.ResourceRequirements{
+							Requests: v1.ResourceList{resources.NvidiaGPU: resource.MustParse("1"), resources.AWSNeuron: resource.MustParse("1")},
+							Limits:   v1.ResourceList{resources.NvidiaGPU: resource.MustParse("1"), resources.AWSNeuron: resource.MustParse("1")},
+						},
+					}))
+				ExpectNotScheduled(ctx, env.Client, pods[0])
 			})
 			It("should not schedule a non-GPU workload on a node w/GPU", func() {
 				Skip("enable after scheduling and binpacking are merged into the same process")
@@ -536,7 +568,8 @@ var _ = Describe("Allocation", func() {
 		})
 		Context("Subnets", func() {
 			It("should default to the cluster's subnets", func() {
-				pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioners, ProvisionerWithProvider(provisioner, provider), test.UnschedulablePod())[0]
+				pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioners, ProvisionerWithProvider(provisioner, provider), test.UnschedulablePod(
+					test.PodOptions{NodeSelector: map[string]string{v1.LabelArchStable: v1alpha5.ArchitectureAmd64}}))[0]
 				ExpectScheduled(ctx, env.Client, pod)
 				Expect(fakeEC2API.CalledWithCreateFleetInput.Cardinality()).To(Equal(1))
 				input := fakeEC2API.CalledWithCreateFleetInput.Pop().(*ec2.CreateFleetInput)
--- a/pkg/controllers/provisioning/scheduling/node.go
+++ b/pkg/controllers/provisioning/scheduling/node.go
@@ -1,0 +1,173 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scheduling
+
+import (
+	"errors"
+
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
+	"github.com/aws/karpenter/pkg/cloudprovider"
+	"github.com/aws/karpenter/pkg/utils/resources"
+)
+
+// Node is a set of constraints, compatible pods, and possible instance types that could fulfill these constraints. This
+// will be turned into one or more actual node instances within the cluster after bin packing.
+type Node struct {
+	Constraints         *v1alpha5.Constraints
+	InstanceTypeOptions []cloudprovider.InstanceType
+	Pods                []*v1.Pod
+}
+
+func NewNode(constraints *v1alpha5.Constraints, instanceTypeOptions []cloudprovider.InstanceType, pods ...*v1.Pod) *Node {
+	n := &Node{Constraints: constraints.DeepCopy()}
+
+	instanceTypes := constraints.Requirements.InstanceTypes()
+	for _, it := range instanceTypeOptions {
+		// provisioner must list the instance type by name (defaults to all instance types)
+		if !instanceTypes.Has(it.Name()) {
+			continue
+		}
+		included := false
+		// and the instance type must have some valid offering combination per the provisioner constraints
+		for _, off := range it.Offerings() {
+			if constraints.Requirements.Zones().Has(off.Zone) && constraints.Requirements.CapacityTypes().Has(off.CapacityType) {
+				included = true
+				break
+			}
+		}
+		if included {
+			n.InstanceTypeOptions = append(n.InstanceTypeOptions, it)
+		}
+	}
+	// technically we might create some invalid construct here if the pod can't be supported by any instance type
+	// but this is ok as we just won't have any valid instance types, this pod won't be schedulable and nothing else
+	// will be compatible with this node
+	for _, p := range pods {
+		n.Add(p)
+	}
+	return n
+}
+func (n Node) Compatible(pod *v1.Pod) error {
+	podRequirements := v1alpha5.NewPodRequirements(pod)
+	if err := n.Constraints.Requirements.Compatible(podRequirements); err != nil {
+		return err
+	}
+
+	// Ensure that at least one instance type of the instance types that we are already narrowed down to based on the
+	// existing pods can support the pod
+	for _, it := range n.InstanceTypeOptions {
+		if n.isPodCompatible(pod, it) {
+			return nil
+		}
+	}
+
+	return errors.New("no matching instance type found")
+}
+
+func (n *Node) Add(pod *v1.Pod) {
+	n.Pods = append(n.Pods, pod)
+	n.Constraints = n.Constraints.Tighten(pod)
+	var instanceTypeOptions []cloudprovider.InstanceType
+	for _, it := range n.InstanceTypeOptions {
+		if n.isPodCompatible(pod, it) {
+			instanceTypeOptions = append(instanceTypeOptions, it)
+		}
+	}
+	n.InstanceTypeOptions = instanceTypeOptions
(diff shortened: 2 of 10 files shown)


pkg/cloudprovider/aws/suite_test.go
```diff
@@ -214,6 +214,37 @@ var _ = Describe("Allocation", func() {
 				}
 				Expect(nodeNames.Len()).To(Equal(2))
 			})
+			It("should launch pods with Nvidia and Neuron resources on different instances", func() {
```
ellistarn, line 217: I remember asking something similar, but when can we move these tests to generic code? Can we at least cut an OE issue?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -45,87 +45,49 @@ func init() {
...
-				c := schedule.Tighten(pod)
-				for _, instanceType := range instanceTypes {
-					if support(instanceType, c) {
-						schedule.Constraints = c
-						schedule.Pods = append(schedule.Pods, pod)
-						isCompatible = true
-						break
-					}
-				}
+		canScheduleOnExistingNode := false
```
ellistarn, line 78: Optional more brief name
```suggestion
		isScheduled := false
```

pkg/controllers/provisioning/scheduling/theoreticalnode.go
```diff
@@ -0,0 +1,137 @@
...
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
+	"github.com/aws/karpenter/pkg/cloudprovider"
+	"github.com/aws/karpenter/pkg/utils/resources"
+)
+
+// TheoreticalNode is a set of constraints, compatible pods, and possible instance types that could fulfill
+// these constraints. This will be turned into one or more actual node instances within the cluster after bin packing.
+type TheoreticalNode struct {
```
ellistarn, line 29: Optional, I think we could reasonably just call this `Node`, which would make it a `scheduling.Node` instead of a `v1.Node`. We're already calling the variable names `node` anyways. I'm a big fan of context driven naming, where the names themselves are very simple, and any contextual adjectives are replaced by the package's context. This works even better with natural package names. It's not a perfect fit like `simulation.Node` or `simulated.Node`, but good enough, I think.

pkg/controllers/provisioning/scheduling/theoreticalnode.go
```diff
@@ -0,0 +1,137 @@
...
+	Constraints           *v1alpha5.Constraints
```
ellistarn, line 30: Thoughts on inlining this type?

pkg/controllers/provisioning/scheduling/theoreticalnode.go
```diff
@@ -0,0 +1,137 @@
...
+	PossibleInstanceTypes []cloudprovider.InstanceType
```
ellistarn, line 31: Optional
```suggestion
	InstanceTypeOptions []cloudprovider.InstanceType
```

pkg/controllers/provisioning/scheduling/theoreticalnode.go
```diff
@@ -0,0 +1,137 @@
...
+
+	if os, ok := nodeSelector[v1.LabelOSStable]; ok {
+		if !it.OperatingSystems().Has(os) {
+			return false
+		}
+	}
+	return true
+}
+
+func (n TheoreticalNode) isPodCompatible(pod *v1.Pod, it cloudprovider.InstanceType) bool {
```
ellistarn, line 106: I might just inline this, since this function name doesn't seem distinguished from `isCompatible`

pkg/controllers/provisioning/scheduling/theoreticalnode.go
```diff
@@ -0,0 +1,137 @@
...
+	podRequirements := v1alpha5.NewPodRequirements(pod)
+	if err := n.Constraints.Requirements.Compatible(podRequirements); err != nil {
+		return err
+	}
+
+	// Ensure that at least one instance type of the instance types that we are already narrowed down to based on the
+	// existing pods can support the pod
+	matching := 0
+	for _, it := range n.PossibleInstanceTypes {
+		if n.isPodCompatible(pod, it) {
```
ellistarn, line 51: This is a tight loop. Better performance to shortcircuit.
```suggestion
		if n.isPodCompatible(pod, it) {
		  return nil
		}
```

pkg/controllers/provisioning/scheduling/theoreticalnode.go
```diff
@@ -0,0 +1,137 @@
...
+			compatibleTypes = append(compatibleTypes, it)
+		}
+	}
+	n.PossibleInstanceTypes = compatibleTypes
+}
+
+// hasCompatibleNodeSelector returns true if a given set of node selectors match the instance type
+func (n TheoreticalNode) hasCompatibleNodeSelector(nodeSelector map[string]string, it cloudprovider.InstanceType) bool {
```
ellistarn, line 75: Can we remove the binpacking validation now that we have this?

pkg/controllers/provisioning/scheduling/theoreticalnode.go
ellistarn, line 106: Seeing you wire through InstanceType everywhere makes me think that this method should potentially live on that type, e.g. `instanceType.Compatible(pod)`. I know it's an interface, so we can't attach behavior. I'm not sure the right answer without playing with it, but it does feel a bit odd to be wiring it through so many methods.

Perhaps most of this collapses once we can genericize resources.

pkg/controllers/provisioning/scheduling/theoreticalnode.go
```diff
@@ -0,0 +1,137 @@
...
+		if !it.OperatingSystems().Has(os) {
+			return false
+		}
+	}
+	return true
+}
+
+func (n TheoreticalNode) isPodCompatible(pod *v1.Pod, it cloudprovider.InstanceType) bool {
+	return n.hasCompatibleResources(resources.RequestsForPods(pod), it) &&
+		n.hasCompatibleResources(resources.LimitsForPods(pod), it) &&
```
ellistarn, line 108: We shouldn't be checking limits.

pkg/controllers/provisioning/scheduling/theoreticalnode.go
```diff
@@ -0,0 +1,137 @@
...
+// hasCompatibleNodeSelector returns true if a given set of node selectors match the instance type
+func (n TheoreticalNode) hasCompatibleNodeSelector(nodeSelector map[string]string, it cloudprovider.InstanceType) bool {
+	if ns, ok := nodeSelector[v1.LabelInstanceTypeStable]; ok {
+		if ns != it.Name() {
+			return false
+		}
+	}
+	if zone, ok := nodeSelector[v1.LabelTopologyZone]; ok {
+		matched := false
+		for _, off := range it.Offerings() {
```
ellistarn, line 83: Seeing this `FindOrDefault` idiom so many times makes me really excited for generics.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -45,87 +45,49 @@ func init() {
...
-		for key, values := range supported {
-			r = append(r, v1.NodeSelectorRequirement{Key: key, Operator: v1.NodeSelectorOpIn, Values: values})
-		}
-		requirements := v1alpha5.NewRequirements(r...)
-		if err := requirements.Compatible(constraints.Requirements); err == nil {
-			return true
+		if !canScheduleOnExistingNode {
+			n := NewTheoreticalNode(constraints, instanceTypes)
+			n.Add(pod)
+			nodes = append(nodes, n)
```
ellistarn, line 89: would be kind of cool if we made Add return the node object (builder pattern) to enable 
`nodes = append(nodes, NewTheoretricalNode(constraints, instanceTypes).Add(pod))`

ellistarn (commented):
Very cool to see this. I'm assuming that the binpacking logic will start to evaporate given the overlap of concerns in this PR. Is that correct?

pkg/cloudprovider/aws/suite_test.go
tzneal: We could do this at the same type as making resource handling generic, that PR I've got will need to be updated after this merge.  I think it's still waiting on a consensus regarding sorting instance types and 'preferred pod instance type compatibility' (e.g. don't schedule a non-GPU workload on a GPU instance type just because there are plenty of GPU instance types available).  My preference is a comparison method on cloud provider for the first, and just letting cloud providers do the second part internally, which they could do now.

Wrote up an issue here: https://github.com/aws/karpenter/issues/1516

pkg/controllers/provisioning/scheduling/scheduler.go
tzneal: SGTM

pkg/controllers/provisioning/scheduling/theoreticalnode.go
tzneal: Yes, but they're all bad 😆 To me it doesn't scream out to be inlined and would expose the Constraints methods on the Node which I think reduces readability:

```go
var n Node
n.Constraints = n.Tighten(pod)
```

pkg/controllers/provisioning/scheduling/theoreticalnode.go
tzneal: Nice, good catch. Especially since there are 300+

pkg/controllers/provisioning/scheduling/theoreticalnode.go
tzneal: I didn't realize that the API server copied the limits over into requests if you left requests empty, fixed.

pkg/controllers/provisioning/scheduling/theoreticalnode.go
tzneal: Yeah, but just think about how good you get at writing the same 6 lines of code 🤣

pkg/controllers/provisioning/scheduling/scheduler.go
tzneal: I like the second better, done.

pkg/controllers/provisioning/scheduling/theoreticalnode.go
tzneal: Sounds good, I went ahead and moved all the logic to node and removed it from packable.

pkg/controllers/provisioning/scheduling/theoreticalnode.go
tzneal: Maybe so, I think generic resources will clean up quite a bit of things.

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -0,0 +1,173 @@
...
+
+	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
+	"github.com/aws/karpenter/pkg/cloudprovider"
+	"github.com/aws/karpenter/pkg/utils/resources"
+)
+
+// Node is a set of constraints, compatible pods, and possible instance types that could fulfill these constraints. This
+// will be turned into one or more actual node instances within the cluster after bin packing.
+type Node struct {
+	Constraints         *v1alpha5.Constraints
```
ellistarn, line 30: Thoughts on turning this into Requirements and Taints and not reusing this API? This is probably the right place to separate out the requirements implementation. Feel free to punt to another PR, just curious about your thoughts.

tzneal: Yeah, Constraints/Requirements stick out as rough edges when writing code.

pkg/utils/resources/resources.go
```diff
@@ -48,19 +48,6 @@ func LimitsForPods(pods ...*v1.Pod) v1.ResourceList {
 	return Merge(resources...)
 }
 
-// GPULimitsFor returns a resource list of GPU limits from a pod
```
ellistarn, line 51: Can we removed LimitsForPods, too?

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -0,0 +1,173 @@
...
+	if os, ok := nodeSelector[v1.LabelOSStable]; ok {
+		if !it.OperatingSystems().Has(os) {
+			return false
+		}
+	}
+	return true
+}
+
+func (n Node) isPodCompatible(pod *v1.Pod, it cloudprovider.InstanceType) bool {
+
```
ellistarn, line 143:
```suggestion
```

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+
+	return errors.New("no matching instance type found")
+}
+
+func (n *Node) Add(pod *v1.Pod) {
+	n.Pods = append(n.Pods, pod)
+	n.Constraints = n.Constraints.Tighten(pod)
+	var compatibleTypes []cloudprovider.InstanceType
```
ellistarn, line 83: Optional: perhaps more idiomatic to have naming parity since this is the same context at different scopes. 
```suggestion
	var instanceTypeOptions []cloudprovider.InstanceType
```

pkg/utils/resources/resources.go
tzneal: It's used in metrics

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Ensure that at least one instance type of the instance types that we are already narrowed down to based on the
+	// existing pods can support the pod
+	for _, it := range n.InstanceTypeOptions {
+		if n.isPodCompatible(pod, it) {
+			return nil
+		}
+	}
+
+	return errors.New("no matching instance type found")
```
ellistarn, line 77: The errors package functionality is now replaced by fmt.Errorf and %w: https://github.com/pkg/errors/issues/202
```suggestion
	return errors.New("no matching instance type found")
```

ellistarn: Also -- would be great to wire through a reason, since this may be a bit hard to reason about for users.

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -0,0 +1,173 @@
...
+type Node struct {
+	Constraints         *v1alpha5.Constraints
+	InstanceTypeOptions []cloudprovider.InstanceType
+	Pods                []*v1.Pod
+}
+
+func NewNode(constraints *v1alpha5.Constraints, instanceTypes []cloudprovider.InstanceType, pods ...*v1.Pod) *Node {
+	n := &Node{Constraints: constraints.DeepCopy()}
+
+	provIts := constraints.Requirements.InstanceTypes()
```
ellistarn, line 38: Optional: you can avoid this naming conflict by calling the method parameter `instanceTypeOptions`
```suggestion
	instanceTypes := constraints.Requirements.InstanceTypes()
```

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -0,0 +1,173 @@
...
+	for _, it := range instanceTypes {
+		// provisioner must list the instance type by name (defaults to all instance types)
+		if provIts.Has(it.Name()) {
```
ellistarn, line 41: Consider a shortcircuitting idiom to avoid deeply nesting lines 42-52

```suggestion
		if !provIts.Has(it.Name()) {
		  continue
		}
```

pkg/controllers/provisioning/scheduling/node.go
tzneal: I think errors.New() is still preferred if you don't have a need for the Sprintf machinery and %w, I've got no other error to pass up. e.g. from stdlib 

```
package io
var EOF = errors.New("EOF")
```

This error is never presented to users, at least not currently.  Compatible could just be a boolean, but I made it return an error to be similar to other Compatible methods.

ellistarn (approved):

ellistarn:
Nice work!

PR #1530, opened 2022-03-16
Remove old archives of the website

spring1843:
**1. Issue, if available:**

Since the web server serves only [subset of versions](https://github.com/aws/karpenter/blob/bb65375ae4f1a6ca1a4a30f217162abf71bd5dcc/website/config.yaml#L69), we do not appear to have a need to maintain these. If for whatever reason someone is interested in the old archives they can always switch their clone of the repo back into the time of the creation of the tag.

- [x] No


ellistarn (commented):
Awesome. Should we change our release process to do this as well?

spring1843:
No you are right I thought it wasn't serving them, I think we ultimately should move these out of the repo to reduce the size.

PR #1541, opened 2022-03-18
allow pod scheduling if they express soft pod affinity / anti affinity

cebernardi:
**1. Issue, if available:**

- Changed method name `Pod.HasPodAffinity` to `Pod.HasRequiredPodAffinity` 
- `Pod.HasRequiredPodAffinity` returns true only if `PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution` is not empty
- Changed method name `Pod.HasPodAntiAffinity` to `Pod.HasRequiredPodAntiAffinity` 
- `Pod.HasRequiredPodAntiAffinity` returns true only if `PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution` is not empty
- added the relevant tests

- UnitTest
- Manual testing in an EKS cluster:
    -  pods with `podAffinity.requiredDuringSchedulingIgnoredDuringExecution` spec, fail to be scheduled by Karpenter with `Ignoring pod, strong pod affinity is not supported` error
    -  pods with `podAntyAffinity.requiredDuringSchedulingIgnoredDuringExecution` spec, fail to be scheduled by Karpenter with `Ignoring pod, strong pod anti affinity is not supported` error
   -  pods with `podAffinity.preferredDuringSchedulingIgnoredDuringExecution` spec are scheduled by Karpenter
   -  pods with `podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution` spec are scheduled by Karpenter
   
- [x] No

I looked at the documentation and didn't find any specific reference to podAffinity and antiAffinity, a part from this bit in the scheduling session:
```
Pod affinity is a key exception to this rule.
```
but I'm happy to add more, if needed

### DIFF
--- a/pkg/controllers/selection/controller.go
+++ b/pkg/controllers/selection/controller.go
@@ -142,11 +142,11 @@ func validateAffinity(p *v1.Pod) (errs error) {
 	if p.Spec.Affinity == nil {
 		return nil
 	}
-	if pod.HasPodAffinity(p) {
-		errs = multierr.Append(errs, fmt.Errorf("pod affinity is not supported"))
+	if pod.HasRequiredPodAffinity(p) {
+		errs = multierr.Append(errs, fmt.Errorf("strong pod affinity is not supported"))
 	}
-	if pod.HasPodAntiAffinity(p) {
-		errs = multierr.Append(errs, fmt.Errorf("pod anti-affinity is not supported"))
+	if pod.HasRequiredPodAntiAffinity(p) {
+		errs = multierr.Append(errs, fmt.Errorf("strong pod anti-affinity is not supported"))
 	}
 	if p.Spec.Affinity.NodeAffinity != nil {
 		for _, term := range p.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution {
--- a/pkg/controllers/selection/suite_test.go
+++ b/pkg/controllers/selection/suite_test.go
@@ -321,19 +321,19 @@ var _ = Describe("Pod Affinity and AntiAffinity", func() {
 		}))[0]
 		ExpectNotScheduled(ctx, env.Client, pod)
 	})
-	It("should not schedule a pod with pod affinity preference", func() {
+	It("should schedule a pod with pod affinity preference", func() {
 		ExpectCreated(ctx, env.Client)
 		pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner, test.UnschedulablePod(test.PodOptions{
 			PodPreferences: []v1.WeightedPodAffinityTerm{{Weight: 1, PodAffinityTerm: v1.PodAffinityTerm{TopologyKey: "foo"}}},
 		}))[0]
-		ExpectNotScheduled(ctx, env.Client, pod)
+		ExpectScheduled(ctx, env.Client, pod)
 	})
-	It("should not schedule a pod with pod anti-affinity preference", func() {
+	It("should schedule a pod with pod anti-affinity preference", func() {
 		ExpectCreated(ctx, env.Client)
 		pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner, test.UnschedulablePod(test.PodOptions{
 			PodAntiPreferences: []v1.WeightedPodAffinityTerm{{Weight: 1, PodAffinityTerm: v1.PodAffinityTerm{TopologyKey: "foo"}}},
 		}))[0]
-		ExpectNotScheduled(ctx, env.Client, pod)
+		ExpectScheduled(ctx, env.Client, pod)
 	})
 	It("should schedule a pod with empty pod affinity and anti-affinity", func() {
 		ExpectCreated(ctx, env.Client)
--- a/pkg/utils/pod/scheduling.go
+++ b/pkg/utils/pod/scheduling.go
@@ -68,16 +68,16 @@ func IsOwnedBy(pod *v1.Pod, gvks []schema.GroupVersionKind) bool {
 	return false
 }
 
-// HasPodAffinity returns true if a non-empty PodAffinity is defined in the pod spec
-func HasPodAffinity(pod *v1.Pod) bool {
+// HasRequiredPodAffinity returns true if a non-empty PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution
+// is defined in the pod spec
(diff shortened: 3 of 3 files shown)


cebernardi:
relates to https://github.com/aws/karpenter/issues/942

pkg/controllers/selection/controller.go
```diff
@@ -142,11 +142,11 @@ func validateAffinity(p *v1.Pod) (errs error) {
 	if p.Spec.Affinity == nil {
 		return nil
 	}
-	if pod.HasPodAffinity(p) {
-		errs = multierr.Append(errs, fmt.Errorf("pod affinity is not supported"))
+	if pod.HasRequiredPodAffinity(p) {
+		errs = multierr.Append(errs, fmt.Errorf("strong pod affinity is not supported"))
```
cebernardi, line 146: I'm not sure if this message is clear enough.
another alternative could be "pod affinity required during scheduling is not supported"

pkg/controllers/selection/controller.go
```diff
@@ -142,11 +142,11 @@ func validateAffinity(p *v1.Pod) (errs error) {
...
 	}
-	if pod.HasPodAntiAffinity(p) {
-		errs = multierr.Append(errs, fmt.Errorf("pod anti-affinity is not supported"))
+	if pod.HasRequiredPodAntiAffinity(p) {
+		errs = multierr.Append(errs, fmt.Errorf("strong pod anti-affinity is not supported"))
```
cebernardi, line 149: same as above

cebernardi:
@ellistarn I read your comment [here](https://github.com/aws/karpenter/issues/942#issuecomment-1071087522)

From my understanding of the code, I was under the impression that Karpenter doesn't manage at the moment any podAffinity / AntiAffinity, so, unlike nodeAffinity, so I thought I didn't have to relax the preferences (iteratively or not), as they're already ignored. 

Happy to revisit, of course :)

pkg/controllers/selection/controller.go
ellistarn, line 146:
```suggestion
		errs = multierr.Append(errs, fmt.Errorf("pod affinity rule  'requiredDuringSchedulingIgnoreDuringExecution' is not supported"))
```

ellistarn (commented):
Nice work! Minor comment.

cebernardi:
thanks!
I've replicated your suggestion on the other error message as well, do you want me to squash?

pkg/controllers/selection/controller.go
```diff
@@ -142,11 +142,11 @@ func validateAffinity(p *v1.Pod) (errs error) {
 	if p.Spec.Affinity == nil {
 		return nil
 	}
-	if pod.HasPodAffinity(p) {
-		errs = multierr.Append(errs, fmt.Errorf("pod affinity is not supported"))
+	if pod.HasRequiredPodAffinity(p) {
+		errs = multierr.Append(errs, fmt.Errorf("pod affinity rule 'requiredDuringSchedulingIgnoreDuringExecution' is not supported"))
```
alanjjenkins, line 146: Is this meant to say 'is not supported'?

https://github.com/aws/karpenter/pull/1541/files#diff-a6635dd6ddb6e0a211850401eb1275a378603ffa15afdcb98c2ba41ab624476dR329

Seems to indicate that you are stating it should be supported in the tests. The same applies to pod Anti Affinity.

alanjjenkins (commented):
Just took a look at the PR and unless I am missing something I think you need to reword the error messages.

pkg/controllers/selection/controller.go
alanjjenkins: Actually reading the comments above looks like this is intentional...

tzneal: Sorry, I didn't follow this comment. This change was making Karpenter ignore preferred pod affinity, while still rejecting scheduling of required pod affinity entirely.

ellistarn: This may be more simply achieved by implementing using preferences.go

cebernardi: Sure, happy that give a try!

ellistarn:
FWIW, I'm happy to accept this PR as is.

pkg/controllers/selection/controller.go
cebernardi: @ellistarn I included pod affinity and anti-affinity preferences relaxation, but I had to maintain the affinity validation as it's proposed in this PR (otherwise the code would return [here](https://github.com/aws/karpenter/blob/14d84eb8ba231fe5966346936374b2ed20b545ca/pkg/controllers/selection/controller.go#L74), and we won't be able to reach `Preferences.Relax` as you were suggesting [here](https://github.com/aws/karpenter/pull/1541#discussion_r830330696)).

Something that I'm not 100% convinced about, is that currently Karpenter doesn't take into consideration any podAffinity / podAntiAffinity to match with a provisioner, so it sounds a bit irrelevant to iteratively remove pod affinity / anti-affinity preferences, it looks to me that it only adds overhead without bringing much value. 

What your thoughts are on this one?

pkg/controllers/selection/preferences.go
```diff
@@ -71,6 +71,8 @@ func (p *Preferences) Relax(ctx context.Context, pod *v1.Pod) {
 
 func (p *Preferences) relax(ctx context.Context, pod *v1.Pod) bool {
 	for _, relaxFunc := range []func(*v1.Pod) *string{
+		func(pod *v1.Pod) *string { return p.removePreferredPodAffinityTerm(pod) },
+		func(pod *v1.Pod) *string { return p.removePreferredPodAntiAffinityTerm(pod) },
```
cebernardi, line 75: I'm not sure about the positioning of these 2 relaxation functions.
Should they go after `removePreferredNodeAffinityTerm`?

tzneal: I think this is fine since especially for now they will always need to be removed.

pkg/controllers/selection/controller.go
tzneal: There shouldn't be too many terms, and it will hopefully align well with future support for required pod affinity/anti-affinity.

cebernardi: thank you @tzneal for your quick reply! :)

tzneal (approved):
LGTM, thanks for the contribution!

robertd:
@adjain131995 0.7.3 got released this morning.

PR #1543, opened 2022-03-18
fix issue regarding not pre-filtering instance types by provider constraints

### LINKED ISSUE #1540: Provisioner is creating some arm64 nodes even with kubernetes.io/arch in  ["amd64"] requirement
## Version

**Karpenter**: v0.7.1

**Kubernetes**: v1.2.1
Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.5-eks-bc4871b", GitCommit:"5236faf39f1b7a7dabea8df12726f25608131aa9", GitTreeState:"clean", BuildDate:"2021-10-29T23:32:16Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}

## Expected Behavior
karpenter should only provision amd64 nodes since it's mentioned in provisioner requirements.

## Actual Behavior
karpenter is adding some arm64 nodes that are causing "standard_init_linux.go:228: exec user process caused: exec format error" for some pods because of architecture incompatibility.
Also 
`k get nodes -l kubernetes.io/arch=arm64` is returning some nodes created by karpenter 

1) create eks cluster using https://karpenter.sh/v0.7.1/getting-started/getting-started-with-eksctl/
2) apply this provisioner
```
kind: Provisioner
metadata:
  name: default
spec:
  requirements:
      operator: In
      values: ["spot"]
    - key: kubernetes.io/arch
      operator: In
      values: ["amd64"]
  limits:
    resources:
      cpu: 1000
  provider:
    subnetSelector:
      karpenter.sh/discovery: ${CLUSTER_NAME}
    securityGroupSelector:
      # karpenter.sh/discovery returns multiple security groups which cause conflict with ingress controller
      # karpenter.sh/discovery: ${CLUSTER_NAME}
      kubernetes.io/cluster/${CLUSTER_NAME}: owned
  ttlSecondsAfterEmpty: 30
```
3) Run some pods to force karpenter to add nodes. 

## Resource Specs and Logs
Will add later. Provisioner spec is already provided.

tzneal:
**1. Issue, if available:**

Fixes #1540

Pre-filter the available instance types by the provider constraints

- [X] No

### DIFF
--- a/pkg/cloudprovider/fake/cloudprovider.go
+++ b/pkg/cloudprovider/fake/cloudprovider.go
@@ -109,6 +109,8 @@ func (c *CloudProvider) GetInstanceTypes(_ context.Context, _ *v1alpha5.Provider
 		NewInstanceType(InstanceTypeOptions{
 			Name:         "arm-instance-type",
 			Architecture: "arm64",
+			CPU:          resource.MustParse("16"),
+			Memory:       resource.MustParse("128Gi"),
 		}),
 	}, nil
 }
--- a/pkg/controllers/provisioning/scheduling/node.go
+++ b/pkg/controllers/provisioning/scheduling/node.go
@@ -19,6 +19,8 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 
+	"github.com/aws/karpenter/pkg/utils/sets"
+
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/utils/resources"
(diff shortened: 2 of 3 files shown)


pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -61,6 +74,17 @@ func NewNode(constraints *v1alpha5.Constraints, instanceTypeOptions []cloudprovi
 	}
 	return n
 }
+
+// TODD(todo): clean up this usage of two different set libraries
+func containsAny(get sets.Set, systems sset.String) bool {
```
felix-zhe-huang, line 79: I think it will be easier to just convert sset to the karpenter set.
```
s := sets.NewSet(systems.UnsortedList()...)

File: pkg/cloudprovider/fake/cloudprovider.go [hunk 2]
```diff
@@ -109,6 +109,8 @@ func (c *CloudProvider) GetInstanceTypes(_ context.Context, _ *v1alpha5.Provider
 		NewInstanceType(InstanceTypeOptions{
 			Name:         "arm-instance-type",
 			Architecture: "arm64",
+			CPU:          resource.MustParse("16"),
```
felix-zhe-huang, line 112: I am curious why do we need to specify the CPU and memory here?

tzneal: I can only verify what it gets scheduled to, so I bumped the cpu/memory up and set a corresponding request in the pod to force it to either fit this instance or not be scheduled.  Now with the filtering, it ends up not scheduled.

pkg/controllers/provisioning/scheduling/node.go
tzneal: Yeah, that's cleaner. Modified.

bwagner5 (approved):
lgtm

PR #1544, opened 2022-03-18
Rm/find broken links

spring1843:
**1. Issue, if available:**

Attempts to find broken links in the docs

- [ ] No


spring1843:
The plugin has an inherent problem of checking files on a disk not from a web server. This will make it hard for links like `foo/#bar` to work, to make the plugin work we would need to convert all such links to the equivalent but less human friendly `foo/index.html#bar`.

PR #1546, opened 2022-03-18
Lint YAML files

spring1843:
**1. Issue, if available:**

Adds linting to yaml files.

YAML is a pretty easy format to mess up, and bad YAML files can cause weird unexpected results, hence it's important to protect the code base from introduction of bad yaml files and also lint our yaml files to enforce some level of standardization and cleanliness.

This adds a yaml linter to the repo, for now it disables checking for all the problems that we already have, we can later on enable them one by one after fixing the issues.

- [x] No

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,6 +14,9 @@ jobs:
       K8S_VERSION: ${{ matrix.k8sVersion }}
     steps:
     - uses: actions/checkout@v2
+    - uses: ibiqlik/action-yamllint@v3
+      with:
+        config_file: .github/workflows/config/.yamllint.yaml
     - uses: actions/setup-go@v2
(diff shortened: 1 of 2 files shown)


ellistarn:
I'd love to include this in `make verify` rather than in a GH action. It's not easy to run GH action code locally, and ideally this would be executed as part of the dev workflow. In general, I lean away from GH actions that do more than trigger scripts that we can run locally. 

Bonus points if such yaml linter can be installed with `go install`. Unfortunately, I'm not finding a good one. There are some benefits to GH actions, such as not needing a cross-platform install story locally. Go based tools are easy, but python/js are nastier to include in the dev loop. Let me know if you find anything, otherwise I'm happy with GH action approach.

spring1843:
Still working on this.

PR #1549, opened 2022-03-19
Resolved two bugs with NodeAffinity and simplified compatibility logic

ellistarn:
### Fixed: A should be compatible to B, <DoesNotExist, NotIn> operator
According to the second bullet at https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement, DoesNotExist and NotIn are compatible. For example, A Pod that requires `foo 'not in' bar` may schedule to a node without a label with key `foo`. 

Previously, we required the node to have the label `foo` with a value of something other than `bar`.

### Fixed: A should be compatible to B, <Exists, Empty> operator
Node labels cannot repel a pod (see: taints), unless a pod has conflicting requirements. In this case, a pod with empty requirements should happily schedule to a node with an Exists requirement on a node.

Previously, the pod needed to have a requirement for label `foo` in order to schedule onto node with requirement `foo: Exists`.

*Note: ` provisioner only supports [In NotIn Exists]`, so it wasn't currently possible to run into this bug.*

### Cleanup: NullSet == EmptySet
Previously, we treated the undefined set to be the same as the universal set. In set theory, the nil set and the empty set are typically modeled equivalently (https://en.wikipedia.org/wiki/Empty_set). Treating NullSet == UniversalSet set was relied upon in `requirements.Add`, but significantly increased the complexity of our compatibility checks. Instead, we intersect requirements in `requirements.Add` if a requirement for the given key already exists, otherwise we use the added requirement as-is.

### Cleanup: Compatibility Checks
Previously we oriented the logic around NodeSelectorOp rules via `requirements.hasRequirement(withKeyAndOperator(key, v1.NodeSelectorOpExists))`. Instead, we can use the properties of the underlying set (e.g. Full, Empty, Positive, Negative). This collapses O(n^2) iteration through requirements to O(n), and simplifies the predicates.

### Cleanup: Requirement Validation
Previously, requirement validation logic redundantly checked that DoesNotExist operators did not coexist with other operators as well as ensuring that valid values exist for any requirement with an operator other than DoesNotExist. This has been collapsed into a single check, and no longer relies on `requirements.hasRequirement(withKeyAndOperator`. This also collapses O(n^2) iteration through requirements to O(n), and simplifies the predicates.

- `make test`
- Manually

```
kind: Provisioner
metadata:
  name: default
spec:
  requirements:
  - key: isolation
    operator: Exists
```
```
apiVersion: v1
kind: Pod
metadata:
  name: inflate-6d7d985f4d-8pkgx
  namespace: default
spec:
  containers:
  - image: public.ecr.aws/eks-distro/kubernetes/pause:3.2
    name: inflate
    resources:
      requests:
        cpu: "1"
  nodeSelector:
    isolation: ellis
```
```
apiVersion: v1
kind: Node
metadata:
  annotations:
    node.alpha.kubernetes.io/ttl: "0"
  creationTimestamp: "2022-03-19T19:47:34Z"
  finalizers:
  - karpenter.sh/termination
  labels:
    isolation: ellis <------------------------------ Here it is!
    karpenter.sh/capacity-type: on-demand
    karpenter.sh/provisioner-name: default
    node.kubernetes.io/instance-type: t3a.micro
    topology.kubernetes.io/zone: us-west-2b
  name: ip-192-168-120-29.us-west-2.compute.internal
```

- [X] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/constraints.go
+++ b/pkg/apis/provisioning/v1alpha5/constraints.go
@@ -71,7 +71,7 @@ func (c *Constraints) GenerateLabels() map[string]string {
 	for key := range c.Requirements.Keys() {
 		if !IsRestrictedNodeLabel(key) {
 			// Ignore cases when values set is empty (i.e., DoesNotExist or <In, NotIn> cancling out)
-			if c.Requirements.Get(key).IsEmpty() {
+			if c.Requirements.Get(key).Len() == 0 {
 				continue
 			}
 			labels[key] = c.Requirements.Label(key)
--- a/pkg/apis/provisioning/v1alpha5/constraints.go
+++ b/pkg/apis/provisioning/v1alpha5/constraints.go
@@ -19,6 +19,8 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/aws/karpenter/pkg/utils/rand"
 )
 
 // Constraints are applied to all nodes created by the provisioner.
@@ -70,11 +72,12 @@ func (c *Constraints) GenerateLabels() map[string]string {
 	}
 	for key := range c.Requirements.Keys() {
 		if !IsRestrictedNodeLabel(key) {
-			// Ignore cases when values set is empty (i.e., DoesNotExist or <In, NotIn> cancling out)
-			if c.Requirements.Get(key).Len() == 0 {
-				continue
+			switch c.Requirements.Get(key).Type() {
+			case v1.NodeSelectorOpIn:
+				labels[key] = c.Requirements.Get(key).Values().UnsortedList()[0]
+			case v1.NodeSelectorOpExists:
+				labels[key] = rand.String(10)
 			}
-			labels[key] = c.Requirements.Label(key)
 		}
 	}
 	return labels
--- a/pkg/apis/provisioning/v1alpha5/requirements.go
+++ b/pkg/apis/provisioning/v1alpha5/requirements.go
@@ -24,7 +24,6 @@ import (
 	stringsets "k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
 
-	"github.com/aws/karpenter/pkg/utils/rand"
 	"github.com/aws/karpenter/pkg/utils/sets"
 )
 
@@ -100,8 +99,6 @@ func (r Requirements) Add(requirements ...v1.NodeSelectorRequirement) Requiremen
 			values = sets.NewComplementSet()
 		case v1.NodeSelectorOpDoesNotExist:
 			values = sets.NewSet()
-		default:
-			values = sets.NewSet()
 		}
 		if existing, ok := r.requirements[requirement.Key]; ok {
 			values = values.Intersection(existing)
@@ -118,21 +115,6 @@ func (r Requirements) Keys() stringsets.String {
 		keys.Insert(requirement.Key)
 	}
 	return keys
-}
-
-// Labels returns value realization for the provided key.
-// If the set is a complement set, return a randomly generated value.
-// If the set is not a complement set, return the first value in the set.
-func (r Requirements) Label(key string) string {
-	values := r.Get(key)
-	if values.IsComplement() {
-		label := rand.String(10)
-		for !values.Has(label) {
-			label = rand.String(10)
-		}
-		return label
-	}
-	return values.Values().UnsortedList()[0]
 }
 
 func (r Requirements) Has(key string) bool {
@@ -180,18 +162,9 @@ func (r Requirements) Validate() (errs error) {
 		if !SupportedNodeSelectorOps.Has(string(requirement.Operator)) {
 			errs = multierr.Append(errs, fmt.Errorf("operator %s not in %s for key %s", requirement.Operator, SupportedNodeSelectorOps.UnsortedList(), requirement.Key))
 		}
-		// Excludes cases when DoesNotExists appears together with In, NotIn, Exists
-		if requirement.Operator == v1.NodeSelectorOpDoesNotExist {
-			if r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpIn)) ||
-				r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpNotIn)) ||
-				r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpExists)) {
-				errs = multierr.Append(errs, fmt.Errorf("operator %s cannot coexist with other operators for key %s", v1.NodeSelectorOpDoesNotExist, requirement.Key))
-			}
-		}
-	}
-	for key := range r.Keys() {
-		if r.Get(key).Len() == 0 && !r.hasRequirement(withKeyAndOperator(key, v1.NodeSelectorOpDoesNotExist)) {
-			errs = multierr.Append(errs, fmt.Errorf("no feasible value for key %s", key))
+		// Combined requirements must have some possible value unless Operator=DoesNotExist.
+		if values := r.Get(requirement.Key); values.Len() == 0 && requirement.Operator != v1.NodeSelectorOpDoesNotExist {
+			errs = multierr.Append(errs, fmt.Errorf("no feasible value for key %s", requirement.Key))
 		}
 	}
 	return errs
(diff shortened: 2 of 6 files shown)


pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -90,16 +90,23 @@ func (r Requirements) Add(requirements ...v1.NodeSelectorRequirement) Requiremen
...
 		case v1.NodeSelectorOpExists:
-			r.requirements[requirement.Key] = r.Get(requirement.Key).Intersection(sets.NewComplementSet())
+			values = sets.NewComplementSet()
 		case v1.NodeSelectorOpDoesNotExist:
-			r.requirements[requirement.Key] = sets.NewSet()
+			values = sets.NewSet()
+		default:
+			values = sets.NewSet()
 		}
+		if existing, ok := r.requirements[requirement.Key]; ok {
```
ellistarn, line 106: This is the core bug. Before this check, values were intersected with the empty set, resulting in incorrect values.

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -128,12 +135,12 @@ func (r Requirements) Label(key string) string {
...
 
-// Get returns the sets of values allowed by all included requirements
-// following a denylist method. Values are allowed except specified
+func (r Requirements) Has(key string) bool {
+	_, ok := r.requirements[key]
+	return ok
+}
+
 func (r Requirements) Get(key string) sets.Set {
-	if _, ok := r.requirements[key]; !ok {
```
ellistarn, line 134: We no longer model nonexistance as the full set. The empty set and undefined set are both equivalent to {}.

pkg/apis/provisioning/v1alpha5/suite_test.go
```diff
@@ -314,20 +314,20 @@ var _ = Describe("Validation", func() {
 			B := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpDoesNotExist})
 			Expect(A.Compatible(B)).ToNot(Succeed())
 		})
-		It("A should fail to be compatible to B, <Exists, Empty> operator, conflicting", func() {
+		It("A should be compatible to B, <Exists, Empty> operator", func() {
 			A := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpExists})
 			B := NewRequirements()
-			Expect(A.Compatible(B)).ToNot(Succeed())
+			Expect(A.Compatible(B)).To(Succeed())
```
ellistarn, line 320: This test was incorrect. Pods are not required to have a nodeSelector in order to schedule onto a node.

pkg/apis/provisioning/v1alpha5/suite_test.go
```diff
@@ -314,20 +314,20 @@ var _ = Describe("Validation", func() {
...
 			A := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpDoesNotExist})
 			B := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpIn, Values: []string{"foo"}})
 			Expect(A.Compatible(B)).ToNot(Succeed())
 		})
-		It("A should fail to be compatible to B, <DoesNotExist, NotIn> operator, conflicting", func() {
+		It("A should be compatible to B, <DoesNotExist, NotIn> operator", func() {
 			A := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpDoesNotExist})
 			B := NewRequirements(v1.NodeSelectorRequirement{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpNotIn, Values: []string{"foo"}})
-			Expect(A.Compatible(B)).ToNot(Succeed())
+			Expect(A.Compatible(B)).To(Succeed())
```
ellistarn, line 330: According to the second bullet at https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement, DoesNotExist and NotIn are compatible. A Pod that says `foo not in bar` may schedule to a node without any label with key `foo`.

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -90,16 +90,23 @@ func (r Requirements) Add(requirements ...v1.NodeSelectorRequirement) Requiremen
...
-			r.requirements[requirement.Key] = r.Get(requirement.Key).Intersection(sets.NewComplementSet(requirement.Values...))
+			values = sets.NewComplementSet(requirement.Values...)
 		case v1.NodeSelectorOpExists:
-			r.requirements[requirement.Key] = r.Get(requirement.Key).Intersection(sets.NewComplementSet())
+			values = sets.NewComplementSet()
 		case v1.NodeSelectorOpDoesNotExist:
-			r.requirements[requirement.Key] = sets.NewSet()
+			values = sets.NewSet()
+		default:
+			values = sets.NewSet()
```
tzneal, line 104: This default case should never happen given what we currently allow, right?

pkg/utils/sets/sets.go
```diff
@@ -53,14 +54,22 @@ func (s Set) DeepCopy() Set {
...
+		if s.Len() < math.MaxInt64 {
+			return v1.NodeSelectorOpNotIn
+		}
+		return v1.NodeSelectorOpExists
+	}
+	if s.Len() > 0 {
+		return v1.NodeSelectorOpIn
+	}
+	return v1.NodeSelectorOpDoesNotExist
 }
```
tzneal, line 73: Neat!

tzneal (approved):

pkg/apis/provisioning/v1alpha5/requirements.go
ellistarn: Correct. I'll rip this out. I only needed it when this was a helper function.

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -173,82 +178,29 @@ func (r Requirements) Validate() (errs error) {
...
-		if requirement.Operator == v1.NodeSelectorOpDoesNotExist && (r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpIn)) ||
-			r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpNotIn)) ||
-			r.hasRequirement(withKeyAndOperator(requirement.Key, v1.NodeSelectorOpExists))) {
-			errs = multierr.Append(errs, fmt.Errorf("operator %s cannot coexist with other operators for key %s", v1.NodeSelectorOpDoesNotExist, requirement.Key))
-		}
-	}
-	for key := range r.Keys() {
-		if r.Get(key).Len() == 0 && !r.hasRequirement(withKeyAndOperator(key, v1.NodeSelectorOpDoesNotExist)) {
-			errs = multierr.Append(errs, fmt.Errorf("no feasible value for key %s", key))
-		}
```
ellistarn, line 186: These correctness checks are deferred to `Compatible()`. Since this is a tight loop, this will improve performance, but more importantly, it enables us to decouple validation of the API `[]v1.NodeSelectorRequirement` from the implementation of the set math used at runtime.

pkg/apis/provisioning/v1alpha5/suite_test.go
```diff
@@ -185,13 +185,6 @@ var _ = Describe("Validation", func() {
 				Expect(provisioner.Validate(ctx)).To(Succeed())
 			}
 		})
-		It("should fail because no feasible value", func() {
```
ellistarn, line 188: This is no longer enforced at the API level, and is caught during compatibility checks.

tzneal (approved):

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -173,82 +178,33 @@ func (r Requirements) Validate() (errs error) {
...
+		if values := r.Get(requirement.Key); values.Len() == 0 && requirement.Operator != v1.NodeSelectorOpDoesNotExist {
+			errs = multierr.Append(errs, fmt.Errorf("no feasible value for key %s", requirement.Key))
 		}
 	}
 	return errs
 }
 
 // Compatible ensures the provided requirements can be met.
-//gocyclo:ignore
 func (r Requirements) Compatible(requirements Requirements) (errs error) {
```
felix-zhe-huang, line 190: This will fail the following test. 
```
		It("should schedule the pod with DoesNotExists operator and a provisioner with NotIn operator", func() {
			provisioner.Spec.Requirements = v1alpha5.NewRequirements(
				v1.NodeSelectorRequirement{Key: "test-key", Operator: v1.NodeSelectorOpNotIn, Values: []string{"test-value"}})
			pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner, test.UnschedulablePod(
				test.PodOptions{NodeRequirements: []v1.NodeSelectorRequirement{
					{Key: "test-key", Operator: v1.NodeSelectorOpDoesNotExist},
				}},
			))[0]
		})
```

felix-zhe-huang: I think we should treat the set values as what it is, it is a set of possible node labels.
Using this logic, We basically just need to identify cases in which Karpenter will not generated a label for the node.
I am proposing this. 
After tightening with pods constraints, generate no label for node:
- When final set is empty (either because key is undefined or intersect with DoesNotExist)
- When the final set is complement but not full set (case of NotIn)

With the above definition, the compatibility check will fall back to the kubernetes logic. [ https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement,]( https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement,) indicates that pods with NotIn and DoesNotExist will be able to run on nodes without the corresponding label. In that case, the compatibility check should return error for all the cases with `requirement.Intersection(r.Get(key)).Len() == 0` except the following cases:

- r is Undefined, requirement.Operator==NotIn
- r is Undefined, requirement.Operator==DoesNotExist
- r is DoesNotExist, requirement.Operator==NotIn
- r is DoesNotExist, requirement.Operator==DoesNotExist
- r is NotIn, requirement.Operator==NotIn
- r is NotIn, requirement.Operator==DoesNotExist


```
if Len()==0 {

}
```

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -189,16 +189,17 @@ func (r Requirements) Validate() (errs error) {
...
-		if requirement.Type() == v1.NodeSelectorOpExists && requirement.Intersection(r.Get(key)).Len() == 0 {
+		if requirement.Type() == v1.NodeSelectorOpExists && intersection.Len() == 0 {
 			errs = multierr.Append(errs, fmt.Errorf("%s not in %s, key %s", requirement, r.Get(key), key))
 		}
-		if requirement.Type() == v1.NodeSelectorOpNotIn && r.Get(key).Type() == v1.NodeSelectorOpIn && requirement.Intersection(r.Get(key)).Len() == 0 {
+		if requirement.Type() == v1.NodeSelectorOpNotIn && r.Get(key).Type() == v1.NodeSelectorOpIn && intersection.Len() == 0 {
 			errs = multierr.Append(errs, fmt.Errorf("%s not in %s, key %s", requirement, r.Get(key), key))
 		}
-		if requirement.Type() == v1.NodeSelectorOpDoesNotExist && r.Get(key).Len() > 0 {
+		if requirement.Type() == v1.NodeSelectorOpDoesNotExist && r.Get(key).Type() == v1.NodeSelectorOpIn {
```
felix-zhe-huang, line 202: I think we should also raise error for r.Get(key).Type() == v1.NodeSelectorOpExists

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -76,7 +76,7 @@ func (c *Constraints) GenerateLabels() map[string]string {
 			case v1.NodeSelectorOpIn:
 				labels[key] = c.Requirements.Get(key).Values().UnsortedList()[0]
 			case v1.NodeSelectorOpExists, v1.NodeSelectorOpNotIn:
-				labels[key] = rand.String(10)
+				labels[key] = rand.String(10) // At size of the universe odds, NotIn could conflict
```
felix-zhe-huang, line 79: Can we resolve the conflict? We use to do it with `for(r.Get(key).has(label)){}`.

felix-zhe-huang: Or better yet, why bother assigning a node label for NotIn? We should just ignore it and k8s will have no problem assigning the pod to the node.

felix-zhe-huang:
I love this new requirement logics! So learn. I would like to go over the unit tests again later according to these new logics to make sure that nothing is missed.

pkg/apis/provisioning/v1alpha5/constraints.go
ellistarn: Wow even better 🤯

felix-zhe-huang:
Love it! Ship it!

felix-zhe-huang (approved):
Love it! Ship it!

PR #1550, opened 2022-03-20
fix bug with pod node affinity testing

tzneal:
**1. Issue, if available:**
N/A

Adds a test and fixes a bug related to not checking pod node affinity.

Unit test + on EKS

- [X] No

### DIFF
--- a/pkg/controllers/provisioning/scheduling/node.go
+++ b/pkg/controllers/provisioning/scheduling/node.go
@@ -19,8 +19,6 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 
-	"github.com/aws/karpenter/pkg/utils/sets"
-
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/utils/resources"
@@ -37,33 +35,12 @@ type Node struct {
 func NewNode(constraints *v1alpha5.Constraints, instanceTypeOptions []cloudprovider.InstanceType, pods ...*v1.Pod) *Node {
 	n := &Node{Constraints: constraints.DeepCopy()}
 
-	instanceTypes := constraints.Requirements.InstanceTypes()
 	for _, it := range instanceTypeOptions {
-		// provisioner must list the instance type by name (defaults to all instance types)
-		if !instanceTypes.Has(it.Name()) {
+		// pre-filter our list of all possible instance types by what the provisioner allows
+		if !compatibleInstanceType(constraints.Requirements, it) {
 			continue
 		}
-
-		// provisioner constraints can't exclude the instance type by arch or OS
-		if !constraints.Requirements.Get(v1.LabelArchStable).Has(it.Architecture()) {
-			continue
-		}
-		if constraints.Requirements.Get(v1.LabelOSStable).Intersection(
-			sets.NewSet(it.OperatingSystems().List()...)).Len() == 0 {
-			continue
-		}
-
-		included := false
-		// and the instance type must have some valid offering combination per the provisioner constraints
-		for _, off := range it.Offerings() {
-			if constraints.Requirements.Zones().Has(off.Zone) && constraints.Requirements.CapacityTypes().Has(off.CapacityType) {
-				included = true
-				break
-			}
-		}
-		if included {
-			n.InstanceTypeOptions = append(n.InstanceTypeOptions, it)
-		}
+		n.InstanceTypeOptions = append(n.InstanceTypeOptions, it)
 	}
 	// technically we might create some invalid construct here if the pod can't be supported by any instance type
 	// but this is ok as we just won't have any valid instance types, this pod won't be schedulable and nothing else
@@ -80,14 +57,14 @@ func (n Node) Compatible(pod *v1.Pod) error {
 		return err
 	}
 
+	tightened := n.Constraints.Requirements.Add(podRequirements.Requirements...)
 	// Ensure that at least one instance type of the instance types that we are already narrowed down to based on the
-	// existing pods can support the pod
+	// existing pods can support the pod resources and combined pod + provider requirements
 	for _, it := range n.InstanceTypeOptions {
-		if n.isPodCompatible(pod, it) {
+		if compatibleInstanceType(tightened, it) && n.hasCompatibleResources(resources.RequestsForPods(pod), it) {
 			return nil
 		}
 	}
-
 	return errors.New("no matching instance type found")
 }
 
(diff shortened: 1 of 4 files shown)


pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -184,3 +164,41 @@ func (n Node) hasCompatibleResources(resourceList v1.ResourceList, it cloudprovi
...
+		}
+	}
+	return true
+}
+
+func compatibleInstanceType(requirements v1alpha5.Requirements, it cloudprovider.InstanceType) bool {
+	if !requirements.Get(v1.LabelInstanceTypeStable).Has(it.Name()) {
+		return false
+	}
+
```
ellistarn, line 187:
```suggestion
```

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -184,3 +164,41 @@ func (n Node) hasCompatibleResources(resourceList v1.ResourceList, it cloudprovi
...
+	if !requirements.Get(v1.LabelArchStable).Has(it.Architecture()) {
+		return false
+	}
+	if requirements.Get(v1.LabelOSStable).Intersection(
+		sets.NewSet(it.OperatingSystems().List()...)).Len() == 0 {
+		return false
+	}
+
```
ellistarn, line 195:
```suggestion
```

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -184,3 +164,39 @@ func (n Node) hasCompatibleResources(resourceList v1.ResourceList, it cloudprovi
...
+}
+
+func compatibleInstanceType(requirements v1alpha5.Requirements, it cloudprovider.InstanceType) bool {
+	if !requirements.Get(v1.LabelInstanceTypeStable).Has(it.Name()) {
+		return false
+	}
+	if !requirements.Get(v1.LabelArchStable).Has(it.Architecture()) {
+		return false
+	}
+	if requirements.Get(v1.LabelOSStable).Intersection(
```
ellistarn, line 190: Can't you just do 

```!it.OperatingSystems().HasAny(requirements.Get(v1.LabelOSStable))```

File: pkg/controllers/provisioning/scheduling/node.go [hunk 4]
```diff
@@ -184,3 +164,39 @@ func (n Node) hasCompatibleResources(resourceList v1.ResourceList, it cloudprovi
...
+		if term.MatchExpressions != nil {
+			if !compatibleInstanceType(v1alpha5.NewRequirements(term.MatchExpressions...), it) {
+				return false
+			}
+		}
+	}
+	return true
+}
+
+func compatibleInstanceType(requirements v1alpha5.Requirements, it cloudprovider.InstanceType) bool {
```
ellistarn, line 183: Your node object already has `Constraints.Requirements`, and you can avoid the static.
```suggestion
func (n *Node) hasCompatibleInstanceType(it cloudprovider.InstanceType) bool {
```

tzneal: It's also used with a generated set of requirements from `hasCompatibleNodeAffinity`

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -184,3 +164,39 @@ func (n Node) hasCompatibleResources(resourceList v1.ResourceList, it cloudprovi
...
+	}
+	if !requirements.Get(v1.LabelArchStable).Has(it.Architecture()) {
+		return false
+	}
+	if requirements.Get(v1.LabelOSStable).Intersection(
+		sets.NewSet(it.OperatingSystems().List()...)).Len() == 0 {
+		return false
+	}
+	included := true
+	for _, off := range it.Offerings() {
```
ellistarn, line 195:
```suggestion
	for _, offering := range it.Offerings() {
```

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -155,7 +134,8 @@ func (n Node) hasCompatibleNodeSelector(nodeSelector map[string]string, it cloud
 
 func (n Node) isPodCompatible(pod *v1.Pod, it cloudprovider.InstanceType) bool {
 	return n.hasCompatibleResources(resources.RequestsForPods(pod), it) &&
-		n.hasCompatibleNodeSelector(pod.Spec.NodeSelector, it)
+		n.hasCompatibleNodeSelector(pod.Spec.NodeSelector, it) &&
+		n.hasCompatibleNodeAffinity(pod.Spec.Affinity, it)
```
ellistarn, line 138: Aren't these checks already modeled by `Requirements`? Does this disappear when we combine scheduling/binpacking? In the mean time, can you use PodRequirements(pod) to handle the selector and affinity parsing?

pkg/controllers/provisioning/scheduling/node.go
tzneal: I don't think so, the requirement may not specify an OS in which case any are ok.  We'd have to call .Values() on the requirements version which would panic due to calling values on a complement set.  The two different set types here are making the code more ugly than it should be.

pkg/controllers/provisioning/scheduling/node.go
tzneal: Yup 😞

pkg/controllers/provisioning/scheduling/node.go
ellistarn: Another option is to implement HasAny on our specialized set type that encapsulates this. Totally optional.

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -80,14 +58,14 @@ func (n Node) Compatible(pod *v1.Pod) error {
 		return err
 	}
 
+	tightened := n.Constraints.Requirements.Add(podRequirements.Requirements...)
```
ellistarn, line 61: It feels like we're tripping over ourselves a bit. This is a tight loop that scales linearly with pods. We're already adding these requirements together inside `n.Constraints.Requirements.Compatible(podRequirements)`, and then we're doing it again here. Effectively, we have two areas of the code base responsible for the same thing (e.g. checking if a pod can schedule to a node), but done at different levels of abstraction.

As you've rightly done, we don't want to push this logic into the API package to create a dependency between `v1alpha5 -> cloudprovider`. Instead, if we could move the compatibility logic up to this layer, I *think* a lot of this impedance mismatch goes away. Obviously, we're blocked on this right now due to https://github.com/aws/karpenter/pull/1549, but it might be worth exploring once that's merged. 

I'm happy to merge this PR as it is currently, but would love to remove this rough edge for simplicity/performance.

pkg/controllers/provisioning/scheduling/node.go
tzneal: Good call, I didn't want to change much when fixing it for 0.7.2, but I've got the time now to make a bit less ugly.

pkg/controllers/provisioning/scheduling/node.go
tzneal: I was debating if the initial compatibility check was required when I wrote this and ended up with the conclusion that I wasn't certain that our Add() was written to handle possibly incompatible requirements, so it felt safer to keep it in.

tzneal: Just verified that the outer compatibility check handles the stuff that's not instance related (e.g. hostname requirements to implement topology spread).

ellistarn: Not currently. You'd need to smoosh em together :D

ellistarn (approved):

PR #1555, opened 2022-03-21
our price ordering no longer provides the guarantee that made this work

tzneal:
**1. Issue, if available:**
N/A

Remove packing optimization that doesn't work with arbitrary price sorting.

Unit tests & by hand on EKS

- [ ] No

### DIFF
--- a/pkg/controllers/provisioning/binpacking/packer.go
+++ b/pkg/controllers/provisioning/binpacking/packer.go
@@ -170,34 +170,19 @@ func (p *Packer) packWithLargestPod(unpackedPods []*v1.Pod, packables []*Packabl
 		return &Packing{Pods: [][]*v1.Pod{bestPackedPods}, InstanceTypeOptions: bestInstances}, remainingPods
 	}
 
-	for i, packable := range packables {
+	for _, packable := range packables {
 		// check how many pods we can fit with the available capacity
 		if result := packable.Pack(unpackedPods); len(result.packed) == maxPodsPacked {
-			// Add all packable nodes that have more resources than this one
-			// Trim the bestInstances so that provisioning APIs in cloud providers are not overwhelmed by the number of instance type options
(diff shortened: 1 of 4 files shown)


pkg/controllers/provisioning/binpacking/packer.go
```diff
@@ -170,34 +170,19 @@ func (p *Packer) packWithLargestPod(unpackedPods []*v1.Pod, packables []*Packabl
...
-				if lvMem.Cmp(rvMem) <= 0 && lvPods.Cmp(rvPods) <= 0 {
-					bestInstances = append(bestInstances, packables[j])
-				}
-			}
+			bestInstances = append(bestInstances, packable)
 			bestPackedPods = result.packed
 			remainingPods = result.unpacked
-			break
+			// reached the max that we care about
+			if len(bestInstances) == MaxInstanceTypes {
```
bwagner5, line 180: since we filter out instance types in the CP now, it might make sense to just move the MaxInstanceTypes to the CP or increase this value.   It might make the bin-packer do a lot more work though...

ellistarn: Alternatively, build this on a branch and whole hog move towards a unified scheduler/binpacker.

tzneal: That's the next step, but right now it sometimes creates instance types that are too small. Was hoping to get this in as a patch before any release.

tzneal: > since we filter out instance types in the CP now, it might make sense to just move the MaxInstanceTypes to the CP or increase this value.   It might make the bin-packer do a lot more work though...  

I think it will with a converged scheduler/bin packer. I think scheduling costs will dominate any bin packing cost anyway.

bwagner5 (approved):
/lgtm

PR #1577, opened 2022-03-25
Nightly release

spring1843:
**1. Issue, if available:**

Runs everyday half an hour after other scheduled daily crons (https://github.com/aws/karpenter/pull/1576)
Releases an image into the karpenter-nightly repository.

Remaining things to do:
1- Add a "Latest" tag
2- Add usage and about docs to the repo
3- Upload a logo to the registry
4- Add an alias in the registry for a friendlier repo name
5- Possibly publish helm charts as well
6- Add the version tag if possible as well

- [ ] No


Makefile
```diff
@@ -77,6 +77,9 @@ codegen: ## Generate code. Must be run if changes are made to ./pkg/apis/...
 release: ## Generate release manifests and publish a versioned container image.
 	$(WITH_GOFLAGS) ./hack/release.sh
 
+nightly: toolchain## Generate nightly release manifests and publish a versioned container image.
+	$(WITH_GOFLAGS) ./hack/nightly.sh
```
ellistarn, line 81: Can we collapse this with the release script?

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,29 @@
+name: Nightly
+on:
+  schedule:
+    - cron: '30 12 * * *'
```
njtran, line 4: Can you comment what time this would be? Is it 12:30 GMT? Do we want to run this on weekends too? Will need to make sure we're on top of failures on the weekend if they arise.

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,29 @@
...
+permissions:
+  id-token: write
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
```
njtran, line 14: Since 1.18 came out, we may need to change this in the near future. Is it possible to query `go.mod` to see which version we're doing?

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,29 @@
...
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+            ~/go/bin/
+            ~/.kubebuilder/bin/k8s
+          key: ${{ runner.os }}-go-cache-${{ hashFiles('**/go.sum') }}
+      - uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: arn:aws:iam::112855948879:role/nightly
+          aws-region: us-east-1
```
njtran, line 26: Is this your role and region? Can we do us-west-2? Should we have builds for different regions? Or is this because of ecr-public?

hack/nightly.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -e
+NIGHTLY_TAG_FMT="+%m%d%y"
+RELEASE_REPO=${RELEASE_REPO:-public.ecr.aws/z3c6l6z9/karpenter-nightly}
+RELEASE_VERSION=${RELEASE_VERSION:-$(date "${NIGHTLY_TAG_FMT}")}
```
njtran, line 4: Do we want to include what the most recent release version was here? Might be easier for tracking rather than having to cross reference release dates and nightly date tags.

njtran (commented):
Nice work!

.github/workflows/nightly.yaml
spring1843: that's something to change in multiple places not planning to include in this PR.

hack/nightly.sh
spring1843: The tag needs to remain guessable, e.g. a simple date, adding more to it would make that impossible. It's easy to find what version was correlated since all version tags have a tag.

.github/workflows/nightly.yaml
spring1843: This is a well known cron tab format, you can use things like https://crontab.guru/ to translate. Not sure about the weekend, but technically since all merges pass build this should never happen. And if something happens it's better to have nightlies for the other days. Since we may merge code during the weekend it's important to have them build every day.

.github/workflows/nightly.yaml
spring1843: It's the default region, the role is made through the infra piece.

spring1843: We don't need to have ECRs in each region, what's the reason for using us-west-2? this is already made and tested it's not super easy to change it. us-east-1 as the default region has a lot of capacity and is arguably better geographically positioned to serve the world although the outcome is most likely cached.

hack/nightly.sh
spring1843: I wonder if we can add multiple tags, we can revisit this when I add the Latest tags as part of the remaining things to do.

Makefile
spring1843: Will do in a follow up PR

ellistarn (approved):

spring1843:
@njtran thank you for your comments lets continue the conversation on these concerns, I'm gonna incorporate them in the following PRs I'm working on.

.github/workflows/nightly.yaml
spring1843: done here https://github.com/aws/karpenter/pull/1578 good suggestion

.github/workflows/nightly.yaml
bwagner5: just to note, us-east-1 is the only region where ecr-public is available anyways, but ecr-public replicates out the images to its CDN.

PR #1589, opened 2022-03-29
Rename role and simplify

spring1843:
**1. Issue, if available:**

* s/nightly/Nightly
* Removes extra lines from nightly.sh

- [ ] No

### DIFF
--- a/hack/nightly.sh
+++ b/hack/nightly.sh
@@ -4,6 +4,11 @@ RELEASE_REPO=${RELEASE_REPO:-public.ecr.aws/z3c6l6z9/karpenter-nightly}
 RELEASE_VERSION=${RELEASE_VERSION:-$(date "${NIGHTLY_TAG_FMT}")}
 RELEASE_PLATFORM="--platform=linux/amd64,linux/arm64"
 
+if [ -z "$CLOUD_PROVIDER" ]; then
+    echo "CLOUD_PROVIDER environment variable is not set: 'export CLOUD_PROVIDER=aws'"
+    exit 1
+fi
+
 image() {
     aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin ${RELEASE_REPO}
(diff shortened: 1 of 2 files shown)


hack/nightly.sh
```diff
@@ -4,19 +4,6 @@ RELEASE_REPO=${RELEASE_REPO:-public.ecr.aws/z3c6l6z9/karpenter-nightly}
 RELEASE_VERSION=${RELEASE_VERSION:-$(date "${NIGHTLY_TAG_FMT}")}
 RELEASE_PLATFORM="--platform=linux/amd64,linux/arm64"
 
-if [ -z "$CLOUD_PROVIDER" ]; then
-    echo "CLOUD_PROVIDER environment variable is not set: 'export CLOUD_PROVIDER=aws'"
-    exit 1
-fi
```
njtran, line 10: Why do we want to remove this? We'd want to know if no cloud provider is set so we don't silently default to a mock environment.

hack/nightly.sh
```diff
@@ -4,19 +4,6 @@ RELEASE_REPO=${RELEASE_REPO:-public.ecr.aws/z3c6l6z9/karpenter-nightly}
...
-
-# TODO restore https://reproducible-builds.org/docs/source-date-epoch/
```
njtran, line 12: Are you removing this because it's been completed?

spring1843: No those were copied from release.sh where there's cosign they are not used in here.

hack/nightly.sh
spring1843: this was removed from https://github.com/aws/karpenter/pull/1587 and it's not needed in the nightly as there's no mock env.

bwagner5: the same thing could happen in the nightly release. If the env var isn't set, Karpenter will build with our "fake" cloud provider. IMO this should be added back to the nightly script.

spring1843: Alright added back.

bwagner5 (approved):
lgtm!

PR #1591, opened 2022-03-29
Collapse release and nightly scripts into one

spring1843:
**1. Issue, if available:**

Collapse release.sh and nightly.sh into one

- [ ] No


hack/release.sh
```diff
@@ -45,6 +53,19 @@ website() {
...
+   injectAndCosign
+   chart
+   website
+}
+
+if [ "$1" = "nightly" ]; then
+  nightly
+else
+  release
+fi
```
ellistarn, line 71: In my mind, release nightly and release should be identical, except for the tag on the image/helm chart. e.g. nightlies should be signed, come with a version of the chart, and perhaps even w/ docs (thought that's covered via preview).

I think this is challenging right now because:
1. the helm chart is hosted in the git repo
2. cosign doesn't work from a GHA (IIUC, maybe it actually does?)

Happy to punt for now if this is out of scope, but I agree with your statement that this script feels brittle as is.

spring1843:
Closing this for now since we both agree as it stands the separation makes more sense.

Issue #1180, opened 2022-01-19
Failed to provision new node


alfianabdi:
## Version

**Karpenter**: v0.5.3

**Kubernetes**: v1.21.2-eks-0389ca3

## Expected Behavior

A new node is provisioned when there are pending pods.

## Actual Behavior

New node is not provisioned. The controller threw these errors
```
2022-01-19T02:44:37.678Z        DEBUG   controller.provisioning Excluding instance type m5a.2xlarge because there are not enough resources for daemons {"commit": "5047f3c", "provisioner": "linux-amd64"}
2022-01-19T02:44:37.680Z        ERROR   controller.provisioning Failed to find instance type option(s) for [default/inflate-6b88c9fb68-wlzxv default/inflate-6b88c9fb68-gbnmk] {"commit": "5047f3c", "provisioner": "linux-amd64"}
```

## Steps to Reproduce the Problem

Create provisioner and deployment as in resource specs

## Resource Specs and Logs

Provisioner
```
apiVersion: karpenter.sh/v1alpha5
kind: Provisioner
metadata:
  name: linux-amd64
spec:
  kubeletConfiguration: {}
  limits: {}
  provider:
    apiVersion: extensions.karpenter.sh/v1alpha1
    instanceProfile: <Redacted>
    kind: AWS
    launchTemplate: Linux-Amd64-WithRole
    securityGroupSelector:
      karpenter.sh/cluster/DevEKS: owned
      kubernetes.io/cluster/DevEKS: '*'
    subnetSelector:
      karpenter.sh/cluster/DevEKS: shared
  requirements:
  - key: node.kubernetes.io/instance-type
    operator: In
    values:
    - m5.2xlarge
    - m5a.2xlarge
  - key: topology.kubernetes.io/zone
    operator: In
    values:
    - ap-southeast-1a
    - ap-southeast-1b
  - key: kubernetes.io/arch
    operator: In
    values:
    - amd64
  - key: karpenter.sh/capacity-type
    operator: In
    values:
    - on-demand
  ttlSecondsAfterEmpty: 30
  ttlSecondsUntilExpired: 2592000
```

Deployment
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: inflate
spec:
  replicas: 0
  selector:
    matchLabels:
      app: inflate
  template:
    metadata:
      labels:
        app: inflate
    spec:
      terminationGracePeriodSeconds: 0
      containers:
        - name: inflate
[...]

Issue #607, opened 2021-08-11
Document best practices for Resource Requests


ellistarn:
## Is an existing page relevant?
No.

## What karpenter features are relevant?
Provisioning

## How should the docs be improved?
We need to document that we expect pods to have resource requests defined. For pods missing these, this can be configured globally per namespace using https://kubernetes.io/docs/concepts/policy/limit-range/.

Community Note
Please vote on this issue by adding a 👍 reaction to the original issue to help the community and maintainers prioritize this request
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
If you are interested in working on this issue or have submitted a pull request, please leave a comment

geoffcline:
Making a note to document LimitRanges best practices https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/ to ensure that karpenter works well with pods that are missing resource requests?

chrisnegus:
I added some [troubleshooting content on LimitRanges](https://karpenter.sh/preview/troubleshooting/#unspecified-resource-requests-cause-schedulingbin-pack-failures) in [PR #1597](https://github.com/aws/karpenter/pull/1597). While this describes the problem, it does not provide guidance on exactly how to set LimitRange values. So I think we should keep this issue open and look for a place to address that.

PR #1597, opened 2022-03-30
Adding troubleshooting docs items

chrisnegus:
**1. Issue, if available:**
Added troubleshooting items related to [Issue #1084](https://github.com/aws/karpenter/issues/1084), [Issue #1180](https://github.com/aws/karpenter/issues/1180) and [Issue #607](https://github.com/aws/karpenter/issues/607), as well as several slack discussions.

The following troubleshooting entries were added:
* Daemonsets can result in deployment failures (@felix-zhe-huang, please review)
* Unspecified resource requests cause scheduling/bin-pack failures (@tzneal, please review)
* Missing discovery tags causing provisioning failures (@bwagner5, please review)

### DIFF
--- a/website/content/en/preview/troubleshooting.md
+++ b/website/content/en/preview/troubleshooting.md
@@ -115,8 +115,8 @@ See the Karpenter [Best Practices Guide](https://aws.github.io/aws-eks-best-prac
 
 ## Missing subnetSelector and securityGroupSelector tags causes provisioning failures
 
-Starting with Karpenter v0.5.5, provisioners require [subnetSelector](./aws/provisioning/#subnetselector) and [securityGroupSelector](./aws/provisioning/#securitygroupselector) tags be set to match your cluster.
-The [Provisioner](./getting-started/getting-started-with-eksctl/#provisioner) example in the Karpenter Getting Started Guide uses the following:
+Starting with Karpenter v0.5.5, provisioners require that [subnetSelector]({{<ref "./aws/provisioning/#subnetselector" >}}) and [securityGroupSelector]({{<ref "./aws/provisioning/#securitygroupselector" >}}) tags be set to match your cluster.
+The [Provisioner](./getting-started/getting-started-with-eksctl/#provisioner) section in the Karpenter Getting Started Guide uses the following example:
 
 ```text
 provider:
--- a/website/content/en/preview/troubleshooting.md
+++ b/website/content/en/preview/troubleshooting.md
@@ -120,6 +120,17 @@ provider:
     securityGroupSelector:
       karpenter.sh/discovery: ${CLUSTER_NAME}
 ```
+To check your subnet and security group selectors, type the following:
+
+```bash
+aws ec2 describe-subnets --filters Name=tag:karpenter.sh/discovery,Values=${CLUSTER_NAME}
+```
+*Returns subnets matching the selector*
+
+```bash
+aws ec2 describe-security-groups --filters Name=tag:karpenter.sh/discovery,Values=${CLUSTER_NAME}
+```
+*Returns security groups matching the selector*
 
 Provisioners created without those tags and run in more recent Karpenter versions will fail with this message when you try to run the provisioner:
 
--- a/website/content/en/preview/troubleshooting.md
+++ b/website/content/en/preview/troubleshooting.md
@@ -89,18 +89,13 @@ approach, and now it's much more restrictive.
 
 ## DaemonSets can result in deployment failures
 
-For Karpenter versions 0.5.3 and earlier, Daemonsets were not properly considered when provisioning nodes.
-This sometimes caused nodes to be deployed that could not meet the needs of the requested Daemonsets.
-The result could be log messages like the following:
+For Karpenter versions 0.5.3 and earlier, DaemonSets were not properly considered when provisioning nodes.
+This sometimes caused nodes to be deployed that could not meet the needs of the requested DaemonSets and workloads.
+This issue no longer occurs after Karpenter version 0.5.3 (see [PR #1155](https://github.com/aws/karpenter/pull/1155)).
 
-```text
-Excluding instance type r3.8xlarge because there are not enough resources for daemons {"commit": "7e79a67", "provisioner": "default"}
-```
-
-One workaround is to set your provisioner to only use larger instance types.
+If you are using a pre-0.5.3 version of Karpenter, one workaround is to set your provisioner to only use larger instance types that you know will be big enough for the DaemonSet and the workload.
 For more information, see [Issue #1084](https://github.com/aws/karpenter/issues/1084).
 Examples of this behavior are included in [Issue #1180](https://github.com/aws/karpenter/issues/1180).
-This issue was addressed in later Karpenter releases by [PR #1155](https://github.com/aws/karpenter/pull/1155).
 
 ## Unspecified resource requests cause scheduling/bin-pack failures
 
@@ -115,7 +110,7 @@ See the Karpenter [Best Practices Guide](https://aws.github.io/aws-eks-best-prac
 
 ## Missing subnetSelector and securityGroupSelector tags causes provisioning failures
 
-Starting with Karpenter v0.5.5, provisioners require that [subnetSelector]({{<ref "./aws/provisioning/#subnetselector" >}}) and [securityGroupSelector]({{<ref "./aws/provisioning/#securitygroupselector" >}}) tags be set to match your cluster.
+Starting with Karpenter v0.5.5, if you are using Karpenter-generated launch template, provisioners require that [subnetSelector]({{<ref "./aws/provisioning/#subnetselector" >}}) and [securityGroupSelector]({{<ref "./aws/provisioning/#securitygroupselector" >}}) tags be set to match your cluster.
 The [Provisioner](./getting-started/getting-started-with-eksctl/#provisioner) section in the Karpenter Getting Started Guide uses the following example:
 
 ```text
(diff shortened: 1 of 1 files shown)


website/content/en/preview/troubleshooting.md
```diff
@@ -86,3 +86,42 @@ Your security groups are not blocking you from reaching your webhook.
...
+```
+
+One workaround is to set your provisioner to only use larger instance types.
+For more information, see [Issue #1084](https://github.com/aws/karpenter/issues/1084).
+Examples of this behavior are included in [Issue #1180](https://github.com/aws/karpenter/issues/1180).
+This issue was addressed in later Karpenter releases by [PR #1155](https://github.com/aws/karpenter/pull/1155).
+
+## Unspecified resource requests cause scheduling/bin-pack failures
+
+Not setting Kubernetes [LimitRanges](https://kubernetes.io/docs/concepts/policy/limit-range/) on pods can cause Karpenter to fail to schedule or properly bin-pack pods.
```
tzneal, line 107:
```suggestion
Not using the Kubernetes [LimitRanges](https://kubernetes.io/docs/concepts/policy/limit-range/) feature to enforce minimum resource request sizes will allow pods with very low or non-existent resource requests to be scheduled.  This can cause issues as Karpenter bin-packs pods based on the resource requests.  If the resource requests do not reflect the actual resource usage of the pod, Karpenter will place too many of these pods onto the same node resulting in the pods getting CPU throttled or terminated due to the OOM killer.  This behavior is not unique to Karpenter and can also occur with the standard `kube-scheduler` with pods that don't have accurate resource requests.
```

tzneal: Do we have any best practice docs that recommend sizing of limits vs requests?

chrisnegus: @tzneal We have a general discussion on limits and pointers to Kubernetes docs on the subject in the [Karpenter Best Practices](https://aws.github.io/aws-eks-best-practices/karpenter/#use-limitranges-to-configure-defaults-for-resource-requests-and-limits) guide. Do you have suggestions for how to set particular limits?

chrisnegus: I incorporated your other comments.

tzneal: I don't know a good limit to request ratio value to recommend, probably worth punting on for now in that case as I can't find any recommendations from K8s or any existing EKS best practices either.

website/content/en/preview/troubleshooting.md
```diff
@@ -86,3 +86,47 @@ Your security groups are not blocking you from reaching your webhook.
...
+
+```text
+provider:
+    subnetSelector:
+      karpenter.sh/discovery: ${CLUSTER_NAME}
+    securityGroupSelector:
+      karpenter.sh/discovery: ${CLUSTER_NAME}
+```
+
+Provisioners created without those tags and run in more recent Karpenter versions will fail with this message when you try to run the provisioner:
```
bwagner5, line 128: This reads like those specific tags are required, but any tags can be used. We just use the `karpenter.sh/discovery` tags in the getting started guide

website/content/en/preview/troubleshooting.md
```diff
@@ -86,3 +86,47 @@ Your security groups are not blocking you from reaching your webhook.
 
 This is especially relevant if you have used `terraform-eks-module` version `>=18` since that version changed its security
 approach, and now it's much more restrictive.
+
+## Daemonsets can result in deployment failures
```
bwagner5, line 90:
```suggestion
## DaemonSets can result in deployment failures
```

website/content/en/preview/troubleshooting.md
```diff
@@ -86,3 +86,47 @@ Your security groups are not blocking you from reaching your webhook.
...
+
+For Karpenter versions 0.53 and earlier, daemonsets were not properly considered when provisioning nodes.
```
bwagner5, line 92:
```suggestion
For Karpenter versions 0.5.3 and earlier, daemonsets were not properly considered when provisioning nodes.
```

website/content/en/preview/troubleshooting.md
chrisnegus: Fixed.

website/content/en/preview/troubleshooting.md
chrisnegus: Good catch. Fixed.

website/content/en/preview/troubleshooting.md
```diff
@@ -86,3 +86,48 @@ Your security groups are not blocking you from reaching your webhook.
...
+
+If the resource requests do not reflect the actual resource usage of the pod, Karpenter will place too many of these pods onto the same node resulting in the pods getting CPU throttled or terminated due to the OOM killer.
+This behavior is not unique to Karpenter and can also occur with the standard `kube-scheduler` with pods that don't have accurate resource requests.
+
+To prevent this, you can set LimitRanges on pod deployments on a per-namespace basis.
+See the Karpenter [Best Practices Guide](https://aws.github.io/aws-eks-best-practices/karpenter/#use-limitranges-to-configure-defaults-for-resource-requests-and-limits) for further information on the use of LimitRanges.
+
+## Missing subnetSelector and securityGroupSelector tags causes provisioning failures
+
+Starting with Karpenter v0.5.5, provisioners require that [subnetSelector]({{<ref "./aws/provisioning/#subnetselector" >}}) and [securityGroupSelector]({{<ref "./aws/provisioning/#securitygroupselector" >}}) tags be set to match your cluster.
```
bwagner5, line 118: `securityGroupSelector` is only required when using Karpenter's generated launch templates. If a user provides a custom launch template, then specifying a `securityGroupSelector` will actually cause a validation error. `subnetSelector` is always required though.  

Other than that, lgtm!

website/content/en/preview/troubleshooting.md
```diff
@@ -86,3 +86,48 @@ Your security groups are not blocking you from reaching your webhook.
...
+
+## DaemonSets can result in deployment failures
+
+For Karpenter versions 0.5.3 and earlier, Daemonsets were not properly considered when provisioning nodes.
+This sometimes caused nodes to be deployed that could not meet the needs of the requested Daemonsets.
+The result could be log messages like the following:
+
+```text
+Excluding instance type r3.8xlarge because there are not enough resources for daemons {"commit": "7e79a67", "provisioner": "default"}
+```
```
felix-zhe-huang, line 98: This doesn't sound right to me. The problem for Daemon set before v0.5.3 is that Karpenter doesn't consider the resource requests of daemon sets when selecting instance type. Karpenter will launch a smaller node and pod will fail to run due to not enough resources. 

The above error log happens when Karpenter bin-packing is filtering instance types according to just the Daemon-set resource requests. It is not an indicator of the problem.

website/content/en/preview/troubleshooting.md
```diff
@@ -86,3 +86,48 @@ Your security groups are not blocking you from reaching your webhook.
...
+
+One workaround is to set your provisioner to only use larger instance types.
```
felix-zhe-huang, line 100: I will suggest to emphasize the workaround is for before v0.5.6.

website/content/en/preview/troubleshooting.md
```diff
@@ -86,3 +86,48 @@ Your security groups are not blocking you from reaching your webhook.
...
+For more information, see [Issue #1084](https://github.com/aws/karpenter/issues/1084).
+Examples of this behavior are included in [Issue #1180](https://github.com/aws/karpenter/issues/1180).
+This issue was addressed in later Karpenter releases by [PR #1155](https://github.com/aws/karpenter/pull/1155).
```
felix-zhe-huang, line 103: This is correct. It has been addressed. We will not see this problem anymore.

website/content/en/preview/troubleshooting.md
chrisnegus: Okay. I'll fix that. Thanks!

website/content/en/preview/troubleshooting.md
```diff
@@ -86,3 +86,45 @@ Your security groups are not blocking you from reaching your webhook.
...
+The [Provisioner](./getting-started/getting-started-with-eksctl/#provisioner) section in the Karpenter Getting Started Guide uses the following example:
+
+```text
+provider:
+    subnetSelector:
+      karpenter.sh/discovery: ${CLUSTER_NAME}
+    securityGroupSelector:
+      karpenter.sh/discovery: ${CLUSTER_NAME}
+```
+
```
bwagner5, line 123: what do you think about adding aws CLI commands for users to check their subnet and security group selectors? 

```
$ aws ec2 describe-subnets --filters Name=tag:karpenter.sh/discovery,Values=${CLUSTER_NAME}
.. Returns subnets matching the selector ...

$ aws ec2 describe-security-groups --filters Name=tag:karpenter.sh/discovery,Values=${CLUSTER_NAME}
.. Returns security groups matching the selector ...
```

chrisnegus (commented):
@felix-zhe-huang , @tzneal , @bwagner5 I've responded to all of the provided comments. Please let me know if further changes are required or provide /lgtm if everything looks okay.

website/content/en/preview/troubleshooting.md
chrisnegus: Sorry, I just saw this. Yes, I think it's a good idea and I'll add them.

chrisnegus: Done.

bwagner5 (approved):
lgtm! nicely written!

website/content/en/preview/troubleshooting.md
njtran: Looks like most instance types are [here](https://aws.amazon.com/ec2/instance-types/#General_Purpose) have a max  cpu:memory ratio of 1:4. GPUs have 1:8. If we need a ratio, 1:4 would be a good place to start thinking about it.

PR #1604, opened 2022-03-31
implement pod affinity & anti-affinity

tzneal:
**1. Issue, if available:**

- implement affinity/anti-affinity
- rework topology spread support

Unit tests and deployed to EKS

- [ ] No


tzneal:
I need to implement something that tracks pod's we've tried to schedule and failed so we don't repeatedly spam the log.

In this case, there is a zonal pod anti-affinity that effectively only lets us schedule as many pods as we have zones.

```
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:45.199Z	INFO	controller.provisioning	Waiting for unschedulable pods	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:51.201Z	INFO	controller.provisioning	Batched 2 pods in 1.000427455s	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:51.205Z	ERROR	controller.provisioning	Scheduling pod, unsatisfiable pod anti-affinity topology constraint for key topology.kubernetes.io/zone	{"commit": "b65bfd3", "provisioner": "default", "pod": "default/inflate-anti-self-zone-54f6d7f85b-9nwqr"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:51.205Z	ERROR	controller.provisioning	Scheduling pod, unsatisfiable pod anti-affinity topology constraint for key topology.kubernetes.io/zone	{"commit": "b65bfd3", "provisioner": "default", "pod": "default/inflate-anti-self-zone-54f6d7f85b-npdlb"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:51.205Z	ERROR	controller.provisioning	Failed to schedule 2 pods	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:51.205Z	INFO	controller.provisioning	Waiting for unschedulable pods	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:57.206Z	INFO	controller.provisioning	Batched 2 pods in 1.000494686s	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:57.287Z	DEBUG	controller.provisioning	Discovered subnets: [subnet-01caedfa832765807 (us-west-2b) subnet-08b04a750cf84d65e (us-west-2a) subnet-0527a666024f87542 (us-west-2b) subnet-0c85abf88a452b4c5 (us-west-2c) subnet-0f102b1ea0880fa0d (us-west-2a) subnet-0c5e0a8bdec952b40 (us-west-2c)]	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:57.291Z	ERROR	controller.provisioning	Scheduling pod, unsatisfiable pod anti-affinity topology constraint for key topology.kubernetes.io/zone	{"commit": "b65bfd3", "provisioner": "default", "pod": "default/inflate-anti-self-zone-54f6d7f85b-9nwqr"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:57.291Z	ERROR	controller.provisioning	Scheduling pod, unsatisfiable pod anti-affinity topology constraint for key topology.kubernetes.io/zone	{"commit": "b65bfd3", "provisioner": "default", "pod": "default/inflate-anti-self-zone-54f6d7f85b-npdlb"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:57.291Z	ERROR	controller.provisioning	Failed to schedule 2 pods	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:33:57.291Z	INFO	controller.provisioning	Waiting for unschedulable pods	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:34:03.292Z	INFO	controller.provisioning	Batched 2 pods in 1.000388795s	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:34:03.296Z	ERROR	controller.provisioning	Scheduling pod, unsatisfiable pod anti-affinity topology constraint for key topology.kubernetes.io/zone	{"commit": "b65bfd3", "provisioner": "default", "pod": "default/inflate-anti-self-zone-54f6d7f85b-9nwqr"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:34:03.296Z	ERROR	controller.provisioning	Scheduling pod, unsatisfiable pod anti-affinity topology constraint for key topology.kubernetes.io/zone	{"commit": "b65bfd3", "provisioner": "default", "pod": "default/inflate-anti-self-zone-54f6d7f85b-npdlb"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:34:03.296Z	ERROR	controller.provisioning	Failed to schedule 2 pods	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:34:03.296Z	INFO	controller.provisioning	Waiting for unschedulable pods	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:34:09.297Z	INFO	controller.provisioning	Batched 2 pods in 1.00064683s	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:34:09.309Z	ERROR	controller.provisioning	Scheduling pod, unsatisfiable pod anti-affinity topology constraint for key topology.kubernetes.io/zone	{"commit": "b65bfd3", "provisioner": "default", "pod": "default/inflate-anti-self-zone-54f6d7f85b-npdlb"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:34:09.309Z	ERROR	controller.provisioning	Scheduling pod, unsatisfiable pod anti-affinity topology constraint for key topology.kubernetes.io/zone	{"commit": "b65bfd3", "provisioner": "default", "pod": "default/inflate-anti-self-zone-54f6d7f85b-9nwqr"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:34:09.309Z	ERROR	controller.provisioning	Failed to schedule 2 pods	{"commit": "b65bfd3", "provisioner": "default"}
karpenter-74c8b6c9d-j8rrj controller 2022-03-31T16:34:09.309Z	INFO	controller.provisioning	Waiting for unschedulable pods	{"commit": "b65bfd3", "provisioner": "default"}
```

pkg/utils/pod/scheduling.go
```diff
@@ -68,16 +68,23 @@ func IsOwnedBy(pod *v1.Pod, gvks []schema.GroupVersionKind) bool {
...
-	return pod.Spec.Affinity.PodAffinity != nil &&
-		(len(pod.Spec.Affinity.PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution) != 0)
-}
-
 // HasRequiredPodAntiAffinity returns true if a non-empty PodAntiAffinity/RequiredDuringSchedulingIgnoredDuringExecution
 // is defined in the pod spec
 func HasRequiredPodAntiAffinity(pod *v1.Pod) bool {
-	return pod.Spec.Affinity.PodAntiAffinity != nil &&
+	return pod.Spec.Affinity != nil && pod.Spec.Affinity.PodAntiAffinity != nil &&
 		(len(pod.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution) != 0)
```
njtran, line 75: Could this work?
```suggestion
	return HasPodAntiAffinity(pod) &&
		len(pod.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution) != 0
```

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -28,42 +32,67 @@ import (
...
+	n.Constraints.Requirements = n.Constraints.Requirements.Add(v1.NodeSelectorRequirement{
+		Key:      v1.LabelHostname,
+		Operator: v1.NodeSelectorOpIn,
+		Values:   []string{n.Name},
+	})
+	return n
 }
 
-func (n *Node) Add(pod *v1.Pod) error {
+func (n *Node) addPod(pod *v1.Pod, commit bool) error {
```
njtran, line 63: It looks like `commit` here is used to show distinguish between a function call that is meant to actually bind the pod to the node but the other just checks constraints. Can you separate this function into one which checks the constraints, and one which binds the pod? It seems weird to me to add a boolean to distinguish when a function will actually do something vs when it's doing a check.

njtran (commented):
I'll resume looking through this tomorrow. Looks nice!

pkg/test/environment.go
```diff
@@ -15,8 +15,12 @@ package test
 import (
 	"context"
 	"fmt"
+	"os"
+	"strings"
 	"sync"
 
+	"github.com/Masterminds/semver"
```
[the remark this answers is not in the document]
tzneal: Yup, didn't know that package existed, thanks!

pkg/controllers/provisioning/scheduling/node.go
tzneal: I reworked it a bit, maybe somewhat cleaner.

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -76,3 +102,583 @@ func (s *NodeSet) getDaemons(ctx context.Context, constraints *v1alpha5.Constrai
 func (s *NodeSet) Add(node *Node) {
 	s.nodes = append(s.nodes, node)
 }
+
+func (s *NodeSet) SchedulePod(ctx context.Context, pod *v1.Pod) error {
```
ellistarn, line 106:
```suggestion
func (s *NodeSet) Schedule(ctx context.Context, pod *v1.Pod) error {
```

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -28,14 +39,29 @@ import (
 )
 
 type NodeSet struct {
```
ellistarn, line 41: It strikes me that NodeSet is a singleton per execution. Does it make sense to just fold this and the scheduler concept together?

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -28,14 +39,29 @@ import (
...
-	client          client.Client
+	kubeClient      client.Client
 	daemonResources v1.ResourceList
 	nodes           []*Node
+	topologies      map[uint64]*Topology
+	// Anti-affinity works both ways (if a zone has a pod foo with anti-affinity to a pod bar, we can't schedule bar to
+	// that zone, even though bar has no anti affinity terms on it. For this to work, we need to separately track the
+	// topologies of pods with anti-affinity terms, so we can prevent scheduling the pods they have anti-affinity to
+	// in some cases.
+	existingAntiAffinities map[uint64]*Topology
```
ellistarn, line 50: For naming parity with topologies 
```suggestion
	antiAffinities map[uint64]*Topology
```

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -76,3 +102,583 @@ func (s *NodeSet) getDaemons(ctx context.Context, constraints *v1alpha5.Constrai
...
 }
+
+func (s *NodeSet) SchedulePod(ctx context.Context, pod *v1.Pod) error {
+	// copy the pod as this method modifies the pod's node selectors to force it to meet topology constraints
+	pod = pod.DeepCopy()
+
+	var possibleNodes []*Node
+	for _, node := range s.nodes {
+		if err := node.Compatible(ctx, pod); err == nil {
+			possibleNodes = append(possibleNodes, node)
```
ellistarn, line 113: If we're convinced that we want to go with this approach, can we map to the kube-scheduler's idioms of `filter` and `score`

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -76,3 +102,583 @@ func (s *NodeSet) getDaemons(ctx context.Context, constraints *v1alpha5.Constrai
...
+		}
+	}
+
+	tightened := s.constraints.Requirements.Add(v1alpha5.NewPodRequirements(pod).Requirements...)
```
ellistarn, line 117: Isn't this duplicating requirements math with `node.Compatible`?

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -76,3 +102,583 @@ func (s *NodeSet) getDaemons(ctx context.Context, constraints *v1alpha5.Constrai
...
+	} else {
+		n := NewNode(s.kubeClient, s.constraints, s.daemonResources, s.instanceTypes)
+		if err := n.Add(ctx, pod); err != nil {
+			return fmt.Errorf("adding pod to node, %w", err)
+		}
+		s.nodes = append(s.nodes, n)
+	}
+
+	if err := s.recordTopologyDecisions(pod); err != nil {
+		s.logger.Errorf("recording topology decision, %s", err)
```
ellistarn, line 138: Why don't we use the logger from context?

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -76,3 +102,583 @@ func (s *NodeSet) getDaemons(ctx context.Context, constraints *v1alpha5.Constrai
...
+		}
+		tsc.AddOwner(client.ObjectKeyFromObject(p))
+	}
+
+	return nil
+}
+
+// buildNamespaceList constructs a unique list of namespaces consisting of the pod's namespace and the optional list of
+// namespaces and those selected by the namespace selector
+func buildNamespaceList(ctx context.Context, c client.Client, p *v1.Pod, namespaces []string, selector *metav1.LabelSelector) []string {
```
ellistarn, line 383: Let's find a way to associate this behavior with a concept. Perhaps as a member of nodeset?

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -76,3 +102,583 @@ func (s *NodeSet) getDaemons(ctx context.Context, constraints *v1alpha5.Constrai
...
+		s.logger.Errorf("recording topology decision, %s", err)
+	}
+
+	return nil
+}
+
+func (s *NodeSet) TrackTopologies(ctx context.Context, pods []*v1.Pod) error {
+	var errs error
+	errs = multierr.Append(errs, s.trackExistingAntiAffinities(ctx))
+	for _, p := range pods {
```
ellistarn, line 147: Thoughts on injecting the logger With("pod", pod) here to avoid doing it multiple times in a tight loop? Also seems more logical to me.

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -76,3 +102,583 @@ func (s *NodeSet) getDaemons(ctx context.Context, constraints *v1alpha5.Constrai
...
+	}
+	return errs
+}
+
+// trackExistingAntiAffinities is used to identify pods with anti-affinity terms so we can track those topologies.  We
+// have to look at every pod in the cluster as there is no way to query for a pod with anti-affinity terms.
+func (s *NodeSet) trackExistingAntiAffinities(ctx context.Context) error {
+	// TODO: if can we determine somehow that the LimitPodHardAntiAffinityTopology admission controller is in place, we
+	// don't need to do any of this.  In that case anti-affinity would be hostname only which makes this tracking un-needed.
+	var nodeList v1.NodeList
```
ellistarn, line 166: Thoughts on separating out topology into a state object to help us avoid calling List nodes so many times?

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,139 +18,283 @@ import (
...
 		if !ok {
 			continue // Don't include pods if node doesn't contain domain https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/#conventions
 		}
-		topologyGroup.Increment(domain)
+		t.RecordUsage(domain)
 	}
 	return nil
 }
 
+func (t *Topology) EmptyDomain(reqts v1alpha5.Requirements) string {
```
ellistarn, line 263: Please use fully spelled variables for readability.
```suggestion
func (t *Topology) EmptyDomain(requirements v1alpha5.Requirements) string {
```

ellistarn (commented):
Took a first pass. First off, well done. This is a big PR and hugely impactful in the short and long term. In general, I agree with what I understood of the core algorithm decisions. I want to discuss the tradeoffs to scoring nodes vs picking the first fit by least full or something similar. If we do go the route of filter/score, I want to model it explicitly. 

I think there are a number of cosmetic changes that could be made to really improve readability.

- Minimize the responsibility of objects. Create separations of concerns and simplify multiple intermingled concepts into single purpose ones.
- Spell out variable names, and keep the names minimal unless context can't be inferred. Seeing names like `z1Pod` really slows me down.
- Be judicious about disabling the linter. There's a reason we have gocyclo.
- Be thoughtful about names. Code is read much more than it is written. Be kind to your readers. As a rule of thumb, I find that if I'm stringing multiple words together, it's because I've failed to construct the right abstractions such that naming is natural. For example `recordTopologyDecisions` could be something like `topology.Record(pod, domain)`. Another example is `existingAntiAffinities` vs `topologies`. Why is one existing and the other not? Is there something conceptually different? I'm really lost on `applyNodeSelectorIfNotControlled`.
- Try to connect concepts together such that nothing is misplaced or "poorly encapsulated". Usage anonymous types and static functions are often good hints that something needs to be better encapsulated.

tzneal:
> Took a first pass. First off, well done. This is a big PR and hugely impactful in the short and long term. In general, I agree with what I understood of the core algorithm decisions. I want to discuss the tradeoffs to scoring nodes vs picking the first fit by least full or something similar. If we do go the route of filter/score, lets model it explicitly.
> 
> I think there are a number of cosmetic changes that could be made to really improve readability.
> 
>  * Minimize the responsibility of objects. Create separations of concerns and simplify multiple intermingled concepts into single purpose ones.
> 
>  * Spell out variable names, and keep the names minimal unless context can't be inferred. Seeing names like `z1Pod` really slows me down.

I'm happy to rename things, but a counterpoint is that longer spelled out names is not idiomatic Go style, particularly for variables with short lifetimes, see https://github.com/golang/go/wiki/CodeReviewComments#variable-names

>  * Be judicious about disabling the linter. There's a reason we have gocyclo.

Happy to consider other ways of writing it, but I didn't see a clean way to break those functions up that improved readability.

>  * Be thoughtful about names. Code is read much more than it is written. Be kind to your readers. As a rule of thumb, I find that if I'm stringing multiple words together, it's because I've failed to construct the right abstractions such that naming is natural. For example `recordTopologyDecisions` could be something like `topology.Record(pod, domain)`. Another example is `existingAntiAffinities` vs `topologies`. Why is one existing and the other not? Is there something conceptually different? I'm really lost on `applyNodeSelectorIfNotControlled`.

Renamed existingAntiAffinities to inverseAntiAffinityTopologies as it's more accurate. 

recordTopologyDecisions loops over all topologies to find ones that match the pod and then record the decision. If it were `topology.Record(pod, domain)`, you would still have the outer loop and are then calling a `topology.Record()` method which needs to check if the topology actually matches the domain, or else it's a no-op.  That seems odd, so you'd pull the check outside of the topology and are then left with `recordTopologyDecisions` :)

> 
>     * Try to connect concepts together such that nothing is misplaced or "poorly encapsulated". Usage anonymous types and static functions are often good hints that something needs to be better encapsulated.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -37,16 +40,23 @@ import (
...
 )
 
+const (
+	failedSchedulingTTL             = 60 * time.Second
+	failedSchedulingCleanupInterval = 10 * time.Minute
+)
+
 func NewProvisioner(ctx context.Context, provisioner *v1alpha5.Provisioner, kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.CloudProvider) *Provisioner {
 	running, stop := context.WithCancel(ctx)
+	failedSchedulingCache := cache.New(failedSchedulingTTL, failedSchedulingCleanupInterval)
```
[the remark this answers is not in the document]
tzneal: Before it was rare for a pod to not schedule as it would have to not match any instance type that the provisioner can create. Now pods can fail to schedule due to skew violations or affinity violations.  If this occurs, they spam the log with their failed scheduling every few seconds.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -95,20 +104,31 @@ func (p *Provisioner) provision(ctx context.Context) error {
 			pods = append(pods, item.(*v1.Pod))
 		}
 	}
+	// if we fail to schedule a pod it remains unschedulable and gets batched a few seconds later, we do this to prevent
```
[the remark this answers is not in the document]
tzneal: I'll remove it, but a failed pod scheduling due to affinity or skew will spam the log with a half dozen lines of logs every few seconds.

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -28,43 +32,72 @@ import (
...
+	n.Constraints.Requirements = n.Constraints.Requirements.Add(v1.NodeSelectorRequirement{
+		Key:      v1.LabelHostname,
+		Operator: v1.NodeSelectorOpIn,
+		Values:   []string{n.Name},
+	})
+	return n
 }
 
-func (n *Node) Add(pod *v1.Pod) error {
+func (n *Node) isCompatibleWith(pod *v1.Pod) (v1alpha5.Requirements, v1.ResourceList, []cloudprovider.InstanceType, error) {
```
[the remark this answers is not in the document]
tzneal: I wrote this three times trying to find a way to make it work that doesn't replicate the compatibility checks in two methods.  The problem is that we need the results of the compatibility checks.

1) The method returns a (commit func(), err error) , and you call the commit function if you want to commit the changes.  The benefit is that you save an additional expensive compatibility check when you finally commit, downside is that it's weird.

2) A method like `func(n *Node) addPod(pod *v1.Pod, commit bool) error` that will only truly commit the changes if commit is true, the exposed Compatible/Add methods are then just wrappers around it.

3) This approach with an internal helper that compatibility checks and returns the results so they can be commited from `Add()` and are ignored from `Compatible()`

In this approach there is a `Compatible` function that mirrors requirements, the method you commented on is the internal helper.  

```go
func (n *Node) Compatible(ctx context.Context, pod *v1.Pod) error {
	_, _, _, err := n.isCompatibleWith(pod)
	return err
}
```

So far I think 3 is cleanest at the expensive of a redundant check, but open to any alternatives.

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
```diff
@@ -44,7 +44,7 @@ import (
 	testclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
-const MinPodsPerSec = 250.0
+const MinPodsPerSec = 1.0
```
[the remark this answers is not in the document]
tzneal: Just changed it so I could run the benchmark without it erroring out, current code is around ~170 pods/sec, with some optimizations that I will do in a different PR I got ~225 pods/sec.

pkg/utils/pod/scheduling.go
```diff
@@ -28,6 +29,10 @@ func FailedToSchedule(pod *v1.Pod) bool {
 	return false
 }
 
+func NamespacedName(p *v1.Pod) string {
```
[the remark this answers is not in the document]
tzneal: Ah, didn't know that had a stringer.

pkg/test/expectations/expectations.go
```diff
@@ -179,7 +182,16 @@ func ExpectProvisioned(ctx context.Context, c client.Client, selectionController
 	// Wait for reconcile
 	ExpectReconcileSucceeded(ctx, provisioningController, client.ObjectKeyFromObject(provisioner))
 	wg := sync.WaitGroup{}
-	for _, pod := range pods {
+
+	unorderedPods := append([]*v1.Pod{}, pods...)
+	// shuffle the pods to try to detect any issues where we rely on pod order within a batch, we shuffle a copy of
+	// the slice so we can return the provisioned pods in the same order that the test supplied them for consistency
+	r := rand.New(rand.NewSource(ginkgo.GinkgoRandomSeed())) //nolint
```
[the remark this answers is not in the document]
tzneal: gosec complains that I'm not using a crypto random number generator to shuffle some pods for a test.

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -28,43 +32,72 @@ import (
 // Node is a set of constraints, compatible pods, and possible instance types that could fulfill these constraints. This
 // will be turned into one or more actual node instances within the cluster after bin packing.
 type Node struct {
+	Name                string
 	Constraints         *v1alpha5.Constraints
 	InstanceTypeOptions []cloudprovider.InstanceType
 	Pods                []*v1.Pod
 
 	requests v1.ResourceList
+	client   client.Client
```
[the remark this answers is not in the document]
tzneal: Nope, good catch.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -80,12 +91,10 @@ func (p *Provisioner) Add(pod *v1.Pod) <-chan struct{} {
 
 func (p *Provisioner) provision(ctx context.Context) error {
 	// Batch pods
-	logging.FromContext(ctx).Infof("Waiting for unschedulable pods")
```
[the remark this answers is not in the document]
tzneal: It gets spammed a lot in the case of having a pod that is not schedulable because of affinity or skew-violation.  It would print out every few seconds.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -51,60 +51,69 @@ func init() {
...
+		if len(unschedulablePods) == 0 || len(unschedulablePods) == previousUnschedulableCount {
+			break
 		}
 	}
 
-	if unschedulableCount != 0 {
-		logging.FromContext(ctx).Errorf("Failed to schedule %d pods", unschedulableCount)
+	if len(unschedulablePods) != 0 {
+		for _, up := range unschedulablePods {
+			logging.FromContext(ctx).With("pod", client.ObjectKeyFromObject(up.pod)).Errorf("Scheduling pod, %s", up.reason)
```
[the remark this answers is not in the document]
tzneal: Agreed, but we don't record events to pods yet.  I think once we do, this goes away and we just record the failed scheduling on the pod so users can see it in `kubectl describe pod pod-name`

pkg/controllers/provisioning/scheduling/nodeset.go
tzneal: I don't think so because the scheduler is re-used.  This also allows a potential defrag to construct a nodeset to evaluate a potential decision.

pkg/controllers/provisioning/scheduling/nodeset.go
tzneal: Renamed to `inverseAntiAffinityTopologies` which is still not great, but more accurate.  The antiAffinities are also stored in `topologies.

pkg/controllers/provisioning/scheduling/nodeset.go
tzneal: We can, but I was trying to minimize changes in this PR since it was already 1k+.

pkg/controllers/provisioning/scheduling/nodeset.go
tzneal: I don't think so, this set of tightened requirements gets modified by the topology decisions we make and is used to inform future topology decisions.

pkg/controllers/provisioning/scheduling/nodeset.go
tzneal: Removing the stored logger, originally I didn't need context for anything but logging in `Schedule`, but now that it's used for more I'll just use it.

pkg/controllers/provisioning/scheduling/nodeset.go
tzneal: Not following this comment.

pkg/controllers/provisioning/scheduling/nodeset.go
tzneal: It would be complicated.  We need to know the current state of the cluster with respect to all pods that have been scheduled by us or kube-scheduler previously that have pod anti-affinity.  We would have to watch for node creation/deletion, and pod creation/update/deletion.

pkg/controllers/provisioning/scheduling/topology.go
tzneal: Happy to change, but it's not idiomatic go IMO.  https://github.com/golang/go/wiki/CodeReviewComments#variable-names

pkg/apis/provisioning/v1alpha5/labels.go
```diff
@@ -27,6 +27,9 @@ var (
 	ArchitectureArm64    = "arm64"
 	OperatingSystemLinux = "linux"
 
+	// ValidTopologyKeys are the topology keys that Karpenter allows for topology spread and pod affinity/anti-affinity
+	ValidTopologyKeys = stringsets.NewString(v1.LabelHostname, v1.LabelTopologyZone, LabelCapacityType)
```
ellistarn, line 31: Is there a reason we have to be aware of this?

pkg/controllers/provisioning/provisioner.go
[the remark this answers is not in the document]
tzneal: Removing for now and may have some excess logs until retry mechanism is reworked.

pkg/controllers/provisioning/provisioner.go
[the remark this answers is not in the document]
ellistarn: It should print exactly once at the beginning of a reconcile loop. Why is the provisioning loop firing so many times?

pkg/controllers/provisioning/provisioner.go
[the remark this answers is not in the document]
ellistarn: Let's fix the requeue behavior, not the logging.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,166 +18,480 @@ import (
...
 
 type Topology struct {
 	kubeClient client.Client
+	topologies map[uint64]*TopologyGroup
+	// Anti-affinity works both ways (if a zone has a pod foo with anti-affinity to a pod bar, we can't schedule bar to
+	// that zone, even though bar has no anti affinity terms on it. For this to work, we need to separately track the
+	// topologies of pods with anti-affinity terms, so we can prevent scheduling the pods they have anti-affinity to
+	// in some cases.
+	inverseTopologies map[uint64]*TopologyGroup
+	constraints       *v1alpha5.Constraints
```
ellistarn, line 44: I don't think this is used anymore. I'm thinking of this topology object as a longlived tracker/manager, so constraints wouldn't make sense anyways. It should even be shared across provisioners.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,166 +18,480 @@ import (
...
 
-// Inject injects topology rules into pods using supported NodeSelectors
-func (t *Topology) Inject(ctx context.Context, constraints *v1alpha5.Constraints, pods []*v1.Pod) error {
-	// Group pods by equivalent topology spread constraints
-	topologyGroups := t.getTopologyGroups(pods)
-	// Compute spread
-	for _, topologyGroup := range topologyGroups {
-		if err := t.computeCurrentTopology(ctx, constraints, topologyGroup); err != nil {
-			return fmt.Errorf("computing topology, %w", err)
+func (t *Topology) TrackTopologies(ctx context.Context, pods ...*v1.Pod) error {
```
ellistarn, line 56: Thoughts on this naming suggestion?
```suggestion
func (t *Topology) Update(ctx context.Context, pods ...*v1.Pod) error {
```

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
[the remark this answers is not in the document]
ellistarn: Perhaps we can set to 100 for now.

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
```diff
@@ -93,7 +93,8 @@ func TestSchedulingProfile(t *testing.T) {
 	totalNodes := 0
 	var totalTime time.Duration
 	for _, instanceCount := range []int{400} {
-		for _, podCount := range []int{10, 100, 500, 1000, 1500, 2000, 2500} {
+		//for _, podCount := range []int{10, 100, 500, 1000, 1500, 2000, 2500} {
+		for _, podCount := range []int{500} {
```
ellistarn, line 97: What if it doesn't scale linearly?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -50,60 +50,77 @@ func init() {
...
-				nodeSet.Add(n)
+			if err := nodeSet.Schedule(ctx, p); err != nil {
+				lastErrors[p] = err
+				newUnschedulablePods = append(newUnschedulablePods, p)
 			}
 		}
+		unschedulablePods = newUnschedulablePods
+		// if there are no more pods to attempt scheduling, or we tried each pod in a scheduling round
+		// and made no progress, we are finished
+		if len(unschedulablePods) == 0 || len(unschedulablePods) == previousUnschedulableCount {
```
ellistarn, line 112: Why not put this in the for loop condition?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -50,60 +50,77 @@ func init() {
...
-		isScheduled := false
-		for _, node := range nodeSet.nodes {
-			if err := node.Add(pod); err == nil {
-				isScheduled = true
-				break
+	if err := topology.TrackTopologies(ctx, pods...); err != nil {
+		return nil, fmt.Errorf("tracking topology counts, %w", err)
+	}
+
+	lastErrors := map[*v1.Pod]error{}
```
ellistarn, line 82: Not obvious what `last` means until you read the impl below. I'd probably lean the former, since we're already in the scheduling loop so it can be inferred from context.
```suggestion
	errs := map[*v1.Pod]error{}
```
```suggestion
	schedulingErrs := map[*v1.Pod]error{}
```

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -50,60 +50,77 @@ func init() {
...
+
+	var unschedulablePods []*v1.Pod
+	unschedulablePods = append(unschedulablePods, pods...)
```
ellistarn, line 85: Why not
```suggestion
	unschedulablePods := append([]*v1.Pod{}, pods...)
```
or if you're looking to avoid an allocate:
```suggestion
	unschedulablePods := append(nil, pods...)
```

ellistarn: Thinking further, why don't we just use the `pods` slice directly? Why do we need something new?

pkg/apis/provisioning/v1alpha5/labels.go
tzneal: We haven't tested with anything else, I think it's ok for this PR with the possibility to expand it in the future.  It's always been this limited for topology spread (capacity type is a recent addition).  Need to figure out what the use case is for other labels since the provider needs to put them on the pod, there probably is a use now considering the exists operator for labels.

pkg/controllers/provisioning/scheduling/scheduler.go
tzneal: We could, I thought it seemed clearer to not mutate the pods slice.

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
tzneal: I'll change it back, modified this while optimizing but decided to wait to optimize until this is landed.

pkg/controllers/provisioning/scheduling/topology.go
tzneal: It's only used once place now which is when creating new topology groups.  We need to pull off the available zones/capacity types to initialize those.  If in the future it's reused across provisioners we could initialize with the union of all provisioner supported constraints.  I think we already tighten by provisioner constraints, so having extra domains there wouldn't hurt normally, it could get weird though:

Assume,
- provisioner 1 can schedule in zone-a, zone-b
- provisioner 2 can schedule in zone-a, zone-b, zone-c

If a deployment with zonal spread/max-skew=1 could only be provisioned by provisioner 1 due to a label seletor, we could run a maximum of four pods.  The topology would know about the zone-c and prevent scheduling more.  If this is correct behavior or not depends on your viewpoint :)

pkg/controllers/provisioning/provisioner.go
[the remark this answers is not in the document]
tzneal: Haven't looked in detail, but if a pod remains unschedulable I believe it hits the selection controller every 5 seconds due to the retry there which causes it to go to the provisioner, fail again, rinse, repeat.

pkg/apis/provisioning/v1alpha5/labels.go
ellistarn: I can imagine some, but we can punt on it.

pkg/controllers/provisioning/provisioner.go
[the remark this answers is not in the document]
ellistarn: Yeah this is the current behavior. Can we leave this as is and decide on a path forward in another PR?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -50,60 +50,70 @@ func init() {
...
 	sort.Slice(instanceTypes, byPrice(instanceTypes))
 
-	// Inject temporarily adds specific NodeSelectors to pods, which are then
-	// used by scheduling logic. This isn't strictly necessary, but is a useful
-	// trick to avoid passing topology decisions through the scheduling code. It
-	// lets us treat TopologySpreadConstraints as just-in-time NodeSelectors.
-	if err := s.topology.Inject(ctx, constraints, pods); err != nil {
-		return nil, fmt.Errorf("injecting topology, %w", err)
-	}
+	topology := NewTopology(s.kubeClient, constraints)
```
ellistarn, line 71: I think eventually we're gonna want to inject this into the scheduler and make it longlived.

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -76,3 +86,42 @@ func (s *NodeSet) getDaemons(ctx context.Context, constraints *v1alpha5.Constrai
...
+	if scheduledNode != nil {
+		if err := s.topology.Record(p, scheduledNode.Constraints.Requirements); err != nil {
+			logging.FromContext(ctx).Errorf("Recording topology decision, %s", err)
+			return err
+		}
+	}
+	return nil
+}
+
+type matchingTopology struct {
```
ellistarn, line 123: Unused now?

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -28,14 +32,20 @@ import (
...
+	kubeClient      client.Client
 	daemonResources v1.ResourceList
 	nodes           []*Node
+	constraints     *v1alpha5.Constraints
+	instanceTypes   []cloudprovider.InstanceType
+	topology        *Topology
 }
 
-func NewNodeSet(ctx context.Context, constraints *v1alpha5.Constraints, client client.Client) (*NodeSet, error) {
+func NewNodeSet(ctx context.Context, topology *Topology, constraints *v1alpha5.Constraints, instanceTypes []cloudprovider.InstanceType, client client.Client) (*NodeSet, error) {
```
ellistarn, line 43: The responsibilities of this class overlaps pretty heavily with Scheduler, and both are pretty small.

For example, 4 of the 6 member variables are injected from topology, and the other two could simply live on the stack. 

I'll see if I can cut you a PR.

pkg/controllers/provisioning/scheduling/nodeset.go
```diff
@@ -59,15 +69,15 @@ func NewNodeSet(ctx context.Context, constraints *v1alpha5.Constraints, client c
...
 		return nil, fmt.Errorf("listing daemonsets, %w", err)
 	}
 	// Include DaemonSets that will schedule on this node
 	var pods []*v1.Pod
 	for _, daemonSet := range daemonSetList.Items {
-		pod := &v1.Pod{Spec: daemonSet.Spec.Template.Spec}
-		if err := constraints.ValidatePod(pod); err == nil {
-			pods = append(pods, pod)
+		p := &v1.Pod{Spec: daemonSet.Spec.Template.Spec}
+		if err := constraints.ValidatePod(p); err == nil {
```
ellistarn, line 79: Just noticed that we're validating this here and again on line 46

pkg/controllers/provisioning/scheduling/nodeset.go
tzneal: No, still used: `func (t *Topology) getMatchingTopologies(p *v1.Pod) []matchingTopology`

pkg/controllers/provisioning/scheduling/nodeset.go
tzneal: I see this on 46: `constraints: constraints,`, wrong line number?

pkg/controllers/provisioning/scheduling/nodeset.go
ellistarn: I worry a bit about aspirational abstraction. If we both agreed, I might be more confident about it.

pkg/controllers/provisioning/scheduling/scheduler.go
tzneal: It will make inverse anti-affinities less expensive, but definitely for another PR due to the complexity of ensuring we track correctly. It may get complicated as well as we have to undo those topology changes if nodes fail to launch 😆

pkg/controllers/provisioning/scheduling/nodeset.go
tzneal: I'm fine merging, I mis-remembered  and was thinking Scheduler was a real controller.

pkg/controllers/provisioning/scheduling/nodeset.go
ellistarn: Ah -- just in the wrong file.

pkg/controllers/provisioning/scheduling/scheduler.go
ellistarn: We will definitely need to be able to update the guessed state of the world with the actual. I'm imagining this looks exactly the same as updating w/ actual and just overwrite keys.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -95,20 +94,20 @@ func (p *Provisioner) provision(ctx context.Context) error {
 			pods = append(pods, item.(*v1.Pod))
 		}
 	}
+	logging.FromContext(ctx).Infof("Batched %d provisionable pod(s) of %d pod(s) in %s", len(pods), len(items), window)
```
ellistarn, line 97: This log line begs the question, "what is this out of". The answer is not satisfying to users, that our cache was stale in one are of the code and not stale in another. 

Here's an option:
```suggestion
        if len(pods) == 0 {
          return nil
        }
	logging.FromContext(ctx).Infof("Batched %d pod(s) in %s", len(pods), window)
```

pkg/controllers/provisioning/scheduling/topology.go
ellistarn: Sounds correct to me

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,506 @@ import (
...
+
+// Initialize scans the pods provided and creates topology groups for any topologies that we need to track based off of
+// topology spreads, affinities, and anti-affinities specified in the pods.
+func (t *Topology) Initialize(ctx context.Context, pods ...*v1.Pod) error {
+	var errs error
+	errs = multierr.Append(errs, t.trackExistingAntiAffinities(ctx))
+	for _, p := range pods {
+		// need to first ensure that we know all of the topology constraints.  This may require looking up
+		// existing pods in the running cluster to determine zonal topology skew.
+		if err := t.trackTopologySpread(ctx, p); err != nil {
```
ellistarn, line 71: I typically try to avoid verb sprawl. In this case `track` `count` `update` `compute` `calculate` `warm` are all viable terms, and I'm sure there are many more I'm not thinking of. I try to minimize the nuance in verbs and instead push all of the context into the noun. This helps constrain the naming space, and results in more similar code from different developers, especially over time. It also helps guide encapsulation best practices, since if your interface has two things it needs to Add, it's a good hint to separate the abstraction.

It's obviously not perfect, but across this codebase we have a relatively small number of simple verbs `Create` `Delete` `Add` `Update` and then a set of "terms of art" from the domain like `Pack` and `Schedule`. I could imagine taking it a step further and replacing those terms of art e.g. `scheduler.Add`, `packing.Add`.

```suggestion
		if err := t.updateTopologySpread(ctx, p); err != nil {
```

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,506 @@ import (
...
+		if err := t.trackPodAffinityTopology(ctx, p); err != nil {
+			errs = multierr.Append(errs, fmt.Errorf("tracking affinity topology, %w", err))
+		}
+	}
+	return errs
+}
+
+// trackExistingAntiAffinities is used to identify pods with anti-affinity terms so we can track those topologies.  We
+// have to look at every pod in the cluster as there is no way to query for a pod with anti-affinity terms.
+func (t *Topology) trackExistingAntiAffinities(ctx context.Context) error {
```
ellistarn, line 83: I typically try to avoid unnecessary context in method names. In this case, how does `existing` clarify the meaning? Are there nonexisting antiaffinities? Should `topologySpread` or `affinity` be prefixed with `existing` as well?
```suggestion
func (t *Topology) trackAntiAffinities(ctx context.Context) error {
```

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,506 @@ import (
...
+	// TODO: if can we determine somehow that the LimitPodHardAntiAffinityTopology admission controller is in place, we
```
ellistarn, line 84: I don't think we can assume this. Let's toss the comment.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,506 @@ import (
...
-// computeZonalTopology for the topology group. Zones include viable zones for
-// the { cloudprovider, provisioner, pod }. If these zones change over time,
-// topology skew calculations will only include the current viable zone
-// selection. For example, if a cloud provider or provisioner changes the viable
-// set of nodes, topology calculations will rebalance the new set of zones.
-func (t *Topology) computeZonalTopology(ctx context.Context, constraints *v1alpha5.Constraints, topologyGroup *TopologyGroup) error {
-	topologyGroup.Register(constraints.Requirements.Zones().UnsortedList()...)
-	if err := t.countMatchingPods(ctx, topologyGroup); err != nil {
-		return fmt.Errorf("getting matching pods, %w", err)
+func (t *Topology) trackPodAffinityTopology(ctx context.Context, p *v1.Pod) error {
```
ellistarn, line 205: I'm a bit confused why this one doesn't have parity with `trackTopologySpread`. How does `topology` clarify things? Are there pod affinities that we don't model as `topology`? Does it make sense to suffix all of these with `topology`, like `trackTopologySpreadTopology`? 

```suggestion
func (t *Topology) trackPodAffinity(ctx context.Context, p *v1.Pod) error {
```

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,506 @@ import (
...
+
+	for _, node := range nodeList.Items {
+		var podlist v1.PodList
+		if err := t.kubeClient.List(ctx, &podlist, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
+			return fmt.Errorf("listing pods on node %s, %w", node.Name, err)
+		}
+		for _, p := range podlist.Items {
+			// lint warning is about taking the address of a variable in the for loop, it would matter if
+			// HasRequiredPodAntiAffinity modified p, but it doesn't
+			//nolint:gosec
```
ellistarn, line 99: We typically avoid this pattern in general, since this will introduce a code bug if `HasRequiredPodAntiAffinity` starts modifying p. The way we typically avoid this is by indexing:

```
for i := range podlist.Items {
  if pod.HasRequiredPodAntiAffinity(&podlist.Items[i]) {
```

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,506 @@ import (
...
+
+		topologyGroup, err := t.newForAffinity(term, namespaces, TopologyTypePodAntiAffinity)
+		if err != nil {
+			return err
+		}
+
+		topologyGroup.isInverse = true
+		hash := topologyGroup.Hash()
+		if existing, ok := t.inverseTopologies[hash]; !ok {
+			t.inverseTopologies[hash] = topologyGroup
```
ellistarn, line 190: Why do we need `topologyGroup.isInverse = true` if this topology is stored in `inverseTopologies`. Can this be deduplicated?

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,506 @@ import (
...
+// pods with the anti-affinity.
+func (t *Topology) trackInverseAntiAffinity(ctx context.Context, node *v1.Node, p *v1.Pod) error {
+	// We intentionally don't track inverse anti-affinity preferences.  We're not required to enforce them so it
+	// just adds complexity for very little value.
+	for _, term := range p.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution {
+		namespaces, err := t.buildNamespaceList(ctx, term.Namespaces, term.NamespaceSelector)
+		if err != nil {
+			return err
+		}
+		namespaces.Insert(p.Namespace)
```
ellistarn, line 180: Is the pod's namespace considered if namespaceSelector is provided? I thought it was just a default.

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -15,54 +15,259 @@ limitations under the License.
...
+	return maxDomain
+}
+
+func (t *TopologyGroup) RegisterDomain(zone string) {
+	if _, ok := t.domains[zone]; !ok {
+		t.domains[zone] = 0
+	}
+}
+
+func (t *TopologyGroup) EmptyDomain(requirements v1alpha5.Requirements) string {
```
ellistarn, line 208: Why does topology expose different interfaces for different topology types `EmptyDomain`, `MaxNonZeroDomain`, `AnyDomain`. It knows it's `ToplogyType`, so can't it just make the right decision with something simple like `topologyGroup.Next()`

pkg/utils/apiobject/apiobject.go
```diff
@@ -27,3 +29,9 @@ func PodNamespacedNames(pods []*v1.Pod) []types.NamespacedName {
 	}
 	return namespacedNames
 }
+
+func MustHash(object interface{}) uint64 {
```
ellistarn, line 33: This package is intended for `apiObject` helpers like `runtime.Object` `client.Object`, etc. I've been trying to deprecate it. I think `PodNamespacedNames` can probably go live in `pod` or w/e we end up naming it. Maybe we can make something like `must.Hash(obj)` or `hash.OrDie(obj)`

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -50,61 +51,106 @@ func init() {
...
+	topology := NewTopology(s.kubeClient, &constraints.Requirements)
+	if err := topology.Initialize(ctx, pods...); err != nil {
+		return nil, fmt.Errorf("tracking topology counts, %w", err)
 	}
-
-	nodeSet, err := NewNodeSet(ctx, constraints, s.kubeClient)
+	daemonOverhead, err := s.getDaemonOverhead(ctx, constraints)
 	if err != nil {
-		return nil, fmt.Errorf("constructing nodeset, %w", err)
+		return nil, err
```
ellistarn, line 75: Worth doing a pass on all of these errors at some point to make sure they're annotated where useful.

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
```diff
@@ -164,7 +196,8 @@ func makeDiversePods(count int) []*v1.Pod {
 	pods = append(pods, makePodAffinityPods(count/7, v1.LabelHostname)...)
 	pods = append(pods, makePodAffinityPods(count/7, v1.LabelTopologyZone)...)
 	pods = append(pods, makePodAntiAffinityPods(count/7, v1.LabelHostname)...)
-	pods = append(pods, makePodAntiAffinityPods(count/7, v1.LabelTopologyZone)...)
+	// We intentionally don't do anti-affinity by zone as that creates tons of unschedulable pods
+	//pods = append(pods, makePodAntiAffinityPods(count/7, v1.LabelTopologyZone)...)
```
ellistarn, line 200: reminder to clean this up

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,396 @@ import (
...
+	for i := range pods {
+		errs = multierr.Append(errs, t.Update(ctx, pods[i]))
+	}
+	return errs
+}
+
+func (t *Topology) Update(ctx context.Context, p *v1.Pod) (errs error) {
+	for _, topology := range t.topologies {
+		if topology.IsOwnedBy(p.UID) {
+			topology.RemoveOwner(p.UID)
```
ellistarn, line 69: Do we need to handle garbage collection where the topology no longer has owners? I guess we potentially don't care since the object doesn't live longer than the batch.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,396 @@ import (
...
 		}
-		for _, pod := range topologyGroup.Pods {
-			domain := topologyGroup.NextDomain(constraints.Requirements.Add(v1alpha5.NewPodRequirements(pod).Requirements...).
-				Get(topologyGroup.Constraint.TopologyKey).
-				Values())
-			pod.Spec.NodeSelector = functional.UnionStringMaps(pod.Spec.NodeSelector, map[string]string{topologyGroup.Constraint.TopologyKey: domain})
+	}
+
+	// need to first ensure that we know all of the topology constraints.  This may require looking up
```
ellistarn, line 73: This comment isn't making sense to me. Remove, maybe, or make it a comment on `Update`?

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,396 @@ import (
...
-				topologyGroupMap[key] = NewTopologyGroup(pod, constraint)
+func (t *Topology) updateTopologySpread(ctx context.Context, p *v1.Pod) error {
+	for _, cs := range p.Spec.TopologySpreadConstraints {
+		topologyGroup, err := t.newForSpread(p.Namespace, cs)
+		if err != nil {
+			return err
+		}
+		hash := topologyGroup.Hash()
+		if existing, ok := t.topologies[hash]; !ok {
+			if err := t.initializeTopologyGroup(ctx, topologyGroup); err != nil {
```
ellistarn, line 115: Thoughts on moving initialization logic to inside `newForSpread` and then only new it after we check for existing?

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,396 @@ import (
...
+		topologyGroup.AddOwner(p.UID)
 	}
-	topologyGroups := []*TopologyGroup{}
-	for _, topologyGroup := range topologyGroupMap {
-		topologyGroups = append(topologyGroups, topologyGroup)
+	return nil
+}
+
+func (t *Topology) newForSpread(namespace string, cs v1.TopologySpreadConstraint) (*TopologyGroup, error) {
+	// validate our label selector that the topology group will later create as needed
```
ellistarn, line 128: This selector is already validated when it's persisted to the API server. Perhaps we can skip?

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,396 @@ import (
...
-	case v1.LabelHostname:
-		return t.computeHostnameTopology(topologyGroup, constraints)
-	case v1.LabelTopologyZone:
-		return t.computeZonalTopology(ctx, constraints, topologyGroup)
-	case v1alpha5.LabelCapacityType:
-		return t.computeLabelCapacity(ctx, constraints, topologyGroup)
-	default:
-		return nil
+func (t *Topology) newForAffinity(term v1.PodAffinityTerm, namespaces sets.String, topoType TopologyType) (*TopologyGroup, error) {
+	_, err := metav1.LabelSelectorAsSelector(term.LabelSelector)
```
ellistarn, line 145: same

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,149 +18,396 @@ import (
...
+func (t *Topology) updateInverseAntiAffinity(ctx context.Context, node *v1.Node, p *v1.Pod) error {
+	// We intentionally don't track inverse anti-affinity preferences.  We're not required to enforce them so it
+	// just adds complexity for very little value.
+	for _, term := range p.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution {
+		namespaces, err := t.buildNamespaceList(ctx, p.Namespace, term.Namespaces, term.NamespaceSelector)
+		if err != nil {
+			return err
+		}
+
+		topologyGroup, err := t.newForAffinity(term, namespaces, TopologyTypePodAntiAffinity)
```
ellistarn, line 172: Same comment on the initialize encapsulation recommendation.

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -15,54 +15,283 @@ limitations under the License.
...
+	utilsets "k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
+	"github.com/aws/karpenter/pkg/utils/sets"
+)
+
+type TopologyType byte
+
+const (
+	TopologyTypeSpread TopologyType = iota
```
ellistarn, line 36: Just a thought, this might be slightly simpler to model with a string, since you can name all of the types and avoid the extra .String() complexity. I'm not sure the performance difference is worth the extra wrinkle.

pkg/controllers/provisioning/scheduling/topology.go
tzneal: I wasn't sure if "made it to the API server" was enough to guarantee no error from the conversion

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -28,37 +29,53 @@ import (
...
+	})
+	return n
+}
 
-	if len(n.Pods) != 0 {
-		// TODO: remove this check for n.Pods once we properly support hostname topology spread
-		if err := n.Constraints.Requirements.Compatible(podRequirements); err != nil {
-			return err
-		}
+func (n *Node) Add(topology *Topology, p *v1.Pod) error {
```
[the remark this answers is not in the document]
tzneal: Seems fine, probably held off because constructor has a bunch of params already, but agree that Add() is looking awkward with this topology parameter.

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -15,54 +15,283 @@ limitations under the License.
...
+	namespaces  utilsets.String
+	rawSelector *metav1.LabelSelector // stored so we can easily hash
+	// Pod Index
+	owners map[types.UID]struct{} // Pods that have this topology as a scheduling rule
+	// Internal state
+	domains map[string]int32 // TODO(ellistarn) explore replacing with a minheap
+}
+
+func (t *TopologyGroup) Matches(namespace string, podLabels labels.Set) bool {
+	selector, _ := metav1.LabelSelectorAsSelector(t.rawSelector)
```
ellistarn, line 68: thoughts on runtime.Must()?

I was even thinking about some sort of helper that was runtime.Must but for multiple return values
```
package hysteria

func Must2(value interface{}, err error) interface{} {
  runtime.Must(error)
  return value
}
```
```
return t.namespaces.Has(namespace) && hysteria.Must2(selector.Matches(podLabels)).(v1.Selector)
```

I wonder if this would be better with generics 🤔

pkg/controllers/provisioning/scheduling/topology.go
tzneal: Yeah, it's short lived.  We actually can't check it here either and would need to rescan at the end of the method.  It may go to zero if the pod was the only owner and the term wasn't relaxed. In which case we would remove it, re-create it, re-query the cluster to get counts, and then add the pod back as the owner.

pkg/controllers/provisioning/scheduling/topology.go
tzneal: The Hash() method is on the result of the new call.  We do a cheap new, essentially just to see if it exists, and if not continue with the real init.

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -15,54 +15,283 @@ limitations under the License.
...
-	Constraint v1.TopologySpreadConstraint
-	Pods       []*v1.Pod
-	// spread is an internal field used to track current spread
-	spread map[string]int // TODO(ellistarn) explore replacing with a minheap
+	// Hashed Fields
+	Key         string
+	Type        TopologyType
+	maxSkew     int32
+	namespaces  utilsets.String
+	rawSelector *metav1.LabelSelector // stored so we can easily hash
```
ellistarn, line 60:
```suggestion
	selector *metav1.LabelSelector
```

pkg/controllers/provisioning/scheduling/topologygroup.go
tzneal: I'll just use runtime.Must on the error, found a few other places in topology listing code that ignored it and added it there too.

pkg/controllers/provisioning/scheduling/topology.go
ellistarn: Oh ofc. yes.

pkg/controllers/provisioning/scheduling/topology.go
tzneal: Removed it and used runtime.Must later when we create it.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,150 +18,373 @@ import (
...
 )
 
 type Topology struct {
 	kubeClient client.Client
+	topologies map[uint64]*TopologyGroup
+	// Anti-affinity works both ways (if a zone has a pod foo with anti-affinity to a pod bar, we can't schedule bar to
+	// that zone, even though bar has no anti affinity terms on it. For this to work, we need to separately track the
+	// topologies of pods with anti-affinity terms, so we can prevent scheduling the pods they have anti-affinity to
+	// in some cases.
+	inverseTopologies map[uint64]*TopologyGroup
```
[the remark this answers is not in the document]
njtran: Or based on L168, it looks like it's the hash of a topologyGroup?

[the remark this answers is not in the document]
tzneal: Yup, it's the second, added.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,150 +18,373 @@ import (
...
+func (t *Topology) updateInverseAffinities(ctx context.Context) error {
+	var nodeList v1.NodeList
+	if err := t.kubeClient.List(ctx, &nodeList); err != nil {
+		return fmt.Errorf("listing nodes, %w", err)
+	}
+	for i := range nodeList.Items {
+		var podlist v1.PodList
+		if err := t.kubeClient.List(ctx, &podlist, client.MatchingFields{"spec.nodeName": nodeList.Items[i].Name}); err != nil {
+			return fmt.Errorf("listing pods on node %s, %w", nodeList.Items[i].Name, err)
+		}
```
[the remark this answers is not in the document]
tzneal: My thought was requesting smaller batches might cause less memory consumption in etcd as we're somewhat paginating the results by node, but I think these are all cached anyway. @ellistarn  what do you think?

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,150 +18,328 @@ import (
...
-	topologyGroup.Register(domains...)
-	// This is a bit of a hack that allows the constraints to recognize viable hostname topologies
-	constraints.Requirements = constraints.Requirements.Add(v1.NodeSelectorRequirement{Key: topologyGroup.Constraint.TopologyKey, Operator: v1.NodeSelectorOpIn, Values: domains})
-	return nil
+		nextDomain, err := topology.Next(requirements, topology.Matches(p.Namespace, p.Labels))
+		if err != nil {
+			return v1alpha5.Requirements{}, err
+		}
+
+		requirements = requirements.Add(v1.NodeSelectorRequirement{
```
ellistarn, line 139: Can we return a set of requirements without mutating the passed in parameter and do the mutation outside?

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,150 +18,328 @@ import (
...
-	domains := []string{}
-	for i := 0; i < int(math.Ceil(float64(len(topologyGroup.Pods))/float64(topologyGroup.Constraint.MaxSkew))); i++ {
-		domains = append(domains, strings.ToLower(randomdata.Alphanumeric(8)))
-	}
-	topologyGroup.Register(domains...)
-	// This is a bit of a hack that allows the constraints to recognize viable hostname topologies
-	constraints.Requirements = constraints.Requirements.Add(v1.NodeSelectorRequirement{Key: topologyGroup.Constraint.TopologyKey, Operator: v1.NodeSelectorOpIn, Values: domains})
-	return nil
+		nextDomain, err := topology.Next(requirements, topology.Matches(p.Namespace, p.Labels))
+		if err != nil {
```
ellistarn, line 135: When is this errorable? Can we instead return a requirements object that is overly constrained? 

```
func (n *Node) Add(p *v1.Pod) error {
	podRequirements := v1alpha5.NewPodRequirements(p)

        requirements := n.Constraints.Requirements.Add(n.topology.Requirements(requirements, n.Name, p))
	if err := n.requirements.Compatible(podRequirements); err != nil {
		return err
	}
	requirements = requirements.Add(podRequirements.Requirements...)
        ...
```

tzneal: No, as this turns a "would violate max-skew" error into a "no zone in []" error.

pkg/controllers/provisioning/scheduling/topology.go
tzneal: Leaving comments here that we talked about for reminder/reference:

It doesn't mutate the incoming requirements, .Add() returns a new set of requirements that we keep adding to as we further tighten the requirements by topology.  We need to keep using this ever-tightening set of requirements as we have to provide it to the topology Next() method so it's aware of the valid domains that it can choose from.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,150 +18,336 @@ import (
...
+			Key:      topology.Key,
+			Operator: v1.NodeSelectorOpIn,
+			Values:   []string{nextDomain},
+		})
+	}
+	return requirements, nil
+}
+
+// RegisterNewHostname registers the new hostname with all registered topologies for the v1.LabelHostname key.
+func (t *Topology) RegisterNewHostname(hostname string) {
```
ellistarn, line 143:
```suggestion
func (t *Topology) Register(key, domain string) {
```

pkg/controllers/provisioning/scheduling/topology.go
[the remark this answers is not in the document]
ellistarn: These calls are hitting the ListWatch cache on the client side.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -18,150 +18,366 @@ import (
...
+	// for i, node := range nodeList.Items {
+	// 	if matched, _ := tg.requiredNodeAffinity.Match(&nodeList.Items[i]); matched {
+	// 		domain, ok := node.Labels[tg.Key]
+	// 		if !ok {
+	// 			continue // Don't include pods if node doesn't contain domain
+	// 		}
+	// 		nodes[node.Name] = &nodeList.Items[i]
+	// 		tg.Register(domain)
+	// 	}
+	// }
```
ellistarn, line 237:
```suggestion
```

tzneal:
Closing PR for a clean review on the next one.

Issue #1573, opened 2022-03-25
New Node creation stuck in a loop


vasu-git:
## Version

**Karpenter**: v0.7.3

**Kubernetes**: v1.21.5-eks-bc4871b

- I have a pretty basic eks cluster in aws. Deployed karpetenter and a default provisioner(only on-demand instances and no restriction on instance type. Cluster also has a few daemonsets deployed(splunk, ebs-csi-driver etc)
- Trying to deploy a new pod (prometheus) which also has a pvc volume. We use ebs-csi-driver addon in our cluster to provision volumes.
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    ieng.proofpoint.com/init-container-inserted: "true"
    kubernetes.io/psp: eks.privileged
  creationTimestamp: "2022-03-25T00:06:25Z"
  generateName: prometheus-prometheus-
  labels:
    app: prometheus
    app.kubernetes.io/instance: prometheus
    app.kubernetes.io/managed-by: prometheus-operator
    app.kubernetes.io/name: prometheus
    app.kubernetes.io/version: 2.24.0
  name: prometheus-prometheus-0
  namespace: prometheus-operator
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: StatefulSet
    name: prometheus-prometheus
    uid: 01cd86c4-3007-4e1b-80fe-207bfefedb02
  resourceVersion: "1977637"
  uid: d3767c88-9fba-45c8-969e-711298019caf
spec:
  containers:
  - args:
    - --web.console.templates=/etc/prometheus/consoles
    - --web.console.libraries=/etc/prometheus/console_libraries
    - --config.file=/etc/prometheus/config_out/prometheus.env.yaml
    - --storage.tsdb.path=/prometheus
    - --storage.tsdb.retention.time=7d
    - --web.enable-lifecycle
    - --query.max-concurrency=20
    - --query.timeout=2m
    - --web.external-url=prometheus-operator.dev-test8-labawsuse.lab.ppops.net
    - --web.route-prefix=/
    - --log.level=warn
    - --web.config.file=/etc/prometheus/web_config/web-config.yaml
    - --storage.tsdb.max-block-duration=2h
    - --storage.tsdb.min-block-duration=2h
    image: repocache.nonprod.ppops.net/dev-docker-local/prometheus/prometheus:v2.24.0
    imagePullPolicy: IfNotPresent
    name: prometheus
    ports:
[...]

PR #1616, opened 2022-04-01
Fix missing requests when only limits are supplied

dewjam:
**1. Issue, if available:**
#1573

Normally when a pod is created, if only Limits are supplied for a specific resource (no Requests), then the PodSpec is updated so the Requests = Limits when the pod event is created.  
For example ,if a manifest contains the below:
```
resources:
  limits:
    cpu: 1
```
The pod will be created as such:
```
resources:
  limits:
    cpu: 1
  requests:
    cpu: 1
```

However, when Karpenter calculates node overhead, it looks at the limits/requests in the PodSpec template of applicable DaemonSets.  If only Limits are supplied in the PodSpec template, then requests are never updated to equal limits for resources where requests are not defined.  The result is pods can be packed and scheduled onto an instance type which doesn't have enough resources and pods stay in a "Pending" state.

This change will merge resource limits into requests if there are no requests defined for a specific resource.

Added a tests to validate the behavior after the change.  Also tested manually with a DaemonSet with only limits to ensure it was scheduled on a node which has enough resources.

- [x] No


pkg/utils/resources/resources.go
```diff
@@ -26,7 +26,7 @@ func RequestsForPods(pods ...*v1.Pod) v1.ResourceList {
 	var resources []v1.ResourceList
 	for _, pod := range pods {
 		for _, container := range pod.Spec.Containers {
-			resources = append(resources, container.Resources.Requests)
+			resources = append(resources, MergeResourceLimitsIntoRequests(container))
```
dewjam, line 29: I decided to merge limits into requests here to catch all scenarios where requests are not populated when limits are set (other than just DaemonSets).

pkg/utils/resources/resources.go
```diff
@@ -63,6 +63,23 @@ func Merge(resources ...v1.ResourceList) v1.ResourceList {
...
 
+// MergeResourceLimitsIntoRequests merges resource limits into requests if no request exists for the given resource
+func MergeResourceLimitsIntoRequests(container v1.Container) v1.ResourceList {
+	if container.Resources.Requests == nil {
+		container.Resources.Requests = v1.ResourceList{}
+	}
+	if container.Resources.Limits == nil {
+		return container.Resources.Requests
+	}
+	resource := container.Resources.Requests
```
tzneal, line 74: I don't think we want to modify the resource list on the container here.  The client we use has a cache, so those modifications will be seen by us later.  It's safer (but slower) to not modify it and just return a new v1.ResourceList that has the correct values.

dewjam: Good call.  I'll take a look at that.

pkg/utils/resources/resources.go
```diff
@@ -63,6 +63,23 @@ func Merge(resources ...v1.ResourceList) v1.ResourceList {
 	return result
 }
 
+// MergeResourceLimitsIntoRequests merges resource limits into requests if no request exists for the given resource
+func MergeResourceLimitsIntoRequests(container v1.Container) v1.ResourceList {
```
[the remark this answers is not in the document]
dewjam: A `Ceiling` method is very similar to the `RequestsForPods` method which already exists.  I opted to stick with the existing method rather than rename it to `Ceiling`.  (although we could easily rename it if we wanted to).

pkg/utils/resources/resources.go
```diff
@@ -24,27 +24,37 @@ import (
...
+		}
+		for _, container := range pod.Spec.InitContainers {
+			init = append(init, MergeResourceLimitsIntoRequests(container))
 		}
 	}
-	merged := Merge(resources...)
-	merged[v1.ResourcePods] = *resource.NewQuantity(int64(len(pods)), resource.DecimalExponent)
-	return merged
+	containerResources := Merge(resources...)
+	initResources := MaxResources(init...)
```
dewjam, line 37: Oops, I don't think this logic is right.

dewjam: For each pod we need to:
1) Sum up non-InitContainer resources for the pod
2) Find the max resources for each resource type for InitContainers for the pod
3) Compare sum of "Container" resources against max of "InitContainer" to get the max resources required for the pod.

(RequestForPods then sums up the max resources for all of the pods.)

New revision incoming.

pkg/utils/resources/resources.go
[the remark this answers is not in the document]
dewjam: I ended up implementing a Ceiling method :).  Right now it takes in a single PodSpec and returns ResourceRequirements.

pkg/utils/resources/resources.go
dewjam: This is corrected in the most recent revision.

pkg/controllers/provisioning/suite_test.go
```diff
@@ -180,6 +180,70 @@ var _ = Describe("Provisioning", func() {
...
+						InitResourceRequirements: v1.ResourceRequirements{
+							Limits:   v1.ResourceList{v1.ResourceCPU: resource.MustParse("10000"), v1.ResourceMemory: resource.MustParse("10000Gi")},
+							Requests: v1.ResourceList{v1.ResourceCPU: resource.MustParse("1")},
+						},
+					}},
+				))
+				pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioningController, provisioner, test.UnschedulablePod(test.PodOptions{}))[0]
+				ExpectNotScheduled(ctx, env.Client, pod)
+			})
+			It("should not schedule if initContainer resources are too large", func() {
```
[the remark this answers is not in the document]
dewjam: ah, yes.  Nice catch.  This is supposed to be InitResourceRequirements.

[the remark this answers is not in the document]
dewjam: Resolved in latest revision.

bwagner5 (approved):
lgtm

PR #1623, opened 2022-04-04
support baremetal instance types

bwagner5:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/1336
https://github.com/aws/karpenter/issues/1371

Support Bare Metal Instance Types. 
 
Before this change, Bare Metal was filtered out of the universe of instance types making it impossible to use Bare Metal (even if directly specified in the Provisioner `node.kubernetes.io/instance-type` requirements). 
 
This PR removes the bare metal filter and instead prioritizes Bare Metal instance types as a last resort in the AWS Cloud Provider (similar to the way GPUs are utilized). Since bare metal instance types have the exact same resource specs as the largest instance type in an instance family, bare metal will only engage if the Provisioner is properly constrained to only bare metal instance types. 

 - Manually in an EKS cluster:
   -  I scaled an inflate deployment without constraining a provisioner and examined the ec2 Fleet request where I saw that metal instance types were not in the request
   - I constrained the provisioner to only m5.metal and scaled an inflate deployment. An m5.metal was provisioned. 

- [ ] No

### DIFF
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -321,20 +321,24 @@ func (p *InstanceProvider) getCapacityType(nodeRequest *cloudprovider.NodeReques
 	return v1alpha1.CapacityTypeOnDemand
 }
 
-// filterInstanceTypes is used to eliminate GPU instance types from the list of possible instance types when a
-// non-GPU instance type will work.  If the list of instance types consists of both GPU and non-GPU types, then only
-// the non-GPU types will be returned.  If it has only GPU types, the list will be returned unaltered.
+// filterInstanceTypes is used to eliminate less desirable instance types (like GPUs) from the list of possible instance types when
+// a set of more appropriate instance types would work. If a set of more desirable instance types is not found, then the original slice
+// of instance types are returned.
 func (p *InstanceProvider) filterInstanceTypes(instanceTypes []cloudprovider.InstanceType) []cloudprovider.InstanceType {
 	var genericInstanceTypes []cloudprovider.InstanceType
 	for _, it := range instanceTypes {
-		itRes := it.Resources()
-		if resources.IsZero(itRes[v1alpha1.ResourceAWSNeuron]) &&
-			resources.IsZero(itRes[v1alpha1.ResourceAMDGPU]) &&
-			resources.IsZero(itRes[v1alpha1.ResourceNVIDIAGPU]) {
-			genericInstanceTypes = append(genericInstanceTypes, it)
+		if aws.BoolValue(it.(*InstanceType).BareMetal) {
+			continue
 		}
+		itRes := it.Resources()
+		if !resources.IsZero(itRes[v1alpha1.ResourceAWSNeuron]) ||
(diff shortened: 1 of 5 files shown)


pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -147,6 +147,10 @@ func (p *InstanceTypeProvider) getInstanceTypes(ctx context.Context) (map[string
 				Name:   aws.String("supported-virtualization-type"),
 				Values: []*string{aws.String("hvm")},
 			},
+			{
+				Name:   aws.String("processor-info.supported-architecture"),
+				Values: aws.StringSlice([]string{"x86_64", "arm64"}),
```
bwagner5, line 152: this filter is needed to remove weird archs like `x86_64_mac`

pkg/cloudprovider/aws/instance.go
```diff
@@ -321,20 +321,24 @@ func (p *InstanceProvider) getCapacityType(nodeRequest *cloudprovider.NodeReques
...
-// non-GPU instance type will work.  If the list of instance types consists of both GPU and non-GPU types, then only
-// the non-GPU types will be returned.  If it has only GPU types, the list will be returned unaltered.
+// filterInstanceTypes is used to eliminate less desirable instance types (like GPUs) from the list of possible instance types when
+// a set of more appropriate instance types would work. If a set of more desirable instance types is not found, then the original slice
+// of instance types are returned.
 func (p *InstanceProvider) filterInstanceTypes(instanceTypes []cloudprovider.InstanceType) []cloudprovider.InstanceType {
 	var genericInstanceTypes []cloudprovider.InstanceType
 	for _, it := range instanceTypes {
+		if aws.BoolValue(it.(*InstanceType).BareMetal) {
+			continue
```
suket22, line 331: So if I had - 
```
  requirements:
    - key: "node.kubernetes.io/instance-type"
      operator: In
      values: ["m5.24xlarge", "m5.metal"]
```

I'm never getting a `m5.metal`, since we'll take it out of the list of instances being passed to Fleet? 


I can't think of why someone would have a combination of gpu and non-gpu instance constraints in a single provisioner, but it makes me wonder why we're doing any filtering at all.

bwagner5: hmmm that's interesting. For GPUs, you could still have non-GPUs and GPUs mixed in instance-type requirements since the pod resource request for a GPU would do a filtering before this func. But you're right that baremetal would never be used if specified with other non-baremetal instance types that also had the resource requirements (which would be likely).  

I think the GPU case is the most convincing argument to do the filtering the way it is now. I also can't think of a reason for mixing baremetal specifically.

tzneal: You can possibly get it in the event that the m5.24xlarge is not available in spot.  Not sure if that logic works the same way for on-demand though.

bwagner5: @tzneal ah you're right and it would work the same for OD

suket22 (approved):

Issue #985, opened 2021-12-14
podAffinity


olemarkus:
**Tell us about your request**
What do you want us to build?

I know you'll love this: Support `podAffinity`

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
This is a sibling issue to #942 , but specifically on `podAffinity`.

Cilium's hubble has component called hubble-relay that needs to run on the same node as a cilium agent.
The cilium install manifests has this podAffinity as non-optional: https://github.com/cilium/cilium/blob/master/install/kubernetes/cilium/templates/hubble-relay/deployment.yaml#L34-L43

This is a fairly essential component to those using Cilium.

On many clusters, one can assume cilium agent on all nodes, but there are clusters with Fargate nodes and Windows nodes using Cilium, where the agent wouldn't run.

**Are you currently working around this issue?**

In our case, we can patch out that podAffinity.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

tmokmss:
+1 to podAffinity support

We're investigating if we can use Karpenter with Agones, and found that Agones use podAffinity for pod scheduling to pack as much as possible into the smallest set of nodes. [doc](https://agones.dev/site/docs/advanced/scheduling-and-autoscaling/#pod-scheduling-strategy)
Because Karpenter currently ignores pods with podAffinity, we cannot use Karpenter to provision nodes for Agones pods.

ellistarn:
What if the pods aren't created at the same time? If Karpenter only sees the first pod, it will launch a node for it. If the next pod c omes in a minute later, it's too late for that pod to be included in the scheduling algorithm. This doesn't exist for anti-affinity because scheduling apart achieves the same objective.

I'd love to make progress on podaffinity, but I don't have a viable path forward. How are you envisioning it?

tmokmss:
Hi thanks for the suggestion. As the doc says 

> The default Kubernetes scheduler doesn’t do a perfect job of packing, but it’s a good enough job for what we need - at least at this stage.

Agones itself admits it isn't perfect, but they seem to know it works well at most cases.

ellistarn:
[...]

PR #1626, opened 2022-04-05
Add affinity and anti-affinity support

### LINKED ISSUE #942: podAntiAffinity

tzneal:
**1. Issue, if available:**

Fixes #942 and #985 

Adds support for pod affinity and anti-affinity

Unit testing and on my local EKS cluster.

- [ ] No


pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -394,13 +394,29 @@ func TopologyListOptions(namespace string, labelSelector *metav1.LabelSelector)
 		selector = selector.Add(*requirement)
 	}
 	for _, expression := range labelSelector.MatchExpressions {
-		requirement, err := labels.NewRequirement(expression.Key, selection.Operator(expression.Operator), expression.Values)
+		requirement, err := labels.NewRequirement(expression.Key, mapOperator(expression.Operator), expression.Values)
```
tzneal, line 397: This branch added the runtime.Must() to the result of NewRequirement, previously we just ignored the error assuming it couldn't happen.  It did, and we ended up just ignoring all match expressions.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -394,13 +394,29 @@ func TopologyListOptions(namespace string, labelSelector *metav1.LabelSelector)
...
 		runtime.Must(err)
 		selector = selector.Add(*requirement)
 	}
 	return &client.ListOptions{Namespace: namespace, LabelSelector: selector}
 }
 
+func mapOperator(operator metav1.LabelSelectorOperator) selection.Operator {
```
ellistarn, line 404: Isn't this the default behavior of what you were doing before?

tzneal: Nope, the strings are different:

```go
package metav1
const (
	LabelSelectorOpIn           LabelSelectorOperator = "In"
	LabelSelectorOpNotIn        LabelSelectorOperator = "NotIn"
	LabelSelectorOpExists       LabelSelectorOperator = "Exists"
	LabelSelectorOpDoesNotExist LabelSelectorOperator = "DoesNotExist"
)
```
```go
package selection
const (
	DoesNotExist Operator = "!"
	Equals       Operator = "="
	DoubleEquals Operator = "=="
	In           Operator = "in"
	NotEquals    Operator = "!="
	NotIn        Operator = "notin"
	Exists       Operator = "exists"
	GreaterThan  Operator = "gt"
	LessThan     Operator = "lt"
)
```

pkg/apis/provisioning/v1alpha5/requirements.go
```diff
@@ -101,12 +101,41 @@ func (r Requirements) Add(requirements ...v1.NodeSelectorRequirement) Requiremen
...
-		r.requirements[requirement.Key] = values
+		cp.requirements[requirement.Key] = values
+	}
+	cp.rebuild()
+	return cp
+}
+
+// rebuild re-generates the node selector requirements based on the set based versions.  This improves scheduling speed
+// as it causes us to not carry around redundant requirements (e.g. 20x copies of instance-type in [it0, it1, ..., it400])
+func (r *Requirements) rebuild() {
```
ellistarn, line 115: I think we should refactor requirements in general. Perhaps worth punting on this until then. Happy to leave this in, but it bloats an already massive PR.

pkg/controllers/provisioning/scheduling/topologynodefilter.go
```diff
@@ -27,102 +27,60 @@ import (
...
-}
+type TopologyNodeFilter []v1alpha5.Requirements
 
-func NewTopologyNodeFilter(p *v1.Pod) *TopologyNodeFilter {
-	selector := &TopologyNodeFilter{
-		nodeSelector: p.Spec.NodeSelector,
-	}
-	if p.Spec.Affinity != nil && p.Spec.Affinity.NodeAffinity != nil && p.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
-		selector.nodeSelectorTerms = p.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+func MakeTopologyNodeFilter(p *v1.Pod) TopologyNodeFilter {
```
ellistarn, line 32:
```suggestion
func NewTopologyNodeFilter(p *v1.Pod) TopologyNodeFilter {
```

pkg/controllers/provisioning/scheduling/topologynodefilter.go
```diff
@@ -27,102 +27,60 @@ import (
...
 	runtime.Must(err)
 	return hash
 
 }
 
-func (t *TopologyNodeFilter) matchesRequirements(requirement v1alpha5.Requirements, node *v1.Node) bool {
-	for key := range requirement.Keys() {
-		if !requirement.Get(key).Has(node.Labels[key]) {
-			return false
+func (t TopologyNodeFilter) isCompatible(requirements v1alpha5.Requirements) bool {
```
ellistarn, line 74: Thoughts on parity with the Requirements interface?
```suggestion
func (t TopologyNodeFilter) Compatible(requirements v1alpha5.Requirements) bool {
```

pkg/controllers/provisioning/scheduling/topologynodefilter.go
```diff
@@ -27,102 +27,60 @@ import (
...
+		return true
+	}
+	// these are an OR, so if any passes the filter passes
+	for _, req := range t {
+		if err := requirements.Compatible(req); err == nil {
+			return true
 		}
 	}
-	return true
+	return false
```
ellistarn, line 85: Optional, you could:
```
  return len(t) == 0
```

pkg/controllers/provisioning/scheduling/topologynodefilter.go
```diff
@@ -27,102 +27,60 @@ import (
...
-				break
-			}
-		}
-		// these terms are OR'd together, so if we match one full set of requirements, the filter passes
-		if matchesAllReqs {
-			return true
-		}
-	}
-	return true
+func (t TopologyNodeFilter) MatchesRequirements(nodeRequirements v1alpha5.Requirements) bool {
```
ellistarn, line 59: Why not just use the Compatible interface?

pkg/controllers/provisioning/scheduling/topologynodefilter.go
```diff
@@ -27,102 +27,60 @@ import (
...
-			return true
-		}
-	}
+	return filter
+}
 
-	// at this point we have node selector terms, but didn't match all of the requirements for any individual term
-	return false
+// Matches returns true if the TopologyNodeFilter doesn't prohibit node from the participating in the topology
+func (t TopologyNodeFilter) Matches(node *v1.Node) bool {
```
ellistarn, line 51: This is used in one place. Thoughts on removing this and just using the Compatible interface in that case? 
```
nodeFilter.Compatible(v1alpha5.NewLabelRequirements(node.Labels))
```

pkg/controllers/provisioning/scheduling/topologynodefilter.go
tzneal: `New.*` is typically reserved for methods that return a pointer interface while `Make.*` is for methods that return value types.  I'm good either way.

pkg/controllers/provisioning/scheduling/topologynodefilter.go
tzneal: The concept of a filter being "Compatible" with a node is less intuitive to me than the concept of a filter "Matching" a node.

pkg/controllers/provisioning/scheduling/topologynodefilter.go
tzneal: I don't think it makes as much sense to ask if a filter is 'Compatible' with something compared to asking if a filter 'Matches' something.

pkg/apis/provisioning/v1alpha5/requirements.go
tzneal: It greatly affects performance, I'd like to leave it in this one for now, or possibly just cherry-pick it to main to reduce this PRs size.  Need to get the benchmark CI in as it's very easy to make a change and send scheduling performance through the floor.

pkg/controllers/provisioning/scheduling/topologynodefilter.go
ellistarn: I'm not familiar with this. Is this from the Go compiler?

pkg/controllers/provisioning/scheduling/topologynodefilter.go
ellistarn: The intent of this comment is to reduce the number of concepts that a reader needs to be aware of. Naming of those concepts can happen later. It's not clear to me that we have multiple concepts we're dealing with.

pkg/controllers/provisioning/scheduling/topologynodefilter.go
tzneal: That's one place, it's not 100% followed consistently but it seems odd to have a New.*() method that doesn't return a pointer which is why I went with Make()

Go compiler
```
src/cmd/compile/internal/syntax/pos.go:24:func MakePos(base *PosBase, line, col uint) Pos { return Pos{base, sat32(line), sat32(col)} }
src/cmd/internal/src/pos.go:381:func makeLicoRaw(line, col uint) lico {
src/reflect/value.go:2912:func MakeSlice(typ Type, len, cap int) Value {
src/reflect/value.go:2931:func MakeChan(typ Type, buffer int) Value {
src/reflect/value.go:2947:func MakeMap(typ Type) Value {
src/reflect/value.go:2953:func MakeMapWithSize(typ Type, n int) Value {
src/reflect/value.go:3438:func makechan(typ *rtype, size int) (ch unsafe.Pointer)
```

Other projects use that idiom it as well:
```
vendor/go.starlark.net/starlark/int.go:43:func MakeInt64(x int64) Int {
vendor/go.starlark.net/starlark/int.go:51:func MakeUint(x uint) Int { return MakeUint64(uint64(x)) }
vendor/go.starlark.net/starlark/int.go:54:func MakeUint64(x uint64) Int {
vendor/go.starlark.net/starlark/int.go:67:func MakeBigInt(x *big.Int) Int {
```

pkg/controllers/provisioning/scheduling/topologynodefilter.go
tzneal: To me it's about readability of the function when used, the first is far more clear than the second:

```go
// nodes may or may not be considered for counting purposes for topology spread constraints depending on if they
// are selected by the pod's node selectors and required node affinities.  If these are unset, the node always counts.
if !tg.nodeFilter.Matches(node) {
        continue
}
```


```go
// nodes may or may not be considered for counting purposes for topology spread constraints depending on if they
// are selected by the pod's node selectors and required node affinities.  If these are unset, the node always counts.
if err := tg.nodeFilter.Compatible(node); err != nil  {
        continue
}
```

pkg/controllers/provisioning/scheduling/topologynodefilter.go
```diff
@@ -0,0 +1,82 @@
...
+	// these are an OR, so if any passes the filter passes
+	for _, req := range t {
+		if err := requirements.Compatible(req); err == nil {
+			return true
+		}
+	}
+	return false
+}
+
+func (t TopologyNodeFilter) Hash() uint64 {
```
ellistarn, line 73: Do we still need this hash function now that it's just an alias type?

tzneal: Good catch, removed

pkg/controllers/provisioning/scheduling/topologynodefilter.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
+)
+
+// TopologyNodeFilter is used to determine if a given actual node or scheduling node matches the pod's node selectors
+// and required node affinity terms.  This is used with topology spread constraints to determine if the node should be
+// included for topology counting purposes. This is only used with topology spread constraints as affinities/anti-affinities
+// always count across all nodes. A nil or zero-value TopologyNodeFilter behaves well and the filter returns true for
+// all nodes.
+type TopologyNodeFilter []v1alpha5.Requirements
```
ellistarn, line 28: Now that this type has been simplified, it occurs to me that it might be more straightforward to fold this into topologygroup e.g. `nodeFilter []v1alpha5.Requirements`. It should be ~40 lines of code.

ellistarn (approved):
Let's get one more approver, but I'm good to go.

pkg/controllers/provisioning/scheduling/preferences.go
```diff
@@ -43,39 +43,45 @@ func NewPreferences() *Preferences {
...
-		func(pod *v1.Pod) *string { return p.removePreferredPodAntiAffinityTerm(pod) },
-		func(pod *v1.Pod) *string { return p.removePreferredNodeAffinityTerm(pod) },
-		func(pod *v1.Pod) *string { return p.removeRequiredNodeAffinityTerm(pod) },
-		func(pod *v1.Pod) *string { return p.toleratePreferNoScheduleTaints(pod) },
+		p.removeRequiredNodeAffinityTerm,
+		p.removePreferredPodAffinityTerm,
+		p.removePreferredPodAntiAffinityTerm,
+		p.removePreferredNodeAffinityTerm,
+		p.removeTopologySpreadScheduleAnyway,
+		p.toleratePreferNoScheduleTaints,
```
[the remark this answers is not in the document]
tzneal: Yes, it can definitely change the result.   There isn't an order specified by K8s, they perform scheduling in a different way from what I can tell (find all compatible nodes, sort by score).  We schedule on the first compatible node treating all preferences as required, and removing one preference at a time until it either schedules or there are no more to remove and it fails.

pkg/controllers/provisioning/scheduling/preferences.go
```diff
@@ -100,6 +106,34 @@ func (p *Preferences) removePreferredNodeAffinityTerm(pod *v1.Pod) *string {
...
+		pod.Spec.Affinity.NodeAffinity == nil ||
+		pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil ||
+		len(pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms) == 0 {
+		return nil
+	}
+	terms := pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+	// Remove the first term if there's more than one (terms are an OR semantic), Unlike preferred affinity, we cannot remove all terms
+	if len(terms) > 1 {
+		pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms = terms[1:]
+		return ptr.String(fmt.Sprintf("removing: spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution[0]=%s", pretty.Concise(terms[0])))
```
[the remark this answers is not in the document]
tzneal: It only gets logged if the pod has failed to schedule.  Most of the time this will never log anything. Preferences is a little odd , look at like 61 or so.  The first time it's called, it doesn't relax anything.

pkg/controllers/provisioning/scheduling/queue.go
```diff
@@ -0,0 +1,88 @@
...
+// pod affinities that occur in a batch of pods if there are enough constraints provided.
+type Queue struct {
+	pods       []*v1.Pod
+	lastPopped *v1.Pod
+	attempts   int
+}
+
+// NewQueue constructs a new queue given the input pods, sorting them to optimize for bin-packing into nodes.
+func NewQueue(pods ...*v1.Pod) *Queue {
+	sort.Slice(pods, byCPUAndMemoryDescending(pods))
```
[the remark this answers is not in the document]
tzneal: Yes, there are some ideas around "vector bin packing" that I want to investigate at some point that will probably improve upon this.  It will take writing some benchmark code to ensure that we are actually packing more densely, but a different sort order could improve.  

We may also want to sort pods with anti-affinity or topology spreads first.  They are more likely to create new nodes which we can then use to fulfill other scheduling constraints.  Lots of options to investigate and determine what works best.

pkg/controllers/provisioning/scheduling/queue.go
```diff
@@ -0,0 +1,88 @@
...
+
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/aws/karpenter/pkg/utils/resources"
+)
+
+// Queue is a queue of pods that is scheduled.  It's used to attempt to schedule pods as long as we are making progress
+// in scheduling. This is sometimes required to maintain zonal topology spreads with constrained pods, and can satisfy
+// pod affinities that occur in a batch of pods if there are enough constraints provided.
+type Queue struct {
```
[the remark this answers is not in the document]
tzneal: Scheduling is currently single-threaded and the queue is created to operate on a single batch of pods.  The launching of nodes later on is parallel, but this part isn't. The topology groups, etc. aren't thread safe and making them that way could potentially lead to more scheduling violations.  E.g. multiple goroutines observe the same topology count and then schedule a pod to the same zone, which would now violate a spread.


This problem exists currently as well as we can race kube-scheduler.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -50,88 +51,115 @@ func init() {
...
 	}
-	return nodeSet.nodes, nil
+	return nodes, nil
 }
 
-func byPrice(instanceTypes []cloudprovider.InstanceType) func(i int, j int) bool {
-	return func(i, j int) bool {
-		return instanceTypes[i].Price() < instanceTypes[j].Price()
+func (s *Scheduler) scheduleExisting(pod *v1.Pod, nodes []*Node) *Node {
+	// Try nodes in ascending order of number of pods to more evenly distribute nodes, 100ms at 2000 nodes.
```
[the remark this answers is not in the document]
tzneal: It may vary depending on CPU, but that should be it.  We're just sorting s list of pointers by the pod count which is a constant operation to retrieve.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -50,88 +51,115 @@ func init() {
...
-		for _, node := range nodeSet.nodes {
-			if err := node.Add(pod); err == nil {
-				isScheduled = true
-				break
-			}
+	// We loop and retrying to schedule to unschedulable pods as long as we are making progress.  This solves a few
+	// issues including pods with affinity to another pod in the batch. We could topo-sort to solve this, but it wouldn't
+	// solve the problem of scheduling pods where a particular order is needed to prevent a max-skew violation. E.g. if we
+	// had 5xA pods and 5xB pods were they have a zonal topology spread, but A can only go in one zone and B in another.
+	// We need to schedule them alternating, A, B, A, B, .... and this solution also solves that as well.
```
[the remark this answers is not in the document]
tzneal: > Finding it hard to understand the code, so I'll ask here. Does this code work with an arbitrary amount of zonal topology requirements? Say if we had `5xA, 5xB, 5xC, 5xD, 5xE` pods, would this eventually schedule them with alternating order: `A,B,C,D,E,A,B,C,D,E,A....`?

Yes, but even the example in the comment here is an edge case.  With a normal topology spread, all of the pods in the deployment have the same restrictions.  This case only occurs if you have multiple deployments that use the same topology spread via label selectors, but each deployment has a different zonal restriction.  It's super unlikely and doesn't make sense, but we will still schedule it, though slowly.  The more common case is pod A has affinity to pod B, in which case we need to schedule B and then make one more pass to schedule A.

> Additionally, does this mean the algorithm is roughly `O(n^m)` where `m` is the cardinality of the topologies (e.g. `cardinality({A,B,C,D,E}) = 5`) and `n` is the amount of pod in each?

I don't think it's that bad as we don't care what order we schedule A,B,C,D,E in. We just have to schedule one of {A,B,C,D,E} before we schedule any other.

So if you try to make a bad input case case (using 3 deployments and 3 pods each for simplicity):

AAABBBCCC

First round we schedule A,B,C and then the same in each subsequent round for three rounds of scheduling.  I think any other sort order will schedule faster. I haven't fully looked at it in detail, but gut feel is O(nm) on these weird edge cases.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -50,88 +51,115 @@ func init() {
...
+			node = NewNode(constraints, topology, daemonOverhead, instanceTypes)
+			// we created a new hostname, so ensure all topologies are aware
+			topology.Register(v1.LabelHostname, node.Hostname)
+
+			if errors[pod] = node.Add(pod); errors[pod] != nil {
+				relaxed := s.preferences.Relax(ctx, pod)
+				if relaxed {
+					// The pod has changed, so topology needs to be recomputed
+					if err := topology.Update(ctx, pod); err != nil {
+						logging.FromContext(ctx).With("pod", client.ObjectKeyFromObject(pod)).Errorf("updating topology, %s", err)
```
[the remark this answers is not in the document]
tzneal: Only logs on topology update error, IIRC that error only occurs when a kube-api server call fails which shouldn't occur often.

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
```diff
@@ -163,8 +195,8 @@ func makeDiversePods(count int) []*v1.Pod {
 	pods = append(pods, makeTopologySpreadPods(count/7, v1.LabelHostname)...)
 	pods = append(pods, makePodAffinityPods(count/7, v1.LabelHostname)...)
 	pods = append(pods, makePodAffinityPods(count/7, v1.LabelTopologyZone)...)
-	pods = append(pods, makePodAntiAffinityPods(count/7, v1.LabelHostname)...)
-	pods = append(pods, makePodAntiAffinityPods(count/7, v1.LabelTopologyZone)...)
+	// We intentionally don't do anti-affinity by zone as that creates tons of unschedulable pods.
+	//pods = append(pods, makePodAntiAffinityPods(count/7, v1.LabelTopologyZone)...)
```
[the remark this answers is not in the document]
tzneal: It's a benchmark test and I left it in for now as I keep adding/removing that line to perform some benchmarking just to get the numbers.

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -17,52 +17,222 @@ package scheduling
...
+	case TopologyTypePodAffinity:
+		return "pod affinity"
+	case TopologyTypePodAntiAffinity:
+		return "pod anti-affinity"
 	}
+	return ""
 }
 
-// TopologyGroup is a set of pods that share a topology spread constraint
+// TopologyGroup is used to track pod counts that match a selector by the topology domain (e.g. SELECT COUNT(*) FROM pods GROUP BY(topology_ke
```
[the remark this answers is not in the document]
tzneal: 😂

pkg/controllers/provisioning/scheduling/scheduler.go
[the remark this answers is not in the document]
ellistarn: Keep in mind. This call is cached. It should only fail due to a code bug.

pkg/controllers/provisioning/scheduling/queue.go
```diff
@@ -0,0 +1,88 @@
...
+	} else {
+		q.attempts--
+	}
+}
+
+func (q *Queue) List() []*v1.Pod {
+	return q.pods
+}
+
+func byCPUAndMemoryDescending(pods []*v1.Pod) func(i int, j int) bool {
```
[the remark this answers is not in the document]
tzneal: Nothing else could use this currently, this is only for bin-packing the pods.  It may be worthwhile moving it to another file if we have multiple strategies and lets users pick, or try multiple and choose the best strategy, but I think it's ok for now.

pkg/controllers/provisioning/scheduling/scheduler.go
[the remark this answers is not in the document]
njtran: Thanks for your detailed response. I wanted to make sure that this was going to work for `m>3` topology groups. Seems good to me.

BeardyBear:
Hi, how long would it take for this to be released as part of new Karpenter version?
Is there a way I could run it now or should I just wait for the release?
We really need this feature.

Thank you :)

ellistarn (approved):

tzneal:
> Hi, how long would it take for this to be released as part of new Karpenter version? Is there a way I could run it now or should I just wait for the release? We really need this feature.
> 
> Thank you :)

We plan to make a snapshot image available soon so it can be tested out before the next release.

joebowbeer:
Also closes #985 ?

PR #1638, opened 2022-04-07
Only apply PreferNoSchedule toleration if it could help

tzneal:
**1. Issue, if available:**

N/A

Check for provisioners before relaxing to prevent confusing error messages and fully relaxing pod before a  provisioner even exists.  Only add the toleration for a `PreferNoSchedule` taint if one of the provisioners would apply the corresponding taint.

Unit test & on EKS

- [X] No

### DIFF
--- a/pkg/controllers/selection/controller.go
+++ b/pkg/controllers/selection/controller.go
@@ -84,17 +84,21 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 }
 
 func (c *Controller) selectProvisioner(ctx context.Context, pod *v1.Pod) (errs error) {
+	// Pick provisioner
+	var provisioner *provisioning.Provisioner
+	provisioners := c.provisioners.List(ctx)
+	if len(provisioners) == 0 {
+		// checking for existing provisioners first prevents relaxing pods which logs messages that make it appear
+		// Karpenter is doing something even when you don't have a provisioner created.  If you create the provisioner
+		// later, the pod may have already been fully relaxed unnecessarily.
+		return nil
+	}
+
 	// Relax preferences if pod has previously failed to schedule.
 	c.preferences.Relax(ctx, pod)
 	// Inject volume topological requirements
 	if err := c.volumeTopology.Inject(ctx, pod); err != nil {
 		return fmt.Errorf("getting volume topology requirements, %w", err)
-	}
-	// Pick provisioner
-	var provisioner *provisioning.Provisioner
-	provisioners := c.provisioners.List(ctx)
-	if len(provisioners) == 0 {
-		return nil
 	}
 	for _, candidate := range c.provisioners.List(ctx) {
 		if err := candidate.Spec.DeepCopy().ValidatePod(pod); err != nil {
--- a/pkg/controllers/selection/controller.go
+++ b/pkg/controllers/selection/controller.go
@@ -89,13 +89,12 @@ func (c *Controller) selectProvisioner(ctx context.Context, pod *v1.Pod) (errs e
 	provisioners := c.provisioners.List(ctx)
 	if len(provisioners) == 0 {
 		// checking for existing provisioners first prevents relaxing pods which logs messages that make it appear
-		// Karpenter is doing something even when you don't have a provisioner created.  If you create the provisioner
-		// later, the pod may have already been fully relaxed unnecessarily.
+		// Karpenter is attempting to schedule even when you don't have a provisioner created.  This also prevents an issue
+		// where if you created a provisioner later, the pod may have already been fully relaxed unnecessarily.
 		return nil
 	}
-
 	// Relax preferences if pod has previously failed to schedule.
-	c.preferences.Relax(ctx, pod)
+	c.preferences.Relax(ctx, pod, getRelaxOptions(provisioners))
 	// Inject volume topological requirements
 	if err := c.volumeTopology.Inject(ctx, pod); err != nil {
 		return fmt.Errorf("getting volume topology requirements, %w", err)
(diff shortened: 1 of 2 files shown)


pkg/controllers/selection/controller.go
```diff
@@ -84,18 +84,33 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	for _, prov := range provisioners {
+		for _, taint := range prov.Spec.Taints {
+			if taint.Effect == v1.TaintEffectPreferNoSchedule {
+				hasPreferNoScheduleTaint = true
+			}
+		}
+	}
+
+	// Relax preferences if pod has previously failed to schedule.
+	c.preferences.Relax(ctx, pod, hasPreferNoScheduleTaint)
```
bwagner5, line 109: If we would expand these conditions later, it might be cleaner to pass the provisioner to the Relax func and throw the `hasPreferNoScheduleTaint` into a func. Wdyt?

tzneal: I thought about that, but it's actually a list of provisioners that we would need to pass in.  That seemed a bit more weird to me, but I'm good either way.  Back to you, Wdyt? 😂

pkg/controllers/selection/preferences.go
```diff
@@ -64,18 +64,23 @@ func (p *Preferences) Relax(ctx context.Context, pod *v1.Pod) {
...
+func (p *Preferences) relax(ctx context.Context, pod *v1.Pod, addPreferNoScheduleToleration bool) bool {
 	for _, relaxFunc := range []func(*v1.Pod) *string{
 		func(pod *v1.Pod) *string { return p.removePreferredPodAffinityTerm(pod) },
 		func(pod *v1.Pod) *string { return p.removePreferredPodAntiAffinityTerm(pod) },
 		func(pod *v1.Pod) *string { return p.removePreferredNodeAffinityTerm(pod) },
 		func(pod *v1.Pod) *string { return p.removeRequiredNodeAffinityTerm(pod) },
-		func(pod *v1.Pod) *string { return p.toleratePreferNoScheduleTaints(pod) },
+		func(pod *v1.Pod) *string {
+			if addPreferNoScheduleToleration {
+				return p.toleratePreferNoScheduleTaints(pod)
```
ellistarn, line 80: Seems like this check out encapsulate `addPreferNoScheduleToleration`, rather than having this logic in the outer for loop.

pkg/controllers/selection/controller.go
bwagner5: I think the slice of provisioners is fine to pass to `Relax()`. To me, it's cleaner to think about Relax taking into account the things passed to it rather than exposing that directly in the `selectProvisioner()` and passing a flag.

pkg/controllers/selection/preferences.go
tzneal: Turned it into a generic options struct, seems to make more sense.

pkg/controllers/selection/controller.go
tzneal: Turned it into a generic options struct, seems to make more sense that way.  We can expand it if needed without making preferences aware of provisioners.

pkg/controllers/selection/controller.go
```diff
@@ -114,6 +117,21 @@ func (c *Controller) selectProvisioner(ctx context.Context, pod *v1.Pod) (errs e
 	return nil
 }
 
+func getRelaxOptions(provisioners []*provisioning.Provisioner) RelaxOptions {
```
ellistarn, line 120: This breaks the encapsulation of `preferences.go`. What can do to maintain that encapsulation?

tzneal: Sorry, not following this comment. RelaxOptions is declared in preferences.go and I introduced it to avoid having preferences know about provisioners.

ellistarn:
> fully relaxing pod before a provisioner even exists.

This signals to me that we should directly fix how relaxation interacts with provisioners.

pkg/controllers/selection/preferences.go
```diff
@@ -60,22 +67,25 @@ func (p *Preferences) Relax(ctx context.Context, pod *v1.Pod) {
...
+func (p *Preferences) relax(ctx context.Context, pod *v1.Pod, options RelaxOptions) bool {
 	for _, relaxFunc := range []func(*v1.Pod) *string{
 		func(pod *v1.Pod) *string { return p.removePreferredPodAffinityTerm(pod) },
 		func(pod *v1.Pod) *string { return p.removePreferredPodAntiAffinityTerm(pod) },
 		func(pod *v1.Pod) *string { return p.removePreferredNodeAffinityTerm(pod) },
 		func(pod *v1.Pod) *string { return p.removeRequiredNodeAffinityTerm(pod) },
-		func(pod *v1.Pod) *string { return p.toleratePreferNoScheduleTaints(pod) },
+		func(pod *v1.Pod) *string {
+			return p.toleratePreferNoScheduleTaints(pod, options)
+		},
```
dewjam, line 88: Do we want to revisit the order in which we relax preferences?  I know when I first pushed this change, the order didn't seem to matter, but maybe it would it make sense to tolerate PreferNoSchedule taints somewhere before relaxing "Required" node affinity terms.

tzneal:
> > fully relaxing pod before a provisioner even exists.
> 
> This signals to me that we should directly fix how relaxation interacts with provisioners creation. I think this may resolve when we refactor selection into the scheduler.

Yeah, when it's there we also won't have this need for a 'prepare for relaxation, but don't do anything' step.

pkg/controllers/selection/preferences.go
tzneal: It probably makes sense, the prefer no schedule should likely be first as it will be what causes the provisioner to not be acceptable.  The required node affinity terms are removed one at a time, until only one is left, the last is never removed.  Still not perfect, but that's the implementation for now.

PR #1641, opened 2022-04-07
Add pdb to helm chart

rothgar:
**1. Issue, if available:**

Adds a pod disruption budget to the helm chart.

It's not exposed as a variable/value. Should it be?

Googling and copy pasting 

- [x] No

No, unless we want to add a variable to disable it or change the value. I'm not sure that's a good idea.

### DIFF
--- a/charts/karpenter/templates/poddisruptionbudget.yaml
+++ b/charts/karpenter/templates/poddisruptionbudget.yaml
@@ -3,7 +3,7 @@ kind: PodDisruptionBudget
 metadata:
   name: karpenter
 spec:
-  minAvailable: 1
+  minAvailable: {{ .Values.podDisruptionBudget.minAvailable }}
   selector:
     matchLabels:
     {{- include "karpenter.labels" . | nindent 4 }}
--- a/charts/karpenter/templates/poddisruptionbudget.yaml
+++ b/charts/karpenter/templates/poddisruptionbudget.yaml
@@ -6,5 +6,5 @@ spec:
   maxUnavailable: {{ .Values.podDisruptionBudget.maxUnavailable }}
   selector:
     matchLabels:
-    {{- include "karpenter.labels" . | nindent 4 }}
+    {{- include "karpenter.selectorLabels" . | nindent 6 }}
 
--- a/charts/karpenter/templates/poddisruptionbudget.yaml
+++ b/charts/karpenter/templates/poddisruptionbudget.yaml
@@ -3,7 +3,7 @@ kind: PodDisruptionBudget
 metadata:
   name: karpenter
 spec:
-  minAvailable: {{ .Values.podDisruptionBudget.minAvailable }}
+  maxUnavailable: {{ .Values.podDisruptionBudget.maxUnavailable }}
   selector:
     matchLabels:
     {{- include "karpenter.labels" . | nindent 4 }}
--- a/charts/karpenter/templates/poddisruptionbudget.yaml
+++ b/charts/karpenter/templates/poddisruptionbudget.yaml
@@ -1,0 +1,10 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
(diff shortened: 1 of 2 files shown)


bwagner5:
`min-available` should probably be in values.yaml IMO

charts/karpenter/templates/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: karpenter
+spec:
+  minAvailable: 1
```
dewjam, line 6: Given we default to just 1 replica, I think this will cause problems when evicting Karpenter pods.  We will always violate the PDB and have to delete the pod manually.

ellistarn (approved):

charts/karpenter/values.yaml
```diff
@@ -37,6 +37,8 @@ podLabels: {}
 # -- Additional annotations for the pod.
 podAnnotations: {}
 # -- SecurityContext for the pod.
```
bwagner5, line 39: This should go below PDB

bwagner5 (changes requested):
just one comment

charts/karpenter/templates/poddisruptionbudget.yaml
dewjam: Perhaps we could do something like this?

```
{{ if eq .Values.replicas 1 }} 
minAvailable: 0
{{ else }}
minAvailable: {{ .Values.podDisruptionBudget.minAvailable }}
{{ end }}
```
(not sure if that's totally correct)

bwagner5: If we included topologySpread like in this PR https://github.com/aws/karpenter/pull/1427, we could increment the controller replicas to 2

charts/karpenter/values.yaml
```diff
@@ -37,6 +37,8 @@ podLabels: {}
...
+podDisruptionBudget:
+  minAvailable: 1
```
bwagner5, line 41:
```suggestion
podDisruptionBudget:
  minAvailable: 1
# -- SecurityContext for the pod.
```

charts/karpenter/templates/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: karpenter
+spec:
+  minAvailable: {{ .Values.podDisruptionBudget.minAvailable }}
```
bwagner5, line 6:
```suggestion
  maxUnavailable: {{ .Values.podDisruptionBudget.maxUnavailable }}
```

charts/karpenter/values.yaml
```diff
@@ -36,6 +36,8 @@ strategy:
 podLabels: {}
 # -- Additional annotations for the pod.
 podAnnotations: {}
+podDisruptionBudget:
+  minAvailable: 1
```
bwagner5, line 40:
```suggestion
  maxUnavailable: 1
```

bwagner5 (approved):
lgtm

charts/karpenter/templates/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: karpenter
+spec:
+  maxUnavailable: {{ .Values.podDisruptionBudget.maxUnavailable }}
+  selector:
+    matchLabels:
+    {{- include "karpenter.labels" . | nindent 4 }}
```
bwagner5, line 9:
```suggestion
    {{- include "karpenter.selectorLabels" . | nindent 6 }}
```

bwagner5 (approved):
lgtm

PR #1650, opened 2022-04-08
Add stale issue closer in dry run mode

spring1843:
**1. Issue, if available:**

This is a different issue closer than the one we used before. The other plugin we used before did not behave correctly with regards to label exemptions. This one should do a better job.

This is gonna run in dry mode so we are gonna see if it would behave properly and if it does we can permanently enable it.

- [x] No


bwagner5 (approved):
lgtm!

PR #1656, opened 2022-04-11
Add topologySpreadConstraints to helm chart

cest-pas-faux:
**1. Issue, if available:**

topologySpreadConstraints was not available in the chart, resulting in karpenter deployment pods scheduled on the same node.

Added topologySpreadConstraints logic to values.yaml and deployment template.
Added a sample config as well, maybe you'll prefer to split on the hostname instead of the zone.

Locally, by deploying the chart and scaling the number of replicas to validate the syntax and the spread result.

- [x] No

### DIFF
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -59,6 +59,10 @@ affinity:
               operator: DoesNotExist
 # -- topologySpreadConstraints to increase the controller resilience
 topologySpreadConstraints: []
+# - maxSkew: 1
+#   topologyKey: topology.kubernetes.io/zone
+#   whenUnsatisfiable: ScheduleAnyway
+
 # -- Tolerations to allow the pod to be scheduled to nodes with taints.
 tolerations: []
 controller:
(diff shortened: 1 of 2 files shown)


charts/karpenter/values.yaml
```diff
@@ -57,6 +57,12 @@ affinity:
         - matchExpressions:
             - key: karpenter.sh/provisioner-name
               operator: DoesNotExist
+# -- topologySpreadConstraints to increase the controller resilience
+topologySpreadConstraints: []
+# - maxSkew: 1
+#   topologyKey: topology.kubernetes.io/zone
+#   whenUnsatisfiable: ScheduleAnyway
```
bwagner5, line 64: What if we just defaulted to the topology spread? In the case of 1 replica, it won't matter, and will spread as the replicas are increased.
```suggestion
topologySpreadConstraints:
- maxSkew: 1
  topologyKey: topology.kubernetes.io/zone
  whenUnsatisfiable: ScheduleAnyway
```

bwagner5 (approved):
lgtm! Thanks!

PR #1657, opened 2022-04-11
manually install golicense from the last release

tzneal:
**1. Issue, if available:**

N/A

golicense doesn't build with go 1.18+, so instead of installing golicense from source, install the last release.

Locally, this PR should test is with VI

- [ ] No

### DIFF
--- a/hack/toolchain.sh
+++ b/hack/toolchain.sh
@@ -6,16 +6,15 @@ KUBEBUILDER_ASSETS="${KUBEBUILDER_ASSETS:="${HOME}/.kubebuilder/bin"}"
 
 main() {
     tools
+    golicense
     kubebuilder
 }
 
 tools() {
-    go install github.com/mitchellh/golicense@v0.2.0
     go install github.com/fzipp/gocyclo/cmd/gocyclo@v0.3.1
     go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.41.1
     go install github.com/google/ko@v0.11.2
     go install github.com/mikefarah/yq/v4@v4.16.1
-    go install github.com/mitchellh/golicense@v0.2.0
     go install github.com/norwoodj/helm-docs/cmd/helm-docs@v1.8.1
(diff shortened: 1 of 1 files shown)


hack/toolchain.sh
```diff
@@ -15,14 +15,23 @@ tools() {
...
-    go install github.com/mitchellh/golicense@v0.2.0
     go install github.com/norwoodj/helm-docs/cmd/helm-docs@v1.7.0
     go install github.com/onsi/ginkgo/ginkgo@v1.16.5
     go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20220113220429-45b13b951f77
     go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.7.0
     go install github.com/sigstore/cosign/cmd/cosign@v1.5.1
     go install github.com/gohugoio/hugo@v0.92.1+extended
 
+    # golicense no longer builds with go 1.18+, for now just install the last release binary
+    if [[ $(go env GOOS) == "darwin" ]]; then
```
ellistarn, line 26: Can we wrap this with a helper like kubebuilder?

ellistarn: Another option would be to keep an older version of go on the host and just use that for the install command.

hack/toolchain.sh
```diff
@@ -15,14 +15,23 @@ tools() {
     go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.41.1
     go install github.com/google/ko@v0.11.2
     go install github.com/mikefarah/yq/v4@v4.16.1
-    go install github.com/mitchellh/golicense@v0.2.0
```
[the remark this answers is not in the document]
tzneal: :face-palm-emoji:

hack/toolchain.sh
tzneal: Just created a helper for now, I viewed this as an interim solution to allow us to move to Go 1.18 whenever we want to.  Hopefully the main repo will accept a PR or a fork will emerge that has this feature fixed.

PR #1670, opened 2022-04-13
rework search indexing

tzneal:
**1. Issue, if available:**
N/A

- make all content non-indexable by default
- during release, ensure everything is non-indexeable and
  only enable indexing for the current release

Went to Netlify preview and searched, only v0.8.2 results displayed.

- [ ] No

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -43,6 +43,9 @@ chart() {
 website() {
     mkdir -p website/content/en/${RELEASE_VERSION} && cp -r website/content/en/preview/* website/content/en/${RELEASE_VERSION}/
     find website/content/en/${RELEASE_VERSION}/ -type f | xargs perl -i -p -e "s/{{< param \"latest_release_version\" >}}/${RELEASE_VERSION}/g;"
+    # make only the latest version searchable
+    find website/content/en/ -type f | xargs perl -i -p -e "s/exclude_search:.*/exclude_search: true/"
+    find website/content/en/${RELEASE_VERSION}/ -type f | xargs perl -i -p -e "s/exclude_search:.*/exclude_search: false/"
 }
 
 image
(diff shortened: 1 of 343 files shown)


spring1843:
@tzneal can you explain what drives this change? What are we trying to accomplish?

I am concerned with our search engine rankings, if we constantly change the index-ability of our pages there's a chance it lowers our ranking. Maybe we should keep the "preview" always indexible.

tzneal:
> @tzneal can you explain what drives this change? What are we trying to accomplish?
> 

The search on karpenter.sh isn't useful as it pulls results from every release we've ever had.  It will only continue to get  worse.
Screen Shot 2022-04-13 at 12 08 27 PM

> I am concerned with our search engine rankings, if we constantly change the index-ability of our pages there's a chance it lowers our ranking. Maybe we should keep the "preview" always indexible.

I didn't realize this also caused meta tags to be added to the HTML output:

```<meta name=ROBOTS content="NOINDEX, NOFOLLOW">```

spring1843:
Thank you for explaining that @tabern.

For this approach to work we need to be able to edit older files. Since we are planning to move the older files away from the repo at some point, doing so will become difficult if not impossible. Is it conceivable to have the same outcome without having to edit the old versions?

tzneal:
> Thank you for explaining that @tzneal
> 
> For this approach to work we need to be able to edit older files. Since we are planning to move the older files away from the repo at some point, doing so will become difficult if not impossible. Is it conceivable to have the same outcome without having to edit the old versions?

This just modifies the markdown input files that are then rendered into HTML.  Presumably we will never lose the ability to modify those files. That there is a directory called "v0.8.2" is that happens to contain our doc snapshots at a point in time is just an artifact of how the directory structure is laid out.  Hugo renders those files each time it generates the site.

bwagner5:
We may want to wait on PR https://github.com/aws/karpenter/pull/1674 since this will only work for the latest version. I'm not sure how far along #1674 is

tzneal:
> We may want to wait on PR #1674 since this will only work for the latest version. I'm not sure how far along #1674 is

I viewed that as a feature, I'm good either way.

PR #1671, opened 2022-04-13
support optional provisioner.Spec.taintsToIgnore

BryanStenson-okta:
**1. Issue, if available:**
#628 

This change allows the provisioner the ability to ignore the listed taints when creating a new node

Added suite test to ensure pods with missing tolerations are still scheduled when the missing taint is in the taintsToIgnore list

- [ ] No

### DIFF
--- a/charts/karpenter/crds/karpenter.sh_provisioners.yaml
+++ b/charts/karpenter/crds/karpenter.sh_provisioners.yaml
@@ -140,7 +140,8 @@ spec:
               taintsToIgnore:
                 description: TaintsToIgnore is a list of taints for the provisioner
                   to ignore when provisioning new nodes and scheduling pods. If specified,
-                  taints included in this list will be ignored when ...
+                  taints included in this list will be ignored when verifying a pod's
+                  constraints.
                 items:
                   description: The node this Taint is attached to has the "effect"
                     on any pod that does not tolerate the Taint.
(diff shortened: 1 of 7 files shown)


dewjam:
@BryanStenson-okta ,
Have you tested this change in a live cluster? 

Correct me if I'm wrong, but in the case of Taints with NoExecute effect, I don't think this will have the desired outcome.  Karpenter will bind a pod to a node, but I believe kube-scheduler will just evict it right away, causing another node to be launched by Karpenter.  The pod would eventually schedule once the NoExecute taint is removed (assuming a DS removes it), but there would be some node churn during the process.

When I get a moment, I'll pull down your code and test in my environment.

BryanStenson-okta:
Hi @dewjam.  I haven't tested in a cluster, but I see how this will cause node churn.  Having said that, it _does_ solve the problem of Karpenter not adding nodes. ;)

I'm not actually suggesting node churn is okay here.  It seems @ellistarn suggests there's discussion about preventing pod binding (which might eliminate the node churn) -- https://github.com/aws/karpenter/issues/628#issuecomment-1098544518

FWIW - I intent to join the working group tomorrow for a discussion on steps forward here.  I'd really like to get this solved with Karpenter and NOT fall back to using CAS.

bwagner5:
Should be taken care of with StartupTaints now https://github.com/aws/karpenter/pull/1727 

Will be released next week.

PR #1673, opened 2022-04-13
Resolved cross-provisioner scheduling bugs by collapsing selection into a unified scheduler

ellistarn:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/975

Now that we have a more advanced scheduler, we move selection logic from the top level to within the scheduler which has a number of nice effects.

- [performance] The scheduler considers the set of possible nodes that could be created by any provisioner, rather than just one at a time
- [bugfix] The scheduler considers topology across all of these provisioners
- [not implemented] we could fallback to provisioners that haven't exceeded limits
- [performance] Preferential fallback is now effective immediately, rather than on a retry. 

- Manually
- `make test`

- [ ] No


pkg/controllers/provisioning/controller.go
```diff
@@ -31,124 +33,125 @@ import (
...
 
-// Register the controller to the manager
 func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 	return controllerruntime.
 		NewControllerManagedBy(m).
 		Named(controllerName).
-		For(&v1alpha5.Provisioner{}).
-		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
+		For(&v1.Pod{}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10_000}).
```
tzneal, line 155: Why is Add() still a blocking call?  We don't use the result, and if it wasn't this could go down to a more reasonable number.

pkg/controllers/provisioning/scheduling/preferences.go
```diff
@@ -34,47 +33,9 @@ const (
 )
```
tzneal, line 33: ExpirationTTL and CleanupInterval aren't used any longer

pkg/controllers/provisioning/provisioner.go
```diff
@@ -120,45 +109,100 @@ func (p *Provisioner) provision(ctx context.Context) error {
...
-			return false, nil
+func (p *Provisioner) getPods(ctx context.Context) ([]*v1.Pod, error) {
+	var podList v1.PodList
+	if err := p.kubeClient.List(ctx, &podList, client.MatchingFields{"spec.nodeName": ""}); err != nil {
+		return nil, fmt.Errorf("listing pods, %w", err)
+	}
+	var pods []*v1.Pod
+	for i := range podList.Items {
+		if isProvisionable(&podList.Items[i]) {
+			pods = append(pods, &podList.Items[i])
```
tzneal, line 120:
```suggestion
			pods = append(pods, podList.Items[i].DeepCopy())
```

We relax the pods later, if we deep copy here we ensure that we don't mess the list cache up.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -81,34 +81,23 @@ func (p *Provisioner) Add(pod *v1.Pod) <-chan struct{} {
 func (p *Provisioner) provision(ctx context.Context) error {
 	// Batch pods
 	logging.FromContext(ctx).Infof("Waiting for unschedulable pods")
-	items, window := p.batcher.Wait()
```
tzneal, line 84: Can easily wait for another PR, but would be nice if batcher didn't collect items and was just a trigger mechanism.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -129,65 +122,65 @@ var _ = Describe("Custom Constraints", func() {
...
+			pod := ExpectProvisioned(ctx, env.Client, controller, provisioner, test.UnschedulablePod(
 				test.PodOptions{NodePreferences: []v1.NodeSelectorRequirement{
 					{Key: "test-key", Operator: v1.NodeSelectorOpIn, Values: []string{"another-value", "test-value"}},
 				}},
 			))[0]
 			node := ExpectScheduled(ctx, env.Client, pod)
 			Expect(node.Labels).To(HaveKeyWithValue("test-key", "test-value"))
 		})
-		It("should not schedule pods with have conflicting preferences", func() {
+		It("should schedule pods with have conflicting preferences", func() {
```
tzneal, line 133:
```suggestion
		It("should schedule pods that have conflicting preferences", func() {
```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -129,65 +122,65 @@ var _ = Describe("Custom Constraints", func() {
...
 			provisioner.Spec.Labels = map[string]string{"test-key": "test-value"}
-			pod := ExpectProvisioned(ctx, env.Client, selectionController, provisioners, provisioner, test.UnschedulablePod(
+			pod := ExpectProvisioned(ctx, env.Client, controller, provisioner, test.UnschedulablePod(
 				test.PodOptions{NodePreferences: []v1.NodeSelectorRequirement{
 					{Key: "test-key", Operator: v1.NodeSelectorOpNotIn, Values: []string{"test-value"}},
 				}},
 			))[0]
-			ExpectNotScheduled(ctx, env.Client, pod)
+			ExpectScheduled(ctx, env.Client, pod)
```
tzneal, line 140: Why does this schedule now?   There are a few tests like this where it looks like just the condition was changed.

pkg/controllers/provisioning/controller.go
ellistarn: Let's catch this on a followup that cleans up the batcher.

pkg/controllers/provisioning/provisioner.go
ellistarn: IIUC it's a copy -- may want to rethink if we care about memory at some point.

pkg/controllers/provisioning/scheduling/suite_test.go
ellistarn: Previously preferences needed to execute twice and the test only executed once.

PR #1678, opened 2022-04-15
dev improvements: helm create-namespace and truncate eksctl domain on CLUSTER_NAME

bwagner5:
**1. Issue, if available:**
N/A

 - use `--create-namespace` with helm upgrade 
 - only use the base domain if a cluster name includes an FQDN like ekctl does

- [x] No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ GOFLAGS ?= -tags=$(CLOUD_PROVIDER) $(LDFLAGS)
 WITH_GOFLAGS = GOFLAGS="$(GOFLAGS)"
 
 ## Extra helm options
-CLUSTER_NAME ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].name}' | rev | cut -d"/" -f1 | rev)
+CLUSTER_NAME ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].name}' | rev | cut -d"/" -f1 | rev | cut -d"." -f1)
 CLUSTER_ENDPOINT ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].cluster.server}')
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -7,7 +7,7 @@ GOFLAGS ?= -tags=$(CLOUD_PROVIDER) $(LDFLAGS)
 WITH_GOFLAGS = GOFLAGS="$(GOFLAGS)"
 
 ## Extra helm options
-CLUSTER_NAME ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].name}' | rev | cut -d"/" -f1 | rev)
+CLUSTER_NAME ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].name}' | rev | cut -d"/" -f1 | rev | cut -d"." -f1)
```
njtran, line 10: Nice! this fixes if you didn't do `export CLUSTER_NAME=<name>` so that it works for weird eksctl behavior.

Makefile
```diff
@@ -57,7 +57,7 @@ licenses: ## Verifies dependency licenses and requires GITHUB_TOKEN to be set
 	golicense hack/license-config.hcl karpenter
 
 apply: ## Deploy the controller from the current state of your git repository into your ~/.kube/config cluster
-	helm upgrade --install karpenter charts/karpenter --namespace karpenter \
+	helm upgrade --create-namespace --install karpenter charts/karpenter --namespace karpenter \
```
njtran, line 60: I don't think this should matter, but in our getting-started guide we do it like this `--namespace karpenter --create-namespace`. Can you make sure that the order doesn't matter?

njtran (approved):
Just a question, but looks good otherwise

Makefile
bwagner5: yep, order doesn't matter:

```
Usage:
  helm upgrade [RELEASE] [CHART] [flags]
```

PR #1689, opened 2022-04-18
Add nightlies, and stable releases to the snapshots repo

spring1843:
**1. Issue, if available:**

Adds nightlies, and stable releases to the same mechanism we use for snapshot releases.
Ded-dupes the code for snapshots/nightlies/stable releases.
Does not change anything about the existing stable release process.

- [x] No


hack/release.sh
```diff
@@ -45,6 +35,9 @@ website() {
     find website/content/en/${RELEASE_VERSION}/ -type f | xargs perl -i -p -e "s/{{< param \"latest_release_version\" >}}/${RELEASE_VERSION}/g;"
 }
 
-image
+requireCloudProvider
+authenticate
+buildImages $RELEASE_VERSION
+cosignImages
```
ellistarn, line 41: Do we not sign nightlies?

Makefile
```diff
@@ -81,12 +81,15 @@ codegen: ## Generate code. Must be run if changes are made to ./pkg/apis/...
...
-	$(WITH_GOFLAGS) ./hack/nightly.sh
+nightly: ## Tag the latest snapshot release with timestamp
+	./hack/add-snapshot-tag.sh $(shell git rev-parse HEAD) $(shell date +"%Y%m%d")
 
-snapshot: ## Generate nightly release manifests and publish a versioned container image.
+snapshot: ## Generate a snapshot release out of the current commit
 	$(WITH_GOFLAGS) ./hack/snapshot.sh
 
+stablerelease: ## Tags the snapshot release of the current commit with the latest tag available, for prod launch
+	./hack/add-snapshot-tag.sh $(shell git rev-parse HEAD) $(shell git tag | sort -V | tail -1)
```
bwagner5, line 91: can we use `git describe --tags --always`. That command will make sure HEAD is on the current release tag and if not append the commit hash.  It might be worth putting a VERSION env var at the top of the Makefile to replace this command and the linker flags that sets the version on line 5.

.github/workflows/release.yaml
```diff
@@ -18,3 +18,9 @@ jobs:
           branch: gh-pages
           folder: charts
           target-folder: charts
+      - run: |
+          make toolchain
```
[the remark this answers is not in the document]
spring1843: done

hack/release.sh
spring1843: It seems like running cosign will require a human interaction the way it works with stable releases, as of now the snapshots and nightlies do not have this process.

bwagner5: it might "just work", maybe worth testing out on a fork: https://github.com/sigstore/cosign/blob/main/KEYLESS.md#identity-tokens

Makefile
spring1843: good idea, switched to `git describe --tags --exact-match` as discussed offline.

hack/release.sh
spring1843: I'm gonna note down to experiment with that, It seems to me like the signing can happen without human interaction but then you need to store a key somewhere and that's slightly more complex, but I can do that as a follow up PR if you don't mind.

hack/add-snapshot-tag.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -e
```
bwagner5, line 1:
```suggestion
#!/bin/bash -euo pipefail
```

hack/release.sh
```diff
@@ -1,13 +1,10 @@
 #!/bin/bash -e
```
bwagner5, line 1:
```suggestion
#!/bin/bash -euo pipefail
```

hack/release_common.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -e
```
bwagner5, line 1:
```suggestion
#!/bin/bash -euo pipefail
```

hack/snapshot.sh
```diff
@@ -1,30 +1,10 @@
 #!/bin/bash -e
```
bwagner5, line 1:
```suggestion
#!/bin/bash -euo pipefail
```

bwagner5 (approved):
lgtm!

PR #1709, opened 2022-04-20
Correct the example provisioner for AWS custom launch template doc

felix-zhe-huang:
**1. Issue, if available:**
The example provisioner is missing the subnet selector. It will fail the validation webhook check.

Include the subnet selector to the example provisioner.

N/A

- [ ] No

### DIFF
--- a/website/content/en/preview/AWS/launch-templates.md
+++ b/website/content/en/preview/AWS/launch-templates.md
@@ -226,6 +226,7 @@ aws cloudformation create-stack \
 
 The LaunchTemplate is ready to be used. Specify it by name in the [Provisioner
 CRD](../../provisioner/). Karpenter will use this template when creating new instances.
+The following is an example of a provisioner using the new template. Please replace the `CLUSTER_NAME` with the correct value.
 
 ```yaml
 apiVersion: karpenter.sh/v1alpha5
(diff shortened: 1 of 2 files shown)


website/content/en/preview/AWS/launch-templates.md
```diff
@@ -226,12 +226,14 @@ aws cloudformation create-stack \
...
+The following is an example of a provisioner using the new template. Please replace the `CLUSTER_NAME` with the correct value.
 
 ```yaml
 kind: Provisioner
 spec:
   provider:
     launchTemplate: KarpenterCustomLaunchTemplate
-
+    subnetSelector:
```
ellistarn, line 237: I think the aim of this was not to provide a full spec, but a snippit. If you want a full spec, you need `securityGroupSelector` too.

felix-zhe-huang: The subnet selector is required for the validation webhook. The original version of the provisioner will fail to be created without it. The securityGroup is defined in the launch template and currently we can either specify the selector or providing the launch template. We can't have both of them together.

The reason that we still need the subnet selector is that fleet will override the subnet defined in the launch template (if there is one at the beginning.) I think this implicit requirements are not properly documented. Or maybe we want to revisit the decision.

ellistarn: Oh TIL. I didn't remember this feature. Thank for explaining.

ellistarn (approved):

PR #1719, opened 2022-04-25
rework batching

tzneal:
**1. Issue, if available:**
N/A

Remove item accumulation which was only used for unit testing.

Unit tests & deployed to EKS with scale up/scale down.

- [X] No

### DIFF
--- a/pkg/controllers/provisioning/batcher.go
+++ b/pkg/controllers/provisioning/batcher.go
@@ -16,84 +16,81 @@ package provisioning
 
 import (
 	"context"
-	"sync"
 	"time"
 )
 
 var (
 	MaxBatchDuration  = time.Second * 10
 	BatchIdleDuration = time.Second * 1
-	// MaxItemsPerBatch limits the number of items we process at one time to avoid using too much memory
-	MaxItemsPerBatch = 2_000
 )
 
-// Batcher separates a stream of Add(item) calls into windowed slices. The
+// Batcher separates a stream of Trigger() calls into windowed slices. The
 // window is dynamic and will be extended if additional items are added up to a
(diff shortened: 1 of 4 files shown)


tzneal:
With Ellis's change to move provisioner selection into scheduling, the batcher was only being used as a triggering mechanism.  The "max items" was solely used for unit test purposes, but caused us to retain those items unnecessarily.

This simplifies the batcher to just be a trigger mechanism.   I just added an exposed public method to control the batch loop which is used during the unit tests.  I also looked at using something like github.com/benbjohnson/clock to mock the timers for testing.  I think it's a bit cleaner to do it that way, but more invasive.  Interested in opinions.

Side benefit of this change is that we eliminate a pool of 10k goroutines and overall use less memory (160Mb -> 100Mb in my 1k pod scale-up testing).

pkg/controllers/provisioning/batcher.go
```diff
@@ -16,84 +16,77 @@ package provisioning
...
-	case b.queue <- item:
-	case <-b.running.Done():
-	}
-	b.RLock()
-	defer b.RUnlock()
-	return b.gate.Done()
+// Trigger causes the batcher to start a batching window, or extend the current batching window if it hasn't reached the
+// maximum length.
+func (b *Batcher) Trigger() {
+	b.trigger <- struct{}{}
```
bwagner5, line 48: To remove the need of setting an arbitrary buffered channel value, you could place this in a select with a default case since we won't need to reset the timer if there are triggers already queued.

bwagner5 (commented):
nice! Just 1 suggestion and a question, otherwise lgtm

bwagner5 (approved):
lgtm!

Issue #1327, opened 2022-02-11
v1beta1 Laundry List


ellistarn:
**Tell us about your request**
This issue contains a list of breaking API changes for v1beta1.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ellistarn:
Fix casing on capacity type to conform w/ k8s value casing:
`on-demand` -> `OnDemand`
`spot` -> `Spot`

ellistarn:
I'd love to explore separating the `spec.provider` into a separate vendor specific CRD and using something like `spec.providerRef`.

ellistarn:
I'd like to rename `spec.kubeletConfiguration` to `spec.kubelet`

ellistarn:
I'd like to explore creating a decoupled `Deprovisioner` resource with `spec.selector` that encapsulates `spec.ttlSecondsAfterEmpty`, `spec.ttlSecondsAfterExpired`, and #1091.

felix-zhe-huang:
I'd like to explore adding `spec.preferenceWeight` to indicate the preference of which provisioner should be selected when multiple are compatible to the pod spec.

ellistarn:
> I'd like to explore adding `spec.preferenceWeight` to indicate the preference of which provisioner should be selected when multiple are compatible to the pod spec.

Check out the related discussion here: https://github.com/aws/karpenter/issues/783

felix-zhe-huang:
Another thought. Remove provisioner labels. Everything will be described as requirements. When Karpenter launches new nodes, a label will be added to the node with a valid value from the combined requirements (provisioner + pods).

robertd:
Thanks for the transparency ... I'll definitely be monitoring this issue for changes that need to be applied to the cdk-karpenter: https://www.npmjs.com/package/cdk-karpenter 👍

ellistarn:
Remove kubeletconfiguration.clusterdnsip in favor of discovery https://github.com/aws/karpenter/issues/1241#issuecomment-1055697350

stevehipwell:
[...]

PR #1720, opened 2022-04-25
Add support for customUserData with Bottlerocket

suket22:
**1. Issue, if available:**
N/A

This change only adds support for customUserData with Bottlerocket AMIFamily. It follows [what's defined in this doc](https://github.com/aws/karpenter/blob/main/designs/aws/custom-user-data-and-amis.md) for the most part. The one main change is that I'm no longer merging labels / taints from the custom userData but rather overwriting them completely. This is because of what we discovered [here regarding kubelet not merging in other labels](https://github.com/aws/karpenter/issues/1681). As long as Karpenter creates the node object itself, there's no point merging in labels / taints etc. 

* I've tested this locally - ran through both the provider level validation checks. 
* The bootstrap logic isn't unit tested but I've been testing it out locally and my nodes are joining the cluster. I haven't tried the full extent of all BR settings but most that I've tried have worked. 

- [X] Yes, issue opened: *link to issue*
- [ ] No

I'm thinking of only adding docs once AL2 support is here. 


pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocket.go
```diff
@@ -15,59 +15,52 @@ limitations under the License.
...
 	s.Settings.Kubernetes.NodeTaints = map[string][]string{}
 	for _, taint := range b.Taints {
 		s.Settings.Kubernetes.NodeTaints[taint.Key] = append(s.Settings.Kubernetes.NodeTaints[taint.Key], fmt.Sprintf("%s:%s", taint.Value, taint.Effect))
 	}
-	script, err := toml.Marshal(s)
+	script := new(bytes.Buffer)
+	tomlEncoder := toml.NewEncoder(script)
+	err := tomlEncoder.Encode(&s)
+	if err != nil {
+		panic(err)
```
tzneal, line 53: Modify Bootstrapper.Script() to return a (string, error) so we can report the error message and error on launch instead of panic()?

pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocket.go
```diff
@@ -15,59 +15,52 @@ limitations under the License.
...
 	if !b.AWSENILimitedPodDensity {
-		s.Settings.Kubernetes.MaxPods = 110
+		s.Settings.Kubernetes.MaxPods = aws.Int(110)
 	}
 	s.Settings.Kubernetes.NodeTaints = map[string][]string{}
 	for _, taint := range b.Taints {
 		s.Settings.Kubernetes.NodeTaints[taint.Key] = append(s.Settings.Kubernetes.NodeTaints[taint.Key], fmt.Sprintf("%s:%s", taint.Value, taint.Effect))
 	}
-	script, err := toml.Marshal(s)
+	script := new(bytes.Buffer)
```
tzneal, line 49:
```suggestion
	script := &bytes.Buffer{}
```

I don't feel strongly either way, but the code base has few instances of `new(Type)`

pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocket.go
```diff
@@ -15,59 +15,52 @@ limitations under the License.
...
+	err := tomlEncoder.Encode(&s)
+	if err != nil {
+		panic(err)
+	}
+	return base64.StdEncoding.EncodeToString(script.Bytes())
+}
+
+func (b Bottlerocket) unmarshalCustomUserData() config {
+	var c config
+	b64DecodedBytes, _ := base64.StdEncoding.DecodeString(*b.CustomUserData)
```
tzneal, line 60: I saw that it's validated prior, but if `Script()` is changed to possibly return an error, I don't think it hurts to check for an error here instead of ignoring it.

Checking just below here will allow removing an additional panic call.

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
```diff
@@ -74,6 +74,11 @@ type LaunchTemplate struct {
 	// BlockDeviceMappings to be applied to provisioned nodes.
 	// +optionals
 	BlockDeviceMappings []*BlockDeviceMapping `json:"blockDeviceMappings,omitempty"`
+	// base64 encoded string that will be applied to provisioned nodes as UserData.
+	// UserData must be in the appropriate format based on the specified AMIFamily. Karpenter will merge certain fields into
+	// this UserData to ensure nodes are being provisioned with the correct configuration.
+	// +optional
+	UserData *string `json:"userData,omitempty"`
```
tzneal, line 81: Nit: Comment should start with the name of the element it describes for automatic doc generation to work best:

```golang
// UserData is a base64 encoded string that will be ....
```
See https://go.dev/blog/godoc

pkg/cloudprovider/aws/apis/v1alpha1/provider_validation.go
```diff
@@ -228,3 +234,18 @@ func (a *AWS) validateVolumeSize(blockDeviceMapping *BlockDeviceMapping) *apis.F
...
+	if a.UserData == nil {
+		return nil
+	}
+	// Only support UserData with Bottlerocket for now.
+	if *a.AMIFamily != AMIFamilyBottlerocket {
+		return apis.ErrInvalidValue(fmt.Sprintf("not supported with %v", *a.AMIFamily), userDataPath)
+	}
+	_, err := base64.StdEncoding.DecodeString(*a.UserData)
+	if err != nil {
+		return apis.ErrInvalidValue("not valid Base64", userDataPath)
```
bwagner5, line 248: nit:

```suggestion
		return apis.ErrInvalidValue("invalid Base64", userDataPath)
```

pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocket.go
bwagner5: We panic'd before since we knew it would always encode properly, but I think expanding the `Script` to return an error now that we're processing user input is a good idea.

pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocket.go
bwagner5, line 60: need a nil check here and short circuit to an empty config

pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocket.go
```diff
@@ -15,59 +15,52 @@ limitations under the License.
...
 	}
 	s.Settings.Kubernetes.NodeTaints = map[string][]string{}
 	for _, taint := range b.Taints {
 		s.Settings.Kubernetes.NodeTaints[taint.Key] = append(s.Settings.Kubernetes.NodeTaints[taint.Key], fmt.Sprintf("%s:%s", taint.Value, taint.Effect))
 	}
-	script, err := toml.Marshal(s)
+	script := new(bytes.Buffer)
+	tomlEncoder := toml.NewEncoder(script)
+	err := tomlEncoder.Encode(&s)
+	if err != nil {
```
bwagner5, line 52: consider combining to one line when a func only returns an err val:

```
if err := tomlEncoder.Encode(&s); err != nil {
...
``

File: pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocket.go (hunk 2 above)
Comment (suket22) [reply pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocket.go]: I was playing around with the encoder but figured the defaults work fine for us so just moved back to using `toml.Marshal` and some of this got cleaned up.

File: pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocket.go (hunk 3 above)
Comment (suket22) [reply pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocket.go]: Ooh, nice catch. Tested manually and verified that too.

Comment (suket22) [reply pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocket.go]: Right yeah, I had left it alone because we do this in our provider's validation code. Changed now.

**Review by tzneal — DISMISSED** [review]
I haven't tested this, but the changes LGTM

File: pkg/cloudprovider/aws/amifamily/bootstrap/bottlerocketsettings.go [hunk 7]
```diff
@@ -0,0 +1,98 @@
...
+	ClusterCertificate        *string              `toml:"cluster-certificate"`
+	ClusterName               *string              `toml:"cluster-name"`
+	ClusterDNSIP              *string              `toml:"cluster-dns-ip,omitempty"`
+	NodeLabels                map[string]string    `toml:"node-labels,omitempty"`
+	NodeTaints                map[string][]string  `toml:"node-taints,omitempty"`
+	MaxPods                   *int                 `toml:"max-pods,omitempty"`
+	StaticPods                map[string]staticPod `toml:"static-pods,omitempty"`
+	EvictionHard              map[string]string    `toml:"eviction-hard,omitempty"`
+	KubeReserved              map[string]string    `toml:"kube-reserved,omitempty"`
+	SystemReserved            map[string]string    `toml:"system-reserved,omitempty"`
```
bwagner5, line 45: should a lot of these be added to the Provisioner spec.kubeletConfig and then merged into the bottlerocket config?

bwagner5 (changes requested):
can you add docs with an example of a simple bottlerocket config in the provisioner? 

And can you add a test too?

suket22:
> can you add docs with an example of a simple bottlerocket config in the provisioner?
> 
> And can you add a test too?

I intentionally left the docs out. I was thinking of adding them once I've got support for both AMI families rather than just one, but can add them now. 

I'll add a few tests :+1:

website/content/en/preview/AWS/provisioning.md
```diff
@@ -1,216 +1,259 @@
...
+          kmsKeyID: "arn:aws:kms:us-west-2:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab"
+          deleteOnTermination: true
+          throughput: 125
+          snapshotID: snap-0123456789
+```
+
+### UserData
+
+The `userData` field in a Provisioner can be used to control the UserData that Karpenter will use when provisioning nodes.
+Your UserData must be a Base64 encoded string, and it will be processed different based on the AMIFamily selected.
```
bwagner5, line 195:
```suggestion
Your UserData must be a Base64 encoded string, and it will be processed differently based on the AMIFamily selected.
```

website/content/en/preview/AWS/provisioning.md
```diff
@@ -1,216 +1,259 @@
...
+
+*This field is currently only supported for the BottleRocket AMIFamily*.
```
bwagner5, line 197:
```suggestion
*This field is currently only supported for the Bottlerocket AMIFamily*.
```

website/content/en/preview/AWS/provisioning.md
```diff
@@ -1,216 +1,259 @@
...
+
+**For BottleRocket**
```
bwagner5, line 199:
```suggestion
**For Bottlerocket**
```

website/content/en/preview/AWS/provisioning.md
```diff
@@ -1,216 +1,259 @@
----
```
bwagner5, line 1: I'm not sure why the diff shows that all of this file was deleted and re-added, but looks like you just added the UserData section?

bwagner5 (commented):
Just a few doc nits, but otherwise lgtm!

website/content/en/preview/AWS/provisioning.md
suket22: I think it's line endings. I think this was originally using CRLF, and I might've switched it to LF. Let me fix that too.

suket22: I think it's line endings. I think this was originally using CRLF, and I might've switched it to LF. Let me fix that too.

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -164,7 +168,7 @@ func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, optio
 				Name: aws.String(options.InstanceProfile),
 			},
 			SecurityGroupIds: aws.StringSlice(options.SecurityGroupsIDs),
-			UserData:         aws.String(options.UserData.Script()),
+			UserData:         aws.String(userData),
```
tzneal, line 171: Launch templates are currently cached just by the LT name, right? It would be difficult to hack on your userdata changes as you need to delete the LT every time or your change won't appear unless the LT has expired.

bwagner5: LT name is a hash of the contents so one should get recreated.

website/content/en/preview/AWS/provisioning.md
```diff
@@ -189,6 +189,49 @@ spec:
...
 
+### UserData
+
+The `userData` field in a Provisioner can be used to control the UserData that Karpenter will use when provisioning nodes.
+Your UserData must be a Base64 encoded string, and it will be processed different based on the AMIFamily selected.
+
+*This field is currently only supported for the BottleRocket AMIFamily*.
+
+**For BottleRocket**
+* Your UserData must be valid TOML.
```
bwagner5, line 200: oops, I think these got reverted when you changed the line breaks

bwagner5:
Have you thought more about the configmap approach for specifying userdata?

website/content/en/preview/AWS/provisioning.md
suket22: Nooooooo

suket22:
> Have you thought more about the configmap approach for specifying userdata?

I did think about it. Most of it seemed good -
* ConfigMaps support both binary and UTF-8 data. For example, kubectl's `--from-file` will automatically detect if the file is UTF-8 and if so puts it under `.data` otherwise under base64 encodes it and puts it under `.binaryData` so that'll work great for us.   
* Makes the ProvisionerSpec read shorter (UserData can go up to 16KB of Base64 content which is quite large). 
* Looks like Karpenter [already had the permissions](https://github.com/aws/karpenter/blob/main/charts/karpenter/templates/clusterrole.yaml#L29) to read ConfigMaps across namespaces so should be good there too.


My worry though is that -
* We're introducing a new call to describe ConfigMaps in the call path to launching instances where we're trying to reduce as much latency as possible - even though I don't think this is double digit ms from what I can tell.
* We can potentially cache the responses of the ConfigMap but that made me wonder if it would lead to confusing UX. The larger issue is that you can modify the nodes being provisioned across provisioners by modifying a single ConfigMap. Maybe that's desirable?

bwagner5:
> > Have you thought more about the configmap approach for specifying userdata?
> 
> I did think about it. Most of it seemed good -
> 
> * ConfigMaps support both binary and UTF-8 data. For example, kubectl's `--from-file` will automatically detect if the file is UTF-8 and if so puts it under `.data` otherwise under base64 encodes it and puts it under `.binaryData` so that'll work great for us.
> * Makes the ProvisionerSpec read shorter (UserData can go up to 16KB of Base64 content which is quite large).
> * Looks like Karpenter [already had the permissions](https://github.com/aws/karpenter/blob/main/charts/karpenter/templates/clusterrole.yaml#L29) to read ConfigMaps across namespaces so should be good there too.
> 
> My worry though is that -
> 
> * We're introducing a new call to describe ConfigMaps in the call path to launching instances where we're trying to reduce as much latency as possible - even though I don't think this is double digit ms from what I can tell.
> * We can potentially cache the responses of the ConfigMap but that made me wonder if it would lead to confusing UX. The larger issue is that you can modify the nodes being provisioned across provisioners by modifying a single ConfigMap. Maybe that's desirable?

As far as the latency in the call-path, we can create a reconciler that watches config maps, so basically a cache but a stronger form since it should only be out of date while fetching the new one on a change.

I think the problem of a single config map causing multiple provisioners to change could be a pro if there's some sort of base bootstrap that the admin is managing for a bunch of provisioners. At the end of the day, it will be up to the admin on how they segment the config maps. They could just as easily duplicate the config map to do 1:1 mappings, but the config map semantic will provide more flexibility in that regard.

pkg/cloudprovider/aws/amifamily/userData.go
```diff
@@ -0,0 +1,56 @@
...
+
+// New constructs a new launch template Resolver
+func NewUserDataProvider(clientSet *kubernetes.Clientset) *UserDataProvider {
+	return &UserDataProvider{
+		clientSet: clientSet,
+	}
+}
+
+// Get returns a set of AMIIDs and corresponding instance types. AMI may vary due to architecture, accelerator, etc
+func (u *UserDataProvider) Get(ctx context.Context, provider *v1alpha1.AWS) (string, error) {
```
bwagner5, line 39: this probably works for now, I think we'll eventually want a controller to watch the config maps to know when they go out-of-sync but that can be part of the out-of-spec node updates scope. 

Actually, maybe we only allow immutable config maps to make this easier to reason about and easier for Karpenter to handle: https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-immutable

pkg/cloudprovider/aws/testdata/br_userdata_input.golden
```diff
@@ -0,0 +1,4 @@
+[settings.kubernetes.eviction-hard]
```
bwagner5, line 1: thoughts on in-lining these in the suite_tests so that you don't have to flip back-and-forth?

bwagner5 (commented):
nice!

pkg/cloudprovider/aws/amifamily/userData.go
suket22: > I think we'll eventually want a controller to watch the config maps to know when they go out-of-sync

Why do we need a controller to detect that? The version we're caching has the entire UserData content, not just the configmap name + namespace so we should use the latest content during every launch.

I think another controller is probably worth it just to introduce some caching and preventing that getConfigMap API call for latency reasons, but I figured that optimization can come later. 

I also just realized that the comment on `Get` is outdated so I'll fix that too :P

pkg/cloudprovider/aws/testdata/br_userdata_input.golden
suket22: I don't mind. I like having the input as dedicated files too because it's easier to edit and maintain rather than large multiline strings but I can move it into the tests

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
```diff
@@ -74,6 +74,25 @@ type LaunchTemplate struct {
 	// BlockDeviceMappings to be applied to provisioned nodes.
 	// +optionals
 	BlockDeviceMappings []*BlockDeviceMapping `json:"blockDeviceMappings,omitempty"`
+	// UserData to be applied to the provisioned nodes.
+	// It must be in the appropriate format based on the specified AMIFamily. Karpenter will merge certain fields into
+	// this UserData to ensure nodes are being provisioned with the correct configuration.
+	// +optional
+	UserData *UserData `json:"userData,omitempty"`
```
ellistarn, line 81: Did we discuss the tradeoffs of doing this inline vs as a configmap (e.g. the design doc?). I worry a bit about resource sprawl. One huge advantage I can see is that separating this into its own resource allows users to configure the same user data for multiple provisioners. 

In #1327, we propose separating the entire Provider object into its own CRD. At that point, the AWS Provider is more or less a kube-native launchTemplate-like abstraction, and shares the benefit mentioned above about being used by multiple provisioners.

suket22: @bwagner5 and I have been discussing this offline for a while (and in this PR too), however it's not captured in the design doc. 

This PR started off by inlining the UserData but it made the Provisioner much too large and quite unreadable. UserData can be 16KB of data, and as I started to play with it more it makes a lot of sense to keep it outside. For example, when using this with AL2, it was really odd looking to have bash scripts inlined inside the provisionerSpec so we decided to separate that out. 


> One huge advantage I can see is that separating this into its own resource allows users to configure the same user data for multiple provisioners.

Completely agree, and that's another reason why we think this approach is better than the inlining. 

> In #1327, we propose separating the entire Provider object into its own CRD. At that point, an AWSProvider CRD is more or less a kube-native launch-template abstraction, and shares the benefit mentioned above about being used by multiple provisioners.

I think I made this exact comment yesterday when demo-ing some of this stuff, and I 100% agree :). The ConfigMap solution is better than inlining, but doesn't go as far as it should. A dedicated CRD that's essentially an EC2 launch template should be next.

pkg/cloudprovider/aws/amifamily/userData.go
suket22: Added a dedicated informer for ConfigMaps. Thanks for the pointers!

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -128,6 +134,7 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = AfterSuite(func() {
+	stop()
```
suket22, line 137: I needed to do this otherwise the watch I set up in UserDataProvider kept a connection open with the API Server and the API Server wouldn't shut down, leading to the tests to fail. Kept running into this - https://github.com/kubernetes-sigs/controller-runtime/issues/1571

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
ellistarn: I might start by inlining it in the provisioner and then migrate to a separate CRD in the future. This is a more natural concept progression, since if it's eventually it's own CRD, it will be alongside other provider fields. This also seems like an easier migration path for customers, since they don't have to worry about cleaning up configmaps from a previous version.

tzneal: The demo @suket22 gave used the [configmap from file](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#create-configmaps-from-files) feature which made for a nice way to create the userdata.  This only works with configmaps as far as I can tell.

ellistarn: Can you paste an example of an unreadable provisioner spec with it inlined? Id image we'd use yaml multi line strings and match indentation. I'd also expect that users provide this at the bottom.

suket22: Something like this for AL2 - 
```
kind: Provisioner
metadata:
  name: default
spec:
  limits:
    resources:
      cpu: 20
  provider:
    amiFamily: AL2
    userData:  |
      MIME-Version: 1.0
      Content-Type: multipart/mixed; boundary="//"

      --//
      Content-Type: text/x-shellscript; charset="us-ascii"

      #!/bin/bash
      echo "Running custom user data script"

      --//--      
    instanceProfile: KarpenterNodeInstanceProfile-karp-cluster
    subnetSelector:
      karpenter.sh/discovery: karp-cluster
    securityGroupSelector:
      karpenter.sh/discovery: karp-cluster
  ttlSecondsAfterEmpty: 30
```

I'd expect most user data scripts to run *far longer* than that however, making it odd to manage IMO.

ellistarn: Does it look cleaner if userdata is less nested (i.e. its own CRD)? What if you do it as the final key in the struct (e.g. alpha order)? Is it possible to default the mime parts to shell script (the way LTs allow today) to remove some of the cruft?

suket22: I think it does make it look cleaner if it's not inlined. This is what it looks like if it's a ConfigMap instead
```
kind: Provisioner
metadata:
  name: default
spec:
  limits:
    resources:
      cpu: 20
  provider:
    amiFamily: AL2
    userData:
      configMap:
        name: al2-config
        namespace: karpenter
    instanceProfile: KarpenterNodeInstanceProfile-karp-cluster
    subnetSelector:
      karpenter.sh/discovery: karp-cluster
    securityGroupSelector:
      karpenter.sh/discovery: karp-cluster
  ttlSecondsAfterEmpty: 30
```

> Is it possible to default the mime parts to shell script

I don't think we can assume that though because someone might give Karpenter a UserData composed of multiple MIME parts already.

pkg/cloudprovider/aws/testdata/br_userdata_merged.golden
```diff
@@ -0,0 +1,15 @@
...
+max-pods = 110
+
+[settings.kubernetes.node-labels]
+'karpenter.sh/capacity-type' = 'on-demand'
+'karpenter.sh/provisioner-name' = '%s'
+
+[settings.kubernetes.node-taints]
+
+[settings.kubernetes.eviction-hard]
+'memory.available' = '12%%'
```
tzneal, line 15: Where does the double percent sign come from?

suket22: I have to format this file at runtime to inject the provisioner-name on #L10 of this file. `%%` was to escape

tzneal (dismissed):

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
ellistarn: This is what I was getting at:
```
kind: Provisioner
metadata:
  name: default
spec:
  limits:
    resources:
      cpu: 20
  ttlSecondsAfterEmpty: 30
  provider:
    amiFamily: AL2
    instanceProfile: KarpenterNodeInstanceProfile-karp-cluster
    subnetSelector:
      karpenter.sh/discovery: karp-cluster
    securityGroupSelector:
      karpenter.sh/discovery: karp-cluster
    userData:  |
      MIME-Version: 1.0
      Content-Type: multipart/mixed; boundary="//"
      --//
      Content-Type: text/x-shellscript; charset="us-ascii"
      #!/bin/bash
      echo "Running custom user data script"
      --//--      
```
If you format the user data at the bottom, I'm not sure it's so bad. 

I think the namespacing of a configmap is a showstopper since provisioners aren't namespaced. I wouldn't be able to recommend a good namespace to customers. Customers may choose `default` or `karpenter`, but both of these are bad choices (happy to elaborate if this is unintuitive to you). We then end up down the road of recommending customers create a namespace like `provisioner-userdata`, but this may be challenging to setup, since many clusters have tight control over namespaces -- this has security implications.

Minimally, if we do go the route of what you're proposing, does it make sense to do something like `userDataConfigMap` instead of `userData.ConfigMap`, since I don't think userData will ever have additional fields.

ellistarn: w/ v1alpha6, we'd end up with something like this
```
apiVersion: karpenter.sh/v1
kind: Provisioner
metadata:
  name: default
spec:
  limits:
    resources:
      cpu: 20
  providerRef:
    apiVersion: karpenter.aws.k8s/v1
    kind: NodeTemplate
    name: default
  ttlSecondsAfterEmpty: 30
---
apiVersion: karpenter.aws.k8s/v1
kind: NodeTemplate
metadata:
  name: default
spec:
  amiFamily: AL2
  instanceProfile: KarpenterNodeInstanceProfile-karp-cluster
  subnetSelector:
    karpenter.sh/discovery: karp-cluster
  securityGroupSelector:
    karpenter.sh/discovery: karp-cluster
  userData:  |
    MIME-Version: 1.0
    Content-Type: multipart/mixed; boundary="//"
    --//
    Content-Type: text/x-shellscript; charset="us-ascii"
    #!/bin/bash
    echo "Running custom user data script"
    --//--      
```

suket22: Agreed on what it would look like with v1alpha6, I think we're on the same page there. 

@bwagner5 and I spoke about the namespacing issue for a while. I don't think either of us believes Karpenter should have the ability to watch for ConfigMaps across namespaces and it's something we might scope down in the future. I don't think the default namespace makes sense either, but I think the namespace that karpenter components are running within seems like an okay choice to me and is something we'd always have access to watch on. I believe @tzneal has another initiative going to move some Karpenter config into a ConfigMap and we're probably going to rely on the karpenter namespace for that too. 

I can change it to `userDataConfigMap`, that sounds okay to me. I don't fully agree that inlining is better just because we push it to the end. I think having large amounts of TOML / Shell scripts embedded within YAML is just not great UX. and makes editing really tough.

ellistarn: > I don't think the default namespace makes sense either, but I think the namespace that karpenter components are running within seems like an okay choice to me and is something we'd always have access to watch on. 

We default to installation to the karpenter namespace, but some customers prefer to install all system components into a different shared system namespace or even `kube-system`. In these cases, karpenter would have access to read config maps like `aws-auth`. 

> I believe @tzneal has another initiative going to move some Karpenter config into a ConfigMap and we're probably going to rely on the karpenter namespace for that too.

The difference is that the name of this configmap is well known vs granting access to all configmaps. In general, I think we're overly permissive with most of our RBAC rules. Really we should be aiming to only have permission to explicitly named objects that we manage (e.g. `config-logging`, `config-features`) and where necessary, we can have broader permissions (e.g. `pods`). In this example, Todd's new configmap is part of the karpenter installation, whereas what you're proposing is user configuration of karpenter. 

> I think having large amounts of TOML / Shell scripts embedded within YAML is just not great UX. 

Agreed -- but I think it's acceptable if we have line of site to a better long term solution within 2-3 months. 

> Agreed on what it would look like with v1alpha6, I think we're on the same page there.

Cool -- so we're just debating the best path to v1alpha6.

I think it's a more natural path to embed into the provisioner in the short term since the v1alpha6 change would simply be breaking out everything under `provider` into its own CRD (including userdata). Further, this avoids introducing a new concept to customers of what/how to configure configmaps that karpenter will read and then deprecating it a couple months later. Finally -- this also avoids adding RBAC scope (configmaps) to the karpenter controller, and then removing it later. RBAC changes need to be made judiciously due to their impact on the cluster's security posture.

suket22: As discussed offline, moving to a new choice where the UserData is specified in an AWS Provider specific CRD.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -47,11 +47,20 @@ type Constraints struct {
...
+	// ProviderRef is a reference to a dedicated CRD for the chosen provider, that holds
+	// additional configuration options
+	// +optional
+	ProviderRef *ProviderRef `json:"providerRef,omitempty"`
 }
 
 // +kubebuilder:object:generate=false
 type Provider = runtime.RawExtension
 
+type ProviderRef struct {
```
ellistarn, line 59: Thoughts on reusing an existing struct? https://pkg.go.dev/k8s.io/api/autoscaling/v1#CrossVersionObjectReference. This may be a bad choice since it's from `autoscaling` api group, but let me look for others.

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -47,11 +47,20 @@ type Constraints struct {
...
+	metav1.TypeMeta `json:",inline"`
+	Name            *string `json:"name,omitempty"`
```
ellistarn, line 61: Pointers are typically used for optional fields.

pkg/apis/provisioning/v1alpha5/constraints.go
ellistarn: Hmm -- maybe this is bad advice. There are multiple places where this is defined similarly. Perhaps we can just copy the syntax? 

https://github.com/kubernetes/api/search?l=Go&q=APIVersion

```
	// Kind of the referent; More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds"
	Kind string `json:"kind" protobuf:"bytes,1,opt,name=kind"`
	// Name of the referent; More info: http://kubernetes.io/docs/user-guide/identifiers#names
	Name string `json:"name" protobuf:"bytes,2,opt,name=name"`
	// API version of the referent
	// +optional
	APIVersion string `json:"apiVersion,omitempty" protobuf:"bytes,3,opt,name=apiVersion"`
```

suket22: Changed it to this ^. I like being able to control what `Kind` and `APIVersion` are described as in the CRD so inlined rather than using typeMeta.

tzneal (approved):

Issue #1531, opened 2022-03-17
Support another registry private credential


KqLLL:
**Tell us about your request**
Configure another(dockerhub, qury.io) private registry credential in kubelet. 

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
We'll replace CA with karpenter, we need to config private registry credential in worker node(provision node).

**Are you currently working around this issue?**
Set kubelet/config.json on bootstrap.sh via [tf-eks-module](https://registry.terraform.io/modules/terraform-aws-modules/eks/aws/latest)
Example:
```
pre_bootstrap_user_data = <<EOF
......
cat <<-BOOTSTRAP > /etc/profile.d/bootstrap.sh
cat >> /var/lib/kubelet/config.json << CONFIG
{
  auths: {
    "quay.io": {}
  }
}
CONFIG
EOF
```
**Additional context**
Support AL and Bottlerocket ami type.
**Attachments**
None

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

KqLLL:
On Bottlerocket
```
bootstrap_extra_args = <<-EOF
[settings.host-containers.admin]
enabled = false

[[settings.container-registry.credentials]]
registry = "quay.io"
auth = "xxxx"

EOF
```

tzneal:
You can now use custom userdata to add any required credentials.

PR #1726, opened 2022-04-26
Support set container registry credentials in Provisioner

KqLLL:
**1. Issue, if available:**
 Issue: #1531
Support set container registry credentials for provision worker node by Provisioner CR.

This change adds support for customUserData with AMIFamily. It would config registry credentials on UserData.
**Reference**:
Bottlerocket: https://github.com/bottlerocket-os/bottlerocket (settings.container-registry.credentials).

Add containerRegistry config to your provisioner CR;
Example
```yaml
  kubeletConfiguration:
    containerRegistry:
      credentials:
        - registry: "registry.rightcapital.io"
          username: "<username>"
          password: "<password>"
        - registry: "quay.io"
          auth: "<your token>"
```

- [x] No


pkg/apis/provisioning/v1alpha5/kubelet_configuration.go
```diff
@@ -22,4 +22,19 @@ type KubeletConfiguration struct {
 	// Note that not all providers may use all addresses.
 	//+optional
 	ClusterDNS []string `json:"clusterDNS,omitempty"`
+	// containerRegistry setting is optional and allows to configure image registry credentials.
+	//+optional
+	ContainerRegistry *ContainerRegistry `json:"containerRegistry,omitempty"`
```
suket22, line 27: Thanks for the PR!

We want to limit the API surface that we expose on the Provisioner and adding every configuration possible for the kubelet or the container-runtime might cause difficulties with maintenance going forward. For example even here, you had to add support both with BottleRocket and the EKS Optimized AMIs but once we add Windows, it introduces another dimension in the matrix. 
We also need to be vendor neutral in the provisioning API, so everything that's added here ideally should apply across all cloud-providers.  

1. Once we get the PR for CustomUserData merged [`in for Bottlerocket`](https://github.com/aws/karpenter/pull/1720), I think we can rely on just setting the container-registry settings there. You could rebase this PR to include your changes in `bottlerocket.go` into `bottlerocketsettings.go`. 

2. For the AL2 AMI part, I think we want folks to update the kubeletConfig via UserData themselves. So a UserData like - 
```
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="//"

--//
Content-Type: text/x-shellscript; charset="us-ascii"
#!/bin/bash

KUBELET_CONFIG=/etc/kubernetes/kubelet/kubelet-config.json
echo "$(jq ".credentialProvider=...." $KUBELET_CONFIG)" > $KUBELET_CONFIG

--//--
```

Issue #972, opened 2021-12-12
Karpenter not setting CCM uninitialized taint


olemarkus:
## Version

**Karpenter**: v0.5.2

**Kubernetes**: v1.23.0

## Expected Behavior

When external cloud controller manager is used, kubelet will taint Nodes with `node.cloudprovider.kubernetes.io/uninitialized`, which the cloud-controller-manager is expected to remove once the Node object has been initialised.

It may be an idea to ensure general "Initialisation taints"–that is taints that Karpenter should set and that is expected will be removed once controller-managers like CCM has initialised the Node.

## Actual Behavior

No taints

## Steps to Reproduce the Problem

Run a cluster using AWS CCM and observe the taint is missing on Karpenter-managed Nodes.

ellistarn:
Interesting -- I'm not familiar with this, but it seems plausible that we should do this. Do you have any links to code/documentation as to how this taint is designed to work?

ellistarn:
Related to: https://github.com/aws/karpenter/issues/628

olemarkus:
https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/#running-cloud-controller-manager explains this particular taint.

https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions this page lists a few others as well. It claims it is node controller that sets this, but it is actually kubelet that does it. 

The code is here https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_node_status.go#L286

ellistarn:
I'll have to play around with this, but as far as I understand, if we set the --external cloud provider flag, then kubelet should add the taint and CCM should be able to remove it. It's not clear how karpenter gets in the way.

olemarkus:
The code around https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_node_status.go#L108-L122
shows what on a Node object is updated by kubelet starting up with the Node object already existing. It patches in things that only kubelet can truly know, like the underlying OS, arch etc. It patches in capacity.

But it does not patch in taints. The appropriate taints have to be added as part of Node creation, and since this is done by karpenter, karpenter has to emulate Kubelet behavior here.

Issue #1584, opened 2022-03-29
Record Karpenter events on Pods


tzneal:
**Tell us about your request**

Use the controller manager's `GetEventRecorderFor()` method to get an event recorder for recording events related to pods directly to the pods.  This will allow us to log our scheduling related activities to the pod.

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
I want to see why Karpenter didn't schedule a pod by looking at `kubectl describe pod pod-name`

**Are you currently working around this issue?**
I need to look at the controller's logs.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

diranged:
Huzzah!

PR #1727, opened 2022-04-26
support awareness of in-flight nodes & startup taints

### LINKED ISSUE #628: Support extra taints not considered when scheduling Pods

tzneal:
**1. Issue, if available:**

Fixes #628 and #972

- don't launch nodes if the pod can schedule on a recently launch
  node that hasn't become ready
- support startup taints

Co-authored-by: Bryan Stenson <bryan.stenson@okta.com>

Unit testing & scaling cluster on EKS. I tested with two EKS clusters, one using Cilium and one standard cluster.

- [ ] No

### DIFF
--- a/charts/karpenter/crds/karpenter.sh_provisioners.yaml
+++ b/charts/karpenter/crds/karpenter.sh_provisioners.yaml
@@ -106,6 +106,39 @@ spec:
                   - operator
                   type: object
                 type: array
+              startupTaints:
+                description: StartupTaints are taints that are applied to nodes upon
+                  startup but which are expected to be removed automatically within
+                  a short period of time, typically by a DaemonSet that tolerates
+                  the taint. These are commonly used by daemonsets to allow initialization
+                  and enforce startup ordering.  StartupTaints are ignored for provisioning
+                  purposes in that pods are not required to tolerate a StartupTaint
+                  in order to have nodes provisioned for them.
+                items:
+                  description: The node this Taint is attached to has the "effect"
+                    on any pod that does not tolerate the Taint.
+                  properties:
+                    effect:
+                      description: Required. The effect of the taint on pods that
+                        do not tolerate the taint. Valid effects are NoSchedule, PreferNoSchedule
+                        and NoExecute.
+                      type: string
+                    key:
+                      description: Required. The taint key to be applied to a node.
+                      type: string
+                    timeAdded:
+                      description: TimeAdded represents the time at which the taint
+                        was added. It is only written for NoExecute taints.
+                      format: date-time
+                      type: string
+                    value:
+                      description: The taint value corresponding to the taint key.
+                      type: string
+                  required:
+                  - effect
+                  - key
+                  type: object
+                type: array
               taints:
                 description: Taints will be applied to every node launched by the
                   Provisioner. If specified, the provisioner will not provision nodes
--- a/charts/karpenter/crds/karpenter.sh_provisioners.yaml
+++ b/charts/karpenter/crds/karpenter.sh_provisioners.yaml
@@ -108,9 +108,9 @@ spec:
                 type: array
               startupTaints:
                 description: StartupTaints are taints that are applied to nodes upon
-                  startup but which are expected to be removed automatically within
-                  a short period of time, typically by a DaemonSet that tolerates
-                  the taint. These are commonly used by daemonsets to allow initialization
+                  startup which are expected to be removed automatically within a
+                  short period of time, typically by a DaemonSet that tolerates the
+                  taint. These are commonly used by daemonsets to allow initialization
                   and enforce startup ordering.  StartupTaints are ignored for provisioning
                   purposes in that pods are not required to tolerate a StartupTaint
                   in order to have nodes provisioned for them.
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/aws/karpenter/pkg/events"
+
 	"github.com/aws/karpenter/pkg/controllers/state"
 	"github.com/aws/karpenter/pkg/utils/project"
 
@@ -77,6 +79,8 @@ func main() {
 
 	logging.FromContext(ctx).Infof("Initializing with version %s", project.Version)
 	// Set up controller runtime controller
+	var recorder events.Recorder = &events.NoOpRecorder{}
+
 	cloudProvider := registry.NewCloudProvider(ctx, cloudprovider.Options{ClientSet: clientSet})
 	cloudProvider = cloudprovidermetrics.Decorate(cloudProvider)
 	manager := controllers.NewManagerOrDie(ctx, config, controllerruntime.Options{
@@ -91,7 +95,7 @@ func main() {
 	cluster := state.NewCluster(ctx, manager.GetClient())
 
 	if err := manager.RegisterControllers(ctx,
-		provisioning.NewController(ctx, manager.GetClient(), clientSet.CoreV1(), cloudProvider, cluster),
+		provisioning.NewController(ctx, manager.GetClient(), clientSet.CoreV1(), recorder, cloudProvider, cluster),
 		state.NewNodeController(manager.GetClient(), cluster),
 		state.NewPodController(manager.GetClient(), cluster),
 		persistentvolumeclaim.NewController(manager.GetClient()),
--- a/pkg/apis/provisioning/v1alpha5/constraints.go
+++ b/pkg/apis/provisioning/v1alpha5/constraints.go
@@ -33,6 +33,12 @@ type Constraints struct {
 	// pod tolerations on a per-node basis.
 	// +optional
 	Taints Taints `json:"taints,omitempty"`
+	// StartupTaints are taints that are applied to nodes upon startup but which are expected to be removed automatically
+	// within a short period of time, typically by a DaemonSet that tolerates the taint. These are commonly used by
+	// daemonsets to allow initialization and enforce startup ordering.  StartupTaints are ignored for provisioning
+	// purposes in that pods are not required to tolerate a StartupTaint in order to have nodes provisioned for them.
+	// +optional
+	StartupTaints Taints `json:"startupTaints,omitempty"`
 	// Requirements are layered with Labels and applied to every node.
 	Requirements Requirements `json:"requirements,inline,omitempty"`
 	// KubeletConfiguration are options passed to the kubelet when provisioning nodes
@@ -61,25 +67,28 @@ func (c *Constraints) ToNode() *v1.Node {
 			}
 		}
 	}
+
+	// both the taints and startup taints are applied to nodes we create
+	taints := append(c.Taints, c.StartupTaints...)
+
+	// Taint karpenter.sh/not-ready=NoSchedule to prevent the kube scheduler
+	// from scheduling pods before we're able to bind them ourselves. The kube
+	// scheduler has an eventually consistent cache of nodes and pods, so it's
+	// possible for it to see a provisioned node before it sees the pods bound
+	// to it. This creates an edge case where other pending pods may be bound to
+	// the node by the kube scheduler, causing OutOfCPU errors when the
+	// binpacked pods race to bind to the same node. The system eventually
+	// heals, but causes delays from additional provisioning (thrash). This
+	// taint will be removed by the node controller when a node is marked ready.
+	taints = append(taints, v1.Taint{Key: NotReadyTaintKey, Effect: v1.TaintEffectNoSchedule})
+
 	return &v1.Node{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:     labels,
 			Finalizers: []string{TerminationFinalizer},
 		},
 		Spec: v1.NodeSpec{
-			// Taint karpenter.sh/not-ready=NoSchedule to prevent the kube scheduler
-			// from scheduling pods before we're able to bind them ourselves. The kube
-			// scheduler has an eventually consistent cache of nodes and pods, so it's
-			// possible for it to see a provisioned node before it sees the pods bound
-			// to it. This creates an edge case where other pending pods may be bound to
-			// the node by the kube scheduler, causing OutOfCPU errors when the
-			// binpacked pods race to bind to the same node. The system eventually
-			// heals, but causes delays from additional provisioning (thrash). This
-			// taint will be removed by the node controller when a node is marked ready.
-			Taints: append(c.Taints, v1.Taint{
-				Key:    NotReadyTaintKey,
-				Effect: v1.TaintEffectNoSchedule,
-			}),
+			Taints: taints,
 		},
 	}
 }
--- a/pkg/apis/provisioning/v1alpha5/constraints.go
+++ b/pkg/apis/provisioning/v1alpha5/constraints.go
@@ -33,7 +33,7 @@ type Constraints struct {
 	// pod tolerations on a per-node basis.
 	// +optional
 	Taints Taints `json:"taints,omitempty"`
-	// StartupTaints are taints that are applied to nodes upon startup but which are expected to be removed automatically
+	// StartupTaints are taints that are applied to nodes upon startup which are expected to be removed automatically
 	// within a short period of time, typically by a DaemonSet that tolerates the taint. These are commonly used by
 	// daemonsets to allow initialization and enforce startup ordering.  StartupTaints are ignored for provisioning
 	// purposes in that pods are not required to tolerate a StartupTaint in order to have nodes provisioned for them.
--- a/pkg/apis/provisioning/v1alpha5/taints.go
+++ b/pkg/apis/provisioning/v1alpha5/taints.go
@@ -27,7 +27,7 @@ type Taints []v1.Taint
 // Has returns true if taints has a taint for the given key and effect
 func (ts Taints) Has(taint v1.Taint) bool {
 	for _, t := range ts {
-		if t.Key == taint.Key && t.Effect == taint.Effect {
+		if t.MatchTaint(&taint) {
 			return true
 		}
 	}
@@ -44,12 +44,16 @@ func (ts Taints) HasKey(taintKey string) bool {
 	return false
 }
 
-// Tolerates returns true if the pod tolerates all taints
-func (ts Taints) Tolerates(pod *v1.Pod) (errs error) {
+// Tolerates returns true if the pod tolerates all taints. 'additional' are extra tolerations for things like startup
+// taints or the standard not-ready taint applied to nodes we have launched.
+func (ts Taints) Tolerates(pod *v1.Pod, additional ...v1.Toleration) (errs error) {
 	for i := range ts {
 		taint := ts[i]
 		tolerates := false
 		for _, t := range pod.Spec.Tolerations {
+			tolerates = tolerates || t.ToleratesTaint(&taint)
+		}
+		for _, t := range additional {
 			tolerates = tolerates || t.ToleratesTaint(&taint)
 		}
 		if !tolerates {
@@ -57,4 +61,14 @@ func (ts Taints) Tolerates(pod *v1.Pod) (errs error) {
 		}
 	}
 	return errs
+}
+
+// TaintToToleration converts a taint to a toleration that tolerates the specified taint.
+func TaintToToleration(taint v1.Taint) v1.Toleration {
+	return v1.Toleration{
+		Key:      taint.Key,
+		Operator: v1.TolerationOpEqual,
+		Value:    taint.Value,
+		Effect:   taint.Effect,
+	}
 }
--- a/pkg/apis/provisioning/v1alpha5/util.go
+++ b/pkg/apis/provisioning/v1alpha5/util.go
@@ -18,7 +18,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-// NodeIsReady is returns true if the all of the startup taints have been removed from the node and it's current status
+// NodeIsReady returns true if all the startup taints have been removed from the node and its current status
 // is set to Ready
 func NodeIsReady(node *v1.Node, provisioner *Provisioner) bool {
 	for _, startupTaint := range provisioner.Spec.StartupTaints {
--- a/pkg/apis/provisioning/v1alpha5/util.go
+++ b/pkg/apis/provisioning/v1alpha5/util.go
@@ -12,13 +12,23 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package node
+package v1alpha5
 
 import (
 	v1 "k8s.io/api/core/v1"
 )
 
-func IsReady(node *v1.Node) bool {
+// NodeIsReady is returns true if the all of the startup taints have been removed from the node and it's current status
+// is set to Ready
+func NodeIsReady(node *v1.Node, provisioner *Provisioner) bool {
+	for _, startupTaint := range provisioner.Spec.StartupTaints {
+		for i := 0; i < len(node.Spec.Taints); i++ {
+			// if the node still has a startup taint applied, it's not ready
+			if startupTaint.MatchTaint(&node.Spec.Taints[i]) {
+				return false
+			}
+		}
+	}
 	return GetCondition(node.Status.Conditions, v1.NodeReady).Status == v1.ConditionTrue
 }
 
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/aws/karpenter/pkg/events"
+
 	"github.com/aws/karpenter/pkg/controllers/state"
 
 	"github.com/imdario/mergo"
@@ -42,7 +44,7 @@ import (
 	"github.com/aws/karpenter/pkg/utils/resources"
 )
 
-func NewProvisioner(ctx context.Context, kubeClient client.Client, coreV1Client corev1.CoreV1Interface, cloudProvider cloudprovider.CloudProvider, cluster *state.Cluster) *Provisioner {
+func NewProvisioner(ctx context.Context, kubeClient client.Client, coreV1Client corev1.CoreV1Interface, recorder events.Recorder, cloudProvider cloudprovider.CloudProvider, cluster *state.Cluster) *Provisioner {
 	running, stop := context.WithCancel(ctx)
 	p := &Provisioner{
 		Stop:           stop,
@@ -52,6 +54,7 @@ func NewProvisioner(ctx context.Context, kubeClient client.Client, coreV1Client 
 		coreV1Client:   coreV1Client,
 		volumeTopology: NewVolumeTopology(kubeClient),
 		cluster:        cluster,
+		recorder:       recorder,
 	}
 	p.cond = sync.NewCond(&p.mu)
 	go func() {
@@ -76,6 +79,7 @@ type Provisioner struct {
 	batcher        *Batcher
 	volumeTopology *VolumeTopology
 	cluster        *state.Cluster
+	recorder       events.Recorder
 
 	mu   sync.Mutex
 	cond *sync.Cond
@@ -191,7 +195,7 @@ func (p *Provisioner) schedule(ctx context.Context, pods []*v1.Pod) ([]*scheduli
 		return nil, fmt.Errorf("getting daemon overhead, %w", err)
 	}
 
-	return scheduling.NewScheduler(provisioners, topology, instanceTypes, daemonOverhead).Solve(ctx, pods)
+	return scheduling.NewScheduler(provisioners, p.cluster, topology, instanceTypes, daemonOverhead, p.recorder).Solve(ctx, pods)
 }
 
 func (p *Provisioner) launch(ctx context.Context, node *scheduling.Node) error {
@@ -203,12 +207,15 @@ func (p *Provisioner) launch(ctx context.Context, node *scheduling.Node) error {
 	if err := latest.Spec.Limits.ExceededBy(latest.Status.Resources); err != nil {
 		return err
 	}
+
+	// apply both the taints and startup taints to the node
+	taints := append(node.Provisioner.Spec.Taints, node.Provisioner.Spec.StartupTaints...)
 	k8sNode, err := p.cloudProvider.Create(ctx, &cloudprovider.NodeRequest{
 		InstanceTypeOptions: node.InstanceTypeOptions,
 		Template: &cloudprovider.NodeTemplate{
 			Provider:             node.Provisioner.Spec.Provider,
 			Labels:               node.Provisioner.Spec.Labels,
-			Taints:               node.Provisioner.Spec.Taints,
+			Taints:               taints,
 			Requirements:         node.Provisioner.Spec.Requirements,
 			KubeletConfiguration: node.Provisioner.Spec.KubeletConfiguration,
 		},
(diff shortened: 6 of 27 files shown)


website/content/en/preview/provisioner.md
```diff
@@ -23,10 +23,18 @@ spec:
...
   # Provisioned nodes will have these taints
-  # Taints may prevent pods from scheduling if they are not tolerated
+  # Taints may prevent pods from scheduling if they are not tolerated by the pod or listed as a startup taint.
   taints:
     - key: example.com/special-taint
       effect: NoSchedule
+      
+      
+  # Startup taints are those that are expected to be temporary. Pods don't need to tolerate these taints to be
+  # considered by the provisioner.  These taints are also ignored for five minutes for the purposes of determing
```
BryanStenson-okta, line 33: ^determing^determining

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -33,6 +33,10 @@ type Constraints struct {
 	// pod tolerations on a per-node basis.
 	// +optional
 	Taints Taints `json:"taints,omitempty"`
+	// StartupTaints are taints that may be applied to nodes upon startup, but are expected to be removed automatically
+	// within a short period of time. These are commonly used by daemonsets to allow initialization.
+	// +optional
```
tzneal, line 38: Clarify that these don't cause taints to be added to nodes.

pkg/controllers/provisioning/scheduling/inflightnode.go
```diff
@@ -0,0 +1,117 @@
...
+		requirements:    v1alpha5.NewLabelRequirements(n.Node.Labels),
+	}
+
+	// TODO(todd): This concept of a 'young' node seems a bit hacky, but not sure of a better approach.  The thought is that
+	// if the node truly fails for some reason, we don't want to wait the full 15 minute timeout before the node is deleted
+	// to schedule pods.
+
+	// If the node is "young", we add tolerations for startup taints.  If the node is older, we assume that these startup
+	// taints will never go away and can't potentially schedule pods to the node
+	if time.Since(node.Node.CreationTimestamp.Time) < 5*time.Minute {
```
tzneal, line 55: Remove this to eliminate the two different timeout values

pkg/controllers/provisioning/scheduling/inflightnode.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// check resource requests first since that's a pretty likely reason the pod won't schedule on an in-flight
+	// node, which at this point can't be increased in size
+	requests := resources.Merge(n.requests, resources.RequestsForPods(pod))
+
+	// TODO(todd): this isn't fully correct, if the node is already 'Ready', then the daemon sets are possibly
+	// already accounted for.  Does this really matter though? Worst case we create more nodes than absolutely needed
+	// but it should unlikely unless daemonsets consume large amounts of resources.  Not sure of a reliable way to
+	// determine if the daemon set pods are already on the node.
+	if !resources.Fits(resources.Merge(n.daemonResources, requests), n.available) {
```
tzneal, line 89: Try to track this in cluster state.

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -34,13 +34,14 @@ type Node struct {
 	InstanceTypeOptions []cloudprovider.InstanceType
 	Pods                []*v1.Pod
 
-	topology *Topology
-	requests v1.ResourceList
+	topology    *Topology
+	requests    v1.ResourceList
+	tolerations []v1.Toleration
```
tzneal, line 39: Rename to startupTolerations

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -45,9 +48,32 @@ type Scheduler struct {
...
 
-func (s *Scheduler) Solve(ctx context.Context, pods []*v1.Pod) ([]*Node, error) {
+func (s *Scheduler) Solve(ctx context.Context, pods []*v1.Pod) ([]*Node, []*InFlightNode, error) {
+	provisionerMap := map[string]*v1alpha5.Provisioner{}
+	for _, provisioner := range s.provisioners {
+		provisionerMap[provisioner.Name] = provisioner
+	}
+
+	// create our in-flight nodes
+	s.cluster.ForEachNode(func(node *state.Node) bool {
```
tzneal, line 62: Look at the deleted timestamp.

tzneal: Already handled by cluster state, just adding a test.

tzneal:
Tested and I am able to launch the cilium connectivity test on a single node cluster. This leads to unschedulable pods for the multi-node tests, so Karpenter launches a second node and doesn't bind the pods to the node due to the startup taint.  Eventually cilium becomes ready, removes the startup taint and kube-scheduler binds the pods.

```
NAMESPACE     NAME                                                     READY   STATUS    RESTARTS   AGE     IP                NODE                                           NOMINATED NODE   READINESS GATES
cilium-test   echo-a-c4cdff77c-jr6bh                                   1/1     Running   0          4m7s    192.168.156.116   ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
cilium-test   echo-b-598c78b9fc-h458m                                  1/1     Running   0          4m6s    192.168.34.22     ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
cilium-test   echo-b-host-5556f9488-rxdkk                              1/1     Running   0          4m5s    192.168.55.185    ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
cilium-test   host-to-b-multi-node-clusterip-7f8c9699d6-ppc62          1/1     Running   0          4m1s    192.168.11.65     ip-192-168-11-65.us-west-2.compute.internal    <none>           <none>
cilium-test   host-to-b-multi-node-headless-7df4d6fdb-zkfc4            1/1     Running   0          4m1s    192.168.11.65     ip-192-168-11-65.us-west-2.compute.internal    <none>           <none>
cilium-test   pod-to-a-679f686cb-frndq                                 1/1     Running   0          4m5s    192.168.50.51     ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
cilium-test   pod-to-a-allowed-cnp-54755cc9c6-7tr2v                    1/1     Running   0          4m3s    192.168.158.115   ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
cilium-test   pod-to-a-denied-cnp-7bfb7d69b8-kv4wf                     1/1     Running   0          4m4s    192.168.138.176   ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
cilium-test   pod-to-b-intra-node-nodeport-6bd8d9468d-4cncq            1/1     Running   0          3m59s   192.168.158.6     ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
cilium-test   pod-to-b-multi-node-clusterip-7984ccf8c6-v9vm9           1/1     Running   0          4m2s    192.168.100.199   ip-192-168-11-65.us-west-2.compute.internal    <none>           <none>
cilium-test   pod-to-b-multi-node-headless-7fb5f5c84b-g4654            1/1     Running   0          4m2s    192.168.117.93    ip-192-168-11-65.us-west-2.compute.internal    <none>           <none>
cilium-test   pod-to-b-multi-node-nodeport-54c89d778-hhggm             1/1     Running   0          4m      192.168.12.141    ip-192-168-11-65.us-west-2.compute.internal    <none>           <none>
cilium-test   pod-to-external-1111-c98db84d4-6qdph                     1/1     Running   0          4m4s    192.168.59.118    ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
cilium-test   pod-to-external-fqdn-allow-google-cnp-5f55d8886b-846hw   1/1     Running   0          4m3s    192.168.62.239    ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
karpenter     karpenter-c4954dbf-wwmtl                                 2/2     Running   0          5m20s   192.168.153.126   ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
kube-system   cilium-52q78                                             1/1     Running   0          12h     192.168.55.185    ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
kube-system   cilium-node-init-9x2xg                                   1/1     Running   0          3m55s   192.168.11.65     ip-192-168-11-65.us-west-2.compute.internal    <none>           <none>
kube-system   cilium-node-init-z9vb7                                   1/1     Running   0          12h     192.168.55.185    ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
kube-system   cilium-operator-6689cf95b6-thpwb                         1/1     Running   0          12h     192.168.55.185    ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
kube-system   cilium-vb6l7                                             1/1     Running   0          3m13s   192.168.11.65     ip-192-168-11-65.us-west-2.compute.internal    <none>           <none>
kube-system   coredns-85d5b4454c-4p9s9                                 1/1     Running   0          13h     192.168.130.87    ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
kube-system   coredns-85d5b4454c-rq4vr                                 1/1     Running   0          34h     192.168.43.71     ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
kube-system   kube-proxy-5lrqf                                         1/1     Running   0          34h     192.168.55.185    ip-192-168-55-185.us-west-2.compute.internal   <none>           <none>
kube-system   kube-proxy-x92cm                                         1/1     Running   0          3m13s   192.168.11.65     ip-192-168-11-65.us-west-2.compute.internal    <none>           <none>
```

pkg/controllers/provisioning/provisioner.go
```diff
@@ -111,11 +111,20 @@ func (p *Provisioner) provision(ctx context.Context) error {
...
-	nodes, err := p.schedule(ctx, pods)
+	nodes, inflight, err := p.schedule(ctx, pods)
 	if err != nil {
 		return err
 	}
 
+	// TODO(todd): do we want to do this?  There may be taints that pods don't tolerate.
+	// Bind pods to in-flight nodes
+	workqueue.ParallelizeUntil(ctx, len(inflight), len(inflight), func(i int) {
+		node := inflight[i]
```
suket22, line 122: If we do do this, should we have a debug log describing a bind to an in-flight node? It might surprise users that are expecting a new node to launch but none were

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -45,9 +48,32 @@ type Scheduler struct {
...
+			// ignoring this node as it wasn't launched by us
+			return true
+		}
+		provisioner, ok := provisionerMap[provisionerName]
+		if !ok {
+			// ignoring this node as it wasn't launched by a provisioner that we recognize
+			return true
+		}
+		s.inflight = append(s.inflight, NewInFlightNode(node, s.topology, provisioner.Spec.StartupTaints, s.daemonOverhead[provisioner]))
+		return true
```
suket22, line 74: If an instance doesn't launch because of an ICE or because the EC2 service is unschedulable, wouldn't we end up considering this node object as `inflight` ? Should we only be doing this once we've kicked off the node, rather than after we've completed successful scheduling?

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -33,6 +33,13 @@ type Constraints struct {
 	// pod tolerations on a per-node basis.
 	// +optional
 	Taints Taints `json:"taints,omitempty"`
+	// StartupTaints are taints that may be applied to nodes upon startup either by the provisioner or a daemonset,
+	// but which are expected to be removed automatically within a short period of time. These are commonly used by
+	// daemonsets to allow initialization.  A startup taint on the provisioner does not cause that taint to be added to
+	// a node. If it appears on a node by also being configured via Taints or by being added by an external mechanism
+	// it will be ignored for provisioning purposes.
+	// +optional
+	StartupTaints Taints `json:"startupTaints,omitempty"`
```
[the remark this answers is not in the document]
tzneal: I agree this was confusing.  I've left it named as 'StartupTaints' for now, but I've modified it so that it now actually causes the taints to be applied to the node.    You'll just list the taint as a startup taint, we put it on the node but ignore it for scheduling and assume someone else will remove it.

pkg/controllers/provisioning/provisioner.go
tzneal: I'm about to push up a new PR that starts to add support for recording events on the pods (at least an interface for it).  We needed something like that for unit tests anyway as we need to know if the scheduling decisions were correct, even without the pod binding.  There is a separate issue at #1584 to turn those into actual events on the pods.  It's more code and requires changing permissions, so I didn't include it in this PR.

pkg/controllers/provisioning/scheduling/scheduler.go
tzneal: This shouldn't occur, we get a success from the instance launch before we create the node object.  The inflight nodes here are ones that were created in previous rounds of scheduling.

There is a scenario where a node fails to become ready and we think it will support pods.  It may prevent provisioning whatever excess capacity it has for 15 minutes, but it will resolve.  The same issue occurs currently where we bind the pods to the node that never comes up and they don't retry for 15 minutes.

pkg/controllers/provisioning/scheduling/inflightnode.go
```diff
@@ -0,0 +1,121 @@
...
+		v1.NodeSelectorRequirement{
+			Key:      v1.LabelHostname,
+			Operator: v1.NodeSelectorOpIn,
+			Values:   []string{hostname},
+		})
+	topology.Register(v1.LabelHostname, hostname)
+	return node
+}
+
+func (n *InFlightNode) Add(pod *v1.Pod) error {
```
bwagner5, line 82: any more thought on combining or at least re-using some of the logic in  `scheduling/node.go`?

bwagner5 (commented):
overall, looks good! No major concerns, just a few minor comments.

charts/karpenter/crds/karpenter.sh_provisioners.yaml
```diff
@@ -106,6 +106,39 @@ spec:
...
+                  properties:
+                    effect:
+                      description: Required. The effect of the taint on pods that
+                        do not tolerate the taint. Valid effects are NoSchedule, PreferNoSchedule
+                        and NoExecute.
+                      type: string
+                    key:
+                      description: Required. The taint key to be applied to a node.
+                      type: string
+                    timeAdded:
```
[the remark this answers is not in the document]
tzneal: We don't, but it's part of the v1.Taint.  The same field is on each of our regular Taints. We would need to create our own type to remove it.

cmd/controller/main.go
```diff
@@ -77,6 +79,8 @@ func main() {
 
 	logging.FromContext(ctx).Infof("Initializing with version %s", project.Version)
 	// Set up controller runtime controller
+	var recorder events.Recorder = &events.NoOpRecorder{}
```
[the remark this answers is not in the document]
tzneal: I did this specifically so that everything that uses `recorder` would have to use the `events.Recorder` interface and we couldn't accidentally use the real type.

ellistarn:
I'm wary of the long term impact of including startup taints in the api. It's discussed in a number of places, but I'll restate here. Upstream has a concept to address this use case (including the node restart case) called the NoAdmit taint effect. This allows scheduling but prevents the kubelet from running the pod, which is exactly what we need. 

I'd hate to include this in karpenters API as a point in time workaround for something that will eventually be fixed upstream. One option is to support this taint effect in karpenter before it's merged upstream and do the upstream work in parallel. 

There may be reasons to not do this, and we can always deprecate a field using an API bump, but I want to make sure we're rigorously evaluating our options for changes that expand the API.

tzneal:
> I'm wary of the long term impact of including startup taints in the api. It's discussed in a number of places, but I'll restate here. Upstream has an unimplemented concept to address this use case (including the node restart case) called the NoAdmit taint effect. This allows scheduling but prevents the kubelet from running the pod, which is exactly what we need. One benefit of doing this upstream is that the change applies to all integrators like CAS, Karpenter, and the Descheduler.
> 

I don't think that NoAdmit solves this problem.  From the docs, NoAdmit is "NoSchedule" + don't start pods that are scheduled.  If it were "don't start pods only", maybe it could work but as it, it doesn't fully convey the concept of a temporary taint applied at startup that will be eventually removed.   Even if it were "don't start pods", I don't think it still fully implies that the taint is temporary.


> I'd hate to include this in karpenters API as a point in time workaround for something that will eventually be fixed upstream. One option is to support this taint effect in karpenter before it's merged upstream and do the upstream work in parallel.

The NoAdmit taint was introduced, commented out, in [2016](https://github.com/kubernetes/kubernetes/commit/52fb89ff7326b3570e9c42a435081dab559f610b) . That appears to be the initial implementation of taints and it hasn't been needed yet.  Agree that it doesn't seem to be a priority and we would need to do the work, but I also don't think it solves this particular problem.

To solve the problem, we would also need to treat 'NoAdmit' as a temporary taint by provisioning capacity for workloads that don't tolerate it and then assume something would remove the taint.  We would have implemented the concept of startup taints only hidden behind a magic effect name. If NoAdmit is ever implemented upstream, Karpenter would then be supporting it incorrectly as nothing in the NoAdmit docs indicate that it will always be removed.

pkg/controllers/provisioning/scheduling/inflightnode.go
tzneal: I didn't see a good way to accomplish it, there are quite a few differences.  I started to refactor out some of the common logic, but it didn't seem to help much.

charts/karpenter/crds/karpenter.sh_provisioners.yaml
```diff
@@ -106,6 +106,39 @@ spec:
                   - operator
                   type: object
                 type: array
+              startupTaints:
+                description: StartupTaints are taints that are applied to nodes upon
+                  startup but which are expected to be removed automatically within
```
njtran, line 111:
```suggestion
                  startup which are expected to be removed automatically within
```

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -33,6 +33,12 @@ type Constraints struct {
 	// pod tolerations on a per-node basis.
 	// +optional
 	Taints Taints `json:"taints,omitempty"`
+	// StartupTaints are taints that are applied to nodes upon startup but which are expected to be removed automatically
```
njtran, line 36: Same thing here
```suggestion
	// StartupTaints are taints that are applied to nodes upon startup which are expected to be removed automatically
```

website/content/en/preview/tasks/provisioning.md
```diff
@@ -78,3 +80,20 @@ spec:
     effect: “NoSchedule”
 ```
 In order for a pod to run on a node defined in this provisioner, it must tolerate `nvidia.com/gpu` in its pod spec.
+
+### Example: Adding the Cilium Startup Taint
+
+Per the Cilium [docs](https://docs.cilium.io/en/stable/gettingstarted/taints/),  it's recommended to place a taint of `node.cilium.io/agent-not-ready=true:NoExecute` on nodes to allow Cilium to configure networking prior to other pods starting.  This can be accomplished via the user of Karpenter `startupTaints`.  These taints are placed on the node, but pods aren't required to tolerate these taints to be considered for provisioning.
```
bwagner5, line 86:
```suggestion
Per the Cilium [docs](https://docs.cilium.io/en/stable/gettingstarted/taints/),  it's recommended to place a taint of `node.cilium.io/agent-not-ready=true:NoExecute` on nodes to allow Cilium to configure networking prior to other pods starting.  This can be accomplished via the use of Karpenter `startupTaints`.  These taints are placed on the node, but pods aren't required to tolerate these taints to be considered for provisioning.
```

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -61,25 +67,28 @@ func (c *Constraints) ToNode() *v1.Node {
 			}
 		}
 	}
+
+	// both the taints and startup taints are applied to nodes we create
+	taints := append(c.Taints, c.StartupTaints...)
```
njtran, line 72: nit: Is there a reason you don't want to inline this with the existing pattern? Is it to have the comment in there?

pkg/controllers/provisioning/scheduling/inflightnode.go
```diff
@@ -0,0 +1,121 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scheduling
+
+import (
+	"fmt"
+
+	"github.com/aws/karpenter/pkg/controllers/state"
```
njtran, line 20: nit: can you move this down to group with the other github.com imports

pkg/apis/provisioning/v1alpha5/taints.go
```diff
@@ -44,17 +44,31 @@ func (ts Taints) HasKey(taintKey string) bool {
...
+		}
 		if !tolerates {
 			errs = multierr.Append(errs, fmt.Errorf("did not tolerate %s=%s:%s", taint.Key, taint.Value, taint.Effect))
 		}
 	}
 	return errs
 }
+
+// TaintToToleration converts a taint to a toleration that tolerates the specified taint.
+func TaintToToleration(taint v1.Taint) v1.Toleration {
```
njtran, line 67: Many tolerations can tolerate the same taint. Should we make the toleration just tolerate the key, not necessarily the value? We also just want NoScheduler here in all cases, right?

pkg/apis/provisioning/v1alpha5/util.go
```diff
@@ -12,13 +12,23 @@ See the License for the specific language governing permissions and
...
 )
 
-func IsReady(node *v1.Node) bool {
+// NodeIsReady returns true if all the startup taints have been removed from the node and its current status
+// is set to Ready
+func NodeIsReady(node *v1.Node, provisioner *Provisioner) bool {
+	for _, startupTaint := range provisioner.Spec.StartupTaints {
+		for i := 0; i < len(node.Spec.Taints); i++ {
+			// if the node still has a startup taint applied, it's not ready
+			if startupTaint.MatchTaint(&node.Spec.Taints[i]) {
```
njtran, line 27: Is it possible the value of a taint would change? This would be out of spec but still have the same taint.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -203,12 +207,15 @@ func (p *Provisioner) launch(ctx context.Context, node *scheduling.Node) error {
 	if err := latest.Spec.Limits.ExceededBy(latest.Status.Resources); err != nil {
 		return err
 	}
+
+	// apply both the taints and startup taints to the node
+	taints := append(node.Provisioner.Spec.Taints, node.Provisioner.Spec.StartupTaints...)
```
njtran, line 212: nit: inline?

pkg/apis/provisioning/v1alpha5/constraints.go
tzneal: Not sure I follow, on line 83 I need to append again and I don't think this is valid syntax:

```go
taints := append(c.Taints, v1.Taint{Key: NotReadyTaintKey, Effect: v1.TaintEffectNoSchedule}, c.StartupTaints...)
```

ellistarn:
> > I'm wary of the long term impact of including startup taints in the api. It's discussed in a number of places, but I'll restate here. Upstream has an unimplemented concept to address this use case (including the node restart case) called the NoAdmit taint effect. This allows scheduling but prevents the kubelet from running the pod, which is exactly what we need. One benefit of doing this upstream is that the change applies to all integrators like CAS, Karpenter, and the Descheduler.
> 
> I don't think that NoAdmit solves this problem. From the docs, NoAdmit is "NoSchedule" + don't start pods that are scheduled. If it were "don't start pods only", maybe it could work but as it, it doesn't fully convey the concept of a temporary taint applied at startup that will be eventually removed. Even if it were "don't start pods", I don't think it still fully implies that the taint is temporary.
> 

Think of these independently:
- NoSchedule: prevent scheduling
- NoExecute: evict any running pods
- NoAdmit: prevent the pod from running

In the initial implementation, there was some concept of combining the taint effects into a single effect, e.g. `NoScheduleNoAdmitNoExecute`. Wisely, upstream never went this direction and instead recommended that users specify both NoSchedule and NoExecute separately if both behaviors are desired. Similarly, I'd expect `NoAdmit` to behave this way, and don't expect that `NoScheduleNoAdmitNoExecute` would be accepted upstream. 

The idea that a taint exists at startup and will eventually be removed is superceded by the more general "don't run if this exists.". `NoAdmit` captures cases of startup and then removed, as well as flip flop (restart) scenarios. If we separate `NoScheduleNoAdmit` from `NoAdmit`, then we can just follow natural scheduling rules and the node will sort everything out at runtime. It's possible that the taint may take hours to be removed (think maintenance windows, or similar mechanisms), and neither karpenter nor the scheduler need to care.  

> > I'd hate to include this in karpenters API as a point in time workaround for something that will eventually be fixed upstream. One option is to support this taint effect in karpenter before it's merged upstream and do the upstream work in parallel.
> 
> The NoAdmit taint was introduced, commented out, in [2016](https://github.com/kubernetes/kubernetes/commit/52fb89ff7326b3570e9c42a435081dab559f610b) . That appears to be the initial implementation of taints and it hasn't been needed yet. Agree that it doesn't seem to be a priority and we would need to do the work, but I also don't think it solves this particular problem.
> 

See above.

> To solve the problem, we would also need to treat 'NoAdmit' as a temporary taint by provisioning capacity for workloads that don't tolerate it and then assume something would remove the taint. We would have implemented the concept of startup taints only hidden behind a magic effect name. If NoAdmit is ever implemented upstream, Karpenter would then be supporting it incorrectly as nothing in the NoAdmit docs indicate that it will always be removed.

Agree that a magical taint that is never implemented upstream could result in an awkward customer experience. I'm not sure whether or not it's worse than a special field to capture this edge case. There are are a couple paths here:

1. We implement NoAdmit, and successfully upstream the concept 🎉 
2. We implement NoAdmit, and upstream does something else -- may require deprecation in Karpenter
3. We implement startupTaints and upstream implements `NoAdmit` -- may require deprecation in Karpenter
4. We implement startupTaints and nothing happens upstream 🥈 .

How hard do you think it would be to get a directional sense from SIG Scheduling on this? My guess is that they would be supportive, since this concept is simple to implement, has precedence, and applies generally to CAS/Karpenter/etc.

tzneal:
> Think of these independently:
> 
>     * NoSchedule: prevent scheduling
> 
>     * NoExecute: evict any running pods
> 
>     * NoAdmit: prevent the pod from running
> 

NoExecute does not just evict running pods though, it also implies NoSchedule.  I would expect NoAdmit to work similarly if it were to be implemented, though it could conceivably behave differently.

> 
> In the initial implementation, there was some concept of combining the taint effects into a single effect, e.g. `NoScheduleNoAdmitNoExecute`. Wisely, upstream never went this direction and instead recommended that users specify both NoSchedule and NoExecute separately if both behaviors are desired. Similarly, I'd expect `NoAdmit` to behave this way, and don't expect that `NoScheduleNoAdmitNoExecute` would be accepted upstream.

This isn't correct, if you put a NoExecute taint on a node it also prevents scheduling any pods unless they tolerate the taint.  If it didn't, the pod would just rapidly get scheduled and evicted immediately. 


> The idea that a taint exists at startup and will eventually be removed is superceded by the more general "don't run if this exists.". `NoAdmit` captures cases of startup and then removed, as well as flip flop (restart) scenarios. If we separate `NoScheduleNoAdmit` from `NoAdmit`, then we can just follow natural scheduling rules and the node will sort everything out at runtime. It's possible that the taint may take hours to be removed (think maintenance windows, or similar mechanisms), and neither karpenter nor the scheduler need to care.
> 

We need awareness of which taints are temporary and even if NoAdmit were to be accepted and implemented upstream, we can't assume that it would always be a temporary taint.  For in-flight nodes we care about the difference between a taint that we place on a node at startup which will exist for a short period of time vs a taint that can exist for hours.  In the latter case, we need to provision a new node.  In the first case, we would not.

The other issue is that for startup taints, we provision capacity even though the pod doesn't tolerate the taint because of the temporary nature of the taint.  This is different from other taints in which we only provision capacity if the pod tolerates the taint.

> Agree that a magical taint that is never implemented upstream could result in an awkward customer experience. I'm not sure whether or not it's worse than a special field to capture this edge case. There are are a couple paths here:
> 
>     1. We implement NoAdmit, and successfully upstream the concept 🎉


Assuming that:
- we can implement NoAdmit and get it merged upstream 
- it doesn't work as described in its initial docs and doesn't work similarly to the existing NoExecute and instead only prevents pods from running

Internally, we would have to add to 'NoAdmit' the concept of provisioning capacity for pods even if they don't tolerate the taint or it doesn't solve the current problem.   For in-flight nodes we have to treat NoAdmit as temporary to prevent us from provisioning excess capacity. If NoAdmit is then supported upstream, customers now can't taint their nodes with "NoAdmit" using Karpenter unless they want us to treat that with our special temporary taint handling. We then need to explain how/why we treat "NoAdmit" differently from every NoSchedule and NoExecute taints and are left in an awkward state.

>     2. We implement NoAdmit, and upstream does something else -- may require deprecation in Karpenter

We still have a special Karpenter only taint effect whose name doesn't clearly describe what it does.

>     3. We implement startupTaints and upstream implements `NoAdmit` -- may require deprecation in Karpenter

See 1, NoAdmit doesn't imply temporary so we still need the concept of a temporary taint.  We need something like startupTaints that lets customers express to Karpenter that a taint is temporary so pods don't need to tolerate it for the purposes of provisioning and so we can recognize it on in-flight nodes and know that the capacity will soon be available.  For the CCM and Cilium case, that capacity will be available faster than we can launch a node since every node goes through the same initialization process.

>     4. We implement startupTaints and nothing happens upstream 🥈 .

> How hard do you think it would be to get a directional sense from SIG Scheduling on this? My guess is that they would be supportive, since this concept is simple to implement, has precedence, and applies generally to CAS/Karpenter/etc.

I could see NoAdmit potentially solving some issues with pre-binding, but not all (e.g. extended resource zero'ing).  I don't think it can solve the 'temporary startup taints' problem at all.

pkg/apis/provisioning/v1alpha5/taints.go
tzneal: I wrote it this way so that if the node taint isn't exactly what we created, it doesn't get the special behavior.  I think that's the safest way to start this.  We place the taint on the node originally, so it will match.  If the taint then changes in some way, I don't think we can safely assume that we should still treat it as a startup taint.

pkg/apis/provisioning/v1alpha5/util.go
tzneal: Same as above.

pkg/controllers/provisioning/provisioner.go
tzneal: I could go either way, but I think it's slightly less clear inlined.

ellistarn:
I'll be offline due to a medical emergency. Don't block on this. 

You make some good points and I've learned a bunch from this discussion. However, I think you're missing the core of what I'm saying. Startup taints is a perfectly reasonable short term (and maybe long term) approach, even if I think we might be able to coordinate a few changes that would collapse the user facing complexity with a more general solution.

pkg/apis/provisioning/v1alpha5/constraints.go
njtran: You're right.

bwagner5 (approved):
lgtm!

PR #1732, opened 2022-04-27
Add support for Shared VPC Architecture

irasnyd:
**1. Issue, if available:**

https://github.com/aws/karpenter/issues/1368

This provides a way to configure the VPC Subnets used by Karpenter for users who have clusters in environments where they do not have access to the VPC Subnet tags.

This can happen in environments which use a [Shared VPC Architecture][1]. To provide a workaround for this type of architecture, the magic keys `subnet-arn` and `subnet-id` are used to indicate that the user is specifying subnets explicitly.

In the implementation, the magic keys indicate that a different type of `ec2.Filter` object should be used in the call to the `ec2.DescribeSubnetsWithContext` method.

[1]: https://docs.aws.amazon.com/vpc/latest/userguide/vpc-sharing.html#vpc-share-limitations

I tested with an AWS EKS Cluster running `Server Version: v1.21.5-eks-bc4871b`

Here is the Provisioner I used:

```yaml
kind: Provisioner
metadata:
  name: default
spec:
  limits:
    resources:
      cpu: "50"
  provider:
    amiFamily: AL2
    instanceProfile: KarpenterNodeInstanceProfile-irasnyd-test
    securityGroupSelector:
      kubernetes.io/cluster/irasnyd-test: "*"
    subnetSelector:
      subnet-id: subnet-0b8acc182f78e5773,subnet-04de6812b38fff67b,subnet-0ea54104b710f7bbc,subnet-0fa94757359cecbf5
    tags:
      Name: karpenter-default
  requirements:
      operator: In
      values:
        - spot
  ttlSecondsAfterEmpty: 120
```

- [ ] Yes, issue opened:
- [ ] No


bwagner5:
Thanks for opening this! We're discussing this one a little bit just to make sure we're making the right decision on expanding the API. I agree that the use-case is valid and we need to solve for this!  

We're currently considering:
 - your approach w/ the magic key strings
 - adding a separate field to the cloud provider spec that takes a list of strings where you could pass arn or id 
 - expanding the selectors to allow for a `tags` section which would match what we have today and then an `ids` section

ellistarn:
One option is to implement the magic using the reserved syntax like "aws:id" see: https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html

This makes it clear that it's syntactic sugar and not something the user can specify while also not introducing a new field just to cover this edge case. 

Whatever we do, we should do the same thing for subnets and security groups. This leads us to something like "id" instead of "subnet-id".

pkg/cloudprovider/aws/subnets.go
```diff
@@ -70,7 +71,13 @@ func getFilters(provider *v1alpha1.AWS) []*ec2.Filter {
 	filters := []*ec2.Filter{}
 	// Filter by subnet
 	for key, value := range provider.SubnetSelector {
```
bwagner5, line 73: can you implement this for `securityGroupSelector` as well?

website/content/en/preview/AWS/provisioning.md
```diff
@@ -70,6 +72,18 @@ Select subnets using wildcards:
 
 ```
 
+Specify subnets explicitly by ARN:
+```yaml
+  subnetSelector:
+    subnet-arn: "arn:aws:ec2:us-west-2:012345678901:subnet/subnet-09fa4a0a8f233a921,arn:aws:ec2:us-west-2:012345678901:subnet/subnet-0471ca205b8a129ae"
```
bwagner5, line 78: let's remove arn support for now since it's not really necessary

bwagner5 (commented):
I've left some comments on changing the key from `subnet-id` to `aws-ids` and adding admission validation since the field is unstructured. 

I'm happy to help out on any of these changes! Just let me know!

bwagner5 (dismissed):
lgtm! Thank you!

bwagner5:
Looks like formatting might be off somewhere, can you run `make verify` and check-in any changes made with that command?

irasnyd:
> Looks like formatting might be off somewhere, can you run `make verify` and check-in any changes made with that command?

Thanks for helping me figure out what went wrong. I was pretty confused. 😄 

I ran `make verify` and it altered the order of imports in `provider_validation.go`. I committed and force pushed to my branch with the fix. It passes `make verify` locally on my workstation now; I think it should pass the CI/CD. Thanks again for the help.

bwagner5 (approved):
lgtm

universam1:
Looking forward to this feature, this is just what we are waiting for!
Thanks @irasnyd @bwagner5 - any release planned?

bwagner5:
Should be next week!

PR #1737, opened 2022-04-28
Split image into repository and tag

dza89:
**1. Issue, if available:**
Follow app image from chart.yaml.
Downside: no more SHA, but you are in control of versioning anyway

value.controller.image and webhook.image get image.repository and image.tag
image.tag gets default chart.appversion

helm install --dryrun

- [ ] No


dza89:
Do i need to bump and package in the MR?

bwagner5:
Curious why it's necessary to break the repo and the version apart? We definitely don't want to lose the SHA.

tuananh:
> 

my guess would be it's convention.

possible use case maybe: in corp context, we have proxy for container registry and we only want to override the repository and keep the tag the same.

bwagner5:
I'm okay with the general change if it's more convenient to override repo this way for the proxy use-case, but we need to keep the SHA and maintain backwards compatibility.

dza89:
Alright, you can just set tag: v0.9.0@sha256:1ce86ff6aa66ba8aa596188966137b13b309f0fbc22663cdc48b01e5f3d2c0fc and it would work, so the functionality is not gone. I dont really think it's necessary as default, as you control both the container registry and helm chart.

However, I've updated it to this:
values.tag and values.digest: image:v0.9.0@sha256:1ce86ff6aa66ba8aa596188966137b13b309f0fbc22663cdc48b01e5f3d2c0fc
only values.digest: image@sha256:1ce86ff6aa66ba8aa596188966137b13b309f0fbc22663cdc48b01e5f3d2c0fc
only values.tag: image:v0.9.0
both empty: image:chart.Version

bwagner5:
Another consideration is that this is a breaking change to the helm chart. Users won't be able to upgrade if they were overriding the image before with the fully qualified `repo:version@digest`

charts/karpenter/README.md
```diff
@@ -42,7 +42,13 @@ helm upgrade --install --namespace karpenter --create-namespace \
 | clusterName | string | `""` | Cluster name. |
 | controller.env | list | `[]` | Additional environment variables for the controller pod. |
 | controller.extraVolumeMounts | list | `[]` | Additional volumeMounts for the controller pod. |
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
dza89: fixed

dza89:
@bwagner5 
I didn't check the history of this chart, but when you use helm (3.8) create chart you get:
```
image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```

Just having only Values.image costs alot of flexibility and is just.. old stuff.

dza89:
@stevehipwell 

Looks like this:
```
{{- define "karpenter.image" -}}
{{- if and .digest .tag }}
{{- printf "%s/%s:%s@%s" .registry .repository .tag .digest -}}
{{- else if .digest }}
{{- printf "%s/%s@%s" .registry .repository .digest -}}
{{- else -}}
{{ printf "%s:%s" .registry .repository (default .Chart.AppVersion .tag) | quote }}
{{- end -}}
{{- end -}}
``` 

But since it's scoped to the controller/webhook.image I can't access Chart.AppVersion anymore in normal ways.
Even more, if you guys are going to use the digest as default, what is the actual use of adding this at all?
So I've created a default and simplified version. If that is not sufficient, you can close this MR.

stevehipwell:
@dza89 you can access the root context with `$` so `$.Chart.AppVersion`. My suggested function would be as below which also supports setting the tag to `-` to skip it which can be useful for automation purposes.

```helm
{{/*
Generate the image name from the context of .Values.controller.image or .Values.webhook.image
*/}}
{{- define "karpenter.image" -}}
{{- if and .digest .tag }}
{{- printf "%s:%s@%s" repository .tag .digest -}}
{{- else if .digest }}
{{- printf "%s@%s" .repository .digest }}
{{- else if and .tag (ne .tag "-") }}
{{- printf "%s" .repository }}
{{- else }}
{{ printf "%s:%s" .repository (default $.Chart.AppVersion .tag) }}
{{- end }}
{{- end }}
```

stevehipwell:
@dza89 good point, this function should work and has been corrected for logic.

```helm
{{/*
Generate the image name from a context with image and appVersion set
*/}}
{{- define "karpenter.image" -}}
{{- if and .image.digest .image.tag (ne .image.tag "-") }}
{{- printf "%s:%s@%s" .image.repository (default .appVersion .image.tag) .image.digest -}}
{{- else if and .image.digest }}
{{- printf "%s@%s" .image.repository .image.digest }}
{{- else if and .image.tag (eq .image.tag "-") }}
{{- printf "%s" .image.repository }}
{{- else }}
{{ printf "%s:%s" .image.repository (default .appVersion .image.tag) }}
{{- end }}
{{- end }}
```

It can be called like this.

```helm
image: {{ include "karpenter.image" (dict "image" .Values.controler.image "appVersion" .Chart.AppVersion) }}
```

dza89:
@stevehipwell  Nice, that's way cleaner than what i had in mind. Thanks alot. 
Implemented

charts/karpenter/values.yaml
```diff
@@ -78,8 +78,12 @@ extraVolumes: []
...
 
 controller:
-  # -- Controller image.
-  image: "public.ecr.aws/karpenter/controller:v0.10.0@sha256:e27cc9fb91f80ed9c5c26202c984a9d7d871ce6008dd4f83f50f3516c9f2ce8e"
+  image:
+    repository: public.ecr.aws/karpenter/controller
+    # -- Defaults to chart.appVersion, skip with "-"
+    tag:
+    # -- sha256:xxx, can be combined with tag
+    digest:
```
stevehipwell, line 86: I suspect that the maintainers will want the digest kept.

charts/karpenter/values.yaml
```diff
@@ -105,8 +109,12 @@ controller:
...
   #   readOnly: true
 webhook:
-  # -- Webhook image.
-  image: "public.ecr.aws/karpenter/webhook:v0.10.0@sha256:9e7ad73c92db5635d43067d4cd0835cbc12b9a02d29bbcde8ed680eeaa55b8f1"
+  image:
+    repository: public.ecr.aws/karpenter/webhook
+    # -- Defaults to chart.appVersion, skip with "-"
+    tag:
+    # -- sha256:xxx, can be combined with tag
+    digest:
```
stevehipwell, line 117: Same as for the controller.

stevehipwell (commented):
@dza89 you probably want to look through the repo for where the Helm chart is used either in docs (preview folder for changes) or automation and change this t o work with your new pattern. You will also need to look at if and how the digest is automatically added to the chart.

charts/karpenter/values.yaml
dza89: If the maintainers want to keep using the digest; there is not much added value in this PR. I think it's better to have a split into repo and tag and that's it. Thoughts?

stevehipwell: I think this PR adds value even if the maintainers want to keep specifying the digest. Personally I think the chart should either have an immutable image with digest or capturing the digest should be a separate concern to the chart.

PR #1773, opened 2022-05-06
don't pre-bind pods to nodes

tzneal:
**1. Issue, if available:**

Will update once it's not a draft

- Eliminate binding of pods to nodes
- Register extended resources as an annotation

- [ ] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/labels.go
+++ b/pkg/apis/provisioning/v1alpha5/labels.go
@@ -34,6 +34,10 @@ var (
 	KarpenterLabelDomain = "karpenter.sh"
 	LabelCapacityType    = KarpenterLabelDomain + "/capacity-type"
 
+	// AnnotationExtendedResources is used to record the expected extended resources on a node that will be created when
+	// device plugins have finished initializing
+	AnnotationExtendedResources = KarpenterLabelDomain + "/extended-resources"
+
 	// RestrictedLabelDomains are either prohibited by the kubelet or reserved by karpenter
 	RestrictedLabelDomains = stringsets.NewString(
 		"kubernetes.io",
--- a/pkg/cloudprovider/aws/instance.go
+++ b/pkg/cloudprovider/aws/instance.go
@@ -29,7 +29,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"go.uber.org/multierr"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/logging"
@@ -262,16 +261,11 @@ func (p *InstanceProvider) instanceToNode(ctx context.Context, instance *ec2.Ins
 				nodeName = aws.StringValue(instance.InstanceId)
 			}
 
+			// Since we no longer pre-bind, we need to ensure that all resources that were possibly considered for scheduling
+			// purposes are placed on the node.  The extended resources will be moved to an annotation, while the
+			// non-extended will be left on the node.
 			resources := v1.ResourceList{}
-			for resourceName, quantity := range map[v1.ResourceName]resource.Quantity{
-				v1.ResourcePods:             instanceType.Resources()[v1.ResourcePods],
-				v1.ResourceCPU:              instanceType.Resources()[v1.ResourceCPU],
-				v1.ResourceMemory:           instanceType.Resources()[v1.ResourceMemory],
-				v1.ResourceEphemeralStorage: instanceType.Resources()[v1.ResourceEphemeralStorage],
-				v1alpha1.ResourceNVIDIAGPU:  instanceType.Resources()[v1alpha1.ResourceNVIDIAGPU],
-				v1alpha1.ResourceAMDGPU:     instanceType.Resources()[v1alpha1.ResourceAMDGPU],
-				v1alpha1.ResourceAWSNeuron:  instanceType.Resources()[v1alpha1.ResourceAWSNeuron],
-			} {
+			for resourceName, quantity := range instanceType.Resources() {
 				if !quantity.IsZero() {
 					resources[resourceName] = quantity
 				}
--- a/pkg/controllers/node/emptiness.go
+++ b/pkg/controllers/node/emptiness.go
@@ -17,6 +17,7 @@ package node
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -33,8 +34,11 @@ import (
 
 // Emptiness is a subreconciler that deletes nodes that are empty after a ttl
 type Emptiness struct {
-	kubeClient client.Client
+	kubeClient     client.Client
+	firstSeenEmpty sync.Map
 }
+
+var initialEmptyDebouncePeriod = 10 * time.Second
 
 // Reconcile reconciles the node
 func (r *Emptiness) Reconcile(ctx context.Context, provisioner *v1alpha5.Provisioner, n *v1.Node) (reconcile.Result, error) {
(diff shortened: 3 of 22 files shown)


pkg/cloudprovider/aws/instance.go
```diff
@@ -262,16 +261,11 @@ func (p *InstanceProvider) instanceToNode(ctx context.Context, instance *ec2.Ins
 				nodeName = aws.StringValue(instance.InstanceId)
 			}
 
+			// Since we no longer pre-bind, we need to ensure that all resources that were possibly considered for scheduling
+			// purposes are placed on the node.  The extended resources will be moved to an annotation, while the
+			// non-extended will be left on the node.
```
tzneal, line 266: I'm not entirely sure how pod-eni ever worked.  We weren't putting the resource on the node, but were just binding the pod. It should have been evicted by the kube-scheduler.

pkg/controllers/node/emptiness.go
```diff
@@ -42,6 +42,14 @@ func (r *Emptiness) Reconcile(ctx context.Context, provisioner *v1alpha5.Provisi
 	if provisioner.Spec.TTLSecondsAfterEmpty == nil {
 		return reconcile.Result{}, nil
 	}
+
+	// If the node just launched, give it some time to fully initialize and for kube-scheduler to assign pods to the
+	// node. This also allows for extended resource device plugins to initialize and assign resource capacity/allocatable
+	// quantities to the node status.
+	if n.CreationTimestamp.After(injectabletime.Now().Add(-5 * time.Minute)) {
```
tzneal, line 49: I don't feel great about this, but I'm not sure of a better solution.  The nvidia driver plugin DS takes some amount of time to start and for the GPU resources to be recorded on the nodes capacity/allocatable.  In my testing this was greater than thirty seconds so the nodes were being deleted before the driver ever finished initializing.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -245,46 +251,68 @@ func (p *Provisioner) launch(ctx context.Context, node *scheduling.Node) error {
...
-			Effect:   v1.TaintEffectNoSchedule,
-		}, {
-			Key:      v1.TaintNodeNotReady,
-			Operator: v1.TolerationOpEqual,
-			Effect:   v1.TaintEffectNoSchedule,
-		}}
-
-	workqueue.ParallelizeUntil(ctx, len(pods), len(pods), func(i int) {
-		pod := pods[i]
-		// Don't bind pods that would immediately get evicted.  We tolerate the two standard taints that are applied for
```
tzneal, line 272: I've removed binding entirely here, but there are other options that are probably worth discussing.

pkg/apis/provisioning/v1alpha5/labels.go
```diff
@@ -34,6 +34,10 @@ var (
 	KarpenterLabelDomain = "karpenter.sh"
 	LabelCapacityType    = KarpenterLabelDomain + "/capacity-type"
 
+	// AnnotationExtendedResources is used to record the expected extended resources on a node that will be created when
```
tzneal, line 37: We need to know about extended resources that should eventually appear on the node so we can schedule against it as an in-flight node (e.g. we need to know there will be 4 NVidia GPUs so we don't launch another GPU instance due to a pending GPU pod).  We can't set the resources on the node as then kube-scheduler races with kubelet and sometimes schedules pods before kubelet zeros resources which leads to an eviction.

pkg/controllers/node/emptiness.go
bwagner5: Could we check that all the extended resource annotations have been added to the node capacity & allocatable as a trigger for when to remove the `karpenter.sh/not-ready` annotation?

pkg/controllers/provisioning/provisioner.go
bwagner5: Did you have other options in mind?  Would knowing what node we think the pod will schedule to help in anyway?

bwagner5 (commented):
nice! haven't played around with it yet, but looking good!

pkg/controllers/provisioning/provisioner.go
tzneal: Other options:
1) Stop pre-binding only for nodes with extended resources (should fix both GPU and SGPP issues)
2) We could bind pods as a last step before removing the `karpenter.sh/not-ready` taint.
3) We could 'nominate a node' for pods as a last step before removing the `karpenter.sh/not-ready` taint.

pkg/controllers/node/emptiness.go
bwagner5: o dang :(   I guess you could still replace the time based check and instead check for the annotated resources to be populated in resources.

spring1843 (approved):
/release

spring1843 (approved):
/snapshot

pkg/controllers/provisioning/provisioner.go
bwagner5: > 1. Stop pre-binding only for nodes with extended resources (should fix both GPU and SGPP issues)

only pre-binding for extended resources would be surprising I would think. 

> 2. We could bind pods as a last step before removing the karpenter.sh/not-ready taint.

This would give us more guarantees if there were a lot of pods coming in that had various resource requests that would throw off proper bin-packing. This would be less of a change too. 

> 3. We could 'nominate a node' for pods as a last step before removing the karpenter.sh/not-ready taint.

Would this really give us anything? Wouldn't opt 2 be the same thing but a stronger form.



It seems to me that option 2 is the most sensible at this point. We fix the issues with extended-resources, Karpenter behaves in roughly the same way of ultimately scheduling the pod reducing the chance of fragmentation from an influx of pods that karpenter is unaware of,  and works the same for all pods. We could always completely remove binding later if we found a "late pre-binding" not needed.

spring1843 (approved):
/snapshot

spring1843 (approved):
ignore this message

spring1843 (approved):
/snapshot

pkg/controllers/provisioning/provisioner.go
tzneal: > > 2. We could bind pods as a last step before removing the karpenter.sh/not-ready taint.
> 
> This would give us more guarantees if there were a lot of pods coming in that had various resource requests that would throw off proper bin-packing. This would be less of a change too.
> 
> > 3. We could 'nominate a node' for pods as a last step before removing the karpenter.sh/not-ready taint.
> 
> Would this really give us anything? Wouldn't opt 2 be the same thing but a stronger form.
> 
> It seems to me that option 2 is the most sensible at this point. We fix the issues with extended-resources, Karpenter behaves in roughly the same way of ultimately scheduling the pod reducing the chance of fragmentation from an influx of pods that karpenter is unaware of, and works the same for all pods. We could always completely remove binding later if we found a "late pre-binding" not needed.

Yeah, 2 and 3 were similar.  After further thought, they may both be broken concepts though.  Device plugin daemonsets which create the extended resources may or may not (e.g. nvidia won't) tolerate our taint.  If they don't, then they won't bind and won't add the extended resources.  

Maybe by the time the node is ready, Kubelet won't zero out the resources so we can optimistically bind the pod and then remove our taint which would allow the DS to schedule.  However, device plugins potentially could zero out resources then which would cause pods to get evicted.

pkg/controllers/node/emptiness.go
tzneal: The one positive side effect of the time based change is that we don't see spurious empty TTL annotations getting added and removed which happens every time as we notice the empty node after it's ready before kube-scheduler binds pods to it.

tzneal: There are two time based checks that have been added in this PR, one is used for emptiness and the other for computing allocatable resources.

They are both working around the same fundamental issue, a node becomes "Ready" and then it takes some amount of time for device plugins to initialize and register the resources on the node so that pods can be scheduled to it.

The emptiness one is needed as if the node starts up and we call it "empty" as soon as it's ready, it may remain empty and be terminated before the device plugin can finish initializing.

The other is used when determining the remaining allocatable resources of an in-flight node. 

One alternative is to increase the time to 15 minutes, but only use it on nodes with extended resources?

ellistarn: Could we use karpenter's readiness taint as a signal and only remove it once all daemonsets are passing readiness probes?

tzneal: The Nvidia plugin daemonset doesn't tolerate our readiness taint, so it doesn't get scheduled until that taint is removed.

ellistarn: I'm really wary of time based solutions, since this varies broadly across vendors and is one of the key pain points of cluster autoscaler that karpenter has managed to avoid so far. I'm open minded to an initial impl that uses timeouts, but I want to find a long term path that fundamentally solves the pod coordination problem.

tzneal: I don't like it as well, but I haven't come up with a better solution yet. Maybe eliminating the time based readiness gate and replacing it with an 'all expected daemonsets are ready' gate would work?  We would need to ensure that all of the extended resource daemonsets that we support don't report 'ready' until they've been able to register the resources on the node:
- aws-vpc-cni
- Nvidia GPU
- AMD GPU
- AWS Neuron

tzneal:
The 5 minute timeouts are now gone along with the startup taint as it no longer served a purpose.  This lets us consider a node being ready when all of the following are true:
1) Kubelet marks it ready
2) All startup taints are removed
3) All extended resources are initialized

If this fails for 15 minutes, we terminate the node.

bwagner5 (approved):
lgtm!

bwagner5 (approved):
Lgtm

spring1843 (approved):
/snapshot

spring1843 (approved):
/snapshot

spring1843 (approved):
/snapshot

spring1843 (approved):
/snapshot

spring1843 (approved):
snapshot

bwagner5 (approved):
lgtm

PR #1787, opened 2022-05-10
Remove duplicate in role helm template

aquarhead:
**1. Issue, if available:**

Saw a duplication in rendered output from helm, looks to me it should be safe to remove.

Not tested yet, but I'm in the process of deploying `karpenter` in one of our cluster for testing. Also, I don't use `helm` so kindly inform me if there is something I can use or run to test it locally.

- [x] No


bwagner5 (approved):
lgtm! thanks!

Issue #1791, opened 2022-05-10
Allow user-specified instance types to pass through filter


thepinetree:
**Tell us about your request**
The filter referenced in #1053 continues to be restrictive (e.g. the `i4` and `z1` instances classes are disallowed).

https://github.com/aws/karpenter/blob/ec2b5c021ca8baa2bebae36fac4f6c5969d2bfa3/pkg/cloudprovider/aws/instancetypes.go#L172

Ideally, as @bwagner5 mentioned in a comment on the aforementioned issue, it would be nice to allow instances specified by the user with `node.kubernetes.io/instance-type` to pass through the filter. 

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
We are trying to schedule jobs on a variety of instance types with Karpenter, in this situation, one-off testing runs.

Unfortunately, the filter prevents Karpenter from provisioning certain node types (`i4`, `z1`) from being provisioned.

**Are you currently working around this issue?**
Not currently working around this, though an alternative certainly would be to add an instance type class to the filter whenever needed.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

bwagner5:
Yes! Thanks for the issue! I just put up a PR to at least allow the use of any instance type that you directly specify in the provisioner spec.  I also set `i*` in the filter rather than just allowing `i3` before. I didn't include `z*` in the standard set, but you can set the provisioner to use `z1d` instance types which it will only use if no other instance types are available. There's probably still work to do in Karpenter to not prioritize when instance type requirements are passed into the provisioner at all, but hopefully this helps in the near term.

bwagner5:
ah! I didn't even see the PR you submitted, apologies. I haven't taken a look at it yet, but if yours solves this problem as well, then I'd be happy to take yours.  I can take a closer look tomorrow.

thepinetree:
ack, thanks for the quick turnaround @bwagner5! Feel free to use whichever PR works out better. 

[...]

PR #1792, opened 2022-05-10
[AWS] Allow instance type specifications to pass through filter

thepinetree:
**1. Issue, if available:** #1791 

**2. Description of changes:** Bypass allow-list filter if instance types are specified by user in `node.kubernetes.io/instance-type`

**3. How was this change tested?** TODO: add unittest

- [x] No

### DIFF
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -35,9 +35,10 @@ import (
 	"github.com/aws/karpenter/pkg/utils/functional"
 	"github.com/aws/karpenter/pkg/utils/injection"
 	"github.com/aws/karpenter/pkg/utils/project"
+	"github.com/aws/karpenter/pkg/utils/sets"
 
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/util/sets"
+	utilsets "k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/transport"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/logging"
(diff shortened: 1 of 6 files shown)


thepinetree:
I'd like to add some unittests testing the new codepath (say, try to schedule a `z1` instance 😃 ). Could someone help point me to how I could best simulate this?

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -110,8 +111,8 @@ func (c *CloudProvider) Create(ctx context.Context, nodeRequest *cloudprovider.N
 }
 
 // GetInstanceTypes returns all available InstanceTypes
-func (c *CloudProvider) GetInstanceTypes(ctx context.Context) ([]cloudprovider.InstanceType, error) {
-	return c.instanceTypeProvider.Get(ctx)
+func (c *CloudProvider) GetInstanceTypes(ctx context.Context, requestedInstanceTypes sets.Set) ([]cloudprovider.InstanceType, error) {
```
bwagner5, line 114: This will break with multiple provisioners. The instanceTypeProvider caches the instance types, so this new param won't take affect if the instance types are cached already or if the cache is empty, the `requestedInstanceTypes` set would cause another provisioner to pickup new instance types that they may not be flexible to.

bwagner5, line 114: This will break with multiple provisioners. The instanceTypeProvider caches the instance types, so this new param won't take affect if the instance types are cached already or if the cache is empty, the `requestedInstanceTypes` set would cause another provisioner to pickup new instance types that they may not be flexible to.

dewjam:
It looks like there hasn't been an update on this PR for quite some time.  And it also looks like #1793 might also solve this problem.  Can you confirm whether this can be closed out?

thepinetree:
I think #1791 requested a fix to this PR that would enable some optimization with the EC2 fleet API on top of #1793. Some higher priority things are taking precedence for a little while but I do plan to revisit and fix this PR.

dewjam:
Gotcha, in that case we can leave it open @thepinetree .  Thanks for your contributions!

dewjam:
Hey @thepinetree .  I'm going to go ahead and close this PR given there hasn't been any updates for over a month.  Go ahead and reopen when you get time to proceed with the change.

Thanks!

PR #1808, opened 2022-05-13
Add labelSelector to topologySpreadConstraints

vadimgusev-codefresh:
**1. Issue, if available:**

Pod Topology Spread Constraints spec [requires](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/#api) self-targeting label selector. Since Deployment's `spec.template.metadata.labels` depends on Helm release name I've reused `karpenter.selectorLabels` Helm template in Deployment's template, rather than adding selector to `values.yaml`.

Deployed modified chart to local k3d cluster and validated that pod will indeed not be scheduled with `whenUnsatisfiable: DoNotSchedule` option:

1. Create 2 node cluster (both having 16 cores, matching my host PC)
2. Create a 15 core Pod to hoard one node completely (more than 1.1 core requests of Karpenter pod)
3. Apply (`helm template . | kubectl apply -f -`) Helm chart with following values different from current defaults:
```yaml
replicas: 2
# RECUIRED SO KARPENTER POD DOESNT PREEMT TEST POD
# priorityClassName: system-cluster-critical
topologySpreadConstraints:
  - maxSkew: 1
    topologyKey: kubernetes.io/hostname
    whenUnsatisfiable: DoNotSchedule
```
4. Observe that one of two Karpenter Pods is stuck in `Pending` with `0/2 nodes are available: 1 Insufficient cpu, 1 node(s) didn't match pod topology spread constraints.`
5. Rollback my changes to Deployment template, apply again
6. Observe two Karpenter Pods landing successfully.

Worth noting that `topologySpreadConstraints` without `labelSelector` has effect when `topologyKey` cannot be met (like `topology.kubernetes.io/zone` is absent on local clusters) and `whenUnsatisfiable` is set to `DoNotSchedule`. The error would be `1 node(s) didn't match pod topology spread constraints (missing required label)`. It kinda short-circuits on node selection, not even getting to Pod selection part. This might give false impression of working in it's entirety without `labelSelector`.

- [x] No

### DIFF
--- a/charts/karpenter/templates/deployment.yaml
+++ b/charts/karpenter/templates/deployment.yaml
@@ -164,5 +164,11 @@ spec:
       {{- end }}
       {{- if .Values.topologySpreadConstraints }}
       topologySpreadConstraints:
-        {{ toYaml .Values.topologySpreadConstraints | nindent 8 }}
+        {{- range $constraint := .Values.topologySpreadConstraints }}
+        -
+        {{- toYaml $constraint | nindent 10 }}
+          labelSelector:
+            matchLabels:
+            {{- include "karpenter.selectorLabels" $ | nindent 14 -}}
+        {{- end }}
       {{- end }}


tzneal (approved):
Tested locally and this works for me, thanks for the contribution!

PR #1816, opened 2022-05-17
Add snapshot PR integration back

spring1843:
**1. Issue, if available:**

Fixed an auth issue resulting from credentials not being set when `id-token: write` is not explicitly defined.

- [x] No

### DIFF
--- a/.github/workflows/snapshot-pr-approval.yaml
+++ b/.github/workflows/snapshot-pr-approval.yaml
@@ -1,7 +1,5 @@
 name: SnapshotsONPRs
 on:
-  push:
-    branches: [ main ]
   pull_request_review:
     types: [ submitted ]
 permissions:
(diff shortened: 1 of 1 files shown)


bwagner5 (approved):
lgtm

bwagner5 (commented):
/snapshot

bwagner5 (approved):
/snapshot

bwagner5:
seems like this requires the approval message to be `/snapshot` not just a comment on a PR that is already approved.  Not a huge deal, but something to note

.github/workflows/snapshot-pr-approval.yaml
```diff
@@ -1,21 +1,46 @@
 name: SnapshotsONPRs
 on:
-  push:
-    branches: [ main ]
   pull_request_review:
```
bwagner5, line 3: maybe triggering on issue_comment too?  The `if` below should short-circuit if the PR is not approved.

spring1843: I think that would become a different Github event i.e. github.event.review.state would not be there anymore, but I'll give it a try

PR #1818, opened 2022-05-17
add an instanceTypeFilter to the provisioner spec

tzneal:
**1. Issue, if available:**

Adds an instance type filter.

Unit testing & deployed on EKS.

- [ ] No


spring1843 (dismissed):
ignore this message

charts/karpenter/crds/karpenter.sh_provisioners.yaml
```diff
@@ -41,6 +41,68 @@ spec:
...
+                  memoryMiB:
+                    description: MemoryMiB allows filtering instance types by the
+                      min and maximum MiB of memory on the instance type with the
+                      minimum and maximum values being inclusive.
+                    properties:
+                      max:
+                        description: Max is the minimum value for the filter
+                        format: int64
+                        minimum: 0
+                        type: integer
```
bwagner5, line 73: why aren't these resource quantitys?

charts/karpenter/crds/karpenter.sh_provisioners.yaml
```diff
@@ -41,6 +41,68 @@ spec:
...
+                description: InstanceTypeFilter allows filtering the instance types
+                  provided by the cloud provider.
+                properties:
+                  cpuCount:
+                    description: CPUCount allows filtering instance types by the min
+                      and maximum number of CPUs on the instance type with the minimum
+                      and maximum values being inclusive.
+                    properties:
+                      max:
+                        description: Max is the minimum value for the filter
```
bwagner5, line 54:
```suggestion
                        description: Max is the maximum value for the filter
```

charts/karpenter/crds/karpenter.sh_provisioners.yaml
```diff
@@ -41,6 +41,68 @@ spec:
...
+                        minimum: 0
+                        type: integer
+                    type: object
+                  memoryMiB:
+                    description: MemoryMiB allows filtering instance types by the
+                      min and maximum MiB of memory on the instance type with the
+                      minimum and maximum values being inclusive.
+                    properties:
+                      max:
+                        description: Max is the minimum value for the filter
```
bwagner5, line 70:
```suggestion
                        description: Max is the maximum value for the filter
```

pkg/apis/provisioning/v1alpha5/provisioner.go
```diff
@@ -43,6 +43,35 @@ type ProvisionerSpec struct {
...
+	// CPUCount allows filtering instance types by the min and maximum number of CPUs on the instance type with the
+	// minimum and maximum values being inclusive.
+	CPUCount *MinMax `json:"cpuCount,omitempty"`
+	// MemoryMiB allows filtering instance types by the min and maximum MiB of memory on the instance type with the
+	// minimum and maximum values being inclusive.
+	MemoryMiB *MinMax `json:"memoryMiB,omitempty"`
+	// MemoryMiBPerCPU allows filtering instance types by the min and maximum MiB of memory per on the instance type
+	// with the minimum and maximum values being inclusive.
+	MemoryMiBPerCPU *MinMax `json:"memoryMiBPerCPU,omitempty"`
+	// NameMatchExpressions are regular expressions to match against instance types names.  If no expressions are
```
bwagner5, line 61: I like this! I was about to ask how we'd handle special CP specific instance type selection, but giving this at the top-level basically takes care of it. nice!

bwagner5 (commented):
overall looks nice!

charts/karpenter/crds/karpenter.sh_provisioners.yaml
tzneal: That makes sense, I was modeling after the ABS but what if we dropped CPU/Memory altogether and made it generic with just min & max resource lists?

Something like:
```yaml
instanceTypeFilter:
  minResources:
    cpu: 2
    memory: 8Gib
  maxResources:
    cpu: 8
    nvidia.com/gpu: 0
```

bwagner5: hmm that is kind of nice... It seems like it's getting a little into the requirements territory though. Wouldn't you want to set a `DoesNotExist` (or `LT`) on `nvidia.com/gpu` in requirements to filter more generically?  Requirements could also support special capabilities like networking or NVMe via labels that the CP adds to nodes.

tzneal: That was probably a bad example. If you wanted only instance types with 2 or more GPUs, you could filter generically using resources.  We can't support that currently unless you list instance types.  I'm not 100% sold on it, but the thought of going to quantities made me think of just generic resource list min/max, which eliminates the need to revisit this if/when someone wants to filter based on some other resource type.

bwagner5: I like the semantic of it. It's growing on me :D

ellistarn:
Can we hold on this until v1alpha6?
/hold

tzneal:
> I don't see draft or a description, so not sure what this is, but can we hold on this until v1alpha6?
> 
> /hold

I just sent you a doc link in slack. I'd rather not wait till v1alpha6, I think there's a large need for this and it's backwards compatible with existing provisioners. Doing this now opens up the possibility of revisiting it in v1alpha6.

vadimgusev-codefresh:
This is indeed highly anticipated, since right now we implement this via long lists of instance types generated by `ec2-instance-selector` and leaving CLI command with exact arguments as a comment.

tzneal:
> This is indeed highly anticipated, since right now we implement this via long lists of instance types generated by `ec2-instance-selector` and leaving CLI command with exact arguments as a comment.

Thanks for the feedback @vadimgusev-codefresh.  The PR is still a WIP as we're reviewing some design choices.

Can you provide some more information on your use cases?
- Do you filter by min/max CPU, memory or both? 
- Is the memory to CPU ratio important

Were considering some other options in addition to, or instead of this PR such as allowing a requirement at the provider and node selectors/affinities at the pod level that filter select specific families and sizes:

```yaml
  requirements:
    - key: "karpenter.sh/instance-family"
      operator: In
      values: ["m5", "r5"]
    - key: "karpenter.sh/instance-size"
      operator: In
      values: ["lage", "xlarge"]
```

Does this mechanism cover your use cases, or some subset of them?

vadimgusev-codefresh:
@tzneal we are still at the early stages of Karpenter adoption and still defining for ourselves node requirements (kinda a mindset shift from single instance type node groups of CAS).
But as of now we have at least 2 requirements, that have to be expressed by explicit lists:
```yaml
- key: "node.kubernetes.io/instance-type"
    # exclude ELB-problematic instances (https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#registered-targets)
    # ec2-instance-selector --max-results 100 --region us-east-1 --service eks -a amd64 --gpus 0 --allow-list '(c1|cc1|cc2|cg1|cg2|cr1|g1|g2|hi1|hs1|m1|m2|m3|t1)\.*'
    operator: NotIn
    values:
    - ...
  - key: "node.kubernetes.io/instance-type"
    # exclude smaller instances
    # ec2-instance-selector --max-results 100 --region us-east-1 --service eks -a amd64 --gpus 0 --allow-list '.+\.(micro|small|medium|large)'
    operator: NotIn
    values:
    - ...
```
I'm dropping actual values for brevity. So those are essentially filters on families and mininum CPUs.

On the horizon I see additional requirement of using Compute-Optimized instance families, since our workloads are CPU bound an single `c6i` core is better than `t3` or even `m6i` one.

On the case of `CPU:Mem` ratio - I see that clusters tend to take a shape of aggregate workload requests, e.g. In cluster with average profile of `1C:2Gi` Karpenter tends to converge to `C`-type instances rather than `M`-type naturally.

tzneal:
> ```
>   requirements:
>     - key: "karpenter.sh/instance-family"
>       operator: In
>       values: ["c5", "c5a", "c6a", "c6i"]
>   requirements:
>     - key: "karpenter.sh/instance-size"
>       operator: NotIn
>       values: ["nano", "micro", "small", "large"]
> ```


The part where it seems less flexible is that you can't use just instance-size to get a roughly equivalent sized nodes. In some cases, it doesn't mean much at all (e.g. an xlarge has from 2 to 8 CPUs (vast majority have 4) with memory from 7.0 GB to 128.0 GB (8.0/16.0/32.0 GB seem evenly split with the others being outliers)).  I think that instance-size really only makes sense in conjunction with an instance-family and in particular similar instance families (e.g. c5/c6).

ellistarn:
> an xlarge has from 2 to 8 CPUs (vast majority have 4) with memory from 7.0 GB to 128.0 GB (8.0/16.0/32.0 GB seem evenly split with the others being outliers)). 

Which types have 2 or 8? Agree that only working on CPU is an unfortunate limitation -- though every time I see customers talk about this feature, it's in terms of "size" and "family".
```
 cat it.json | jq -r ".InstanceTypes[] | select( .VCpuInfo.DefaultVCpus==4 ) | .InstanceType"                                                                                                           ✔ │ at 11:17:34
r6i.xlarge
m6gd.xlarge
x2iedn.xlarge
r6gd.xlarge
m5d.xlarge
c6gn.xlarge
is4gen.xlarge
r5a.xlarge
c7g.xlarge
inf1.xlarge
m5.xlarge
z1d.xlarge
a1.xlarge
t3.xlarge
c6g.xlarge
im4gn.xlarge
c5a.xlarge
m5dn.xlarge
c5ad.xlarge
m2.2xlarge
m4.xlarge
c5.xlarge
m6g.xlarge
c6gd.xlarge
x2gd.xlarge
r5d.xlarge
d3.xlarge
r5dn.xlarge
i2.xlarge
r5b.xlarge
c3.xlarge
m1.xlarge
c6a.xlarge
m3.xlarge
r5n.xlarge
m5n.xlarge
m5zn.xlarge
i3.xlarge
c5d.xlarge
m6i.xlarge
r5.xlarge
r6g.xlarge
g4dn.xlarge
t4g.xlarge
m5ad.xlarge
r4.xlarge
c4.xlarge
g5.xlarge
t2.xlarge
r5ad.xlarge
r3.xlarge
x1e.xlarge
d3en.xlarge
i4i.xlarge
p2.xlarge
c6i.xlarge
d2.xlarge
m6a.xlarge
c5n.xlarge
m5a.xlarge
g3s.xlarge
g4ad.xlarge
i3en.xlarge
t3a.xlarge
g5g.xlarge
```

tzneal:
> > an xlarge has from 2 to 8 CPUs (vast majority have 4) with memory from 7.0 GB to 128.0 GB (8.0/16.0/32.0 GB seem evenly split with the others being outliers)).
> 
> Which types have 2 or 8? Agree that only working on CPU is an unfortunate limitation -- though every time I see customers talk about this feature, it's in terms of "size" and "family".


`m2.xlarge` has 2 CPUs and `c1.xlarge` has 8 CPUs, super-old but they make things weirder so that "xlarge" means 4 CPUs normally, except for a few cases, and probably  8-32 GB of memory...usually 😆

vadimgusev-codefresh:
@ellistarn 
>   requirements:
    - key: "karpenter.sh/instance-family"
      operator: In
      values: ["c5", "c5a", "c6a", "c6i"]
  requirements:
    - key: "karpenter.sh/instance-size"
      operator: NotIn
      values: ["nano", "micro", "small", "large"]
This approach would include support for the pod level, as well as reuse our existing "well known labels + layered constraints" paradigm.

On one hand it looks neat, but at the same time it looks like some sort anti-pattern that pod can request particular node types explicitly. It's too abstraction-breaching in my opinion, and conflicts with resource requests. I mean, how much sense this pod spec makes:
```yaml
resources:
  requests:
    cpu: "100m"
    memory: "128Mi"
  limits:
    cpu: "100m"
    memory: "256Mi"
affinity:
  nodeAffinity:
      nodeSelectorTerms:
        - matchExpressions:
          - key: "karpenter.sh/instance-size"
            operator: "In"
            values: ["c6i.24xlarge"]
```

ellistarn:
>but at the same time it looks like some sort anti-pattern that pod can request particular node types explicitly. It's too abstraction-breaching in my opinion,

Selecting nodes based on an instance type label selector is common for specialized workloads (e.g ML compiled for a particular GPU type) and is used broadly outside of karpenter.

vadimgusev-codefresh:
@ellistarn if it's a norm in some environments - so be it. In general I would wish for a single unified interface for node properties selection, but I see two major shortcomings of having only `requirements`:
1. It opens doors for contradiction I've presented in my previous comment. Traditionally, at least in my opinion, `requests` are responsible for **quantity** of the resources node should have (e.g. `give me 10 cores`), while `affinity` is responsible for **qualities** of the nodes (e.g. `it's ok if it's Spot`). In my example I show two concepts trying to control **quantity**, creating contradiction.
2. Limitation of `operator`, not allowing for ranges (`1-4 cores`) or regexes (`^c[5-6].*` for modern compute-optimized families)

ellistarn:
> It opens doors for contradiction I've presented in my previous comment. Traditionally, at least in my opinion, requests are responsible for quantity of the resources node should have (e.g. give me 10 cores), while affinity is responsible for qualities of the nodes (e.g. it's ok if it's Spot). In my example I show two concepts trying to control quantity, creating contradiction.

I don't quite see it this way. Quantity and quality are deeply intermingled. I think of it as a universe of instance types with different properties (including name, family, etc) and karpenter is solving for a set of constraints. As you restrict qualities of the node, you end up with different quantities. Sometimes there is no solution. For example, you may want a burstable workload with 16 cores, but it's simply not on the menu. 

> Limitation of operator, not allowing for ranges (1-4 cores) or regexes (^c[5-6].* for modern compute-optimized families)

I think of regexes as an orthogonal question. We can choose to implement regex support in `provisioner.spec.requirements`. 

However, I'm not a fan of regexes. As far as I'm aware, they're not used much in the k8s ecosystem due to their unpredictability. When I write regexes in my code, I make sure to test them extensively since it's easy to miss edge cases. With instance type selection, this is potentially dangerous. Imagine a `m5.*` selector that unintentionally matches `m5a.*`. The only way to test that what you've configured is acceptable would be to use some client side tool to validate the configuration. For broad regexes, this could change over time in surprising ways as AWS releases more instance types.

I share a similar criticism of the proposed instanceTypeFilter, where the situation is ripe for surprising outcomes. The filter is abstract enough that users are going to need some tool to make sure the instance types that they're getting are what they expect.

vadimgusev-codefresh:
Yeah, I agree that regex is a total can of worms. I just don't see much better solution here: the alternative to regex is either explicit node list or some other wildcard mechanism that users will have to learn and possibly make even more mistakes in a process.

The only alternative to regex I can think of is [shell globbing](https://tldp.org/LDP/abs/html/globbingref.html). It's relatively familiar to wide auditory and far less explosive. Basically what you can `ls` in a directory - you can also select in the instance catalog.

PR #1822, opened 2022-05-18
Adds AWSNodeTemplate CRD for AWS specific config

suket22:
**1. Issue, if available:**
N/A

As follow-up from https://github.com/aws/karpenter/pull/1720 this attempts to create a dedicated CRD for the AWSProvider where we will house fields like UserData, ImageId and so on. 

Installed the CRD and verified that the webhooks were working as well. 

- [x] No


pkg/apis/awsprovider/v1alpha1/awsprovider.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// AWSProviderSpec is the top level specification for the AWS Karpenter Provider.
+// This will contain configuration necessary to launch instances in AWS.
+type AWSProviderSpec struct {
+	// UserData to be applied to the provisioned nodes.
+	// It must be in the appropriate format based on the AMIFamily in use. Karpenter will merge certain fields into
+	// this UserData to ensure nodes are being provisioned with the correct configuration.
+	// +optional
+	UserData *string `json:"userData,omitempty"`
```
suket22, line 28: I was tempted to inline `v1alpha1.AWS` in here from `pkg/cloudprovider/aws/apis/v1alpha1` but I decided not since then we'd end up complicating a lot of our logic trying to read values from either of the two places and uniting the defaulting / validating webhook logic.

 I figured it's easier to always read `UserData` from this CRD and all the other configuration values remain within the Provisioner CRD until we do a clean switch as part of an API upgrade.

pkg/apis/awsprovider/v1alpha1/awsprovider_validation.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func (a *AWSProvider) Validate(ctx context.Context) (errs *apis.FieldError) {
+	return errs.Also(
+		apis.ValidateObjectMetadata(a).ViaField("metadata"),
+		a.Spec.validate(ctx).ViaField("spec"),
+	)
+}
+
+func (a *AWSProviderSpec) validate(ctx context.Context) (errs *apis.FieldError) {
```
suket22, line 30: There's not a lot useful you can do here since the rest of the AWS Provider spec is in a different CRD. (I can't validate my UserData based off AMIFamily because the latter lives elsewhere).

spring1843 (approved):
/snapshot

pkg/apis/awsprovider/v1alpha1/awsprovider.go
ellistarn: +1!

pkg/apis/awsprovider/v1alpha1/awsprovider.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// AWSProvider is the Schema for the AWSProviders API
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=awsproviders,scope=Cluster,categories=karpenter
+type AWSProvider struct {
```
ellistarn, line 34: Naming is hard. `kubectl get awsprovider` feels wrong to me, since it's not clearly scoped to anything about nodes. You could easily imagine a completely different CRD called AWSProvider for a different domain. Musing more on this, what do you think about calling it a `launchTemplate`? I guess this could conflict with an ACK `launchTemplate` CRD, but it's a better conceptual match.

bwagner5: I'd lean towards `NodeConfig` or `AWSNodeConfig` or `AWSProvisionerConfig`, since a lot of the AWS specific instance config terminology is overloaded and confusing. LaunchTemplate or LaunchConfiguration are actual things which do not map to what our parameters are in the AWS CP.

ellistarn: I like this. A couple other thoughts:
- AWSNodeConfiguration (fully spell it out, pretend configmaps are fully spelled :P )
- AWSNodeTemplate (this one has nice parity where NodeTemplate + AWSNodeTemplate make up a node)

suket22: Hmm I chose Config because it's lesser to type :P 
With that I'd prefer `AWSNodeTemplate` than `AWSNodeConfiguration`. Agreed though that the work template might be better suited here. I'll make the change.

ellistarn (approved):

PR #1825, opened 2022-05-18
apply pod induced labels to nodes

tzneal:
**1. Issue, if available:**
N/A

With pre-binding, this wasn't strictly necessary as we bound
the pod regardless. Without pre-binding we need to label the
nodes to allow kube-scheduler to bind the pods and for us to
know the pods will bind.

Unit testing & deployed to EKS.

- [X] No

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/constraints.go
+++ b/pkg/apis/provisioning/v1alpha5/constraints.go
@@ -58,13 +58,17 @@ func (c *Constraints) ToNode() *v1.Node {
 		labels[key] = value
 	}
 	for key := range c.Requirements.Keys() {
-		if !IsRestrictedNodeLabel(key) {
-			switch c.Requirements.Get(key).Type() {
-			case v1.NodeSelectorOpIn:
+		if IsRestrictedNodeLabel(key) {
+			continue
+		}
+		switch c.Requirements.Get(key).Type() {
+		case v1.NodeSelectorOpIn:
+			// something restricted us down to a single label
+			if c.Requirements.Get(key).Values().Len() > 1 {
 				labels[key] = c.Requirements.Get(key).Values().UnsortedList()[0]
-			case v1.NodeSelectorOpExists:
-				labels[key] = rand.String(10)
 			}
+		case v1.NodeSelectorOpExists:
+			labels[key] = rand.String(10)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


spring1843 (dismissed):
/snapshot

tzneal:
Converting to a draft as we look at the root cause of this (creating node objects)

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -168,6 +168,29 @@ var _ = Describe("Custom Constraints", func() {
...
+						{
+							Key:      v1.LabelOSStable,
+							Operator: v1.NodeSelectorOpIn,
+							Values:   []string{"linux", "windows"},
+						},
+					}}))[0]
+			node := ExpectScheduled(ctx, env.Client, pod)
+			value := node.Labels[v1.LabelOSStable]
+			// Ugh, this doesn't really seem correct, but not sure at the moment about it
+			Expect(value == "linux" || value == "windows").To(BeTrue())
```
ellistarn, line 192: To(Or(BeEqual("linux"), BeEqual("windows"))

pkg/apis/provisioning/v1alpha5/constraints.go
```diff
@@ -58,13 +58,17 @@ func (c *Constraints) ToNode() *v1.Node {
 		labels[key] = value
 	}
 	for key := range c.Requirements.Keys() {
-		if !IsRestrictedNodeLabel(key) {
-			switch c.Requirements.Get(key).Type() {
-			case v1.NodeSelectorOpIn:
+		if IsRestrictedNodeLabel(key) {
+			continue
+		}
+		switch c.Requirements.Get(key).Type() {
```
ellistarn, line 64: Thoughts on only applying this logic to custom labels? Well known labels should be determined by the cloud provider, because only it can know specifically which node it launched given the zone/instance type flexibility.

tzneal: The problem is that us creating node objects breaks that. If we create a node and don't put an OS label on it, we don't know that a pod with a node selector for OS can schedule to it. We then launch extra nodes.

ellistarn: I meant here: https://github.com/aws/karpenter/blob/main/pkg/cloudprovider/aws/instance.go#L283. I *think* we can know OS from the AMI type. https://aws.amazon.com/premiumsupport/knowledge-center/ec2-linux-find-os-platform-or-version/

tzneal: Yeah, that's better, I'm going to delete this and create similar tests in the "no node creation/pod binding" branch

tzneal:
Reconsidering node creation.

Issue #1795, opened 2022-05-11
Karpenter fails to provision any new nodes when a PVC is missing


centr0nix:
## Version

**Karpenter**: v0.9.1

**Kubernetes**: v1.21

## Expected Behavior
Karpenter should provision new nodes as expected. 

## Actual Behavior
If a PVC is missing/deleted from a deployment Karpenter fails to provision any new nodes from any of the 
applied provisioners.
## Steps to Reproduce the Problem
Delete/Remove a PVC from a deployment.

## Resource Specs and Logs
The Karpenter controller logs show the following error message:
`Provisioning failed, getting volume topology requirements, getting persistent volume claim my-vol, PersistentVolumeClaim “my-“vol not found	{"commit": "a7e26e6"}`

dewjam:
Hey @centr0nix ,
Thanks for reporting this issue.  I'm going to work on reproducing this problem today.  I'll give an update once I have more info.

dewjam:
Hello again.  I believe I was able to reproduce this.  

To clarify the issue, if a PVC is missing/deleted for a single deployment or pod, Karpenter will fail to provision nodes for all future pods, even those which aren't dependent on the missing PVC.

centr0nix:
> Hello again. I believe I was able to reproduce this.
> 
> To clarify the issue, if a PVC is missing/deleted for a single deployment or pod, Karpenter will fail to provision nodes for all future pods, even those which aren't dependent on the missing PVC.

Hello! That is correct.

dewjam:
OK great.  This certainly is not intended.  I'll be working towards a fix for this shortly.

PR #1831, opened 2022-05-19
Fix: validate pod pvc before including in batch

dewjam:
**1. Issue, if available:**
#1795

This fixes an issue where no pods can be batched if a single pod references a PVC which does not exist.  Additionally, ensures pod affinity and node selector terms are validated when including pods into a batch.

Added a test with a combination of schedulable pods and pods referencing "invalid" PVCs.  Also tested manually by batching pods referencing an invalid PVC with schedulable pods.

```
2022-05-19T16:52:55.249Z	ERROR	controller	Unable to batch pod inflate-7d56fc4858-dz9rr, getting persistent volume claim "myclaim", PersistentVolumeClaim "myclaim" not found	{"commit": "aaab0c9"}
2022-05-19T16:52:55.249Z	INFO	controller	Batched 1 pod(s) in 1.00026088s	{"commit": "aaab0c9"}
```

- [x] No


pkg/controllers/provisioning/provisioner.go
```diff
@@ -141,8 +142,17 @@ func (p *Provisioner) getPods(ctx context.Context) ([]*v1.Pod, error) {
...
 	for i := range podList.Items {
-		if isProvisionable(&podList.Items[i]) {
-			pods = append(pods, &podList.Items[i])
+		pod := podList.Items[i]
+		errs := multierr.Combine(
+			validate(&pod),
+			p.volumeTopology.validatePvcs(ctx, &pod),
+		)
+		if errs != nil {
+			logging.FromContext(ctx).Errorf("Unable to batch pod %s, %s", pod.Name, errs)
```
dewjam, line 151: Noticing I didn't log the errors in the reconcile loop, but I do here.  It's certainly valuable to know that a pod wasn't scheduled due to a missing PVC, but it's not really due to Karpenter misbehaving.  And people can always describe the pending pod to see why it wasn't scheduled.

spring1843 (dismissed):
LGTM

pkg/controllers/provisioning/provisioner.go
```diff
@@ -141,8 +143,17 @@ func (p *Provisioner) getPods(ctx context.Context) ([]*v1.Pod, error) {
...
 	for i := range podList.Items {
-		if isProvisionable(&podList.Items[i]) {
-			pods = append(pods, &podList.Items[i])
+		pod := podList.Items[i]
+		errs := multierr.Combine(
+			validate(&pod),
+			p.volumeTopology.validatePersistentVolumeClaims(ctx, &pod),
+		)
+		if errs != nil {
+			logging.FromContext(ctx).With("pod", fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)).Errorf("Unable to batch pod, %s", errs)
```
tzneal, line 152: I don't think we want the errors logged here.  This error will just be repeated for a bad pod a ton of times if you have lots of pod churn.

dewjam: You're right it will log at every attempt to batch pods.  But we can't necessarily return the error otherwise we will fail the entire batch of pods.  Or are you suggesting we just make it a debug log instead?

tzneal: Does the error show up on the pod events from kube-scheduler? If so, maybe we don't need to log anything and just consider it a non-provisionable pod.

dewjam: yep, it does.  I was thinking either Debug level logging, or don't log at all.
```
Events:
  Type     Reason            Age              From               Message
  ----     ------            ----             ----               -------
  Warning  FailedScheduling  7s (x2 over 8s)  default-scheduler  0/1 nodes are available: 1 persistentvolumeclaim "myclaim" not found.
```

tzneal: Either sounds good to me too.

tzneal (approved):
LGTM

PR #1841, opened 2022-05-23
Reconciling Nodes on Provisioner Spec changes

ra-grover:
**1. Issue, if available:**
https://github.com/aws/karpenter/issues/1457
https://github.com/aws/karpenter/issues/1018

The implementation stamps a label on Node signifying the provisioner version, when the provisioner spec changes so does it's hash and as soon Nodes are reconciled, they are made to compare with the current provisioner hash. 

If the hash turns out to be different, they are put in a `DelayedQueue` which enables the scheduling of nodes in to batches for termination. So entire set of nodes doesn't go down at once, but only `5%` of nodes at one time, with a 10 minute interval between in each batch. (Can be also taken dynamically in the spec).

The same approach can also be used in `Expiration` Sub-reconciler, where all nodes are terminated at once. Instead they could be put in this delayed queue and terminated in batches thereafter.

To batch the nodes, the total number of nodes provisioned by a specific provisioner is needed, to enable that there is a new field on `ProvisionerStatus`. So let's suppose 100 nodes are provisioned, when they are reconciled, we consider putting first 5 nodes reconciles into first batch, and consequently others. `DelayedQueue` in the` client-go` allows us to put the items at a later point of time.

Through test cases and also on a live EKS cluster.

- [x] No

### DIFF
--- a/charts/karpenter/crds/karpenter.sh_provisioners.yaml
+++ b/charts/karpenter/crds/karpenter.sh_provisioners.yaml
@@ -241,6 +241,9 @@ spec:
                   x-kubernetes-int-or-string: true
                 description: Resources is the list of resources that have been provisioned.
                 type: object
+              totalNodesProvisioned:
+                description: Total number of nodes provisioned by the provisioner
+                type: integer
             type: object
         type: object
     served: true
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -100,7 +100,7 @@ func main() {
 		state.NewPodController(manager.GetClient(), cluster),
 		persistentvolumeclaim.NewController(manager.GetClient()),
 		termination.NewController(ctx, manager.GetClient(), clientSet.CoreV1(), cloudProvider),
-		node.NewController(manager.GetClient()),
+		node.NewController(ctx, manager.GetClient(), clientSet.CoreV1()),
 		metricspod.NewController(manager.GetClient()),
 		metricsnode.NewController(manager.GetClient()),
 		counter.NewController(manager.GetClient()),
--- a/pkg/apis/provisioning/v1alpha5/provisioner_status.go
+++ b/pkg/apis/provisioning/v1alpha5/provisioner_status.go
@@ -33,6 +33,9 @@ type ProvisionerStatus struct {
 
 	// Resources is the list of resources that have been provisioned.
 	Resources v1.ResourceList `json:"resources,omitempty"`
+
+	//Total number of nodes provisioned by the provisioner
+	TotalNodesProvisioned int `json:"totalNodesProvisioned,omitempty"`
 }
 
 func (p *Provisioner) StatusConditions() apis.ConditionManager {
--- a/pkg/apis/provisioning/v1alpha5/register.go
+++ b/pkg/apis/provisioning/v1alpha5/register.go
@@ -45,6 +45,7 @@ var (
 	DoNotEvictPodAnnotationKey      = Group + "/do-not-evict"
 	EmptinessTimestampAnnotationKey = Group + "/emptiness-timestamp"
 	TerminationFinalizer            = Group + "/termination"
+	ProvisionerVersionKey           = Group + "/provisioner-version"
 )
 
 const (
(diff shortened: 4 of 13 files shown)


njtran:
Hey @ra-grover, thanks for the PR! I can tell you've thought deeply about the implementation and I have some comments that I'll post here. Would love to see you at the next Working Group so we can also discuss this if you we don't find consensus here. 

> The implementation stamps a label on Node signifying the provisioner version, when the provisioner spec changes so does it's hash and as soon Nodes are reconciled, they are made to compare with the current provisioner hash.

Since we use watch-based controllers, I think this controller should watch provisioners, and shouldn't be in the controller/node folder. This could look like [this](https://github.com/aws/karpenter/blob/main/pkg/controllers/counter/controller.go#L98). Implementing this means we also need to think about if Nodes should be reconciled if a user/controller updates it to be out-of-spec.

> If the hash turns out to be different, they are put in a DelayedQueue which enables the scheduling of nodes in to batches for termination. So entire set of nodes doesn't go down at once, but only 5% of nodes at one time, with a 10 minute interval between in each batch. (Can be also taken dynamically in the spec).
   The same approach can also be used in Expiration Sub-reconciler, where all nodes are terminated at once. Instead they could be put in this delayed queue and terminated in batches thereafter.
   To batch the nodes, the total number of nodes provisioned by a specific provisioner is needed, to enable that there is a new field on ProvisionerStatus. So let's suppose 100 nodes are provisioned, when they are reconciled, we consider putting first 5 nodes reconciles into first batch, and consequently others. DelayedQueue in the client-go allows us to put the items at a later point of time.

I agree that rate-limiting here is definitely necessary, but this solution should be able to scale to all the related use-cases. Check [this issue](https://github.com/aws/karpenter/issues/1738) for more information on how we're tracking all related concepts together. Also, I don't think a DelayedQueue TerminationQueue implementation is the right solution here, as it does something very similar to the Termination controller, and takes away from the centralized termination model designed [here](https://github.com/aws/karpenter/blob/main/designs/termination.md#triggering-termination). Implementing it like this means future controllers could need to do the same, which could create a lot of deduplicable code.

W.r.t. the 5% disruption budget and 10 minute interval, this needs to be configurable, so a default like this might not make sense, especially if the controller is enabled by default. To help drive us to good values (if they exist) we can ask some questions:
- How many nodes are you thinking would this controller reconcile? 
- How often do you anticipate Provisioner Spec changes? 
- Are there times you would want this disabled? (e.g. Maintenance windows)
- Do we need to reconciliation to end with the same cluster state? With complex enough scheduling constraints, rolling these could result in a different bin-packing and scheduling decisions (e.g. more AZ node skew, different instance types, different bin-packing decisions)

Happy to discuss more on slack or here. I'm currently working on fleshing out the [NodeDisruptionBudget](https://github.com/aws/karpenter/blob/main/designs/termination.md#parallel-node-termination) which is deeply connected to the DelayedQueue you've implemented here.

njtran:
> Does it mean that you will plan a BackoffQueue for Nodes (similar to what I have implemented as DelayedQueue) and you will keep retrying to terminate the nodes from the queue, in case if the deletions are rejected from the Webhook considering NodeDisruption budgets ?

The excerpt you quoted was from the design doc, right? This is a little different from what we do now. More [here](https://karpenter.sh/v0.10.1/tasks/deprovisioning/). Can't comment on when the design for NDB will come out, but we'll need to consider which if not all deletion requests will be limited by. 

> Now considering the scope of this PR and my intention to contribute to Karpenter in someway, do you think in order to address https://github.com/aws/karpenter/issues/1457 , I will be able to do that in this PR only after you are done with your NodeDisruptionBudget Implementation ?

We'll need to definitely have NDB's design out to know how we'd go forward with this. My hope with NDBs is that it can act as a CRD that acts in tandem with the Karpenter controller so that going forward, any additional features to deprovision nodes do not have to additionally consider how it fits with other deprovisioning features. I'll put this as blocked right now and get back to you when we have more information on this.  Does this sound alright?

ra-grover:
> Does this sound alright?

Sure thanks @njtran

njtran:
Hey @ra-grover, I cut a PR with the design doc on how we're thinking of this. We're initially only going to be reconciling nodes when the AMI that Karpenter discovers does not match the nodes. https://github.com/aws/karpenter/pull/2569

Closing this, but would love to have feedback from you on this.

PR #1847, opened 2022-05-24
during scheduling, filter instance types to those that won't violate limits

### LINKED ISSUE #1728: Enhance resources limit constraint consistency
Enhance ["spec.limits.resources"](https://karpenter.sh/preview/tasks/set-resource-limits/) provisioner resource limiter to be geared towards stronger consistency

Karpenter's resources limit checking is eventually consistent, which can result in overrun of compute resources during rapid scale outs.
ASGs Max Capacity is a hard limit, which makes this implementations strongly consistent

In environments were capacity limits require stronger consistency, EKS Managed nodegroup are used 

**Additional context**

**Attachments**

### Community Note

tzneal:
**1. Issue, if available:**
Fixes #1728

During the scheduling process, only provision from instance types that
won't violate limits.

Unit testing & deployed to EKS. I scaled up pods with an 8Gb memory limit set on my provisioner and the usage never exceeded 8Gb.

- [ ] No

### DIFF
--- a/pkg/controllers/counter/controller.go
+++ b/pkg/controllers/counter/controller.go
@@ -22,7 +22,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
 	controllerruntime "sigs.k8s.io/controller-runtime"
-
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -33,6 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
+	"github.com/aws/karpenter/pkg/utils/resources"
 )
 
 // Controller for the resource
@@ -74,16 +74,20 @@ func (c *Controller) resourceCountsFor(ctx context.Context, provisionerName stri
 	if err := c.kubeClient.List(ctx, &nodes, client.MatchingLabels{v1alpha5.ProvisionerNameLabelKey: provisionerName}); err != nil {
 		return nil, err
 	}
-	var cpu = resource.NewScaledQuantity(0, 0)
-	var memory = resource.NewScaledQuantity(0, resource.Giga)
-	for _, node := range nodes.Items {
-		cpu.Add(*node.Status.Capacity.Cpu())
-		memory.Add(*node.Status.Capacity.Memory())
+
+	// record all resources provisioned by the provisioners
+	provisioned := []v1.ResourceList{
+		{
+			// record some zero values so the status will display something even if no nodes are provisioned
+			v1.ResourceCPU:    resource.MustParse("0"),
+			v1.ResourceMemory: resource.MustParse("0"),
+		},
 	}
-	return v1.ResourceList{
-		v1.ResourceCPU:    *cpu,
-		v1.ResourceMemory: *memory,
-	}, nil
+
+	for _, node := range nodes.Items {
+		provisioned = append(provisioned, node.Status.Capacity)
+	}
+	return resources.Merge(provisioned...), nil
 }
 
 // Register the controller to the manager
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -206,7 +206,7 @@ func (p *Provisioner) schedule(ctx context.Context, pods []*v1.Pod) ([]*schedule
 		return nil, fmt.Errorf("getting daemon overhead, %w", err)
 	}
 
-	return scheduler.NewScheduler(nodeTemplates, p.cluster, topology, instanceTypes, daemonOverhead, p.recorder).Solve(ctx, pods)
+	return scheduler.NewScheduler(nodeTemplates, provisionerList.Items, p.cluster, topology, instanceTypes, daemonOverhead, p.recorder).Solve(ctx, pods)
 }
 
 func (p *Provisioner) launch(ctx context.Context, node *scheduler.Node) error {
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -16,15 +16,10 @@ package scheduling
 
 import (
 	"context"
+	"fmt"
 	"sort"
 
 	"github.com/samber/lo"
-
-	"github.com/aws/karpenter/pkg/events"
-	"github.com/aws/karpenter/pkg/scheduling"
-
-	"github.com/aws/karpenter/pkg/controllers/state"
-
 	"go.uber.org/multierr"
 	v1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/logging"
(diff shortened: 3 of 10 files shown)


pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -159,13 +169,67 @@ func (s *Scheduler) add(pod *v1.Pod) error {
...
+			// we will launch this node and need to track its maximum possible resource usage against our remaining resources
+			s.remainingResources[nodeTemplate.ProvisionerName] = subtractMax(s.remainingResources[nodeTemplate.ProvisionerName], node.InstanceTypeOptions)
 			return nil
 		}
 		errs = multierr.Append(errs, err)
 	}
 	return errs
 }
+
+// subtractMax returns the remaining resources after subtracting the max resource quantity per instance type. To avoid
```
tzneal, line 195: This could cause us to require multiple scheduling rounds as you approach the limits configured on a provisioner. We still launch the max node sizes possible, but to avoid overshooting we have to assume worst case.

pkg/controllers/counter/controller.go
```diff
@@ -74,16 +74,20 @@ func (c *Controller) resourceCountsFor(ctx context.Context, provisionerName stri
 	if err := c.kubeClient.List(ctx, &nodes, client.MatchingLabels{v1alpha5.ProvisionerNameLabelKey: provisionerName}); err != nil {
 		return nil, err
 	}
-	var cpu = resource.NewScaledQuantity(0, 0)
-	var memory = resource.NewScaledQuantity(0, resource.Giga)
+
+	// record all resources provisioned by the provisioners
+	provisioned := []v1.ResourceList{
+		{
+			// record some zero values so the status will display something even if no nodes are provisioned
```
ellistarn, line 81: Can you elaborate? Is this just for kubectl?

pkg/controllers/counter/controller.go
```diff
@@ -74,16 +74,20 @@ func (c *Controller) resourceCountsFor(ctx context.Context, provisionerName stri
 	if err := c.kubeClient.List(ctx, &nodes, client.MatchingLabels{v1alpha5.ProvisionerNameLabelKey: provisionerName}); err != nil {
 		return nil, err
 	}
-	var cpu = resource.NewScaledQuantity(0, 0)
-	var memory = resource.NewScaledQuantity(0, resource.Giga)
+
```
ellistarn, line 77: I'm wondering if we can collapse this logic into the core provisioning loop. Trying to remember why it was decoupled.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -16,40 +16,43 @@ package scheduling
...
-		recorder:       recorder,
-		preferences:    &Preferences{},
+		nodeTemplates:      nodeTemplates,
+		topology:           topology,
+		cluster:            cluster,
+		instanceTypes:      instanceTypes,
+		daemonOverhead:     daemonOverhead,
+		recorder:           recorder,
+		preferences:        &Preferences{},
+		remainingResources: map[string]v1.ResourceList{},
```
ellistarn, line 46: Thoughts on inverting this to `resources`?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -64,21 +67,28 @@ func NewScheduler(nodeTemplates []*scheduling.NodeTemplate, cluster *state.Clust
 			return true
 		}
 		s.inflight = append(s.inflight, NewInFlightNode(node, s.topology, nodeTemplate.StartupTaints, s.daemonOverhead[nodeTemplate]))
+
+		// We don't use the status field and instead recompute the remaining resources to ensure we have a consistent view
```
ellistarn, line 71: Why don't we use the status field?

pkg/scheduling/nodetemplate.go
```diff
@@ -27,22 +27,26 @@ import (
 // the fields in Provisioner. These structs are maintained separately in order
 // for fields like Requirements to be able to be stored more efficiently.
 type NodeTemplate struct {
+	ProvisionerName      string
```
ellistarn, line 30: This is currently used by extracting the name from `labels` elsewhere in the code.

pkg/controllers/provisioning/scheduling/scheduler.go
tzneal: It's not guaranteed to be updated yet.  At least this way we have a consistent view of the cluster when scheduling (or as consistent as we can be).  The other benefit is that this will continue working as I expect when we stop creating nodes that `ForEachNode` will still iterate over nodes that are in the process of being launched, meaning this will continue to work.

pkg/scheduling/nodetemplate.go
```diff
@@ -27,22 +27,26 @@ import (
...
 	Provider             *v1alpha5.Provider
 	Labels               map[string]string
 	Taints               Taints
 	StartupTaints        Taints
 	Requirements         Requirements
 	KubeletConfiguration *v1alpha5.KubeletConfiguration
+	Limits               *v1alpha5.Limits
```
ellistarn, line 37: It feels wrong to me to include this in the node template. We're converging this object towards `provisioner`. When I designed this struct, I was careful to limit it to the parameters inherent to the node, and removed any that relate to the orchestration of that node (e.g. limits, ttls, knowledge of provisioner).

pkg/controllers/state/cluster.go
```diff
@@ -57,6 +57,9 @@ func NewCluster(ctx context.Context, client client.Client) *Cluster {
 // compute topology information.
 type Node struct {
 	Node *v1.Node
+	// Capacity is the total amount of resources on the node.  The available resources are the capacity minus overhead
+	// minus anything allocated to pods.
+	Capacity v1.ResourceList
```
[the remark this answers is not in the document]
tzneal: Capacity never changes, Available and DaemonSetRequested do change.  We need both Available and DaemonSetRequested as for in-flight nodes we know that there is some set of capacity that will be reduced by the daemonsets, but it's not the full set of daemonset resources as they schedule over time and we observe the node during this scheduling.

pkg/scheduling/nodetemplate.go
ellistarn: Yeah I felt really dirty doing this. I think we might need another concept in scheduler.go that sits outside of NodeTemplate.

pkg/controllers/provisioning/scheduling/scheduler.go
tzneal: If it's just the resources used, then you need to also store the provisioner limits so you can calculate.  It would work, but it's an extra variable and this way you just need to filter for instance types smaller than the remaining resources which is pretty simple as opposed to calculating that amount each time and then filtering.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -206,7 +206,7 @@ func (p *Provisioner) schedule(ctx context.Context, pods []*v1.Pod) ([]*schedule
 		return nil, fmt.Errorf("getting daemon overhead, %w", err)
 	}
 
-	return scheduler.NewScheduler(nodeTemplates, p.cluster, topology, instanceTypes, daemonOverhead, p.recorder).Solve(ctx, pods)
+	return scheduler.NewScheduler(nodeTemplates, provisionerList.Items, p.cluster, topology, instanceTypes, daemonOverhead, p.recorder).Solve(ctx, pods)
```
ellistarn, line 209: Thoughts on generating the node templates inside the scheduler, now that we're passing the provisioner in?

pkg/controllers/provisioning/scheduling/scheduler.go
ellistarn: So IIUC, in a followon PR you'll update the limit in the provisioning loop and then replace `remainingResources` using provisioner.status.limits?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -159,13 +170,67 @@ func (s *Scheduler) add(pod *v1.Pod) error {
...
 		errs = multierr.Append(errs, err)
 	}
 	return errs
 }
+
+// subtractMax returns the remaining resources after subtracting the max resource quantity per instance type. To avoid
+// overshooting out, we need to pessimistically assume that if e.g. we request a 2, 4 or 8 CPU instance type
+// that the 8 CPU instance type is all that will be available.  This could cause a batch of pods to take multiple rounds
+// to schedule.
+func subtractMax(remaining v1.ResourceList, instanceTypes []cloudprovider.InstanceType) v1.ResourceList {
```
ellistarn, line 200: There are a couple of similar helpers in cloudprovider/requirements.go that filter instance types based off of requirements. Thoughts on them living together?

ellistarn (approved):
Couple optional comments

pkg/controllers/provisioning/scheduling/scheduler.go
tzneal: At some point, and likely in the same PR that I fix our metrics code scale problems.

pkg/controllers/provisioning/provisioner.go
tzneal: NodeTemplates are used by topology and getDaemonOverhead which are both in the provisioner.  The NodeTemplate being a subset of a provisioner with the added functionality of a tightened set of requirements causes some complications. If you need the requirements, you need the NodeTemplate.  If you need anything outside the subset that's in NodeTemplate, you need both the provisioner and the NodeTemplate.

PR #1852, opened 2022-05-24
Simplified instance type requirements to live behind cloudprovider API

ellistarn:
**1. Issue, if available:**

This simplification is a preliminary change for the Machine CR.

- [ ] No

### DIFF
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -120,20 +120,26 @@ func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
 }
 
 func (c *CloudProvider) GetRequirements(ctx context.Context, provider *v1alpha5.Provider) (scheduling.Requirements, error) {
+	instanceTypes, err := c.GetInstanceTypes(ctx)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -120,20 +120,26 @@ func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
...
+
 	awsprovider, err := v1alpha1.Deserialize(provider)
 	if err != nil {
-		return scheduling.NewRequirements(), apis.ErrGeneric(err.Error())
+		return nil, apis.ErrGeneric(err.Error())
 	}
 	// Constrain AZs from subnets
 	subnets, err := c.subnetProvider.Get(ctx, awsprovider)
 	if err != nil {
-		return scheduling.NewRequirements(), err
```
suket22, line 130: Not sure I understand why we were returning `scheduling.NewRequirements()` earlier. `nil` makes more sense to me.

suket22 (approved):

pkg/cloudprovider/aws/cloudprovider.go
ellistarn: Because it used to be a concrete type. Now it's a map which is nillable.

Issue #1615, opened 2022-04-01
Parameterising the container runtime


himanshurajput32:
Karpenter: v0.8.0
Kubernetes: v1.21

Currently **containerd** is fixed as a runtime in case of non-GPU instances. 

https://github.com/aws/karpenter/blob/476804d3962586fe53a784315d56315e078f387d/pkg/cloudprovider/aws/amifamily/al2.go#L65-L76

Can we introduce a new field under provisioner's configuration so that one can use **dockerd** with Non-GPU instances ?(Until Kubernetes 1.23)

ellistarn:
Can you explain you use case a little bit more? I'm wary of expanding Karpenter's complexity to handle deprecated features, since we have limited resources.

cebernardi:
Thank you for your prompt reply. 

We would like to control the container runtime of the worker nodes Amis, instead of defaulting on `containerd` (if not GPU), as this would give us time to prepare for the transition (currently we're running on docker).

We're running on EKS 1.21, and the AMIs provide opt-in support for contained, but we haven't opted in yet, and we would like to not go down the route of using LaunchTemplates.

We were wondering if introducing this control over the runtime could be a common use case that the community could benefit from.

We can help deliver this if you see value in it.

ellistarn:
> (currently we're running on docker)

I'm curious what specifically is your dependency on docker. In general, I'd prefer to encourage users to use containerd when using karpenter. We have some work in progress that will enable you to run arbitrary scripts (e.g. swap containerd for dockerd) before we start the node. We're hoping that this can be a catchall for users without needing to define their own launch templates. Because of this, I'm not sold on introducing first class support for dockerd in the provisioner API. 

Is this acceptable to you? Are any of my assumptions wrong?

cebernardi:
> I'm curious what specifically is your dependency on docker
we have a few metrics that depend on docker runtime, that we're migrating from, and it just requires time, which we wanted to buy proposing this feature. 

For the rest, yes, I see your point of supporting this particular feature as not in line with the current roadmap, and yes, the bootstrap scripts support that you're mentioning would absolutely supersede this "pick-your-runtime".

Is there any way we could help with this?

ellistarn:
@suket22 is driving this work. You can find him on the k/k slack

tzneal:
Closed with #1855

PR #1855, opened 2022-05-25
Adding container-runtime to kubeletConfiguration

suket22:
- Adding container-runtime to kubeletConfiguration

**1. Issue, if available:**
#1615 

* After a little back and forth, I think I've decided to let the container-runtime be parameterized. Letting folks set this via UserData will be a little complicated given you'd need to modify [kubelet CLI args and it's not part of their JSON config](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/). 
* I want each cloud-provider to dictate which container runtimes are supported so I moved the validation and runtime defaulting into the provider. 
* We no longer use Docker with Nvidia and AMD GPUs. This is a breaking change of sorts and **will need to be called out in the Release Notes.**
  * We can safely make this breaking change because there's now a way to override the container-runtime to be used.

Some unit tests, and also tested some of this locally.

WIP.
- [ ] No


tzneal (approved):

AswinKumarVenketesan:
Hi,
As you mentioned I have specified the container runtime but I have received the below error:

error validating data: ValidationError(Provisioner.spec.kubeletConfiguration): unknown field "containerRuntime" in sh.karpenter.v1alpha5.Provisioner.spec.kubeletConfiguration; if you choose to ignore these errors, turn validation off with --validate=false

PR #1856, opened 2022-05-25
don't pre-bind pods to nodes (#1773)

tzneal:
**1. Issue, if available:**

Several, I'll link/close them all later once we get approval for a PR that eliminates binding.

* don't pre-bind pods to nodes
- Eliminate binding of pods to nodes
- remove the Karpenter specific not-ready taint
- debounce the emptiness controller to handle the initial node
  startup and prevent confusing log messages before kube-scheduler
  has had an opportunity to schedule pods

Unit testing & deployed to EKS, including GPU nodes.

- [ ] No


pkg/apis/provisioning/v1alpha5/register.go
```diff
@@ -41,7 +41,7 @@ var (
 		return nil
 	})
 	ProvisionerNameLabelKey         = Group + "/provisioner-name"
-	NotReadyTaintKey                = Group + "/not-ready"
```
tzneal, line 44: We can't place a taint on the node (or at least would need to remove it fairly quickly) as it prevents daemonsets from binding so we don't get any device plugins and extended resources never register.

pkg/utils/options/options.go
```diff
@@ -44,6 +44,7 @@ func MustParse() Options {
 	flag.StringVar(&opts.AWSNodeNameConvention, "aws-node-name-convention", env.WithDefaultString("AWS_NODE_NAME_CONVENTION", string(IPName)), "The node naming convention used by the AWS cloud provider. DEPRECATION WARNING: this field may be deprecated at any time")
 	flag.BoolVar(&opts.AWSENILimitedPodDensity, "aws-eni-limited-pod-density", env.WithDefaultBool("AWS_ENI_LIMITED_POD_DENSITY", true), "Indicates whether new nodes should use ENI-based pod density")
 	flag.StringVar(&opts.AWSDefaultInstanceProfile, "aws-default-instance-profile", env.WithDefaultString("AWS_DEFAULT_INSTANCE_PROFILE", ""), "The default instance profile to use when provisioning nodes in AWS")
+	flag.BoolVar(&opts.AWSEnablePodENI, "aws-enable-pod-eni", env.WithDefaultBool("AWS_ENABLE_POD_ENI", false), "If true then instances that support pod ENI will report a vpc.amazonaws.com/pod-eni resource")
```
tzneal, line 47: The pod-eni resources only appear from the vpc-cni if it's configured appropriately.  We need the customer to tell us if they configure it that way or not as I don't see a way to determine it ourselves.  If we always reported the resource, the node would never become "ready" from our POV as the ENI resource would never register.

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -157,6 +160,26 @@ var _ = Describe("Allocation", func() {
 	})
 
 	Context("Reconciliation", func() {
+		Context("Standard Labels", func() {
+			It("should apply OS label based on the AMI Family", func() {
+				ExpectApplied(ctx, env.Client, provisioner)
```
tzneal, line 165: This was the issue Brandon saw, we've never labeled the nodes correctly but it didn't matter since we pre-bound and the kubelet would generate these labels anyway on startup.  With our inflight support we need to know the correct final labels.

pkg/utils/options/options.go
tzneal: Technically possibly, but not easily.  The docs for enabling the pod ENI are [here](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html). 

To do it, you set an environment variable on the aws-node DS.
```sh
kubectl set env daemonset aws-node -n kube-system ENABLE_POD_ENI=true
```

So to have a heterogeneous cluster, you'd need to edit the existing aws-node DS to not select all nodes and then create a duplicate DS which selected the other nodes and have only one of them configured with `ENABLE_POD_ENI=true`

pkg/apis/provisioning/v1alpha5/util.go
```diff
@@ -15,21 +15,26 @@ limitations under the License.
...
-			}
-		}
+// NodeIsReady returns true if:
+// a) its current status is set to Ready
+// b) all the startup taints have been removed from the node
+// c) all extended resources have been registered
+// This method handles both nil provisioners and nodes without extended resources gracefully.
+func NodeIsReady(ctx context.Context, node *v1.Node, provisioner *Provisioner) bool {
+	// fast checks first
+	if GetCondition(node.Status.Conditions, v1.NodeReady).Status != v1.ConditionTrue {
```
ellistarn, line 34: Shortcircuitting would let you inline this
```
return GetCondition(node.Status.Conditions, v1.NodeReady).Status == v1.ConditionTrue && 
  isStartupTaintRemoved(node, provisioner) 
  && isExtendedResourceRegistered(ctx, node)
```

pkg/apis/provisioning/v1alpha5/util.go
```diff
@@ -40,3 +45,43 @@ func GetCondition(conditions []v1.NodeCondition, match v1.NodeConditionType) v1.
 	}
 	return v1.NodeCondition{}
 }
+
+// isStartupTaintRemoved returns true if there are no startup taints registered for the provisioner, or if all startup
+// taints have been removed from the node
+func isStartupTaintRemoved(node *v1.Node, provisioner *Provisioner) bool {
+	if provisioner != nil {
```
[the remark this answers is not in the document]
tzneal: Cluster state uses this too. If you've deleted a provisioner, but the node it launched still exists we won't be able to look up startup taint information.

pkg/cloudprovider/aws/instance.go
```diff
@@ -254,35 +254,31 @@ func (p *InstanceProvider) getInstance(ctx context.Context, id string) (*ec2.Ins
...
 	for _, instanceType := range instanceTypes {
 		if instanceType.Name() == aws.StringValue(instance.InstanceType) {
 			nodeName := strings.ToLower(aws.StringValue(instance.PrivateDnsName))
 			if injection.GetOptions(ctx).GetAWSNodeNameConvention() == options.ResourceName {
 				nodeName = aws.StringValue(instance.InstanceId)
 			}
 
+			// Since we no longer pre-bind, we need to ensure that all resources that were possibly considered for scheduling
+			// purposes are placed on the node.  The extended resources will be moved to an annotation, while the
+			// non-extended will be left on the node.
```
ellistarn, line 267: It's generally easier to reason about things if we can avoid bimodal behavior (e.g. ABC except for XYZ). Why can't we write these to the node capacity? Is it because the device driver may briefly zero it out and cause overscaling? Alternatively can we store them all in the annotation?

ellistarn: We could also not store them and just retrieve them from the instance type

tzneal: Not following, this change just ensures that all resources reported by the instance are also reported on the node object it returns. It was a bug in our AWS provider that we didn't report all resources on the node.  We later peel off the extended resources and store them in an annotation.  We're not guaranteed to see the same instances returned by the instance provider, if an instance type becomes unlaunchable it will disappear.  We've then lost knowledge of what extended resource it had, so I thought it was cleaner just to store the resources on the node as an annotation.

tzneal: Yes, kubelet zeros out both the capacity and the allocatable for extended resources on startup.

ellistarn (commented):
Discussed offline:

- Create the node object with just metadata (no taints or extended resources)
- Track InstanceType in cluster state instead of relying on the node object
- To make the above work, return all instance types but with empty offerings when iced
- Don't terminate the node if failed to startup (runaway scaling is not possible w/ in flight)

This means that we:
- Don't need extended resources annotation
- Don't need readiness annotation
- Minimize the amount of behavior we can get wrong w/ the node object (closer to not creating the node)

Tradeoffs:
- We lose the ability to recover from transient node failures if they happen during startup
- This should be exceedingly rare and we can follow up with a more comprehensive autorepair feature

cmd/controller/main.go
```diff
@@ -106,15 +106,19 @@ func main() {
 		logging.FromContext(ctx).Errorf("watching configmaps, config changes won't be applied immediately, %s", err)
 	}
 
-	cluster := state.NewCluster(ctx, manager.GetClient())
+	instanceTypes, err := cloudProvider.GetInstanceTypes(ctx)
```
ellistarn, line 109: This is going to collide with @dewjam's changes

pkg/controllers/state/cluster.go
```diff
@@ -67,11 +79,13 @@ type Node struct {
...
 	DaemonSetRequested v1.ResourceList
-
 	// HostPort usage of all pods that are bound to the node
 	HostPortUsage *HostPortUsage
+	// Provisioner is the provisioner used to create the node.
+	Provisioner *v1alpha5.Provisioner
 
-	podRequests map[types.NamespacedName]v1.ResourceList
+	podRequests  map[types.NamespacedName]v1.ResourceList
+	InstanceType cloudprovider.InstanceType
```
ellistarn, line 88: How expensive would it be to retrieve these (cached) per node. This will avoid the conflict w/ @dewjam

tzneal: We cache them for 5 minutes, so 12 times per hour * # of provisioners should be the max number of requests. Shouldn't be terrible.

pkg/controllers/state/cluster.go
```diff
@@ -146,10 +172,41 @@ func (c *Cluster) newNode(node *v1.Node) *Node {
...
 	n.DaemonSetRequested = resources.Merge(daemonsetRequested...)
 	n.Capacity = n.Node.Status.Capacity
-	n.Available = resources.Subtract(n.Node.Status.Allocatable, resources.Merge(requested...))
+	n.Available = resources.Subtract(c.getNodeAllocatable(node, n.Provisioner), resources.Merge(requested...))
 	return n
 }
 
+// getNodeAllocatable gets the allocatable resources for the node.
+func (c *Cluster) getNodeAllocatable(node *v1.Node, provisioner *v1alpha5.Provisioner) v1.ResourceList {
+	allocatable := node.Status.Allocatable
```
ellistarn, line 181: Looks like we aren't using this variable everywhere in this scope.

pkg/cloudprovider/aws/instance.go
```diff
@@ -299,6 +295,16 @@ func (p *InstanceProvider) instanceToNode(ctx context.Context, instance *ec2.Ins
 					},
 				},
 			}
+
+			if amiFamily != nil {
```
ellistarn, line 299: How does scheduling work if we don't know AMIFamily? Is it possible for the customer to get this wrong? Should we default to linux? Should we require that the customer sets an OS requirement if they don't set AMIFamily? Sounds like you're arguing we should just not set it and customers may get it wrong.

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -43,10 +43,10 @@ type InstanceType struct {
 	provider           *v1alpha1.AWS
 }
 
-func newInstanceType(info ec2.InstanceTypeInfo, provider *v1alpha1.AWS) *InstanceType {
+func newInstanceType(info ec2.InstanceTypeInfo, includePodENI bool, provider *v1alpha1.AWS) *InstanceType {
```
ellistarn, line 46: IMO we should pass context down and extract from there.

pkg/cloudprovider/util.go
```diff
@@ -0,0 +1,78 @@
+/*
```
ellistarn, line 1: This package seems like the wrong home. WDYT about `scheduling`?

pkg/cloudprovider/util.go
```diff
@@ -0,0 +1,78 @@
...
+// b) all the startup taints have been removed from the node
+// c) all extended resources have been registered
+// This method handles both nil provisioners and nodes without extended resources gracefully.
+func NodeIsReady(node *v1.Node, provisioner *v1alpha5.Provisioner, instanceType InstanceType) bool {
+	// fast checks first
+	return GetCondition(node.Status.Conditions, v1.NodeReady).Status == v1.ConditionTrue &&
+		isStartupTaintRemoved(node, provisioner) && isExtendedResourceRegistered(node, instanceType)
+}
+
+func GetCondition(conditions []v1.NodeCondition, match v1.NodeConditionType) v1.NodeCondition {
```
ellistarn, line 35: Can we make this private?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -232,6 +233,9 @@ func (p *Provisioner) launch(ctx context.Context, node *scheduler.Node) error {
 	if err := mergo.Merge(k8sNode, node.ToNode()); err != nil {
 		return fmt.Errorf("merging cloud provider node, %w", err)
 	}
+	// ensure we clear out the status
+	k8sNode.Status = v1.NodeStatus{}
```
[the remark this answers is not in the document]
tzneal: I don't think we should allow the cloud provider return a v1.Node to begin with :) I can remove it, but it at least preserves the invariant that  cloud provider status's don't get recorded on nodes.

pkg/cloudprovider/aws/instancetype.go
tzneal: Not much difference to me, this way is more explicit about what is being used. Technically using that for the context is a misuse of context per the Go docs since it's not request scoped:

> Use context values only for request-scoped data that transits processes and API boundaries, not for passing optional parameters to functions.

pkg/controllers/node/emptiness.go
```diff
@@ -42,9 +45,19 @@ func (r *Emptiness) Reconcile(ctx context.Context, provisioner *v1alpha5.Provisi
 	if provisioner.Spec.TTLSecondsAfterEmpty == nil {
 		return reconcile.Result{}, nil
 	}
-	if !v1alpha5.NodeIsReady(n, provisioner) {
+
+	instanceType, err := r.getInstanceType(ctx, provisioner, n.Labels[v1.LabelInstanceTypeStable])
+	if err != nil {
+		// We only report the error, but let the emptiness check continue.  This situation shouldn't occur, but if it
```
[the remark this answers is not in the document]
ellistarn: What if the node hasn't booted yet? Outages tend to be correlated due to shared dependencies, so it's possible that instances starting during the outage would result in a ton of churn, applying further load to the EC2 APIs. When something unexpected is going on, the best thing is to slow down and wait for things to recover before taking actions -- especially destructive actions like terminating nodes.

pkg/cloudprovider/aws/instancetype.go
ellistarn: I think this may have been written without deep thought towards usage of go. Controllers may not have been the initial use case and the language is very server specific. Given that we're already storing options in context, I think we might have already violated this rule, if we think it's the correct interpretation. 

That said, I'm not fussed either way.

pkg/cloudprovider/aws/instance.go
```diff
@@ -254,51 +254,39 @@ func (p *InstanceProvider) getInstance(ctx context.Context, id string) (*ec2.Ins
...
-						Architecture:    v1alpha1.AWSToKubeArchitectures[aws.StringValue(instance.Architecture)],
-						OSImage:         aws.StringValue(instance.ImageId),
-						OperatingSystem: v1alpha5.OperatingSystemLinux,
-					},
-				},
 			}
+
+			n.Labels[v1.LabelOSStable] = "linux"
+			if amiFamily != nil {
+				switch *amiFamily {
```
ellistarn, line 282: If you use
`aws.StringValue(amiFamily)`

This will let you drop the extra nil check and layer of indentation/complexity.

pkg/cloudprovider/util.go
tzneal: There's an import cycle again if it's moved there.  The problem is that the NodeIsReady now takes an instance type to know about possible extended resources, which is a cloudprovider type.

ellistarn: Ok. If the method is deeply tied to an instance type, cloud provider sounds like a good home. I'm a bit wary of util sprawl, so something to keep in mind.

pkg/apis/provisioning/v1alpha5/labels.go
```diff
@@ -30,6 +30,7 @@ var (
 	// Karpenter specific domains and labels
 	KarpenterLabelDomain = "karpenter.sh"
 	LabelCapacityType    = KarpenterLabelDomain + "/capacity-type"
+	LabelNodeReady       = KarpenterLabelDomain + "/node-ready"
```
ellistarn, line 33: Ready is a slightly different semantic in k8s. It's possible to go ready -> not ready -> ready. This feature is a latch semantic, so maybe:
```suggestion
	LabelNodeReady       = KarpenterLabelDomain + "/initialized"
```

ellistarn (approved):

ellistarn:
Approved

ellistarn (approved):

Issue #1467, opened 2022-03-03
Bin-Pack Pod Ephemeral Storage Requests 


bwagner5:
**Tell us about your request**
What do you want us to build?

Support provisioning time node volume adjustments when pods request ephemeral storage (emptyDir).

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
What outcome are you trying to achieve, ultimately, and why is it hard/impossible to do right now? What is the impact of not having this problem solved? The more details you can provide, the better we'll be able to understand and solve the problem.

Karpenter defaults the node volume to 20GiB, but pods can request ephemeral storage (emptyDir) that Karpenter needs to bin-pack on to instances. 

**Are you currently working around this issue?**
How are you currently solving this problem?

There is not a workaround currently, but there will soon (https://github.com/aws/karpenter/pull/1420) be a workaround by statically setting the volume size at the provisioner level. 

**Additional context**
Anything else we should know?

**Attachments**
If you think you might have additional information that you'd like to include via an attachment, please do - we'll take a look. (Remember to remove any personally-identifiable information.)

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

dewjam:
Per #1859 , we are now bin-packing based on either our default ephemeral storage value (20GB), or the BlockDeviceMappings provided in the AWS Provider Spec.  This should be included in the upcoming Karpenter release.

PR #1859, opened 2022-05-25
consider ephemeral-storage when bin-packing

dewjam:
**1. Issue, if available:**
#1467

Prior to this change, Karpenter would report an arbitrarily large value for the `ephemeral-storage` resource when discovering AWS instanceTypes.  This can result in Pending pods and node churn if the sum of the `ephemeral-storage` resources for pods being bound to a node exceeds that of the disk size of the launched node.  This change accurately reflects storage attached to instanceTypes so that binpacking of pods based on the `ephemeral-storage` resource type functions as expected.

The amount of storage used for binpacking is based on the size of the ~~"root" device~~ the device where `/var/lib/kubelet` resides.  If a BlockDeviceMapping is defined for the device in the AWS Provider Spec, then Karpenter will use the given BlockDevice size, otherwise it will revert to the default size (20GB).  

Added:
`EphemeralBlockDeviceOverhead` method which returns values specific to the AMIFamily.  How we calculate this overhead is up in the air, so more discussion is needed as to what it should be set to.

Implemented automated testing and also tested manually.  

- [x] No

### DIFF
--- a/pkg/cloudprovider/aws/amifamily/al2.go
+++ b/pkg/cloudprovider/aws/amifamily/al2.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	core "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
@@ -78,7 +79,15 @@ func (a AL2) containerRuntime(instanceTypes []cloudprovider.InstanceType) string
 // DefaultBlockDeviceMappings returns the default block device mappings for the AMI Family
 func (a AL2) DefaultBlockDeviceMappings() []*v1alpha1.BlockDeviceMapping {
 	return []*v1alpha1.BlockDeviceMapping{{
-		DeviceName: aws.String("/dev/xvda"),
-		EBS:        &defaultEBS,
+		DeviceName: a.EphemeralBlockDevice(),
+		EBS:        &DefaultEBS,
 	}}
+}
+
+func (a AL2) EphemeralBlockDevice() *string {
+	return aws.String("/dev/xvda")
+}
+
+func (a AL2) EphemeralBlockDeviceOverhead() resource.Quantity {
+	return resource.MustParse("5Gi")
 }
--- a/pkg/cloudprovider/aws/amifamily/resolver.go
+++ b/pkg/cloudprovider/aws/amifamily/resolver.go
@@ -31,7 +31,7 @@ import (
 	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
 )
 
-var defaultEBS = v1alpha1.BlockDevice{
+var DefaultEBS = v1alpha1.BlockDevice{
 	Encrypted:  aws.Bool(true),
 	VolumeType: aws.String(ec2.VolumeTypeGp3),
 	VolumeSize: resource.NewScaledQuantity(20, resource.Giga),
@@ -73,6 +73,8 @@ type AMIFamily interface {
 	SSMAlias(version string, instanceType cloudprovider.InstanceType) string
 	DefaultBlockDeviceMappings() []*v1alpha1.BlockDeviceMapping
 	DefaultMetadataOptions() *v1alpha1.MetadataOptions
+	EphemeralBlockDevice() *string
+	EphemeralBlockDeviceOverhead() resource.Quantity
 }
 
 // New constructs a new launch template Resolver
@@ -93,7 +95,7 @@ func (r Resolver) Resolve(ctx context.Context, provider *v1alpha1.AWS, nodeReque
 	if err != nil {
 		return nil, err
 	}
-	amiFamily := r.getAMIFamily(provider.AMIFamily, options)
+	amiFamily := GetAMIFamily(provider.AMIFamily, options)
 	amiIDs := map[string][]cloudprovider.InstanceType{}
 	for _, instanceType := range nodeRequest.InstanceTypeOptions {
 		amiID, err := r.amiProvider.Get(ctx, instanceType, amiFamily.SSMAlias(options.KubernetesVersion, instanceType))
@@ -123,7 +125,7 @@ func (r Resolver) Resolve(ctx context.Context, provider *v1alpha1.AWS, nodeReque
 	return resolvedTemplates, nil
 }
 
-func (r Resolver) getAMIFamily(amiFamily *string, options *Options) AMIFamily {
+func GetAMIFamily(amiFamily *string, options *Options) AMIFamily {
 	switch aws.StringValue(amiFamily) {
 	case v1alpha1.AMIFamilyBottlerocket:
 		return &Bottlerocket{Options: options}
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -111,8 +111,12 @@ func (c *CloudProvider) Create(ctx context.Context, nodeRequest *cloudprovider.N
 }
 
 // GetInstanceTypes returns all available InstanceTypes
-func (c *CloudProvider) GetInstanceTypes(ctx context.Context) ([]cloudprovider.InstanceType, error) {
-	return c.instanceTypeProvider.Get(ctx)
+func (c *CloudProvider) GetInstanceTypes(ctx context.Context, provider *v1alpha5.Provider) ([]cloudprovider.InstanceType, error) {
+	awsprovider, err := v1alpha1.Deserialize(provider)
+	if err != nil {
+		return nil, apis.ErrGeneric(err.Error())
+	}
+	return c.instanceTypeProvider.Get(ctx, awsprovider)
 }
 
 func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
@@ -120,7 +124,7 @@ func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
 }
 
 func (c *CloudProvider) GetRequirements(ctx context.Context, provider *v1alpha5.Provider) (scheduling.Requirements, error) {
-	instanceTypes, err := c.GetInstanceTypes(ctx)
+	instanceTypes, err := c.GetInstanceTypes(ctx, provider)
 	if err != nil {
 		return nil, fmt.Errorf("getting instance types, %w", err)
 	}
--- a/pkg/cloudprovider/aws/instancetype.go
+++ b/pkg/cloudprovider/aws/instancetype.go
@@ -26,6 +26,7 @@ import (
 	"knative.dev/pkg/ptr"
 
 	"github.com/aws/karpenter/pkg/cloudprovider"
+	"github.com/aws/karpenter/pkg/cloudprovider/aws/amifamily"
 	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
 	"github.com/aws/karpenter/pkg/utils/resources"
 )
@@ -39,10 +40,12 @@ type InstanceType struct {
 	MaxPods            *int32
 	resources          v1.ResourceList
 	overhead           v1.ResourceList
+	provider           *v1alpha1.AWS
 }
 
-func newInstanceType(info ec2.InstanceTypeInfo) *InstanceType {
+func newInstanceType(info ec2.InstanceTypeInfo, provider *v1alpha1.AWS) *InstanceType {
 	it := &InstanceType{InstanceTypeInfo: info}
+	it.provider = provider
 	it.resources = it.computeResources()
 	it.overhead = it.computeOverhead()
 	return it
@@ -135,9 +138,18 @@ func (i *InstanceType) memory() resource.Quantity {
 	)
 }
 
-// Setting ephemeral-storage to be arbitrarily large so it will be ignored during binpacking
+// Setting ephemeral-storage to be either the default value or what is defined in blockDeviceMappings
 func (i *InstanceType) ephemeralStorage() resource.Quantity {
-	return resource.MustParse("100Pi")
+	ephemeralBlockDevice := amifamily.GetAMIFamily(i.provider.AMIFamily, &amifamily.Options{}).EphemeralBlockDevice()
+	if i.provider.BlockDeviceMappings != nil {
+		for _, blockDevice := range i.provider.BlockDeviceMappings {
+			// If a block device mapping exists in the provider for the root volume, set the volume size specified in the provider
+			if *blockDevice.DeviceName == *ephemeralBlockDevice {
+				return *blockDevice.EBS.VolumeSize
+			}
+		}
+	}
+	return *amifamily.DefaultEBS.VolumeSize
 }
 
 func (i *InstanceType) pods() resource.Quantity {
@@ -210,6 +222,7 @@ func (i *InstanceType) computeOverhead() v1.ResourceList {
 				// eviction threshold https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/kubelet/apis/config/v1beta1/defaults_linux.go#L23
 				100,
 		)),
+		v1.ResourceEphemeralStorage: amifamily.GetAMIFamily(i.provider.AMIFamily, &amifamily.Options{}).EphemeralBlockDeviceOverhead(),
 	}
 	// kube-reserved Computed from
 	// https://github.com/bottlerocket-os/bottlerocket/pull/1388/files#diff-bba9e4e3e46203be2b12f22e0d654ebd270f0b478dd34f40c31d7aa695620f2fR611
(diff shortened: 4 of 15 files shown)


pkg/controllers/provisioning/provisioner.go
```diff
@@ -161,32 +161,34 @@ func (p *Provisioner) getPods(ctx context.Context) ([]*v1.Pod, error) {
...
 			return nil, fmt.Errorf("getting provider requirements, %w", err)
 		}
-		nodeTemplates = append(nodeTemplates, scheduling.NewNodeTemplate(&provisionerList.Items[i], requirements))
+		// Get instance type options
+		instanceType, err := p.cloudProvider.GetInstanceTypes(ctx, provisioner.Spec.Provider)
+		if err != nil {
+			return nil, fmt.Errorf("getting instance types, %w", err)
+		}
+		instanceTypes[provisioner.Name] = append(instanceTypes[provisioner.Name], instanceType...)
+		nodeTemplates = append(nodeTemplates, scheduling.NewNodeTemplate(provisioner, requirements))
```
dewjam, line 183: Curious if it would make sense to add instanceTypes as a field to the nodeTemplates type.  I think that would prevent having to use a map to store provider specific instanceTypes.

(Similarly, maybe we could add DaemonSet overhead as well)

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -924,10 +924,62 @@ var _ = Describe("Allocation", func() {
 				pod := ExpectProvisioned(ctx, env.Client, controller,
 					test.UnschedulablePod(test.PodOptions{ResourceRequirements: v1.ResourceRequirements{
 						Requests: map[v1.ResourceName]resource.Quantity{
-							v1.ResourceEphemeralStorage: resource.MustParse("1Pi"),
+							v1.ResourceEphemeralStorage: resource.MustParse("18Gi"),
 						}}}))
 				ExpectScheduled(ctx, env.Client, pod[0])
 			})
+			It("should launch multiple nodes if sum of pod ephemeral-storage requests exceesd a single nodes capacity", func() {
```
dewjam, line 931: ugh, spelling

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -125,7 +126,7 @@ func (p *InstanceTypeProvider) getInstanceTypeZones(ctx context.Context) (map[st
 }
 
 // getInstanceTypes retrieves all instance types from the ec2 DescribeInstanceTypes API using some opinionated filters
-func (p *InstanceTypeProvider) getInstanceTypes(ctx context.Context) (map[string]*InstanceType, error) {
+func (p *InstanceTypeProvider) getInstanceTypes(ctx context.Context, provider *v1alpha1.AWS) (map[string]*InstanceType, error) {
 	if cached, ok := p.cache.Get(InstanceTypesCacheKey); ok {
```
tzneal, line 130: This isn't correct I don't think.  There's a single cache entry for all instance types, but their Resources() vary by the provider.  Could probably write a test to verify and ensure this stays fixed.

pkg/cloudprovider/aws/amifamily/al2.go
```diff
@@ -78,7 +78,11 @@ func (a AL2) containerRuntime(instanceTypes []cloudprovider.InstanceType) string
...
 func (a AL2) DefaultBlockDeviceMappings() []*v1alpha1.BlockDeviceMapping {
 	return []*v1alpha1.BlockDeviceMapping{{
-		DeviceName: aws.String("/dev/xvda"),
-		EBS:        &defaultEBS,
+		DeviceName: a.RootBlockDevice(),
+		EBS:        &DefaultEBS,
 	}}
 }
+
+func (a AL2) RootBlockDevice() *string {
```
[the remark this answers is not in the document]
dewjam: Good call.  Not sure why I thought `/dev/xvdb` was the root device for BR.

pkg/cloudprovider/aws/instancetypes.go
dewjam: Nice catch, I believe you're right.  I'm thinking about hashing the provider spec to generate a unique cache key.   I think I have a test which can reproduce it as well.

pkg/cloudprovider/aws/amifamily/al2.go
```diff
@@ -78,7 +79,15 @@ func (a AL2) containerRuntime(instanceTypes []cloudprovider.InstanceType) string
...
+		EBS:        &DefaultEBS,
 	}}
 }
+
+func (a AL2) EphemeralBlockDevice() *string {
+	return aws.String("/dev/xvda")
+}
+
+func (a AL2) EphemeralBlockDeviceOverhead() resource.Quantity {
+	return resource.MustParse("2Gi")
```
dewjam, line 92: Probably worth some more discussion how we calculate this overhead.  I'm starting to think we should take a static minValue or a percentage of the size of the `EphemeralBlockDevice`.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -161,32 +161,34 @@ func (p *Provisioner) getPods(ctx context.Context) ([]*v1.Pod, error) {
...
 	for i := range provisionerList.Items {
-		requirements, err := p.cloudProvider.GetRequirements(ctx, provisionerList.Items[i].Spec.Provider)
+		provisioner := &provisionerList.Items[i]
+		requirements, err := p.cloudProvider.GetRequirements(ctx, provisioner.Spec.Provider)
 		if err != nil {
 			return nil, fmt.Errorf("getting provider requirements, %w", err)
 		}
-		nodeTemplates = append(nodeTemplates, scheduling.NewNodeTemplate(&provisionerList.Items[i], requirements))
+		// Get instance type options
+		instanceType, err := p.cloudProvider.GetInstanceTypes(ctx, provisioner.Spec.Provider)
```
ellistarn, line 178:
```suggestion
		instanceTypes, err := p.cloudProvider.GetInstanceTypes(ctx, provisioner.Spec.Provider)
```

pkg/cloudprovider/aws/amifamily/al2.go
tzneal: It will vary by the AMI, not sure of the best way to do it.  For AL2 it looks to be 2.8G.  Maybe just bump it to 5G for all for now? @bwagner5 may have an opinion.

```
/dev/nvme0n1p1   20G  2.8G   18G  14% /
```

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -189,3 +190,11 @@ func (p *InstanceTypeProvider) CacheUnavailable(ctx context.Context, fleetErr *e
 func UnavailableOfferingsCacheKey(capacityType string, instanceType string, zone string) string {
 	return fmt.Sprintf("%s:%s:%s", capacityType, instanceType, zone)
 }
+
+func InstanceTypesCacheKey(provider *v1alpha1.AWS) string {
+	providerHash, err := hashstructure.Hash(provider, hashstructure.FormatV2, nil)
+	if err != nil {
+		panic(fmt.Sprintf("hashing provider spec, %s", err))
+	}
+	return fmt.Sprintf("types:%d", providerHash)
```
ellistarn, line 199: Don't try to make this human readable
```suggestion
	return fmt.Sprintf("%d", providerHash)
```

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -1015,10 +1015,91 @@ var _ = Describe("Allocation", func() {
...
+						Requests: map[v1.ResourceName]resource.Quantity{
+							v1.ResourceEphemeralStorage: resource.MustParse("150Gi"),
+						},
+					},
+					}))[0]
+				ExpectNotScheduled(ctx, env.Client, pod)
+			})
+			It("should pack pods using the blockdevicemappings from the provider spec when defined", func() {
+				provider, _ = v1alpha1.Deserialize(provisioner.Spec.Provider)
+				provider, _ := v1alpha1.Deserialize(provisioner.Spec.Provider)
```
dewjam, line 1085: Where did this come from?

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -210,6 +223,7 @@ func (i *InstanceType) computeOverhead() v1.ResourceList {
 				// eviction threshold https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/kubelet/apis/config/v1beta1/defaults_linux.go#L23
 				100,
 		)),
+		v1.ResourceEphemeralStorage: amiFamily.EphemeralBlockDeviceOverhead(),
```
ellistarn, line 226: consider inlining `amifamily.GetAMIFamily...`

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -111,16 +111,20 @@ func (c *CloudProvider) Create(ctx context.Context, nodeRequest *cloudprovider.N
 }
 
 // GetInstanceTypes returns all available InstanceTypes
-func (c *CloudProvider) GetInstanceTypes(ctx context.Context) ([]cloudprovider.InstanceType, error) {
-	return c.instanceTypeProvider.Get(ctx)
+func (c *CloudProvider) GetInstanceTypes(ctx context.Context, provider *v1alpha5.Provider) ([]cloudprovider.InstanceType, error) {
+	awsprovider, err := v1alpha1.Deserialize(provider)
+	if err != nil {
+		return []cloudprovider.InstanceType{}, apis.ErrGeneric(err.Error())
```
ellistarn, line 117:
```suggestion
		return nil, apis.ErrGeneric(err.Error())
```

ellistarn (commented):
LGTM other than the comments. We seriously need to rethink the NodeTemplate abstraction, it's becoming an unweildy god object with unclear encapsulatio and purpose.

pkg/cloudprovider/aws/instancetypes.go
dewjam: Solved this by hashing the provider and adding it to the key.

dewjam:
I profiled memory usage of Karpenter before/after this change.  Because we are now getting instance types per Provisioner, memory usage will increase along with the number of Provisioners. 

I've gathered some data to show what the impact on memory usage looks like with this code change.  The data was gathered via running top directly on a K8S node.  

***(NOTE: Memory usage collected from v0.10.1 was stable at 95-100M in my testing regardless of provisioner count)***

|# of provisioners| Resident Memory (MB)|Top Output|
|-|-|-|
|1|113|`4761 65532     20   0  811.0m 113.7m  36.1m S   0.7  2.9   0:02.72 controller`|
|2|122.1|`0932 65532     20   0  811.8m 122.1m  35.8m S   0.7  3.8   0:01.74 controller`|
|3|147.8|`8835 65532     20   0  811.8m 147.8m  35.8m S   0.7  3.8   0:01.71 controller`|
|4|149.7|`0850 65532     20   0  811.5m 149.7m  35.8m S   0.3  3.9   0:01.59 controller`|
|5|171.2|`32265 65532     20   0  812.3m 171.2m  35.7m S   0.7  4.4   0:02.55 controller`|

pkg/cloudprovider/aws/amifamily/al2.go
dewjam: I bumped this to 5GB for now.

pkg/controllers/provisioning/provisioner.go
dewjam: changed this to InstanceTypeOptions as we are appending this to a dictionary later in the code.

pkg/cloudprovider/aws/instancetypes.go
dewjam: Fixed.

pkg/cloudprovider/aws/cloudprovider.go
dewjam: Fixed

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -135,9 +138,18 @@ func (i *InstanceType) memory() resource.Quantity {
 	)
 }
 
-// Setting ephemeral-storage to be arbitrarily large so it will be ignored during binpacking
+// Setting ephemeral-storage to be either the default value or what is defined in blockDeviceMappings
 func (i *InstanceType) ephemeralStorage() resource.Quantity {
-	return resource.MustParse("100Pi")
+	EphemeralBlockDevice := amifamily.GetAMIFamily(i.provider.AMIFamily, &amifamily.Options{}).EphemeralBlockDevice()
```
ellistarn, line 143: Local variables should use lower camel case. UpperCamel is reserved for globals and public interfaces
```suggestion
	ephemeralBlockDevice := amifamily.GetAMIFamily(i.provider.AMIFamily, &amifamily.Options{}).EphemeralBlockDevice()
```

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -189,3 +190,11 @@ func (p *InstanceTypeProvider) CacheUnavailable(ctx context.Context, fleetErr *e
 func UnavailableOfferingsCacheKey(capacityType string, instanceType string, zone string) string {
 	return fmt.Sprintf("%s:%s:%s", capacityType, instanceType, zone)
 }
+
+func InstanceTypesCacheKey(provider *v1alpha1.AWS) string {
```
ellistarn, line 194: I think we can make this private
```suggestion
func instanceTypesCacheKey(provider *v1alpha1.AWS) string {
```

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -189,3 +190,11 @@ func (p *InstanceTypeProvider) CacheUnavailable(ctx context.Context, fleetErr *e
...
+	providerHash, err := hashstructure.Hash(provider, hashstructure.FormatV2, nil)
```
ellistarn, line 195: Totally optional, but I avoid qualifying a variable with multiple words unless it helps clarify. In this case, it's unambiguous given the context. 
```suggestion
	hash, err := hashstructure.Hash(provider, hashstructure.FormatV2, nil)
```

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -189,3 +190,11 @@ func (p *InstanceTypeProvider) CacheUnavailable(ctx context.Context, fleetErr *e
...
+	if err != nil {
+		panic(fmt.Sprintf("hashing provider spec, %s", err))
```
ellistarn, line 197: I'm not sure users refer to this as "spec" anywhere else in the codebase
```suggestion
		panic(fmt.Sprintf("hashing provider, %s", err))
```

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -135,9 +138,18 @@ func (i *InstanceType) memory() resource.Quantity {
...
+	if len(i.provider.BlockDeviceMappings) > 0 {
```
tzneal, line 144: Nit, but I don't think this check on the slice length is necessary, you can iterate over empty/nil slices.

pkg/cloudprovider/aws/amifamily/resolver.go
```diff
@@ -93,7 +95,7 @@ func (r Resolver) Resolve(ctx context.Context, provider *v1alpha1.AWS, nodeReque
 	if err != nil {
 		return nil, err
 	}
-	amiFamily := r.getAMIFamily(provider.AMIFamily, options)
+	amiFamily := GetAMIFamily(provider.AMIFamily, options)
```
tzneal, line 98: Does `GetAMIFamily` need to be exported? It's fine if so, but I couldn't tell from the diff.

pkg/cloudprovider/aws/amifamily/resolver.go
```diff
@@ -31,7 +31,7 @@ import (
 	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
 )
 
-var defaultEBS = v1alpha1.BlockDevice{
+var DefaultEBS = v1alpha1.BlockDevice{
```
tzneal, line 34: Does `DefaultEBS` need to be exported? It's fine if so, but I couldn't tell from the diff.

pkg/controllers/provisioning/provisioner.go
ellistarn: Yeah this sounds reasonable to me. We might want to decouple it from what's passed to the cloudprovider, but it's not the end of the world to expose this information. I'd start by thinking about the fundamental role that NodeTemplate plays. Try explaining to a 5 year old what its role is.

ellistarn: IIUC, this will create a circular dependency unless we make other changes.

pkg/cloudprovider/aws/instancetype.go
dewjam: oops, that was an oversight.  Fixing.

pkg/cloudprovider/aws/instancetypes.go
dewjam: ugh, yes correct.

pkg/cloudprovider/aws/amifamily/resolver.go
dewjam: Yeah, it's used in the CloudProvider package when getting instance types, so we do have to export.

pkg/cloudprovider/aws/amifamily/resolver.go
dewjam: Yeah it does.  It's used in the `aws` package.

pkg/cloudprovider/aws/instancetypes.go
dewjam: I don't mind removing for the next rev.

tzneal (approved):
LGTM

PR #1867, opened 2022-05-27
Terraform getting started - replace gavinbunney/kubectl with hashicorp/kubernetes provider

GregCKrause:
**1. Issue, if available:**
None

Replaces gavinbunney/kubectl provider with hashicorp/kubernetes provider

Locally

- [ ] No

### DIFF
--- a/website/content/en/preview/getting-started/getting-started-with-terraform/_index.md
+++ b/website/content/en/preview/getting-started/getting-started-with-terraform/_index.md
@@ -63,8 +63,8 @@ terraform {
       version = "~> 2.4"
     }
     kubectl = {
-      source  = "gavinbunney/kubectl"
-      version = "~> 1.14"
+      source  = "hashicorp/kubernetes"
+      version = "~> 2.11"
     }
   }
 }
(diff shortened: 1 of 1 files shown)


bryantbiggs:
Thanks for the PR @GregCKrause ! Would you mind sharing a bit of the motivation around changing from the `kubectl` provider to the `kubernetes` provider for this getting started guide?

GregCKrause:
> Thanks for the PR @GregCKrause ! Would you mind sharing a bit of the motivation around changing from the `kubectl` provider to the `kubernetes` provider for this getting started guide?

Hi @bryantbiggs! The biggest motivation is that terraform-provider-kubernetes is the official Kubernetes provider from Hashicorp. I do not believe there is a net difference in functionality, but I was a little surprised when following the "Getting Started with Terraform" guide and seeing an alternative provider.

It also seems that there is more activity/support behind the official provider when [comparing the two](https://www.githubcompare.com/gavinbunney/terraform-provider-kubectl+hashicorp/terraform-provider-kubernetes).

bryantbiggs:
Thanks for sharing @GregCKrause - while I agree that it would be great to use the official K8s provider, it currently suffers from a [number of issues for the manifest resource] (https://github.com/hashicorp/terraform-provider-kubernetes/issues?q=is%3Aissue+is%3Aopen+kubernetes_manifest+sort%3Areactions-%2B1-desc) and the one that affects us the most is https://github.com/hashicorp/terraform-provider-kubernetes/issues/1380 which is a [very old issue](https://github.com/hashicorp/terraform-provider-kubernetes-alpha/issues/123). The issue is that the provider will make dry-run API calls to Kubernetes during the plan/apply step and for this getting started guide, the cluster does not exist yet and fails. The kubectl provider does not suffer from this same issue and allows the getting started example to deploy successful in one `terraform apply`. 

Therefore I think its ok if we use the kubectl provider. This is just one example of how a Karpenter provisioner can be deployed, there are various ways that it can be accomplished and this is left up to users to choose. But for an easy to use getting started guide that works without any additional fuss, I think its ok as is. What do you think?

bryantbiggs:
@GregCKrause let me know if we're ok to close this or not, thank you!

suket22:
Closing this PR based on the discussion above.

Issue #1738, opened 2022-04-28
Mega Issue: Deprovisioning Controls


njtran:
**Tell us about your request**
Karpenter provisions nodes and deprovisions nodes as described in the [docs](https://karpenter.sh/preview/concepts/deprovisioning/).

Some users are asking for more control on how to specify when Karpenter should disrupt nodes, and controls on how to rate-limit these disruptions. Opening this issue to aggregate the existing issues.

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**

** Karpenter disruption conditions:**
- [x] aws/karpenter#1716 - Karpenter should watch for AMI version updates and roll nodes to update to the new version.
- [x] aws/karpenter#1495 - allow specifying AMI Release version in Provisioner
- [x] aws/karpenter#1457 - Provisioner Drift
- [x] aws/karpenter#702 - Consume EC2 events similar to Node Termination Handler
- [x] aws/karpenter#1268 - Support for orphaned node destruction (both present in the cluster and outside)
- [x] aws/karpenter#1884 - Include `jitter` for TTLSecondsUntilExpired

**Control over how Karpenter should disrupt nodes:**
- [x] aws/karpenter#1018 - TTLSecondsUntilExpired can result in a lot of node churn at unexpected times. Rate-limit some termination decisions to reduce downtime.
- [x] aws/karpenter-core#753, aws/karpenter#903 - Time-based disruption controls
- [x] aws/karpenter-core#679 - Node Disruption Budget
- [ ] aws/karpenter-core#696 - Use a defined "expected cluster size" to drive deprovisioning
- [x] aws/karpenter-core#710 - Consider deprovisioning differently per-deprovisioning method?
- [x] aws/karpenter-core#735 - Per-node TTL for Consolidation
- [ ] kubernetes-sigs/karpenter#1030 - MinAge for NodeClaims

**Control over Karpenter's Eviction Policy:**
- [ ] aws/karpenter-core#752 - Allow specifying a non-permanent time frame `do-not-evict` for a pod, for un-programmatic removal use-cases. 
- [x] aws/karpenter#1532 - Include `allow-evict` annotation for pods for emptiness condition.
- [ ] aws/karpenter-core#701 - Don't evict running jobs
- [x] aws/karpenter#3052 - Max eviction tries
- [x] aws/karpenter-core#743 - Drain/Eviction Timeout

**Additional context**

**Attachments**
[...]

Issue #1302, opened 2022-02-09
Feature Request - Ability to terminate the node at the specified time


vikas027:
**Tell us about your request**
We already have a way to terminate the instance after X amount of time. It would be great to have a way of doing the same at a specified time as well. For example, say I want to terminate a node(s) at 2:00 AM UTC to save running costs.

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
We are looking to have the ability to scale down nodes at certain times, with ASGs it is easy using the scheduled actions but since Karpenter is managed outside of an ASG (which is a good thing), we can't automatically scale them down to save costs.

**Are you currently working around this issue?**
I may end up writing a lambda function to do the same but it would be great to have this feature inbuilt

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

bwagner5:
This is an interesting request, thanks for posting! We have considered adding a maintenance window to the node expiration, so that if a node expires, it will wait to terminate the node until the maintenance window. We haven't committed to building this feature but it's a similar idea although solving different end-goals. 

I suppose your use-case is to spin down dev workloads or something that is not being used for cost?  I don't think this use-case is something that Karpenter should necessarily deal with directly. But Karpenter would enable something like this to be built really easily with a Kubernetes cron-job that simply `kubectl delete node <nodes>` or to `kubectl scale deploy/<deployment> --replicas=0` which can scale down the non-critical workloads and if the nodes are empty, Karpenter will automatically delete them.  A lambda would be super easy to build as well, as you suggested.

vikas027:
>  I don't think this use-case is something that Karpenter should necessarily deal with directly. 

Yeah, I agree. Just raised this request as we already have a TTL feature :)

albertjunkim:
Thanks for raising this issue, @vikas027. 

I want to add another use case of having a maintenance windows feature. 
Cost-saving for dev workloads is for sure but having a specific window allows more control and resiliency when nodes need to be updated(new patched AMI...etc.)
We wouldn't want to bounce the nodes while there are incoming live traffics, especially in production env. 
[...]

PR #1920, opened 2022-06-10
Adding Maintenance Windows feature for advanced node control TTL (#1302)

albertjunkim:
**1. Issue, if available:**
#1302 - Maintenance Windows
#1738 

The change introduces ```maintenanceWindows``` in Provisioner, allowing users to control node disruption in a more expected manner. Users may provide specific time windows with date,time, and duratuon. Then,Karpenter controller will delete nodes after ```ttlSecondsUntilExpired``` is passed if time falls within given window. This feature helps users minimize the unexpected disruptions in running applications with live traffic in the production environment. A typical example would be setting the maintenance window for the weekends or off-hours when the traffic is low. 

Test cases and manual deployment. 

- [x] Yes, issue opened: *link to issue*
- [ ] No


ellistarn:
I'll let @njtran chime in on this one. tl;dr, we have some work in flight on this.

njtran:
Hey @albertjunkim, thanks for the contribution. 

We're thinking about Node Disruption Budgets (think Pod Disruption Budgets for nodes) to drive some of the more complex situations to block node deletion. One way to achieve Maintenance Windows with a Node Disruption Budget could be to use K8s [Cronjobs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) to then update these NDBs as desired. 

All in all, I'm a little fearful of adding these crontabs into the provisioner spec as bloats the Provisioner Spec. We would essentially be trying to re-invent a concept already possible with Kubernetes objects. In addition, adding it into the provisioner spec limits the scope that these maintenance windows can affect. This may be desired behavior, but complex workloads could require cross-provisioner maintenance windows. 

I can't comment on how far along the design for Node Disruption Budgets are, but we're thinking about this to make sure we can get this and other node control decisions right.

njtran:
Hey @albertjunkim, I opened a PR for the node upgrades design [here](https://github.com/aws/karpenter/pull/2569/files). At a high level, the proposed solution would do upgrades serially, and would plug into a mechanism of drift. Relying on the same stabilization (cooldown) windows for consolidation, Karpenter will upgrade serially. In addition, if a node shouldn't be consolidated, a blocking PDB or `do-not-evict` pod will prevent it, and Karpenter will move onto another. 

I'm going to close this for now. If this design doesn't cover your use-case, I'd be happy to discuss this more at length with you and figure out a path forward.

PR #1923, opened 2022-06-12
update provisioner status upon node creation

tzneal:
**1. Issue, if available:**
 N/A

Instead of waiting for kubelet to become ready to report the allocated
resources on the provisioner status, pulll it from the cluster state
so it can be updated as soon as the node is launched.

Deployed to EKS and scale-up/scale-down/

- [X] No


ellistarn (commented):
I was hoping we could deprecate the counter controller entirely in favor of computing node limits in the provisioning loop (depending on how long it adds).

tzneal:
> I was hoping we could deprecate the counter controller entirely in favor of computing node limits in the provisioning loop (depending on how long it adds).

It's just there for user status, we don't use it internally for any reason.

Issue #1888, opened 2022-06-02
Nodes hitting max-volume limit still considered eligible for pods by Karpenter?


diranged:
## Version

**Karpenter**: v0.10/1

**Kubernetes**: v1.22.6-eks-7d68063

## Background

We have certain workloads that mount a lot of PVCs to our EC2 nodes. Because these EC2 nodes are Nitro nodes, they share their "attachment limits" between ENI and EBS volumes. To try to minimize the situation where we run out of ENIs because of EBS volume mounts, we pass in `--volume-attach-limit=15` to our aws-ebs-csi-driver chart, expecting that the chart will set the "attachable-volumes-aws-ebs" status on the nodes to 15. However, as I have reported in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1258, we still see that value set to 39.

## Expected Behavior

To some extent, I expect that Karpenter should be able to monitor for _all_ reasons why a pod may not be scheduled onto a host ... and once it has not been scheduled for long enough, it should try to launch new hardware if one of the provisioners matches the general requirements for the pod.

In this specific case, I expect Karpenter to fully understand the way the aws-ebs-csi-driver works and honor whatever max-volume setting we have applied.

## Actual Behavior

When our nodes have available CPU and Memory, but have hit the 15 volume limit, we expect that Karpenter will understand this and know that it needs to scale up new hardware. However, instead we see the controller go into a loop where it just keeps indicating that these pods should be scheduled against the "existing  capacity":

```
2022-06-02T00:24:18.865Z	INFO	controller	Batched 15 pod(s) in 10.000533501s	{"commit": "10fc37b"}
2022-06-02T00:24:18.875Z	INFO	controller	0 pod(s) will schedule against new capacity, 15 pod(s) against existing capacity	{"commit": "10fc37b"}
2022-06-02T00:24:18.875Z	INFO	controller	Waiting for unschedulable pods	{"commit": "10fc37b"}
```

Here is the pod-event log:
```
Warning  FailedScheduling   6s (x2 over 22s)  default-scheduler   0/13 nodes are available: 1 Too many pods, 1 node(s) had taint {kubernetes.io/arch: arm64}, that the pod didn't tolerate, 1 node(s) were unschedulable, 5 node(s) exceed max volume count, 5 node(s) had taint {nd.group-name: kube-system}, that the pod didn't tolerate.
```
[...]

PR #1927, opened 2022-06-13
track volume mounts per node

### LINKED ISSUE #919: Volume Limits as an InstanceType property
The current state of [InstanceType](https://github.com/aws/karpenter/blob/c3b37d99bdfa84ed5fc28c544d5f02d640b3f70c/pkg/cloudprovider/aws/instancetype.go) accommodates CPU, Memory, Pods and some specific resources.  However, it does not seem to account for [Instance volume limits](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html).

We find that volume limits are typically overlooked or mishandled as a scarce resource by all existing schedulers and cluster scalers.  We have some applications that mount multiple PVCs to a single Pod and this causes problems for schedulers that do not account for the Instance volume attachment limits.

[AWS has low volume attachment limits compared to other cloud providers](https://kubernetes.io/docs/concepts/storage/storage-limits/#dynamic-volume-limits) and this causes significant over-provisioning of CPU/Memory in order to accommodate volume attachment constraints.

According to https://github.com/kubernetes/kubernetes/issues/80967#issuecomment-621447246 this could be considered an ENI resource constraint in AWS, but the [current calculation](https://github.com/aws/karpenter/blob/c3b37d99bdfa84ed5fc28c544d5f02d640b3f70c/pkg/cloudprovider/aws/instancetype.go#L70) only considers ENI resource to be a Pod constraint

We had been using `KUBE_MAX_PD_VOLS` until Kubernetes 1.17 to artificially lower the limits.  With CSI drivers the volume limits are accounted in the Kubernetes scheduler, and therefore cluster-autoscaler - which we are using to workaround this issue.

**Additional context**
It's great to see a new attempt at cluster autoscaling and we're looking forward to seeing if the complexity can be managed while handling the use cases.  We're also curious to see the uptake for cloud providers other than AWS.

### Community Note

tzneal:
**1. Issue, if available:**
Fixes #919 and #1888

track volume mounts per node

Assume nodes can support infinite volumes until they launch. Once the CSI driver is
reporting the current number of mountable volumes, use that value which may require
launching more nodes.

Unit testing & deployed to EKS with my EBS CSI driver configured with `--volume-attach-limit=3`. When the driver created the CSINode Karpenter launched another node for the remaining volumes.

- [X] No


pkg/utils/options/options.go
```diff
@@ -66,6 +67,8 @@ type Options struct {
 	AWSENILimitedPodDensity   bool
 	AWSDefaultInstanceProfile string
 	AWSEnablePodENI           bool
+	// AWSVolumeAttachLimit is the setting of the volume-attach-limit from the ebs CSI driver
```
ellistarn, line 70: Not sure this comment is adding much value beyond what's written on line 48.

pkg/utils/options/options.go
```diff
@@ -45,6 +45,7 @@ func MustParse() Options {
 	flag.BoolVar(&opts.AWSENILimitedPodDensity, "aws-eni-limited-pod-density", env.WithDefaultBool("AWS_ENI_LIMITED_POD_DENSITY", true), "Indicates whether new nodes should use ENI-based pod density")
 	flag.StringVar(&opts.AWSDefaultInstanceProfile, "aws-default-instance-profile", env.WithDefaultString("AWS_DEFAULT_INSTANCE_PROFILE", ""), "The default instance profile to use when provisioning nodes in AWS")
 	flag.BoolVar(&opts.AWSEnablePodENI, "aws-enable-pod-eni", env.WithDefaultBool("AWS_ENABLE_POD_ENI", false), "If true then instances that support pod ENI will report a vpc.amazonaws.com/pod-eni resource")
+	flag.IntVar(&opts.AWSVolumeAttachLimit, "aws-volume-attach-limit", env.WithDefaultInt("AWS_VOLUME_ATTACH_LIMIT", -1), "If set to a value >=0, this is the maximum number of volumes attachable per node and should match the value passed to the EBS CSI driver.")
```
ellistarn, line 48: Worth linking to https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/92feaef772b7c93c7da3a5b282a1f9bac819a319/docs/options.md?

pkg/cloudprovider/types.go
```diff
@@ -74,6 +74,8 @@ type InstanceType interface {
 	// Overhead is the amount of resource overhead expected to be used by kubelet and any other system daemons outside
 	// of Kubernetes.
 	Overhead() v1.ResourceList
+	// MaxVolumes returns the maximum number of volumes that can be mounted to the node.
+	MaxVolumes() int
```
ellistarn, line 78: Can this be expressed via `Resources()`? 
From a real node:
```
    capacity:
      attachable-volumes-aws-ebs: "25"
      cpu: "8"
      ephemeral-storage: 83873772Ki
      hugepages-1Gi: "0"
      hugepages-2Mi: "0"
      memory: 32521312Ki
      pods: "58"
```

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -89,13 +89,17 @@ func (p *InstanceTypeProvider) newInstanceType(ctx context.Context, info *ec2.In
...
-	instanceType.resources = instanceType.computeResources(injection.GetOptions(ctx).AWSEnablePodENI)
+	instanceType.resources = instanceType.computeResources(opts.AWSEnablePodENI)
 	instanceType.overhead = instanceType.computeOverhead()
 	instanceType.requirements = instanceType.computeRequirements()
-	if !injection.GetOptions(ctx).AWSENILimitedPodDensity {
+	if !opts.AWSENILimitedPodDensity {
 		instanceType.maxPods = ptr.Int32(110)
 	}
+	if opts.AWSVolumeAttachLimit >= 0 {
+		instanceType.maxVolumes = ptr.Int32(int32(opts.AWSVolumeAttachLimit))
```
ellistarn, line 101: This seems wrong. IIUC, the EBS CSI driver estimates this if `=-1`. Can we try to approximate?

pkg/controllers/provisioning/scheduling/inflightnode.go
```diff
@@ -35,20 +35,23 @@ type InFlightNode struct {
...
+	maxVolumes         int
 }
 
 func NewInFlightNode(n *state.Node, topology *Topology, startupTaints []v1.Taint, daemonResources v1.ResourceList) *InFlightNode {
 	// the remaining daemonResources to schedule are the total daemonResources minus what has already scheduled
 	remainingDaemonResources := resources.Subtract(daemonResources, n.DaemonSetRequested)
-
 	node := &InFlightNode{
 		Node:          n.Node,
+		maxVolumes:    n.InstanceType.MaxVolumes(),
```
ellistarn, line 48: Does it not make sense to store the entire state.Node if we're using multiple piece of data from it?

pkg/controllers/provisioning/scheduling/inflightnode.go
```diff
@@ -87,9 +90,18 @@ func (n *InFlightNode) Add(pod *v1.Pod) error {
 		return err
 	}
 
-	if err := n.hostPortUsage.Add(pod); err != nil {
+	if err := n.hostPortUsage.Validate(pod); err != nil {
+		return err
+	}
+
+	// determine the number of volumes that will be mounted if the pod schedules
+	mountedVolumeCount, err := n.volumeUsage.Validate(pod)
```
ellistarn, line 98: It's not clear to me why this can't be modeled as part of the node's capacity.

pkg/scheduling/volumelimits.go
```diff
@@ -0,0 +1,94 @@
...
+		ctx:        ctx,
+		volumes:    sets.NewString(),
+		podVolumes: map[types.NamespacedName]sets.String{},
+	}
+}
+
+func (v *VolumeLimits) Add(pod *v1.Pod) {
+	podVolumes, err := v.validate(pod)
+	if err != nil {
+		logging.FromContext(v.ctx).Errorf("inconsistent state registering volume li mit, %s", err)
```
ellistarn, line 48: This seems like an antipattern to have all of these ERROR messages complaining about invariants. What's a customer to do about it? Can we structure the code such that this simply can't happen?

pkg/scheduling/volumelimits.go
```diff
@@ -0,0 +1,94 @@
...
+	v.volumes = v.volumes.Union(podVolumes)
+}
+
+func (v *VolumeLimits) Validate(pod *v1.Pod) (int, error) {
+	podVolumes, err := v.validate(pod)
+	result := v.volumes.Union(podVolumes)
+	return len(result), err
+}
+
+func (v *VolumeLimits) validate(pod *v1.Pod) (sets.String, error) {
```
ellistarn, line 60: Looks like this can never return an error?

pkg/scheduling/volumelimits.go
```diff
@@ -0,0 +1,94 @@
...
+	podPVCs := sets.String{}
+	for _, volume := range pod.Spec.Volumes {
+		if volume.PersistentVolumeClaim != nil {
+			podPVCs.Insert(fmt.Sprintf("%s/%s", pod.Namespace, volume.PersistentVolumeClaim.ClaimName))
+		} else if volume.Ephemeral != nil {
+			// generated name per https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#persistentvolumeclaim-naming
+			pvcName := fmt.Sprintf("%s-%s", pod.Name, volume.Name)
+			podPVCs.Insert(fmt.Sprintf("%s/%s", pod.Namespace, pvcName))
```
ellistarn, line 68: Why is this one not inlined, but the one on line 64 is?

pkg/controllers/provisioning/scheduling/inflightnode.go
```diff
@@ -35,20 +35,23 @@ type InFlightNode struct {
 	requirements       scheduling.Requirements
 	available          v1.ResourceList
 	startupTolerations []v1.Toleration
-	hostPortUsage      *state.HostPortUsage
+	hostPortUsage      *scheduling.HostPortUsage
```
ellistarn, line 38: Why did we move these into the top level `scheduling` package? Seems like it may be better encapsulated in this package?

pkg/controllers/provisioning/scheduling/inflightnode.go
tzneal: It's not [reported correctly by the EBS CSI plugin](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1258) on the node and pods don't request volumes so it doesn't get reduced in the capacity.

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -91,7 +94,12 @@ func (n *Node) Add(pod *v1.Pod) error {
 
 	// Check instance type combinations
 	requests := resources.Merge(n.requests, resources.RequestsForPods(pod))
-	instanceTypes := filterInstanceTypes(n.InstanceTypeOptions, nodeRequirements, requests)
+	// determine the number of volumes that will be mounted if the pod schedules
+	mountedVolumeCount, err := n.volumeUsage.Validate(pod)
+	if err != nil {
+		return err
+	}
+	instanceTypes := filterInstanceTypes(n.InstanceTypeOptions, nodeRequirements, requests, mountedVolumeCount)
```
ellistarn, line 102: I'd expect the volume count to be modeled using the resource requests.

pkg/cloudprovider/types.go
tzneal: See below, that number is incorrect.

pkg/controllers/provisioning/scheduling/node.go
tzneal: They're pretty different.  Pods don't have volumes in their resource requests, they are implicit based off of volumes that are mounted.  They are also shared, ten pods can use the same PVC which will reduce the number of volumes by one since only one volume has been mounted.

pkg/controllers/provisioning/scheduling/inflightnode.go
tzneal: They're both used in pkg/controllers/state/ and pkg/controllers/provisioning/scheduling/ so I thought they fit better at the higher level pkg/scheduling/.

pkg/scheduling/volumelimits.go
tzneal: No reason, I'll merge.

pkg/scheduling/volumelimits.go
tzneal: It did previously, I was validating the PVC and volume, but decided not to check the volume.  Removing the error.

pkg/controllers/provisioning/scheduling/inflightnode.go
tzneal: A few of the pieces we use are intentionally deep copied.

pkg/scheduling/volumelimits.go
tzneal: This particular one is gone, there is still one or two remaining in the host port usage.  It indicates that an assumption we've made has not held true, ideally customers don't see it and if they do we know to re-look at the assumption vs troubleshooting a hard to find logic error.

ellistarn: Can we make the message something like "Invariant violated, please cut an issue"

pkg/controllers/provisioning/scheduling/node.go
ellistarn: Is `node.capacity[ebs...]` simply node used then?

pkg/controllers/provisioning/scheduling/inflightnode.go
```diff
@@ -35,20 +35,23 @@ type InFlightNode struct {
...
+	volumeUsage        *scheduling.VolumeLimits
+	maxVolumes         scheduling.VolumeCount
```
ellistarn, line 40:
```suggestion
	volumeLimits         scheduling.VolumeCount
```

pkg/controllers/provisioning/scheduling/inflightnode.go
```diff
@@ -87,9 +90,18 @@ func (n *InFlightNode) Add(pod *v1.Pod) error {
...
+		return err
+	}
+
+	// determine the number of volumes that will be mounted if the pod schedules
+	mountedVolumeCount, err := n.volumeUsage.Validate(pod)
+	if err != nil {
 		return err
 	}
+	if mountedVolumeCount.Exceeds(n.maxVolumes) {
+		return fmt.Errorf("would exceed max volume")
```
ellistarn, line 103: Does this print something useful to the user when unwrapped all the way?
```suggestion
		return fmt.Errorf("would exceed node volume limits")
```

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -91,7 +96,13 @@ func (n *Node) Add(pod *v1.Pod) error {
 
 	// Check instance type combinations
 	requests := resources.Merge(n.requests, resources.RequestsForPods(pod))
```
ellistarn, line 98: Worth moving down to line 104 now w/ the filter?

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -91,7 +96,13 @@ func (n *Node) Add(pod *v1.Pod) error {
...
-	instanceTypes := filterInstanceTypes(n.InstanceTypeOptions, nodeRequirements, requests)
+	// determine the number of volumes that will be mounted if the pod schedules
+	mountedVolumeCount, err := n.volumeUsage.Validate(pod)
```
ellistarn, line 100: We call this variable `mountedVolumeCount` and `volumesPerNode` in different parts of the code. Can we simplify to a single concept to minimize cognitive load for the reader? Perhaps just `volumeCount` with "mounted" and "node" implied.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -196,7 +200,7 @@ func (s *Scheduler) add(pod *v1.Pod) error {
 			}
 		}
 
-		node := NewNode(nodeTemplate, s.topology, s.daemonOverhead[nodeTemplate], instanceTypes)
+		node := NewNode(s.ctx, s.kubeClient, nodeTemplate, s.topology, s.daemonOverhead[nodeTemplate], instanceTypes)
```
ellistarn, line 203: Why are we using a different context object here? How does this differ from `ctx` on line 103?

pkg/scheduling/volumelimits.go
```diff
@@ -0,0 +1,202 @@
...
+			existing = sets.NewString()
+			cp[k] = existing
+		}
+		existing.Insert(v.List()...)
+	}
+	return cp
+}
+
+func (u *volumeUsage) insert(c volumeUsage) {
+	for k, v := range c {
```
ellistarn, line 68: I'm having a hard time reading this. What's `c` stand for?

pkg/scheduling/volumelimits.go
```diff
@@ -0,0 +1,202 @@
...
+type VolumeLimits struct {
+	ctx        context.Context
+	volumes    volumeUsage
+	podVolumes map[types.NamespacedName]volumeUsage
+	kubeClient client.Client
+}
+
+type volumeUsage map[string]sets.String
+
+func (u *volumeUsage) Add(provisioner string, pvcID string) {
```
ellistarn, line 42: IIUC, since you type is itself a pointer (i.e. map), you don't need to use pointer receivers.
```suggestion
func (u volumeUsage) Add(provisioner string, pvcID string) {
```

pkg/scheduling/volumelimits.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (u *volumeUsage) copy() volumeUsage {
+	cp := volumeUsage{}
+	for k, v := range *u {
+		cp[k] = sets.NewString(v.List()...)
+	}
+	return cp
+}
+
+func NewVolumeLimits(ctx context.Context, kubeClient client.Client) *VolumeLimits {
```
ellistarn, line 86: likewise
```suggestion
func NewVolumeLimits(ctx context.Context, kubeClient client.Client) VolumeLimits {
```

pkg/controllers/provisioning/scheduling/inflightnode.go
```diff
@@ -87,9 +90,18 @@ func (n *InFlightNode) Add(pod *v1.Pod) error {
 		return err
 	}
 
-	if err := n.hostPortUsage.Add(pod); err != nil {
+	if err := n.hostPortUsage.Validate(pod); err != nil {
```
ellistarn, line 93: Since we're taking a copy, doesn't it simplify things to collapse Add and Validate into the same method like we had before?

```if err := n.hostPortUsage.Add(pod); err != nil {`

File: pkg/controllers/provisioning/scheduling/node.go (hunk 15 above)
Comment (tzneal) [reply pkg/controllers/provisioning/scheduling/node.go:100]: The results of `Validate()` is called `mountedVolumeCount` both times it's used.  This is the number of volumes that would be mounted on the node if the pod were added.  I think renaming this to volumeCount is slightly ess clear.  I renamed `volumesPerNode` to `mountedVolumeCount`.

File: pkg/controllers/provisioning/scheduling/scheduler.go (hunk 16 above)
Comment (tzneal) [reply pkg/controllers/provisioning/scheduling/scheduler.go:203]: It just wasn't passed down and was only used for logging. Modified to pass it through add().

File: pkg/scheduling/volumelimits.go (hunk 17 above)
Comment (tzneal) [reply pkg/scheduling/volumelimits.go:68]: collection, just renamed it to volumes.

File: pkg/controllers/provisioning/scheduling/inflightnode.go (hunk 20 above)
Comment (tzneal) [reply pkg/controllers/provisioning/scheduling/inflightnode.go:93]: No, this introduces a bug that can cause us to create more nodes than necessary.  The pods are registered to host port usage, and then if a few lines later the pod is rejected for scheduling by requirements or instance type compatibility issues we would need to then remove it from the host port usage.  We could do that, but there's a few places and it's cleaner to compatibility check everything with respect to scheduling, and if it all passes, commit the changes.

File: pkg/scheduling/hostportusage.go [hunk 21]
```diff
@@ -56,22 +59,36 @@ func (e entry) matches(rhs entry) bool {
...
 
-func NewHostPortUsage() *HostPortUsage {
-	return &HostPortUsage{}
+func NewHostPortUsage(ctx context.Context) *HostPortUsage {
+	return &HostPortUsage{ctx: ctx}
 }
 
 // Add adds a port to the HostPortUsage, returning an error in the case of a conflict
-func (u *HostPortUsage) Add(pod *v1.Pod) error {
+func (u *HostPortUsage) Add(pod *v1.Pod) {
```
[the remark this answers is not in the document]
tzneal: Modified, I started this code in cluster state which had a stored context already for the same reason provisioner does.

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -293,6 +300,36 @@ func (i *InstanceType) eniLimitedPods() int64 {
 	return *i.NetworkInfo.MaximumNetworkInterfaces*(*i.NetworkInfo.Ipv4AddressesPerInterface-1) + 2
 }
 
+func (i *InstanceType) computeVolumeLimits(ebsVolumeAttachLimit int) map[string]int {
+	result := map[string]int{
```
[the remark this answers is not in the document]
tzneal: I wrote it that way to support adding other provisioner limits if needed, my thought was to set the values based on the instance type and then apply the overrides.

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -293,6 +300,36 @@ func (i *InstanceType) eniLimitedPods() int64 {
...
+		ebsCsiProvisioner: 27,
+	}
+	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/volume_limits.html
+	switch i.Name() {
+	case "d3.8xlarge", "d3en.12xlarge":
```
diranged, line 309: Am I right that AWS exposes no API for this?

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -293,6 +300,36 @@ func (i *InstanceType) eniLimitedPods() int64 {
 	return *i.NetworkInfo.MaximumNetworkInterfaces*(*i.NetworkInfo.Ipv4AddressesPerInterface-1) + 2
 }
 
+func (i *InstanceType) computeVolumeLimits(ebsVolumeAttachLimit int) map[string]int {
```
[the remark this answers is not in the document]
tzneal: The downside is that we'll have slower provisioning, but it probably doesn't matter much as its somewhat unlikely to hit the volume limits.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -16,6 +16,8 @@ package scheduling_test
 
 import (
 	"context"
+	"fmt"
+	sv1 "k8s.io/api/storage/v1"
```
ellistarn, line 20: The common pattern I've seen in the k8s ecosystem is to move backwards along the path until there isn't a naming conflict (e.g. storagev1). Being consistent across the codebase with this really helps with search/replace.

tzneal: SGMT

pkg/controllers/state/cluster.go
```diff
@@ -180,7 +181,32 @@ func (c *Cluster) newNode(node *v1.Node) *Node {
 		n.Capacity = n.InstanceType.Resources()
 	}
 	n.Available = resources.Subtract(c.getNodeAllocatable(node, n.Provisioner), resources.Merge(requested...))
-	return n
+}
+
+func (c *Cluster) populateVolumeLimits(node *v1.Node, n *Node) {
+	var csiNode storagev1.CSINode
+	if err := c.kubeClient.Get(c.ctx, client.ObjectKey{Name: node.Name}, &csiNode); err != nil && !errors.IsNotFound(err) {
+		logging.FromContext(c.ctx).Errorf("getting CSINode, %s", err)
```
ellistarn, line 189: Since we're not follow normal error patterns, is it possible that we can get stuck due to a transient networking error?

pkg/controllers/state/cluster.go
```diff
@@ -221,10 +250,12 @@ func (c *Cluster) deleteNode(nodeName string) {
 }
 
 // updateNode is called for every node reconciliation
-func (c *Cluster) updateNode(node *v1.Node) {
+func (c *Cluster) updateNode(ctx context.Context, node *v1.Node) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	c.nodes[node.Name] = c.newNode(node)
+	var err error
+	c.nodes[node.Name], err = c.newNode(ctx, node)
```
ellistarn, line 257: If an error occurs, will we inadvertently set c.nodes[node.Name] to nil? This will cause it to show up when looping over the nodes.

pkg/controllers/state/cluster.go
```diff
@@ -313,14 +345,15 @@ func (c *Cluster) updateNodeUsageFromPod(pod *v1.Pod) {
...
-		if err := c.kubeClient.Get(c.ctx, client.ObjectKey{Name: pod.Spec.NodeName}, &node); err != nil {
-			logging.FromContext(c.ctx).Errorf("getting node, %s", err)
+		if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: pod.Spec.NodeName}, &node); err != nil && !errors.IsNotFound(err) {
+			return fmt.Errorf("getting node, %w", err)
 		}
 
 		// node didn't exist, but creating it will pick up this newly bound pod as well
-		n = c.newNode(&node)
+		var err error
+		n, err = c.newNode(ctx, &node)
```
ellistarn, line 354: same re: setting if err != nil

pkg/scheduling/hostportusage.go
```diff
@@ -61,17 +63,31 @@ func NewHostPortUsage() *HostPortUsage {
 }
 
 // Add adds a port to the HostPortUsage, returning an error in the case of a conflict
-func (u *HostPortUsage) Add(pod *v1.Pod) error {
+func (u *HostPortUsage) Add(ctx context.Context, pod *v1.Pod) {
+	newUsage, err := u.validate(pod)
+	if err != nil {
+		logging.FromContext(ctx).Errorf("invariant violated registering host port usage, %s, please file an issue", err)
```
ellistarn, line 69: I'm wondering if we should do a similar error propagation pattern here.

ellistarn (dismissed):

pkg/controllers/state/cluster.go
```diff
@@ -179,15 +181,42 @@ func (c *Cluster) newNode(node *v1.Node) *Node {
...
+		n.Provisioner = &provisioner
+	}
+	return nil
 }
 
 // getNodeAllocatable gets the allocatable resources for the node.
-func (c *Cluster) getNodeAllocatable(node *v1.Node, provisioner *v1alpha5.Provisioner) v1.ResourceList {
-	instanceType, err := c.getInstanceType(c.ctx, provisioner, node.Labels[v1.LabelInstanceTypeStable])
+func (c *Cluster) getNodeAllocatable(ctx context.Context, node *v1.Node, provisioner *v1alpha5.Provisioner) v1.ResourceList {
+	instanceType, err := c.getInstanceType(ctx, provisioner, node.Labels[v1.LabelInstanceTypeStable])
```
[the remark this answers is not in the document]
tzneal: If we can't find the instance type, it only matters for not ready nodes.  We then treat them as having no capacity, once kubelet is running, not finding an instance type means something weird has happened (instance type is no longer reported by EC2 API or some other odd error), but it doesn't affect anything so we just log it.

pkg/controllers/state/cluster.go
tzneal: Previously newNode() always returned a node object, even if it failed to populate some portion of it.  I could see it going both ways as some semi-failure modes aren't real failures in most cases (e.g. unable to read the CSINode), but for now I've modified it to return nil in the event of error for consistency sake.

dewjam:
>This might be a silly question, but do we no longer need the [PersistentVolumeClaim controller](https://github.com/aws/karpenter/blob/main/pkg/controllers/persistentvolumeclaim/controller.go) given we're no longer pre-binding pods?
Discussed with @tzneal offline and the short answer is "yes".

pkg/controllers/state/cluster.go
```diff
@@ -253,16 +280,18 @@ func (c *Cluster) updateNodeUsageFromPodDeletion(podKey types.NamespacedName) {
...
 	// with this separately, but if a daemonset pod is being deleted, it usually means the node is going down.  In the
 	// worst case we will resync to correct this.
 }
 
 // updatePod is called every time the pod is reconciled
-func (c *Cluster) updatePod(pod *v1.Pod) {
-	c.updateNodeUsageFromPod(pod)
+func (c *Cluster) updatePod(ctx context.Context, pod *v1.Pod) error {
+	err := c.updateNodeUsageFromPod(ctx, pod)
 	c.updatePodAntiAffinities(pod)
```
ellistarn, line 293: Should we update if there's an error?

ellistarn (approved):

pkg/test/storage.go
```diff
@@ -92,10 +95,13 @@ func StorageClass(overrides ...StorageClassOptions) *storagev1.StorageClass {
 	if options.Zones != nil {
 		allowedTopologies = []v1.TopologySelectorTerm{{MatchLabelExpressions: []v1.TopologySelectorLabelRequirement{{Key: v1.LabelTopologyZone, Values: options.Zones}}}}
 	}
+	if options.Provisioner == nil {
```
ellistarn, line 98: Consider just checking "" and not making options.Provisioner a pointer to avoid type juggling.

PR #1932, opened 2022-06-14
chore: added rules to a chart which dynamicly appear

AndrewChubatiuk:
**Description**
When using ArgoCD both `defaulting.webhook.provisioners.karpenter.sh` and `validation.webhook.provisioners.karpenter.sh` are always in outofsync state because awsnodetemplates rules appear dynamicly after helm chart deployment

Just need to deploy a helm chart

**Release Note**
```release-note

```


ellistarn (approved):
nice fix! I believe this happens from knative/pkg's webhook reconcilers.

PR #1943, opened 2022-06-16
chore: added missed DELETE method for provisioner

AndrewChubatiuk:
**Description**
Missed DELETE method in #1932

### DIFF
--- a/charts/karpenter/templates/webhooks.yaml
+++ b/charts/karpenter/templates/webhooks.yaml
@@ -41,6 +41,7 @@ webhooks:
         operations:
           - CREATE
           - UPDATE
+          - DELETE
 ---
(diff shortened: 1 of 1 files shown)


ellistarn (dismissed):

charts/karpenter/templates/webhooks.yaml
```diff
@@ -41,6 +41,7 @@ webhooks:
         operations:
           - CREATE
           - UPDATE
+          - DELETE
```
suket22, line 44: The commit says `for awsnodetemplates` but I think this is for the `provisioners` resource. Could you update the commit message and we'll merge it in

AndrewChubatiuk: done

ellistarn (approved):

PR #1946, opened 2022-06-16
docs: add tool to generate instance type docs

tzneal:
Fixes # 1329

**Description**

Displays both labels and the resources that we use for bin-packing after overhead calculations.

* Preview with netlify.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/hack/docs/instancetypes_gen_docs.go
+++ b/hack/docs/instancetypes_gen_docs.go
@@ -1,0 +1,160 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
+	"github.com/aws/karpenter/pkg/cloudprovider"
+	"github.com/aws/karpenter/pkg/cloudprovider/aws"
+	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
+	"github.com/aws/karpenter/pkg/utils/resources"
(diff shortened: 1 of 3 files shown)


hack/docs/instancetypes_gen_docs.go
```diff
@@ -0,0 +1,152 @@
...
+weight: 100
+
+description: >
+  Evaluate Instance Type Resources
+---
+`)
+	fmt.Fprintln(f, "<!-- this document is generated from hack/docs/instancetypes_gen_docs.go -->")
+	fmt.Fprintln(f, `AWS instance types offer varying resources and can be selected by labels. In particular the
+ephemeral-storage is the default value for an AL2 family assuming no specific block device mappings have been configured.`)
+
```
bwagner5, line 72: Might be worth calling out the the resources have overhead taken out.  Although it looks like ephemeral storage does not and maybe not CPU either?  https://github.com/aws/karpenter/blob/205c4db5743f1f6824253e92304a5df2f875c90c/pkg/cloudprovider/aws/amifamily/al2.go#L94

Might also be worth calling out that pods are assuming the `aws-eni-limited-pod-density = true`

bwagner5 (commented):
NICE! just one comment

hack/docs/instancetypes_gen_docs.go
tzneal: SGTM

bwagner5 (approved):
lgtm! Big improvement! to docs!

PR #1951, opened 2022-06-16
feat: Include all Provider fields in the AWSNodeTemplate CRD

suket22:
**Description**
As part of releasing custom UserData support, this moves all of the provider fields into ProviderRef as well. This should allow everyone to specify everything related to the AWS provider within the AWSNodeTemplate CRD. 

Made a couple more changes while I was on this - 
* Got rid of ProviderRefNamespace - dunno why I did this given the CRD was cluster scoped anyway. 
* Fixed the CRD. I had set `paths` to `awsnodetemplate` which is the annotation kubebuilder is looking for to determine the plural name of the CRD. Anyone running a pre-release version of Karpenter should be unaffected as long as they update the charts. (new CRD works, and old provisioner yaml is fine). 

Tested a few things on this front - 
* Made sure that the validation webhook is working on the Provider fields inside the AWSNodeTemplate CRD.
* Made sure you can't specify provider and providerRef. 
* Validated that old behavior works (spinning up AL2 and Bottlerocket nodes)
* Validated that new behavior works (spinning up AL2 and Bottlerocket nodes with UserData using the new CRD).

Example Manifest to try this change out - 
```
kind: Provisioner
metadata:
  name: default
spec:
  limits:
    resources:
      cpu: 20
  providerRef:
    name: default
  ttlSecondsAfterEmpty: 30
---
kind: AWSNodeTemplate
metadata:
  name: default
spec:
  amiFamily: Bottlerocket
  instanceProfile: KarpenterNodeInstanceProfile-karp-cluster
  subnetSelector:
    karpenter.sh/discovery: karp-cluster
  securityGroupSelector:
    karpenter.sh/discovery: karp-cluster
  userData:  |
    [settings.kubernetes]
    kube-api-qps = 30
```
    
    
*Docs will be added in a subsequent PR since this change is large enough as-is.*
I'll need to add - 
1. Some examples on CRD and Provisioner Usage. 
2. Undo the changes that hide the user data documentation. 
3. Additions to the upgrade guide. 

With that, a separate PR makes sense.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
feat: Include all Provider fields in the AWSNodeTemplate CRD
```


pkg/cloudprovider/aws/apis/v1alpha1/provider_validation.go
```diff
@@ -44,16 +43,9 @@ var (
...
 
-func (a *AWS) Validate(provisioner v1alpha5.Provisioner) (errs *apis.FieldError) {
+func (a *AWS) Validate() (errs *apis.FieldError) {
 	return errs.Also(
 		a.validate().ViaField("provider"),
-		a.validateProvisioner(provisioner),
-	)
-}
-
-func (a *AWS) validateProvisioner(provisioner v1alpha5.Provisioner) (errs *apis.FieldError) {
```
suket22, line 54: This semantic doesn't work with ProviderRef. We can't have a validating webhook validate the AWSNodeTemplate CRD based on `kubeletConfiguration` which is a property of the Provisioner CRD. 

This check was pretty shallow in any case (making sure you don't pass `dockerd` with Bottlerocket so I don't think we lose too much by removing this).

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -96,11 +101,21 @@ func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *Cloud
...
 		},
+		kubeClient: options.KubeClient,
 	}
 }
 
 // Create a node given the constraints.
 func (c *CloudProvider) Create(ctx context.Context, nodeRequest *cloudprovider.NodeRequest) (*v1.Node, error) {
+	if nodeRequest.Template.ProviderRef != nil {
+		var awsnodetemplate awsv1.AWSNodeTemplate
+		providerRef := nodeRequest.Template.ProviderRef
```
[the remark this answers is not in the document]
suket22: Sure, I'll inline it.

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -926,11 +928,14 @@ var _ = Describe("Allocation", func() {
 					}
 					nodeTemplate := test.AWSNodeTemplate(test.AWSNodeTemplateOptions{
 						UserData:   aws.String(string(content)),
+						AWS:        provider,
 						ObjectMeta: metav1.ObjectMeta{Name: providerRefName}})
 					ExpectApplied(ctx, env.Client, nodeTemplate)
+					env.Client.Get(ctx, client.ObjectKeyFromObject(nodeTemplate), nodeTemplate)
```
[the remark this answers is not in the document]
suket22: Ha! Missed removing this line from my test. I was losing my mind in the middle when I'd apply the node template but the tests wouldn't recognize it and realize I hadn't run codegen :|

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -109,8 +124,17 @@ func (c *CloudProvider) Create(ctx context.Context, nodeRequest *cloudprovider.N
 }
 
 // GetInstanceTypes returns all available InstanceTypes
-func (c *CloudProvider) GetInstanceTypes(ctx context.Context, provider *v1alpha5.Provider) ([]cloudprovider.InstanceType, error) {
-	awsprovider, err := v1alpha1.Deserialize(provider)
+func (c *CloudProvider) GetInstanceTypes(ctx context.Context, provisioner *v1alpha5.Provisioner) ([]cloudprovider.InstanceType, error) {
+	if provisioner.Spec.ProviderRef != nil {
```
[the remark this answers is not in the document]
suket22: I actually had written that too but you barely DRY out a couple lines. Plus you have to repeat the err handling so it doesn't even get any shorter. I'll do it anyway.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -96,11 +101,21 @@ func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *Cloud
...
 	}
 }
 
 // Create a node given the constraints.
 func (c *CloudProvider) Create(ctx context.Context, nodeRequest *cloudprovider.NodeRequest) (*v1.Node, error) {
+	if nodeRequest.Template.ProviderRef != nil {
+		var awsnodetemplate awsv1.AWSNodeTemplate
+		providerRef := nodeRequest.Template.ProviderRef
+		if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: providerRef.Name}, &awsnodetemplate); err != nil {
+			logging.FromContext(ctx).Errorf("retrieving provider reference, %s", err)
```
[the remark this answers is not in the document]
suket22: Good point, removing this.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -96,11 +101,21 @@ func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *Cloud
...
+			return nil, err
```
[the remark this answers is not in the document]
suket22: Done. I think this looks better now - 
```
provisioning    provisioning/provisioner.go:101 Provisioning failed, getting instance types, retrieving provider reference awsnodetemplates.karpenter.k8s.aws "foo" not found
```

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -36,7 +37,10 @@ import (
 	"github.com/aws/karpenter/pkg/utils/injection"
 	"github.com/aws/karpenter/pkg/utils/project"
 
+	k8sClient "sigs.k8s.io/controller-runtime/pkg/client"
```
ellistarn, line 40: this is just called client elsewhere. Any reason to deviate?

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -96,11 +101,19 @@ func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *Cloud
...
 			),
 		},
+		kubeClient: options.KubeClient,
 	}
 }
 
 // Create a node given the constraints.
 func (c *CloudProvider) Create(ctx context.Context, nodeRequest *cloudprovider.NodeRequest) (*v1.Node, error) {
+	if nodeRequest.Template.ProviderRef != nil {
+		aws, err := c.getProvider(ctx, nodeRequest.Template.ProviderRef.Name)
```
ellistarn, line 111: Thoughts on encapsulating `if nodeRequest.Template.ProviderRef != nil {` inside c.getProvider?

You've duplicated some logic around 
`c.instanceProvider.Create(ctx, aws, nodeRequest)`

ellistarn: Essentially, c.getProvider returns you an `*AWS` from the node template, and the caller figures out what to do with it.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -109,8 +122,15 @@ func (c *CloudProvider) Create(ctx context.Context, nodeRequest *cloudprovider.N
 }
 
 // GetInstanceTypes returns all available InstanceTypes
-func (c *CloudProvider) GetInstanceTypes(ctx context.Context, provider *v1alpha5.Provider) ([]cloudprovider.InstanceType, error) {
-	awsprovider, err := v1alpha1.Deserialize(provider)
+func (c *CloudProvider) GetInstanceTypes(ctx context.Context, provisioner *v1alpha5.Provisioner) ([]cloudprovider.InstanceType, error) {
+	if provisioner.Spec.ProviderRef != nil {
```
ellistarn, line 126: Similar deduplication is possible here.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -200,3 +223,11 @@ func getCABundle(ctx context.Context) *string {
 	logging.FromContext(ctx).Debugf("Discovered caBundle, length %d", len(transportConfig.TLS.CAData))
 	return ptr.String(base64.StdEncoding.EncodeToString(transportConfig.TLS.CAData))
 }
+
+func (c *CloudProvider) getProvider(ctx context.Context, providerRefName string) (*v1alpha1.AWS, error) {
+	var ant awsv1alpha1.AWSNodeTemplate
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: providerRefName}, &ant); err != nil {
+		return nil, fmt.Errorf("retrieving provider reference %w", err)
```
ellistarn, line 230: Typical error annotation uses a comma before the next piece of the error. Similarly, I'd keep the concepts lined up for simplicity. Customers are already aware of `providerRef` since it's part of the API, so better to make it apparent that it's the same concept. Similarly, `get` means something, so might as well reuse the concept.
```suggestion
		return nil, fmt.Errorf("getting providerRef, %w", err)
```

pkg/test/environment.go
```diff
@@ -76,7 +76,7 @@ func NewEnvironment(ctx context.Context, options ...EnvironmentOption) *Environm
 		Environment: envtest.Environment{
 			CRDDirectoryPaths: []string{
 				project.RelativeToRoot("charts/karpenter/crds/karpenter.sh_provisioners.yaml"),
-				project.RelativeToRoot("charts/karpenter/crds/karpenter.k8s.aws_awsnodetemplate.yaml"),
+				project.RelativeToRoot("charts/karpenter/crds/karpenter.k8s.aws_awsnodetemplates.yaml"),
```
ellistarn, line 79: This should just work.
```suggestion
				project.RelativeToRoot("charts/karpenter/crds"),
```

pkg/controllers/state/cluster.go
```diff
@@ -339,11 +339,11 @@ func (c *Cluster) updateNodeUsageFromPod(pod *v1.Pod) {
 }
 
 func (c *Cluster) getInstanceType(ctx context.Context, provisioner *v1alpha5.Provisioner, instanceTypeName string) (cloudprovider.InstanceType, error) {
-	if provisioner == nil || provisioner.Spec.Provider == nil {
+	if provisioner == nil || (provisioner.Spec.Provider == nil && provisioner.Spec.ProviderRef == nil) {
```
ellistarn, line 342: Is this necessary given our validation? We can assume one of these is populated as an invariant of the codebase. 
```suggestion
	if provisioner == nil {
```

suket22: Yeah makes sense. With the earlier check, what was happening was that we'd return early since Provider was indeed nil (since ProviderRef was being used instead) and that was causing in-flight support to break.

ellistarn (approved):

PR #1952, opened 2022-06-17
Remove hardcoded resource values from Helm chart

mogggggg:
**Description**
I've found that most Helm charts tend to leave the `resources` section empty and leave it up to the operator to configure. In it's current state there's no way for me to remove the CPU limit unless I'm using something like Kustomize (which I'm not 😢 ) as omitting the limit forces it to use the default.

```bash
cd charts/karpenter
helm template -s templates/deployment.yaml .
```

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Removed hardcoded resources from Helm chart values
```

### DIFF
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -95,13 +95,8 @@ controller:
   #   value: eu-west-1
 
   # -- Resources for the controller pod.
-  resources:
-    requests:
-      cpu: 1
-      memory: 1Gi
-    limits:
-      cpu: 1
-      memory: 1Gi
+  resources: {}
+
   # -- Controller log level, defaults to the global log level
   logLevel: ""
   # -- Controller log encoding, defaults to the global log encoding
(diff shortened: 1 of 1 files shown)


vadimgusev-codefresh:
@mogggggg not denying your PR, but just a quick tip: `resources.requests.cpu: null` to remove the built-ins.

ellistarn:
I'm a bit wary of this resourceless controllers may pack anywhere any OOM or resource starve. Karpenter requires nontrivial memory and compute to operate, unlike other more simple operators. It's definitely the case that the default value are too high for most use cases. Can you provide examples of controllers that do this?

mogggggg:
I wasn't actually aware that you could override the defaults by setting `null` so that's my original issue solved! Two controllers that come to mind that don't set resources are sealed-secrets and prometheus-operator, although compared to Karpenter they do use less resources.

I'll close this issue.

PR #1953, opened 2022-06-17
feat: make vm memory overhead configurable

bwagner5:
Fixes # 

**Description**
 - Custom AMIs require different VM memory overhead due to other software or agents running on the machine. This PR makes the VM memory overhead parameter configurable via a CLI arg or env var.
 - I've also added a script to generate a table with the env var / cli flag  and usage description for configuring the karpenter controller

* `make docgen` and observed the overhead calculation is the same as before.
* Varied the vm overhead parameter between 0, 1, and 0.075, observing correct docs gen of memory. 

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
VM Memory Overhead is now configurable via an environment variable to the Karpenter controller (see the options here: https://karpenter.sh/preview/tasks/configuration/)
```

### DIFF
--- a/hack/docs/instancetypes_gen_docs.go
+++ b/hack/docs/instancetypes_gen_docs.go
@@ -15,6 +15,8 @@ import (
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/cloudprovider/aws"
 	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
+	"github.com/aws/karpenter/pkg/utils/injection"
+	"github.com/aws/karpenter/pkg/utils/options"
 	"github.com/aws/karpenter/pkg/utils/resources"
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
@@ -31,7 +33,9 @@ func main() {
 
 	os.Setenv("AWS_SDK_LOAD_CONFIG", "true")
 	os.Setenv("AWS_REGION", "us-east-1")
-	ctx := context.Background()
+	os.Setenv("CLUSTER_NAME", "docs-gen")
+	os.Setenv("CLUSTER_ENDPOINT", "https://docs-gen.aws")
+	ctx := injection.WithOptions(context.Background(), options.MustParse())
 
 	cp := aws.NewCloudProvider(ctx, cloudprovider.Options{
 		ClientSet:  nil,
--- a/hack/docs/instancetypes_gen_docs.go
+++ b/hack/docs/instancetypes_gen_docs.go
@@ -35,7 +35,7 @@ func main() {
 	os.Setenv("AWS_REGION", "us-east-1")
 	os.Setenv("CLUSTER_NAME", "docs-gen")
 	os.Setenv("CLUSTER_ENDPOINT", "https://docs-gen.aws")
-	ctx := injection.WithOptions(context.Background(), options.MustParse())
+	ctx := injection.WithOptions(context.Background(), options.New().MustParse())
 
 	cp := aws.NewCloudProvider(ctx, cloudprovider.Options{
 		ClientSet:  nil,
--- a/pkg/utils/env/env.go
+++ b/pkg/utils/env/env.go
@@ -33,6 +33,20 @@ func WithDefaultInt(key string, def int) int {
 	return i
 }
 
+// WithDefaultFloat64 returns the float64 value of the supplied environment variable or, if not present,
+// the supplied default value. If the float64 conversion fails, returns the default
+func WithDefaultFloat64(key string, def float64) float64 {
+	val, ok := os.LookupEnv(key)
+	if !ok {
+		return def
+	}
+	f, err := strconv.ParseFloat(val, 64)
+	if err != nil {
+		return def
+	}
+	return f
+}
+
 // WithDefaultString returns the string value of the supplied environment variable or, if not present,
 // the supplied default value.
 func WithDefaultString(key string, def string) string {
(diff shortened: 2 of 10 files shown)


hack/docs/instancetypes_gen_docs.go
```diff
@@ -31,7 +33,9 @@ func main() {
 
 	os.Setenv("AWS_SDK_LOAD_CONFIG", "true")
 	os.Setenv("AWS_REGION", "us-east-1")
-	ctx := context.Background()
+	os.Setenv("CLUSTER_NAME", "docs-gen")
+	os.Setenv("CLUSTER_ENDPOINT", "https://docs-gen.aws")
```
ellistarn, line 37: :O What's this?

pkg/utils/env/env.go
```diff
@@ -33,6 +33,20 @@ func WithDefaultInt(key string, def int) int {
 	return i
 }
 
+// WithDefaultFloat64 returns the float64 value of the supplied environment variable or, if not present,
+// the supplied default value. If the float64 conversion fails, returns the default
+func WithDefaultFloat64(key string, def float64) float64 {
```
ellistarn, line 38: Is there a generics way to do this?

pkg/utils/options/options.go
```diff
@@ -31,41 +32,57 @@ const (
...
 	KubeClientQPS             int
 	KubeClientBurst           int
+	VMMemoryOverhead          float64
 	AWSNodeNameConvention     string
 	AWSENILimitedPodDensity   bool
 	AWSDefaultInstanceProfile string
 	AWSEnablePodENI           bool
+
+	// Flagset is the backing flagset that is used to fill in the Options fields
+	Flagset *flag.FlagSet
```
ellistarn, line 50: Worth subclassing?

type Options struct {
  *flag.FlagSet
   ...
}

pkg/utils/options/options.go
```diff
@@ -31,41 +33,58 @@ const (
...
+	f.StringVar(&opts.ClusterEndpoint, "cluster-endpoint", env.WithDefaultString("CLUSTER_ENDPOINT", ""), "The external kubernetes cluster endpoint for new nodes to connect with")
+	f.IntVar(&opts.MetricsPort, "metrics-port", env.WithDefaultInt("METRICS_PORT", 8080), "The port the metric endpoint binds to for operating metrics about the controller itself")
+	f.IntVar(&opts.HealthProbePort, "health-probe-port", env.WithDefaultInt("HEALTH_PROBE_PORT", 8081), "The port the health probe endpoint binds to for reporting controller health")
+	f.IntVar(&opts.KubeClientQPS, "kube-client-qps", env.WithDefaultInt("KUBE_CLIENT_QPS", 200), "The smoothed rate of qps to kube-apiserver")
+	f.IntVar(&opts.KubeClientBurst, "kube-client-burst", env.WithDefaultInt("KUBE_CLIENT_BURST", 300), "The maximum allowed burst of queries to the kube-apiserver")
+	f.Float64Var(&opts.VMMemoryOverhead, "vm-memory-overhead", env.WithDefaultFloat64("VM_MEMORY_OVERHEAD", 0.075), "The VM memory overhead as a percent that will be subtracted from the total memory for all instance types")
```
suket22, line 65: I remember you doing some testing on this - did we want to tune 0.075 any further or is that still the best approximation we've got by default?

Have we also given up trying to do this per instance type?

hack/docs/instancetypes_gen_docs.go
bwagner5: just dummy params to allow for flag parsing to take place. This allowed me to plug in other params that could change the generated instance types resources like max-pods and vm overhead easily. Figured I'd leave in how I did it so that if we need to or want to change some of those parameters, it's a simple `os.Setenv(...`

pkg/utils/options/options.go
bwagner5: Testing found `0.075` to be very close to the best param for AL2, but this will at least give users control if they want to tune it themselves which could make sense if the instance types are constrained too. 

I gave up on per instance type, there's too much variability and we could be wrong as new ones are released, the percentage is a conservative, safe value.

pkg/utils/env/env.go
bwagner5: I don't think so.  The float64 func uses `strconv.ParseFloat`, the Bool one uses strconv.ParseBool and the Int one uses `strconv.Atoi`.  I think the only way you could do it is to use a generic return type and accept a default interface{} as the parameter so that you could do a type switch:

```
func WithDefault[V float64 | int | string | bool](key string, def V) V {
    var err error
    var p V
    val, ok := os.LookupEnv(key)
    if !ok {
        return def
    }
    switch def.(type) {
        case float64:
           p, err = strconv.ParseFloat(val, 64)
        case int:
           p, err = strconv.Atoi(val)
        case bool:
           p, err = strconv.ParseBool(val)
       default:
           panic("not supported")
    }
    if err != nil {
        return def.(V)
    }
    return p
}
```

I think they're better as separate funcs :)

ellistarn (approved):

pkg/utils/options/options.go
suket22: As a customer, how do I know what value to be setting here? 

I'd probably see that 0.075 is too high and therefore my nodes aren't as densely packed as they are with just nodegroups + k8s scheduler, but how do I tune this to  say 0.072 etc? Is it just guesswork?

If I choose a really bad number, the worst that will happen is that Karpenter may launch an incorrect number of nodes to satisfy my pending pods but it should eventually converge. Is my understanding correct?

PR #1957, opened 2022-06-20
feat: use real pricing data

tzneal:
Fixes # 

**Description**

Initialize with a list of generated pricing data from us-east-1. Refresh
this list on startup (if possible) and then again periodically. In
profiling the price generation code, main uses about 2.5MB of memory.

* Unit testing
* Deployed to EKS with no permissions to price APIs
* Deployed to EKS with permissions to price APIs

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
action required: Update your KarpenterControllerPolicy to add ec2:DescribeSpotPriceHistory and pricing:GetProducts permissions. If this is not done, Karpenter will still work but will be forced to use its static copy of pricing data which may not include new instance types.
```


Makefile
```diff
@@ -76,6 +76,7 @@ codegen: ## Generate code. Must be run if changes are made to ./pkg/apis/...
 		crd \
 		paths="./pkg/..." \
 		output:crd:artifacts:config=charts/karpenter/crds
+	go run hack/code/prices_gen.go -- pkg/cloudprovider/aws/zz_generated.pricing.go
```
tzneal, line 79: This makes price generation part of the `verify` target, which may not be ideal. The on-demand prices don't update often, but the spot prices do so this will result in some churn on the static price list.  I could also just move the price generation into the release target.

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -46,13 +46,21 @@ type InstanceType struct {
 	resources    v1.ResourceList
 	provider     *v1alpha1.AWS
 	maxPods      *int32
+	price        float64
 }
 
-func NewInstanceType(ctx context.Context, info *ec2.InstanceTypeInfo, provider *v1alpha1.AWS, offerings []cloudprovider.Offering) *InstanceType {
+func NewInstanceType(ctx context.Context, info *ec2.InstanceTypeInfo, pricing *PricingProvider, provider *v1alpha1.AWS, offerings []cloudprovider.Offering) (*InstanceType, error) {
+	// TODO: move pricing information from the instance type down into offerings
```
tzneal, line 53: Leaving this for a future PR.  Our pricing information is at the wrong level at the moment (instance vs offering), it also varies by OS.

pkg/cloudprovider/aws/pricing.go
```diff
@@ -0,0 +1,313 @@
...
+		ServiceCode: aws.String("AmazonEC2")}, p.onDemandPage(prices))
+}
+
+// turning off cyclo here, it measures as a 12 due to all of the type checks of the pricing data which returns a deeply
+// nested map[string]interface{}
+// nolint: gocyclo
+func (p *PricingProvider) onDemandPage(prices map[string]float64) func(output *pricing.GetProductsOutput, b bool) bool {
+	return func(output *pricing.GetProductsOutput, b bool) bool {
+		for _, outer := range output.PriceList {
+			product, ok := outer["product"].(map[string]interface{})
```
tzneal, line 224: I felt bad writing this code.

tzneal:
Existing Karpenter users that haven't added the two new permissions to their controller role will see the following errors on startup and every 12 hours.  Karpenter still works correctly as we fall back to the last collected pricing data when we have an error updating pricing, including using the static scraped pricing data if we can never update.

```
karpenter-69685567db-zc57f controller 2022-06-20T17:13:00.460Z  INFO    controller.aws.pricing  updating pricing information    {"commit": "a49161f"}
karpenter-69685567db-zc57f controller 2022-06-20T17:13:00.846Z  ERROR   controller.aws.pricing  updating on-demand pricing, AccessDeniedException: User: arn:aws:sts::399082912554:assumed-role/tnealt-karpenter-demo-karpenter/1655745180460675630 is not authorized to perform: pricing:GetProducts because no identity-based policy allows the pricing:GetProducts action
karpenter-69685567db-zc57f controller   status code: 400, request id: e5a7bf0b-7821-47a9-a7a8-6c6e696be2e6, using existing pricing data from 20 Jun 22 12:11 -0500      {"commit": "a49161f"}
karpenter-69685567db-zc57f controller 2022-06-20T17:13:00.888Z  ERROR   controller.aws.pricing  updating spot pricing, UnauthorizedOperation: You are not authorized to perform this operation.
karpenter-69685567db-zc57f controller   status code: 403, request id: efc6b7d2-e021-41e5-8786-9f1809c5664d, using existing pricing data from 20 Jun 22 12:11 -0500      {"commit": "a49161f"}
```

pkg/cloudprovider/aws/zz_generated.pricing.go
```diff
@@ -0,0 +1,546 @@
...
+	"x2iedn.metal": 26.676000, "x2iedn.xlarge": 0.833630,
+	// x2iezn family
+	"x2iezn.12xlarge": 10.008000, "x2iezn.2xlarge": 1.668000, "x2iezn.4xlarge": 3.336000,
+	"x2iezn.6xlarge": 5.004000, "x2iezn.8xlarge": 6.672000, "x2iezn.metal": 10.008000,
+	// z1d family
+	"z1d.12xlarge": 4.464000, "z1d.2xlarge": 0.744000, "z1d.3xlarge": 1.116000, "z1d.6xlarge": 2.232000,
+	"z1d.large": 0.186000, "z1d.metal": 4.464000, "z1d.xlarge": 0.372000,
+}
+
+var initialSpotPrices = map[string]float64{
```
bwagner5, line 293: I don't think there's much of a purpose to store static spot pricing data since, unlike OD, it won't be the same relative ranking within different regions. I think it might be better to just assume OD pricing if not able to pull spot pricing.

hack/code/prices_gen.go
```diff
@@ -0,0 +1,123 @@
...
+	"runtime/pprof"
+	"sort"
+	"strings"
+	"time"
+)
+
+func main() {
+	flag.Parse()
+	if flag.NArg() != 1 {
+		log.Printf("Usage: %s -- pkg/cloudprovider/aws/zz_generated.pricing.go", os.Args[0])
```
bwagner5, line 25: why `--` here?

bwagner5 (commented):
nice!  A few comments

pkg/cloudprovider/aws/pricing.go
bwagner5: do you think it would be better to convert to json and then unmarshal to a struct?  You couldn't clean-up all of it with that, but might make the structure easy to see as the reader and limit all the `!ok` checks

hack/code/prices_gen.go
```diff
@@ -0,0 +1,123 @@
...
+		log.Fatalf("formatting generated source, %s", err)
+	}
+
+	if err := ioutil.WriteFile(flag.Arg(0), formatted, 0644); err != nil {
+		log.Fatalf("writing output, %s", err)
+	}
+
+	runtime.GC()
+	if err := pprof.WriteHeapProfile(f); err != nil {
+		log.Fatal("could not write memory profile: ", err)
```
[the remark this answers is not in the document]
tzneal: I added it just so I could see how much memory this was taking, the hack code uses the same code for fetching prices that Karpenter uses.  I don't think it hurts to leave it, the generated file is in the .gitignore already.

hack/code/prices_gen.go
tzneal: Otherwise `go run` thinks that the generated file is part of the compilation package.

pkg/cloudprovider/aws/zz_generated.pricing.go
tzneal: Ah, that makes sense, I didn't think to fall back to OD pricing if we can't get spot. We'd have to reconsider once we support pricing per offering,  maybe spot = od*some discount at that point?

hack/code/prices_gen.go
bwagner5: oooooo because the file ends in `.go` I guess?  Was trying to compare against the docgen args

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -76,7 +78,13 @@ func (p *InstanceTypeProvider) Get(ctx context.Context, provider *v1alpha1.AWS)
 	}
 	var result []cloudprovider.InstanceType
 	for _, i := range instanceTypes {
-		result = append(result, NewInstanceType(ctx, i, provider, p.createOfferings(i, instanceTypeZones[aws.StringValue(i.InstanceType)])))
+		instanceType, err := NewInstanceType(ctx, i, p.pricingProvider, provider, p.createOfferings(i, instanceTypeZones[aws.StringValue(i.InstanceType)]))
```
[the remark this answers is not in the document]
tzneal: The problem is that we don't know what that instance type costs relative to any of the others so we can't order them.    It should be somewhat rare for this to occur (e.g. isolated VPC + old version that doesn't have a new instance type baked in).

pkg/cloudprovider/aws/pricing.go
```diff
@@ -0,0 +1,313 @@
...
+		onDemandPrices:     initialOnDemandPrices,
+		spotUpdateTime:     initialPriceUpdate,
+		spotPrices:         initialSpotPrices,
+		ec2:                ec2Api,
+		pricing:            pricing,
+	}
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("pricing"))
+
+	// perform an initial price update at startup
+	p.updatePricing(ctx)
```
bwagner5, line 82: how long does this take? This blocks reconcilers from starting right?

tzneal: < 10 seconds, but no problem to move it to being parallel and run on the first iteration of the periodic loop. The downside there is you may provision the first time with stale pricing data.  WDYT?

pkg/cloudprovider/aws/pricing.go
tzneal: So we'd go []byte -> map[string]interface{} -> []byte -> new custom struct 😆   I could go either way, as long as the pricing API doesn't change much I think this is ok.  I suspect the API response is basically set in stone or everyone would be breaking.

Makefile
```diff
@@ -42,7 +42,7 @@ deflake:
 battletest: strongertests
 	go tool cover -html coverage.out -o coverage.html
 
-verify: codegen ## Verify code. Includes dependencies, linting, formatting, etc
+verify: ## Verify code. Includes dependencies, linting, formatting, etc
```
ellistarn, line 45: We need to run codegen on `make ci` to ensure we catch any uncommitted changes.

ellistarn:
Can you post a pprof image?

hack/code/prices_gen.go
ellistarn: You could always use something like https://github.com/spf13/cobra to parse the args. Totally optional, since I kinda hate cobra.

Makefile
ellistarn: I see the impl now. We probably need to split this to avoid losing codegen checks for static code.

hack/code/prices_gen.go
```diff
@@ -0,0 +1,123 @@
...
+	fmt.Fprintln(src, "package aws")
+	fmt.Fprintln(src, `import "time"`)
+	now := time.Now().Format(time.RFC822Z)
+	fmt.Fprintf(src, "// generated at %s for %s\n\n\n", time.Now().Format(time.RFC822Z), region)
+	fmt.Fprintf(src, "var initialPriceUpdate, _ = time.Parse(time.RFC822Z, \"%s\")\n", now)
+
+	instanceTypes := pricingProvider.InstanceTypes()
+	sort.Strings(instanceTypes)
+
+	writePricing(src, instanceTypes, "initialOnDemandPrices", pricingProvider.OnDemandPrice)
```
ellistarn, line 54: We could simply hydrate the maps and then dynamically update it, rather than maintaining a concept of initial vs updated.
```suggestion
	writePricing(src, instanceTypes, "DemandPrices", pricingProvider.OnDemandPrice)
```

hack/code/prices_gen.go
```diff
@@ -0,0 +1,123 @@
...
+	writePricing(src, instanceTypes, "initialSpotPrices", pricingProvider.SpotPrice)
```
ellistarn, line 55:
```suggestion
	writePricing(src, instanceTypes, "SpotPrices", pricingProvider.SpotPrice)
```

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -82,7 +82,8 @@ func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *Cloud
 	logging.FromContext(ctx).Debugf("Using AWS region %s", *sess.Config.Region)
 	ec2api := ec2.New(sess)
 	subnetProvider := NewSubnetProvider(ec2api)
-	instanceTypeProvider := NewInstanceTypeProvider(ec2api, subnetProvider)
+	pricingProvider := NewPricingProvider(ctx, NewPricingAPI(sess, *sess.Config.Region), ec2api, *sess.Config.Region)
```
ellistarn, line 85: There's a lot of unnecessary duplication in this. Seems like we need the sess and ec2api.
```suggestion
	pricingProvider := NewPricingProvider(ctx, sess, ec2api)
```

pkg/cloudprovider/aws/pricing.go
ellistarn: Do we know why our current startup time is so high? Ideally I'd like to get it down to a couple seconds, but if it's not possible, it may be worth being more correct.

pkg/cloudprovider/aws/pricing.go
```diff
@@ -0,0 +1,313 @@
...
+// fails, the previous pricing information is retained and used which may be the static initial pricing data if pricing
+// updates never succeed.
+type PricingProvider struct {
+	ec2     ec2iface.EC2API
+	pricing pricingiface.PricingAPI
+	region  string
+
+	mu                 sync.RWMutex
+	onDemandUpdateTime time.Time
+	onDemandPrices     map[string]float64
```
ellistarn, line 48: Thoughts on combining into something like:
```
map[string]Pricing

type struct Pricing {
  lastUpdated time.Time
  offering map[string]float64
}
```

Might be more extensible if there are ever future pricing models other than spot/od.

pkg/cloudprovider/aws/pricing.go
```diff
@@ -0,0 +1,313 @@
...
+	defer p.mu.RUnlock()
+	price, ok := p.spotPrices[instanceType]
+	if !ok {
+		return 0.0, fmt.Errorf("instance type %s not found", instanceType)
+	}
+	return price, nil
+}
+
+func (p *PricingProvider) updatePricing(ctx context.Context) {
+	logging.FromContext(ctx).Infof("updating pricing information")
```
ellistarn, line 130: Most of our logs start with an Upper case letter. Are you advocating for a lower case pattern? Should we consider migrating the rest of the codebase (over time) to this new pattern?

This line needs to be more explicit about EC2 I think. Look at other log lines and try to keep the concepts aligned
```
karpenter-54564ccf6-dqtpv controller 2022-06-20T21:56:57.492Z	DEBUG	controller.provisioning.cloudprovider	Discovered security groups: [sg-000d68e67c3fe53d8]	{"commit": "cd5f75f", "provisioner": "default"}
karpenter-54564ccf6-dqtpv controller 2022-06-20T21:56:57.494Z	DEBUG	controller.provisioning.cloudprovider	Discovered kubernetes version 1.22	{"commit": "cd5f75f", "provisioner": "default"}
karpenter-54564ccf6-dqtpv controller 2022-06-20T21:56:57.600Z	DEBUG	controller.provisioning.cloudprovider	Discovered ami-04db1f008385aae5e for query "/aws/service/eks/optimized-ami/1.22/amazon-linux-2/recommended/image_id"	{"commit": "cd5f75f", "provisioner": "default"}
```

pkg/cloudprovider/aws/pricing.go
```diff
@@ -0,0 +1,313 @@
...
+			Field: aws.String("productFamily"),
+			Type:  aws.String("TERM_MATCH"),
+			Value: aws.String("Compute Instance"),
+		})
+	if err != nil {
+		return err
+	}
+
+	// bare metal on-demand prices
+	err = p.fetchOnDemandPricing(ctx, prices,
```
ellistarn, line 158: Thoughts on parallelizing these update functions to speed startup time?

ellistarn: Same with spot prices.

pkg/cloudprovider/aws/pricing.go
```diff
@@ -0,0 +1,313 @@
...
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.onDemandPrices = prices
+	p.onDemandUpdateTime = time.Now()
+	logging.FromContext(ctx).Infof("updated on-demand pricing with %d instance types", len(p.onDemandPrices))
+	return nil
+}
+
+func (p *PricingProvider) fetchOnDemandPricing(ctx context.Context, prices map[string]float64, additionalFilters ...*pricing.Filter) error {
+	filters := append([]*pricing.Filter{
```
ellistarn, line 181: consider inlining.

pkg/cloudprovider/aws/pricing.go
ellistarn: +1 to structifying this.

pkg/cloudprovider/aws/cloudprovider.go
ellistarn: Oh -- you're using the NewPricingProvider function in tests, rather than building out the struct. Makes sense to me.

pkg/cloudprovider/aws/zz_generated.pricing.go
ellistarn: If we assume OD pricing, we'll still be able to leverage fleet for price selection on provisioning, we just lose the ability to defrag spot until pricing comes back, which seems acceptable. We can also do something like label the instance with its price when its created to avoid bad defrags.

pkg/cloudprovider/aws/pricing.go
tzneal: The pricing APIs aren't very performant.  I've got the price requests filtered as much as I can which helps, and I've just updated them to be in parallel. It takes < 3 seconds now.

```
{"level":"info","ts":1655781861.763466,"logger":"fallback.aws.pricing","caller":"aws/pricing.go:142","msg":"updating pricing information"}
{"level":"info","ts":1655781862.709313,"logger":"fallback.aws.pricing","caller":"aws/pricing.go:231","msg":"fetch pricing 945.664169ms took Compute Instance (bare metal)"}
{"level":"info","ts":1655781864.187744,"logger":"fallback.aws.pricing","caller":"aws/pricing.go:382","msg":"updated spot pricing with 537 instance types"}
{"level":"info","ts":1655781864.187788,"logger":"fallback.aws.pricing","caller":"aws/pricing.go:328","msg":"update spot pricing took 2.424180538s"}
{"level":"info","ts":1655781864.348746,"logger":"fallback.aws.pricing","caller":"aws/pricing.go:231","msg":"fetch pricing 2.585108153s took Compute Instance"}
{"level":"info","ts":1655781864.348879,"logger":"fallback.aws.pricing","caller":"aws/pricing.go:224","msg":"updated on-demand pricing with 545 instance types"}
{"level":"info","ts":1655781864.348892,"logger":"fallback.aws.pricing","caller":"aws/pricing.go:171","msg":"update on-demand pricing took 2.585278096s"}
{"level":"info","ts":1655781864.348908,"logger":"fallback.aws.pricing","caller":"aws/pricing.go:145","msg":"update pricing took 2.585310603s"}
```

hack/code/prices_gen.go
tzneal: It's actually not needed in that usage line, only used with `go run`. I updated the comment.

pkg/cloudprovider/aws/pricing.go
tzneal: It's called twice, one for standard on-demand and once for bare metal.

hack/code/prices_gen.go
tzneal: I didn't want the static data exposed, I track the staleness time intentionally as well.  I plan on using it in a PR at some point when launching spot instances so we don't launch a spot that costs more than the cheapest on-demand.

pkg/cloudprovider/aws/pricing.go
tzneal: I don't consider either to be better than the other. Initial capitalization of a log is the level of detail I don't think it's worth us having an opinion on.

pkg/cloudprovider/aws/instancetypes.go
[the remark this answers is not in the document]
bwagner5: For those cases, could we just order it last and only use it if needed?

pkg/cloudprovider/aws/pricing.go
bwagner5: I suspect if you're requesting pricing from Europe or Asia, it will take substantially longer since there's not a close endpoint

hack/code/prices_gen.go
ellistarn: >I didn't want the static data exposed

Didn't mean to imply this should be public.

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -76,7 +78,13 @@ func (p *InstanceTypeProvider) Get(ctx context.Context, provider *v1alpha1.AWS)
...
+		if err != nil {
+			// this can occur often in an isolated VPC with a stale price list, so we only report it at the debug level
+			logging.FromContext(ctx).Debugf("skipping instance type %s due to %s", aws.StringValue(i.InstanceType), err)
```
ellistarn, line 84: IIUC, printing the string value of an error results in a pretty human unfriendly string. Here's the pattern we use broadly in the code base. 
```suggestion
			logging.FromContext(ctx).Debugf("skipping instance type %s, %s", aws.StringValue(i.InstanceType), err.Error())
```

pkg/cloudprovider/aws/pricing.go
```diff
@@ -0,0 +1,368 @@
...
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+	price, ok := p.spotPrices[instanceType]
+	// if there is no spot price available, fall back to the on-demand price
+	if !ok {
+		price, ok = p.onDemandPrices[instanceType]
+	}
+	if !ok {
+		return 0.0, fmt.Errorf("instance type %s not found", instanceType)
+	}
```
ellistarn, line 137: Sugg: more idiomatic, fewer lines, easier to read
```suggestion
	if price, ok := p.spotPrices[instanceType]; ok {
		return price, nil
	}
	// if there is no spot price available, fall back to the on-demand price
	if price, ok := p.onDemandPrices[instanceType]; ok {
		return price, nil 
	}
	return 0.0, fmt.Errorf("instance type %s not found", instanceType)
```

pkg/cloudprovider/aws/pricing.go
ellistarn: Oh I meant the filters.

ellistarn (commented):
Great changes! A few minor nits/comments.

pkg/cloudprovider/aws/instancetypes.go
tzneal: printf calls err.Error internally automatically for  %v, %s, %x, %X and %q see https://github.com/golang/go/blob/668041ef66ddafffccf1863e6180b83ea1ad30c9/src/fmt/print.go#L651

pkg/cloudprovider/aws/pricing.go
tzneal: I prefer to keep function invocations as succinct as possible.

Makefile
```diff
@@ -82,8 +82,15 @@ docgen: ## Generate docs
 	go run hack/docs/metrics_gen_docs.go pkg/ website/content/en/preview/tasks/metrics.md
 	go run hack/docs/instancetypes_gen_docs.go website/content/en/preview/AWS/instance-types.md
 	go run hack/docs/configuration_gen_docs.go website/content/en/preview/tasks/configuration.md
-     
-release: ## Generate release manifests and publish a versioned container image.
+
+release-gen: docgen ## Generate any materials which should be updated prior to release
```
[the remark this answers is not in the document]
tzneal: We generated new source code, the next step in release is to start building images.  We want to lint at a minimum and it doesn't hurt to run tidy and cache any new modules. I doubt the pricing code will ever import any new module, but there will potentially be future code generation that's performed at release time.

pkg/cloudprovider/aws/pricing.go
tzneal: They could differ if one update succeeds and the other fails.

pkg/cloudprovider/aws/instancetypes.go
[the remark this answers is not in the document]
tzneal: I'll change it to return max float 64 as the price in those cases.  That will both deprioritize it for regular launches and prevent consolidation from ever considering it as a replacement.

pkg/cloudprovider/aws/pricing.go
```diff
@@ -0,0 +1,368 @@
...
+
+	prices := map[string]*pricingInfo{}
+	if err := p.ec2.DescribeSpotPriceHistoryPagesWithContext(ctx, &ec2.DescribeSpotPriceHistoryInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("product-description"),
+				Values: []*string{aws.String("Linux/UNIX")},
+			},
+		},
```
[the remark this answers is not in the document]
tzneal: SGTM, didn't notice that parameter.

pkg/cloudprovider/aws/pricing.go
```diff
@@ -0,0 +1,363 @@
...
+	if sess == nil {
+		return nil
+	}
+	// pricing API doesn't have an endpoint in all regions
+	pricingAPIRegion := "us-east-1"
+	if strings.HasPrefix(region, "cn-") {
+		pricingAPIRegion = "cn-north-1"
+	} else if strings.HasPrefix(region, "ap-") {
+		pricingAPIRegion = "ap-south-1"
+	}
```
suket22, line 72: Not sure I understand this - there's 6 regions that start with `ap` so wondering why we're redirecting all their traffic to `ap-south-1` ? Is this just for lower latencies or something?

pkg/cloudprovider/aws/pricing.go
```diff
@@ -0,0 +1,363 @@
...
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+	if price, ok := p.spotPrices[instanceType]; ok {
+		return price, nil
+	}
+	// if there is no spot price available, fall back to the on-demand price
+	if price, ok := p.onDemandPrices[instanceType]; ok {
+		return price, nil
+	}
+	return 0.0, fmt.Errorf("instance type %s not found", instanceType)
```
suket22, line 137: Hmm, so there's a 12 hour window after an instance is newly launched where Karpenter might fail all launches because the cached Pricing API values don't have that type?

pkg/cloudprovider/aws/pricing.go
```diff
@@ -0,0 +1,363 @@
...
+
+func (p *PricingProvider) updatePricing(ctx context.Context) {
+	logging.FromContext(ctx).Infof("Updating EC2 pricing information")
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if err := p.updateOnDemandPricing(ctx); err != nil {
+			logging.FromContext(ctx).Errorf("updating on-demand pricing, %s, using existing pricing data from %s", err, p.onDemandUpdateTime.Format(time.RFC822Z))
```
suket22, line 148: We don't expect this to work in fully private subnets right? (no IGW or NATGW). I wonder if we'll end up spamming logs in those controllers. 

I like that we're failing open, but should we have a way to turn this off?

pkg/cloudprovider/aws/pricing.go
tzneal: The only two endpoints listed at https://docs.aws.amazon.com/general/latest/gr/billing.html for pricing are us-east-1 and ap-south-1.

pkg/cloudprovider/aws/pricing.go
tzneal: It's at startup and once every 12 hours so the logs shouldn't be excessive.  If there's some other feature that doesn't work in non-connected Karpenter, maybe add an environment variable then?

Makefile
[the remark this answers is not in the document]
bwagner5: What if you just added verify as a target dependency after release-gen? 

```
release: release-gen verify
...
```

If anything has changed, then it would fail, but I think that's okay, the generated code just needs to be committed and then rerun. Probably best to have that looked at someone separately anyways to ensure nothing looks wrong.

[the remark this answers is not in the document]
bwagner5: https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html

PR #1978, opened 2022-06-22
test: fix deflake

bwagner5:
Fixes # 

**Description**

 - Fix deflake workflow, since GHA env wasn't parsing the for loop `{1..10}` syntax. So the deflake tests were only being executed twice rather than 10 times (it worked fine on a MBP) 
 - Added `workflow_dispatch` so that we can manually trigger if we want to investigate a test

* tested on my fork within GHA

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/.github/workflows/deflake.yml
+++ b/.github/workflows/deflake.yml
@@ -2,6 +2,7 @@ name: Deflake
 on:
   schedule:
     - cron: '0 12 * * *'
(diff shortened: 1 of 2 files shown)


spring1843 (approved):
Nice

PR #1980, opened 2022-06-23
fix: prefer to satisfy self affinity from an existing inflight node

### LINKED ISSUE #1975: Pod Affinity impacts Karpenter tracking in-flight nodes
## Version

**Karpenter**: v0.12.0 (actively working on upgrading from 0.10)

**Kubernetes**: v1.19.15

## Expected Behavior
As of Karpenter 0.11, pods are no longer pre-bound to nodes. However, per https://github.com/aws/karpenter/pull/1727 , Karpenter is supposed to track newly launched nodes and not launch new nodes for unbound pods.

## Actual Behavior
I have a business reason to require certain pods to be in the same availability zone. When I include PodAffinity in my spec, Karpenter launches nodes, then continues to launch nodes while the first ones become ready. Eventually the first ones become ready, the pods schedule, and Karpenter deletes the extraneous nodes.

Here is an anonymized version of my podSpec. I've picked resource and instance class so that each pod gets its own node, I don't think they're important otherwise.
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: "inflate-deployment"
  namespace: "inflate-ns"
spec:
  replicas: 2
  selector:
    matchLabels:
      app: "TESTING-KARPENTER"
  template:
    metadata:
      labels:
        app: "TESTING-KARPENTER"
      name: "inflate"
      namespace: "inflate-ns"
    spec:
      containers:
        - name: inflate
          image: public.ecr.aws/eks-distro/kubernetes/pause:3.2
          resources:
            requests:
              cpu: 800m
              memory: 8000M
      nodeSelector:
        node.kubernetes.io/instance-type: "r5a.large"
      affinity:
        podAffinity:
            - topologyKey: "topology.kubernetes.io/zone"
              labelSelector:
                matchLabels:
                  app: "TESTING-KARPENTER"
```

If I remove the podAffinity block, everything behaves as I expect. Or, it behaves in a superficially similar way to 0.10 - Karpenter launches exactly the right number of nodes. With the podAffinity block, Karpenter launches nodes, then continues to launch more nodes until the pods schedule. I am fairly certain that the node has the right label `topology.kubernetes.io/zone: us-east-1d` before it becomes Ready in the cluster.

As a workaround, I've tried increasing the global setting `batchMaxDuration` . The longer the interval, the less time for Karpenter to launch extra nodes, but also the more time it takes to launch the first node and get the pod scheduled.

tzneal:
Fixes #1975

**Description**

For satisfying self affinity, prefer the domain of an inflight node.

Unit testing & deployed to EKS with a workload that exhibited the bug.

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/provisioning/scheduling/suite_test.go
+++ b/pkg/controllers/provisioning/scheduling/suite_test.go
@@ -3850,6 +3850,35 @@ var _ = Describe("No Pre-Binding", func() {
 		Expect(env.Client.List(ctx, &nodeList)).To(Succeed())
 		Expect(nodeList.Items).To(HaveLen(1))
 	})
+	It("should respect self pod affinity without pod binding (zone)", func() {
+		// Issue #1975
+		affLabels := map[string]string{"security": "s2"}
+
+		pods := MakePods(2, test.PodOptions{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: affLabels,
+			},
+			PodRequirements: []v1.PodAffinityTerm{{
+				LabelSelector: &metav1.LabelSelector{
+					MatchLabels: affLabels,
+				},
+				TopologyKey: v1.LabelTopologyZone,
+			}},
+		})
+		ExpectApplied(ctx, env.Client, provisioner)
+		ExpectProvisionedNoBinding(ctx, env.Client, controller, pods[0])
+		var nodeList v1.NodeList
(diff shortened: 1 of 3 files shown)


bwagner5 (approved):
nice!

PR #1990, opened 2022-06-23
feat: move to lease only elections

bwagner5:
Fixes # 

**Description**

 - Switch to a `lease` only lock rather than a `ConfigMapLease` lock. 
   - controller-runtime switched from a configmap lease pre v0.7.0 and starting w/ v0.7.0 went to a `ConfigMapLease` lock that does both in order to support a migration to a `Lease` only lock. Karpenter adopted v0.7.0 starting with v0.1.0 https://github.com/aws/karpenter/commit/2e3a63063e0faf6f37ef23dd3423b78bdec9aaf0 so we should be safe to switch to Lease only locks. This will remove one of the configmaps that Karpenter creates and gives permissions to.  

* Installed in an EKS cluster and deleted all leases and configmaps (related to election) and saw that the Karpenter pod created the lease resources. 

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/charts/karpenter/templates/role.yaml
+++ b/charts/karpenter/templates/role.yaml
@@ -27,7 +27,6 @@ rules:
     verbs: ["update", "patch", "delete"]
     resourceNames:
       - karpenter-global-settings
-      - karpenter-leader-election
       - config-logging
   - apiGroups: ["coordination.k8s.io"]
     resources: ["leases"]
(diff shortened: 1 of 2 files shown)


njtran (commented):
LGTM

charts/karpenter/templates/role.yaml
```diff
@@ -40,7 +39,7 @@ rules:
       - "webhook.webhookcertificates.00-of-01"
   # Cannot specify resourceNames on create
   # https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources
-  - apiGroups: ["coordinations.k8s.io"]
+  - apiGroups: ["coordination.k8s.io"]
```
ellistarn, line 42: hilarious

ellistarn (approved):

PR #1993, opened 2022-06-23
fix: hydrate lt cache on election win and invalidate LTs in the cache that do not exist in the CP

bwagner5:
Fixes # 

**Description**
 - This change requires that Karpenter has the leader lease before hydrating the LT cache on startup. This is necessary since the LT provider in the AWS CP manages the deletion lifecycle of LTs. So if Karpenter is deployed in an HA setup, a Karpenter pod that is not the leader could hydrate the cache with existing LTs created by Karpenter and then they would be deleted once the TTL expires (currently 1 min). The other Karpenter pods would have the LT still in the cache and fail when trying to launch an instance.  
 - In the case where the cache would get out-of-sync due to an operator manually deleting an LT, this change will invalidate the LTs in the cache that have caused errors which will result in a retry and a new LT being generated. 

* Tested manually on an EKS cluster.
  * Deployed Karpenter with some existing LTs. Noted that hydration of the LT cache occurred after the lease had been acquired. 
  * Manually deleted the LTs in the AWS console and tried an inflate scale-up
  * Karpenter invalidated the LTs causing the errors in the LT cache and retried the provisioning.   

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -94,24 +94,6 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, provider *v1alpha1.AWS
 	if provider.LaunchTemplateName != nil {
 		return map[string][]cloudprovider.InstanceType{ptr.StringValue(provider.LaunchTemplateName): nodeRequest.InstanceTypeOptions}, nil
 	}
-	// If launch template is not specified then resolve the correct AMI family launch templates
-	resolvedLaunchTemplates, err := p.getResolvedLTs(ctx, provider, nodeRequest, additionalLabels)
-	if err != nil {
-		return nil, err
-	}
-	launchTemplates := map[string][]cloudprovider.InstanceType{}
-	for _, resolvedLaunchTemplate := range resolvedLaunchTemplates {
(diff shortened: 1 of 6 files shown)


pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -217,6 +225,16 @@ func (p *LaunchTemplateProvider) volumeSize(quantity *resource.Quantity) *int64
 	return aws.Int64(quantity.ScaledValue(resource.Giga))
 }
 
+// Invalidate deletes a launch template from cache if it exists
+func (p *LaunchTemplateProvider) Invalidate(ctx context.Context, ltName string) {
+	p.Lock()
+	defer p.Unlock()
+	defer p.cache.OnEvicted(p.onCacheEvicted)
```
tzneal, line 232: Ugh, does Delete() call the OnEvicted handler?

tzneal (approved):

pkg/cloudprovider/aws/launchtemplate.go
bwagner5: yeah :(

PR #2008, opened 2022-06-28
fix: restrict iam:PassRole to just the node role we create

tzneal:
Fixes # 

**Description**

restrict iam:PassRole to just the node role we create

* Tested in a new cluster using this cloudformation template and was unable to launch instances using an instance profile that was not the default.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/website/content/en/preview/getting-started/getting-started-with-eksctl/cloudformation.yaml
+++ b/website/content/en/preview/getting-started/getting-started-with-eksctl/cloudformation.yaml
@@ -62,4 +62,4 @@ Resources:
           - Effect: Allow
             Action:
               - iam:PassRole
-            Resource: !Sub "arn:aws:iam::${AWS::AccountId}:role/KarpenterNodeRole-${ClusterName}"
+            Resource: !Sub "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/KarpenterNodeRole-${ClusterName}"
--- a/website/content/en/preview/getting-started/getting-started-with-eksctl/cloudformation.yaml
+++ b/website/content/en/preview/getting-started/getting-started-with-eksctl/cloudformation.yaml
@@ -46,7 +46,6 @@ Resources:
               - ec2:CreateFleet
               - ec2:RunInstances
               - ec2:CreateTags
-              - iam:PassRole
               - ec2:TerminateInstances
               - ec2:DeleteLaunchTemplate
               # Read Operations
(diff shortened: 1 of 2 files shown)


website/content/en/preview/getting-started/getting-started-with-eksctl/cloudformation.yaml
```diff
@@ -60,3 +59,7 @@ Resources:
               - ec2:DescribeSpotPriceHistory
               - ssm:GetParameter
               - pricing:GetProducts
+          - Effect: Allow
+            Action:
+              - iam:PassRole
+            Resource: !Sub "arn:aws:iam::${AWS::AccountId}:role/KarpenterNodeRole-${ClusterName}"
```
njtran, line 65: Is it possible to !Ref the role and get the ARN of the output of it?

tzneal: This doesn't work:

```
          - Effect: Allow
            Action:
              - iam:PassRole
            Resource: !Ref "KarpenterNodeRole-${ClusterName}"
```

njtran: Ah just trying it myself, doesn't output as ARN. Looks like the join func works here though.
```suggestion
            Resource: !Join [ "/", [ !Sub "arn:aws:iam::${AWS::AccountId}:role", !Ref "KarpenterNodeRole" ] ]
```

bwagner5: make sure to use `${AWS::Partition}` as well:

`!Sub "${AWS::Partition}:iam::${AWS::AccountId}:role/KarpenterNodeRole-${ClusterName}"`

tzneal: If you change the name, you'll have to change it multiple times anyway, we reference that name with the `eksctl create iamidentitymapping`. I think the `!Join` is a lot less readable.

njtran: I'm only talking about the CFN resources, rather than the whole setup process. If I need to change the name of a resource, then I only need to do it where the name is defined with the join. I agree it's less readable, but it will reduce user errors if they need to configure this, especially since this might be hard to debug.

bwagner5: I'd agree that the readability trade-off isn't really worth it here. This is just an example infra setup, so it should bias towards readability rather than extensibility.

bwagner5 (approved):
lgtm

PR #2013, opened 2022-06-28
test: add a test for inflight nodes combined with providerRef

tzneal:
Fixes # 

**Description**

 add a test for inflight nodes combined with providerRef

* unit testing

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/provisioning/scheduling/suite_test.go
+++ b/pkg/controllers/provisioning/scheduling/suite_test.go
@@ -3803,6 +3803,28 @@ var _ = Describe("In-Flight Nodes", func() {
 		})
 		Expect(nodesWithCPUFree).To(BeNumerically("<=", 1))
 	})
+	It("should not launch a second node if there is an in-flight node that can support the pod (#2011)", func() {
+		opts := test.PodOptions{ResourceRequirements: v1.ResourceRequirements{
+			Limits: map[v1.ResourceName]resource.Quantity{
+				v1.ResourceCPU: resource.MustParse("10m"),
+			},
+		}}
+
+		// there was a bug in cluster state where we failed to identify the instance type resources when using a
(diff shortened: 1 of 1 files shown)


suket22 (dismissed):

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3803,6 +3803,28 @@ var _ = Describe("In-Flight Nodes", func() {
...
+		provisioner.Spec.ProviderRef = &v1alpha5.ProviderRef{}
+
+		ExpectApplied(ctx, env.Client, provisioner)
+		initialPod := ExpectProvisioned(ctx, env.Client, controller, test.UnschedulablePod(opts))
+		node1 := ExpectScheduled(ctx, env.Client, initialPod[0])
+		ExpectReconcileSucceeded(ctx, nodeStateController, client.ObjectKeyFromObject(node1))
+
+		secondPod := ExpectProvisioned(ctx, env.Client, controller, test.UnschedulablePod(opts))
+		node2 := ExpectScheduled(ctx, env.Client, secondPod[0])
+		Expect(node1.Name).To(Equal(node2.Name))
```
suket22, line 3826: For sanity, can you check if this test fails on v0.13.0?

As I understand it, a single pod kept causing multiple nodes to be launched so we shouldn't need the second pod?
I wonder if there's a way to just force the reconcile loop again in our test and make sure that 
```
node2 := ExpectScheduled(ctx, env.Client, secondPod[0])
Expect(node1.Name).To(Equal(node2.Name))
```
succeeds.

tzneal: I had verified that the test failed without the fix when I wrote it, the same mechanism caused the failure since the two pods were identical.  I reworked it to be a bit closer in spirit to the actual bug though.

suket22 (approved):

PR #2014, opened 2022-06-28
feat: Adding support for custom AMIs

### LINKED ISSUE #1008: An ability to specify AMI in Provisioner spec

It would be convenient if we were able to specify custom `AMI` in the `Provisioner` spec instead of creating a lot of custom launch templates just to change the AMI in them.

Currently I should create a lot of launch templates and change only `AMI` in them.
It is annoying because `Karpenter` already creates launch templates using taints/labels from the `Provisioner` spec.
Why can't we add `AMI` as well?

Thanks!

### Community Note

suket22:
Adding support for custom AMIs

Fixes #1008

**Description**
This adds initial support for custom AMIs. 
* This works in conjunction with `AMIFamily` and the bootstrapping logic of that family will apply to these worker nodes. 
* TODO - Introduce a `Custom` AMIFamily where users can provide the entire content of the UserData instead (similar to what you'd do with launch templates today). 

Behavior - 
* If you specify an `awsnodetemplate` with no AMIs or

* Unit tests
* I ran a `make apply` and created a node template that had two custom AMIs - one for ARM and one for x86 and made sure pods requesting both could be satisfied. 

**Docs will land in a PR at the end, once everything is complete (including custom AMI families)**
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
feat: Adding support for custom AMIs
```


pkg/cloudprovider/aws/amifamily/ami.go
```diff
@@ -22,18 +22,55 @@ import (
...
+			if err != nil {
+				return nil, err
+			}
+			amiIDs[amiID] = append(amiIDs[amiID], instanceType)
+		}
+	}
+	return amiIDs, nil
+}
+
+func (p *AMIProvider) GetDefaultAMIFromSSM(ctx context.Context, instanceType cloudprovider.InstanceType, ssmQuery string) (string, error) {
```
bwagner5, line 73: Does this need to be exported?

pkg/cloudprovider/aws/amifamily/ami.go
```diff
@@ -46,3 +83,35 @@ func (p *AMIProvider) Get(ctx context.Context, instanceType cloudprovider.Instan
 	logging.FromContext(ctx).Debugf("Discovered %s for query %q", ami, ssmQuery)
 	return ami, nil
 }
+
+func (p *AMIProvider) GetAMIRequirements(ctx context.Context, providerRef *v1alpha5.ProviderRef) (map[string]scheduling.Requirements, error) {
```
bwagner5, line 87: I don't think this needs to be exported either?

pkg/cloudprovider/aws/amifamily/ami.go
```diff
@@ -46,3 +83,35 @@ func (p *AMIProvider) Get(ctx context.Context, instanceType cloudprovider.Instan
...
+	for amiID, requirements := range amiRequirements {
+		if err := instanceType.Requirements().Compatible(requirements); err == nil {
+			return amiID
+		}
+	}
+	return ""
+}
+
+// Keys returns a slice of all the keys in a map
+func Keys(m map[string]scheduling.Requirements) []string {
```
bwagner5, line 111: same here on not exporting?

pkg/cloudprovider/aws/amifamily/ami.go
```diff
@@ -46,3 +83,35 @@ func (p *AMIProvider) Get(ctx context.Context, instanceType cloudprovider.Instan
 	logging.FromContext(ctx).Debugf("Discovered %s for query %q", ami, ssmQuery)
 	return ami, nil
 }
+
+func (p *AMIProvider) GetAMIRequirements(ctx context.Context, providerRef *v1alpha5.ProviderRef) (map[string]scheduling.Requirements, error) {
+	amiRequirements := make(map[string]scheduling.Requirements)
```
bwagner5, line 88:
```suggestion
	amiRequirements := map[string]scheduling.Requirements{}
```

charts/karpenter/crds/karpenter.k8s.aws_awsnodetemplates.yaml
```diff
@@ -42,6 +42,22 @@ spec:
               amiFamily:
                 description: AMIFamily is the AMI family that instances use.
                 type: string
+              amis:
+                description: AMIs is a list of custom AMIs that Karpenter will choose
+                  from.
+                items:
+                  properties:
+                    id:
```
bwagner5, line 50: Is there going to be another field for an AMI Alias ?

pkg/cloudprovider/aws/amifamily/ami.go
```diff
@@ -46,3 +83,35 @@ func (p *AMIProvider) Get(ctx context.Context, instanceType cloudprovider.Instan
...
+		for _, ami := range ant.Spec.AMIs {
+			amiRequirements[ami.ID] = scheduling.NewAMIRequirements(ami)
+		}
+	}
+	return amiRequirements, nil
+}
+
+func getAMIOverride(instanceType cloudprovider.InstanceType, amiRequirements map[string]scheduling.Requirements) string {
+	for amiID, requirements := range amiRequirements {
+		if err := instanceType.Requirements().Compatible(requirements); err == nil {
```
bwagner5, line 103: neat!

pkg/scheduling/requirements.go
```diff
@@ -97,6 +98,14 @@ func NewPodRequirements(pod *v1.Pod) Requirements {
 	return NewNodeSelectorRequirements(requirements...)
 }
 
+func NewAMIRequirements(ami v1alpha1.AMI) Requirements {
```
bwagner5, line 101: We don't want AWS specific things leaking into the core scheduling logic. Seems like this is just a small helper, can you move this to the AWS cloudprovider?

pkg/cloudprovider/aws/amifamily/resolver.go
```diff
@@ -96,13 +97,9 @@ func (r Resolver) Resolve(ctx context.Context, provider *v1alpha1.AWS, nodeReque
...
 	}
 	amiFamily := GetAMIFamily(provider.AMIFamily, options)
-	amiIDs := map[string][]cloudprovider.InstanceType{}
-	for _, instanceType := range nodeRequest.InstanceTypeOptions {
-		amiID, err := r.amiProvider.Get(ctx, instanceType, amiFamily.SSMAlias(options.KubernetesVersion, instanceType))
-		if err != nil {
-			return nil, err
-		}
-		amiIDs[amiID] = append(amiIDs[amiID], instanceType)
+	amiIDs, err := r.amiProvider.Get(ctx, provider, nodeRequest, options, amiFamily)
```
bwagner5, line 100: I think it's going to be required to do an inspection of the AMI to see what the block device mappings are and potentially other metadata that we'll need for custom AMIs. It's probably better to do that anyways for the default AMIs too. The default AMIs could potentially be updated to use different block device mappings that we should really discover.

bwagner5 (commented):
nice, looking good so far!

pkg/cloudprovider/aws/amifamily/ami.go
tzneal: This is just the `lo.Keys` function I think.

charts/karpenter/crds/karpenter.k8s.aws_awsnodetemplates.yaml
```diff
@@ -42,6 +42,22 @@ spec:
               amiFamily:
                 description: AMIFamily is the AMI family that instances use.
                 type: string
+              amis:
+                description: AMIs is a list of custom AMIs that Karpenter will choose
```
tzneal, line 46: Maybe a note about how we pick one at random if multiple AMIs match?

pkg/cloudprovider/aws/amifamily/ami.go
suket22: The Requirements abstraction is really neat!

pkg/cloudprovider/aws/amifamily/resolver.go
suket22: > The default AMIs could potentially be updated to use different block device mappings that we should really discover.


I know AMIs are versioned so backwards compatibility is less of a concern, but even so I don't expect something like the default device name / other metadata to change for a given AMI family. Can break everyone's existing launch templates pretty easily :D

> I think it's going to be required to do an inspection of the AMI to see what the block device mappings are and potentially other metadata that we'll need for custom AMIs

The thinking I had on this was that when you specify the AMIFamily to be `AL2` or `Bottlerocket` then we'll just assume the same settings as that AMIFamily. If you want to override other metadata and things, then you have to specify `AMIFamily` as `Custom` and then we'll start inspecting the AMI and get the `RootDeviceName` from there. It's a fair amount of changes to support that AMIFamily (new IAM permissions for ec2:DescribeImages too) which is why I thought it deserved it's own PR. 

Thoughts?

bwagner5: 👍 sgtm on an additional PR if it becomes necessary. The block device mappings I was thinking of that could change is Ubuntu. Since we pick up latest, it might break unintentionally. But I'm not overly concerned.

bwagner5:
Should be able to rebase to fix the failing test. Todd fixed in https://github.com/aws/karpenter/pull/2020

pkg/apis/awsnodetemplate/v1alpha1/awsnodetemplate.go
```diff
@@ -29,6 +31,17 @@ type AWSNodeTemplateSpec struct {
...
+	// one will be chosen at random.
+	// +optional
+	AMIs []AMI `json:"amis,omitempty"`
+}
+
+type AMI struct {
+	// Id is an identifier for an AMI.
+	ID string `json:"id,omitempty"`
+	// Properties defines all the properties of an AMI like architecture, GPU compatibility and so on.
+	Properties map[string]string `json:"properties,omitempty"`
```
ellistarn, line 44: See design guidelines. Reuse k8s concepts where applicable.
```suggestion
	Selector map[string]string `json:"properties,omitempty"`
```

ellistarn (commented):
Can you add an integration test?

pkg/apis/awsnodetemplate/v1alpha1/awsnodetemplate.go
suket22: I thought about this and the word `Selector` didn't make sense to me. `nodeSelector` makes sense in a podSpec -> you're selecting a set of nodes that the pod can be assigned to. 

This is not helping you select an AMI since the AMI is already defined by this struct. You're only defining the requirements of an AMI to be used / or it's properties.

ellistarn: Selector (LabelSelector, NodeSelector, etc) is a broadly used concept in the k8s that is identical to the semantic you're defining here. I could see an argument for `AMISelector`, similar to `SubnetSelector`, but given that this is already a property of AMI, to avoid stuttering you'd need `ami.selector`

suket22: I still don't agree. `SubnetSelector` helps you choose / select a subnet based on something. This isn't helping you select an AMI, but is rather just defining what the AMI represents. Same for label selector / node selector. 

I don't think it's the same semantic.

ellistarn: You're right, it's the inverse semantic. The pattern is typically `Selector` -> `Labels`. The thing selecting (e.g. the pod) uses a selector, where the thing targeted has labels. Perhaps we can simply call this `labels`?

bwagner5: As is, I would agree w/ @suket22 that this is not the same concept and would be confusing to conflate the two. It could be modified to be the same and might be useful at the expense of describing the AMIs and discovering the attributes on them.

ellistarn: Not sure I'm tracking @bwagner5. Perhaps we can discuss offline?

suket22: Adding amiSelector in a subsequent commit.

test/suites/launchTemplates/suite_test.go
```diff
@@ -0,0 +1,93 @@
+package launchtemplates
```
ellistarn, line 1: Thoughts on keeping this in the integration suite? Right now suites are parallel, but pay a large startup cost. Unless this relies on addtl dependencies (CSI, etc), maybe best to keep them together.

pkg/cloudprovider/aws/amifamily/ami.go
```diff
@@ -19,30 +19,178 @@ import (
...
+func (p *AMIProvider) Get(ctx context.Context, provider *awsv1alpha1.AWS, nodeRequest *cloudprovider.NodeRequest, options *Options, amiFamily AMIFamily) (map[string][]cloudprovider.InstanceType, error) {
+	amiIDs := map[string][]cloudprovider.InstanceType{}
+	amiRequirements, err := p.getAMIRequirements(ctx, nodeRequest.Template.ProviderRef)
+	if err != nil {
+		return nil, err
+	}
+	var amiID string
+	if len(amiRequirements) > 0 {
+		for _, instanceType := range nodeRequest.InstanceTypeOptions {
+			amiID = getCompatibleAMI(instanceType, amiRequirements)
```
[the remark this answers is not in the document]
suket22: Yup makes sense. Dunno why I've done this roundabout way of an empty string + extra if.

test/suites/launchTemplates/suite_test.go
suket22: I can go either way. I expect this entire suite of tests to call EC2 + SSM to verify that the actual configuration on the EC2 instances matches our requirements in the provisioner, but shouldn't be other dependencies.

ellistarn: Cool. If no other deps, let's keep it in the integ suite.

pkg/cloudprovider/aws/amifamily/ami.go
```diff
@@ -19,30 +19,170 @@ import (
...
-func (p *AMIProvider) Get(ctx context.Context, instanceType cloudprovider.InstanceType, ssmQuery string) (string, error) {
-	if id, ok := p.cache.Get(ssmQuery); ok {
+// If AMI overrides are specified in the AWSNodeTemplate, then only those AMIs will be chosen.
+func (p *AMIProvider) Get(ctx context.Context, provider *awsv1alpha1.AWS, nodeRequest *cloudprovider.NodeRequest, options *Options, amiFamily AMIFamily) (map[string][]cloudprovider.InstanceType, error) {
+	amiIDs := map[string][]cloudprovider.InstanceType{}
+	amiRequirements, err := p.getAMIRequirements(ctx, nodeRequest.Template.ProviderRef)
+	if err != nil {
+		return nil, err
+	}
+	var amiID string
```
ellistarn, line 60: This stray variable is a bit odd to me, since you shadow it on line 63, but don't shadow it on line 64. I'm surprised the linter isn't complaining, since we only allow shadowing ctx/err.

pkg/cloudprovider/aws/amifamily/ami.go
```diff
@@ -19,30 +19,170 @@ import (
...
+	// Always add the architecture of an image as a requirement, irrespective of what's specified in EC2 tags.
+	architecture := *ec2Image.Architecture
+	if value, ok := awsv1alpha1.AWSToKubeArchitectures[architecture]; ok {
+		architecture = value
+	}
+	requirements.Add(scheduling.Requirements{v1.LabelArchStable: sets.NewSet(architecture)})
+	return requirements
+}
+
+func amiIDs(amis []*ec2.Image) []string {
```
ellistarn, line 182: optional: depending on your proclivity for functional programming, and could be done inline above.

`lo.Map(amis, func(ami *ec2.Image) string { return ami.ImageId })`

pkg/cloudprovider/aws/amifamily/resolver.go
```diff
@@ -96,13 +100,9 @@ func (r Resolver) Resolve(ctx context.Context, provider *v1alpha1.AWS, nodeReque
...
```
ellistarn, line 103: optional: you could inline `amiFamily` abovem without the performance issue that existed prior, now that you're no longer calling it in a loop.

pkg/cloudprovider/aws/fake/ec2api.go
```diff
@@ -187,6 +191,25 @@ func (e *EC2API) DescribeInstancesWithContext(_ context.Context, input *ec2.Desc
...
+		return nil, e.NextError.Get()
+	}
+	e.CalledWithDescribeImagesInput.Add(input)
+	if !e.DescribeImagesOutput.IsNil() {
+		return e.DescribeImagesOutput.Clone(), nil
+	}
+	return &ec2.DescribeImagesOutput{
+		Images: []*ec2.Image{
+			{
+				ImageId:      aws.String(randomdata.SillyName()),
```
ellistarn, line 206: Check out the test.RandomName() helper I just wrote.

pkg/cloudprovider/aws/suite_test.go
```diff
@@ -1177,6 +1181,175 @@ var _ = Describe("Allocation", func() {
...
+					nodeTemplate := test.AWSNodeTemplate(test.AWSNodeTemplateOptions{
+						UserData:    nil,
+						AMISelector: map[string]string{"karpenter.sh/discovery": "my-cluster"},
+						AWS:         *provider,
+						ObjectMeta:  metav1.ObjectMeta{Name: providerRefName}})
+					fakeEC2API.DescribeImagesOutput.Set(&ec2.DescribeImagesOutput{Images: []*ec2.Image{
+						{ImageId: aws.String("ami-123"), Architecture: aws.String("x86_64")},
+					}})
+					ExpectApplied(ctx, env.Client, nodeTemplate)
+					controller := provisioning.NewController(injection.WithOptions(ctx, opts), cfg, env.Client, clientSet.CoreV1(), recorder, cloudProvider, cluster)
```
ellistarn, line 1201: Is it necessary to make a new controller here?

test/suites/launchTemplates/suite_test.go
```diff
@@ -0,0 +1,93 @@
...
+	RegisterFailHandler(Fail)
+	BeforeSuite(func() {
+		var err error
+		env, err = environment.NewEnvironment(t)
+		Expect(err).ToNot(HaveOccurred())
+	})
+	RunSpecs(t, "LaunchTemplates")
+}
+
+var _ = BeforeEach(func() {
```
ellistarn, line 35:
```suggestion
var _ = BeforeEach(env.BeforeEach)
```

test/suites/launchTemplates/suite_test.go
```diff
@@ -0,0 +1,93 @@
...
+	env.BeforeEach()
+})
+var _ = AfterEach(func() {
```
ellistarn, line 38:
```suggestion
var _ = AfterEach(env.AfterEach)
```

test/suites/launchTemplates/suite_test.go
```diff
@@ -0,0 +1,93 @@
...
+		env.ExpectCreatedNodeCount("==", 0)
+		env.ExpectCreated(pod, provider, provisioner)
+
+		env.EventuallyExpectHealthy(pod)
+		env.ExpectCreatedNodeCount("==", 1)
+
+		var node v1.Node
+		env.Client.Get(env.Context, types.NamespacedName{Name:  pod.Spec.NodeName}, &node)
+		instanceID := instanceID(node.Spec.ProviderID)
+		ec2api := ec2.New(&env.AWSSession)
```
ellistarn, line 69: Thoughts on baking this into the env so that it always has an EC2 client we can use?

`env.ec2api.DescribeInstances`

test/suites/launchTemplates/suite_test.go
```diff
@@ -0,0 +1,93 @@
...
+	})
+})
+
+func instanceID(providerID string) string {
+	providerIDSplit := strings.Split(providerID, "/")
+	return providerIDSplit[len(providerIDSplit)-1]
+}
+
+func customAMI() string {
+	ssmApi := ssm.New(&env.AWSSession)
```
ellistarn, line 86: same re: attaching to env

ellistarn (dismissed):
Couple of optional comments. Nice work!

pkg/cloudprovider/aws/amifamily/ami.go
suket22: Nice!

pkg/cloudprovider/aws/amifamily/resolver.go
suket22: amiFamily gets used a few lines below too so can't inline it here I think

pkg/cloudprovider/aws/suite_test.go
suket22: I don't think so. I think I might've copied from some other test and didn't realize this was in. Taking it out, nice catch!

test/suites/launchTemplates/suite_test.go
suket22: Sounds good!

test/suites/launchTemplates/suite_test.go
suket22: Didn't spend too long to debug, but this was making our tests fail so left it out.

test/suites/launchTemplates/suite_test.go
suket22: Done!

charts/karpenter/crds/karpenter.k8s.aws_awsnodetemplates.yaml
```diff
@@ -42,6 +42,11 @@ spec:
               amiFamily:
                 description: AMIFamily is the AMI family that instances use.
                 type: string
+              amiSelector:
+                additionalProperties:
```
bwagner5, line 46: should this still have `additionalProperties`?

bwagner5 (commented):
nice!! Just two small comments and then I think it's ready.

charts/karpenter/crds/karpenter.k8s.aws_awsnodetemplates.yaml
suket22: Ooh, this is auto-generated, I wonder what happened here. Lemme fix.

suket22: I deleted and recreated the CRD and I think this is just part of the default CRD spec. You'll see it exists for subnetSelector and securityGroupSelector too

suket22:
> Can you also add the upgrade guide instructions for the new permission and a kubectl replace the AWSNodeTemplateCRD?

I plan on adding all the documentation once we've merged in support for the `Custom` AMIFamily too. I'll make updates to 
* Upgrade Guide
* Documentation on how this feature will work (defining requirements via EC2 Tags), how UserData will be interpreted, how EBS volume sizing will be determined etc.
* Examples to the /examples directory.

charts/karpenter/crds/karpenter.k8s.aws_awsnodetemplates.yaml
bwagner5: oh weird! ok 👍

bwagner5 (approved):
lgtm! nice work!

ellistarn (commented):
Awesome work!

PR #2067, opened 2022-07-07
fix: update initial pricing async (#2063)

tzneal:
Fixes # 

**Description**
    Updating synchronously with startup avoids potentially launching
    any initial nodes with the old pricing information.  I think this
    is causing more problems than it was solving though and this change
    moves the initial update to be async so it doesn't block startup.

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


bwagner5 (approved):
lgtm

PR #2086, opened 2022-07-08
test: fix batcher test flake

tzneal:
**Description**

TriggerAndWait() is used in unit tests to immediately trigger the batcher
and then wait on the condition variable.  It locks the mutex, and then
waits which unlocks it.  By locking on the mutex before broadcasting,
we ensure that we don't broadcast until the unit test is already
waiting on the condition variable.

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


bwagner5 (approved):
lgtm

PR #2105, opened 2022-07-11
feat: make emptiness controller aware of scheduling

tzneal:
**Description**

    This eliminates some problems seen with short emptiness TTL
    where we can delete the node too early (e.g. PVC binding) and
    the 'boucing' of the emptiness annotation for newly launched nodes.

* Unit tests & `make e2etests`

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/node/emptiness.go
+++ b/pkg/controllers/node/emptiness.go
@@ -26,6 +26,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
+	"github.com/aws/karpenter/pkg/controllers/state"
 	"github.com/aws/karpenter/pkg/utils/functional"
 	"github.com/aws/karpenter/pkg/utils/injectabletime"
 	"github.com/aws/karpenter/pkg/utils/pod"
@@ -34,6 +35,7 @@ import (
 // Emptiness is a subreconciler that deletes nodes that are empty after a ttl
 type Emptiness struct {
 	kubeClient client.Client
+	cluster    *state.Cluster
 }
 
 // Reconcile reconciles the node
@@ -52,6 +54,11 @@ func (r *Emptiness) Reconcile(ctx context.Context, provisioner *v1alpha5.Provisi
 	empty, err := r.isEmpty(ctx, n)
 	if err != nil {
 		return reconcile.Result{}, err
+	}
+
+	// node is empty, but it is in-use per the last scheduling round so we don't consider it empty
+	if r.cluster.IsNodeNominated(n.Name) {
+		return reconcile.Result{}, nil
 	}
 
 	emptinessTimestamp, hasEmptinessTimestamp := n.Annotations[v1alpha5.EmptinessTimestampAnnotationKey]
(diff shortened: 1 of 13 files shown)


tzneal:
This simplifies consolidation a good amount while also eliminating the bouncing emptiness TTL that we currently see when nodes launch.

pkg/controllers/node/emptiness.go
```diff
@@ -54,6 +56,11 @@ func (r *Emptiness) Reconcile(ctx context.Context, provisioner *v1alpha5.Provisi
 		return reconcile.Result{}, err
 	}
 
+	// node is empty, but it is in-use per the last scheduling round so we don't consider it empty
+	if r.cluster.IsNodeNominated(n.Name) {
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
ellistarn, line 61: Do we need to requeue? The reconcile loop should be re-triggered by the node's status updates or any related pods on this node.

pkg/controllers/provisioning/scheduling/queue.go
```diff
@@ -83,6 +83,19 @@ func byCPUAndMemoryDescending(pods []*v1.Pod) func(i int, j int) bool {
 		} else if memCmp > 0 {
 			return true
 		}
-		return false
+		// If all else is equal, give a consistent ordering. This reduces the number of NominatePod events as we
+		// de-duplicate those based on identical content.
+		lhsPod := pods[i]
```
ellistarn, line 88: consider setting this at the top of the function, since these helper vars could be used by lines 69/70

pkg/controllers/state/cluster.go
```diff
@@ -55,6 +59,7 @@ func NewCluster(client client.Client, cp cloudprovider.CloudProvider) *Cluster {
 	c := &Cluster{
 		kubeClient:    client,
 		cloudProvider: cp,
+		nodesInUse:    cache.New(10*time.Second, 10*time.Second),
```
ellistarn, line 62: Should we be sharing this 10 second constant with the emptiness controller?

ellistarn: Thoughts on tracking this per pod, instead of via a cache? We could maintain `NominatedNodes map[pod]*Node` and clean this up as we learn about the pods real binding.

pkg/controllers/node/emptiness.go
tzneal: It will eventually retrigger, so it's probably fine to not re-queue.

pkg/controllers/state/cluster.go
tzneal: Removed requeue from emptiness controller.  This is a more inexpensive way to track if we expect a node to be used in the near future and conveniently avoids adding more work that would be performed while holding the cluster state lock.

pkg/controllers/state/cluster.go
```diff
@@ -53,10 +55,11 @@ type Cluster struct {
 
 func NewCluster(client client.Client, cp cloudprovider.CloudProvider) *Cluster {
 	c := &Cluster{
-		kubeClient:    client,
-		cloudProvider: cp,
-		nodes:         map[string]*Node{},
-		bindings:      map[types.NamespacedName]string{},
+		kubeClient:     client,
+		cloudProvider:  cp,
+		nominatedNodes: cache.New(10*time.Second, 10*time.Second),
```
ellistarn, line 60: Do you mind adding a comment about why 10 seconds here? IIUC, this is the amount of time we'd expect the kube scheduler to kick in after a node is ready.

ellistarn (approved):
Nice! One minor question

ellistarn (approved):

ellistarn:
Nice!!

PR #2112, opened 2022-07-12
chore: add notification when a release occurs

spring1843:
Fixes # 

Adds a notification message when releases occur. 

*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/hack/release_common.sh
+++ b/hack/release_common.sh
@@ -41,3 +41,12 @@ cosignImages() {
     COSIGN_EXPERIMENTAL=1 cosign sign ${COSIGN_FLAGS} ${CONTROLLER_DIGEST}
     COSIGN_EXPERIMENTAL=1 cosign sign ${COSIGN_FLAGS} ${WEBHOOK_DIGEST}
 }
+
+notifyRelease(){
+  RELEASE_TYPE=$1
+  RELEASE_IDENTIFIER=$2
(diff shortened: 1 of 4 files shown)


hack/release_common.sh
```diff
@@ -41,3 +41,13 @@ cosignImages() {
     COSIGN_EXPERIMENTAL=1 cosign sign ${COSIGN_FLAGS} ${CONTROLLER_DIGEST}
     COSIGN_EXPERIMENTAL=1 cosign sign ${COSIGN_FLAGS} ${WEBHOOK_DIGEST}
 }
+
+notifyRelease(){
+  RELEASE_TYPE=$1
```
ellistarn, line 46: Thoughts on not exposing the concept of RELEASE_TYPE and instead let it simply be the tag. Customers can delineate the release type by the name. This enables us to build new types of releases and just tag them as we see fit.

spring1843: I rather keep it, notifications can possibly used in many different ways, and it's good to make it easy for the client to not have to do extra work. For example a simple check for the release type can be done via text search making it easy for a basic client like a mail client to do different things depending on the type.

We should still be able to tag things the way we see fit. Also it's good to know what type of a tag was supposed to be released so that we can then compare it to what was actually release. This maybe helpful in troubleshooting and forensics.

Finally a blob of data is not as helpful, this makes the message more human friendly too, it's somewhat like the "kind" in Kubernetes, we are not just defining attributes but also specifying what this even is.

ellistarn (approved):

PR #2123, opened 2022-07-13
feat: cluster consolidation

### LINKED ISSUE #1091: Karpenter workload consolidation/defragmentation
As a cluster admin, I want Karpenter to consolidation the application workloads by moving pods to a fewer worker nodes and scale down the cluster so that I can improve the cluster resource utilization rate.

In an under-utilized cluster, application pods are spread across worker nodes with exceeding among of resources. This wasteful situation can be improved by carefully packing the pods to a smaller number of worker nodes with the right size. Current version of Karpenter does not support rearranging pods and continuously improve the  cluster utilization. The workload consolidation feature is the missing important component to complete the cluster scaling life cycle management loop. 

This workload consolidation feature is nontrivial because of the following coupling problems.
- **Pod Packing:**
    The pod packing problem determines which pods should be hosted together by the same worker node according to their taints and constraints. The goal is to produce a fewer well-balanced groups of pods that can be hosted by worker nodes with just the right size. 
- **Instance Type Selection:** 
    According to the pod packing solution, the instance type selection problem determines which combination of instance types should be used to host the pods after the rearrangement.

The above problems deeply couple together so that one solution affect the other. Together the problem is a variant of the bin packing problem which is NP-complete. A practical solution will implements a quick heuristic algorithm that utilizes the special structure of the problem for specific use cases and user preferences. Therefore, thorough discussions with the customer is important.

Currently Karpenter will scale down empty nodes automatically. However, it does not actively move pods around to create empty nodes.

**Additional context**
Currently the workload consolidation feature is in the design phase. We should gather inputs from the customers about their objectives and preferences.

**Attachments**

### Community Note

tzneal:
Fixes #1091

**Description**

Implements cluster consolidation via:
- removing nodes if their workloads can run on other nodes
- replacing nodes with cheaper instances

Unit testing, e2e tests and deployed to EKS.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/controllers/consolidation/pdblimits.go
```diff
@@ -0,0 +1,86 @@
...
+limitations under the License.
+*/
+
+package consolidation
+
+import (
+	"context"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/policy/v1beta1"
```
olemarkus, line 21: Use policy/v1? v1beta1 has been removed from k/k now.

olemarkus (changes requested):

pkg/controllers/consolidation/pdblimits.go
tzneal: Thanks, updated.

tzneal:
From @offzale
> I will leave it running and keep an eye on it to gather some feedback. What I have noticed so far an increment on cpu usage by +640%, in comparison to the resources it normally takes.

This should be resolved with the 010e04fb7826eae80b6c62984083e99cfe52ce30 snapshot, I updated the instructions on the issue.

pkg/controllers/consolidation/pdblimits.go
tzneal: I need to revert this change, we currently test against 1.19, 1.20, 1.21 and 1.22 and this breaks on older K8s.

website/content/en/preview/tasks/deprovisioning.md
```diff
@@ -86,9 +101,17 @@ Review what [disruptions are](https://kubernetes.io/docs/concepts/workloads/pods
...
 
 If you want to terminate a node with a `do-not-evict` pod, you can simply remove the annotation and the deprovisioning process will continue.
+
+### Scheduling Constraints (Consolidation Only)
+
+Consolidation will be unable to consolidate a node if, as a result of its scheduling simulation, it determines that the pods on a node cannot run on other nodes due to inter-pod affinity/anti-affinity, topology spread constraints, or some other scheduling restriction that couldn't be fulfilled.
+
+### Controllerless Pods (Consolidation Only)
+
+Consolidation will not attempt to consolidate a node that is running pods that are not owned by a controller (e.g. a `ReplicAset`).  In general we cannot assume that these pods would be recreated if they were evicted from the node that they are currently running on.
```
ellistarn, line 117:
```suggestion
Consolidation will not attempt to consolidate a node that is running pods that are not owned by a controller (e.g. a `ReplicaSet`).  In general we cannot assume that these pods would be recreated if they were evicted from the node that they are currently running on.
```

charts/karpenter/templates/clusterrole.yaml
```diff
@@ -22,12 +22,15 @@ rules:
   - apiGroups: ["storage.k8s.io"]
     resources: ["storageclasses", "csinodes"]
     verbs: ["get", "watch", "list"]
-  - apiGroups: ["apps"]
-    resources: ["daemonsets"]
+  - apiGroups: ["apps", ""]
+    resources: ["daemonsets","replicasets", "replicationcontrollers"]
```
ellistarn, line 26: Did we end up using these permissions?

charts/karpenter/values.yaml
```diff
@@ -92,7 +92,9 @@ controller:
   # -- SecurityContext for the controller container.
   securityContext: {}
   # -- Additional environment variables for the controller pod.
-  env: []
+  env: 
+  - name: ENABLE_PROFILING
```
ellistarn, line 96: Reminder to disable this by default.

pkg/apis/provisioning/v1alpha5/labels.go
```diff
@@ -30,10 +30,11 @@ var (
 	// Karpenter specific domains and labels
 	KarpenterLabelDomain = "karpenter.sh"
 
-	ProvisionerNameLabelKey         = Group + "/provisioner-name"
-	DoNotEvictPodAnnotationKey      = Group + "/do-not-evict"
-	EmptinessTimestampAnnotationKey = Group + "/emptiness-timestamp"
-	TerminationFinalizer            = Group + "/termination"
+	ProvisionerNameLabelKey           = Group + "/provisioner-name"
+	DoNotEvictPodAnnotationKey        = Group + "/do-not-evict"
+	DoNotConsolidateNodeAnnotationKey = KarpenterLabelDomain + "/do-not-consolidate"
```
ellistarn, line 35: I'm not sure if this is a good idea yet, but just a thought. If we make this a label, there are a couple of interactions here that are *very* interesting. 

A pod can request a node that doesn't consolidate by node selector. You can enable/disable consolidation at the provisioner level via standard requirements.

pkg/apis/provisioning/v1alpha5/provisioner.go
```diff
@@ -70,6 +70,14 @@ type ProvisionerSpec struct {
...
 	// Limits define a set of bounds for provisioning capacity.
 	Limits *Limits `json:"limits,omitempty"`
+	// Consolidation are the consolidation parameters
+	// +optional
+	Consolidation *Consolidation `json:"consolidation,omitempty"`
+}
+
+type Consolidation struct {
+	// Enabled enables consolidation if it has been set
+	Enabled *bool `json:"enabled,omitempty"`
```
ellistarn, line 80: Booleans in Kubernetes APIs are pretty rare. Things are typically modeled via ptr existence or enums, rather than a big panel of booleans. I'd like to explore configuration alternatives that are more familiar to existing applications in the k8s space.

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -62,6 +64,10 @@ func (s *ProvisionerSpec) validateTTLSecondsAfterEmpty() (errs *apis.FieldError)
 	if ptr.Int64Value(s.TTLSecondsAfterEmpty) < 0 {
 		return errs.Also(apis.ErrInvalidValue("cannot be negative", "ttlSecondsAfterEmpty"))
 	}
+	// TTLSecondsAfterEmpty and consolidation are mutually exclusive
+	if s.Consolidation != nil && aws.BoolValue(s.Consolidation.Enabled) && s.TTLSecondsAfterEmpty != nil {
+		return errs.Also(apis.ErrGeneric("must not set the field if consolidation is enabled", "ttlSecondsAfterEmpty"))
```
ellistarn, line 69: Does it make sense to use apis.ErrMultipleOneOf()

charts/karpenter/templates/clusterrole.yaml
tzneal: Yes, they're used to reduce our stabilization time (i.e. if there are no pending pods and daemonsets, replicasets, replicationcontrollers are all reporting the correct number of ready replicas then we can consider consolidating now as opposed to continuing to wait).

ellistarn: Generally, the less coupled Karpenter is to higher level controllers the better.

pkg/controllers/consolidation/pdblimits.go
```diff
@@ -0,0 +1,86 @@
...
+				if pdb.disruptionsAllowed == 0 {
+					return false
+				}
+			}
+		}
+	}
+	return true
+}
+
+type pdbItem struct {
```
ellistarn, line 70: Why not just rely on policyv1.PodDisruptionBudget if it's all the same data, just formatted differently.

charts/karpenter/templates/clusterrole.yaml
ellistarn: Could we track pod creation rates? e.g., if we haven't seen a pod created in n seconds, we can assume that all controllers are idle instead of checking specifically for a set of controllers.

pkg/controllers/consolidation/pdblimits.go
ellistarn: Check out https://github.com/kubernetes/apimachinery/blob/97e5df2d0258ac077093867fabf508c1fc31f53b/pkg/labels/labels.go#L37.

Should be able to convert from a map into a selector without an errorable in the middle.

pkg/controllers/consolidation/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+
+var consolidationNodesCreatedCounter = prometheus.NewCounter(
+	prometheus.CounterOpts{
+		Namespace: metrics.Namespace,
+		Subsystem: "consolidation",
+		Name:      "nodes_created",
+		Help:      "Number of nodes created in total by consolidation.",
+	},
+)
+var consolidationNodesTerminatedCounter = prometheus.NewCounter(
```
ellistarn, line 60: This collides with Jason's PR. Worth syncing with him.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -133,23 +133,35 @@ func (p *Provisioner) Provision(ctx context.Context) error {
 		return nil
 	}
 
+	_, err = p.LaunchNodes(ctx, LaunchOptions{RecordPodNomination: true}, nodes...)
+	return err
+}
+
+type LaunchOptions struct {
+	// RecordPodNomination causes nominate pod events to be recorded against the node.
+	RecordPodNomination bool
```
ellistarn, line 142: Thoughts on checking `pod.spec.nodeName` instead of wiring through this struct.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -33,7 +34,25 @@ import (
 	"github.com/aws/karpenter/pkg/utils/resources"
 )
 
-func NewScheduler(ctx context.Context, kubeClient client.Client, nodeTemplates []*scheduling.NodeTemplate, provisioners []v1alpha5.Provisioner, cluster *state.Cluster, topology *Topology, instanceTypes map[string][]cloudprovider.InstanceType, daemonOverhead map[*scheduling.NodeTemplate]v1.ResourceList, recorder events.Recorder) *Scheduler {
+// SchedulerOptions can be used to control the scheduling, these options are currently only used during consolidation.
+type SchedulerOptions struct {
+	// SimulationMode if true will prevent recording of the pod nomination decisions as events
+	SimulationMode bool
```
ellistarn, line 40: Same here re: pod.spec.nodeName?

offzale:
@tzneal thanks for the quick fix, the new version has reduced the CPU usage to a 50%.

charts/karpenter/templates/clusterrole.yaml
ellistarn: Otherwise we probably also want to add stateful sets to this list.

tzneal: Thanks, added statefulsets to what we check for.

pkg/controllers/consolidation/pdblimits.go
tzneal: That takes a `labels.Set` not a `metav1.LabelSelector`

pkg/controllers/consolidation/metrics.go
tzneal: This counts nodes terminated for consolidation purposes, there's another counter for nodes created for consolidation.  The idea was that you can track over time what consolidation specifically is doing to your cluster.

pkg/controllers/provisioning/provisioner.go
tzneal: I prefer it explicit.

pkg/controllers/consolidation/metrics.go
ellistarn: I wonder if we should be marking the node with a deletion reason e.g., emptiness, consolidation, expiry, and track that with a single gaugevec.

pkg/controllers/consolidation/controller.go
```diff
@@ -0,0 +1,665 @@
...
+	return ProcessResultNothingToDo
+}
+
+// candidateNodes returns nodes that appear to be currently consolidatable based off of their provisioner
+func (c *Controller) candidateNodes(ctx context.Context) []candidateNode {
+	var nodes []candidateNode
+	uninitializedNodeExists := false
+	c.cluster.ForEachNode(func(n *state.Node) bool {
+		// skip any nodes that we can't determine the instance of or for which we don't have consolidation enabled
+		if n.InstanceType == nil || n.Provisioner == nil || n.Provisioner.Spec.Consolidation == nil || !aws.BoolValue(n.Provisioner.Spec.Consolidation.Enabled) {
```
ellistarn, line 180: What do you think about using a label on the node instead of looking up the provisioner? We used to do a similar approach for emptiness/expiry. This helps to decouple consolidation from provisioners and gives users a hook to elect (or exempt) a node from consolidation. 

We will have the issue that a provisioner that toggles consolidation may drift from nodes, but we have other work in flight to catch this case.

pkg/controllers/consolidation/controller.go
```diff
@@ -0,0 +1,665 @@
...
+		}
+
+		// skip nodes that the scheduler thinks will soon have pending pods bound to them
+		if c.cluster.IsNodeNominated(n.Node.Name) {
+			return true
+		}
+
+		pods, err := c.getNodePods(ctx, n.Node.Name)
+		if err != nil {
+			logging.FromContext(ctx).Errorf("Determining node pods, %s", err)
```
ellistarn, line 204: How spammy is this log?

pkg/controllers/consolidation/controller.go
```diff
@@ -0,0 +1,665 @@
...
+	case consolidateResultUnknown:
+		return "Unknown"
+	case consolidateResultNotPossible:
+		return "Not Possible"
+	case consolidateResultDelete:
+		return "Delete"
+	case consolidateResultDeleteEmpty:
+		return "Delete (empty node)"
+	case consolidateResultShrink:
+		return "Shrink"
```
ellistarn, line 386: Since this is a customer facing term, I'd like to advocate for calling this "Replace". Specifically, we won't necessarily be shrinking the node, especially not across all dimensions. i.e., I'd expect something in our  docs like "Karpenter will replace a node if the workloads on it would be supported by a cheaper one."

pkg/apis/provisioning/v1alpha5/provisioner.go
tzneal: They seem fairly common:

```
find ~/go/pkg/mod/k8s.io/api@v0.24.3 -iname "*.proto" -exec grep "optional bool" {} + | wc -l
     144
```

There is a boolean in Node, Pod, Deployment, ConfigMap, and Secret among others.  This was also the interface in the review design and I don't plan to diverge from the reviewed design in this PR unless there is a clear improvement that can be made.

pkg/cloudprovider/types.go
```diff
@@ -48,6 +48,13 @@ type CloudProvider interface {
 	// availability, the GetInstanceTypes method should always return all instance types,
 	// even those with no offerings available.
 	GetInstanceTypes(context.Context, *v1alpha5.Provisioner) ([]InstanceType, error)
+	// ShouldConsolidate should return true if the cloud provider believes that consolidation should replace the given
+	// instance type with one of the instance type options provided. The instance types provided will all be cheaper
+	// than the node existing node.  This method allows the cloud provider to provide insight to the consolidation
+	// algorithm beyond price, specifically regarding if a given node replacement decision is good with respect to the
+	// availability of the instance type options provided. Both the newNodeRequirements and instanceTypeOptions will be
+	// nil in the case of deleting a node for consolidation.
+	ShouldConsolidate(existingNode *v1.Node, newNodeRequirements scheduling.Requirements, instanceTypeOptions []InstanceType) bool
```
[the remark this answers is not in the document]
tzneal: I'm open to changing this as it wasn't covered in the design.  There are two uses for this currently:

- Not replacing a spot node with a cheaper spot node due to inability to determine if the cheaper node is as or more available than the node being replaced
- Not replacing a node when all of the replacements are de-prioritized.  In provisioning we remove de-prioritized nodes from our instance types unless they are the only available option. When shrinking a node, we filter by price which can cause the normally de-prioritized node to not be de-prioritized.

pkg/controllers/consolidation/metrics.go
tzneal: There is a `consolidationActionsPerformedCounter` that tracks the consolidations actions performed (shrink/delete/delete empty)  by action.  If someone wants to rework metrics in a follow-up PR, I'm fine with that but I want to limit the scope of this one.

pkg/controllers/consolidation/controller.go
tzneal: That's a large departure from the design document.  Someone can experiment with other options in a follow-on PR.

pkg/controllers/consolidation/controller.go
tzneal: Seems fine to me.

pkg/apis/provisioning/v1alpha5/labels.go
tzneal: I prefer an explicit enabling/disabling of consolidation rather than having a magic label to turn it on/off.  That's also large departure from the design document. Someone can experiment with other options in a follow-on PR.

pkg/controllers/consolidation/controller.go
tzneal: Sure, this is the result of a scale-down that ended up deleting two nodes and replacing one:

```
karpenter-68d787db56-6x6q8 controller 2022-07-18T02:44:31.214Z	INFO	controller.consolidation	Consolidating via Delete, terminating 1 nodes ip-192-168-166-214.us-west-2.compute.internal/c6g.4xlarge	{"commit": "63daf4b"}
karpenter-68d787db56-6x6q8 controller 2022-07-18T02:44:31.248Z	INFO	controller.termination	Cordoned node	{"commit": "63daf4b", "node": "ip-192-168-166-214.us-west-2.compute.internal"}
karpenter-68d787db56-6x6q8 controller 2022-07-18T02:44:31.577Z	INFO	controller.termination	Deleted node	{"commit": "63daf4b", "node": "ip-192-168-166-214.us-west-2.compute.internal"}
karpenter-68d787db56-6x6q8 controller 2022-07-18T02:44:41.344Z	INFO	controller.consolidation	Consolidating via Delete, terminating 1 nodes ip-192-168-190-133.us-west-2.compute.internal/t4g.xlarge	{"commit": "63daf4b"}
karpenter-68d787db56-6x6q8 controller 2022-07-18T02:44:41.385Z	INFO	controller.termination	Cordoned node	{"commit": "63daf4b", "node": "ip-192-168-190-133.us-west-2.compute.internal"}
karpenter-68d787db56-6x6q8 controller 2022-07-18T02:44:41.678Z	INFO	controller.termination	Deleted node	{"commit": "63daf4b", "node": "ip-192-168-190-133.us-west-2.compute.internal"}
karpenter-68d787db56-6x6q8 controller 2022-07-18T02:45:12.882Z	INFO	controller.consolidation	Consolidating via Replace, terminating 1 nodes ip-192-168-191-127.us-west-2.compute.internal/c6g.4xlarge and replacing with a node from types a1.2xlarge, t4g.2xlarge, c6g.2xlarge, c7g.2xlarge, t3a.2xlarge and 37 other(s)	{"commit": "63daf4b"}
karpenter-68d787db56-6x6q8 controller 2022-07-18T02:45:15.281Z	INFO	controller.consolidation.cloudprovider	Launched instance: i-01882e52c50acf61b, hostname: ip-192-168-31-116.us-west-2.compute.internal, type: t4g.2xlarge, zone: us-west-2d, capacityType: on-demand	{"commit": "63daf4b", "provisioner": "default"}
karpenter-68d787db56-6x6q8 controller 2022-07-18T02:45:15.286Z	INFO	controller.consolidation	Created node with 5 pods requesting {"cpu":"5125m","pods":"10"} from types a1.2xlarge, t4g.2xlarge, c6g.2xlarge, c7g.2xlarge, t3a.2xlarge and 37 other(s)	{"commit": "63daf4b", "provisioner": "default"}
karpenter-68d787db56-6x6q8 controller 2022-07-18T02:46:19.967Z	INFO	controller.termination	Deleted node	{"commit": "63daf4b", "node": "ip-192-168-191-127.us-west-2.compute.internal"}
```

charts/karpenter/templates/clusterrole.yaml
ellistarn: > Could we track pod creation rates? e.g., if we haven't seen a pod created in n seconds, we can assume that all controllers are idle instead of checking specifically for a set of controllers.

I'd love to get an answer for this.

tzneal: That's an alternative approach but I don't see it as being a better solution. A pause in pod creation doesn't mean that those pods are now ready and their requisite controllers healthy. There are trade-offs to both approaches.

pkg/controllers/consolidation/controller.go
tzneal: No, I can collect another log with debug.

ellistarn:
Discussed offline:

Cloud Provider API
- We can apply the spot assumption (do not consolidate spot) to all vendors, until it becomes a problem. 
- We can solve the prioritization problem by defaulting families/sizes in our vendor specific webhook. This improves UX in other ways. Not default instance types may still be configured at the provisioner level

TTLAfterUnderutilized
- Currently we're losing the ability to wait after nodes become empty. The cluster drains very quickly.
- There are algorithmic complexities to applying this rule to nodes with >0 pods
- Having a TTL for 0 pods on a node, but no TTL for 1 pod on a node is a very jarring UX
- We'll take this one as a fast follow, since it's nontrivial work that needs to be explored.

pkg/controllers/consolidation/controller.go
ellistarn: https://github.com/aws/karpenter/pull/2123#issuecomment-1188258308

pkg/controllers/consolidation/metrics.go
ellistarn: SGTM. Let's cut a followup issue.

pkg/cloudprovider/types.go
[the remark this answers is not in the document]
ellistarn: https://github.com/aws/karpenter/pull/2123#issuecomment-1188258308

pkg/apis/provisioning/v1alpha5/provisioner.go
ellistarn: https://github.com/aws/karpenter/pull/2123#issuecomment-1188258308

pkg/controllers/termination/eviction.go
```diff
@@ -112,6 +112,5 @@ func (e *EvictionQueue) evict(ctx context.Context, nn types.NamespacedName) bool
 		logging.FromContext(ctx).Error(err)
 		return false
 	}
-	logging.FromContext(ctx).Debug("Evicted pod")
```
njtran, line 115: I don't think we want to remove this. Have you found that these logs are way too spammy? If we don't have the logs here, we should have record that we're evicting them somewhere else.

pkg/events/dedupe.go
```diff
@@ -34,29 +34,52 @@ type dedupe struct {
...
 	d.rec.NodeFailedToDrain(node, err)
 }
+
+func (d *dedupe) shouldCreateEvent(key string) bool {
+	if _, exists := d.cache.Get(key); exists {
+		return false
+	}
+	d.cache.SetDefault(key, nil)
+	return true
+}
```
njtran, line 85: nit: need a newline

pkg/controllers/consolidation/controller.go
```diff
@@ -0,0 +1,665 @@
...
+			c.recorder.WaitingOnReadinessForConsolidation(&k8Node)
+			return errors.New("node is not initialized")
+		}
+		return nil
+	}, retry.Delay(2*time.Second),
+		retry.LastErrorOnly(true),
+		retry.Attempts(30),
+		retry.MaxDelay(10*time.Second), // ~ 4.5 minutes in total
+	); err != nil {
+		// node never become ready, so uncordon the node we were trying to delete and report the error
```
njtran, line 312: Can we check that the nodeDeletionTimestamp is not nil here before we uncordon? Worried about the case where a user decides to `kubectl delete node` a node and we uncordon it, allowing pods to schedule while we're trying to drain it.

njtran (commented):
Really nice work! I had some comments and questions, but excited to see this get merged!

pkg/controllers/consolidation/controller.go
```diff
@@ -0,0 +1,665 @@
...
+
+// ProcessCluster is exposed for unit testing purposes
+func (c *Controller) ProcessCluster(ctx context.Context) ProcessResult {
+	candidates := c.candidateNodes(ctx)
+	if len(candidates) == 0 {
+		return ProcessResultNothingToDo
+	}
+
+	emptyNodes := lo.Filter(candidates, func(n candidateNode, _ int) bool { return len(n.pods) == 0 })
+	// first see if there are empty nodes that we can delete immediately, and if so delete them all at once
```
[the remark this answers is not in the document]
tzneal: Adding a `ttlUntilConsolidatable` is complicated and diverges from the reviewed design.  See Ellis's note above where it's called `TTLAfterUnderutilized`.

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
```diff
@@ -62,6 +64,10 @@ func (s *ProvisionerSpec) validateTTLSecondsAfterEmpty() (errs *apis.FieldError)
 	if ptr.Int64Value(s.TTLSecondsAfterEmpty) < 0 {
 		return errs.Also(apis.ErrInvalidValue("cannot be negative", "ttlSecondsAfterEmpty"))
 	}
+	// TTLSecondsAfterEmpty and consolidation are mutually exclusive
```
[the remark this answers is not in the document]
tzneal: As far as I know, controller-gen doesn't support this.

pkg/controllers/consolidation/controller.go
tzneal:
```
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:30.491Z	INFO	controller.consolidation	Consolidating via Delete (empty node), terminating 1 nodes ip-192-168-189-114.us-west-2.compute.internal/c6g.8xlarge	{"commit": "ece47b1"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:30.491Z	DEBUG	controller.events	Normal	{"commit": "ece47b1", "object": {"kind":"Node","name":"ip-192-168-189-114.us-west-2.compute.internal","uid":"1e1bc41d-e1c4-402a-a4b0-22ee79d8c882","apiVersion":"v1","resourceVersion":"52764954"}, "reason": "ConsolidateTerminateNode", "message": "Consolidating node via Delete (empty node), terminating 1 nodes ip-192-168-189-114.us-west-2.compute.internal/c6g.8xlarge"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:30.542Z	INFO	controller.termination	Cordoned node	{"commit": "ece47b1", "node": "ip-192-168-189-114.us-west-2.compute.internal"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:30.726Z	INFO	controller.termination	Deleted node	{"commit": "ece47b1", "node": "ip-192-168-189-114.us-west-2.compute.internal"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:40.815Z	INFO	controller.consolidation	Consolidating via Replace, terminating 1 nodes ip-192-168-81-216.us-west-2.compute.internal/c4.8xlarge and replacing with a node from types c6g.8xlarge, c7g.8xlarge, c6a.8xlarge, c6gd.8xlarge, m6g.8xlarge and 10 other(s)	{"commit": "ece47b1"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:41.036Z	DEBUG	controller.consolidation.cloudprovider	Discovered security groups: [sg-05c364fe5ccfc49b4 sg-05f88204faa8cf2d4]	{"commit": "ece47b1", "provisioner": "default"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:41.038Z	DEBUG	controller.consolidation.cloudprovider	Discovered kubernetes version 1.21	{"commit": "ece47b1", "provisioner": "default"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:41.087Z	DEBUG	controller.consolidation.cloudprovider	Discovered ami-064ac13bc8e3c2ee1 for query "/aws/service/eks/optimized-ami/1.21/amazon-linux-2-arm64/recommended/image_id"	{"commit": "ece47b1", "provisioner": "default"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:41.113Z	DEBUG	controller.consolidation.cloudprovider	Discovered ami-00cf63b12c53803a5 for query "/aws/service/eks/optimized-ami/1.21/amazon-linux-2/recommended/image_id"	{"commit": "ece47b1", "provisioner": "default"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:43.191Z	INFO	controller.consolidation.cloudprovider	Launched instance: i-0d136b6c50015d855, hostname: ip-192-168-187-186.us-west-2.compute.internal, type: c6g.8xlarge, zone: us-west-2a, capacityType: on-demand	{"commit": "ece47b1", "provisioner": "default"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:43.250Z	INFO	controller.consolidation	Created node with 22 pods requesting {"cpu":"22125m","pods":"27"} from types c6g.8xlarge, c7g.8xlarge, c6a.8xlarge, c6gd.8xlarge, m6g.8xlarge and 10 other(s)	{"commit": "ece47b1", "provisioner": "default"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:43.250Z	DEBUG	controller.events	Normal	{"commit": "ece47b1", "object": {"kind":"Node","name":"ip-192-168-187-186.us-west-2.compute.internal","uid":"0dbb680e-2c7c-4fe0-aef6-e625e95e1eef","apiVersion":"v1","resourceVersion":"52765122"}, "reason": "ConsolidateLaunchNode", "message": "Launching node for Replace, terminating 1 nodes ip-192-168-81-216.us-west-2.compute.internal/c4.8xlarge and replacing with a node from types c6g.8xlarge, c7g.8xlarge, c6a.8xlarge, c6gd.8xlarge, m6g.8xlarge and 10 other(s)"}
karpenter-68d7cfc48c-rfdg2 controller 2022-07-19T16:17:43.250Z	DEBUG	controller.events	Normal	{"commit": "ece47b1", "object": {"kind":"Node","name":"ip-192-168-187-186.us-west-2.compute.internal","uid":"0dbb680e-2c7c-4fe0-aef6-e625e95e1eef","apiVersion":"v1","resourceVersion":"52765122"}, "reason": "ConsolidateWaiting", "message": "Waiting on readiness to continue consolidation"}
```

pkg/controllers/counter/controller.go
```diff
@@ -79,6 +79,10 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	}
 	provisioner.Status.Resources = resourceCounts
 	if err := c.kubeClient.Status().Patch(ctx, provisioner, client.MergeFrom(persisted)); err != nil {
+		// provisioner was deleted
+		if errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
njtran: Ahhh I see. Thanks 👍

pkg/controllers/consolidation/controller.go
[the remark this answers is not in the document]
njtran: Ah thanks. sgtm

pkg/controllers/termination/eviction.go
njtran: Makes sense. Some thoughts:
- Do we want to maybe investigate a different level of logging? I know we've been only using debug and info, but we're making a lot of logging now, so may be useful. 
- Maybe a good in-between is when we decide to delete a node, we just log the list of pods we're going to evict, so at least we can find this in the logs tied to the node. Another idea is to export these as events somewhere. Mostly just thinking out loud.

charts/karpenter/templates/clusterrole.yaml
```diff
@@ -22,12 +22,18 @@ rules:
   - apiGroups: ["storage.k8s.io"]
     resources: ["storageclasses", "csinodes"]
     verbs: ["get", "watch", "list"]
+  - apiGroups: [""]
+    resources: ["replicationcontrollers"]
```
ellistarn, line 26: Nitpicking because these files are critically important to customers. Can we move this to line 20. This file is organized into objects that share the same API Group and Permissions, and replication controllers is equivalent to pods/nodes/persistentvolumes/persistentvolumeclaims.

List is a superset of get, so you could remove the `get` permission above, or explicitly add it to the `replicationcontroller` rule here.

pkg/controllers/termination/eviction.go
ellistarn: > Do we want to maybe investigate a different level of logging? I know we've been only using debug and info, but we're making a lot of logging now, so may be useful.

I'm cautious of this, as it gets quite subjective when to make something trace or debug. Further, it adds an additional thing for customers to think about. Logging is notoriously subjective and requires judgement required on the part of the author. In this case, I think there's enough logging via kube events re: eviction that it seems reasonable to drop this log if it's spammy.

pkg/apis/provisioning/v1alpha5/provisioner_validation.go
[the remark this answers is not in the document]
ellistarn: Great suggestion, but I'm cautious of scheme-level validation. It feels a bit half baked and it's challenging to model enough of your validation using it that it becomes worth it. Given that we're already using a webhook, I'd lean towards making our validation logic explicit and avoiding schema-level validation (one thing to think about vs two)

pkg/controllers/consolidation/controller.go
```diff
@@ -0,0 +1,552 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+func (c *Controller) stabilizationWindow(ctx context.Context) time.Duration {
+	// no pending pods, and all replica sets/replication controllers are reporting ready so quickly consider another consolidation
+	if !c.hasPendingPods(ctx) && c.replicaSetsReady(ctx) &&
+		c.replicationControllersReady(ctx) && c.statefulSetsReady(ctx) {
```
mikkeloscar, line 524: What is the motivation for this check and to have a 5 min. delay if it doesn't pass?

I worry that a single broken deployment can block/delay consolidation, which could be quite expensive if you have a huge scale up and later a big scale down leading to many nodes needing to be consolidated.

In our environment we have many teams deploying to a single cluster and some teams may have a broken deployment which is not getting fixed because there is no impact from their perspective, ofc. this could be considered a problem, but that it can block/delay the consolidation is probably a side effect that is not obvious to those teams/users.

The check for pending pods seems reasonable as it doesn't consider pods that are unschedulable to my understanding. :+1:

mikkeloscar: I imagine this is also a problem if you have rolling updates of deployments. We have many teams deploying all the time during the day, so if a rolling update would lead to a 5 min. delay that wouldn't be ideal.

tzneal: We are trying to be conservative during consolidation so that we don't disrupt the cluster too much but still drive towards a lower cost.  This looks at pending pods, deployments, etc. in an attempt to speed up consolidation if the overall cluster state looks good with respect to the workloads running on it.

Even with the 5 minute delay, you're still looking at consolidating ~12 nodes per hour or ~288 nodes per day.  We also consolidate empty nodes in a single batch which should help.

There may be tunables in the future, but in the initial version we are trying to provide a simple interface.

mikkeloscar: We currently can scale 100s of nodes within minutes and expect them to be gone/consolidated soon after. Waiting hours/days for complete consolidation seems like a lot :)

This hardcoded 5 min. also have the effect that if nodes are empty within this time they have to wait the full window before being consolidated from my understanding. With `TTLSecondsAfterEmpty` you could reduce this to seconds, but that is no longer possible with consolidation enabled.

I can also fully understand starting with a simple interface, just trying to give a bit of feedback from a user perspective as we're eagerly awaiting this consolidation feature for likely replacing the cluster-autoscaler. For context we run 200+ Kubernetes clusters (test+prod) with up to 1000 nodes per cluster (scaling a lot during the day), so reducing the time nodes are running empty has a big compounding cost efficiency effect. :)


I'm really impressed with what karpenter has become by now with this feature, and the potential that it has compared to the cluster-autoscaler. Great work! :+1:

tzneal: There's definitely some trade-off in just about everything. I hope to get another snapshot out tomorrow (they're linked on the issue at #1091).  If you've got a dev/test cluster available that isn't critical, we would welcome some feedback on how it performs for you.

pkg/controllers/consolidation/controller.go
```diff
@@ -0,0 +1,665 @@
...
+	}
+	for i := range podList.Items {
+		if pod.IsProvisionable(&podList.Items[i]) {
+			return true
+		}
+	}
+	return false
+}
+
+func (c *Controller) replicaSetsReady(ctx context.Context) bool {
```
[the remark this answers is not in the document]
ellistarn: If you really wanted to be fancy, you could use the https://pkg.go.dev/k8s.io/api/autoscaling/v1#Scale subresource to check this generically, but 🤷

pkg/controllers/consolidation/pdblimits.go
```diff
@@ -0,0 +1,86 @@
...
+limitations under the License.
+*/
+
+package consolidation
+
+import (
+	"context"
+
+	v1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1beta1"
```
olemarkus, line 21: Just not to lose this: https://github.com/aws/karpenter/pull/2123#discussion_r921393771

What's the strategy for supporting newer k8s versions where this has been removed?

psuj:
hello @tzneal, great job, I am really looking forward for this improvement - are you able to share any estimation about how much effort is left before releasing this awesome feature?

bwagner5 (approved):
Nice! LGTM

OrIlani055:
@tzneal we're experiencing problem with the new consolidation feature, while using the consolidation flag true on the provisioner.yaml 

this is our provisioner file:

```
kind: Provisioner
metadata:
  name: provisioner
spec:
   consolidation:
     enabled: true
  requirements:
    - key: "topology.kubernetes.io/zone"
      operator: In
      values: {{ .Values.karpenter.zones }}
    - key: "karpenter.sh/capacity-type" # If not included, the webhook for the AWS cloud provider will default to on-demand
      operator: In
      values: ["spot"]
      operator: In
      values: ["t3a.2xlarge", "t3a.xlarge","c5.2xlarge"]
  limits:
    resources:
      cpu: 1000
  provider:
    launchTemplate: {{ .Values.karpenter.lt }}
    subnetSelector:
      Name: {{ .Values.karpenter.subnets }}
    tags:
      Name: {{ .Values.karpenter.tagname }}
  ttlSecondsAfterEmpty: 30
  ttlSecondsUntilExpired: 86400
```


and the error we're getting is :
`one or more objects failed to apply, reason: admission webhook "defaulting.webhook.provisioners.karpenter.sh" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "consolidation". `

francis-jjk:
Is there any design doc for this feature?

FernandoMiguel:
> Is there any design doc for this feature?

@francis-jjk https://github.com/aws/karpenter/blob/main/designs/consolidation.md

psuj:
@OrIlani055 perhaps you need to update CRDs on your cluster as stated in the upgrade guide - https://karpenter.sh/v0.15.0/upgrade-guide/#custom-resource-definition-crd-upgrades

also keep in mind that you need to delete `ttlSecondsAfterEmpty` if consolidation is enabled

PR #2127, opened 2022-07-13
Test: Implemented pipeline for integration test suite

ellistarn:
Fixes # 

**Description**
Added a test pipeline for the integration suite

* Manually using tekton

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
None
```


test/tasks/karpenter-test-suite.yaml
```diff
@@ -0,0 +1,127 @@
...
+kind: Task
+metadata:
+  name: karpenter-test-suite
+  namespace: karpenter-tests
+spec:
+  description: |
+    Execute a karpenter test suite
+  params:
+    - name: cluster-name
+      default: $(context.taskRun.name)
```
njtran, line 12: Can we pre-fix this with more metadata? If we have many of these running in our cluster, we'll need an easier way to distinguish different which cluster was created with which parameters.

test/tasks/karpenter-test-suite.yaml
```diff
@@ -0,0 +1,127 @@
...
+    - name: repository
+      default: github.com/aws/karpenter
+    - name: tag
+      default: v0.13.2
```
njtran, line 16: Can we default to latest since we're testing ourselves?

test/tasks/karpenter-test-suite.yaml
```diff
@@ -0,0 +1,127 @@
...
+      git clone https://$(params.repository).git .
+      git checkout $(params.tag)
+
+  - name: create-iam-policies
+    image: amazon/aws-cli:2.7.11
+    workingDir: $(workspaces.ws.path)/$(params.repository)
+    script: |
+      aws cloudformation deploy \
+        --stack-name "$(params.cluster-name)" \
+        --template-file website/content/en/preview/getting-started/getting-started-with-eksctl/cloudformation.yaml \
```
njtran, line 48:
```suggestion
        --template-file website/content/en/preview/$(params.tag)/getting-started-with-eksctl/cloudformation.yaml \
```

test/tasks/karpenter-test-suite.yaml
```diff
@@ -0,0 +1,127 @@
...
+        --capabilities CAPABILITY_NAMED_IAM \
+        --parameter-overrides "ClusterName=$(params.cluster-name)"
+
+  - name: create-cluster-configuration
+    image: alpine/curl:3.14
+    workingDir: $(workspaces.ws.path)/$(params.repository)
+    script: |
+      cat <<EOF > cluster.yaml
```
njtran, line 56: Any problems with putting this in the same step as create-cluster? The only time we should be getting an error here is with the image.

test/tasks/karpenter-test-suite.yaml
```diff
@@ -0,0 +1,127 @@
...
+    image: weaveworks/eksctl:v0.94.0
+    workingDir: $(workspaces.ws.path)/$(params.repository)
+    script: |
+      eksctl create cluster -f cluster.yaml || eksctl upgrade cluster -f cluster.yaml
+
+  - name: update-kubeconfig
+    image: amazon/aws-cli:2.7.11
+    script: |
+      aws eks update-kubeconfig --name $(params.cluster-name)
+      cat /root/.kube/config > $(results.kubeconfig.path)
```
njtran, line 87: I think this gives the world-readable and group-readable warnings. Can we chmod 600 just for good measure here?

test/tasks/karpenter-test-suite.yaml
```diff
@@ -0,0 +1,127 @@
...
+  workspaces:
+    - name: ws
+  steps:
+
+  - name: detect-environment
+    image: alpine/curl:3.14
+    workingDir: $(workspaces.ws.path)/$(params.repository)
+    script: |
+      TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")
+      curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/identity-credentials/ec2/info | grep "AccountId" | cut -f4 -d"\"" > $(results.account-id.path)
```
njtran, line 32: nice!

test/tasks/karpenter-test-suite.yaml
```diff
@@ -0,0 +1,127 @@
...
+      aws eks update-kubeconfig --name $(params.cluster-name)
+      cat /root/.kube/config > $(results.kubeconfig.path)
+
+  - name: create-iam-identity-mapping
+    image: weaveworks/eksctl:v0.94.0
+    script: |
+      eksctl create iamidentitymapping \
+      --username system:node:{{EC2PrivateDNSName}} \
+      --cluster "$(params.cluster-name)" \
+      --arn "arn:aws:iam::$(cat $(results.region.account-id)):role/KarpenterNodeRole-$(params.cluster-name)" \
```
njtran, line 95:
```suggestion
      --arn "arn:aws:iam::$(cat $(results.account-id.path)):role/KarpenterNodeRole-$(params.cluster-name)" \
```

test/tasks/karpenter-test-suite.yaml
```diff
@@ -0,0 +1,127 @@
...
+    image: alpine/helm:3.9.0
+    script: |
+      helm repo add karpenter https://charts.karpenter.sh/
+      helm repo update
+      helm upgrade --install --namespace karpenter --create-namespace karpenter karpenter/karpenter \
+        --kubeconfig $(results.kubeconfig.path) \
+        --version $(params.tag) \
+        --set serviceAccount.annotations."eks\.amazonaws\.com/role-arn"="arn:aws:iam::$(cat $(results.account-id.path)):role/$(params.cluster-name)-karpenter" \
+        --set clusterName=$(params.cluster-name) \
+        --set clusterEndpoint=$(cat $(results.kubeconfig.path) | grep server | awk '{print $2}') \
```
njtran, line 119: Since Tekton bypasses the image entry point, does this work? Should we #!/usr/bin/env bash the top of this step to make sure?

test/tasks/karpenter-test-suite.yaml
```diff
@@ -0,0 +1,127 @@
...
+    Execute a karpenter test suite
+  params:
+    - name: cluster-name
+      default: $(context.taskRun.name)
+    - name: repository
+      default: github.com/aws/karpenter
+    - name: tag
+      default: v0.13.2
+    - name: command
+      default: go test ./pkg/...
```
njtran, line 18: Should this be the following?
```suggestion
      default: go test ./test/pkg/...
```

test/tasks/karpenter-test-suite.yaml
ellistarn: Good idea. I'm no longer curling in here.

test/tasks/karpenter-test-suite.yaml
ellistarn: This is an `eks update-kubeconfig` issue that will eventually be fixed (i hope). I'd prefer to keep it simple, even if it logs.

test/tasks/karpenter-test-suite.yaml
ellistarn: It should have pipeline-name-task-name prefixed once this is fully wired up.

test/tasks/after-suite.yaml
```diff
@@ -0,0 +1,26 @@
...
+    Executes after a test suite to clean up resources
+  params:
+    - name: cluster-name
+      description: Name of the cluster under test.
+  workspaces:
+    - name: ws
+  steps:
+
+  - name: delete-cluster
+    image: 767520670908.dkr.ecr.us-west-2.amazonaws.com/karpenter-test-infrastructure:v4
```
njtran, line 18: This looks like private ECR. Is it possible for us to use public-ecr here?

test/tasks/after-suite.yaml
```diff
@@ -0,0 +1,26 @@
...
+    script: |
+      eksctl delete cluster --name $(params.cluster-name) --wait || true
+
+  - name: delete-iam-policies
```
njtran, line 22: Should we reverse the order of these steps? IIUC we should delete in the reverse order that we create to ensure that there are no blocking dependencies.

test/tasks/before-suite.yaml
```diff
@@ -0,0 +1,107 @@
...
+    - name: region
+  workspaces:
+    - name: ws
+  steps:
+  - name: detect-environment
+    image: 767520670908.dkr.ecr.us-west-2.amazonaws.com/karpenter-test-infrastructure:v4
+    workingDir: $(workspaces.ws.path)
+    script: |
+      TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")
+      curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/identity-credentials/ec2/info | grep "AccountId" | cut -f4 -d"\"" > $(results.account-id.path)
```
njtran, line 26: https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/instance-identity-documents.html
Can we use jq instead of grep here for robustness?

test/tasks/before-suite.yaml
```diff
@@ -0,0 +1,107 @@
...
+      curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/placement/region > $(results.region.path)
+
+  - name: checkout-repository
+    image: alpine/git
+    workingDir: $(workspaces.ws.path)
+    script: |
+      git clone https://github.com/aws/karpenter.git .
```
njtran, line 33: Just a thought: since we don't have an ssh key here, can we get throttled?

test/tasks/before-suite.yaml
```diff
@@ -0,0 +1,107 @@
...
+      git checkout $(params.git-ref)
+
+  - name: create-iam-policies
+    image: 767520670908.dkr.ecr.us-west-2.amazonaws.com/karpenter-test-infrastructure:v4
+    workingDir: $(workspaces.ws.path)
+    script: |
+      aws cloudformation deploy \
+        --stack-name KarpenterIAM-$(params.cluster-name) \
```
njtran, line 41: Is there a reason to divert from the naming scheme in our [getting-started guides](https://karpenter.sh/v0.13.2/getting-started/getting-started-with-eksctl/#create-the-karpenternode-iam-role)? 
```
  --stack-name "Karpenter-${CLUSTER_NAME}" \
```

test/tasks/before-suite.yaml
```diff
@@ -0,0 +1,107 @@
...
+        --template-file website/content/en/preview/getting-started/getting-started-with-eksctl/cloudformation.yaml \
```
njtran, line 42: Any thoughts here? https://github.com/aws/karpenter/pull/2127#discussion_r922401825

test/tasks/before-suite.yaml
```diff
@@ -0,0 +1,107 @@
+# ginkgo-suite is a generic task that executes ginkgo on the given repository and ginkgo suite.
```
njtran, line 1: This looks very similar to the `create-eks.yaml` and `install-karpenter-eks.yaml` files in the examples folder. If we're going to use this, I think we can go ahead and remove those examples, since this looks much better catered to the dev loop.

test/tasks/before-suite.yaml
```diff
@@ -0,0 +1,107 @@
...
+          maxSize: 2
+          taints:
+          - key: CriticalAddonsOnly
+            value: "true"
+            effect: NoSchedule
+      iam:
+        withOIDC: true
+      EOF
+
+      eksctl create iamidentitymapping \
```
njtran, line 77: nit: Can we put this and the following command in another step? Something like `add-authentication`, since it's not actually creating a cluster.

test/tasks/suite.yaml
```diff
@@ -0,0 +1,33 @@
...
+  name: suite
+  namespace: karpenter-tests
+spec:
+  description: |
+    Execute a karpenter test suite
+  params:
+    - name: cluster-name
+      description: Name of the cluster under test.
+    - name: suite-path
+      description: Path to the test suite
```
njtran, line 14: nit: at root
```suggestion
      description: Path to the test suite at root
```

test/tasks/suite.yaml
```diff
@@ -0,0 +1,33 @@
...
+      description: Git commit, tag, or branch to check out. Requires a corresponding Karpenter snapshot release.
+  workspaces:
+    - name: ws
+  steps:
+
+  - name: checkout-repository
+    image: alpine/git
+    workingDir: $(workspaces.ws.path)
+    script: |
+      git clone https://github.com/aws/karpenter.git .
```
njtran, line 25: we're cloning in both the `before-suite.yaml` and the `suite.yaml`. Do you know how long each clone takes? If we use PVCs for our workspaces instead of emptyDir, we can share the cloned repo and make it so we only do this call once, which I believe would probably speed up every test by a good amount. 

IMO, this logically isolates all setup work from testing better.

You can do this with volumeClaimTemplate: https://tekton.dev/docs/pipelines/workspaces/#example-pipelinerun-definition-using-workspaces

Note that this will also use hostname affinity to schedule the tasks onto the same nodes.

test/tools/Dockerfile
```diff
@@ -0,0 +1,48 @@
+# Please forgive me for this -- build AWS SDK v2 from source, see https://github.com/aws/aws-cli/issues/4685#issuecomment-1094307056
```
njtran, line 1: 🧙

test/tasks/after-suite.yaml
ellistarn: This is now `public.ecr.aws/karpenter-testing/tools:latest`

test/tasks/after-suite.yaml
ellistarn: The IAM policies are in use by the cluster, which I believe will block deletion.

test/tasks/after-suite.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - name: cluster-name
+      description: Name of the cluster under test.
+  workspaces:
+    - name: ws
+  steps:
+
+  - name: delete-cluster
+    image: 767520670908.dkr.ecr.us-west-2.amazonaws.com/karpenter-test-infrastructure:v4
+    script: |
+      eksctl delete cluster --name $(params.cluster-name) --wait || true
```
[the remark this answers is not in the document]
ellistarn: It does.

test/tasks/after-suite.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+  - name: delete-iam-policies
+    image: 767520670908.dkr.ecr.us-west-2.amazonaws.com/karpenter-test-infrastructure:v4
+    script: |
+      aws cloudformation delete-stack --stack-name KarpenterIAM-$(params.cluster-name)
+      aws cloudformation wait stack-delete-complete --stack-name KarpenterIAM-$(params.cluster-name)
```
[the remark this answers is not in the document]
ellistarn: It's possible, but we should be able to retry this task.

test/tasks/before-suite.yaml
ellistarn: Yes. I expect we'll deprecate all examples at some point. They were helpful for getting me started.

test/tasks/before-suite.yaml
ellistarn: I originally had this broken out into single command steps, but this ends up being really verbose in tekton. I'm not sure it's materially different, since it's all part of the core cluster configuration (and eksctl even handles the cascading deletes)

test/tasks/suite.yaml
ellistarn: I'm really wary of depending on EBS volumes due to the latency, leak potential, failure modes, etc. These clones are quite a bit faster than mounting the volume 

image

test/tasks/karpenter-test-suite.yaml
njtran: Since we're cloning the repo at a certain commit, this is resolved.

test/tasks/after-suite.yaml
njtran: You bring up a good point, if we have some instances that didn't get cleaned up in a test still using the policies, I believe it blocks.

test/tasks/before-suite.yaml
njtran: sgtm

test/tasks/suite.yaml
njtran: Wow, didn't realize it took only 2 seconds. Fine by me here.

test/suites/common/setup.yaml
```diff
@@ -0,0 +1,98 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: setup
+  namespace: karpenter-tests
+spec:
+  description: |
+    Execute a karpenter test suite
```
njtran, line 8: It's unclear to me when we'd use `setup.yaml` vs `install-karpenter.yaml`. Can you update the descriptions of these files to explain this? Would this be used to create an EKS cluster and install karpenter, and `install-karpenter.yaml` would be used on an existing cluster? If that's the case, is there use in de-duplicating this into a `create-eks.yaml` and `install-karpenter-eks.yaml` like in the examples?

test/tasks/karpenter-test-suite.yaml
ellistarn: Tekton allows UID, but the strings get so long that IAM starts to complain (max len 64)

test/suites/common/cleanup.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - name: cluster-name
+      description: Name of the cluster under test.
+  workspaces:
+    - name: ws
+  steps:
+
+  - name: delete-cluster
+    image: public.ecr.aws/karpenter-testing/tools:latest
+    script: |
+      eksctl delete cluster --name $(params.cluster-name) --wait || true
```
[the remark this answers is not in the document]
ellistarn: I actually don't want to continue on error. I really need an idempotent eksctl delete, but decided to not include in this PR. Let's iterate/improve on this.

test/suites/common/setup.yaml
```diff
@@ -0,0 +1,98 @@
...
+      cmd="create"
+      eksctl get cluster --name $(params.cluster-name) && cmd="upgrade"
+      eksctl ${cmd} cluster -f - <<EOF
+      ---
+      apiVersion: eksctl.io/v1alpha5
+      kind: ClusterConfig
+      metadata:
+        name: $(params.cluster-name)
+        region: ${REGION}
+        version: "1.21"
```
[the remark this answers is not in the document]
ellistarn: No -- I think this should be hard coded unless/until we do multi version testing. It's less complexity to just modify this in place and tekton is verbose enough that I'm trying to minimize parameterization wherever possible.

test/suites/common/setup.yaml
ellistarn: Meant to delete install-karpenter. It's deleted now.

test/tools/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+    helm@testing \
+    kubectl@testing
+
+# Install AWS CLI
+COPY --from=installer /usr/local/aws-cli/ /usr/local/aws-cli/
+COPY --from=installer /aws-cli-bin/ /usr/local/bin/
+RUN aws --version
+
+# Install eksctl
+RUN curl --silent --location "https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_$(uname -s)_amd64.tar.gz" | tar xz -C /tmp
```
njtran, line 47: nit: Can we pin this to a version? I think we tested v0.94.0 for the setup instructions. Latest is v0.105.0, and as more k8s/eksctl versions come out, older versions become unsupported.

https://github.com/weaveworks/eksctl/releases/download/v0.94.0/eksctl_Darwin_amd64.tar.gz

njtran (approved):
Looks good. One nit.

njtran (approved):
Cool! Thanks for the help :)

PR #2139, opened 2022-07-14
feat: add node termination metrics

jmdeal:
**Description**
Adds the `karpenter_nodes_termination_provisioner_time_seconds` and `karpenter_nodes_termination_time_seconds` metrics detailed in the metrics design doc. Also includes updates to the metric design docs.

* Tested scale-up / scale-down events in development cluster

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```


designs/metrics.md
```diff
@@ -72,10 +72,11 @@ Scheduling statistics consist of actions not representable by a single controlle
...
-| `karpenter_pods_startup_time_seconds`      | Summary | `provisioner`, `zone`, `arch`, `capacity_type`, `instance_type` | [Measurement Definitions](#measurment-definitions) |
+| Name                                                   | Type      | Labels                                                          | Description                                        |
+|--------------------------------------------------------|-----------|-----------------------------------------------------------------|----------------------------------------------------|
+| `karpenter_nodes_termination_provisioner_time_seconds` | Summary   | `provisioner`                                                   | [Measurement Definitions](#measurment-definitions) |
+| `karpenter_nodes_termination_time_seconds`             | Summary   |                                                                 | [Measurement Definitions](#measurment-definitions) |
+| `karpenter_pods_startup_time_seconds`                  | Summary   | `provisioner`, `zone`, `arch`, `capacity_type`, `instance_type` | [Measurement Definitions](#measurment-definitions) |
```
njtran, line 79: Like we discussed offline: let's keep these labels in, but when you add the metric in, filter down the labels to match the implementation.

designs/metrics.md
```diff
@@ -101,6 +102,10 @@ represents the number of occurrences that were less than that buckets value. Sin
 histograms are a poor choice for unbounded data. Summaries instead directly track quantiles. This involves more
 computation but works better for unbounded data.
 
+Summaries do have a flaw: they can not be aggregated over multiple metrics. An example of this problem can be found in 
+the metric for node termination latencies. If we want to track the termination latencies of all nodes and still be able
+to filter over provisioners, this must be instrumented as two seperate metric families.
```
njtran, line 107: Just a suggestion on wording that pulls from Prometheus Docs. Feel free to use what you want. 
```suggestion
Prometheus `Summary` metric types compute running quantiles for a given metric for given label values. Unlike histograms, summaries cannot be aggregated since it can result in statistically non-sensical results. Theoretically, summaries can be tracked for many labels, but we would need a number of summary metrics equal to the size of the cartesian product of all the labels, which is not viable.

Thus, if we want to track the termination latencies of all nodes and still be able
to filter over provisioners, this must be instrumented as two separate metric families.
```

pkg/metrics/constants.go
```diff
@@ -37,6 +37,15 @@ func DurationBuckets() []float64 {
 		1.25, 1.5, 1.75, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30, 40, 50, 60}
 }
 
+func SummaryObjectives() map[float64]float64 {
+	const epsilon = 0.01
```
njtran, line 41: What is this supposed to do? Can you add a comment here?

njtran (commented):
Nice. Can you add some images of the dashboards you've used to test? You could also create a restrictive PDB to add some higher variance to the numbers you've been getting.

pkg/controllers/metrics/termination/controller.go
```diff
@@ -0,0 +1,139 @@
...
+		[]string{},
+	)
+)
+
+func init() {
+	crmetrics.Registry.MustRegister(terminationProvisionerSummaryVec)
+	crmetrics.Registry.MustRegister(terminationSummaryVec)
+}
+
+type Controller struct {
```
ellistarn, line 72: I worry a bit about the additional processing and memory of this controller. We're caching the full cluster's nodes here, and reconciling on every single node object update. Does it make more sense to put these metrics in the termination controller, where we're directly handling the termination of nodes?

pkg/controllers/metrics/termination/controller.go
```diff
@@ -0,0 +1,139 @@
...
+	provisioner    = "provisioner"
+)
+
+var (
+	terminationProvisionerSummaryVec = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Namespace:  "karpenter",
+			Subsystem:  "nodes",
+			Name:       "termination_provisioner_time_seconds",
+			Help:       "The time taken between a node's deletion request and its termination, labeled by provisioner",
```
ellistarn, line 49: Thoughts on labeling this metric with all well known labels instead of just provisioner? e.g. How long is it taking large instances to scale down compared to small ones? Does it vary by zone? etc.

jmdeal: The problem is it is not possible to aggregate summary metrics. Because of this, for summaries to be useful, we would have a combinitorial explosion of metric families. @njtran and I decided that provisioner was likely the most useful label to be able to filter on so I included an additional metric family to allow filtering on provisioners. A more detailed explanation is in the updated doc's appendix.

ellistarn (dismissed):

ellistarn (approved):

PR #2146, opened 2022-07-18
feat: add pod startup summary metric

jmdeal:
**Description**
Instruments the `karpenter_pods_startup_time_seconds` metric as detailed in the metrics design doc.

* Scale-up / scale-down tests in EKS cluster.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/metrics/pod/controller.go
+++ b/pkg/controllers/metrics/pod/controller.go
@@ -26,6 +26,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -33,6 +34,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
+	"github.com/aws/karpenter/pkg/metrics"
 )
 
 const (
@@ -46,6 +48,8 @@ const (
 	podHostCapacityType = "capacity_type"
 	podHostInstanceType = "instance_type"
 	podPhase            = "phase"
+
+	phasePending = "Pending"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/controllers/metrics/pod/controller.go
```diff
@@ -107,14 +126,24 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	// Record pods state metric
 	labels := c.labels(ctx, pod)
 	podGaugeVec.With(labels).Set(float64(1))
 	c.labelsMap.Store(client.ObjectKeyFromObject(pod), labels)
+
+	// Record pods startup time metric
+	podKey := client.ObjectKeyFromObject(pod).String()
+	if pod.Status.Phase == phasePending {
+		c.pendingPods.Insert(podKey)
+	} else if c.pendingPods.Has(podKey) {
```
njtran, line 143: can you explicitly make this check if the pending pod phase is Running?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -107,14 +126,24 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		c.pendingPods.Delete(podKey)
+		podStartupTimeSummary.Observe(timestamp.Sub(pod.CreationTimestamp.Time).Seconds())
```
njtran, line 145: Can you switch these two lines, we should delete the pod from the list after we observe it.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -88,13 +105,15 @@ func labelNames() []string {
...
-		kubeClient: kubeClient,
+		kubeClient:  kubeClient,
+		pendingPods: sets.NewString(),
 	}
 }
 
 // Reconcile executes a termination control loop for the resource
 func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("podmetrics").With("pod", req.Name))
+	ts := time.Now()
```
ellistarn, line 116: Watch events are eventually consistent. Can we use the timestamp on a pod's status condition?

jmdeal: Yep, I didn't think there was a recorded timestamp but there is a `LastTransitionTime` field that looks like it should do the trick.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -112,9 +129,28 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+
+	// Record pods startup time metric
+	getReadyCondition := func() *v1.PodCondition {
+		for _, condition := range pod.Status.Conditions {
+			if condition.Type == v1.PodReady {
+				return &condition
+			}
+		}
+		return nil
+	}
```
njtran, line 145: I'd either separate this out to another helper function or just do the for loop like normal instead of creating a function to call it.

njtran (commented):
lgtm: just one style thing before approving.

njtran (approved):

vadimgusev-codefresh:
@jmdeal great job on adding this useful metric. I've got couple feedback items:
- If you scale up karpenter then you will have N sets of summaries, produced by every pod, but only leader will have actual numbers on them.
- Is it possible/does it make sense to add labels like provisioner/capacity-type/node to align with the rest of the metrics?

PR #2148, opened 2022-07-19
docs: fix misleading description on controller max-pods flag

offzale:
**Description**
Fixes a misleading description in the controller flag `aws-eni-limited-pod-density` which enables nodes to have a higher pod density, up to 110 pods.  Updating description to the current behaviour.

https://github.com/aws/karpenter/blob/e37909b223fd7625b57b59d28681561895643f09/pkg/cloudprovider/aws/launchtemplate.go#L364-L368

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/website/content/en/v0.10.0/tasks/pod-density.md
+++ b/website/content/en/v0.10.0/tasks/pod-density.md
@@ -20,13 +20,15 @@ Do not use the `max-pods` argument to kubelet. Karpenter is not aware of this va
 
 *☁️ AWS Specific*
 
-The number of pods on a node is limited by the number of networking interfaces (ENIs) that may be attached to a node.
+By default, the number of pods on a node is limited by both the number of networking interfaces (ENIs) that may be attached to an instance type and the number of IP addresses that can be assigned to each ENI.  See [IP addresses per network interface per instance type](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI) for a more detailed information on these instance types' limits.
 
-[AWS VPC CNI v1.9 introduced prefix assignment.](https://aws.amazon.com/blogs/containers/amazon-vpc-cni-increases-pods-per-node-limits/) In short, a single ENI can provide IP addresses for multiple pods. Much higher pod densities are now supported.
-
-Run the Karpenter controller with the environment variable `AWS_ENI_LIMITED_POD_DENSITY` (or the argument  `--aws-eni-limited-pod-density=true`) to enable nodes with more than 110 pods.
+Karpenter can be configured to disable nodes' ENI-based pod density.  This is specially useful for small to medium instance types which have a lower ENI-based pod density.  Set the environment variable `AWS_ENI_LIMITED_POD_DENSITY: "false"` (or the argument  `--aws-eni-limited-pod-density=false`) in the Karpenter controller to allow nodes to host up to 110 pods.
 
 Environment variables for the Karpenter controller may be specified as [helm chart values](https://github.com/aws/karpenter/blob/c73f425e924bb64c3f898f30ca5035a1d8591183/charts/karpenter/values.yaml#L15).
+
+{{% alert title="Note" color="primary" %}}
+When using small instance types, it may be necessary to enable [prefix assignment mode](https://aws.amazon.com/blogs/containers/amazon-vpc-cni-increases-pods-per-node-limits/) in the AWS VPC CNI plugin to support 110 pods per node.  Prefix assignment mode was introduced in AWS VPC CNI v1.9 and allows ENIs to manage a broader set of IP addresses.  Much higher pod densities are supported as a result.
+{{% /alert %}}
 
 ## Limit Pod Density
 
--- a/website/content/en/preview/tasks/pod-density.md
+++ b/website/content/en/preview/tasks/pod-density.md
@@ -20,12 +20,14 @@ Do not use the `max-pods` argument to kubelet. Karpenter is not aware of this va
 
 *☁️ AWS Specific*
 
-By default, the number of pods on a node is limited by the number of networking interfaces (ENIs) that may be attached to an instance type. By running the Karpenter controller with the environment variable `AWS_ENI_LIMITED_POD_DENSITY=false` (or the argument  `--aws-eni-limited-pod-density=false`) you can set the maximum number of pods per node to 110 instead.
+By default, the number of pods on a node is limited by both the number of networking interfaces (ENIs) that may be attached to an instance type and the number of IP addresses that can be assigned to each ENI.  See [IP addresses per network interface per instance type](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI) for a more detailed information on these instance types' limits.
+
+Karpenter can be configured to disable nodes' ENI-based pod density.  This is specially useful for small to medium instance types which have a lower ENI-based pod density.  Set the environment variable `AWS_ENI_LIMITED_POD_DENSITY: "false"` (or the argument  `--aws-eni-limited-pod-density=false`) in the Karpenter controller to allow nodes to host up to 110 pods.
 
 Environment variables for the Karpenter controller may be specified as [helm chart values](https://github.com/aws/karpenter/blob/c73f425e924bb64c3f898f30ca5035a1d8591183/charts/karpenter/values.yaml#L15).
 
 {{% alert title="Note" color="primary" %}}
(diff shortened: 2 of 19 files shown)


website/content/en/v0.10.0/tasks/pod-density.md
```diff
@@ -24,7 +24,7 @@ The number of pods on a node is limited by the number of networking interfaces (
 
 [AWS VPC CNI v1.9 introduced prefix assignment.](https://aws.amazon.com/blogs/containers/amazon-vpc-cni-increases-pods-per-node-limits/) In short, a single ENI can provide IP addresses for multiple pods. Much higher pod densities are now supported. 
 
-Run the Karpenter controller with the enviornment variable `AWS_ENI_LIMITED_POD_DENSITY` (or the argument  `--aws-eni-limited-pod-density=true`) to enable nodes with more than 110 pods. 
+Run the Karpenter controller with the enviornment variable `AWS_ENI_LIMITED_POD_DENSITY: "false"` (or the argument  `--aws-eni-limited-pod-density=false`) to disable ENI-based pod density and allow nodes to host up to 110 pods.
```
ellistarn, line 27: Nice catch! The polarity of this boolean is backwards (defaults should be false). Let's fix in #1327

ellistarn (changes requested):
Looks good. Make sure you add this to `preview`, which is where docs for new versions get generated from.

ellistarn:
Hey @offzale , sorry for the slow response here!

website/content/en/v0.10.0/tasks/pod-density.md
FernandoMiguel: Does this mean I dont have to do this ?
https://github.com/aws/karpenter/issues/2162

ellistarn: Not sure what you mean. This value currently defaults to true.

offzale:
Thanks for the review @ellistarn, I missed that file indeed.

I noticed there were a few changes in the pod-density's preview docs that were not in the current/previous release docs. Those are for clarifying a bit more the requirements to get this feature working, so I believe it would be handy to have them available in the docs for all versions which have the feature. So, I rewrote a little part and took the changes to all versions' docs.

FernandoMiguel:
as someone who is actively working in scaling the number of pods in their cluster, can some of this be even further clarified?
with `AWS_ENI_LIMITED_POD_DENSITY` set to `false` karpenter is setting a limit of 110 pods on any ec2 size instance?
and if set to `true` what is the limit?
is it hardcoded or is it using a shell script to determine the max number of IPs available and as such limiting the number of pods?

I believe that 110 number is coming from k8s recommendation, right? can that docs be linked here too?

website/content/en/v0.10.0/tasks/pod-density.md
offzale: @FernandoMiguel correct, after https://github.com/aws/karpenter/issues/1327#issuecomment-1195605857 is applied ENI-based pod density will be disabled by default. Therefore, you will not need to set the environment variable to false anymore. However, the parsing issue you're experiencing with that helm module will still be there unless it gets fixed as well

FernandoMiguel: thanks @offzale

offzale:
> with AWS_ENI_LIMITED_POD_DENSITY set to false karpenter is setting a limit of 110 pods on any ec2 size instance?
Correct, you can see this in the code snippet in the description of the PR

> and if set to true what is the limit? is it hardcoded or is it using a shell script to determine the max number of IPs available and as such limiting the number of pods?
That depends on the instance types, each have a specific number of ENIs that can get attached and a maximum number of IP addresses that each ENI can get assigned. You can check the [updated documentation](https://github.com/aws/karpenter/blob/d20f788db8a6331c2d2f3d4cf636a99e858bf72a/website/content/en/preview/tasks/pod-density.md), there are links to both the details on how this ENI/IP limits work ([here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI)), and the default pod density limit AWS gives to each instance type ([here](https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt)).

> I believe that 110 number is coming from k8s recommendation, right? can that docs be linked here too?
https://kubernetes.io/docs/setup/best-practices/cluster-large/
This is mentioned in Karpenter's pod-density docs as well. However, I have not seen any online documentation yet the reasons for this limit in Kubernetes nodes.

FernandoMiguel:
> That depends on the instance types, each have a specific number of ENIs that can get attached and a maximum number of IP addresses that each ENI can get assigned. You can check the [updated documentation](https://github.com/aws/karpenter/blob/d20f788db8a6331c2d2f3d4cf636a99e858bf72a/website/content/en/preview/tasks/pod-density.md), t[here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI) are links to both the details on how this ENI/IP limits work ([here](https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt)), and the default pod density limit AWS gives to each instance type (here).

thanks for the detailed info.
what happens if you set (remove?) the limit to 110 pods, and the instance size is really small (like c7g.medium) which will only have one ENI, and can host 8 EIP (before prefix delegation being enable)?
does it try to allocate 110 pods? 8? fails?

even with prefix delegation, that gives it 13 IPs, if my math is correct for /28, which leave 12 IPs available for Pods.

ellistarn (dismissed):

ellistarn:
Looks like this needs a rebase. Sorry!

offzale:
No problem, happy to see a project with so many active contributors :smile: I just rebased

offzale:
@FernandoMiguel 

> what happens if you set (remove?) the limit to 110 pods, and the instance size is really small (like c7g.medium) which will only have one ENI, and can host 8 EIP (before prefix delegation being enable)?
does it try to allocate 110 pods? 8? fails?

Assuming you only remove this configuration in the Karpenter controller to set `--max-pods 110` on the nodes' kubelet, and not also the configuration in the CNI plugin. The current nodes will not notice the change since they have already been bootstrapped with this kubelet argument. Therefore, they will still be able to host up to 110 pods. However, Karpenter will bootstrap all new nodes, from the moment you change this setting, without the `--max-pods 110` argument in the kubelet. So, new nodes will use ENI-based pod density limits, e.g. 8 pods limit for the c7g.medium instance type that you mentioned as example.

However, if you change both Karpenter controller's setting and the CNI plugin configuration, that could cause some disruption. I have not tested it, and I do not know the AWS ANI plugin that well. But I could imagine either of the following scenarios playing out:
* The CNI plugin will reconfigure the ENIs dropping the /28 CIDRs it got assigned, needing to terminate all pods using IPs from those CIDRs (which may potentially be all non DaemonSet pods in the node) to be able to release the IPs. Therefore, causing a major disruption on the cluster where all pods running on those nodes will need to be re-scheduled, potentially having to wait for a few nodes to join the cluster first. This would be assuming the worst case, where the termination of pods is forced. However, if the pods are terminated gracefully to release the IPs, the re-scheduling would happen gently and without disruption (with HA workloads and PDB in place).
* The CNI plugin does not reconfigure the ENIs, so there is no impact on the current nodes and scheduled pods. However, for all new ENIs that get attached to either current or new nodes the CNI plugin will not configure prefix delegation. So, each of those new ENIs will be able to provision only 4 IPs for pods (taking c7g.medium instance type as example, which can get attached 2 ENIs that handle 4 IPs each).

But again, I have not proven that any of this actually happen when you do such changes, it is only my best guess. So, take it with a grain of salt :)

> even with prefix delegation, that gives it 13 IPs, if my math is correct for /28, which leave 12 IPs available for Pods.

All IPs in the /28 prefixes assigned to ENIs are available to be used by pods. Although ENIs require to get assigned a primary IP address to operate, this primary IP does not take any of the available IPv4/6 addresses per ENI. For those, either a prefix or a single private IP can be assigned. So, all 4 IP/prefix allocations per ENI are available for pods.

In practice, once you enable prefix delegation the CNI plugin always allocate a prefix instead of an IP, if there are /28 prefixes available in the subnet.

So, if you have plenty of room in your subnet, the potential total IPs you could get for pods would be (for c7g.medium):
`2 ENI * 4 prefixes per ENI * 16 IPs per prefix = 128 IPs`

ellistarn:
> The current nodes will not notice the change since they have already been bootstrapped with this kubelet argument. Therefore, they will still be able to host up to 110 pods. However, Karpenter will bootstrap all new nodes, from the moment you change this setting, without the --max-pods 110 argument in the kubelet. So, new nodes will use ENI-based pod density limits, e.g. 8 pods limit for the c7g.medium instance type that you mentioned as example.

We should be able to solve this in the future with node reconciliation when a provisioner change is made: https://github.com/aws/karpenter/issues/1738

offzale:
@ellistarn I am sorry for the often conflicts with files that were not changed in the PR. I have just reset the branch with the latest status from main and added the changes once again.

> We should be able to solve this in the future with node reconciliation when a provisioner change is made: #1738

That would be great! :smile:

ellistarn (approved):

PR #2153, opened 2022-07-19
feat: in the absence of instance-type/family filtering, provide a default

tzneal:
**Description**

Provide a more opinionated set of default instance types if the provisioner
doesn't supply its own list of instance types/families.

Unit testing & deployed to EKS.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/aws/cloudprovider.go
+++ b/pkg/cloudprovider/aws/cloudprovider.go
@@ -29,6 +29,14 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ssm"
 	"github.com/patrickmn/go-cache"
+	"github.com/samber/lo"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/transport"
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/ptr"
+	k8sClient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	awsv1alpha1 "github.com/aws/karpenter/pkg/apis/awsnodetemplate/v1alpha1"
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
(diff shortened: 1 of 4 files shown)


ellistarn (changes requested):
Why not apply defaults using the defaulting webhook and persisting explicitly to the API?

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -140,7 +139,38 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context, provisioner *v1alp
 	if err != nil {
 		return nil, err
 	}
-	return c.instanceTypeProvider.Get(ctx, aws)
+	instanceTypes, err := c.instanceTypeProvider.Get(ctx, aws)
```
ellistarn, line 142: The responsibilities of this top level interface are more or less limited to delegating our requests to other components (e.g. instanceTypeProvider). Does it make more sense to encapsulate this there alongside instance-type specific logic?

tzneal: The instance type provider maintains a cache of instance types, while this does provisioner specific filtering.  Seems best to keep the filtering here at the CP and let instance type provider be just a generic cached fetcher of instance types.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -205,3 +235,39 @@ func (c *CloudProvider) getProvider(ctx context.Context, provider *runtime.RawEx
 	}
 	return aws, nil
 }
+
+func (c *CloudProvider) useOpinionatedInstanceFilter(provisionerRequirements ...v1.NodeSelectorRequirement) bool {
+	var instanceTypeRequirement, familyRequirement v1.NodeSelectorRequirement
```
ellistarn, line 240: optional: Check out the scheduling.Requirements library to potentially remove a bunch of undifferentiated code. 
```
requirements := scheduling.NewNodeSelectorRequirements(provisionerRequirements)
for _, label := range string[]{ v1.LabelInstanceTypeStable, v1alpha1.LabelInstanceFamily } {
  if requirements.Has(label) && requirements.Get(label).Operator != v1.NodeSelectorOpIn {
    return false
  }
}
return true
```

ellistarn (approved):
We can revisit serverside vs implicit defaulting in v1alpha6. Cut an issue here: https://github.com/aws/karpenter/issues/1327

PR #2154, opened 2022-07-19
feat: Adding Custom AMIFamily and validations

suket22:
- feat: Adding Custom AMIFamily and validations

**Description**
This adds a new AMIFamily - `Custom`. When the AMIFamily is `Custom` - 
* We'll use the UserData that's provided to the awsnodetemplate as is. No merging whatsoever. 
* When we binpack, we'll assume that the default EBS volume is of size 20G, and that the `kubeReserved`, `sysReserved` values etc match what the EKS AMI has. This needs to be done better but is outside the scope of this initial PR IMO and has to be solved for all AMI families.

* Unit tests and integ tests

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
feat: Adding Custom AMIFamily and validations
```


pkg/cloudprovider/aws/instancetype.go
```diff
@@ -169,8 +169,13 @@ func (i *InstanceType) memory() *resource.Quantity {
 
 // Setting ephemeral-storage to be either the default value or what is defined in blockDeviceMappings
 func (i *InstanceType) ephemeralStorage() *resource.Quantity {
-	ephemeralBlockDevice := amifamily.GetAMIFamily(i.provider.AMIFamily, &amifamily.Options{}).EphemeralBlockDevice()
 	if i.provider.BlockDeviceMappings != nil {
```
bwagner5, line 172: Checking length here seems a little safer although block device mappings shouldn't be an empty slice.  Line 176 would panic on an out-of-bounds error if it somehow was an empty slice though. 

```suggestion
	if len(i.provider.BlockDeviceMappings) != 0 {
```

bwagner5 (commented):
just one minor comment, otherwise LGTM

pkg/cloudprovider/aws/instancetype.go
suket22: Make sense!

PR #2161, opened 2022-07-21
fix: permit custom device requests

### LINKED ISSUE #1900: Karpenter doesnt support custom resources requests/limit
## Version

**Karpenter**: v0.10.1

**Kubernetes**: v1.20.15

## Expected Behavior
Karpenter should be able to trigger an autoscale

## Actual Behavior
Karpenter isnt able to trigger an autoscale

We're using Karpenter on EKS. We have pods that has custom resource requests/limits in their spec definition - `smarter-devices/fuse: 1`. Karpenter seems to not respecting this resource and fails to autoscale and the pod remains to be in pending state 

## Resource Specs and Logs

Provisioner spec

```
kind: Provisioner
metadata:
  name: default
spec:
  limits:
    resources:
      cpu: "100"
  provider:
    launchTemplate: xxxxx
    subnetSelector:
      xxxxx: xxxxx
  requirements:
    operator: In
    values:
    - on-demand
    operator: In
    values:
    - m5.large
    - m5.2xlarge
    - m5.4xlarge
    - m5.8xlarge
    - m5.12xlarge
  - key: kubernetes.io/arch
    operator: In
    values:
    - amd64
  ttlSecondsAfterEmpty: 30
status:
  resources:
    cpu: "32"
    memory: 128830948Ki
```

pod spec

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: fuse-test
  labels:
    app: fuse-test
spec:
  replicas: 1
  selector:
    matchLabels:
      name: fuse-test
  template:
    metadata:
      labels:
        name: fuse-test
    spec:
      containers:
      - name: fuse-test
        image: ubuntu:latest
        ports:
          - containerPort: 8080
            name: web
            protocol: TCP
        securityContext:
          capabilities:
            add:
              - SYS_ADMIN
        resources:
          limits:
            cpu: 32
            memory: 4Gi
            smarter-devices/fuse: 1  # Custom resource
          requests:
            cpu: 32
            memory: 2Gi
            smarter-devices/fuse: 1  # Custom resource
        env:
        - name: S3_BUCKET
          value: test-s3
        - name: S3_REGION
          value: eu-west-1
```

karpenter controller logs:

controller 2022-06-06T15:59:00.499Z    ERROR    controller    no instance type satisfied resources {"cpu":"32","memory":"2Gi","pods":"1","smarter-devices/fuse":"1"} and requirements kubernetes.io/os In [linux], karpenter.sh/capacity-type In [on-demand], kubernetes.io/hostname In [hostname-placeholder-3403], node.kubernetes.io/instance-type In [m5.12xlarge m5.2xlarge m5.4xlarge m5.8xlarge m5.large], karpenter.sh/provisioner-name In [default], topology.kubernetes.io/zone In [eu-west-1a eu-west-1b], kubernetes.io/arch In [amd64];

universam1:
Fixes #1900

**Description**

Karpenter is negative towards custom device requests it is unaware of, erroneously assuming those cannot be scheduled. Instead, ignore unmanaged resources from Karpenter perspective in best effort.

This changes the request handling to be scoped only to resource request that Karpenter is actively managing. 
The reasoning here is that it cannot influence those resource requests anyways, they come into existence by means of other concepts such as the device plugin manager that might be late bound and thus out of Karpenter scope.
Hence, instead of being negative, it should be neutral regarding unknown resources.

It is related with the recent concept change _not to assign pods_ to a node but let control plane perform the final scheduling which takes into account runtime available resources.

* integration tests in AWS lab account

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
fix: supporting custom device requests
```

### DIFF
--- a/pkg/controllers/provisioning/scheduling/suite_test.go
+++ b/pkg/controllers/provisioning/scheduling/suite_test.go
@@ -2912,8 +2912,8 @@ var _ = Describe("Instance Type Compatibility", func() {
 	})
 	It("should launch pods with resources that aren't on any single instance type on different instances", func() {
 		cloudProv.InstanceTypes = fake.InstanceTypes(5)
-		const fakeGPU1 = "karpenter.sh/super-great-gpu"
-		const fakeGPU2 = "karpenter.sh/even-better-gpu"
+		fakeGPU1 := v1alpha1.ResourceNVIDIAGPU
+		fakeGPU2 := v1alpha1.ResourceAMDGPU
 		cloudProv.InstanceTypes[0].Resources()[fakeGPU1] = resource.MustParse("25")
 		cloudProv.InstanceTypes[1].Resources()[fakeGPU2] = resource.MustParse("25")
 
@@ -2935,24 +2935,6 @@ var _ = Describe("Instance Type Compatibility", func() {
 			nodeNames.Insert(node.Name)
 		}
 		Expect(nodeNames.Len()).To(Equal(2))
-	})
-	It("should fail to schedule a pod with resources requests that aren't on a single instance type", func() {
-		cloudProv.InstanceTypes = fake.InstanceTypes(5)
-		const fakeGPU1 = "karpenter.sh/super-great-gpu"
-		const fakeGPU2 = "karpenter.sh/even-better-gpu"
-		cloudProv.InstanceTypes[0].Resources()[fakeGPU1] = resource.MustParse("25")
-		cloudProv.InstanceTypes[1].Resources()[fakeGPU2] = resource.MustParse("25")
-
-		ExpectApplied(ctx, env.Client, provisioner)
-		pods := ExpectProvisioned(ctx, env.Client, controller,
-			test.UnschedulablePod(test.PodOptions{
-				ResourceRequirements: v1.ResourceRequirements{
-					Limits: v1.ResourceList{
-						fakeGPU1: resource.MustParse("1"),
-						fakeGPU2: resource.MustParse("1")},
-				},
-			}))
-		ExpectNotScheduled(ctx, env.Client, pods[0])
 	})
 	Context("Provider Specific Labels", func() {
 		It("should filter instance types that match labels", func() {
@@ -3288,11 +3270,22 @@ var _ = Describe("Binpacking", func() {
 		pod := ExpectProvisioned(ctx, env.Client, controller,
 			test.UnschedulablePod(test.PodOptions{
 				ResourceRequirements: v1.ResourceRequirements{
-					Requests: v1.ResourceList{"foo.com/weird-resources": resource.MustParse("0")},
-					Limits:   v1.ResourceList{"foo.com/weird-resources": resource.MustParse("0")},
+					Requests: v1.ResourceList{v1alpha1.ResourceAWSNeuron: resource.MustParse("0")},
+					Limits:   v1.ResourceList{v1alpha1.ResourceAWSNeuron: resource.MustParse("0")},
 				},
 			}))
 		// requesting a resource of quantity zero of a type unsupported by any instance is fine
+		ExpectScheduled(ctx, env.Client, pod[0])
+	})
+	It("should be neutral regarding custom, unknown resource requests", func() {
+		ExpectApplied(ctx, env.Client, provisioner)
+		pod := ExpectProvisioned(ctx, env.Client, controller,
+			test.UnschedulablePod(test.PodOptions{
+				ResourceRequirements: v1.ResourceRequirements{
+					Requests: v1.ResourceList{"smarter-device/fuse": resource.MustParse("1")},
+					Limits:   v1.ResourceList{"smarter-device/fuse": resource.MustParse("1")},
+				},
+			}))
 		ExpectScheduled(ctx, env.Client, pod[0])
 	})
 	It("should not schedule pods that exceed every instance type's capacity", func() {
--- a/pkg/controllers/provisioning/scheduling/inflightnode.go
+++ b/pkg/controllers/provisioning/scheduling/inflightnode.go
@@ -108,7 +108,8 @@ func (n *InFlightNode) Add(ctx context.Context, pod *v1.Pod) error {
 
 	// check resource requests first since that's a pretty likely reason the pod won't schedule on an in-flight
 	// node, which at this point can't be increased in size
-	requests := resources.Merge(n.requests, resources.RequestsForPods(pod))
+	podRequests := state.FilterWellKnownRequests(ctx, resources.RequestsForPods(pod))
+	requests := resources.Merge(n.requests, podRequests)
 
 	if !resources.Fits(requests, n.available) {
 		return fmt.Errorf("exceeds node resources")
--- a/pkg/controllers/provisioning/scheduling/node.go
+++ b/pkg/controllers/provisioning/scheduling/node.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
+	"github.com/aws/karpenter/pkg/controllers/state"
 	"github.com/aws/karpenter/pkg/scheduling"
 	"github.com/aws/karpenter/pkg/utils/resources"
 	"github.com/aws/karpenter/pkg/utils/sets"
@@ -90,10 +91,11 @@ func (n *Node) Add(ctx context.Context, pod *v1.Pod) error {
 	nodeRequirements.Add(topologyRequirements)
 
 	// Check instance type combinations
-	requests := resources.Merge(n.requests, resources.RequestsForPods(pod))
+	podRequests := state.FilterWellKnownRequests(ctx, resources.RequestsForPods(pod))
+	requests := resources.Merge(n.requests, podRequests)
 	instanceTypes := filterInstanceTypes(n.InstanceTypeOptions, nodeRequirements, requests)
 	if len(instanceTypes) == 0 {
-		return fmt.Errorf("no instance type satisfied resources %s and requirements %s", resources.String(resources.RequestsForPods(pod)), nodeRequirements)
+		return fmt.Errorf("no instance type satisfied resources %s and requirements %s", resources.String(podRequests), nodeRequirements)
 	}
 
 	// Update node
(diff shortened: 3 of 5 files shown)


universam1:
@ellistarn created a suggested fix as requested - tested it in lab which works and solves our use case with custom devices. Of course this is a naive implementation being new to the project.
Thanks for feedback

ellistarn:
Thanks for the PR @universam1. The main concern I'd have with this feature is that it's possible to get into scheduling loops if the node fails to populate unknown custom resources. We need to make sure that both the scheduling.Node and scheduling.InFlightNode agree about the schedulability of the pod. 

If for some reason the custom resources don't get advertised on the node, we either need to terminate the node (via initialization failure https://github.com/aws/karpenter/blob/main/pkg/controllers/node/initialization.go#L72) or leave the node around (and the pods would stay stuck pending).

I'm curious what @tzneal thinks about this.

As mentioned on the issue, I'd really like to explore the design space of doing this properly (via a configmap or CRD). Perhaps @billrayburn can help us with the roadmap on this.

universam1:
Thank you @ellistarn for reviewing and the provided context, appreciated, also for the hint with the scheduling.InFlightNode. This is improved and I added a unit test. Confirmed that it works in our lab clusters.

> it's possible to get into scheduling loops if the node fails to populate unknown custom resources. 

Valid point, however I cannot see a difference to a whitelisted (CRD) concept for this issue. If a request is whitelisted and the resource is not populated, the same condition of a potential loop would arise, no? Isn't this the case already for GPU requests where this can happen too?

>  doing this properly (via a configmap or CRD)

If I may reason that properly™ might look from the operators perspective differently:

- having an extra whitelist is apparently overhead. I as operator must form then a contract with the devs about a provisioner they are to use, which requires now another place to keep in sync. The deployment needs to request a specific provisioner to work with Karpenter where it just works OOB with ClusterAutoscaler. This PR demonstrates this overhead is not necessary
- _ClusterAutoscaler_ does this transparently, gracefully ignoring unknown requests, without a need for a specific whitelist. So it falls behind
- whitelist requires CRD change which might be harder to get going than without?

Eventually this missing support for custom devices is a blocking issue for us and we are happy to help getting it solved, thanks!

ellistarn:
> The deployment needs to request a specific provisioner to work with Karpenter where it just works OOB with ClusterAutoscaler. This PR demonstrates this overhead is not necessary

IIUC, this is because cluster autoscaler works by discovering extended resources on node groups from preexisting nodes. If you want to scale from zero (which karpenter does), you have to annotate this using ASG tags.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3288,13 +3270,24 @@ var _ = Describe("Binpacking", func() {
...
-					Requests: v1.ResourceList{"foo.com/weird-resources": resource.MustParse("0")},
-					Limits:   v1.ResourceList{"foo.com/weird-resources": resource.MustParse("0")},
+					Requests: v1.ResourceList{v1alpha1.ResourceAWSNeuron: resource.MustParse("0")},
+					Limits:   v1.ResourceList{v1alpha1.ResourceAWSNeuron: resource.MustParse("0")},
 				},
 			}))
 		// requesting a resource of quantity zero of a type unsupported by any instance is fine
 		ExpectScheduled(ctx, env.Client, pod[0])
 	})
+	It("should be neutral regarding custom, unknown resource requests", func() {
```
ellistarn, line 3280:
```suggestion
	It("should ignore unknown resource requests", func() {
```

ellistarn:
I *think* this may be able to be achieved by modifying the `Fits` function
```
// Fits returns true if the candidate set of resources is less than or equal to the total set of resources.
func Fits(candidate, resources v1.ResourceList) bool {
	for resourceName, quantity := range candidate {
                // -----------------------------------------> If unknown, don't even check and fall through to true below
		if available, ok := resources[resourceName]; ok && Cmp(quantity, available) > 0 {
			return false
		}
	}
	return true
}
```

ellistarn:
As mentioned above, the other drawback I can think of is that we can't track node initialization of extended resources, so we will be unable to terminate automatically fix a node that comes online and cannot register its extended resource due to a hardware failure or similar. Additionally, you'll need to ensure that your `ttlSecondsAfterEmpty` is set high enough such that the node isn't terminated, since we'll start counting the node as initialized before it is.

ellistarn:
The other drawback I just realized is that if we consider existing or in flight capacity to have infinite of a custom resource, karpenter will fail to provision beyond the first node (unless other resources are full on the node). If we assume infinite for new nodes and zero for existing, then karpenter will launch nodes as fast as it can while the pod remains pending.


I don't see a maintainable path forward with the assumption as proposed, as there are enough common cases that break. 

I'd like to pursue an instance type override approach, but this has some caveats with multi-os clusters. It may be worth bringing this idea to working group to discuss further.

universam1:
> I don't see a maintainable path forward with the assumption as proposed, as there are enough common cases that break.

I understand that it might not be the perfect solution, but it is certainly unblocking the showstopper for now. In our case we are running this fork in production clusters, otherwise we'd have to rollback from Karpenter.
The issues you mentioned might appear but do not happen at least for us. One could see this as an iterative solution as it is not involving any API changes and thus there is _any_ path forward possible...

> I'd like to pursue an instance type override approach, 

What time frame are you considering, since this is a blocking issue for certain users?

ellistarn:
> I understand that it might not be the perfect solution, but it is certainly unblocking the showstopper for now. In our case we are running this fork in production clusters, otherwise we'd have to rollback from Karpenter.

Can you share some setup instructions and I can try to reproduce?

universam1:
> > I understand that it might not be the perfect solution, but it is certainly unblocking the showstopper for now. In our case we are running this fork in production clusters, otherwise we'd have to rollback from Karpenter.
> 
> Can you share some setup instructions and I can try to reproduce?

Eks cluster, containerD node runtime. Run smarter-device daemonset setup to provide fuse device. Run podman pod unprivileged with volume device request of /dev/fuse. Build a container inside podman pod.

ellistarn:
> Need a spec file?

Any yaml / links you can share would help me get to this faster :D.

universam1:
> Any yaml / links you can share would help me get to this faster :D.

Sure @ellistarn  this would be the test setup:

- EKS
- containerD node runtime
- deploy this smarter-device-manager and the test pod
<details>
<summary>spec file</summary>
  
```yaml

apiVersion: v1
kind: ConfigMap
metadata:
  name: smarter-device-manager
  namespace: kube-system
data:
  conf.yaml: |
    - devicematch: ^fuse$
      nummaxdevices: 1000
---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: smarter-device-manager
  namespace: kube-system
  labels:
    name: smarter-device-manager
spec:
  selector:
    matchLabels:
      name: smarter-device-manager
  template:
    metadata:
      labels:
        name: smarter-device-manager
    spec:
      hostname: smarter-device-management
      hostNetwork: true
      containers:
      - name: smarter-device-manager
        image: registry.gitlab.com/arm-research/smarter/smarter-device-manager:v1.20.10
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop: ["ALL"]
        volumeMounts:
          - name: device-plugin
            mountPath: /var/lib/kubelet/device-plugins
          - name: dev-dir
            mountPath: /dev
          - name: sys-dir
            mountPath: /sys
          - name: config
            mountPath: /root/config
      volumes:
        - name: device-plugin
          hostPath:
            path: /var/lib/kubelet/device-plugins
        - name: dev-dir
          hostPath:
            path: /dev
        - name: sys-dir
          hostPath:
            path: /sys
        - name: config
          configMap:
            name: smarter-device-manager
---
apiVersion: v1
kind: Pod
metadata:
  name: testpod
  namespace: default
spec:
  affinity:
    nodeAffinity:
        nodeSelectorTerms:
        - matchExpressions:
            operator: Exists
  containers:
    - name: testpod
      image: quay.io/podman/stable:latest
      command: ["/bin/sh"]
      args:
        - -c
        - >-
            id &&
            echo 'FROM alpine' >/tmp/Dockerfile &&
            echo 'RUN id' >>/tmp/Dockerfile &&
            echo 'RUN apk add -U curl' >>/tmp/Dockerfile &&
            echo 'RUN curl --version' >>/tmp/Dockerfile &&
            podman build -f /tmp/Dockerfile .
      resources:
        limits:
          smarter-devices/fuse: 1
      securityContext:
        runAsUser: 1000
        runAsGroup: 1000
        privileged: false
```
</details>

ellistarn:
Just as a heads up, I'm about to go on vacation. Trying to find an owner for this.

jonathan-innis:
Hi @universam1 , we are currently working on a long-term solution to this by allowing users to be able to specify these kind of custom device requests at the instance type level as @ellistarn has mentioned. We will have a design doc out shortly around how we are thinking about achieving this and will share it with the community. Will update this thread when we have the doc out, would love to receive your feedback on the PR

universam1:
> Will update this thread when we have the doc out, would love to receive your feedback on the PR

Thank you @jonathan-innis for the heads up and the feedback about the road map. Looking forward to your solution, happy to test it out!

ellistarn:
Closing in favor of #2390

cvpatel:
As the long-term solution may take some time, as discussed in #2390, any chance this or some variation of it can be revived as an intermediate solution? 

@ellistarn @universam1

universam1:
> As the long-term solution may take some time, as discussed in #2390, any chance this or some variation of it can be revived as an intermediate solution?
 
@ellistarn @jonathan-innis Trying to get a feeling for the time to expect for your long-term solution. We are wondering if we should maintain a fork with an intermediate solution for the time being or if there are other options possible?
As an idea, could we guard this naive implementation behind an alpha flag to unblock those users willing to take the risk for the time being? Would that be a feasable approach?

jonathan-innis:
> unblock those users willing to take the risk for the time being

I think my biggest concern here is with the deprovisioning algorithm when assuming unlimited resources. Consolidation would assume that it could consolidate more pods down into a single node than it might actually be able to, at which point it would effectively hang since not all pods that we anticipated we could schedule to the new node would actually be able to schedule. @universam1 I'm curious what you have seen in your environment when running your forked version with consolidation. Have you seen consolidation just sit there after making a bad decision based on the infinite resource assumption or at least overschedule nodes for the number of pods on the cluster?

Another potential here remedy is assuming the extended resource has some fixed value by default (like 1) but again this doesn't work as a one-size-fits-all solution for users as there are definitely other extended resources and workloads that will need to request more than one of a resource and; therefore, never schedule.

There's of course still the more complicated "specify all extended resource type mappings to instance types" solution but we're a little wary of introducing a more complex solution that people will most likely take dependency on with something that we know probably doesn't work as the long-term robust solution

PR #2175, opened 2022-07-22
fix: add in a grace period when checking if pods are stuck terminating

ellistarn:
Fixes # 

**Description**
Currently Karpenter waits until pods are outside of their graceful termination period before terminating the instance. However, the kubelet doesn't start forcibly killing pods until this exact point in time, so Karpenter doesn't give the node the opportunity to force kill the pods. This causes the pods to become unowned and be cleaned up by the pod lifecycle controller's garbage collection mechanism.

I ran into this issue when testing large scale downs (50 pods) that would exceed the default 30 second grace period, and then the subsequent test would notice existing pods.

Long term, we probably should trigger node termination in EC2 without forcibly deleting the node object and build a mechanism for the instance to deregister from the API Server, but this change is outside of the scope of this fix.

* `make e2etest`

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
None
```


bwagner5 (approved):
lgtm

PR #2184, opened 2022-07-25
Test: Fixed a few bugs with running integration tests in the integration environment

ellistarn:
Fixes # 

**Description**

* Uses the correct OCI helm chart
* Wires up kubeconfig correctly
* Speeds up testing by vendoring go modules

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
None
```

### DIFF
--- a/test/pkg/environment/environment.go
+++ b/test/pkg/environment/environment.go
@@ -17,6 +17,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	loggingtesting "knative.dev/pkg/logging/testing"
+	"knative.dev/pkg/ptr"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -25,7 +26,7 @@ import (
 	"github.com/aws/karpenter/pkg/utils/project"
 )
 
-var clusterNameFlag = *flag.String("cluster-name", env.WithDefaultString("CLUSTER_NAME", ""), "Cluster name enables discovery of the testing environment")
+var clusterNameFlag = flag.String("cluster-name", env.WithDefaultString("CLUSTER_NAME", ""), "Cluster name enables discovery of the testing environment")
 
 type Environment struct {
 	context.Context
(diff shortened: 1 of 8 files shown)


test/pkg/environment/environment.go
```diff
@@ -25,7 +26,7 @@ import (
 	"github.com/aws/karpenter/pkg/utils/project"
 )
 
-var clusterNameFlag = *flag.String("cluster-name", env.WithDefaultString("CLUSTER_NAME", ""), "Cluster name enables discovery of the testing environment")
+var clusterNameFlag = flag.String("cluster-name", env.WithDefaultString("CLUSTER_NAME", ""), "Cluster name enables discovery of the testing environment")
```
bwagner5, line 29: Why is this using flag? It looks like this will only ever get the env variable since you never call flag.Parse. Might as well just use:

```suggestion
var clusterName = env.WithDefaultString("CLUSTER_NAME", "")
```
or directly in `DiscoverClusterName()`

test/tools/Dockerfile
```diff
@@ -46,3 +46,8 @@ RUN aws --version
 # Install eksctl
 RUN curl --silent --location "https://github.com/weaveworks/eksctl/releases/download/v0.105.0/eksctl_$(uname -s)_amd64.tar.gz" | tar xz -C /tmp
 RUN mv /tmp/eksctl /usr/local/bin
+
+# Install toolchain
```
bwagner5, line 50: seems like you're just caching the modules not installing the "toolchain"?

test/tools/Dockerfile
```diff
@@ -46,3 +46,8 @@ RUN aws --version
...
+RUN git clone https://github.com/aws/karpenter.git /karpenter
+WORKDIR /karpenter
+RUN GOPRIVATE=* go mod tidy
```
bwagner5, line 53: You shouldn't set `GOPRIVATE=*` since this causes go to skip checksum verifications... Do you just want the proxy off or is there more you're trying to do here?

test/tools/Dockerfile
ellistarn: Good point. I was originally going to install the toolchain, but it took foreverrrrr

test/pkg/environment/environment.go
ellistarn: The test library calls flags.Parse() for us =/. I can't call it again without creating problem. I like your solution though.

bwagner5 (approved):
lgtm

PR #2199, opened 2022-07-27
chore: update default value of `dnsPolicy` in `values.yaml` to `Default`

### LINKED ISSUE #2186: Use dnsPolicy=Default by default in helm chart

I think the karpenter helm chart should use `dnsPolicy=Default` (not `ClusterFirst`) by default.

Karpenter has a boostrapping challenge.  You need to run karpenter somewhere in order to have karpenter manage all the rest of your capacity.  One solution is to run karpenter on fargate, as a special case.

If you _only_ run karpenter on fargate, then at initial cluster bootstrap, karpenter is _the only pod running_.  In particular, coreDNS is still Pending.  At this moment, karpenter fails because it tries to do simple DNS lookups to find the AWS and k8s API endpoints - which go to the _cluster_ DNS clusterIP, which fails (times out with no response) because coreDNS is not yet functional.

However! Karpenter doesn't actually need to talk to other pods or internal ClusterIPs, so doesn't actually need to use the cluster-internal DNS resolver.  Setting `pod.spec.dnsPolicy=Default` (note this is not the k8s default, despite the name) on karpenter pods tells kubelet to configure the karpenter pod with the _host's_ (route53/droplet) resolver, and allows karpenter to lookup these non-cluster hostnames even before coreDNS is functional.  Karpenter is then able to create nodes for coreDNS as usual and everyone is happy.

Manually set `dnsPolicy=Default` on every karpenter helm chart install.   I think we can help others get past this hurdle by making this the default, even for non-fargate.

**Additional context**

Note https://karpenter.sh/v0.13.2/getting-started/getting-started-with-eksctl/#create-a-cluster "Example 2" gives an example of running karpenter on fargate.  This example adds `kube-system` to the fargate profile, presumably so coreDNS (and everything else in kube-system!) runs on fargate.  This is unnecessary with the above `dnsPolicy` change, and this documentation should be updated.

### Community Note

jonathan-innis:
Fixes #2186 

**Description**

Updates the default `dnsPolicy` for the `karpenter-controller` pod spec to be `Default` as opposed to `ClusterFirst`

* `make test`
* Manual deployment validation with `make apply`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ coverage.html
 
 # Common in OSs and IDEs
 .idea
+.vscode
 .DS_Store
 
 # Go Benchmarking
--- a/Makefile
+++ b/Makefile
@@ -7,17 +7,19 @@ CLOUD_PROVIDER ?= aws
 GOFLAGS ?= -tags=$(CLOUD_PROVIDER) $(LDFLAGS)
 WITH_GOFLAGS = GOFLAGS="$(GOFLAGS)"
 
+## Values for applying Karpenter locally
+AWS_DEFAULT_REGION ?= us-west-2
+KO_DOCKER_REPO ?= ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com/karpenter
+
 ## Extra helm options
 CLUSTER_NAME ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].name}' | rev | cut -d"/" -f1 | rev | cut -d"." -f1)
 CLUSTER_ENDPOINT ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].cluster.server}')
 AWS_ACCOUNT_ID ?= $(shell aws sts get-caller-identity --query Account --output text)
-AWS_DEFAULT_REGION ?= us-west-2
 KARPENTER_IAM_ROLE_ARN ?= arn:aws:iam::${AWS_ACCOUNT_ID}:role/${CLUSTER_NAME}-karpenter
 HELM_OPTS ?= --set serviceAccount.annotations.eks\\.amazonaws\\.com/role-arn=${KARPENTER_IAM_ROLE_ARN} \
       		--set clusterName=${CLUSTER_NAME} \
 			--set clusterEndpoint=${CLUSTER_ENDPOINT} \
 			--set aws.defaultInstanceProfile=KarpenterNodeInstanceProfile-${CLUSTER_NAME}
-KO_DOCKER_REPO ?= ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com/karpenter
 TEST_FILTER ?= .*
 help: ## Display help
 	@awk 'BEGIN {FS = ":.*##"; printf "Usage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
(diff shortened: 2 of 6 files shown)


charts/karpenter/values.yaml
```diff
@@ -52,7 +52,7 @@ terminationGracePeriodSeconds:
 # This is required when using a custom CNI.
 hostNetwork: false
 # -- Configure the DNS Policy for the pod
-dnsPolicy: ClusterFirst
+dnsPolicy: Default
```
FernandoMiguel, line 55: Wouldn't this be preferred to be overwritten from the fargate examples `values.yaml` or a `set` rather than change it for everyone?

jonathan-innis: Karpenter doesn't have a need to reach out to coreDNS since it's not interacting with other services and there are no services that are interacting with Karpenter directly. With this in mind, I believe this is fine to take as the default

bwagner5: I think this is generally applicable to all Karpenter installs since Karpenter doesn't rely on cluster internal dns lookups from core-dns.

hack/toolchain.sh
```diff
@@ -17,7 +17,7 @@ tools() {
     go install github.com/norwoodj/helm-docs/cmd/helm-docs@v1.8.1
     go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20220421205612-c162794a9b12
     go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0
-    go install github.com/sigstore/cosign/cmd/cosign@v1.9.0
+    go install github.com/sigstore/cosign/cmd/cosign@v1.10.0
```
bwagner5, line 20: did you test this by making a snapshot release?

Makefile
```diff
@@ -3,9 +3,14 @@ export KUBEBUILDER_ASSETS ?= ${HOME}/.kubebuilder/bin
 
 ## Inject the app version into project.Version
 LDFLAGS ?= -ldflags=-X=github.com/aws/karpenter/pkg/utils/project.Version=$(shell git describe --tags --always)
+CLOUD_PROVIDER ?= aws
 GOFLAGS ?= -tags=$(CLOUD_PROVIDER) $(LDFLAGS)
 WITH_GOFLAGS = GOFLAGS="$(GOFLAGS)"
 
+## Values for applying Karpenter locally
+AWS_DEFAULT_REGION ?= us-west-2
```
ellistarn, line 11: You should maintain these values in your own personal environment, since many developers on the project don't use this region

jonathan-innis: Is there a better region to pick as the default? `us-east-1`. Seems useful to me that this would be set for users already when they are configuring development.

.gitignore
```diff
@@ -6,6 +6,7 @@ coverage.html
 
 # Common in OSs and IDEs
 .idea
+.vscode
```
ellistarn, line 9: Happy to include this in karpenter, but I also recommend using a .gitignore.global in your ~/.gitconfig 
```
	excludesfile = /Users/etarn/.gitignore.global
```
This avoids needing to add all of these env specific files to every project's gitignore.

ellistarn (approved):
Nice work!

ellistarn (approved):

hack/toolchain.sh
jonathan-innis: Validated making a snapshot release with personal public ECR

snorlaX-sleeps:
Just a note incase this causes an issue for other users:
When using a network stack in Kubernetes e.g Istio, Karpenter will fail to start as dnsPolicy: default is connecting to the VPC based DNS servers which don't have any kubernetes services listed.
The istio-proxy needs to connect to the istiod.istio-system.svc which is a kubernetes service and unavailable with the current setting.
The fix is to revert to dnsPolicy: ClusterFirst in the Helm values file

johngmyers:
Another situation this breaks: running Karpenter on an ipv6-only cluster on a dual-stack node. Using the node's DNS servers, without DNS64, causes Karpenter to attempt connecting to STS using IPv4. Reverting to ClusterFirst causes it to use coredns, with DNS64, so the connections to STS go through NAT64.

PR #2204, opened 2022-07-27
fix: login to Public ECR Should Occur at `public.ecr.aws`

jonathan-innis:
Fixes # 

**Description**

`make snapshot` effectively expects two values:
1. The Registry path to the public repo where the images/chart will be published
2. The Public login endpoint for `docker login`

These values are currently under a single value `$RELEASE_REPO` which breaks some things when running `make snapshot`  

* Run `make snapshot` against a custom $RELEASE_REPO

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/hack/release_common.sh
+++ b/hack/release_common.sh
@@ -24,7 +24,7 @@ requireCloudProvider(){
 }
 
 authenticate() {
-  aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin ${RELEASE_REPO}
+  aws ecr-public get-login-password --region us-east-1 | ko login --username AWS --password-stdin public.ecr.aws
 }
 
 buildImages() {


hack/release_common.sh
```diff
@@ -24,7 +24,7 @@ requireCloudProvider(){
 }
 
 authenticate() {
-  aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin ${RELEASE_REPO}
+  aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws
```
ellistarn, line 27: This is a great simplification if it works!

ellistarn: Also -- you can rely on ko instead of docker to reduce our dependency closure
```suggestion
  aws ecr-public get-login-password --region us-east-1 | ko login --username AWS --password-stdin public.ecr.aws
```

jonathan-innis: Ah, didn't realize you could do this, neat suggestion 🎉

ellistarn (approved):

bwagner5 (approved):
lgtm

PR #2205, opened 2022-07-28
fix: Drain terminated Pods without owner reference

ese:
**Description**

Since this https://github.com/aws/karpenter/pull/2092 Pods without owner reference are not drained but pods in terminated state should be safe to be evicted.

```
2022-07-28T11:21:43.167Z	DEBUG	controller.events	Warning	{"commit": "362ea52", "object": {"kind":"Node","name":"ip-172-21-98-19.eu-west-1.compute.internal","uid":"465876f3-6732-4e9e-a89c-a4fa3c1e6bae","apiVersion":"v1","resourceVersion":"1528597833"}, "reason": "FailedDraining", "message": "Failed to drain node, pod test/pod-name does not have any owner references"}
```
```
kubectl -n test get pods
NAME       READY   STATUS      RESTARTS   AGE
pod-name   0/1     Completed   0          16m
```
When a no long-running process Pod has no ownerReference but is in a Terminated state (Failed or Completed) Its expected that the lifecycle is ended so we can drain it to remove the node. In another case, we end with empty nodes that are not running any workload stuck to be removed 

* Deployed in a test cluster, created a test pod 

- [X] No

### DIFF
--- a/pkg/controllers/termination/terminate.go
+++ b/pkg/controllers/termination/terminate.go
@@ -77,7 +77,7 @@ func (t *Terminator) drain(ctx context.Context, node *v1.Node) (bool, error) {
 	// Skip node due to pods that are not able to be evicted
 	for _, p := range pods {
 		// if a pod doesn't have owner references then we can't expect a controller to manage its lifecycle
-		if len(p.ObjectMeta.OwnerReferences) == 0 {
+		if len(p.ObjectMeta.OwnerReferences) == 0 && !IsTerminatedStatus(p) {
 			return false, NodeDrainErr(fmt.Errorf("pod %s/%s does not have any owner references", p.Namespace, p.Name))
 		} else if val := p.Annotations[v1alpha5.DoNotEvictPodAnnotationKey]; val == "true" {
 			return false, NodeDrainErr(fmt.Errorf("pod %s/%s has do-not-evict annotation", p.Namespace, p.Name))
@@ -160,4 +160,11 @@ func IsStuckTerminating(pod *v1.Pod) bool {
 		return false
 	}
 	return injectabletime.Now().After(pod.DeletionTimestamp.Time.Add(1 * time.Minute))
+}
+
+func IsTerminatedStatus(pod *v1.Pod) bool {
+	if pod.Status.Phase == v1.PodFailed || pod.Status.Phase == v1.PodSucceeded {
+		return true
+	}
+	return false
 }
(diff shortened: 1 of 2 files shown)


pkg/controllers/termination/terminate.go
```diff
@@ -77,7 +77,7 @@ func (t *Terminator) drain(ctx context.Context, node *v1.Node) (bool, error) {
 	// Skip node due to pods that are not able to be evicted
 	for _, p := range pods {
 		// if a pod doesn't have owner references then we can't expect a controller to manage its lifecycle
-		if len(p.ObjectMeta.OwnerReferences) == 0 {
+		if len(p.ObjectMeta.OwnerReferences) == 0 && !IsTerminatedStatus(p) {
```
bwagner5, line 80: Thanks for catching this! We can use an already defined pod util here:

```
		if len(p.ObjectMeta.OwnerReferences) == 0 && !pod.IsTerminal(p) {
```

pkg/controllers/termination/terminate.go
```diff
@@ -161,3 +161,10 @@ func IsStuckTerminating(pod *v1.Pod) bool {
 	}
 	return injectabletime.Now().After(pod.DeletionTimestamp.Time.Add(1 * time.Minute))
 }
+
+func IsTerminatedStatus(pod *v1.Pod) bool {
+	if pod.Status.Phase == v1.PodFailed || pod.Status.Phase == v1.PodSucceeded {
+		return true
+	}
+	return false
+}
```
bwagner5, line 170:
```suggestion
```

bwagner5 (commented):
Thanks for catching this case! We need to add this to our test cases to catch this in the future.

ellistarn (approved):
Nice!!

ellistarn:
Not sure why these tests are failing. Pulling it down to run myself

ellistarn:
It looks like there's a race condition that doesn't occur with other tests since these pods are terminal. Normally, pods have a 30 second grace period, so the test has enough time to check their terminating status before we manually delete them (simulating kubelet behavior). However, I'm seeing these pods deleted immediately (causing the race), so there must be a special check in the API Server that's doing this.

This got me thinking. Why do we even need to evict terminal pods? These pods aren't counted towards disruption budgets, and there's nothing to evict if the pod has already stopped. I cut a PR to your branch -- curious for you to try it out and see if it accomplishes the same goal.

https://github.com/ese/karpenter/pull/1

bwagner5:
https://github.com/aws/karpenter/pull/2255

PR #2210, opened 2022-07-28
fix: Emit evictions as events

ellistarn:
Fixes # 

**Description**
Following up from a previous PR: Emit evictions as events instead of logging. Also renamed NominatePod to Nominate to be more in line with existing pod events (Pulled, Created, Started, etc), since they're already scoped to an object type

`16m         Normal    Evict                  pod/ogreveil-4-cjtyd9gst2-6d998fb8b6-z7xd8           Evicted pod`

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
None
```


tzneal (approved):
lgtm

PR #2211, opened 2022-07-28
Revert "fix: login to Public ECR Should Occur at `public.ecr.aws`"

spring1843:
Reverts aws/karpenter#2204

This breaks the nightlies https://github.com/aws/karpenter/actions/runs/2753842941 let's revert and do it in a way that doesn't break the nightly builds if we need to.


jonathan-innis (approved):
🚀

PR #2220, opened 2022-07-29
Release v0.14.0-rc.0

suket22:
Fixes # 

**Description**
Adds v0.14.0-rc.0

* I ran `make website` to make sure everything looks okay.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Release v0.14.0-rc.0
```


website/config.yaml
```diff
@@ -66,13 +66,13 @@ params:
         url: 'https://slack.k8s.io/'
         icon: fab fa-slack
         desc: 'Chat with us on Slack in the #aws-provider channel'
-  latest_release_version: v0.13.2
+  latest_release_version: v0.14.0-rc.0
   versions:
+    - "v0.14.0-rc.0"
```
suket22, line 71: We could potentially leave `latest_release_version` as `v0.13.2` and keep all redirects to that version too and only add `v0.14.0-rc.0` to the dropdown.

spring1843 (approved):

website/static/_redirects
```diff
@@ -2,4 +2,4 @@
 /docs/tasks/running-pods/       /docs/tasks/scheduling/
 /docs/provisioner-crd/	        /docs/provisioner/
 /docs/faqs/	                    /docs/faq/
-/docs/*     	                /v0.13.2/:splat
+/docs/*     	                /v0.14.0-rc.0/:splat
```
ellistarn, line 5: I don't think we want to redirect the latest version to 0.14

website/config.yaml
ellistarn: Agreed! Perhaps not even cut new docs and just let users rely on preview?

website/static/_redirects
suket22: Yup I was thinking that too. I'll change.

website/config.yaml
suket22: Yeah I can definitely go either way on this. Let me change.

suket22: Alright so if you were to manually go to `karpenter.sh/v0.14.0-rc.0/` the docs will show, but we won't redirect you there, and it's not in our dropdowns. You also get the warning that it's not the latest stable release which works out

ellistarn: Beautiful

ellistarn (approved):

PR #2223, opened 2022-07-29
chore: add make target for setup which creates IAM roles

jonathan-innis:
Fixes # 

**Description**

- Adds role creation as a make target `make setup` so that the first time karpenter-controller is onboarded with `make apply`, `make setup` will cause the role creation to succeed

* Ran `make setup`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```


hack/setup-roles.sh
```diff
@@ -0,0 +1,14 @@
+SCRIPT_DIR=website/content/en/preview/getting-started/getting-started-with-eksctl/scripts
+TEMP_DIR=$(mktemp -d)
```
ellistarn, line 2: Do you even need the temp dir, since you can just execute these straight out of the file system?

ellistarn (approved):

PR #2224, opened 2022-07-29
fix: Consider unready nodes as in flight

ellistarn:
Fixes #https://github.com/aws/karpenter/issues/2164 

**Description**
During large scale ups, nodes sometimes flip/flop from ready to not read as things come online. This change considers unready nodes as in flight to avoid overscaling. 

In the case of hardware, networking, or other failure, it's possible for a node to transition from ready to unready. Previously, Kubernetes would evict the pods on the node after 5 minutes (default), triggering additional scale out. Instead, Karpenter will consider these taints as ephemeral and the node as in flight. Intervention (automated or otherwise) is required to remove these nodes. 

This may also be considered as a safety feature. If some issue caused nodes to consistently transition Ready -> NotReady, Karpenter would keep scaling out to provisioner limits. Instead, this waits for the nodes to recover or an operator to intervene. Additional automation (e.g. https://github.com/aws/karpenter/issues/2021) can address this case.

* `TEST_FILTER=TestUtilization make e2etests`

WIP

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```

### DIFF
--- a/pkg/controllers/provisioning/scheduling/inflightnode.go
+++ b/pkg/controllers/provisioning/scheduling/inflightnode.go
@@ -18,10 +18,11 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
-
 	v1 "k8s.io/api/core/v1"
 
+	"github.com/samber/lo"
+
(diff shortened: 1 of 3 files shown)


tzneal (approved):
lgtm

PR #2227, opened 2022-07-30
feat: add ebsOptimized bool to AWSNodeTemplate spec

joemiller:
**Description**
Add new boolean `ebsOptimized` to the `AWSNodeTemplate` spec for creating EBS-Optimized nodes.

* Unit tests
* manual testing in a dev EKS cluster

I am not sure. Looking for advise on this.

**Release Note**

```release-note 
New boolean `ebsOptimized` in `AWSNodeTemplate` to select EBS-Optimized nodes.
```


pkg/cloudprovider/aws/apis/v1alpha1/provider.go
```diff
@@ -78,6 +78,13 @@ type LaunchTemplate struct {
 	// BlockDeviceMappings to be applied to provisioned nodes.
 	// +optionals
 	BlockDeviceMappings []*BlockDeviceMapping `json:"blockDeviceMappings,omitempty"`
+	// EBSOptimized indicates whether the instance is optimized for Amazon EBS I/O.
+	// This optimization provides dedicated throughput to Amazon EBS and an optimized
+	// configuration stack to provide optimal Amazon EBS I/O performance. This
+	// optimization isn't available with all instance types. Additional usage charges
+	// apply when using an EBS-optimized instance.
+	// +optional
+	EBSOptimized *bool `json:"ebsOptimized,omitempty"`
```
ellistarn, line 87: Can you add a comment to the LaunchTemplate struct that this contains a subset of fields from https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RequestLaunchTemplateData.html

pkg/cloudprovider/aws/apis/v1alpha1/provider.go
```diff
@@ -78,6 +78,13 @@ type LaunchTemplate struct {
 	// BlockDeviceMappings to be applied to provisioned nodes.
 	// +optionals
```
ellistarn, line 79:
```suggestion
	// +optional
```

ellistarn (approved):
Awesome feature! Thanks! A couple of optional documentation requests

pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -173,6 +173,7 @@ func (p *LaunchTemplateProvider) createLaunchTemplate(ctx context.Context, optio
 		LaunchTemplateName: aws.String(launchTemplateName(options)),
 		LaunchTemplateData: &ec2.RequestLaunchTemplateData{
 			BlockDeviceMappings: p.blockDeviceMappings(options.BlockDeviceMappings),
+			EbsOptimized:        aws.Bool(options.EBSOptimized),
```
bwagner5, line 176: I believe you get an error if the launch template turns on ebs optimized and the instance type doesn't support it. I think we'd also need to add a scheduling label so that a user can easily filter for instance types that support EBS optimized.

bwagner5 (changes requested):

pkg/cloudprovider/aws/launchtemplate.go
joemiller: @bwagner5 I believe you're right on that. Is there some special handling of such an error case that we should do in the code? Or documentation we should add (which might already be covered in the struct comment pulled from the public AWS docs which indicates  `this isn't available with all instance types`) ?

joemiller: @bwagner5 There is a comment in the linked file `// Well Known to AWS`. Are these well-known labels documented somewhere? My searches are not finding anything. If there's a well-known label for ebs-optimized I'd like to match it, or if not, then I would put the definition into the `// Well known to Karpenter` section instead.

bwagner5: After looking into this more, I doubt there are cases where you'd want both instance types that support EBS optimized and those that don't. In fact, it seems like all new instance types `T3+`,  `i3+`, `*4+` all have EBS Optimized on by default.  Is an ebs-optimized attribute actually necessary? Which instance types are you trying to use where you need to manually turn on ebs optimized?

bwagner5: Below are all the instance types that do not support EBS Optimized (`unsupported`) or `supported` but not enabled by default. All other instance types have EBS Optimized on by default. These are all legacy instance types, so I'd recommend that we not add ebsOptimized to the NodeTemplate. 

|Instance type | EBS optimization support|
-- | --
  t1.micro | unsupported
   t2.nano | unsupported
   t2.micro | unsupported
   t2.small | unsupported
   t2.medium | unsupported
   t2.large | unsupported
   t2.xlarge | unsupported
   t2.2xlarge | unsupported
  c1.medium | unsupported
   c3.large | unsupported
   c3.8xlarge | unsupported
   cc2.8xlarge | unsupported
   g2.8xlarge | unsupported
   i2.8xlarge | unsupported
   m1.small | unsupported
   m1.medium | unsupported
   m2.xlarge | unsupported
  m3.medium | unsupported
   m3.large | unsupported
   r3.large | unsupported
   r3.8xlarge | unsupported
   c1.xlarge | supported
   c3.xlarge | supported
   c3.2xlarge | supported
   c3.4xlarge | supported
   g2.2xlarge | supported
   i2.xlarge | supported
   i2.2xlarge | supported
   i2.4xlarge | supported
   m1.large | supported
   m1.xlarge | supported
   m2.2xlarge | supported
   m2.4xlarge | supported
   m3.xlarge | supported
   m3.2xlarge | supported
   r3.xlarge | supported
   r3.2xlarge | supported
   r3.4xlarge | supported

FernandoMiguel: Out of that list, t2 are still fairly familiar.
But we can inch folks to use t3 instead

bwagner5: Yeah t3's are definitely preferable to t2's, in fact, Karpenter filters out t2's by default unless directly specified in the provisioner spec.

FernandoMiguel: > Yeah t3's are definitely preferable to t2's, in fact, Karpenter filters out t2's by default unless directly specified in the provisioner spec. 

Good to know.
Is this documented anywhere?

bwagner5: It doesn't look like it, we should definitely add that somewhere.   Here are the filters that we apply by default:

https://github.com/aws/karpenter/blob/f4594705f9d20d3d1568fe5f989112dabbc51579/pkg/cloudprovider/aws/cloudprovider.go#L156-L178
https://github.com/aws/karpenter/blob/f4594705f9d20d3d1568fe5f989112dabbc51579/pkg/cloudprovider/aws/instancetypes.go#L153-L197

JTCunning:
Hey @bwagner5, I'm one of @joemiller's teammates. I'd like to shed some light on why we're explicitly setting this boolean in the launch template even though AWS advertises that the instances we personally plan to use have it enabled by default.
We're up for any suggestions about [dis]allowing the boolean to be set on certain instance types, and not supplying the boolean at all by default, but the following behavior is why we believe it needs to be set in the launch template itself.

I've attached screenshots of the behavior after redacting the irrelevant identifying information.

After our initial installation of Karpenter, we noticed that root volumes of instances having `EBS optimization: disabled` in the Root device details:
disabled_instance_1

After manually stopping the instance, we confirmed that the checkbox was indeed set, and that there was no way to unset it:
disabled_instance_checkbox_1

Before Karpenter removed the instance's launch template after the grace period, we made sure that the boolean was unset:
instance_disabled_lt_1
disabled_lt_1

This led us to Joe's change. After adding the new CRD and our forked build of the Karpenter image, we were able to explicitly set the boolean in the launch template, leading to a different machine getting the root volume properly set to true:
enabled_lt_1
enabled_instance_lt_1
enabled_instance_1

We have not contacted AWS about this behavior because we've previously explicitly set the boolean to true for all of our CAS-controlled ASGs + launch templates.

Happy to answer any additional questions on what specific combinations of settings we're using for Karpenter, but I'm inclined to believe that Karpenter has nothing to do with the current behavior we're observing.

bwagner5:
Thanks for the details @JTCunning! I will follow up and look into this behavior.

bwagner5:
> Thanks for the details @JTCunning! I will follow up and look into this behavior.

Even though the field says "Disabled" for EBSOptimized, it is actually enabled for all instance types that support it by default. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-optimized.html#:~:text=EBS%20optimized%20by%20default

PR #2235, opened 2022-08-01
feat: Add a node liveness TTL that terminates NotReady nodes if optional field is set on provisioner

htoo97:
Fixes #2021 

**Description**
A new optional field `TTLSecondsAfterNotReady` is added to the provisioner spec to provide an option to terminate nodes that never become ready after starting up within the TTL. 

* Unit tests

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/charts/karpenter/crds/karpenter.sh_provisioners.yaml
+++ b/charts/karpenter/crds/karpenter.sh_provisioners.yaml
@@ -218,6 +218,17 @@ spec:
                   not set."
                 format: int64
                 type: integer
+              ttlSecondsAfterNotReady:
+                description: 'TTLSecondsAfterNotReady is the number of seconds before
+                  a node that does not become Ready will be terminated. It is disabled
+                  if this field is not set, which means a node that joins the cluster
+                  but never become ready will cause pods assigned to the node to never
+                  start Running. Note: Seeing the TTL too low can hastily terminate
+                  nodes that otherwise would''ve joined given some time, especially
+                  for large clusters. It may also hinder debugging the underlying
+                  root problem of the node if it goes away too soon.'
+                format: int64
+                type: integer
               ttlSecondsUntilExpired:
                 description: "TTLSecondsUntilExpired is the number of seconds the
                   controller will wait before terminating a node, measured from when
--- a/charts/karpenter/crds/karpenter.sh_provisioners.yaml
+++ b/charts/karpenter/crds/karpenter.sh_provisioners.yaml
@@ -222,8 +222,8 @@ spec:
                 description: 'TTLSecondsAfterNotReady is the number of seconds before
                   a node that does not become Ready will be terminated. It is disabled
                   if this field is not set, which means a node that joins the cluster
-                  but never become ready will cause pods assigned to the node to never
-                  start Running. Note: Seeing the TTL too low can hastily terminate
+                  but never becomes Ready will cause pods assigned to the node to never
+                  start Running. Note: Setting the TTL too low can hastily terminate
                   nodes that otherwise would''ve joined given some time, especially
                   for large clusters. It may also hinder debugging the underlying
                   root problem of the node if it goes away too soon.'
--- a/pkg/controllers/node/controller.go
+++ b/pkg/controllers/node/controller.go
@@ -17,6 +17,7 @@ package node
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"go.uber.org/multierr"
 	v1 "k8s.io/api/core/v1"
@@ -36,6 +37,7 @@ import (
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/controllers/state"
+	"github.com/aws/karpenter/pkg/utils/injectabletime"
 	"github.com/aws/karpenter/pkg/utils/result"
 )
 
@@ -150,4 +152,22 @@ func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
 		).
 		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
 		Complete(c)
+}
+
+// TriggerTerminationIfExpired is a helper function to be called by other controllers that want to trigger the
+// termination of nodes.
+func TriggerTerminationIfExpired(ctx context.Context, ttlSeconds int64, node *v1.Node, kubeClient client.Client,
+	reason string) (
+	time.Time, error) {
+	expirationTTL := time.Duration(ttlSeconds) * time.Second
+	expirationTime := node.CreationTimestamp.Add(expirationTTL)
+	if injectabletime.Now().After(expirationTime) {
+		logging.FromContext(ctx).Infof("Triggering termination for %s node after %s (+%s)",
+			reason, expirationTTL, time.Since(expirationTime))
+		if err := kubeClient.Delete(ctx, node); err != nil {
+			return expirationTime, fmt.Errorf("deleting node, %w", err)
+		}
+	}
+
+	return expirationTime, nil
 }
--- a/pkg/controllers/node/controller.go
+++ b/pkg/controllers/node/controller.go
@@ -17,7 +17,6 @@ package node
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"go.uber.org/multierr"
 	v1 "k8s.io/api/core/v1"
@@ -37,7 +36,6 @@ import (
 	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/controllers/state"
-	"github.com/aws/karpenter/pkg/utils/injectabletime"
 	"github.com/aws/karpenter/pkg/utils/result"
 )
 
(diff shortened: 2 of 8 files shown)


ellistarn:
Excited to see this feature. I spent a couple minutes with @bwagner5 trying to decipher the difference between 
https://github.com/kubernetes/api/blob/master/core/v1/well_known_taints.go#L27 and https://github.com/kubernetes/api/blob/master/core/v1/well_known_taints.go#L22, and which one we should be using for this feature.

I think it's correct behavior for Karpenter to not touch nodes experiencing CPU/Memory/PID pressure.

pkg/controllers/node/controller.go
```diff
@@ -150,3 +152,21 @@ func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
 		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
 		Complete(c)
 }
+
+// TriggerTerminationIfExpired is a helper function to be called by other controllers that want to trigger the
+// termination of nodes.
+func TriggerTerminationIfExpired(ctx context.Context, ttlSeconds int64, node *v1.Node, kubeClient client.Client,
```
ellistarn, line 158: This comment is totally optional and if you disagree with it, I'm happy to merge w/ it as is.

We generally try to avoid static helpers that pass injected dependencies (e.g. kubeclient), parameters only used for logging (e.g. reason), and parameters that are only used by some callers (e.g. return time.Time). Given the relative simplicity of this code if duplicated, being DRY in this case may not be worth it.

pkg/controllers/node/initialization.go
```diff
@@ -47,6 +48,15 @@ func (r *Initialization) Reconcile(ctx context.Context, provisioner *v1alpha5.Pr
 		return reconcile.Result{}, fmt.Errorf("determining instance type, %w", err)
 	}
 	if !r.isInitialized(n, provisioner, instanceType) {
```
ellistarn, line 50: I think we want to terminate nodes irrespective of whether or not they've been initialized. 

If a node flip flops from NotReady (initializing) -> Ready -> NotReady (network issue), then this TTL should kick in on this node as well.

pkg/controllers/node/controller.go
htoo97: No strong opinion from me either way! I'll refactor to adhere to the style of the rest of the repo.

pkg/controllers/node/initialization.go
htoo97: Hmm I was imagining this feature to be more restricted to startup issues preventing nodes from becoming ready. I've seen nodes flap readiness in the past when a certain resource like disk comes under pressure, and terminating the nodes in that case, I'm afraid would hinder them even more - it would just be moving the pods around instead of giving them an opportunity to finish.

Another point also is that if we implement termination here without the init check, a control plane failure or network issues causing segmentation between the control plane and workers (which would result in nodes becoming `NotReady`) would cause all karpenter provisioned nodes in the cluster to start terminating, which is potentially dangerous behavior IMO.

Even without those drastic failure scenarios, I would be concerned with a controller having the potential to take away nodes at any point in time without having any bounds (e.g. no more than 5% of nodes within x hours, etc).

ellistarn: > I've seen nodes flap readiness in the past when a certain resource like disk comes under pressure, and terminating the nodes in that case, I'm afraid would hinder them even more - it would just be moving the pods around instead of giving them an opportunity to finish.

Agree. We definitely don't want to flag here. Can the `Unreachable` condition help with this?

> Another point also is that if we implement termination here without the init check, a control plane failure or network issues causing segmentation between the control plane and workers (which would result in nodes becoming NotReady) would cause all karpenter provisioned nodes in the cluster to start terminating, which is potentially dangerous behavior IMO.

Agreed! We typically call this static stability. e.g., how long should things continue to run/drift after (some part of) the dataplane gets partitioned from the control plane. In Kubernetes, pods get default tolerations for Unreachable and Unready that give them a static stability of 5 minutes. I was imagining that the TTL for this use case would typically be set to something like 6 hours+, to clean up dead nodes, but allow for long recovery windows.

> Even without those drastic failure scenarios, I would be concerned with a controller having the potential to take away nodes at any point in time without having any bounds (e.g. no more than 5% of nodes within x hours, etc).

We've talked at length about a NodeDisruptionBudget, but haven't found time for. it yet: https://github.com/aws/karpenter/issues/1738

After this discussion, I'm a little confused with what's motivating this feature if not to support liveness termination after readiness. Karpenter currently terminates nodes that fail to connect to the control plane after 15 minutes, but this is unconfigurable. Can you describe your use case where you want to configure this?

htoo97: > Can the Unreachable condition help with this?
I might be mistaken and I haven't looked into this code in a while, but anecdotally, `Unreachable` taints are added to nodes that get segmented away from the control plane. The [docs](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions) also seem to suggest the same. Making karpenter terminate all unreachable nodes after some TTL exactly means static stability would be disrupted, if the control plane was just momentarily down. While the default k8s eviction does have an aggressive 5min TTL, in practice, the node controller in kube-controller-manager is also segmented away with the control plane and as soon as the control plane <> data plane connectivity recovers, we've seen the control plane remove the taints and not go through with the evictions. If this feature starts allowing nodes to get terminated, Karpenter would disrupt this behavior, in the sense that whatever safeguards for network partitions the control plane has by default would not be disrupted and the nodes would get terminated immediately once the karpenter pods can talk to the control plane again after a hypothetical partition.

> Karpenter currently terminates nodes that fail to connect to the control plane after 15 minutes, but this is unconfigurable. Can you describe your use case where you want to configure this?
Yeah absolutely. The exact situation we ran into was not the node being NonReady because of a network issue, so the termination you mentioned never kicked in for us. It was an issue with the subnet that the node is on running out of IPs. In this case, we would like Karpenter to tear the node down and reprovision so that there's a chance for it to land on a non-empty subnet instead of just being stuck. The example mentioned here is a very specific one, but this feature would apply for any other potential init issues.

Having a `NodeDisruptionBudget` sounds very exciting for some time the future!

ellistarn: I'm on vacation next week. Will try to get another maintainer to help with this.

bwagner5: A node that never gets past Initialization sounds like a good place to start with at least. The stability of a cluster after it goes ready may require some more thought and testing. 

Also, on the "out of IPs" issue you faced; I've heard this from several users and I think we can do something a little better in the AWS Cloud Provider to temporarily remove the zonal offerings when IP exhaustion happens.

charts/karpenter/crds/karpenter.sh_provisioners.yaml
```diff
@@ -218,6 +218,17 @@ spec:
                   not set."
                 format: int64
                 type: integer
+              ttlSecondsAfterNotReady:
```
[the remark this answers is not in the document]
htoo97: I'm personally open to changing it to be an internal implementation and not an API, but as a user that does need this feature, I'd prefer having the ability to configure it flexibly on each provisioner, or even just on the karpenter controller globally. That said, I don't know what the request frequency for this feature that you folks have seen on your end (i.e. if you think not many users will find this API useful), so I'll defer to your decision.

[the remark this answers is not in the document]
htoo97: I can also rename it to `ttlSecondsAfterNotReadyOnInit` if that makes it clearer.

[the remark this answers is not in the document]
ellistarn: We typically delay as much as possible before adding API surface, especially if we expect that most users won't change the setting. Once we find an example of two users with valid but conflicting use cases, that's typically when we expose new API params.

tzneal:
Additional request from a customer this feature to also handle the situation where nodes go not ready after initialization. This would be a `ttlSecondsAfterNotReady` which worked for both initialization and post-initialization if the node goes NotReady for some reason.

charts/karpenter/crds/karpenter.sh_provisioners.yaml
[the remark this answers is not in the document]
dewjam: #2544 Seems somewhat related.

dewjam:
Discussed with other Karpenter folks and the general consensus is exposing `ttlSecondsAfterNotReady` through the API is an appropriate path.  The initial implementation could remain focused on terminating new nodes that never enter "Ready" state (after the Karpenter initialization process).  This could be extended later to include recovery of "Ready" nodes which then enter a "Not Ready" state.

dewjam:
Hello @htoo97 ,
We are working on a design doc to help guide our direction on this feature.  I should hopefully have a draft available sometime early next week.  Until then, we should probably hold off on merging this code.  Just so we're sure it aligns with the intended design.

Let me know if you have any questions/concerns.  Thanks!

htoo97:
Sounds perfect, thanks!

omgapuppy:
Quite keen on this feature - is there any movement with the design doc or merging this?

omgapuppy:
> @omgapuppy, we're currently working on a design for this. It has pretty broad implications, so we're taking care to get it right.

Thanks for the quick update!

ajaykumarmandapati:
Quite interested in this feature - is there any movement with the design doc or merging this PR?

jonathan-innis:
Closing in favor of https://github.com/aws/karpenter-core/pull/191. Once this feature is released along with the work associated with [Node Ownership](https://github.com/aws/karpenter/blob/main/designs/node-ownership.md), we expect that a large number of user issues around this should go away.

PR #2238, opened 2022-08-01
docs: Add a docs example for weighted provisioners

jonathan-innis:
Fixes # 

**Description**

- Adds an example to the docs for deploying two provisioners, one with priority and the other as a backup

* 

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```

### DIFF
--- a/website/content/en/preview/tasks/provisioning.md
+++ b/website/content/en/preview/tasks/provisioning.md
@@ -113,7 +113,40 @@ spec:
 ```
 In order for a pod to run on a node defined in this provisioner, it must tolerate `nvidia.com/gpu` in its pod spec.
 
-### Example: Adding the Cilium Startup Taint
+## Example: Creating Provisioner Hierarchies with Weights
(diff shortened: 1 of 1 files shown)


website/content/en/preview/tasks/provisioning.md
```diff
@@ -113,7 +113,40 @@ spec:
 ```
 In order for a pod to run on a node defined in this provisioner, it must tolerate `nvidia.com/gpu` in its pod spec.
 
-### Example: Adding the Cilium Startup Taint
+## Example: Creating Provisioner Hierarchies with Weights
```
ellistarn, line 116: I think this is better framed as Preferences to map to existing concepts of pod level preferences. This should include some discussion about whether or not the instance types are available. 

Thinking more on this feature -- how does this relate to @tzneal's consolidation work? Will consolidations be considered across weights?

PR #2242, opened 2022-08-02
fix: get node state before pending pods

tzneal:
**Description**

Get the node state before the list of pending pods and copy the
node state prior to using it so that binding pods don't affect it
mid-schedule.

* Unit testing & deployed to EKS.  Deflakes the Utilization e2e test

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


pkg/controllers/state/cluster.go
```diff
@@ -107,6 +107,26 @@ type Node struct {
 	PodTotalLimits v1.ResourceList
 }
 
+// Copy creates a copy of the state node suitable for capturing a snapshot of the current node state. It's not a true
+// deep copy, but does contain a deep copy of all of the values that are updated over time by the cluster state. This
+// allows clients to use a .Copy() of the node to represent a snapshot of the node state which won't be modified
+// (e.g. the amount of Allocatable resources won't diminish as pods continue to bind to the node)
+func (n *Node) Copy() *Node {
```
ellistarn, line 114: Not blocking. We've seen a couple of cases (cloud provider cache, instance types) where structs like this can drift, introducing bugs. Do you have any ideas for how we can protect against this? Maybe a comment on the state.Node struct?

ellistarn (dismissed):
Nice!

pkg/controllers/state/cluster.go
tzneal: I'm starting to think we should really deep-copy it all.  It's probably not a lot more expensive and is certainly less weird.  We'll do it once per node per scheduling batch and we already need to deep copy most of the node anyway.

ellistarn (approved):

PR #2246, opened 2022-08-02
Provisioner Prioritization One Pager

jonathan-innis:
Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


designs/provisioner-priority.md
```diff
@@ -0,0 +1,196 @@
...
+      name: arm64
+    spec:
+      weight: 50
+      requirements:
+      - key: kubernetes.io/arch
+        operator: In
+        values: ["arm64"]
+    ```
+
+2. Creating a strongly preferred default provisioner for enterprise users that have `reserved` instance types and want to prefer having these instance types scheduled prior to other instance types. This assumes that we eventually support the `karpenter.sh/capacity-type=reserved` label.
```
bwagner5, line 45: we don't need to assume the reserved label. It would be reasonable for users to just fill in the instance types that they have purchases RIs for in the provisioner.

designs/provisioner-priority.md
```diff
@@ -0,0 +1,196 @@
...
+
+These constraints are consistent with pod affinity/pod anti-affinity preference weights described [here](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#an-example-of-a-pod-that-uses-pod-affinity).
+
+## Considerations
+
+### Provisioner Scheduling Ordering
+
+__Current State__: Scheduling calls the Kubernetes LIST API for the `karpenter.sh/v1alpha5/Provisioner` and receives a random ordering of Provisioners back for scheduling. When processing pods, Karpenter will find the first provisioner that it is able to schedule the pod onto given the instance type options that are constrained by the pod scheduling requirements (affinity, anti-affinity, topology spread) and the provisioner requirements. This creates some randomness for users and leads to some unpredictability with respect to the provisioner that will be used for a given pod. Currently, the best practices that are recommended by the [Karpenter/EKS docs](https://aws.github.io/aws-eks-best-practices/karpenter/#create-provisioners-that-are-mutually-exclusive) recommend that we create mutually exclusive provisioners such that we reduce this unpredictability.
+
+**Improvement Options**
```
bwagner5, line 134: You might add an option where `Provisioner` matching is scored for an implicit weight rather than the simple `Compatible` check we do today. This would solve for preferring more requirements matches like in the taints case above.

bwagner5 (commented):
nice doc! just a comment and 1 suggestion from me

designs/provisioner-priority.md
jonathan-innis: Ah, got it makes sense. My assumption is that these instance types fall under the `on-demand` capacity type

designs/provisioner-priority.md
jonathan-innis: I'm a bit concerned about how this will affect the bin-packing logic. In general, since we are bin-packing on nodes we have already decided to provision, if we get new pods, we are going to add to the current node, not prefer other provisioners over this current one

designs/provisioner-priority.md
bwagner5: yes, RIs are a billing concept, not a capacity type, but they do map to on-demand instances.

designs/provisioner-priority.md
```diff
@@ -0,0 +1,268 @@
+# Prioritizing Provisioners
+
+## Goals
+
+- Allowing a user to describe a logical ordering to their provisioners
+- Allowing a user to set defaults and/or fallback orderings for their provisioners
+
+## Background
+
+Currently, there is no logical ordering to how the Karpenter scheduling receives and processes Provisioners. Provisioners are processed in the order they are received from the Kubernetes LIST API, which is randomly. Like pod affinities, users want to convey a preferential relationship to some Provisioners over others. Rather than having to manually map a pod to a specific Provisioner through a `nodeSelector` or `affinity` (which could be cumbersome for clusters with large workloads), a user should be able to specify global preferences to pod assignment to provisioners.
```
ellistarn, line 10: Worth mentioning that we currently recommend that provisioners are mutually exclusive in the docs. We'll need to update this.

designs/provisioner-priority.md
```diff
@@ -0,0 +1,268 @@
...
+    metadata:
+      name: fallback
+    spec:
+      requirements:
+      - key: "node.kubernetes.io/instance-type"
+        operator: In
+        values: ["m5.large", "m5.2xlarge"]
+    ```
+
+3. Allowing provisioner with taints to be attempted first so that pods that have tolerations for these taints can be scheduled to specific instance types. Without this ordering, it is possible these pods will be scheduled to nodes that have no tolerations.
```
bwagner5, line 81:
```suggestion
3. Allowing provisioner with taints to be attempted first so that pods that have tolerations for these taints can be scheduled to specific instance types. Without this ordering, it is possible these pods will be scheduled to nodes that have no taints.
```

designs/provisioner-priority.md
```diff
@@ -0,0 +1,268 @@
...
+    apiVersion: karpenter.sh/v1alpha5
+    kind: Provisioner
+    metadata:
+      name: arm64
+    spec:
+      weight: 50
+      requirements:
+      - key: kubernetes.io/arch
+        operator: In
+        values: ["arm64"]
```
ellistarn, line 45: Even better, you could just be flexible w/ this one using `Exists`

designs/provisioner-priority.md
```diff
@@ -0,0 +1,268 @@
...
+    ```
+
+2. Creating a strongly preferred default provisioner for enterprise users that have `reserved` instance types and want to prefer having these instance types scheduled prior to other instance types.
+
+   **Example**
+
+    ```yaml
+    # I have c5.large instance types as reserved instances, so I want to schedule to these nodes first
```
tzneal, line 53: Does this make sense?  Don't you have a fixed number of RI and this will keep preferring that regardless until they fail to launch.

designs/provisioner-priority.md
jonathan-innis: Sure, I think I mentioned this somewhere down below. Could probably bring this up into the background section.

tzneal: I wonder if that shouldn't still be the recommendation.  Weighting provisioners has some use cases, but in the event that we launch nodes from multiple provisioners in a single scheduling batch with pods that could go to either, we don't control what gets scheduled where.

designs/provisioner-priority.md
jonathan-innis: At the point that it fails to launch this instance type due to ICE, we will null it out in our `getInstaceTypes` call so at that point it should no longer recommend this instance, right? What is the issue with this being the user experience?

designs/provisioner-priority.md
```diff
@@ -0,0 +1,268 @@
...
+
+Because we scheduled with this provisioner and the node has capacity, all the other pods that were part of this batch may end up on that single node, even though there were some pods in this batch that would have tolerated the taints.
+
+**Options**
+
+1. Use a `karpenter.sh/weight` annotation on pods to tell Karpenter to order these pods first by weight ahead of any resource ordering that is performed for bin-packing.
+2. Do not change any of the existing logic and document that this is a consideration to take into account when prioritizing provisioners.
+
+**Recommendation:** Do not change the existing logic. In general, the fact that the scheduler may choose to schedule a second provisioner first due to constraints is a general concern with the kube-scheduler as we can't even guarantee which node each pod will end up on. Instead, document that provisioner weight is just a preference given in scheduling and may not actually mean that all pods that could be scheduled to a provisioner with higher weight will always end up on a provisioner with higher weight.
```
ellistarn, line 192: We should document that users should use both tolerations and nodeselectors to enforce isolation

designs/provisioner-priority.md
bwagner5: The experience isn't great, but there's not a great way around it with RIs. You wouldn't get ICE'd once you run out, you'd just lose your discount.

tzneal: Magic: Put a CPU limit on the provisioner that only lets you launch up to the # of RIs you have

designs/provisioner-priority.md
jonathan-innis: In the case that we have `Exists`, we are saying that the pod requirement must have this label or that the node must have this label? If it's the former, makes sense. If it's the latter, not sure how this helps the scenario that we are suggested

bwagner5 (approved):
lgtm

Issue #2100, opened 2022-07-11
node cleanup fails when instance doesn't exist any more


universam1:
## Version

**Karpenter**: v0.13.2

**Kubernetes**: v1.21.12-eks-a64ea69 

## Expected Behavior
Karpenter detects that instance is not existent in AWS, and skips the terminate action.

**Problems found:**
1. NodeTerminationHandler is not possible to be used: https://github.com/aws/aws-node-termination-handler/issues/654
2. When node is gone, the **tag filter is not matching**:  https://karpenter.sh/v0.13.2/getting-started/migrating-from-cas/
```json
{
            "Action": "ec2:TerminateInstances",
            "Condition": {
                "StringLike": {
                    "ec2:ResourceTag/Name": "*karpenter*"
```
3. Karpenter tries to delete a node that is not existent and fails.

## Actual Behavior
When the node is terminated in AWS due to spot interruption and K. was not aware of it, the node is not deleted in Kubernetes and Karpenter loops into this error:
```
terminating instance ip-172-16-101-161.ec2.internal, UnauthorizedOperation: You are not authorized to perform this operation
```

## Steps to Reproduce the Problem
- scale K. to zero instance
- terminate the node, wait until gone
- scale K. up

## Resource Specs and Logs

```json
{
"level": "ERROR",
"time": "2022-07-11T11:45:47.555Z",
"logger": "controller.controller.termination",
"message": "Reconciler error",
"commit": "1f7a67b",
"reconciler group": "",
"reconciler kind": "Node",
"name": "ip-172-16-101-161.ec2.internal",
"namespace": "",
"error": "terminating node ip-172-16-101-161.ec2.internal, terminating cloudprovider instance, terminating instance ip-172-16-101-161.ec2.internal, UnauthorizedOperation: You are not authorized to perform this operation. Encoded authorization failure message: xxxxx\n\tstatus code: 403, request id: xxxx"
}
```

decoded failure message:
```json
{
    "allowed": false,
    "explicitDeny": false,
    "matchedStatements": {
        "items": []
    },
    "failures": {
        "items": []
    },
    "context": {
        "principal": {
            "id": "xxx:be41e23e-xxxx-KarpenterControllerIamRo-6HZ0",
            "arn": "arn:aws:sts::xxxx:assumed-role/xxxx-KarpenterControllerIamRo-6HZ0EFA4MZ0X/be41e23e-xxxx-KarpenterControllerIamRo-6HZ0"
        },
[...]

Issue #1796, opened 2022-05-11
Errors continue despite terminating nodes manually


andrewhibbert:
## Version

**Karpenter**: v0.0.0

**Kubernetes**: v1.0.0

## Expected Behavior

Node terminated manually and reconciler recognizes this and stops erroring

## Actual Behavior

Keeps erroring

```
[...]

PR #2253, opened 2022-08-04
fix: check instance state on termination failure

dewjam:
**Fixes** # 
#2100 and #1796

**Description**
If tag conditions are used in Karpenter IAM policies for the TerminateInstances action, Karpenter can get stuck in a loop trying to terminate instances which were previously terminated (and reaped) without Karpenter's knowledge.  This fix adds a step where if instance Termination fails, Karpenter will then try to call DescribeInstances to determine the instances state.  If the state is "terminated" or if the instance isn't found, then skip the TerminateInstance call and proceed with removing the node in K8S.

Manually reproduced the issue by adding tag conditions to the Karpenter IAM policies for the TerminateInstances action.  Then launched an instance with said tags.  Then, manually removed the tags from the instance.  This simulates the case where a terminated instance no longer has the tags required for Karpenter to terminate.

Next, manually deleted the node in K8S, which caused Karpenter to continuously attempt to delete the EC2 instance, but fails with an UnauthorizedException error (this is intended behavior).
```
2022-08-04T14:30:27.498Z	ERROR	controller.controller.termination	Reconciler error	{"commit": "d94b7ac", "reconciler group": "", "reconciler kind": "Node", "name": "ip-192-168-178-4.us-west-2.compute.internal", "namespace": "", "error": "terminating node ip-192-168-178-4.us-west-2.compute.internal, terminating cloudprovider instance, terminating instance ip-192-168-178-4.us-west-2.compute.internal, UnauthorizedOperation: You are not authorized to perform this operation. Encoded authorization failure message: <message>\n\tstatus code: 403, request id: <id>"}
```

Finally, manually terminated the EC2 instance, leaving it in "terminated" state.  The next time Karpenter tries to terminate the instance, it also calls DescribeInstances and sees the instance is "terminated".  Karpenter then skips the TerminateInstance call and removes the node object.
```
2022-08-04T14:32:20.144Z	DEBUG	controller.termination	Instance already terminated, ip-192-168-178-4.us-west-2.compute.internal	{"commit": "d94b7ac", "node": "ip-192-168-178-4.us-west-2.compute.internal"}
2022-08-04T14:32:20.162Z	INFO	controller.termination	Deleted node	{"commit": "d94b7ac", "node": "ip-192-168-178-4.us-west-2.compute.internal"}
2022-08-04T14:32:35.692Z	DEBUG	controller.aws.launchtemplate	Deleted launch template Karpenter-dewaard-karpenter-demo-9688201425030023995 (lt-0a942f88cae582d73)	{"commit": "d94b7ac"}
```

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Adds support for deleting a node for an EC2 instance which was manually Terminated without Karpenter's knowledge
```

### DIFF
--- a/pkg/cloudprovider/aws/errors.go
+++ b/pkg/cloudprovider/aws/errors.go
@@ -43,12 +43,24 @@ var (
 
 type SpotFallbackError error
 
+type InstanceTerminatedError struct {
+	error
+}
+
 func isSpotFallback(err error) bool {
 	if err == nil {
 		return false
 	}
 	var sfbErr SpotFallbackError
 	return errors.As(err, &sfbErr)
+}
+
+func isInstanceTerminated(err error) bool {
+	if err == nil {
+		return false
+	}
+	var itErr InstanceTerminatedError
+	return errors.As(err, &itErr)
 }
 
 // isNotFound returns true if the err is an AWS error (even if it's
--- a/pkg/cloudprovider/aws/fake/ec2api.go
+++ b/pkg/cloudprovider/aws/fake/ec2api.go
@@ -136,6 +136,7 @@ func (e *EC2API) CreateFleetWithContext(_ context.Context, input *ec2.CreateFlee
 			if skipInstance {
 				continue
 			}
+			instanceState := ec2.InstanceStateNameRunning
 			for i := 0; i < int(*input.TargetCapacitySpecification.TotalTargetCapacity); i++ {
 				instance := &ec2.Instance{
 					InstanceId:            aws.String(test.RandomName()),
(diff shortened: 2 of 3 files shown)


pkg/cloudprovider/aws/errors.go
```diff
@@ -43,6 +43,10 @@ var (
 
 type SpotFallbackError error
 
+type InstanceTerminatedError struct {
```
bwagner5, line 46: can you just make this an error instead of a struct like `SpotFallbackError`?

dewjam: I actually started with that, but I think the `InstanceTerminatedError ` error ends up being the `*errors.errorString` type.  Which means that the `IsTerminatedError` would return true for errors when we expect them to return false.  Now that I think of it, I'm not 100% sure the `isSpotFallback` method is doing what we intend either.  Will have to test it further.

Here's an example, btw.
https://go.dev/play/p/cswSjwGQFL2

bwagner5: o wow, good catch!

bwagner5 (approved):
lgtm

jwcesign:
I meet this problem again, and it looks like the modification lost in https://github.com/aws/karpenter-provider-aws/pull/3129/files

jwcesign:
cc @dewjam @jonathan-innis

jwcesign:
OK, looks the error is different.

PR #2254, opened 2022-08-05
feature: Override `--system-reserved` when scheduling and in kubelet parameter startup

jonathan-innis:
Fixes # 

**Description**

Allow users to specify 

```yaml
systemReserved:
   cpu: 2
   memory: 5Gi
   ephemeral-storage: 10Gi
```

in their `.spec.kubeletConfiguration` to override kubelet configuration passed into bootstrap scripts and used for shceduling

* `make test`
* Manual validation of allocatable resources on deployed nodes

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```


pkg/apis/provisioning/v1alpha5/provisioner.go
```diff
@@ -98,6 +98,8 @@ type KubeletConfiguration struct {
 	// ContainerRuntime is the container runtime to be used with your worker nodes.
 	// +optional
 	ContainerRuntime *string `json:"containerRuntime,omitempty"`
+	// SystemReserved contains resources reserved for OS system daemons and kernel memory.
+	SystemReserved v1.ResourceList `json:"systemReserved,omitempty"`
```
ellistarn, line 102: Can you point to where this comes from? I don't see it defined in: https://github.com/kubernetes/kubernetes/blob/9f82d81e55cafdedab619ea25cabf5d42736dacf/cmd/kubelet/app/options/options.go#L53

jonathan-innis: https://github.com/kubernetes/kubernetes/blob/9f82d81e55cafdedab619ea25cabf5d42736dacf/cmd/kubelet/app/options/options.go#L498

ellistarn: Ah here it is specifically:

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/types.go#L360.

We try to map this configuration as close as possible to being a subset of what's upstream. Technically this is a map[string]string. I'd like to include the comment as well as a link to the source at the top of the struct where the other guidance/link is.

```
	// A set of ResourceName=ResourceQuantity (e.g. cpu=200m,memory=150G,ephemeral-storage=1G,pid=100) pairs
	// that describe resources reserved for non-kubernetes components.
	// Currently only cpu, memory and local ephemeral storage for root file system are supported.
	// See http://kubernetes.io/docs/user-guide/compute-resources for more detail.
	SystemReserved map[string]string
```

ellistarn:
One minor nit about the API. Otherwise nice work!

pkg/cloudprovider/aws/amifamily/bootstrap/eksbootstrap.go
```diff
@@ -107,6 +112,23 @@ func (e EKS) nodeLabelArg() string {
 	return fmt.Sprintf("%s%s", nodeLabelArg, strings.Join(labelStrings, ","))
 }
 
+// systemReservedArg gets the kubelet-defined arguments for any valid resource
+// values that are specified within the system reserved resource list
+func (e EKS) systemReservedArg() string {
+	var args []string
+	if e.KubeletConfig.SystemReserved != nil {
+		for _, name := range []v1.ResourceName{v1.ResourceCPU, v1.ResourceMemory, v1.ResourceEphemeralStorage} {
```
ellistarn, line 120: Any reason to limit to these types? Could we pass whatever's present in the map?

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -146,7 +146,7 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context, provisioner *v1alp
 	if err != nil {
 		return nil, err
 	}
-	instanceTypes, err := c.instanceTypeProvider.Get(ctx, aws)
+	instanceTypes, err := c.instanceTypeProvider.Get(ctx, aws, provisioner)
```
ellistarn, line 149: Thoughts on scoping this to v1alpha5.KubeletConfiguration?

pkg/cloudprovider/aws/amifamily/bootstrap/eksbootstrap.go
jonathan-innis: It's in the comment that you posted above. Only cpu, memory, and ephemeral storage are supported

pkg/apis/provisioning/v1alpha5/provisioner.go
jonathan-innis: The spec describes this as `ResourceName=ResourceQuantity` which is what `v1.ResourceList` is. Any reason we want to take restrictions that this imposes off of the spec? I think in general we want the strings to match a specific resource format.

pkg/cloudprovider/aws/cloudprovider.go
jonathan-innis: Sure, only thought was if we add details about VMOverhead, etc. later, we would potentially need the whole provisioner, but we can scope it down until we need it

pkg/apis/provisioning/v1alpha5/provisioner.go
ellistarn: It's a finer detail, I'm fine either way. Under the hood since resource list is just an alias type of map[string]string so it's technically equivalent. The conventional wisdom so far has been to defer to the upstream definitions so that we avoid coming up with our own concept for kubelet configuration. It's muddied by the fact that the code for kubelet is spread across so many places that it's not as coherent as it could be. I'm happy to drop this convention if it's more trouble than it's worth.

jonathan-innis: We are going to have to do the same validation on the the resource quantity format anyways if we don't take v1.ResourceList from upstream so my opinion is that we just go with the ResourceList since we get the validation and the conversion for free

bwagner5 (approved):
lgtm! nice work

PR #2260, opened 2022-08-08
fix: spot error type comparison

bwagner5:
Fixes # 
N/A

**Description**
 - Fixes an issue in the type comparison for identifying a spot flexibility error where the type was actually `errors.ErrorString` rather than SpotFallbackError. This would cause a `launchInstance` retry and constrain requirements to just spot even if on-demand was a valid requirement. This could also cause Spot to become a requirement if there was an error other than `LaunchTemplateNotFound` returned by `launchInstance` even if the provisioner was not flexible to Spot. 

* `make test`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/aws/errors.go
+++ b/pkg/cloudprovider/aws/errors.go
@@ -41,7 +41,9 @@ var (
 	}
 )
 
-type SpotFallbackError error
+type SpotFallbackError struct {
+	error
+}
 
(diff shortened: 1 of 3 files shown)


jonathan-innis (dismissed):
🚀

jonathan-innis (approved):

PR #2265, opened 2022-08-09
fix: only count active pods towards node requests

njtran:
**Description**
https://github.com/aws/karpenter/pull/2212

* Created a job with Resource requests and completions: 1000. Confirmed that resource metrics remained the same over the course of jobs completing.

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -255,10 +255,12 @@ func (c *Cluster) populateResourceRequests(ctx context.Context, node *v1.Node, n
 	var daemonsetLimits []v1.ResourceList
 	for i := range pods.Items {
 		pod := &pods.Items[i]
+		if podutils.IsTerminal(pod) {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


tzneal (approved):
lgtm

pkg/controllers/state/cluster.go
```diff
@@ -255,10 +255,12 @@ func (c *Cluster) populateResourceRequests(ctx context.Context, node *v1.Node, n
 	var daemonsetLimits []v1.ResourceList
 	for i := range pods.Items {
 		pod := &pods.Items[i]
+		if podutils.IsTerminal(pod) {
```
ellistarn, line 258: We've had this bug in so many places now. I wonder if there's a way for us to simply ignore terminal pods as a blanket policy. I think the cluster autoscaler might do this.

tzneal (approved):
lgtm

ellistarn (approved):

PR #2283, opened 2022-08-11
chore: Add pricing at the offerings level for spot instance types

### LINKED ISSUE #2256: Move Pricing to Offerings
What do you want us to build?

Currently, pricing is modeled at the InstanceType level. However, pricing is actually per availability zone and capacity type.    Moving the pricing information to the instance type offerings levels will provide more granular pricing information. 

This is required for consolidation efforts and resolves an issue where spot usage-class availability is only at the regional level from the DescribeInstanceTypes API, so pricing is an authoritative source for the spot offering being launched in a specific zone if a price exists. 

N/A

**Additional context**
Anything else we should know?

**Attachments**

### Community Note

jonathan-innis:
Fixes #2256 

**Description**

- Adds pricing data to the offerings level for spot and od instance types
- Sorting the instance types based on price is now achieved by determining the cheapest offering for the instance type and using that as the representative price
- Determining whether something is consolidatable is now determined at the offering-level for both spot and on-demand

* `make test`
* Manual validation of consolidation

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```


pkg/cloudprovider/aws/instancetype.go
```diff
@@ -92,25 +92,38 @@ func (i *InstanceType) Resources() v1.ResourceList {
 	return i.resources
 }
 
+// Price will get the minimum price of all offerings based on the passed filter
+// so that we can use the cheapest offerings at scheduling time
+func (i *InstanceType) Price(filters ...func(o cloudprovider.Offering) bool) float64 {
```
bwagner5, line 97: I don't think we need this `Price` func anymore since pricing is on Offerings and that is already exposed via the CloudProvider interface.  Rather than doing this filtering, I think the consolidation logic can do the offering parsing itself outside of the cloud provider. If it's shifted outside of the cloud provider, it will make onboarding new cloud providers easier, shifting complexity to the core.

pkg/cloudprovider/types.go
```diff
@@ -80,14 +80,29 @@ type InstanceType interface {
...
+	// for the instance type, or just a weighting where lower 'prices' are preferred.
+	Price(filters ...func(Offering) bool) float64
 }
 
 // An Offering describes where an InstanceType is available to be used, with the expectation that its properties
 // may be tightly coupled (e.g. the availability of an instance type in some zone is scoped to a capacity type)
-type Offering struct {
-	CapacityType string
-	Zone         string
+type Offering interface {
```
bwagner5, line 90: I don't think this needs to be an interface if the price parsing is shifted out of the cloud provider.

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -44,21 +44,21 @@ var (
...
+	fallbackPrice float64
 }
 
 func NewInstanceType(ctx context.Context, info *ec2.InstanceTypeInfo, kc *v1alpha5.KubeletConfiguration, price float64, provider *v1alpha1.AWS, offerings []cloudprovider.Offering) *InstanceType {
 	instanceType := &InstanceType{
 		InstanceTypeInfo: info,
 		provider:         provider,
 		offerings:        offerings,
-		price:            price,
+		fallbackPrice:    price,
```
bwagner5, line 61: I don't think we need fallbackPrice at the instance type level.

pkg/cloudprovider/aws/pricing.go
```diff
@@ -52,9 +53,13 @@ type PricingProvider struct {
 	onDemandUpdateTime time.Time
 	onDemandPrices     map[string]float64
 	spotUpdateTime     time.Time
-	spotPrices         map[string]float64
+	spotPrices         map[string]map[string]float64
 }
 
+// Default zone is used to store the pricing data based off of instance
+// type level information until zonal information is available
+const defaultZone = "default"
```
bwagner5, line 61: I don't quite understand this, when is zonal information not available?

pkg/cloudprovider/aws/pricing.go
```diff
@@ -137,7 +142,7 @@ func (p *PricingProvider) OnDemandLastUpdated() time.Time {
 func (p *PricingProvider) SpotLastUpdated() time.Time {
 	p.mu.RLock()
 	defer p.mu.RUnlock()
-	return p.onDemandUpdateTime
+	return p.spotUpdateTime
```
bwagner5, line 145: oh nice catch!

pkg/cloudprovider/aws/pricing.go
```diff
@@ -152,13 +157,17 @@ func (p *PricingProvider) OnDemandPrice(instanceType string) (float64, error) {
 	return price, nil
 }
 
-// SpotPrice returns the last known spot price for a given instance type, returning an error if there is no
+// SpotPrice returns the minimum spot price for a given instance type across zones, returning an error if there is no
 // known spot pricing for the instance type.
 func (p *PricingProvider) SpotPrice(instanceType string) (float64, error) {
```
bwagner5, line 162: do we still need this?

pkg/cloudprovider/aws/pricing.go
```diff
@@ -52,9 +53,13 @@ type PricingProvider struct {
 	onDemandUpdateTime time.Time
 	onDemandPrices     map[string]float64
 	spotUpdateTime     time.Time
-	spotPrices         map[string]float64
+	spotPrices         map[string]map[string]float64
```
bwagner5, line 56: it might be good to break this out into a type since it's a little weird to read. 
Alternatively, a comment indicating the mapping might be sufficient.

pkg/cloudprovider/aws/pricing.go
```diff
@@ -383,11 +419,29 @@ func (p *PricingProvider) updateSpotPricing(ctx context.Context) error {
...
 	p.spotUpdateTime = time.Now()
-	logging.FromContext(ctx).Infof("updated spot pricing with %d instance types", len(p.spotPrices))
+	logging.FromContext(ctx).Infof("updated spot pricing with %d instance types and %d offerings", len(p.spotPrices), totalOfferings)
 	return nil
 }
+
+// initialSpotPricing creates spot zonal pricing information based off of the
+// initial on-demand pricing information coming at the regional level
+func initialSpotPricing(odPricing map[string]float64) map[string]map[string]float64 {
+	m := make(map[string]map[string]float64)
```
bwagner5, line 441: I think it reads a little easier using the shorthand instantiation syntax (we tend to use this elsewhere in the codebase too).

```suggestion
	m := map[string]map[string]float64{}
```

pkg/controllers/consolidation/controller.go
```diff
@@ -192,6 +195,18 @@ func (c *Controller) candidateNodes(ctx context.Context) ([]candidateNode, error
...
 
+		// skip any nodes that we can't determine the capacity type or the topology zone for
+		ct, ok := n.Node.Labels[v1alpha5.LabelCapacityType]
+		if !ok {
+			logging.FromContext(ctx).Errorf("Getting node capacity type, %v", err)
+			return true
+		}
+		az, ok := n.Node.Labels[v1.LabelTopologyZone]
+		if !ok {
+			logging.FromContext(ctx).Errorf("Getting topologzy zone, %v", err)
```
bwagner5, line 206:
```suggestion
			logging.FromContext(ctx).Errorf("Getting topology zone, %v", err)
```

bwagner5 (changes requested):
Nice work! I think we can shift some of the logic around pricing parsing of the offerings into Karpenter core and get rid of some of the complexity of parsing pricing in each cloud provider.

pkg/cloudprovider/aws/pricing.go
jonathan-innis: I was thinking that if Karpenter's role didn't have permission on the SpotPriciningHistory API than we are never going to be able to get spot pricing information but we should still give back pricing based off of the on-demand pricing that we grabbed at startup. 

I would want to have static pricing data but we don't know what region this cluster is sitting in and therefore don't have specific data on the AZ-based data, unless we wanted to grab all the AZ-based data from all AZs across all regions. Seems like a lot of data when the instance will have permissions most of the time

jonathan-innis: I guess alternatively the call to `Price()` itself can fallback to the initial on-demand data when the data isn't there for spot

bwagner5: > I guess alternatively the call to Price() itself can fallback to the initial on-demand data when the data isn't there for spot

Exactly, and this is what's implemented on initialization right?

pkg/cloudprovider/aws/pricing.go
```diff
@@ -363,12 +385,18 @@ func (p *PricingProvider) updateSpotPricing(ctx context.Context) error {
...
-			existing, ok := prices[instanceType]
+			// on a per-zone basis
+			_, ok := prices[instanceType]
+			if !ok {
+				prices[instanceType] = make(map[string]*pricingInfo)
+			}
+			existing, ok := prices[instanceType][az]
 			if !ok || timeStamp.After(existing.timestamp) {
-				prices[instanceType] = &pricingInfo{
+				prices[instanceType][az] = &pricingInfo{
```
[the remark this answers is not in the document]
jonathan-innis: Just AZs in the region that we are in. The DecribeSpotPriceHistory API is regional

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -58,11 +58,6 @@ func NewScheduler(ctx context.Context, kubeClient client.Client, nodeTemplates [
 		}
 	}
 
-	for provisioner := range instanceTypes {
```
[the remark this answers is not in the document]
jonathan-innis: Done after the scheduling so we can use the nodeTemplate requirements to order the instances with offerings in mind

pkg/cloudprovider/types.go
```diff
@@ -80,14 +80,13 @@ type InstanceType interface {
...
-	Price() float64
 }
 
 // An Offering describes where an InstanceType is available to be used, with the expectation that its properties
 // may be tightly coupled (e.g. the availability of an instance type in some zone is scoped to a capacity type)
 type Offering struct {
 	CapacityType string
 	Zone         string
+	Price        float64
+	Available    bool
```
ellistarn, line 91: Previously we used lack of existence as a signal for availability. Is there a reason this doesn't work anymore?

pkg/controllers/provisioning/scheduling/node.go
```diff
@@ -20,6 +20,8 @@ import (
 	"strings"
 	"sync/atomic"
 
+	cputils "github.com/aws/karpenter/pkg/utils/cloudprovider"
```
ellistarn, line 23: sugg: Aim to name your packages such that the name naturally makes sense in context. This helps minimize the number of decisions that developers need to make, and allows you to import the package unaliased in almost all situations. In some cases, there will be name conflicts, which can be resolved by simply moving up the path. e.g. `import kubeclient"github.com/kubernetes/pkg/kube/client"`. If your folder structure is coherent, then this should read naturally(ish) as well.

pkg/cloudprovider/types.go
jonathan-innis: The thought was that we could send back all the offerings that have ever existed and keep the pricing information in the offerings this way to keep it decoupled from the pricing provider. This was needed so that we could get a close price to use to calculate the savings on the node during consolidation.

In my mind there's a couple of options:
1. We keep track of the last price of the offering through this model, keeping availability in the offering and filtering out the offerings that are available when we needs to
2. Return a "close enough" price when we are trying to get the offering price in the consolidation loop (by getting either the exact offering or one within the same capacity type if that exact offering is no longer available (due to ICE or otherwise)), else return an error. This means we could leave the `Offerings()` call the same and we wouldn't have to change a ton of logic in the pricing provider.

Personally, I'm more in favor of the second since I think it keeps our `Offerings()` call much simpler but I know that @bwagner5 was thinking there may be other use-cases for having details around whether an offering _ever_ existed as part of the offerings logic.

pkg/controllers/provisioning/scheduling/node.go
jonathan-innis: So to clarify you are thinking a more "sensical" name here would be `utilscloudprovider`

ellistarn: Given the naming conflict, that's probably the fastest way to resolve it. I don't want to be too prescriptive about the solution and I'm happy to merge as is, just sharing principles for thinking about this.

You'll notice that you're aliasing this utility in every occurrence in the code base. This is because each place you need it, you happen to also need the primary cloudprovider package. Is this a coincidence? No -- these helpers are deeply coupled to our cloudprovider `Offering` and `InstanceType` structs. By definition, if you need the utils, you also need the types. What's the point of utility classes? When should we make them? What are the alternatives? 

Utility classes are a thing that come up regularly in lists of software antipatterns, but they continually emerge in almost every project. Why is this? Often, we can encapsulate logic is member functions of structs, which allows for straightforward code reuse, but sometimes we don't own the struct or it's not natural to act on the struct so we need *something else* to own the logic. 

Often these static helpers simply can be encapsulated in the package that owns the structs. In this case, you could expose all of these helpers as part of the original cloudprovider package, which resolves your conceptual conflict. Another way to do this is to make an alias type `type Offering []Offerings` and hang these methods off of it.

Sometimes we're not so lucky as to have such a natural home for static methods, either because of large dependency closures, circular dependencies, or we don't own the home. If the code is only duplicated once (or potentially even twice), it might make sense to simply encapsulate the method in the same context where it's needed (e.g. `HasZone` is a good candidate to live inside `leastExpensiveInstanceWithZone`, it's only caller). 

If we find that code is used in multiple contexts, risks bugs from duplicating error prone lines, and doesn't have a natural home, it's time to build a home (e.g. utility package) for it.

pkg/controllers/consolidation/helpers.go
```diff
@@ -50,10 +55,11 @@ func GetPodEvictionCost(ctx context.Context, p *v1.Pod) float64 {
 	return clamp(-10.0, cost, 10.0)
 }
 
-func filterByPrice(options []cloudprovider.InstanceType, price float64, inclusive bool) []cloudprovider.InstanceType {
+func filterByPrice(options []cloudprovider.InstanceType, reqs scheduling.Requirements, price float64) []cloudprovider.InstanceType {
 	var result []cloudprovider.InstanceType
 	for _, it := range options {
-		if (it.Price() < price) || (inclusive && it.Price() == price) {
+		cheapestOffering := cputils.CheapestOfferingWithReqs(cputils.AvailableOfferings(it), reqs)
```
tzneal, line 61: I don't think this is correct as we aren't guaranteed to launch the cheapest offering here, but the problem is pretty rare since we don't do spot->spot replacement.  However I have seen a spot price be equal to the on-demand price which makes this scenario possible

Suppose you have an instance that is an On-Demand at 2.00 and offerings:
- Spot/Zone-A  1.00
- Spot/Zone-B  2.00

this code would consider the instance type as a valid replacement as the offering Spot/Zone-A is cheaper.  However since there's no requirement that gets imposed on the node request, CreateFleet could technically still launch a Spot/Zone-B at the same price which won't save money but is a downgrade as the node went from on-demand to spot at the same price.

I think this bug exists in the original launch code as well as we would prefer to launch the spot node and should prefer to launch the on-demand since it costs the same. It also occurs where we have multiple instance types that can suffice and we launch a more expensive spot node than the cheapest on-demand node would have been.  I've been meaning to fix that for a while.

pkg/controllers/provisioning/scheduling/node.go
jonathan-innis: Makes sense. Good feedback, having this rationale is helpful in making the decision. For many of these methods, the thought is also not just if they are used once or twice now but if they could potentially be useful functions to capture for the future. I admit that that means a bit of guesswork at what is and isn't a useful function for future use, but the thought was that some of these helper functions could be potentially used later in other parts of the codebase or by others implementing their version of `cloudprovider`

pkg/cloudprovider/types.go
ellistarn: > This was needed so that we could get a close price to use to calculate the savings on the node during consolidation.

Ah -- I see. The problem is that we need to know the existing price of a node when performing a consolidation action. This actually sounds like a bug, since the price returned by the cloud provider can be different. 

What if we made price a well known label? This would enable us to consolidate based on an actual known value of the node, rather than guessing by trying to match with a result from Instance Types. IIUC from your statements above, this lets us remove the concept of availability from the API. 

This is extremely interesting from a selector perspective using the new GT/LT operators, since pods/provisioners could request capacity as long as it's under a certain price 😮.

> @bwagner5 was thinking there may be other use-cases for having details around whether an offering ever existed as part of the offerings logic.

I'm hesitant to build features/api surface before we know (or can at least guess) real use cases for them. There's some discussion on this in https://karpenter.sh/docs/contributing/design-guide/

pkg/controllers/consolidation/helpers.go
jonathan-innis: Sure, but I'm not sure the alternative is much better where we average out the offerings price or we assume the worst-case and take the most expensive spot offering.

Perhaps we filter out the offerings here too so that we only end up passing through the offerings that are lower in price than the current offering?

jonathan-innis: Also, the same problem exists in scheduling; however, on some level, we have to take some heuristic to order the instance types and determine what is a "cheaper" instance type. This is; however, less of a problem with initial scheduling since we are making no claim about us launching a node making the nodeset cheaper overall

pkg/controllers/provisioning/scheduling/node.go
ellistarn: > could potentially be useful functions to capture for the future.

I'm personally very skeptical of premature abstraction. I usually wait until n=3 (occasionally n=2) before abstracting. Instead, I like to treat the code as a living document that best represents the solution at present, and continually refine things as new use cases emerge.

jonathan-innis: Good feedback. These are good words to live by :)

pkg/cloudprovider/types.go
FernandoMiguel: > What if we made price a well known label? 
Isn't the price in the fleet request?
Can't we access that one that is linked to the ec2 instance metadata?

jonathan-innis: Yes, this would let us remove `.Availability` from the API. @bwagner5 and I discussed adding this label. It sounds like this is something we would have to add post-node creation since we won't know the instance type that will be created until after the node is actually joined and brought up. 

> This actually sounds like a bug, since the price returned by the cloud provider can be different.

I'm not quite getting how this would be a bug since the current price of the offering should be someone is currently getting charged (save special enterprise pricing). The issue I see with applying a node label is that pricing is variable and the pricing that is in the node label that is first used upon node creation may vary greatly over a long period of time (particularly for spot pricing).

bwagner5: The price can also change per hour for Spot, so we'd need to reconcile the nodes with price.

pkg/controllers/consolidation/helpers.go
tzneal: I think assuming the max price is safer, but it wouldn't do an on-demand to spot node replacement.  It seems like we really need to add a requirement to the node that forces the cheapest offering to be selected, but I'm a bit worried of the implications of that.  We're removing some flexibility from CreateFleet due to up to 12-hour old pricing information.

tzneal: For some more information, the scenario I want to avoid is a second cloud provider comes along and consolidation gets in a loop repeatedly trying to replace a node with cheaper instance which continually fails since the cheaper instance doesn't launch. I don't think it will happen for us currently due to spot to spot replacement avoidance and on-demand prices be >= spot prices.

pkg/cloudprovider/types.go
ellistarn: It could be an interesting example of the first provider specific reconciler.

pkg/controllers/consolidation/helpers.go
jonathan-innis: It seems like filtering the offerings may solve the issue here since we won't generate any launch template overrides for capacity type/zone that don't appear in the offerings for an instance type, meaning that if we filter them away from the instance types given to the nodeTemplate, we will not be producing instances that wouldn't satisfy our pricing requirement.

pkg/cloudprovider/types.go
jonathan-innis: Resolved this discussion, the plan is to keep the `.Available` as a portion of the offerings and return all `Offerings` in the interface and filter out availability as a first-class helper function

pkg/cloudprovider/aws/instancetypes.go
```diff
@@ -92,28 +91,38 @@ func (p *InstanceTypeProvider) Get(ctx context.Context, provider *v1alpha1.AWS,
...
+			_, isUnavailable := p.unavailableOfferings.Get(UnavailableOfferingsCacheKey(*instanceType.InstanceType, zone, capacityType))
+			var price float64
+			var err error
+			switch capacityType {
+			case ec2.UsageClassTypeSpot:
+				price, err = p.pricingProvider.SpotPrice(*instanceType.InstanceType, zone)
+			case ec2.UsageClassTypeOnDemand:
+				price, err = p.pricingProvider.OnDemandPrice(*instanceType.InstanceType)
+			default:
+				logging.FromContext(ctx).Errorf("Received unknown capacity type %s for instance type %s", capacityType, *instanceType.InstanceType)
```
ellistarn, line 116: We used to have this comment:
`// don't warn as this can occur extremely often` in the error handling. Is this potentially really noisy now?

pkg/cloudprovider/types.go
```diff
@@ -75,14 +78,32 @@ type InstanceType interface {
...
```
ellistarn, line 89: Make sure you add a comment why available is included in the API.

pkg/cloudprovider/types.go
```diff
@@ -75,14 +78,32 @@ type InstanceType interface {
...
+// and returns the offerings marked as available
+func AvailableOfferings(it InstanceType) []Offering {
+	return lo.Filter(it.Offerings(), func(o Offering, _ int) bool {
+		return o.Available
+	})
+}
+
+// GetOffering gets the offering from passed offerings that matches the
+// passed zone and capacity type
+func GetOffering(ofs []Offering, ct, zone string) (Offering, error) {
```
ellistarn, line 102: nit: maybe slightly more intuitive to hang this off of instanceType, instead of []Offering.

ellistarn: nit: I also noticed that you're never actually consuming the error as an error, you're just consuming it as a boolean. Consider returning `Offering, bool`, which also happens to be an identical interface to `lo.Find`, which can simplify your code/concepts.

pkg/controllers/consolidation/controller.go
```diff
@@ -104,6 +104,7 @@ func (c *Controller) run(ctx context.Context) {
 				if err != nil {
 					logging.FromContext(ctx).Errorf("consolidating cluster, %s", err)
 				} else if result == ProcessResultNothingToDo {
+					logging.FromContext(ctx).Debugf("Attepted consolidation but nothing to consolidate")
```
ellistarn, line 107: What impact does this have on the logs long term?
```suggestion
					logging.FromContext(ctx).Debugf("Attempted consolidation but nothing to consolidate")
```

pkg/controllers/consolidation/controller.go
```diff
@@ -116,6 +117,8 @@ func (c *Controller) run(ctx context.Context) {
 type candidateNode struct {
 	*v1.Node
 	instanceType   cloudprovider.InstanceType
+	capacityType   string
```
ellistarn, line 120: nit: Consider just relying on node.Labels instead of hoisting them up onto this struct.

pkg/controllers/consolidation/controller.go
```diff
@@ -192,6 +195,18 @@ func (c *Controller) candidateNodes(ctx context.Context) ([]candidateNode, error
 			return true
 		}
 
+		// skip any nodes that we can't determine the capacity type or the topology zone for
+		ct, ok := n.Node.Labels[v1alpha5.LabelCapacityType]
+		if !ok {
+			logging.FromContext(ctx).Errorf("Getting node capacity type, %v", err)
```
ellistarn, line 201: This has the potential to be quite noisy in the logs. Consider a set of nodes w/o capacity type. It will spew error logs on every consolidation loop. There are lots of other cases where we skip the node, is there a reason to log here, but not there?

pkg/controllers/consolidation/controller.go
```diff
@@ -456,7 +474,10 @@ func (c *Controller) nodeConsolidationOptionReplaceOrDelete(ctx context.Context,
 			schedulableCount += len(inflight.Pods)
 		}
 		if len(node.pods) == schedulableCount {
-			savings := node.instanceType.Price()
+			savings, err := getNodePrice(node)
+			if err != nil {
+				return consolidationAction{result: consolidateResultUnknown}, fmt.Errorf("getting offering price from candidate node, %w", err)
```
ellistarn, line 479: This deviates from the standard go error return pattern. Typically, the `foo, err` pattern is an OR semantic. Either you get a value OR you get an error. The pattern is typically to return the uninitialized type value if there's an error. For example: nil, 0, "", or in this case, consolidationAction{}.

pkg/controllers/consolidation/controller.go
```diff
@@ -456,7 +474,10 @@ func (c *Controller) nodeConsolidationOptionReplaceOrDelete(ctx context.Context,
 			schedulableCount += len(inflight.Pods)
 		}
 		if len(node.pods) == schedulableCount {
-			savings := node.instanceType.Price()
+			savings, err := getNodePrice(node)
```
ellistarn, line 477: Reading through the code, I'm struggling to see where savings is even used.

pkg/controllers/consolidation/helpers.go
```diff
@@ -67,3 +76,42 @@ func disruptionCost(ctx context.Context, pods []*v1.Pod) float64 {
 	}
 	return cost
 }
+
+// getNodePrice gets the last known node price for the candidate node
+// This price is used to determine the savings that can be achieved through consolidation
+func getNodePrice(node candidateNode) (float64, error) {
```
ellistarn, line 82: How is this method different than just calling `cloudprovider.GetOffering(node.instanceType.Offerings(), node.capacityType, node.zone)` directly?

pkg/controllers/consolidation/controller.go
jonathan-innis: Out of curiosity what's the concern with having this data in the struct since we need to have this data as part of the consolidation savings calculation and we are relying on things like capacity type in other places in the consolidation code like for checking spot -> spot consolidation

pkg/controllers/consolidation/controller.go
jonathan-innis: Actually a good question. @tzneal what was the intention of this `.savings` value? It looks like this value is currently not getting used anywhere in the code

pkg/controllers/consolidation/controller.go
ellistarn: Mostly just asking what an idle karpenter looks like from a logging perspective. Does this trigger on every loop? How frequently?

pkg/controllers/consolidation/controller.go
ellistarn: It's completely optional. Generally, I try to deduplicate as many concepts as possible, and just rely on a few well known concepts. Nodes naturally have a set of properties expressed in their labels, and if I was writing code in some function and had a node, I wouldn't hesitate to use those labels unless I couldn't for some reason. Since candidateNode is itself a node, you're more or less guaranteed to always have those properties available wherever you need them.

pkg/controllers/consolidation/controller.go
ellistarn: Maybe we can write up some guidelines together. This comes up fairly often.

pkg/controllers/consolidation/controller.go
jonathan-innis: Yeah, I can imagine this would trigger too frequently. Let's discuss the logging constructs we think make sense

pkg/controllers/consolidation/controller.go
jonathan-innis: This is consistent with how it was written in other places throughout the code when an error is returned. At this point, it looks like this error is being logged at the level above but then the action is what is being used to determine what to do with the consolidation reconciliation loop so it looks like the error isn't passed all the way up

ellistarn: huh, ok.

ellistarn:
Nice work!

pkg/controllers/consolidation/controller.go
tzneal: At one point you could configure it to consolidate based on maximizing savings or minimizing cluster disruption. It's fine to remove it.

pkg/controllers/consolidation/controller.go
tzneal: In this case it's the same thing, consolidateResultUnknown = 0

pkg/controllers/consolidation/controller.go
tzneal: Minimum of every 5 mins, I'm working on a PR to reduce the log clutter from the repeated discovering instance types, subnets, pricing, etc.

pkg/cloudprovider/aws/instancetypes.go
jonathan-innis: Shouldn't be noisy since we are getting the capacity types from the UsageClasses above from AWS. This default case was just there as an extra added measure in case we happen to get a capacity type that we don't know if somehow AWS were to change their SDK

pkg/cloudprovider/types.go
```diff
@@ -75,14 +77,31 @@ type InstanceType interface {
...
-	Price() float64
 }
 
 // An Offering describes where an InstanceType is available to be used, with the expectation that its properties
 // may be tightly coupled (e.g. the availability of an instance type in some zone is scoped to a capacity type)
 type Offering struct {
 	CapacityType string
 	Zone         string
+	Price        float64
+	// Available is added so that Offerings() can return all offerings that have ever existing for an instance type
```
bwagner5, line 88:
```suggestion
	// Available is added so that Offerings() can return all offerings that have ever existed for an instance type
```

pkg/cloudprovider/types.go
```diff
@@ -75,14 +77,31 @@ type InstanceType interface {
...
+
+// AvailableOfferings filters the offerings on the passed instance type
+// and returns the offerings marked as available
+func AvailableOfferings(it InstanceType) []Offering {
+	return lo.Filter(it.Offerings(), func(o Offering, _ int) bool {
+		return o.Available
+	})
+}
+
+// GetOffering gets the offering from passed offerings that matches the
```
bwagner5, line 101:
```suggestion
// GetOffering gets the offering from passed instance type that matches the
```

bwagner5 (changes requested):

bwagner5 (dismissed):
lgtm! Let's wait for the release and probably merge tomorrow

tzneal:
LGTM as well, can we get a consolidation test in a follow-on PR that does an OD to Spot replace consolidation?

ellistarn (approved):

PR #2307, opened 2022-08-17
Weighted provisioners

### LINKED ISSUE #2024: weight provisioners?
**What karpenter features are relevant?**
provisioners

**How should the docs be improved?**
We currently deploy a single provisioner with both X64 and ARMv8 archs
```
kind: Provisioner
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
...
  generation: 1
  name: default-dev-ue1-008-flying-possum
  resourceVersion: "4109"
spec:
  limits:
    resources:
      cpu: 1k
  provider:
    amiFamily: AL2
    blockDeviceMappings:
    - deviceName: /dev/xvda
      ebs:
        deleteOnTermination: true
        iops: 10000
        throughput: 125
        volumeSize: 100Gi
        volumeType: gp3
    securityGroupSelector:
      aws-ids: sg-xxxx
    subnetSelector:
      aws-ids: subnet-xxx
    tags:
      Environment: dev
  requirements:
  - key: topology.kubernetes.io/zone
    operator: In
    values:
    - us-east-1a
    - us-east-1b
    - us-east-1c
    - us-east-1d
    - us-east-1f
    operator: In
    values:
    - spot
    - on-demand
    operator: In
    values:
    - c6a.large
    - c6g.large
    - c6i.large
  - key: kubernetes.io/arch
    operator: In
    values:
    - amd64
    - arm64
  ttlSecondsAfterEmpty: 120
  ttlSecondsUntilExpired: 2592000
status:
  resources:
    attachable-volumes-aws-ebs: "39"
    cpu: "2"
    ephemeral-storage: 113224684Ki
    memory: 3915268Ki
    pods: "29"
```

Our question is if there is an option to have two provisioners, one per arch but with different weights in each.
We want to offer both arch but not all workloads work on ARM, but we can't currently modify the Deployment for all of them to pin an arch.
If we could influence Karpenter to pick x64 over ARMv8 (it tends to pick graviton cause it is cheaper, I believe), and then specific workloads that pick arm arch would ask for those specific nodes.

Thanks

### Community Note

jonathan-innis:
Fixes #2024 

**Description**

- Allows for the ordering of Provisioners based off a `.spec.weight` which will be used during scheduling

* `make test`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```


debugger24:
If there are 2 provisioner

Provisioner 1

- weight: 100
- instance types: large instances

Provisioner 2
- weight: 50
- instance types: medium + large instances

----

Now, if there is requirement to add a single pod for which even small instance size is enough, will it still create resource using provisioner 1 config because of higher weight? or it will use provisoner 2 config since karpenter also tries to create least size instance when possible.

jonathan-innis:
> Now, if there is requirement to add a single pod for which even small instance size is enough, will it still create resource using provisioner 1 config because of higher weight? or it will use provisoner 2 config since karpenter also tries to create least size instance when possible.

In this case, it would choose provisioner 1 regardless if the cost would be cheaper on provisioner 2. As long as the pod can be scheduled to that first provisioner, Karpenter will launch the node with the highest priority. 

The way to solve the issue you have described above is to make sure that you have enough flexibility in your instance types in your provisioners such that the scheduler is not forced into a situation like this

pkg/controllers/provisioning/suite_test.go
```diff
@@ -748,4 +748,33 @@ var _ = Describe("Multiple Provisioners", func() {
 		node := ExpectScheduled(ctx, env.Client, pod)
 		Expect(node.Labels[v1alpha5.ProvisionerNameLabelKey]).ToNot(Equal(provisioner.Name))
 	})
+	Describe("Weighted Provisioners", func() {
```
ellistarn, line 751: Technically I think this should be a Context() block. Describe is typically used at the top level. Ginkgo is a little messy/undefined in this area =/

website/content/en/preview/provisioner.md
```diff
@@ -184,6 +189,12 @@ Karpenter prioritizes Spot offerings if the provisioner allows Spot and on-deman
 
 Karpenter also allows `karpenter.sh/capacity-type` to be used as a topology key for enforcing topology-spread.
 
+## spec.weight
+
+Karpenter allows you to describe provisioner preferences through a `weight` mechanism similar to how weight is described with [pod and node affinities](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity). 
+
+For more information on weighting provisioners, see the [Scheduling Details](../tasks/scheduling/).
```
ellistarn, line 196: Can you link directly to #weighting-provisioners?

website/content/en/preview/tasks/scheduling.md
```diff
@@ -366,6 +366,91 @@ The EBS CSI driver uses `topology.ebs.csi.aws.com/zone` instead of the standard
...
+  requirements:
+  - key: karpenter.sh/capacity-type
+    operator: In
+    values: ["spot", "on-demand"]
+  - key: kubernetes.io/arch
+    operator: In
+    values: ["arm64"]
+```
+
+_Note: Based on the way that Karpenter does pod batching and bin packing, it is not guaranteed that Karpenter will always choose the highest priority provisioner given specific requirements. For example, it is possible that a pod that has a node selector on a an `arm64` architecture may be provisioned first and all other workload pods from that batch may also schedule on that node._
```
ellistarn, line 452:
```suggestion
_Note: Based on the way that Karpenter does pod batching and bin packing, it is not guaranteed that Karpenter will always choose the highest priority provisioner given specific requirements. For example, it is possible that a pod that has a node selector on a an `arm64` architecture may be provisioned first and all other workload pods from that batch may also schedule on that node. The behavior may also occur if existing capacity is available, as the kube-scheduler will schedule the pods instead of allowing Karpenter to provision a new node.   _
```

ellistarn (dismissed):
LGTM. A few nits.

ellistarn (approved):

Adityashar:
Hey @jonathan-innis, 

I have a question regarding the usability of these weighted provisioners.  

Suppose I have two provisioners -  

1. provisioner-spot with weight: 40
2. provisioner-on-demand with weight: 60

then as per the [code](https://github.com/jonathan-innis/karpenter/blob/608d9a2e0de775e6648e9ea4efaf9d77cd4f0a75/pkg/apis/provisioning/v1alpha5/provisioner.go#L151), since our provisioners are getting sorted with respect to their weights, there may arise a situation where only my `provisioner-on-demand` is being used for provisioning nodes and the other provisioner becomes stale. Is this how we are intending weighted provisioners to work?

Also, could we work upon changing this logic to something like a percentage based usage of provisioners? Eg. If my `provisioner-on-demand` has a weight equal to 60, then it means that this provisioner will be used for fulfilling 60% of my node provisioning requests and similarly the other provisioner responds to the remaining 40% of the requests.

PR #2326, opened 2022-08-19
feat: Adds ability to get most recent ami

andrewhibbert:
Fixes # 

**Description**

Adds support for picking the most recent AMI found with filters (aws-ids, name or tags).

By default it will create launch templates for all found AMIs. If `mostRecent` is true then it will just create a launch template for the most recent AMI. This ensures users who use their own custom built AMIs can use the most latest ami without having to use launch templates or hardcode the AMI in aws-ids.

* Unit tests
* Running make apply in cluster and ensuring that when `mostRecent` is true the launch template created is with the latest AMI found

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Existing filters in `AWSNodeTemplate.spec.amiSelector` will need moving to `AWSNodeTemplate.spec.amiSelector.amiFilters`. Optionally if you want just the most recent you can set `AWSNodeTemplate.spec.amiSelector.mostRecent` to true
```


pkg/apis/awsnodetemplate/v1alpha1/awsnodetemplate.go
```diff
@@ -29,9 +29,15 @@ type AWSNodeTemplateSpec struct {
...
+	// AMISelector discovers AMIs to be used by filters.
 	// +optional
-	AMISelector map[string]string `json:"amiSelector,omitempty"`
+	AMISelector *AMISelector `json:"amiSelector,omitempty"`
+}
+
+// AMISelector discovers AMIs to be used by filters
+type AMISelector struct {
+	AMIFilters map[string]string `json:"amiFilters,omitempty"`
+	MostRecent *bool             `json:"mostRecent,omitempty"`
```
ellistarn, line 40: This is a backwards incompatible change, and will require a v1alpha6 bump. See: #1327. One of our tenets in https://karpenter.sh/docs/contributing/design-guide/ is that we should only expose parameters/functionality as necessary. Does it make sense to just default to most recent? This seems like a reasonable experience to me, and doesn't require API changes.

ellistarn:
Great feature!

pkg/apis/awsnodetemplate/v1alpha1/awsnodetemplate.go
andrewhibbert: Okay. Yes I think it should default to the most recently created AMI. I'll push up another change to set `AMISelector` back to `map[string]string` with the logic to use the most recently created AMI.

PR #2330, opened 2022-08-20
feat: AMISelector uses most recently built AMI

andrewhibbert:
Fixes # 

**Description**

In this PR `AMISelector` will create a launch template with the most recently built AMI found matching using tags and/or name. If `aws-ids` key is used it will create a launch template for each matching AMI.

Relates to https://github.com/aws/karpenter/pull/2326#discussion_r950551896

* Unit tests
* Running make apply in cluster and ensuring that the most recently built AMI is chosen when using tags and/or name

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/aws/amifamily/ami.go
+++ b/pkg/cloudprovider/aws/amifamily/ami.go
@@ -17,6 +17,8 @@ package amifamily
 import (
 	"context"
 	"fmt"
+	"sort"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -109,6 +111,7 @@ func (p *AMIProvider) getAMIRequirements(ctx context.Context, providerRef *v1alp
 }
 
 func (p *AMIProvider) selectAMIs(ctx context.Context, amiSelector map[string]string) (map[string]scheduling.Requirements, error) {
+	var amis []*ec2.Image
 	ec2AMIs, err := p.fetchAMIsFromEC2(ctx, amiSelector)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/aws/amifamily/ami.go
```diff
@@ -109,16 +111,24 @@ func (p *AMIProvider) getAMIRequirements(ctx context.Context, providerRef *v1alp
...
 	if err != nil {
 		return nil, err
 	}
 	if len(ec2AMIs) == 0 {
 		return nil, fmt.Errorf("no amis exist given constraints")
 	}
+	if _, exists := amiSelector["aws-ids"]; exists {
+		amis = ec2AMIs
+	} else {
+		mostRecentAMI := getMostRecentAMI(ec2AMIs)
```
bwagner5, line 125: This would limit the AMIs returned even if they have separate requirements. The most prevalent would be if the tag selector return both arm64 and x86 AMIs. Both would need to be returned from this func.  I think it makes sense to move this prioritization into the `Get` func above when the amiID is being added to the mapping of AMI ID -> InstanceTypes:

```
		for _, instanceType := range nodeRequest.InstanceTypeOptions {
			for amiID, requirements := range amiRequirements {
				if err := instanceType.Requirements().Compatible(requirements); err == nil {
					amiIDs[amiID] = append(amiIDs[amiID], instanceType)
				}
			}
		}
```

You'd need to modify the selectAMIs func to return more info than just the IDs to be able to do a creation timestamp sort within `Get`

andrewhibbert: I have made some changes so that the most recent ami is chosen after it is found to be compatible

njtran:
Hey @andrewhibbert, it looks like @bwagner5 had some comments for you to implement and it looks like you have some conflicting files. Let me know if you can fix those, and I can help review this and get it merged in. In the meantime, I'll pull your code down and start making changes on my own branch.

njtran:
Hey @andrewhibbert, thanks for the contribution! I pulled your code down and made some changes [here](https://github.com/aws/karpenter/pull/2653). Since we haven't heard from you in a bit, I'll close this out. Feel free to leave a review.

PR #2341, opened 2022-08-23
fix: improve health checks

tzneal:
**Description**
Acquire and release some locks during health checks to identify deadlock.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -113,9 +113,11 @@ func Initialize(injectCloudProvider func(context.Context, cloudprovider.Options)
 		utilruntime.Must(registerPprof(manager))
 	}
 
-	cloudProvider := cloudprovidermetrics.Decorate(
-		injectCloudProvider(ctx, cloudprovider.Options{ClientSet: clientSet, KubeClient: manager.GetClient(), StartAsync: manager.Elected()}),
-	)
+	cloudProvider := injectCloudProvider(ctx, cloudprovider.Options{ClientSet: clientSet, KubeClient: manager.GetClient(), StartAsync: manager.Elected()})
+	if hp, ok := cloudProvider.(HealthCheck); ok {
+		utilruntime.Must(manager.AddHealthzCheck("cloud-provider", hp.LivenessProbe))
+	}
+	cloudProvider = cloudprovidermetrics.Decorate(cloudProvider)
 
 	cfg, err := config.New(ctx, clientSet, cmw)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -512,3 +513,10 @@ func (c *Cluster) Synchronized(ctx context.Context) error {
 func (c *Cluster) recordConsolidationChange() {
 	atomic.StoreInt64(&c.consolidationState, c.clock.Now().UnixMilli())
 }
+
+func (c *Cluster) LivenessProbe(req *http.Request) error {
+	c.mu.RLock()
```
bwagner5, line 518: why not take a write lock here?

bwagner5 (approved):
nicee!

pkg/controllers/controllers.go
```diff
@@ -113,9 +113,11 @@ func Initialize(injectCloudProvider func(context.Context, cloudprovider.Options)
 		utilruntime.Must(registerPprof(manager))
 	}
 
-	cloudProvider := cloudprovidermetrics.Decorate(
-		injectCloudProvider(ctx, cloudprovider.Options{ClientSet: clientSet, KubeClient: manager.GetClient(), StartAsync: manager.Elected()}),
-	)
+	cloudProvider := injectCloudProvider(ctx, cloudprovider.Options{ClientSet: clientSet, KubeClient: manager.GetClient(), StartAsync: manager.Elected()})
+	if hp, ok := cloudProvider.(HealthCheck); ok {
```
ellistarn, line 117: Another option is to write an AWS specific controller that checks the liveness of the cloudprovider. I've been wanting to build a simple mechanism to wire additional controllers through Initialize()

pkg/controllers/state/cluster.go
tzneal: Yeah, that's better. I did the read locks at first to be less intrusive, but I agree that write locks is better at finding the problem quickly.

pkg/controllers/controllers.go
tzneal: I was hoping to make these health checks better when I started this, but couldn't really come up with much apart from preventing deadlock.  If we can come up with any scenario where restarting is the best corrective action, we can just hook it up to a health check and return it as an error which will provide a bit more info on why we restarted.

ellistarn (dismissed):
Let's hold for the release. But great change!

bwagner5 (approved):
lgtm!

PR #2342, opened 2022-08-23
fix: dynamically set go 1.19 runtime soft memory limit to better avoid OOM conditions

tzneal:
**Description**

This informs the go runtime of the amount of memory available to the
container.  It allows the runtime to sometimes delay GC when there is
plenty of headroom and to run GC more often to avoid exceeding the
container memory limit.

* 13k pod / 400 node scaleup

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/charts/karpenter/templates/deployment.yaml
+++ b/charts/karpenter/templates/deployment.yaml
@@ -83,6 +83,12 @@ spec:
             - name: AWS_DEFAULT_INSTANCE_PROFILE
               value: {{ .Values.aws.defaultInstanceProfile }}
           {{- end }}
+            - name: MEMORY_LIMIT
+              valueFrom:
+                resourceFieldRef:
+                  containerName: controller
+                  divisor: "0"
+                  resource: limits.memory
           {{- with .Values.controller.env }}
             {{- toYaml . | nindent 12 }}
           {{- end }}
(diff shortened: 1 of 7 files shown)


tzneal:
This requires go 1.19, but in testing allowed me to do a 13k pod/400 node scale-up without an OOM.

tzneal:
This follows the guidance in the gc guide at  https://tip.golang.org/doc/gc-guide 


> Do take advantage of the memory limit when the execution environment of your Go program is entirely within your control, and the Go program is the only program with access to some set of resources (i.e. some kind of memory reservation, like a container memory limit).
>
> A good example is the deployment of a web service into containers with a fixed amount of available memory.
>
> In this case, a good rule of thumb is to leave an additional 5-10% of headroom to account for memory sources the Go runtime is unaware of.

charts/karpenter/templates/deployment.yaml
```diff
@@ -83,6 +83,12 @@ spec:
             - name: AWS_DEFAULT_INSTANCE_PROFILE
               value: {{ .Values.aws.defaultInstanceProfile }}
           {{- end }}
+            - name: MEMORY_LIMIT
+              valueFrom:
+                resourceFieldRef:
+                  containerName: controller
+                  divisor: "0"
+                  resource: limits.memory
           {{- with .Values.controller.env }}
```
bwagner5, line 92: Would setting the env var here override the above?

bwagner5 (approved):
lgtm

tzneal:
@bwagner5 Will this break any E2E tests that assume go 1.18? I looked around and didn't see anything that referenced 1.18 specifically/

bwagner5:
> @bwagner5 Will this break any E2E tests that assume go 1.18? I looked around and didn't see anything that referenced 1.18 specifically/

I don't believe so.

PR #2361, opened 2022-08-25
docs: do not tag managed nodegroup

mcanevet:
Fixes # 

**Description**
IMHO, this should not be required because Karpenter creates its own launch template.

* I actually do not set this tag and it works.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Do not tag initial managed nodegroup with Karpenter discovery tag in reference documentation for Terraform.
```


bwagner5:
I agree that this doesn't seem necessary. It may have been tagged just for discovery purposes.

bwagner5:
tActually this may actually be important

PR #2363, opened 2022-08-25
perf: Provisioner changes trigger consolidation to be re-considered

jonathan-innis:
Fixes # 

**Description**

- Allows consolidation to automatically be triggered if there is a change to the provisioner spec
- This allows immediate reconciliation when consolidation is turned on or when there is a change to provisioner requirements

* Manual validation of E2E
* `make test`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/pkg/controllers/state/provisioner.go
+++ b/pkg/controllers/state/provisioner.go
@@ -17,11 +17,10 @@ package state
 import (
 	"context"
 
-	"k8s.io/apimachinery/pkg/api/errors"
-	"knative.dev/pkg/logging"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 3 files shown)


pkg/controllers/state/provisioner.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// If the provisioner is deleted, no reason to re-consider consolidation at this point
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, stored); err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+	// Something changed in the provisioner so we should re-consider consolidation
+	c.cluster.recordConsolidationChange()
```
tzneal, line 59: Have you added a log to see how often this is called?  There's a counter controller that updates provisioner status which is triggered off node updates, which I think occur once a minute.

jonathan-innis: Provisioner status won't trigger this reconciliation loop, only `.spec` updates (see the `GenerationChangedPredicate` below)

tzneal: Ah, nice I missed that!

tzneal (approved):
lgtm

tzneal (approved):
lgtm

pkg/controllers/state/provisioner.go
```diff
@@ -0,0 +1,67 @@
...
+	return reconcile.Result{}, nil
+}
+
+func (c *ProvisionerController) Register(ctx context.Context, m manager.Manager) error {
+	return controllerruntime.
+		NewControllerManagedBy(m).
+		Named(provisionerControllerName).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
```
njtran, line 56: Do we need max reconciles to be 10 here or are you just following pattern from other controllers?

pkg/controllers/state/provisioner.go
```diff
@@ -0,0 +1,67 @@
...
+		WithEventFilter(predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
```
njtran, line 58: You could consider using the vended function for this here

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/predicate/predicate.go#L156-L172

njtran:
Looks like i was a little late :(

pkg/controllers/state/provisioner.go
jonathan-innis: Following the same pattern, not sure why we would have an issue with concurrent reconciles here.

PR #2377, opened 2022-08-26
feat: Update poddisruptionbudget name

arunanandhan:
Fixes # 

**Description**
Update the name of the pod disruption budget to be same as other resources. 

This PR doesn't have any major changes. Testing done in locally to make sure there is no errors.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


bwagner5:
This has the potential to break current users if they were relying on the configurable PDB name for some reason.  We could make this change later, but not until we do a breaking release. There are some upcoming helm changes that we were planning on making for specifying images and tags. We could revisit later.

PR #2384, opened 2022-08-28
fix: reduce pod nomination event counts

tzneal:
**Description**
These aren't useful during large scale-ups, so drop them instead
of loading the API server.  The case where these are useful is
if we believe a pod will schedule, but kube-scheduler won't.  In
this case, the pod nominations will be repeated indefinitely and
will be available on the pod.

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


tzneal:
I could be convinced there is a better way to do this.

jonathan-innis (approved):
lgtm 🚀

Issue #1131, opened 2022-01-12
Windows support


shirwahersi:
**Tell us about your request**
Windows OS support

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
We have mixed OS workload (Linux,Windows), we would like to managed both Linux/Windows nodes with Karpenter

ellistarn:
This should be straightforward to support as a new value for key `kubernetes.io/os`.

shirwahersi:
Hi all, Is there any plans to support Windows in near future?

halovanic:
I believe @bwagner closed this patch attempt for the understandable difficulties keeping up with other changes around AMI handling in Karpenter.
Having gone back over the patches to test on my EKS cluster with post-0.15, we could aim for very basic support that gets Windows working, just not as its own first-class AMI Family:

1. Support vpc.amazonaws.com/PrivateIPv4Address resources that are needed for Windows-scheduled pods
2. Remove or replace a hardcoded requirement getting added in the instance type calculation that forbids non-Linux OSs

This does require the user to define a launch template in the Windows provisioner, but this is the same requirement Windows EKS users already had with [Self-Managed node groups](https://docs.aws.amazon.com/eks/latest/userguide/launch-windows-workers.html).  I also think a Windows provisioner probably should define the system reserved memory and restrict instance types more explicitly, but the effort is still quite minimal.

This has worked out quite well in my admittedly limited tests: without even really understanding anything specific about Windows, Karpenter seems to have done a more effective job of getting all of my large workloads scheduled.

Is there interest in pursuing these more limited changes?  I can submit reworked patches if there is.

ellistarn:
> define a launch template in the Windows provisioner

We'll likely be deprecating custom launch template support. Any chance you use can `amiSelector` and `userData` to meet the use case instead?

halovanic:
That would require a new user data format and corresponding AMI Family defined. That was already in the original patches, but seemed to be the part that was running into difficulties with a fast moving Karpenter target state.
What I am proposing are to tackle these more stable prerequisites for Windows separately, then we could come back to dedicated Windows AMI support before the launch template support goes away,

PrivateIPv4Address support was in the previous patches (see function `awsPodPrivateIPv4()` in either) and is needed because [Windows supports fewer pods](https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html#windows-support-prerequisites) and EKS's webhook adds this calculation to prevent overscheduling.

[...]

PR #2388, opened 2022-08-29
feat: Private IPv4 address for AWS Windows. (#1131)

halovanic:
Adds private IPv4 address resource to support Windows AWS workloads

`make test` pass with just this change as it does not affect existing behaviors.

I tested end-to-end with mixed Windows and Linux provisioners scoped to OS by also removing the hardcoded Linux scheduling requirement in this file: actually allowing the OS target is not covered by this pull request.

- [ ] Yes, issue opened: # 
- [X ] No

**Release Note**

```release-note
Private Ipv4 address prerequisite toward scheduling Windows workloads on AWS
```


jonathan-innis:
Hi @halovanic, thanks for your contribution! We are excited to take this change but want to be thorough with testing our support for it. I'll be validating the change over the next week or so.

jonathan-innis:
Hi @halovanic. We discussed that before taking this change, we wanted to pull in the Windows scheduling requirement so that we have a way to effectively test this change both with unit tests and E2E. 

I'm opening an issue to add the windows OS as a scheduling requirement to the instanceTypes/offerings. We'll track the progress of that issue first and then have this PR as a fast-follow once that one gets merged.

halovanic:
Thank you @jonathan-innis, allowing alternate OS scheduling before this AWS-specific resource need sounds good.

jonathan-innis:
We've discussed and it sounds like we will unblock by just allowing wind

jonathan-innis:
This is going to be partial support since much of the burden will be on the user to setup their configuration so that everything works

jonathan-innis:
@halovanic I've been testing this/adding E2E testing and there's another snag. Enabling IPAM for Windows only adds the `vpc.amazonaws.com/PrivateIPv4Address` resource to Windows nodes, which breaks Karpenter's initializing logic. 

Essentially, Karpenter expects all extended resources to be registered on an instance for that instance to be considered initialized (initialization affects a lot of things in Karpenter like emptiness, expiration, scheduling, etc.). We don't currently have a way to define resources at a per-OS level (resources currently exist at the instance type level, which you have probably seen). 

What this means, is that we can't differentiate the extended resources that will be registered for an instance type that is a Windows node vs one that is a Linux node. Since Linux nodes, don't have the `vpc.amazonaws.com/PrivateIPv4Address` resource added to them, this causes problems.

I'm re-adding the blocked label since this PR is contingent on discussions on whether we should flatten instance type offerings into instance types or we should allow resources at the offerings level.

halovanic:
@jonathan-innis Can you clarify what problems this causes? When I first started with this patch, I agree it would cause Karpenter to falsely believe that all Linux and Windows nodes had this resource.  That would cause incorrect scheduling behavior if someone were to _manually_ schedule against this resource, such as a workload that asked for PrivateIPv4Address while not specifying OS=windows.  But a PrivateIPv4Address constraint are not applied manually, it really only is added when a workload is already destined for Windows.  From a practical standpoint (if not strictly correct), I never saw any issues scheduling AWS Linux workloads under this configuration.  Unless something has changed to check the presence of this resource after node startup?

jonathan-innis:
> I never saw any issues scheduling AWS Linux workloads under this configuration. Unless something has changed to check the presence of this resource after node startup?

You wouldn't necessarily observe an issue at first glance, but there is an issue with our initialization logic (including a code link [here](https://github.com/aws/karpenter-core/blob/main/pkg/controllers/node/initialization.go#L83)). We check if all extended resources that we expect on the instance are properly registered before considering the instance in a good state.

This is checked all over the Karpenter codebase like in the case of [any deprovisioning logic](https://github.com/aws/karpenter-core/blob/main/pkg/controllers/deprovisioning/helpers.go#L202)(emptiness/consolidation/expiration). This means that once nodes were launched, they wouldn't be spun down, unless they were manually deleted, which fundamentally breaks Karpenter's assumptions.

This is solved by understanding that an InstanceType can support two different OSs that have different resource requirements, but we don't currently allow an instance type to have different extended resources on a per-OS type basis today.

Let me know if any of this doesn't make sense. I'm also happy to hop on a call next week and discuss more.

njtran:
Closing in favor of #3698

Issue #2180, opened 2022-07-25
Limit nodes by amount of pods


devopsjnr:
Not too long ago I asked to add an option to limit nodes by pods limitation. 
I would like to know from which version this feature should be available. Thanks!

spring1843:
Is this about having a maximum number of pods scheduled on a node?

There are some ways to [control pod density](https://karpenter.sh/v0.13.2/tasks/pod-density/) in Karpenter, do any of those help?

spring1843:
Additionally Karpenter doesn't really schedule pods as mentioned in https://github.com/aws/karpenter/issues/1759 . It's kube-scheduler as mentioned [in the other issue](https://github.com/aws/karpenter/issues/1759) that does that job.

devopsjnr:
@spring1843 I think I saw somewhere in the Karpenter documentation that limiting by amount of pods is yet to be supported but will be in future versions. I'll try to look for it again

ellistarn:
We could consider exposing a MaxPods override (https://github.com/kubernetes/kubelet/blob/v0.24.3/config/v1beta1/types.go#L415) in Karpenter's KubeletConfiguration to help control this.

devopsjnr:
@ellistarn Though this was already in process. Anyway, it would be great exposing it. Could you think of overriding MaxPods from the provisioner?

bwagner5:
@devopsjnr we've talked about this a lot before and I'm very supportive of it. The KubeletConfiguration that @ellistarn is referring to is apart of the provisioner spec: https://karpenter.sh/v0.13.2/provisioner/#:~:text=and%20use%20cases.-,kubeletConfiguration,-%3A%0A%20%20%20%20clusterDNS%3A

Issue #2129, opened 2022-07-14
Allow passing --kube-reserved/system-reserved/eviction-hard flags


gkaskonas:
**Tell us about your request**
I would like to set reserved resources for kube and system. Maybe you could expand KubeletConfiguration to allow any flag to be passed?

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
We have noticed that some nodes become unresponsive and karpenter doesn't know what to do. They go into NodeReady state and pods get stuck in Terminating state with "Kubelet stopped sending updates" message.
These flags can be used to allocate enough resources for kubelet and system to prevent these issues.

**Are you currently working around this issue?**
Yes

**Additional context**
Anything else we should know?

**Attachments**
If you think you might have additional information that you'd like to include via an attachment, please do - we'll take a look. (Remember to remove any personally-identifiable information.)

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ellistarn:
I'm supportive of this.

jonathan-innis:
`--system-reserved` override is now supported as of #2254

Issue #1995, opened 2022-06-24
Allow allocating storage capacity based on instance type 


tonylsw:
**Tell us about your request**
When using custom launch template with Karpenter, a single node storage capacity (specified in the launch template) is used by all types of instances launched by Karpenter. We'd like to able to allocate storage capacity based on instance type.

For example:
m5.xlarge: /dev/sdf=100
m5.2xlarge /dev/sdf=200

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**

Currently Karpenter using a single volume capacity for all types of instances, it's not possible to allocate volume based on instance type.

Using multiple provisioners, one for each instance type, can support this, but the instance launched in such case is undeterministic  
and not optimal.

**Are you currently working around this issue?**

Have to specify the maximum volume size in the launch template in order to avoid running out of disk.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

dewjam:
Hey @tonylsw,
Thanks for creating this issue and sorry for the delayed response.  Would you be willing to share more info regarding your use-case?  Are you looking to scale storage based on instance types to accommodate container images and filesystems?  Or are you allocating local ephemeral storage to pods/containers?

tonylsw:
Hi @dewjam ideally, we'd like to have a solution to scale up/down node storage on-demand, much like Karpenter does for pod :-) , short of that, we'd like to allocate storage based on instance type for container images.

carlesjavierre:
@dewjam Do we have any news regarding this/these (maybe they can be considered different) features?

dewjam:
Hey @carlesjavierre ,
My apologies for the delayed response.  For your use-case, you could expand a nodes storage from the default (20GB) by defining a Block Device Mapping in the Karpenter Provisioner.  Take a look at the docs and let me know if that helps:  https://karpenter.sh/v0.14.0/aws/provisioning/#block-device-mappings

dewjam:
>ideally, we'd like to have a solution to scale up/down node storage on-demand, much like Karpenter does for pod :-)

[...]

PR #2390, opened 2022-08-29
RFC: Add instance type settings design proposal

jonathan-innis:
Fixes # 

**Description**

- Adds Instance Type settings design doc and RFC.

*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/designs/instance-type-settings.md
+++ b/designs/instance-type-settings.md
@@ -1,0 +1,341 @@
+# Instance Type Settings
+
+## Goals
+
+Allow users to:
+
+* Define [custom device resources](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/) on an instance type
+* Define allocatable memory, cpu, ephemeral storage, pods on an instance type
+* Define pricing overrides for enterprise pricing support on an instance type or instance family
+* Define pricing discounts as raw values and percentages off of the default pricing on an instance type or instance family
+* Expand ephemeral storage for launching instances that contain pods that surpass the block device mappings default ephemeral storage
+
+## Background
+
+Kubelet-specific configuration for nodes deployed by Karpenter is passed through to the `bootstrap.sh` script and is used when bootstrapping the new node to the cluster. Karpenter currently supports certain static default values that are used when bootstrapping the node. In particular, values like `--max-pods=110` are used when the `AWS_ENABLE_POD_ENI` is set to `false`. Users currently have no way to specify extra arguments to the kubelet beyond `--max-pods` and `--system-reserved`. Additionally, there is currently no supported way to define per-instance type or per-instance family memory/cpu/ephemeral-storage requirements for overhead that is used during scheduling decision-making.
+
+We currently surface all of these values on a per-provisioner basis through `.spec.kubeletConfiguration` for configuration that is passed down into the userData when bootstrapping the cluster and on a per-controller basis for values like `VM_MEMORY_OVERHEAD` values. Because instance types are vastly different in their resources, we need a way to expand this logic so that users can specify different metadata overrides on a per-instance type basis as opposed to a per-provisioner or per-controller basis.
+
+Finally, we make consolidation decision-making based on the approximated cost-savings that we will get by either removing a node and rescheduling its pods on existing nodes or by replacing an existing node that has extra capacity with the pods that are currently scheduled on it with a smaller node that is strictly cheaper than the existing node.
+ 
+Items with a 🔑 symbol call out key points or things of note
+
+## 🔑 Proposed Updated API Surface
+
+**Introduce `v1alpha1/InstanceType` CRD**
+
+```yaml
+apiVersion: karpenter.sh/v1alpha1
+kind: InstanceType
+metadata:
+  name: "c5.large"
+spec:   
+  offerings: # These define the requirement offering possibilities for the instance type
+    - karpenter.sh/capacity-type: spot
+      topology.kubernetes.io/zone: us-west-2b
+      karpenter.sh/price: 0.40
+    - karpenter.sh/capacity-type: spot
+      topology.kubernetes.io/zone: us-west-2c
+  resources: # This defines the resources that the instance has for bin-packing, will be merged with known resource quantities
+    hardware-vendor.example/foo: "2"
+    hardware-vendor.example/bar: "20"
+    pods: 10
+  overhead: # Any overhead that should be subtracted from the instance type away from resources
+    memory: 200Mi    
+```
+
+
+**Extend `v1alpha5/Provisioner` CRD**
+
+In particular, we will extend the `.spec.kubeletConfiguration` with `kubeReserved` and `evictionThreshold` that will be passed down into the bootstrap scripts.
+
+```yaml
+apiVersion: karpenter.sh/v1alpha5
+kind: Provisioner
+metadata:
+  name: default
+spec:
+  requirements:
+    - key: karpenter.sh/capacity-type
+      operator: In
+      values: ["spot"]
+  limits:
+    resources:
+      cpu: "1000"
+  kubeletConfiguration:
+    kubeReserved:
+      memory: "1Gi"
+      cpu: "300m"
+      ephemeral-storage: "1Gi"
+    systemReserved:
+      memory: "1Gi"
+      cpu: "300m"
+      ephemeral-storage: "1Gi"
+    evictionThreshold:
+      hard:
+        memoryAvailable: "1Gi"
+        imagefsAvailable: "10%"
+        nodefsAvailable: "15%"
+        nodefsInodesFree: "5%"
+      soft:
+        ...
+```
+
+## Instance Type Setting Examples
+
+1. Configure the VM memory overhead values for `c5.large`
+
+```yaml
+apiVersion: karpenter.sh/v1alpha1
+kind: InstanceType
+metadata:
+   name: "c5.large"
+spec:
+   overhead:
+      memory: 200Mi
+```
+
+1. Configure custom device resources for custom device resources for `c4.large`
+
+```yaml
+apiVersion: karpenter.sh/v1alpha1
+kind: InstanceType
+metadata:
+  name: "c4.large"
+spec:
+   resources:
+      hardware-vendor.example/foo: "2"
+```
+
+1. Specify custom offerings with custom pricing information for `c4.large`
+
+```yaml
+apiVersion: karpenter.sh/v1alpha1
+kind: InstanceType
+metadata:
+  name: "c4.large"
+spec:
+  offerings:
+  - karpenter.sh/capacity-type: on-demand
+    topology.kubernetes.io/zone: us-west-2b
+    karpenter.sh/price: "0.60"
+  - karpenter.sh/capacity-type: spot
+    karpenter.sh/price: "0.40"
+```
+
+## Considerations
+
+### Instance Type Overrides for Other Cloud Providers (Azure/CAPI/etc.)
+
+Because Karpenter is intended to be cloud-agnostic, we need to keep this in mind when adding changes to the API surface. For the initial iteration of the `InstanceType` CRD, we will assume that the instance type will be selected by the `.metadata.name` field which will map to the well-defined Kubernetes node field `node.kubernetes.io/instance-type`.
+
+For other clouds, this is expected to be mapped on the nodes as it is in AWS and the `GetInstanceType` call for other cloud providers is expected to retrieve and return all the instance type values that map to the `node.kubernetes.io/instance-type` value. In this way, Karpenter can perform instance type overrides for other clouds when instance types are surfaced.
+
+### Custom Resource Cardinality
+
+One of the primary questions presented is whether we should create a 1-1 mapping between settings and a given instance type or whether we should allow the ability to create more complex structures using set relationships and label selectors. Options are listed below combined with the pros and cons for each:
+
+**Options**
+
+1. Create an `InstanceType` CRD that specifies a 1-1 relationship between the settings and the instance type
+    1. Pros
+        1. Clearly defined relationship between an instance type and the settings that are defined for it
+        2. Users that only have a few instance types will not have an issue configuring the instance types with these values
+        3. We can consider adding an `InstanceTypeClass` or similar at a later date that would allow a grouping of instance types to be assigned the same settings
+    2. Cons
+        1. May not scale well as users that have high levels of customizations across large numbers of instance types may have to maintain a large number of `InstanceType` CRDs
+        2. I may want to define settings over an instance family without having to individually configure the instances that exist within that family
+2. Create an `InstanceTypeSetting` CRD that contains a list of instance types that allow a 1-many relationship between settings and instance types. These instance types could be selected either with a `selector` or `regex`
+    1. Pros
+        1. Allows users who want to apply the same instance type settings across a wide range of instance types or across an instance type family to do so
+        2. Lower maintenance burden from maintaining less `InstanceTypeSetting` CRDs
+    2. Cons
+        1. Can create complex relationships that may be difficult to track for users
+        2. If we were to allow overlapping `InstanceTypeSettings`, we would have to create a `weight` on the `InstanceTypeSettings` which would further convolute details for the user
+        3. If we were to deny overlapping `InstanceTypeSettings` , then the feature of specifying a group of instance types may not be used that often as configuration between instance types may rarely be shared
+        4. It’s unclear whether users would want have enough common configuration that they would assign across instance types. It makes more sense to iterate on a grouping mechanism later if there is a user ask for it
+
+**Recommendation:** Use `InstanceType` CRD for initial iteration for 1-1 mappings before considering introducing more complex constructs to the user. While specifying rulesets might reduce the overhead burden for customers to keep track of separate `InstanceTypeSettings` , ultimately without an API surface to view exactly what settings that ruleset will produce, it will ultimately be confusing for users.
+
+### 🔑 Modeling Offerings as Requirements
+
+Currently, we model Offerings with the following static properties in the internal offering struct:
+
+```go
+type Offering struct {
+  CapacityType string
+  Zone string
+  Price float64
+  Available bool
+}
+```
+
+Rather than model these values as static properties, we can extend the requirements model that we leverage in the InstanceType internal struct to transform offerings into
+
+```go
+type Offering struct {
+  Requirements scheduling.Requirements
+}
+```
+
+Requirements in Offerings would then be modeled by default for a given offering type with Capacity Type: “spot”, Zone: “us-west-2a” and Price per hour in USD of “0.40” as:
+
+```yaml
+offerings:
+    - requirements:
+      - key: "topology.kubernetes.io/zone"
+        operator: "In"
+        values: ["us-west-2a"]
+      - key: "karpenter.sh/capacity-type"
+        operator: "In"
+        values: ["spot"]
+      - key: "karpenter.sh/price"
+        operator: "In"
+        values: ["0.40"]
+```
+
+For an instance type that has no InstanceType overrides, we would use the default values used in the `createOfferings` call that is generated from the `DescribeInstanceType` call. This set of offerings would describe the full combination of options that are available from the cloud provider for a given instance type. When we schedule and attempt to launch an instance, we validate that there is an offering that matches the nodeTemplate requirements that are added from the scheduler.
+
+In the case that the user provides an Instance Type offerings overrides (currently the scenario that they would do this would be for pricing overrides), we will use the provided offerings from the `InstanceType` CR as the offering values as opposed to the offering requirement values that are used by default for the cloud provider. To simplify the semantic, we assume the “In” operator for the offerings and we only allow a single value for each label requirement.
+
+```yaml
+offerings:
+  - karpenter.sh/capacity-type: spot
+    topology.kubernetes.io/zone: us-west-2b
+    karpenter.sh/price: "0.40"
+  - karpenter.sh/capacity-type: on-demand # No zonal constraint specified
+    karpenter.sh/price: "0.60"
+```
+
+This implies that I can create offerings that don’t contain all constraints that are imposed by the default offering implementation. For instance, in the above `offerings` overrides, I don’t need a zonal constraint on the second offering constraint set which is essentially saying “give me an on-demand offering in any zone and all on-demand offerings regardless of zone cost ~$6.00 per day.”
+
+The upshot of this is that we become far more flexible with our offerings in that we can now have arbitrary constraints specified by users that can differentiate the pricing for a set of labels (i.e. I can specify different pricing for different OS for instance)
+
+🔑  Offerings specified in `.spec.offerings` would *fully override* the offerings that would come by default for the AWS cloud provider. This means that the ownership is on the user to fully define the set of offerings that they want to make available for a given instance type.
+
+**🔑  Recommendation:** Convert static offering constraints into offering requirements that give more flexibility into how to constrain offerings on instance types. Allow users to define which offerings they want to surface per `InstanceType` using the offering requirements/constraints.
+
+### ✨ Reserved Resources/Resource Overhead
+
+For any given node that is launched as part of Kubernetes, kubelet can be configured with both [system reserved resources](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#system-reserved)  and [kubelet reserved resources.](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#kube-reserved) System reserved resources are more likely to be tied to the system daemons that are shipped with an OS image and will most likely be flat resource overhead values on a per-provisioner basis. As such, they are more closely tied to the AMI that the node is launched with, which is tied to the Provisioner `provider` directly.
+
+Kubelet reserved resources are much more a function of the overhead of the pods that run on the node. Kubelet reserved resources can be affected by:
+
+1. Number of pods running on a given node
+2. The size of the images running on a given node
+
+Thus, we can break down resource overhead into two main categories:
+
+1. Flat resource overhead
+    1. This includes systemReserved values for system daemons, OS overhead, etc.
+    2. Hard eviction threshold for the node
+    3. OS memory overhead (assigned to **VM_MEMORY_OVERHEAD** now) to run the instance
+2. Dynamic Resource Overhead (per-pod overhead)
+    1. This includes kubeReserved values needed for the container runtime. These values are typically dependent on the number of pods that are scheduled to the instance (or at worst the maximum number of pods that could be scheduled to the instance)
+    2. This includes needed ephemeral storage since large images that come from multiple pods scheduled to a node can increase the needed ephemeral storage and affect the need to have larger blockDeviceMappings on the provisioner
+
+Based on these two categories, we can tackle user-defined overhead/resource values in the following way:
+
+1. Flat Resource Overhead
+    1. `kubeletConfiguration` can live at the `Provisioner` level. If there is need to specify flat `kubeReserved` or `systemReserved` values based on the user’s knowledge of the OS/AMI, they are able to set this at the Provisioner-level
+    2. Hard eviction thresholds can be set within the `kubeletConfiguration` at the `Provisioner` level as this will often be some globally defaulted value or vary based on knowledge of the OS or the system daemons
+    3. OS memory overhead (**VM_MEMORY_OVERHEAD**) can be moved into this `InstanceType` CRD. Since this overhead in bytes will vary on a per-instance type level, we can allow the user to set a custom value to reserve for VM_OVERHEAD rather than defaulting to the percentage value that is currently reserved for overhead.
+2. Dynamic Resource Overhead (per-pod overhead)
+    1. [RuntimeClass](https://kubernetes.io/docs/concepts/containers/runtime-class/) provides a mechanism for users to specify per-pod overhead values that can be reserved for each pod. This is the recommended way that Kubernetes provides for specifying any pod overhead values that are dynamic and based on the CRI.
+       1. 🔑 Dynamic pod overhead that is tied to the container runtime should move to a `RuntimeClass` if there is pod overhead unique to the CRI that is used by your nodes.
+    2. Other per-pod overhead that may be tied to things like image size can be tied to the pod `ephemeral-stroage` requests which will help Karpenter ensure that it accurately provisions pods to nodes while also ensuring that there is parity between the scheduling decisions of Karpenter and the kube-scheduler.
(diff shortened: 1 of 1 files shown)


jonathan-innis:
#2129, #2180, #1490, #2129, #1803, #1995, and #2161 addressed by this design. RFC from issue owners and others that are interested in the design.

designs/instance-type-settings.md
```diff
@@ -0,0 +1,354 @@
+# Instance Type Settings
+
+## Goals
+
+Allow users to:
+
+* Define [custom device resources](https://github.com/aws/karpenter/pull/2161) on an instance type
```
njtran, line 7: nit: Can we reference the issue instead of the pull request here? 
```suggestion
* Define [custom device resources](https://github.com/aws/karpenter/issues/1900) on an instance type
```

njtran (dismissed):
LGTM. Just one nit

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+3. Max Pods
+    1. In general, it appears to be an anti-pattern in the Kubernetes community to strictly fix maxPods values (for kubeletConfiguration) on a per-instance type level. Users should be restricting the number of pods on their instances based on pod requests and size of instances as opposed to strict pod density sizing. If users are concerned with Provisioners creating instances that are too large, there is now the option to specify GT or LT operators for setting limits for how large an instance type that is provisioned can get.
+    2. 🔑 For users that continue to utilize and/or need `--max-pods`, we will surface a `pods` value in the `.spec.allocatable` of the `InstanceType`. This `pods` value will be used for pod density for bin-packing and scheduling. Since there would now be a `max-pods` value surfaced at both the `InstanceType` level and at the `Provisioner` level, we will take the `min(provisioner.maxPods, instanceType.maxPods)` to be the pod density for the instance.
```
ellistarn, line 249: IIUC, this is the only field of the instancetype CRD that is capable of influencing user data. I'm a bit wary of this coupling, how it might limit our design decisions in the future, as well as the security implications. Put differently, InstanceType is currently a readonly type and you're proposing to give it write privileges. It currently isn't capable of modifying the configuration of nodes, which is a special privilege of the Provisioner. I'm comfortable using it for binpacking, since this relationship exists today. 

Here's a less controversial alternative that functions identically w/ how `type InstanceType interface` works today, and doesn't sacrifice any use cases.

1. The InstanceType CRD can expose `pods` as a value in `Allocatable`
2. The `maxpods` value can be hardcoded in the provisioner's kubelet configuration
3. Our binpacker will default to what the `InstanceType` says (which may be overriden by the CRD), but if the provisioner sets a value, that's what we'll use.
4. If the customer wants to compute a dynamic (w.r.t. instancetype) max pods value, they can do so in user data, and they just need to make sure that this function agrees with the values they've overridden in the instancetype CRD. 
5. If you need different dynamic max pods values for the same instance type but different provisioners, stop and rethink your architecture.

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+    karpenter.sh/price: "0.60"
+  - karpenter.sh/capacity-type: spot
+    karpenter.sh/price: "0.40"
+```
+
+## Considerations
+
+### Instance Type Overrides for Other Cloud Providers (Azure/CAPI/etc.)
+
+Because Karpenter is intended to be cloud-agnostic, we need to keep this in mind when adding changes to the API surface. For the initial iteration of the `InstanceType` CRD, we will assume that the instance type will be selected by the `.metadata.name` field which will map to the well-defined Kubernetes node field `node.kubernetes.io/instance-type`.
```
ellistarn, line 129:
```suggestion
Because Karpenter is cloud-agnostic, we need to keep this in mind when adding changes to the API surface. For the initial iteration of the `InstanceType` CRD, we will assume that the instance type will be selected by the `.metadata.name` field which will map to the well-defined Kubernetes node field `node.kubernetes.io/instance-type`.
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+    2. 🔑 For users that continue to utilize and/or need `--max-pods`, we will surface a `pods` value in the `.spec.allocatable` of the `InstanceType`. This `pods` value will be used for pod density for bin-packing and scheduling. Since there would now be a `max-pods` value surfaced at both the `InstanceType` level and at the `Provisioner` level, we will take the `min(provisioner.maxPods, instanceType.maxPods)` to be the pod density for the instance. 
+       1. These per-instance type pod density values will be passed into the userData that is used to bootstrap the node. Instance type discovery will be performed in the userData script to determine the max pods value to pass to `bootstrap.sh`.
+
+**🔑 Open Questions**
+
+1. There is still [users asks](https://github.com/aws/karpenter/issues/1803) for calculating the kubeReserved resources value based on the GKE calculation of the kubeReserved resources value as opposed to the EKS default which is based on pod density. Should we allow some flag on the provisioner that can determine which way to base the calculation on? i.e. `kubeReservedMemoryStrategy`
```
ellistarn, line 254: I'd love for us to do the work and figure out containerd overhead in 2022, and then provide customers with an accurate runtime class.

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+
+
+**🔑 Recommendation:** Based on the additional overhead and muddiness of putting `blockDeviceMappings` in the `InstanceType` CR, we will punt on block device mappings for now. For now, we will consider extending ephermal-storage through some dynamic mechanism out of scope, though the hope is that we will have some mechanism as a future feature that calculates the combined needed ephemeral-storage based off of the pod requests, kubeReserved and systemReserved.
+
+### Launch Templates
+
+For now, we are avoiding the need for Instance Type changes to affect launch template cardinality. Configuration at the `InstanceType` level is intended to configure the starting point for the scheduler to know how many resources can be used by the bin-packer. The expectation is that when the node comes up this is roughly the allocatable amount of resources the node will have, less kubeReserved, systemReserved, and evictionThresholds. Thus, the calculation we will do will be:
+
+```
+Allocatable = Instance Type Resources - kubeReserved - systemReserved - evictionThreshhold
```
ellistarn, line 296: Now that I see this, my recommendation elsewhere to have the field be `InstanceType.Spec.Allocatable` is inaccurate. Is it more correct to make it `InstanceType.Spec.Capacity` and then we substract off these additional 3 values that are controlled at the provisioner level?

ellistarn:
🚀 🎸 🎸 🚀

designs/instance-type-settings.md
jonathan-innis: > as well as the security implications
I'm not quite sure I get the security implications that you are trying to call out here. Maybe you could describe it further. I see that this changes some read/write assumptions about the `InstanceType` and how it interacts with the current code logic but I'm not quite sure how this has security implications for the project.

> Here's a less controversial alternative that functions
I think the other option to call out is that we can surface `podsPerCore` in the `kubeletConfiguration` in the `Provisioner` which would give users another method to have this be a dynamic value without the overhead of having to write custom `userData`. If we are going to say that `pods` in the `.sepc.capacity` can't affect `--max-pods`, I would rather just punt having `pods` included entirely.

designs/instance-type-settings.md
jonathan-innis: Yes, you are completely right. This should be `Capacity` not `Allocatable`

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+apiVersion: karpenter.sh/v1alpha1
+kind: InstanceType
+metadata:
+   name: "c5.large"
+spec:
+   capacity:
+      memory: 200Mi
+```
+
+1. Configure custom device resources for custom device resources for `c4.large`
```
chrisnegus, line 97:
```suggestion
1. Configure custom device resources for `c4.large`
```
I think that's what you mean.

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+
+### Instance Type Overrides for Other Cloud Providers (Azure/CAPI/etc.)
+
+Because Karpenter is intended to be cloud-agnostic, we need to keep this in mind when adding changes to the API surface. For the initial iteration of the `InstanceType` CRD, we will assume that the instance type will be selected by the `.metadata.name` field which will map to the well-defined Kubernetes node field `node.kubernetes.io/instance-type`.
+
+For other clouds, this is expected to be mapped on the nodes as it is in AWS and the `GetInstanceType` call for other cloud providers is expected to retrieve and return all the instance type values that map to the `node.kubernetes.io/instance-type` value. In this way, Karpenter can perform instance type overrides for other clouds when instance types are surfaced.
+
+### Custom Resource Cardinality
+
+One of the primary questions presented is whether we should create a 1-1 mapping between settings and a given instance type or whether we should allow the ability to create more complex structures using set relationships and label selectors. Options are listed below combined with the pros and cons for each:
```
chrisnegus, line 135:
```suggestion
One of the primary questions presented is whether we should create a one-to-one mapping between settings and a given instance type or whether we should allow the ability to create more complex structures using set relationships and label selectors. Options are listed below, combined with the pros and cons for each:
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+
+**Options**
+
+1. Create an `InstanceType` CRD that specifies a 1-1 relationship between the settings and the instance type
```
chrisnegus, line 139:
```suggestion
1. Create an `InstanceType` CRD that specifies a one-to-one relationship between the settings and the instance type
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+    1. Pros
+        1. Clearly defined relationship between an instance type and the settings that are defined for it
+        2. Users that only have a few instance types will not have an issue configuring the instance types with these values
+        3. We can consider adding an `InstanceTypeClass` or similar at a later date that would allow a grouping of instance types to be assigned the same settings
+    2. Cons
+        1. May not scale well as users that have high levels of customizations across large numbers of instance types may have to maintain a large number of `InstanceType` CRDs
+        2. I may want to define settings over an instance family without having to individually configure the instances that exist within that family
```
chrisnegus, line 146:
```suggestion
        2. Someone may want to define settings over an instance family without having to individually configure the instances that exist within that family
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+2. Create an `InstanceTypeSetting` CRD that contains a list of instance types that allow a 1-many relationship between settings and instance types. These instance types could be selected either with a `selector` or `regex`
```
chrisnegus, line 147:
```suggestion
2. Create an `InstanceTypeSetting` CRD that contains a list of instance types that allow a one-to-many relationship between settings and instance types. These instance types could be selected either with a `selector` or `regex`
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+    1. Pros
+        1. Allows users who want to apply the same instance type settings across a wide range of instance types or across an instance type family to do so
+        2. Lower maintenance burden from maintaining less `InstanceTypeSetting` CRDs
```
chrisnegus, line 150:
```suggestion
        2. Lower maintenance burden from maintaining fewer `InstanceTypeSetting` CRDs
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+    2. Cons
+        1. Can create complex relationships that may be difficult to track for users
+        2. If we were to allow overlapping `InstanceTypeSettings`, we would have to create a `weight` on the `InstanceTypeSettings` which would further convolute details for the user
+        3. If we were to deny overlapping `InstanceTypeSettings` , then the feature of specifying a group of instance types may not be used that often as configuration between instance types may rarely be shared
```
chrisnegus, line 154:
```suggestion
        3. If we were to deny overlapping `InstanceTypeSettings`, then the feature of specifying a group of instance types may not be used that often, as configuration between instance types may rarely be shared
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+        4. It’s unclear whether users would want have enough common configuration that they would assign across instance types. It makes more sense to iterate on a grouping mechanism later if there is a user ask for it
+
+**Recommendation:** Use `InstanceType` CRD for initial iteration for 1-1 mappings before considering introducing more complex constructs to the user. While specifying rulesets might reduce the overhead burden for customers to keep track of separate `InstanceTypeSettings` , ultimately without an API surface to view exactly what settings that ruleset will produce, it will ultimately be confusing for users.
```
chrisnegus, line 157:
```suggestion
**Recommendation:** Use `InstanceType` CRD for initial iteration for one-to-one mappings before considering introducing more complex constructs to the user. While specifying rulesets might reduce the overhead burden for customers to keep track of separate `InstanceTypeSettings`, ultimately without an API surface to view exactly what settings that ruleset will produce, it will ultimately be confusing for users.
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+```yaml
+offerings:
+  - karpenter.sh/capacity-type: spot
+    topology.kubernetes.io/zone: us-west-2b
+    karpenter.sh/price: "0.40"
+  - karpenter.sh/capacity-type: on-demand # No zonal constraint specified
+    karpenter.sh/price: "0.60"
+```
+
+This implies that I can create offerings that don’t contain all constraints that are imposed by the default offering implementation. For instance, in the above `offerings` overrides, I don’t need a zonal constraint on the second offering constraint set which is essentially saying “give me an on-demand offering in any zone and all on-demand offerings regardless of zone cost ~$6.00 per day.”
```
chrisnegus, line 209: Instead of "I", perhaps use "you" or "someone" in this paragraph.

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+    2. Hard eviction threshold for the node
+    3. OS memory overhead (assigned to **VM_MEMORY_OVERHEAD** now) to run the instance
+2. Dynamic Resource Overhead (per-pod overhead)
+    1. This includes kubeReserved values needed for the container runtime. These values are typically dependent on the number of pods that are scheduled to the instance (or at worst the maximum number of pods that could be scheduled to the instance)
+    2. This includes needed ephemeral storage since large images that come from multiple pods scheduled to a node can increase the needed ephemeral storage and affect the need to have larger blockDeviceMappings on the provisioner
+
+Based on these two categories, we can tackle user-defined overhead/resource values in the following way:
+
+1. Flat Resource Overhead
+    1. `kubeletConfiguration` can live at the `Provisioner` level. If there is need to specify flat `kubeReserved` or `systemReserved` values based on the user’s knowledge of the OS/AMI, they are able to set this at the Provisioner-level
```
chrisnegus, line 239:
```suggestion
    1. `kubeletConfiguration` can live at the `Provisioner` level. If there is need to specify flat `kubeReserved` or `systemReserved` values based on the user’s knowledge of the OS/AMI, they are able to set this at the Provisioner level
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+    3. OS memory overhead (**VM_MEMORY_OVERHEAD**) can be moved into this `InstanceType` CRD. Since this overhead in bytes will vary on a per-instance type level, we can allow the user to set a custom value to reserve for VM_OVERHEAD rather than defaulting to the percentage value that is currently reserved for overhead.
+2. Dynamic Resource Overhead (per-pod overhead)
+    1. [RuntimeClass](https://kubernetes.io/docs/concepts/containers/runtime-class/) provides a mechanism for users to specify per-pod overhead values that can be reserved for each pod. This is the recommended way that Kubernetes provides for specifying any pod overhead values that are dynamic and based on the CRI.
+       1. 🔑 Dynamic pod overhead that is tied to the container runtime should move to a `RuntimeClass` if there is pod overhead unique to the CRI that is used by your nodes.
+    2. Other per-pod overhead that may be tied to things like image size can be tied to the pod `ephemeral-stroage` requests which will help Karpenter ensure that it accurately provisions pods to nodes while also ensuring that there is parity between the scheduling decisions of Karpenter and the kube-scheduler.
```
chrisnegus, line 245:
```suggestion
    2. Other per-pod overhead that may be tied to things like image size can be tied to the pod `ephemeral-storage` requests which will help Karpenter ensure that it accurately provisions pods to nodes while also ensuring that there is parity between the scheduling decisions of Karpenter and the kube-scheduler.
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+    2. 🔑 For users that continue to utilize and/or need `--max-pods`, we will surface a `pods` value in the `.spec.capacity` of the `InstanceType`. This `pods` value will be used for pod density for bin-packing and scheduling. Since there would now be a `max-pods` value surfaced at both the `InstanceType` level and at the `Provisioner` level, we will take the `min(provisioner.maxPods, instanceType.maxPods)` to be the pod density for the instance. 
+       1. These per-instance type pod density values will be passed into the userData that is used to bootstrap the node. Instance type discovery will be performed in the userData script to determine the max pods value to pass to `bootstrap.sh`.
+
+**🔑 Open Questions**
+
+1. There is still [users asks](https://github.com/aws/karpenter/issues/1803) for calculating the kubeReserved resources value based on the GKE calculation of the kubeReserved resources value as opposed to the EKS default which is based on pod density. Should we allow some flag on the provisioner that can determine which way to base the calculation on? i.e. `kubeReservedMemoryStrategy`
```
chrisnegus, line 254:
```suggestion
1. There is still [users asks](https://github.com/aws/karpenter/issues/1803) for calculating the kubeReserved resources value based on the GKE calculation of the kubeReserved resources value, as opposed to the EKS default which is based on pod density. Should we allow some flag on the provisioner that can determine which way to base the calculation on (i.e. `kubeReservedMemoryStrategy`)?
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+        iops: 10000
+        encrypted: true
+        kmsKeyID: "arn:aws:kms:us-west-2:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab"
+        deleteOnTermination: true
+        throughput: 125
+        snapshotID: snap-0123456789
+```
+
+
+**🔑 Recommendation:** Based on the additional overhead and muddiness of putting `blockDeviceMappings` in the `InstanceType` CR, we will punt on block device mappings for now. For now, we will consider extending ephermal-storage through some dynamic mechanism out of scope, though the hope is that we will have some mechanism as a future feature that calculates the combined needed ephemeral-storage based off of the pod requests, kubeReserved and systemReserved.
```
chrisnegus, line 289:
```suggestion
**🔑 Recommendation:** Based on the additional overhead and muddiness of putting `blockDeviceMappings` in the `InstanceType` CR, we will punt on block device mappings for now. For now, we will consider extending ephemeral-storage through some dynamic mechanism out of scope, though the hope is that we will have some mechanism as a future feature that calculates the combined needed ephemeral-storage based off of the pod requests, kubeReserved and systemReserved.
```

designs/instance-type-settings.md
```diff
@@ -0,0 +1,340 @@
...
+    1. Pros
+        1. Creates clean separation between pricing configuration for instances and the system/requirements specific information for instances (keep separation between pricing and `InstanceTypeSettings`
+    2. Cons
+        1. Creates another CR/configMap that the user has to configure, understand, and maintain
+2. Add pricing information to the `InstanceType` CRD
+    1. Pros
+        1. Maintains all instance-specific information in one location
+        2. Less CRDs for the user to maintain
+    2. Cons
+        1. Pricing information is instance-type specific which means we are necessarily requiring a 1-1 mapping between some CRD and an instance type. This creates less flexibility if a user wants to describe common settings across an instance family or across an architecture type.
```
chrisnegus, line 316:
```suggestion
        1. Pricing information is instance-type specific which means we are necessarily requiring a one-to-one mapping between some CRD and an instance type. This creates less flexibility if a user wants to describe common settings across an instance family or across an architecture type.
```

chrisnegus (commented):
This reads really well. I just have few small suggestions.

designs/instance-type-settings.md
AndreasSko, line 157: For our use case (similar to #1900 where nodes offer fuse devices via the [smarter device manager](https://gitlab.com/arm-research/smarter/smarter-device-manager)) the recommended approach of having a 1-1 mapping would create quite a significant overhead and burden: For every instance type and size, we would need to create an `InstanceType` resource and configure it. Even with templating this is quite a bit of effort and seems prone to errors. Also, we would need to keep the list of available instance types up to date, increasing the maintenance work. So going with the second option sounds like a more sustainable approach to me personally.

jonathan-innis: How would you get around having this burden? My assumption is that each instance type wouldn't have a static number of resources that match the `smarter-devices/fuse` so you would need some dynamic way to configure the quantities for each of the different instance types, or at least a collection of instance types.

This becomes a problem when extending out to other resources. Even assuming that there is a consistent quantity of resources for the `smarter-devices/fuse` resource across _all_ instance types, that certainly wouldn't be true for other extended resources.

> Also, we would need to keep the list of available instance types up to date
This works as an override mechanism, so you would still have those instance types available if they didn't have a matching CR for them. The only difference there would be that the extended resource wouldn't be mapped onto it, which would still be something that would need to be addressed with the other "collection-based" approach.

jonathan-innis: Also, the other thing that we were thinking with the approach proposed in this design, is we would offer some tooling/guidance around how to programmatically generate these values into CRs, so that the maintenance burden is reduced.

designs/instance-type-settings.md
jonathan-innis: Good catch 👍

jonathan-innis:
Migrating a thread from #2404:

> My assumption is that all instance types wouldn't have the same number of extended resources

Let me picture our two use cases:

For a large videoconferencing system (up to several hundreds of nodes per cluster) we leverage the ALSA /dev/snd device exposed by the kernel (module). Obviously the amount of these kernel devices available per host are not in relation to an instance size or type and probably not even on CPU architecture, but rather on linux kernel config. For this amount of instances Karpenter is the answer, overcomimg AWS provider limits of the availablity of a certain type/size, so the loosely tied flexibilty of types and sizes possible by Karpenter is what we have been looking for. Now having to constrain on few types and sizes by the current manual declaration cancels the whole benefits out.
We have a container build farm that [requires /dev/fuse device](https://www.redhat.com/sysadmin/podman-inside-kubernetes). The number of fuse devices provided by the kernel is also not tied to instance type or size, per default 1000 or more. Also here it is impossible for us to define all instance types times instance sizes to match all provider offers.
Granted there are other use cases that I might not be aware of, but I am struggling to picture a case where a one-to-one relation of this config would shine. In our case it would be a management nightmare so to speek. Maybe this context was missing to explain why my simple https://github.com/aws/karpenter/pull/2161 is sufficient for us. Thank you for considering!

jonathan-innis:
Summarizing most feedback that we've heard. It seems like we might want to consider a mechanism for "selecting" sets of instance types so that we can ease the customer burden of having to maintain a 1:1 mapping.

An alternative solution here may be to keep the 1:1 mapping and provide some documentation around manifest generation for these `InstanceType` CRDs.

Though this might be more confusing to reason about; in general, it will cause less overhead for users.

haarchri:
any chance to see this in one of the coming releases ?

universam1:
@jonathan-innis We would really like to move out of the fork of #2161 for a significant number of clusters. We are forced to where otherwise it is be a blocking no-go for Karpenter.
In case the final solution is pending longer, would you please reconsider merging #2161 for the time being, as it does provide an MVP and it is tested well with hundreds of nodes to be working as expected?
Thank you

jonathan-innis:
@universam1 We've talked a bit about this internally. I think we're willing to take a short-term experimental path that would unblock you in the main project. Ideally, the hope is that we can get a more robust solution that should solve for all the problems in the RFC going forward, but right now we haven't put this at super high priority.

Here's the thoughts:
1. The long-term solution is still under design and is deproiritized for the time-being but one idea that we have is to expose the InstanceTypes offered by a cloudprovider as CRs on the cluster. This would allow you to either SSA changes to the InstanceTypes or to create mutating webhooks that could receive the InstanceTypes and then edit them with anything that you need injected into them (like additional extended resources). Thoughts on this one?
2. As a short-term solution, we’d like to add a ConfigMap that exposes any extended resources (on top of the known ones) through a mapping from instance types to a v1.ResourceList like the example below. This would be an experimental feature and would be highly likely to change at some point in the future in favor of something more akin to 1 that would be a more robust long-term solution. Would you be interested in contributing to this short-term solution?

```yaml
"c5.large": "{"fuse": "1", "other-extended-resource": "4"}"
```

ellistarn:
Perhaps we can model it based off of the instance type struct? ```{ "name": "c5.large", "capacity": {"fuse": "1", "other-extended-resource": "4"} }```

universam1:
Thank you @jonathan-innis for considering this request and taking into account several options to solve it. Very much appreciated!

We are certainly happy to contribute to a solution - if you can provide some more details where you would see it fit into and what you'd specifically like to see, that would help getting us started. 
No issue with alpha grade solution that is likely going to change.

One requirement though from our side is to be agnostic to instance type and size. It is impossible for us to predefine all variants we like to support ( _type_ x _size_ ).
A simple globbing or regex support in `aws.instanceTypeOverrides` matcher doesn't sound hard for me to add and should meet our needs just fine. Thank you very much for consideration!

jonathan-innis:
> It is impossible for us to predefine all variants we like to support

Can you explain why this is the case? If you want all types of the size `.xlarge`, you could just get the instance type names from an `aws ec2 describe-instance-types` call and then include these in the settings using some programatic means.

> A simple globbing or regex support

I'm not really a fan of opening up this settings schema to regex-based pattern matching. It doesn't seem particularly explicit in its design and implementation and seems overly-complicated given what we are trying to implement.

universam1:
> include these in the settings using some programatic m

The instances we permit the devs. to request or allow Karpenter to choose, sum up at roughly 600 types and versions. Say anything >1 core, >3GB, incl. amd64 or arm64. Having to manage that, means updating a huge list and/or keeping hundreds of developers informed about what instances are whitelisted or why not.
Offloading this type decision is the selling point of Karpenter, and we love it for that. Feels like this undermines the whole concept.

> seems overly-complicated given what we are trying to implement

Could you explain please why this is complicated to implement? I believe few lines of code would suffice. In contrast coding an API or generator for Karpenter definition is a huge effort to implement.
It would mean that Karpenter now submits to an external API that it would have to comply with, which could easily become an unpleasant dependency for the project!?

jonathan-innis:
I'm less a fan of the regex solution and might be swayed toward doing a requirements-based solution that allows you to create disparate objects that contain requirements (of which instance-type name can be one) and capacity details with a merge semantic. Would this suite the case you are describing and be more maintainable for you?

```json
[
   {
      "requirements":[
         {
            "key":"karpenter.k8s.aws/instance-size""operator":"In""values":[
               "large"
            ]
         }
      ],
      "capacity":{
         "fuse":"1",
         "other-extended-resource":"4"
      }
   }
]
```

In the ConfigMap, it would look like

```yaml
"aws.instanceTypeOverrides": "[{"requirements":[{"key":"karpenter.k8s.aws/instance-size","operator":"In","values":["large"]}],"capacity":{"fuse":"1","other-extended-resource":"4"}}]"
```

universam1:
> might be swayed toward doing a requirements-based solution

That's a great one! I think the list of instance sizes is not that of an issue to manage. In fact I could imagine a hack to invert to a "NotIn: small" if we really want to permit all sizes but those irrelevant ones.

Thank you!
How do you like us to proceed?

jonathan-innis:
>  I think the list of instance sizes is not that of an issue to manage

On this same train of thought, what if there was a `resources` field in the Provisioner API that allowed you to specify the amount of extended resource that a Node that a Provisioner launched would have (we would assume this amount for inflight scheduling until the Node resources actually registered), do you feel like this type of solution would also work well for you? I'm assuming if you have a set of requirements that are something like "only use instance types that aren't of the small variety" that this would map naturally to Provisioner requirements, where you could add these extended resources.

```yaml
kind: Provisioner
metadata:
  name: default
spec:
  resources:
    requests:
      fuse: "2" # We assume this much until the node registers
      some.example.com/extended-resource: "500Mi" # We assume this much until the node registers
```

universam1:
> what if there was a resources field

I like this semantic actually most, a single authoritative place for Karpenter config! 👏🏻 

> we would assume this amount for inflight scheduling until the Node resources actually registered

AFAIR Karpenter does not actively assign pods to nodes any more like it used to, leaving that to kubescheduler? Great, we'll only have to ensure that the `smarter-device-manager` daemonset is up in time and has patched the custom resources on the nodes, _before_ Karpenter enters a new loop for unscheduled pods.... I guess we can tame such a flapping via `startupTaints` to have a defined cool down delay!? I think this should work fine!

> I'm assuming if you have a set of requirements that are something like "only use instance types that aren't of the small variety" that this would map naturally to Provisioner requirements

Indeed, main part of our specs contain
```yaml
spec:
  requirements:
    - key: karpenter.k8s.aws/instance-size
      operator: NotIn
      values:
        - nano
        - micro
        - small
```
Great idea!

jonathan-innis:
> before Karpenter enters a new loop for unscheduled pods

We handle this the concept of "node initialization". We wait for the node to go ready and have resources registered before we consider new pods for scheduling to new nodes so you shouldn't have to handle this cool down period yourself. Once we are aware that a resource _should_ appear on the node, we wait until we see it before we consider it empty or not eligible to schedule pods.

haarchri:
thanks for this RFC - i wonder what is the current status ?

jonathan-innis:
@haarchri We're keeping this open to keep the conversation going, since this is the largest item that is currently tracking and creating pointers to all the open issues. There's been a lot of discussion among the maintainers and contributors around how we should attack a lot of these disparate problems, so the RFC is still unsettled and probably won't be until the project settles and merges a lot of the other current feature work:
1. https://github.com/aws/karpenter-core/pull/176
2. https://github.com/aws/karpenter-core/pull/221

jonathan-innis:
As mentioned [here](https://github.com/aws/karpenter/pull/2390#issuecomment-1456581131), there is still a lot of uncertainty on the plan of attack with how to handle custom instance type settings. The issues mentioned in this initial RFC will be addressed at a later time, with a rewritten RFC. Closing this for now.

universam1:
@ellistarn @jonathan-innis Trying to get a feeling for the time to expect for your long-term solution. We are wondering if we should maintain a fork with an intermediate solution for the time being or if there are other options possible?
As an idea, could we guard this  #2390 implementation behind an alpha flag to unblock those users willing to take the risk for the time being? Would that be a feasable approach?

PR #2400, opened 2022-08-30
docs: fix bottlerocket node template for GPU volume size

lamdor:
**Description**

the GPU bottlerocket AMI requires at least 3GB. otherwise you run into:

> 2022-08-30T13:27:52.897Z        ERROR   controller.provisioning Provisioning failed, launching node, creating cloud provider instance, with fleet error(s), InvalidBlockDeviceMapping: Volume of size 3GB is smaller than  snapshot 'snap-0c5c999fe56c48b6f', expect size >= 4GB        {"commit": "639756a"}

The normal bottlerocket AMI size is still 2GB, just GPU is a bit bigger.

There was also a typo with the volume size value.

* N/A

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/examples/provisioner/bottlerocket.yaml
+++ b/examples/provisioner/bottlerocket.yaml
@@ -23,7 +23,7 @@ spec:
     - deviceName: /dev/xvda
       ebs:
         volumeType: gp3
-        volumeSize: 2G # if using GPU, replace with 3G since the GPU bottlerocket AMI requires at least 3G
+        volumeSize: 2G # if using GPU, replace with 4G since the GPU bottlerocket AMI requires at least 4G
         deleteOnTermination: true
     - deviceName: /dev/xvdb
       ebs:
--- a/examples/provisioner/bottlerocket.yaml
+++ b/examples/provisioner/bottlerocket.yaml
@@ -23,7 +23,7 @@ spec:
     - deviceName: /dev/xvda
       ebs:
         volumeType: gp3
-        volumeSize: 2Gi
+        volumeSize: 2G # if using GPU, replace with 3G since the GPU bottlerocket AMI requires at least 3G
         deleteOnTermination: true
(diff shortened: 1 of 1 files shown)


examples/provisioner/bottlerocket.yaml
```diff
@@ -23,7 +23,7 @@ spec:
     - deviceName: /dev/xvda
       ebs:
         volumeType: gp3
-        volumeSize: 2Gi
+        volumeSize: 2G # if using GPU, replace with 4G since the GPU bottlerocket AMI requires at least 4G
```
tzneal, line 26:
```suggestion
        volumeSize: 4Gi
```
Since 4Gi works consistently, I'd be fine just changing it to 4GI by default to reduce the chances people running into this.

FernandoMiguel, line 26: Gi is the correct format, right?
the typo is in the one with 20GiBs

tzneal: They both work, G is a power of 10 version and Gi is the power of 2 version. https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

FernandoMiguel: right... but helps if we align with what AWS EC2 and AMIs provide.
If EBSs with 20 gigabytes are 20Gi, that is what I feel we should use, since if we were to use 20G, the volume would be smaller than the AMI and fail.

Similar for RAM. If an ec2 has 16 gigabytes of RAM, using 16G is better than Gi ?

and yeah, I'm aware that HDD and SSD use base 2 and base 10 each.... _OEMs_ ....

FernandoMiguel (approved):

bwagner5 (approved):
lgtm!

PR #2404, opened 2022-08-30
feat: Introduce `InstanceType` CRD for overriding 

### LINKED ISSUE #1900: Karpenter doesnt support custom resources requests/limit

jonathan-innis:
Fixes #1900, Fixes #2161 

**Description**

- Add `InstanceType` CR for specifying per-instance type settings. For now, you can override the allocatable resources to specify extended resources that have corresponding [custom device plugins](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/) for specific instance types.

```yaml
apiVersion: karpenter.sh/v1alpha1
kind: InstanceType
metadata:
  name: c5.large
spec:
  resources:
    hardware.vendor.com/resource: 2
    hardware.vendor.com/other-resource: 10
```

- [x] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Add `InstanceType` CRD for customizing instance type resources
```


pkg/apis/instancetype/v1alpha1/instancetype.go
```diff
@@ -0,0 +1,50 @@
...
+
+// InstanceTypeSpec is the instance type setting override specification
+// for specifying custom values on a per-instance type basis for scheduling and
+// launching of nodes
+type InstanceTypeSpec struct {
+	// Resources contains a map of allocatable resources for the instance type
+	// used by the scheduler. This resource list can contain known resources (cpu, memory, etc.)
+	// or it may also contain unknown custom device resources for custom device plugins
+	// +optional
+	Resources v1.ResourceList `json:"resources,omitempty"`
```
ellistarn, line 30: Thoughts on calling this `Allocatable` for parity with v1.Node?

universam1:
@jonathan-innis @ellistarn 
Thank you for this proposed solution, but we have serious concerns this does not solve #1900 for us as it requires a definition for each `kind: InstanceType`!?
We understand that the core idea of Karpenter is to solve the problem of  preselected instance types to dynamically select from the magnitude of various types available in AWS. This CRD however defeats this idea and thus does not provide the quality of #2161 - in contrast we see an extreme overhead here.

The additional complexity introduced here that multiplies with instance type times instance size results into thousands of variations, which is unmanageable!

jonathan-innis:
> @jonathan-innis @ellistarn Thank you for this proposed solution, but we have serious concerns this does not solve #1900 for us, let me explain.
> 
> ### unmanageable?
> We understand that the core idea of Karpenter is to solve the problem of preselected instance types to dynamically select from the magnitude of various types available in AWS. This CRD however defeats this idea and thus does not provide the quality of #2161 - in contrast we see an extreme overhead here.
> 
> The additional complexity introduced here that multiplies with _instance type_ **times** _instance size_ results into thousands of variations, which is unmanageable!
> 
> ### divergence with `kind: Provisioner`
> This concept requires a manual, logical sync with `kind: Provisioner` to pin Karpenter to _only use those defined instances_ of `kind: InstanceType`, which is prone to drift. There is the logical connection of the choice of types with the definition of their features - but this is not given as there is no coupling btw. those two independent CRD.
> 
> ### idea
> It looks like `kind: InstanceType` would better be part of extended `kind: Provisioner` CRD to be closely coupled, and then a missing (regex) selector to solve the unmanageability would be solved as well. Or, like with `awsnodetemplates.karpenter.k8s.aws` reference it like `providerRef` Thank you for your consideration

Hi @universam1, thanks for your feedback. I'm interested to know how you would deal with the difference in resource capacity for extended resources across different instance types. My assumption is that all instance types wouldn't have the same number of extended resources so we didn't want to couple it too closely to the provisioner as it might also cause a provisioner explosion.

Let me know if this understanding is incorrect, though. I'm curious to know how much commonality would exist across your instance types for extended resources and if an overlay concept using something like a label selector would be better.

universam1:
Hi @jonathan-innis , thank you for reaching out and acquiring feedback, appreciated!

> My assumption is that all instance types wouldn't have the same number of extended resources 

Let me picture our two use cases:
1. For a large videoconferencing system (up to several hundreds of nodes per cluster) we leverage the ALSA `/dev/snd` device exposed by the kernel (module). Obviously the amount of these kernel devices available per host are not in relation to an instance size or type and probably not even on CPU architecture, but rather on linux kernel config. For this amount of instances Karpenter is the answer, overcomimg AWS provider limits of the availablity of a certain type, so the loosely tied flexibilty of  types and sizes possible by Karpenter is what we have been looking for. Now having to constrain on few types and sizes by the current manual declaration cancels the whole benefits out.
2. We have a container build farm that [requires `/dev/fuse` device](https://www.redhat.com/sysadmin/podman-inside-kubernetes). The number of fuse devices available is also not tied to instance type or size, per default 1000 or more. Also here it is impossible for us to define all instance types times instance sizes to match all provider offers.

Granted there are other use cases that I might not be aware of, but I am struggling to picture a case where a one-to-one relation of this config would shine. In our case it would be a management nightmare so to speek. Thank you for considering!

jonathan-innis:
Hi @universam1. We are going to close this PR for now. I'm going to migrate our discussion to the RFC PR (#2404) so that we can keep this feedback in a common place

PR #2411, opened 2022-08-31
Fix ami selector example

dschaaff:
**Description**

The example files in the repo incorrectly specify the amiSelector when using ami ids.


bwagner5 (approved):
Nice catch!

PR #2415, opened 2022-08-31
fix: increase the liveness/readiness probe timeouts

tzneal:
**Description**

The healthchecks check locks which cluster state may hold for
a period of time.  We should solve that, but for now prevent
the controller from restarting if you get unlucky during a
large scale-up and the health checks hit when the lock is held.

* Discovered via an E2E test, I could reproduce locally and ran multiple times with the new deployment values with no reproductions.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


bwagner5 (approved):
nice!

PR #2429, opened 2022-08-31
fix: Remove comment after wait command

eichlerr:
The comment after the wait command causes the `helm upgrade` command to fail with the following message:
> Error: "helm upgrade" requires 2 arguments

Removing the comment fixes this error.

Fixes # NA

**Description**
This is a fix to the `helm upgrade` script in the getting started documentation. The documentation doesn't make it clear that users should remove this comment before running the command/script, otherwise it will fail.

* Removing the comment after the last command in the script resolved my issue.

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
This is a change to the Getting Started document where developers could experience issues trying to install the Helm chart needed for Karpenter.
```


jonathan-innis (approved):
lgtm 🚀

PR #2444, opened 2022-09-01
feat: Added `evictionHard`, `kubeReserved` and `podsPerCore` to Provisioner

### LINKED ISSUE #2129: Allow passing --kube-reserved/system-reserved/eviction-hard flags

jonathan-innis:
Fixes #2129 

**Description**

- Adds the ability to pass `evictionHard`, `kubeReserved` and `podsPerCore` in Provisioner spec

```yaml
kind: Provisioner
metadata:
  name: default
spec:
  ...
  kubeletConfiguration:
    podsPerCore: 3
    kubeReserved:
      cpu: 100m
      memory: 50Mi
    evictionHard:
      memory.available: 10%
```

* `make test`
* `TEST_FILTER="integration" make e2etests`
* Manual validation

- [x] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Added `evictionHard`, `kubeReserved` and `podsPerCore` to Provisioner
```


pkg/cloudprovider/aws/amifamily/bootstrap/eksbootstrap.go
```diff
@@ -59,16 +59,21 @@ func (e EKS) Script() (string, error) {
...
-	// Backwards compatibility for AWSENILimitedPodDensity flag
 	if e.KubeletConfig != nil && e.KubeletConfig.MaxPods != nil {
 		userData.WriteString(" \\\n--use-max-pods false")
 		kubeletExtraArgs += fmt.Sprintf(" --max-pods=%d", ptr.Int32Value(e.KubeletConfig.MaxPods))
 	} else if !e.AWSENILimitedPodDensity {
 		userData.WriteString(" \\\n--use-max-pods false")
 		kubeletExtraArgs += " --max-pods=110"
 	}
+	if e.KubeletConfig != nil && e.KubeletConfig.PodsPerCore != nil {
+		kubeletExtraArgs += fmt.Sprintf(" --pods-per-core=%d", ptr.Int32Value(e.KubeletConfig.PodsPerCore))
```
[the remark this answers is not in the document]
tzneal: Found the old issue, https://github.com/aws/karpenter/pull/1520/ , it's only the bootstrap args that it causes problems with.

tzneal (dismissed):
lgtm

pkg/cloudprovider/aws/amifamily/bootstrap/eksbootstrap.go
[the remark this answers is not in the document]
bwagner5: ^^ yep, kubelet args will work fine w/ `=`

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -319,19 +312,65 @@ func (i *InstanceType) kubeReservedResources(pods int64) v1.ResourceList {
...
+	}
 	return resources
 }
 
-func (i *InstanceType) miscResources(vmMemOverhead float64) v1.ResourceList {
+func (i *InstanceType) evictionThreshold(kc *v1alpha5.KubeletConfiguration, vmMemoryOverhead resource.Quantity) v1.ResourceList {
+	overhead := v1.ResourceList{
+		v1.ResourceMemory: resource.MustParse("100Mi"),
+	}
+	if kc != nil && kc.EvictionHard != nil {
```
bwagner5, line 325: might be better in this case to invert this logic and return overhead early rather than doing 4-nested ifs

pkg/cloudprovider/aws/instancetype.go
```diff
@@ -319,19 +312,65 @@ func (i *InstanceType) kubeReservedResources(pods int64) v1.ResourceList {
...
+	case kc != nil && kc.MaxPods != nil:
+		mp = ptr.Int64(int64(ptr.Int32Value(kc.MaxPods)))
+	case !injection.GetOptions(ctx).AWSENILimitedPodDensity:
+		mp = ptr.Int64(110)
+	default:
+		mp = ptr.Int64(i.eniLimitedPods())
+	}
+
+	if kc != nil && ptr.Int32Value(kc.PodsPerCore) > 0 && amiFamily.PodsPerCoreEnabled() {
+		mp = ptr.Int64(lo.Min([]int64{int64(ptr.Int32Value(kc.PodsPerCore)) * ptr.Int64Value(i.VCpuInfo.DefaultVCpus), ptr.Int64Value(mp)}))
```
bwagner5, line 371: This shouldn't use `DefaultVCpus` since those are virtual CPUs and the kubelet uses physical cores. You can use `DefaultCores` instead.  

https://github.com/kubernetes/kubernetes/blob/442574f3a7321ff0f34dec9bae67f28651dbabd2/pkg/kubelet/nodestatus/setters.go#L282

https://github.com/kubernetes/kubernetes/blob/55e1d2f9a73a0c78fc0c266236c3e31261635b90/pkg/kubelet/cm/cpumanager/topology/topology.go#L221

pkg/test/options.go
```diff
@@ -0,0 +1,28 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import "github.com/aws/karpenter/pkg/utils/options"
+
+func Options() options.Options {
```
bwagner5, line 19: Why is this broken out now? A lot of these are AWS specific flags, seems better to keep in the specific test suite and just reset or deep copy on each execution.

bwagner5 (changes requested):
nice! A few comments but mainly one issue in the podsPerCore calc.

bwagner5:
Does it make sense to support soft eviction thresholds as well?

pkg/test/options.go
jonathan-innis: Good catch, I forgot that these could be highly AWS-specific

jonathan-innis:
> Does it make sense to support soft eviction thresholds as well?

It probably does, I figured we could decouple these for now, but we should probably follow-up this PR with support for `evictionSoft` and respect the higher of `evictionHard` and `evictionSoft` for scheduling

pkg/cloudprovider/aws/instancetype.go
jonathan-innis: This actually isn't the case. Weirdly enough, kubelet is using `DefaultVCPus` from the instance when testing with instances that have a differing number of cpus and cores.

bwagner5 (dismissed):
lgtm

bwagner5 (approved):
lgtm

PR #1752, opened 2022-05-03
Remove extra steps from release GA

spring1843:
**1. Issue, if available:**

**2. Description of changes:**
This is very similar to nightlies in what it does, hence it doesn't require the toolchain. The toolchain inclusion was failing because it would require Go runtime.

**3. How was this change tested?**

**4. Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: *link to issue*
- [x] No

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2457, opened 2022-09-02
chore: Enable mod upgrade of `controller-runtime`, `knative`, and `logr`

jonathan-innis:
Fixes # 

**Description**

- Enable mod upgrade of `controller-runtime`, `knative`, and `logr`

* `make test`
* `make build`
* Manual Validation of Build

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```


jonathan-innis:
This PR is blocked until [knative](https://github.com/knative/pkg) upgrades its packages to support `v0.24.x` of kubernetes api packages

olemarkus:
Do you actually need v0.24.x support for the API in knative? As far as I know, Karpenter doesn't use any 1.25.0 resources.
I used v0.25.x for #2354 and that works just fine.

jonathan-innis:
> Do you actually need v0.24.x support for the API in knative? As far as I know, Karpenter doesn't use any 1.25.0 resources.
> I used v0.25.x for #2354 and that works just fine.

It's `client-go` support, not necessarily the server K8s version. Your PR #2354 only upgrades the `client-go` package (and other k8s apis) to `v0.22.13`. There is an incompatibility with an interface that Knative implements starting in `v0.24.0` of K8s api packages.

jonathan-innis:
To be more explicit, the crux of the issue is the following:
1. We can't upgrade to `v0.11.x` of `controller-runtime` due to [#1752](https://github.com/kubernetes-sigs/controller-runtime/issues/1752)
2. Upgrade to `v0.12.x` has a minimum requirement of `v0.24.2` of the k8s api client packages
3. Upgrade of `v0.24.2` of k8s api client packages breaks compatability with knative because the latest knative version does not support the `StorageV1Interface`

```console
# knative.dev/pkg/client/injection/kube/client
../../go/pkg/mod/knative.dev/pkg@v0.0.0-20220826162920-93b66e6a8700/client/injection/kube/client/client.go:13323:9: cannot use &wrapStorageV1{…} (value of type *wrapStorageV1) as type "k8s.io/client-go/kubernetes/typed/storage/v1".StorageV1Interface in return statement:
        *wrapStorageV1 does not implement "k8s.io/client-go/kubernetes/typed/storage/v1".StorageV1Interface (missing CSIStorageCapacities method)
Error: failed to publish images: error building "ko://github.com/aws/karpenter/cmd/controller": exit status 2
```

olemarkus:
Ah. Thanks for the info. I now remember I could not go up as far as I like on k8s deps due to knative and just had to go for one that had the entire `policy/v1`. Sorry for the noise.

jonathan-innis:
Knative package upgrade being tracked [here](https://github.com/knative/pkg/pull/2586)

jonathan-innis:
Knative package is upgraded 🎉

pkg/controllers/controllers.go
```diff
@@ -114,6 +114,7 @@ func Initialize(injectCloudProvider func(context.Context, cloudprovider.Options)
 		Scheme:                     scheme,
 		MetricsBindAddress:         fmt.Sprintf(":%d", opts.MetricsPort),
 		HealthProbeBindAddress:     fmt.Sprintf(":%d", opts.HealthProbePort),
+		BaseContext:                func() context.Context { return ctx },
```
tzneal, line 117: When I had looked into this previously, I wrote this code:
```go
BaseContext:            newRunnableContext(config, opts, logging.FromContext(ctx))
```
```go
func newRunnableContext(config *rest.Config, options options.Options, logger zap.SugaredLogger) func() context.Context {
       return func() context.Context {
               ctx := context.Background()
               ctx = logging.WithLogger(ctx, logger)
               ctx = injection.WithConfig(ctx, config)
               ctx = injection.WithOptions(ctx, options)
               return ctx
       }
}
```

I don't think it matters at the moment, but it seems to be safer for the BaseContext func to return new contexts.

jonathan-innis: Can you explain why you think this is safer? I'm failing to think of issues with passing in the same context object into the func

tzneal: It's just that BaseContext is a function and not just a context and that the default implementation returns a new context via `context.Background()` for each call. Combining that with the behavior of the cancels cancelling child contexts:
```go
// The WithCancel, WithDeadline, and WithTimeout functions take a
// Context (the parent) and return a derived Context (the child) and a
// CancelFunc. Calling the CancelFunc cancels the child and its
// children, removes the parent's reference to the child, and stops
// any associated timers. Failing to call the CancelFunc leaks the
// child and its children until the parent is canceled or the timer
// fires. The go vet tool checks that CancelFuncs are used on all
// control-flow paths.
```

A cancel on that top-level context would cancel all of the runner's contexts.  Maybe that's ok? It seemed safer to use new contexts than figure that out.

jonathan-innis: You're completely correct. Since they are just calling the `baseContext` function to get the parent context for the `context.WithCancel` used by the manager runner

tzneal (dismissed):
lgtm

tzneal (approved):
lgtm

PR #2470, opened 2022-09-07
While using Custome AMI, Controller Role cannot describe images

iflan7744:
Fixes # 

**Description**
While using Launch Template with Custom AMI the controller role cannot describe the AMI image ID and create Instance out of it.

I include ec2:DescribeImages to KarpenterControllerPolicy and Karpenter created the Instance with custom AMI.
*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/website/content/en/v0.16.1/getting-started/migrating-from-cas/scripts/step05-controller-iam.sh
+++ b/website/content/en/v0.16.1/getting-started/migrating-from-cas/scripts/step05-controller-iam.sh
@@ -26,6 +26,7 @@ echo '{
             "Action": [
                 "ssm:GetParameter",
                 "iam:PassRole",
+                "ec2:DescribeImages",
                 "ec2:RunInstances",
                 "ec2:DescribeSubnets",
                 "ec2:DescribeSecurityGroups",


jonathan-innis:
Thanks @iflan7744 for your proposed contribution! We are thinking it might be better to rely on cloudformation rather than having two separate places in the docs where we manage policy definitions, to point to the same location across the docs. i.e. we have a policy definition defined in cloudformation [here](https://github.com/aws/karpenter/blob/main/website/content/en/preview/getting-started/getting-started-with-eksctl/cloudformation.yaml). 

If we could reference this cloudformation policy (similar to how it's referenced [here](https://karpenter.sh/preview/getting-started/getting-started-with-eksctl/#create-the-karpenternode-iam-role)) and then add the detail of the OIDC trust policy relationship, we could cut down on docs duplication quite a bit!

jonathan-innis (approved):
LGTM 🚀

jonathan-innis:
Unblocking some users by taking this change in

PR #2475, opened 2022-09-08
feat: reconcile tags with current node template (wip)

blakepettersson:
NB: This is a draft at the moment since I still need to work on adding tests as well as determining what the actual API for reconciling tags will actually look like. This is more of an inquiry to see if this direction makes sense at all. This is intended for updating the EC2 tags on all nodes affected by a change in a node template, without needing to re-provision the nodes.

Fixes # 

**Description**

The idea here is that any update on an `AWSNodeTemplate` will trigger a node reconcile. In `controller.go`, we lookup all provisioners which have a matching `ProviderRef` (this means that in order for this to work `ProviderRef` would need to be used by the provisioner(s) instead of the embedded version).

We then go through each matching provisioner and enqueue any nodes making use of said provisioner and for each node we then lookup the existing tags on the EC2 instance and set its tags to the latest values found in the node template, and removes any tags which are not a part of the node template.

* Running this in an EKS cluster

- [x] Yes, need to add docs if this gets accepted
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/apis/provisioning/v1alpha5/labels.go
+++ b/pkg/apis/provisioning/v1alpha5/labels.go
@@ -35,6 +35,7 @@ var (
 	DoNotConsolidateNodeAnnotationKey = KarpenterLabelDomain + "/do-not-consolidate"
 	EmptinessTimestampAnnotationKey   = Group + "/emptiness-timestamp"
 	TerminationFinalizer              = Group + "/termination"
+	TagsVersionKey                    = Group + "/tags-version"
 
 	LabelCapacityType    = KarpenterLabelDomain + "/capacity-type"
 	LabelNodeInitialized = KarpenterLabelDomain + "/initialized"
@@ -46,7 +47,7 @@ var (
 		KarpenterLabelDomain,
 	)
 
-	// LabelDomainException are sub-domains of the RestrictedLabelDomains but allowed because
+	// LabelDomainExceptions are sub-domains of the RestrictedLabelDomains but allowed because
 	// they are not used in a context where they may be passed as argument to kubelet.
 	LabelDomainExceptions = sets.NewString(
 		"kops.k8s.io",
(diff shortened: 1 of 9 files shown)


njtran:
Hey @blakepettersson. Thanks for the PR. We're currently working on an upgrade mechanism to handle cases of spec-drift from the nodes and provisioners. In all, modifying the tags in-place can be a scary operation to default to especially if there are any applications or systems that depend on these tags. My thought that in all cases of spec drift, we should recreate the node before we terminate it.

There's more to come on the upgrades, but we're currently looking at doing AMIs first and hoping that we can get any other upgrade triggers while we're at it. Can you explain a little more about why you'd want to do an essentially in-place upgrade with tags?

blakepettersson:
> Hey @blakepettersson. Thanks for the PR. We're currently working on an upgrade mechanism to handle cases of spec-drift from the nodes and provisioners. In all, modifying the tags in-place can be a scary operation to default to especially if there are any applications or systems that depend on these tags. My thought that in all cases of spec drift, we should recreate the node before we terminate it.
>
> There's more to come on the upgrades, but we're currently looking at doing AMIs first and hoping that we can get any other upgrade triggers while we're at it

I'm happy for _any_ mechanism which would allow the desired state of the node templates to align with the state of the nodes. :smile:  I don't really have any better reason for having in-place updates of tags than the fact that for such a "small" thing, it seemed a bit scary/overkill to need to re-provision a node for that purpose. 

If you're guys are working on something which will supercede this, I'm happy to close this and call it a day. Do you happen to have an ETA for when this upgrade mechanism would be in place?

njtran:
Thanks! For the ETA, I don't have any dates I can share with you, but working on the design currently, and will open it up for review with the community when it's ready.

njtran:
Hey @blakepettersson, I cut a PR with the design doc on node AMI upgrades. This should include a way to eventually consider reconciling tags on nodes. https://github.com/aws/karpenter/pull/2569. As discussed above, I'll close this, but feel free to leave a review on the linked PR.

PR #2488, opened 2022-09-09
feat: consolidation enhancements

tzneal:
**Description**
- Monitor for deployment/stateful set updates when reducing the stabilitization window
- Consider the last node creation time as part of the stabilization window
- Reduce some more log spam
- Add a FAQ entry explaining why sometimes an extra node appears when updating a deployment image

* Unit tested and deployed to EKS.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -160,7 +160,7 @@ func (p *LaunchTemplateProvider) ensureLaunchTemplate(ctx context.Context, optio
 	} else if len(output.LaunchTemplates) != 1 {
 		return nil, fmt.Errorf("expected to find one launch template, but found %d", len(output.LaunchTemplates))
 	} else {
-		if p.cm.HasChanged("launchtemplate"+name, name) {
+		if p.cm.HasChanged("launchtemplate-"+name, name) {
 			logging.FromContext(ctx).Debugf("Discovered launch template %s", name)
 		}
 		launchTemplate = output.LaunchTemplates[0]
--- a/pkg/cloudprovider/aws/launchtemplate.go
+++ b/pkg/cloudprovider/aws/launchtemplate.go
@@ -160,7 +160,9 @@ func (p *LaunchTemplateProvider) ensureLaunchTemplate(ctx context.Context, optio
 	} else if len(output.LaunchTemplates) != 1 {
 		return nil, fmt.Errorf("expected to find one launch template, but found %d", len(output.LaunchTemplates))
 	} else {
-		logging.FromContext(ctx).Debugf("Discovered launch template %s", name)
+		if p.cm.HasChanged("launchtemplate"+name, name) {
+			logging.FromContext(ctx).Debugf("Discovered launch template %s", name)
+		}
 		launchTemplate = output.LaunchTemplates[0]
 	}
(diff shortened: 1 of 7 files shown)


pkg/cloudprovider/aws/launchtemplate.go
```diff
@@ -160,7 +160,9 @@ func (p *LaunchTemplateProvider) ensureLaunchTemplate(ctx context.Context, optio
 	} else if len(output.LaunchTemplates) != 1 {
 		return nil, fmt.Errorf("expected to find one launch template, but found %d", len(output.LaunchTemplates))
 	} else {
-		logging.FromContext(ctx).Debugf("Discovered launch template %s", name)
+		if p.cm.HasChanged("launchtemplate"+name, name) {
```
jonathan-innis, line 163: nit: maybe a dash after the launchtemplate if we ever need to debug the keys

pkg/controllers/consolidation/controller.go
```diff
@@ -611,7 +632,7 @@ func (c *Controller) statefulSetsReady(ctx context.Context) bool {
 			// unspecified defaults to 1
```
jonathan-innis, line 632: Should the `List` failure on line 625 be `return false` instead?

website/content/en/preview/faq.md
```diff
@@ -163,3 +163,9 @@ error: error validating "provisioner.yaml": error validating data: ValidationErr
 ```
 
 The `startupTaints` parameter was added in v0.10.0.  Helm upgrades do not upgrade the CRD describing the provisioner, so it must be done manually. For specific details, see the [Upgrade Guide]({{< ref "./upgrade-guide/#upgrading-to-v0100" >}})
+
+## Consolidation
+
+### Why do I sometimes see an extra node get launched when updating a deployment that then remains empty and is removed?
```
jonathan-innis, line 169:
```suggestion
### Why do I sometimes see an extra node get launched when updating a deployment that remains empty and is later removed?
```

website/content/en/preview/faq.md
```diff
@@ -163,3 +163,9 @@ error: error validating "provisioner.yaml": error validating data: ValidationErr
...
+
+Consolidation packs pods tightly onto nodes which can leave little free allocatable CPU/memory on your nodes.  If a deployment uses a deployment strategy with a non-zero `maxSurge`, such as the default 25%, those surge pods may not have anywhere to run. In this case Karpenter will launch a new node so that the surge pods can run and then remove it soon after if it's not needed.
```
jonathan-innis, line 171:
```suggestion
Consolidation packs pods tightly onto nodes which can leave little free allocatable CPU/memory on your nodes.  If a deployment uses a deployment strategy with a non-zero `maxSurge`, such as the default 25%, those surge pods may not have anywhere to run. In this case, Karpenter will launch a new node so that the surge pods can run and then remove it soon after if it's not needed.
```

jonathan-innis (approved):
Small nits, otherwise lgtm 🚀

pkg/controllers/consolidation/controller.go
tzneal: Yup, missed one

jonathan-innis (approved):
:rocket

PR #2518, opened 2022-09-16
feat: Add LB exclusion label when deleting node

DWSR:
Fixes # N/A

**Description**

Currently, when Karpenter drains and then deletes a Node from the
cluster, if that node is registered in a Target Group for an ALB/NLB the
corresponding EC2 instance is not removed. This leads to the potential
for increased errors when deleting nodes via Karpenter.

In order to help resolve this issue, this change adds the well-known
`node.kubernetes.io/exclude-from-external-balancers` label, which will
case the AWS LB controller to remove the node from the Target Group
while Karpenter is draining the node. This is similar to how the AWS
Node Termination Handler works (see
aws/aws-node-termination-handler#316).

In future, Karpenter might be enhanced to be able to wait for a
configurable period before deleting the Node and terminating the
associated instance as currently there's a race condition between the
Pods being drained off of the Node and the EC2 instance being removed
from the target group.

* Running Karpenter against a manual cluster that had the AWS LB Controller installed and observing the de-registration in progress.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/termination/suite_test.go
+++ b/pkg/controllers/termination/suite_test.go
@@ -103,6 +103,25 @@ var _ = Describe("Termination", func() {
 			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(node))
 			ExpectNotFound(ctx, env.Client, node)
 		})
+		It("should exclude nodes from load balancers when terminating", func() {
+			// This is a kludge to prevent the node from being deleted before we can
+			// inspect its labels
+			podNoEvict := test.Pod(test.PodOptions{
+				NodeName: node.Name,
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations:     map[string]string{v1alpha5.DoNotEvictPodAnnotationKey: "true"},
+					OwnerReferences: defaultOwnerRefs,
+				},
+			})
+
+			ExpectApplied(ctx, env.Client, node, podNoEvict)
+
+			Expect(env.Client.Delete(ctx, node)).To(Succeed())
+			node = ExpectNodeExists(ctx, env.Client, node.Name)
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(node))
+			node = ExpectNodeExists(ctx, env.Client, node.Name)
+			Expect(node.Labels[v1.LabelNodeExcludeBalancers]).Should(Equal("karpenter"))
+		})
 		It("should not evict pods that tolerate unschedulable taint", func() {
 			podEvict := test.Pod(test.PodOptions{NodeName: node.Name, ObjectMeta: metav1.ObjectMeta{OwnerReferences: defaultOwnerRefs}})
 			podSkip := test.Pod(test.PodOptions{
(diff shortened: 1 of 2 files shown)


bwagner5:
Thanks for this! Would you be able to add a test as well within the termination suite_test.go?

pkg/controllers/termination/terminate.go
```diff
@@ -61,6 +61,14 @@ func (t *Terminator) cordon(ctx context.Context, node *v1.Node) error {
 	// 2. Cordon node
 	persisted := node.DeepCopy()
 	node.Spec.Unschedulable = true
+	// Handle nil map
+	if node.Labels == nil {
+		node.Labels = map[string]string{
+			"node.kubernetes.io/exclude-from-external-load-balancers": "karpenter",
```
jonathan-innis, line 67: Can use constant [`v1.LoadNodeExcludeLoadBalancers`](https://github.com/kubernetes/api/blob/master/core/v1/well_known_labels.go#L67) instead

pkg/controllers/termination/terminate.go
```diff
@@ -61,6 +61,14 @@ func (t *Terminator) cordon(ctx context.Context, node *v1.Node) error {
 	// 2. Cordon node
 	persisted := node.DeepCopy()
 	node.Spec.Unschedulable = true
+	// Handle nil map
+	if node.Labels == nil {
```
jonathan-innis, line 65: Personal preference: you can take it or leave it. I prefer,

```go
if node.Labels == nil {
  node.Labels = map[string]string{}
}
node.Labels["node.kubernetes.io/exclude-from-external-load-balancers"] = "karpenter"
```

so that the map indexing isn't repeated

jonathan-innis (changes requested):

pkg/controllers/termination/terminate.go
DWSR: Nice, didn't know that existed!

pkg/controllers/termination/terminate.go
DWSR: Good call. Updated

DWSR:
> Thanks for this! Would you be able to add a test as well within the termination suite_test.go?

Added, but I'm not entirely happy with the fact that I'm (ab)using a `do-not-evict` Pod. Feel free to suggest a better way to perform the test

pkg/controllers/termination/terminate.go
ellistarn: optional syntactic sugar to do this.

`node.Labels = lo.Assign(node.Labels, map[string]string{"key":"value"})`

pkg/controllers/termination/suite_test.go
```diff
@@ -103,6 +103,25 @@ var _ = Describe("Termination", func() {
 			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(node))
 			ExpectNotFound(ctx, env.Client, node)
 		})
+		It("should exlude nodes from load balancers when terminating", func() {
```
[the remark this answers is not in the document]
DWSR: Fixed, thank you for catching this. :)

PR #2527, opened 2022-09-17
test: Add cluster name to upgrade crd task

jonathan-innis:
Fixes # 

**Description**

- Fix miss in Tekton `upgrade-crd` task

*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/test/suites/common/setup.yaml
+++ b/test/suites/common/setup.yaml
@@ -103,4 +103,5 @@ spec:
         --set clusterName=$(params.cluster-name) \
         --set clusterEndpoint=$(cat /root/.kube/config | grep server | awk '{print $2}') \
         --set aws.defaultInstanceProfile=KarpenterNodeInstanceProfile-$(params.cluster-name) \
+        --set controller.batchIdleDuration=10s \
         --wait
--- a/test/suites/common/upgrade-crds.yaml
+++ b/test/suites/common/upgrade-crds.yaml
@@ -7,6 +7,8 @@ spec:
   description: |
     Upgrade Karpenter CRDs
   params:
+    - name: cluster-name
+      description: Name of the cluster under test.
     - name: git-ref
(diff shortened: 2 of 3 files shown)


test/suites/common/setup.yaml
```diff
@@ -103,4 +103,5 @@ spec:
         --set clusterName=$(params.cluster-name) \
         --set clusterEndpoint=$(cat /root/.kube/config | grep server | awk '{print $2}') \
         --set aws.defaultInstanceProfile=KarpenterNodeInstanceProfile-$(params.cluster-name) \
+        --set controller.batchIdleDuration=10s \
```
bwagner5, line 106: do we need this? Make me a little uneasy to set this to something that's not the default.

test/suites/common/upgrade-crds.yaml
```diff
@@ -15,5 +17,6 @@ spec:
   - name: upgrade-crds
     image: public.ecr.aws/karpenter-testing/tools:latest
     script: |
+      aws eks update-kubeconfig --name $(params.cluster-name)
```
bwagner5, line 20: idr if the region is already set at this point, did you try this out within tekton?

jonathan-innis: Yes, tested this against a cluster. This is the same call that occurs in the `setup.yaml`

test/suites/common/setup.yaml
jonathan-innis: I was seeing issues where pods would get batched into two separate nodes when deploying with deployment controller which made the tests unreliable.

jonathan-innis: Not sure if there's a better way to handle this since we are relying on strict node count comparison for things like Expiration testing

bwagner5 (approved):
lgtm

PR #2530, opened 2022-09-19
test: Update `WithOffset` so that failures point to outer call to helper function

jonathan-innis:
Fixes # 

**Description**

- Updates all helper functions to have `WithOffset` options so that the caller function line number will be referenced on failure rather than just pointing to the inner failure in the helper function which may not be useful.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


tzneal (approved):
lgtm

PR #2534, opened 2022-09-20
docs: adds more pod-eni info

dewjam:
Fixes # 
N/A

**Description**
Adds some more contextual information for Pod ENIs and Security Groups for pods.  Also adds a Helm value for enablePodENIs to make configuration a bit easier.

* Followed instructions in the documentation to enable POD ENI support.
* Tested that Karpenter provisions nodes when pod-eni resource is requested.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
adds more information regarding Pod ENI support
```

### DIFF
--- a/website/content/en/preview/AWS/provisioning.md
+++ b/website/content/en/preview/AWS/provisioning.md
@@ -341,3 +341,23 @@ Refer to general [Kubernetes GPU](https://kubernetes.io/docs/tasks/manage-gpus/s
 * `amd.com/gpu`: [AMD GPU device plugin for Kubernetes](https://github.com/RadeonOpenCompute/k8s-device-plugin)
 * `aws.amazon.com/neuron`: [Kubernetes environment setup for Neuron](https://github.com/aws/aws-neuron-sdk/blob/master/neuron-deploy/tutorials/tutorial-k8s.rst)
 {{% /alert %}}
+
+### Pod ENI (Security Groups for Pods)
+[Pod ENI](https://github.com/aws/amazon-vpc-cni-k8s#enable_pod_eni-v170) is a feature of the AWS VPC CNI Plugin which allows an Elastic Network Interface (ENI) to be allocated directly to a Pod. When enabled, the `vpc.amazonaws.com/pod-eni` extended resource is added to supported nodes. The Pod ENI feature can be used independently, but is most often used in conjunction with Security Groups for Pods.  Follow the below instructions to enable support for Pod ENI and/or Security Groups for Pods in Karpenter.
+
+{{% alert title="Note" color="primary" %}}
+You must enable Pod ENI support in the AWS VPC CNI Plugin before enabling Pod ENI support in Karpenter.  Please refer to the [Security Groups for Pods documentation](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html) for instructions.
+{{% /alert %}}
+
+Now that Pod ENI support is enabled in the AWS VPC CNI Plugin, you can enable Pod ENI support in Karpenter by setting the `settings.aws.enablePodENI` Helm chart value to `true`.
+
+Here is an example of a pod-eni resource defined in a deployment manifest:
+```
+spec:
+  template:
+    spec:
+      containers:
+      - resources:
+          limits:
+            vpc.amazonaws.com/pod-eni: "1"
+```


charts/karpenter/values.yaml
```diff
@@ -162,3 +162,4 @@ clusterEndpoint: ""
 aws:
   # -- The default instance profile to use when launching nodes on AWS
   defaultInstanceProfile: ""
+  enablePodENI: ""
```
jonathan-innis, line 165: Would it be better to just set the default as `false` in the helm values for visibility since we know what the default is? My understanding is that this is consistent with how other projects handle boolean helm values.

website/content/en/preview/AWS/provisioning.md
```diff
@@ -358,3 +358,26 @@ Refer to general [Kubernetes GPU](https://kubernetes.io/docs/tasks/manage-gpus/s
 * `amd.com/gpu`: [AMD GPU device plugin for Kubernetes](https://github.com/RadeonOpenCompute/k8s-device-plugin)
 * `aws.amazon.com/neuron`: [Kubernetes environment setup for Neuron](https://github.com/aws/aws-neuron-sdk/blob/master/neuron-deploy/tutorials/tutorial-k8s.rst)
 {{% /alert %}}
+
+### Pod ENI (Security Groups for Pods)
+[Pod ENI](https://github.com/aws/amazon-vpc-cni-k8s#enable_pod_eni-v170) is a feature of the AWS VPC CNI Plugin which allows an Elastic Network Interface (ENI) to be allocated directly to a Pod. When enabled, the `vpc.amazonaws.com/pod-eni` extended resource is added to supported nodes. Pod ENI support must be enabled to use the Security Groups for Pods feature. To allow Karpenter to provision nodes with the `vpc.amazonaws.com/pod-eni` resource, you must enable Pod ENI support in both the AWS VPC CNI plugin as well as in Karpenter.
```
jonathan-innis, line 363: > Pod ENI support must be enabled to use the Security Groups for Pods feature

Link here to show the user how to enable it?

jonathan-innis, line 363: > you must enable Pod ENI support in both the AWS VPC CNI plugin as well as in Karpenter

Another link to documentation would probably be helpful here

jonathan-innis (changes requested):

charts/karpenter/values.yaml
dewjam: Sure, we certainly can do that.

website/content/en/preview/AWS/provisioning.md
dewjam: There are two sections below which describe how to enable pod-eni in the VPC CNI plugin as well as in Karpenter. I also linked to the Security Groups for Pods docs site at the bottom.  Perhaps I could move it up to this line instead.

chrisnegus (dismissed):
Reads very well.
/lgtm

website/content/en/preview/AWS/provisioning.md
jonathan-innis: Ah, got it. Yeah, probably no need for links then.

Still, since the below section is describing the process of how to enable this in your cluster, we should phrase these as steps. i.e.

### Enable Security Groups for Pods
1. Enable Pod ENI support in the AWS VPC CNI plugin
2. Enable Pod ENI support in Karpenter

website/content/en/preview/AWS/provisioning.md
```diff
@@ -358,3 +358,26 @@ Refer to general [Kubernetes GPU](https://kubernetes.io/docs/tasks/manage-gpus/s
 * `amd.com/gpu`: [AMD GPU device plugin for Kubernetes](https://github.com/RadeonOpenCompute/k8s-device-plugin)
 * `aws.amazon.com/neuron`: [Kubernetes environment setup for Neuron](https://github.com/aws/aws-neuron-sdk/blob/master/neuron-deploy/tutorials/tutorial-k8s.rst)
 {{% /alert %}}
+
+### Pod ENI (Security Groups for Pods)
```
jonathan-innis, line 362: Could we also show how this can be specified in the `resources` section of the  `podSpec` similar to how we do with `nvidia` above

```yaml
spec:
  template:
    spec:
      containers:
      - resources:
          requests:
            vpc.amazonaws.com/pod-eni: "1"
```

jonathan-innis (changes requested):

website/content/en/preview/AWS/provisioning.md
dewjam: Sure thing.

dewjam: >Still, since the below section is describing the process of how to enable this in your cluster, we should phrase these as steps. i.e.
> Enable Security Groups for Pods

I see what you mean, though there are additional steps required to enable Security Groups for pods.  Pod-eni is just a prerequisite.  Perhaps we should just show how to enable pod-eni resources in Karpenter and then refer to the [Security Groups for Pods deployment guide]((https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html#security-groups-pods-deployment)) for all the remaining steps.

Something like:
"If you intend to use Security Groups for Pods with Karpenter, refer to the Security Groups for Pods deployment guide for additional steps".

A little background.  The reason I added instructions for how to enable pod-eni in the CNI plugin is to prevent an issue where Pods get stuck in a "Pending" state.  In short, if we enable pod-eni in Karpenter, we must enable it in the VPC CNI plugin as well.

jonathan-innis: I think prefacing this as these are the steps to enable pod-eni resources in Karpenter is reasonable

charts/karpenter/values.yaml
jonathan-innis: Still looks like this is a string value

website/content/en/preview/AWS/provisioning.md
dewjam: Updated.  Take another look now.

charts/karpenter/values.yaml
dewjam: I'm pretty sure it has to be a string.

website/content/en/preview/AWS/provisioning.md
dewjam: Updated now.  Take a look.

charts/karpenter/values.yaml
dewjam: oh, nope, I'm wrong.  K8S expects the value to be a string, but we handle that with 
```
{{ quote .Values.aws.enablePodENI }}
```

dewjam: Fixed now.

website/content/en/preview/tasks/configuration.md
```diff
@@ -15,7 +15,7 @@ There are two main configuration mechanisms that can be used to configure Karpen
 | Environment Variable | CLI Flag | Description |
 |--|--|--|
 | AWS_DEFAULT_INSTANCE_PROFILE | \-\-aws-default-instance-profile | The default instance profile to use when provisioning nodes in AWS|
-| AWS_ENABLE_POD_ENI | \-\-aws-enable-pod-eni | If true then instances that support pod ENI will report a vpc.amazonaws.com/pod-eni resource (default = false)|
+| AWS_ENABLE_POD_ENI | \-\-aws-enable-pod-eni | If true then instances that support pod ENI will report a vpc.amazonaws.com/pod-eni resource (default = false). **Read the [Pod ENI documentation](../../aws/provisioning/#pod-eni-security-groups-for-pods) before enabling.**|
```
jonathan-innis, line 18: Adding to this file manually will get overriden by `docgen` since this file is completely auto-generated. Either add the relevant content to the generation script directly or add it to the CLI help args

jonathan-innis (dismissed):

website/content/en/preview/tasks/configuration.md
dewjam: ah, didn't realize that.  I will take a look.

pkg/utils/project/project.go
```diff
@@ -30,3 +31,10 @@ func RelativeToRoot(path string) string {
 	manifestsRoot := filepath.Join(filepath.Dir(file), "..", "..", "..")
 	return filepath.Join(manifestsRoot, path)
 }
+
+func GetReleaseVersion() string {
+	if strings.Contains(Version, "-") {
```
dewjam, line 36: Debated on whether I should actually do this.  I'm guessing we can't rely on `-` to be in every non-production release.

pkg/utils/options/options.go
```diff
@@ -77,7 +78,7 @@ func New() *Options {
...
 	f.BoolVar(&opts.AWSENILimitedPodDensity, "aws-eni-limited-pod-density", env.WithDefaultBool("AWS_ENI_LIMITED_POD_DENSITY", true), "Indicates whether new nodes should use ENI-based pod density. DEPRECATED: Use `.spec.kubeletConfiguration.maxPods` to set pod density on a per-provisioner basis")
 	f.StringVar(&opts.AWSDefaultInstanceProfile, "aws-default-instance-profile", env.WithDefaultString("AWS_DEFAULT_INSTANCE_PROFILE", ""), "The default instance profile to use when provisioning nodes in AWS")
-	f.BoolVar(&opts.AWSEnablePodENI, "aws-enable-pod-eni", env.WithDefaultBool("AWS_ENABLE_POD_ENI", false), "If true then instances that support pod ENI will report a vpc.amazonaws.com/pod-eni resource")
+	f.BoolVar(&opts.AWSEnablePodENI, "aws-enable-pod-eni", env.WithDefaultBool("AWS_ENABLE_POD_ENI", false), fmt.Sprintf("If true then instances that support pod ENI will report a vpc.amazonaws.com/pod-eni resource. Refer to Karpenter Pod ENI documentation before enabling: https://karpenter.sh/%s/aws/provisioning/#pod-eni-security-groups-for-pods", project.GetReleaseVersion()))
```
dewjam, line 81: Interested to hear what people think about putting the version in the help message.  I could instead just refer to the "latest" docs page, which would be `https://karpenter.sh/docs/`.

hack/docs/configuration_gen_docs.go
```diff
@@ -24,6 +24,10 @@ import (
 	"github.com/aws/karpenter/pkg/utils/options"
 )
 
+var docsLinks = map[string]string{
+	"https://karpenter.sh/unspecified/aws/provisioning/#pod-eni-security-groups-for-pods": "[Pod ENI documentation](../../aws/provisioning/#pod-eni-security-groups-for-pods)",
```
njtran, line 28: Should unspecified be `preview`? or just `docs`?

pkg/utils/options/options.go
njtran: I think referring to latest might be easiest. Makes us have to call less functions.

pkg/utils/project/project.go
njtran: I'm advocating for using karpenter.sh/docs for latest rather than this, although I think this is a fair assumption.

njtran (commented):
Nice work. One nit

hack/docs/configuration_gen_docs.go
dewjam: "unspecified" is the default value if no tag is discovered (https://github.com/aws/karpenter/blob/main/pkg/utils/project/project.go#L25).  

But if we change the link in `options.go` to be `karpenter.sh/docs`, then we will replace `karpenter.sh/docs` instead.

pkg/utils/options/options.go
dewjam: Yeah, that makes sense to me also.  I think the only caveat is that `docs` refers to the latest version of documentation.  Which, if the docs page is new, will only be available in `preview` until the next release.  Essentially the link will be broken until the next stable release.

pkg/utils/project/project.go
dewjam: I just pushed an update to use the karpenter.sh/docs site instead.

pkg/utils/options/options.go
dewjam: I updated the link

website/content/en/preview/AWS/provisioning.md
```diff
@@ -341,3 +341,40 @@ Refer to general [Kubernetes GPU](https://kubernetes.io/docs/tasks/manage-gpus/s
...
+[Pod ENI](https://github.com/aws/amazon-vpc-cni-k8s#enable_pod_eni-v170) is a feature of the AWS VPC CNI Plugin which allows an Elastic Network Interface (ENI) to be allocated directly to a Pod. When enabled, the `vpc.amazonaws.com/pod-eni` extended resource is added to supported nodes. The Pod ENI feature can be used independently, but is most often used in conjunction with Security Groups for Pods.  Follow the below instructions to enable support for Pod ENI and/or Security Groups for Pods in Karpenter.
+
+{{% alert title="Note" color="primary" %}}
+You must enable Pod ENI support in the AWS VPC CNI Plugin before enabling Pod ENI support in Karpenter.  Please refer to the [Security Groups for Pods documentation](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html) for instructions.
+{{% /alert %}}
+
+Once Pod ENI support is enabled in the AWS VPC CNI Plugin, you can enable Pod ENI support in Karpenter.  This can be handled via an environmental variable, command line flag, or via Helm.
+
+* Environmental Variable: `AWS_ENABLE_POD_ENI=true`
```
ellistarn, line 354: Common mistake
```suggestion
* Environment Variable: `AWS_ENABLE_POD_ENI=true`
```

website/content/en/preview/AWS/provisioning.md
```diff
@@ -341,3 +341,40 @@ Refer to general [Kubernetes GPU](https://kubernetes.io/docs/tasks/manage-gpus/s
...
 {{% /alert %}}
+
+### Pod ENI (Security Groups for Pods)
+[Pod ENI](https://github.com/aws/amazon-vpc-cni-k8s#enable_pod_eni-v170) is a feature of the AWS VPC CNI Plugin which allows an Elastic Network Interface (ENI) to be allocated directly to a Pod. When enabled, the `vpc.amazonaws.com/pod-eni` extended resource is added to supported nodes. The Pod ENI feature can be used independently, but is most often used in conjunction with Security Groups for Pods.  Follow the below instructions to enable support for Pod ENI and/or Security Groups for Pods in Karpenter.
+
+{{% alert title="Note" color="primary" %}}
+You must enable Pod ENI support in the AWS VPC CNI Plugin before enabling Pod ENI support in Karpenter.  Please refer to the [Security Groups for Pods documentation](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html) for instructions.
+{{% /alert %}}
+
+Once Pod ENI support is enabled in the AWS VPC CNI Plugin, you can enable Pod ENI support in Karpenter.  This can be handled via an environmental variable, command line flag, or via Helm.
```
ellistarn, line 352: Thoughts on exposing this through the karpenter-settings configmap instead? I'm hesitant to use the word "OR" unless we have to. One way to do this is to keep the docs simple, even if we happen to have a little more powerful support under the hood.

website/content/en/preview/AWS/provisioning.md
dewjam: oof, I'll get that fixed.

website/content/en/preview/AWS/provisioning.md
dewjam: So you mean to keep the helm `enablePodENI` value, but populate the config map instead of pass an ENV, right?  Or am I misinterpreting.

I think we could remove the other config options in Preview docs, and leave just Helm.  For previous versions of docs we would need to provide at least one of the config options (flag or env), though.

dewjam: This is now in the global settings configmap

website/content/en/preview/AWS/provisioning.md
dewjam: Removed that entirely

dewjam:
Updated PR to account for repo split.

website/content/en/preview/AWS/provisioning.md
```diff
@@ -341,3 +341,36 @@ Refer to general [Kubernetes GPU](https://kubernetes.io/docs/tasks/manage-gpus/s
...
+helm upgrade --install --namespace karpenter --create-namespace \
+  karpenter karpenter/karpenter \
+  --version ${KARPENTER_VERSION} \
+  --set serviceAccount.annotations."eks\.amazonaws\.com/role-arn"=${KARPENTER_IAM_ROLE_ARN} \
+  --set clusterName=${CLUSTER_NAME} \
+  --set clusterEndpoint=${CLUSTER_ENDPOINT} \
+  --set aws.defaultInstanceProfile=KarpenterNodeInstanceProfile-${CLUSTER_NAME} \
+  --set aws.enablePodENI=true \
+  --wait
+```
```
bwagner5, line 365: It's probably okay to leave this example out and just have the sentence above with the value they need to set. The example has changed for v0.19.0 and it might be difficult to keep this spot up-to-date with the getting started guide: https://karpenter.sh/preview/getting-started/getting-started-with-eksctl/#install-karpenter-helm-chart

dewjam: ok, good call.

website/content/en/preview/AWS/provisioning.md
```diff
@@ -341,3 +341,23 @@ Refer to general [Kubernetes GPU](https://kubernetes.io/docs/tasks/manage-gpus/s
...
 {{% /alert %}}
+
+### Pod ENI (Security Groups for Pods)
+[Pod ENI](https://github.com/aws/amazon-vpc-cni-k8s#enable_pod_eni-v170) is a feature of the AWS VPC CNI Plugin which allows an Elastic Network Interface (ENI) to be allocated directly to a Pod. When enabled, the `vpc.amazonaws.com/pod-eni` extended resource is added to supported nodes. The Pod ENI feature can be used independently, but is most often used in conjunction with Security Groups for Pods.  Follow the below instructions to enable support for Pod ENI and/or Security Groups for Pods in Karpenter.
+
+{{% alert title="Note" color="primary" %}}
+You must enable Pod ENI support in the AWS VPC CNI Plugin before enabling Pod ENI support in Karpenter.  Please refer to the [Security Groups for Pods documentation](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html) for instructions.
+{{% /alert %}}
+
+Now that Pod ENI support is enabled in the AWS VPC CNI Plugin, you can enable Pod ENI support in Karpenter by setting the `aws.enablePodENI` Helm chart value to `true`.
```
bwagner5, line 352: This was moved under `settings.aws.enablePodENI` (https://github.com/aws/karpenter/blob/main/charts/karpenter/values.yaml#L146)

```suggestion
Now that Pod ENI support is enabled in the AWS VPC CNI Plugin, you can enable Pod ENI support in Karpenter by setting the `settings.aws.enablePodENI` Helm chart value to `true`.
```

dewjam: ah, so it was.

bwagner5 (approved):
LGTM

bwagner5 (approved):
LGTM

PR #2546, opened 2022-09-23
feat: Add Native Spot Termination Handling

### LINKED ISSUE #702: Native Support for Spot Termination
 - Native Support for Spot Termination

Using aws-node-termination-handler up to this point (https://github.com/awslabs/karpenter/issues/105)

**Additional context**
 - This will be an aws specific spot termination controller.

**Attachments**

### Community Note

jonathan-innis:
Fixes #702 

**Description**

- Adds native handling for spot termination by introducing two new controllers to the AWS cloud provider: `aws/controllers/nodetemplate` and `aws/controllers/interruption`
- `aws/controllers/nodetemplate` is responsible for reconciling the desired state of Karpenter AWS infrastructure with the current state
- `aws/controllers/nodetemplate` is also responsible for cleaning up the infrastructure when all `AWSNodeTemplates` are removed
- `aws/controllers/interruption` is responsible for polling the SQS queue and cordoning and draining instances that it receives notifications for from the SQS queue

* `make test`
* `FOCUS=Interruption make e2etests`

- [x] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Add Native Spot Termination Handling
```


pkg/apis/provisioning/v1alpha5/labels.go
```diff
@@ -31,6 +31,7 @@ var (
 	KarpenterLabelDomain = "karpenter.sh"
 
 	ProvisionerNameLabelKey           = Group + "/provisioner-name"
+	DiscoveryLabelKey                 = Group + "/discovery"
```
ellistarn, line 34: Is this vendor neutral, or specific?

ellistarn:
Discussed offline. Thoughts on a vendor neutral `AwaitDisruption()` cloudprovider API to guarantee behavior conformance between providers. This should help us if we ever want to deal with voluntary disruptions, integrate with consolidation, or change how `kubectl delete nodes` works under the hood.

bwagner5 (commented):
Have you done a load test on the queue to make sure the controller keeps up?  Especially in the case where the queue is getting a lot of messages for nodes that are not in the cluster.

pkg/cloudprovider/aws/sqs.go
```diff
@@ -0,0 +1,258 @@
...
+			v1alpha5.DiscoveryLabelKey: aws.String(injection.GetOptions(ctx).ClusterName),
+		},
+	}
+	provider.getQueueURLInput = &sqs.GetQueueUrlInput{
+		QueueName: aws.String(provider.queueName),
+	}
+	provider.receiveMessageInput = &sqs.ReceiveMessageInput{
+		MaxNumberOfMessages: aws.Int64(10),
+		VisibilityTimeout:   aws.Int64(10), // Seconds
+		WaitTimeSeconds:     aws.Int64(10), // Seconds, maximum for long polling
```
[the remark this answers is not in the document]
jonathan-innis: I guess I'll ask the same question back. What was the reason for the original 20 second heuristic?

pkg/cloudprovider/aws/eventbridge.go
```diff
@@ -0,0 +1,190 @@
...
+func NewEventBridgeProvider(eb eventbridgeiface.EventBridgeAPI, metadata *Metadata, queueName string) *EventBridgeProvider {
+	return &EventBridgeProvider{
+		client:    eb,
+		metadata:  metadata,
+		queueName: queueName,
+	}
+}
+
+func (eb *EventBridgeProvider) CreateEC2NotificationRules(ctx context.Context) (err error) {
+	wg := &sync.WaitGroup{}
```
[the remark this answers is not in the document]
jonathan-innis: I thought the same after a bit, actually implemented an `atomic.MultiErr` but then realized that this creates weird behavior when it comes to checking the "nullity" of errors based on convention.

```go
err := &atomic.MultiErr{}
go func(){
  e := DoSomeAction()
  err.Append(e)
}
go func(){
  e := DoSomeOtherAction()
  err.Append(e)
}
// This is the incorrect call but feels natural based on how errs are typically checked
if err != nil {
  // Comes here even if there is no error
}
```

I assume a better abstraction is to just abstract the entire goroutine process similary to `golang.org/x/sync/errgroup` but I wanted to return the full error set rather than just the first error. Probably just roll our own abstraction. It's not that difficult.

pkg/cloudprovider/aws/controllers/notification/controller.go
```diff
@@ -0,0 +1,326 @@
...
+		c.backoff.Reset() // We succeeded so reset the backoff period
+		select {
+		case <-ctx.Done():
+			return
+		case <-c.clock.After(pollingPeriod):
+		}
+	}
+}
+
+func (c *Controller) PollSQS(ctx context.Context) error {
```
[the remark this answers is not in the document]
jonathan-innis: I disagree with this pretty strongly, primarily because `controller-runtime` reconcilers are meant to respond to Kubernetes object events to trigger reconciliation. 

This is evident from the fact that the `Reconcile` method takes in a `reconcile.Request` which is a wrapper around `types.NamespacedName`. It isn't well-suited for periodic polling and feels unnatural to me.

The way I would bucketize it is:

1. If you need to reconcile based on some Kubernetes object or some object event, use `controller-runtime`
2. Otherwise, write a periodic poller using a controller written in the style of `pkg/consolidation/controller`

I think this is a reasonable pattern to use with the project going forward and I, personally, don't find it any less readable. In fact, I would argue that it's more readable based on the context.

pkg/cloudprovider/aws/controllers/infrastructure/controller.go
```diff
@@ -0,0 +1,342 @@
...
+	"knative.dev/pkg/system"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/aws/karpenter/pkg/cloudprovider/aws"
+	"github.com/aws/karpenter/pkg/cloudprovider/aws/events"
+	"github.com/aws/karpenter/pkg/metrics"
+	"github.com/aws/karpenter/pkg/utils/injection"
+)
+
+// Controller is the AWS infrastructure controller.  It is not a standard controller-runtime controller in that it doesn't
```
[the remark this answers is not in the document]
jonathan-innis: See comment above

pkg/cloudprovider/aws/controllers/infrastructure/controller.go
```diff
@@ -0,0 +1,343 @@
...
+
+// DeleteInfrastructure removes the infrastructure that was stood up and reconciled
+// by the infrastructure controller for SQS message polling
+func (c *Controller) DeleteInfrastructure(ctx context.Context) (err error) {
+	logging.FromContext(ctx).Infof("Deprovisioning the infrastructure...")
+	wg := &sync.WaitGroup{}
+	m := &sync.Mutex{}
+
+	wg.Add(2)
+	go func() {
```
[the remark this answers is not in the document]
jonathan-innis: Primarily because we are performing an asynchronous action (i.e. calling out to an API) and these calls don't rely on each other, so why not make it quicker?

pkg/controllers/controllers.go
```diff
@@ -200,6 +235,28 @@ func RegisterControllers(ctx context.Context, m manager.Manager, controllers ...
 	return m
 }
 
+// StartCleanupWatcher monitors the signal channel for termination, closes the cleanupAsync channel when
+// there is a signal for pod termination, waits for the cleanup operation to complete, and then cancels all contexts
+// Only the leader will perform the cleanup operation
+func StartCleanupWatcher(ctx context.Context, cancel context.CancelFunc, elected <-chan struct{},
```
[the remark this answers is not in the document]
jonathan-innis: I'm a little confused, even if the `controller-runtime` catches `SIGTERM` and closes the `Done` channel, that doesn't affect the parent context. 

Also, from what I can tell, `controller-runtime` does not do this by default. They have support using the `SetupSignalHandler` to be passed into the `mgr.Start` call but they don't catch these signals internally. Point to me a link if my understanding is wrong because I could definitely be missing something.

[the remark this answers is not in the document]
tzneal: As far as I can tell, the signal handling context is created [here](https://github.com/aws/karpenter/blob/b1ef730df63cea61a970fe3f4c046edbeec2925d/pkg/utils/injection/injection.go#L36), which is passed to manager.Start, which [returns](https://github.com/kubernetes-sigs/controller-runtime/blob/7399a3a595bf254add9d0c96c49af462e1aac193/pkg/manager/internal.go#L501) when the context is closed by the signal which is is the only thing that keeping main from exiting as it's the last thing called by `controllers.Initialize` [here](https://github.com/aws/karpenter/blob/b1ef730df63cea61a970fe3f4c046edbeec2925d/pkg/controllers/controllers.go#L160).

When I tested I didn't see the log line indicating that the cleanup functions had finished running.

pkg/cloudprovider/aws/sqs.go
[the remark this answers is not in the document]
bwagner5: Discussed offline, NTH used 20 to reduce network calls since there was no real downside to maxing out the wait period that I know of at least.  

Also I wasn't sure which method did a full query when we chatted, but looks like that is the long polling method:

```
With long polling, the ReceiveMessage request queries all of the servers for messages. Amazon SQS sends a response after it collects at least one available message, up to the maximum number of messages specified in the request. Amazon SQS sends an empty response only if the polling wait time expires.
```

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-short-and-long-polling.html

pkg/events/recorder.go
```diff
@@ -45,40 +47,40 @@ type Recorder interface {
 }
 
 type recorder struct {
-	rec record.EventRecorder
+	record.EventRecorder
```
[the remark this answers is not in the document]
jonathan-innis: Ok, the alternative here is we expose the event recorder as a method on the Recorder interface. I am trying to make the CloudProvider recorder a wrapper around this common recorder that way we get all the common recording methods for free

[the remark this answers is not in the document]
jonathan-innis: I somewhat disagree that embedding a type communicates an "is" relationship. I get that the methods end up being exposed on the container struct directly but historically my experience has been that type embedding can often be interpreted as a "has" relationship as well. Thoughts?

pkg/controllers/controllers.go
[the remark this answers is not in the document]
jonathan-innis: Ah, I completely missed that we used `signal.NewContext()` in `EnableInjectionOrDie`. Good catch!

cmd/controller/main.go
```diff
@@ -19,11 +19,15 @@ import (
...
+	awscontrollers "github.com/aws/karpenter/pkg/cloudprovider/aws/controllers"
 	"github.com/aws/karpenter/pkg/controllers"
 )
 
 func main() {
-	controllers.Initialize(func(ctx context.Context, options cloudprovider.Options) cloudprovider.CloudProvider {
-		return aws.NewCloudProvider(ctx, options)
+	controllers.Initialize(func(ctx context.Context, options cloudprovider.Options) (cloudprovider.CloudProvider, controllers.ControllerInitFunc) {
+		provider := aws.NewCloudProvider(ctx, options)
+		return provider, func(c context.Context, opts *controllers.ControllerOptions) []<-chan struct{} {
```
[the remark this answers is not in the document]
jonathan-innis: Can you clarify the part that doesn't seem straightforward to you? Is it the cleanup portion or just the injection in general?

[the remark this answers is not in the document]
jonathan-innis: Can you clarify the part that doesn't seem straightforward to you? Is it the cleanup portion or just the injection in general?

go.mod
```diff
@@ -6,9 +6,11 @@ require (
 	github.com/Pallinder/go-randomdata v1.2.0
 	github.com/avast/retry-go v3.0.0+incompatible
 	github.com/aws/aws-sdk-go v1.44.91
+	github.com/cenkalti/backoff/v4 v4.1.3
 	github.com/deckarep/golang-set v1.8.0
 	github.com/go-logr/logr v1.2.2
 	github.com/go-logr/zapr v1.2.3
+	github.com/google/uuid v1.3.0
```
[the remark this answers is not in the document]
jonathan-innis: Good find! Didn't know that uuid existed as a package inside `apimachinery`!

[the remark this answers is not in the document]
jonathan-innis: Actually, now I'm realizing that realistically this uses the same dependency tree under the hood so there really isn't much of a difference here.

pkg/cloudprovider/aws/controllers/notification/controller.go
```diff
@@ -0,0 +1,327 @@
...
+		if err != nil {
+			logging.FromContext(ctx).Errorf("Handling notification messages from SQS queue, %v", err)
+			select {
+			case <-ctx.Done():
+				return
+			case <-c.clock.After(c.backoff.NextBackOff()):
+				continue
+			}
+		}
+		c.backoff.Reset() // We succeeded so reset the backoff period
```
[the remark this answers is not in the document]
jonathan-innis: Discussed offline, will migrate these controllers to `controller-runtime`-based controllers

pkg/cloudprovider/aws/controllers/notification/controller.go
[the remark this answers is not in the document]
jonathan-innis: Had a separate discussion today offline, decision was made to migrate to `controller-runtime`-based controllers

pkg/controllers/controllers.go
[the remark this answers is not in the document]
jonathan-innis: Discussed offline: cleanup is now going to be handled when there are no Provisioners managed by the controller. We will modify the infrastructure controller to run while there are > 0 Provisioners on the cluster and delete the queue when there are <= 0 Provisioners on the cluster

pkg/cloudprovider/aws/eventbridge.go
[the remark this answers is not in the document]
jonathan-innis: Using `workqueue.ParallelizeUntil` instead

pkg/events/recorder.go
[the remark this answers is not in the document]
tzneal: IMO, the type embedding is just exposing methods that no one should be calling.  If no one should be calling them it's better to have them not exported to make it not possible to call them. Essentially, there's no benefit to embedding and the downside is you expose methods that shouldn't be.

cmd/controller/main.go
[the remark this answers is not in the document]
tzneal: The interface we are providing to a cloud provider is that they give us:
1) A function
2) That returns a function
3) That returns a channel of structs.

To figure out how to use that you're going to have to spend a good amount of time.

We should probably be using the controller-manager's `Runnable` and `LeaderElectionRunnable` interfaces both internally and for allowing cloudproviders to register their own application lifetime related processing.  That would hook them up into the lifetime that the controller-manager is already managing.

pkg/events/recorder.go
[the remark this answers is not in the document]
jonathan-innis: Updated to not expose them in this case and to wrap the `events.Recorder` in the cloudprovider

cmd/controller/main.go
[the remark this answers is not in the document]
jonathan-innis: I went back and forth between exposing Runnable on the CloudProvider interface so it was more discoverable; however, there was a dependency issue with injecting things like ClusterState into these controllers since cluster state uses the CloudProvider interface. There may be a better way to decouple these things but, in my mind, it's not unreasonable to think that cloudproviders might rely on cluster-state, and potentially other components from the common controller set/packages in the future.

The interface is simplified now to they give us:
1. A function
2. That returns a function that initializes the controllers

Fundamentally though, at this point, we need the concept of a `CloudProvider` interface and a `CloudProviderControllerSet` interface and they have to be completely decoupled for the dependency tree to work

pkg/controllers/types.go
```diff
@@ -0,0 +1,91 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/aws/karpenter/pkg/config"
+	"github.com/aws/karpenter/pkg/controllers/state"
+	"github.com/aws/karpenter/pkg/events"
+)
+
+type ControllerGetterFunc func(context.Context, *ControllerOptions) []Controller
+
+type ControllerOptions struct {
```
ellistarn, line 33: Thoughts on using a context injection pattern like we do with our logger and other options constructs. 

```
package controllers

func OptionsFromContext(context.Context) {}

func WithOptions() context.context {}
```

pkg/controllers/polling/controller.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/aws/karpenter/pkg/controllers"
+	"github.com/aws/karpenter/pkg/metrics"
+)
+
+// Immediate isn't exactly immediate for a reconcile result. But it should be passed to the RequeueAfter if you want
+// effectively immediate re-reconciliation. This can't be 0 because otherwise controller-runtime won't treat it as a
+// valid RequeueAfter value
+const Immediate = time.Nanosecond
+
+type ControllerInterface interface {
```
ellistarn, line 49: Worth discussing what it would look like to natively use controller runtime + RequeueAfter. This is nontrivial complexity that we'll need to maintain.

go.mod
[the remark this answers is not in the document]
ellistarn: The difference is ownership, licensing, CLA, etc. Let's rely on the k8s ecosystem where we can.

pkg/cloudprovider/aws/apis/v1alpha1/register.go
```diff
@@ -71,6 +71,8 @@ var (
 	LabelInstanceGPUManufacturer = LabelDomain + "/instance-gpu-manufacturer"
 	LabelInstanceGPUCount        = LabelDomain + "/instance-gpu-count"
 	LabelInstanceGPUMemory       = LabelDomain + "/instance-gpu-memory"
+
+	DiscoveryTagKey = v1alpha5.Group + "/discovery"
```
ellistarn, line 75: Should this concept be neutral? Worth a discussion.

pkg/cloudprovider/aws/controllers/infrastructure/provider.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func NewProvider(sqsProvider *aws.SQSProvider, eventBridgeProvider *aws.EventBridgeProvider) *Provider {
+	return &Provider{
+		sqsProvider:         sqsProvider,
+		eventBridgeProvider: eventBridgeProvider,
+	}
+}
+
+func (p *Provider) CreateInfrastructure(ctx context.Context) error {
```
ellistarn, line 40: Optional simplification. 
```suggestion
func (p *Provider) Create(ctx context.Context) error {
```

pkg/cloudprovider/aws/controllers/nodetemplate/controller.go
```diff
@@ -0,0 +1,104 @@
...
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named(Name))
+	nt := &v1alpha1.AWSNodeTemplate{}
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, nt); err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+	list := &v1alpha1.AWSNodeTemplateList{}
+	if err := c.kubeClient.List(ctx, list); err != nil {
```
ellistarn, line 65: Do we want to do a consistent read here?

pkg/cloudprovider/aws/controllers/nodetemplate/controller.go
```diff
@@ -0,0 +1,104 @@
...
+		return reconcile.Result{}, err
+	}
+
+	// Handle removing the finalizer and also cleaning up the infrastructure on the last AWSNodeTemplate deletion
+	if !nt.DeletionTimestamp.IsZero() {
+		if len(list.Items) == 1 {
+			c.infraController.Stop(ctx)
```
ellistarn, line 72: What if we made these controllers aware of AWSNodeTemplates, so they could simply do the right thing, rather than stopping and starting them in response to AWSNodeTemplates?

Thinking a bit further. This makes me think that we might be able to collapse the nodetemplate controller and the infrastructure controller into a single thing.

- On an interval, or NodeTemplate change (e.g. traditional NodeTemplateReconciler)
- Ensure the queue if >1
- Delete the queue if finalizing && == 1

pkg/cloudprovider/aws/controllers/nodetemplate/controller.go
```diff
@@ -0,0 +1,104 @@
...
+		controllerutil.RemoveFinalizer(nt, v1alpha5.TerminationFinalizer)
+		if err := c.kubeClient.Patch(ctx, nt, mergeFrom); err != nil {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
+	if len(list.Items) >= 1 {
+		// Start reconciling the infrastructure controller. This also waterfalls the starting of the
+		// notification controller once the infra is healthy
+		c.infraController.Start(ctx)
```
ellistarn, line 88: It's a bit odd that the infraController is responsible for starting the notificationController but not stopping it. This makes me think that the ownership boundary between these components isn't clearly defined.

pkg/cloudprovider/aws/controllers/notification/reconciler.go
```diff
@@ -0,0 +1,281 @@
...
+	statechangev0 "github.com/aws/karpenter/pkg/cloudprovider/aws/controllers/notification/event/statechange"
+	"github.com/aws/karpenter/pkg/cloudprovider/aws/events"
+	"github.com/aws/karpenter/pkg/cloudprovider/aws/utils"
+	"github.com/aws/karpenter/pkg/controllers"
+	"github.com/aws/karpenter/pkg/controllers/polling"
+	"github.com/aws/karpenter/pkg/controllers/state"
+	"github.com/aws/karpenter/pkg/metrics"
+)
+
+type Action = string
```
ellistarn, line 47: Why not follow the enum pattern from https://go.dev/doc/effective_go
```
In Go, enumerated constants are created using the iota enumerator. Since iota can be part of an expression and expressions can be implicitly repeated, it is easy to build intricate sets of values.

type ByteSize float64

const (
    _           = iota // ignore first value by assigning to blank identifier
    KB ByteSize = 1 << (10 * iota)
    MB
    GB
    TB
    PB
    EB
    ZB
    YB
)
```
or
```
type Action string

const (
 Cordon Action = "Cordon"
 CordonAndDrain Action = "CordonAndDrain"
 NoAction Action = "NoAction" 
)
```

Thinking more on this, why do we even need these actions?

pkg/cloudprovider/aws/events/recorder.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/aws/karpenter/pkg/events"
+)
+
+type recorder struct {
+	rec record.EventRecorder
+	events.Recorder
+}
+
+type Recorder interface {
```
ellistarn, line 29: This events recorder pattern we use across the codebase isn't very DRY. Can we take an AI to investigate how to implement this without so much boilerplate?

ellistarn:
Any thoughts on how to break up this PR into individually reviewable chunks?

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -120,10 +127,13 @@ func NewCloudProvider(ctx context.Context, options cloudprovider.Options) *Cloud
 				options.StartAsync,
 			),
 		),
-		kubeClient: options.KubeClient,
+		sqsProvider:         sqsProvider,
+		eventBridgeProvider: eventBridgeProvider,
```
ellistarn, line 131: IIUC, the cloudprovider doesn't need these, they're only consumed by the notification/infrastructure controllers.

pkg/cloudprovider/aws/cloudprovider.go
```diff
@@ -212,6 +222,23 @@ func (*CloudProvider) Validate(ctx context.Context, provisioner *v1alpha5.Provis
...
+func (c *CloudProvider) EventBridgeProvider() *EventBridgeProvider {
+	return c.eventBridgeProvider
+}
+
+func (c *CloudProvider) InstanceTypeProvider() *InstanceTypeProvider {
+	return c.instanceTypeProvider
+}
+
+// Name returns the CloudProvider implementation name.
+func (c *CloudProvider) Name() string {
```
ellistarn, line 238: Is this used?

pkg/cloudprovider/aws/controllers/nodetemplate/controller.go
jonathan-innis: Does that work? IIRC, we will still have to keep track of the number of times that we have triggered a requeue when there is a new event. I was trying to abstract away this bookkeeping so that could be completely up to a controller that is purely responsible to responding to trigger events and polling on an interval

pkg/cloudprovider/aws/events/recorder.go
jonathan-innis: Sounds good, I can definitely take up an AI to look into this

pkg/cloudprovider/aws/controllers/nodetemplate/controller.go
jonathan-innis: We can definitely start both controllers from here. Let's discuss the alternatives, I somewhat disagree that the boundaries aren't clearly defined, primarily because I think the alternative proposed implementation will most likely be messier

pkg/controllers/polling/controller.go
jonathan-innis: Let's discuss. I think the concept of having a single reconciliation loop run that maintains a singleton poll isn't really something that native controller runtime is built for. I think this abstraction layer helps ease the burden of handling the boilerplate of maintaining singleton reconciles while also giving a triggering mechanism for having shorter requeues if needed

pkg/cloudprovider/aws/cloudprovider.go
jonathan-innis: Make sense. We definitely still need the AWS cloudprovider, though, since we need to use the instanceTypeProvider for the ICE cache. Probably best that we keep this as a singleton rather than generating another instance of it

pkg/controllers/polling/controller.go
ellistarn: Agreed for notification, but infrastructure is tied to AWSNodeTemplate and could benefit from the watch.

pkg/cloudprovider/aws/controllers/nodetemplate/controller.go
ellistarn: > we will still have to keep track of the number of times that we have triggered a requeue when there is a new event. 

I'm imagining this thing is serialized and threadsafe. The worker shouldn't be aware of the triggering mechanism.

pkg/apis/config/settings/settings.go
```diff
@@ -35,27 +35,34 @@ var Registration = &config.Registration{
 	DefaultData:   lo.Must(defaultSettings.Data()),
 }
 
-var defaultSettings = Settings{}
+var defaultSettings = Settings{
```
ellistarn, line 38: Does this need hydrating? IIUC, the zero value of these are the correct default.

pkg/controllers/interruption/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+const (
+	subsystem              = "aws_interruption_controller"
+	messageTypeLabel       = "message_type"
+	actionableTypeLabel    = "actionable"
+	actionTypeLabel        = "action_type"
+	terminationReasonLabel = "interruption"
+)
+
+var (
+	reconcileDuration = prometheus.NewHistogram(
```
ellistarn, line 33: Should this live in the controller framework?

ellistarn (dismissed):

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,275 @@
...
+func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontroller.Builder {
+	return operatorcontroller.NewSingletonManagedBy(m).
+		Named("interruption")
+}
+
+func (c *Controller) LivenessProbe(_ *http.Request) error {
+	return nil
+}
+
+// handleMessage gets the node names of the instances involved in the queue message and takes the
```
ellistarn, line 122: > and

When I see this, I think https://en.wikipedia.org/wiki/Single-responsibility_principle

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,275 @@
...
+	// No message to parse in this case
+	if raw == nil || raw.Body == nil {
+		return nil
+	}
+	msg, err := c.parser.Parse(*raw.Body)
+	if err != nil {
+		// In the scenario where we can't parse the message, we log that we have an error and then are
+		// forced to just delete the message from the queue
+		logging.FromContext(ctx).Errorf("parsing sqs message, %v", err)
+		err = c.provider.DeleteSQSMessage(ctx, raw)
```
ellistarn, line 134: Any reason to deviate from the common idiom? `if err := ...; err != nil`

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,275 @@
...
+func (c *Controller) handleMessage(ctx context.Context, instanceIDMap map[string]*v1.Node, raw *sqsapi.Message) error {
+	// No message to parse in this case
+	if raw == nil || raw.Body == nil {
+		return nil
+	}
+	msg, err := c.parser.Parse(*raw.Body)
+	if err != nil {
+		// In the scenario where we can't parse the message, we log that we have an error and then are
+		// forced to just delete the message from the queue
+		logging.FromContext(ctx).Errorf("parsing sqs message, %v", err)
```
ellistarn, line 133: Can we let the outer context handling the error bubbling do the logging for us?

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,275 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to delete message from queue, %w", err)
+		}
+		deletedMessages.Inc()
+		return nil
+	}
+	receivedMessages.WithLabelValues(msg.Kind().String(), "true").Inc()
+
+	nodeNames := lo.Map(nodes, func(n *v1.Node, _ int) string { return n.Name })
+	logging.FromContext(ctx).Infof("Received actionable event from SQS queue for node(s) [%s%s]",
```
ellistarn, line 159: Thoughts on logging the action when we take it, rather than exposing the implementation details of how we're receiving and implementing these actions.

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Mark the offering as unavailable in the ICE cache since we got a spot interruption warning
+	if msg.Kind() == messages.SpotInterruptionKind {
+		zone := node.Labels[v1.LabelTopologyZone]
+		instanceType := node.Labels[v1.LabelInstanceTypeStable]
+		if zone != "" && instanceType != "" {
+			c.unavailableOfferingsCache.MarkUnavailable(ctx, msg.Kind().String(), instanceType, zone, v1alpha1.CapacityTypeSpot)
+		}
+	}
+	if action != NoAction {
```
ellistarn, line 196: Thoughts on simply not having a "NoAction" message? Seems like we could model it more implicitly, such that there simply aren't any actions to take from the message.

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,275 @@
...
+			return fmt.Errorf("failed to delete message from queue, %w", err)
+		}
+		deletedMessages.Inc()
+		return nil
+	}
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("event", msg.Kind()))
+
+	nodes := getInvolvedNodes(msg.EC2InstanceIDs(), instanceIDMap)
+	// There's no action to take here since the event doesn't pertain to any of our instances
+	if len(nodes) == 0 {
```
ellistarn, line 145: optional: I think a lot of code would collapse by using implicit zero looping. i.e, loop over a list of things, which implicitly doesn't take an action if the list is emtpy
```
for _, instanceId := msg.EC2InstanceID {
   node, ok := instanceIDMap[instanceIdMap]
   if !ok {
       continue
   }
}
// Always do this if handled.
c.provider.DeleteSQSMessage
```

pkg/apis/config/settings/settings.go
ellistarn: The counterpoint is that this could drift, and users might be more confused by partial hydration.

pkg/fake/recorder.go
```diff
@@ -0,0 +1,54 @@
...
+	"sync/atomic"
+
+	"github.com/aws/karpenter-core/pkg/events"
+	"github.com/aws/karpenter-core/pkg/test"
+	interruptionevents "github.com/aws/karpenter/pkg/controllers/interruption/events"
+)
+
+// EventRecorder is a mock event recorder that is used to facilitate testing.
+type EventRecorder struct {
+	InstanceSpotInterruptedCalled         atomic.Int64
```
ellistarn, line 27: optional: use a thread safe map?

pkg/apis/config/settings/settings.go
jonathan-innis: I'm more inclined to keep it as is with explicit defaults

charts/karpenter/templates/clusterrole.yaml
```diff
@@ -35,6 +35,9 @@ rules:
   - apiGroups: ["karpenter.sh"]
     resources: ["provisioners/status"]
     verbs: ["create", "delete", "patch"]
+  - apiGroups: ["karpenter.k8s.aws"]
+    resources: ["awsnodetemplates"]
```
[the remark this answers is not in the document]
jonathan-innis: There's no immediate reason to patch the status at present.

pkg/controllers/interruption/controller.go
jonathan-innis: I think I'm ok with it since the action abstraction is encapsulating the behavior of what to do based on the message type. I don't want to leak those details into the reconcile loop

pkg/controllers/nodetemplate/controller.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	nodeTemplate := stored.DeepCopy()
+	var results []reconcile.Result
+	var errs error
+	for _, r := range []interface {
+		Reconcile(context.Context, *v1alpha1.AWSNodeTemplate) (reconcile.Result, error)
+	}{
+		c.infrastructure,
+		c.finalizer,
```
ellistarn, line 81: Is this an opportunity to leverage the generic controller abstraction you'd proposed, with Reconcile() and Finalize() APIs?

jonathan-innis: Sure, let's do it in a follow-up.

website/content/en/preview/getting-started/getting-started-with-eksctl/cloudformation.yaml
```diff
@@ -4,6 +4,9 @@ Parameters:
   ClusterName:
     Type: String
     Description: "EKS cluster name"
+  ClusterNameSHA:
```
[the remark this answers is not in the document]
jonathan-innis: Unfortunately, there's no in-built CloudFormation function for generating this value.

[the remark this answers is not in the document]
jonathan-innis: Talked offline, the decision was made to use the truncated CLUSTER_NAME for the SQS queue and tag-based discovery for the EventBridge rules

website/content/en/preview/getting-started/getting-started-with-eksctl/scripts/step05-controller-iam.sh
```diff
@@ -2,6 +2,7 @@ eksctl create iamserviceaccount \
   --cluster "${CLUSTER_NAME}" --name karpenter --namespace karpenter \
   --role-name "${CLUSTER_NAME}-karpenter" \
   --attach-policy-arn "arn:aws:iam::${AWS_ACCOUNT_ID}:policy/KarpenterControllerPolicy-${CLUSTER_NAME}" \
+  --attach-policy-arn "arn:aws:iam::${AWS_ACCOUNT_ID}:policy/KarpenterEventPolicy-${CLUSTER_NAME}" \
```
bwagner5, line 5: Were you planning on doing the upgrade notes in a different PR? 

We'll need to make clear that to upgrade Karpenter they'll need to attach the new policy to the controller role since eksctl create iamserviceaccount doesn't add attachments if the role is already created.

```
> aws iam attach-role-policy --role-name "${CLUSTER_NAME}-karpenter" --policy-arn "arn:aws:iam::${AWS_ACCOUNT_ID}:policy/KarpenterEventPolicy-${CLUSTER_NAME}"
```

The upgrade guide or maybe a dedicated docs section on Spot should include migration from NTH to Karpenter native interruption handling since that will be useful for new installs of Karpenter and the upgrade case.

pkg/controllers/providers/infrastructure.go
```diff
@@ -0,0 +1,112 @@
...
+	queueExists, err := p.sqsProvider.QueueExists(ctx)
+	if err != nil {
+		return fmt.Errorf("checking queue existence, %w", err)
+	}
+	if !queueExists {
+		logging.FromContext(ctx).Debugf("Queue not found, creating the SQS notification queue...")
+		if err := p.sqsProvider.CreateQueue(ctx); err != nil {
+			return fmt.Errorf("creating sqs queue with policy, %w", err)
+		}
+		logging.FromContext(ctx).Debugf("Successfully created the SQS notification queue")
```
bwagner5, line 94: can you add the queue name to this message?

pkg/controllers/interruption/metrics.go
```diff
@@ -0,0 +1,71 @@
...
+	terminationReasonLabel = "interruption"
+)
+
+var (
+	active = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: subsystem,
+			Name:      "active",
+			Help:      "Whether the message polling is currently active.",
```
bwagner5, line 38: does this actually mean `active` or `enabled`?

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,271 @@
...
+		parser:                    NewEventParser(DefaultParsers...),
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	queueExists, err := c.sqsProvider.QueueExists(ctx)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("checking queue existence, %w", err)
+	}
+	if settings.FromContext(ctx).EnableInterruptionHandling && queueExists {
```
bwagner5, line 89: why isn't this at the top to bail early, before the QueueExists call?

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,271 @@
...
+	"github.com/aws/karpenter-core/pkg/metrics"
+	corecontroller "github.com/aws/karpenter-core/pkg/operator/controller"
+)
+
+func init() {
+	lo.Must0(apis.AddToScheme(scheme.Scheme))
+}
+
+// Controller is an AWS interruption controller.
+// It continually polls an provisioned SQS queue for events from aws.ec2 and aws.health that
```
bwagner5, line 60:
```suggestion
// It continually polls an SQS queue for events from aws.ec2 and aws.health that
```

pkg/controllers/interruption/metrics.go
```diff
@@ -0,0 +1,71 @@
...
+			Subsystem: subsystem,
+			Name:      "actions_performed",
+			Help:      "Number of notification actions performed. Labeled by action",
+		},
+		[]string{actionTypeLabel},
+	)
+)
+
+func init() {
+	crmetrics.Registry.MustRegister(receivedMessages, deletedMessages, actionsPerformed)
```
bwagner5, line 70: did you mean to not register `active`?

bwagner5 (commented):
I did another pass with a few comments and questions. Will take another pass later today. Looking good!

website/content/en/preview/getting-started/getting-started-with-eksctl/scripts/step05-controller-iam.sh
jonathan-innis: Yeah, I was planning to follow-up this PR with a docs PR as a fast-follow. I figured there would be a lot of discussion on wording for migration guide and the docs PR would be relatively large.

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,276 @@
...
+		clk:                       clk,
+		recorder:                  recorder,
+		sqsProvider:               sqsProvider,
+		unavailableOfferingsCache: unavailableOfferingsCache,
+		parser:                    NewEventParser(DefaultParsers...),
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	if settings.FromContext(ctx).EnableInterruptionHandling {
```
ellistarn, line 85: Consider using a shortcircuitting pattern to minimize nesting.
```suggestion
	if !settings.FromContext(ctx).EnableInterruptionHandling {
	   return reconcile.Result{}, nil
	}
```

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,276 @@
...
+		queueExists, err := c.sqsProvider.QueueExists(ctx)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("checking queue existence, %w", err)
+		}
+		if !queueExists {
+			enabled.Set(0)
```
ellistarn, line 91: What's the use case for this metric?

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,276 @@
...
+		kubeClient:                kubeClient,
+		clk:                       clk,
+		recorder:                  recorder,
+		sqsProvider:               sqsProvider,
+		unavailableOfferingsCache: unavailableOfferingsCache,
+		parser:                    NewEventParser(DefaultParsers...),
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
```
ellistarn, line 84: Let's create an AI to add a NodeTemplate status condition for the queue.

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	if len(sqsMessages) == 0 {
+		return reconcile.Result{}, nil
+	}
+	instanceIDMap, err := c.makeInstanceIDMap(ctx)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("making instance id map, %w", err)
+	}
+	errs := make([]error, len(sqsMessages))
+	workqueue.ParallelizeUntil(ctx, 10, len(sqsMessages), func(i int) {
```
ellistarn, line 117: Worth a peek into lo/parallel to see if there's something more idiomatic.

pkg/controllers/interruption/controller.go
```diff
@@ -0,0 +1,283 @@
...
+
+// parseMessage parses the passed SQS message into an internal Message interface
+func (c *Controller) parseMessage(raw *sqsapi.Message) (messages.Message, error) {
+	// No message to parse in this case
+	if raw == nil || raw.Body == nil {
+		return nil, fmt.Errorf("message or message body is nil")
+	}
+	msg, err := c.parser.Parse(*raw.Body)
+	if err != nil {
+		deletedMessages.Inc()
```
ellistarn, line 151: Why is the metric incremented here?

jonathan-innis: Good catch!

pkg/controllers/interruption/controller.go
jonathan-innis: I'd rather overhaul this all at once in a separate PR rather than making a change to this PR at this point.

pkg/controllers/interruption/suite_test.go
```diff
@@ -0,0 +1,393 @@
...
+
+var _ = BeforeEach(func() {
+	settingsStore := test.SettingsStore{
+		settings.ContextKey: test.Settings(),
+		awssettings.ContextKey: awssettings.Settings{
+			EnableInterruptionHandling: true,
+		},
+	}
+	ctx = settingsStore.InjectSettings(ctx)
+	env = test.NewEnvironment(ctx, func(e *test.Environment) {
```
ellistarn, line 99:
```suggestion
var _ = BeforeSuite(func() {
```

pkg/controllers/nodetemplate/infrastructure.go
```diff
@@ -0,0 +1,153 @@
...
+	sqsProvider         *providers.SQS
+	eventBridgeProvider *providers.EventBridge
+
+	lastInfrastructureReconcile time.Time // Keeps track of the last reconcile time for infra, so we don't keep calling APIs
+}
+
+// Reconcile reconciles the infrastructure based on whether interruption handling is enabled and deletes
+// the infrastructure by ref-counting when the last AWSNodeTemplate is removed
+func (i *Infrastructure) Reconcile(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate) (reconcile.Result, error) {
+	if awssettings.FromContext(ctx).EnableInterruptionHandling {
```
ellistarn, line 45: same re: shortcircuitting

pkg/controllers/nodetemplate/infrastructure.go
```diff
@@ -0,0 +1,153 @@
...
+			i.lastInfrastructureReconcile = time.Time{}
+			return reconcile.Result{}, nil
+		} else if len(list.Items) >= 1 {
+			if i.lastInfrastructureReconcile.Add(time.Hour).Before(time.Now()) {
+				if err := i.CreateInfrastructure(ctx); err != nil {
+					infrastructureHealthy.Set(0)
+					return reconcile.Result{}, err
+				}
+				i.lastInfrastructureReconcile = time.Now()
+				infrastructureHealthy.Set(1)
```
ellistarn, line 63: This seems like a status object, not a metric.

pkg/controllers/nodetemplate/infrastructure.go
```diff
@@ -0,0 +1,153 @@
...
+// CreateInfrastructure provisions an SQS queue and EventBridge rules to enable interruption handling
+func (i *Infrastructure) CreateInfrastructure(ctx context.Context) error {
+	defer metrics.Measure(infrastructureCreateDuration)()
+	if err := i.ensureQueue(ctx); err != nil {
+		return fmt.Errorf("ensuring queue, %w", err)
+	}
+	if err := i.ensureEventBridge(ctx); err != nil {
+		return fmt.Errorf("ensuring eventBridge rules and targets, %w", err)
+	}
+	logging.FromContext(ctx).Infof("Completed reconciliation of infrastructure")
```
ellistarn, line 83:
```suggestion
	logging.FromContext(ctx).Infof("Ensured existence of interruption infrastructure")
```

pkg/controllers/nodetemplate/infrastructure.go
```diff
@@ -0,0 +1,153 @@
...
+	errs := make([]error, len(funcs))
+	workqueue.ParallelizeUntil(ctx, len(funcs), len(funcs), func(i int) {
+		errs[i] = funcs[i](ctx)
+	})
+
+	err := multierr.Combine(errs...)
+	if err != nil {
+		return err
+	}
+	logging.FromContext(ctx).Infof("Completed deprovisioning the infrastructure")
```
ellistarn, line 105: I worry users might take the current message out of context.
```suggestion
	logging.FromContext(ctx).Infof("Ensured cleanup of EC2 interruption handling infrastructure")
```

pkg/controllers/nodetemplate/suite_test.go
```diff
@@ -0,0 +1,365 @@
...
+	AWSDefaultInstanceProfile: "test-instance-profile",
+}
+
+func TestAPIs(t *testing.T) {
+	ctx = TestContextWithLogger(t)
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "AWS Node Template")
+}
+
+var _ = BeforeEach(func() {
```
ellistarn, line 74: BeforeSuite

pkg/controllers/nodetemplate/suite_test.go
```diff
@@ -0,0 +1,365 @@
...
+	AWSNodeNameConvention:     string(options.IPName),
+	AWSENILimitedPodDensity:   true,
+	AWSEnablePodENI:           true,
+	AWSDefaultInstanceProfile: "test-instance-profile",
+}
+
+func TestAPIs(t *testing.T) {
+	ctx = TestContextWithLogger(t)
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "AWS Node Template")
```
ellistarn, line 71: This will be used by FOCUS
```suggestion
	RunSpecs(t, "AWSNodeTemplate")
```

pkg/controllers/nodetemplate/infrastructure.go
```diff
@@ -0,0 +1,153 @@
...
+		list := &v1alpha1.AWSNodeTemplateList{}
+		if err := i.kubeClient.List(ctx, list); err != nil {
+			return reconcile.Result{}, err
+		}
+		if !nodeTemplate.DeletionTimestamp.IsZero() && len(list.Items) == 1 {
+			if err := i.DeleteInfrastructure(ctx); err != nil {
+				return reconcile.Result{}, err
+			}
+			i.lastInfrastructureReconcile = time.Time{}
+			return reconcile.Result{}, nil
```
ellistarn, line 55: Should this logic be handled in a finalizer? You have one elsewhere.

pkg/controllers/nodetemplate/infrastructure.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+	"github.com/aws/karpenter/pkg/controllers/providers"
+)
+
+type Infrastructure struct {
+	kubeClient          client.Client
+	sqsProvider         *providers.SQS
+	eventBridgeProvider *providers.EventBridge
+
+	lastInfrastructureReconcile time.Time // Keeps track of the last reconcile time for infra, so we don't keep calling APIs
```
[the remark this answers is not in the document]
jonathan-innis: I think this is a fast-follow as part of the discussion about what should be stored in the AWSNodeTemplate but maybe shouldn't be captured as part of this PR.

pkg/controllers/nodetemplate/infrastructure.go
jonathan-innis: I don't know what you mean here. This entire sub-reconciler handles the infrastructure so that it decouples that from the finalizer creation/deletion logic

test/suites/interruption/suite_test.go
```diff
@@ -0,0 +1,293 @@
...
+		env, err = aws.NewEnvironment(t)
+		Expect(err).ToNot(HaveOccurred())
+	})
+	RunSpecs(t, "Interruption")
+}
+
+var _ = BeforeEach(func() { env.BeforeEach() })
+var _ = AfterEach(func() { env.Cleanup() })
+var _ = AfterEach(func() { env.ForceCleanup() })
+var _ = AfterEach(func() { env.AfterEach() })
```
ellistarn, line 58: Why are these different?

ellistarn (dismissed):
Finished the review. Mostly comments on metrics/logging. Otherwise amazing work, this was a huge effort!

pkg/controllers/nodetemplate/infrastructure.go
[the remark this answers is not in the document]
ellistarn: SGTM

test/suites/interruption/suite_test.go
jonathan-innis: If Cleanup fails due to some condition, forceCleanup and AfterEach will still run because they are in separate blocks

pkg/controllers/providers/eventbridge.go
```diff
@@ -0,0 +1,269 @@
...
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/aws/karpenter-core/pkg/apis/provisioning/v1alpha5"
+	"github.com/aws/karpenter-core/pkg/operator/injection"
+	awssettings "github.com/aws/karpenter/pkg/apis/config/settings"
+	awserrors "github.com/aws/karpenter/pkg/errors"
+)
+
+func init() {
+	rand.Seed(time.Now().Unix())
```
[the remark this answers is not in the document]
jonathan-innis: Good catch, and we don't really have a need to seed here in all honesty.

ellistarn (approved):

ellistarn (approved):

jonathan-innis:
🎉

FernandoMiguel:
here's the 1st part for IAM policies https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/1145
once 0.19 comes out , I can test the ConfigMap and the renamed SETs

auyer:
Hi ! Does this PR mean that it is no longer recommended (or necessary) to install the aws/aws-node-termination-handler resource ?

Thanks!

ellistarn:
> Hi ! Does this PR mean that it is no longer recommended (or necessary) to install the aws/aws-node-termination-handler resource ?

Yep. You can continue to use it, if you prefer.

bnbrijesh:
Is there any new setup for this , I do not see any docs about the new interruption controller, any help if be awesome.

ellistarn:
https://karpenter.sh/v0.21.1/concepts/deprovisioning/#interruption

sap147:
are there updated docs on how to implement this? The above link goes to a page not found. Also, what is the minimum supported version of karpenter that this was introduced in?

thepoppingone:
https://karpenter.sh/v0.30/concepts/deprovisioning/#consolidation
I think its under here

thepoppingone:
@sap147 
> https://karpenter.sh/v0.30/concepts/deprovisioning/#consolidation I think its under here

PR #2569, opened 2022-09-30
docs: Added node AMI upgrade and drift design docs

njtran:
**Description**
These design docs detail automatic node upgrades for AMIs. This includes implementing a mechanism to detect drift between the provisioning requirements and extending the current controllers.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


designs/node-upgrades.md
```diff
@@ -0,0 +1,185 @@
...
+If this does not prove enough, we can implement Imperative Upgrades or Maintenance Windows in the future.
+
+## User Stories
+
+* Karpenter will be able to automatically upgrade node AMIs to match their Provisioner requirements.
+* Karpenter will be able to enable other providers to implement their own drift mechanisms.
+
+### Background - What is node drift?
+
+Drifted nodes are nodes whose spec does not match the ProvisionerSpec and ProviderRefSpec. Automatic upgrades will monitor the cluster for drifted nodes and reconcile that node when necessary. A node can become drifted when a user changes their Provisioner or ProviderRef. For example, if a user who changes their architecture requirements to only `amd64` from `arm64`, any existing `arm64` nodes would be drifted. In another example, if a user changes their tags in the  will have underlying instances with “out-of-spec” tags. These are both actions done within the cluster.
```
bryantbiggs, line 22: Last sentence - missing something after `if a user changes their tags in the <something> will have ..."

designs/node-upgrades.md
```diff
@@ -0,0 +1,185 @@
...
+
+Yet, a node can drift even if the Provisioner or ProviderRef do not change. Underlying infrastructure in the Cloud Provider can be changed outside of the cluster. More specifically, using anything except the `aws-ids` filter in the `AWSNodeTemplate` `AMISelector` allows the AMI to change when a new AL2 EKS Optimized AMI is released, making nodes drift. Additionally, if a user changes the tags on their AMIs and uses arbitrary tags in the AWSNodeTemplate, nodes can also become drifted.
```
bryantbiggs, line 24: Instead of "nodes can also become drifted." -> "these nodes will have drifted."

designs/node-upgrades.md
```diff
@@ -0,0 +1,185 @@
...
+
+## How will we upgrade nodes?
+
+When upgrading a node, Karpenter will minimize the downtime of the applications on the node by first creating the replacement node before terminating the old node. Once the replacement is ready, Karpenter will cordon and drain the old node, terminating it when it’s fully drained. Once the old node has been terminated, the upgrade will be complete.
```
bryantbiggs, line 28: I think there is a bit more nuance to this - specifically, there are two paths to reconcilliation and they depend on the underlying operation(s) that need to be performed:
- In place reconciliation - the node does not need to be replaced and can be simply updated in-place. Updating tags is the perfect example, or modifying the security groups
- Replacement reconciliation - the node needs to be replaced in order to fulfill the change(s). Changing the underlying OS arch would require a different node for example

This nuance is important because the first path is the "happier" of the two paths and is more lightweight, whereas the second path will need to account for moving pods and also brings the burden of additional resources while reconciling

designs/node-upgrades.md
```diff
@@ -0,0 +1,185 @@
...
+
+Upgrading nodes without a dedicated controller to orchestrate the logic means relying on the termination controller to scale down and the provisioning controller to scale back up. Since the termination controller is asynchronous and terminates in quick succession with no PDBs. upgrades would happen in parallel, with rate-limiting happening at the pod level. In creating a dedicated controller to orchestrate the logic, there are two options to upgrading nodes - serially and in parallel.
+
+### Option 1 - Upgrade Serially
+
+Karpenter will only upgrade one node at a time. Each node's upgrade will be complete when a its replacement is ready and the old node is terminated. This naturally rate-limits node upgrades to node initialization time and means that users will only have to worry about one node disruption at a time with upgrades.
```
bryantbiggs, line 34: nit: "complete when a its replacement" -> "complete when its replacement"

designs/node-upgrades.md
```diff
@@ -0,0 +1,185 @@
...
+
+In order to not permanently block serial upgrades, upgrades will have timeouts. Relying on Kubernetes native concepts, Karpenter will compute a drain timeout that respects the GracefulTerminationPeriodSeconds on the pods on the nodes by summing all GracefulTerminationPeriodSeconds fields. This represents the worst case where Karpenter can only evict one pod at a time, taking as long as possible. If the replacement node fails to create, we can terminate the failed upgraded node and try another one, where rolling back is already complete.
+
+### Option 2 - Upgrade in Parallel
+
+For some use cases, upgrading serially may be too slow. To combat this, Karpenter can upgrade in parallel, allowing more than one node to upgrade at a time. In contrast to upgrading serially,  upgrading in parallel requires taking an opinion on hyper-parameters or allowing an API for users to configure them.
+
+#### Option 2a - In parallel with controls
```
bryantbiggs, line 44: This feels like the "correct" approach and native to what users expect when it comes to K8s resource management

designs/node-upgrades.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Recommendation
+
+*Upgrade serially*. If we upgrade in parallel and use something like an NDB, users will need to use NDBs to safely onboard to automatic updates. Newer users may find this too complicated and disable automatic upgrades altogether. If Karpenter hard-codes defaults into the process, some cases may end up working poorly for existing users.
+
+Upgrading serially requires no additional API or hard-coded hyper-parameters. It creates simple requirements on upgrades, and solves some of the existing problems of jitter (https://github.com/aws/karpenter/issues/1884), more granular expiration conditions (https://github.com/aws/karpenter/issues/903), and rate-limiting (https://github.com/aws/karpenter/issues/1018). Upgrading serially is additionally extendable to various forms of the parallel upgrades as called out in each of the sections. Should users want quicker upgrades in the future, Karpenter can easily implement more controls where defaults retain the same assumptions.
```
bryantbiggs, line 79: Regarding quicker upgrades - I think the worst case scenario to keep in mind is large clusters upgrading to a new AMI (either new K8s version or newly patched AMI detected) which is a common exercise and not a niche edge case

designs/node-upgrades.md
```diff
@@ -0,0 +1,185 @@
...
+
+## What can interfere with upgrades?
```
bryantbiggs, line 81: not specific to ICE but resource exhaustion in general - specifically with parallel upgrades (going back to worst case scenario of new AMI rolling out on large cluster) due to the theoretical doubling of resources during the reconciliation process:
- IP exhaustion
- AWS service limits (EC2 limits)

designs/node-upgrades.md
```diff
@@ -0,0 +1,185 @@
...
+
+A cluster may undergo less churn if Karpenter prioritizes upgrading nodes that are close to being recycled. In addition, Karpenter will also not upgrade nodes that it will be unable to upgrade successfully. With this in mind, Karpenter will not upgrade nodes that fit these conditions:
+
+* Has a pod with a do-not-evict: true annotation. Karpenter will not attempt to drain a node with a pod with this annotation. For the same reason, this will hold true for upgrades. This will be included in Upgrade Best Practices.
+* Nodes with pods that have misconfigured PDBs, or blocking PDBs altogether. If a pod is owned by two PDBs, the [eviction API blocks eviction](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/#how-api-initiated-eviction-works)of that pod. This requires user intervention.
+
+On top of this, Karpenter’s current AMI selection picks a random AMI out of a list of allowed AMIs. Karpenter will now implement AMI selection to get the latest AMI out of the allowed list. If users need a method of not having the latest EKS Optimized AMI, Karpenter can implement a way to include native AMI Versioning controls in the future.
```
bryantbiggs, line 132: Side question - it picks AMIs by random from a list? There is no ordering at all, not even on AMI date?

designs/consolidation-and-drift.md
```diff
@@ -0,0 +1,33 @@
...
+
+* Order of Node Evaluation (after emptiness) - LifecycleActions will handle expired nodes then drifted nodes before doing cost-based consolidation, prioritizing lower disruption costs for drift and cost-based consolidation.
+* Stabilization Window - This will be different based on what the last `LifecycleAction` was. Currently it’s 5 minutes for all actions. We will keep Emptiness, Expiration, Drift, and Cost at 5 minutes, but can allow it to be tunable in the future.
+* Replacement Node Timeout - Currently this is [9.5 minutes](https://github.com/aws/karpenter/blob/main/pkg/controllers/consolidation/controller.go#L70). Karpenter will not change this.
+
+## Benefits of Expanding Consolidation
+
+Expanding consolidation for drift and expiration allows Karpenter to innately behavior between the different controllers. For instance, Consolidation’s stabilization window which acts as a cooldown period can be configured differently for each mechanism. Karpenter could also more efficiently execute a series of LifecycleActions by considering future actions.
```
johngmyers, line 31: I think the first sentence is missing its verb.

designs/node-upgrades.md
johngmyers, line 22: Missing noun in "in the  will".

Perhaps change "does not match" into "does not satisfy"? A better example might be removing one value from a multi-valued instance-type requirement. Nodes of the removed type would be drifted, whereas nodes of a type still present would not.

designs/node-upgrades.md
johngmyers, line 24: I recommend consistently using either "drift" or "become drifted", not switching between the two.

designs/node-upgrades.md
johngmyers, line 28: Suggest either "by creating the replacement node before terminating the old node" or "by first creating the replacement node then terminating the old node".

designs/node-upgrades.md
```diff
@@ -0,0 +1,185 @@
...
+
+## How will we upgrade nodes?
+
+When upgrading a node, Karpenter will minimize the downtime of the applications on the node by first creating the replacement node before terminating the old node. Once the replacement is ready, Karpenter will cordon and drain the old node, terminating it when it’s fully drained. Once the old node has been terminated, the upgrade will be complete.
+
+Upgrading nodes without a dedicated controller to orchestrate the logic means relying on the termination controller to scale down and the provisioning controller to scale back up. Since the termination controller is asynchronous and terminates in quick succession with no PDBs. upgrades would happen in parallel, with rate-limiting happening at the pod level. In creating a dedicated controller to orchestrate the logic, there are two options to upgrading nodes - serially and in parallel.
```
johngmyers, line 30: This paragraph is hard to parse. I think "means" should be changed to "would mean" since the sentence appears to be describing an option not taken.

In the second sentence, the period after "PDBs" doesn't parse. It's unclear what "terminates ... with no PDBs" means. 

That there is a "dedicated controller" seems to be an implementation detail, tangental to the purpose of describing how Kerpenter will upgrade nodes.

johngmyers:
One possible alternative to NodeDisruptionBudgets would be to attach the budget to the Provisioner. Consider whether there is a use case for NDBs having coverage other than that of a Provisioner.

designs/node-upgrades.md
```diff
@@ -0,0 +1,185 @@
+# Karpenter Node Upgrades - Automatic Upgrades
```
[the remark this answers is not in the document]
njtran: I can see it both ways. In a way, Drift is the symptom and Upgrading is the cure. I could understand using Detection & Remediation for this, but I think Upgrading gets that across more tersely, regardless of the underlying implementation.

johngmyers:
In our kOps deployment, we generally limit parallelism to 30 nodes, though I'm not confident this is well-tuned. For node groups that support workloads with restrictive PDBs we also set a node-level budget of 1.

kOps does not do parallelism across node groups. If there were two Provisioners, each with its own NDB limiting to 30, I would not want a total of 60 nodes updating at once.

designs/node-upgrades.md
njtran: I agree that there are cases where in-place upgrades would work better. We could hope that applications don't take dependencies on things like labels and tags, but I think that requires more investigation and research. Therefore, initially, replacement is the safest way to introduce this node drift mechanism - and we'll be sticking to just doing node drift on AMIs. This will definitely expand in the future.

designs/node-upgrades.md
njtran: I'd be interested to see if there are any large clusters that have strict requirements on how many nodes can be upgraded at a time, and how long each of those should take. That'd be good evidence in the future to extend the process to be in parallel in the future.

designs/node-upgrades.md
njtran: Good callout. In a way, this makes a bigger case for upgrading serially, as we will only ever pre-spin one more node than the cluster's application's need (theoretically).

designs/node-upgrades.md
njtran: Currently it's at random. We'll adjust this to pick the latest, native AMI versioning in the future could include pinning versions of the EKS Optimized AMI or some sort of mandated bake time. Would be great to get feedback here as well.

designs/node-upgrades.md
njtran: Good idea. Will do.

designs/node-upgrades.md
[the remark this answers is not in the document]
johngmyers: The reconciliation could be replacement with an older AMI. The term "upgrading" has implications which might not be desirable.

designs/node-upgrades.md
johngmyers: In-place reconciliation is an optimization.

designs/node-upgrades.md
njtran: Fixed it.

designs/node-upgrades.md
johngmyers: Replacing too many nodes at once can overstress the CNI.

Serial reconciliation is a problem when humans have to manually fix stuck PDBs. It is better to be able to fix a bunch of problems at once.

njtran:
Thanks for the review @johngmyers! So it sounds like for your kOps deployments, you need global NDBs, but also some more granular controls for each node group. Neatly, upgrading seriallly would make parallelism limited to the amount of Provisioners you have, one for each.

johngmyers:
@njtran that might be accurate. Put another way, we want to rate-limit infrastructure change across the cluster, but then have a subset of groups where PDBs prevent any benefit of parallelism within the group.

I could see a case where a PDB crosses multiple Provisioners. This would be one Provisioner per availability zone, all supporting the same stateful workload. Karpenter might be able to support this with a single Provisioner, though. I've also seen a single workload backed by half EBS-backed instances and half storage-optimized instances.

designs/node-upgrades.md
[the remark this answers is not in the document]
njtran: In what ways? We'll always use the latest AMI out of the ones specified by the user

designs/node-upgrades.md
bryantbiggs: If the focus initially is just on AMIs then I agree completely.

designs/node-upgrades.md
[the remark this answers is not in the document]
olemarkus: I would explicitly call it AMI upgrades and not specifically not call it Node Upgrades. Node upgrades suggest the K8s version is increasing, in my mind. Update may also be more suitable word than upgrade.

designs/node-upgrades.md
```diff
@@ -0,0 +1,187 @@
...
+Starting with an API, Karpenter could implement a Node Disruption Budget. Similar to PDBs, NDBs rate-limit the termination process of nodes, gating termination requests from the controller or users. Similar to a restrictive PDB, a restrictive NDB can be used to stop termination altogether. Upgrading serially and using a NDB with maxUnavailable to 1 are equivalent.
+
+```
+apiVersion: karpenter.sh/v1alpha1
+kind: NodeDisruptionBudget
+metadata:
+  name: default
+spec:
+  # Merged labelSelectors for multiple provisioners
+  labelSelectors:
```
olemarkus, line 55: Maybe use the more expressive `matchExpression` syntax?

designs/node-upgrades.md
```diff
@@ -0,0 +1,187 @@
...
+    - karpenter.sh/provisioner-name: default
+    - karpenter.sh/provisioner-name: app-name
+  # Following two are mutually exclusive:
+  minAvailable: 10% # intOrString
+  maxUnavailable: 30 # intOrString
```
olemarkus, line 60: How about surging? I.e scaling the cluster up by X extra nodes before tearing down old ones? A minimal extra cost to significantly safer upgrade (mitigates potential capacity exhaustion challenges)

designs/node-upgrades.md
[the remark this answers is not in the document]
johngmyers: The user might have changed the specification to refer to an older AMI than previously. This would typically be when the newer AMI shows unexpected behavior.

designs/consolidation-and-drift.md
```diff
@@ -0,0 +1,33 @@
...
+
+Karpenter will implement a mechanism to detect and reconcile nodes that drift from Provisioner and AWSNodeTemplate specs. Karpenter will merge Consolidation, Expiration, Emptiness, and Drift into a `NodeLifecycle` controller which will be responsible for all controllers executing scale-down logic on existing nodes.
+
+## Consolidation Background
+
+Consolidation implements node emptiness, deletion, and replacement as `consolidationActions` named `deleteEmpty`, `delete`, and `replace`, respectively. Consolidation takes one `consolidationAction` per reconcile loop, where any empty nodes are deleted first. After ordering the consolidatable nodes by `disruptionCost`, Karpenter computes a `consolidationAction` for one node at a time.
+
+The `deleteEmpty` operation is by definition able to delete a node immediately without considering PDBs or `do-not-evicts` since no pods exist on the node. In this case, no replacement node needs to be created.
+
+A node cannot be considered for the `delete` and `replace` operations if the node has a blocking PDB, isn’t already being deleted, or has a pod that cannot be evicted. A replacement node is created only for the replace operation.
```
blakepettersson, line 11: > isn’t already being deleted

This should be "is already being deleted" right? Might sound better with 

> in the process of being deleted

tzneal (approved):
lgtm

designs/node-upgrades.md
njtran: That's a good idea! Since this section is explicitly called out as not the suggestion for now, I was thinking that we can come back to this in the future, if the current design is not good enough for users.

designs/node-upgrades.md
njtran: Surging also sounds like a good idea. I think this is something we should definitely consider once NDBs go through a design review.

designs/node-upgrades.md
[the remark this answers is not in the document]
hawkesn: AWS has previously called something like this an [instance refresh](https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-instance-refresh.html). Not sure if this comment is helpful, but just adding another name option 😄

PR #2590, opened 2022-10-03
feat: make consolidation a bit more conservative

### LINKED ISSUE #2370: make consolidation wait a bit
i’ve noticed that consolidation  is a tad toooo quick removing nodes.
sometimes a pod is being replaced, and before the new one even get the change of being scheduled, karpenter kills the node just for 1 or 2 seconds later spinning a new one… this makes the deployment take 1 min for the node to be ready

## Expected Behavior
for karpenter to keep nodes a bit longer so workloads being replaced still have where to go

## Actual Behavior
nodes get ripped too quick

tzneal:
Fixes #2370

**Description**

- Allow pod creation/deletion to affect the stabilization window calculation.
- Increase minimum stabilization window time from 0 seconds to 15 seconds
- Reduce max stabilization window time from 5 minutes to 3 minutes

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


pkg/controllers/state/cluster.go
```diff
@@ -464,6 +476,16 @@ func (c *Cluster) updatePod(ctx context.Context, pod *v1.Pod) error {
 		err = c.updateNodeUsageFromPod(ctx, pod)
 	}
 	c.updatePodAntiAffinities(pod)
+	if pod.DeletionTimestamp != nil {
+		podDeletionTime := pod.DeletionTimestamp.UnixMilli()
+		if podDeletionTime > atomic.LoadInt64(&c.lastPodDeletionTime) {
```
jonathan-innis, line 481: Don't these comparisons create races where two loads happen between the store call on 482?

pkg/controllers/state/cluster.go
```diff
@@ -464,6 +476,16 @@ func (c *Cluster) updatePod(ctx context.Context, pod *v1.Pod) error {
...
+			atomic.StoreInt64(&c.lastPodDeletionTime, podDeletionTime)
+		}
+	}
+	podCreationTime := pod.CreationTimestamp.UnixMilli()
+	if podCreationTime > atomic.LoadInt64(&c.lastNodeCreationTime) {
```
jonathan-innis, line 486: Same question with race condition here

pkg/controllers/state/cluster.go
jonathan-innis: A way to resolve this might be to use the `atomic.CompareAndSwap...` call

tzneal: I don't think it matters. The deletion timestamp has a one-second resolution IIRC and if something else stores with a different value, it won't be off by more than a second.  Changing the code to do compare and swap makes it even more complicated:

```go
	if pod.DeletionTimestamp != nil {
		podDeletionTime := pod.DeletionTimestamp.UnixMilli()
		for {
			lastPodDeletionTime := atomic.LoadInt64(&c.lastPodDeletionTime)
			if (podDeletionTime > lastPodDeletionTime && atomic.CompareAndSwapInt64(&c.lastPodDeletionTime, lastPodDeletionTime, podDeletionTime)) ||
				podDeletionTime <= lastPodDeletionTime {
				break
			}
		}
	}
```

jonathan-innis: Makes sense. Yeah, I guess in this case, the race condition is warranted. Can we make a comment to note that a race could happen in these two locations?

My main concern is someone coming in the future and assuming that this code section is synchronous without being aware of this possible race.

jonathan-innis (changes requested):

pkg/controllers/state/cluster.go
```diff
@@ -464,6 +476,19 @@ func (c *Cluster) updatePod(ctx context.Context, pod *v1.Pod) error {
 		err = c.updateNodeUsageFromPod(ctx, pod)
 	}
 	c.updatePodAntiAffinities(pod)
+	if pod.DeletionTimestamp != nil {
+		// We could do a more complex cmp/swap operation, but since deletion/creation timestamps have a resolution of
```
jonathan-innis, line 480: nit: Sorry, maybe I'm being confusing. I just meant that this comment should note that there is a potential race here. I personally don't think you need to go into full detail into how we could improve.

pkg/controllers/state/cluster.go
```diff
@@ -464,6 +476,19 @@ func (c *Cluster) updatePod(ctx context.Context, pod *v1.Pod) error {
...
+		// one-second and worst case we'll be delay by a second too little, it's more straightforward to just
+		// load/compare/store vs load/compareswap in a loop.
+		podDeletionTime := pod.DeletionTimestamp.UnixMilli()
+		if podDeletionTime > atomic.LoadInt64(&c.lastPodDeletionTime) {
+			atomic.StoreInt64(&c.lastPodDeletionTime, podDeletionTime)
+		}
+	}
+	podCreationTime := pod.CreationTimestamp.UnixMilli()
+	if podCreationTime > atomic.LoadInt64(&c.lastPodCreationTime) {
```
jonathan-innis, line 489: Same comment should be made here

jonathan-innis (changes requested):

pkg/controllers/state/cluster.go
tzneal: Sorry, I think it's fine as is.

pkg/controllers/consolidation/stabilizationwindow.go
```diff
@@ -0,0 +1,201 @@
...
+type StabilizationWindow struct {
+	clock clock.Clock
+
+	lastConsolidationTime time.Time
+	cluster               *state.Cluster
+	kubeClient            client.Client
+}
+
+// fastWindowTime is the  window used if all the cluster workloads are reporting ready
+const fastWindowTime = 15 * time.Second
```
ellistarn, line 40: I wonder if this is large enough that we'll see significant slow down for large scale downs. IIUC, all voluntary terminations will eventually be subject to this window, and this could really add up. 

IIUC, there are two cases here. 1/ allow controllers to recreate pods, and 2/, cover the delete -> create deployment case. I'm a bit hesitant of optimizing too heavily towards the latter case. WDYT about shortening this to something like 5 seconds?

pkg/controllers/consolidation/stabilizationwindow.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// or has our stabilization window exceeded its max length?
+	exceededMaxStabWindow := w.lastConsolidationTime.Add(maxWindowTime).Before(w.clock.Now())
+
+	return clusterStable || exceededMaxStabWindow
+}
+
+// Reset resets the controller's stabilization window, we must get out of this window again before
+// IsStabilized will return true
+func (w *StabilizationWindow) Reset() {
```
ellistarn, line 88: Thoughts on setting this in IsStabilized, so callers don't need to think about it?

pkg/controllers/consolidation/stabilizationwindow.go
tzneal: 5 seconds is probably ok, trying to balance letting controllers react but since we trigger on pod deletion I think it will be fine.

ellistarn (approved):

tzneal:
Closing since I prefer #2614

PR #2614, opened 2022-10-06
feat: add a node TTL for consolidation

### LINKED ISSUE #2370: make consolidation wait a bit

tzneal:
Fixes #2370

**Description**

- Use a consolidation TTL of 15 seconds which allows removing the stabilization window

* Unit testing, E2E tests and deployed to EKS.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/consolidation/controller.go
+++ b/pkg/controllers/consolidation/controller.go
@@ -192,7 +192,7 @@ func (c *Controller) consolidateNodes(ctx context.Context, candidates []candidat
 			logging.FromContext(ctx).Errorf("calculating consolidation option, %s", err)
 			continue
 		}
-		if cmd.action == lifecycleActionDelete || cmd.action == lifecycleActionReplace {
+		if cmd.action == deprovisionActionDelete || cmd.action == deprovisionActionReplace {
 			isValid, err := c.validateCommand(ctx, cmd)
 			if err != nil {
 				logging.FromContext(ctx).Errorf("validating command, %s", err)
@@ -226,7 +226,7 @@ func (c *Controller) deleteEmptyNodes(ctx context.Context, candidates []candidat
 
 	cmd := lifecycleCommand{
 		nodesToRemove: lo.Map(emptyNodes, func(n candidateNode, _ int) *v1.Node { return n.Node }),
-		action:        lifecycleActionDeleteEmpty,
+		action:        deprovisionActionDeleteEmpty,
 		created:       c.clock.Now(),
 	}
 	isValid, err := c.validateCommand(ctx, cmd)
@@ -350,9 +350,9 @@ func (c *Controller) buildProvisionerMap(ctx context.Context) (map[string]*v1alp
 }
 
 func (c *Controller) executeCommand(ctx context.Context, action lifecycleCommand) {
-	if action.action != lifecycleActionDelete &&
-		action.action != lifecycleActionReplace &&
-		action.action != lifecycleActionDeleteEmpty {
+	if action.action != deprovisionActionDelete &&
+		action.action != deprovisionActionReplace &&
+		action.action != deprovisionActionDeleteEmpty {
 		logging.FromContext(ctx).Errorf("Invalid disruption action calculated: %s", action.action)
 		return
 	}
@@ -362,7 +362,7 @@ func (c *Controller) executeCommand(ctx context.Context, action lifecycleCommand
 	// action's stringer
 	logging.FromContext(ctx).Infof("Consolidating via %s", action.String())
 
-	if action.action == lifecycleActionReplace {
+	if action.action == deprovisionActionReplace {
 		if err := c.launchReplacementNode(ctx, action); err != nil {
 			// If we failed to launch the replacement, don't consolidate.  If this is some permanent failure,
 			// we don't want to disrupt workloads with no way to provision new nodes for them.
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -32,6 +32,7 @@ jobs:
         make ci
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GO_TEST_TIMEOUT_SCALE: 2
```
ellistarn, line 35: Does it make sense to set this in the Makefile to prevent CI and local from drifting?

pkg/controllers/consolidation/controller.go
```diff
@@ -685,3 +630,94 @@ func (c *Controller) setNodeUnschedulable(ctx context.Context, nodeName string,
 	}
 	return nil
 }
+
+// validateCommand is used to validate that a lifecycleCommand is still able to be performed by re-inspecting nodes and
+// re-simulating scheduling
+// nolint:gocyclo
```
ellistarn, line 636: I'm noticing a lot of disabling of gocyclo. We originally added it as a hint to developers to factor their code into smaller chunks to improve readability. 

I'm open minded about this though. Are you finding it getting in your way? Do we want to remove it from the project?

tzneal: I'll revisit and see if I can refactor. In general I think it's pretty useful, it's sort of a judgement call to see if breaking it up makes it worse or not.  I'll mark this as WIP, working on figuring out why it's failing 100% in CI and passing 100% locally.

pkg/controllers/consolidation/controller.go
```diff
@@ -468,198 +525,130 @@ func (c *Controller) calculateLifetimeRemaining(node candidateNode) float64 {
 	return remaining
 }
 
-// nolint:gocyclo
-func (c *Controller) nodeConsolidationOptionReplaceOrDelete(ctx context.Context, node candidateNode) (consolidationAction, error) {
+//nolint:gocyclo
+func (c *Controller) nodeConsolidationOptionReplaceOrDelete(ctx context.Context, node candidateNode) (lifecyleCommand, error) {
```
njtran, line 529: nit: It might be more apt to name this `computeNodeConsolidationOption` since it returns more than just a replace or delete action. Happy to do this in my follow up refactor though.

pkg/controllers/consolidation/controller.go
```diff
@@ -468,198 +525,130 @@ func (c *Controller) calculateLifetimeRemaining(node candidateNode) float64 {
...
-		return consolidationAction{result: consolidateResultNotPossible}, nil
+		return lifecyleCommand{action: lifecycleActionNonePossible}, nil
 	}
 
 	// get the current node price based on the offering
 	// fallback if we can't find the specific zonal pricing data
 	offering, ok := cloudprovider.GetOffering(node.instanceType, node.capacityType, node.zone)
 	if !ok {
-		return consolidationAction{result: consolidateResultUnknown}, fmt.Errorf("getting offering price from candidate node, %w", err)
+		return lifecyleCommand{action: lifeCycleActionUnknown}, fmt.Errorf("getting offering price from candidate node, %w", err)
```
njtran, line 562: This seems like it was implemented like this previously, but shouldn't this be a `lifeCycleActionFailed` and we should reserve the Unknown action for any other unexpected errors? It looks like we'd never return that if not here anyway.

njtran (dismissed):
LGTM just some nits.

pkg/controllers/consolidation/types.go
```diff
@@ -20,70 +20,99 @@ import (
...
 
+	"github.com/aws/karpenter/pkg/apis/provisioning/v1alpha5"
 	"github.com/aws/karpenter/pkg/controllers/provisioning/scheduling"
 )
 
-// ProcessResult is used to indicate the result of consolidating so we can optimize by not trying to consolidate if
+// LifecycleResult is used to indicate the action of consolidating so we can optimize by not trying to consolidate if
 // we were unable to consolidate the cluster and it hasn't changed state with respect to pods/nodes.
-type ProcessResult byte
+type LifecycleResult byte
```
ellistarn, line 29: Nick, Bill and I discussed this name, and we ended up really liking the following for how we talk about this in the docs
```suggestion
type DeprovisioningResult byte
```

pkg/controllers/consolidation/controller.go
```diff
@@ -139,56 +136,120 @@ type candidateNode struct {
...
+		default:
+			// shouldn't occur
+			return LifecycleResultFailed, fmt.Errorf("unhandled lifecycle result %d", result)
+		}
+	}
+	return LifecycleResultNothingToDo, nil
+}
 
+// consolidateNodes looks at the non-empty nodes and executes the first viable consolidation action
+func (c *Controller) consolidateNodes(ctx context.Context, candidates []candidateNode) (LifecycleResult, error) {
```
ellistarn, line 182: Discussed offline, we probably need to take a new cluster snapshot for after each simulation, especially if we're waiting for n seconds in between consolidations.

pkg/controllers/consolidation/controller.go
```diff
@@ -37,10 +38,11 @@ import (
 	"github.com/aws/karpenter/pkg/cloudprovider"
 	"github.com/aws/karpenter/pkg/cloudprovider/aws/apis/v1alpha1"
 	"github.com/aws/karpenter/pkg/controllers/provisioning"
-	"github.com/aws/karpenter/pkg/controllers/provisioning/scheduling"
+	pscheduling "github.com/aws/karpenter/pkg/controllers/provisioning/scheduling"
```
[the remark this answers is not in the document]
tzneal: Maybe so, unless you're just instantiating a provisioner you probably need the scheduling stuff as well.

pkg/controllers/consolidation/types.go
```diff
@@ -20,70 +20,102 @@ import (
...
+type lifecyleAction byte
 
 const (
-	consolidateResultUnknown consolidateResult = iota
-	consolidateResultNotPossible
-	consolidateResultDelete
-	consolidateResultDeleteEmpty
-	consolidateResultReplace
-	consolidateResultNoAction
+	lifeCycleActionUnknown lifecyleAction = iota
```
njtran, line 56: Since you changed some of the types to be `DeprovisioningResult....` it'd make sense to change the actions as well.

njtran (commented):
One nit on the naming, but otherwise LGTM.

ellistarn:
LGTM, needs a rebase.

njtran (approved):
Nice!

FernandoMiguel:
@tzneal https://github.com/aws/karpenter/blob/main/charts/karpenter/crds/karpenter.sh_provisioners.yaml#L260 needs to be updated

PR #2621, opened 2022-10-07
fix: go mod linting errors and add dependabot for the test module

ellistarn:
Fixes # 

**Description**

Head broken due to dependabot not being caught up to latest makefile changes.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


Makefile
```diff
@@ -118,10 +118,6 @@ docgen: ## Generate docs
 api-code-gen: ## Auto generate files based on AWS APIs response
 	$(WITH_GOFLAGS) ./hack/api-code-gen.sh
 
-release-gen: tidy download docgen ## Generate any materials which should be updated prior to release
```
jonathan-innis, line 121: This is part of the stable release workflow, so I don't think that we can delete this

jonathan-innis (changes requested):

Makefile
ellistarn: See: https://github.com/aws/karpenter/pull/2621/files#diff-b7b8f3ad0055e283745ba109392228345271045251ea72b5d1ce2da2f3ac2bcfR58.

The observation is that release-gen is really just doing docgen at this point. There aren't any code changes happening in that workflow, as they would have had to happen prior to the release.

PR #2657, opened 2022-10-12
chore: Add an implementation of `atomic.Lazy` which caches values internally

jonathan-innis:
Fixes # 

**Description**

- Adds an implementation of lazy evaluation which allows us to cache an api call persistently internally with lazy evaluation

```go
var queueURL atomic.Lazy[string]

queueURL.Resolve = func(ctx context.Context) (string, error) {
	ret, err := provider.client.GetQueueUrlWithContext(ctx, provider.getQueueURLInput)
	if err != nil {
		return "", fmt.Errorf("fetching queue url, %w", err)
	}
	return aws.StringValue(ret.QueueUrl), nil
}
queueURL.Set("new-url")
```

* `make test`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/utils/atomic/cached_val.go
+++ b/pkg/utils/atomic/cached_val.go
@@ -1,0 +1,85 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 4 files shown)


ellistarn (approved):
These are really cool!

PR #2676, opened 2022-10-17
chore: initialization touchups

ellistarn:
Fixes # 

**Description**

- Subclasses context.Context w/ operator.Context that contains process-wide accessible objects.
- Collapses operator specific options/injection into `operator` package

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/operator/options.go
```diff
@@ -53,9 +53,10 @@ func init() {
 	utilruntime.Must(apis.AddToScheme(scheme))
 }
 
-// Options exposes shared components that are initialized by the startup.Initialize() call
-type Options struct {
-	Ctx        context.Context
+// Context is the root context for the operator and exposes shared components used by the entire process
```
jonathan-innis, line 56: We should rename this file to `Context` if we are renaming the concept

cmd/controller/main.go
```diff
@@ -27,20 +27,15 @@ import (
...
-		manager,
-		controllers.GetControllers(options, cloudProvider)...,
-	).Start(options.Ctx); err != nil {
+	ctx := operator.NewContextOrDie()
+	cloudProvider := awscloudprovider.NewCloudProvider(ctx, cloudprovider.Options{
+		ClientSet:  ctx.Clientset,
+		KubeClient: ctx.KubeClient,
+		StartAsync: ctx.StartAsync,
+	})
+	utilruntime.Must(ctx.Manager.AddHealthzCheck("cloud-provider", cloudProvider.LivenessProbe))
```
jonathan-innis, line 36: Can we remove the `HealthCheck` interface now?

cmd/controller/main.go
```diff
@@ -27,20 +27,15 @@ import (
...
-	if hp, ok := cloudProvider.(operator.HealthCheck); ok {
-		utilruntime.Must(manager.AddHealthzCheck("cloud-provider", hp.LivenessProbe))
-	}
-	cloudProvider = cloudprovidermetrics.Decorate(cloudProvider)
-	if err := operator.RegisterControllers(options.Ctx,
-		manager,
-		controllers.GetControllers(options, cloudProvider)...,
-	).Start(options.Ctx); err != nil {
+	ctx := operator.NewContextOrDie()
+	cloudProvider := awscloudprovider.NewCloudProvider(ctx, cloudprovider.Options{
```
jonathan-innis, line 31: I have some small concerns about exposing the manager through the context in this way. If someone does a typed conversion to the `operator.Context` object within the cloudprovider, they could then have access to the manager which seems like something we want to avoid

jonathan-innis (changes requested):

cmd/controller/main.go
ellistarn: Agree. Will play with this.

pkg/operator/manager.go
```diff
@@ -53,16 +49,16 @@ type HealthCheck interface {
...
+		LeaderElection:             options.EnableLeaderElection,
 		LeaderElectionID:           "karpenter-leader-election",
 		LeaderElectionResourceLock: resourcelock.LeasesResourceLock,
 		Scheme:                     scheme,
-		MetricsBindAddress:         fmt.Sprintf(":%d", opts.MetricsPort),
-		HealthProbeBindAddress:     fmt.Sprintf(":%d", opts.HealthProbePort),
-		BaseContext:                newRunnableContext(config, opts, logging.FromContext(ctx)),
+		MetricsBindAddress:         fmt.Sprintf(":%d", options.MetricsPort),
+		HealthProbeBindAddress:     fmt.Sprintf(":%d", options.HealthProbePort),
+		BaseContext:                func() context.Context { return ctx },
```
jonathan-innis, line 61: Base context was introduced to avoid top-level context cancellation from stopping runnable cleanup as shown [here](https://github.com/kubernetes-sigs/controller-runtime/pull/1846). We shouldn't be passing the top-level context in a closure here since this gets back to the issue that controller-runtime was trying to avoid

pkg/utils/injection/injection.go
```diff
@@ -18,30 +18,10 @@ import (
...
-	ctx, startinformers := knativeinjection.EnableInjectionOrDie(signals.NewContext(), config)
-	logger, atomicLevel := sharedmain.SetupLoggerOrDie(ctx, componentName)
-	ctx = logging.WithLogger(ctx, logger)
-	rest.SetDefaultWarningHandler(&logging.WarningHandler{Logger: logger})
-	sharedmain.WatchLoggingConfigOrDie(ctx, cmw, logger, atomicLevel, componentName)
-	startinformers()
-	return ctx
-}
-
 func WithNamespacedName(ctx context.Context, namespacedname types.NamespacedName) context.Context {
```
jonathan-innis, line 25: Why do these two methods stay in injection?

jonathan-innis (changes requested):

jonathan-innis (dismissed):
LGTM 🚀

cmd/controller/main.go
ellistarn: Used all over, unfortunately.

pkg/operator/manager.go
ellistarn: Fixed and added a comment

pkg/utils/injection/injection.go
ellistarn: They're relied upon in the AWS API which creates a circular dependency. I need to make some other changes to remove this.

jonathan-innis (dismissed):
LGTM 🚀

jonathan-innis (dismissed):
LGTM 🚀

jonathan-innis (approved):
LGTM 🚀

pkg/operator/context.go
```diff
@@ -0,0 +1,114 @@
...
+	scheme = runtime.NewScheme()
+)
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(apis.AddToScheme(scheme))
+}
+
+// Context is the root context for the operator and exposes shared components used by the entire process
+type Context struct {
```
jonathan-innis, line 57: After some thought, I want to discuss this decision a bit more. Primarily because I am concerned with the ramifications of passing fields that would normally be passed through to controller constructors as part of the `context.Context` struct.

In general, the `Context` object is used for request-specific context and cancellation and not for instantiation/performing dependency injection. 

I'm curious for what your thought is in where else this `Context` might be passed; otherwise, I'm leaning more towards continuing calling this `Options`

jonathan-innis (changes requested):

pkg/operator/context.go
ellistarn: I think we need to sort out the neutrality of these bits too, since options currently co-mingles AWS specific and vendor agnostic parameters. Let's spend some time pair programming this.

PR #2683, opened 2022-10-18
chore: Extend AWS event recorder to specific event types

jonathan-innis:
Fixes # 

**Description**

- Adds the `events.Event` interface for AWS cloudprovider events
- Adds the `events.EventTemplate` interface that can be used to create EventTemplates for events that are often fired
- Adds individual event types that will be used by `InterruptionController`

```go
n := &v1.Node{}
c.recorder.Publish(InstanceSpotInterrupted.For(n))
```

* `make battletest`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/pkg/cloudproviders/aws/controllers/interruption/events/events.go
+++ b/pkg/cloudproviders/aws/controllers/interruption/events/events.go
@@ -1,0 +1,68 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package events
+
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/aws/karpenter/pkg/cloudproviders/aws/events"
+)
+
+var (
+	InstanceSpotInterrupted = events.EventTemplate[*v1.Node]{
+		Type:   v1.EventTypeWarning,
+		Reason: "InstanceSpotInterrupted",
+		MessageTemplate: func(n *v1.Node) string {
+			return fmt.Sprintf("Node %s event: A spot interruption warning was triggered for the node", n.Name)
+		},
+	}
+	InstanceSpotRebalanceRecommendation = events.EventTemplate[*v1.Node]{
+		Type:   v1.EventTypeNormal,
+		Reason: "InstanceSpotRebalanceRecommendation",
+		MessageTemplate: func(n *v1.Node) string {
+			return fmt.Sprintf("Node %s event: A spot rebalance recommendation was triggered for the node", n.Name)
+		},
+	}
+	InstanceStopping = events.EventTemplate[*v1.Node]{
+		Type:   v1.EventTypeWarning,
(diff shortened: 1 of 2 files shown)


pkg/cloudproviders/aws/events/recorder.go
```diff
@@ -15,19 +15,52 @@ limitations under the License.
...
 
 import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 )
 
-// Recorder is used to record events to the Kubernetes Event API
+type EventType byte
```
ellistarn, line 24: If you make this a string, you get String() for free.

pkg/cloudproviders/aws/events/recorder.go
```diff
@@ -15,19 +15,52 @@ limitations under the License.
...
 
 func NewRecorder(r record.EventRecorder) Recorder {
-	return recorder{
+	return &recorder{
 		rec: r,
 	}
 }
+
+func (r *recorder) Create(evt Event) {
+	r.rec.Eventf(evt.InvolvedObject(), evt.Type().String(), evt.Reason(), evt.Message())
```
ellistarn, line 65:
```suggestion
	r.rec.Event(evt.InvolvedObject(), evt.Type().String(), evt.Reason(), evt.Message())
```

pkg/cloudproviders/aws/events/recorder.go
```diff
@@ -15,19 +15,52 @@ limitations under the License.
...
+		return "Normal"
+	case WarningType:
+		return "Warning"
+	default:
+		return fmt.Sprintf("Unsupported EventType %d", e)
+	}
+}
+
+type Event interface {
+	InvolvedObject() runtime.Object
```
ellistarn, line 44: Thoughts?
```suggestion
	For() runtime.Object
```

pkg/cloudproviders/aws/events/recorder.go
ellistarn: Also, corev1 has 
```
// Valid values for event types (new types could be added in future)
const (
	// Information only and will not cause any problems
	EventTypeNormal string = "Normal"
	// These events are to warn that something might go wrong
	EventTypeWarning string = "Warning"
)
```

jonathan-innis: This is the general way to do enums in Go, though. It is a more efficient memory representation of the type.

pkg/cloudproviders/aws/events/recorder.go
jonathan-innis: I like `InvolvedObject` since it aligns with what `corev1.Event` calls the field

pkg/cloudproviders/aws/controllers/interruption/events.go
```diff
@@ -0,0 +1,193 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/aws/karpenter/pkg/cloudproviders/aws/events"
+)
+
+var _ events.Event = EC2SpotInterruptionWarning{}
+
+type EC2SpotInterruptionWarning struct {
```
ellistarn, line 29: Including Type() in the type name feels a bit leaky. WDYT:
```suggestion
type EC2SpotInterruption struct {
```

pkg/cloudproviders/aws/events/recorder.go
ellistarn: Oh I missed that since the Event() API calls it object. SGMT.

ellistarn (approved):

ellistarn (approved):

ellistarn (approved):

PR #2690, opened 2022-10-19
fix: cycle through pods multiple times as required when scheduling

tzneal:
**Description**

This is a minor fix, but allows us to schedule to lager nodes in the case where pods have inter-dependencies that can be satisfied as long as they are scheduled in the correct order. There was a test for this, but the test was incorrect in that it didn't actually verify that the pods scheduled.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


ellistarn (approved):
Nice catch!

PR #2695, opened 2022-10-19
fix: provide a mock k8s client for use in generating instance type docs

tzneal:
Fixes # 

**Description**
The current AWS CP needs the K8s DNS ip for detecting IPV6, so provide a  mock K8s client that returns it.

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/hack/docs/instancetypes_gen_docs.go
+++ b/hack/docs/instancetypes_gen_docs.go
@@ -20,7 +20,9 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io"
 	"log"
+	"net/http"
 	"os"
 	"sort"
 	"strings"
@@ -29,13 +31,15 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 
 	"github.com/aws/karpenter-core/pkg/apis/provisioning/v1alpha5"
+	awscloudprovider "github.com/aws/karpenter/pkg/cloudproviders/aws/cloudprovider"
 	"github.com/aws/karpenter/pkg/operator/injection"
 	"github.com/aws/karpenter/pkg/operator/options"
 
 	"github.com/aws/karpenter/pkg/cloudproviders/aws/apis/v1alpha1"
-	awscloudprovider "github.com/aws/karpenter/pkg/cloudproviders/aws/cloudprovider"
 	"github.com/aws/karpenter/pkg/cloudproviders/common/cloudprovider"
 	"github.com/aws/karpenter/pkg/utils/resources"
 )
@@ -56,7 +60,7 @@ func main() {
 	opts = opts.MustParse()
 	ctx := injection.WithOptions(context.Background(), *opts)
 
-	cp := awscloudprovider.NewCloudProvider(ctx, cloudprovider.Options{})
+	cp := NewAWSCloudProviderForCodeGen()
 	provider := v1alpha1.AWS{SubnetSelector: map[string]string{
 		"*": "*",
 	}}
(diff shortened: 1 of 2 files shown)


spring1843 (approved):
LGTM

hack/docs/instancetypes_gen_docs.go
```diff
@@ -195,3 +199,47 @@ below are the resources available with some assumptions and after the instance o
...
+		StatusCode: http.StatusOK,
+		Proto:      "http/1.0",
+		ProtoMajor: 1,
+		ProtoMinor: 0,
+		Body:       io.NopCloser(bytes.NewBufferString(kubeDNS)),
+	}, nil
+}
+
+func NewAWSCloudProviderForCodeGen() *awscloudprovider.CloudProvider {
+	cs, _ := kubernetes.NewForConfigAndClient(&rest.Config{}, &http.Client{Transport: &kubeDnsTransport{}})
```
ellistarn, line 241: This requires you to have access to an authenticated cluster to generate the docs right? How does this work in the codegen?

hack/docs/instancetypes_gen_docs.go
```diff
@@ -56,7 +60,7 @@ func main() {
 	opts = opts.MustParse()
 	ctx := injection.WithOptions(context.Background(), *opts)
 
-	cp := awscloudprovider.NewCloudProvider(ctx, cloudprovider.Options{})
+	cp := NewAWSCloudProviderForCodeGen()
```
jonathan-innis, line 63: Why not inject `ctx` into this function?

pkg/cloudproviders/aws/cloudprovider/cloudprovider.go
```diff
@@ -293,6 +293,9 @@ func getCABundle(ctx context.Context) *string {
 }
 
 func kubeDNSIP(ctx context.Context, clientSet *kubernetes.Clientset) (net.IP, error) {
+	if clientSet == nil {
+		return nil, fmt.Errorf("no K8s client provided")
+	}
```
ellistarn, line 298: I'd lean away from nil checking for interfaces that we have full control over and don't expose as a core library. This pattern leads us down a road of checking all nilable parameters in every method, which I don't think is sustainable.

hack/docs/instancetypes_gen_docs.go
```diff
@@ -195,3 +199,47 @@ below are the resources available with some assumptions and after the instance o
 		}
 	}
 }
+
+type kubeDnsTransport struct {
+}
+
+const kubeDNS = `{
```
ellistarn, line 206: Thoughts on ripping out most of this config and just configuring what's necessary for the mock to work?

pkg/cloudproviders/aws/cloudprovider/cloudprovider.go
```diff
@@ -293,6 +293,9 @@ func getCABundle(ctx context.Context) *string {
 }
 
 func kubeDNSIP(ctx context.Context, clientSet *kubernetes.Clientset) (net.IP, error) {
+	if clientSet == nil {
```
jonathan-innis, line 296: What is the need for this check? Why not just allow the panic and crash?

jonathan-innis (changes requested):

pkg/cloudproviders/aws/cloudprovider/cloudprovider.go
tzneal: It still crashes, only with a better error message.

hack/docs/instancetypes_gen_docs.go
tzneal: I originally had it in a different package.  There is zero benefit in this scenario to passing the context in, it's just holding up a PR.

hack/docs/instancetypes_gen_docs.go
tzneal: I spent a few minutes and minimized it to this from the full output.  When it broke, I added back what I had removed and quit trying.

ellistarn (approved):

jonathan-innis (approved):

Issue #2480, opened 2022-09-08
Karpenter replicas cause metrics service to wrongly route requests


tlk-kayrros:
## Version

**Karpenter**: v0.16.1

**Kubernetes**: v1.22.11-eks-18ef993

## Expected Behavior
When a controller is using replicas=2 or more the metrics should be available in any of the replicas.

## Actual Behavior
By adding replicas in #2311 the idea was to implement the rolling update strategy and improve the availability of the Karpenter controller. When implementing this change we realised our Karpenter metrics were no longer properly collected by Prometheus. We discovered that the controller which is not currently active doesn't export metrics correctly and exports mainly zeroes and NaN. If the service route the traffic of http-metrics to the inactive controller, it wrongly exposes incomplete metrics causing metric collection disruption.

## Steps to Reproduce the Problem
Add replicas=2 and check metric exposition at the service level.

jonathan-innis:
Hi @tlk-kayrros, can you describe what the specific issue is here? If you are using a `ServiceMonitor`, Prometheus will just use the service for endpoint discovery, but will still scrape metrics from all pods so I don't believe the routing issue that you described above will occur with Prometheus `ServiceMonitor`

dewjam:
Hello @tlk-kayrros ,
I looked into this a bit further and this seems to be related to how our Prometheus scrape targets are configured.  The "karpenter" service is added as a static target which means we are not scraping each Karpenter pod individually. In short, only one of the pods will be scraped at each scrape interval.  

Here is the config I'm referring to:  https://github.com/aws/karpenter/blob/main/website/content/en/preview/getting-started/getting-started-with-eksctl/prometheus-values.yaml#L10-L14

I think the best option here is use the [pod](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#pod) or [endpoint](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#endpoints) service discovery mechanism to scrape both pods at every scrape interval.  Then make sure the predefined queries for Karpenter are aggregating those metrics properly.

dewjam:
Hello tlk-kayrros@,
I just pushed up a PR which changes the scrape target to use endpoint discovery instead of a static scrape target.  Let me know if you think this will solve your problem.

tlk-kayrros:
It will probably solve the issue, yes. Thanks a lot

dewjam:
Closing this out as the PR is now merged.  Feel free to reopen if you feel the issue is not solved.

PR #2710, opened 2022-10-20
fix: updating prom scrape target

dewjam:
Fixes # 
#2480

**Description**
Prior to this change, static scrape targets were configured for the Karpenter service in the Prometheus scrape config.  When only a single Karpenter pod exists, this was fine given Prom would always scrape the one existing Karpenter pod.  When more than one replica exists, Prometheus will potential hit one or the other pod at each scrape interval which can skew metrics.  This change uses the Endpoint discovery mechanism to discover the service endpoints and then scrape them each directly.

* Manually validated that queries used by Grafana are aggregating metrics.  
* I also manually validated that Grafana dashboards still are returning what looks like accurate data.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Update Karpenter scrape target in prometheus to support multiple replicas
```

### DIFF
--- a/website/content/en/preview/getting-started/getting-started-with-eksctl/prometheus-values.yaml
+++ b/website/content/en/preview/getting-started/getting-started-with-eksctl/prometheus-values.yaml
@@ -9,6 +9,12 @@ server:
 
 extraScrapeConfigs: |
     - job_name: karpenter
-      static_configs:
-      - targets:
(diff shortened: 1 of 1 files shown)


ellistarn (approved):

jonathan-innis (approved):
🎉 LGTM

PR #2720, opened 2022-10-21
fix: resolve a bug where AWSNodeTemplate webhooks were no longer registered

ellistarn:
This bug was introduced in the recent code split between karpenter and `karpenter-core`

Fixes # 

**Description**

* `make e2etest`

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -15,13 +15,16 @@ limitations under the License.
 package main
 
 import (
+	"github.com/aws/karpenter-core/pkg/apis"
 	"github.com/aws/karpenter-core/pkg/cloudprovider"
 	"github.com/aws/karpenter-core/pkg/webhooks"
+	"github.com/aws/karpenter/pkg/apis/awsnodetemplate/v1alpha1"
 	awscloudprovider "github.com/aws/karpenter/pkg/cloudprovider"
 	awscontext "github.com/aws/karpenter/pkg/context"
 )
 
 func main() {
+	apis.Resources[v1alpha1.SchemeGroupVersion.WithKind("AWSNodeTemplate")] = &v1alpha1.AWSNodeTemplate{}
 	webhooks.Initialize(func(ctx cloudprovider.Context) cloudprovider.CloudProvider {
 		return awscloudprovider.New(awscontext.NewOrDie(ctx))
 	})
--- a/go.sum
+++ b/go.sum
@@ -66,8 +66,8 @@ github.com/avast/retry-go v3.0.0+incompatible h1:4SOWQ7Qs+oroOTQOYnAHqelpCO0biHS
 github.com/avast/retry-go v3.0.0+incompatible/go.mod h1:XtSnn+n/sHqQIpZ10K1qAevBhOOCWBLXXy3hyiqqBrY=
 github.com/aws/aws-sdk-go v1.44.114 h1:plIkWc/RsHr3DXBj4MEw9sEW4CcL/e2ryokc+CKyq1I=
 github.com/aws/aws-sdk-go v1.44.114/go.mod h1:y4AeaBuwd2Lk+GepC1E9v0qOiTws0MIWAX4oIKwKHZo=
-github.com/aws/karpenter-core v0.0.2-0.20221020221326-8e09dcecd08e h1:LqlztkDuph9HuGWcIqhnBzvu59wq0WhcqmVJmuBzu6w=
-github.com/aws/karpenter-core v0.0.2-0.20221020221326-8e09dcecd08e/go.mod h1:inbUQ43WPgGrDn2YTlHLOvxYmH6Lv5Uhim63H5FeK58=
+github.com/aws/karpenter-core v0.0.2-0.20221021235942-051216a25fdc h1:jYeeAYgQ4oc7gAsq71K0MiAB6fFfWae1Fg4nEbFkcAE=
+github.com/aws/karpenter-core v0.0.2-0.20221021235942-051216a25fdc/go.mod h1:inbUQ43WPgGrDn2YTlHLOvxYmH6Lv5Uhim63H5FeK58=
 github.com/benbjohnson/clock v1.1.0 h1:Q92kusRqC1XV2MjkWETPvjJVqKetz1OzxZB7mHJLju8=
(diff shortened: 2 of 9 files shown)


go.sum
```diff
@@ -66,8 +66,6 @@ github.com/avast/retry-go v3.0.0+incompatible h1:4SOWQ7Qs+oroOTQOYnAHqelpCO0biHS
 github.com/avast/retry-go v3.0.0+incompatible/go.mod h1:XtSnn+n/sHqQIpZ10K1qAevBhOOCWBLXXy3hyiqqBrY=
 github.com/aws/aws-sdk-go v1.44.114 h1:plIkWc/RsHr3DXBj4MEw9sEW4CcL/e2ryokc+CKyq1I=
 github.com/aws/aws-sdk-go v1.44.114/go.mod h1:y4AeaBuwd2Lk+GepC1E9v0qOiTws0MIWAX4oIKwKHZo=
-github.com/aws/karpenter-core v0.0.2-0.20221020221326-8e09dcecd08e h1:LqlztkDuph9HuGWcIqhnBzvu59wq0WhcqmVJmuBzu6w=
```
jonathan-innis, line 69: Why did these entries get removed?

test/go.sum
```diff
@@ -64,8 +64,6 @@ github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d/go.mod h1:rBZYJk5
 github.com/antihax/optional v1.0.0/go.mod h1:uupD/76wgC+ih3iEmQUL+0Ugr19nfwCT1kdvxnR2qWY=
 github.com/aws/aws-sdk-go v1.44.114 h1:plIkWc/RsHr3DXBj4MEw9sEW4CcL/e2ryokc+CKyq1I=
 github.com/aws/aws-sdk-go v1.44.114/go.mod h1:y4AeaBuwd2Lk+GepC1E9v0qOiTws0MIWAX4oIKwKHZo=
-github.com/aws/karpenter-core v0.0.2-0.20221020221326-8e09dcecd08e h1:LqlztkDuph9HuGWcIqhnBzvu59wq0WhcqmVJmuBzu6w=
```
jonathan-innis, line 67: Why did these entries get removed?

test/pkg/environment/common/expectations.go
```diff
@@ -34,11 +33,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const TestLabelName = "testing.karpenter.sh/test-id"
-
 func (env *Environment) ExpectCreated(objects ...client.Object) {
 	for _, object := range objects {
-		object.SetLabels(lo.Assign(object.GetLabels(), map[string]string{TestLabelName: env.ClusterName}))
```
jonathan-innis, line 41: Why are we removing these labels?

cmd/webhook/main.go
```diff
@@ -15,13 +15,16 @@ limitations under the License.
...
+	"github.com/aws/karpenter-core/pkg/apis"
 	"github.com/aws/karpenter-core/pkg/cloudprovider"
 	"github.com/aws/karpenter-core/pkg/webhooks"
+	"github.com/aws/karpenter/pkg/apis/awsnodetemplate/v1alpha1"
 	awscloudprovider "github.com/aws/karpenter/pkg/cloudprovider"
 	awscontext "github.com/aws/karpenter/pkg/context"
 )
 
 func main() {
+	apis.Resources[v1alpha1.SchemeGroupVersion.WithKind("AWSNodeTemplate")] = &v1alpha1.AWSNodeTemplate{}
```
jonathan-innis, line 27: This feels like a strange, round-a-bout way to register the resources, is there a better registration flow that we could introduce?

pkg/apis/apis.go
```diff
@@ -33,9 +30,4 @@ var (
 	)
 	// AddToScheme may be used to add all resources defined in the project to a Scheme
 	AddToScheme = Builder.AddToScheme
-	// Resources defined in the project
-	Resources = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
```
jonathan-innis, line 37: Why not register these resources with an init or similar?

test/suites/integration/webhook_test.go
```diff
@@ -28,170 +28,162 @@ import (
...
-			provider := awstest.AWSNodeTemplate(v1alpha1.AWSNodeTemplateSpec{AWS: v1alpha1.AWS{
-				SecurityGroupSelector: map[string]string{"karpenter.sh/discovery": env.ClusterName},
-				SubnetSelector:        map[string]string{"karpenter.sh/discovery": env.ClusterName},
-			}})
-			provisioner := test.Provisioner(test.ProvisionerOptions{
-				ProviderRef: &v1alpha5.ProviderRef{Name: provider.Name},
-			})
-			env.ExpectCreated(provisioner)
-			env.ExpectFound(provisioner)
+	Context("Provisioner", func() {
```
jonathan-innis, line 31: I think I'm missing where we actually create the AWSNodeTemplate in some of these tests. Seems like we don't, shouldn't we be?

cmd/webhook/main.go
ellistarn: I want to overhaul the webhooks in a followon when I combine things. This is largely to unblock next week's release if there is one. I'm thinking about something like
```
operator.New().RegisterWebhooks(...).RegisterControllers(...) 
```
so you can do a builder pattern in both neutral and specific code.

test/go.sum
ellistarn: No clue.. just go mod tidy.

test/pkg/environment/common/expectations.go
ellistarn: They're added here now: https://github.com/aws/karpenter-core/pull/27. We could add in cluster name, but right now, our tests assume non-parallel (and probably need to to avoid provisioner collision), so the label value doesn't really matter.

test/suites/integration/webhook_test.go
ellistarn: Validation for the provisioner doesn't care about the AWS Node Template. Resources should never validate each other's existence.

jonathan-innis (approved):
LGTM 🚀

jonathan-innis (approved):
LGTM 🚀

PR #2728, opened 2022-10-24
feat: Combine webhooks and controllers into a single binary

ellistarn:
Fixes # 

**Description**

@jonathan-innis and I discovered that we could avoid the --kubeconfig conflict by relying on controllerruntime libraries to initialize the rest.Config for both webhooks and the controller. This enables us to collapse these into a single pod, simplifying operations and release infrastructure.

Some caveats:

- Helm chart flags to configure the webhook will no longer work
- Webhook logging has been turned down to ERROR, since INFO/Debug is so spammy.

Partner PR: https://github.com/aws/karpenter-core/pull/29

* `make test`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```


jonathan-innis (dismissed):
🚀  LGTM!

jonathan-innis:
🎸 🎸 🎸 🎸 🎸 🎸 🎸 🎸

jonathan-innis (approved):
LGTM

ellistarn:
This PR was WIP and should not have been merged.

maximmold:
I am having an issue as well with a connection refused error presumably in 0.19.3 presumably due to hostNetwork port overriding being broken in this PR. This documentation describes the environment variable WEBHOOK_PORT which can override the default. Even with this set, it still uses the default of 8443.

We are using `hostNetwork: true` to get around issues with the control plane contacting the webhook. When using cilium, the control plane doesn't seem to be able to reach the service otherwise. I'd honestly prefer to have the opposite direction where the webhook lives in a separate pod so I can reduce the radius of ports that end up having a hostPort when using host networking. Perhaps multiple CNIs would be a better option?

anguslees:
> When using cilium, the control plane doesn't seem to be able to reach the service otherwise

If you're using encapsulation, don't do that.  :)
The EKS apiserver won't know about your CNI's encapsulation scheme and won't be able to talk to pod IPs directly - as you've discovered.  If this is your problem, and you switch to 'eni' mode with cilium (for example), then this will work fine.

If you stick with encapsulation, you'll need to use hostNetwork (ie: listen on host IPs directly), or hostPort (ie: DNAT a port from the local host to pod) - and modify the webhook config accordingly.  Fwiw, EKS uses `apiserver --enable-aggregator-routing` so you _can't_ use Service type=LoadBalancer or NodePort (iirc).

PR #2731, opened 2022-10-25
feat: Add global `aws.tags...` settings to AWS config

jonathan-innis:
Fixes # 

**Description**

- Adds `aws.tags` to the `karpenter-global-settings` to provision all AWS infrastructure resources (LaunchTemplates, Instances, SQS queues, etc...) with these global tag values

* `make battletest`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```


pkg/apis/config/settings/settings.go
```diff
@@ -0,0 +1,126 @@
...
+	if err := configmap.Parse(cm.Data,
+		AsMap("aws.tags", &s.Tags),
+	); err != nil {
+		// Failing to parse means that there is some error in the Settings, so we should crash
+		panic(fmt.Sprintf("parsing config data, %v", err))
+	}
+	return s, nil
+}
+
+func ToContext(ctx context.Context, s Settings) context.Context {
```
ellistarn, line 85: WDYT about making a generic for this? We could use a `*new(T)` for the context key. You have to make sure the injection is a singleton, but should work for most cases.

pkg/test/settings.go
```diff
@@ -0,0 +1,23 @@
...
+limitations under the License.
+*/
+
+package test
+
+import (
+	awssettings "github.com/aws/karpenter/pkg/apis/config/settings"
+)
+
+func Settings() awssettings.Settings {
```
ellistarn, line 21: Is there a way to collapse this with the API settings package instead of having a test specific one?

website/content/en/preview/AWS/configuration.md
```diff
@@ -0,0 +1,41 @@
+---
```
ellistarn, line 1: I'd love to eventually deprecate this section and maintain an AWS specific docs site and neutral docs site. Can you collapse with tasks/Configuration (and perhaps rename to Operator Configuration)

pkg/cloudprovider/cloudprovider.go
```diff
@@ -60,6 +61,7 @@ const (
 func init() {
 	v1alpha5.NormalizedLabels = lo.Assign(v1alpha5.NormalizedLabels, map[string]string{"topology.ebs.csi.aws.com/zone": v1.LabelTopologyZone})
 	coreapis.Resources = lo.Assign(coreapis.Resources, apis.Resources)
+	coreapis.Settings = coreapis.Settings.Union(apis.Settings)
```
ellistarn, line 64: musing: I'd love to find a better path than this global injection, but this is fine for now.

jonathan-innis: I'm curious why you think this global injection is bad? Scheme registration is done through the same mechanism and I'd argue these are very close concepts.

pkg/apis/config/settings/settings.go
jonathan-innis: I messed with it a bit and I'm not convinced I like it. The code seems to become less idiomatic like


```go
context.FromContext[settings.Settings](ctx).BatchIdleDuration.Duration

settings.FromContext(ctx).BatchIdleDuration.Duration
```

I could maybe go either way, but I'm not seeing a huge benefit to making it generic and the type checking actually becomes more confusing in a lot of ways

pkg/cloudprovider/cloudprovider.go
ellistarn: Injection hides where values come from. Sometimes this is good (encapsulation) sometimes it's bad (obfuscation). I'm fine with it here, but it feels a touch obfuscated to me.

ellistarn (approved):

PR #2770, opened 2022-11-01
fix: Split the CI workflow into ci-test and ci to eliminate redundant work…

spring1843:
… and speed up testing feedback

Fixes # 

**Description**

* Separate CI activities (license check, lint and etc..) other than testing into a different workflow, these activities need to happen only once whereas before they were happening multiple times on each k8s version matrix  (5 times) which was prolonging test execution
* Separate CI tests into a different workflow so that they can start to execute in parallel and immediately without having to wait for other checks, allowing them to finish quicker and notify the user faster 
* Upload coverage data to coveralls only once, previously this was happening on each
* Fix formatting for conditions in `release-tag.yml`
* No logical change to `make ci` to preserve the same dev experience and limit the changes to Workflows

* Different fork

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


.github/workflows/ci-test.yml
```diff
@@ -15,9 +15,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: ./.github/actions/install-go-and-dependencies
-    - run: make ci
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - run: make ci-test
     - uses: shogo82148/actions-goveralls@v1
+      if: env.K8S_VERSION == '1.24.x'
```
ellistarn, line 20: Why restrict this?

spring1843: Because otherwise it's going to do it as many times (5 times), while this needs to happen onyly once

Makefile
```diff
@@ -37,7 +37,11 @@ help: ## Display help
 
 presubmit: verify test ## Run all steps in the developer loop
 
-ci: toolchain verify licenses vulncheck battletest coverage ## Run all steps used by continuous integration
+ci-test: battletest coverage ## Runs tests and submits coverage
+
+ci-non-test: verify licenses vulncheck ## Runs checks other than tests
```
ellistarn, line 42: Trying to think of more idiomatic names for this. I'd like to make sure we share as much as possible between karpenter and karpenter-core. 

I like this separation, since it enables developers to run `ci-test` locally without significant time investment of the license/vuln checks. I think we might be able to collapse `battletest` and `test`, since I'm not sure it actually takes very much more time 

`presubmit` Runs codegen, linting, testing, coverage, etc. It's fast 
`presubmit-slow` Runs license checks, vulnerabilities, slow stuff.  

This makes me think that we might actually want to run the slow stuff on postsubmit. These fail so rarely. It would be really nice to dramatically improve our merge times.

That would leave us with

`presubmit` Devs run before submission, Runs before CI checks. SLA <5min, ideally 3m?
`postsubmit` GH runs after submission, may take longer

Happy to chat offline

.github/workflows/ci-test.yml
ellistarn: We're going to need to constantly maintain this whenever we change Kubernetes versions.

spring1843: We only need to change it if we get rid of `1.24.x` I'll have to figure out how to access the dimensions from here and reference the last one instead of this to avoid that.

Makefile
spring1843: Those are good idea, we also don't really need vulnerability scanning right here as we are doing it in another workflow. I'll do them in a follow up PR if you don't mind.

ellistarn (approved):

PR #2778, opened 2022-11-01
test: Always print test logs, even if cleanup fails

jonathan-innis:
Fixes # 

**Description**

- Always prints test logs now since log dump is in a separate `AfterEach` block
- Always do ForceCleanup to avoid test pollution

*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/test/pkg/environment/aws/setup.go
+++ b/test/pkg/environment/aws/setup.go
@@ -34,22 +34,25 @@ var (
 )
 
 func (env *Environment) BeforeEach(opts ...common.Option) {
-	options := common.ResolveOptions(opts)
-	if options.EnableDebug {
-		fmt.Println("------- START AWS BEFORE -------")
-		defer fmt.Println("------- END AWS BEFORE -------")
-	}
 	env.ExpectSettingsCreatedOrUpdated(settings.Registration.DefaultData, awssettings.Registration.DefaultData)
 	env.Environment.BeforeEach(opts...)
 }
 
-func (env *Environment) AfterEach(opts ...common.Option) {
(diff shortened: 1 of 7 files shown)


test/pkg/environment/aws/setup.go
```diff
@@ -34,22 +34,25 @@ var (
 )
 
 func (env *Environment) BeforeEach(opts ...common.Option) {
-	options := common.ResolveOptions(opts)
-	if options.EnableDebug {
-		fmt.Println("------- START AWS BEFORE -------")
```
njtran, line 39: If we're keeping the AWS after log line, why are we deleting the before?

test/pkg/environment/common/setup.go
```diff
@@ -233,28 +259,26 @@ func (env *Environment) CleanupObjects(cleanableObjects []functional.Pair[client
 	Expect(env.Client.List(env, namespaces)).To(Succeed())
 	wg := sync.WaitGroup{}
 	for _, p := range cleanableObjects {
-		if !lo.ContainsBy(options.IgnoreCleanupFor, func(o client.Object) bool { return reflect.TypeOf(o) == reflect.TypeOf(p.First) }) {
```
njtran, line 236: Seems odd that we'd do this in the first place. Are you aware if we were doing any of this? If so, might want to keep this logic in just for good measure.

njtran (approved):
LGTM, just two nits

test/pkg/environment/common/setup.go
jonathan-innis: Wasn't used, was intending to use it but then found a different way

PR #2781, opened 2022-11-02
test: Fix flaky expiration test

jonathan-innis:
Fixes # 

**Description**

- Expiration tests are flaking because a previous test is provisioning a node in the middle of two tests. This causes an extra node to exist on the cluster when the test starts

* `FOCUS=Expiration make e2etests`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/test/pkg/environment/common/setup.go
+++ b/test/pkg/environment/common/setup.go
@@ -229,13 +229,13 @@ func (env *Environment) ForceCleanup(opts ...Option) {
 
 	// Delete all the nodes if they weren't deleted by the provisioner propagation
 	Expect(env.Client.DeleteAllOf(env, &v1.Node{},
-		client.HasLabels([]string{test.DiscoveryLabel}),
+		client.HasLabels([]string{v1alpha5.ProvisionerNameLabelKey}),
 		client.PropagationPolicy(metav1.DeletePropagationForeground),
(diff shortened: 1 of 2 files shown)


test/pkg/environment/common/setup.go
```diff
@@ -229,13 +229,13 @@ func (env *Environment) ForceCleanup(opts ...Option) {
 
 	// Delete all the nodes if they weren't deleted by the provisioner propagation
 	Expect(env.Client.DeleteAllOf(env, &v1.Node{},
-		client.HasLabels([]string{test.DiscoveryLabel}),
```
bwagner5, line 232: Was this intended to be a different discovery label than the provisioner name label when we looked at doing concurrent tests in the same cluster?

bwagner5 (approved):
LGTM, just one question on the discovery label

test/pkg/environment/common/setup.go
ellistarn: Yeah it's scoped to a test execution. Every object has this injected to it.

jonathan-innis: Nodes don't have this label since they aren't created directly by the tests.

PR #2793, opened 2022-11-03
fix: kube-dns: allow setting ip for clusters without kube-dns

grosser:
Fixes https://github.com/aws/karpenter/issues/2787

* deployed to cluster with hardcoded coredns ip

- [X] No

**Release Note**

```release-note
- allow setting dns ip via KARPENTER_KUBERNETES_DNS_IP for clusters without kube-dns
```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"os"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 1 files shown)


bwagner5:
I'm thinking rather than adding a flag, we could switch the discovery logic to use the `kubernetes` service instead of kube-dns. We assumed everyone was running `kube-dns`, but I guess we were wrong in that assumption :D

grosser:
kubernetes service should work fine

pkg/cloudprovider/cloudprovider.go
```diff
@@ -166,6 +167,11 @@ func getCABundle(ctx context.Context, restConfig *rest.Config) (*string, error)
 }
 
 func kubeDNSIP(ctx context.Context, kubernetesInterface kubernetes.Interface) (net.IP, error) {
+	envIp := os.Getenv("KARPENTER_KUBERNETES_DNS_IP")
```
ellistarn, line 170: Can we expose this via the recently introduced Settings configmap? This is where all other env vars are defined. It has the benefit of hot reloading without restarting karpenter.

bwagner5:
let us know if this PR fixes it for you: https://github.com/aws/karpenter/pull/2794

grosser:
that should work

PR #2798, opened 2022-11-04
fix: Fix cloudformation and sqs queue backoff

jonathan-innis:
Fixes # 

**Description**

- Fixes cloudformation template to use tags
- Fixes backoff when queue was recently deleted

* 

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/pkg/controllers/nodetemplate/infrastructure.go
+++ b/pkg/controllers/nodetemplate/infrastructure.go
@@ -29,6 +29,7 @@ import (
 	awssettings "github.com/aws/karpenter/pkg/apis/config/settings"
 	"github.com/aws/karpenter/pkg/apis/v1alpha1"
 	"github.com/aws/karpenter/pkg/controllers/providers"
+	"github.com/aws/karpenter/pkg/errors"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/controllers/nodetemplate/infrastructure.go
```diff
@@ -67,6 +68,10 @@ func (i *InfrastructureReconciler) Reconcile(ctx context.Context, nodeTemplate *
 	} else if len(list.Items) >= 1 {
 		if i.lastInfrastructureReconcile.Add(time.Minute * 5).Before(time.Now()) {
 			if err := i.CreateInfrastructure(ctx); err != nil {
+				if errors.IsRecentlyDeleted(err) {
```
bwagner5, line 71: can we mitigate this by using some randomness in the queue name?

jonathan-innis: We rely on the queue name. It's not tag-based, it's static so we can't add randomness to the name.

bwagner5 (approved):
LGTM

PR #2812, opened 2022-11-07
 feat:  Drift controller implementation

ra-grover:
Fixes # 

**Description**
This PR implements Drift in the AWS Cloud Provider
- Node Controller that annotates nodes as drifted when the AMI used doesn't match Provisioning specification
- Updates karpenter-core dependency to https://github.com/aws/karpenter-core/pull/124
- Implements the IsMachineDrifted() cloud provider function
- Adds in e2e tests for drift
- Includes a fallback call to EC2 when the machine does not have the AMI label
- Decouples some AMI specific logic from the launch template provider

This was tested in an EKS cluster. Following things were tested:
- `make apply` then `make e2etests`
- `make presubmit

The implementation was able to mark the already provisioned nodes as Drifted. 

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
```


njtran (commented):
Leaving some comments to checkpoint my progress

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,120 @@
...
+				}
+				for _, provisioner := range provisionerList.Items {
+					if provisioner.Spec.ProviderRef != nil && provisioner.Spec.ProviderRef.Name == x.Name {
+						reconcileList = append(reconcileList, reconcile.Request{NamespacedName: types.NamespacedName{Name: provisioner.GetName()}})
+					}
+				}
+				return reconcileList
+			}),
+		).WithOptions(controller.Options{MaxConcurrentReconciles: 10})
+}
```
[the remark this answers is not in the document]
ra-grover: Is there any particular reason, if primary watch should be for the nodes instead of the provisioner ? 
I think it's better to provide the cloudprovider a collection of nodes to process on together rather than individual nodes. In future we might have a case where we need to query node attributes from the cloudprovider(for example DescibeInstances), and collection of nodes gives us the opportunity to batch them all in a single api call.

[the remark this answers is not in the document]
ra-grover: To me, the source of truth for the state of the a is the `Provisioner` or `AwsNodeTemplate` and not the actual Node itself. 
Please correct me if I am wrong :)

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,120 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	"time"
+)
+
+const controllerName = "drift"
+
+type DriftController struct {
```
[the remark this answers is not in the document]
ra-grover: Cool, thanks.

pkg/controllers/controllers.go
```diff
@@ -17,9 +17,13 @@ package controllers
...
 	"github.com/aws/karpenter-core/pkg/controllers/state"
 	"github.com/aws/karpenter-core/pkg/operator/controller"
+	"github.com/aws/karpenter/pkg/cloudprovider"
 	awscontext "github.com/aws/karpenter/pkg/context"
+	"github.com/aws/karpenter/pkg/controllers/drift"
 )
 
-func NewControllers(ctx awscontext.Context, cluster *state.Cluster) []controller.Controller {
-	return []controller.Controller{}
+func NewControllers(ctx awscontext.Context, cluster *state.Cluster, cloudProvider *cloudprovider.CloudProvider) []controller.Controller {
```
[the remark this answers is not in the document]
ra-grover: Done, thanks.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,120 @@
...
+// NewController constructs a controller instance
+func NewController(kubeClient client.Client, cluster *state.Cluster, cloudProvider *cloudprovider.CloudProvider) *DriftController {
+	return &DriftController{
+		kubeClient:    kubeClient,
+		cluster:       cluster,
+		cloudProvider: cloudProvider,
+	}
+}
+
+func (r DriftController) Reconcile(ctx context.Context, req controllerruntime.Request) (reconcile.Result, error) {
```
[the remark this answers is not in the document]
ra-grover: Reverted on the other thread for the discussion on primarily reconciling node vs provisioner.

pkg/cloudprovider/launchtemplate.go
```diff
@@ -115,31 +157,57 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, provider *v1alpha1.AWS
...
+		amiId = ami.(string)
+		logging.FromContext(ctx).Debugf("Found %s from cache for launch template %s", ami, launchTemplateName)
+		return amiId, nil
 	}
-	launchTemplates := map[string][]cloudprovider.InstanceType{}
-	for _, resolvedLaunchTemplate := range resolvedLaunchTemplates {
-		// Ensure the launch template exists, or create it
-		ec2LaunchTemplate, err := p.ensureLaunchTemplate(ctx, resolvedLaunchTemplate)
+
+	output, err := p.ec2api.DescribeLaunchTemplatesWithContext(ctx, &ec2.DescribeLaunchTemplatesInput{
```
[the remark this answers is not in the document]
ra-grover: Removed, as we discussed on DM.

pkg/cloudprovider/amidrift.go
```diff
@@ -0,0 +1,47 @@
...
+	aws, err := ad.c.getProvider(ad.ctx, provisioner.Spec.Provider, provisioner.Spec.ProviderRef)
+	if err != nil {
+		return []v1.Node{}
+	}
+	amis, err := ad.c.instanceProvider.launchTemplateProvider.GetAmisForProvisioner(ad.ctx, aws, provisioner.Spec.ProviderRef, instanceTypes)
+	if err != nil {
+		logging.FromContext(ad.ctx).Errorf("getting drift amis from provisioner %w", err)
+		return []v1.Node{}
+	}
+	if len(amis) == 0 {
```
njtran, line 33: We should make this an invariant in the code so that we don't have to worry about this.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,6 +100,32 @@ func New(ctx awscontext.Context) *CloudProvider {
 	}
 }
 
+func (c *CloudProvider) GetDriftedNodes(ctx context.Context, provisioner *v1alpha5.Provisioner, nodes []v1.Node) []v1.Node {
+	driftedNodesMap := map[string]bool{}
+	var driftedNodes []v1.Node
+	//Consider if provisioner should be a part of this interface
```
njtran, line 106: I think this is a good direction to be thinking about, but an in-line interface here might be setting up too much structure to the code that we may not need in the future.

Is it possible to restructure this code as a `IsNodeDrifted()` so that we just simply return a boolean? This way we can leave getting any node objects to the controller that wants to call this function, and this helper function can be a way to get the information.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,6 +100,32 @@ func New(ctx awscontext.Context) *CloudProvider {
...
+		//Execute Drifters
+		for _, node := range drifter.GetDriftedNodes(nodes) {
+			//Check first in the map, one node can be drifted by multiple drifter implementations
+			if _, ok := driftedNodesMap[node.Name]; !ok {
+				driftedNodes = append(driftedNodes, node)
+			}
+			driftedNodesMap[node.Name] = true
+		}
+	}
+	logging.FromContext(ctx).Infof("Found %d drifted nodes", len(driftedNodes))
```
njtran, line 125: My hunch is that this will be unnecessarily noisy in logs. I don't think that we'll need to log the number of drifted nodes we find, since this could be called many times in rapid succession. I think it'd be better if we just log if we annotate any node as drifted in the controller.go file. WDYT?

pkg/cloudprovider/amidrift.go
```diff
@@ -0,0 +1,47 @@
...
+		//Ideally this would only happen if the provisioner/awsnodetemplate are wrongly configured, just return with no drift.
+		logging.FromContext(ad.ctx).Infof("no amis while calculating drift")
+		return []v1.Node{}
+	}
+	logging.FromContext(ad.ctx).Infof("%v amis found for drift", amis)
```
njtran, line 38: Likewise to my comment about logging when we find drifted nodes: we should probably just log when we label a node as drifted, since finding a node as drifted doesn't necessarily equal something changing in the cluster until the deprovisioner catches it.

pkg/cloudprovider/amidrift.go
```diff
@@ -0,0 +1,47 @@
...
+	return lo.Filter(nodes, func(node v1.Node, _ int) bool {
+		nodeAmi, exists := node.Labels[v1alpha1.LabelInstanceAMIID]
+		if nodeAmi == "" || !exists {
```
njtran, line 41: I think we'll need to assert that it fits a `ami-%s` regex to ensure it fits any string-based requirements. Additionally, if possible, we should think about doing an EC2 DescribeInstances call as a fallback in case the node isn't labeled with the AMI ID. 

In the case that a user decides to annotate their node with the Karpenter label, they shouldn't need to be aware of the fact that AMI ID needs to be annotated for the node to be considered for drift, so a call to get the Image ID probably saves us from some errors.

pkg/cloudprovider/amidrift.go
```diff
@@ -0,0 +1,47 @@
...
+			logging.FromContext(ad.ctx).Debugf("No ami id found on node %s, while calculating drift", node.Name)
```
njtran, line 42: Same here about logging.

njtran (commented):
Nice Work! I have some structural comments. Let's talk about this on slack.

pkg/cloudprovider/amidrift.go
FernandoMiguel: Nit: should we use `AMI` and `AMIs` instead of their lowercase counterparts?

pkg/cloudprovider/amidrift.go
ra-grover: Done that but need your inputs. I dont think I have that piece perfectly in place

pkg/cloudprovider/amidrift.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/aws/karpenter-core/pkg/apis/provisioning/v1alpha5"
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+	"github.com/samber/lo"
+	v1 "k8s.io/api/core/v1"
+	"knative.dev/pkg/logging"
+)
+
+type AmiDrifter struct {
+	provisioner *v1alpha5.Provisioner
+	ctx         context.Context
```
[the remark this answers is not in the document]
ra-grover: Removed this drifter altogether without an inline interface

pkg/cloudprovider/launchtemplate.go
```diff
@@ -102,6 +104,39 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, provider *v1alpha1.AWS
...
+		ec2LaunchTemplate, err := p.ensureLaunchTemplate(ctx, resolvedLaunchTemplate)
+		if err != nil {
+			return nil, err
+		}
+		launchTemplates[*ec2LaunchTemplate.LaunchTemplateName] = resolvedLaunchTemplate.InstanceTypes
+	}
+	return launchTemplates, nil
+}
+
+func (p *LaunchTemplateProvider) GetAmisForProvisioner(ctx context.Context, ant *v1alpha1.AWS, providerRef *v1alpha5.ProviderRef, instanceTypes []cloudprovider.InstanceType) ([]string, error) {
```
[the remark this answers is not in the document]
ra-grover: Done

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,30 +103,31 @@ func New(ctx awscontext.Context) *CloudProvider {
...
-		},
-	} {
-		//Execute Drifters
-		for _, node := range drifter.GetDriftedNodes(nodes) {
-			//Check first in the map, one node can be drifted by multiple drifter implementations
-			if _, ok := driftedNodesMap[node.Name]; !ok {
-				driftedNodes = append(driftedNodes, node)
-			}
-			driftedNodesMap[node.Name] = true
+func (c *CloudProvider) IsNodeDrifted(ctx context.Context, provisioner *v1alpha5.Provisioner, node v1.Node) bool {
```
ellistarn, line 106: I'm not convinced that this should be part of the cloud provider. Is there a reason we can't encapsulate all of this logic into the drift controller?

pkg/controllers/drift/controller.go
```diff
@@ -39,74 +38,94 @@ func NewController(kubeClient client.Client, cloudProvider *cloudprovider.CloudP
 
 func (d Drift) Reconcile(ctx context.Context, req controllerruntime.Request) (reconcile.Result, error) {
 	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named(controllerName).With("provisioner", req.Name))
-
-	provisioner := &v1alpha5.Provisioner{}
-	if err := d.kubeClient.Get(ctx, req.NamespacedName, provisioner); err != nil {
+	node := &v1.Node{}
+	if err := d.kubeClient.Get(ctx, req.NamespacedName, node); err != nil {
```
ellistarn, line 42: Don't we want to retry on error here?

njtran: Yeah I was going to say the same. We shouldn't be masking any errors without a good reason.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,135 @@
...
+
+	provisionerName, provisionerExists := node.Labels[v1alpha5.ProvisionerNameLabelKey]
+	if !provisionerExists {
+		//Only karpenter owned Nodes.
+		return reconcile.Result{}, nil
+	}
+
+	provisioner := &v1alpha5.Provisioner{}
+	if err := d.kubeClient.Get(ctx, types.NamespacedName{Name: provisionerName}, provisioner); err != nil {
+		return reconcile.Result{}, nil
```
njtran, line 58: Same for the error here, we shouldn't mask any errors here.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,135 @@
...
+	node := &v1.Node{}
+	if err := d.kubeClient.Get(ctx, req.NamespacedName, node); err != nil {
+		return reconcile.Result{}, nil
+	}
+
+	if _, drifted := node.Annotations[v1alpha5.DriftedAnnotationKey]; drifted {
+		return reconcile.Result{}, nil
+	}
+
+	provisionerName, provisionerExists := node.Labels[v1alpha5.ProvisionerNameLabelKey]
```
njtran, line 50: We should check to see if the provisioner label exists first, since that's the first condition where we know if we can do anything to this node.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,135 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	provisioner := &v1alpha5.Provisioner{}
+	if err := d.kubeClient.Get(ctx, types.NamespacedName{Name: provisionerName}, provisioner); err != nil {
+		return reconcile.Result{}, nil
+	}
+
+	drifted := d.cloudProvider.IsNodeDrifted(ctx, provisioner, *node)
+	if drifted {
```
njtran, line 62: nit: you should be able to in-line this.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,135 @@
...
+	return reconcile.Result{RequeueAfter: 30 * time.Minute}, nil
+}
+
+func (d Drift) Builder(ctx context.Context, m manager.Manager) operatorcontroller.Builder {
+	if err := m.GetFieldIndexer().IndexField(ctx, &v1alpha5.Provisioner{}, ".spec.providerRef.name", func(rawObj client.Object) []string {
+		provisioner := rawObj.(*v1alpha5.Provisioner)
+		return []string{provisioner.Spec.ProviderRef.Name}
+	}); err != nil {
+		//todo:Need inputs on this, or this indexer needs to be put somewhere else or removed altogether
+	}
```
njtran, line 79: I'm not sure I understand what you're doing here. Is this another pattern you've seen somewhere else? I think we should be able to do everything within the return statement you have at line 81.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -111,6 +141,31 @@ func checkEC2Connectivity(ctx context.Context, api *ec2.EC2) error {
 	return err
 }
 
+// todo:Need some input on this from maintainers.
+func getAmiFromEC2Instance(ctx context.Context, api *ec2.EC2, node v1.Node) (string, error) {
```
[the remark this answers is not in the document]
njtran: We should also be able to use the https://github.com/aws/karpenter/blob/main/pkg/cloudprovider/instance.go#L292 getInstance() call here and parse the values we need. Here's the spec if you need it. https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2@v1.70.0/types#Instance

njtran (commented):
Looking better! Good work

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,6 +103,33 @@ func New(ctx awscontext.Context) *CloudProvider {
 	}
 }
 
+func (c *CloudProvider) IsNodeDrifted(ctx context.Context, provisioner *v1alpha5.Provisioner, node v1.Node) bool {
+	instanceTypes, _ := c.GetInstanceTypes(context.Background(), provisioner)
```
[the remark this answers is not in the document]
ra-grover: Done, silly mistake! You want me to capture it's error too ? For now I have just used `_`.

pkg/controllers/drift/controller.go
ra-grover: So its related to `d.kubeClient.List(ctx, provisioners, client.MatchingFields{".spec.providerRef.name": o.GetName()})` which is down below in `Watches(
			&source.Kind{Type: &v1alpha1.AWSNodeTemplate{}},`, if the field is not indexed, it wont work and returns an error that the index doesn't exist. 

Is there a way we can do away with this index ? We need it to get provisioners that are related to a specific `AwsNodeTemplate`.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,6 +103,33 @@ func New(ctx awscontext.Context) *CloudProvider {
...
+		return false
+	}
+
+	amis, err := c.instanceProvider.launchTemplateProvider.GetAmisForProvisioner(ctx, aws, provisioner.Spec.ProviderRef, instanceTypes)
+	if err != nil {
+		logging.FromContext(ctx).Errorf("getting drift amis %w", err)
+		return false
+	}
+
+	nodeAmi, _ := node.Labels[v1alpha1.LabelInstanceAMIID]
```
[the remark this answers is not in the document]
ra-grover: I have removed the `isValidAmi, _ := regexp.MatchString("ami-[^\\s]*", nodeAmi)` then, does that sound good ?

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (d Drift) Reconcile(ctx context.Context, req controllerruntime.Request) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named(controllerName).With("provisioner", req.Name))
+	node := &v1.Node{}
+	if err := d.kubeClient.Get(ctx, req.NamespacedName, node); err != nil {
+		return reconcile.Result{}, nil
+	}
+
+	if _, drifted := node.Annotations[v1alpha5.DriftedAnnotationKey]; drifted {
```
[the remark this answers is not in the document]
ra-grover: You want me to put drifted as a label ? 
And yeah sure I can check the value.

pkg/controllers/drift/controller.go
ra-grover: Sorry missed those errors.

pkg/cloudprovider/cloudprovider.go
ra-grover: I think the drift is tied to a cloudprovider in the sense that it needs information related to what cloudprovider has. Going forward, multiple drifters can be implemented in this function which need more access to cloudprovider. Ami is just one of the things which was rooted in cloudprovider -> instanceProvider -> launchTemplateProvider -> amiFamily.

ra-grover: I am okay to convert it to `GetAmiForProvisioner` sorts, but I think we will have to expose more functions like this to get attributes related to cloudprovider going ahead.

pkg/controllers/drift/controller.go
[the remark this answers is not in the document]
ra-grover: And also, not sure if you have noticed but `v1alpha5.DriftedAnnotationKey` belongs to karpenter-core. I hope that's the right thing?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,6 +103,33 @@ func New(ctx awscontext.Context) *CloudProvider {
 	}
 }
 
+func (c *CloudProvider) IsNodeDrifted(ctx context.Context, provisioner *v1alpha5.Provisioner, node v1.Node) bool {
```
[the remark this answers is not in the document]
ra-grover: Done.

pkg/controllers/drift/controller.go
ra-grover: Done, sounds good.

pkg/controllers/drift/controller.go
ra-grover: done.

pkg/controllers/drift/controller.go
ra-grover: > Don't we want to retry on error here?

Should I check for a not found error, if not then requeue ?

pkg/cloudprovider/cloudprovider.go
ellistarn: Sorry -- I don't have this fully clear in my head, so I may be giving bad advice. 

Previously, this cloudprovider.go file was a strict implementation of https://github.com/aws/karpenter-core/blob/main/pkg/cloudprovider/types.go#L47. I'd like to keep these 1:1 if at all possible. This raises the question of whether or not this drift controller implementation should be vendor neutral (e.g. karpenter core) or vendor specific.

I could see an argument that the drift logic could be standardized across vendors, with this as the extensibility point.

I see our options as:
1. Make drift an AWS specific controller for now (as you have it here), and promote it to be neutral when it makes sense
2. Make IsDrifted(v1.Node) part of the cloudprovider interface, and implement a vendor neutral controller that calls it

The more I think on this, the more I have a slight preference for #2. In either solution, I think that we should communicate this state via a node annotation as you have it today.

pkg/cloudprovider/amifamily/resolver.go
```diff
@@ -116,6 +116,15 @@ func New(kubeClient client.Client, ssm ssmiface.SSMAPI, ec2api ec2iface.EC2API,
 	}
 }
 
+func (r Resolver) ResolveAmis(ctx context.Context, provider *v1alpha1.AWS, providerRef *v1alpha5.ProviderRef, instanceTypes []cloudprovider.InstanceType, options *Options) (map[string][]cloudprovider.InstanceType, error) {
```
ellistarn, line 119: My gut tells me that we should resolve provider/providerRef higher up in the stack.

ellistarn: Thinking on this, I'm struggling to see the point of this function since it has the same signature as r.amiProvider.Get. Is there a reason we can't call GetAMIFamily at the callsites?

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,133 @@
...
+		return reconcile.Result{}, fmt.Errorf("getting provisioner, %w", err)
+	}
+
+	if drifted, err := d.cloudProvider.IsNodeDrifted(ctx, provisioner, *node); err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting drift for node, %w", err)
+	} else if drifted {
+		stored := node.DeepCopy()
+		node.Annotations[v1alpha5.DriftedAnnotationKey] = "true"
+		if err := d.kubeClient.Patch(ctx, node, client.MergeFrom(stored)); err != nil {
+			return reconcile.Result{RequeueAfter: 60 * time.Second}, fmt.Errorf("patching node, %w", err)
```
njtran, line 66: i think since we're returning an error, we shouldn't need to put in a requeueAfter.

pkg/controllers/drift/controller.go
njtran: I think you should be able to do this within the EnqueueRequestsFromMapFunc function and just return an empty array if it's not indexed.

njtran (commented):
Nice Work! Looking good. Some comments

pkg/controllers/drift/controller.go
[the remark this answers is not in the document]
njtran: I don't see this annotation in Karpenter-core right now, this must be in your fork. I think we'll want to make this a `v1alpha5.DriftedLabelKey` if anything so that we can filter on this.

pkg/cloudprovider/cloudprovider.go
ra-grover: Based on the slack conversation, we decided to move ahead with the approach that is implemented in this PR, i.e. having a function not on the cloudprovider interface. 

It can be later added to the interface when there is more clarity on Machine CRD. 
Resolving this conversation, please feel free to open :)

pkg/cloudprovider/amifamily/resolver.go
ra-grover: `GetAMIFamily` was not a method on Resolver, I added that and see your point that there is no need for this function. 
Will fix it, thanks!

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,6 +104,34 @@ func New(ctx awscontext.Context) *CloudProvider {
 	}
 }
 
+func (c *CloudProvider) IsNodeDrifted(ctx context.Context, provisioner *v1alpha5.Provisioner, node v1.Node) (bool, error) {
+	instanceTypes, _ := c.GetInstanceTypes(ctx, provisioner)
```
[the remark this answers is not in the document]
ra-grover: Done. Thanks!

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,6 +104,34 @@ func New(ctx awscontext.Context) *CloudProvider {
...
+	aws, err := c.getProvider(ctx, provisioner.Spec.Provider, provisioner.Spec.ProviderRef)
+	if err != nil {
+		return false, err
```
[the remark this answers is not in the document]
ra-grover: Perfect, was thinking the same. Resolving.

pkg/controllers/drift/controller.go
ra-grover: Done.

pkg/controllers/drift/controller.go
ra-grover: I think, `IndexField` is a cost intensive operation and can't be done on each reconcile. I am thinking to handle it differently by only adding a `Watches(
			&source.Kind{Type: &v1alpha1.AWSNodeTemplate{}},`, if the index error is not nil by exposing a builder function. Will put that in the next commit for you to verify.

pkg/cloudprovider/amifamily/resolver.go
ra-grover: Fixed, this function is removed now.

pkg/controllers/drift/controller.go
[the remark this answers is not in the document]
ra-grover: Its there on https://github.com/aws/karpenter-core/pull/57/files

pkg/controllers/drift/controller.go
ra-grover: @ellistarn Since Nick is out, can you please check if current implementation is fine ?

pkg/controllers/drift/controller.go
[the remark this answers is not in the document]
ra-grover: @ellistarn How do we manage where the changes belong to karpenter-core also ? The drifted label is part of karpenter-core actually. Do we need to merge that in first ?

pkg/cloudprovider/launchtemplate.go
```diff
@@ -102,6 +104,40 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, provider *v1alpha1.AWS
...
+func (p *LaunchTemplateProvider) GetAmisForProvider(ctx context.Context, provider *v1alpha1.AWS, providerRef *v1alpha5.ProviderRef, instanceTypes []cloudprovider.InstanceType) ([]string, error) {
+	if provider.LaunchTemplateName != nil {
+		return nil, fmt.Errorf("using a custom Launch Template which is deprecated")
+	}
+	options, err := p.createAmiOptions(ctx, provider, map[string]string{})
+	if err != nil {
+		return nil, err
+	}
+	amifamily := amifamily.GetAMIFamily(provider.AMIFamily, options)
+	amiIds, err := p.amiFamily.AmiProvider.Get(ctx, providerRef, options, instanceTypes, amifamily)
```
ellistarn, line 136: Same re: dependency chaining
```suggestion
	amiIds, err := p.amiProvider.Get(ctx, providerRef, options, instanceTypes, amifamily)
```

ra-grover: Sorry, I can't get it to work. `p` is the launchTemplateProvider. It has amiFamily which is a ` *amifamily.Resolver` type, which inturn has an amiProvider, [here](https://github.com/aws/karpenter/blob/main/pkg/cloudprovider/amifamily/resolver.go#L47)

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,137 @@
...
+func (d Drift) Builder(ctx context.Context, m manager.Manager) operatorcontroller.Builder {
+	builder := controllerruntime.
+		NewControllerManagedBy(m).
+		Named(controllerName).
+		For(&v1.Node{}).
+		Watches(
+			&source.Kind{Type: &v1alpha5.Provisioner{}},
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) (requests []reconcile.Request) {
+				provisioner := o.(*v1alpha5.Provisioner)
+				if provisioner.Spec.ProviderRef == nil {
```
jonathan-innis, line 82: I'm wondering if we should even do this check here? It seems to overcomplicate the code and it's an edge case that can be checked in the Reconcile loop. The Provisioner admission controller should ensure that the providerRef is set for new Provisioners

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,137 @@
...
+					return requests
+				}
+				for _, node := range nodes.Items {
+					requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Name: node.Name}})
+				}
+				return requests
+			})).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10})
+
+	if err := m.GetFieldIndexer().IndexField(ctx, &v1alpha5.Provisioner{}, ".spec.providerRef.name", func(rawObj client.Object) []string {
```
jonathan-innis, line 103: We shouldn't handle this field indexing here. Since this is a generic value, we can probably move this indexing over to karpenter-core

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,137 @@
...
+				logging.FromContext(ctx).Errorf("listing provisioners for AWSNodeTemplate reconciliation %w", err)
+				return requests
+			}
+			for _, provisioner := range provisioners.Items {
+				nodes := &v1.NodeList{}
+				if err := d.kubeClient.List(ctx, nodes, client.MatchingLabels(map[string]string{v1alpha5.ProvisionerNameLabelKey: provisioner.Name})); err != nil {
+					logging.FromContext(ctx).Errorf("listing nodes when mapping drift watch events, %s", err)
+					return requests
+				}
+				for _, node := range nodes.Items {
```
jonathan-innis, line 126: There's probably ways to abstract this since I see you doing this twice in both `Enqueues...`

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,137 @@
+package drift
```
jonathan-innis, line 1: Have we load tested this at all? I have some concern how frequently we may be hitting the APIs considering how many different changes can enqueue requests on this controller.

pkg/cloudprovider/amifamily/resolver.go
```diff
@@ -44,7 +44,7 @@ var DefaultEBS = v1alpha1.BlockDevice{
 
 // Resolver is able to fill-in dynamic launch template parameters
 type Resolver struct {
-	amiProvider      *AMIProvider
+	AmiProvider      *AMIProvider
```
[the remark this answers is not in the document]
ra-grover: I had written a function in resolver.go so that we dont need to expose it and the launch template provider was calling that function. But since the args of the function were same as `r.amiProvider.Get`, as a part of this [conv](https://github.com/aws/karpenter/pull/2812#discussion_r1024415985) I removed that function and exposed this amiProvider.

pkg/cloudprovider/amifamily/ami.go
```diff
@@ -59,16 +59,16 @@ type AMI struct {
 
 // Get returns a set of AMIIDs and corresponding instance types. AMI may vary due to architecture, accelerator, etc
 // If AMI overrides are specified in the AWSNodeTemplate, then only those AMIs will be chosen.
-func (p *AMIProvider) Get(ctx context.Context, provider *v1alpha1.AWS, nodeRequest *cloudprovider.NodeRequest, options *Options, amiFamily AMIFamily) (map[string][]cloudprovider.InstanceType, error) {
+func (p *AMIProvider) Get(ctx context.Context, providerRef *v1alpha5.ProviderRef, options *Options, instanceTypes []cloudprovider.InstanceType, amiFamily AMIFamily) (map[string][]cloudprovider.InstanceType, error) {
```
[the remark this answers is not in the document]
ra-grover: No we weren't using it in the function. https://github.com/aws/karpenter/blob/main/pkg/cloudprovider/amifamily/ami.go#L62

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,6 +104,37 @@ func New(ctx awscontext.Context) *CloudProvider {
...
+	if err != nil {
+		return false, fmt.Errorf("getting provider, %w", err)
+	}
+
+	amis, err := c.instanceProvider.launchTemplateProvider.GetAmisForProvider(ctx, aws, provisioner.Spec.ProviderRef, instanceTypes)
+	if err != nil {
+		return false, fmt.Errorf("getting amis, %w", err)
+	}
+
+	nodeAmi, ok := node.Labels[v1alpha1.LabelInstanceAMIID]
```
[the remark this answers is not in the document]
ra-grover: I was returning a `false` initially, I think I had a slack conversation with @njtran to handle the case of a no ami on the node.

pkg/cloudprovider/launchtemplate.go
```diff
@@ -102,6 +104,40 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, provider *v1alpha1.AWS
...
+		ec2LaunchTemplate, err := p.ensureLaunchTemplate(ctx, resolvedLaunchTemplate)
+		if err != nil {
+			return nil, err
+		}
+		launchTemplates[*ec2LaunchTemplate.LaunchTemplateName] = resolvedLaunchTemplate.InstanceTypes
+	}
+	return launchTemplates, nil
+}
+
+func (p *LaunchTemplateProvider) GetAmisForProvider(ctx context.Context, provider *v1alpha1.AWS, providerRef *v1alpha5.ProviderRef, instanceTypes []cloudprovider.InstanceType) ([]string, error) {
```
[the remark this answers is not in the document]
ra-grover: Done, thanks.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,137 @@
...
+	if drifted, ok := node.Labels[v1alpha5.DriftedLabelKey]; ok && drifted == "true" {
+		return reconcile.Result{}, nil
+	}
+
+	provisioner := &v1alpha5.Provisioner{}
+	if err := d.kubeClient.Get(ctx, types.NamespacedName{Name: provisionerName}, provisioner); err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting provisioner, %w", err)
+	}
+
+	if drifted, err := d.cloudProvider.IsNodeDrifted(ctx, provisioner, *node); err != nil {
```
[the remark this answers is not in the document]
ra-grover: If we are using the EventFilter that you mentioned [here](https://github.com/aws/karpenter/pull/2812#discussion_r1030947663) or the currently implemented `if` check , I dont think we will be able to remove the label. Let me know if you want to check drift for a drifted node, and that case can be handled then by removing the drift label.

pkg/controllers/drift/controller.go
ra-grover: I can do that, let me know if I need to. Sorry not aware of that handling in Provisioner admission controller.

pkg/controllers/drift/controller.go
ra-grover: Can you please help me figure out the right place to put that in karpenter-core ? 
I think [here](https://github.com/aws/karpenter-core/blob/main/pkg/controllers/provisioning/provisioner.go#L108), since it's related to provisioner, but not sure of the error handling.

pkg/controllers/drift/controller.go
ra-grover: Sure, I can create a function in the controller. Does that sound good ?

pkg/cloudprovider/amifamily/ami.go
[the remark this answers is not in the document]
ra-grover: Worked in the end to end test that I did against an EKS cluster. `provider` wasn't used in this function so I removed it. Can you please elaborate if I am missing any edge case here ?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,6 +104,37 @@ func New(ctx awscontext.Context) *CloudProvider {
 	}
 }
 
+func (c *CloudProvider) IsNodeDrifted(ctx context.Context, provisioner *v1alpha5.Provisioner, node v1.Node) (bool, error) {
+	instanceTypes, err := c.GetInstanceTypes(ctx, provisioner)
```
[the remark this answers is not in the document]
ra-grover: Thanks for pointing this. 
The instanceTypes are coming from cache in the `instanceTypeProvider.Get`, there are two options:
1. I can expose a `instanceTypeProvider.GetSpecificInstanceType` and will return only that, but in that function it has to get all from the cache and then filter by the specific instance type and return only that. 
2. I don't create additional functionality in the instanceTypeProvider but filter that instanceType here in-line and pass it on to `launchTemplateProvider.GetAMIsForProvider` so we get only get only one ami in the array (but the return type would still be an array, if we dont want to refactor the existing code).

But both of these approaches, still return an array of ami with one element, otherwise we will have to create more functions to get just a single ami, according to me that may be an overkill. Happy to hear your thoughts.  :)

pkg/controllers/drift/controller.go
ra-grover: I haven't personally, but I think most of the calls in the `IsNodeDrifted` are being called from a cache or the `kubeClient.Get`. Is there anything that you see which can be a bottleneck in terms of api call ?
I would require assistance from someone in Karpenter team to load test this one.

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
jonathan-innis: My point was more that we can pass in the specific InstanceType into the `GetAmisForProvider` call so that we don't have to pass in all the instance types.

I don't think you should add a new interface method to the cloudprovider. Ideally, the behavior would be:
1. `GetInstanceTypes`
2. Find the instanceType that matches the one labeled on the node
3. `GetAmisForProvider` for that specific instance type

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,6 +104,37 @@ func New(ctx awscontext.Context) *CloudProvider {
 	}
 }
 
+func (c *CloudProvider) IsNodeDrifted(ctx context.Context, provisioner *v1alpha5.Provisioner, node v1.Node) (bool, error) {
```
jonathan-innis, line 107:
```suggestion
func (c *CloudProvider) IsNodeDrifted(ctx context.Context, provisioner *v1alpha5.Provisioner, node *v1.Node) (bool, error) {
```

Should be a pointer to the node so that we don't copy the node object around

pkg/cloudprovider/launchtemplate.go
```diff
@@ -102,6 +104,40 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, provider *v1alpha1.AWS
...
+	options, err := p.createAmiOptions(ctx, provider, map[string]string{})
+	if err != nil {
+		return nil, err
+	}
+	amifamily := amifamily.GetAMIFamily(provider.AMIFamily, options)
+	amiIds, err := p.amiFamily.AmiProvider.Get(ctx, providerRef, options, instanceTypes, amifamily)
+	return lo.Keys(amiIds), nil
+}
+
+func (p *LaunchTemplateProvider) createAmiOptions(ctx context.Context, provider *v1alpha1.AWS, additionalLabels map[string]string) (*amifamily.Options, error) {
```
jonathan-innis, line 140: Can we move this into the `resolver.go` so that we can use the standard go pattern of `NewOptions`?

pkg/cloudprovider/launchtemplate.go
jonathan-innis, line 136: Why is this `p.amiFamily` and not `amiFamily` directly? It looks like you aren't using the correct family here.

pkg/cloudprovider/launchtemplate.go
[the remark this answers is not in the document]
jonathan-innis: Got it, then can we just pass in the entire `AWSNodeTemplate` into this function since that is essentially the dependency here

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,137 @@
...
+
+const controllerName = "drift"
+
+type Drift struct {
+	kubeClient    client.Client
+	cloudProvider *cloudprovider.CloudProvider
+}
+
+// NewController constructs a controller instance
+func NewController(kubeClient client.Client, cloudProvider *cloudprovider.CloudProvider) *Drift {
```
jonathan-innis, line 32: There's the introduction of the `controller.Typed` into the `karpenter-core` project which means that you can create this controller as a `controller.Typed[*v1.Node]()` with automatic getting and patching.

Here's an example since we're trying to migrate all controllers over to the framework if we can:

https://github.com/aws/karpenter-core/blob/main/pkg/controllers/node/controller.go#L58

jonathan-innis (commented):
Great work overall! Super excited to see this feature go in!

pkg/cloudprovider/launchtemplate.go
ra-grover: Ohh, sorry for the confusion. But `p.amiFamily` is actually `*amifamily.Resolver` type. Would you mind if I rename `p.amiFamily` to `p.amiResolver` ?

pkg/cloudprovider/launchtemplate.go
ra-grover: This is using three functions from launchTemplateProvider, if we move then we will have to pass `LaunchTemplateProvider` in the `NewOptions` func, is that the right direction to go ?

pkg/cloudprovider/cloudprovider.go
ra-grover: Done, thanks. Will push it.

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
ra-grover: Done.

pkg/controllers/drift/controller.go
ra-grover: Done.

pkg/controllers/drift/controller.go
ra-grover: Done. Created a function `getReconcileRequest` which is being used in both `Enqueues...`

pkg/controllers/drift/controller.go
ra-grover: @njtran You might want to check this when you come back. I think @jonathan-innis mentioned in some other comment that we move this indexField to karpenter-core, but I am not sure where.

pkg/controllers/drift/controller.go
ellistarn: If there are perf concerns, let's set `MaxConcurrentReconciles: 1`

pkg/controllers/drift/controller.go
ra-grover: Done.

pkg/cloudprovider/launchtemplate.go
ra-grover: @ellistarn If you can chime in here, that will be great :) I wasn't able to get the suggestion to work.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -82,15 +86,18 @@ func New(ctx awscontext.Context) *CloudProvider {
 	}
 	subnetProvider := NewSubnetProvider(ec2api)
 	instanceTypeProvider := NewInstanceTypeProvider(ctx, ctx.Session, ec2api, subnetProvider, ctx.UnavailableOfferingsCache, ctx.StartAsync)
+	amiProvider := amifamily.NewAmiProvider(ctx.KubeClient, ssm.New(ctx.Session), ec2api, cache.New(awscontext.CacheTTL, awscontext.CacheCleanupInterval), cache.New(awscontext.CacheTTL, awscontext.CacheCleanupInterval), ctx.KubernetesInterface)
```
ra-grover, line 89: Hey @jonathan-innis , I have refactored amiProvider out and it now has `GetAMIsForProvider`.

ra-grover: Moved `KubeServerVersion(ctx)` to amiProvider too, I think it belongs there only. As only for ami we need that for now.

pkg/cloudprovider/launchtemplate.go
```diff
@@ -102,6 +104,40 @@ func (p *LaunchTemplateProvider) Get(ctx context.Context, provider *v1alpha1.AWS
...
+func (p *LaunchTemplateProvider) GetAMIsForProvider(ctx context.Context, provider *v1alpha1.AWS, providerRef *v1alpha5.ProviderRef, instanceTypes []*cloudprovider.InstanceType) ([]string, error) {
+	if provider.LaunchTemplateName != nil {
+		return nil, fmt.Errorf("using a custom Launch Template which is deprecated")
+	}
+	options, err := p.createAmiOptions(ctx, provider, map[string]string{})
+	if err != nil {
+		return nil, err
+	}
+	amifamily := amifamily.GetAMIFamily(provider.AMIFamily, options)
+	amiIds, err := p.amiFamily.AmiProvider.Get(ctx, providerRef, options, instanceTypes, amifamily)
```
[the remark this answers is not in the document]
ra-grover: I moved this to amiProvider, and its checked there. 
https://github.com/aws/karpenter/pull/2812/files#diff-9ef5e430b50c50c45152aa4ffbb7bbdb55ee8d587d68235dc55f1e169f4c837bR106

pkg/cloudprovider/cloudprovider.go
```diff
@@ -67,6 +69,7 @@ type CloudProvider struct {
 	instanceTypeProvider *InstanceTypeProvider
 	instanceProvider     *InstanceProvider
 	kubeClient           k8sClient.Client
+	ec2Api               *ec2.EC2
```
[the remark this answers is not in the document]
ra-grover: Done, thanks.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,134 @@
...
+
+var _ operatorcontroller.TypedController[*v1.Node] = (*Drift)(nil)
+
+type Drift struct {
+	kubeClient    client.Client
+	cloudProvider *cloudprovider.CloudProvider
+}
+
+// NewController constructs a controller instance
+func NewController(kubeClient client.Client, cloudProvider *cloudprovider.CloudProvider) operatorcontroller.Controller {
```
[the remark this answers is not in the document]
ra-grover: Done, thanks.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+			for _, provisioner := range provisioners.Items {
+				return getReconcileRequest(ctx, &provisioner, d.kubeClient)
+			}
+			return requests
+		}),
+	))
+}
+
+func getReconcileRequest(ctx context.Context, provisioner *v1alpha5.Provisioner, kubeClient client.Client) (requests []reconcile.Request){
```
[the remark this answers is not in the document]
ra-grover: Done, thanks.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,134 @@
...
+
+	provisioner := &v1alpha5.Provisioner{}
+	if err := d.kubeClient.Get(ctx, types.NamespacedName{Name: provisionerName}, provisioner); err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting provisioner, %w", err)
+	}
+
+	if drifted, err := d.cloudProvider.IsNodeDrifted(ctx, provisioner, node); err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting drift for node, %w", err)
+	} else if drifted {
+		stored := node.DeepCopy()
```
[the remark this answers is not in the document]
ra-grover: Done, thanks.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,134 @@
...
+		cloudProvider: cloudProvider,
+	})
+}
+
+func (c *Drift) Name() string {
+	return "drift"
+}
+
+func (d Drift) Reconcile(ctx context.Context, node *v1.Node) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named(d.Name()).With("node", node.Name))
```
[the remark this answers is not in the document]
ra-grover: Done, thanks.

pkg/cloudprovider/suite_test.go
```diff
@@ -255,6 +258,56 @@ var _ = Describe("Allocation", func() {
 			Expect(createFleetInput.Context).To(BeNil())
 		})
 	})
+	Context("Node Drift", func() {
+		It("should detect drift if ami gets changed", func() {
```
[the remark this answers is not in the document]
njtran: Additionally, we should add tests for this watch mechanism in controllers/drift/suite_test.go to test this individually.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -100,6 +106,44 @@ func New(ctx awscontext.Context) *CloudProvider {
...
+	if nodeTemplate.Spec.LaunchTemplateName != nil {
+		return false, fmt.Errorf("using a custom Launch Template which is deprecated")
+	}
+	amis, err := c.amiProvider.Get(ctx, nodeTemplate, nodeInstanceType, amifamily.GetAMIFamily(nodeTemplate.Spec.AMIFamily, &amifamily.Options{}))
+	if err != nil {
+		return false, fmt.Errorf("getting amis, %w", err)
+	}
+	nodeAmi, ok := node.Labels[v1alpha1.LabelInstanceAMIID]
+	if !ok {
+		instanceID, err := utils.ParseInstanceID(node)
```
njtran, line 133: Since we're doing the fallback, can you add a suite test for when the node does not have the AMI ID?

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,131 @@
...
+
+	provisioner := &v1alpha5.Provisioner{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: provisionerName}, provisioner); err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting provisioner, %w", err)
+	}
+
+	if drifted, err := c.cloudProvider.IsNodeDrifted(ctx, provisioner, node); err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting drift for node, %w", err)
+	} else if drifted {
+		node.Labels["drifted"] = "true"
```
njtran, line 78: Change this to the annotation as well!

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,131 @@
...
+		Watches(
+			&source.Kind{Type: &v1alpha5.Provisioner{}},
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) (requests []reconcile.Request) {
+				provisioner := o.(*v1alpha5.Provisioner)
+				// Ensure provisioner has a defined AWSNodeTemplate
+				nodeTemplate := &v1alpha1.AWSNodeTemplate{}
+				if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: provisioner.Spec.ProviderRef.Name}, nodeTemplate); err != nil {
+					logging.FromContext(ctx).Errorf("getting AWSNodeTemplates when mapping drift watch events, %s", err)
+					return requests
+				}
```
njtran, line 97: We're also checking if the node template exists in the reconcile method, so we can just remove this in favor of the reconcile method since we bubble up the error differently there.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (c *Controller) Reconcile(ctx context.Context, node *v1.Node) (reconcile.Result, error) {
+	provisionerName, provisionerExists := node.Labels[v1alpha5.ProvisionerNameLabelKey]
+	if !provisionerExists {
+		return reconcile.Result{}, nil
+	}
+
+	// TODO: change this label to the karpenter-core drift label
+	if drifted, ok := node.Labels["drifted"]; ok && drifted == "true" {
```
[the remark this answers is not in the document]
ra-grover: I have to import the key/value for the annotation from karpenter-core , right ?

pkg/cloudprovider/cloudprovider.go
ra-grover: Done, but not sure if my implementation is 100% right. You might have to look.

cmd/controller/main.go
```diff
@@ -45,6 +47,15 @@ func main() {
 	lo.Must0(operator.AddHealthzCheck("cloud-provider", awsCloudProvider.LivenessProbe))
 	cloudProvider := metrics.Decorate(awsCloudProvider)
 
+	// Setup field indexers for AWS controllers to save computation for looking up providerRef
+	lo.Must0(operator.Manager.GetFieldIndexer().IndexField(ctx, &v1alpha5.Provisioner{}, ".spec.providerRef.name", func(o client.Object) []string {
```
ellistarn, line 51: I don't think name is enough
```
type ProviderRef struct {
	// Kind of the referent; More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds"
	Kind string `json:"kind,omitempty"`
	// Name of the referent; More info: http://kubernetes.io/docs/user-guide/identifiers#names
	Name string `json:"name,omitempty"`
	// API version of the referent
	// +optional
	APIVersion string `json:"apiVersion,omitempty"`
}
```

pkg/cloudprovider/cloudprovider.go
```diff
@@ -165,6 +169,50 @@ func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
 	return c.instanceProvider.Terminate(ctx, node)
 }
 
+func (c *CloudProvider) IsMachineDrifted(ctx context.Context, machine *corev1alpha1.Machine) (bool, error) {
+	// Not needed when GetInstanceTypes removes provisioner dependency
+	provisioner := &v1alpha5.Provisioner{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: machine.Labels[v1alpha5.ProvisionerNameLabelKey]}, provisioner); err != nil {
+		return false, fmt.Errorf("getting provisioner, %w", err)
```
ellistarn, line 176: Do we need an IgnoreNotFound here?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -165,6 +169,50 @@ func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
...
+	}
+	instanceTypes, err := c.GetInstanceTypes(ctx, provisioner)
+	if err != nil {
+		return false, fmt.Errorf("getting instanceTypes, %w", err)
+	}
+	nodeInstanceType := lo.Filter(instanceTypes, func(instType *cloudprovider.InstanceType, _ int) bool {
```
ellistarn, line 182: lo.Find is what you're looking for.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -165,6 +169,50 @@ func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
...
+		return instType.Name == machine.Labels[v1.LabelInstanceTypeStable]
+	})
+	if len(nodeInstanceType) == 0 {
+		return false, fmt.Errorf("getting nodeInstanceType")
+	}
+	nodeTemplate, err := c.resolveNodeTemplate(ctx, []byte(machine.
```
ellistarn, line 188: thoughts on:
```suggestion
	nodeTemplate, err := c.resolveNodeTemplate(ctx, machine)
```

pkg/cloudprovider/cloudprovider.go
```diff
@@ -165,6 +169,50 @@ func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
...
+		Annotations[v1alpha5.ProviderCompatabilityAnnotationKey]), machine.
+		Spec.MachineTemplateRef)
+	if err != nil {
+		return false, fmt.Errorf("resolving node template, %w", err)
+	}
+	if nodeTemplate.Spec.LaunchTemplateName != nil {
+		return false, fmt.Errorf("using a custom Launch Template which is deprecated")
```
ellistarn, line 195: Should this return an error, or just false,nil? 
```suggestion
		return false, fmt.Errorf("using a custom Launch Template which is deprecated")
```

pkg/cloudprovider/cloudprovider.go
```diff
@@ -165,6 +169,50 @@ func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
...
+	}
+	amis, err := c.amiProvider.Get(ctx, nodeTemplate, nodeInstanceType, amifamily.GetAMIFamily(nodeTemplate.Spec.AMIFamily, &amifamily.Options{}))
+	if err != nil {
+		return false, fmt.Errorf("getting amis, %w", err)
+	}
+	nodeAmi, ok := machine.Labels[v1alpha1.LabelInstanceAMIID]
```
ellistarn, line 201: What if we only supported this for machines that had this label?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -165,6 +169,50 @@ func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
 	return c.instanceProvider.Terminate(ctx, node)
 }
 
+func (c *CloudProvider) IsMachineDrifted(ctx context.Context, machine *corev1alpha1.Machine) (bool, error) {
```
ellistarn, line 172: This logic is going to grow as we add in things like instanceProfile, securityGroup, etc. How do you see this evolving?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -165,6 +169,50 @@ func (c *CloudProvider) Delete(ctx context.Context, node *v1.Node) error {
...
+		return instType.Name == machine.Labels[v1.LabelInstanceTypeStable]
+	})
+	if len(nodeInstanceType) == 0 {
+		return false, fmt.Errorf("getting nodeInstanceType")
+	}
+	nodeTemplate, err := c.resolveNodeTemplate(ctx, []byte(machine.
+		Annotations[v1alpha5.ProviderCompatabilityAnnotationKey]), machine.
+		Spec.MachineTemplateRef)
+	if err != nil {
+		return false, fmt.Errorf("resolving node template, %w", err)
```
ellistarn, line 192: What if we just returned false if no node template?

pkg/cloudprovider/cloudprovider.go
ellistarn, line 201: I'd love to just have one mechanism for detecting this. Which do you think is best if you could only choose one?

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,141 @@
...
+					return requests
+				}
+				return getReconcileRequests(ctx, provisioner, c.kubeClient)
+			})).
+		Watches(
+			&source.Kind{Type: &v1alpha1.AWSNodeTemplate{}},
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) (requests []reconcile.Request) {
+				provisioners := &v1alpha5.ProvisionerList{}
+				if err := c.kubeClient.List(ctx, provisioners, client.MatchingFields{".spec.providerRef.name": o.GetName()}); err != nil {
+					logging.FromContext(ctx).Errorf("listing provisioners for AWSNodeTemplate reconciliation %w", err)
```
ellistarn, line 115: Consider collapsing error handling

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,141 @@
...
+	return "drift"
+}
+
+func (c *Controller) Reconcile(ctx context.Context, node *v1.Node) (reconcile.Result, error) {
+	provisionerName, provisionerExists := node.Labels[v1alpha5.ProvisionerNameLabelKey]
+	if !provisionerExists {
+		return reconcile.Result{}, nil
+	}
+
+	if drifted, ok := node.Annotations[v1alpha5.VoluntaryDisruptionAnnotationKey]; ok && drifted == v1alpha5.VoluntaryDisruptionDriftedAnnotationValue {
```
ellistarn, line 69:
```suggestion
	if node.Annotations[v1alpha5.VoluntaryDisruptionAnnotationKey] == v1alpha5.VoluntaryDisruptionDriftedAnnotationValue {
```

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,141 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	provisioner := &v1alpha5.Provisioner{}
+	err := c.kubeClient.Get(ctx, types.NamespacedName{Name: provisionerName}, provisioner)
+	if errors.IsNotFound(err) {
+		return reconcile.Result{}, nil
+	}
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting provisioner, %w", err)
```
ellistarn, line 79: client.IgnoreNotFound(err)

cmd/controller/main.go
njtran: Cut an issue for this as discussed: https://github.com/aws/karpenter/issues/3023

pkg/cloudprovider/cloudprovider.go
njtran: Putting notes from our discussion here: 
- Moving most of the AMI logic to a separate function. 
- Maybe we can move some of the additional drift functions as a sort of requirements so that we can easily cache resolved values of the AWSNodeTemplate settings as some sort of easily comparable values.

pkg/cloudprovider/cloudprovider.go
njtran: Done

pkg/cloudprovider/cloudprovider.go
njtran: returning false, nil

pkg/cloudprovider/cloudprovider.go
njtran: Returning false if IsNotFound.

pkg/cloudprovider/cloudprovider.go
njtran: As discussed offline, just doing an EC2 call here instead of the AMI label.

cmd/controller/main.go
```diff
@@ -45,6 +47,15 @@ func main() {
...
+		provisioner := o.(*v1alpha5.Provisioner)
+		if provisioner.Spec.ProviderRef == nil {
+			return []string{""}
```
ellistarn, line 54: Should this be 
```suggestion
			return nil
```

pkg/cloudprovider/cloudprovider.go
ellistarn: This usage of IgnoreNotFound introduces a bug. If the provisioner is not found we'll continue as if nothing happened. 
```suggestion
		return false, k8sClient.IgnoreNotFound(err)
```

Thinking further on this, I don't know if we want to error, since this is likely unrecoverable. Perhaps we should just print an error log and return false (not drifted).

pkg/cloudprovider/cloudprovider.go
ellistarn:
```suggestion
		return false, k8sClient.IgnoreNotFound(err)
```

pkg/cloudprovider/cloudprovider.go
```diff
@@ -201,6 +226,37 @@ func kubeDNSIP(ctx context.Context, kubernetesInterface kubernetes.Interface) (n
...
+	// Get InstanceID to fetch from EC2
+	instanceID, err := utils.ParseMachineInstanceID(machine)
+	if err != nil {
+		return false, err
+	}
+	instance, err := c.instanceProvider.getInstance(ctx, *instanceID)
+	if err != nil {
+		return false, fmt.Errorf("getting instance, %w", err)
+	}
+	return !lo.Contains(lo.Keys(amis), *instance.ImageId), nil
```
bwagner5, line 257: wouldn't this say there's no drift if a new version of the AMI is released but is still included in the return of `amiProvider.Get` ? 

i.e.

```
amis, _ := amiProvider.Get(ctx,  nodeTemplate, []*cloudprovider.InstanceType{nodeInstanceType}, amiFamily...) 
-> ["ami-123"]

... time passes ...

amis, _ := amiProvider.Get(ctx,  nodeTemplate, []*cloudprovider.InstanceType{nodeInstanceType}, amiFamily...) 
-> ["ami-234", "ami-123"]
```

In the case above, "ami-123" was first published and then "ami-234" was published, so the amiProvider sorted by most recent, but the ami would be marked as not being drited since "ami-123" is in the list still. 

Should you just check the first element in the amiID's return?

bwagner5 (commented):
very nice work! A few comments

pkg/cloudprovider/cloudprovider.go
```diff
@@ -201,6 +226,37 @@ func kubeDNSIP(ctx context.Context, kubernetesInterface kubernetes.Interface) (n
...
+	if nodeTemplate.Spec.LaunchTemplateName != nil {
+		return false, nil
+	}
+	amis, err := c.amiProvider.Get(ctx, nodeTemplate, []*cloudprovider.InstanceType{nodeInstanceType},
+		amifamily.GetAMIFamily(nodeTemplate.Spec.AMIFamily, &amifamily.Options{}))
+	if err != nil {
+		return false, fmt.Errorf("getting amis, %w", err)
+	}
+	// Get InstanceID to fetch from EC2
+	instanceID, err := utils.ParseMachineInstanceID(machine)
```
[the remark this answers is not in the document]
njtran: I agree, I copied this over previously. Changed it.

ellistarn (approved):
Amazing work all around from all contributors.

bwagner5 (approved):
LGTM

PR #2828, opened 2022-11-09
test: Add test for adding global tags to interruption infra

jonathan-innis:
Fixes # 

**Description**

- Adds test for checking that global tags exist on interruption infra

* `FOCUS=Infrastructure make e2etests`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/test/pkg/environment/aws/environment.go
+++ b/test/pkg/environment/aws/environment.go
@@ -21,10 +21,10 @@ import (
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/eventbridge"
 	"github.com/aws/aws-sdk-go/service/iam"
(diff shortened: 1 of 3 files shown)


test/pkg/environment/aws/environment.go
```diff
@@ -35,28 +36,36 @@ type Environment struct {
...
 
-	return &Environment{
+	ret := &Environment{
 		Region:          *session.Config.Region,
 		Environment:     env,
 		EC2API:          *ec2.New(session),
 		SSMAPI:          *ssm.New(session),
 		IAMAPI:          *iam.New(session),
+		SQSAPI:          sqs.New(session),
+		EventBridgeAPI:  eventbridge.New(session),
```
bwagner5, line 62: nit: is there a reason these are pointers? I don't think it really matters which way we go, but I think they should all be values or all pointers :D

jonathan-innis: They should probably all be pointers since that's what the New function produces

bwagner5 (approved):
LGTM

PR #2839, opened 2022-11-11
docs: Update provisioner configuration

mbevc1:
Match existing `securityGroupSelector` and avoid confusion.

Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/website/content/en/preview/getting-started/getting-started-with-eksctl/scripts/step12-add-provisioner.sh
+++ b/website/content/en/preview/getting-started/getting-started-with-eksctl/scripts/step12-add-provisioner.sh
@@ -23,5 +23,5 @@ spec:
   subnetSelector:
     karpenter.sh/discovery: ${CLUSTER_NAME}
   securityGroupSelector:
-    karpenter.sh/discovery: ${CLUSTER_NAME}
+    kubernetes.io/cluster/${CLUSTER_NAME}: '*'
 EOF


jonathan-innis:
@mbevc1 Can you explain the confusion that you feel like is being caused by the `karpenter.sh/discovery` tag being used for security group discovery?

mbevc1:
I was using `eksctl` last week for some demons and it tripped me following the Getting started guides as default behaviour wouldn't set those tags on nodes SGs. I took me some time to backtrack why my LoadBalances are not provisioning to Karpenter nodes. If we use the suggested tags this works out of the box for the examples provided in the docs.

FernandoMiguel:
you do realise this is a breaking change for everyone else who already has this deployed, correct?

mbevc1:
Not sure if I'm missing some context here @FernandoMiguel or this being used anywhere else, but since this is part of getting started section, wouldn't this only affect the docs? :thinking:

FernandoMiguel:
@mbevc1 correct. but folks followed those instructions to setup their clusters

mbevc1:
I'm probably being daft here, but I still fail to see why that would break anything. Could you please elaborate why this would breaks things for whoever followed those instructions? I'd assume getting started is used for initial setups and you wouldn't come back to this config afterwards?

FernandoMiguel:
@mbevc1 during upgrades folks come back to see if this has changed and now it drifts from their remaining infra

bwagner5:
> Since it's a non-working default setup I'd be inclined to add an upgrade warning to this change 🤔 How does that sound?

I'm not sure it's a non-working default. Does using the kubernetes.io/cluster tag result in more resources being selected than just the sg? 

I like the explicit karpenter tag for selection (at least as a default). Maybe going another direction, could just add a note about tagging the SG if using the AWS Load Balancer Controller? I think the note would demystify the experience anyways. If we just change the tag, I still wouldn't really know what it's selecting and could easily change that tag to something else and then break my setup.

mbevc1:
Seems using default config with `eksctl` results in SG without any usable tags for LB to select (only suggested one is present). Provisioned nodes also don't have the `karpenter.sh/discovery` tag present. Am I missing anything obvious here? :thinking:

tzneal:
We're going to close this for now, the current method seems to be working and we want to cause as little churn for current Karpenter users as we can.

mbevc1:
I was distracted with some other bits. I might test some more in the following months and provide more feedback.

PR #2858, opened 2022-11-16
fix: helm chart webhook sort order

smcavallo:
Signed-off-by: smcavallo <smcavallo@hotmail.com>

Fixes # 

**Description**

Small change to update the webhook rules to be sorted alphabetically.
The sorting was changed as part of this commit ->  https://github.com/aws/karpenter/commit/b358e0040153183d2df720e39b9530f50251889f
The api group names are similar enough to confuse the way ArgoCD detects differences, which means these webhooks are constantly appearing as OutOfSync.  I am assuming this is also causing issues with Flux and other tools.
Either way, reverting the order so they are now properly sorted alphabetically.

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/charts/karpenter/templates/webhooks.yaml
+++ b/charts/karpenter/templates/webhooks.yaml
@@ -19,16 +19,6 @@ webhooks:
     sideEffects: None
     rules:
       - apiGroups:
-          - karpenter.sh
-        apiVersions:
-          - v1alpha5
-        resources:
-          - provisioners
-          - provisioners/status
-        operations:
-          - CREATE
-          - UPDATE
-      - apiGroups:
           - karpenter.k8s.aws
         apiVersions:
           - v1alpha1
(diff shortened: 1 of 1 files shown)


bwagner5:
I can see how the ordering would make them out-of-sync, but the PR referenced changes other things that would need to be updated like the name. I don't understand how after the webhook was updated, it still says they are out of sync?

smcavallo:
@bwagner5 - when the webhooks are deployed to kubernetes, kubernetes sorts them alphabetically ascending.  They are sorted in descending order in the helm chart.   Normally this would not be an issue, but it an issue with however these deployment tools are doing diffs.  The names are close enough where their diff tools are getting confused.
Normally I would say this is an issue with the deployment tool diffs which should get fixed.  Except it doesn't hurt to sort them alphabetically in source - it is such a simple change.  Especially when the older webhooks used to be sorted alphabetically.
 
image

ellistarn (approved):

bwagner5 (approved):
LGTM

PR #2884, opened 2022-11-18
test: Added New test to check Resource name DNS A record

engedaam:
Fixes # 

**Description**
Added an integration test to make sure that only one node with the correct node name would be created when a subnet has Resource name DNS A record Enabled

Ran the test using `FOCUS="should use the subnet-id selector with resource based naming" make e2etests`
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


test/suites/integration/subnet_test.go
```diff
@@ -54,6 +54,29 @@ var _ = Describe("Subnets", func() {
...
+		})
+		provisioner := test.Provisioner(test.ProvisionerOptions{ProviderRef: &v1alpha5.ProviderRef{Name: provider.Name}})
+		pod := test.Pod()
+
+		env.ExpectCreated(pod, provider, provisioner)
+		env.EventuallyExpectHealthy(pod)
+		env.ExpectCreatedNodeCount("==", 1)
+
+		env.ExpectInstance(pod.Spec.NodeName).To(HaveField("SubnetId", HaveValue(Equal(firstSubnet))))
+		disabledResourceNameDNSArecord(firstSubnet)
```
jonathan-innis, line 78: This reset should be in a `DeferCleanup` statement so that it is always executed when the test cleans up. Otherwise, if this test fails, L78 will never be executed.

engedaam: Okay, makes sense

test/suites/integration/subnet_test.go
```diff
@@ -54,6 +54,32 @@ var _ = Describe("Subnets", func() {
 
 		env.ExpectInstance(pod.Spec.NodeName).To(HaveField("SubnetId", HaveValue(Equal(firstSubnet))))
 	})
+	It("should use the subnet-id selector with resource based naming", func() {
+		subnets := getSubnets(map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName})
+		Expect(len(subnets)).ToNot(Equal(0))
+		shuffledAZs := lo.Shuffle(lo.Keys(subnets))
+		firstSubnet := subnets[shuffledAZs[0]][0]
+
+		enableResourceNameDNSArecord(firstSubnet)
```
bwagner5, line 63: instead of picking one subnet, can we configure all the subnets with the selector to use resource based names? This will help to prevent flakey tests when there are operational issues occurring in one AZ. 

You can change the func signature of `enableResourceNameDNSArecord` to a variadic: 

```
func enableResourceNameDNSArecord(subnetIDs ...string) {
```

test/suites/integration/subnet_test.go
```diff
@@ -54,6 +54,32 @@ var _ = Describe("Subnets", func() {
...
+				SecurityGroupSelector: map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName},
+				SubnetSelector:        map[string]string{"aws-ids": firstSubnet},
+			},
+		})
+		provisioner := test.Provisioner(test.ProvisionerOptions{ProviderRef: &v1alpha5.ProviderRef{Name: provider.Name}})
+		pod := test.Pod()
+
+		env.ExpectCreated(pod, provider, provisioner)
+		env.EventuallyExpectHealthy(pod)
+		env.ExpectCreatedNodeCount("==", 1)
```
bwagner5, line 79: Can you also do an Expect for the node name to include the instance ID ?

bwagner5 (commented):
nice! just a few suggestions to make this even better

test/suites/integration/subnet_test.go
```diff
@@ -54,6 +54,36 @@ var _ = Describe("Subnets", func() {
 
 		env.ExpectInstance(pod.Spec.NodeName).To(HaveField("SubnetId", HaveValue(Equal(firstSubnet))))
 	})
+	It("should use the subnet-id selector with resource based naming", func() {
```
bwagner5, line 57:
```suggestion
	It("should use resource based naming as node names", func() {
```

test/suites/integration/subnet_test.go
```diff
@@ -54,6 +54,36 @@ var _ = Describe("Subnets", func() {
...
+		subnets := getSubnets(map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName})
+		Expect(len(subnets)).ToNot(Equal(0))
+		availabilityZone := lo.Keys(subnets)
```
bwagner5, line 60: I think instead of this and the looping below, you could do:

```
allSubnets := lo.Flatten(lo.Values(subnets))
```

which should give you a slice of subnet ID: `[]string{"subnet-id1", "subnet-id2"} 

This is from the lo lib w/ docs here: https://github.com/samber/lo#values

test/pkg/environment/aws/expectations.go
```diff
@@ -31,6 +31,11 @@ func (env *Environment) ExpectInstance(nodeName string) Assertion {
 	return Expect(env.GetInstance(nodeName))
 }
 
+func (env *Environment) ExceptNodeNameToContainInstanceID(nodeName string) {
```
bwagner5, line 34: since we're not using this func anywhere else, can you inline the two lines in this func in the actual test?  We can always move it to its own function later if we add more tests that use this Expectation.

test/pkg/environment/common/expectations.go
```diff
@@ -193,10 +193,6 @@ func (env *Environment) EventuallyExpectNotFoundAssertionWithOffset(offset int,
 	})
 }
 
-func (env *Environment) ExpectDeploymentCreatedAndHealthy(numPods int) {
```
bwagner5, line 196: Nice!

test/suites/integration/subnet_test.go
```diff
@@ -54,6 +54,36 @@ var _ = Describe("Subnets", func() {
...
+
+		var allSubnets []string
+		for zoneIndex := 0; zoneIndex < len(availabilityZone); zoneIndex++ {
+			allSubnets = append(allSubnets, subnets[availabilityZone[zoneIndex]]...)
+		}
```
bwagner5, line 65: I think we can remove this based on the comment above, but in the future, you can use a `range` which is a lot more common in go for-loops. 

```
for _, zone := range availabilityZone {
    allSubnets = append(allSubnets, subnets[zone]...)
} 
```

or even better, `range` supports maps as well:

```
for zone, subnetIDs := range subnets {
    allSubnets = append(allSubnets, subnetIDs...)
}
```

Here's some more examples: https://gobyexample.com/range

test/suites/integration/subnet_test.go
```diff
@@ -108,3 +138,43 @@ func getSubnets(tags map[string]string) map[string][]string {
...
+		inputOne := &ec2.ModifySubnetAttributeInput{
+			EnableResourceNameDnsARecordOnLaunch: &ec2.AttributeBooleanValue{
+				Value: aws.Bool(false),
+			},
+			SubnetId: aws.String(subnetIDs[subnetID]),
+		}
+		inputTwo := &ec2.ModifySubnetAttributeInput{
+			PrivateDnsHostnameTypeOnLaunch: aws.String("ip-name"),
+			SubnetId:                       aws.String(subnetIDs[subnetID]),
+		}
```
bwagner5, line 174: rather than setting these as variables, can you inline these in the ModifySubnetAttribute arguments, like this?

```
		_, err := env.EC2API.ModifySubnetAttribute(&ec2.ModifySubnetAttributeInput{
			EnableResourceNameDnsARecordOnLaunch: &ec2.AttributeBooleanValue{
				Value: aws.Bool(false),
			},
			SubnetId: aws.String(subnetIDs[subnetID]),
		})
		Expect(err).To(BeNil())
``` 

and same for the enable func above.

bwagner5 (changes requested):
Great work on this! Just a few comments on some simplifications and style suggestions! :)

bwagner5:
Let's hold on merging until we fix the resource based naming properly: https://github.com/aws/karpenter/issues/2892

This test is still valid, but we'll need to modify it a bit to expect only the instance and to enable the external cloud-provider.

test/suites/integration/subnet_test.go
```diff
@@ -108,3 +134,43 @@ func getSubnets(tags map[string]string) map[string][]string {
...
+		Expect(err).To(BeNil())
+		_, err = env.EC2API.ModifySubnetAttribute(&ec2.ModifySubnetAttributeInput{
+			PrivateDnsHostnameTypeOnLaunch: aws.String("resource-name"),
+			SubnetId:                       aws.String(subnetIDs[subnetID]),
+		})
+		Expect(err).To(BeNil())
+	}
+}
+
+func disableResourceBasedNaming(subnetIDs ...string) {
```
jonathan-innis, line 155: Can these be formatted as `ExpectResourceBasedNamingDisabled`? Generally, we are formatting things that have expect statements inside of them to start with `Expect...` to have visibility into the fact there is an Expect statement there at all

engedaam: Updated!

bwagner5 (approved):
LGTM!

PR #2754, opened 2022-10-28
chore: Separated webhooks from karpenter-core

ellistarn:
Fixes # 

**Description**

**How was this change tested?**

* make apply && TEST_FILTER=TestIntegration make e2etests

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2891, opened 2022-11-19
fix: add missing DELETE method for validating webhook

skdltmxn:
Fixes # 

**Description**

Minor fix for `validation.webhook.karpenter.k8s.aws` webhook.
It seems like DELETE method for karpenter.k8s.aws group was removed in #2754.
I see no reason for the deletion, and think it was a mistake.

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/charts/karpenter/templates/webhooks.yaml
+++ b/charts/karpenter/templates/webhooks.yaml
@@ -63,6 +63,7 @@ webhooks:
         operations:
           - CREATE
           - UPDATE
+          - DELETE
         resources:
           - awsnodetemplates
           - awsnodetemplates/status


ellistarn:
I don't see the usecase for validating delete requests. Can't remember, but probably why I removed it. Does this break something?

skdltmxn:
> I don't see the usecase for validating delete requests. Can't remember, but probably why I removed it. Does this break something?

I am currently running two EKS clusters with karpenter.
I used helm to install karpenter, but directly installed it using helm cli on one cluster and argocd installed on the another.

On both clusters, deployed webhooks have DELETE method unlike the template. It is fine on cluster where I used helm cli, but on cluster with argocd it seems argocd is constantly re-syncing the webhook. This makes karpenter pods emit reconcile errors every few second and no node is spawned.

I still don't understand why deployed webhooks get DELETE method back again, but so far I believe this PR could fix current problem.

ellistarn:
I think this is due to knative's webhook reconciliation behavior. Fix SGTM

ellistarn (approved):

PR #2910, opened 2022-11-22
Do no fail when EC2 credentials are missing

grosser:
fix: karpenter crashing with kube2iam

**Description**

issue was introduced in https://github.com/aws/karpenter/pull/1605
with the good intention of making broken ec2 api fail fast

we run kube2iam to get ec2 permissions for our pods, but that seems to count as "Checking EC2 API connectivity, NoCredentialProviders: no valid providers in chain. Deprecated.\n\tFor verbose messaging see aws.Config.CredentialsChainVerboseErrors"

TODO: inspect the actual error type or message an then decide if it's error or fatal

deploy to cluster, see if it crashes

- [X] No

**Release Note**

```release-note
Fix: do not fail when using ec2 api without credential chain
```


grosser:
permissions were just missing 🤦

PR #2926, opened 2022-11-23
feat: instanceProfileSelector-enhancement

smcavallo:
Signed-off-by: smcavallo <smcavallo@hotmail.com>

Fixes # 

**Description**

Ability to configure `role` instead of `instanceProfile`
*  `instanceProfile` are difficult to look up in AWS
*  easier to configure as `role` name is more well known
* this is more aligned with how nodeGroups are set up - you provide a role name
* more resilient - if you delete a role and recreate it with the same name this code will auto-detect the new instance profile
   *  this is much safer as otherwise the nodes will not be able to join the cluster with the wrong instance profile

Role is not "guaranteed" to work since....per the docs - https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html
_An instance profile can contain only one IAM role, although a role can be included in multiple instance profiles._

**Aside from the above, role is safer and more convenient as it protects against the underlying instance profile of a role changing.  Using a role is much easier to wire up and maintain considering the role is not created or maintained by karpenter.**

* Unit tests

- [X ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/apis/crds/karpenter.k8s.aws_awsnodetemplates.yaml
+++ b/pkg/apis/crds/karpenter.k8s.aws_awsnodetemplates.yaml
@@ -189,6 +189,10 @@ spec:
                       credentials are not available."
                     type: string
                 type: object
+              role:
+                description: Role is the AWS Role associated with the InstanceProfile
+                  which should be used
+                type: string
               securityGroupSelector:
                 additionalProperties:
                   type: string
--- a/pkg/apis/v1alpha1/provider.go
+++ b/pkg/apis/v1alpha1/provider.go
@@ -40,6 +40,9 @@ type AWS struct {
 	// InstanceProfile is the AWS identity that instances use.
 	// +optional
 	InstanceProfile *string `json:"instanceProfile,omitempty"`
+	// Role is the AWS Role associated with the InstanceProfile which should be used
+	// +optional
+	Role *string `json:"role,omitempty"`
 	// SubnetSelector discovers subnets by tags. A value of "" is a wildcard.
 	// +optional
 	SubnetSelector map[string]string `json:"subnetSelector,omitempty"`
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -25,6 +25,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/ssm"
 	"github.com/patrickmn/go-cache"
 	"github.com/samber/lo"
@@ -80,6 +81,8 @@ func New(ctx awscontext.Context) *CloudProvider {
 	if err := checkEC2Connectivity(ctx, ec2api); err != nil {
 		logging.FromContext(ctx).Fatalf("Checking EC2 API connectivity, %s", err)
 	}
+	iamapi := iam.New(ctx.Session)
+
 	subnetProvider := NewSubnetProvider(ec2api)
 	instanceTypeProvider := NewInstanceTypeProvider(ctx, ctx.Session, ec2api, subnetProvider, ctx.UnavailableOfferingsCache, ctx.StartAsync)
 	return &CloudProvider{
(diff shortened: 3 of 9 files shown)


ellistarn:
I think this approach just makes things more complex. This is an unfortunate wrinkle on the EC2 capacity APIs, but I don't see how we can meaningfully abstract instanceProfile, since the user will need to be sure the role maps to instanceprofile. I see this potentially resulting in lots of confusion for users.

Is it possible to just name the instanceprofilename == rolename, so that this more or less works transparently?

smcavallo:
Having to set the `instanceProfile` has been a challenge to maintain.  The instance profile ends up being something like `eks-z6c1fd8c-1f20-44b2-6c48-cba9dfaafbc5` - some random GUID-like string which doesn't mean much and isn't apparent which role it is mapped to.  I know that's what a `LaunchTemplate` expects - a name or arn ->  https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata.html#cfn-ec2-launchtemplate-launchtemplatedata-iaminstanceprofile

It is not too painful to use instanceprofiles when you are using launch templates in cloudformation or terraform since you are normally creating both the role AND the launch template using the same infrastructure-as-code template so you can easily reference it.

But for karpenter it's a bit painful since the role is managed by infrastructure-as-code, but we need to set the instanceProfile in a helm chart or other - 2 separate systems - which makes it not easily referenced and becomes a dependency issue.
Another scenario is where a security or operations team manages IAM but an engineering team manages karpenter - this can be tricky.

`NodeGroups` are easier to manage since they do allow a Role to be used (which ends up using the InstanceProfile) ->  https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-eks-nodegroup.html#cfn-eks-nodegroup-noderole
Also - the roleArn (not the instanceProfile) is what is added to the aws-auth configmap.
And for *most* people, the role to instanceprofile is a 1:1 mapping.

From a usability perspective it is so much nicer to set `role: dev-eks-worker` than it is to set `instanceProfile: eks-z6c1fd8c-1f20-44b2-6c48-cba9dfaafbc5` - it can be convention based and you can immediately know it is using the correct role just by the name.  It eliminates a lot of config.
**In general it is easiest to maintain when users don't need to hard-code IDs in config.**

Maybe I am not understanding what you are asking?
I was hoping the `instanceProfile` could stay the same.  But for people who wanted to not have to use a secret-decoder-ring to look up and configure the instance profile, that they could just define a role name, and the code could magically look up the instance profile associated with that role and send that to the launch template API.

Maybe there is a better way to achieve this?
Maybe the code could log the `instanceProfile` which was resolved by the role name?
Or maybe we could write back the instanceProfile and role names or arns to the kubernetes object?

Maybe it would make sense to define the instanceProfile or role on an `AWSNodeTemplate` object?
That would have the additional benefit of being able to use different roles for different types of nodes.
And we could use the Drift Detection feature if it changes? 

We love love love karpenter - it is absolutely amazing :heart: .
We are trying to brainstorm and find ways to simplify our configuration and make it as low maintenance as possible.

FernandoMiguel:
this all makes sense to me

ellistarn:
> Maybe it would make sense to define the instanceProfile or role on an AWSNodeTemplate object?
That would have the additional benefit of being able to use different roles for different types of nodes.
And we could use the Drift Detection feature if it changes?

This is already supported https://karpenter.sh/v0.19.2/aws/provisioning/#instanceprofile.

> The instance profile ends up being something like eks-z6c1fd8c-1f20-44b2-6c48-cba9dfaafbc5 - some random GUID-like string which doesn't mean much and isn't apparent which role it is mapped to. I know that's what a LaunchTemplate expects - a name or arn ->

I'm a touch confused by this one. Karpenter lets you use instance profile name (or ID) via the settings configmap or in the awsnodetemplate. Whatever IAC setup you're using, just make sure that the instanceprofilename==rolename and things should "just work". Since you control the name, it should never change, which allows you to do discovery in case you want to swap out the underlying instance profile policy.

Apologies if I'm missing the point. Maybe you can explain in a different way? What do you use to create instanceprofiles?

FernandoMiguel:
What we provide karpenter is an instances (execution) role (name).
That's invisible from the console.
You can query it via api sdk cli etc.

What the OP would like to provide karpenter instead is the iam role (name or arm) and have karpenter query aws api for the instance role (id?) to attach to the instances.

Hope that helps.

ellistarn:
Yeah I understand the feature ask, but I'm trying to understand why it's necessary. The AWS console hides lots of things, and does a fair bit of magic. You're saying that you're creating roles manually using the console, and it's resulting in the creation of an instance profile with a random guid name, such that when you go to manually install karpenter, it's hard to know the instance profile name? 

I'm having a hard time tracking the user experience through the console or the automation that's installing karpenter.

ellistarn:
Closing for now. Feel free to reopen if you want to continue the discussion.

Issue #2832, opened 2022-11-10
Deregistered Node Not Terminated


cmukai-cisco:
### Version

__Karpenter Version:__ 0.18.1

__Kubernetes Version:__ v1.23

### Expected Behavior

Nodes deregisted from the K8s cluster should be terminated.

### Actual Behavior

A node that was deregistered from EKS was not terminated.

### Steps to Reproduce the Problem

1. Modify provisioner manifest to use smaller instance types.
2. Terminate worker node.
3. Karpenter launched 4 new nodes but consolidated 1 so that only 3 were registered with the cluster.
4. The node that was deregistered was not terminated (see screen shots from EC2 and EKS consoles).
<img width="758" alt="Screenshot 2022-11-10 at 8 31 46 AM" src="https://user-images.githubusercontent.com/30554470/201156810-230426b8-8fa4-4454-933a-df65fd501156.png">
<img width="734" alt="Screenshot 2022-11-10 at 8 39 25 AM" src="https://user-images.githubusercontent.com/30554470/201156855-f4c89a56-63a5-4956-9dea-c56020a4bc74.png">

### Resource Specs and Logs

The node that was not terminated is: ip-10-135-235-20.ec2.internal
[karpenter.log](https://github.com/aws/karpenter/files/9983058/karpenter.log)

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ellistarn:
Karpenter uses a finalizer to ensure that the instance is deleted before we allow the node object to delete. There's a known race condition, where if the node fails to register to the API server and the karpenter controller crashes, it's possible to leak a node. Do you have any logs you can share that might let us know what's going on?

cmukai-cisco:
I posted the Karpenter logs in the "Resource Specs and Logs" section.

jonathan-innis:
It look like the node was cordoned prior to the launching of a new node with the same host name. Do you know what caused that cordoning? Was this a manual action that you performed against the node to delete the node and the new node that was launched just happened to get the same IP?

cmukai-cisco:
[...]

Issue #2831, opened 2022-11-10
Deleting provisioner while draining a node orphans the node


yizhang-zen:
### Tell us about your request

We started deleting a node, then deleted its provisioner. This left the node in deletion status since nothing removed its finalizer.

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

Trying to prevent orphan node when replacing or removing provisioners.

### Are you currently working around this issue?

We manually remove the finalizer to get the orphaned node deleted

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

dewjam:
Hey @yizhang-zen ,
Can you clarify?  Are you asking that Karpenter keep track of Ready nodes even if the ownerRef has been removed?  (after deleting a Provisioner)

yizhang-zen:
Hi @dewjam , I updated the description. Does it make more sense now?

ellistarn:
Makes complete sense.

jonathan-innis:
@yizhang-zen, we have a design doc up now that should address this issue: #2944

grosser:
that looks great 👍

jonathan-innis:
@yizhang-zen @grosser  Have you been able to install version `v0.28.0-rc.1` to try and see if the orphaned node issue you were seeing is resolved?

grosser:
no, we no longer run karpenter, will retry if we pick it back up

Issue #2734, opened 2022-10-26
Karpenter leaves orphaned ec2 instances


ig-leonov:
### Version

v0.18.0

### Expected Behavior

No ec2 instances left when no pods scheduled in cluster

### Actual Behavior

We have and EKS cluster running for 10 days. After that we decided to delete it and found out that 20 ec2 instances are still running. We verified that all these 20 instances have the same pattern in logs.
As I see it, the process is following:
1. Karpenter spins up a node (instance1). Node name is like "123.123.123.123.ec2.internal"
2. after some time EC2 instance health check fails and it's shut down
3. Karpenter spins up a new instance (instance2) which gets the same IP and thus the same node name
4. Karpenter can't register new instance in API server, getting "node already registered" 
5. Karpenter deletes "123.123.123.123.ec2.internal" node but with the old instance id (instance1)
6. instance2 is running forever

### Steps to Reproduce the Problem

I haven't tried to do that but I see the following steps:
1. Have a subnet with 1 address
2. Make karpenter spin up a node with that address
3. kill the instance
4. have another pod to be scheduled so karpenter will reuse IP to create a new node

### Resource Specs and Logs

"2022-10-10T10:10:37.000-0400","Stopping Getty on tty1..."
"2022-10-10T10:10:37.000-0400","Stopping Serial Getty on ttyS0..."
"2022-10-10T10:10:37.000-0400","Stopped target Login Prompts."
"2022-10-10T10:10:37.000-0400","Stopping NTP client/server..."
"2022-10-10T10:10:37.000-0400","Stopping Postfix Mail Transport Agent..."
"2022-10-10T10:10:37.000-0400","chronyd exiting"
"2022-10-10T10:10:37.000-0400","I1010 14:10:37.486257    4863 dynamic_cafile_content.go:170] "Shutting down controller" name="client-ca-bundle::/etc/kubernetes/pki/ca.crt""
"2022-10-10T10:10:37.000-0400","Stopping D-Bus System Message Bus..."
"2022-10-10T10:10:37.000-0400","Stopping Command Scheduler..."
"2022-10-10T10:10:37.000-0400","Stopping Kubernetes Kubelet..."
"2022-10-10T10:10:37.000-0400","Stopped target Multi-User System."
"2022-10-10T10:10:37.000-0400","Stopped target Graphical Interface."
"2022-10-10T10:10:37.000-0400","Unmounting RPC Pipe File System..."
[...]

Issue #2385, opened 2022-08-29
Node provisioning invalid capacity 0 on image filesystem


njohnstone2:
## Version

**Karpenter**: v0.16.0

**Kubernetes**: v1.23.7-eks-4721010

## Notes
This is more of a query and does not affect the functionality of the node. I am only logging as node event monitoring logs this every time a new node is provisioned.

## Expected Behavior

Karpenter creates the EBS volume before initializing the node.

## Actual Behavior

The new node does not detect the EBS volume size and logs `InvalidDiskCapacity` warnings on startup.

## Steps to Reproduce the Problem

- Trigger karpenter node scale up
-  Check node events

### Node Events
    Events:
      Type     Reason                   Age                    From        Message
      ----     ------                   ----                   ----        -------
      Normal   Starting                 2m10s                  kube-proxy  
      Warning  InvalidDiskCapacity      2m18s                  kubelet     invalid capacity 0 on image filesystem
      Normal   NodeAllocatableEnforced  2m18s                  kubelet     Updated Node Allocatable limit across pods
      Normal   Starting                 2m18s                  kubelet     Starting kubelet.
      Normal   NodeHasSufficientMemory  2m15s (x3 over 2m18s)  kubelet     Node ip-10-202-144-6.ap-southeast-2.compute.internal status is now: NodeHasSufficientMemory
      Normal   NodeHasNoDiskPressure    2m15s (x3 over 2m18s)  kubelet     Node ip-10-202-144-6.ap-southeast-2.compute.internal status is now: NodeHasNoDiskPressure
      Normal   NodeHasSufficientPID     2m15s (x3 over 2m18s)  kubelet     Node ip-10-202-144-6.ap-southeast-2.compute.internal status is now: NodeHasSufficientPID
      Normal   NodeReady                105s                   kubelet     Node ip-10-202-144-6.ap-southeast-2.compute.internal status is now: NodeReady

mogggggg:
We just had one node that seems to have had this same issue. After 20 minutes of being NotReady is was removed and there have been instances since then that have started without problems.

Karpenter: v0.15.0
Kubernetes: v1.23.7-eks-7709a84

haarchri:
we have the same issue on v0.19.1 any ideas what happen?

ellistarn:
[...]

PR #2944, opened 2022-11-30
RFC: Karpenter Node Ownership

jonathan-innis:
**Description**

-  Add Node Creation Removal design with RFC from users

Relates to Issues: #1681, #2385, #2831, #2832, #2734, 

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```


designs/node-creation-removal.md
```diff
@@ -0,0 +1,369 @@
+# Karpenter Node Creation Removal
+
+*NOTE: Provisioning is used throughout this doc to reference Karpenter scheduling and spinning up a node. Note that this can apply to both the pending pod scheduling loop and also the consolidation scheduling loop. Essentially, any scheduling loop where Karpenter launches pods.*
+
+## 🔑  Summary
+
+Karpenter’s current node creation logic causes consistency issues as well as race conditions between instance and Node synchronization due to the provisioning loop’s lack of atomicity. Karpenter should remove node creation from its provisioning flow and allow the Kubelet to register the Node on the apiserver. As a result, Karpenter needs some way to model capacity that has been launched but has not yet been registered as a Node on the api-server.
```
ellistarn, line 7:
```suggestion
Karpenter’s current node creation logic causes consistency issues as well as race conditions that lead to leaked instances and reduced node launch performance. With these changes, Karpenter will no longer create nodes, and instead allow the Kubelet to register and deregister the Node with the apiserver. As a result, Karpenter needs a new way to model in flight capacity: instances that have been launched, but has not yet been registered as a Node.
```

designs/node-creation-removal.md
```diff
@@ -0,0 +1,369 @@
...
+
+The recommendation is that this in-flight capacity should be modeled as a Machine CR that represents a Karpenter provisioner/consolidation scheduling decision. This has the following benefits:
```
ellistarn, line 9: Be clear that this API is internal and not to be relied upon by humans.

designs/node-creation-removal.md
```diff
@@ -0,0 +1,369 @@
...
+
+1. Karpenter stops creating the Node object, avoiding race conditions and conflicts with the kube-apiserver and the kubelet that are causing orphaned nodes and instances
+2. Karpenter does not need to know the node name at the end of its provisioning loop, shortening the provisioning loop for clouds that do not know the node name at instance launch time
+3. Kaprenter stops significantly mutating the Node object, avoiding causing conflicts with other Kubernetes tooling which relies on a well-known state for the Node
```
ellistarn, line 13: Significantly is ambiguous here. I think we should be clear about ownerrferences and finalizers, which is another source of leakage (e.g. nodes can't be deleted if karpenter is uninstalled)

```suggestion
3. Karpenter stops significantly mutating the Node object, avoiding causing conflicts with other Kubernetes tooling which relies on a well-known state for the Node
```

designs/node-creation-removal.md
```diff
@@ -0,0 +1,369 @@
...
+4. Karpenter makes its scheduling and de-scheduling decisions observable to users through well-known Kubernetes APIs
```
ellistarn, line 14: I think this is more about inflight capacity than (de)scheduling

designs/node-ownership.md
```diff
@@ -0,0 +1,224 @@
+# Karpenter Node Ownership
+
+*NOTE: Provisioning is used throughout this doc to reference Karpenter scheduling and spinning up a node. Note that this can apply to both the pending pod scheduling loop and also the consolidation scheduling loop. Essentially, any scheduling loop where Karpenter launches pods.*
```
grosser, line 3: 🤷 
```suggestion
Provisioning is used throughout this doc to reference Karpenter scheduling and spinning up a node. This can apply to both the pending pod scheduling loop and also the consolidation scheduling loop. Essentially, any scheduling loop where Karpenter launches pods.
```

designs/node-ownership.md
```diff
@@ -0,0 +1,225 @@
...
+
+The recommendation is that this in-flight capacity should be modeled as an internal Machine CR that represents a Karpenter provisioner/consolidation scheduling decision. This has the following benefits:
+
+1. Karpenter stops creating the Node object, avoiding race conditions and conflicts with the kube-apiserver and the kubelet that are causing orphaned nodes and instances
+2. Karpenter does not need to know the node name at the end of its provisioning loop, shortening the provisioning loop for clouds that do not know the node name at instance launch time
+3. Karpenter stops mutating the Node object by adding labels, ownerReferences and finalizers, avoiding conflicts with other Kubernetes tooling which relies on a well-known state for the Node.
```
olemarkus, line 13: This should not really be a problem. It's perfectly fine for Karpenter to at least modify finalizers and labels. This should not interfere with other tooling. Given that Karpenter owns the label "namespace" and don't try to add/mutate labels owned by others.

designs/node-ownership.md
```diff
@@ -0,0 +1,225 @@
...
+### Kubernetes CRD Object Store (Machine CRD)
+
+Karpenter will no longer create node objects or launch instances as part of the provisioning loop, but, instead, will create Machine CRs at the completion of the provisioning loop. This machine CR will then be picked up by a separate controller that will launch capacity based on the requirements passed from the provisioning loop and will resolve the static values from the **CreateFleet** response into its status.
+
+#### Impact
+
+1. Introduction of the Machine CR as a representation of a “scheduling decision” to launch by the CloudProvider
+2. Migration of Deprovisioning mechanisms from watching and monitoring the Node object to watching and monitoring the Machine CR
+3. Migration of the Node termination finalizer to Machine to orchestrate the cordoning, draining, and deletion
+4. Required Garbage Collection of any Machines that don’t have Nodes mapped to them
```
olemarkus, line 154: Similar to what CCM does, I would suggest htat Karpenter GC Machine objects by watching the cloud API. E.g when an AWS EC2 instance has been terminated, the corresponding Machine CR is removed. Any Machine would exist without a corresponding Node before the Node's creation.

However, I suspect this line is also about terminating the underlying instance when a user terminates the Node object. In which case, you may want to do both.

designs/node-ownership.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Kubernetes CRD Object Store (Machine CRD)
+
+Karpenter will no longer create node objects or launch instances as part of the provisioning loop, but, instead, will create Machine CRs at the completion of the provisioning loop. This machine CR will then be picked up by a separate controller that will launch capacity based on the requirements passed from the provisioning loop and will resolve the static values from the **CreateFleet** response into its status.
+
+#### Impact
+
+1. Introduction of the Machine CR as a representation of a “scheduling decision” to launch by the CloudProvider
+2. Migration of Deprovisioning mechanisms from watching and monitoring the Node object to watching and monitoring the Machine CR
+3. Migration of the Node termination finalizer to Machine to orchestrate the cordoning, draining, and deletion
```
olemarkus, line 153: Do you really want the finalizer on this object? Just thinking it would be better to handle this async rather than making the delete action actually hang waiting for the instance termination.

designs/node-ownership.md
```diff
@@ -0,0 +1,225 @@
...
+4. Required Garbage Collection of any Machines that don’t have Nodes mapped to them
+5. Liveness checks for Nodes that don’t register or become ready after a certain time to delete the backing instance
```
olemarkus, line 155: Delete the backing Machine resource?

designs/node-ownership.md
```diff
@@ -0,0 +1,225 @@
+# Karpenter Node Ownership
+
+_Provisioning is used throughout this doc to reference Karpenter scheduling and spinning up a node. This can apply to both the pending pod scheduling loop and also the consolidation scheduling loop. Essentially, any scheduling loop where Karpenter launches pods._
```
tallaxes, line 3:
```suggestion
_Provisioning is used throughout this doc to reference Karpenter scheduling and spinning up a node. This can apply to both the pending pod scheduling loop and also the consolidation scheduling loop. Essentially, any scheduling loop where Karpenter launches nodes._
```

designs/node-ownership.md
```diff
@@ -0,0 +1,225 @@
...
+## Current Node Creation Problems
+
+This section describes the primary issues that have arisen from users and from maintainer testing due to Karpenter creating the node object after launching the VM instance in the provisioning loop.
+
+### Synchronization
+
+Node objects currently capture the in-flight details for Karpenter’s scheduler to use (allocatable and capacity details for nodes). As a result, the node object *must* be created by the end of the provisioning loop before provisioning can continue. For AWS, this is not an issue since node names can be resolved based on the EC2 **CreateFleet** response; however, for cloud providers such as Azure, this becomes more of an issue, as node name cannot be resolved from the VM create response but can only be resolved when the kubelet creates the node.
+
+While node launch times might be reduced with time, this means that for some cloudproviders (such as Azure), creating the node ties our inflight logic directly to how long the node takes to launch and register with the api-server. This is a non-starter for our scheduling logic since part of Karpenter’s value proposition is its ability to quickly provision capacity for pending pods.
```
tallaxes, line 76: On Azure, as long as VM is created directly - rather than via VM Scale Set - one can generate and specify the host name (node name) upfront, as input to Create VM API call. In this mode of operation (which is very likely to be used by Karpenter Azure provider), waiting to find out the name of the node is not a concern.

jonathan-innis: Got it, I can edit this. It seems like an either/or in this case. Either your cloudprovider has to lookup the node name after creation _or_ it has to know the node name before lauch. Looked into it a bit and talked to some of the AWS cloud-controller-manager folks. The TLDR is that this doesn't look like a solution that adequately scales to all cloudproviders and what it comes down to is that we would still create the node before the kubelet has registered it, meaning we are still ending up with strange error messages (like the one mentioned in [this issue](https://github.com/aws/karpenter/issues/2385)) that result from the node being registered before the underlying instance is actually there.

designs/node-ownership.md
olemarkus: If I understand this correctly, the first thing that happens is that the Machine CR is created. The actual launch of the instance happens consequently. The Machine resource would then require a random name, unrelated to the instance. If the instance is tagged/labelled with `karpenter.sh/machine-name` we have a direct relationship between the CR and the instance itself.

With this relationship it should be no problem for a "controller" to track instances with no corresponding CR, which would mean the CR has been deleted, and the controller can ensure the instance itself is terminated.

I assume the Machine object's status will also be populated both with the cloud provider's instance ID and the node object once these are available.

jonathan-innis: I think the benefit of having the finalizer on the Machine is that we can migrate the cordon, drain, terminate flow of the Node over to the Machine eventually. In general, we can say Machine Delete results in:

1. Cordon of Node
2. Drain of Node
3. Deletion of Instance
4. Deletion of Node

I could potentially see an argument for handling 3&4 async but if we want to do 1&2, then we require a termination finalizer on the Machine already.

> Just thinking it would be better to handle this async rather than making the delete action actually hang waiting for the instance termination

I'm curious what your concern here is. Are you worried that we may hit failure which could leave the Machine in a hung state due to the finalizer?

designs/node-ownership.md
jonathan-innis: I think I get your point here. Machine owns the instance and is solely responsible for it. From a philosophical perspective, it makes sense that the Machine should only care about the instance it owns and nothing else.

jonathan-innis: Updated the RFC based on your suggestion.

jonathan-innis: Updated the RFC based on your suggestion.

designs/node-ownership.md
```diff
@@ -0,0 +1,227 @@
...
+
+This document describes the current node creation flow for Karpenter as well as the rationale for why Karpenter originally created the node object. It then calls out the specific problems with this approach and recommends an alternative approach to creating the Node object that solves for the current approach’s problems.
+
+### Current Instance & Node Creation Flow
+
+Karpenter currently performs the following operations when it detect pending pods and launches capacity to the cluster:
+
+1. Performs a **CreateFleet** request against EC2 APIs which leads to VM Instance Creation
```
sftim, line 29: Would that really be a MachineSet rather than a Machine (both could still be custom resources that Karpenter defines)?

designs/node-ownership.md
```diff
@@ -0,0 +1,227 @@
...
+
+**This is wrong, since it means that the node provider ID will map to the old instance and the termination finalizer logic will no longer handle deletion of the instance that Karpenter provisioned.**
+
+This issue is primarily caused by a few issues:
+
+1. cloud-controller-manager that is responsible for cleaning up nodes that no longer have instances tied to it does not have a quick enough reconciliation loop to cleanup the node before Karpenter provisions the new capacity. *NOTE: In general, we shouldn’t rely on this timing anyways since Kubernetes is eventually consistent.*
+2. The subnet is highly constrained leading to IP hostname naming conflicts with the node
+3. EKS nodes rely on hostnaming, which ultimately causes this conflict. This problem could theoretically be circumvented by using [resource name-based naming](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-naming.html) that would ensure each node has a unique name, regardless of IP.
```
sftim, line 94: Perhaps we should explicitly recommend resource based naming (we can still support the other way).

designs/node-ownership.md
```diff
@@ -0,0 +1,227 @@
...
+3. Migration of the Node termination finalizer to Machine to orchestrate the cordoning, draining, and deletion
+4. Required Garbage Collection of any Machines that don’t have instances mapped to them
+5. Liveness checks for Nodes that don’t register or become ready after a certain TTL to delete the backing Machine
+6. In-flight representations of Nodes for scheduling become the Machine `.status.allocatable` until the Node registers and initializes its `.status.allocatable`
+7. Mirroring Machine labels and annotations onto Nodes through a reconciliation mechanism after Node join
+
+### In-Memory Node Store
+
+An alternative to storing persistent data on the Kubernetes API server is to utilize the cloud provider’s instance data from a **ListMachines** call and utilize the instance data that is stored in cloud provider’s instance store to understand our in-flight capacity. This remote store means we no longer need to create the node for inflight capacity, but we still need to operate on the node object for details like ownership and node termination/provisioner ownership.
```
sftim, line 161: Another alternative: store pending instance information in the `.status` of each Provisioner, and limit the number of in-flight VM launches.

Analogy: the Job API in Kubernetes uses `.status` to track completed Jobs prior to being able to track them, making atomic updates to the Job to record a count of completions.

sftim: This approach also gives Karpenter a way to recover from an outage that occurs after creating the Fleet and before recording the new instance: on detecting a suitably tagged fleet, Karpenter can record that detection in the provisioner's `.status`.

This still might need a CR that is specific to the particular cloud provider (like an AWSNodeTemplate, but for provider-specific status). I'm not sure.

sftim (commented):
Overall, I wonder if we've considered using `.status` on each Provisioner to store node launch state.

designs/node-ownership.md
jonathan-innis: I don't think we can just rip out IP-based naming, particularly because it is only supported on a later version of AWS K8s that not all customers have migrated to, and also because just solving this doesn't cover the other problems that are outlined in the design.

Yes, we can recommend that users use this naming convention and/or use interruption handling to circumvent this race, but I think it's a non-starter to force everyone over to a naming convention.

designs/node-ownership.md
jonathan-innis: I think this still relies on a reconciliation mechanism to LIST out the instances that have already been created based on tag in the case of failure to store into the status like you have described.

This, to me, seems like an extension of keeping track of details in-memory (notably, a more observable one) since we will still have to perform essentially a LIST/WATCH in order to deal with crashes/restarts/outages.

designs/node-ownership.md
sftim: Yep, agreed. “Explicitly recommend” is different from ”insist upon”.

jonathan-innis:
> Overall, I wonder if we've considered using `.status` on each Provisioner to store node launch state.

This is an interesting approach. Overall, I think we leaned away from doing it this way since it closely ties the instance's lifecycle in with the Provisioner's lifecycle. Ideally, it would be nice if we could decouple instance launch and lifecycle from Provisioner, since we could see the Provisioner as an extension point of instance launch i.e. "a way to managed flexible instances".

jonathan-innis:
Good feedback @sftim!

designs/node-ownership.md
olemarkus: I am actually more concerned about the client having to wait for the termination to complete before carrying on. If I want to delete a list of Machines, the process would be sequential.

1&2 doesn't require a finalizer. At least I don't quite understand why it would require one.

I actually question if Karpenter should ever cordon & drain, or if this should be up to kubelet (graceful node shutdown) or NTH (which is now also shipped with karpenter). I.e Karpenter should just directly terminate the instance and then rely on other things for 1 & 2.

4. should never be done by Karpenter. It's CCM that deletes Node objects.

sftim: We want _something_ to take care of load skewing to empty the nodes that Karpenter no longer wants running. We can do that by tainting the node, or via other means. We definitely do want the descheduling to happen.

IMO a finalizer on a Machine is a good way to let Karpenter record (for its own use) an intent to stop using a node. To me, this feels like what finalizers are intended for.

As a response to the scheduled Machine deletion, Karpenter could take descheduling steps such as tainting the node (either `PreferNoSchedule` or `NoSchedule`, or even both at different times) and requesting Pod eviction.

olemarkus: Maybe take a step back to consider the ordering @jonathan-innis mentioned:

1. Delete Machine with finalizer
2. Cordon of Node
3. Drain of Node
4. Deletion of Instance
5. Remove finalizer, completing Machine deletion
6. Deletion of Node

As mentioned, 6 is not a Karpenter concern, so I think we can all agree to drop that one.

If we drop the finalizer, the actual condition for instance termination is the same: machine object must be deleted, Node must be cordoned and drained. The intent is the deletion of the machine object and the reconciliation loop picks up this one and acts on it the same way as before. It just doesn't have to remove the finalizer.

Since Karpenter also needs to explicitly track state of EC2 and Machine objects, the finalizer's purpose is _only_ the blocking of deletion.

sftim: If we didn't have a finalizer on Machine, the problem then becomes that Karpenter loses its record that it owns the node, _before_ cleanup is complete. Much easier to keep the Machine around until the Node actually disappears (or Karpenter gives up).

This also helps with the case where the primary Karpenter is running on a node that we decide to drain, right? If the design makes it easy for one instance of Karpenter to start the scale-in and another to finish the work, that's ideal.

I don't think we need a finalizer on Node.

olemarkus: You mean that Karpenter owns the _instance_? The instance ownership should be tracked with a tag in EC2. EC2 has to be the source of truth for that. If a Karpenter pod is started without state, it has to check k8s and EC2 and reconcile. E.g EC2 instance exists, machine object does not means EC2 needs to be drained and terminated.

sftim: Sort of. How does Karpenter drain a cloud instance when the record of the node-to-instance mapping has already been deleted?

We can't assume that we can use the EC2 API or knowledge about EC2 instance hostnames, because Karpenter has an ambition to support more than one cloud provider.

olemarkus: All cloud providers should have an idea of tags or labels on the instance side. So EC2 for AWS, compute engine API on GCP etc. Cloud Controller Manager already does this for Node objects (instance terminated equals Node object removal).

olemarkus: Can add that finalizers _may_ be used to track state, but consider it more informational and perhaps use the information to save some cloud API calls. But it can't rely on finalizers for the actual state and Machine cannot be considered the source of truth of whether an instance runs or not.

jonathan-innis: > You mean that Karpenter owns the instance?

Your point is that when Karpenter restarts in the middle of instance deletion/node cordon/drain that we should rediscover it through the tag and re-create the machine? This effectively requires a LIST call on the CloudProvider API to hydrate in any machines that exist that Karpenter isn't tracking. With the finalizer, we don't have to worry about the LIST call since we can be assured that the Machines that exist on the cluster will always be a superset of the instances that exist in EC2

jonathan-innis: > I actually question if Karpenter should ever cordon & drain, or if this should be up to kubelet

I think it's highly important that Karpenter performs this logic to ensure that we don't delete the underlying instance until all the workloads running on that node are removed from that cluster. In the graceful termination case, there is a fixed time for the workload to cleanup, but it won't block indefinitely which is a much riskier proposition to put the customer into.

I could potentially see an argument for forcing deletion of the instance assuming that a user is performing manual deletion of the machine (other cases involve Karpenter automation simulating the termination and not even starting the termination if we know that the drain won't succeed).

designs/node-ownership.md
jonathan-innis: I can clarify in the doc. The bulk of the point here is that we can't Create the node with these values, which means they have to reconcile on. When these pieces are the basis for ownership and termination, this causes races where some dependency goes away before the reconciliation can occur (this is what is mentioned in the race condition section)

designs/node-ownership.md
jonathan-innis: Agreed, what we have settled on is relying on the `.spec.providerID` as the mapping between node/machine on the cluster, which is a known mapping (CCM uses this for instance). If node name == machine name == instance name, that's just a benefit for user observability

designs/node-ownership.md
jonathan-innis: Can you describe what you mean? I'm not seeing how the line you are pointing at here links to the question?

sftim: Ah, what I meant is that `CreateFleet` creates a _set_ of machines, so the CR should be a MachineSet not a Machine.

Maybe there's a good explanation why this is wrong; for example, we tell people writing cloud provider integrations that Karpenter must only ever create fleets (and similar abstractions) with a max of 1 running server.

designs/node-ownership.md
olemarkus: I'll do this one first since it is simple.

> > I actually question if Karpenter should ever cordon & drain, or if this should be up to kubelet
> 
> I think it's highly important that Karpenter performs this logic to ensure that we don't delete the underlying instance until all the workloads running on that node are removed from that cluster. In the graceful termination case, there is a fixed time for the workload to cleanup, but it won't block indefinitely which is a much riskier proposition to put the customer into.
> 
> I could potentially see an argument for forcing deletion of the instance assuming that a user is performing manual deletion of the machine (other cases involve Karpenter automation simulating the termination and not even starting the termination if we know that the drain won't succeed).

I agree ... I think it is important that the instance is fully drained before termination. Mainly due to the compaction feature and that Karpenter it is not improbable for Karpenter to assume Pods can quickly reschedule when they can't.

olemarkus: > > You mean that Karpenter owns the instance?
> 
> Your point is that when Karpenter restarts in the middle of instance deletion/node cordon/drain that we should rediscover it through the tag and re-create the machine? This effectively requires a LIST call on the CloudProvider API to hydrate in any machines that exist that Karpenter isn't tracking. With the finalizer, we don't have to worry about the LIST call since we can be assured that the Machines that exist on the cluster will always be a superset of the instances that exist in EC2

I don't think Machine objects should be recreated (just like Nodes are not recreated). I think the instance should be terminated. I also don't see finalizers as safe enough to guarantee one doesn't leak resources. As mentioned CCM does LIST calls every X seconds against the cloud provider to ensure it's internal list of instances match the cloud provider.

sftim: In the sketched design from this RFC PR, Karpenter is a controller for Nodes, but only indirectly. It's fine for c-c-m to continue with this behavior, but I'm still expecting that Karpenter's primary reconcile mechanism becomes one where it matches Machines against Nodes and taking appropriate action. Karpenter also triggers indirect Node creations by driving Machines to their desired (ie, running) state.

Kubernetes offers [finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) for cleanup and that's seems to be like a good API mechanism to use here. If we find a Machine with no finalizer set, we can consider adopting it providing it matches the right selector. If someone force-deletes a Machine, we can semi-gracefully terminate the appropriate instance without draining it first (and that person might regret their API call choices).

If Karpenter wants to shut down a server, Karpenter can use its registered finalizer on the Machine to mark the _intent_ to delete and carry out a controlled drain, etc. In my view, that's a really good architectural fit with Kubernetes' API conventions. It also allows any other controller to hook into the same mechanism - for example, a spot interruption, or something like one, could also trigger a delete on the Machine. It's nice to have a design that lets us keep extra complexity apart from Karpenter where we can.

designs/node-ownership.md
jonathan-innis: In our case, when we make a `CreateFleet` call, we aren't asking for a set of machines, we're really asking for a single machine at a time, which we batch together if the requirements are "similar enough"

It also makes the management experience post-launch difficult to reason about since there isn't a way to represent each Machine as a single entity

sftim: We _could_ model this as a MachineSet _and_ a set of Machines, but it sounds like there's little need.

designs/node-ownership.md
sftim: Storing status into the cluster's API is _not_ the same as keeping it withing Karpenter's memory. In the first case, a cluster operator can observe Karpenter's view of the actual state.

tallaxes (commented):
LGTM

bwagner5 (approved):
LGTM

ellistarn:
🥳

PR #2964, opened 2022-12-01
docs: RFC v1beta1

### LINKED ISSUE #1327: v1beta1 Laundry List

ellistarn:
Fixes #1327 

**Description**

v1beta1 proposal. 

**Release Note**

```release-note

```

### DIFF
--- a/designs/v1beta1.md
+++ b/designs/v1beta1.md
@@ -1,0 +1,93 @@
+# v1beta11
+
+This document formalizes the [v1beta1 laundry list](https://github.com/aws/karpenter/issues/1327) into a release strategy and concrete set of proposed changes.
+
+### Migration path from v1alpha5 to v1beta1
+
+Customers will be able to migrate from v1alpha5 to v1beta1 in a single cluster using a single Karpenter version.
+
+Kubernetes custom resources have built in support for API version compatibility. CRDs with multiple versions must define a “storage version”, which controls the data stored in etcd. Other versions are views onto this data and converted using [conversion webhooks](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion). However, there is a fundamental limitation that all versions must be safely [round-tripable through each other](https://book.kubebuilder.io/multiversion-tutorial/api-changes.html). This means that it must be possible to define a function that converts a v1alpha5 Provisioner into a v1beta1 Provisioner and vise versa.
+
+Unfortunately, multiple proposed changes in v1beta1 are not round-trippable. Below, we propose two deprecations of legacy fields in favor more modern mechanisms that have seen broad adoption in v1alpha5. These changes remove sharp edges that regularly cause customer surprise and production pain.
+
+To work around this limitation, we have three options:
+
+1. [Recommended] Rename Provisioner to something like NodeProvisioner, to avoid being subject to round-trippable requirements
+2. Require that users delete the existing v1alpha5 Provisioner CRD and then install the v1beta1 Provisioner CRD. This will result in all capacity being shut down, and cannot be done in place if the customer has already launched nodes.
+3. Keep the legacy fields in our API forever
+
+Option 2 is untenable and easily discarded. We must provide a migration path for existing customers. Option 3 minimizes immediate impact to customers, but results in long term customer pain. There are other benefits to renaming Provisioner, so while it does cause some churn, it results in long term value.
+
+Following option #1, customers would upgrade as follows:
+
+1. Install Karpenter version 0.x.0, this version supports both v1alpha5.Provisioner and v1beta1.NodeProvisioner
+    1. We may choose to support both APIs in more than one Karpenter version at the cost of developer velocity
+2. Create a v1beta1.NodeProvisioner, and add a taint to the old v1alpha5.Provisioner
+3. Depending on disruption tolerance, trigger rolling restarts to all deployments, delete the provisioner, or simply wait.
+
+### Implementation Plan
+
+The Karpenter codebase currently references v1alpha5.Provisioner in most of its controllers. While v1alpha5 and v1beta1 are not roundtrippable, we can restore the broken relationship between v1alpha5.Provisioner.spec.provider and v1beta1.AWSNodeTemplate in memory. This can be achieved by json encoding provisioner.spec.provider into an annotation karpenter.sh/compatibility/provider. Any area of code that needs to maintain compatibility will be required to parse out these values. Luckily, this is contained to the AWS cloud provider, as spec.provider predates prototyping efforts from other vendors.
+
+Alternatively, we can choose to replicate existing v1alpha5 controllers into a new folder controllers/v1beta1/... where each set of versioned controllers operates independently on Provisioners of different API version. This will result in a much larger volume of code in the short term, but keep the complexity isolated. Developers will need to be aware of this duplication as long as it exists, and backport fixes to both places where necessary. Because of this, we will limit the number of Karpenter releases that support both versions and eventually deprecate support for v1alpha5.
+
+Either approach impacts the velocity of the project — we will need to decide when is the best time to make these changes.
+
+## Proposed Changes
+
+### Rename: Provisioner → NodeProvisioner
+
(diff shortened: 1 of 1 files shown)


designs/v1alpha6.md
```diff
@@ -0,0 +1,99 @@
...
+2. [Recommended] Require that users delete the existing v1alpha5 Provisioner CRD and then install the v1alpha6 Provisioner CRD. This will result in all capacity being shut down, and cannot be done in place if the customer has already launched nodes.
+3. Keep the legacy fields in our API forever
+
+Option 2 is untenable and easily discarded. We must provide a migration path for existing customers. Option 3 minimizes immediate impact to customers, but results in long term customer pain. There are other benefits to renaming Provisioner, so while it does cause some churn, it results in long term value.
+
+Following option #1, customers would upgrade as follows:
+
+1. Install Karpenter version 0.x.0, this version supports both v1alpha5.Provisioner and v1alpha6.NodeProvisioner
+    1. We may choose to support both APIs in more than one Karpenter version at the cost of developer velocity
+2. Create a v1alpha6.NodeProvisioner, and add a taint to the old v1alpha5.Provisioner
```
sftim, line 25: Could the new Karpenter version automate that transition? I think it's technically possible.

If not, we could provide a container image that runs a one-time migration job (either in-cluster with appropriate access, or to run locally).

designs/v1alpha6.md
```diff
@@ -0,0 +1,99 @@
...
+* https://kubernetes.slack.com/archives/C02SFFZSA2K/p1669605517500229?thread_ts=1669289308.636989&cid=C02SFFZSA2K
+
+### Rename: Move provisioner.spec to provisioner.spec.template
+
+Currently fields that control node properties, such as Labels, Taints, StartupTaints, Requirements, Kubelet, ProviderRef, are top level members of provisioner.spec. These are siblings of fields that control provisioning behavior, such as Limits, TTLSecondsUntilExpired, and Weight.
+
+Instead, we should separate out properties of the node in the same way that pod properties are separated from their parent controllers. This has a nice effect of simultaneously creating a place for users to define node metadata. This design mirrors that of kubernetes objects that contain podtemplatespecs, like Deployment, StatefulSet, Job, and CronJob.
+```
+spec:
+  ttlSecondsUntilExpired: ...
```
sftim, line 59: **Opinion**

Consider moving this into a containing field such as `nodeLifecycle`. We could do that for `v1beta1` though as the API change would survive round trips.

designs/v1alpha6.md
```diff
@@ -0,0 +1,99 @@
...
+      providerRef: ...
+```
+
+### Rename: kubeletConfiguration → kubelet
+
+This is a minor change that favors a less verbose naming style, while conveying the same information to the user.
+
+### Deprecate: provisioner.spec.provider
+
+We’ve recommended that customers leverage spec.providerRef in favor of spec.provider since Q2 2022. Documentation for this feature has been removed since Q3 2022. We will take the opportunity to remove the feature entirely to minimize code bugs/complexity and user confusion.
```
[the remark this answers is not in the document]
sftim: If deprecating (ie, retaining, but asking cluster operators not to use the field, then optionally: emit a `Warning:` if that field is set on a NodeProvisioner (via an admission webhook).

See https://kubernetes.io/blog/2020/09/03/warnings/ for how these work.

This `Warning:` behavior could also arrive via a Karpenter release that comes out ahead of the switch to v1alpha6.

designs/v1alpha6.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Deprecate: awsnodetemplate.spec.launchTemplate
+
+Direct launch template support is problematic for many reasons, outlined in the [design](https://github.com/aws/karpenter/blob/main/designs/aws/aws-launch-templates-v2.md). Customers continue to run into issues when directly using launch templates. From many conversations with customers, our AWSNodeTemplate design has achieved feature parity with launch templates. The only gap is for users who maintain external workflows for launch template management. This requirement is in direct conflict with users who run into this sharp edge.
+
+This change simply removes legacy support for launch templates in favor of the AWSNodeTemplate design.
```
sftim, line 86: I'm not sure if this is a removal or a deprecation; the text is ambiguous.

designs/v1alpha6.md
```diff
@@ -0,0 +1,99 @@
...
+1. Rename Provisioner to something like NodeProvisioner, to avoid being subject to round-trippable requirements
+2. [Recommended] Require that users delete the existing v1alpha5 Provisioner CRD and then install the v1alpha6 Provisioner CRD. This will result in all capacity being shut down, and cannot be done in place if the customer has already launched nodes.
+3. Keep the legacy fields in our API forever
+
+Option 2 is untenable and easily discarded. We must provide a migration path for existing customers. Option 3 minimizes immediate impact to customers, but results in long term customer pain. There are other benefits to renaming Provisioner, so while it does cause some churn, it results in long term value.
+
+Following option #1, customers would upgrade as follows:
+
+1. Install Karpenter version 0.x.0, this version supports both v1alpha5.Provisioner and v1alpha6.NodeProvisioner
+    1. We may choose to support both APIs in more than one Karpenter version at the cost of developer velocity
```
sftim, line 24: **Opinion**

We could support both APIs for several minor releases that are essentially the same underlying code but with different release-time behaviors around API support. The way I picture it, those changes _mostly_ cover three things: updates to Helm charts, default parameters, and docs.

Doing that lets people pick the release that gets the behavior they like. Eg:
- v0.23.0 for **Provisioner favored, NodeProvisioner available** for early migration
   - this is a big release because it adds the v1alpha6 API group version and any related features
-  v0.24.0 for **NodeProvisioner favored, Provisioner supported**, no warnings emitted
-  v0.25.0 for **NodeProvisioner favored, Provisioner supported**, warnings on creating a Provisioner
-  v0.25.0 for **NodeProvisioner favored, Provisioner supported, creating a new Provisioner _prohibited_** by default
- v0.26.0 for **NodeProvisioner available**, Provisioner API no longer present
  - this is an important release because it drops the v1alpha5 API group version

The way I imagine this, those releases could come out 1 calendar week apart. (I don't know how automated the release process is). Or even more frequently, if folks want to. Once made, those releases stay published and people who want to stick with v1alpha5 have a few options about that.

_PS._ https://kubernetes.io/blog/2020/09/03/warnings/ has some details about `Warning:` headers; an admission webhook can set these whilst still admitting an object. It's a nice way to tell ends users that they're not doing something the best way.

sftim (changes requested):
Thanks for the PR. I recommend some clarifications, and I've added other opinions.

sftim:
Related: https://github.com/kubernetes/kubernetes/issues/115139

> As far as I can tell, we don't hav e acomprehensive doc which covers the expected lifecycle of nodes in Kubernetes.
> 
> Specifically, we have lots of intersecting, async things which involve nodes. For example:
> 
> -   Many environments have VMs "behind" Nodes. Those VMs can be deleted without telling k8s.
> -   Many environments have subsystems which cross-reference things which need to coordinate with node lifecycle. E.g. the service controller puts VMs into LBs.
> -   Some components manage nodes (e.g. Cluster Autoscaler, Karpenter).

&hellip;

> For an example of things that I think are "weird" for lack of docs, look at https://github.com/kubernetes/autoscaler/issues/5201#issuecomment-1385949774 . ClusterAutoscaler defines a taint which it uses to prevent work from landing on "draining" nodes (even though we have the unschedulable field already). The service LB controller currently uses that taint to manage LBs. Cluster autoscaler removes the VM from the cloud, and leaves the Node object around for someone else to clean up.
> 
> The discussion is about the MEANING of the taint, when it happens, and how to be more graceful. What we want is a clear signal that "this node is going away" and a way for 3rd parties to indicate they have work to do when that happens. It strikes me that we HAVE such a mechanism - delete and finalizers. But CA doesn't do that. I don't know why, but I suspect there are reasons. Should it evolve?

designs/v1beta1.md
```diff
@@ -0,0 +1,93 @@
+# v1beta11
```
njtran, line 1:
```suggestion
# v1beta1
```

jonathan-innis:
This RFC is going to go under a re-write before it's re-proposed as we're still trying to capture more items from #1327. Closing this for now since the discussion has gone a bit stale since the initial RFC was opened. I'll re-open with the revised version when the proposed items for v1beta1 are finalized.

PR #3013, opened 2022-12-09
docs: Retain current path in new version docs link

rothgar:
Fixes # 

**Description**
When clicking on "latest verison" from the old version pop-up this should retain the current path so the user isn't send back to the main docs page. If the path doesn't exist in the new version they'll get the standard 404 page.

* running docs locally with multiple versions

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


tzneal:
I think you may have written then against an old version of the docs, can you rebase?

rothgar:
Rebased. We may want to handle some redirects for people as content has moved around a lot. This would avoid internal and external links breaking.

spring1843 (approved):
LGTM

PR #3017, opened 2022-12-09
docs: Revise Karpenter Getting Started into one guide

chrisnegus:
Fixes Issue #[1574](https://github.com/aws/karpenter/issues/1574)

**Description**: This PR proposes consolidating the Karpenter Getting Started experience into a single guide. It assumes that the user will bring their own cluster, or choose from a list links to different cluster setup instructions outside of Karpenter that we provide. 

Because this is a major overhaul of the docs, I'm marking it as a work in progress for now. I would like lots of people to try it and help decide if we are going down the right path.

Keep in mind that I've left the other Getting Started guides in for the time being since removing them was breaking lots of links and preventing the docs from building.


website/content/en/preview/getting-started/_index.md
```diff
@@ -13,3 +13,241 @@ Learn more about Karpenter and how to get started below.
...
+* Your workloads have pod disruption budgets that adhere to [EKS best practices](https://aws.github.io/aws-eks-best-practices/karpenter/)
+* Your cluster has an [OIDC provider](https://docs.aws.amazon.com/eks/latest/userguide/enable-iam-roles-for-service-accounts.html) for service accounts
+* Your nodes may be part of one or more node groups
+
+This guide will also assume you have the `aws` and `kubectl` CLI tools installed.
+You can also perform many of these steps in the console, but we will use the command line for simplicity.
+
+## Get a cluster
+To use Karpenter, you must be running a Kubernetes cluster on a supported cloud provider.
+Currently, only AWS is supported.
```
ellistarn, line 33: Thoughts on scoping this guide to just karpenter on EKS and not mention other options or lack thereof -- this will help the content survive the test of time.

website/content/en/preview/getting-started/_index.md
```diff
@@ -13,3 +13,241 @@ Learn more about Karpenter and how to get started below.
...
+
+If you delete a node with kubectl, Karpenter will gracefully cordon, drain,
+and shutdown the corresponding instance. Under the hood, Karpenter adds a
+finalizer to the node object, which blocks deletion until all pods are
+drained and the instance is terminated. Keep in mind, this only works for
+nodes provisioned by Karpenter.
+
+{{% script file="./scripts/step15-delete-node.sh" language="bash"%}}
+
+## Cleanup
```
ellistarn, line 243: Maybe worth removing this section now since you don't know how the user created the cluster.

ellistarn:
Nice!

website/content/en/preview/getting-started/_index.md
chrisnegus: Okay. How should we handle kOps then? Should I just make it clear that the steps provided in this guide are only supported for EKS? That way we could still mention the kOps procedure, since it handles adding Karpenter itself. Or should we just drop the link to kOps?

website/content/en/preview/getting-started/scripts/step11-create-provisioner.sh
```diff
@@ -0,0 +1,26 @@
...
+apiVersion: karpenter.sh/v1alpha5
+kind: Provisioner
+metadata:
+  name: default
+spec:
+  requirements:
+    - key: karpenter.k8s.aws/instance-family
+      operator: In
+      values: [c, m, r]
+    - key: karpenter.k8s.aws/instance-generation
```
FernandoMiguel, line 11: we should filter to only be nitro instances too

website/content/en/preview/getting-started/_index.md
chrisnegus: I'll remove the `eksctl` command for deleting the cluster. I think the other steps would be a good way to remind them how to clean up the Karpenter stuff they added from the guide.

website/content/en/preview/getting-started/scripts/step11-create-provisioner.sh
ellistarn: IMO we should just rely on Karpenter's defaults, here.

website/content/en/preview/getting-started/_index.md
```diff
@@ -13,3 +13,241 @@ Learn more about Karpenter and how to get started below.
...
+
+If you have not already configured Karpenter when you set up your Kubernetes cluster in AWS, follow the instructions below to add Karpenter to your existing cluster.
+
+## Add Karpenter to an existing cluster
+
+To add Karpenter to your cluster, you need to cread IAM roles, add tags to subnets and security groups, and update the aws-auth ConfigMap.
+
+### Create IAM roles
+
+To get started, create two new IAM roles for nodes provisioned with Karpenter and the Karpenter controller.
```
bryantbiggs, line 49: this can be simplified further to only requiring an IRSA role. Since Karpenter needs something to run on, then there already exists some form of compute that is connected to the control plane. tl;dr - there has to be at least one node with an IAM role and instance profile and the necessary permissions before Karpenter can even be provisioned. Users can simply re-use this role/instance profile. Just FYI

website/content/en/preview/getting-started/scripts/step02-node-policies.sh
```diff
@@ -0,0 +1,11 @@
+aws iam attach-role-policy --role-name KarpenterInstanceNodeRole \
+    --policy-arn arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy
+
+aws iam attach-role-policy --role-name KarpenterInstanceNodeRole \
+    --policy-arn arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy
+
+aws iam attach-role-policy --role-name KarpenterInstanceNodeRole \
+    --policy-arn arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly
+
+aws iam attach-role-policy --role-name KarpenterInstanceNodeRole \
```
bryantbiggs, line 10: this is not required so I would consider leaving this out. I will go out on a limb and say that I, like others, assumed it was required after seeing it in the guide

spring1843 (dismissed):
/karpenter snapshot

website/content/en/preview/getting-started/_index.md
chrisnegus: Okay. I changed the text to say that only EKS on AWS is currently supported. I said that if they wanted to try Karpenter on unsupported Kubernetes distributions on AWS, they could try kOps. I left the link to kOps instructions in.

website/content/en/preview/getting-started/scripts/step11-create-provisioner.sh
chrisnegus: Correct me if I'm wrong, but these seem to be the instance-family and instance-generation defaults. If that is true, do you want me to cut them, rather than show them here explicitly? Or should I just leave them as they are?

website/content/en/preview/getting-started/_index.md
```diff
@@ -13,3 +13,241 @@ Learn more about Karpenter and how to get started below.
...
+
+* [Creating an Amazon EKS cluster](https://docs.aws.amazon.com/eks/latest/userguide/create-cluster.html): Use `eksctl` or `aws` CLI tools or the AWS Management console to deploy an EKS cluster.
+* [Amazon EKS Blueprints for Terraform](https://aws-ia.github.io/terraform-aws-eks-blueprints): Follow a basic [Getting Started](https://aws-ia.github.io/terraform-aws-eks-blueprints/v4.18.0/getting-started/) guide and also add modules and add-ons. This includes a [Karpenter](https://aws-ia.github.io/terraform-aws-eks-blueprints/v4.18.0/add-ons/karpenter/) add-on that lets you bypass the instructions in this guide for setting up Karpenter.
+* [kOps](https://kops.sigs.k8s.io/operations/karpenter/): These instructions describe how to create a kOps Kubernetes cluster in AWS that includes Karpenter.
+
+If you have not already configured Karpenter when you set up your Kubernetes cluster in AWS, follow the instructions below to add Karpenter to your existing cluster.
+
+## Add Karpenter to an existing cluster
+
+To add Karpenter to your cluster, you need to cread IAM roles, add tags to subnets and security groups, and update the aws-auth ConfigMap.
```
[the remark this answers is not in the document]
chrisnegus: Good catch! Fixed.

website/content/en/preview/getting-started/_index.md
chrisnegus: I'm not sure exactly how to say that. Can you suggest specific wording?

chrisnegus:
I made the change from instance-family to instance-category, as suggested in https://github.com/aws/karpenter/pull/3079

website/content/en/preview/getting-started/scripts/step02-node-policies.sh
chrisnegus: Would deleting lines 7-10 take care of that then?

website/content/en/preview/getting-started/_index.md
bryantbiggs: Here is what that could look like (changes based on your PR here) https://github.com/aws/karpenter/commit/b4ec84d65af52bc8c35cbdd871696577ad63e1ea

This cuts out quite a bit of overhead for users since there is already compute required before provisioning Karpenter which that means there is an existing IAM instance profile that is already in the `aws-auth` configmap (unless Karpenter is running on Fargate, in which case an IAM instance profile would need to be created). 

I will leave it up to you all to decided if this is worthwhile or not - just a suggestion to make this easier and with fewer resources/steps

website/content/en/preview/getting-started/scripts/step02-node-policies.sh
bryantbiggs: yes

website/content/en/preview/getting-started/_index.md
chrisnegus: Thanks @bryantbiggs . I've incorporated the changes you suggested.

website/content/en/preview/getting-started/scripts/step02-node-policies.sh
chrisnegus: I've deleted the whole step, per the suggestion in [b4ec84d](https://github.com/aws/karpenter/commit/b4ec84d65af52bc8c35cbdd871696577ad63e1ea)

website/content/en/scripts/step11-create-provisioner.sh
```diff
@@ -0,0 +1,26 @@
...
+      operator: Gt
+      values: ["2"]
+  providerRef:
+    name: default
+---
+apiVersion: karpenter.k8s.aws/v1alpha1
+kind: AWSNodeTemplate
+metadata:
+  name: default
+spec:
```
ellistarn, line 21: WDYT about configuring instanceProfile here, since we're considering removing it from the global configmap in our next api version.

website/content/en/preview/getting-started/_index.md
```diff
@@ -13,3 +13,211 @@ Learn more about Karpenter and how to get started below.
...
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: karpenter.sh/provisioner-name
+                operator: DoesNotExist
+            - matchExpressions:
+              - key: eks.amazonaws.com/nodegroup
+                operator: In
+                values:
+                - ${NODEGROUP}
```
ellistarn, line 116: I wonder if this is necessary? It should be enough to simply say "don't run karpenter on itself", which is already included in the helm chart.

website/content/en/scripts/step10-deploy.sh
```diff
@@ -0,0 +1,6 @@
+kubectl create namespace karpenter
```
ellistarn, line 1: Did you mean to include this? This must already exist when we do the helm install.

website/content/en/scripts/step09-generate-chart.sh
```diff
@@ -0,0 +1,6 @@
+helm template karpenter oci://public.ecr.aws/karpenter/karpenter --version ${KARPENTER_VERSION} --namespace karpenter \
```
ellistarn, line 1: Why not use `helm install --upgrade --create-namespace`?

website/content/en/scripts/step11-create-provisioner.sh
```diff
@@ -0,0 +1,26 @@
+cat <<EOF | kubectl apply -f -
+apiVersion: karpenter.sh/v1alpha5
+kind: Provisioner
+metadata:
+  name: default
+spec:
+  requirements:
```
ellistarn, line 7:
```suggestion
  # If requirements are left empty, karpenter will default to the following. They may also be set explicitly
  requirements:
```

website/content/en/preview/getting-started/_index.md
```diff
@@ -13,3 +13,211 @@ Learn more about Karpenter and how to get started below.
...
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: eks.amazonaws.com/nodegroup
+                operator: In
+                values:
+                - ${NODEGROUP}
+```
+
+## Remove Cluster Autoscaler
```
ellistarn, line 154: I wonder if it's worth including this in the getting started guide. Actually migrating from CAS requires a lot more intention/planning.

website/content/en/preview/getting-started/_index.md
```diff
@@ -13,3 +13,211 @@ Learn more about Karpenter and how to get started below.
...
+{{% script file="./scripts/step10-deploy.sh" language="bash" %}}
+
+## Create default provisioner
+
+We need to create a default provisioner so Karpenter knows what types of nodes we want for unscheduled workloads.
+You can refer to some of the [example provisioners](https://github.com/aws/karpenter/tree{{< githubRelRef >}}examples/provisioner) for specific needs.
+
+{{% script file="./scripts/step11-create-provisioner.sh" language="bash" %}}
+
+## Set nodeAffinity for critical workloads (optional)
```
ellistarn, line 130: WDYT about excluding this from guidance? This is more of a personal choice. Personally, I'd be happy to run both of these on karpenter nodes, and I don't see a great reason to recommend against it.

website/content/en/preview/getting-started/_index.md
```diff
@@ -13,3 +13,211 @@ Learn more about Karpenter and how to get started below.
...
+To get started, create a new IAM role for the Karpenter controller.
+
+The controller will be using [IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) which requires an OIDC endpoint.
+
+If you have another option for using IAM credentials with workloads (e.g. [kube2iam](https://github.com/jtblin/kube2iam)) your steps will be different.
+
+Set a variable for your cluster name.
+
+```bash
+CLUSTER_NAME=<your cluster name>
```
njtran, line 59: nit:
```suggestion
export CLUSTER_NAME=<your cluster name>
```

website/content/en/scripts/step05-controller-iam.sh
```diff
@@ -0,0 +1,66 @@
+echo "{
```
njtran, line 1: Can we refer to an already created file rather than an in-line bash blurb? I can imagine someone might forget to update this file (and others) or may mess up the formatting, where it should be simple to check by eye in a separate file. 

For reference, I mean something similar to 
https://github.com/aws/karpenter/blob/main/website/content/en/v0.22.0/getting-started/getting-started-with-eksctl/cloudformation.yaml. 

Let me know if I'm missing something here.

website/content/en/preview/getting-started/_index.md
```diff
@@ -13,3 +13,211 @@ Learn more about Karpenter and how to get started below.
...
+
+### Add tags to subnets and security groups
+
+Next add tags to your nodegroup subnets so Karpenter will know which subnets to use.
+
+{{% script file="./scripts/step06-tag-subnets.sh" language="bash" %}}
+
+Add tags to the security groups.
+This command only tags the security groups for the first nodegroup in the cluster.
+If you have multiple nodegroups or multiple security groups you will need to decide which one Karpenter should use.
```
njtran, line 78: The wording seems misleading here to me. From reading this, I would assume that Karpenter isn't going to be a groupless node autoscaler. Maybe we can do 
```suggestion
This command only tags the security groups used by the first nodegroup in the cluster.
If you have multiple nodegroups or multiple security groups you will need to decide which one Karpenter should use.
```

website/content/en/scripts/step10-deploy.sh
njtran: Seems like the previous step creates a helm template then applies it with kubectl. That sounds like we need this line.

website/content/en/preview/getting-started/_index.md
njtran: IIUC, this is from the CAS migration guide so that the user can install Karpenter on a nodegroup that the user isn't planning on deleting.

website/content/en/preview/getting-started/_index.md
njtran: ^ Agree, I don't think we'll need this just for the getting-started guide, since it's more of a "do you want more availability on your workloads"? Which is more of a Kubernetes good idea rather than something integral to the installation.

website/content/en/scripts/step11-create-provisioner.sh
```diff
@@ -0,0 +1,26 @@
...
+    - key: karpenter.k8s.aws/instance-category
+      operator: In
+      values: [c, m, r]
+    - key: karpenter.k8s.aws/instance-generation
+      operator: Gt
+      values: ["2"]
+  providerRef:
```
njtran, line 14: the main guide says that after scaling down with step14-deprovisioning says that emptiness will kick in but it's not here. 
```suggestion
  ttlSecondsAfterEmpty: 30
  providerRef:
```

website/content/en/preview/getting-started/_index.md
njtran: I think the other commands work, but I still see the eksctl command there. Can we remove it from the bottom of the file?

njtran (commented):
Nice work! Some requests.

website/content/en/scripts/step11-create-provisioner.sh
chrisnegus: Okay. Here's what I added (I got the example from another part of our docs):
`  instanceProfile: MyInstanceProfile # optional, if already set in controller args`

website/content/en/preview/getting-started/_index.md
chrisnegus: Yes, this was taken from the CAS migration guide. Should I change it?

website/content/en/scripts/step10-deploy.sh
chrisnegus: Should I remove the line or not?

website/content/en/scripts/step09-generate-chart.sh
chrisnegus: Okay. Is this right then?
`helm upgrade --install karpenter oci://public.ecr.aws/karpenter/karpenter --version ${KARPENTER_VERSION} --namespace karpenter --create-namespace \`

website/content/en/scripts/step10-deploy.sh
chrisnegus: I removed this line and replaced and added --create-namespace to step01. Let me know if I need to do more than that.

website/content/en/scripts/step11-create-provisioner.sh
chrisnegus: Done.

website/content/en/scripts/step11-create-provisioner.sh
chrisnegus: Okay. I added that line.

website/content/en/preview/getting-started/_index.md
chrisnegus: Done.

chrisnegus:
I'm closing this PR, which was seeking to create a "bring-your-own-cluster" approach to Karpenter Getting Started. Instead, I opened [PR #3312](https://github.com/aws/karpenter/pull/3312), which has the less lofty goal of removing two getting started guides (replacing them with external links) and simplifying the existing eksctl getting started guide. I'll revisit the "bring-your-own-cluster" idea in a later PR.

PR #1234, opened 2022-01-28
fix snapshot docs versions

bwagner5:
**1. Issue, if available:**
N/A

**2. Description of changes:**
- Snapshot docs versions pinned at a particular version were referencing the `latest_release` param. This change backfills the correct versions at each snapshot. [Here's an example of where the docs are wrong](https://karpenter.sh/v0.5.5/getting-started/#:~:text=serviceAccount.create%3Dfalse-,%2D%2Dversion%20v0.5.6%20%5C,-%2D%2Dset%20controller.clusterName) 
- I've also added a command to the `make website` target to replace the `latest_release` param with the version being released.  (used perl since that's more vendor neutral than sed, since mac's default sed is not gnu-sed)

**3. How was this change tested?**
`make website` resulted in the correct snapshot with a hardcoded version.

**4. Does this change impact docs?**
- [x] Yes, PR includes docs updates
- [ ] Yes, issue opened: *link to issue*
- [ ] No

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3054, opened 2022-12-16
test: add parameter to task to allow fetching things from origin

spring1843:
Fixes # 

**Description**

Will allow us to do things like running tests on PRs

* Local copy

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


test/suites/common/run-test.yaml
```diff
@@ -25,8 +29,16 @@ spec:
     workingDir: $(workspaces.ws.path)
     script: |
       echo "Cloning $(params.git-repo-url)"
+      echo "Fetching - Optional: $(params.git-fetch-origin)"
       echo "Checking out git ref: $(params.git-ref)"
-      git clone $(params.git-repo-url).git .
+
+      git clone $(params.git-repo-url).git
+
+      if [ "$(params.git-fetch-origin)" != "" ]; then
```
bwagner5, line 37: I don't think we need this parameter. Why not just fetch all the time and build the fetch command from the git-ref param?  

You could also just expect the `pull/ID/head:#1234` to be the input to git-ref if you want a PR and then `git fetch ${params.git-ref}` all the time.

test/suites/common/run-test.yaml
```diff
@@ -26,8 +26,11 @@ spec:
     script: |
       echo "Cloning $(params.git-repo-url)"
       echo "Checking out git ref: $(params.git-ref)"
-      git clone $(params.git-repo-url).git .
-      git checkout $(params.git-ref)
+
+      git clone $(params.git-repo-url).git
+
+      git fetch origin ${params.git-ref}
+      git checkout $(echo ${params.git-ref} | cut -d":" -f2)
```
bwagner5, line 33:
```suggestion
      git fetch origin $(params.git-ref)
      git checkout $(echo $(params.git-ref) | cut -d":" -f2)
```

test/suites/common/run-test.yaml
spring1843: done good idea

bwagner5 (approved):
LGTM

PR #3068, opened 2022-12-19
test: Bump ginkgo timeout so that ginkgo doesn't panic

jonathan-innis:
Fixes # 

**Description**

- Bump ginkgo timeout so that it doesn't panic on 60m default timeout

* `make e2etests`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ battletest: ## Run randomized, racing, code coveraged, tests
 		-tags random_test_delay
 
 e2etests: ## Run the e2e suite against your local cluster
-	cd test && CLUSTER_NAME=${CLUSTER_NAME} go test -p 1 -count 1 -timeout 180m -v ./suites/... --ginkgo.focus="${FOCUS}"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -69,7 +69,7 @@ battletest: ## Run randomized, racing, code coveraged, tests
 		-tags random_test_delay
 
 e2etests: ## Run the e2e suite against your local cluster
-	cd test && CLUSTER_NAME=${CLUSTER_NAME} go test -p 1 -count 1 -timeout 180m -v ./suites/... --ginkgo.focus="${FOCUS}"
+	cd test && CLUSTER_NAME=${CLUSTER_NAME} go test -p 1 -count 1 -timeout 180m -v ./suites/... --ginkgo.timeout=180m --ginkgo.focus="${FOCUS}"
```
ellistarn, line 72: Do we need both ginkgo timeout and go test timeout?

jonathan-innis: I think the answer is yes since the "go test" runs in a different context from the "ginkgo" and they have different timeouts. I think I'm thinking of favoring just calling `ginkgo` instead of `go test` in all places we do this so we avoid having weird CLI args that apply to different contexts.

PR #3069, opened 2022-12-19
Revert "docs: Retain current path in new version docs link"

spring1843:
Reverts aws/karpenter#3013

Avoid DOM-based cross site scripting risk


spring1843:
@rothgar PTAL, maybe we can find a way to do this statically?

ellistarn (approved):

PR #3104, opened 2022-12-23
docs: Update README to account for other regions

dennisotugo:
Hello, I updated the README default region to somewhere other than us-east-1, which allows us to create a secondary provider pointing to the us-east-1 where the ECR registry is stored. Covering varied use-cases

Fixes # 

**Description**

*

- [X ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/website/content/en/v0.20.0/getting-started/getting-started-with-terraform/_index.md
+++ b/website/content/en/v0.20.0/getting-started/getting-started-with-terraform/_index.md
@@ -46,7 +46,7 @@ After setting up the tools, set the following environment variables to store
 commonly used values.
 
 ```bash
-export AWS_DEFAULT_REGION="us-east-1"
+export AWS_DEFAULT_REGION="eu-west-1"
 ```
 
 The first thing we need to do is create our `main.tf` file and place the following in it.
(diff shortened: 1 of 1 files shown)


FernandoMiguel (commented):
lgtm

website/content/en/v0.20.0/getting-started/getting-started-with-terraform/_index.md
```diff
@@ -71,8 +71,13 @@ terraform {
   }
 }
 
+provider "aws" {
+  region = "eu-west-1"
+}
+
 provider "aws" {
   region = "us-east-1"
+  alias = "virginia"
```
FernandoMiguel, line 80: pls fmt

ellistarn:
The getting started guide is just a getting started guide. We expect folks to modify it to meet their needs. The more complicated this guide is, the harder it is to test, and the more likely it is to break. Changing the region is something we expect users to do when customizing the guide to meet their use case.

ellistarn (changes requested):
The getting started guide is just a getting started guide. We expect users to customize it to meet their use cases. The more branching logic we include in this guide, the harder it is to test, and the less likely it is to be correct.

jonathan-innis:
Closing since we are consolidating these docs as part of #3017

PR #3123, opened 2022-12-29
chore: add v0.21.1 docs

njtran:
Fixes # 

**Description**
Adds v0.21.1 docs 

* make website

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


website/config.yaml
```diff
@@ -66,9 +66,9 @@ params:
         url: 'https://slack.k8s.io/'
         icon: fab fa-slack
         desc: 'Chat with us on Slack in the #aws-provider channel'
-  latest_release_version: v0.21.0
+  latest_release_version: v0.21.1
```
ellistarn, line 69: Is it enough to just update this file, and the docs will update appropriately?

ellistarn (approved):
I think it makes sense to merge as is, but we should really move to minor version docs.

website/config.yaml
njtran: Are you talking about if we just change this file but not add the directory? Unfortunately not.

PR #3125, opened 2022-12-29
chore: Migrate to `v1alpha5` and cleanup some provider cross-pollination

jonathan-innis:
Fixes # 

**Description**

- Abstract provider calls into interface to enable the underlying implementation to be swapped out for use-cases like for testing. This also makes the public API smaller and the interface cleaner

* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"strings"
 
 	"github.com/aws/karpenter/pkg/apis"
 	"github.com/aws/karpenter/pkg/apis/v1alpha1"
@@ -47,7 +48,6 @@ import (
 	awscontext "github.com/aws/karpenter/pkg/context"
 
 	coreapis "github.com/aws/karpenter-core/pkg/apis"
-	corev1alpha1 "github.com/aws/karpenter-core/pkg/apis/v1alpha1"
 	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
 	"github.com/aws/karpenter-core/pkg/cloudprovider"
 )
@@ -91,7 +91,13 @@ func New(ctx awscontext.Context) *CloudProvider {
 		kubeClient:           ctx.KubeClient,
 		instanceTypeProvider: instanceTypeProvider,
 		amiProvider:          amiProvider,
-		instanceProvider: NewInstanceProvider(ctx, ec2api, instanceTypeProvider, subnetProvider,
+		instanceProvider: NewInstanceProvider(
+			ctx,
+			aws.StringValue(ctx.Session.Config.Region),
+			ec2api,
+			ctx.UnavailableOfferingsCache,
+			instanceTypeProvider,
+			subnetProvider,
 			NewLaunchTemplateProvider(
 				ctx,
 				ec2api,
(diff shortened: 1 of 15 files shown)


pkg/cloudprovider/instance.go
```diff
@@ -52,17 +51,28 @@ var (
 	instanceTypeFlexibilityThreshold = 5 // falling back to on-demand without flexibility risks insufficient capacity errors
 )
 
-type InstanceProvider struct {
+// InstanceProvider enables creating, getting and deleting EC2 instances based on Karpenter specs
+type InstanceProvider interface {
```
bwagner5, line 55: What is this enabling us to do? It seems like this is just abstraction without a compelling reason. There may be one that I'm not seeing. I think think this helps much in testing since we'd either be mocking out the whole cloud provider interface or we'd be testing the providers themselves.

pkg/cloudprovider/instance.go
```diff
@@ -52,17 +51,28 @@ var (
...
+	Create(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate, machine *v1alpha5.Machine, instanceTypes []*cloudprovider.InstanceType) (*ec2.Instance, error)
+	Get(ctx context.Context, id string) (*ec2.Instance, error)
+	Delete(ctx context.Context, machine *v1alpha5.Machine) error
+}
+
+type instanceProviderImpl struct {
```
bwagner5, line 61: I'm not a fan of the `Impl` suffix, but I'm struggling to come up with something better since I generally prefer the `er` suffix on the interface and then without it on the implementation, but that doesn't really work here...

bwagner5 (commented):
I'm not sure if this would enable much for us. Can you implement what you were thinking in the testing so I could see how this is better than what we have now?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -145,31 +155,28 @@ func (c *CloudProvider) LivenessProbe(req *http.Request) error {
 func (c *CloudProvider) GetInstanceTypes(ctx context.Context, provisioner *v1alpha5.Provisioner) ([]*cloudprovider.InstanceType, error) {
 	var nodeTemplate *v1alpha1.AWSNodeTemplate
 	var err error
```
bwagner5, line 157: Should be able to bring not define these here and just use the shorthand assignment below now.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -145,31 +155,28 @@ func (c *CloudProvider) LivenessProbe(req *http.Request) error {
...
-			Name:       provisioner.Spec.ProviderRef.Name,
-		})
-	} else {
-		nodeTemplate, err = c.resolveNodeTemplate(ctx, provisioner.Spec.Provider.Raw, nil)
+
+	var rawProvider []byte
+	if provisioner.Spec.Provider != nil {
+		rawProvider = provisioner.Spec.Provider.Raw
 	}
+	nodeTemplate, err = c.resolveNodeTemplate(ctx, rawProvider, provisioner.Spec.ProviderRef)
```
bwagner5, line 163: resolveNodeTemplate could probably just take the provisioner spec and deal with the `Raw` being nil since you have special logic in this func and below `nil` is passed directly to raw which is awkward.

bwagner5 (commented):
just a few minor comments, otherwise LGTM!

pkg/cloudprovider/cloudprovider.go
jonathan-innis: Good eye

pkg/cloudprovider/cloudprovider.go
jonathan-innis: I'd agree if Machine wasn't also using this function. The parameters need to be abstract enough to work for both

bwagner5: oooo right 👍

bwagner5 (approved):
LGTM

PR #3128, opened 2022-12-30
chore: Hydrate `v1alpha5/Machine` CRs from Karpenter-owned Nodes

jonathan-innis:
Fixes # 

**Description**

NOTE: This controller is not initialized in `pkg/controllers/controllers.go` so it will not run with the binary. The `machinehydration` controller is added for staging out the changes for migrating instance ownership in EC2 from the `Node` resource to the `v1alpha5/Machine` CR

* `make presubmit`
* Scale tested with 500 nodes and saw 500 machines propagated off of them on start

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```


pkg/cache/unavailableofferings.go
```diff
@@ -26,7 +26,8 @@ import (
 )
 
 const (
-	UnavailableOfferingsTTL = 3 * time.Minute
+	UnavailableOfferingsTTL             = 3 * time.Minute
+	UnavailableOfferingsCleanupInterval = 10 * time.Minute
```
ellistarn, line 30: In other places we've used the same TTL for both. Any thoughts on this?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -193,6 +177,36 @@ func (c *CloudProvider) IsMachineDrifted(ctx context.Context, machine *v1alpha5.
...
+// has the corresponding tag value for that machine
+func (c *CloudProvider) Hydrate(ctx context.Context, machine *v1alpha5.Machine) error {
+	instanceID, err := utils.ParseInstanceID(machine.Status.ProviderID)
+	if err != nil {
+		return fmt.Errorf("parsing instance id, %w", err)
+	}
+	instance, err := c.instanceProvider.Get(ctx, instanceID)
+
+	if err != nil {
+		if awserrors.IsInstanceTerminated(err) {
```
ellistarn, line 190: What if awserrs.IsNotFound(err) == true ?

pkg/cache/unavailableofferings.go
jonathan-innis: I'm not sure what you mean. You're asking if we should use a minute for unavailable offerings in this case. Or are you asking if the cleanup interval should be the same as the TTL? I'm not sure I fully understand the effects/scope of reducing the cleanup interval for expired entries. I'd imagine it's just a memory/cpu burden tradeoff and more of a heuristic in general?

pkg/controllers/machinehydration/namegenerator.go
```diff
@@ -0,0 +1,40 @@
...
+
+var MachineNameGenerator machineNameGenerator
+
+var (
+	maxNameLength          = 63
+	randomLength           = 12
+	MaxGeneratedNameLength = maxNameLength - randomLength
+)
+
+func (machineNameGenerator) GenerateName(base string) string {
```
ellistarn, line 35: Not seeing the reason to make a struct here. Why not just a function?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -236,6 +219,37 @@ func (c *CloudProvider) IsMachineDrifted(ctx context.Context, machine *v1alpha5.
...
+	}
+	instance, err := c.instanceProvider.GetByID(ctx, instanceID)
+	if err != nil {
+		return fmt.Errorf("getting instance for instanceID '%s', %w", instanceID, err)
+	}
+	// Update the machine with the name stored in the tag value of the instance if it exists
+	// We do this so that when we create the Machine after hydration, the machine-controller is aware
+	// that the instance already exists, so it doesn't create another one
+	if tag, ok := lo.Find(instance.Tags, func(tag *ec2.Tag) bool {
+		return aws.StringValue(tag.Key) == v1alpha5.MachineNameLabelKey
```
ellistarn, line 237: brevity
```suggestion
		return lo.FromPtr(tag.Key) == v1alpha5.MachineNameLabelKey
```

pkg/cloudprovider/instance.go
```diff
@@ -367,6 +368,48 @@ func (p *InstanceProvider) getOverrides(instanceTypes []*cloudprovider.InstanceT
 	return overrides
 }
 
+func (p *InstanceProvider) Update(ctx context.Context, machine *v1alpha5.Machine) (*ec2.Instance, error) {
```
ellistarn, line 371: Consider adding a comment that you can remove this in v1beta1.

pkg/controllers/interruption/suite_test.go
```diff
@@ -102,10 +99,11 @@ var _ = BeforeEach(func() {
 		}),
 	}
 	ctx = settingsStore.InjectSettings(ctx)
+	unavailableOfferingsCache.Flush()
```
ellistarn, line 102: Looking at lines 93/94, instantiating controllers on each test loop seems expensive.

pkg/controllers/machinehydration/controller.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	logging.FromContext(ctx).Debugf("hydrated machine from node")
+	return nil
+}
+
+func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontroller.Builder {
+	return corecontroller.Adapt(controllerruntime.
+		NewControllerManagedBy(m).
+		For(&v1.Node{}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10}))
```
ellistarn, line 113: consider importing at ctrl, so the other one can take the more natural name controller

pkg/controllers/machinehydration/controller.go
```diff
@@ -0,0 +1,125 @@
...
+		if corecloudprovider.IsMachineNotFoundError(err) {
+			return nil
+		}
+		return fmt.Errorf("hydrating machine, %w", err)
+	}
+	if err := c.kubeClient.Create(ctx, machine); err != nil {
+		if errors.IsAlreadyExists(err) {
+			return nil
+		}
+		return err
```
ellistarn, line 103: I think there's `return IgnoreAlreadyExists(err)`

pkg/fake/cloudprovider.go
```diff
@@ -72,6 +71,10 @@ func (c *CloudProvider) Get(context.Context, string, string) (*v1alpha5.Machine,
 	return nil, nil
 }
 
+func (c *CloudProvider) HydrateMachine(_ context.Context, _ *v1alpha5.Machine) error {
+	return nil
+}
```
ellistarn, line 76: Can't we make this part of the AWS API, but exclude it from other providers?

jonathan-innis: That's what the current state is. We have this only in the AWS API but not a part of the generic interface.

pkg/cloudprovider/cloudprovider.go
jonathan-innis: Interesting point of note here that @tzneal pointed out to me. Since the `lo.FromPtr` function is a generic function, it's an order of magnitude slower than the non-generic verison.

pkg/fake/cloudprovider.go
ellistarn: Why include it in the fake, then?

jonathan-innis: Good point, that's an artifact of an older commit. I'll remove.

pkg/cloudprovider/cloudprovider.go
ellistarn: We do a lot of filtering in our tight scheduling loop. Maybe worth specifying?

crds/karpenter.k8s.aws_awsnodetemplates.yaml
```diff
@@ -0,0 +1,230 @@
+---
```
ellistarn, line 1: Why is this here? Needs rebase?

ellistarn (approved):

ellistarn (approved):

PR #3143, opened 2023-01-03
feat: Changes to make logging easier with karpenter

endriu0:
**Description**
Currently if karpenter is running in fargate the only way to use 3rd party logging would be to send all the logs via fargate logging on the cluster to firehose and then to actual app.

With this PR it's now possible to override zap logger to send logs to stdout and also to a file. With addition of sideCar config as well we can now run logging sidecar which shares volumes with karpenter container and use e.g. fluentbit sidecar to forward the logs to our monitoring tool of choice.

Tested POC of this in our environment / tested that helm chart doesn't see any changes with no changes to values

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


charts/karpenter/templates/_helpers.tpl
```diff
@@ -92,4 +92,26 @@ Flatten Values Map using "." syntax
   {{- end -}}
 {{- end -}}
 {{- end -}}
+{{- end -}}
+
+{{/*
+Flatten the stdout logging outputs from args provided
+*/}}
+{{- define "outputPathsList" -}}
```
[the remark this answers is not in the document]
endriu0: Sure - makes sense.

charts/karpenter/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: karpenter
 description: A Helm chart for Karpenter, an open-source node provisioning project built for Kubernetes.
 type: application
-version: 0.21.1
+version: 0.21.2
```
ellistarn, line 5: Can you revert these changes? It's circumventing our normal release process.

ellistarn (changes requested):
lgtm -- one blocking comment

charts/karpenter/Chart.yaml
endriu0: Done now, sorry wasn't completely clear on the release process

ellistarn (approved):
Great changes. Thanks!

spring1843 (approved):
/karpenter snapshot

charts/karpenter/templates/deployment.yaml
```diff
@@ -113,6 +113,12 @@ spec:
           volumeMounts:
             {{- toYaml . | nindent 12 }}
           {{- end }}
+        {{- with .Values.controller.sidecarContainer }}
+          {{- with .Values.controller.extraVolumeMounts }}
+          volumeMounts:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+        {{- end }}
```
Flydiverny, line 121: How is this supposed to be used? `sidecarContainer` value will never be injected here, this seems like some copy-paste gone wrong here
```
❯ helm template . --set controller.sidecarContainer.name="hello world" --debug

Error: template: karpenter/templates/deployment.yaml:117:26: executing "karpenter/templates/deployment.yaml" at <.Values.controller.extraVolumeMounts>: nil pointer evaluating interface {}.controller
helm.go:84: [debug] template: karpenter/templates/deployment.yaml:117:26: executing "karpenter/templates/deployment.yaml" at <.Values.controller.extraVolumeMounts>: nil pointer evaluating interface {}.controller
```

PR #3161, opened 2023-01-05
docs: Add details on permissions required for interruption handling

jonathan-innis:
Fixes # 

**Description**

- Add details on permissions required for interruption handling

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```

### DIFF
--- a/website/content/en/preview/concepts/deprovisioning.md
+++ b/website/content/en/preview/concepts/deprovisioning.md
@@ -129,6 +129,15 @@ data:
   ...
 ```
 
+#### Permissions
+
+Karpenter requires the following IAM permissions to be assigned to the Karpenter controller IAM role to enable interruption handling on the cluster. These permissions should only be assigned access the SQS queue resource that you have provisioned for Karpenter interruption event handling. 
+
+- sqs:GetQueueAttributes
+- sqs:GetQueueUrl
+- sqs:ReceiveMessage
+- sqs:DeleteMessage
+
 ## Drift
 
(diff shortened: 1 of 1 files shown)


website/content/en/preview/concepts/deprovisioning.md
```diff
@@ -129,6 +129,15 @@ data:
   ...
 ```
 
+#### Permissions
```
ellistarn, line 132: I wonder if this is the right place to document this. Maybe we need a separate page where we document all of the permissions that karpenter needs?

jonathan-innis: Might make sense. Have a low-level detailing of permissions that is stored in a `Reference` section in the docs. Since this was a specific feature that required a specific set of additional permissions, I was thinking it might make sense to include it here in the short-term.

jonathan-innis: What are your thoughts on including this in this section of the docs in the short-term and then moving this over to a separate Reference section in the docs that includes all the details of which features need which permissions?

jonathan-innis: I can add a task to create this reference section.

ellistarn: preference for users to discover this in another way, rather than teaching them to discover the permissions here.

jonathan-innis: Not sure if I understand what you mean,, "discover this in another way"? Are you referring to a permissions reference section? If it's not documented, it's not clear to me how the users are going to see a correlation between the permissions they need to set in IAM and the features that are backing these permissions.

PR #3169, opened 2023-01-06
test: trigger tekton pipelinerun after each release (snapshot, stable, from a PR)

spring1843:
Fixes # 

**Description**

* Add release-notification-listener that listens for notifications and reacts to them by running tekton tasks
* Reorganizes the tools directory to contain multiple tools that we have

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


test/cmd/release-notification-listener/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+# Build release-notification-listener
+RUN mkdir /release-notification-listener
+COPY go.mod /release-notification-listener
+COPY go.sum /release-notification-listener
+COPY main.go /release-notification-listener
+COPY listener /release-notification-listener/listener
+
+WORKDIR /release-notification-listener
+
+RUN GOPROXY=direct go build .
```
ellistarn, line 56: Do we need all this for the tekton client to work?

spring1843: Basically yes.

ellistarn: We could base it on the tools image, as another option.

spring1843: I did think of that, that image is larger and has a clone of the Karpenter code base and all dependencies in it (1.6GB) and we don't need all of that here. This image is ~400MB

spring1843: This and that docker image could both be from a base image, or have scripts instead of commands to help with not having duplication. Can do that as a follow up.

test/infrastructure/clusters/test-infra/release-notification-listener/release-notification-listener.yaml
```diff
@@ -0,0 +1,73 @@
...
+  selector:
+    matchLabels:
+      app: release-notification-listener
+  template:
+    metadata:
+      labels:
+        app: release-notification-listener
+    spec:
+      serviceAccountName: release-notification-listener
+      automountServiceAccountToken: true
```
ellistarn, line 61: this is the default, if you wanted to skip setting this.

test/cmd/release-notification-listener/listener/tekton.go
```diff
@@ -0,0 +1,118 @@
...
+			"Chaos",
+			"Drift",
+		},
+		"ipv6": []string{},
+	}
+)
+
+func runTests(message *notificationMessage, args ...string) error {
+	log.Printf("running tests with args %v for release type: %s, release identifier: %s ", args, message.ReleaseType, message.ReleaseIdentifier)
+	cmd := exec.Command(tektonCLICommandPath, args...)
```
ellistarn, line 46: It strikes me that using the kubernetes go client and the tekton apis is probably easier/cleaner. Curious why you went this route -- if you're not familiar with k8s client-go, it's worth the investment to learn it.  Happy to merge as is.

test/cmd/release-notification-listener/Dockerfile
ellistarn: if you didn't need the aws/tkn clis, you could use ko to build this, which would be much faster to build/push/pull.

ellistarn (approved):

test/cmd/release-notification-listener/listener/tekton.go
spring1843: I did attempt that there are some obstacles. Should be able to replace these parts when it's ready.

PR #3190, opened 2023-01-10
feat: Added additional topology spread constraint

wpbeckwith:
Fixes # 

**Description**

While a production cluster should definitely be multi-zonal, some dev/test clusters may only be in a single zone.  Regardless, we don't want the multiple karpenter pods on a single host if possible.

Deployed to multiple clusters with one or multiple zones and verified karpenter pods are on different hosts

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Added additional topology spread constraint for the kubernetes.io/hostname topologyKey
```


charts/karpenter/values.yaml
```diff
@@ -75,6 +75,9 @@ topologySpreadConstraints:
   - maxSkew: 1
     topologyKey: topology.kubernetes.io/zone
     whenUnsatisfiable: ScheduleAnyway
+  - maxSkew: 1
```
ellistarn, line 78: Can we use hostname antiaffinity instead? If there's 1 node, the scheduler will spread all karpenter pod replicas across all of the nodes, or in this case, just a single node.

wpbeckwith: Sure, we'll update the PR.

jonathan-innis:
@wpbeckwith Ping for the requested changes. Otherwise, PR looks good!

jonathan-innis:
@runningman84 This PR has gone stale. If you want to open a PR that adds preferred hostname anti-affinity, we would be happy to accept it.

wpbeckwith:
Sorry been busy with a cloud migration.  We'll get this submitted this weekend.

charts/karpenter/values.yaml
sftim: We might want to let users choose beween `requiredDuringSchedulingIgnoredDuringExecution` and `preferredDuringSchedulingIgnoredDuringExecution`, as different cluster admins might have different preferences.

For example, some clusters bootstrap with one initial node and the admin might want to see their cluster-critical Pods go healthy pronto. Meanwhile, they have a separate mechanism that warns about anti-affinity issues, or even they have their own custom descheduler to take care of that.

charts/karpenter/values.yaml
```diff
@@ -72,14 +72,20 @@ affinity:
...
             - key: karpenter.sh/provisioner-name
               operator: DoesNotExist
+  podAntiAffinity:
+    preferredDuringSchedulingIgnoredDuringExecution:
+      - labelSelector:
+          matchExpressions:
+            - key: app.kubernetes.io/name
+              operator: In
+              values:
+              - karpenter
```
sftim, line 82: Does this always match with: `app.kubernetes.io/name: {{ include "karpenter.name" . }}`

If I customise the name, I'd expect that to change the affinity rules automatically - at least by default.

charts/karpenter/values.yaml
wpbeckwith: I was looking to stick with the current defaults where karpenter will still be scheduled on the same node if that's all available.  I think a default rule to naturally spread out is good though.  If users want to take a harder line, would overriding the values be out of order?

charts/karpenter/values.yaml
wpbeckwith: From running some tests this looks to be fine.  If I do a 
```
helm template kpop ./karpenter
...
  template:
    metadata:
      labels:
        app.kubernetes.io/name: karpenter
        app.kubernetes.io/instance: kpop
...
```
The above are the labels for the karpenter deployment.  The app name is constant at karpenter while the instance name will change to whatever the release name is.

sftim: What happens if you set `nameOverride` explicitly?

wpbeckwith: Ok, in that instance it will change the name.  I'll see what can be done here.  Helm-fu is not my normal specialty.

wpbeckwith: This was way harder than I would imagine to get the helm to do what is required.  Regardless, I learned something new today.

jonathan-innis (approved):
LGTM 🚀 Thanks for contribution! 🎉

runningman84 (approved):

PR #2978, opened 2022-12-02
fix: remove name filter

tzneal:
Fixes # 

**Description**

**How was this change tested?**

*

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3204, opened 2023-01-12
feat: Provide ability to specify owners when selecting an AMI

### LINKED ISSUE #3022: Can no longer use amiSelector to filter on name
### Version

__Karpenter Version:__ v0.19.3 / v0.20.0

__Kubernetes Version:__ v1.22

### Expected Behavior

Use `name: AMI_NAME` to be able to select AMI. We centrally create our AMIs and share them out cross account, where the tags do not propagate. We were able to use this functionally in 0.19.2 but it broke in 0.19.3 after being removed.

### Actual Behavior

This functionality was removed in 0.19.3: https://github.com/aws/karpenter/pull/2978/files. I'd expect to be able to use the ami name to search.

### Steps to Reproduce the Problem

use
```
amiSelector:
  name: ....
```

### Resource Specs and Logs

NA

### Community Note

willthames:
Fixes #3022 

**Description**

This allows `owners` to be passed when requesting an AMI.

Default is `self,amazon` - anything else replaces that list.

Created new test suite for AMI selectors

It will impact docs, and this PR will be updated to contain docs if the idea is acceptable

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Add `aws::name` as a special-case field in AMISelector to look up AMI by name (this is different to passing `Name` which uses the `Name` tag that isn't shared across accounts)

Add `aws::owners` as a special-case field in AMISelector. Note that, when `aws::ids`/`aws-ids` are not set, this value defaults to `self,amazon`. Setting `aws::owners` will override (rather than extend) this list.
```


njtran:
Hey @willthames, thanks for the PR. Is it possible for you to write a design doc for this? Although this is a fairly simple implementation, it would be good to consider what the alternatives are or what this may affect. 
Here is the design guide. https://karpenter.sh/v0.22.1/contributing/design-guide/

A member of the Karpenter team should be able to help guide you through the design process.

Some things off the top of my head are: 
1) using "owners" will then make this a magic string, where other uses who use "owners" as a tag will be unable. Maybe we can circumvent this by taking this out of the amiSelector and making it a field similar to amiFamily?
2) It's important to note that AMI Tags are private, so if you're trying to use any tags on the AMI, those won't be shared for AMI usage unless CI/CD systems sync up.

dudicoco:
Hi @njtran, can you please elaborate regarding the decision making process on removal/addition of features within the karpenter project?

There was a [breaking change introduced](https://github.com/aws/karpenter/pull/2978) without specifying any context or background and it was merged with no problem, yet the fix for this breaking change requires a design doc even though the context was specified within #3022, with puts an unneeded strain on contributors.

I've seen at least one other such breaking change that has zero context on why it was introduced: https://github.com/aws/karpenter/pull/2746.

In an open source repository there should be more transparency on PRs, rather than perhaps talking about them behind the scenes in slack or another messaging system as I have no other explanation on how the PRs in question were merged with zero context.

On a side note, it seems that there are many breaking changes constantly being introduced to karpenter and i'm not sure if the users who have karpenter running in production are taken into consideration, after all no one wants to deal with constant migrations.

ellistarn:
> There was a https://github.com/aws/karpenter/pull/2978 without specifying any context or background and it was merged with no problem, yet the fix for this breaking change requires a design doc even though the context was specified within https://github.com/aws/karpenter/issues/3022, with puts an unneeded strain on contributors.

Hey @dudicoco, we acted swiftly on #2978 as it was an undocumented, unintended insecure-by-default feature.

@njtran's request for a 1 pager is reasonable, as this feature has its complicated history. That said, if that feels like a big burden to you, I think we have reasonably clarity in the proposed direction such that it may not require a doc.

> In an open source repository there should be more transparency on PRs

💯 agree. We're taking steps to be more public about our development. You can see the public development slack channel we just created here: https://kubernetes.slack.com/archives/C04JW2J5J5P. 

> after all no one wants to deal with constant migrations.

Completely agree -- the project is quickly maturing towards 1.0, which will change our policy on breaking changes. As it stands, we do our best to minimize breaking changes, but do make some in the best long term interest of the project. It's essential for a project to be able to be refined to benefit the user experience in the long term -- of course, there's always a turning point where the disruption of existing users outweighs the future benefit. We're fast approaching that point w/ 1.0.

willthames:
@njtran @ellistarn design document is now attached to this PR (probably the easiest way to provide feedback).

I'm willing to do the work to make the v1alpha2 change (that is my preferred outcome, this PR was an easy stake in the ground and generated the discussion I wanted to have anyway)

dudicoco:
> > There was a #2978 without specifying any context or background and it was merged with no problem, yet the fix for this breaking change requires a design doc even though the context was specified within #3022, with puts an unneeded strain on contributors.
> 
> Hey @dudicoco, we acted swiftly on #2978 as it was an undocumented, unintended insecure-by-default feature.
> 
> @njtran's request for a 1 pager is reasonable, as this feature has its complicated history. That said, if that feels like a big burden to you, I think we have reasonably clarity in the proposed direction such that it may not require a doc.
> 
> > In an open source repository there should be more transparency on PRs
> 
> 💯 agree. We're taking steps to be more public about our development. You can see the public development slack channel we just created here: https://kubernetes.slack.com/archives/C04JW2J5J5P.
> 
> > after all no one wants to deal with constant migrations.
> 
> Completely agree -- the project is quickly maturing towards 1.0, which will change our policy on breaking changes. As it stands, we do our best to minimize breaking changes, but do make some in the best long term interest of the project. It's essential for a project to be able to be refined to benefit the user experience in the long term -- of course, there's always a turning point where the disruption of existing users outweighs the future benefit. We're fast approaching that point w/ 1.0.

@ellistarn thanks for reply.

Hopefully all future PRs will have explanation and context within them so breaking changes can be discussed/addressed before the PRs are merged and it will be easier for everyone to understand what was the purpose of these PRs.

designs/ami-selector.md
```diff
@@ -0,0 +1,62 @@
...
+create a new special case of `owners` that would conflict with any existing use of `owners` as
+an AMI tag.
+
+Solution 3 provides a workaround to the `owners` tag conflict but seems unnecessary and inelegant
+just to provide a way to avoid a theoretical conflict.
+
+Solution 2 is the most work, but removes any of the special cases, and provides a much more flexible
+approach to AMI filtering in general - by allowing filters other than `tag:`, users will gain
+more power.
+
```
njtran, line 62: Writing down my pros and cons here for each solution. Unclear to me which one we should go for, or if theres is a better alternative. Can you come present this to Working Group?

Solution 1
- Pros: Easily solves the problem
- Cons: Can break users that have `owners` and `name` tags in their AWSNodeTemplate already. Adds in multi-case runtime defaulting that may be ambiguous to new users

Solution 2: 
- Pros: Scales to the future a bit better as any EC2 API user can easily put their configs in.
- Cons: Relies heavily on the structure of the EC2 API. Any diff in EC2 APIs in the future will be confusing. This gives less control to Karpenter for the future of how we want this API to look.

Solution 3: 
- Pros: Doesn't break any users
- Cons: Adds config bloat into the AWSNodeTemplate. The field would be misplaced.

njtran (commented):
Thanks for the Design doc. Appreciate you working on this. Would you be able to come to the Working Group tomorrow to discuss this design doc? This way we can get consensus from the team and get a quick round of reviews on this.

willthames:
@njtran since the working group is at an AU-friendly time tomorrow, I'm definitely intending to be there.

designs/ami-selector.md
```diff
@@ -0,0 +1,62 @@
...
+# Solutions
+
+1. Restore the `name` special case and add a new `owners` special case that is passed
+as the `Owners` argument to `DescribeImages` (this arguments supports both aliases
+such as `self` and `amazon` as well as AWS account IDs). Make `self,amazon` the default
+for this solution.
+
+2. Create a v1alpha2 AMISelector with a much more flexible type than go's `map[string]string` and
+deprecate v1alpha1. The new AMISelector could have an interface much closer to `DescribeImages`.
+
```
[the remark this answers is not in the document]
willthames: I believe it should be round-trippable as long as you make certain assumptions about the old configuration (the main being what to assume the owners are if they're not specifed)

designs/ami-selector.md
```diff
@@ -0,0 +1,62 @@
...
+While the `Name` tag is still usable, AMI tags are private to an account (so if you want
+to use the AWS bottlerocket AMIs, you'd need to tag them yourself in each account that
+you use)
+
+# Solutions
+
+1. Restore the `name` special case and add a new `owners` special case that is passed
+as the `Owners` argument to `DescribeImages` (this arguments supports both aliases
+such as `self` and `amazon` as well as AWS account IDs). Make `self,amazon` the default
+for this solution.
```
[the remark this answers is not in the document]
willthames: > Users that have multiple accounts must now add this to every AMISelector they have to include a list of Account IDs that they trust in order to enable their AMISelector to discover these.

This is true, but this should have always been true if you wish to avoid the security hole that specifying owners fixes. (Also you *can* copy AMIs into an account, you don't necessarily need to trust them - but that's the chore we're trying to avoid by allowing `name` to be specified again - you can just use `tag:Name`, if you first tag them in your own account)

willthames:
@njtran @ellistarn I've updated this PR to implement version 1 with `aws:` prefixed AMISelector parameters.

As discussed in the Working Group meeting, this should be considered an interim implementation until a more fully fleshed out NodeTemplate v1alpha2 design is complete.

If this meets your approval I'll update docs and tests.

ellistarn:
> @njtran @ellistarn I've updated this PR to implement version 1 with `aws:` prefixed AMISelector parameters.
> 
> As discussed in the Working Group meeting, this should be considered an interim implementation until a more fully fleshed out NodeTemplate v1alpha2 design is complete.
> 
> If this meets your approval I'll update docs and tests.

See: https://kubernetes.slack.com/archives/C02SFFZSA2K/p1674679860160339?thread_ts=1674676734.526229&cid=C02SFFZSA2K

designs/ami-selector.md
```diff
@@ -0,0 +1,62 @@
...
+by owner, AMI impersonation could happen with various risks (e.g. security vulnerabilities, malware
+such as mining, or cost implications).
+
+While the `Name` tag is still usable, AMI tags are private to an account (so if you want
+to use the AWS bottlerocket AMIs, you'd need to tag them yourself in each account that
+you use)
+
+# Solutions
+
+1. Restore the `name` special case and add a new `owners` special case that is passed
```
njtran, line 18: Hey, can you change this option and update the docs based on our discussion? Just some points for refresher:
    1. Change all special casing to be `aws::` prefixed.
    2. `aws-ids` and `aws::ids` will be supported to not break users.
    3. *Caveat:* System tags are not currently tagged on AMIs used with EKS Optimized AMIs. This could be added in the future. We’ll make it `aws::` instead of `aws:` so that we can still use the special case, and be flexible to a possible system tag addition in the future.
    4. Long term we need a design doc to implement given the feature request here: https://github.com/aws/karpenter/issues/3326 (Option 2)

pkg/cloudprovider/amifamily/ami.go
```diff
@@ -193,23 +193,41 @@ func (p *AMIProvider) fetchAMIsFromEC2(ctx context.Context, amiSelector map[stri
...
-		if key == "aws-ids" {
+		switch key {
+		case "aws-ids":
 			filterValues := functional.SplitCommaSeparatedString(value)
 			filters = append(filters, &ec2.Filter{
 				Name:   aws.String("image-id"),
 				Values: aws.StringSlice(filterValues),
 			})
-		} else {
+		case "aws:owners":
```
njtran, line 207:
```suggestion
		case "aws::owners":
```

pkg/cloudprovider/amifamily/ami.go
```diff
@@ -193,23 +193,41 @@ func (p *AMIProvider) fetchAMIsFromEC2(ctx context.Context, amiSelector map[stri
...
 }
 
-func getFilters(amiSelector map[string]string) []*ec2.Filter {
+func getFiltersAndOwners(amiSelector map[string]string) ([]*ec2.Filter, []*string) {
 	filters := []*ec2.Filter{}
+	owners := []*string{}
 	for key, value := range amiSelector {
-		if key == "aws-ids" {
+		switch key {
+		case "aws-ids":
```
njtran, line 201: Can you make it so that aws::ids and aws-ids are both special cased so that we don't break users?
```suggestion
		case "aws-ids", "aws::ids":
```

pkg/cloudprovider/amifamily/ami.go
```diff
@@ -193,23 +193,41 @@ func (p *AMIProvider) fetchAMIsFromEC2(ctx context.Context, amiSelector map[stri
...
 				Name:   aws.String("image-id"),
 				Values: aws.StringSlice(filterValues),
 			})
-		} else {
+		case "aws:owners":
+			ownerValues := functional.SplitCommaSeparatedString(value)
+			for _, owner := range ownerValues {
+				owners = append(owners, &owner)
+			}
+		case "aws:name":
```
njtran, line 212:
```suggestion
		case "aws::name":
```

pkg/cloudprovider/amifamily/ami.go
```diff
@@ -193,23 +193,41 @@ func (p *AMIProvider) fetchAMIsFromEC2(ctx context.Context, amiSelector map[stri
...
 				Values: []*string{aws.String(value)},
 			})
 		}
 	}
-	return filters
+	if owners == nil {
+		self := aws.String("self")
+		amazon := aws.String("amazon")
+		owners = []*string{self, amazon}
+	}
```
njtran, line 228: nit: you can collapse this into one line if you'd like.
```suggestion
	if owners == nil {
		owners = []*string{aws.String("self"), aws.String("amazon")}
	}
```

njtran (changes requested):
Hey @willthames, thanks for updating the PR. Can you update the design doc too and fix the CI check? Additionally can you add some unit tests? I can help out with that if you need some guidance.  Feel free to take the PR out of draft when you're ready too.

willthames:
I've added documentation for the new fields, but this has made me realise that we might want to tweak the case where `aws::ids`/`aws-ids` is set but `aws::owners` is not set - it's a lot safer to look up an image-id without an owner - in that case it likely makes more sense not to set `owners`.

willthames:
I've added the change to not have default owners when AMI IDs are explicitly set.

njtran (changes requested):
Sorry this might make it a little bit harder, but I’d like to keep the tests consistent with how we run other tests with ginkgo. I’m not familiar with stretchr/testify, so it might still work with the CI and other Makefile targets that are used to run the unit tests. Any of the other suite_test files that you see should make help guide you how they should be formatted.

willthames:
@njtran Tests rewritten in ginkgo/gomega.

pkg/cloudprovider/amifamily/ami_test.go
```diff
@@ -0,0 +1,131 @@
...
+		}))
+	})
+	It("should not set owners when legacy ids are passed", func() {
+		amiSelector := map[string]string{
+			"aws::name": "my-ami",
+			"aws-ids":   "ami-abcd1234,ami-cafeaced",
+		}
+		filters, owners := getFiltersAndOwners(amiSelector)
+		Expect(owners).Should(BeNil())
+		Expect(filters).Should(ConsistOf([]*ec2.Filter{
```
njtran, line 57: Sanity check on `ConsistOf()`, this does a deep comparison of the elements rather than comparing the addresses, right?

pkg/cloudprovider/amifamily/ami_test.go
```diff
@@ -0,0 +1,131 @@
...
+			{
+				Name: aws.String("image-id"),
+				Values: aws.StringSlice([]string{
+					"ami-abcd1234",
+					"ami-cafeaced",
+				}),
+			},
+		}))
+	})
+	It("should have work with prefixed id and prefixed name and owners", func() {
```
njtran, line 92: nit on wording:
```suggestion
	It("should allow prefixed id, prefixed name, and prefixed owners", func() {
```

pkg/cloudprovider/amifamily/ami_test.go
```diff
@@ -0,0 +1,131 @@
...
+			{
+				Name: aws.String("image-id"),
+				Values: aws.StringSlice([]string{
+					"ami-abcd1234",
+					"ami-cafeaced",
+				}),
+			},
+		}))
+	})
+	It("should have work with prefixed name and owners", func() {
```
njtran, line 114: nit on wording: 
```suggestion
	It("should allow prefixed name and prefixed owners", func() {
```

pkg/cloudprovider/amifamily/ami_test.go
willthames: Definitely (this was one of my concerns) but the tests simply wouldn't pass if this wasn't the case

pkg/cloudprovider/amifamily/ami.go
```diff
@@ -171,16 +171,20 @@ func (p *AMIProvider) selectAMIs(ctx context.Context, amiSelector map[string]str
...
 	if err != nil {
 		return nil, err
 	}
 	if amis, ok := p.ec2Cache.Get(fmt.Sprint(hash)); ok {
 		return amis.([]*ec2.Image), nil
 	}
+	describeImagesInput := &ec2.DescribeImagesInput{Filters: filters}
+	if owners != nil {
+		describeImagesInput.Owners = owners
+	}
```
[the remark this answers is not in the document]
willthames: looking at it, it should have the identical functionality due to go's unset value defaults.

pkg/cloudprovider/amifamily/ami_test.go
```diff
@@ -0,0 +1,131 @@
...
+			{
+				Name: aws.String("image-id"),
+				Values: aws.StringSlice([]string{
+					"ami-abcd1234",
+					"ami-cafeaced",
+				}),
+			},
+		}))
+	})
+	It("should not set owners when prefixed ids are passed", func() {
```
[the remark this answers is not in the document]
willthames: It doesn't make any sense to pass the two together, in general, but I also don't wish to over-complicate the logic - if people do want to pass them both, as long as the result is a working AMI filter, they'll be happy.

For this test I could remove the `aws::name` parameter (as it's not adding anything to the test)

go.mod
```diff
@@ -52,13 +52,15 @@ require (
 	github.com/go-openapi/swag v0.19.15 // indirect
 	github.com/go-playground/locales v0.14.1 // indirect
 	github.com/go-playground/universal-translator v0.18.1 // indirect
+	github.com/go-task/slim-sprig v0.0.0-20210107165309-348f09dbbbc0 // indirect
```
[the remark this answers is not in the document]
willthames: Not sure to be honest - might have been due to a `go run ginkgo` - a `go mod tidy` seems to have fixed this back up.

pkg/cloudprovider/amifamily/ami_test.go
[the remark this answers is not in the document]
njtran: If you'd like to, I think moving the name filter out of each of the tests would make sense.

njtran (commented):
Thanks for all the updates! I'll let the tests run now.

ellistarn:
Looks like there's some formatting that needs checking in. @njtran, can you help out @willthames so we can get this merged, he's gone above and beyond already. :)

njtran (approved):

njtran:
Thanks for all the work @willthames! I'll add in the e2e test.

PR #3255, opened 2023-01-23
fix: a crash when instance can't be looked up on node creation

bwagner5:
Fixes # 
N/A

**Description**
 - If the DescribeInstances call fails after 6 tries, the err message tries to dereference the `instance` which is nil if it can't be looked up which results in Karpenter crashing rather than just reporting an error back to the provisioner. 

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


jonathan-innis (approved):
LGTM 🚀

PR #3260, opened 2023-01-24
feat: Add CRD charts as a dependency

spring1843:
Fixes # 

**Description**
* Adds CRDs of karpenter and karpenter-core as a dependency to the Helm package so that they are managed by the Helm installation
* Use `none` instead of `no tag` for space matching problems

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


charts/karpenter/Chart.yaml
```diff
@@ -14,3 +14,10 @@ home: https://karpenter.sh/
 icon: https://repository-images.githubusercontent.com/278480393/dab059c8-caa1-4b55-aaa7-3d30e47a5616
 sources:
   - https://github.com/aws/karpenter/
+dependencies:
```
ellistarn, line 17: Do we want to update these as part of the release PR? Or should we upgrade manually?

ellistarn (approved):

charts/karpenter/Chart.yaml
spring1843: Right now it's manual, but I will have PR to update that soon.

PR #3314, opened 2023-01-29
feat: allow associate public ipAddress on nodes

nami-su:
Fixes # https://github.com/aws/karpenter/issues/2026

**Description**

Adds support to the AWS Node Template for enabling associate public ipAddress.

* Unit tests

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
The AWS Node Template now has a associatePublicIpAddress field that can be used to enable public ipAddress association on EC2 instances that are launched.
```


website/content/en/preview/concepts/node-templates.md
```diff
@@ -34,6 +34,7 @@ spec:
   metadataOptions: { ... }       # optional, configures IMDS for the instance
   blockDeviceMappings: [ ... ]   # optional, configures storage devices for the instance
   detailedMonitoring: "..."      # optional, configures detailed monitoring for the instance
+  associatePublicIpAddress: "..."      # optional, configures public address association for the instance
```
bwagner5, line 37:
```suggestion
  associatePublicIPAddress: "..."      # optional, configures public address association for the instance
```

website/content/en/preview/concepts/node-templates.md
```diff
@@ -358,6 +359,13 @@ spec:
   detailedMonitoring: true
 ```
 
+## spec.associatePublicIpAddress
+
+Enabling associate public ipAddress on the node template controls the [EC2 associate public ipaddress](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-networkinterface.html) feature. If you enable this option, EC2 instance associates a public IPv4 address with eth0 for a new network interface.
```
bwagner5, line 364:
```suggestion
## spec.associatePublicIPAddress

Enabling associate public ipAddress on the node template controls the [EC2 associate public ipAddress](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-networkinterface.html) feature. If you enable this option, EC2 instance associates a public IPv4 address with eth0 for a new network interface.
```

website/content/en/preview/concepts/node-templates.md
```diff
@@ -358,6 +359,13 @@ spec:
...
+```yaml
+spec:
+  associatePublicIpAddress: true
```
bwagner5, line 367:
```suggestion
  associatePublicIPAddress: true
```

bwagner5 (changes requested):
Thanks for the PR!  

There are a few changes needed on casing of `IPAddress` to pass linting and our general API naming standards with acronyms.  I think there is also an issue with the default since it looks like it will default to `false` which would result in a public IP address not being assigned to an instance even if it's launched in a subnet with `auto-assign-public-ips`. 

I'm also curious about the use-case that makes it necessary to use `associatePublicIPAddress` rather than just setting the subnet to auto assign IPs and use those public subnets?

bwagner5:
We may actually want to add NetworkInterfaces to the AWSNodeTemplate where we'd get this for free, see https://github.com/aws/karpenter/issues/2026

if you think this PR should still be valid, please reopen and we can discuss.

PR #3331, opened 2023-01-31
feature: add CRDs to helm chart

spring1843:
Fixes # 

**Description**

* Optionally adds CRDs to helm charts, defaults to false for a transitionary period

* My fork

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


Makefile
```diff
@@ -21,6 +21,7 @@ HELM_OPTS ?= --set serviceAccount.annotations.eks\\.amazonaws\\.com/role-arn=${K
 			--set settings.aws.clusterEndpoint=${CLUSTER_ENDPOINT} \
 			--set settings.aws.defaultInstanceProfile=KarpenterNodeInstanceProfile-${CLUSTER_NAME} \
 			--set settings.aws.interruptionQueueName=${CLUSTER_NAME} \
+			--set crd-helm-chart-inclusion.enabled=true \
```
ellistarn, line 24:
```suggestion
			--set include-crds=true \
```

charts/karpenter/Chart.yaml
```diff
@@ -14,3 +14,12 @@ home: https://karpenter.sh/
 icon: https://repository-images.githubusercontent.com/278480393/dab059c8-caa1-4b55-aaa7-3d30e47a5616
 sources:
   - https://github.com/aws/karpenter/
+dependencies:
+  - name: karpenter-crd
+    version: "v0.23.0"
```
bwagner5, line 19: does this get auto-updated?

Makefile
ellistarn: Could we support the `--skip-cords` flag? https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

charts/karpenter/Chart.yaml
spring1843: Not at the moment but will do soon. Have to test that a bit.

Makefile
spring1843: I don't think so, I believe you can do that when you use the other option of putting CRDs in a directory like we were doing in past. Now that CRDs are dependencies that flag is ignored.

charts/karpenter/Chart.yaml
spring1843: A few other things such as the go.mod need to be updated that's why it's better suited as a follow up: will most likely happen here https://github.com/aws/karpenter/blob/main/hack/release/common.sh#L53

stevehipwell:
I don't want to come across as negative but I'm not convinced that this pattern is valid for the two reasons I'll list below; plus I'm concerned that usability will be impacted by changing away from the idiomatic pattern of CRDs defaulting to being in the `crds` folder.

Reason 1 - I'm not 100% sure if this is an issue in this case; but if the two dependencies have dependencies between each other then this will not work.

Reason 2 - This pattern is explicitly called out as incorrect in the [Helm CRD documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) and reads as follows.

> Another way to do this is to put the CRD definition in one chart, and then put any resources that use that CRD in another chart.
> 
> In this method, each chart must be installed separately. However, this workflow may be more useful for cluster operators who have admin access to a cluster

Note the vital use of "separately", dependent charts aren't separate. There are plenty of issues and threads out there showing how people have tried this pattern or similar and broke things.

spring1843:
>Reason 1 - I'm not 100% sure if this is an issue in this case; but if the two dependencies have dependencies between each other then this will not work.

There's no circular dependencies here. Nor are there going to be any since these dependencies are solely CRDs.

> Reason 2 - This pattern is explicitly called out as incorrect in the [Helm CRD documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) and reads as follows.

>Another way to do this is to put the CRD definition in one chart, and then put any resources that use that CRD in another chart.

This is not the path that is taken here, we have added CRDs as a dependency so they are not two separate charts.

Not sure if you noticed or not @stevehipwell but this will only optionally take the CRDs as part of dependencies and is disabled by default. So this should theoretically not cause any changes to what you are doing right now.

stevehipwell:
> This is not the path that is taken here, we have added CRDs as a dependency so they are not two separate charts.

@spring1843 exactly, this goes directly against the Helm guidance for CRDs as linked above. If you use the CRD charts independently you're fine. Behaviour wise as well as opening up race conditions you've also removed the ability for people to uninstall Karpenter and re-install it leaving any custom provisioners in place; this pattern will delete the CRDs on uninstall.

I know I can opt out of this behaviour but that just fixes it for me not everyone else.

[Helm CRD Architecture](https://github.com/helm/community/blob/f9e06c16d89ccea1bea77c01a6a96ae3b309f823/architecture/crds.md)

spring1843:
Related: https://github.com/aws/karpenter-core/pull/199

spring1843:
@stevehipwell 
> I know I can opt out of this behaviour but that just fixes it for me not everyone else.

It shouldn't really affect anyone since it's disabled by default for now. This has been tested against variety scenarios and it seems to be producing expected outcome.

1. For a fresh install, if `crd-helm-chart-inclusion.enabled=true` is used it will install the CRDs, if it's not it will not and the user will have to do it manually

2. For upgrading an already existing version. If `crd-helm-chart-inclusion.enabled=true` it will produce an error asking the user to remove the already existing chart before upgrading. If `crd-helm-chart-inclusion.enabled=false` it will not attempt to install the CRDs and will just use what already exists. 

3. When we update charts if `crd-helm-chart-inclusion.enabled=true` is used they will be upgraded automatically, and ignored otherwise.

charts/karpenter/Chart.yaml
spring1843: donen in https://github.com/aws/karpenter/pull/3331/commits/f55598213fd527bf131385d9c11f9e43298184e6

website/content/en/preview/upgrade-guide.md
```diff
@@ -92,6 +92,7 @@ By adopting this practice we allow our users who are early adopters to test out
 # Released Upgrade Notes
 
 ## Upgrading to v0.24.0+
+* The CRDs can be optionally managed by helm when using `include-crds.enabled=true` during installation or upgrade. If CRDs already exist they must be removed before running helm with this option.`kubectl delete CustomResourceDefinition awsnodetemplates.karpenter.k8s.aws` and `kubectl delete CustomResourceDefinition provisioners.karpenter.sh`
```
bwagner5, line 95: We need to note that the instances will be terminated when doing these steps.

spring1843 (commented):
/karpenter snapshot

spring1843:
Closing this because we decided to publish independent charts for CRDs rather than including them in as in this approach.

PR #3332, opened 2023-01-31
fix: remove terminateinstancesbatcher log line

bwagner5:
Fixes # 

**Description**
 - Remove unneeded error log in terminateinstancesbatcher

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,6 @@ coverage:
 verify: tidy download ## Verify code. Includes dependencies, linting, formatting, etc
 	go generate ./...
 	hack/boilerplate.sh
-	curl https://raw.githubusercontent.com/aws/karpenter-core/main/pkg/apis/crds/karpenter.sh_provisioners.yaml > pkg/apis/crds/karpenter.sh_provisioners.yaml
 	$(foreach dir,$(MOD_DIRS),cd $(dir) && golangci-lint run $(newline))
 	@git diff --quiet ||\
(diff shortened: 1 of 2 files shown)


jonathan-innis (approved):
🚀

Makefile
```diff
@@ -101,7 +101,6 @@ coverage:
 verify: tidy download ## Verify code. Includes dependencies, linting, formatting, etc
 	go generate ./...
 	hack/boilerplate.sh
-	curl https://raw.githubusercontent.com/aws/karpenter-core/main/pkg/apis/crds/karpenter.sh_provisioners.yaml > pkg/apis/crds/karpenter.sh_provisioners.yaml
```
spring1843, line 104: This may be used in the crds directory of helm charts at generation time. Best to leave it here for now and remove with https://github.com/aws/karpenter/pull/3331

spring1843 (changes requested):

Makefile
bwagner5: This is incorrectly downloading the latest CRD rather than the core-dependency and is in the verification step. I don't think we can trust it either way.

bwagner5: After looking at it, I'm fairly certain this is safe to remove now.  If there were any updates, the verification check would fail since there are modifications. If it's up-to-date, then we don't need to do anything. So the only way there is a side effect is if the repo is not up-to-date with the CRDs from core, but since it's at HEAD of main rather than the dep, it will likely be a false positive anyways.

Issue #1495, opened 2022-03-10
Support specifying AMI family release version in provisioner


stevehipwell:
**Tell us about your request**
When specifying an AMI family in the provisioner I'd also like to be able to specify the release version, with the default behaviour falling back to the latest available version. This might need to be modelled as a map if the provisioner is supporting multiple families so each ones version can be specified.

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
We need to validate each AMI release before we're willing to allow any of or clusters to use it, we've been caught out too many times by AMIs that cause issues. We also want to be able to keep our node group node versions aligned with our Karpenter versions.

**Are you currently working around this issue?**
n/a

**Additional context**
This is related to #1008 and the logic for this could be used when an AMI ID is specified which is an official one.

**Attachments**
n/a

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

njtran:
Sorry for the late response on this. We're thinking of implementing this as part of an extension of the current AMI controls after implementing a mechanism for enforcing AMI Drift [here](https://github.com/aws/karpenter/issues/1716). Linking it in the #1738 issue as well for tracking.

njtran:
Hey all, if you're tracking this, come give some feedback on https://github.com/aws/karpenter-core/pull/207!

stevehipwell:
@njtran I think this has been implemented hasn't it?

jonathan-innis:
> I think this has been implemented hasn't it

This isn't in there yet. We pin to the K8s control-plane minor version but still auto-upgrade to the latest patch without giving any control just based on a semver selector. You can really only pin today by grabbing the AMI id and pinning that in your `amiSelectorTerms` or you are just forced to take the latest patch version.

If you are using your own AMIs, you're in a bit of a worse spot since you would have to tag the version using image tags on the image that you created AND you would have to specify that same version tag in the `amiSelectorTerms`. There are some ideas here around templating the discovered control-plane K8s version so you could do something like the example below.

```
amiSelectorTerms:
  - version: "{{ .ClusterVersion.Major}}.{{ .ClusterVersion.Minor }}"
```

[...]

PR #3356, opened 2023-02-06
RFC: added native ami versioning design doc

njtran:
**Description**
- Design doc for  #1495 

*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/designs/native-ami-versioning.md
+++ b/designs/native-ami-versioning.md
@@ -1,6 +1,6 @@
 # Native AMI Versioning
 
-Karpenter uses the latest EKS Optimized AMI releases as defaults in their AMI Requirements. With [Drift](https://karpenter.sh/preview/concepts/deprovisioning/#drift) implemented, when a new EKS Optimized AMI is released users who have enabled drift could see their fleet rolled out to the newly released AMI without warning.
+Karpenter uses the latest EKS Optimized AMI releases for Bottlerocket, Amazon Linux 2, and Ubuntu as the defaults for AMI Requirements. With [Drift](https://karpenter.sh/preview/concepts/deprovisioning/#drift) implemented, when a new EKS Optimized AMI is released, users who have enabled drift could see the newly released AMIs rolled out to their fleet without warning.
 
 [EKS Optimized AMI Releases](#eks-optimized-ami-releases) are irregular and versioned differently. Without validation, irregular release cadences could introduce unwanted AMIs in production. This doc proposes two functional improvements to AMI selection logic to ease customer pain.
 
--- a/designs/native-ami-versioning.md
+++ b/designs/native-ami-versioning.md
@@ -27,7 +27,7 @@ While this isn’t a difficult process, it isn't viable to require users to jump
 
 ### 🔑 Introduce a minimumAgeDays field in the amiSelector.
 
-While this hasn’t been asked for explicitly, bake time is a common concept in phasing in new changes. Whether a user has a mandated bake time for production or developers have an estimated time to validate workloads in before rolling out to prod, bake time provides a sliding window of leeway. Users can be cautious and set a high value, or leave out the setting for no bake time requirements.
+While this hasn’t been asked for explicitly, bake time is a common concept in phasing in new changes. Whether a user has a mandated bake time for production or developers have an estimated time to validate workloads before rolling out to prod, bake time provides a sliding window of leeway. Users can be cautious and set a high value, or leave out the setting for no bake time requirements.
 
 ```
 apiVersion: karpenter.k8s.aws/v1alpha1
--- a/designs/native-ami-versioning.md
+++ b/designs/native-ami-versioning.md
@@ -57,7 +57,7 @@ spec:
 
 ### DriftTTL in the ProvisionerSpec
 
-Karpenter could include a drift TTL in the Provisioner. Nodes would have to be drifted for the TTL's duration to be deprovisioned. This can implement bake time, but will also apply for all future methods of Drift, such as zonal and instance permission requirements. Since other Drift methods may need more immediate actions when it occurs, setting a Drift TTL may not be granular as needed. This
+Karpenter could include a drift TTL in the Provisioner. Nodes would have to be drifted for the TTL's duration to be deprovisioned. This can implement bake time, but will also apply for all future methods of Drift, such as zonal and instance permission requirements. Since other Drift methods may need more immediate actions when it occurs, setting a Drift TTL may not be granular as needed.
 
 ### 🔑 Suggestion
 
--- a/designs/native-ami-versioning.md
+++ b/designs/native-ami-versioning.md
@@ -2,7 +2,7 @@
 
 Karpenter uses the latest EKS Optimized AMI releases for Bottlerocket, Amazon Linux 2, and Ubuntu as the defaults for AMI Requirements. With [Drift](https://karpenter.sh/preview/concepts/deprovisioning/#drift) implemented, when a new EKS Optimized AMI is released, users who have enabled drift could see the newly released AMIs rolled out to their fleet without warning.
 
-[EKS Optimized AMI Releases](#eks-optimized-ami-releases) are irregular and versioned differently. Without validation, irregular release cadences could introduce unwanted AMIs in production. This doc proposes two functional improvements to AMI selection logic to ease customer pain.
+[EKS Optimized AMI Releases](#eks-optimized-ami-releases) are irregular and versioned differently. Without validation, [irregular release cadences](#release-cadences) could introduce unwanted AMIs in production. This doc proposes two functional improvements to AMI selection logic to ease customer pain.
 
 ## What could happen when a new EKS Optimized AMI is released?
 
@@ -25,7 +25,7 @@ While this isn’t a difficult process, it isn't viable to require users to jump
 
 ## Bake Time
 
-### 🔑 Introduce a minimumAgeDays field in the amiSelector.
+### 🔑 Introduce a minimumAgeDuration field in the amiSelector.
 
 While this hasn’t been asked for explicitly, bake time is a common concept in phasing in new changes. Whether a user has a mandated bake time for production or developers have an estimated time to validate workloads before rolling out to prod, bake time provides a sliding window of leeway. Users can be cautious and set a high value, or leave out the setting for no bake time requirements.
 
@@ -36,15 +36,15 @@ metadata:
   name: default
 spec:
   amiSelector:
-    minimumAgeDuration: 10
+    minimumAgeDuration: 10d  // metav1.Duration
 ```
 
 ### How `minimumAgeDuration` interacts with the other methods of AMI selection
 
 #### *EKS Optimized AMI (no other AMI Selector field)*
 
-1. The newest EKS Optimized AMI that fits the `minimumAgeDuration` selector will be chosen.
-2. An AMI will be considered eligible at the time and day that it was created as shown in aws ec2 describe-images.
+1. Karpenter will compare `minimumAgeDuration` to when the SSM Parameter was last modified instead of using the Image Creation Date that's returned in `aws ec2 describe-images`. This is because the SSM Parameter is updated independently of image creation.
+2. The newest EKS Optimized AMI that fits the `minimumAgeDuration` selector will be chosen.
 
 #### *Arbitrary tags*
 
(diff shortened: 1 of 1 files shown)


designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
+# Native AMI Versioning
+
+Karpenter uses the latest EKS Optimized AMI releases as defaults in their AMI Requirements. With [Drift](https://karpenter.sh/preview/concepts/deprovisioning/#drift) implemented, when a new EKS Optimized AMI is released users who have enabled drift could see their fleet rolled out to the newly released AMI without warning.
```
bwagner5, line 3: The sentence structure is a bit confusing here.  I think it's worth calling out the different "EKS Optimized" AMIs in this intro section too since that can be another point of confusion on exactly what we're talking about here. 

```suggestion
Karpenter uses the latest EKS Optimized AMI releases for Bottlerocket, Amazon Linux 2, and Ubuntu as the defaults for AMI Requirements. With [Drift](https://karpenter.sh/preview/concepts/deprovisioning/#drift) implemented, when a new EKS Optimized AMI is released, users who have enabled drift could see the newly released AMIs rolled out to their fleet without warning.
```

designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
...
+```
+spec:
+    amiFamily: Ubuntu # Or Bottlerocket or AL2
+```
+
+While this isn’t a difficult process, it isn't viable to require users to jump through these hoops. They must be aware of new EKS Optimized AMI releases, understand SSM aliases to fetch the AMI IDs, and modify their AWSNodeTemplate values in their CI/CD pipelines multiple times. Some users have [requested these improvements](https://github.com/aws/karpenter/issues/1495). In order to improve the experience for all users, we’ll introduce (1) Bake Time and (2) Release Version Pinning improvements.
+
+## Bake Time
+
+### 🔑 Introduce a minimumAgeDays field in the amiSelector.
```
bwagner5, line 28: Days are not necessarily all the same length of time, would hours be better here and then we could use a go duration? 

I'm not sure minimumAge is a good concept for the feature we're trying to create here. ttlAfterCreated seems nicer since it matches the other ttl semantic, with the obvious downside that "Created" is somewhat ambiguous Is it SSM parameter created or AMI created? I think it needs to the SSM parameters changed since it's the standard practice to build the AMI early and then flip the SSM parameter when actually released.

designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
...
+
+While this hasn’t been asked for explicitly, bake time is a common concept in phasing in new changes. Whether a user has a mandated bake time for production or developers have an estimated time to validate workloads in before rolling out to prod, bake time provides a sliding window of leeway. Users can be cautious and set a high value, or leave out the setting for no bake time requirements.
```
bwagner5, line 30:
```suggestion
While this hasn’t been asked for explicitly, bake time is a common concept in phasing in new changes. Whether a user has a mandated bake time for production or developers have an estimated time to validate workloads before rolling out to prod, bake time provides a sliding window of leeway. Users can be cautious and set a high value, or leave out the setting for no bake time requirements.
```

designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
...
+
+```
+apiVersion: karpenter.k8s.aws/v1alpha1
+kind: AWSNodeTemplate
+metadata:
+  name: default
+spec:
+  amiSelector:
+    minimumAgeDuration: 10
```
bwagner5, line 39: I thought we had talked about duration! :) Was the above just outdated?

designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+### How `minimumAgeDuration` interacts with the other methods of AMI selection
+
+#### *EKS Optimized AMI (no other AMI Selector field)*
+
+1. The newest EKS Optimized AMI that fits the `minimumAgeDuration` selector will be chosen.
+2. An AMI will be considered eligible at the time and day that it was created as shown in aws ec2 describe-images.
```
bwagner5, line 47: I think this is probably not right. See above comment.

designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
...
+1. An AMI tagged and discovered via the amiSelector tags will not be chosen unless it fits the `minimumAgeDuration` filter.
+2. This has the downside of resulting in no AMIs found if users are too restrictive with tags and the `minimumAgeDuration` filter. The upside is that users will know any AMI created before x days ago will not be included.
+
+#### *AMI IDs*
+
+1. When using AMI IDs, other tags specified are not respected. To keep explicit AMI ID usage as an override, using `minimumAgeDuration` and AMI IDs will result in `minimumAgeDuration` not being respected.
+
+### DriftTTL in the ProvisionerSpec
+
+Karpenter could include a drift TTL in the Provisioner. Nodes would have to be drifted for the TTL's duration to be deprovisioned. This can implement bake time, but will also apply for all future methods of Drift, such as zonal and instance permission requirements. Since other Drift methods may need more immediate actions when it occurs, setting a Drift TTL may not be granular as needed. This
```
bwagner5, line 60:
```suggestion
Karpenter could include a drift TTL in the Provisioner. Nodes would have to be drifted for the TTL's duration to be deprovisioned. This can implement bake time, but will also apply for all future methods of Drift, such as zonal and instance permission requirements. Since other Drift methods may need more immediate actions when it occurs, setting a Drift TTL may not be granular as needed.
```

designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
...
+metadata:
+  name: default
+spec:
+  amiFamily: Bottlerocket
+  amiVersion: v1.9.1 # Defaults to latest, requires AMIFamily if specified
+```
+
+### 🔑 Suggestion
+
+Implement only Option 1. This keeps configuration within the amiFamily field, does not add new fields to be managed, and works well for users who are aware of EKS Optimized semantics, which is the audience for this feature.
```
bwagner5, line 107: This is easily moved too if we decide to move the amiFamily logic into the amiSelector or some other richer API like in the suggestion [here](https://github.com/aws/karpenter/pull/3204/files#diff-e08ed94660726c8fa676b3fdf3d9c1640b4207419768252f2f19f07825d2fe85).

designs/native-ami-versioning.md
njtran: Sorry, this is from a previous version of the doc. I had changed all occurrences to `minimumAgeDuration` except this one 🤦 . Does your comment still hold after it being changed to `minimumAgeDuration`?

designs/native-ami-versioning.md
njtran: Yes! Sorry about that.

designs/native-ami-versioning.md
njtran: Discussed offline, the SSM Parameter is updated in a varied amount of time since the image creation. Because of that we can special case this `minimumAgeDuration` to work based off the SSM Parameter creation.

designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
...
+This may be more intuitive to users that are using EKS Optimized AMIs, but has no value add for those who aren’t. This may be confusing to users looking to extend the amiVersion field into other methods of custom AMI selection, especially since it’s a top level field. One alternative is to make this mutually exclusive with AMISelector so that there’s no room for confusion.
+
+```
+apiVersion: karpenter.k8s.aws/v1alpha1
+kind: AWSNodeTemplate
+metadata:
+  name: default
+spec:
+  amiFamily: Bottlerocket
+  amiVersion: v1.9.1 # Defaults to latest, requires AMIFamily if specified
```
willthames, line 102: I love this if it can be part of the AMISelector API change - this would meet our needs perfectly

alex-berger, line 102: I am also in favor of this, as we could further extend `amiVersion` to support [semver version ranges](https://devhints.io/semver). Some examples: 
- `1.9.1` pin to exactly this version
- `^1.9` allow for any bug-fix release upgrades
- `^1` allow for minor release upgrades
-  ...

designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
...
+
+Use an AMI Family with an appended magic string. This allows users who only know of version strings to modify AWSNodeTemplates easily to their liking.
+
+```
+apiVersion: karpenter.k8s.aws/v1alpha1
+kind: AWSNodeTemplate
+metadata:
+  name: default
+spec:
+  amiFamily: AL2-v20220824 # No version suffix defaults to latest
```
ellistarn, line 82: While this is a cool semantic that maps to container image semantics (e.g. amiFamily: AL2:v20220824), I think we're mixing up concepts here.

amiFamily is currently a runtime contract that tells Karpenter what to expect for the user data, ebs volumes, etc. Users can specify amiFamily as well as specific amiIDs and this semantic would enable them to specify both simultaneously. 

Does it make sense to support pinning through amiID instead?

designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
...
+
+### 🔑 Suggestion
+
+Implement only Option 1. This keeps configuration within the amiFamily field, does not add new fields to be managed, and works well for users who are aware of EKS Optimized semantics, which is the audience for this feature. Additionally, since this doesn't add another API field, this can be easily moved to Option 2 or another option if that's a better place in the future.
+
+## Appendix
+
+### EKS Optimized AMI Releases
+
+EKS Optimized AMIs are not released on a regular cadence, and do not all use the same versioning semantic.
```
ellistarn, line 113: Did we explore changing this?

designs/native-ami-versioning.md
njtran: This would be a cool extension! For specifically the EKS Optimized AMIs, this would only work for Bottlerocket. My only qualm with this option is that we'd need to add some way to plug this into custom AMIs for it to be useful for all users.

designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
...
+spec:
+  amiSelector:
+    minimumAgeDuration: 10d  // metav1.Duration
+```
+
+### How `minimumAgeDuration` interacts with the other methods of AMI selection
+
+#### *EKS Optimized AMI (no other AMI Selector field)*
+
+1. Karpenter will compare `minimumAgeDuration` to when the SSM Parameter was last modified instead of using the Image Creation Date that's returned in `aws ec2 describe-images`. This is because the SSM Parameter is updated independently of image creation.
```
alex-berger, line 46: This might be a bit tricky, as the corresponding SSM Parameters have no value history and only the most recent value is available. Let's assume `minimumAgeDuration: 15d` and that there is a new AMI published every `7d`, then the bake time requirement will never be met and the nodes would be stuck on the current AMI version in use.

sftim:
How about another control loop that can manage an AWSNodeTemplate? Either have more than one Provisioner, or change the way that Provisioners reference node templates so it's done using a selector.

That'd feel more like how other Kubernetes APIs work: if you want to use a static AWSNodeTemplate, you do that; if you want automated customization, you use a controller to make this happen.

designs/native-ami-versioning.md
wkaczynski: > Does it make sense to support pinning through amiID instead?
as amiIDs are different in different regions (and for different architectures, gpu / non-gpu setups, etc), if we allow to pin only using amiID, we force users to perform tedious and error-prone lookups (while we could allow to specify some selected optional, potentially even amiFamily-specific constraints)

allowing to specify a version could be a great addition to the `minimumAgeDuration` bake time
(not saying that specifying these optional constraints in the `amiFamily` field itself is the best option)

designs/native-ami-versioning.md
wkaczynski: also `minimumAgeDuration` bake time should then probably be ignored only for a specific version (similarly like when amiID is specified) but not for a range

designs/native-ami-versioning.md
```diff
@@ -0,0 +1,121 @@
...
+    amiFamily: Ubuntu # Or Bottlerocket or AL2
+```
+
+While this isn’t a difficult process, it isn't viable to require users to jump through these hoops. They must be aware of new EKS Optimized AMI releases, understand SSM aliases to fetch the AMI IDs, and modify their AWSNodeTemplate values in their CI/CD pipelines multiple times. Some users have [requested these improvements](https://github.com/aws/karpenter/issues/1495). In order to improve the experience for all users, we’ll introduce (1) Bake Time and (2) Release Version Pinning improvements.
+
+## Bake Time
+
+### 🔑 Introduce a minimumAgeDuration field in the amiSelector.
+
+While this hasn’t been asked for explicitly, bake time is a common concept in phasing in new changes. Whether a user has a mandated bake time for production or developers have an estimated time to validate workloads before rolling out to prod, bake time provides a sliding window of leeway. Users can be cautious and set a high value, or leave out the setting for no bake time requirements.
```
ellistarn, line 30: I'm cautious about whether or not this is a good idea.

PR #3370, opened 2023-02-08
feat:add-extra-manifest-objects-to-helm-chart

evanshoshan:
…sources

**Description**

Added extra objects deployment to helm chart to support dynamic creation of K8s manifests via helm values files

- helm template validation 
- helm lint 

- [ x] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```

### DIFF
--- a/charts/karpenter/README.md
+++ b/charts/karpenter/README.md
@@ -48,6 +48,7 @@ helm upgrade --install --namespace karpenter --create-namespace \
 | dnsConfig | object | `{}` | Configure DNS Config for the pod |
 | dnsPolicy | string | `"Default"` | Configure the DNS Policy for the pod |
 | extraVolumes | list | `[]` | Additional volumes for the pod. |
+| extraObjects | list | `[]` | Array of extra K8s manifests to deploy. |
 | fullnameOverride | string | `""` | Overrides the chart's computed fullname. |
 | hostNetwork | bool | `false` | Bind the pod to the host network. This is required when using a custom CNI. |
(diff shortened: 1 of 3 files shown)


jonathan-innis:
@evanshoshan Can you explain why this is needed? Why not just deploy your K8s objects through some other mechanism like helm chart dependencies rather than injecting these K8s manifests directly into the chart?

nxf5025:
> @evanshoshan Can you explain why this is needed? Why not just deploy your K8s objects through some other mechanism like helm chart dependencies rather than injecting these K8s manifests directly into the chart?

@jonathan-innis - Not the PR author, but I can chime in here...This is starting to become a common trend with helm charts. A few other examples can be found here:

- [Grafana](https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/extra-manifests.yaml)
- [Traefik](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/extra-objects.yaml)
- [ArgoCD](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/extra-manifests.yaml)

This is really just a convenience update so you don't need to worry about a subchart for an extra file or two. For more advanced cases I definitely agree setting up a dependency would be the best route, but I don't see any downside to adding this in.

jonathan-innis:
@nxf5025 Thanks for the context here. You're right that this seems to be more and more of a pattern in the chart community. Sounds good to me.

jonathan-innis (approved):
LGTM 🚀

PR #3394, opened 2023-02-13
feat: Add envFrom to container spec and expose in values file for use in proxy environments

jebbens:
Fixes # 3388 

**Description**
Many of our customers route traffic through network proxies, which requires deploying a configMap with http_proxy environment variables in the cluster, and referencing that configMap in each container that is part of a deployment. Adding the envFrom property in the container spec and exposing it in the values file makes it easy for customers to use the upstream directly without having to maintain their own charts with that modification.
I made the changes, packaged as a local chart, then tested in an environment with a proxy. 

- [ ] Yes, issue opened: # 
- [x ] No

**Release Note**

```release-note
NONE
```


charts/karpenter/templates/deployment.yaml
```diff
@@ -89,6 +89,10 @@ spec:
           {{- with .Values.controller.env }}
             {{- toYaml . | nindent 12 }}
           {{- end }}
+          envFrom:
+          {{- with .Values.controller.envFrom }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
```
bwagner5, line 95: I think this should be like below so that it works with the default empty list:

```suggestion
          {{- with .Values.controller.envFrom }}
          envFrom:
            {{- toYaml . | nindent 12 }}
          {{- end }}
```

bwagner5 (changes requested):
Can you remove the checked in `charts/karpenter-0.24.1.tgz` file?

jebbens:
Removed .tgz in b26d7ea684fb8eea7e8e7f30e57f49bcf11ec161

charts/karpenter/values.yaml
```diff
@@ -101,6 +101,10 @@ controller:
   env: []
   # - name: AWS_REGION
   #   value: eu-west-1
+  envFrom: []
+  # - configMapRef:
+  #     name: proxy-environment-variables
+  #     optional: true
```
ellistarn, line 107:
```suggestion
```

charts/karpenter/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: karpenter
 description: A Helm chart for Karpenter, an open-source node provisioning project built for Kubernetes.
 type: application
-version: 0.24.0
+version: 0.24.1
```
ellistarn, line 5: Let's leave this as is, and catch it in the next release.

bwagner5, line 5:
```suggestion
version: 0.24.0
```

bwagner5 (approved):
LGTM!

PR #3395, opened 2023-02-13
perf: Reduce multiple describe instances calls on start

jonathan-innis:
Fixes # 

**Description**

Currently, we will hit cache misses on startup for the DescribeInstanceTypes calls due to parallel calls to GetInstanceTypes. This results in a larger memory footprint (> 1GB) at startup that eventually reduces over time to around ~300MB. This ensures that the calls to the API are synchronous so we reduce our cache misses which means we are consistently sitting around 300MB throughout the pod's lifetime.

* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/pkg/cloudprovider/instancetypes.go
+++ b/pkg/cloudprovider/instancetypes.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"sync"
 	"sync/atomic"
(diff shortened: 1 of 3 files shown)


bwagner5 (approved):
LGTM

PR #3398, opened 2023-02-14
BREAKING CHANGE: Refactor controller image values

nxf5025:
**Description**
This updates the `controller.image` value in the helm chart to be broken out to a map consisting of `controller.image.repository`, `controller.image.tag`, and `controller.image.digest`.

This enables users to provide just a repository override while leaving the tag and digest as the chart default. This is a very common approach that can be found in other widely adopted open source charts such as the [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L490) and [cluster-autoscaler](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/values.yaml#L235)

* `helm lint`
* `helm template` with test values

- [x] Yes, PR includes docs updates (upgrade guide)
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Breaking the helm `controller.image` value out to a map consisting of `controller.image.repository`, `controller.image.tag`, and `controller.image.digest`.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,11 +128,16 @@ setup: ## Sets up the IAM roles needed prior to deploying the karpenter-controll
 
 build: ## Build the Karpenter controller images using ko build
 	$(eval CONTROLLER_IMG=$(shell $(WITH_GOFLAGS) KO_DOCKER_REPO="$(KO_DOCKER_REPO)" ko build -B github.com/aws/karpenter/cmd/controller))
+	$(eval IMG_REPOSITORY=$(shell echo $(CONTROLLER_IMG) | cut -d "@" -f 1 | cut -d ":" -f 1))
+	$(eval IMG_VERSION=$(shell echo $(CONTROLLER_IMG) | cut -d "@" -f 1 | cut -d ":" -f 2 -s))
+	$(eval IMG_DIGEST=$(shell echo $(CONTROLLER_IMG) | cut -d "@" -f 2))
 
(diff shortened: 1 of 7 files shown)


jonathan-innis:
@nxf5025 This PR affects some of our release automation processes so we need to test some things on our side before we can take this in. Otherwise LGTM 🎉

bwagner5:
Here's a reference to the release automation that will need to be updated to include this change: https://github.com/aws/karpenter/blob/main/hack/release/common.sh#L53

Makefile
```diff
@@ -9,6 +9,12 @@ else
 LDFLAGS ?= -ldflags=-X=github.com/aws/karpenter/pkg/utils/project.Version=$(shell git describe --tags --always)
 endif
 
+## Alias "gcut" to "cut" if running on Darwin
+UNAME := $(shell uname)
```
bwagner5, line 13: why do you need this?  Seems like you're not using any gnu cut specific things?

bwagner5 (commented):
just one comment on the gcut alias in the Makefile, otherwise LGTM

Makefile
jonathan-innis: regular Darwin cut doesn't let you do `cut -d`

bwagner5 (dismissed):
LGTM

bwagner5 (approved):
LGTM

PR #3399, opened 2023-02-14
chore: Remove Machine Hydration Controller and Add Machine Link Controller

jonathan-innis:
Fixes # 

**Description**

Adds Machine Link Controller which is responsible for querying the AWS cloudprovider for exsiting instances and using any instances that have the `karpenter.sh/provisioner-name` instance tag to create Machines that will then be picked up and hydrated by the `karpenter-core` machine controller

* `make presubmit`
* Manual Testing

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```


pkg/cloudprovider/cloudprovider.go
```diff
@@ -84,7 +89,7 @@ func New(ctx awscontext.Context) *CloudProvider {
 	}
 	kubeDNSIP, err := kubeDNSIP(ctx, ctx.KubernetesInterface)
 	if err != nil {
-		logging.FromContext(ctx).Debugf("unable to detect the IP of the kube-dns service, %s", err)
```
bwagner5, line 87: crashing caused customers issues here with custom DNS setups. https://github.com/aws/karpenter/issues/2787

pkg/cloudprovider/cloudprovider.go
```diff
@@ -340,20 +291,55 @@ func (c *CloudProvider) resolveInstanceTypes(ctx context.Context, machine *v1alp
...
+			resources.Fits(machine.Spec.Resources.Requests, i.Allocatable())
 	}), nil
 }
 
-// TODO @joinnis: Remove name resolution when Machine migration is completed
-func (c *CloudProvider) instanceToMachine(ctx context.Context, instance *ec2.Instance, instanceTypes []*cloudprovider.InstanceType) (*v1alpha5.Machine, error) {
-	machine := &v1alpha5.Machine{}
-	instanceType, found := lo.Find(instanceTypes, func(i *cloudprovider.InstanceType) bool {
-		return aws.StringValue(instance.InstanceType) == i.Name
+func (c *CloudProvider) resolveInstanceTypesFromInstance(ctx context.Context, instance *ec2.Instance) ([]*cloudprovider.InstanceType, error) {
```
bwagner5, line 298: this is a weird function because it's not obvious how it's resolving instance types from an instance. I would expect, based on the name, to return a single instance type. Is it even worth wrapping this functionality, rather than just resolving the provisioner and then getting instance types from that in-line?

pkg/cloudprovider/instance.go
```diff
@@ -140,21 +139,13 @@ func (p *InstanceProvider) GetByID(ctx context.Context, id string) (*ec2.Instanc
...
-func (p *InstanceProvider) List(ctx context.Context, machineName string) ([]*ec2.Instance, error) {
+func (p *InstanceProvider) List(ctx context.Context) ([]*ec2.Instance, error) {
 	// Use the machine name data to determine which instances match this machine
 	out, err := p.ec2api.DescribeInstancesWithContext(ctx, &ec2.DescribeInstancesInput{
 		Filters: []*ec2.Filter{
 			{
-				Name:   aws.String(fmt.Sprintf("tag:%s", v1alpha5.MachineNameLabelKey)),
-				Values: aws.StringSlice([]string{machineName}),
+				Name:   aws.String(fmt.Sprintf("tag:%s", v1alpha5.ProvisionerNameLabelKey)),
+				Values: aws.StringSlice([]string{"*"}),
```
bwagner5, line 148: use the tag-key filter instead:

```suggestion
				Name:   aws.String("tag-key"),
				Values: aws.StringSlice([]string{v1alpha5.ProvisionerNameLabelKey}),
```

pkg/controllers/machine/link/controller.go
```diff
@@ -0,0 +1,129 @@
...
+		return corecloudprovider.IgnoreMachineNotFoundError(c.cloudProvider.Delete(ctx, retrieved))
+	}
+	provisioner := &v1alpha5.Provisioner{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: provisionerName}, provisioner); err != nil {
+		if errors.IsNotFound(err) {
+			return corecloudprovider.IgnoreMachineNotFoundError(c.cloudProvider.Delete(ctx, retrieved))
+		}
+		return err
+	}
+	machine := machineutil.New(&v1.Node{}, provisioner)
```
bwagner5, line 93: why does this take a pointer to a node?

pkg/controllers/machine/link/controller.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// shouldLink checks if the cloudprovider machine is owned by a provisioner but if it wasn't created
+// by a machine. If both of these are true, then we should generate a Machine
+func (c *Controller) shouldLink(retrieved *v1alpha5.Machine, existingMachines []v1alpha5.Machine) bool {
+	if _, ok := retrieved.Labels[v1alpha5.ManagedByLabelKey]; ok {
+		return false
+	}
+	// Machine was already linked but controller-runtime cache didn't update
+	if _, ok := c.cache.Get(retrieved.Status.ProviderID); ok {
```
bwagner5, line 114: Is there harm in just linking again rather than yet-another-cache?

bwagner5 (commented):
nice! A few comments and questions

pkg/cloudprovider/cloudprovider.go
jonathan-innis: Sounds good, I'll add a comment linking the issue because I certainly won't remember this reason down the line.

pkg/controllers/machine/link/controller.go
jonathan-innis: I think so since we would get two Machines that have the same provider id. There may be room to clean this up in a garbage collection flow but I'd like to avoid it for now if we can. Plus, the cache shouldn't be that expensive from a memory perspective and we should be ripping this code out at some point when the migration is done

pkg/controllers/machine/link/controller.go
jonathan-innis: This was originally there so that the machineutil.New() func could use the NewFromNode for status details and then layer that with provisioner spec details. There may be room to clean this up later but I think it's reasonable to keep it for now

bwagner5 (dismissed):
LGTM

bwagner5 (approved):
LGTM

PR #3419, opened 2023-02-17
fix: track inflight ip usage for subnet selection between ec2 api resyncs

bwagner5:
Fixes # 

https://github.com/aws/karpenter/issues/3364
https://github.com/aws/karpenter/issues/3380

**Description**
 - When a provisioner has two subnets within the same AZ, Karpenter selects the subnet with the most available IPs when creating the instance. However, the availableIPs comes from the EC2 DescribeSubnets API which is only synchronized every minute. This causes issues when multiple instances are launched with the stale available IPs such that the IP drops below 0. This PR introduces inflight IP tracking which approximates the IPs that will be used by the instance that is being launched by using the minimum maxPods value in the instance type options list. The inflight IP tracking is best effort and is resynchronized when the subnet cache is updated. The inflight IP tracking can also go negative since it's only an approximation of magnitude of IP usage and is compared relatively. 

* Created two subnets in the same AZ with a /24 CIDR (251 usable IPs). Scaled up an inflate deployment to 10 with pod anti-affinity so that 1 pod per node is brought up.  Observed that the subnets were draining IPs equally (151 IPs were available in each subnet after the scale-up). The increased IP usage from the 1-pod per node is due to the VPC CNI allocating warm IPs at launch. 

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/instance.go
+++ b/pkg/cloudprovider/instance.go
@@ -86,11 +86,20 @@ func (p *InstanceProvider) Create(ctx context.Context, nodeTemplate *v1alpha1.AW
 		instanceTypes = instanceTypes[0:MaxInstanceTypes]
 	}
 
-	id, err := p.launchInstance(ctx, nodeTemplate, machine, instanceTypes)
+	// Get subnets given the constraints
+	subnets, err := p.subnetProvider.List(ctx, nodeTemplate)
+	if err != nil {
+		return nil, fmt.Errorf("getting subnets, %w", err)
+	}
+	if len(subnets) == 0 {
+		return nil, fmt.Errorf("no subnets matched selector %v", nodeTemplate.Spec.SubnetSelector)
+	}
+
+	id, err := p.launchInstance(ctx, nodeTemplate, machine, subnets, instanceTypes)
 	if awserrors.IsLaunchTemplateNotFound(err) {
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/instance.go
```diff
@@ -86,11 +86,20 @@ func (p *InstanceProvider) Create(ctx context.Context, nodeTemplate *v1alpha1.AW
...
+	subnets, err := p.subnetProvider.List(ctx, nodeTemplate)
+	if err != nil {
+		return nil, fmt.Errorf("getting subnets, %w", err)
+	}
+	if len(subnets) == 0 {
+		return nil, fmt.Errorf("no subnets matched selector %v", nodeTemplate.Spec.SubnetSelector)
+	}
+
+	id, err := p.launchInstance(ctx, nodeTemplate, machine, subnets, instanceTypes)
 	if awserrors.IsLaunchTemplateNotFound(err) {
```
ellistarn, line 99: Why isn't this encapsulated in `p.launchInstance(` 🤔

bwagner5: you're right, from a previous implementation, forgot to move it back. fixed

ellistarn (approved):

ellistarn (approved):

PR #3439, opened 2023-02-22
chore: Move pricingProvider to providers to prevent dependency cycles

engedaam:
Fixes 

**Description**
The goal is to move the pricing provider to its own package to prevent import dependency that may be caused when trying to access the providers from controllers. This will also instantiate the pricingProvider as a singleton in the context. 

Manually tested

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/context/context.go
```diff
@@ -68,15 +71,26 @@ func NewOrDie(ctx cloudprovider.Context) Context {
...
 	logging.FromContext(ctx).With("region", *sess.Config.Region).Debugf("discovered region")
+
 	subnetProvider := subnet.NewProvider(ec2api)
 	securityGroupProvider := securitygroup.NewProvider(ec2api)
+	pricingProvider := pricing.NewProvider(
+		ctx,
+		pricing.NewAPI(sess, *sess.Config.Region),
+		ec2api,
+		*sess.Config.Region,
+		settings.FromContext(ctx).IsolatedVPC,
```
jonathan-innis, line 82: Can we remove this from the constructor since we can just grab this from the context?

pkg/cloudprovider/instancetypes.go
```diff
@@ -69,19 +69,12 @@ type InstanceTypeProvider struct {
 }
 
 func NewInstanceTypeProvider(ctx context.Context, sess *session.Session, ec2api ec2iface.EC2API, subnetProvider *subnet.Provider,
-	unavailableOfferingsCache *awscache.UnavailableOfferings, startAsync <-chan struct{}) *InstanceTypeProvider {
+	unavailableOfferingsCache *awscache.UnavailableOfferings, pricingProvider *pricing.Provider, startAsync <-chan struct{}) *InstanceTypeProvider {
```
jonathan-innis, line 72:
```suggestion
	unavailableOfferingsCache *awscache.UnavailableOfferings, pricingProvider *pricing.Provider) *InstanceTypeProvider {
```

I don't think we need this anymore

pkg/cloudprovider/instancetypes.go
```diff
@@ -69,19 +69,12 @@ type InstanceTypeProvider struct {
 }
 
 func NewInstanceTypeProvider(ctx context.Context, sess *session.Session, ec2api ec2iface.EC2API, subnetProvider *subnet.Provider,
```
jonathan-innis, line 71:
```suggestion
func NewInstanceTypeProvider(sess *session.Session, ec2api ec2iface.EC2API, subnetProvider *subnet.Provider,
```

I don't think we need this anymore

pkg/providers/pricing/pricing.go
```diff
@@ -38,13 +38,13 @@ import (
 	"github.com/aws/karpenter-core/pkg/utils/pretty"
 )
 
-// PricingProvider provides actual pricing data to the AWS cloud provider to allow it to make more informed decisions
+// Provider provides actual pricing data to the AWS cloud provider to allow it to make more informed decisions
```
jonathan-innis, line 41: We should think about doing a similar model here to our cluster state where we have a internal data representation and then we have a controller that is informing that data representation. 

This code was added before we had the `controller.SingletonController`. Now that we have that, I think that we can leverage that here. Probably a good follow-up PR.

jonathan-innis (approved):
LGTM 🚀

PR #3447, opened 2023-02-23
fix: security group selector wildcard key

bwagner5:
Fixes # 

**Description**
 - Fixes a case on processing the security group selector to ec2 filter where a wildcard key was not transformed to a tag-key filter and instead was a full `tag:key -> value` filter which never returned security groups since wildcards are not supported within the Filter name, only Values.

* Fixed the ec2 mock for security groups to correctly evaluate wildcard filters. Ran the tests before the fix, observed failure, and after the fix, passed. 

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


tzneal (approved):
lgtm

PR #3509, opened 2023-03-02
chore: Move instance to providers

engedaam:
Fixes # 

**Description**

The goal is to move the instance provider to its own package to prevent import dependency that may be caused when trying to access the providers from controllers. This will also instantiate the instance as a singleton in the context.

* Manually tested

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/controllers/machine/link/suite_test.go
```diff
@@ -80,9 +97,30 @@ var _ = BeforeSuite(func() {
...
+		amiResolver,
+		securityGroupProvider,
+		ptr.String("ca-bundle"),
+		make(chan struct{}),
+		net.ParseIP("10.0.100.10"),
+		"https://test-cluster",
+	)
+	instanceProvider = instance.NewProvider(ctx, "", ec2API, unavailableOfferingsCache, instanceTypeProvider, subnetProvider, launchTemplateProvider)
 	instanceTypesProvider = instancetype.NewProvider("", cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval), ec2API, subnetProvider, unavailableOfferingsCache, pricingProvider)
 	cloudProvider = cloudprovider.New(awscontext.Context{
```
jonathan-innis, line 125: I think we only need to instantiate the things in these tests that the test has dependency on. So in this case, I'm not sure that we need to take dependency on all of these other providers

pkg/controllers/machine/garbagecollect/suite_test.go
```diff
@@ -75,7 +95,29 @@ var _ = BeforeSuite(func() {
 	ctx = settings.ToContext(ctx, test.Settings())
 	env = coretest.NewEnvironment(scheme.Scheme, coretest.WithCRDs(apis.CRDs...))
 	unavailableOfferingsCache = awscache.NewUnavailableOfferings()
+	launchTemplateCache = cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval)
```
jonathan-innis, line 98: I think we only need to instantiate the things in these tests that the test has dependency on. So in this case, I'm not sure that we need to take dependency on all of these other providers

pkg/providers/instance/instance.go
```diff
@@ -46,11 +46,13 @@ import (
 	"github.com/aws/karpenter-core/pkg/utils/resources"
 
 	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
-	"github.com/aws/karpenter-core/pkg/cloudprovider"
+	corecloudprovider "github.com/aws/karpenter-core/pkg/cloudprovider"
```
jonathan-innis, line 49: Why not just cloudprovider?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -65,7 +61,7 @@ var _ cloudprovider.CloudProvider = (*CloudProvider)(nil)
 
 type CloudProvider struct {
```
jonathan-innis, line 62: nit: Can we order things in the struct in the same order that they are in the return statement on L70?

pkg/providers/instance/instance.go
```diff
@@ -69,8 +71,8 @@ type InstanceProvider struct {
 	ec2Batcher             *batcher.EC2API
 }
 
-func NewInstanceProvider(ctx context.Context, region string, ec2api ec2iface.EC2API, unavailableOfferings *cache.UnavailableOfferings, instanceTypeProvider *instancetype.Provider, subnetProvider *subnet.Provider, launchTemplateProvider *launchtemplate.Provider) *InstanceProvider {
-	return &InstanceProvider{
+func NewProvider(ctx context.Context, region string, ec2api ec2iface.EC2API, unavailableOfferings *cache.UnavailableOfferings, instanceTypeProvider *instancetype.Provider, subnetProvider *subnet.Provider, launchTemplateProvider *launchtemplate.Provider) *Provider {
```
jonathan-innis, line 74: nit: newline here so this line doesn't extend so long

jonathan-innis (commented):
We should think about adding tests as a `suite_test.go` for instance package. Can you create an issue to create tests for this package so that we track it properly?

engedaam:
> We should think about adding tests as a `suite_test.go` for instance package. Can you create an issue to create tests for this package so that we track it properly?

#3514

pkg/controllers/machine/garbagecollect/suite_test.go
engedaam: Unfortunately there is a dependency between the providers. So the extra providers that are instantiated are need for the providers that the tests has a dependency on.

pkg/controllers/machine/link/suite_test.go
engedaam: Unfortunately there is a dependency between the providers. So the extra providers that are instantiated are need for the providers that the tests has a dependency on.

jonathan-innis:
Discussed offline. Are planning to create a `test.Context` function that instantiates the common context members for testing so that users don't have to individually define all the members when initializing a cloudprovider

pkg/test/environment.go
```diff
@@ -0,0 +1,152 @@
...
+	SecurityGroupProvider  *securitygroup.Provider
+	PricingProvider        *pricing.Provider
+	AMIProvider            *amifamily.Provider
+	AMIResolver            *amifamily.Resolver
+	LaunchTemplateProvider *launchtemplate.Provider
+	CloudProvider          *cloudprovider.CloudProvider
+}
+
+func NewEnvironment(ctx context.Context, env *coretest.Environment) *Environment {
+	//
```
jonathan-innis, line 68: Why the empty comment?

pkg/test/environment.go
```diff
@@ -0,0 +1,152 @@
...
+		SecurityGroupProvider:  securityGroupProvider,
+		PricingProvider:        pricingProvider,
+		AMIProvider:            amiProvider,
+		AMIResolver:            amiResolver,
+		LaunchTemplateProvider: launchTemplateProvider,
+		CloudProvider:          cloudProvider,
+	}
+}
+
+func (env *Environment) ResetCache() {
```
jonathan-innis, line 139:
```suggestion
func (env *Environment) Reset() {
```

This isn't just resetting the cache. Also, I think a short, simpler name is better

pkg/providers/amifamily/ami.go
```diff
@@ -259,3 +259,9 @@ func (p *Provider) getRequirementsFromImage(ec2Image *ec2.Image) scheduling.Requ
 	requirements.Add(scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, architecture))
 	return requirements
 }
+
+func (p *Provider) Reset() {
```
jonathan-innis, line 263: Why not pass it through? Why is this needed?

pkg/test/environment.go
```diff
@@ -0,0 +1,152 @@
...
+	// Providers
+	InstanceTypesProvider  *instancetype.Provider
+	InstanceProvider       *instance.Provider
+	SubnetProvider         *subnet.Provider
+	SecurityGroupProvider  *securitygroup.Provider
+	PricingProvider        *pricing.Provider
+	AMIProvider            *amifamily.Provider
+	AMIResolver            *amifamily.Resolver
+	LaunchTemplateProvider *launchtemplate.Provider
+	CloudProvider          *cloudprovider.CloudProvider
```
jonathan-innis, line 64: I think declaring cloudprovider here is a step too far. I think this is similar to a controller where you should let the user declare this. IMO, this isn't a low-level enough component to declare it here

jonathan-innis (commented):
Looks like you have an error generated from `make docgen` that you need to fix. Otherwise, really good work! Also, a red PR 🎉

pkg/test/environment.go
engedaam: That's fair, I'll update that

pkg/controllers/machine/garbagecollect/suite_test.go
```diff
@@ -78,15 +78,15 @@ var _ = AfterSuite(func() {
...
+	awsEnv.Reset()
 })
 
 var _ = Describe("MachineGarbageCollect", func() {
 	var instance *ec2.Instance
 	var providerID string
 
 	BeforeEach(func() {
-		awsEnv.ResetCache()
+		awsEnv.Reset()
```
jonathan-innis, line 92: If L81 does this, we don't also need to do it here

pkg/controllers/machine/link/suite_test.go
```diff
@@ -81,7 +81,7 @@ var _ = Describe("MachineLink", func() {
 	var nodeTemplate *v1alpha1.AWSNodeTemplate
 
 	BeforeEach(func() {
-		awsEnv.ResetCache()
+		awsEnv.Reset()
```
jonathan-innis, line 87: If L74 does this, we don't also need to do it here

jonathan-innis (commented):
Couple nits

jonathan-innis (approved):
LGTM 🚀

PR #3516, opened 2023-03-02
feat: support reservedENIs for vpc cni custom networking

bwagner5:
Fixes # 
https://github.com/aws/karpenter/issues/2273

**Description**
 - Support `aws.reservedENIs` in global settings to allow for proper max-pods and kube-reserved calculations when using the VPC CNI Custom Networking mode (primary CIDR for nodes; secondary CIDR for pods).
 

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


jonathan-innis (approved):
🚀 Should we think about adding an E2E test to test this new --max-pods calculation as well?

FernandoMiguel:
can some design page be written about this ?
we separate our network in this manner, by having having the nodes in one set of subnets and pods on much bigger CGNat subnets.
So having some documentation and design around this feature would be appreciated so we could best configure things on our side.
thanks

bwagner5:
> can some design page be written about this ? we separate our network in this manner, by having having the nodes in one set of subnets and pods on much bigger CGNat subnets. So having some documentation and design around this feature would be appreciated so we could best configure things on our side. thanks

I added some docs in the latest update to the PR @FernandoMiguel. Do these docs seem sufficient? I think the best reference (which I liked to in the docs) is the EKS Best Practices guide for Custom Networking https://aws.github.io/aws-eks-best-practices/networking/custom-networking

jonathan-innis (approved):
LGTM 🚀

FernandoMiguel:
much better and clear @bwagner5 .
thanks
is there anyway this could be discoverable so a manual config is not required?

bwagner5:
> much better and clear @bwagner5 . thanks is there anyway this could be discoverable so a manual config is not required?

Looked into an auto-discovery for a while, and there's simply not a great way to do it. If you have multiple subnets in an AZ, it's not clear if that should be used for Custom Networking or you're just scaling out your IPs in different subnets, or you have special routing on the subnets. Looked into discovering the VPC CNI configuration, but it's hard to detect VPC CNI from a permissions perspective and if you rename the VPC CNI or some other change to the deployment, it may not work. The pro of a manual configuration is that other CNIs or strategies can use this to tune Karpenter to the general approach without having to rely on a specific component's configuration (i..e VPC CNI vs Calico)

bwagner5 (commented):
/karpenter snapshot

bwagner5 (commented):
/karpenter snapshot

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -304,7 +304,10 @@ func (p *Provider) cachedEvictedFunc(ctx context.Context) func(string, interface
 			logging.FromContext(ctx).Errorf("Unable to delete launch template, %v", err)
 			return
 		}
-		logging.FromContext(ctx).Debugf("deleted launch template")
+		logging.FromContext(ctx).With(
+			"launch-template-id", aws.StringValue(launchTemplate.LaunchTemplateId),
```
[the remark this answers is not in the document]
jonathan-innis: You've already specified that you are deleting a launch template in the logging message so I think the scope is already defined

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -304,7 +304,10 @@ func (p *Provider) cachedEvictedFunc(ctx context.Context) func(string, interface
...
+			"launch-template-name", aws.StringValue(launchTemplate.LaunchTemplateName),
```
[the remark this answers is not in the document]
jonathan-innis: Same comment

pkg/providers/amifamily/resolver.go
```diff
@@ -113,29 +114,44 @@ func (r Resolver) Resolve(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTem
...
-			DetailedMonitoring:  aws.BoolValue(nodeTemplate.Spec.DetailedMonitoring),
-			AMIID:               amiID,
-			InstanceTypes:       instanceTypes,
+		maxPodsToInstanceTypes := map[int][]*cloudprovider.InstanceType{}
+		for _, instanceType := range instanceTypes {
+			maxPodsToInstanceTypes[int(instanceType.Capacity.Pods().Value())] = append(maxPodsToInstanceTypes[int(instanceType.Capacity.Pods().Value())], instanceType)
 		}
-		if resolved.BlockDeviceMappings == nil {
-			resolved.BlockDeviceMappings = amiFamily.DefaultBlockDeviceMappings()
+		for maxPods, instanceTypes := range maxPodsToInstanceTypes {
```
jonathan-innis, line 121: Do we want to add comments here to describe why we are doing this logic since it's not immediately intuitive to me reading it for the first time?

pkg/fake/ec2api.go
```diff
@@ -140,7 +141,7 @@ func (e *EC2API) CreateFleetWithContext(_ context.Context, input *ec2.CreateFlee
 					e.CalledWithCreateLaunchTemplateInput.Add(lt)
 				}
 				instanceState := ec2.InstanceStateNameRunning
-				for i := 0; i < int(*input.TargetCapacitySpecification.TotalTargetCapacity); i++ {
+				for ; fulfilled < int(*input.TargetCapacitySpecification.TotalTargetCapacity); fulfilled++ {
```
jonathan-innis, line 144: Why is this change needed?

pkg/cloudprovider/suite_test.go
```diff
@@ -454,7 +456,7 @@ var _ = Describe("CloudProvider", func() {
 			ExpectScheduled(ctx, env.Client, pod)
 			Expect(awsEnv.EC2API.CreateFleetBehavior.CalledWithInput.Len()).To(Equal(1))
 			input := awsEnv.EC2API.CreateFleetBehavior.CalledWithInput.Pop()
-			Expect(input.LaunchTemplateConfigs).To(HaveLen(1))
+			Expect(len(input.LaunchTemplateConfigs)).To(BeNumerically(">=", 1))
```
jonathan-innis, line 459: It seems like a lot of our testing logic relies on a single launch template and validating that that first launch template fulfills a given condition. Should we change our logic now to ensure that _all_ launch templates now fulfill this condition?

pkg/fake/ec2api.go
bwagner5: We overfulfill instances when multiple launch templates are involved. Adding more launch templates made it a little worse, so noticed it in some of the tests related to this change.

pkg/cloudprovider/suite_test.go
bwagner5: The logic should be updated everywhere to check all LTs in this PR. I've started writing some higher-level expectations  for LTs that we could use to trim the code down a bit though.

website/content/en/preview/concepts/settings.md
```diff
@@ -68,6 +68,9 @@ data:
   aws.interruptionQueueName: karpenter-cluster
   # Global tags are specified by including a JSON object of string to string from tag key to tag value
   aws.tags: '{"custom-tag1-key": "custom-tag-value", "custom-tag2-key": "custom-tag-value"}'
+  # Reserved ENIs are not included in the calculations for max-pods or kube-reserved
+  # This is most often used in the VPC CNI custom networking setup https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html
+  aws.reservedENIs: "1"
```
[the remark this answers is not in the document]
bwagner5: As discussed offline, I can't think of a use-case where you'd need to adjust this in an AWSNodeTemplate. The VPC CNI custom networking configuration would be for all nodes in the cluster unless you've deployed multiple CNIs that select different groups of nodes which is possible, but probably not widely done in practice. I'm not strictly opposed though since I like the NodeTemplate API better than throwing things in Settings.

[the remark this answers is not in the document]
jonathan-innis: We can always move it later if we decide to rip out the settings at a later date.

pkg/fake/ec2api.go
jonathan-innis: > We overfulfill instances when multiple launch templates are involved

Maybe we can discuss offline but I'm not quite sure what you mean here. You're saying that our mocked API will try to create an instance for each LT that's passed through and this causes us to mock more instances in the API than we actually need?

bwagner5: correct

pkg/cloudprovider/suite_test.go
jonathan-innis: Is that meant to be a follow-up item to this PR?

bwagner5: yep, updated

pkg/providers/instancetype/types.go
```diff
@@ -250,12 +250,13 @@ func habanaGaudis(info *ec2.InstanceTypeInfo) *resource.Quantity {
 // The number of pods per node is calculated using the formula:
 // max number of ENIs * (IPv4 Addresses per ENI -1) + 2
 // https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt#L20
-func ENILimitedPods(info *ec2.InstanceTypeInfo) *resource.Quantity {
+func ENILimitedPods(ctx context.Context, info *ec2.InstanceTypeInfo) *resource.Quantity {
 	// VPC CNI only uses the default network interface
 	// https://github.com/aws/amazon-vpc-cni-k8s/blob/3294231c0dce52cfe473bf6c62f47956a3b333b6/scripts/gen_vpc_ip_limits.go#L162
 	networkInterfaces := *info.NetworkInfo.NetworkCards[*info.NetworkInfo.DefaultNetworkCardIndex].MaximumNetworkInterfaces
+	usableNetworkInterfaces := lo.Max([]int64{(networkInterfaces - int64(awssettings.FromContext(ctx).ReservedENIs)), 0})
```
jonathan-innis, line 257: sanity check: in the case that this value is 0, I'm assuming that we error out during scheduling since none of the instance types would have sufficient resources to schedule pods.

pkg/providers/amifamily/resolver.go
```diff
@@ -127,29 +130,47 @@ func (r Resolver) Resolve(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTem
...
-				options.CABundle,
-				instanceTypes,
-				nodeTemplate.Spec.UserData,
-			),
-			BlockDeviceMappings: nodeTemplate.Spec.BlockDeviceMappings,
-			MetadataOptions:     nodeTemplate.Spec.MetadataOptions,
-			DetailedMonitoring:  aws.BoolValue(nodeTemplate.Spec.DetailedMonitoring),
-			AMIID:               amiID,
-			InstanceTypes:       instanceTypes,
+		maxPodsToInstanceTypes := map[int][]*cloudprovider.InstanceType{}
```
jonathan-innis, line 133: optional: `lo.GroupBy`?

bwagner5 (commented):
/karpenter snapshot

pkg/test/expectations.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"encoding/base64"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	. "github.com/onsi/gomega" // nolint:revive,stylecheck
+)
+
+func (env Environment) ExpectLaunchTemplatesCreatedWithUserDataContaining(substrings ...string) {
+	Expect(env.EC2API.CalledWithCreateLaunchTemplateInput.Len()).To(BeNumerically(">=", 1))
```
jonathan-innis, line 26: I'm wondering if we should couple this method to the environment as opposed to the API directly. Thoughts? Like a method on the EC2API that validates all launch templates? Or even maybe just a method to range over the input like `ForEach`

pkg/providers/instancetype/suite_test.go
```diff
@@ -1246,12 +1288,14 @@ var _ = Describe("Instance Types", func() {
...
 			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
 			ExpectScheduled(ctx, env.Client, pod)
-			Expect(awsEnv.EC2API.CalledWithCreateLaunchTemplateInput.Len()).To(Equal(1))
-			input := awsEnv.EC2API.CalledWithCreateLaunchTemplateInput.Pop()
-			Expect(*input.LaunchTemplateData.MetadataOptions.HttpEndpoint).To(Equal(ec2.LaunchTemplateInstanceMetadataEndpointStateDisabled))
-			Expect(*input.LaunchTemplateData.MetadataOptions.HttpProtocolIpv6).To(Equal(ec2.LaunchTemplateInstanceMetadataProtocolIpv6Enabled))
-			Expect(*input.LaunchTemplateData.MetadataOptions.HttpPutResponseHopLimit).To(Equal(int64(1)))
-			Expect(*input.LaunchTemplateData.MetadataOptions.HttpTokens).To(Equal(ec2.LaunchTemplateHttpTokensStateOptional))
+			Expect(awsEnv.EC2API.CalledWithCreateLaunchTemplateInput.Len()).To(BeNumerically(">=", 1))
+			for i := 0; i < awsEnv.EC2API.CalledWithCreateLaunchTemplateInput.Len(); i++ {
```
jonathan-innis, line 1292: nit: Reason you didn't use your abstracted expectation function `ExpectLaunchTemplates` here?

pkg/providers/amifamily/resolver.go
bwagner5: good callll

pkg/providers/instancetype/types.go
bwagner5: Yes, the max-pods count would actually be 2 if usableNetworkInterfaces is 0 which is correct for kube-proxy and vpc-cni. But we wouldn't launch a node that only has room for DaemonSets. I guess there are situations where kube-proxy is a static pod in which case maybe I'll just check if it's two and set it to 0 explicitly.

pkg/providers/instancetype/suite_test.go
bwagner5: missed it, I found some other spots I missed too, updated

pkg/providers/instancetype/types.go
bwagner5: updated to return 0 pods if usableNetworkInterfaces is 0.

pkg/test/expectations.go
bwagner5: hmm I don't have a strong opinion. I slightly favor the way I have it here since it matches the e2e test setup. There are differences in why it's setup like that in the e2e tests, but I kind of like the decoupling of an expectations helper on the env rather than coupling directly with the mocks.

jonathan-innis: I'm not sure that setting the expectation on the mocks makes sense but I think an iteration mechanism on the input as a general concept makes a lot of sense. WDYT?

bwagner5:
The maximum number of LTs that we could create for an instance launch is ~10 for c, m, and r instance types. This would be for small provisioning where we include the smallest instance types and all larger ones across amd64 and arm64.

```
$ for it in $(aws ec2 describe-instance-types | jq -r '.InstanceTypes[] .InstanceType' | grep -i '^[cmr]'); do ./max-pods-calculator.sh --cni-version 1.7.4 --instance-type $it >> /tmp/maxpods; done 
$ cat /tmp/maxpods | sort | uniq -c | wc -l
       7
```

So there are 7 unique max-pods buckets for c,m,and r instance types. Adding in arm64 instance types gives us not quite double that number. 

In the average case, we're only creating 3 - 5 LTs (up from our 1 - 2 LTs created today).

bwagner5 (commented):
/karpenter snapshot

bwagner5 (commented):
/karpenter snapshot

pkg/test/expectations.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"encoding/base64"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	. "github.com/onsi/gomega" // nolint:revive,stylecheck
+)
+
+func (env Environment) ExpectLaunchTemplatesCreatedWithUserDataContaining(substrings ...string) {
```
jonathan-innis, line 25: nit: Are these functions that need to be in the generic test/expectations file or can they be scoped to the launchtemplate/suite_test?

pkg/providers/instancetype/types.go
jonathan-innis: Can we add a comment around why we are explicitly setting it to 0 here?

pkg/fake/atomic.go
```diff
@@ -159,3 +159,11 @@ func (a *AtomicPtrSlice[T]) Pop() *T {
 	a.values = a.values[0 : len(a.values)-1]
 	return last
 }
+
+func (a *AtomicPtrSlice[T]) ForEach(fn func(*T)) {
+	a.mu.Lock()
```
jonathan-innis, line 164: nit: Doesn't this just need a ReadLock since you aren't actually writing against anything that's being guarded?

bwagner5:
:white_check_mark: utilization-pr-3516-px6lg tests passed
:white_check_mark: ipv6-pr-3516-t5nbq tests passed
:white_check_mark: drift-pr-3516-dvbm7 tests passed
:white_check_mark: interruption-pr-3516-c4j8t tests passed
:white_check_mark: chaos-pr-3516-5glms tests passed
:white_check_mark: consolidation-pr-3516-kh298 tests passed
:white_check_mark: machine-pr-3516-jvv9h tests passed
:white_check_mark: integration-pr-3516-slvsz tests passed

jonathan-innis (approved):
LGTM 🚀

rijoparakkal:
As per above fix we are seeing limited number of pods since it is not considering IP address delegation . example
for c5.large it is only showing max pods as 20 even though we could attach 3 enis and we can use around 60+ IPs

PR #3568, opened 2023-03-09
fix: Update InstanceType Overhead Calculation

jonathan-innis:
Fixes # 

**Description**

Updates the instance type overhead calculation by:
1. Ensuring that there are no defaults for `systemReserved` values since AL2, Ubuntu, BR and other images (including kubelet itself) does not assume defaults for system reserved values
2. Updates the eviction threshold overhead calculation to ensure that `nodefs.available` is taken into account when determining the amount of available ephemeral storage that can be leveraged by pods

* `make presubmit`
* Manual deployment and validation of allocatable on AL2, Ubuntu, and BR

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```

### DIFF
--- a/charts/karpenter/README.md
+++ b/charts/karpenter/README.md
@@ -96,3 +96,4 @@ helm upgrade --install --namespace karpenter --create-namespace \
 | topologySpreadConstraints | list | `[{"maxSkew":1,"topologyKey":"topology.kubernetes.io/zone","whenUnsatisfiable":"ScheduleAnyway"}]` | Topology spread constraints to increase the controller resilience by distributing pods across the cluster zones. If an explicit label selector is not provided one will be created from the pod selector labels. |
 | webhook.logLevel | string | `"error"` |  |
 | webhook.port | int | `8443` | The container port to use for the webhook. |
+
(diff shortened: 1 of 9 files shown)


jonathan-innis (commented):
/karpenter snapshot

jonathan-innis:
Here's the results of the overhead diff after analyzing an array of instance types and sizes:

## Before PR (0 VMMemoryOverhead)
[allocatable-diff.csv](https://github.com/aws/karpenter/files/10985550/allocatable-diff.csv)


### After PR (0 VMMemoryOverhead)
[allocatable-diff-after.csv](https://github.com/aws/karpenter/files/10985551/allocatable-diff-after.csv)

You can see that the allocatable calculation is now almost exactly correct but there needs to be some refinement of the vmOverhead estimation

bwagner5 (approved):
LGTM

tzneal:
This broke t4g.small, prior to this change:

1442111488 - Actual Node Allocatable
1486880768 - Current estimate (-44,769,280)
1382023168 - Previous estimate (60,088,320)

We're actually closer to the real allocatable value, but the problem is we now over-estimate and launch nodes that don't fit the pod.

See https://github.com/aws/karpenter/issues/3937

PR #3605, opened 2023-03-16
docs: Enrich examples on how to handle AWSNodeTemplates and Provisioners inside the chart

cest-pas-faux:
Fixes # 

**Description**

Managing the Provisioners and AWSNodeTemplate as extraObjects generate a race condition since the karpenter pods are not ready to handle calls.

`Error: Internal error occurred: failed calling webhook "defaulting.webhook.karpenter.k8s.aws": failed to call webhook: Post "[https://karpenter.karpenter.svc:443/?timeout=10s](https://karpenter.karpenter.svc/?timeout=10s)": no endpoints available for service "karpenter"`

Adding the helm hooks annotation has fixed this issue on 100% of the tests.
Not sure if you want to document this behavior elsewhere.

* Deployed on 2 production clusters

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -101,11 +101,26 @@ extraObjects: []
 #  kind: AWSNodeTemplate
 #  metadata:
 #    name: default
+#    annotations:
+#      "helm.sh/hook": "post-install, post-upgrade"
+#      "helm.sh/hook-weight": "1"
+#      "helm.sh/hook-delete-policy": before-hook-creation
 #  spec:
 #    subnetSelector:
 #      karpenter.sh/discovery: {CLUSTER_NAME}
 #    securityGroupSelector:
 #      karpenter.sh/discovery: {CLUSTER_NAME}
+#- apiVersion: karpenter.sh/v1alpha5
+#  kind: Provisioner
+#  metadata:
+#    name: default
+#    annotations:
+#      "helm.sh/hook": "post-install, post-upgrade"
+#      "helm.sh/hook-weight": "5"
+#      "helm.sh/hook-delete-policy": before-hook-creation
+#  spec:
+#    providerRef:
+#      name: default
 
 controller:
   image:
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -114,7 +114,7 @@ extraObjects: []
 #  metadata:
 #    name: default
 #    annotations:
-#      "helm.sh/hook": "post-install, post-upgrade"
+#      "helm.sh/hook": "post-install,post-upgrade"
 #      "helm.sh/hook-delete-policy": before-hook-creation
 #  spec:
 #    providerRef:
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -103,7 +103,6 @@ extraObjects: []
 #    name: default
 #    annotations:
 #      "helm.sh/hook": "post-install, post-upgrade"
-#      "helm.sh/hook-weight": "1"
 #      "helm.sh/hook-delete-policy": before-hook-creation
 #  spec:
 #    subnetSelector:
@@ -116,7 +115,6 @@ extraObjects: []
 #    name: default
 #    annotations:
 #      "helm.sh/hook": "post-install, post-upgrade"
-#      "helm.sh/hook-weight": "5"
 #      "helm.sh/hook-delete-policy": before-hook-creation
 #  spec:
 #    providerRef:
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -102,7 +102,7 @@ extraObjects: []
 #  metadata:
 #    name: default
 #    annotations:
-#      "helm.sh/hook": "post-install, post-upgrade"
+#      "helm.sh/hook": "post-install,post-upgrade"
 #      "helm.sh/hook-delete-policy": before-hook-creation
 #  spec:
 #    subnetSelector:
--- a/website/content/en/preview/troubleshooting.md
+++ b/website/content/en/preview/troubleshooting.md
@@ -139,7 +139,6 @@ If you see this issue happens while using the`extraObjects` key from the values 
     name: default
     annotations:
       "helm.sh/hook": "post-install, post-upgrade"
-      "helm.sh/hook-weight": "5"
       "helm.sh/hook-delete-policy": before-hook-creation
 ```
 
--- a/website/content/en/preview/troubleshooting.md
+++ b/website/content/en/preview/troubleshooting.md
@@ -128,6 +128,21 @@ Your security groups are not blocking you from reaching your webhook.
 This is especially relevant if you have used `terraform-eks-module` version `>=18` since that version changed its security
 approach, and now it's much more restrictive.
 
+If you see this issue happens while using the`extraObjects` key from the values file, ensure that:
+ * The helm install/upgrade command has the `--wait` flag (or `wait: true` when using helmfile)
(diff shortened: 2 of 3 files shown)


charts/karpenter/values.yaml
```diff
@@ -101,11 +101,26 @@ extraObjects: []
...
 #      karpenter.sh/discovery: {CLUSTER_NAME}
 #    securityGroupSelector:
 #      karpenter.sh/discovery: {CLUSTER_NAME}
+#- apiVersion: karpenter.sh/v1alpha5
+#  kind: Provisioner
+#  metadata:
+#    name: default
+#    annotations:
+#      "helm.sh/hook": "post-install, post-upgrade"
+#      "helm.sh/hook-weight": "5"
```
jonathan-innis, line 119:
```suggestion
#      "helm.sh/hook-weight": "1"
```

Any reason this has a different weight?

jonathan-innis (commented):
Really good find! Do you mind adding the details of adding the Helm hooks in the install section of the troubleshooting guide?

charts/karpenter/values.yaml
cest-pas-faux: Since the provisioner relies on the awsnodetemplate it made more sense for me to create the template first, and if it fails, to not continue.
That's more an OCD-compliant setting than a real issue, if you think everything should be applied at once that's fine too.

cest-pas-faux:
Done, feel free to rewrite if my english is a bit weird :)

charts/karpenter/values.yaml
jonathan-innis: I'd remove the ordering. If we can't find the node template for the provisioner we'll just ignore the provisioner until the node template gets applied

website/content/en/v0.27.0/troubleshooting.md
```diff
@@ -128,6 +128,22 @@ Your security groups are not blocking you from reaching your webhook.
 This is especially relevant if you have used `terraform-eks-module` version `>=18` since that version changed its security
 approach, and now it's much more restrictive.
 
+If this issue happens when installing karpenter for the first time and the provisionners are defined using the `extraObjects` key from the values file, ensure that :
```
jonathan-innis, line 131:
```suggestion
If you see this issue happens while using the`extraObjects` key from the values file, ensure that:
```

website/content/en/v0.27.0/troubleshooting.md
```diff
@@ -128,6 +128,22 @@ Your security groups are not blocking you from reaching your webhook.
...
+ * The helm install/upgrade command has the `--wait` flag (or `wait: true` when using helmfile)
```
jonathan-innis, line 132: Why do we need the wait flag? I thought helm just orchestrates the hooks in the background and the `wait` construct is more of a client-side thing?

website/content/en/v0.27.0/troubleshooting.md
jonathan-innis, line 131: Can you make this update in the `content/en/preview` doc rather than in a specific version? That is the version that will propagate changes between versions

cest-pas-faux: Done

website/content/en/v0.27.0/troubleshooting.md
cest-pas-faux: The wait ensures the karpenter pods are up before starting to create the helm hooks.
If the pods are not running, the webhook has no endpoints, resulting in the error message from the description.

cest-pas-faux:
All updated :
- Moved the doc from the 0.27.0 folder to preview
- Removed the ordering

charts/karpenter/values.yaml
```diff
@@ -101,11 +101,24 @@ extraObjects: []
 #  kind: AWSNodeTemplate
 #  metadata:
 #    name: default
+#    annotations:
+#      "helm.sh/hook": "post-install, post-upgrade"
```
jonathan-innis, line 105:
```suggestion
#      "helm.sh/hook": "post-install,post-upgrade"
```

nit: I think the space should be removed here

charts/karpenter/values.yaml
```diff
@@ -101,11 +101,24 @@ extraObjects: []
...
 #    subnetSelector:
 #      karpenter.sh/discovery: {CLUSTER_NAME}
 #    securityGroupSelector:
 #      karpenter.sh/discovery: {CLUSTER_NAME}
+#- apiVersion: karpenter.sh/v1alpha5
+#  kind: Provisioner
+#  metadata:
+#    name: default
+#    annotations:
+#      "helm.sh/hook": "post-install, post-upgrade"
```
jonathan-innis, line 117:
```suggestion
#      "helm.sh/hook": "post-install,post-upgrade"
```

nit: space should be removed

website/content/en/preview/troubleshooting.md
```diff
@@ -128,6 +128,20 @@ Your security groups are not blocking you from reaching your webhook.
...
+ * The helm install/upgrade command has the `--wait` flag (or `wait: true` when using helmfile)
+ * Your Provisioners and AWSNodeTemplates definitions have the proper [helm hooks annotations](https://helm.sh/docs/topics/charts_hooks/) to install them *after* the karpenter pods are running
+
+```yaml
+- apiVersion: karpenter.sh/v1alpha5
+  kind: Provisioner
+  metadata:
+    name: default
+    annotations:
+      "helm.sh/hook": "post-install, post-upgrade"
```
jonathan-innis, line 141:
```suggestion
      "helm.sh/hook": "post-install,post-upgrade"
```

jonathan-innis (commented):
Just comments on spacing, otherwise LGTM 🚀

website/content/en/v0.27.0/troubleshooting.md
jonathan-innis: Is that what `--wait` does? Have you tried this without the `--wait`? I was under the impression that it's just a client-side foreground wait vs a background run of the hooks

cest-pas-faux: I didn't try without the wait option since the wait switch is taken from the install doc [here](https://karpenter.sh/v0.27.1/getting-started/getting-started-with-karpenter/#3-create-a-cluster), should be consistent.

I'm not a helm guru and lack time to dig up the code to have the final word, but [that's the expected behavior according to the doc](https://helm.sh/docs/intro/using_helm/#helpful-options-for-installupgraderollback).

jonathan-innis: Got it, yes you are right. Otherwise, it won't wait for resource readiness before it applies the extraObjects

jonathan-innis (approved):
LGTM 🚀

sidewinder12s:
Was `before-hook-creation` tested that widely? I just added this to an existing Provisioner and NodeTemplate and it then tried to delete both on me.

jonathan-innis:
> Was before-hook-creation tested that widely? I just added this to an existing Provisioner and NodeTemplate and it then tried to delete both on me.

@sidewinder12s Yep, this definitely looks wrong. Doing the `before-hook-creation` on `post-upgrade` will cause the hook resources to get deleted whenever an upgrade is executed. This definitely needs a revert to pull out these annotations. I'm not event completely convinced anymore that these resources should be hooks given Helm's definition of hook resources.

sidewinder12s:
Luckily I only ran this on a dev environment, but it blew away almost all nodes in less than 30 seconds. I raised this in Slack but I am also going to open another ticket that I think Karpenter is way too trigger happy when deleting nodes, this is the second time I've had this type of scenario.

PR #3518, opened 2023-03-02
docs:  multiple improvements and fixes to migrate from CAS tutorial  

davivcgarcia:
Fixes #3517

**Description**

This PR contains fixes and improvements to the [Migrating from Cluster Autoscaler](https://karpenter.sh/v0.26.0/getting-started/migrating-from-cas/) documentation page.

**How was this change tested?**

This was reproduced during real world migration from CAS on a production environment.

**Does this change impact docs?**
- [X] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3630, opened 2023-03-21
docs: updated cluster name case to capital case

karadia10:
See this document with the clustername used as snake case in few places instead of capital case

Impact on code - None
Update in - Website/scripts in document

Release Note

> Website/scripts in document where the cluster name is in snake case instead of capital case which causes failed commands while running them from doc


jonathan-innis (commented):
Great changes! Thanks for the contribution! Do you mind making these changes across versions, including in the `content/en/preview` docs? These are ultimately the docs that we use to propagate changes forward

jonathan-innis:
These were fixed by #3518 and are updated in the latest version

PR #3637, opened 2023-03-22
chore: Add singleton pricing controller

### LINKED ISSUE #3459: Migrate Custom Async Reconcile Flows to Controller-Based Reconcile Flows

Both the [Pricing Provider](https://github.com/aws/karpenter/blob/08949e35ef9d493d5c8346acd43d598353eebd69/pkg/providers/pricing/pricing.go#L124) and the [Launch Template Provider](https://github.com/aws/karpenter/blob/08949e35ef9d493d5c8346acd43d598353eebd69/pkg/cloudprovider/launchtemplate.go#L78) have custom async reconcile flows to hydrate their caches on startup. In the pricing provider's case, it continually refreshes the cache.

This seems like a great fit for a standard `Reconcile` flow with the [`NewSingletonManagedBy`](https://github.com/aws/karpenter-core/blob/2fe0c7634cc43cc0321abf0a020c570400eee6e3/pkg/operator/controller/singleton.go#L38) singleton controller to reconcile and register similar to a standard controller. This unification will make the code easier to parse for new users.

An example of a similar "Singleton Controller" can be found [here](https://github.com/aws/karpenter/blob/08949e35ef9d493d5c8346acd43d598353eebd69/pkg/controllers/interruption/controller.go#L84).

N/A

N/A

### Additional Context

_No response_

### Attachments

_No response_

### Community Note

engedaam:
Fixes #3459

**Description**

The pricingProvider contains async reconcile flows to hydrate its cache. This seems like a good fit for a standard Reconcile flow.

* `make battletest`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/pricing/controller.go
+++ b/pkg/providers/pricing/controller.go
@@ -1,0 +1,49 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pricing
+
+import (
+	"context"
+
+	corecontroller "github.com/aws/karpenter-core/pkg/operator/controller"
+	"knative.dev/pkg/logging"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type Controller struct {
+	pricingProvider *Provider
+}
+
+func NewController(pricingProvider *Provider) *Controller {
+	return &Controller{
+		pricingProvider: pricingProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("pricing"))
+	c.pricingProvider.UpdatePricing(ctx)
+
+	return reconcile.Result{RequeueAfter: pricingUpdatePeriod}, nil
+}
+
+func (c *Controller) Name() string {
+	return "pricing"
+}
+
+func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontroller.Builder {
+	return corecontroller.NewSingletonManagedBy(m)
+}
--- a/pkg/providers/pricing/controller.go
+++ b/pkg/providers/pricing/controller.go
@@ -16,11 +16,14 @@ package pricing
 
 import (
 	"context"
+	"time"
 
-	corecontroller "github.com/aws/karpenter-core/pkg/operator/controller"
-	"knative.dev/pkg/logging"
+	lop "github.com/samber/lo/parallel"
+	"go.uber.org/multierr"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	corecontroller "github.com/aws/karpenter-core/pkg/operator/controller"
 )
 
 type Controller struct {
@@ -34,10 +37,7 @@ func NewController(pricingProvider *Provider) *Controller {
 }
 
 func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
-	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("pricing"))
-	c.pricingProvider.UpdatePricing(ctx)
-
-	return reconcile.Result{RequeueAfter: pricingUpdatePeriod}, nil
+	return reconcile.Result{RequeueAfter: 12 * time.Hour}, c.updatePricing(ctx)
 }
 
 func (c *Controller) Name() string {
@@ -46,4 +46,19 @@ func (c *Controller) Name() string {
 
 func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontroller.Builder {
 	return corecontroller.NewSingletonManagedBy(m)
+}
+
+func (c *Controller) updatePricing(ctx context.Context) error {
+	work := []func(ctx context.Context) error{
+		c.pricingProvider.UpdateOnDemandPricing,
+		c.pricingProvider.UpdateSpotPricing,
+	}
+	errs := make([]error, len(work))
+	lop.ForEach(work, func(f func(ctx context.Context) error, i int) {
+		if err := f(ctx); err != nil {
+			errs[i] = err
+		}
+	})
+
+	return multierr.Combine(errs...)
 }
--- a/pkg/providers/pricing/pricing.go
+++ b/pkg/providers/pricing/pricing.go
@@ -79,9 +79,6 @@ func newZonalPricing(defaultPrice float64) zonal {
 	return z
 }
 
-// pricingUpdatePeriod is how often we try to update our pricing information after the initial update on startup
-const pricingUpdatePeriod = 12 * time.Hour
-
 // NewPricingAPI returns a pricing API configured based on a particular region
 func NewAPI(sess *session.Session, region string) pricingiface.PricingAPI {
 	if sess == nil {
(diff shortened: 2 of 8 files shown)


pkg/providers/pricing/controller.go
```diff
@@ -0,0 +1,50 @@
...
+	return &Controller{
+		pricingProvider: pricingProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("pricing"))
+	c.pricingProvider.UpdatePricing(ctx)
+
+	return reconcile.Result{RequeueAfter: pricingUpdatePeriod}, nil
```
jonathan-innis, line 41: nit: Maybe just use a static value here rather than referencing a variable?

pkg/providers/pricing/controller.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func NewController(pricingProvider *Provider) *Controller {
+	return &Controller{
+		pricingProvider: pricingProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("pricing"))
```
jonathan-innis, line 38: This already occurs as part of the framework so no need for this line

pkg/providers/pricing/suite_test.go
```diff
@@ -82,14 +83,18 @@ var _ = AfterEach(func() {
 var _ = Describe("Pricing", func() {
 	It("should return static on-demand data if pricing API fails", func() {
 		awsEnv.PricingAPI.NextError.Set(fmt.Errorf("failed"))
-		p := pricing.NewProvider(ctx, awsEnv.PricingAPI, awsEnv.EC2API, "", make(chan struct{}))
+		p := pricing.NewProvider(ctx, awsEnv.PricingAPI, awsEnv.EC2API, "")
+		controller := pricing.NewController(p)
```
jonathan-innis, line 87: Is there a reason to re-create the controller every time? Why not create the controller once and just reset the data when you move between tests?

pkg/providers/pricing/pricing.go
```diff
@@ -201,7 +167,7 @@ func (p *Provider) SpotPrice(instanceType string, zone string) (float64, bool) {
 	return 0.0, false
 }
 
-func (p *Provider) updatePricing(ctx context.Context) {
+func (p *Provider) UpdatePricing(ctx context.Context) {
```
jonathan-innis, line 170: Maybe move this func into the controller portion? Also, can we do a `workqueue.ParallelizeUntil` to abstract away waitGroup logic?

jonathan-innis (commented):
Nice work! Make sure to update the `prices_gen` as well

pkg/providers/pricing/suite_test.go
engedaam: We create a new provider for each test, and I don't think it would be a good idea to expose switching out the pricing provider. That's why I just made a new controller for each test.

jonathan-innis: If I recall, I think we re-created the provider every time since we didn't have control over the async logic. I don't think we need this anymore now that it's controller based

jonathan-innis: My assumption is that we could just reset the data inside the provider and then re-reconcile the controller when we need to update the pricing at the beginning of the test

pkg/providers/pricing/controller.go
```diff
@@ -0,0 +1,63 @@
...
+func (c *Controller) Name() string {
+	return "pricing"
+}
+
+func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontroller.Builder {
+	return corecontroller.NewSingletonManagedBy(m)
+}
+
+func (c *Controller) updatePricing(ctx context.Context) {
+	workqueue.ParallelizeUntil(ctx, 10, 1, func(i int) {
```
jonathan-innis, line 54: `workqueue.ParallelizeUntil` should parallelize these two things that return errors. You can think of it like this

```go
work := []func(context.Context) error {
  c.pricingProvider.UpdateOnDemandPricing,
  c.pricingProvider.UpdateSpotPricing,
}
errs := make([]error, len(work))
workqueue.ParallizeUntil(ctx, len(work), len(work), func(i int) {
   errs[i] = work[i](ctx)
})
return multierr.Combine(errs...)
```

pkg/providers/pricing/pricing.go
```diff
@@ -201,27 +164,6 @@ func (p *Provider) SpotPrice(instanceType string, zone string) (float64, bool) {
...
-	go func() {
-		defer wg.Done()
-		if err := p.UpdateSpotPricing(ctx); err != nil {
-			logging.FromContext(ctx).Errorf("updating spot pricing, %s, using existing pricing data from %s", err, err.lastUpdateTime.Format(time.RFC3339))
-		}
-	}()
-
-	wg.Wait()
-}
-
```
jonathan-innis, line 224: You could also consider doing a `workqueue.ParalleizeUntil` on the `UpdateOnDemandPricing` section here

pkg/providers/pricing/pricing.go
```diff
@@ -451,3 +393,17 @@ func populateInitialSpotPricing(pricing map[string]float64) map[string]zonal {
...
+	// see if we've got region specific pricing data
+	staticPricing, ok := initialOnDemandPrices[p.region]
+	if !ok {
+		// and if not, fall back to the always available us-east-1
+		staticPricing = initialOnDemandPrices["us-east-1"]
+	}
+
+	p.onDemandPrices = staticPricing
+	p.spotPrices = populateInitialSpotPricing(staticPricing)
+	p.onDemandUpdateTime = initialPriceUpdate
```
jonathan-innis, line 407: Is this correct? If it's getting reset, shouldn't the update time be `time.Now()`?

pkg/providers/pricing/pricing.go
```diff
@@ -451,3 +393,17 @@ func populateInitialSpotPricing(pricing map[string]float64) map[string]zonal {
 	}
 	return m
 }
+
+func (p *Provider) Reset() {
```
jonathan-innis, line 397: Can we reuse this to populate the pricing initially so that we don't duplicate the code here?

pkg/providers/pricing/controller.go
ellistarn: Checkout Lo's parallel libraries. Easier to use imo

pkg/providers/pricing/pricing.go
engedaam: `initialPriceUpdate` it is has a preset value, which is generated with the pricing data. I don't believe that `time.Now()` would be the correct value in this case.

pkg/providers/pricing/controller.go
```diff
@@ -0,0 +1,67 @@
...
+func NewController(pricingProvider *Provider) *Controller {
+	return &Controller{
+		pricingProvider: pricingProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	err := c.updatePricing(ctx)
+
+	return reconcile.Result{RequeueAfter: 12 * time.Hour}, err
```
jonathan-innis, line 42:
```suggestion
	return reconcile.Result{RequeueAfter: 12 * time.Hour}, c.updatePricing(ctx)
```
Why not just combine L40 and L42?

pkg/providers/pricing/controller.go
```diff
@@ -0,0 +1,67 @@
...
+	return corecontroller.NewSingletonManagedBy(m)
+}
+
+func (c *Controller) updatePricing(ctx context.Context) error {
+	work := []func(ctx context.Context) *Err{
+		c.pricingProvider.UpdateOnDemandPricing,
+		c.pricingProvider.UpdateSpotPricing,
+	}
+	errs := make([]error, len(work))
+	lop.ForEach(work, func(x func(ctx context.Context) *Err, i int) {
```
jonathan-innis, line 59:
```suggestion
	lop.ForEach(work, func(f func(ctx context.Context) *Err, i int) {
```

It's a nit but shortname functions are typically represented with `f`

jonathan-innis, line 59: 🎉

pkg/providers/pricing/controller.go
```diff
@@ -0,0 +1,67 @@
...
+func (c *Controller) Name() string {
+	return "pricing"
+}
+
+func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontroller.Builder {
+	return corecontroller.NewSingletonManagedBy(m)
+}
+
+func (c *Controller) updatePricing(ctx context.Context) error {
+	work := []func(ctx context.Context) *Err{
```
jonathan-innis, line 54:
```suggestion
	work := []func(ctx context.Context) error{
```

I think `*Err` implements `error`; so, I think you can just use it

pkg/providers/pricing/controller.go
```diff
@@ -0,0 +1,67 @@
...
+		c.pricingProvider.UpdateOnDemandPricing,
+		c.pricingProvider.UpdateSpotPricing,
+	}
+	errs := make([]error, len(work))
+	lop.ForEach(work, func(x func(ctx context.Context) *Err, i int) {
+		err := x(ctx)
+		if err != nil {
```
jonathan-innis, line 61: nit: Why not combine this and just directly assign?

pkg/providers/pricing/pricing.go
```diff
@@ -97,57 +93,15 @@ func NewAPI(sess *session.Session, region string) pricingiface.PricingAPI {
...
-			case <-startAsync:
-			case <-ctx.Done():
-				return
-			}
-			// if it took many hours to be elected leader, we want to re-fetch pricing before we start our periodic
-			// polling
-			if time.Since(startup) > pricingUpdatePeriod {
-				p.updatePricing(ctx)
-			}
+	p.Reset()
```
jonathan-innis, line 103: nit: Maybe a comment here to describe why we are doing this `Reset`?

pkg/providers/pricing/pricing.go
```diff
@@ -201,37 +155,10 @@ func (p *Provider) SpotPrice(instanceType string, zone string) (float64, bool) {
...
-	// standard on-demand instances
-	var wg sync.WaitGroup
-	var onDemandPrices, onDemandMetalPrices map[string]float64
-	var onDemandErr, onDemandMetalErr error
-
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		onDemandPrices, onDemandErr = p.fetchOnDemandPricing(ctx,
+	filters := map[string][]*pricing.Filter{
```
jonathan-innis, line 159: Looking at this now, I think my initial intuition was wrong. This is more obfuscated than the original. Sticking with the waitGroups might be the best call here

jonathan-innis (approved):
Great work! Nice cleanup! 🚀

PR #3664, opened 2023-03-30
test: Fix pre-upgrade release tag pinning

jonathan-innis:
Fixes # 

**Description**

Sets a static preUpgradeVersion for the release-listener and updates this static version any time there is a new release through the `stable-pr.sh`

* Manual testing of script changes in local environment

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/hack/release/stable-pr.sh
+++ b/hack/release/stable-pr.sh
@@ -1,6 +1,31 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+# updateKarpenterCoreGoMod bumps the karpenter-core go.mod to the release version so that the
+# karpenter and karpenter-core release versions match
+updateKarpenterCoreGoMod(){
+  RELEASE_VERSION=$1
+  if [[ $GITHUB_ACCOUNT != $MAIN_GITHUB_ACCOUNT ]]; then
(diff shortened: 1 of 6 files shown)


hack/release/stable-pr.sh
```diff
@@ -1,6 +1,31 @@
...
+  cd test
+  go get -u "github.com/aws/karpenter-core@${RELEASE_VERSION}"
+  cd ..
+  make tidy
+}
+
+# updateTektonPreUpgradeVersion updates the version that we use for the pre-upgrade E2E test suite
+# so that we are constantly testing against the last minor version of Karpenter
+updateTektonPreUpgradeVersion(){
+  LAST_MINOR_VERSION=$(git tag --sort=committerdate | grep -v "v${RELEASE_VERSION_MAJOR}.${RELEASE_VERSION_MINOR}" | tail -1)
```
bwagner5, line 22: I don't think this works.

for example, if we release `v0.28.0` 

```
git tag --sort=committerdate | grep -v "v0.28" | tail -1
v0.28.0
```

And we want it to be `v0.27.1`. I think you'd want to subtract 1 from RELEASE_VERSION_MINOR, and then it should work.

PR #3668, opened 2023-03-30
chore: Allow batching calls to run in parallel

jonathan-innis:
Fixes # 

**Description**

Allows worker requests to be fired in parallel for batching requests. This reduces the time that it takes to run requests, particularly `CreateFleet` requests that were erroring and taking > 30s, blocking the batcher during this entire timeframe.

* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```


pkg/batcher/batcher.go
```diff
@@ -118,26 +130,39 @@ func (b *Batcher[T, U]) run() {
...
 	}
 }
 
 func (b *Batcher[T, U]) waitForIdle() {
 	timeout := time.NewTimer(b.options.MaxTimeout)
 	idle := time.NewTimer(b.options.IdleTimeout)
-	count := 0
-	for {
+	count := 1 // we already got a single trigger
+	for b.options.MaxItems == 0 || count < b.options.MaxItems {
```
bwagner5, line 160: This should be `<=` right? 

```suggestion
	for b.options.MaxItems == 0 || count <= b.options.MaxItems {
```

jonathan-innis: If it's less than or equal to, then we would always be over by 1. We would continue looping when count == b.options.MaxItems and then would only exit after we get triggered one more time

bwagner5 (approved):
LGTM

PR #3671, opened 2023-03-31
test: Fix issue where `MockedFunction` results would leak into other calls

jonathan-innis:
Fixes # 

**Description**

Calling any `MockedFunction` with `WithDefault()` simultaneously could result in a race where the the `defaultOutput` value would be used for both calls to the function by the time Invoke was called. This is because using the result from the `WithDefault` function wasn't an atomic operation with the `Invoke` function. 

This PR removes the `WithDefault()` call and adjust `Invoke` so that it takes in the `input` value as well as a `defaultTransformer` which transforms the input into some expected output

* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```


engedaam (approved):
LGTM

PR #3698, opened 2023-04-04
feat: Add windows support

### LINKED ISSUE #1131: Windows support

topikachu:
Fixes #1131

**Description**
This change try to integrate with the latest EKS window support.

Add two new Windows AMI Family, Windows2019 and Windows2022.
Only Core are supported OOTB. 
Windows relevant code, "kubernetes.io/os" and "vpc.amazonaws.com/PrivateIPv4Address" is only active when Windows AMI Family defined in the AwsNodeTemplate

Users can use amiSelector to choose other AMIs.
Test on a Linux and Windows mixed system for provision and de-provision.

* Test on a live system with mixed Windows and Linux workload.
* Both provision and de-provision are tested

- [ ] Yes, issue opened: # 
- [ ] No

New document is required after this change is approved.
**Release Note**

```release-note
Add Windows support. A new Windows AMI Family is introduced.
```


ellistarn:
Wow, thanks for writing this up. @jonathan-innis can you review?

topikachu:
@jonathan-innis @ellistarn 
Please tell me if more tests are required.
Not sure what kind of tests to add.

pkg/apis/v1alpha1/register.go
```diff
@@ -41,23 +41,41 @@ var (
...
+	Windows2022              = "2022"
+	WindowsCore              = "Core"
+	WindowsFull              = "Full"
+	SupportedWindowsVersions = []string{
+		Windows2019,
+		Windows2022,
+	}
+	SupportedWindowsVariants = []string{
+		WindowsCore,
+		Windows2022,
```
rawahars, line 70: I am assuming this was meant to be `WindowsFull`.

pkg/providers/amifamily/bootstrap/windows.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	kubeletExtraArgs := strings.Join([]string{w.nodeLabelArg(), w.nodeTaintArg()}, " ")
+	if kubeletExtraArgs = strings.Trim(kubeletExtraArgs, " "); len(kubeletExtraArgs) > 0 {
+		userData.WriteString(fmt.Sprintf(` -KubeletExtraArgs "%s"`, kubeletExtraArgs))
+	}
+	if w.KubeletConfig != nil && len(w.KubeletConfig.ClusterDNS) > 0 {
+		userData.WriteString(fmt.Sprintf(` -DNSClusterIP "%s"`, w.KubeletConfig.ClusterDNS[0]))
+	}
+	userData.WriteString("\n</powershell>")
```
rawahars, line 51: In addition to above, there is a parameter `ServiceCIDR` which needs to be present in the userdata. This is required in case the user is using a custom CIDR range instead of default one.

Reference: https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html#bootstrap-script-configuration-parameters

pkg/providers/amifamily/windows.go
```diff
@@ -0,0 +1,78 @@
...
+			CustomUserData:  customUserData,
+		},
+	}
+}
+
+// DefaultBlockDeviceMappings returns the default block device mappings for the AMI Family
+func (w Windows) DefaultBlockDeviceMappings() []*v1alpha1.BlockDeviceMapping {
+	return []*v1alpha1.BlockDeviceMapping{{
+		DeviceName: w.EphemeralBlockDevice(),
+		EBS:        &DefaultEBS,
```
rawahars, line 72: By default on EKS Optimized Windows AMIs, the EBS configuration is-
```
{
  "blockDeviceMappings": [
    {
      "DeviceName": "/dev/sda1",
      "Ebs": {
        "DeleteOnTermination": true,
        "Encrypted": false,
        "VolumeSize": 50,
        "VolumeType": "gp2"
      }
    }
  ]
}
```

Should we change this to make them consistent?

pkg/providers/instancetype/types.go
```diff
@@ -46,12 +46,15 @@ const (
 
 var (
 	instanceTypeScheme = regexp.MustCompile(`(^[a-z]+)(\-[0-9]+tb)?([0-9]+).*\.`)
+
+	instanceTypesSupportedForWindows = []string{"m4.16xlarge"}
```
rawahars, line 50: Windows support other instance types as well in addition to `m4.16xlarge`.

An exhaustive list would be https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/pkg/aws/vpc/limits.go.

I see the documentation is behind- https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html. Let me check with the team to test and get this part rectified.

pkg/providers/instancetype/types.go
```diff
@@ -137,6 +141,39 @@ func computeRequirements(ctx context.Context, info *ec2.InstanceTypeInfo, offeri
...
 }
 
+func getOs(info *ec2.InstanceTypeInfo, family amifamily.AMIFamily) string {
+
+	//We don't know the AMIFamily. Suppose user need a linux system.
+	if family == nil {
+		return string(v1.Linux)
+	}
+
+	if _, ok := family.(*amifamily.Windows); !ok {
```
rawahars, line 151: We should maybe do direct checks here-

- Check if the AMI family requested is Linux. If so-
  - Check architecture if needed
  - Check supported instance type
If any is true, return `Linux`.
- Check if the AMI family supported is Windows. If so-
  - Check architecture
  - Check supported instance type
If any is true, return `Windows`.

By default, Linux would be returned.

pkg/providers/instancetype/types.go
```diff
@@ -192,6 +233,23 @@ func ephemeralStorage(amiFamily amifamily.AMIFamily, blockDeviceMappings []*v1al
 	return amifamily.DefaultEBS.VolumeSize
 }
 
+func privateIPv4Address(name string) *resource.Quantity {
+	// https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html
+	// The number of pods that you can run per Windows node
+	// is equal to the number of IP addresses available per elastic network interface for the node's instance type,
+	// minus one.
+	// however, https://aws.github.io/aws-eks-best-practices/windows/docs/networking/
```
rawahars, line 241: For Windows, IPs are managed by VPC Resource controller and therefore, the limit is `Number of supported IPs - 1`. 
Reference: https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/83e4e5f88ede891e05b3d7f80a9cc0633df07567/pkg/provider/ip/provider.go#L290

Also, why do we need this value?

As new features are launched for Windows, such as multi-eni support or Prefix delegation support, this value would change based on a lot of factors and would be partially customer configurable. Such information might not be present with Karpenter. Therefore, maybe we should work around this.

rawahars (changes requested):

pkg/apis/v1alpha1/register.go
topikachu: My bad. Fixed!

pkg/providers/instancetype/types.go
topikachu: Wait for a clear rule to refine this getOS code

pkg/providers/amifamily/windows.go
topikachu: 50G is required. Let's use a similar config as other AmiFamily. for example
https://github.com/aws/karpenter/blob/07c1ebf5942e4630659a4de25f58b7b6244201cc/pkg/providers/amifamily/bottlerocket.go#L73

pkg/providers/instancetype/types.go
topikachu: Refine the code like.

```
if isWindowsAmiFamily && isSupportedWindows(instanceType) {
  return Windows
}

return Linux

func isSupportedWindows(instanceType) bool {
  //check explicated instance type
  //check architecture
  //check instance family
}

```

pkg/providers/instancetype/types.go
topikachu: 1. This is a must for Windows pod, because "vpc.amazonaws.com/PrivateIPv4Address" is a resource in pod.spec.resource.requirement. And Karpenter must define such capability for Windows pod, otherwise the pod's requirement is not met.
2. From Karpenter perspective, the capacity must minus 3, since the IP addresses are consumed by the node, CNI and kube-proxy already.

pkg/providers/amifamily/bootstrap/windows.go
topikachu: This is reasonable, however, there's no such ServiceCIDR option in Kapenter framework. Not sure how to get this config in Kapenter

pkg/providers/amifamily/windows.go
jonathan-innis: @rawahars Is there any particular reason that the volumeSize is larger for windows in the default case? Would it be fine to default to `20Gi` and `gpt3`?

pkg/providers/instancetype/types.go
```diff
@@ -137,6 +141,45 @@ func computeRequirements(ctx context.Context, info *ec2.InstanceTypeInfo, offeri
 	return requirements
 }
 
+func getOs(info *ec2.InstanceTypeInfo, family amifamily.AMIFamily) string {
```
jonathan-innis, line 144: Do we need to take a dependency on the AMIFamily here? Can we just special-case the instance types that support windows and add a `windows` value to their `os` requirement?

pkg/providers/instancetype/types.go
jonathan-innis: I'm also not clear on this. We are aware of the DS that exist in your cluster. It's possible that a user isn't running `vpc-cni` or isn't running `kube-proxy`, in which case, the `- 3` calculation would be off. This should be `- 1` as @rawahars said and Karpenter will handle the pod calculation for the DS.

pkg/providers/amifamily/windows.go
```diff
@@ -0,0 +1,82 @@
...
+
+type Windows struct {
+	DefaultFamily
+	*Options
+	template *v1alpha1.AWSNodeTemplate
+}
+
+// SSMAlias returns the AMI Alias to query SSM
+func (w Windows) SSMAlias(version string, _ *cloudprovider.InstanceType) string {
+	windowsVersion := getValueFromAnnotation(w.template.Annotations, v1alpha1.AnnoInstanceWindowsVersion, v1alpha1.Windows2019)
```
jonathan-innis, line 42: Can we just specify a single Windows version as the default version and have anyone who wants to use a custom windows version use the `AMISelector` mechanism for picking AMIs?

pkg/providers/amifamily/resolver.go
```diff
@@ -140,12 +140,14 @@ func (r Resolver) Resolve(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTem
...
 
-func GetAMIFamily(amiFamily *string, options *Options) AMIFamily {
+func GetAMIFamily(amiFamily *string, options *Options, template *v1alpha1.AWSNodeTemplate) AMIFamily {
 	switch aws.StringValue(amiFamily) {
 	case v1alpha1.AMIFamilyBottlerocket:
 		return &Bottlerocket{Options: options}
 	case v1alpha1.AMIFamilyUbuntu:
 		return &Ubuntu{Options: options}
+	case v1alpha1.AMIFamilyWindows:
+		return &Windows{Options: options, template: template}
```
jonathan-innis, line 150: If we remove the annotation mechanism, then we remove the need for `template` to be passed through here

pkg/providers/amifamily/windows.go
topikachu: That works!! I remove the code of annotations.

Users can use AMISelector to choose the right Windows version (a document may be required).

pkg/providers/instancetype/types.go
topikachu: A two-values requirement prevents Karpenter from creating the node labels.
So the test launchtemplate_test.suite_test.go `should apply labels to the node` fails at 
`Expect(node.Labels).To(HaveKey(v1.LabelOSStable)`

Can we remove this "expect" since kubelet should label node correct at runtime? We don't need any tests for these Well-Known labels from my understanding.

pkg/providers/amifamily/resolver.go
topikachu: Removed in the new code

pkg/providers/amifamily/bootstrap/windows.go
rawahars: This setting can be obtained similar to `DNSClusterIP` and is part of the response of DescribeCluster API.

@jonathan-innis can probably suggest the location from where we can plumb this value.

pkg/providers/amifamily/windows.go
rawahars: @jonathan-innis This might not be the ideal way. For Windows, the AMIs are updated monthly and therefore, if someone is using Karpenter then there would be additional overhead of figuring out the latest AMIs using the SSM parameter.

Also, each Windows version is it's own AMI family as per the customer usage. For example- Customers would have their Apps supported on Windows Server 2019 and they would hence run their workloads on Windows Server 2019 only. I don't think customers mix and match different versions of Windows in a single cluster.

As per the initial discussion with @bwagner5, we were considering one of the two approaches-
- We add `WindowsServer2019Full`, `WindowsServer2019Core`, ``WindowsServer2022Full`, and ``WindowsServer2022Core` as new AMI families. This is done by `eksctl` as well. Reference [here](https://eksctl.io/usage/schema/#nodeGroups-amiFamily). I am more inclined towards this approach.
- We make the users add annotations on pods about the target Windows version and edition. Then Karpenter would spin up the node as required for the given Windows version and edition.

What was done by @topikachu also works well in my opinion.

pkg/providers/amifamily/windows.go
rawahars: So Windows EKS AMIs do have cached container images which take substantial space. Also, Windows base images are larger than Linux ones. For example- `mcr.microsoft.com/windows:ltsc2019` is around 11.2 Gi when uncompressed.

Therefore, we keep the default size to `50Gi`. I think we can default the `VolumeType` to `gpt3`.

pkg/providers/amifamily/windows.go
topikachu: > @jonathan-innis This might not be the ideal way. For Windows, the AMIs are updated monthly and therefore, if someone is using Karpenter then there would be additional overhead of figuring out the latest AMIs using the SSM parameter.
> 
> Also, each Windows version is it's own AMI family as per the customer usage. For example- Customers would have their Apps supported on Windows Server 2019 and they would hence run their workloads on Windows Server 2019 only. I don't think customers mix and match different versions of Windows in a single cluster.
> 
> As per the initial discussion with @bwagner5, we were considering one of the two approaches-
> 
> * We add `WindowsServer2019Full`, `WindowsServer2019Core`, `` WindowsServer2022Full`, and  ``WindowsServer2022Core`as new AMI families. This is done by`eksctl` as well. Reference [here](https://eksctl.io/usage/schema/#nodeGroups-amiFamily). I am more inclined towards this approach.

I like this. In fact, I create four Windows AMIFamiliy strings at the beginning but remove them and use annotation when filing the pull request.
I can add them back.
My implementation only needs to add 4 strings but uses a single Windows class to handle them all. 

> * We make the users add annotations on pods about the target Windows version and edition. Then Karpenter would spin up the node as required for the given Windows version and edition.
> 
> What was done by @topikachu also works well in my opinion.

pkg/providers/amifamily/bootstrap/windows.go
topikachu: @rawahars 
I think this requires changing the v1alpha5/machines.go KubeletConfiguration in the karpenter core first.

If Service CIDR is required, it's better to change the core first.

pkg/providers/amifamily/windows.go
topikachu: > @rawahars Is there any particular reason that the volumeSize is larger for windows in the default case? Would it be fine to default to `20Gi` and `gpt3`?

The Windows AMI root block is 50Gi. So 20Gi is not an option to launch the instance and will be rejected by ec2.

pkg/providers/instancetype/types.go
topikachu: > * This is correct that `vpc.amazonaws.com/PrivateIPv4Address` is a required resource. How is the value calculated here used by Karpenter?
> * For Windows nodes, the IPs are not consumed by CNI and kube-proxy since they run on the node itself. I'm not sure what you mean by Karpenter perspective.


Below is a node with only one Windows pod. 3 secondary private IPv4 addresses are allocated here. One is for the pod. I guess CNI and kube-proxy consume another two.

image

topikachu:
@jonathan-innis @rawahars 
I need your suggestion about these:
1. Is it OK to give four new Windows AMI Family names? If yes, shall we use the same long name as eksctl, like WindowsServer2019CoreContainer? Or shorter ones, like WindowsServer2019Core
2. Shall we use a single value of the node requirement "kubernetes.io/os" or two values that have both Linux and Windows? If we choose two values, how to handle the default node labels created by schedule? This change may break the test `should apply labels to the node`.
3. Is there a clear rule about the ec2 instances supported for Windows workload? 

Thanks

pkg/providers/instancetype/types.go
rawahars: ah no that's not the case. Basically, at any time, VPC Resource Controller would keep 3 or so IPs warmed up on the node so that as soon as a pod is created, an IP can be allocated to it. The additional secondary IPs which you see are the IPs which have been warmed up to be allocated to the pods.

There would be one primary IP for the interface which is not used by the pods (Hence -1 in the calculation).

pkg/providers/instancetype/types.go
jonathan-innis: > two-values requirement prevents Karpenter from creating the node labels

I think these labels should constrain once the instance is launched. If it doesn't do that right now, then we should constrain and add the labels at launch-time for OS (possibly from the AMI), even if there are two values in the requirement.

I don't think constraining the requirements on the instance type to get labels to propagate on the node is the right separation of concerns though.

topikachu: Done. AMIFamily is not required.

pkg/providers/amifamily/windows.go
topikachu: Done!
Add 4 new Windows Families.

jonathan-innis:
> Is it OK to give four new Windows AMI Family names

Our general feeling is that an AMIFamily should be centered around the userData that is used to startup the node and attach it to the cluster. An AMIFamily should be a grouping of AMIs that have effectively the same configuration. With that in mind, I think it would be preferable to have `Windows` be a single AMIFamily and we can differentiate the different versions using a different mechanism.

Musing: I wonder if we can do this through the windows build number that we know will get populated to the node when the node registers with the cluster. We can support different AMI SSM aliases and have them be compatible with different sets of `node.kubernetes.io/windows-build` requirements

jonathan-innis:
> Is there a clear rule about the ec2 instances supported for Windows workload

@rawahars Can you comment on this one?

topikachu:
> Our general feeling is that an AMIFamily should be centered around the userData that is used to startup the node and attach it to the cluster. An AMIFamily should be a grouping of AMIs that have effectively the same configuration. With that in mind, I think it would be preferable to have `Windows` be a single AMIFamily and we can differentiate the different versions using a different mechanism.

It's reasonable but we need to first decide on the source of AMI.  There are two options
1. The AMI can be resolved by the AWSNodeTemplate. In this case, all parameters must be defined in this file. This can be implemented by multiple AMIFamilies, annotations, or new fields in AWSNodeTemplate
2. The AMI can be resolved by the AWSNodeTemplate and Provisioner together. User only defines a single AMI Family and declares the OS requirement in the provisioner. This needs to change both the code of provision and drifte.

All are doable, and please shed some light on this.

jonathan-innis:
> declares the OS requirement in the provisioner. This needs to change both the code of provision and drift.

I'd argue either the Provisioner _or_ the workload is the best place. At a high-level, the philosophy behind the different components is:
1. Provisioner defines a "world of possibilities" at the Cluster Admin level for a group of workloads
2. Workload-based scheduling defines the "world of possibilities" at the App Dev level for single workloads. I think in general, most users are going to be nodeSelecting on a version of windows if their workloads only support a specific version of windows.

You can imagine a scenario where a version of Windows isn't defined in the Provisioner at all and its just defined at the workload level, which will be passed into the launched machine requirements and be used to select the correct Windows AMI version for that workload

jonathan-innis:
Referencing the [Upstream Kubernetes Windows User Guide](https://kubernetes.io/docs/concepts/windows/user-guide/) for more info on the windows build version and windows best-practices

topikachu:
> Referencing the [Upstream Kubernetes Windows User Guide](https://kubernetes.io/docs/concepts/windows/user-guide/) for more info on the windows build version and windows best-practices

I'm afraid I have no idea to implement the Windows build version without AWS EKS team support.
There's no enough information to resolve the matched AMI until AWS can provide more metadata on each AMI.

I notice that the new AMIFamily `DefaultAMIs` method returns multiple AMIs and each AMI output has its own requirements. https://github.com/aws/karpenter/blob/629859c790b8410b78d9b9c134b02fe02e32165f/pkg/providers/amifamily/resolver.go#L75
It's possible that a single Windows AMI Family returns all the Windows AMIs with 2019/2022, core/full as part of the requirements.
However, it still needs some code change around https://github.com/aws/karpenter/blob/629859c790b8410b78d9b9c134b02fe02e32165f/pkg/cloudprovider/cloudprovider.go#L224 and https://github.com/aws/karpenter/blob/629859c790b8410b78d9b9c134b02fe02e32165f/pkg/providers/amifamily/resolver.go#L122 

@jonathan-innis Any idea?

topikachu:
@jonathan-innis Please check this new change.
It uses two ami families with core as default.

jonathan-innis:
@topikachu Can you rebase the change? It looks like you have a few conflicts in a few files in the PR

pkg/apis/v1alpha1/register.go
```diff
@@ -51,23 +51,41 @@ var (
...
 	}
 	SupportedContainerRuntimesByAMIFamily = map[string]sets.String{
 		AMIFamilyBottlerocket: sets.NewString("containerd"),
 		AMIFamilyAL2:          sets.NewString("dockerd", "containerd"),
 		AMIFamilyUbuntu:       sets.NewString("dockerd", "containerd"),
+		AMIFamilyWindows2019:  sets.NewString("containerd"),
+		AMIFamilyWindows2022:  sets.NewString("containerd"),
+	}
+
+	Windows2019              = "2019"
```
jonathan-innis, line 73: Do we need these values now that we are separating out the AMi families and we aren't supporting the variant or version labels?

pkg/providers/amifamily/windows.go
```diff
@@ -0,0 +1,84 @@
...
+	Version string
+	Variant string
+}
+
+func (w Windows) DefaultAMIs(version string) []DefaultAMIOutput {
+	return []DefaultAMIOutput{
+		{
+			Name:  fmt.Sprintf("windows-server-%s-%s-%s", w.Version, w.Variant, version),
+			Query: fmt.Sprintf("/aws/service/ami-windows-latest/Windows_Server-%s-English-%s-EKS_Optimized-%s/image_id", w.Version, w.Variant, version),
+			Requirements: scheduling.NewRequirements(
```
jonathan-innis, line 49: There should also be a `windows-build` requirement here since if the build number doesn't match for any deployment that is going to be joining this node, the AMI will be incompatible.

pkg/providers/amifamily/windows.go
```diff
@@ -0,0 +1,84 @@
...
+	DefaultFamily
+	*Options
+	Version string
+	Variant string
+}
+
+func (w Windows) DefaultAMIs(version string) []DefaultAMIOutput {
+	return []DefaultAMIOutput{
+		{
+			Name:  fmt.Sprintf("windows-server-%s-%s-%s", w.Version, w.Variant, version),
```
jonathan-innis, line 47: I don't know if we need the variant since we are only supporting "Core" as the default

pkg/providers/instancetype/types.go
```diff
@@ -247,10 +265,16 @@ func habanaGaudis(info *ec2.InstanceTypeInfo) *resource.Quantity {
...
 
-// The number of pods per node is calculated using the formula:
-// max number of ENIs * (IPv4 Addresses per ENI -1) + 2
-// https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt#L20
-func ENILimitedPods(info *ec2.InstanceTypeInfo) *resource.Quantity {
+func ENILimitedPods(info *ec2.InstanceTypeInfo, amiFamily amifamily.AMIFamily) *resource.Quantity {
+	// https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html
+	// The number of pods that you can run per Windows node is equal to the number of IP addresses
+	// available per elastic network interface for the node's instance type, minus one
+	if _, ok := amiFamily.(*amifamily.Windows); ok {
```
jonathan-innis, line 272: Ah, apologize. You are correct that we should have this resource capacity in a separate function. This calculation is based off of the calculation here: https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/ecbd6965a0100d9a070110233762593b16023287/pkg/provider/ip/provider.go#L297.

Can you link the ref to the code where this is determined in the resource initialization? I do agree that we should do the resource calculation for `PrivateIPv4` address in a different function

pkg/providers/instancetype/types.go
```diff
@@ -335,13 +359,16 @@ func evictionThreshold(memory *resource.Quantity, storage *resource.Quantity, am
 
 func pods(ctx context.Context, info *ec2.InstanceTypeInfo, amiFamily amifamily.AMIFamily, kc *v1alpha5.KubeletConfiguration) *resource.Quantity {
 	var count int64
+	_, isWindows := amiFamily.(*amifamily.Windows)
 	switch {
+	case isWindows:
+		count = ENILimitedPods(info, amiFamily).Value()
 	case kc != nil && kc.MaxPods != nil:
```
jonathan-innis, line 366: Are maxPods and ENILimitedPodDensity not supported by default in the default windows install?

pkg/providers/amifamily/windows.go
topikachu: Used as const to create SSM query.

pkg/providers/amifamily/windows.go
topikachu: keep for the future.

pkg/providers/instancetype/types.go
topikachu: I don't think so.
Neither [custom networking](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html) nor [IP prefixes](https://docs.aws.amazon.com/eks/latest/userguide/cni-increase-ip-addresses.html) is supported

pkg/providers/instancetype/types.go
topikachu: Done

pkg/apis/v1alpha1/register.go
topikachu: Used as const to create SSM query

pkg/apis/v1alpha1/register.go
```diff
@@ -55,23 +55,48 @@ var (
...
 		AMIFamilyUbuntu,
+		AMIFamilyWindows2019,
+		AMIFamilyWindows2022,
 		AMIFamilyCustom,
 	}
 	SupportedContainerRuntimesByAMIFamily = map[string]sets.String{
 		AMIFamilyBottlerocket: sets.NewString("containerd"),
 		AMIFamilyAL2:          sets.NewString("dockerd", "containerd"),
 		AMIFamilyUbuntu:       sets.NewString("dockerd", "containerd"),
+		AMIFamilyWindows2019:  sets.NewString("containerd"),
```
rawahars, line 73: On Windows, similar to AL2, both `dockerd` and `containerd` runtimes are supported but the same are based on the EKS version.

As discussed with @jonathan-innis, the container runtime validation is obsolete code. Even so, Should we make Windows consistent with AL2?

jonathan-innis: Yes, I think let's pull it out or do the validation in a separate PR. @topikachu We should have both container runtimes supported here for both windows versions.

pkg/apis/v1alpha1/register.go
jonathan-innis: optional: If they're only being referenced in a single place, I'd prefer to not have them as constants in this file, but I'd also take the current state as-is

pkg/apis/v1alpha1/register.go
topikachu: Fixed.
Add both dockerd and containerd

pkg/providers/amifamily/bootstrap/windows.go
```diff
@@ -0,0 +1,79 @@
...
+	Options
+}
+
+// nolint:gocyclo
+func (w Windows) Script() (string, error) {
+	var userData bytes.Buffer
+	userData.WriteString("<powershell>\n")
+	userData.WriteString("[string]$EKSBootstrapScriptFile = \"$env:ProgramFiles\\Amazon\\EKS\\Start-EKSBootstrap.ps1\"\n")
+	userData.WriteString(fmt.Sprintf(`& $EKSBootstrapScriptFile -EKSClusterName "%s" -APIServerEndpoint "%s"`, w.ClusterName, w.ClusterEndpoint))
+	if w.CABundle != nil {
```
jonathan-innis, line 40: Can we support the `-ContainerRuntime` argument since it's supported by the Windows AMI: https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html?

pkg/providers/amifamily/bootstrap/windows.go
```diff
@@ -0,0 +1,79 @@
...
+		userData.WriteString(fmt.Sprintf(" -Base64ClusterCA \"%s\"", *w.CABundle))
+	}
+	kubeletExtraArgs := strings.Join([]string{w.nodeLabelArg(), w.nodeTaintArg()}, " ")
```
jonathan-innis, line 43: Can the `kubeletExtraArgs` here pass the rest of the Provisioner's `spec.kubeletConfiguration` through? For an example of this, you could look at: https://github.com/aws/karpenter/blob/main/pkg/providers/amifamily/bootstrap/eksbootstrap.go#L90

pkg/providers/amifamily/bootstrap/windows.go
```diff
@@ -0,0 +1,79 @@
...
+		userData.WriteString(fmt.Sprintf(` -KubeletExtraArgs "%s"`, kubeletExtraArgs))
+	}
+	if w.KubeletConfig != nil && len(w.KubeletConfig.ClusterDNS) > 0 {
+		userData.WriteString(fmt.Sprintf(` -DNSClusterIP "%s"`, w.KubeletConfig.ClusterDNS[0]))
+	}
+	userData.WriteString("\n</powershell>")
+	return base64.StdEncoding.EncodeToString(userData.Bytes()), nil
+}
+
+func (w Windows) nodeTaintArg() string {
```
jonathan-innis, line 54: nit: Can this function look like the updated version of the one in the `eksbootstrap.go` file?

pkg/providers/amifamily/bootstrap/windows.go
```diff
@@ -0,0 +1,79 @@
...
+	var taintStrings []string
+	var once sync.Once
+	for _, taint := range w.Taints {
+		once.Do(func() { nodeTaintsArg = "--register-with-taints=" })
+		taintStrings = append(taintStrings, fmt.Sprintf("%s=%s:%s", taint.Key, taint.Value, taint.Effect))
+	}
+	return fmt.Sprintf("%s%s", nodeTaintsArg, strings.Join(taintStrings, ","))
+}
+
+func (w Windows) nodeLabelArg() string {
```
jonathan-innis, line 65: nit: Can this function look like the updated version of the one in the `eksbootstrap.go` file?

pkg/providers/instancetype/suite_test.go
```diff
@@ -179,8 +181,9 @@ var _ = Describe("Instance Types", func() {
 			"topology.ebs.csi.aws.com/zone": "test-zone-1a",
 		}
 
+		expected := v1alpha5.WellKnownLabels.Clone().Delete(v1.LabelWindowsBuild)
```
jonathan-innis, line 184: Any reason this label needs to be pulled out and can't be put in the label set for the test case?

pkg/providers/instancetype/suite_test.go
```diff
@@ -229,12 +232,13 @@ var _ = Describe("Instance Types", func() {
...
 		// Ensure that we're exercising all well known labels except for accelerator labels
-		expectedLabels := append(v1alpha5.WellKnownLabels.Difference(sets.NewString(
-			v1alpha1.LabelInstanceAcceleratorCount,
-			v1alpha1.LabelInstanceAcceleratorName,
-			v1alpha1.LabelInstanceAcceleratorManufacturer,
-		)).UnsortedList(), lo.Keys(v1alpha5.NormalizedLabels)...)
-		Expect(lo.Keys(nodeSelector)).To(ContainElements(expectedLabels))
+		Expect(lo.Keys(nodeSelector)).To(ContainElements(
+			append(
+				v1alpha5.WellKnownLabels.Clone().Delete(v1.LabelWindowsBuild).Difference(sets.NewString(
```
jonathan-innis, line 237: I think you might be able to throw this in the bucket with the other labels like 

```go
expectedLabels := append(v1alpha5.WellKnownLabels.Difference(sets.NewString(
			v1alpha1.LabelInstanceAcceleratorCount,
			v1alpha1.LabelInstanceAcceleratorName,
			v1alpha1.LabelInstanceAcceleratorManufacturer,
                         v1.LabelWindowsBuild,
		)).UnsortedList(), lo.Keys(v1alpha5.NormalizedLabels)...)
```

pkg/providers/instancetype/suite_test.go
```diff
@@ -276,7 +280,7 @@ var _ = Describe("Instance Types", func() {
 		}
 
 		// Ensure that we're exercising all well known labels except for gpu labels and nvme
-		expectedLabels := append(v1alpha5.WellKnownLabels.Difference(sets.NewString(
+		expectedLabels := append(v1alpha5.WellKnownLabels.Clone().Delete(v1.LabelWindowsBuild).Difference(sets.NewString(
```
jonathan-innis, line 283: Same comment here about putting it in the same bucket as the other labels that we are excluding here

pkg/providers/amifamily/windows.go
```diff
@@ -0,0 +1,85 @@
...
+
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
+	"github.com/aws/karpenter-core/pkg/cloudprovider"
+
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+)
+
+type Windows struct {
```
jonathan-innis, line 37: nit: Does the variant change? I think we might be able to hard-code in "Core"

pkg/apis/v1alpha1/register.go
```diff
@@ -55,23 +55,48 @@ var (
...
+		AMIFamilyWindows2019:  sets.NewString("dockerd", "containerd"),
+		AMIFamilyWindows2022:  sets.NewString("dockerd", "containerd"),
+	}
+
+	Windows2019              = "2019"
+	Windows2022              = "2022"
+	WindowsCore              = "Core"
+	Windows2019Build         = "10.0.17763"
+	Windows2022Build         = "10.0.20348"
+	SupportedWindowsVersions = []string{
```
jonathan-innis, line 82: nit: How are these "Supported" values being used. I'm not sure if anything from L82-L92 is being used

jonathan-innis (commented):
Looking good! Just some high-level nits

pkg/providers/instancetype/types.go
topikachu: I'm not sure if we can remove this.
I find the v1.ResourcePods is used a lot in the core Karpenter and it is computed by ENILimitedPods.
If we remove this, the ResourcePods size is larger than the PrivateIPv4Address. I'm not sure if Karpenter will do something wrong in this case.

Please confirm again, if you think it's safe to remove this.

pkg/providers/instancetype/types.go
```diff
@@ -257,6 +282,19 @@ func ENILimitedPods(ctx context.Context, info *ec2.InstanceTypeInfo) *resource.Q
 	return resources.Quantity(fmt.Sprint(usableNetworkInterfaces*(addressesPerInterface-1) + 2))
 }
 
+func privateIPv4Address(info *ec2.InstanceTypeInfo, amiFamily amifamily.AMIFamily) *resource.Quantity {
+	//https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/ecbd6965a0100d9a070110233762593b16023287/pkg/provider/ip/provider.go#L297
+
+	var capacity int64
+	if _, ok := amiFamily.(*amifamily.Windows); ok {
+		capacity = aws.Int64Value(info.NetworkInfo.Ipv4AddressesPerInterface) - 1
```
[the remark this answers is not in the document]
topikachu: Exactly.
I don't want to introduce something new to the Linux node.
Only Windows requires this.

pkg/providers/instancetype/suite_test.go
topikachu: I now understand the “Difference” method now. :)
Will fix the test code.

pkg/providers/instancetype/types.go
topikachu: I'm not sure if we can remove this.
I find the v1.ResourcePods is used a lot in the core Karpenter and it is computed by ENILimitedPods.
If we remove this, the ResourcePods size is larger than the PrivateIPv4Address. I'm not sure if Karpenter will do something wrong in this case.

Please confirm again, if you think it's safe to remove this.

pkg/providers/instancetype/suite_test.go
```diff
@@ -179,8 +181,11 @@ var _ = Describe("Instance Types", func() {
 			"topology.ebs.csi.aws.com/zone": "test-zone-1a",
 		}
 
+		expected := v1alpha5.WellKnownLabels.Difference(sets.NewString(
```
jonathan-innis, line 184: I don't think we need to remove the WindowsBuild requirement here. I think we should definitely test that label if we are saying that it's a `WellKnownLabel`

pkg/providers/instancetype/types.go
```diff
@@ -132,6 +132,13 @@ func computeRequirements(ctx context.Context, info *ec2.InstanceTypeInfo, offeri
 	return requirements
 }
 
+func getOS(info *ec2.InstanceTypeInfo, amiFamily amifamily.AMIFamily) string {
+	if _, ok := amiFamily.(*amifamily.Windows); ok && getArchitecture(info) == v1alpha5.ArchitectureAmd64 {
```
jonathan-innis, line 136: Does this make sense? i.e. If the family is Windows, I think we should return:
1. Windows if the architecture is `amd64`
2. Empty string (no value) if it's anything else (meaning that windows doesn't support it)
In the case that the family isn't windows, then we always return `linux`

pkg/providers/instancetype/types.go
```diff
@@ -184,6 +196,12 @@ func ephemeralStorage(amiFamily amifamily.AMIFamily, blockDeviceMappings []*v1al
 			}
 		}
 	}
+	//Return the ephemeralBlockDevice size if defined in ami
+	if ephemeralBlockDevice, ok := lo.Find(amiFamily.DefaultBlockDeviceMappings(), func(item *v1alpha1.BlockDeviceMapping) bool {
```
jonathan-innis, line 200: I'm not sure that this is correct. I think we need to get the last `DefaultBlockDeviceMapping` to consider that one for the ephemeral-storage capacity. In general, we are taking the last blockDeviceMapping to be the ephemeral block device.

pkg/providers/instancetype/types.go
jonathan-innis: We should validate it, but, in general, we should trust that a resource capacity cannot be exceeded. I would expect that we should see pods (which should have a privateIPv4Address resource injected into them by the vpc-resource-controller mutating webhook) and we should get constrained on the instance type by this resource before pods

pkg/providers/instancetype/types.go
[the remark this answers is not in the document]
jonathan-innis: But you are returning a value when the AMIFamily isn't windows. Does that make sense? Maybe `privateIPv4Address` can always return `aws.Int64Value(info.NetworkInfo.Ipv4AddressesPerInterface) - 1` and then we can control whether the instanceType has this value at the top-level on L165 where you already gated this resource on amiFamily

pkg/providers/instancetype/types.go
jonathan-innis: I just tested this by spinning up a few Windows nodes. Looks like `eniLimitedPodDensity` isn't a concept in Windows; so, a Windows node always has `--max-pods` set to 110 by default. `spec.kubeletConfiguration.maxPods` still applies because I can pass this through to `-KubeletExtraArgs` and I am able to change the pod capacity value. 

With that in mind, I think what would make the most sense is to add a `amiFamily.FeatureFlags()` that tells us whether we support `ENILimiatedPodDensity`:

```go
func pods(ctx context.Context, info *ec2.InstanceTypeInfo, amiFamily amifamily.AMIFamily, kc *v1alpha5.KubeletConfiguration) *resource.Quantity {
	var count int64
	switch {
	case kc != nil && kc.MaxPods != nil:
		count = int64(ptr.Int32Value(kc.MaxPods))
        case awssettings.FromContext(ctx).EnableENILimitedPodDensity && amiFamily.FeatureFlags().SupportsENILimitedPodDensity():
                count = ENILimitedPods(ctx, info).Value()
        default:
		count = 110
	}
	if kc != nil && ptr.Int32Value(kc.PodsPerCore) > 0 && amiFamily.FeatureFlags().PodsPerCoreEnabled {
		count = lo.Min([]int64{int64(ptr.Int32Value(kc.PodsPerCore)) * ptr.Int64Value(info.VCpuInfo.DefaultVCpus), count})
	}
	return resources.Quantity(fmt.Sprint(count))
}
```

pkg/providers/instancetype/suite_test.go
topikachu: Well,It's  
v1.LabelOSStable:                 "linux",

If nobody concerns the mixing of Linux with Windows Build together, I can add it back.

pkg/providers/instancetype/types.go
topikachu: Yours is more accurate. :)

pkg/providers/instancetype/types.go
topikachu: So, what's the purpose of AMIFamily.EphemeralBlockDevice?
Is it the device name of the ephemeral storage?

pkg/providers/instancetype/types.go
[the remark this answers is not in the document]
topikachu: My bad. How about return 0 if not windows?

pkg/providers/amifamily/bootstrap/windows.go
topikachu: Will add this parameter

pkg/providers/amifamily/bootstrap/windows.go
topikachu: Will align with the Linux nodes.

pkg/providers/amifamily/bootstrap/windows.go
topikachu: Will align with the Linux nodes.

pkg/providers/instancetype/types.go
topikachu: Removed

pkg/providers/instancetype/types.go
topikachu: fixed

pkg/providers/amifamily/bottlerocket.go
```diff
@@ -141,5 +141,6 @@ func (b Bottlerocket) FeatureFlags() FeatureFlags {
 		UsesENILimitedMemoryOverhead: false,
 		PodsPerCoreEnabled:           false,
 		EvictionSoftEnabled:          false,
+		SupportsENILimitedPodDensity: true,
```
jonathan-innis, line 144: nit: I'm not sure that this is needed if it uses the default value

pkg/providers/amifamily/bootstrap/bootstrap.go
```diff
@@ -15,7 +15,17 @@ limitations under the License.
 package bootstrap
```
jonathan-innis, line 15: nit: label this file `args.go` since i has to do with passing args to the bootstrap scripts

pkg/providers/amifamily/bootstrap/eksbootstrap.go
```diff
@@ -81,80 +75,12 @@ func (e EKS) eksBootstrapScript() string {
 	if (e.KubeletConfig != nil && e.KubeletConfig.MaxPods != nil) || !e.AWSENILimitedPodDensity {
 		userData.WriteString(" \\\n--use-max-pods false")
 	}
-	if args := e.kubeletExtraArgs(); len(args) > 0 {
+	if args := e.asKubeletExtraArgs(); len(args) > 0 {
```
jonathan-innis, line 78:
```suggestion
	if args := e.kubeletExtraArgs(); len(args) > 0 {
```

nit: why change this name?

pkg/providers/amifamily/bootstrap/windows.go
```diff
@@ -0,0 +1,53 @@
...
+// nolint:gocyclo
+func (w Windows) Script() (string, error) {
+	var userData bytes.Buffer
+	userData.WriteString("<powershell>\n")
+	userData.WriteString("[string]$EKSBootstrapScriptFile = \"$env:ProgramFiles\\Amazon\\EKS\\Start-EKSBootstrap.ps1\"\n")
+	userData.WriteString(fmt.Sprintf(`& $EKSBootstrapScriptFile -EKSClusterName '%s' -APIServerEndpoint '%s'`, w.ClusterName, w.ClusterEndpoint))
+	if w.CABundle != nil {
+		userData.WriteString(fmt.Sprintf(` -Base64ClusterCA '%s'`, *w.CABundle))
+	}
+	if w.ContainerRuntime != "" {
```
jonathan-innis, line 38: It looks like you have `w.ContainerRuntime` twice. Any reason to specify it twice? I _think_ you only need to have `w.ContainerRuntime`.

pkg/providers/amifamily/windows.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (w Windows) EphemeralBlockDevice() *string {
+	return aws.String("/dev/sda1")
+}
+
+func (w Windows) FeatureFlags() FeatureFlags {
+	return FeatureFlags{
+		UsesENILimitedMemoryOverhead: false,
+		PodsPerCoreEnabled:           false,
```
jonathan-innis, line 94: I think it supports `PodsPerCore` and `EvictionSoft`. You can try it, but I believe I passed these into the `kubeletExtraArgs` and they worked correctly.

pkg/providers/instancetype/suite_test.go
```diff
@@ -177,10 +177,12 @@ var _ = Describe("Instance Types", func() {
 			"beta.kubernetes.io/os":         "linux",
 			v1.LabelInstanceType:            "g4dn.8xlarge",
 			"topology.ebs.csi.aws.com/zone": "test-zone-1a",
+			v1.LabelWindowsBuild:            "some-build",
 		}
 
+		expected := v1alpha5.WellKnownLabels
```
jonathan-innis, line 183: nit: any reason to re-label this? Why not just leave this as-is?

pkg/providers/instancetype/types.go
```diff
@@ -132,6 +132,17 @@ func computeRequirements(ctx context.Context, info *ec2.InstanceTypeInfo, offeri
 	return requirements
 }
 
+func getOS(info *ec2.InstanceTypeInfo, amiFamily amifamily.AMIFamily) string {
```
jonathan-innis, line 135: I think we need to process this requirement a bit differently. If we return an empty string here e.g. `""`. That means that the scheduling requirement says that the instance type supports an OS with the value of `""`. Rather, I think what we want to say is that it doesn't support this OS at all.

We may need to modify this function to return the OS scheduling requirement like

```go
func getOSRequirement(info *ec2.InstanceTypeInfo, amiFamily amifamily.AMIFamily) string {
   if _, ok := amiFamily.(*amifamily.Windows); ok {
	if getArchitecture(info) == v1alpha5.ArchitectureAmd64 {
		return scheduling.NewRequirement(v1.LabelOSStable, v1.OperatorIn, string(v1.Windows)
	}
        return scheduling.NewRequirement(v1.LabelOSStable, DoesNotExist)
  }
  return scheduling.NewRequirement(v1.LabelOSStable, v1.OperatorIn, string(v1.Linux))
}

File: pkg/providers/instancetype/types.go (hunk 28 above)
Comment (jonathan-innis) [reply pkg/providers/instancetype/types.go:204]: Yep, you are completely right. I didn't realize that this was a re-write of existing code. Good call

File: pkg/providers/instancetype/types.go [hunk 36]
```diff
@@ -337,10 +367,11 @@ func pods(ctx context.Context, info *ec2.InstanceTypeInfo, amiFamily amifamily.A
 	switch {
 	case kc != nil && kc.MaxPods != nil:
 		count = int64(ptr.Int32Value(kc.MaxPods))
-	case !awssettings.FromContext(ctx).EnableENILimitedPodDensity:
-		count = 110
-	default:
+	case awssettings.FromContext(ctx).EnableENILimitedPodDensity && amiFamily.FeatureFlags().SupportsENILimitedPodDensity:
```
jonathan-innis, line 370: optional: Is there a good way to add a test for these changes?

pkg/apis/v1alpha1/register.go
```diff
@@ -55,23 +55,36 @@ var (
...
 		AMIFamilyAL2:          sets.NewString("dockerd", "containerd"),
 		AMIFamilyUbuntu:       sets.NewString("dockerd", "containerd"),
+		AMIFamilyWindows2019:  sets.NewString("dockerd", "containerd"),
+		AMIFamilyWindows2022:  sets.NewString("dockerd", "containerd"),
 	}
+
+	Windows2019                                           = "2019"
+	Windows2022                                           = "2022"
+	WindowsCore                                           = "Core"
+	Windows2019Build                                      = "10.0.17763"
```
jonathan-innis, line 80: Just checking: This value isn't defined anywhere in `corev1` in any upstream file. It would be nice if this mapped to an upstream constant, but, if not, that's also fine. Doesn't hurt to check

pkg/providers/amifamily/bottlerocket.go
topikachu: Does Bottlerocket support ENILimitedPodDensity?

pkg/providers/amifamily/bootstrap/windows.go
topikachu: My bad. I had a brain fart at that time and wrote the wrong code. :(

pkg/providers/instancetype/suite_test.go
topikachu: You are right.
I should revert to its original version and only add windows build in the above code.

pkg/providers/instancetype/types.go
topikachu: sure!

pkg/providers/amifamily/bootstrap/eksbootstrap.go
topikachu: I try to avoid misleading that Options has a kubeletExtraArgs member/copy.
This kubeletExtraArgs is computed from other members, so use the "as" prefix here.

pkg/providers/amifamily/windows.go
topikachu: Check windows kubelet and it supports these parameters.
fixed.

pkg/apis/v1alpha1/register.go
topikachu: No such const in the upstream project. :( 
Have to add by ourselves.

pkg/providers/amifamily/bootstrap/bootstrap.go
topikachu: ok.

pkg/providers/instancetype/types.go
topikachu: Add one new test!

rawahars (approved):

pkg/providers/amifamily/bootstrap/args.go
```diff
@@ -0,0 +1,119 @@
...
+		args = append(args, fmt.Sprintf("--cpu-cfs-quota=%t", lo.FromPtr(o.KubeletConfig.CPUCFSQuota)))
+	}
+	return lo.Compact(args)
+}
+func (o Options) maxPodsArg() string {
+	if o.KubeletConfig != nil && o.KubeletConfig.MaxPods != nil {
+		return fmt.Sprintf("--max-pods=%d", ptr.Int32Value(o.KubeletConfig.MaxPods))
+	}
+	if !o.AWSENILimitedPodDensity {
+		return "--max-pods=110"
```
rawahars, line 81: So if the customer doesn't specify `max-pods`, it would be set as `110` right?
If so, this is in contrast with Windows where the default would be `500` going forward.

pkg/providers/amifamily/bootstrap/args.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (o Options) asKubeletExtraArgs() (args []string) {
+	args = append(args, o.nodeLabelArg(), o.nodeTaintArg(), o.maxPodsArg())
+
+	if o.KubeletConfig == nil {
+		return lo.Compact(args)
+	}
+	if o.KubeletConfig.PodsPerCore != nil {
+		args = append(args, fmt.Sprintf("--pods-per-core=%d", ptr.Int32Value(o.KubeletConfig.PodsPerCore)))
```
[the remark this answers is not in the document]
jonathan-innis: I think `pods-per-core` should still work then? If someone sets `pods-per-core` and then sets a higher `max-pods` value, this should all still work like normal, right?

pkg/providers/amifamily/bootstrap/args.go
```diff
@@ -0,0 +1,119 @@
...
+	if o.KubeletConfig.CPUCFSQuota != nil {
+		args = append(args, fmt.Sprintf("--cpu-cfs-quota=%t", lo.FromPtr(o.KubeletConfig.CPUCFSQuota)))
+	}
+	return lo.Compact(args)
+}
+func (o Options) maxPodsArg() string {
+	if o.KubeletConfig != nil && o.KubeletConfig.MaxPods != nil {
+		return fmt.Sprintf("--max-pods=%d", ptr.Int32Value(o.KubeletConfig.MaxPods))
+	}
+	if !o.AWSENILimitedPodDensity {
```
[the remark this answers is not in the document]
jonathan-innis: `AWSENILimitedPodDensity` is an option in the `karpenter-global-settings` that applies more towards other AMIFamilies than the Windows family. Basically, some AMIs, like EKS-optimized AMI, bases their `--max-pods` value based on the number of network interfaces combined with the number of IPs supported per interface. This is used to set `--max-pods` in the bootstrap script

pkg/providers/amifamily/bootstrap/args.go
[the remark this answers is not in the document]
rawahars: Yes this is correct. This would work like normal.

pkg/providers/amifamily/bootstrap/args.go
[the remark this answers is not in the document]
rawahars: ah got it. I see that this is set as `false` for Windows. Therefore, this shouldn't impact Windows.

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis:
@topikachu Wanted to update you here. Everything is passing on our E2Es 🎉 We're holding this change until our upcoming stable release but should take it shortly after. Nice work 🥇

andrewleech:
I would like to extend my huge thanks to @topikachu ! Great work and diligence getting this all working and looking fantastic.

I've been eagerly awaiting windows support in karpenter myself to use as on-demand CI worker nodes. My existing Linux ones work great, it'll be fantastic to add windows jobs to my builds too!

cudders:
Many thanks for adding this support, we've been following this closely.

I've used the following to install (here illustrative of the version used)

"helm upgrade --install karpenter oci://public.ecr.aws/karpenter/karpenter --version 'v0-c990a2d9fb10c1bfeffd5c6af64bf8575536d67e' --namespace karpenter --create-namespace <other params>"

This successfully added a windows node to the cluster on EKS version 1.24.

Be great if this pull request could be completed?

rawahars (approved):
Thanks @topikachu for putting this together. Awesome work!

topikachu:
> We will perform some additional testing from Windows perspective and we should be ready to merge very soon.
> 
> Also, there are a couple of enhancements which we will undertake as a fast followup to this PR-
> - Adding support for automatically plumbing `ServiceCIDR` which sets up the required networking configuration for service endpoints
> - Allowing merge of custom user data with the required user data. Ref: https://karpenter.sh/preview/concepts/node-templates/#specuserdata
> - Add support for `ExcludedSnatCIDRs` which disables SNAT for certain CIDRs. Ref: https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html#bootstrap-script-configuration-parameters

About ServiceCIDR and ExcludedSnatCIDRs, how user defines these two parameters? In the node template?

About userdata, I have no idea how to merge windows power shell script. As I tested, Windows node doesnt support multiple scripts in userdata.

rawahars:
@topikachu `ServiceCIDR` and `ExcludedSnatCIDRs` parameters are passed on to the bootstrap script. Reference: https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html#bootstrap-script-configuration-parameters

`ServiceCIDR` can be fetched using DescribeCluster call. As discussed earlier in this PR, that would need changes in Core as well. `ExcludedSnatCIDRs` is passed manually by the customers. We will make changes to ensure that we can set it via Karpenter as well.

Userdata merge can possibly happen when we generate the script i.e. between `<poweshell> </powershell>`. We can take raw string as user input and prefix it to the generated script.

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

cudders:
Apologies for chasing but do we have an ETA on when this is an official release please? We're keen to get this into our non production EKS world where we have quite a lot of windows workloads.
Thanks again for implementing this.

rawahars:
@cudders We are really close to wrapping this up. Optimistically, we should be done with our internal processes/testing in the next couple of weeks.

cudders:
> @cudders We are really close to wrapping this up. Optimistically, we should be done with our internal processes/testing in the next couple of weeks.

Brilliant - thanks

jonathan-innis:
E2Es are passing

:white_check_mark: ipv6-pr-3698-bwwq7 tests passed
:white_check_mark: drift-pr-3698-k7q8q tests passed
:white_check_mark: utilization-pr-3698-bzmq9 tests passed
:white_check_mark: interruption-pr-3698-mxg2x tests passed
:white_check_mark: chaos-pr-3698-ndqkd tests passed
:white_check_mark: consolidation-pr-3698-cw7vc tests passed
:white_check_mark: upgrade-pr-3698-n4scb tests passed
:white_check_mark: integration-pr-3698-986mc tests passed

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (approved):
LGTM 🚀

bwagner5 (approved):
lgtm

rawahars (approved):
LGTM :ship:

topikachu:
Thank you for accepting my code change. I'm excited that the addition of Windows support to Karpeneter will benefit all users.

If there are any further modifications needed, please let me know. I'm eager to contribute more to the Karpeneter project and assist in any way I can.

Thanks again for the opportunity.

PR #3700, opened 2023-04-04
chore: Default AMI Requirements

engedaam:
Fixes # 

**Description**
- Separating out the fetching of AMIs and the mapping of AMI to instance types

* `make battletest`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/apis/v1alpha1/register.go
+++ b/pkg/apis/v1alpha1/register.go
@@ -53,11 +53,13 @@ var (
 		AMIFamilyAL2:          sets.NewString("dockerd", "containerd"),
 		AMIFamilyUbuntu:       sets.NewString("dockerd", "containerd"),
 	}
-	ResourceNVIDIAGPU   v1.ResourceName = "nvidia.com/gpu"
-	ResourceAMDGPU      v1.ResourceName = "amd.com/gpu"
-	ResourceAWSNeuron   v1.ResourceName = "aws.amazon.com/neuron"
-	ResourceHabanaGaudi v1.ResourceName = "habana.ai/gaudi"
-	ResourceAWSPodENI   v1.ResourceName = "vpc.amazonaws.com/pod-eni"
+	ResourceNVIDIAGPU             v1.ResourceName         = "nvidia.com/gpu"
+	ResourceAMDGPU                v1.ResourceName         = "amd.com/gpu"
+	ResourceAWSNeuron             v1.ResourceName         = "aws.amazon.com/neuron"
+	ResourceHabanaGaudi           v1.ResourceName         = "habana.ai/gaudi"
+	ResourceAWSPodENI             v1.ResourceName         = "vpc.amazonaws.com/pod-eni"
+	NVIDIAacceleratorManufacturer AcceleratorManufacturer = "nvidia"
+	AWSAcceleratorManufacturer    AcceleratorManufacturer = "aws"
 
 	LabelInstanceHypervisor                   = LabelDomain + "/instance-hypervisor"
 	LabelInstanceEncryptionInTransitSupported = LabelDomain + "/instance-encryption-in-transit-supported"
@@ -127,3 +129,5 @@ func init() {
 		LabelInstanceAcceleratorMemory,
 	)
 }
+
+type AcceleratorManufacturer string
--- a/pkg/fake/ec2api.go
+++ b/pkg/fake/ec2api.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/Pallinder/go-randomdata"
 	"github.com/aws/aws-sdk-go/aws"
@@ -316,10 +317,15 @@ func (e *EC2API) DescribeImagesWithContext(_ context.Context, input *ec2.Describ
 	if !e.DescribeImagesOutput.IsNil() {
 		return e.DescribeImagesOutput.Clone(), nil
 	}
+	if aws.StringValue(input.Filters[0].Values[0]) == "invalid" {
+		return &ec2.DescribeImagesOutput{}, nil
+	}
 	return &ec2.DescribeImagesOutput{
 		Images: []*ec2.Image{
 			{
+				Name:         aws.String(test.RandomName()),
 				ImageId:      aws.String(test.RandomName()),
+				CreationDate: aws.String(time.Now().Format(time.UnixDate)),
 				Architecture: aws.String("x86_64"),
 			},
 		},
--- a/pkg/providers/amifamily/al2.go
+++ b/pkg/providers/amifamily/al2.go
@@ -21,9 +21,9 @@ import (
 	v1 "k8s.io/api/core/v1"
 
 	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
+	"github.com/aws/karpenter-core/pkg/scheduling"
 
 	"github.com/aws/karpenter-core/pkg/cloudprovider"
-	"github.com/aws/karpenter-core/pkg/utils/resources"
 	"github.com/aws/karpenter/pkg/apis/v1alpha1"
 	"github.com/aws/karpenter/pkg/providers/amifamily/bootstrap"
 )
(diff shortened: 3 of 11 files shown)


pkg/apis/v1alpha1/register.go
```diff
@@ -58,6 +58,8 @@ var (
 	ResourceAWSNeuron   v1.ResourceName = "aws.amazon.com/neuron"
 	ResourceHabanaGaudi v1.ResourceName = "habana.ai/gaudi"
 	ResourceAWSPodENI   v1.ResourceName = "vpc.amazonaws.com/pod-eni"
+	NVIDIAaccelerator                   = "nvidia"
```
jonathan-innis, line 61: Is there an upstream naming convention in `v1`? If not, maybe we can create a type for Nvidia and AWS for AcceleratorManufacturer?

i.e.

```go
type AcceleratorManufacturer string
```

pkg/fake/ec2api.go
```diff
@@ -316,10 +317,29 @@ func (e *EC2API) DescribeImagesWithContext(_ context.Context, input *ec2.Describ
 	if !e.DescribeImagesOutput.IsNil() {
 		return e.DescribeImagesOutput.Clone(), nil
 	}
+	if aws.StringValue(input.Filters[0].Name) == "image-id" {
```
jonathan-innis, line 320: Have we thought about doing a similar filtering mechanisms to [how we do filtering with describeInstances](https://github.com/aws/karpenter/blob/main/pkg/fake/ec2api.go#L265). This way it more closely aligns with how the upstream API performs

pkg/providers/amifamily/al2.go
```diff
@@ -33,16 +33,34 @@ type AL2 struct {
...
-// SSMAlias returns the AMI Alias to query SSM
-func (a AL2) SSMAlias(version string, instanceType *cloudprovider.InstanceType) string {
-	amiSuffix := ""
-	if !resources.IsZero(instanceType.Capacity[v1alpha1.ResourceNVIDIAGPU]) || !resources.IsZero(instanceType.Capacity[v1alpha1.ResourceAWSNeuron]) {
-		amiSuffix = "-gpu"
-	}
-	if instanceType.Requirements.Get(v1.LabelArchStable).Has(v1alpha5.ArchitectureArm64) {
-		amiSuffix = fmt.Sprintf("%s-%s", amiSuffix, v1alpha5.ArchitectureArm64)
+// DefaultAMIs returns the AMI name, and Requirements, with an SSM query
+func (a AL2) DefaultAMIs(version string) []DefaultAMIOutput {
```
jonathan-innis, line 37: Sanity check: Do we know what happens here when we request something that is using ARM64 and a GPU? I think we should have no compatible AMIs and just return an error with no AMIs that are matching, but just want to double-check

pkg/providers/amifamily/ami.go
```diff
@@ -94,45 +97,80 @@ func (p *Provider) KubeServerVersion(ctx context.Context) (string, error) {
 	return version, nil
 }
 
-// Get returns a set of AMIIDs and corresponding instance types. AMI may vary due to architecture, accelerator, etc
+// MapInstanceTypes returns a set of AMIIDs and corresponding instance types. AMI may vary due to architecture, accelerator, etc
```
jonathan-innis, line 100: nit: Update this comment so that it more accurately reflects what the function is doing

pkg/providers/amifamily/ami.go
```diff
@@ -94,45 +97,80 @@ func (p *Provider) KubeServerVersion(ctx context.Context) (string, error) {
...
+				break
 			}
 		}
-		if len(amiIDs) == 0 {
-			return nil, fmt.Errorf("no instance types satisfy requirements of amis %v,", lo.Keys(amiRequirements))
+	}
+	return amiIDs
+}
+
+// Returning a list of AMIs with its associated requirements
```
jonathan-innis, line 116: nit: Always start your comments with the name of the function, this is a Go language best practice

pkg/providers/amifamily/ami.go
```diff
@@ -94,45 +97,80 @@ func (p *Provider) KubeServerVersion(ctx context.Context) (string, error) {
...
+// the AMIs are sorted by creation date in descending order.
+// If creation date is nil or two AMIs have the same creation date, the AMIs will be sorted by name in ascending order.
```
jonathan-innis, line 118: nit: This part of the function comment is too specific for this function. To me, this should be part of the description of the sorting function

pkg/providers/amifamily/ami.go
```diff
@@ -233,14 +264,18 @@ func getFiltersAndOwners(amiSelector map[string]string) ([]*ec2.Filter, []*strin
...
 
-func sortAMIsByCreationDate(amiRequirements map[AMI]scheduling.Requirements) []AMI {
-	amis := lo.Keys(amiRequirements)
-
+func sortAMIsByCreationDate(amis []AMI) []AMI {
 	sort.Slice(amis, func(i, j int) bool {
-		itime, _ := time.Parse(time.RFC3339, amis[i].CreationDate)
-		jtime, _ := time.Parse(time.RFC3339, amis[j].CreationDate)
-		return itime.Unix() >= jtime.Unix()
+		if amis[i].CreationDate != "" || amis[j].CreationDate != "" {
```
jonathan-innis, line 269: Will this fail if one of the creationDates isn't set?

pkg/providers/amifamily/bottlerocket.go
```diff
@@ -38,17 +38,42 @@ type Bottlerocket struct {
...
-	if !resources.IsZero(instanceType.Capacity[v1alpha1.ResourceNVIDIAGPU]) {
-		amiSuffix = "-nvidia"
-	}
-	if instanceType.Requirements.Get(v1.LabelArchStable).Has(v1alpha5.ArchitectureArm64) {
-		arch = v1alpha5.ArchitectureArm64
+// DefaultAMIs returns the AMI name, and Requirements, with an SSM query
+func (b Bottlerocket) DefaultAMIs(version string) []DefaultAMIOutput {
+	return []DefaultAMIOutput{
+		{
+			Name:  fmt.Sprintf("bottlerocket-aws-k8s-%s%s", version, "-x86_64"),
```
jonathan-innis, line 45:
```suggestion
			Name:  fmt.Sprintf("bottlerocket-aws-k8s-%s-x86_64", version),
```

If you have a static string in the `fmt.Sprintf`, why not just put that static string in the templated string? I think this applies to all the `fmt.Sprintf`s so check this across all the files

pkg/providers/amifamily/bottlerocket.go
```diff
@@ -38,17 +38,42 @@ type Bottlerocket struct {
...
+			Query: fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/%s/latest/image_id", version, "x86_64"),
```
jonathan-innis, line 46:
```suggestion
			Query: fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/x86_64/latest/image_id", version),
```
Same comment here and throughout the code

pkg/providers/amifamily/custom.go
```diff
@@ -38,8 +39,8 @@ func (c Custom) UserData(_ *v1alpha5.KubeletConfiguration, _ []v1.Taint, _ map[s
 	}
 }
 
-func (c Custom) SSMAlias(_ string, _ *cloudprovider.InstanceType) string {
-	return "/unknown"
+func (c Custom) DefaultAMIs(_ string) []DefaultAMIOutput {
```
jonathan-innis, line 42: Any reason we have to return anything here at all?

pkg/providers/amifamily/ami.go
engedaam: Yeah it will fail, just made the adjustment

pkg/providers/amifamily/al2.go
engedaam: Yeah, that is the behavior we see here

engedaam (commented):
/karpenter snapshot

pkg/apis/v1alpha1/register.go
```diff
@@ -127,3 +129,9 @@ func init() {
 		LabelInstanceAcceleratorMemory,
 	)
 }
+
+type AcceleratorManufacturer string
+
+func (a AcceleratorManufacturer) Strings() string {
```
jonathan-innis, line 135: Why do we need this?

pkg/providers/amifamily/ami.go
```diff
@@ -94,45 +97,77 @@ func (p *Provider) KubeServerVersion(ctx context.Context) (string, error) {
...
 }
 
-// Get returns a set of AMIIDs and corresponding instance types. AMI may vary due to architecture, accelerator, etc
-// If AMI overrides are specified in the AWSNodeTemplate, then only those AMIs will be chosen.
-func (p *Provider) Get(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate, instanceTypes []*cloudprovider.InstanceType, amiFamily AMIFamily) (map[string][]*cloudprovider.InstanceType, error) {
-	kubernetesVersion, err := p.KubeServerVersion(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("getting kubernetes version %w", err)
-	}
+// MapInstanceTypes returns a map of AMIIDs to compatible instancetypes
```
jonathan-innis, line 100: I'd add a comment that we expect this map to be sorted such that the first AMI that matches is chosen to be compatible AMI for the instance type

pkg/providers/amifamily/ami.go
```diff
@@ -94,45 +97,77 @@ func (p *Provider) KubeServerVersion(ctx context.Context) (string, error) {
...
-			}
-			amiIDs[amiID] = append(amiIDs[amiID], instanceType)
+		amis, err = p.getAMIsFromSelector(ctx, nodeTemplate)
+		if err != nil {
+			return nil, err
+		}
+	}
+	amis = sortAMIsByCreationDate(amis)
+	if len(nodeTemplate.Spec.AMISelector) != 0 {
+		amis = groupAMIsByRequirements(amis)
```
jonathan-innis, line 132: Can't we just do this always? Why special case this?

jonathan-innis (approved):
LGTM 🚀

engedaam (commented):
/karpenter snapshot

PR #3710, opened 2023-04-06
docs: Add volume mount troubleshooting for volume limits

jonathan-innis:
Fixes # 

**Description**

Add volume mount troubleshooting for volume limits

Relates to https://github.com/aws/karpenter-core/issues/260

* Manually viewed docs

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```

### DIFF
--- a/website/content/en/preview/troubleshooting.md
+++ b/website/content/en/preview/troubleshooting.md
@@ -212,6 +212,30 @@ spec:
   ttlSecondsAfterEmpty: 30
 ```
 
+### Pods using PVCs can hit volume limits and fail to scale-up
+
+When attempting to schedule a large number of pods with PersistentVolumes, it's possible that these pods will co-locate on the same node. Pods will report the following errors in their events using a `kubectl describe pod` call 
+
+```console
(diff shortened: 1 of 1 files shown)


website/content/en/preview/troubleshooting.md
```diff
@@ -212,6 +212,30 @@ spec:
...
+```
+
+#### Pods were scheduled due to a race condition in Kubernetes
+
+Due to [this race condition in Kubernetes](https://github.com/kubernetes/kubernetes/issues/95911), it's possible that the scheduler and the CSINode can race during node registration such that the scheduler assumes that a node can mount infinite volumes. In this case, the scheduler will schedule more pods to the node than the node attachments support. There is currently no solve for this problem other than enforcing `toplogySpreadConstraints` and `podAntiAffinity` on your workloads that use PVCs such that you attempt to reduce the number of PVCs that schedule to a given node.
```
tzneal, line 237: more volumes than the CSI driver allows

tzneal (approved):
lgtm

PR #3724, opened 2023-04-07
test: Fix node leaking through expiration tests

jonathan-innis:
Fixes # 

**Description**

* `make presubmit`
* `FOCUS=Expiration make e2etests`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/test/suites/integration/expiration_test.go
+++ b/test/suites/integration/expiration_test.go
@@ -61,10 +61,23 @@ var _ = Describe("Expiration", func() {
 		node := env.Monitor.CreatedNodes()[0]
 		env.Monitor.Reset()
 
-		// Eventually expect the node to be gone and a new one to come up
-		env.EventuallyExpectNotFound(node)
+		// Eventually the node will be set as unschedulable, which means its actively being deprovisioned
+		Eventually(func(g Gomega) {
+			n := &v1.Node{}
+			g.Expect(env.Client.Get(env.Context, types.NamespacedName{Name: node.Name}, n)).Should(Succeed())
+			g.Expect(n.Spec.Unschedulable).Should(BeTrue())
+		}).Should(Succeed())
+
+		// Remove the TTLSecondsUntilExpired to make sure new node isn't deleted
+		// This is CRITICAL since it prevents nodes that are immediately spun up from immediately being expired and
+		// racing at the end of the E2E test, leaking node resources into subsequent tests
 		provisioner.Spec.TTLSecondsUntilExpired = nil
 		env.ExpectUpdated(provisioner)
+
+		// After the deletion timestamp is set and all pods are drained
+		// the node should be gone
+		env.EventuallyExpectNotFound(node)
+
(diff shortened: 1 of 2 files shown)


test/suites/integration/expiration_test.go
```diff
@@ -61,10 +61,19 @@ var _ = Describe("Expiration", func() {
 		node := env.Monitor.CreatedNodes()[0]
 		env.Monitor.Reset()
 
-		// Eventually expect the node to be gone and a new one to come up
-		env.EventuallyExpectNotFound(node)
```
ellistarn, line 65: Why isn't it enough to just wait?

test/suites/integration/expiration_test.go
```diff
@@ -61,10 +61,19 @@ var _ = Describe("Expiration", func() {
...
+		// Eventually the node will be set as unschedulable, which means its actively being deprovisioned
+		Eventually(func(g Gomega) {
+			n := &v1.Node{}
+			g.Expect(env.Client.Get(env.Context, types.NamespacedName{Name: node.Name}, n)).Should(Succeed())
+			g.Expect(n.Spec.Unschedulable).Should(BeTrue())
+		}).Should(Succeed())
```
ellistarn, line 69: What if the node is gone and this check never succeeds?

test/suites/integration/expiration_test.go
jonathan-innis: Because we aren't fast enough. If we wait to update the provisioner after the node is gone, the deprovisioner already gets into another loop of deprovisioning before it sees the provisioner update

test/suites/integration/expiration_test.go
jonathan-innis: Why would that happen? We pre-spin so it should always make it here

jonathan-innis:
This PR works around a few races that currently exist in Karpenter. These are the following:
1. Karpenter will end up in a deprovsioining loop for short expirations immediately after a test succeeds. The test will be cleaning up while the deprovisoining loop is running and launching a new node. That node _should_ have the ownerReference to the owning Provisioner that created it, but the provisioner has been cleaned up by the test suite. Because of #2831, this leaves a node around which will exist in subsequent test cases
2. EC2 will allow the userData script to run through even if the EC2 instance is deleted while the instance is launching. This means that an instance in a `shutting-down` state may still have a Node register to the cluster since the kubelet will register the node to the cluster. This can leave a zombie node on the cluster until CCM comes around and deletes the node that doesn't have a backing instance.

Number 1 above should be solved going forward with the #3408
Number 2 needs to be addressed separately so that we don't Delete machines until after they are fully removed from EC2 and are in a terminated state. This one needs more investigation since this will most likely increase the time that it takes for a Machine/Node to be fully terminated.

jonathan-innis (commented):
/karpenter snapshot

ellistarn (approved):

PR #3769, opened 2023-04-16
feat: Add windows support (single amifamily)

topikachu:
This change integrates with the latest EKS window support.

Add a new Windows AMI Family.
Windows relevant code, "kubernetes.io/os" and "vpc.amazonaws.com/PrivateIPv4Address" is only active when Windows AMI Family defined in the AwsNodeTemplate

Add two new provisioner requirement well-known labels karpenter.k8s.aws/windows-version and karpenter.k8s.aws/windows-variant

Test on a Linux and Windows mixed system for provision and de-provision.

Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/apis/v1alpha1/register.go
+++ b/pkg/apis/v1alpha1/register.go
@@ -38,26 +38,45 @@ var (
 	RestrictedLabelDomains = []string{
 		LabelDomain,
 	}
-	AMIFamilyBottlerocket = "Bottlerocket"
-	AMIFamilyAL2          = "AL2"
-	AMIFamilyUbuntu       = "Ubuntu"
-	AMIFamilyCustom       = "Custom"
-	SupportedAMIFamilies  = []string{
+	AMIFamilyBottlerocket  = "Bottlerocket"
+	AMIFamilyAL2           = "AL2"
+	AMIFamilyUbuntu        = "Ubuntu"
+	AMIFamilyWindowsServer = "WindowsServer"
+	AMIFamilyCustom        = "Custom"
+	SupportedAMIFamilies   = []string{
 		AMIFamilyBottlerocket,
 		AMIFamilyAL2,
 		AMIFamilyUbuntu,
+		AMIFamilyWindowsServer,
 		AMIFamilyCustom,
 	}
 	SupportedContainerRuntimesByAMIFamily = map[string]sets.String{
-		AMIFamilyBottlerocket: sets.NewString("containerd"),
-		AMIFamilyAL2:          sets.NewString("dockerd", "containerd"),
-		AMIFamilyUbuntu:       sets.NewString("dockerd", "containerd"),
+		AMIFamilyBottlerocket:  sets.NewString("containerd"),
+		AMIFamilyAL2:           sets.NewString("dockerd", "containerd"),
+		AMIFamilyUbuntu:        sets.NewString("dockerd", "containerd"),
+		AMIFamilyWindowsServer: sets.NewString("containerd"),
 	}
-	ResourceNVIDIAGPU             v1.ResourceName         = "nvidia.com/gpu"
-	ResourceAMDGPU                v1.ResourceName         = "amd.com/gpu"
-	ResourceAWSNeuron             v1.ResourceName         = "aws.amazon.com/neuron"
-	ResourceHabanaGaudi           v1.ResourceName         = "habana.ai/gaudi"
-	ResourceAWSPodENI             v1.ResourceName         = "vpc.amazonaws.com/pod-eni"
+
+	Windows2019              = "2019"
+	Windows2022              = "2022"
+	WindowsCore              = "Core"
+	WindowsFull              = "Full"
+	SupportedWindowsVersions = []string{
+		Windows2019,
+		Windows2022,
+	}
+	SupportedWindowsVariants = []string{
+		WindowsCore,
+		WindowsFull,
+	}
+
+	ResourceNVIDIAGPU          v1.ResourceName = "nvidia.com/gpu"
+	ResourceAMDGPU             v1.ResourceName = "amd.com/gpu"
+	ResourceAWSNeuron          v1.ResourceName = "aws.amazon.com/neuron"
+	ResourceHabanaGaudi        v1.ResourceName = "habana.ai/gaudi"
+	ResourceAWSPodENI          v1.ResourceName = "vpc.amazonaws.com/pod-eni"
+	ResourcePrivateIPv4Address v1.ResourceName = "vpc.amazonaws.com/PrivateIPv4Address"
+
 	NVIDIAacceleratorManufacturer AcceleratorManufacturer = "nvidia"
 	AWSAcceleratorManufacturer    AcceleratorManufacturer = "aws"
 
@@ -85,6 +104,9 @@ var (
 	LabelInstanceAcceleratorManufacturer = LabelDomain + "/instance-accelerator-manufacturer"
 	LabelInstanceAcceleratorCount        = LabelDomain + "/instance-accelerator-count"
 	LabelInstanceAcceleratorMemory       = LabelDomain + "/instance-accelerator-memory"
+
+	LabelWindowsVersion = LabelDomain + "/windows-version"
+	LabelWindowsVariant = LabelDomain + "/windows-variant"
 
 	InterruptionInfrastructureFinalizer = Group + "/interruption-infrastructure"
 )
@@ -127,6 +149,8 @@ func init() {
 		LabelInstanceAcceleratorManufacturer,
 		LabelInstanceAcceleratorCount,
 		LabelInstanceAcceleratorMemory,
+		LabelWindowsVersion,
+		LabelWindowsVariant,
 	)
 }
 
--- a/pkg/apis/v1alpha1/register.go
+++ b/pkg/apis/v1alpha1/register.go
@@ -38,23 +38,23 @@ var (
 	RestrictedLabelDomains = []string{
 		LabelDomain,
 	}
-	AMIFamilyBottlerocket  = "Bottlerocket"
-	AMIFamilyAL2           = "AL2"
-	AMIFamilyUbuntu        = "Ubuntu"
-	AMIFamilyWindowsServer = "WindowsServer"
-	AMIFamilyCustom        = "Custom"
-	SupportedAMIFamilies   = []string{
+	AMIFamilyBottlerocket = "Bottlerocket"
+	AMIFamilyAL2          = "AL2"
+	AMIFamilyUbuntu       = "Ubuntu"
+	AMIFamilyWindows      = "Windows"
+	AMIFamilyCustom       = "Custom"
+	SupportedAMIFamilies  = []string{
 		AMIFamilyBottlerocket,
 		AMIFamilyAL2,
 		AMIFamilyUbuntu,
-		AMIFamilyWindowsServer,
+		AMIFamilyWindows,
 		AMIFamilyCustom,
 	}
 	SupportedContainerRuntimesByAMIFamily = map[string]sets.String{
-		AMIFamilyBottlerocket:  sets.NewString("containerd"),
-		AMIFamilyAL2:           sets.NewString("dockerd", "containerd"),
-		AMIFamilyUbuntu:        sets.NewString("dockerd", "containerd"),
-		AMIFamilyWindowsServer: sets.NewString("containerd"),
+		AMIFamilyBottlerocket: sets.NewString("containerd"),
+		AMIFamilyAL2:          sets.NewString("dockerd", "containerd"),
+		AMIFamilyUbuntu:       sets.NewString("dockerd", "containerd"),
+		AMIFamilyWindows:      sets.NewString("containerd"),
 	}
 
 	Windows2019              = "2019"
--- a/pkg/apis/v1alpha1/register.go
+++ b/pkg/apis/v1alpha1/register.go
@@ -57,13 +57,15 @@ var (
 		AMIFamilyWindows:      sets.NewString("containerd"),
 	}
 
-	Windows2019              = "2019"
-	Windows2022              = "2022"
-	WindowsCore              = "Core"
-	WindowsFull              = "Full"
-	SupportedWindowsVersions = []string{
-		Windows2019,
-		Windows2022,
+	Windows2019                            = "2019"
+	Windows2019Build                       = "10.0.17763"
+	Windows2022                            = "2022"
+	Windows2022Build                       = "10.0.20348"
+	WindowsCore                            = "Core"
+	WindowsFull                            = "Full"
+	SupportedWindowsVersionAndBuildMapping = map[string]string{
+		Windows2019: Windows2019Build,
+		Windows2022: Windows2022Build,
 	}
 	SupportedWindowsVariants = []string{
 		WindowsCore,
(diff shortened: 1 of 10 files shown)


pkg/apis/v1alpha1/register.go
```diff
@@ -38,26 +38,45 @@ var (
...
 	}
-	AMIFamilyBottlerocket = "Bottlerocket"
-	AMIFamilyAL2          = "AL2"
-	AMIFamilyUbuntu       = "Ubuntu"
-	AMIFamilyCustom       = "Custom"
-	SupportedAMIFamilies  = []string{
+	AMIFamilyBottlerocket  = "Bottlerocket"
+	AMIFamilyAL2           = "AL2"
+	AMIFamilyUbuntu        = "Ubuntu"
+	AMIFamilyWindowsServer = "WindowsServer"
```
jonathan-innis, line 44:
```suggestion
	AMIFamilyWindows = "Windows"
```

I think we probably just want to keep the naming as short as possible

pkg/apis/v1alpha1/register.go
```diff
@@ -38,26 +38,45 @@ var (
...
+		AMIFamilyUbuntu:        sets.NewString("dockerd", "containerd"),
+		AMIFamilyWindowsServer: sets.NewString("containerd"),
 	}
-	ResourceNVIDIAGPU             v1.ResourceName         = "nvidia.com/gpu"
-	ResourceAMDGPU                v1.ResourceName         = "amd.com/gpu"
-	ResourceAWSNeuron             v1.ResourceName         = "aws.amazon.com/neuron"
-	ResourceHabanaGaudi           v1.ResourceName         = "habana.ai/gaudi"
-	ResourceAWSPodENI             v1.ResourceName         = "vpc.amazonaws.com/pod-eni"
+
+	Windows2019              = "2019"
```
jonathan-innis, line 60: Is it possible to introduce a single windows version as the default windows version and then build off of that once we get the initial version out and tested by users?

pkg/providers/amifamily/windows.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (w Windows) DefaultAMIs(version string) (results []DefaultAMIOutput) {
+	for _, windowsVersion := range v1alpha1.SupportedWindowsVersions {
+		for _, windowsVariant := range v1alpha1.SupportedWindowsVariants {
+			results = append(results, DefaultAMIOutput{Name: fmt.Sprintf("windows-server-%s-%s-%s", windowsVersion, windowsVariant, version),
+				Query: fmt.Sprintf("/aws/service/ami-windows-latest/Windows_Server-%s-English-%s-EKS_Optimized-%s/image_id", windowsVersion, windowsVariant, version),
+				Requirements: scheduling.NewRequirements(
+					scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, v1alpha5.ArchitectureAmd64),
+					scheduling.NewRequirement(v1.LabelOSStable, v1.NodeSelectorOpIn, string(v1.Windows)),
+					scheduling.NewRequirement(v1alpha1.LabelWindowsVersion, v1.NodeSelectorOpIn, windowsVersion),
```
jonathan-innis, line 50: Can these scheduling requirements be based on the [upstream scheduling requirements](https://kubernetes.io/docs/concepts/windows/user-guide/#ensuring-os-specific-workloads-land-on-the-appropriate-container-host) that users are probably already putting on their workloads?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -321,12 +322,21 @@ func (c *CloudProvider) instanceToMachine(ctx context.Context, ec2instance *ec2.
...
 	labels[v1alpha5.LabelCapacityType] = instance.GetCapacityType(ec2instance)
+	labels[v1.LabelOSStable] = instance.GetOS(ec2instance)
+
 	if tag, ok := lo.Find(ec2instance.Tags, func(t *ec2.Tag) bool { return aws.StringValue(t.Key) == v1alpha5.ProvisionerNameLabelKey }); ok {
 		labels[v1alpha5.ProvisionerNameLabelKey] = aws.StringValue(tag.Value)
 	}
 	if tag, ok := lo.Find(ec2instance.Tags, func(t *ec2.Tag) bool { return aws.StringValue(t.Key) == v1alpha5.ManagedByLabelKey }); ok {
 		labels[v1alpha5.ManagedByLabelKey] = aws.StringValue(tag.Value)
 	}
+	if tag, ok := lo.Find(ec2instance.Tags, func(t *ec2.Tag) bool { return aws.StringValue(t.Key) == v1alpha1.LabelWindowsVersion }); ok {
```
jonathan-innis, line 333: Do we need to add this? I would have assumed that this would have been satisfied by L313 above that adds the label if there's only one value used?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -321,12 +322,21 @@ func (c *CloudProvider) instanceToMachine(ctx context.Context, ec2instance *ec2.
 	labels[v1alpha1.LabelInstanceAMIID] = aws.StringValue(ec2instance.ImageId)
 	labels[v1.LabelTopologyZone] = aws.StringValue(ec2instance.Placement.AvailabilityZone)
 	labels[v1alpha5.LabelCapacityType] = instance.GetCapacityType(ec2instance)
+	labels[v1.LabelOSStable] = instance.GetOS(ec2instance)
```
jonathan-innis, line 325: Double-checking, but I'm assuming that this needs to be done because we no longer can know the OS based on instance type requirements?

pkg/providers/instancetype/types.go
```diff
@@ -160,6 +196,10 @@ func computeCapacity(ctx context.Context, info *ec2.InstanceTypeInfo, amiFamily
 		v1alpha1.ResourceAWSNeuron:   *awsNeurons(info),
 		v1alpha1.ResourceHabanaGaudi: *habanaGaudis(info),
 	}
+	if amiFamily.IsWindows() {
```
jonathan-innis, line 199: Can we just do a raw type check here rather than having to implement a new method on the interface? Like `if _, ok := amiFamily.(amifamily.Windows); ok {`

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -215,15 +215,15 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, options *amifamily.
...
+			ImageId:          aws.String(options.Ami.AmiID),
 			MetadataOptions: &ec2.LaunchTemplateInstanceMetadataOptionsRequest{
 				HttpEndpoint:            options.MetadataOptions.HTTPEndpoint,
 				HttpProtocolIpv6:        options.MetadataOptions.HTTPProtocolIPv6,
 				HttpPutResponseHopLimit: options.MetadataOptions.HTTPPutResponseHopLimit,
 				HttpTokens:              options.MetadataOptions.HTTPTokens,
 			},
 			TagSpecifications: []*ec2.LaunchTemplateTagSpecificationRequest{
-				{ResourceType: aws.String(ec2.ResourceTypeNetworkInterface), Tags: utils.MergeTags(options.Tags)},
+				{ResourceType: aws.String(ec2.ResourceTypeInstance), Tags: utils.MergeTags(options.Tags, options.Ami.GetLabelsFromRequirements())},
```
jonathan-innis, line 226: I think this is correct and not a type but double-checking with another maintainer who added this code.

pkg/apis/v1alpha1/register.go
```diff
@@ -86,6 +105,9 @@ var (
 	LabelInstanceAcceleratorCount        = LabelDomain + "/instance-accelerator-count"
 	LabelInstanceAcceleratorMemory       = LabelDomain + "/instance-accelerator-memory"
 
+	LabelWindowsVersion = LabelDomain + "/windows-version"
```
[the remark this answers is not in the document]
topikachu: There is no such mapping between Windows build and eks ssm config.
Can eks team give some new mappings?
For example
/aws/service/ami-windows-latest/Windows_Server-${windows-build}-English-${Core/Full}-EKS_Optimized-${version}/image_id

https://docs.aws.amazon.com/eks/latest/userguide/retrieve-windows-ami-id.html

pkg/providers/amifamily/windows.go
topikachu: Reasonable
But ssm has no such mapping

pkg/providers/amifamily/windows.go
```diff
@@ -0,0 +1,89 @@
...
+			CABundle:        caBundle,
+			CustomUserData:  customUserData,
+		},
+	}
+}
+
+// DefaultBlockDeviceMappings returns the default block device mappings for the AMI Family
+func (w Windows) DefaultBlockDeviceMappings() []*v1alpha1.BlockDeviceMapping {
+	sda1EBS := DefaultEBS
+	sda1EBS.VolumeSize = lo.ToPtr(resource.MustParse("50Gi"))
```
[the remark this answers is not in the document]
topikachu: It's required by Windows AMI :)

pkg/cloudprovider/cloudprovider.go
topikachu: I don't like this either. However, there's some problem I haven't found the root cause of.

1. These tags is used by drift cycle. yes, it sounds unreasonable but it is!
2. In the drift cycle, the controller recovers all machine requirements from node labels.
3. In the current implementation, the LabelWindowsVersion and LabelWindowsVariant are very important. They must be presented.
4. There are two options to add these label
    1. Pass as kubelet arguments. I tried, but to my surprise, it doesn't work. I believe, there's some hidden issue in [karpenter core](https://github.com/aws/karpenter-core/blob/ab4eccca6536fa89c751953cad2c0611c5a9018c/pkg/controllers/provisioning/provisioner.go#L362) that the node labels are overwritten by karpenter core.
    2.  Pass as node tags and restore by karpenter. As current code

pkg/apis/v1alpha1/register.go
topikachu: Sure.

pkg/cloudprovider/cloudprovider.go
topikachu: Not only double-checking. :)

Karpenter returns "Linux" so it breaks the drift cycle in some cases.

pkg/providers/instancetype/types.go
topikachu: Sure.
Will follow your team convention.

pkg/providers/instancetype/types.go
```diff
@@ -192,6 +232,17 @@ func ephemeralStorage(amiFamily amifamily.AMIFamily, blockDeviceMappings []*v1al
 	return amifamily.DefaultEBS.VolumeSize
 }
 
+func privateIPv4Address(name string) *resource.Quantity {
```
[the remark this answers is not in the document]
topikachu: Sure. But a new amiFamily argument is required.

pkg/providers/launchtemplate/launchtemplate.go
topikachu: Please tell me if I'm wrong.
I can add new TagSpecifications if both ResourceTypeNetworkInterface and ResourceTypeInstance are needed.

pkg/apis/v1alpha1/register.go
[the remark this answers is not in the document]
topikachu: This is "2019/2022" used by Windows AMI.
Not the build version.
And no idea to map the build version to the Windows server version.

pkg/providers/amifamily/windows.go
topikachu: The current upstream well-known labels don't meet the Windows AMI selecting requirements.
For pods, it may be enough (however, from my experience, a Kubernetes admin still needs to know the detail of the workload).
For EKS AMI, it needs three arguments to query from SSM
https://docs.aws.amazon.com/eks/latest/userguide/retrieve-windows-ami-id.html
Windows_Server-${2019/2022}-English-${Core/Full}-EKS_Optimized-${k8s_version}

pkg/providers/instancetype/types.go
```diff
@@ -137,6 +141,38 @@ func computeRequirements(ctx context.Context, info *ec2.InstanceTypeInfo, offeri
...
+
+	//Non amd64
+	if getArchitecture(info) != v1alpha5.ArchitectureAmd64 {
+		return false
+	}
+
+	//https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html
+	//Amazon EC2 instance types C3, C4, D2, I2, M4 (excluding m4.16xlarge), M6a.x, and R3 instances aren't supported for Windows workloads.
+	//From AWS support, M6a.x means all types in m6a family
+	split := strings.Split(*info.InstanceType, ".")
```
jonathan-innis, line 167: Is it possible to create a generic function for getting the instance type parts since we do it once here and we also do it again in `computeRequirements`?

pkg/providers/instancetype/types.go
```diff
@@ -46,6 +46,9 @@ const (
 
 var (
 	instanceTypeScheme = regexp.MustCompile(`(^[a-z]+)(\-[0-9]+tb)?([0-9]+).*\.`)
+
+	instanceTypesSupportedForWindows = []string{"m4.16xlarge"}
```
jonathan-innis, line 50: optional: scope these values to the `isSupportedForWindows` function since they are only used there

pkg/providers/instancetype/types.go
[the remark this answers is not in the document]
jonathan-innis: Why is a new AMI family argument value required? I think `ENILimitedPods` should return the same value as this function.

pkg/apis/v1alpha1/register.go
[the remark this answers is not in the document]
jonathan-innis: From my testing these are the build versions:

```
Windows Server 2022 Full: “10.0.20348”
Windows Server 2022 Core: “10.0.20348”
Windows Server 2019 Full: “10.0.17763”
Windows Server 2019 Core: “10.0.17763”
```

Based on that, it doesn't look like we can differentiate the variant of Windows based off of the build version, but it doesn't look like we have a need to define a new "/window-version" label. I think we can rely on the existing K8s concept here.

pkg/providers/amifamily/resolver.go
```diff
@@ -65,7 +65,7 @@ type LaunchTemplate struct {
 	UserData            bootstrap.Bootstrapper
 	BlockDeviceMappings []*v1alpha1.BlockDeviceMapping
 	MetadataOptions     *v1alpha1.MetadataOptions
-	AMIID               string
+	Ami                 AMI
```
jonathan-innis, line 68: What's the reason you need to pass the full AMI struct through here now?

jonathan-innis (commented):
Nice! Just some comments to simplify!

pkg/providers/amifamily/ami.go
```diff
@@ -302,3 +303,48 @@ func (p *Provider) getRequirementsFromImage(ec2Image *ec2.Image) scheduling.Requ
...
+		}
+	}
+
+	if len(results) != 0 {
+		return results
+	}
+	return amis
+}
+
+func (ami AMI) GetLabelsFromRequirements() map[string]string {
```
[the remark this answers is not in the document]
topikachu: For the drift cycle.
As said in https://github.com/aws/karpenter/pull/3769#discussion_r1169399806
the drift cycle recovers the machine requirements from node labels.
ec2 tag is the only solution I found that I can label nodes.
The kubelet args "--node-labels" doesn't work because of some unknown reasons. Perhaps https://github.com/aws/karpenter-core/blob/ab4eccca6536fa89c751953cad2c0611c5a9018c/pkg/controllers/provisioning/provisioner.go#L362 has some hidden issue.

PR #3786, opened 2023-04-20
docs: Update AWSNodeTemplate docs for ami selectors

vazkarvishal:
**Description**

The AMI Selector as shown on the [merged pr](https://github.com/aws/karpenter/pull/3204) uses the `aws::` prefixes on the tags. 

We tested using the AMISelector without the `aws::` prefix and it breaks. So thought of updating the docs to reflect this.

**Release Note**

Update AWSNodeTemplate docs for ami selectors to reflect the `aws::` prefix requirements.

### DIFF
--- a/website/content/en/preview/concepts/node-templates.md
+++ b/website/content/en/preview/concepts/node-templates.md
@@ -222,7 +222,14 @@ Select all AMIs with a specified tag:
 Select AMIs by name:
 ```yaml
   amiSelector:
-    Name: my-ami
+    aws::name: my-ami
+```
+
+Select AMIs by name and a specific owner:
+```yaml
+  amiSelector:
+    aws::name: my-ami
+    owners: self/ownerAccountID
 ```
 
 Select AMIs by an arbitrary AWS tag key/value pair:
@@ -234,7 +241,7 @@ Select AMIs by an arbitrary AWS tag key/value pair:
 Specify AMIs explicitly by ID:
 ```yaml
   amiSelector:
-    aws-ids: "ami-123,ami-456"
+    aws::ids: "ami-123,ami-456"
 ```
 
 ## spec.tags
--- a/website/content/en/v0.27.3/concepts/node-templates.md
+++ b/website/content/en/v0.27.3/concepts/node-templates.md
@@ -222,7 +222,14 @@ Select all AMIs with a specified tag:
 Select AMIs by name:
 ```yaml
   amiSelector:
-    Name: my-ami
+    aws::name: my-ami
+```
+
+Select AMIs by name and a specific owner:
+```yaml
+  amiSelector:
+    aws::name: my-ami
+    owners: self/ownerAccountID
 ```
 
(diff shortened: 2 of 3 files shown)


website/content/en/v0.27.3/concepts/node-templates.md
```diff
@@ -222,7 +222,14 @@ Select all AMIs with a specified tag:
 Select AMIs by name:
 ```yaml
   amiSelector:
-    Name: my-ami
+    aws::name: my-ami
```
jonathan-innis, line 225: Weirdly enough, there's the name tag and the actual name of the AMI that you can select on. We should probably specify that this is the name tag on L222 but keep an example in the docs with the Name tag as well as with `aws::name`

jonathan-innis (commented):
Do you mind backporting these changes across docs versions? Also, we take the `preview` version forward for new versions so do you mind also making these changes there?

vazkarvishal:
> Do you mind backporting these changes across docs versions? Also, we take the `preview` version forward for new versions so do you mind also making these changes there?

 Of course. no Problem.

website/content/en/v0.27.3/concepts/node-templates.md
jonathan-innis: It works if you use `create-tags` with `Name` being the key. If you create an AMI and view this screen, you can see there's a dileniation between the "Name" (which is the name tag) and the AMI Name, which is the actual "name" attribute. I admit it's incredibly confusing, but I think we should probably callout both in the docs.

Screenshot 2023-04-21 at 4 09 41 PM

vazkarvishal: @jonathan-innis  okay, just so that I add the right parts back in, am I right in understanding that adding the below block in is what you mean by use of **create tags** with name?
 
```yaml
amiSelector:
    Name: my-ami
```

jonathan-innis: Yes, that's correct. I think the examples should be:
1. Select AMIs by Name Tag (Name)
2. Select AMIs by AMI Name (aws::name)
3. Select AMIs with a specific owner (aws::owners)
4. Select AMIs explicitly with id (aws::ids)

vazkarvishal: Okay I have updated the docs now. Can you re-review please? ta.

website/content/en/preview/concepts/node-templates.md
```diff
@@ -219,12 +219,24 @@ Select all AMIs with a specified tag:
...
 ```yaml
   amiSelector:
     Name: my-ami
 ```
 
+Select AMIs by name and a specific owner:
+```yaml
+  amiSelector:
+    aws::name: my-ami
+    owners: self/ownerAccountID
```
jonathan-innis, line 237:
```suggestion
    aws::owners: self/ownerAccountID
```

jonathan-innis (approved):
LGTM 🚀

PR #3790, opened 2023-04-21
test: Fix upgrading CRDS from older version

jonathan-innis:
Fixes # 

**Description**

The `kubectl replace` directive does not work when the object doesn't already exist. Anyways, we should not have to specify new CRDs when we add them. This change applies _every_ CRD from the pkg/apis/crds directory

* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/test/manifests/upgrade-crds.yaml
+++ b/test/manifests/upgrade-crds.yaml
@@ -9,15 +9,29 @@ spec:
   params:
     - name: cluster-name
       description: Name of the cluster under test.
+    - name: git-repo-url
+      description: the git repo url can be used to run tests on a fork
+      default: "https://github.com/aws/karpenter"
     - name: git-ref
       description: Git commit, tag, or branch to check out. Requires a corresponding Karpenter snapshot release.
   workspaces:
     - name: ws
   steps:
+  - name: checkout-repository
+    image: alpine/git
+    workingDir: $(workspaces.ws.path)
+    script: |
+      echo "Cloning $(params.git-repo-url)"
+      git clone $(params.git-repo-url).git .
+      
+      echo "Fetching: $(params.git-ref)"
+      git fetch origin $(params.git-ref)
+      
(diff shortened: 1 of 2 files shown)


test/manifests/upgrade-crds.yaml
```diff
@@ -9,15 +9,29 @@ spec:
...
+      echo "Checking out git ref: $(echo $(params.git-ref) | cut -d":" -f2)"
+      git checkout $(echo $(params.git-ref)| cut -d":" -f2)
   - name: upgrade-crds
     image: public.ecr.aws/karpenter/tools:latest
+    workingDir: $(workspaces.ws.path)
     script: |
       aws eks update-kubeconfig --name $(params.cluster-name)
-      kubectl replace -f https://raw.githubusercontent.com/aws/karpenter/$(params.git-ref)/pkg/apis/crds/karpenter.sh_provisioners.yaml
-      kubectl replace -f https://raw.githubusercontent.com/aws/karpenter/$(params.git-ref)/pkg/apis/crds/karpenter.sh_machines.yaml
-      kubectl replace -f https://raw.githubusercontent.com/aws/karpenter/$(params.git-ref)/pkg/apis/crds/karpenter.k8s.aws_awsnodetemplates.yaml
```
bwagner5, line 23: the reason we replace is to take updates from additions to these CRDs like we suggest in the upgrade guide.  Will apply do the same thing w/ `-f`?

jonathan-innis: Right, `apply` is the same behavior, just with an override mechanism and it doesn't care if the resource already exists. See: https://stackoverflow.com/questions/47241626/what-is-the-difference-between-kubectl-apply-and-kubectl-replace

jonathan-innis: This also means that we don't have to callout the resources by name in our docs

bwagner5 (approved):
Lgtm

PR #3803, opened 2023-04-24
BREAKING CHANGE: Remove `.Values.extraObjects`

jonathan-innis:
Fixes # 

**Description**

Reverts #3370 which allowed the ability to add extra manifests into the helm chart. This is being removed from the Karpenter project for a few reasons:
1. It couples the install of Karpenter with the layer that manages and controls the nodes. Since Provisioners own Machines, Karpenter cannot be uninstalled without carefully tearing down the Provisioners before uninstalling the Provisioners and the Karpenter deployment
2. The `extraObjects` field did not work for new installs since the webhooks were not up before the objects were applied. This meant that `helm installs` would often fail for users that were installing their manifests in this way.

Since Helm offers no in-chart dependency mechanisms (beyond hooks, which do not work in this case), the Karpenter maintainer team recommends that, if you want to install Karpenter manifests as a helm chart (alongside Karpenter), it should be done using a separate chart that can take the `aws/karpenter` chart as a dependency so there is a strict ordering for install.

* `make apply`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
None
```

### DIFF
--- a/charts/karpenter/README.md
+++ b/charts/karpenter/README.md
@@ -50,7 +50,6 @@ helm upgrade --install --namespace karpenter --create-namespace \
 | controller.sidecarVolumeMounts | list | `[]` | Additional volumeMounts for the sidecar - this will be added to the volume mounts on top of extraVolumeMounts |
 | dnsConfig | object | `{}` | Configure DNS Config for the pod |
 | dnsPolicy | string | `"Default"` | Configure the DNS Policy for the pod |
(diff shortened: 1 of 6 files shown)


bwagner5 (approved):
lgtm

pierluigilenoci:
FYI https://github.com/helm/helm/issues/12653

PR #3808, opened 2023-04-25
test: E2ETesting Delete launch templates in cleanup step

jonathan-innis:
Fixes #

**Description**

Delete launch templates associated with the cluster during the cleanup step on the E2E test run. This ensures that we don't leak launch templates in our account when tearing down the cluster

* Running the command manually

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/test/hack/cleanup/cleanup.sh
+++ b/test/hack/cleanup/cleanup.sh
@@ -13,10 +13,12 @@ aws cloudformation list-stacks --stack-status-filter CREATE_COMPLETE ROLLBACK_FA
 
 
 # Clean up old launch templates
-launch_templates=$(aws ec2 describe-launch-templates |  jq -r '.LaunchTemplates[].LaunchTemplateName' | grep Karpenter-karpenter-tests)
-n_lts=$(echo ${launch_templates} | wc -w)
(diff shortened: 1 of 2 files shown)


test/hack/cleanup/cleanup.sh
```diff
@@ -13,10 +13,12 @@ aws cloudformation list-stacks --stack-status-filter CREATE_COMPLETE ROLLBACK_FA
...
-for lt in ${launch_templates}; do
+
+# Iterate through every line in the launch template describe response
+while IFS= read -r lt; do
   echo "Deleting LT ${lt}"
-  aws ec2 delete-launch-template --no-cli-pager --launch-template-name ${lt}
-done
\ No newline at end of file
+  aws ec2 delete-launch-template --no-cli-pager --launch-template-name "${lt}"
+done <<< "$launch_templates"
```
bwagner5: changing it like this doesn't handle the case where there are 0 LTs to delete and causes the script to exit 1. 

```suggestion
# Iterate through every line in the launch template describe response
 for lt in ${launch_templates}; do
   echo "Deleting LT ${lt}"
   aws ec2 delete-launch-template --no-cli-pager --launch-template-name "${lt}"
 done
```

bwagner5 (approved):
LGTM

PR #3814, opened 2023-04-26
fix: Explicitly configure AssignPublicIp in EC2 Launch Templates

### LINKED ISSUE #3815: Explicitly define `AssociatePublicIpAddress` in Launch Templates

Explicitly set `AssociatePublicIpAddress` in launch templates. The missing key in launch templates results in the rejection of EC2 instance creation requests in environments where `AssociatePublicIpAddress` is required to be false by the use of a `BoolIfExists` condition, an AWS-recommended best practice, [an instance of which can be found here under the subheading "Recommended Combination".](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-multifactorauthpresent) To be explicit, this problem is occurring with a subnet which is *not* configured to assign public IP addresses.

I am aware that the project would prefer to have subnet settings determine whether an attempt to associate a public IP address is actually made. Ensuring that a value exists, either true or false, does not change this design decision. 

See above.

I am using custom launch templates, which introduces a great deal of overhead for a basic and philosophically-aligned feature.

### Additional Context

Example policy statement, demonstrating the cause of the conflict:

```json
{
    "Effect": "Deny",
    "Action": [
        "ec2:RunInstances"
    ],
    "Condition": {
        "BoolIfExists": {
            "ec2:AssociatePublicIpAddress": "true"
        }
    },
    "Resource": "arn:aws:ec2:*:*:network-interface/*"
}
```

Sketch of a possible solution:
https://github.com/aws/karpenter/compare/main...lorenzadia:karpenter:lorenzo.explictly-disable-public-ipaddr-in-launch-template

### Attachments

_No response_

### Community Note

lorrrrrrrenzo:
Fixes #3815

**Description**
See issue.

* I added two tests covering three possible values of `AssignPublicIP`, `true`, `false` and undefined.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/suite_test.go
+++ b/pkg/cloudprovider/suite_test.go
@@ -419,7 +419,7 @@ var _ = Describe("CloudProvider", func() {
 			ExpectScheduled(ctx, env.Client, pod)
 			Expect(awsEnv.EC2API.CalledWithCreateLaunchTemplateInput.Len()).To(BeNumerically(">=", 1))
 			awsEnv.EC2API.CalledWithCreateLaunchTemplateInput.ForEach(func(ltInput *ec2.CreateLaunchTemplateInput) {
-				Expect(aws.StringValueSlice(ltInput.LaunchTemplateData.NetworkInterfaces[0].Groups)).To(ConsistOf("sg-test1"))
+				Expect(aws.StringValueSlice(ltInput.LaunchTemplateData.SecurityGroupIds)).To(ConsistOf("sg-test1"))
 			})
 		})
 		It("should discover security groups by ID in the LT when no network interfaces are defined", func() {


pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -218,6 +221,11 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, options *amifamily.
 				HttpPutResponseHopLimit: options.MetadataOptions.HTTPPutResponseHopLimit,
 				HttpTokens:              options.MetadataOptions.HTTPTokens,
 			},
+			NetworkInterfaces: []*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest{
+				{
+					AssociatePublicIpAddress: aws.Bool(options.AssociatePublicIPAddress),
```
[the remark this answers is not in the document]
bwagner5: instead of passing this as an option, I think it would be better to depend on the subnetProvider within the launchTemplate provider and check if all subnets returned by subnetProvider.List() have the default public IP association set to false. If that holds true, then explicitly set AssociatePublicIPAddress to false when creating the launch template. 

This way, when we add support for NetworkInterfaces, the API doesn't change and we can incorporate the user provided network interface configuration.  

Let me know if you think this would solve your issue and is the right way to go

pkg/providers/subnet/subnet.go
```diff
@@ -86,6 +86,23 @@ func (p *Provider) List(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTempl
...
+	}
+	onlyPrivateSubnets := true
+	for _, sb := range subnets {
+		if sb.MapPublicIpOnLaunch != nil {
+			if *sb.MapPublicIpOnLaunch {
+				onlyPrivateSubnets = false
+			}
+		}
+	}
+	return onlyPrivateSubnets, nil
```
bwagner5, line 103:
```suggestion
	for _, sb := range subnets {
		if aws.BoolValue(sb.MapPublicIpOnLaunch) {
                 	return false
		}
	}
	return true, nil
```

.golangci.yaml
```diff
@@ -34,7 +34,7 @@ linters:
 
 linters-settings:
   gocyclo:
-    min-complexity: 11
+    min-complexity: 15
```
bwagner5, line 37:
```suggestion
    min-complexity: 11
```

pkg/providers/subnet/subnet.go
```diff
@@ -86,6 +86,23 @@ func (p *Provider) List(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTempl
 	return output.Subnets, nil
 }
 
+// OnlyPrivateSubnets returns a bool indicating whether all referenced subnets do *not* assign public IPv4 addresses to instances created therein
+func (p *Provider) OnlyPrivateSubnets(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate) (bool, error) {
```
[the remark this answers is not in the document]
bwagner5: There's an EC2 API mock that outputs hardcoded subnets here https://github.com/aws/karpenter/blob/main/pkg/fake/ec2api.go#L366-L410

You could add the associatePublicIP bool on 2 of those subnets and then write tests selecting the two that have false and then another selecting the one that is true.

One should also be in the launch templates suite test to check the LT gets a propagated value which would follow a similar pattern of selecting subnets in the AWSNodeTemplate: https://github.com/aws/karpenter/blob/main/pkg/providers/launchtemplate/suite_test.go

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -148,16 +151,21 @@ func (p *Provider) createAMIOptions(ctx context.Context, nodeTemplate *v1alpha1.
 	if len(securityGroupsIDs) == 0 {
 		return nil, fmt.Errorf("no security groups exist given constraints")
 	}
+	associatePublicIpv4Addrs, err := p.subnetProvider.AssociatePublicIpv4Addrs(ctx, nodeTemplate)
```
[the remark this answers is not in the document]
jonathan-innis: Should we just pass:
1. `true` if all are `true`
2. `false` if all are `false`
3. `nil` if there is a mix

This would give us nice parity among all the return conditions

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -148,16 +151,21 @@ func (p *Provider) createAMIOptions(ctx context.Context, nodeTemplate *v1alpha1.
...
+		ClusterName:              settings.FromContext(ctx).ClusterName,
+		ClusterEndpoint:          p.ClusterEndpoint,
+		AWSENILimitedPodDensity:  settings.FromContext(ctx).EnableENILimitedPodDensity,
+		InstanceProfile:          instanceProfile,
+		SecurityGroupsIDs:        securityGroupsIDs,
+		Tags:                     tags,
+		Labels:                   labels,
+		CABundle:                 p.caBundle,
+		KubeDNSIP:                p.KubeDNSIP,
+		AssociatePublicIPv4Addrs: associatePublicIPv4Addrs,
```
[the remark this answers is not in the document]
jonathan-innis: Responded in the comment that you linked

pkg/providers/amifamily/resolver.go
```diff
@@ -69,6 +70,7 @@ type LaunchTemplate struct {
 	AMIID               string
 	InstanceTypes       []*cloudprovider.InstanceType `hash:"ignore"`
 	DetailedMonitoring  bool
+	NetworkInterface    []*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest
```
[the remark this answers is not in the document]
jonathan-innis: Ah, looking at it now I think this is probably the best way to organize the code:
1. Put the `CheckIPv4AddressAssociations...` call inside of the `createAMIOptions` call and store the resolved value as a `associatePublicIpAddress` with a `*bool` type in the `amifamily.Options`.
2. This value in the `amifamily.Options` will automatically get passed through to the `amifamily.LaunchTemplate` so there's no need to handle it in the amifamily code
3. Using that value in `createLaunchTemplate` to generate the `networkInterfaces` if specified

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -236,6 +254,23 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, options *amifamily.
 	return output.LaunchTemplate, nil
 }
 
+func (p *Provider) generateNetworkInterface(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate, options *amifamily.LaunchTemplate) ([]*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest, error) {
+	associatePublicIPv4Addrs, err := p.subnetProvider.CheckAnyPublicIPv4Associations(ctx, nodeTemplate)
+	if err != nil {
+		return nil, err
+	}
+	if !associatePublicIPv4Addrs {
+		return []*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest{
```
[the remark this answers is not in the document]
jonathan-innis: Yep, I think this makes sense. 👍

pkg/providers/subnet/subnet.go
```diff
@@ -89,6 +89,18 @@ func (p *Provider) List(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTempl
 	return output.Subnets, nil
 }
 
+// CheckAnyPublicIPv4Associations returns a bool indicating whether all referenced subnets assign public IPv4 addresses to EC2 instances created therein
+func (p *Provider) CheckAnyPublicIPv4Associations(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate) (bool, error) {
+	subnets, err := p.List(ctx, nodeTemplate)
+	if err != nil {
+		return false, err
+	}
+	_, ok := lo.Find(subnets, func(s *ec2.Subnet) bool {
```
jonathan-innis, line 98: optional: Update this to `lo.Any()` so that we return `true` if we have any IPv4 subnets set, otherwise, we return false

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -236,6 +256,19 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, options *amifamily.
 	return output.LaunchTemplate, nil
 }
 
+func (p *Provider) generateNetworkInterface(options *amifamily.LaunchTemplate) []*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest {
+	if !aws.BoolValue(options.AssociatePublicIpv4Address) {
+		return []*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest{
+			{
+				AssociatePublicIpAddress: aws.Bool(false),
```
jonathan-innis, line 263:
```suggestion
				AssociatePublicIpAddress: options. AssociatePublicIpv4Address,
```

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -148,16 +151,25 @@ func (p *Provider) createAMIOptions(ctx context.Context, nodeTemplate *v1alpha1.
...
-		Tags:                    tags,
-		Labels:                  labels,
-		CABundle:                p.caBundle,
-		KubeDNSIP:               p.KubeDNSIP,
+		ClusterName:                settings.FromContext(ctx).ClusterName,
+		ClusterEndpoint:            p.ClusterEndpoint,
+		AWSENILimitedPodDensity:    settings.FromContext(ctx).EnableENILimitedPodDensity,
+		InstanceProfile:            instanceProfile,
+		SecurityGroupsIDs:          securityGroupsIDs,
+		AssociatePublicIpv4Address: aws.Bool(anyPublicIPv4Associations),
```
jonathan-innis, line 168:
```suggestion
		AssociatePublicIpAddress: aws.Bool(anyPublicIPv4Associations),
```

We should probably align this name with the API

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -148,16 +151,25 @@ func (p *Provider) createAMIOptions(ctx context.Context, nodeTemplate *v1alpha1.
...
 	}
+	if err != nil {
+		return nil, err
+	}
+	anyPublicIPv4Associations, err := p.subnetProvider.CheckAnyPublicIPv4Associations(ctx, nodeTemplate)
+	if err != nil {
+		return nil, err
+	}
+
 	return &amifamily.Options{
```
jonathan-innis, line 162: How do you feel about being really explicit about how this is being set in code combined with a comment on why we are doing this explicit set logic? i.e. you do the following in code

```go
options := amifamily.Options{...} // leave the AssociatePublicIPAddress nil
if ok, err := p.subnetProvider.CheckAnyPublicIPAssociations(ctx, nodeTemplate); err != nil {
  return nil, err
} else if !ok {
  // provide some comment here on why we are doing this
  options.AssociatePublicIpAddress = aws.Bool(false)
}
return options
```

pkg/providers/subnet/subnet.go
```diff
@@ -89,6 +89,18 @@ func (p *Provider) List(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTempl
 	return output.Subnets, nil
 }
 
+// CheckAnyPublicIPv4Associations returns a bool indicating whether all referenced subnets assign public IPv4 addresses to EC2 instances created therein
+func (p *Provider) CheckAnyPublicIPv4Associations(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate) (bool, error) {
```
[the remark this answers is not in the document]
jonathan-innis: Makes sense. Does it make sense to just align the names still?

pkg/providers/launchtemplate/launchtemplate.go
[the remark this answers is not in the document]
jonathan-innis: Discussed with @bwagner5 offline. We are going with the way that he has proposed. We'll do a check and return `false` if _all_ of the subnets do not have `AssociatePublicIpAddress` set, otherwise, it will be `nil` and we will defer to the subnet setting

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -236,6 +256,19 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, options *amifamily.
 	return output.LaunchTemplate, nil
 }
 
+func (p *Provider) generateNetworkInterface(options *amifamily.LaunchTemplate) []*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest {
+	if !aws.BoolValue(options.AssociatePublicIpv4Address) {
```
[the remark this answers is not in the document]
jonathan-innis: I responded to the comment. @bwagner5 and I are aligned now on the approach.

pkg/providers/amifamily/resolver.go
```diff
@@ -48,11 +48,12 @@ type Resolver struct {
...
-	ClusterEndpoint         string
-	AWSENILimitedPodDensity bool
-	InstanceProfile         string
-	CABundle                *string `hash:"ignore"`
+	ClusterName              string
+	ClusterEndpoint          string
+	AWSENILimitedPodDensity  bool
+	InstanceProfile          string
+	CABundle                 *string `hash:"ignore"`
+	AssociatePublicIpAddress *bool   //nolint:revive,stylecheck
```
jonathan-innis, line 56: Why is there a need for a `//nolint` here?

pkg/providers/launchtemplate/launchtemplate.go
jonathan-innis: > The snippet above changes the logic, and in the wrong place.

How do you mean that it changes the logic in the wrong place?

1. I would expect that the concept of `checkAnyPublicIPAddresses` would return `true` if any IPs have a public mapping, I think that makes sense as we have it.
2. I would expect that the value `AssociatePublicIpAddress` in the `amifamily.Options` would appropriately map to the same value in the `networkInterface` (since it has the same name). In this case, we are _not_ doing that and I think that it's likely to cause some confusion. To me it makes more sense to handle setting that value to whatever it should be, outside the context of the options, so that you properly abstract the handling of passing that value down to `generateNetworkInterfaces`. i.e. `generateNetworkInterfaces` doesn't leak logic into this feature. If things change later and we do something different, like surface `v1alpha1.NetworkInterfaces` through the API, `generateNetworkInterfaces` doesn't have to change since it still handles passing the `AssociatePublicIpAddress` value through in an abstracted, non-leaky way.

pkg/providers/subnet/subnet.go
jonathan-innis: Just checking: did you intend to resolve this one?

pkg/providers/amifamily/resolver.go
jonathan-innis: Got it. I'm more inclined to side with the linter and not have `//nolint`s in the code if we can avoid it, even if the naming is slightly off

jonathan-innis (approved):
LGTM 🚀

bwagner5 (commented):
just some minor stuff, overall LGTM! Thanks for contributing!

jonathan-innis (approved):
LGTM 🚀

bwagner5 (approved):
/karpenter snapshot

jonathan-innis (dismissed):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

bwagner5 (commented):
/karpenter snapshot

bwagner5 (approved):
lgtm

tuananh:
awesome. Thanks. Just what we need. 

we have scp to deny associatepublicip and without this, karpenter cannot provision new nodes.

PR #3819, opened 2023-04-27
feat: add networkInterfaces configuration to launchTemplate

### LINKED ISSUE #2026: configure network interfaces
configure the network interfaces. In particular enable [AssociatePublicIpAddress](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-networkinterface.html#cfn-ec2-launchtemplate-networkinterface-associatepublicipaddress)

Our particular requirement is to associate public IPs on some node's interfaces for UDP ingress where there is no LB support avail.
We do not want to enable public IPs an all nodes, that's why we cannot use a public subnet.

None, docs even state  "Do not configure network instances in the launch template"

### Community Note

myaser:
Fixes #2026

**Description**
Add a networkInterfaces struct to the AWSNodeTemplate CRD, and use the parameters to create AWS launch templates

* unit and integration tests were created; also tested manually on a test cluster

- [x] Yes, issue opened: TBD
- [ ] No

**Release Note**

```release-note
TBD
```


jonathan-innis:
[Responded in the K8s slack](https://kubernetes.slack.com/archives/C02SFFZSA2K/p1682676053491149). I think we need a one-pager design on this feature add to consider all the different inflection points here. Any surface added to the AWSNodeTemplate probably needs to consider how the change interacts with things like drift (@njtran) as well as how it affects our subnet and security group resolution logic (@engedaam).

myaser:
@jonathan-innis 
I added the design doc as part of this PR

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,69 @@
...
+iterative approach where we start by exposing subset of the fields of networkInterfaces and then expand the support based on the demand.
+those fields are:
+```yaml
+  AssociateCarrierIpAddress: Boolean
+  AssociatePublicIpAddress: Boolean
+  DeleteOnTermination: Boolean
+  Description: String
+  DeviceIndex: Integer
+  InterfaceType: String
+  NetworkCardIndex: Integer
```
bwagner5, line 58: thoughts on exposing ipv4PrefixCount and ipv6PrefixCount?
Those seem useful to mitigate the need of the VPC CNI to allocate IP addresses at startup and gets a head start allocating IPs.

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,69 @@
...
+for example if the provisioner allows the `m5.large` and `m5.xlarge` instance types, while the `AWSNodeTemplate` specifies 4 network interfaces, karpenter controller needs to know that it can only provision `m5.xlarge` instances.
+
+## node drift
+there are requirements to reconcile nodes as they drift from the specs defined in the provisioner resource as defined [here](https://github.com/aws/karpenter/issues/1738) and [here](https://github.com/aws/karpenter/issues/1457).
+the topic of having multiple network interfaces per node adds to the complexity of the drift detection logic.
+
+# Proposed Solution:
+iterative approach where we start by exposing subset of the fields of networkInterfaces and then expand the support based on the demand.
+those fields are:
+```yaml
```
bwagner5, line 51: can you add this example to a minimal AWSNodeTemplate to make it clearer where this will live and that it's a list of this info?

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,69 @@
...
+```
+having both options (`AWSNodeTemplate` level and networkInterface level) will complicate the API and will cause confusion for the users. It shall also complicate the logic on karpenter controller side.   
+In addition, removing the `AWSNodeTemplate`'s level subnet and security group selectors is a breaking change from API backward compatibility perspective.
+
+one more thing to consider is that subnets are bound to a single AZ. that's why the resolution happen in runtime instead of being part of the AWS launchTemplate (to avoid creating as many launch templates as the AZs in the zone).
+moving the resolution logic to the networkInterface level has the following complications: 
+1. more resolutions and interface allocations to do
+2. NetworkInterfaces can have conflicting requirements (e.g. different subnets in different AZs). which we need to handle in runtime.
+
+## networkInterfaces with specific IP addresses
```
bwagner5, line 37: Agree on leaving this out!

I'm not entirely sure how this would work, but there is an ask for a selector on interfaceID such that you could assign an EIP, and could extend that to a NetworkInterfaceID https://github.com/aws/karpenter/issues/3899 

I'm not sure if Karpenter is the right place for that logic though, seems more like the CNI should allocate the EIP rather than Karpenter.

bwagner5 (commented):
nice doc! I agree with the direction that this is taking. I haven't looked at the code or taken it for a spin yet though. 

We'll definitely need some e2e tests for these. I'm super curious if this would speed up large scale-ups too if we add ipv4Prefixes that are allocated at launch! :)

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,69 @@
...
+```yaml
+  AssociateCarrierIpAddress: Boolean
+  AssociatePublicIpAddress: Boolean
+  DeleteOnTermination: Boolean
+  Description: String
+  DeviceIndex: Integer
+  InterfaceType: String
+  NetworkCardIndex: Integer
+```
+in addition, we will only support two networkInterfaces per node, this way we guarantee that the node will be able to run on all instance types.
```
bwagner5, line 60: I think for this case, we just have to trust the user to craft a provisioner and AWSNodeTemplate that are compatible. The 2 limit is safe, but it's limiting if you've configured your provisioner to only have c6i.48xlarge instance types or something like that.

designs/aws-launch-template-network-interfaces.md
myaser: There is no way to assign an EIP to launchTemplate. The best we can do is 

1. use a selector to select a NetworkInterface and assign it to the launchTemplate using [NetworkInterfaceId](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-networkinterface.html#cfn-ec2-launchtemplate-networkinterface-networkinterfaceid); in this case the user will have to create the network interfaces and assign the EIP to them beforehand. Karpenter do not allocate EIPs in that case
2. alternatively, regardless of how the networkInterfaces are created (dynamically, or using a networkInterface selector to pick an existing interface); users can specify an EIP tag/selector that karpenter will use it to associate the EIP to the EC2 instance after it is created.

I guess 1 is more in the scope of this change, but it does not fully solve the problem for the user. they will need to pre-create network interfaces and associate EIPs with them.

in addition, the [workaround](https://github.com/aws-samples/eip-and-dns-auto-binding-for-karpenter-worker#solution-introduction) specified by the user in the above issue suggests more than just assigning an EIP which go beyond the scope of Karpenter

designs/aws-launch-template-network-interfaces.md
myaser: added

designs/aws-launch-template-network-interfaces.md
myaser: added

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -260,16 +260,36 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, options *amifamily.
 // This is done to help comply with AWS account policies that require explicitly setting that field to 'false'.
 // https://github.com/aws/karpenter/issues/3815
 func (p *Provider) generateNetworkInterface(options *amifamily.LaunchTemplate) []*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest {
```
demonCoder95, line 262: Probably should rename the function to `generateNetworkInterfaces` (plural), since now we're generating multiple of them in a loop.

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,129 @@
...
+```yaml
+subnetSelector: String
+securityGroupSelector: String
+```
+
+this way we can support different subnets and security groups for each network interface, we can also remove the limit
+of having only 2 network interfaces per LaunchTemplate.
+
+The drift design doc #366 suggests classifying CRD fields as  (1) Static, (2) Dynamic, or (3) Behavioral. in this
+context, the networkInterfaces would be classified as following:
```
myaser, line 107: @jonathan-innis

I added this section about drift detection

designs/aws-launch-template-network-interfaces.md
bwagner5: Yeah, I thin runtime errors are okay for the complex validation failures having to do with AWS APIs parameter combinations. Anything that we can validate in a straight forward way on the CRD is good, but I'd stop short of any complicated validations and definitely stopping short at any AWS API calls to perform validation.

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,129 @@
...
+  networkInterfaces:                      # optional, configures network interfaces for the instance
+    - AssociateCarrierIpAddress: false    # optional, Indicates whether to associate a Carrier IP address with eth0 for a new network interface.
+      AssociatePublicIpAddress: false     # optional, Indicates whether to assign a public IPv4 address to eth0 for a new network interface.
+      DeleteOnTermination: false          # optional, Indicates whether the network interface is deleted when the instance is terminated.
+      Description: "..."                  # optional, A description for the network interface.
+      DeviceIndex: 0                      # required, The device index for the network interface attachment.
+      InterfaceType: "interface"          # optional, The interface type for the network interface. To create an Elastic Fabric Adapter (EFA), specify efa
+      NetworkCardIndex: 0                 # optional, The index of the network card. Some instance types support multiple network cards. The primary network interface must be assigned to network card index 0. The default is network card index 0.
```
bwagner5, line 86: just some casing nits: 
```suggestion
    - associateCarrierIPAddress: false    # optional, Indicates whether to associate a Carrier IP address with eth0 for a new network interface.
      associatePublicIPAddress: false     # optional, Indicates whether to assign a public IPv4 address to eth0 for a new network interface.
      deleteOnTermination: false          # optional, Indicates whether the network interface is deleted when the instance is terminated.
      description: "..."                  # optional, A description for the network interface.
      deviceIndex: 0                      # required, The device index for the network interface attachment.
      interfaceType: "interface"          # optional, The interface type for the network interface. To create an Elastic Fabric Adapter (EFA), specify efa
      networkCardIndex: 0 
```

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,129 @@
...
+      ipv4PrefixCount: 1                  # optional, The number of IPv4 prefixes that AWS automatically assigns to the network interface.
+      ipv6PrefixCount: 1                  # optional, The number of IPv6 prefixes that AWS automatically assigns to the network interface.
+    - { ... }
+```
+
+in addition, we will only support two networkInterfaces per node, this way we guarantee that the node will be able to
```
bwagner5, line 92: can we remove this since we're going to leave it up to the user to make sure the instance types they specify can support the number of network interface they specify?

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,129 @@
...
+| AWSNodeTemplate.NetworkInterfaces[@].Description               | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].DeviceIndex               | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].InterfaceType             | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].NetworkCardIndex          | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].ipv4PrefixCount           | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].ipv6PrefixCount           | x      |         |            |
+
+since `AWSNodeTemplate.NetworkInterfaces` is a list, Drift would be triggered if:
+
+- the number of network interfaces of the node do not match the length of `AWSNodeTemplate.NetworkInterfaces`
```
bwagner5, line 125:
```suggestion
- the number of network interfaces of the node does not contain the specified `AWSNodeTemplate.NetworkInterfaces`
```

pkg/apis/crds/karpenter.k8s.aws_awsnodetemplates.yaml
```diff
@@ -193,6 +193,61 @@ spec:
...
+                      type: string
+                    ipv4PrefixCount:
+                      description: The number of IPv4 prefixes to be automatically
+                        assigned to the network interface. You cannot use this option
+                        if you use the Ipv4Prefix option.
+                      format: int64
+                      type: integer
+                    ipv6PrefixCount:
+                      description: The number of IPv6 prefixes to be automatically
+                        assigned to the network interface. You cannot use this option
```
bwagner5, line 238: can you remove the last sentence since we're not supporting the `ipv6Prefix` property?

pkg/apis/crds/karpenter.k8s.aws_awsnodetemplates.yaml
```diff
@@ -193,6 +193,61 @@ spec:
...
+                      description: "The type of network interface. To create an Elastic
+                        Fabric Adapter (EFA), specify efa. For more information, see
+                        Elastic Fabric Adapter (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa.html)
+                        in the Amazon Elastic Compute Cloud User Guide. \n If you
+                        are not creating an EFA, specify interface or omit this parameter.
+                        \n Valid values: interface | efa"
+                      type: string
+                    ipv4PrefixCount:
+                      description: The number of IPv4 prefixes to be automatically
+                        assigned to the network interface. You cannot use this option
```
bwagner5, line 232: can you remove the last sentence here since we're not supporting the `ipv4Prefix` property?

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,129 @@
...
+The [drift design doc #366](https://github.com/aws/karpenter-core/pull/366) suggests classifying CRD fields as  (1) Static, (2) Dynamic, or (3) Behavioral. in this
+context, the networkInterfaces would be classified as following:
+
+|                                                                | Static | Dynamic | Behavioral |
+|----------------------------------------------------------------|--------|---------|------------|
+| AWSNodeTemplate.NetworkInterfaces.length                       | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces.order                        | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].AssociateCarrierIpAddress | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].AssociatePublicIpAddress  | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].DeleteOnTermination       |        |         | x          |
```
bwagner5, line 115: I think this should be static as well. This is describable via an EC2 DescribeInstances call: https://docs.aws.amazon.com/cli/latest/reference/ec2/run-instances.html#:~:text=%22DeleteOnTermination%22%3A%20true%2C

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,129 @@
...
+| AWSNodeTemplate.NetworkInterfaces[@].ipv4PrefixCount           | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].ipv6PrefixCount           | x      |         |            |
+
+since `AWSNodeTemplate.NetworkInterfaces` is a list, Drift would be triggered if:
+
+- the number of network interfaces of the node do not match the length of `AWSNodeTemplate.NetworkInterfaces`
+- the order of network interfaces of a node do not match the order of `AWSNodeTemplate.NetworkInterfaces` when sorted
+  by `NetworkCardIndex` and `DeviceIndex`.
+  `AWSNodeTemplate.NetworkInterfaces` do not define `NetworkCardIndex` or `DeviceIndex`, their values are assumed to
+  equal the index of the  `v1alpha1.NetworkInterface` object in th list
```
bwagner5, line 129:
```suggestion
  equal the index of the  `v1alpha1.NetworkInterface` object in the list
```

test/suites/integration/networkinterfaces_test.go
```diff
@@ -0,0 +1,116 @@
...
+				SecurityGroupSelector: map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName},
+				SubnetSelector:        map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName},
+				LaunchTemplate: v1alpha1.LaunchTemplate{
+					NetworkInterfaces: []*v1alpha1.NetworkInterface{
+						{
+							AssociatePublicIPAddress: aws.Bool(true),
+							Description:              aws.String(desc),
+							DeviceIndex:              aws.Int64(0),
+							NetworkCardIndex:         aws.Int64(0),
+							IPv4PrefixCount:          aws.Int64(2),
```
bwagner5, line 49: since prefix delegation is only available for nitro instance types, you'll need to add a Provisioner requirement:

```
key: karpenter.k8s.aws/instance-hypervisor
operator: In
values: ["nitro"]
```

also, Is it valid to do both an ipv4 and ipv6 prefix?

bwagner5 (commented):
nice work! This is looking really good! 

Can you add some suite_tests as well, probably in https://github.com/aws/karpenter/blob/main/pkg/providers/launchtemplate/suite_test.go

designs/aws-launch-template-network-interfaces.md
myaser: > I think this should be static as well. This is describable via an EC2 DescribeInstances call

True, I just thought it does not benefit much to have it. 
if the current instance has the value `false`, while the CRD changed to `true`. decommissioning the node for the drift will leave the networkInterface un deleted anyway. and vice versa. it makes more sense for an in-place drift fixing but this is not going to be supported.

bwagner5: ah I see your point

myaser:
> Can you add some suite_tests as well, probably in https://github.com/aws/karpenter/blob/main/pkg/providers/launchtemplate/suite_test.go

I am not sure what test scenarios to be added there, I also added some tests to `karpenter/test/suites/integration/networkinterfaces_test.go`.

Are they not sufficient or if you prefer them moved to `karpenter/pkg/providers/launchtemplate/suite_test.go`

please advice

test/suites/integration/networkinterfaces_test.go
myaser: I added the requirements

> Is it valid to do both an ipv4 and ipv6 prefix?
I do not know 🤔

bwagner5:
> > Can you add some suite_tests as well, probably in https://github.com/aws/karpenter/blob/main/pkg/providers/launchtemplate/suite_test.go
> 
> I am not sure what test scenarios to be added there, I also added some tests to `karpenter/test/suites/integration/networkinterfaces_test.go`.
> 
> Are they not sufficient or if you prefer them moved to `karpenter/pkg/providers/launchtemplate/suite_test.go`
> 
> please advice

The suite_tests are more like unit-tests (i.e. mocked dependencies) that run faster, so I think we should check that the parameters are getting set properly in the suite tests in addition to the e2e tests that operate on a real cluster.

myaser:
> > > Can you add some suite_tests as well, probably in https://github.com/aws/karpenter/blob/main/pkg/providers/launchtemplate/suite_test.go
> > 
> > 
> > I am not sure what test scenarios to be added there, I also added some tests to `karpenter/test/suites/integration/networkinterfaces_test.go`.
> > Are they not sufficient or if you prefer them moved to `karpenter/pkg/providers/launchtemplate/suite_test.go`
> > please advice
> 
> The suite_tests are more like unit-tests (i.e. mocked dependencies) that run faster, so I think we should check that the parameters are getting set properly in the suite tests in addition to the e2e tests that operate on a real cluster.
> 
> For example, these tests were recently added https://github.com/aws/karpenter/blob/main/pkg/providers/launchtemplate/suite_test.go#L1626-L1646
> 
> A good test for this feature would be to create a NodeTemplate with NetworkInterfaces specified and then check the LaunchTemplate mock to see if the parameters were set properly.

Tests have been added.

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,126 @@
...
+this way we can support different subnets and security groups for each network interface, we can also remove the limit
+of having only 2 network interfaces per LaunchTemplate.
+
+The [drift design doc #366](https://github.com/aws/karpenter-core/pull/366) suggests classifying CRD fields as  (1) Static, (2) Dynamic, or (3) Behavioral. in this
+context, the networkInterfaces would be classified as following:
+
+|                                                                | Static | Dynamic | Behavioral |
+|----------------------------------------------------------------|--------|---------|------------|
+| AWSNodeTemplate.NetworkInterfaces.length                       | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces.order                        | x      |         |            |
```
njtran, line 109: Does the order in which network interfaces are defined define some priority on how they're being used? If so, then this makes sense to me, but otherwise, I don't know if we need to consider the order.

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,126 @@
...
+| AWSNodeTemplate.NetworkInterfaces[@].NetworkCardIndex          | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].ipv4PrefixCount           | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].ipv6PrefixCount           | x      |         |            |
+
+since `AWSNodeTemplate.NetworkInterfaces` is a list, Drift would be triggered if:
+
+- the number of network interfaces of the node does not contain the specified `AWSNodeTemplate.NetworkInterfaces`
+- the order of network interfaces of a node do not match the order of `AWSNodeTemplate.NetworkInterfaces` when sorted
+  by `NetworkCardIndex` and `DeviceIndex`.
+  `AWSNodeTemplate.NetworkInterfaces` do not define `NetworkCardIndex` or `DeviceIndex`, their values are assumed to
```
njtran, line 125: Sorry if I'm missing the point here, but I see that you've included both NetworkCardIndex and DeviceIndex in the specifications. Do you mean to add an if here? 
```suggestion
  If `AWSNodeTemplate.NetworkInterfaces` do not define `NetworkCardIndex` or `DeviceIndex`, their values are assumed to
```

njtran (commented):
Just looked at the design doc. Let's discuss on slack if you'd like

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,126 @@
...
+  tags: { ... }                           # optional, propagates tags to underlying EC2 resources
+  metadataOptions: { ... }                # optional, configures IMDS for the instance
+  blockDeviceMappings: [ ... ]            # optional, configures storage devices for the instance
+  detailedMonitoring: "..."               # optional, configures detailed monitoring for the instance
+  networkInterfaces:                      # optional, configures network interfaces for the instance
+    - associateCarrierIPAddress: false    # optional, Indicates whether to associate a Carrier IP address with eth0 for a new network interface.
+      associatePublicIPAddress: false     # optional, Indicates whether to assign a public IPv4 address to eth0 for a new network interface.
+      deleteOnTermination: false          # optional, Indicates whether the network interface is deleted when the instance is terminated.
+      description: "..."                  # optional, A description for the network interface.
+      deviceIndex: 0                      # required, The device index for the network interface attachment.
```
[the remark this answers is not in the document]
myaser: It is not required; it was documented wrong! I am correcting it in the docs

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,126 @@
...
+      interfaceType: "interface"          # optional, The interface type for the network interface. To create an Elastic Fabric Adapter (EFA), specify efa
+      networkCardIndex: 0
```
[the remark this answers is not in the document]
myaser: optional, added a comment

designs/aws-launch-template-network-interfaces.md
myaser: yes, thanks for spotting it

designs/aws-launch-template-network-interfaces.md
```diff
@@ -0,0 +1,126 @@
...
+| AWSNodeTemplate.NetworkInterfaces[@].AssociatePublicIpAddress  | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].DeleteOnTermination       |        |         | x          |
+| AWSNodeTemplate.NetworkInterfaces[@].Description               | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].DeviceIndex               | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].InterfaceType             | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].NetworkCardIndex          | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].ipv4PrefixCount           | x      |         |            |
+| AWSNodeTemplate.NetworkInterfaces[@].ipv6PrefixCount           | x      |         |            |
+
+since `AWSNodeTemplate.NetworkInterfaces` is a list, Drift would be triggered if:
```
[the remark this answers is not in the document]
myaser: > The initial design is to default the behavior to terminate and replace. Maybe this is a reasonable case to consider in-place as it would be its own subspec? Would terminate and replace be too unreasonable as a initial landing place for this design?

Agree, it is a reasonable case for the in-place update, but it was mentioned by the original design doc for drift as a no-go. I think it is still reasonable for terminate and replace, except for DeleteOnTermination ([reason](https://github.com/aws/karpenter/pull/3819#discussion_r1229400842)) which is marked as `behavioral`

designs/aws-launch-template-network-interfaces.md
myaser: Yea, the order does not change the way they are used. It is the device index that matters.

after reconsiderations, the order in which they are defined does not matter, we should only compare the network interfaces where the device index match

I think in this case it is better to add the validation to have unique value per device index in the `AWSNodeTemplate` (already enforced by AWS API but better to have the AWSNodeTemplate in a valid state as well)

bwagner5:
I'm having trouble getting this new functionality to work. It seems that CreateFleet ignores (no errors) the LT parameters and provisions an instance with a default primary network interface. @myaser Have you had luck getting instances provisioned with the parameters from the AWSNodeTemplate?

myaser:
> I'm having trouble getting this new functionality to work. It seems that CreateFleet ignores (no errors) the LT parameters and provisions an instance with a default primary network interface. @myaser Have you had luck getting instances provisioned with the parameters from the AWSNodeTemplate?

🤔, yes I could provision instances with the desired parameters. 
for that we also have the `pkg/providers/launchtemplate/suite_test.go`  checking the parameters match between the `AWSNodeTemplate` and `LaunchTemplate`

bwagner5:
> > I'm having trouble getting this new functionality to work. It seems that CreateFleet ignores (no errors) the LT parameters and provisions an instance with a default primary network interface. @myaser Have you had luck getting instances provisioned with the parameters from the AWSNodeTemplate?
> 
> 🤔, yes I could provision instances with the desired parameters. for that we also have the `pkg/providers/launchtemplate/suite_test.go` checking the parameters match between the `AWSNodeTemplate` and `LaunchTemplate`

The tests work fine and even the real requests to EC2 are what I expect. The problem is that EC2 Fleet is not respecting the LaunchTemplate parameters. So even if I try to allocate Ipv4PrefixCount, Fleet ignores it and the instance is not launched with a prefix allocation even though it was requested in the launch template.

bwagner5:
> > > I'm having trouble getting this new functionality to work. It seems that CreateFleet ignores (no errors) the LT parameters and provisions an instance with a default primary network interface. @myaser Have you had luck getting instances provisioned with the parameters from the AWSNodeTemplate?
> > 
> > 
> > 🤔, yes I could provision instances with the desired parameters. for that we also have the `pkg/providers/launchtemplate/suite_test.go` checking the parameters match between the `AWSNodeTemplate` and `LaunchTemplate`
> 
> The tests work fine and even the real requests to EC2 are what I expect. The problem is that EC2 Fleet is not respecting the LaunchTemplate parameters. So even if I try to allocate Ipv4PrefixCount, Fleet ignores it and the instance is not launched with a prefix allocation even though it was requested in the launch template.

Just wanted to ping in and say this is still being worked by the EC2 Fleet team.  It appears if a SubnetID override is specified in the Fleet request, the NetworkInterfaces block is completely ignored. Will update here as we get more info.

myaser:
> > > > I'm having trouble getting this new functionality to work. It seems that CreateFleet ignores (no errors) the LT parameters and provisions an instance with a default primary network interface. @myaser Have you had luck getting instances provisioned with the parameters from the AWSNodeTemplate?
> > > 
> > > 
> > > 🤔, yes I could provision instances with the desired parameters. for that we also have the `pkg/providers/launchtemplate/suite_test.go` checking the parameters match between the `AWSNodeTemplate` and `LaunchTemplate`
> > 
> > 
> > The tests work fine and even the real requests to EC2 are what I expect. The problem is that EC2 Fleet is not respecting the LaunchTemplate parameters. So even if I try to allocate Ipv4PrefixCount, Fleet ignores it and the instance is not launched with a prefix allocation even though it was requested in the launch template.
> 
> Just wanted to ping in and say this is still being worked by the EC2 Fleet team. It appears if a SubnetID override is specified in the Fleet request, the NetworkInterfaces block is completely ignored. Will update here as we get more info.


perhaps it is worth mentioning that I tested it with a private subnet (`MapPublicIpOnLaunch: false`) and I set a networkInterface with `AssociatePublicIpAddress` true.
I see the nodes already have auto-assigned public IPs

so maybe not all fields are ignored. perhaps we can bitesize this and defer the problematic fields.
I am currently running this patch version on some of our clusters, specifically for the usecase of assigning public IP addresses

myaser:
@bwagner5 

Since this is blocked and needs modifications on EC2 fleet API, I would like to bitesize this and craft a PR to carry only one very important field to us right now, `AssociatePublicIpAddress`.
We can not convert the subnets to public for security and compliance reasons, so we are currently running this patch version on some of our test clusters. 

I am afraid this patch will soon become outdated with latest karpenter versions and will be hard to maintain.

If this design/approach is  is approved, I can create a PR that contain this field alone. when the fleet API support the other cases we can expand

WDYT?

bwagner5:
> @bwagner5
> 
> Since this is blocked and needs modifications on EC2 fleet API, I would like to bitesize this and craft a PR to carry only one very important field to us right now, `AssociatePublicIpAddress`. We can not convert the subnets to public for security and compliance reasons, so we are currently running this patch version on some of our test clusters.
> 
> I am afraid this patch will soon become outdated with latest karpenter versions and will be hard to maintain.
> 
> If this design/approach is is approved, I can create a PR that contain this field alone. when the fleet API support the other cases we can expand
> 
> WDYT?

Apologies for the delayed reply, I've been on vacation for the last few weeks. 

Interesting! I think it's fine to implement this in pieces until the EC2 Fleet team is able to fix the issue we observed with the NetworkInterfaces. I'll take a look at your other PR and reproduce the Fleet propagation of the param. Thanks for your patience!

bwagner5:
going to close this one in-favor of https://github.com/aws/karpenter/pull/4353

PR #3824, opened 2023-04-28
feat: Adds vcpu and memory metrics per instance type

valdisrigdon:
**Description**

This adds `karpenter_cloudprovider_instance_type_cpu_cores` and `karpenter_cloudprovider_instance_type_memory_bytes` metrics for each instance type that Karpenter understands.

These new metrics can be combined with the `karpenter_cloudprovider_instance_price_estimate` metric to give you a price per VCPU and price per 1 GiB of memory to help calculate your hourly cost per Pod.

* `make test`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/test/environment.go
+++ b/pkg/test/environment.go
@@ -33,6 +33,8 @@ import (
 	"github.com/aws/karpenter/pkg/providers/subnet"
 
 	coretest "github.com/aws/karpenter-core/pkg/test"
+
+	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
 type Environment struct {
@@ -146,4 +148,15 @@ func (env *Environment) Reset() {
 	env.LaunchTemplateCache.Flush()
 	env.SubnetCache.Flush()
 	env.SecurityGroupCache.Flush()
+
+	mfs, err := crmetrics.Registry.Gather()
+	if err != nil {
+		for _, mf := range mfs {
+			for _, metric := range mf.GetMetric() {
+				if metric != nil {
+					metric.Reset()
+				}
+			}
+		}
+	}
 }
--- a/website/content/en/preview/concepts/metrics.md
+++ b/website/content/en/preview/concepts/metrics.md
@@ -94,7 +94,13 @@ Pod state is the current state of pods. This metric can be used several ways as 
 ### `karpenter_cloudprovider_duration_seconds`
 Duration of cloud provider method calls. Labeled by the controller, method name and provider.
 
-### `karpenter_cloudprovider_instance_price_estimate`
+### `karpenter_cloudprovider_instance_type_cpu_cores`
+VCPUs cores for a given instance type.
+
+### `karpenter_cloudprovider_instance_type_memory_bytes`
+Memory, in bytes, for a given instance type.
+
+### `karpenter_cloudprovider_instance_type_price_estimate`
 Estimated hourly price used when making informed decisions on node cost calculation. This is updated once on startup and then every 12 hours.
 
 ## Cloudprovider Batcher Metrics
--- a/pkg/providers/instancetype/instancetype.go
+++ b/pkg/providers/instancetype/instancetype.go
@@ -21,6 +21,8 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/prometheus/client_golang/prometheus"
+
 	awscache "github.com/aws/karpenter/pkg/cache"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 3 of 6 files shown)


valdisrigdon:
@jonathan-innis Thoughts on this? Instead of trying to break out the instance types and make them available to `pricing` I ended up just generating a file with `cpu` and `memory`, which seems reasonable to me.

jonathan-innis:
> Thoughts on this?

Is it possible to get this info directly from GetInstanceTypes() call inside of the InstanceType provider. This has all the EC2 information for the instance type and we could use that for updating the metrics. Essentially, any time there is an update to that response, we would update the metric info. I think that would avoid us having to store static files that have these values.

valdisrigdon:
> Is it possible to get this info directly from GetInstanceTypes() call inside of the InstanceType provider. This has all the EC2 information for the instance type and we could use that for updating the metrics. Essentially, any time there is an update to that response, we would update the metric info. I think that would avoid us having to store static files that have these values.

It is, but separating this out from `instancetype` without coupling `pricing` is a bit challenging, as most of `instancetype` is dealing with the cache.

I'll take another stab at that -- I sort of figured the code gen solution wasn't the way forward, but wanted to throw it out there.

jonathan-innis:
> I sort of figured the code gen solution wasn't the way forward

I think starting looking in [this area](https://github.com/aws/karpenter/blob/main/pkg/providers/instancetype/instancetype.go#L85) is a great place to start!

website/content/en/preview/concepts/metrics.md
```diff
@@ -97,6 +97,12 @@ Duration of cloud provider method calls. Labeled by the controller, method name
 ### `karpenter_cloudprovider_instance_price_estimate`
```
jonathan-innis, line 97:
```suggestion
### `karpenter_cloudprovider_instance_type_price_estimate`
```

Maybe change this so that we align all of these under the `instance_type` namespace

pkg/test/environment.go
```diff
@@ -146,4 +148,15 @@ func (env *Environment) Reset() {
 	env.LaunchTemplateCache.Flush()
 	env.SubnetCache.Flush()
 	env.SecurityGroupCache.Flush()
+
+	mfs, err := crmetrics.Registry.Gather()
```
jonathan-innis, line 152: Can you explain what this is here for?

website/content/en/preview/concepts/metrics.md
valdisrigdon: done

pkg/providers/pricing/metrics.go
```diff
@@ -34,7 +34,7 @@ var (
 		prometheus.GaugeOpts{
 			Namespace: metrics.Namespace,
 			Subsystem: cloudProviderSubsystem,
-			Name:      "instance_price_estimate",
+			Name:      "instance_type_price_estimate",
```
jonathan-innis, line 37:
```suggestion
			Name:      "instance_type_price",
```

Thinking about this now. Do you think there's any reason to label the metric as estimate and not just put price? It shortens the name and the description mentions that this is an estimate.

pkg/test/environment.go
jonathan-innis: Got it. Then, this might be better suited to the `ExpectCleanedUp` function. I'm fine with this here for now, but we should consider resetting metrics generally from the `karpenter-core` test package

website/content/en/preview/concepts/metrics.md
```diff
@@ -94,9 +94,15 @@ Pod state is the current state of pods. This metric can be used several ways as
...
 Duration of cloud provider method calls. Labeled by the controller, method name and provider.
 
-### `karpenter_cloudprovider_instance_price_estimate`
+### `karpenter_cloudprovider_instance_type_memory`
+Memory, in Mib, for a given instance type.
+
+### `karpenter_cloudprovider_instance_type_price_estimate`
 Estimated hourly price used when making informed decisions on node cost calculation. This is updated once on startup and then every 12 hours.
 
+### `karpenter_cloudprovider_instance_type_vcpu`
```
jonathan-innis, line 103:
```suggestion
### `karpenter_cloudprovider_instance_type_cpu_cores`
```

`kube_state` calls them `cpu_cores` and `memory_bytes`. I wonder if we should align with that: https://github.com/kubernetes/kube-state-metrics/blob/main/docs/pod-metrics.md

website/content/en/preview/concepts/metrics.md
```diff
@@ -94,9 +94,15 @@ Pod state is the current state of pods. This metric can be used several ways as
 ### `karpenter_cloudprovider_duration_seconds`
 Duration of cloud provider method calls. Labeled by the controller, method name and provider.
 
-### `karpenter_cloudprovider_instance_price_estimate`
+### `karpenter_cloudprovider_instance_type_memory`
```
jonathan-innis, line 97:
```suggestion
### `karpenter_cloudprovider_instance_type_memory_bytes`
```

`kube_state` calls them `cpu_cores` and `memory_bytes`. I wonder if we should align with that: https://github.com/kubernetes/kube-state-metrics/blob/main/docs/pod-metrics.md

pkg/providers/pricing/metrics.go
valdisrigdon: I'd like to leave `estimate` as it's more than likely not the actual price paid for that spot instance, but rather what Karpenter thinks the market price is.

website/content/en/preview/concepts/metrics.md
valdisrigdon: It does this for `overhead` which is going to typically be much smaller numbers. Thinking in terms of `bytes` for an ec2 instance seems like the wrong scale -- I was leaning towards the other way and converting MiB to Gb, but passed on that.

I can change this if you feel strongly about it, but IMO reading really large numbers makes it difficult to understand this metric.

jonathan-innis: I think for visualization solutions like Grafana, the assumption is that you will provide the raw value and then the visualization solution will be responsible for doing the conversion for you. That gives you the most flexibility with your data: https://community.grafana.com/t/grafana-unit-for-mb-gb-etc/68495/4

valdisrigdon: Fair enough! I've changed this to `memory_bytes`.

jonathan-innis (approved):
LGTM 🚀

jonathan-innis (dismissed):
LGTM 🚀

valdisrigdon:
@jonathan-innis fixed the flakey test, ready for re-review

jonathan-innis (approved):
LGTM 🚀 Great work!

PR #3845, opened 2023-05-02
chore: Restore GPU labels and use DoesNotExist

jonathan-innis:
Fixes # 

**Description**

Changes the instance type labels to more closely align with the `describe-instance-types` API response for EC2. It also changes the AMI selection logic to use the `DoesNotExist` and `Exist` requirement rather than selecting AMIs on very specific labels by manufacturer name

* `make presubmit`
* `FOCUS=Scheduling make e2etests`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/apis/v1alpha1/register.go
+++ b/pkg/apis/v1alpha1/register.go
@@ -38,6 +38,10 @@ var (
 		"x86_64":                   v1alpha5.ArchitectureAmd64,
 		v1alpha5.ArchitectureArm64: v1alpha5.ArchitectureArm64,
 	}
+	WellKnownArchitectures = sets.NewString(
+		v1alpha5.ArchitectureAmd64,
+		v1alpha5.ArchitectureArm64,
+	)
 	RestrictedLabelDomains = []string{
 		LabelDomain,
 	}
@@ -82,21 +86,14 @@ var (
 	LabelInstanceMemory                       = LabelDomain + "/instance-memory"
 	LabelInstanceNetworkBandwidth             = LabelDomain + "/instance-network-bandwidth"
 	LabelInstancePods                         = LabelDomain + "/instance-pods"
-	// TODO: will deprecate at v1beta1, remove at v1
-	LabelInstanceGPUName = LabelDomain + "/instance-gpu-name"
-	// TODO: will deprecate at v1beta1, remove at v1
-	LabelInstanceGPUManufacturer = LabelDomain + "/instance-gpu-manufacturer"
-	// TODO: will deprecate at v1beta1, remove at v1
-	LabelInstanceGPUCount = LabelDomain + "/instance-gpu-count"
-	// TODO: will deprecate at v1beta1, remove at v1
-	LabelInstanceGPUMemory               = LabelDomain + "/instance-gpu-memory"
-	LabelInstanceAMIID                   = LabelDomain + "/instance-ami-id"
-	LabelInstanceAcceleratorName         = LabelDomain + "/instance-accelerator-name"
-	LabelInstanceAcceleratorManufacturer = LabelDomain + "/instance-accelerator-manufacturer"
-	LabelInstanceAcceleratorCount        = LabelDomain + "/instance-accelerator-count"
-	LabelInstanceAcceleratorMemory       = LabelDomain + "/instance-accelerator-memory"
-
-	InterruptionInfrastructureFinalizer = Group + "/interruption-infrastructure"
+	LabelInstanceGPUName                      = LabelDomain + "/instance-gpu-name"
+	LabelInstanceGPUManufacturer              = LabelDomain + "/instance-gpu-manufacturer"
+	LabelInstanceGPUCount                     = LabelDomain + "/instance-gpu-count"
+	LabelInstanceGPUMemory                    = LabelDomain + "/instance-gpu-memory"
+	LabelInstanceAMIID                        = LabelDomain + "/instance-ami-id"
+	LabelInstanceAcceleratorName              = LabelDomain + "/instance-accelerator-name"
+	LabelInstanceAcceleratorManufacturer      = LabelDomain + "/instance-accelerator-manufacturer"
+	LabelInstanceAcceleratorCount             = LabelDomain + "/instance-accelerator-count"
 )
 
 var (
@@ -136,7 +133,6 @@ func init() {
 		LabelInstanceAcceleratorName,
 		LabelInstanceAcceleratorManufacturer,
 		LabelInstanceAcceleratorCount,
-		LabelInstanceAcceleratorMemory,
 	)
 }
 
--- a/pkg/providers/instancetype/suite_test.go
+++ b/pkg/providers/instancetype/suite_test.go
@@ -167,10 +167,9 @@ var _ = Describe("Instance Types", func() {
 			v1alpha1.LabelInstanceGPUCount:                     "1",
 			v1alpha1.LabelInstanceGPUMemory:                    "16384",
 			v1alpha1.LabelInstanceLocalNVME:                    "900",
-			v1alpha1.LabelInstanceAcceleratorName:              "t4",
-			v1alpha1.LabelInstanceAcceleratorManufacturer:      "nvidia",
+			v1alpha1.LabelInstanceAcceleratorName:              "inferentia",
+			v1alpha1.LabelInstanceAcceleratorManufacturer:      "aws",
 			v1alpha1.LabelInstanceAcceleratorCount:             "1",
-			v1alpha1.LabelInstanceAcceleratorMemory:            "16384",
 			// Deprecated Labels
 			v1.LabelFailureDomainBetaRegion: "",
 			v1.LabelFailureDomainBetaZone:   "test-zone-1a",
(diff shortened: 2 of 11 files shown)


jonathan-innis (commented):
/karpenter snapshot

pkg/apis/v1alpha1/register.go
```diff
@@ -82,21 +86,14 @@ var (
...
-	LabelInstanceGPUCount = LabelDomain + "/instance-gpu-count"
-	// TODO: will deprecate at v1beta1, remove at v1
-	LabelInstanceGPUMemory               = LabelDomain + "/instance-gpu-memory"
-	LabelInstanceAMIID                   = LabelDomain + "/instance-ami-id"
-	LabelInstanceAcceleratorName         = LabelDomain + "/instance-accelerator-name"
-	LabelInstanceAcceleratorManufacturer = LabelDomain + "/instance-accelerator-manufacturer"
-	LabelInstanceAcceleratorCount        = LabelDomain + "/instance-accelerator-count"
-	LabelInstanceAcceleratorMemory       = LabelDomain + "/instance-accelerator-memory"
-
-	InterruptionInfrastructureFinalizer = Group + "/interruption-infrastructure"
```
njtran, line 99: Do we want to remove this?

pkg/apis/v1alpha1/register.go
```diff
@@ -136,7 +133,6 @@ func init() {
 		LabelInstanceAcceleratorName,
 		LabelInstanceAcceleratorManufacturer,
 		LabelInstanceAcceleratorCount,
-		LabelInstanceAcceleratorMemory,
```
njtran, line 139: Just to confirm, is this being removed in favor of LabelInstanceGPUMemory?

pkg/providers/instancetype/suite_test.go
```diff
@@ -167,10 +167,9 @@ var _ = Describe("Instance Types", func() {
 			v1alpha1.LabelInstanceGPUCount:                     "1",
 			v1alpha1.LabelInstanceGPUMemory:                    "16384",
 			v1alpha1.LabelInstanceLocalNVME:                    "900",
-			v1alpha1.LabelInstanceAcceleratorName:              "t4",
-			v1alpha1.LabelInstanceAcceleratorManufacturer:      "nvidia",
+			v1alpha1.LabelInstanceAcceleratorName:              "inferentia",
+			v1alpha1.LabelInstanceAcceleratorManufacturer:      "aws",
```
njtran, line 171: Does this actually change anything? Shouldn't this work identically to before?

pkg/providers/instancetype/suite_test.go
```diff
@@ -231,8 +228,6 @@ var _ = Describe("Instance Types", func() {
 			"topology.ebs.csi.aws.com/zone": "test-zone-1a",
 		}
 
-		// Ensure that we're exercising all well known labels
-		Expect(lo.Keys(nodeSelector)).To(ContainElements(append(v1alpha5.WellKnownLabels.UnsortedList(), lo.Keys(v1alpha5.NormalizedLabels)...)))
```
njtran, line 235: Why are we removing this check in the suite_test? I think combined instance type labels and individual instance type labels should both make sure we're exercising all well known labels.

pkg/providers/instancetype/suite_test.go
```diff
@@ -218,10 +219,6 @@ var _ = Describe("Instance Types", func() {
 			v1alpha1.LabelInstanceGPUCount:                     "1",
 			v1alpha1.LabelInstanceGPUMemory:                    "16384",
 			v1alpha1.LabelInstanceLocalNVME:                    "900",
-			v1alpha1.LabelInstanceAcceleratorName:              "t4",
-			v1alpha1.LabelInstanceAcceleratorManufacturer:      "nvidia",
-			v1alpha1.LabelInstanceAcceleratorCount:             "1",
-			v1alpha1.LabelInstanceAcceleratorMemory:            "16384",
```
njtran, line 224: Don't we only need to remove this line?

pkg/providers/instancetype/types.go
```diff
@@ -118,10 +117,6 @@ func computeRequirements(ctx context.Context, info *ec2.InstanceTypeInfo, offeri
 	// GPU Labels
 	if info.GpuInfo != nil && len(info.GpuInfo.Gpus) == 1 {
 		gpu := info.GpuInfo.Gpus[0]
-		requirements.Get(v1alpha1.LabelInstanceAcceleratorName).Insert(lowerKabobCase(aws.StringValue(gpu.Name)))
-		requirements.Get(v1alpha1.LabelInstanceAcceleratorManufacturer).Insert(lowerKabobCase(aws.StringValue(gpu.Manufacturer)))
-		requirements.Get(v1alpha1.LabelInstanceAcceleratorCount).Insert(fmt.Sprint(aws.Int64Value(gpu.Count)))
-		requirements.Get(v1alpha1.LabelInstanceAcceleratorMemory).Insert(fmt.Sprint(aws.Int64Value(gpu.MemoryInfo.SizeInMiB)))
```
njtran, line 124: Same here, don't we only need to remove this line?

test/suites/integration/scheduling_test.go
```diff
@@ -191,14 +191,10 @@ var _ = Describe("Scheduling", Ordered, ContinueOnFailure, func() {
 	})
 	It("should support well-known labels for an accelerator (nvidia)", func() {
 		nodeSelector := map[string]string{
-			v1alpha1.LabelInstanceGPUName:                 "t4",
-			v1alpha1.LabelInstanceGPUMemory:               "16384",
-			v1alpha1.LabelInstanceGPUManufacturer:         "nvidia",
-			v1alpha1.LabelInstanceGPUCount:                "1",
-			v1alpha1.LabelInstanceAcceleratorName:         "t4",
-			v1alpha1.LabelInstanceAcceleratorMemory:       "16384",
```
njtran, line 199: Same here

pkg/apis/v1alpha1/register.go
jonathan-innis: I'm not sure which thing you are refering to here but the only thing that's being removed is the `LabelInstanceAcceleratorMemory` and the `InterruptionInfrastructureFinalizer` which is an artifact of code changes that are no longer valid

pkg/apis/v1alpha1/register.go
jonathan-innis: GPU memory was always being used. The accelerator info that is returned by the EC2 api does not have any information on memory

jonathan-innis: See an example response for an inferentia instance

```console
"InferenceAcceleratorInfo": {
                "Accelerators": [
                    {
                        "Count": 1,
                        "Name": "Inferentia",
                        "Manufacturer": "AWS"
                    }
                ]
            },
```

pkg/providers/instancetype/suite_test.go
jonathan-innis: Yes, it does but "nvidia" is no longer a valid accelerator manufacturer since we aren't propagating that label for GPUs with this change

pkg/providers/instancetype/suite_test.go
jonathan-innis: No, since the groups of instances that have accelerators and those that have GPUs are distinct, there's no way to combine all the labels and satisfy a single instance type

pkg/providers/instancetype/suite_test.go
jonathan-innis: Removing since there's no way to ensure that we are exercising all the labels through combination anymore since there's no instance type that will satisfy all the labels that we need to test.

I think ensuring we are testing all the labels on the individual label testing should be sufficient to ensure that we make sure that all the labels work

test/suites/integration/scheduling_test.go
jonathan-innis: We are testing accelerators down below

pkg/providers/instancetype/suite_test.go
njtran: Ah I see, can we make one suite_test for the `LabelInstanceGPU*` labels like you have here and one with the `LabelInstanceAccelerator*` for completeness then?

pkg/providers/instancetype/suite_test.go
njtran: Ah makes sense, I'd still like to make sure that this fails in the case of a new WellKnownLabel that's added, but not included in this test. Can we check that all the well known labels are used and either the set of GPU/Accelerator Labels are used? Something like
```
Expect(lo.Keys(nodeSelector)).To(ContainElements(append(v1alpha5.WellKnownLabels.UnsortedList().WithoutGPULabels(), lo.Keys(v1alpha5.NormalizedLabels)...)))
Expect(lo.Keys(nodeSelector)).To(ContainElements(append(v1alpha5.WellKnownLabels.UnsortedList().OnlyGPULabels(), lo.Keys(v1alpha5.NormalizedLabels)...)))

```

pkg/providers/instancetype/types.go
jonathan-innis: We're creating distinct groups of instance types that have accelerators and those that have gpus

njtran (approved):
LGTM!

PR #3874, opened 2023-05-10
test: Scale testing

jonathan-innis:
Fixes # 

**Description**

Adds scale testing for the following scenarios to Karpenter's test suites to validate Karpenter running against higher scale than is performed against the E2Es:
1. Node-dense (500 node) scale-up
2. Pod-dense (~110 pod/node) scale-up
3. Multi-Deprovisioner scale-down (10 nodes each with 20 pods each)
4. Consolidation Deletion scale-down (200 nodes with 20 pods each)
5. Consolidation Replacement scale-down (20 nodes with 20 pods each)
6. Emptiness Deletion scale-down (200 nodes with 20 pods each)
7. Expiration Replacement scale-down (20 nodes with 20 pods each)
8. Drift Replacement scale-down (20 nodes with 20 pods each)
9. Interruption Replacement scale-down (200 nodes with 20 pods each)

This change also enable pushing CloudWatch duration metrics for sub-pieces of the test through `env.MeasureDurationFor` and enabling the `ENABLE_CLOUDWATCH` environment variable to `true` in your environment.

* `FOCUS="Scale Provisioning make e2etests`
* `FOCUS="Scale Deprovisioning make e2etests`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```


test/suites/scale/deprovisioning_test.go
```diff
@@ -92,11 +92,41 @@ var _ = Describe("Deprovisioning", Label(debug.NoWatch), Label(debug.NoEvents),
...
+			env.EventuallyExpectCreatedNodeCount("==", expectedNodeCount)
+			env.EventuallyExpectInitializedNodeCount("==", expectedNodeCount)
+			env.EventuallyExpectHealthyPodCount(selector, replicas)
+
+			replicas = int(float64(replicas) * 0.2)
+			deployment.Spec.Replicas = lo.ToPtr[int32](int32(replicas))
+			env.ExpectUpdated(deployment)
+
+			provisioner.Spec.Consolidation = &v1alpha5.Consolidation{Enabled: lo.ToPtr(true)}
+			env.ExpectUpdated(provisioner)
```
njtran, line 124: Should have some assertion of an eventually so that this test doesn't just end immediately after updating the provisioner.  

Maybe this is good in the meantime? 
```suggestion
			env.ExpectUpdated(provisioner)
			env.EventuallyExpectDeletedNodeCount("==", createdNodes * 0.2)
```

jonathan-innis: Yep, this is still in draft so I hadn't had the chance to add it yet. Added now!

test/suites/scale/provisioning_test.go
```diff
@@ -0,0 +1,149 @@
...
+			env.EventuallyExpectCreatedNodeCount("==", expectedNodeCount)
+			env.EventuallyExpectInitializedNodeCount("==", expectedNodeCount)
+			env.EventuallyExpectHealthyPodCount(selector, replicas)
+		}, aws.ProvisioningEventType, testGroup, "pod-dense")
+	}, SpecTimeout(10*time.Minute))
+	It("should scale successfully on a pod-dense scale-up", func(_ context.Context) {
+		replicasPerNode := 110
+		maxPodDensity := replicasPerNode + dsCount
+		expectedNodeCount := 60
+		replicas := replicasPerNode * expectedNodeCount
```
njtran, line 122: nit: if we want to make it exactly 110 pods on the node, regardless of dsCount, we could do something like this. Don't think it matters much though.
```suggestion
                 maxPodDensity := 110
		replicasPerNode := maxPodDensity - dsCount
		expectedNodeCount := 60
		replicas := replicasPerNode * expectedNodeCount
```

test/suites/scale/provisioning_test.go
```diff
@@ -0,0 +1,149 @@
...
+		// Get the DS pod count and use it to calculate the DS pod overhead
+		dsCount = env.GetDaemonSetCount(provisioner)
+	})
+	It("should scale successfully on a node-dense scale-up", func(_ context.Context) {
+		replicasPerNode := 1
+		expectedNodeCount := 500
+		replicas := replicasPerNode * expectedNodeCount
+
+		deployment.Spec.Replicas = lo.ToPtr[int32](int32(replicas))
+		// Hostname anti-affinity to require one pod on each node
```
njtran, line 92: question: any reason you chose anti-affinity over choosing pod density? anti-affinity is definitely more expensive, so it'll be more of a strain on Karpenter, if that's what you want.
```
		provisioner.Spec.KubeletConfiguration = &v1alpha5.KubeletConfiguration{
			MaxPods: lo.ToPtr[int32](int32(maxPodDensity)),
		}
```

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,631 @@
...
+						env.EventuallyExpectHealthyPodCount(labels.SelectorFromSet(dep.Spec.Selector.MatchLabels), int(lo.FromPtr(dep.Spec.Replicas)))
+					}(d)
+				}
+				wg.Wait()
+			}, aws.ProvisioningEventType, multipleDeprovisionersTestGroup, defaultTestName)
+
+			env.Monitor.Reset() // Reset the monitor so that we now track the nodes starting at this point in time
+
+			By("scaling down replicas across deployments")
+			deploymentMap[consolidationValue].Spec.Replicas = lo.ToPtr[int32](int32(int(float64(replicas) * 0.2)))
```
njtran, line 237: Thoughts on just scaling this down by 50%? Should be easier to assert a number of nodes after consolidation with a factor of 2 rather than 5, considering that instance type sizes also change by a factor of 2 as well.

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,631 @@
...
+				provisionerMap[consolidationValue].Spec.Consolidation = &v1alpha5.Consolidation{Enabled: lo.ToPtr(true)}
+				provisionerMap[emptinessValue].Spec.TTLSecondsAfterEmpty = lo.ToPtr[int64](0)
+				provisionerMap[expirationValue].Spec.TTLSecondsUntilExpired = lo.ToPtr[int64](0)
+				provisionerMap[expirationValue].Spec.Limits = disableProvisioningLimits
+				provisionerMap[driftValue].Spec.Limits = disableProvisioningLimits
+				for _, p := range provisionerMap {
+					env.ExpectCreatedOrUpdated(p)
+				}
+				driftNodeTemplate.Spec.AMIFamily = &v1alpha1.AMIFamilyBottlerocket
+				env.ExpectUpdated(driftNodeTemplate)
```
njtran, line 257: Just an observation: it's funny that we'd drift a node given a node template, only to re-provision it as the drifted value for another provisioner

test/pkg/environment/aws/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+				MetricName: aws.String(name),
+				Dimensions: lo.MapToSlice(labels, func(k, v string) *cloudwatch.Dimension {
+					return &cloudwatch.Dimension{
+						Name:  aws.String(k),
+						Value: aws.String(v),
+					}
+				}),
+				Unit:      aws.String(unit),
+				Value:     aws.Float64(value),
+				Timestamp: aws.Time(time.Now()),
```
njtran, line 73: I think these should be pretty similar, but in the case of delays, consider parameterizing this as when the test started/finished

test/pkg/environment/aws/environment.go
```diff
@@ -64,7 +70,16 @@ func NewEnvironment(t *testing.T) *Environment {
...
 		SSMAPI:          ssm.New(session),
 		IAMAPI:          iam.New(session),
+		CloudwatchAPI:   GetCloudWatchAPI(session),
 		InterruptionAPI: itn.New(lo.Must(config.LoadDefaultConfig(env.Context))),
 		SQSProvider:     interruption.NewSQSProvider(sqs.New(session)),
 	}
 }
+
+func GetCloudWatchAPI(session *session.Session) cloudwatchiface.CloudWatchAPI {
+	if lo.Must(env.GetBool("ENABLE_CLOUDWATCH", false)) {
```
njtran, line 80: just inferring, but shouldn't this be the following?
```suggestion
	if lo.Must(env.GetBool("ENABLE_CLOUDWATCH", true)) {
```

test/pkg/environment/aws/setup.go
```diff
@@ -34,23 +31,18 @@ var (
...
 	persistedSettings = env.ExpectSettings()
-	env.Environment.BeforeEach(opts...)
+	env.Environment.BeforeEach()
 }
 
-func (env *Environment) Cleanup(opts ...common.Option) {
-	options := common.ResolveOptions(opts)
-	if !options.DisableDebug {
-		fmt.Println("------- START AWS CLEANUP -------")
-		defer fmt.Println("------- END AWS CLEANUP -------")
```
njtran, line 46: Why remove these? Is it because debug is always enabled now?

test/pkg/environment/common/expectations.go
```diff
@@ -118,16 +148,14 @@ func (env *Environment) ExpectSettingsReplaced(data ...map[string]string) {
 	}
 	// Update the configMap to update the settings
 	env.ExpectCreatedOrUpdated(cm)
-
-	// Get the karpenter pods and delete them to restart the containers
-	env.ExpectKarpenterPodsDeletedWithOffset(1)
-	env.EventuallyExpectKarpenterPodsHealthyWithOffset(1)
```
njtran, line 124: Maybe I'm missing where this is plugged in, but don't we still need to restart the deployment to get the configmap changes?

test/pkg/environment/common/expectations.go
```diff
@@ -159,6 +203,15 @@ func (env *Environment) EventuallyExpectHealthy(pods ...*v1.Pod) {
 	}
 }
 
+func (env *Environment) EventuallyExpectKarpenterRestarted() {
+	env.EventuallyExpectKarpenterRestartedWithOffset(1)
+}
+
+func (env *Environment) EventuallyExpectKarpenterRestartedWithOffset(offset int) {
```
njtran, line 212: We don't need this with GinkgoHelper() right? If you're opting to not migrate it all in this PR, fine with me too :)

test/pkg/environment/common/setup.go
```diff
@@ -56,15 +56,8 @@ var (
 )
 
 // nolint:gocyclo
-func (env *Environment) BeforeEach(opts ...Option) {
-	options := ResolveOptions(opts)
-	if !options.DisableDebug {
-		fmt.Println("------- START BEFORE -------")
-		defer fmt.Println("------- END BEFORE -------")
```
njtran, line 63: Same question about lines here

test/pkg/environment/common/setup.go
```diff
@@ -95,26 +88,14 @@ func (env *Environment) ExpectCleanCluster() {
 	}
 }
 
-func (env *Environment) Cleanup(opts ...Option) {
-	options := ResolveOptions(opts)
-	if !options.DisableDebug {
-		fmt.Println("------- START CLEANUP -------")
-		defer fmt.Println("------- END CLEANUP -------")
```
njtran, line 102: And lastly here

test/pkg/environment/common/setup.go
```diff
@@ -95,26 +88,14 @@ func (env *Environment) ExpectCleanCluster() {
...
 	env.CleanupObjects(CleanableObjects...)
 	env.eventuallyExpectScaleDown()
 	env.ExpectNoCrashes()
 }
 
-func (env *Environment) AfterEach(opts ...Option) {
-	options := ResolveOptions(opts)
-	if !options.DisableDebug {
-		fmt.Println("------- START AFTER -------")
-		defer fmt.Println("------- END AFTER -------")
```
njtran, line 116: Or I guess also here too

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,631 @@
...
+
+				By("waiting for the nodes across all deprovisoiners to get deleted")
+				wg = sync.WaitGroup{}
+				wg.Add(4)
+				go func() {
+					defer GinkgoRecover()
+					defer wg.Done()
+
+					env.MeasureDurationFor(func() {
+						env.EventuallyExpectDeletedNodeCountWithSelector("==", int(float64(nodeCountPerProvisioner)*0.8), labels.SelectorFromSet(map[string]string{
```
njtran, line 267: If you change the factor of .2 to .5, as per my suggestion above make sure to change this to .5 as well.

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,631 @@
...
+				provisionerMap[expirationValue].Spec.TTLSecondsUntilExpired = lo.ToPtr[int64](0)
+				provisionerMap[expirationValue].Spec.Limits = disableProvisioningLimits
+				provisionerMap[driftValue].Spec.Limits = disableProvisioningLimits
+				for _, p := range provisionerMap {
+					env.ExpectCreatedOrUpdated(p)
+				}
+				driftNodeTemplate.Spec.AMIFamily = &v1alpha1.AMIFamilyBottlerocket
+				env.ExpectUpdated(driftNodeTemplate)
+
+				By("waiting for the nodes across all deprovisoiners to get deleted")
```
njtran, line 259:
```suggestion
				By("waiting for the nodes across all deprovisioners to get deleted")
```

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,630 @@
...
+				wg = sync.WaitGroup{}
+				wg.Add(4)
+				go func() {
```
njtran, line 262: Musing: Is it possible to de-dupe these following lines with some function like this? Maybe it's not worth it, but you could also add some sort of expectations map in too. 

`env.ExpectDurationMeasuredForDeprovisioners(provisionerMap, deploymentMap, deprovisioners string...)`

njtran: Or maybe just making the function that you pass into the env.MeasureDurationFor() a defined variable to pass in for each call

userData.txt
```diff
@@ -0,0 +1,4 @@
+<powershell>
```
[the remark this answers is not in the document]
jonathan-innis: Nope!

test/suites/scale/deprovisioning_test.go
jonathan-innis: Yeah, this is changed. I realized that this is wrong.

test/pkg/environment/aws/environment.go
jonathan-innis: No, we want to disable this by default, unless someone explicitly overrides it.

test/pkg/environment/aws/setup.go
jonathan-innis: Yes, WDYT about using `ginkgo.vv` instead of relying on our own mechanism for printing the stages of the test?

test/pkg/environment/common/expectations.go
jonathan-innis: Yep, just re-added it back since I realized that I missed this

test/suites/scale/deprovisioning_test.go
jonathan-innis: Argh, spelling is hard 😛

test/suites/scale/deprovisioning_test.go
jonathan-innis: Yeah, I got that gut feeling too. I'll try to take a look and see if I can make it work

test/suites/scale/suite_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/aws/karpenter/test/pkg/environment/aws"
+)
+
+var env *aws.Environment
+
+func TestScale(t *testing.T) {
+	RegisterFailHandler(Fail)
+	BeforeSuite(func() {
+		env = aws.NewEnvironment(t)
+		SetDefaultEventuallyTimeout(30 * time.Minute)
```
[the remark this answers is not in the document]
jonathan-innis: I'm more convinced that we should bump this timeout when we get there rather than prematurely bumping this timeout when we know that none of our `Eventually` statements take that long

test/suites/scale/provisioning_test.go
jonathan-innis: I think this makes the "math" of how many nodes that we are going to have spun up a bit more complicated. I'm fine with going slightly over 110 for this test-case. The 110 pods is more of a guideline anyways and going slightly over it doesn't seem like it would cause much of a problem.

test/suites/scale/deprovisioning_test.go
jonathan-innis: I'm not convinced that it really matters one way or the other. I think we will eventually do different levels of utilization anyways so starting with 20% seems reasonable to me.

test/pkg/environment/aws/metrics.go
jonathan-innis: I'm not convinced that this really makes a difference here since the metric is generally fired when the event happens.

test/suites/scale/suite_test.go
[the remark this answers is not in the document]
jonathan-innis: Actually, I take it back, I'm realizing that we have a few tests that are cycling 20 nodes with a 2m/node cycle, so I think 1h makes sense with spec-based timeouts for things we expect to be relatively short.

test/suites/scale/provisioning_test.go
jonathan-innis: Do you think it should be one over the other? I think anti-affinity makes sense, since it should be a higher bar for scheduling performance.

test/pkg/environment/aws/setup.go
jonathan-innis: Discussed offline and were fine with removing these lines in favor of `ginkgo.vv`

test/pkg/environment/aws/environment.go
njtran: My bad here, misunderstood what the env.GetBool() second parameter meant.

test/suites/scale/deprovisioning_test.go
jonathan-innis: Created a struct that captures data for each of the provisioners called `testAssertions`. We can do expectations against this instead of having to generate 4 separate functions.

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,630 @@
...
+	consolidationValue    = "consolidation"
+	emptinessValue        = "emptiness"
+	expirationValue       = "expiration"
+	noExpirationValue     = "noExpiration"
+	driftValue            = "drift"
+	noDriftValue          = "noDrift"
+)
+
+const (
+	multipleDeprovisionersTestGroup = "multipleDeprovsioners"
```
njtran, line 55:
```suggestion
	multipleDeprovisionersTestGroup = "multipleDeprovisioners"
```

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,630 @@
...
+		provisioner = test.Provisioner(provisionerOptions)
+		deploymentOptions = test.DeploymentOptions{
+			PodOptions: test.PodOptions{
+				ResourceRequirements: v1.ResourceRequirements{
+					Requests: v1.ResourceList{
+						v1.ResourceCPU:    resource.MustParse("100m"),
+						v1.ResourceMemory: resource.MustParse("100Mi"),
+					},
+				},
+				TerminationGracePeriodSeconds: lo.ToPtr[int64](0),
```
njtran, line 125: Just a thought, I think this is fine for now, but we should really start testing with some non-pause pods and some non-deterministic pod shutdown times. We're taking the assumption that pod graceful shutdown goes well here, which is definitely a place where deprovisioning can be bottlenecked.

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,630 @@
...
+			By("scaling down replicas across deployments")
+			deploymentMap[consolidationValue].Spec.Replicas = lo.ToPtr[int32](int32(int(float64(replicas) * 0.2)))
+			deploymentMap[emptinessValue].Spec.Replicas = lo.ToPtr[int32](0)
+			for _, d := range deploymentMap {
+				env.ExpectUpdated(d)
+			}
+
+			env.MeasureDurationFor(func() {
+				By("enabling deprovisioning across provisioners")
+				// Create provisioners for expiration and drift so that these provisioners can do replacement
```
njtran, line 249:
```suggestion
				// Create a provisioner for expiration so that expiration can do replacement
```

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,630 @@
...
+				provisionerMap[expirationValue].Spec.TTLSecondsUntilExpired = lo.ToPtr[int64](0)
+				provisionerMap[expirationValue].Spec.Limits = disableProvisioningLimits
+				for _, p := range provisionerMap {
+					env.ExpectCreatedOrUpdated(p)
+				}
+				driftNodeTemplate.Spec.AMIFamily = &v1alpha1.AMIFamilyBottlerocket
+				env.ExpectUpdated(driftNodeTemplate)
+
+				By("waiting for the nodes across all deprovisioners to get deleted")
+				type testAssertions struct {
```
njtran, line 264: nice! nit: Consider moving this out to a different file.

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,630 @@
...
+								selector = labels.SelectorFromSet(map[string]string{v1alpha5.ProvisionerNameLabelKey: provisionerMap[d].Name})
+							}
+							env.EventuallyExpectNodeCountWithSelector("==", assertions.nodeCount, selector)
+							env.EventuallyExpectHealthyPodCount(labels.SelectorFromSet(deploymentMap[d].Spec.Selector.MatchLabels), int(lo.FromPtr(deploymentMap[d].Spec.Replicas)))
+						}, aws.DeprovisioningEventType, multipleDeprovisionersTestGroup, defaultTestName, map[string]string{aws.TestSubEventTypeDimension: d})
+					}(k, v)
+				}
+				wg.Wait()
+			}, aws.DeprovisioningEventType, multipleDeprovisionersTestGroup, defaultTestName)
+		}, SpecTimeout(30*time.Minute))
```
njtran, line 318: Forget what we decided here. Did we say 30 minutes was enough? At 200 nodes each I feel like this will take more than half an hour, but less than an hour, but that's just a guess. 

Edit: looking at the later tests, it says 20 nodes takes 45 minutes for consolidation and drift.So this makes me think this needs to be bumped up.

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,630 @@
...
+				env.ExpectUpdated(driftNodeTemplate)
+
+				By("waiting for the nodes across all deprovisioners to get deleted")
+				type testAssertions struct {
+					deletedCount             int
+					deletedNodeCountSelector labels.Selector
+					nodeCount                int
+					nodeCountSelector        labels.Selector
+				}
+				assertionMap := map[string]testAssertions{
```
njtran, line 270: nit: Further de-dupe you can do is to define this assertion map at the top and just re-use it for each test, parameterizing the `env.MeasureDurationFor()` with a `assertions map[string]testAssertions`

test/suites/scale/deprovisioning_test.go
```diff
@@ -0,0 +1,630 @@
...
+
+			env.MeasureDurationFor(func() {
+				By("kicking off deprovisioning expiration by setting the ttlSecondsUntilExpired value on the provisioner")
+				// Change Provisioner limits so that replacement nodes will use another provisioner.
+				provisioner.Spec.Limits = disableProvisioningLimits
+				// Enable Expiration
+				provisioner.Spec.TTLSecondsUntilExpired = lo.ToPtr[int64](0)
+
+				noExpireProvisioner := test.Provisioner(test.ProvisionerOptions{
+					Requirements: []v1.NodeSelectorRequirement{
```
njtran, line 493: Consider just copying the requirements from the base provisioner here.
```suggestion
					Requirements: provisioner.Spec.Requirements
```

njtran (commented):
Looks good! some nits, but one comment on timing for the multiple deprovisioners.

njtran (commented):
Looks good! some nits, but one comment on timing for the multiple deprovisioners.

test/suites/scale/deprovisioning_test.go
jonathan-innis: Agreed. Makes sense. We can do this as we start to expand the world of what we do in scale testing.

test/suites/scale/deprovisioning_test.go
jonathan-innis: Agreed, though (like I mentioned above), I think that this use-case right now is highly contextualized. I can see the point for generalizing this instantiation once we have n > 1 use-cases for it.

test/suites/scale/deprovisioning_test.go
jonathan-innis: Good callout. You're right that we should bump this

njtran (approved):

PR #3890, opened 2023-05-15
test: Cleanup Subnet tests

engedaam:
Fixes # 

**Description**
- Improving the `nodetempate.status` testing for subnets by validating the whole API surface 

* Unite tests

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/nodetemplate/suite_test.go
+++ b/pkg/controllers/nodetemplate/suite_test.go
@@ -103,124 +103,136 @@ var _ = Describe("AWSNodeTemplateController", func() {
 			ExpectApplied(ctx, env.Client, nodeTemplate)
 			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(nodeTemplate))
 			nodeTemplate = ExpectExists(ctx, env.Client, nodeTemplate)
-			subnet, _ := awsEnv.SubnetProvider.List(ctx, nodeTemplate)
-			subnetIDs := lo.Map(subnet, func(ec2subnet *ec2.Subnet, _ int) string {
-				return *ec2subnet.SubnetId
-			})
-			sort.Strings(subnetIDs)
-			subnetIDsInStatus := lo.Map(nodeTemplate.Status.Subnets, func(subnet v1alpha1.Subnet, _ int) string {
-				return subnet.ID
-			})
-			sort.Strings(subnetIDsInStatus)
-			Expect(subnetIDsInStatus).To(Equal(subnetIDs))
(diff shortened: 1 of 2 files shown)


pkg/providers/subnet/suite_test.go
```diff
@@ -116,63 +117,139 @@ var _ = Describe("Subnet Provider", func() {
 		subnets, err := awsEnv.SubnetProvider.List(ctx, nodeTemplate)
 
 		Expect(err).To(BeNil())
-		Expect(subnets).To(HaveLen(1))
-		Expect(aws.StringValue(subnets[0].SubnetId)).To(Equal("subnet-test1"))
-		Expect(aws.StringValue(subnets[0].AvailabilityZone)).To(Equal("test-zone-1a"))
+		Expect(subnets).To(Equal([]*ec2.Subnet{
```
jonathan-innis, line 120: For these tests, I'm less convinced that we need to ensure the entire API surface. I had the opinion that we should ensure the entire API surface for the `nodeTemplate.status` because:
1. We control the API, so we should ensure the API
2. It's a smaller API surface, so it's easier to ensure every field

For these test, I think it might be sufficient to ensure the pieces that we care about and rely on. If we rely on all these fields in other places, then this is fine, we can keep it as you have it. But, if I recall, we really only care about the AZs and the IDs for these tests.

pkg/controllers/nodetemplate/suite_test.go
```diff
@@ -103,124 +103,136 @@ var _ = Describe("AWSNodeTemplateController", func() {
...
-				return int(*subnet[i].AvailableIpAddressCount) > int(*subnet[j].AvailableIpAddressCount)
-			})
-			correctSubnetIDs := lo.Map(subnet, func(ec2subnet *ec2.Subnet, _ int) string {
-				return *ec2subnet.SubnetId
-			})
-			subnetIDsInStatus := lo.Map(nodeTemplate.Status.Subnets, func(subnet v1alpha1.Subnet, _ int) string {
-				return subnet.ID
-			})
-			Expect(subnetIDsInStatus).To(Equal(correctSubnetIDs))
+			Expect(nodeTemplate.Status.Subnets).To(Equal([]v1alpha1.Subnet{
```
jonathan-innis, line 130: Nice 🎉

pkg/providers/subnet/suite_test.go
engedaam: Ah I see, we only really rely on  `SubnetId`, `AvailabilityZone`, and `AvailableIpAddressCount` for choosing the subnet, so I'll just add that pieces.

jonathan-innis (approved):
Nice 🎉

PR #3893, opened 2023-05-16
test: Enable GitHub Actions for E2Es

jonathan-innis:
Fixes # 

**Description**

This PR enables Github Actions to trigger our E2E test runners which enables us to get visibility into our test runners directly from the GH UI. The GHA assumes that you have followed the instructions to deploy the relevant CloudFormation stacks into your account and enabled a `SLACK_WEBHOOK_URL` in the Github action secrets.

This change serves to replace the current Tekton run flow where we trigger and orchestrate our test runners in infrastructure that we manage.

* Running on main in `jonathan-innis/karpenter`

Screenshot 2023-05-26 at 10 02 44 AM

__Note: Upgrade is intentionally failing and should pass when it is pushed to upstream__

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```


jonathan-innis (commented):
/karpenter snapshot

.github/actions/e2e/install-helm/action.yaml
```diff
@@ -0,0 +1,15 @@
+name: InstallHelm
```
ellistarn, line 1: IIUC you could build a tools image (like the one we have) and run the container with gha so you have all your deps packaged up.

.github/workflows/e2e-matrix.yaml
```diff
@@ -0,0 +1,33 @@
+name: E2ETestMatrix
```
ellistarn, line 1: Nit: does not match file name

test/gha/README.md
```diff
@@ -0,0 +1,17 @@
+## Enabling Github Action Runs in Your AWS Account
+
+1. Deploy the [Cloudformation stacks](https://github.com/aws/karpenter/tree/main/test/gha/cloudformation/README.md) into your account that enable Managed Prometheus and the Github Actions runner policies.
+2. Remove the `if: github.repository == 'aws/karpenter'` lines from the top of the `e2e-matrix.yaml` file so that the jobs are allowed to run in other repositories besides Karpenter
```
ellistarn, line 4: Wonder if we could drive this via another mechanism. Maybe the existence of a gh secret?

ellistarn (dismissed):
Very cool!!!

test/gha/README.md
jonathan-innis: We _could_ just not gate it. This will outright fail if it tries to assume the role outlined in the action. I'm fine either way.

.github/actions/e2e/install-helm/action.yaml
jonathan-innis: Yep, you're right. Was planning to take care of it in a follow-up. Wanted to get the initial iteration out and then follow-up in another PR to add the image.

test/gha/README.md
jonathan-innis: I'm going to do a little look-around and see if there's a best practice on how to handle actions like this for forks.

.github/actions/e2e/install-eksctl/action.yaml
```diff
@@ -0,0 +1,16 @@
+name: InstallEKSCTL
```
njtran, line 1: I'm worried about how many different ways this opens us up to failures. 

Looking briefly for some ways to use ECR images, I saw this https://agileek.github.io/software/aws/2020/10/28/using-an-ecr-image-in-github-actions/#:~:text=The%20complete%20GitHub%20workflow%3A but I don't know how viable it is.

njtran: I feel the effort is worth it to just use the tools image we already have, rather than creating some other tool installation containers. WDYT?

.github/actions/e2e/install-karpenter/action.yaml
```diff
@@ -0,0 +1,64 @@
...
+    with:
+      repository: ${{ inputs.git_repo }}
+      ref: ${{ inputs.git_ref }}
+  - uses: actions/checkout@v3
+    with:
+      path: "actions"
+  - uses: ./actions/.github/actions/e2e/install-helm
+  - name: install-karpenter
+    shell: bash
+    # TODO @joinnis: Change the --version argument back to --version v0-$(git rev-parse HEAD)
```
njtran, line 40: Seems like this TODO is done here, but not in the next step

.github/actions/e2e/install-prometheus/action.yaml
```diff
@@ -0,0 +1,43 @@
...
+      run: |
+        kubectl create ns prometheus
+        kubectl label ns prometheus scrape=enabled
+    - name: install prometheus
+      shell: bash
+      run: |
+        helm upgrade --install prometheus prometheus-community/kube-prometheus-stack \
+          -n prometheus --create-namespace \
+          -f ./actions/.github/actions/e2e/install-prometheus/values.yaml \
+          --set prometheus.prometheusSpec.remoteWrite[0].url=https://aps-workspaces.${{ inputs.region }}.amazonaws.com/workspaces/${{ inputs.workspace_id }}/api/v1/remote_write \
```
njtran, line 41: Is this URL going to be for a managed Prometheus?

.github/workflows/e2e-matrix.yaml
```diff
@@ -0,0 +1,33 @@
...
+      git_repo: jonathan-innis/karpenter
+      event_name: ${{ github.event_name }}
+    secrets:
+      SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+  e2e-upgrade:
+    if: github.repository == 'aws/karpenter'
+    uses: ./.github/workflows/e2e-upgrade.yaml
+    with:
+      from_git_ref: v0.26.1
+      git_repo: jonathan-innis/karpenter
```
njtran, line 30: Probably should change this

.github/workflows/e2e-matrix.yaml
```diff
@@ -0,0 +1,33 @@
...
+  e2e:
+    if: github.repository == 'aws/karpenter'
+    strategy:
+      fail-fast: false
+      matrix:
+        suite: [Integration, Machine, Consolidation, Utilization, Interruption, Drift, Chaos, IPv6]
+    uses: ./.github/workflows/e2e.yaml
+    with:
+      suite: ${{ matrix.suite }}
+      git_repo: jonathan-innis/karpenter
```
njtran, line 21: Should probably change this too

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,101 @@
...
+        required: true
+      event_name:
+        type: string
+        required: true
+    secrets:
+      SLACK_WEBHOOK_URL:
+        required: true
+env:
+  AWS_REGION: ${{ inputs.region }}
+  K8s_VERSION: 1.25
```
njtran, line 24: Would be nice to parameterize this

.github/workflows/e2e-matrix.yaml
jonathan-innis: Good catch

.github/workflows/e2e.yaml
jonathan-innis: There isn't a good way to pass these down from their top-level caller. i.e. even if we passed these down as inputs from the top-level workflow caller, we couldn't define these values only once. I tried it and you can't specify environment variables declared once in a `with:` statement like

```yaml
  e2e-upgrade:
    if: github.repository == 'aws/karpenter'
    uses: ./.github/workflows/e2e-upgrade.yaml
    with:
      region: ${{ env.AWS_REGION }} # This fails with an error
      from_git_ref: v0.26.1
      git_repo: jonathan-innis/karpenter
```

.github/actions/e2e/install-prometheus/action.yaml
jonathan-innis: Yep, that's the workspace ID that's passed through.

.github/actions/e2e/install-eksctl/action.yaml
jonathan-innis: > I'm worried about how many different ways this opens us up to failures

I'm curious what you're thinking that these failures could be. This is the prescribed download method for `eksctl`

The tools image that we have doesn't currently work out-of-the-box and needs some tweaking to get it working properly with GHA. I can take that as an AI and fast-follow but I'm not convinced that this needs to be blocking for this PR to go in.

.github/workflows/e2e.yaml
jonathan-innis: I also would say that `input` values for wokrflows should be things that are likely to change while environment variables are going to be consistent and generally the same throughout a given environment. I don't see these values as likely to change as often so I think they are fine as environment variables.

.github/actions/e2e/install-eksctl/action.yaml
njtran: > I'm curious what you're thinking that these failures could be.

Really just thinking about unexpected bash issues. I'm fine to merge this and do a fast follow, but don't want us to forget about this.

jonathan-innis: Sounds good, not planning to forget about this. Part of the follow-up for this is going to be creating the testing image for GHA to run the tests with

test/gha/cloudformation/iam_cloudformation.yaml
```diff
@@ -0,0 +1,475 @@
...
+              - sqs:GetQueueAttributes
+              - sqs:GetQueueUrl
+              - sqs:SendMessage
+              - sqs:ReceiveMessage
+            Resource: "*"
+          - Effect: Allow
+            Action: iam:PassRole
+            Resource:
+              - !Sub "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/KarpenterNodeRole-*"
+              - !Sub "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/aws-fis-itn"
```
njtran, line 409: FISInterruptionRole

test/gha/cloudformation/iam_cloudformation.yaml
```diff
@@ -0,0 +1,475 @@
...
+        Statement:
+          - Effect: Allow
+            Principal:
+              Federated: !Ref GithubOIDCProvider
+            Action: "sts:AssumeRoleWithWebIdentity"
+            Condition:
+              StringEquals:
+                token.actions.githubusercontent.com:aud: sts.amazonaws.com
+              StringLike:
+                token.actions.githubusercontent.com:sub: !Sub "repo:${Repository}:ref:refs/heads/main"
```
njtran, line 451: synced offline, need to validate after this PR is merged to check that this can work for snapshots before merge

.github/workflows/e2e-matrix.yaml
```diff
@@ -0,0 +1,33 @@
...
+      suite: ${{ matrix.suite }}
+      git_repo: aws/karpenter
+      event_name: ${{ github.event_name }}
+    secrets:
+      SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+  e2e-upgrade:
+    if: github.repository == 'aws/karpenter'
+    uses: ./.github/workflows/e2e-upgrade.yaml
+    with:
+      from_git_ref: v0.26.1
```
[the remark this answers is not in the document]
jonathan-innis: Added

hack/release/stable-pr.sh
```diff
@@ -21,9 +21,11 @@ updateKarpenterCoreGoMod(){
 updateTektonPreUpgradeVersion(){
   LAST_MINOR_VERSION=$(git tag --sort=committerdate | grep -v "v${RELEASE_VERSION_MAJOR}.${RELEASE_VERSION_MINOR}" | tail -1)
   TEKTON_RELEASE_LISTENER_PATH="tools/release-notification-listener/listener/tekton.go"
+  GITHUB_ACTION_E2E_MATRIX_PATH=".github/workflows/e2e-matrix.yaml"
 
   # This command goes into the tekton release-listener file and replaces the preUpgradeVersion with the last minor version
   sed -i "s/preUpgradeVersion = \".*\"/preUpgradeVersion = \"$LAST_MINOR_VERSION\"/" "$TEKTON_RELEASE_LISTENER_PATH"
+  sed -i "s/from_git_ref:.*/from_git_ref: $LAST_MINOR_VERSION/" "$GITHUB_ACTION_E2E_MATRIX_PATH"
```
njtran, line 28: Wow nice! Thanks for doing this :D

njtran (approved):

PR #3897, opened 2023-05-16
fix: Update E2E old drift AMI

engedaam:
Fixes # 

**Description**

- The newer AMI contains some security updates that would be good to include for our testing

* Manually tested

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/test/suites/drift/suite_test.go
+++ b/test/suites/drift/suite_test.go
@@ -66,7 +66,7 @@ var _ = Describe("Drift", Label("AWS"), func() {
 		})
 		// choose an old static image
 		parameter, err := env.SSMAPI.GetParameter(&ssm.GetParameterInput{
-			Name: awssdk.String("/aws/service/eks/optimized-ami/1.23/amazon-linux-2/amazon-eks-node-1.23-v20221101/image_id"),
+			Name: awssdk.String("/aws/service/eks/optimized-ami/1.23/amazon-linux-2/amazon-eks-node-1.23-v20230322/image_id"),
 		})
 		Expect(err).To(BeNil())
 		oldCustomAMI := *parameter.Parameter.Value
@@ -112,7 +112,7 @@ var _ = Describe("Drift", Label("AWS"), func() {
 		})
 		// choose an old static image
 		parameter, err := env.SSMAPI.GetParameter(&ssm.GetParameterInput{
-			Name: awssdk.String("/aws/service/eks/optimized-ami/1.23/amazon-linux-2/amazon-eks-node-1.23-v20221101/image_id"),
+			Name: awssdk.String("/aws/service/eks/optimized-ami/1.23/amazon-linux-2/amazon-eks-node-1.23-v20230322/image_id"),
 		})
 		Expect(err).To(BeNil())
 		oldCustomAMI := *parameter.Parameter.Value
--- a/test/suites/integration/ami_test.go
+++ b/test/suites/integration/ami_test.go
@@ -65,7 +65,7 @@ var _ = Describe("AMI", func() {
 	It("should use the most recent AMI when discovering multiple", func() {
 		// choose an old static image
 		parameter, err := env.SSMAPI.GetParameter(&ssm.GetParameterInput{
-			Name: aws.String("/aws/service/eks/optimized-ami/1.23/amazon-linux-2/amazon-eks-node-1.23-v20221101/image_id"),
+			Name: aws.String("/aws/service/eks/optimized-ami/1.23/amazon-linux-2/amazon-eks-node-1.23-v20230322/image_id"),
 		})
 		Expect(err).To(BeNil())
 		oldCustomAMI := *parameter.Parameter.Value


njtran (approved):
Approving, but musing: maybe we can programmatically pick the second latest SSM AMI release, and pick that so that we don't have to come back and update this file later. 

I.e. this command returns all the SSM aliases, that we could sort
```
aws ssm get-parameters-by-path --path /aws/service/eks/optimized-ami/1.23/amazon-linux-2 --recursive --max-items=50
```

jonathan-innis (approved):

test/suites/drift/suite_test.go
```diff
@@ -66,7 +66,7 @@ var _ = Describe("Drift", Label("AWS"), func() {
 		})
 		// choose an old static image
 		parameter, err := env.SSMAPI.GetParameter(&ssm.GetParameterInput{
-			Name: awssdk.String("/aws/service/eks/optimized-ami/1.23/amazon-linux-2/amazon-eks-node-1.23-v20221101/image_id"),
+			Name: awssdk.String("/aws/service/eks/optimized-ami/1.23/amazon-linux-2/amazon-eks-node-1.23-v20230322/image_id"),
```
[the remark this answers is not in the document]
engedaam: Updated the PR

test/suites/integration/ami_test.go
```diff
@@ -64,8 +64,13 @@ var _ = Describe("AMI", func() {
 	})
 	It("should use the most recent AMI when discovering multiple", func() {
 		// choose an old static image
+		parameters, err := env.SSMAPI.GetParametersByPath(&ssm.GetParametersByPathInput{
```
jonathan-innis, line 67: We should also expect a certain length from these responses. Else, we might get a panic when we try to index if the length is too short

test/suites/drift/suite_test.go
```diff
@@ -65,8 +65,13 @@ var _ = Describe("Drift", Label("AWS"), func() {
 			"featureGates.driftEnabled": "true",
 		})
 		// choose an old static image
+		parameters, err := env.SSMAPI.GetParametersByPath(&ssm.GetParametersByPathInput{
+			Path:      awssdk.String("/aws/service/eks/optimized-ami/1.23/amazon-linux-2"),
```
njtran, line 69: Can you also parameterize the kubernetes version to use the one that's discovered by the environment?

test/suites/drift/suite_test.go
```diff
@@ -65,8 +65,13 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+			Recursive: awssdk.Bool(true),
+		})
+		Expect(err).To(BeNil())
 		parameter, err := env.SSMAPI.GetParameter(&ssm.GetParameterInput{
-			Name: awssdk.String("/aws/service/eks/optimized-ami/1.23/amazon-linux-2/amazon-eks-node-1.23-v20221101/image_id"),
+			Name: awssdk.String(fmt.Sprintf("%s/image_id", *parameters.Parameters[len(parameters.Parameters)-2].Name)),
```
njtran, line 74: Is this output guaranteed to be sorted?

test/suites/drift/suite_test.go
njtran: Additionally, you could consider putting this in a function to get the latest-nth version with a `n int` parameter.

test/suites/drift/suite_test.go
engedaam: It has a specific hierarchy, according the documentation, but I will sort just in case

test/suites/drift/suite_test.go
```diff
@@ -66,7 +66,7 @@ var _ = Describe("Drift", Label("AWS"), func() {
 		})
```
jonathan-innis, line 66: I think we should replace the `selectCustomAMI` function in this file with the `GetCustomAMI` function in the `aws/expectations.go` file so that we aren't duplicating the code throughout.

test/pkg/environment/aws/expectations.go
```diff
@@ -214,3 +215,18 @@ func (env *Environment) GetCustomAMI(amiPath string, versionOffset int) string {
 	Expect(err).To(BeNil())
 	return *parameter.Parameter.Value
 }
+
+func (env *Environment) GetNthLastAMIVersion(k8sVersion string, n int) string {
```
jonathan-innis, line 219: Is it possible to extend this off of the `GetCustomAMI` function rather than creating an entirely different function or to create a function similar to it so that it makes sense that the functions are related?

test/suites/drift/suite_test.go
```diff
@@ -66,7 +66,7 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
 		// choose an old static image
 		parameter, err := env.SSMAPI.GetParameter(&ssm.GetParameterInput{
```
jonathan-innis, line 68: musing: Can we create a function that completely encapsulates the behavior of getting an older AMI version and gives the id back rather than leaking the SSM parameter details into the test context?

engedaam:
The `ssm get-parameters-by-path` is not a viable solution for automating for picking an AMI that is not the latest AMI, as `get-parameters-by-path` does not respond with a sorted list of AMIs. To unblock for now, just gonna update the SSM path and make a follow up PR to automate the AMI selection.

test/suites/drift/suite_test.go
engedaam: Will implement in a follow up PR

test/pkg/environment/aws/expectations.go
engedaam: Will implement in a follow up PR

test/suites/drift/suite_test.go
engedaam: Will implement in a follow up PR

jonathan-innis (approved):
LGTM 🚀

PR #3904, opened 2023-05-17
chore: Move AMI drift to a drift.go

engedaam:
Fixes # 

**Description**
- Created a file to place all the `AWSNodeTemplate` components for drift

* Manually tested

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


njtran (approved):

pkg/cloudprovider/drift.go
```diff
@@ -0,0 +1,75 @@
...
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
+	"github.com/aws/karpenter-core/pkg/cloudprovider"
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+	"github.com/aws/karpenter/pkg/providers/amifamily"
+	"github.com/aws/karpenter/pkg/utils"
+)
+
+func (c *CloudProvider) isNodeFromDriftedNodeTemplate(ctx context.Context, machine *v1alpha5.Machine, provisioner *v1alpha5.Provisioner, nodeTemplate *v1alpha1.AWSNodeTemplate) (bool, error) {
```
jonathan-innis, line 31: nit: This naming doesn't make a _ton_ of sense to me. Why not just, `isNodeTemplateDrifted`? I assume you are doing all of this breaking up so that you can do an `isDrifted` call on each aspect of the spec?

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (approved):
One nit, but otherwise LGTM 🚀  

Let's wait for the E2Es to run before merging

pkg/cloudprovider/drift.go
engedaam: I was thinking that would be confusing because that seems like we are checking if an AWSNodeTemplate is drifted, as opposed to checking if a machine is drifted against a NodeTemplate. Would `isMachineFromDriftedNodeTemplate` be better?

engedaam: And yes all the calls  from the NodeTemplate spec will be placed in that function

jonathan-innis (approved):
Nice! 🎉

PR #3926, opened 2023-05-22
test: Fix failures in Scale Testing Suite

jonathan-innis:
Fixes # 

**Description**

1. Updates the `SpecTimeouts` for each of the shorter tests since the current `SpecTimeouts` were too constrained
2. Updates the `--grace-period` for Ginkgo so that clean-ups on `SpecTimeouts` have enough time to clean-up after a failure
3. Updates the E2E test suite manifests to use a different instance type when running scale testing

* Tekton runs of Scale Testing

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/test/manifests/setup.yaml
+++ b/test/manifests/setup.yaml
@@ -32,7 +32,6 @@ spec:
       TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")
       curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/identity-credentials/ec2/info | jq -r ".AccountId" | tr -d '\n' > $(results.account-id.path)
       curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/placement/region > $(results.region.path)
-
   - name: checkout-repository
     image: alpine/git
     workingDir: $(workspaces.ws.path)
@@ -59,7 +58,7 @@ spec:
   - name: create-cluster
     image: public.ecr.aws/karpenter/tools:latest
     script: |
-      cmd="create"
+      cmd="create"  
       eksctl get cluster --name $(params.cluster-name) && cmd="upgrade"
       eksctl ${cmd} cluster -f - <<EOF
       ---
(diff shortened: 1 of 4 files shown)


jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

test/manifests/setup.yaml
```diff
@@ -60,6 +61,13 @@ spec:
     image: public.ecr.aws/karpenter/tools:latest
     script: |
       cmd="create"
+      
+      # Use a bigger instance type when running scale tests so we can have more cpu/memory
+      instanceType="c5.large"
+      if [[ "$(params.test-filter)" == "Scale" ]]; then
+        instanceType="c5.2xlarge"
+      fi
```
njtran, line 69: I feel like it's fine just to remove this and just use a `.2xlarge` for all tests, no need to complicate the logic

test/manifests/setup.yaml
```diff
@@ -114,6 +122,14 @@ spec:
...
     workingDir: $(workspaces.ws.path)
     script: |
+      # Use more cpu/memory when running scale tests
+      cpuCount=1
+      memoryCount="1Gi"
+      if [[ "$(params.test-filter)" == "Scale" ]]; then
+        cpuCount=4
+        memoryCount="4Gi"
+      fi
+
```
njtran, line 132: Is this value tuned for 200 node scale tests? Thoughts on fully parameterizing this? Since we'll probably be testing many different dimensions, probably helps to have this knob in too.

test/manifests/setup.yaml
jonathan-innis: Sounds good. This is a short-term change anyways since removing Tekton is probably down-the-pipe.

test/manifests/setup.yaml
jonathan-innis: That seems like a change that we can make down-the-line. I think using these values as an upper-bound is fine for now

njtran (approved):

PR #3942, opened 2023-05-25
test: fix consolidation test timing and add commit ref in metrics

njtran:
Fixes # 

**Description**
- adds `git_ref` as a dimension into exported cloudwatch metrics
- fixes a timing bug with consolidation delete scale tests
- adds consolidation delete empty scale test

* `TEST_SUITE=scale ENABLE_CLOUDWATCH=true make e2etests`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


Makefile
```diff
@@ -79,7 +80,7 @@ battletest: ## Run randomized, racing, code-covered tests
 		-tags random_test_delay
 
 e2etests: ## Run the e2e suite against your local cluster
-	cd test && CLUSTER_NAME=${CLUSTER_NAME} go test \
+	cd test && GIT_REF=${GIT_REF} CLUSTER_NAME=${CLUSTER_NAME} go test \
```
njtran, line 83: I guess this could be wrong if someone isn't running in the checked out directory as their applied image

test/pkg/environment/aws/metrics.go
```diff
@@ -50,17 +50,19 @@ const (
 	TestSubEventTypeDimension   = "subEventType"
 	TestGroupDimension          = "group"
 	TestNameDimension           = "name"
+	CommitHashDimension         = "commitHash"
 )
 
 // MeasureDurationFor observes the duration between the beginning of the function f() and the end of the function f()
-func (env *Environment) MeasureDurationFor(f func(), eventType EventType, group, name string, additionalLabels ...map[string]string) {
+func (env *Environment) MeasureDurationFor(f func(), eventType EventType, group, name, commit string, additionalLabels ...map[string]string) {
```
jonathan-innis, line 57: Rather than passing it through to this function, can this just be a global that the test writer doesn't have to reason about? As-in, you just get this for free when you run the suite and use that function? You could event put it into `ExpectEventDurationMetric` function directly so it's completely abstracted away.

Makefile
jonathan-innis: Yeah, I think the main use-case here is to pass this through during automation. I'm not sure if it's necessarily required to have a good local dev story for these metrics beyond being able to test that they are firing reasonable values.

test/suites/scale/deprovisioning_test.go
```diff
@@ -438,8 +479,7 @@ var _ = Describe("Deprovisioning", Label(debug.NoWatch), func() {
 			env.Monitor.Reset() // Reset the monitor so that we now track the nodes starting at this point in time
 
 			By("waiting for all deployment pods to be deleted")
-			// Fully scale down all pods to make nodes empty
-			deployment.Spec.Replicas = lo.ToPtr[int32](0)
+			// Delete deployment to make nodes empty
```
jonathan-innis, line 482: Nice 🎉

test/pkg/environment/aws/metrics.go
```diff
@@ -66,7 +69,8 @@ func (env *Environment) MeasureDurationFor(f func(), eventType EventType, group,
 
 func (env *Environment) ExpectEventDurationMetric(d time.Duration, labels map[string]string) {
 	GinkgoHelper()
-	env.ExpectMetric("eventDuration", cloudwatch.StandardUnitSeconds, d.Seconds(), labels)
+	gitRef := lo.Ternary(env.Context.Value(common.GitRefContextKey) != nil, env.Value(common.GitRefContextKey).(string), "n/a")
```
jonathan-innis, line 72:
```suggestion
	gitRef := lo.Ternary(env.Context.Value(common.GitRefContextKey) != nil, env.Value(common.GitRefContextKey).(string), "")
```

I think we generally have been taking empty string to be the `null` value for metrics. WDYT about making that consistent throughout?

njtran: IIUC, this should fail, because it can't be an empty string: 
https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_Dimension.html#:~:text=Length%20Constraints%3A%20Minimum%20length%20of%201.%20Maximum%20length%20of%20255.

jonathan-innis (approved):
LGTM 🚀

PR #3943, opened 2023-05-25
fix: Ensure that our cluster Machine list is more updated than our CloudProvider Machine list

jonathan-innis:
Fixes # 

**Description**

Prior to this change, it was possible for a call to `ec2.DescribeInstances` to take a long time. Machines were listed prior to this API call, which meant that if this `DescribeInstances` call took a long time, that the instances returned from the EC2 API would be more updated than the Machines that were retrieved from the List. To solve this, we should reverse the order of execution. This way, if the `DescribeInstances` takes longer, we will still get the Machines at the end of the call.

With this new order of execution, even if the Machine LIST takes a lot longer and the EC2 data is old, Machines should be a superset of the instances data, meaning that we will only ever delete _less_ instances than we should, not _more_.

* `make presubmit`
* `/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/pkg/controllers/machine/garbagecollection/controller.go
+++ b/pkg/controllers/machine/garbagecollection/controller.go
@@ -58,6 +58,16 @@ func (c *Controller) Name() string {
 }
 
 func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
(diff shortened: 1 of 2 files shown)


jonathan-innis (commented):
/karpenter snapshot

njtran (approved):

PR #3948, opened 2023-05-26
chore: Scope CloudFormation Permissions

jonathan-innis:
Fixes # 

**Description**

Updates CloudFormation permissions in the "Getting Started Guide" to be scoped to sufficiently restrict the tags that can be used when provisioning nodes with Karpenter. It also enforce tag-based authz on the karpenter controller by default.

Note that the CloudFormation file and the permissions that are provided there-in only serve as a starting point for users. The permissions will most likely have to be extended to some degree for most use-cases.

* `/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -109,10 +109,6 @@ func (p *Provider) Link(ctx context.Context, id, provisionerName string) error {
 				Value: aws.String(settings.FromContext(ctx).ClusterName),
 			},
 			{
-				Key:   aws.String(fmt.Sprintf("kubernetes.io/cluster/%s", settings.FromContext(ctx).ClusterName)),
-				Value: aws.String("owned"),
-			},
-			{
 				Key:   aws.String(v1alpha5.ProvisionerNameLabelKey),
 				Value: aws.String(provisionerName),
 			},
--- a/website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
+++ b/website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
@@ -108,9 +108,12 @@ Resources:
               "Effect": "Allow",
               "Resource": "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:launch-template/*",
               "Action": "ec2:DeleteLaunchTemplate",
-              "Condition": {
+               "Condition": {
                 "StringEquals": {
-                  "aws:ResourceTag/karpenter.k8s.aws/cluster": "${ClusterName}"
+                  "aws:ResourceTag/kubernetes.io/cluster/${ClusterName}": "owned"
+                },
+                "StringLike": {
+                  "aws:ResourceTag/karpenter.sh/provisioner-name": "*"
                 }
               }
             },
--- a/website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
+++ b/website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
@@ -35,7 +35,7 @@ Resources:
     Type: AWS::IAM::ManagedPolicy
     Properties:
       ManagedPolicyName: !Sub "KarpenterControllerPolicy-${ClusterName}"
-      # The PolicyDocument must be in JSON string format because we use a StringEquals condition that uses a interpolated
+      # The PolicyDocument must be in JSON string format because we use a StringEquals condition that uses an interpolated
       # value in one of its key parameters which isn't natively supported by CloudFormation
       PolicyDocument: !Sub |
         {
@@ -111,7 +111,7 @@ Resources:
               }
             },
             {
-              "Sid": "AllowScopedResourceTagging",
+              "Sid": "AllowScopedResourceCreationTagging",
               "Effect": "Allow",
               "Resource": [
                 "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:fleet/*",
@@ -196,25 +196,17 @@ Resources:
               "Action": "ssm:GetParameter"
             },
             {
-              "Sid": "AllowScopedInstanceTermination",
+              "Sid": "AllowScopedDeletion",
               "Effect": "Allow",
-              "Resource": "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:instance/*",
-              "Action": "ec2:TerminateInstances",
+              "Resource": [
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:instance/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:launch-template/*"
+              ],
+              "Action": [
+                "ec2:TerminateInstances",
+                "ec2:DeleteLaunchTemplate"
+              ],
               "Condition": {
-                "StringEquals": {
-                  "aws:ResourceTag/kubernetes.io/cluster/${ClusterName}": "owned"
-                },
-                "StringLike": {
-                  "aws:ResourceTag/karpenter.sh/provisioner-name": "*"
-                }
-              }
-            },
-            {
-              "Sid": "AllowScopedLaunchTemplateDeletion",
-              "Effect": "Allow",
-              "Resource": "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:launch-template/*",
-              "Action": "ec2:DeleteLaunchTemplate",
-               "Condition": {
                 "StringEquals": {
                   "aws:ResourceTag/kubernetes.io/cluster/${ClusterName}": "owned"
                 },
--- a/website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
+++ b/website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
@@ -11,7 +11,7 @@ Resources:
       InstanceProfileName: !Sub "KarpenterNodeInstanceProfile-${ClusterName}"
       Path: "/"
       Roles:
-        - Ref: "KarpenterNodeRole"
+        - !Ref "KarpenterNodeRole"
   KarpenterNodeRole:
     Type: "AWS::IAM::Role"
     Properties:
@@ -39,16 +39,24 @@ Resources:
         Version: "2012-10-17"
         Statement:
           - Effect: Allow
-            Resource: "*"
+            Resource:
+              - !Sub "arn:${AWS::Partition}:ec2:${AWS::Region}::image/*"
+              - !Sub "arn:${AWS::Partition}:ec2:${AWS::Region}::snapshot/*"
+              - !Sub "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:fleet/*"
+              - !Sub "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:instance/*"
+              - !Sub "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:spot-instances-request/*"
+              - !Sub "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:security-group/*"
+              - !Sub "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:volume/*"
+              - !Sub "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:network-interface/*"
+              - !Sub "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:subnet/*"
+              - !Sub "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:launch-template/*"
             Action:
-              # Write Operations
               - ec2:CreateFleet
               - ec2:CreateLaunchTemplate
               - ec2:CreateTags
-              - ec2:DeleteLaunchTemplate
-              - ec2:RunInstances
-              - ec2:TerminateInstances
-              # Read Operations
+          - Effect: Allow
+            Resource: "*"
+            Action:
               - ec2:DescribeAvailabilityZones
               - ec2:DescribeImages
               - ec2:DescribeInstances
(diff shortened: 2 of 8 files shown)


jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

FernandoMiguel:
I wonder if it is too much to ask that once you finish the cloudformation policy updates, if you wouldnt mind also updating the terraform ones?
https://github.com/terraform-aws-modules/terraform-aws-iam/blob/9cda42890147ee5c29c2c8eea3352e7f8d7446b9/modules/iam-role-for-service-accounts-eks/policies.tf#L586

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/f741db1b2c7c4ef0409d7c2e4f588bbc639cbaf2/modules/karpenter/main.tf#L70

https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/80bd3b92e0f82f45a187805d6d847db320a01c31/modules/kubernetes-addons/karpenter/data.tf#L3

cc @bryantbiggs

jonathan-innis:
> if you wouldnt mind also updating the terraform ones

@bryantbiggs There may be some room to align here. There are some cluster-scoped tags that Karpenter already uses by default that might be good to use as part of the default policy rather than asking the user to pass through an IRSA tag that they would also have to pass through to their `AWSNodeTemplate`.

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
```diff
@@ -58,8 +66,39 @@ Resources:
...
+            Action: pricing:GetProducts
+          - Effect: Allow
+            Resource: !Sub "arn:${AWS::Partition}:ssm:${AWS::Region}::parameter/aws/service/*"
+            Action: ssm:GetParameter
+          - Effect: Allow
+            Resource: !Sub "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:instance/*"
+            Action: ec2:TerminateInstances
+            Condition:
+              StringEquals:
+                "aws:ResourceTag/karpenter.sh/managed-by": !Sub "${ClusterName}"
```
bwagner5, line 83: this won't work if someone updates infra before moving to Machines right?

jonathan-innis (commented):
/karpenter snapshot

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
jonathan-innis: Yep, you're right. Discussed offline. Since CloudFormation doesn't natively allow you to do string interpolation with Map keys in YAML, we've transformed the `PolicyDocument` into a JSON string where we can do the interpolation. This allow us to do a StringEquals and StringLike condition like

```yaml
StringEquals:
  "aws:ResourceTag/kubernetes.io/cluster/${ClusterName}": "owned"
StringLike
  "aws:ResourceTag/karpenter.sh/provisioner-name": "*"
```

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
```diff
@@ -35,48 +35,125 @@ Resources:
...
+      # The PolicyDocument must be in JSON string format because we use a StringEquals condition that uses a interpolated
+      # value in one of its key parameters which isn't natively supported by CloudFormation
+      PolicyDocument: !Sub |
+        {
+          "Version": "2012-10-17",
+          "Statement": [
+            {
+              "Effect": "Allow",
+              "Resource": [
+                "arn:${AWS::Partition}:ec2:${AWS::Region}::image/*",
```
jonathan-innis, line 47: One question: Is it possible for users to do cross-region images for instances? I think the answer is no given that our `DescribeImages` call region-scoped.

bwagner5: no images are regional

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
```diff
@@ -35,48 +35,125 @@ Resources:
...
+                "sqs:DeleteMessage",
+                "sqs:GetQueueAttributes",
+                "sqs:GetQueueUrl",
+                "sqs:ReceiveMessage"
+              ]
+            },
+            {
+              "Effect": "Allow",
+              "Resource": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/KarpenterNodeRole-${ClusterName}",
+              "Action": "iam:PassRole"
```
bwagner5, line 148: need an iam:PassedToService condition for EC2 as well: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_iam-passrole-service.html

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
jonathan-innis: Sorry, is this "no, images are regional" or "no images are regional"?

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
```diff
@@ -35,48 +35,125 @@ Resources:
...
+                "arn:${AWS::Partition}:ec2:${AWS::Region}::snapshot/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:instance/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:spot-instances-request/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:security-group/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:volume/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:network-interface/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:subnet/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:launch-template/*"
+              ],
+              "Action": "ec2:RunInstances"
```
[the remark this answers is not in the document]
jonathan-innis: I'm curious how this is tangibly different than specifying that the resource must be in the region like in the above.

pkg/providers/instance/instance.go
```diff
@@ -108,10 +108,6 @@ func (p *Provider) Link(ctx context.Context, id, provisionerName string) error {
 				Key:   aws.String(v1alpha5.ManagedByLabelKey),
 				Value: aws.String(settings.FromContext(ctx).ClusterName),
 			},
-			{
-				Key:   aws.String(fmt.Sprintf("kubernetes.io/cluster/%s", settings.FromContext(ctx).ClusterName)),
-				Value: aws.String("owned"),
-			},
```
micahhausler, line 114: Why remove the cluster tag?

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
```diff
@@ -35,48 +35,224 @@ Resources:
...
-            Resource: !GetAtt KarpenterInterruptionQueue.Arn
-          - Effect: Allow
-            Action:
-              - iam:PassRole
-            Resource: !Sub "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/KarpenterNodeRole-${ClusterName}"
-          - Effect: Allow
-            Action:
-              - eks:DescribeCluster
-            Resource: !Sub "arn:${AWS::Partition}:eks:${AWS::Region}:${AWS::AccountId}:cluster/${ClusterName}"
+      # The PolicyDocument must be in JSON string format because we use a StringEquals condition that uses a interpolated
```
micahhausler, line 38:
```suggestion
      # The PolicyDocument must be in JSON string format because we use a StringEquals condition that uses an interpolated
```

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
micahhausler: ImageIDs only exist within in a region. You cannot use a cross-region image

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
```diff
@@ -35,48 +35,224 @@ Resources:
...
+                    "karpenter.sh/managed-by",
+                    "karpenter.sh/provisioner-name",
+                    "kubernetes.io/cluster/${ClusterName}",
+                    "karpenter.k8s.aws/cluster"
+                  ]
+                }
+              }
+            },
+            {
+              "Sid": "AllowScopedEC2InstanceActionsWithTags",
```
micahhausler, line 84: With the above `AllowScopedEC2InstanceActions` Statement, isn't this one unnecessary?

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
```diff
@@ -35,48 +35,224 @@ Resources:
...
+              "Action": "ec2:DeleteLaunchTemplate",
+               "Condition": {
+                "StringEquals": {
+                  "aws:ResourceTag/kubernetes.io/cluster/${ClusterName}": "owned"
+                },
+                "StringLike": {
+                  "aws:ResourceTag/karpenter.sh/provisioner-name": "*"
+                }
+              }
+            },
```
micahhausler, line 225: Are we running into a policy/template size problem? Can these two statements be combined?

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
```diff
@@ -35,48 +35,224 @@ Resources:
...
+          "Statement": [
+            {
+              "Sid": "AllowScopedEC2InstanceActions",
+              "Effect": "Allow",
+              "Resource": [
+                "arn:${AWS::Partition}:ec2:${AWS::Region}::image/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}::snapshot/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:spot-instances-request/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:security-group/*",
+                "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:subnet/*",
```
micahhausler, line 52: This prevents cross-account subnets using RAM (resource access manager), and should either use `*` or just empty `::` instead of an account ID. (I can't remember which)

pkg/providers/instance/instance.go
jonathan-innis: This is for CreateTag tagging after the fact on the migration path. If we want to scope the permission such that Karpenter can't modify the cluster tag on instances that don't already belong to the cluster, then we have to assert that this tag already exists on the instance. In that case, it's superfluous to modify this tag since we know it's already there.

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
jonathan-innis: There is a _slight_ difference. Maybe we can discuss if it makes sense to consolidate them. If you were to combine these two statements, then you would require the request that is generated by the RunInstances call to the LaunchTemplate to _require_ the `kubernetes.io/cluster/<cluster-name>` request tag. In my testing, I wasn't able to get this to work since whatever request is generated to the launch template by fleet doesn't seem to propagate the tag, even if I specify the tag in the tagSpecifications (maybe @bwagner5 knows more about this). 

By separating it out, I'm able to allow the CreateFleet/RunInstances call to succeed to make its request to the launch template while also enforcing that whichever launch templates that Karpenter creates must have the relevant tags

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
jonathan-innis: Makes sense. Yes, they should be combined

pkg/providers/instance/instance.go
micahhausler: 👍

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
micahhausler: > you would require the request that is generated by the RunInstances call to the LaunchTemplate to require the kubernetes.io/cluster/<cluster-name> request tag.

I do think this is something we want right?

> I wasn't able to get this to work since whatever request is generated to the launch template by fleet doesn't seem to propagate the tag, even if I specify the tag in the tagSpecifications 

We may need to follow up with EC2 folks on this.

jonathan-innis: Yeah, I think so. If it's not propagating right now, do you think it's fine to iterate on this template moving forward?

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
jonathan-innis: Is there anything else that's cross-account that we should support by default beyond subents?

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

bwagner5 (approved):
lgtm

bwagner5 (dismissed):
LGTM

bwagner5 (approved):
lgtm

jonathan-innis (commented):
/karpenter snapshot

njtran (approved):
lgtm

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

PR #3949, opened 2023-05-26
feat: Securitygroup Drift

engedaam:
**Description**
* Security group drift will check drift between `AWSNodeTemplate.Spec.SecurityGroupSelector` and `machines` with resolved values returned from EC2

* Unite Tests
* Integration Tests 
* Manually Tested

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


engedaam:
Related to #1738

website/content/en/preview/concepts/deprovisioning.md
```diff
@@ -142,12 +142,15 @@ data:
...
 
-If users annotate their own nodes with `karpenter.sh/voluntary-disruption: "drifted"`, Karpenter will respect the annotation and deprovision the nodes.
+The current fields that support drift:
+ - AWSNodeTemplate.Spec.amiFamily 
+ - AWSNodeTemplate.Spec.amiSelector
+ - AWSNodeTemplate.Spec.securityGroupSelector  
 
 {{% alert title="Note" color="primary" %}}
-Karpenter will only automatically mark nodes as drifted in the case of a drifted AMI. More methods of drift will be implemented in the future. Please cut a feature request if you'd like to see more methods implemented.
+If users annotate their own nodes with `karpenter.sh/voluntary-disruption: "drifted"`, Karpenter will respect the annotation and deprovision the nodes.
```
njtran, line 153: This is actually incorrect since we changed it. https://github.com/aws/karpenter-core/blob/main/pkg/controllers/node/drift.go#L49-L65 I'll make the change

njtran (commented):
Nice work. I think the tests need to be adjusted too

pkg/cloudprovider/drift.go
```diff
@@ -73,3 +83,28 @@ func (c *CloudProvider) isAMIDrifted(ctx context.Context, machine *v1alpha5.Mach
 	}
 	return !lo.Contains(lo.Keys(mappedAMIs), instance.ImageID), nil
 }
+
+func (c *CloudProvider) isSecurityGroupDrifted(ec2Instance *instance.Instance, nodeTemplate *v1alpha1.AWSNodeTemplate) bool {
+	securityGroupIds := lo.Map(nodeTemplate.Status.SecurityGroups, func(sg v1alpha1.SecurityGroup, _ int) string { return sg.ID })
```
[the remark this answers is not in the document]
engedaam: Do you mean in the event that securityGroup is update upstream?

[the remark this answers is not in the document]
engedaam: The status will reconcile as soon as the `nodetemplate` is updated. The drift check will not happen if the `nodetemplate` gets updated, because drift check will occurs every 5 min

test/suites/drift/suite_test.go
```diff
@@ -148,20 +148,83 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+						},
+					},
+				},
+			},
+		}
+		newSecurityGroup, err := env.EC2API.CreateSecurityGroup(createSecurityGroup)
+		Expect(err).To(BeNil())
+		DeferCleanup(func() {
+			// Need to make sure that all instance launched with the security group are terminated
+			time.Sleep(time.Minute)
```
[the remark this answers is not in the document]
engedaam: We need to wait until the instances if fully terminated before we can delete the security group

pkg/cloudprovider/suite_test.go
```diff
@@ -230,6 +232,9 @@ var _ = Describe("CloudProvider", func() {
 					CreationDate: aws.String("2022-08-15T12:00:00Z"),
 				}},
 			})
+			awsEnv.EC2API.DescribeSecurityGroupsOutput.Set(&ec2.DescribeSecurityGroupsOutput{
+				SecurityGroups: []*ec2.SecurityGroup{{GroupId: aws.String(validSecurityGroup)}},
+			})
```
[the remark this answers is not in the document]
engedaam: Added the controller. We still need to make sure the fake EC2API has the new Securitygroup, no?

pkg/cloudprovider/drift.go
```diff
@@ -73,3 +83,28 @@ func (c *CloudProvider) isAMIDrifted(ctx context.Context, machine *v1alpha5.Mach
...
+
+	for _, instanceSecurityGroup := range ec2Instance.SecurityGroupIDs {
```
jonathan-innis, line 90: `lo.Every`?

pkg/cloudprovider/drift.go
```diff
@@ -25,16 +25,26 @@ import (
...
+	ec2Instance, err := c.getInstance(ctx, machine.Status.ProviderID)
+	if err != nil {
+		return false, cloudprovider.IgnoreMachineNotFoundError(err)
+	}
+
 	amiDrifted, err := c.isAMIDrifted(ctx, machine, provisioner, nodeTemplate)
 	if err != nil {
 		return false, cloudprovider.IgnoreMachineNotFoundError(fmt.Errorf("calculating ami drift, %w", err))
 	}
+	securitygroupDrifted := c.isSecurityGroupDrifted(ec2Instance, nodeTemplate)
```
jonathan-innis, line 42:
```suggestion
	securitygroupDrifted := c.areSecurityGroupsDrifted(ec2Instance, nodeTemplate)
```

nit: It's plural since we are validating that all security groups are propagated to the instance

pkg/cloudprovider/drift.go
```diff
@@ -73,3 +83,28 @@ func (c *CloudProvider) isAMIDrifted(ctx context.Context, machine *v1alpha5.Mach
 	}
 	return !lo.Contains(lo.Keys(mappedAMIs), instance.ImageID), nil
 }
+
+func (c *CloudProvider) isSecurityGroupDrifted(ec2Instance *instance.Instance, nodeTemplate *v1alpha1.AWSNodeTemplate) bool {
```
jonathan-innis, line 87: nit: Can we have a function comment to describe what this function does

pkg/cloudprovider/suite_test.go
```diff
@@ -314,6 +327,23 @@ var _ = Describe("CloudProvider", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(isDrifted).To(BeTrue())
 		})
+		It("should return drifted if the securitygroup is not valid", func() {
+			node := coretest.Node(coretest.NodeOptions{
```
jonathan-innis, line 331: Can this test be based on a machine now that we are relying on Machines instead of nodes here?

pkg/cloudprovider/suite_test.go
```diff
@@ -314,6 +327,23 @@ var _ = Describe("CloudProvider", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(isDrifted).To(BeTrue())
 		})
+		It("should return drifted if the securitygroup is not valid", func() {
```
jonathan-innis, line 330: We should ensure the opposite of this. That is, that we don't return drifted when the security groups are equal

jonathan-innis, line 330: Can we make sure we cover all the edge cases? What about when the security groups on the instance are a subset of the valid security groups?

pkg/providers/instance/types.go
```diff
@@ -46,7 +47,10 @@ func NewInstance(out *ec2.Instance) *Instance {
 		Type:         aws.StringValue(out.InstanceType),
 		Zone:         aws.StringValue(out.Placement.AvailabilityZone),
 		CapacityType: lo.Ternary(out.SpotInstanceRequestId != nil, v1alpha5.CapacityTypeSpot, v1alpha5.CapacityTypeOnDemand),
-		Tags:         lo.SliceToMap(out.Tags, func(t *ec2.Tag) (string, string) { return aws.StringValue(t.Key), aws.StringValue(t.Value) }),
+		SecurityGroupIDs: lo.Map(out.SecurityGroups, func(securitygroup *ec2.GroupIdentifier, _ int) string {
```
jonathan-innis, line 50: Can this value also be set for `NewInstanceFromFleet`? It's probably fine if not, but we should keep this consistent if we can

test/suites/drift/suite_test.go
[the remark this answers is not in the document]
jonathan-innis: Can we do an Eventually to at least have a polling mechanism to check that the instance is gone before we try to delete the security group?

test/suites/drift/suite_test.go
```diff
@@ -148,20 +148,83 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+			})
+			Expect(err).To(BeNil())
+			_, err = env.EC2API.DeleteSecurityGroup(&ec2.DeleteSecurityGroupInput{
+				GroupId: newSecurityGroup.GroupId,
+			})
+			Expect(err).To(BeNil())
+		})
+		provider := awstest.AWSNodeTemplate(v1alpha1.AWSNodeTemplateSpec{AWS: v1alpha1.AWS{
+			SecurityGroupSelector: map[string]string{"aws-ids": fmt.Sprintf("%s,%s,%s", *securitygroup[0].GroupId, *securitygroup[1].GroupId, awssdk.StringValue(newSecurityGroup.GroupId))},
+			SubnetSelector:        map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName},
```
jonathan-innis, line 196: What's the need to specify a custom AMI in this case? I think we just need the security group customization, right?

test/suites/drift/suite_test.go
```diff
@@ -148,20 +148,83 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+				},
+			},
+		})
+
+		env.ExpectCreated(pod, provider, provisioner)
+		env.EventuallyExpectHealthy(pod)
+		env.ExpectCreatedNodeCount("==", 1)
+
+		node := env.Monitor.CreatedNodes()[0]
+		provider.Spec.SecurityGroupSelector = map[string]string{"aws-ids": fmt.Sprintf("%s,%s", *securitygroup[0].GroupId, *securitygroup[1].GroupId)}
```
jonathan-innis, line 218: Nice 🎉

test/suites/drift/suite_test.go
```diff
@@ -148,20 +148,83 @@ var _ = Describe("Drift", Label("AWS"), func() {
 			g.Expect(env.Client.Get(env.Context, client.ObjectKeyFromObject(node), &v1.Node{})).To(Succeed())
 		}).WithTimeout(time.Minute).Should(Succeed())
 	})
-})
+	It("should deprovision nodes that have drifted due to securitygroup", func() {
```
jonathan-innis, line 151: Can you consider using `By()` statements? We've started using those more and that gives us some indication of the step that is occurring at a given time for free from Ginkgo

test/suites/drift/suite_test.go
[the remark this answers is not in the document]
jonathan-innis: Why does a straight `Delete` not work here?

website/content/en/preview/concepts/deprovisioning.md
```diff
@@ -142,9 +142,12 @@ data:
 
 ## Drift
 
-If drift is enabled, Karpenter will deprovision nodes that have been marked as drifted with the annotation `karpenter.sh/voluntary-disruption: "drifted"`. Karpenter will automatically cordon, drain, and terminate nodes, while respecting any PDBs or `do-not-evict` pods that are configured. Karpenter will automatically mark nodes as drifted if the AMI that is used on the instance does not match the AMI set by the AWSNodeTemplate. Check the [AWSNodeTemplate Docs]({{<ref "./node-templates" >}}) settings for more.
+If drift is enabled, Karpenter will deprovision nodes that have been marked as drifted with the annotation `karpenter.sh/voluntary-disruption: "drifted"`. Karpenter will automatically cordon, drain, and terminate nodes, while respecting any PDBs or `do-not-evict` pods that are configured. Karpenter will automatically mark nodes as drifted if instances contain configurations does not match the resolved sets from the AWSNodeTemplate. Check the [AWSNodeTemplate Docs]({{<ref "./node-templates" >}}) settings for more.
```
jonathan-innis, line 145: Should we start outlining the difference between the different types of drift in the docs. i.e. Here's the types of drift (static, dynamic). Here's the fields in each bucket and how they are handled

test/pkg/environment/aws/expectations.go
```diff
@@ -192,7 +192,8 @@ func (env *Environment) GetSubnetNameAndIds(tags map[string]string) []SubnetInfo
 
 type SecurityGroup struct {
 	ec2.GroupIdentifier
-	Tags []*ec2.Tag
+	VpcID *string
```
jonathan-innis, line 195: Why do we need this additional info?

pkg/cloudprovider/drift.go
jonathan-innis, line 87: Security groups can also be defined through an external launch template with launcTemplateName. I think this will break those users that are using that, so we need to make sure that we gate this drift to avoid drifting for those users

pkg/cloudprovider/drift.go
```diff
@@ -25,16 +25,26 @@ import (
...
 	"github.com/aws/karpenter/pkg/apis/v1alpha1"
 	"github.com/aws/karpenter/pkg/providers/amifamily"
+	"github.com/aws/karpenter/pkg/providers/instance"
 	"github.com/aws/karpenter/pkg/utils"
 )
 
 func (c *CloudProvider) isNodeTemplateDrifted(ctx context.Context, machine *v1alpha5.Machine, provisioner *v1alpha5.Provisioner, nodeTemplate *v1alpha1.AWSNodeTemplate) (bool, error) {
+	ec2Instance, err := c.getInstance(ctx, machine.Status.ProviderID)
+	if err != nil {
+		return false, cloudprovider.IgnoreMachineNotFoundError(err)
```
njtran, line 35: Would be nice if we could do the `IgnoreMachineNotFoundError()` call in the parent call and just return err here and in the other places, since it seems like we always want to ignore not found

pkg/cloudprovider/drift.go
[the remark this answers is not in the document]
njtran: I meant in the case that the AWSNodeTemplate Status reconciliation fails, and there's no value in the `nodeTemplate.Status.SecurityGroups` field.

pkg/cloudprovider/drift.go
```diff
@@ -73,3 +83,28 @@ func (c *CloudProvider) isAMIDrifted(ctx context.Context, machine *v1alpha5.Mach
 	}
 	return !lo.Contains(lo.Keys(mappedAMIs), instance.ImageID), nil
 }
+
+// Checks if the security groups are drifted, by comparing the AWSNodeTemplate.Status.SecurityGroups
+// to the ec2 instance security groups
+func (c *CloudProvider) areSecurityGroupsDrifted(ec2Instance *instance.Instance, nodeTemplate *v1alpha1.AWSNodeTemplate) bool {
+	securityGroupIds := lo.Map(nodeTemplate.Status.SecurityGroups, func(sg v1alpha1.SecurityGroup, _ int) string { return sg.ID })
+
+	lo.EveryBy(securityGroupIds, func(id string) bool {
```
njtran, line 92: Don't we need to return this?

website/content/en/preview/concepts/deprovisioning.md
njtran: I added a design doc for this https://github.com/aws/karpenter-core/pull/366

website/content/en/preview/concepts/deprovisioning.md
njtran: Actually can you make the change and just remove this here?

pkg/cloudprovider/suite_test.go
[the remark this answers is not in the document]
njtran: Not if we're relying on the AWSNodeTemplate status as the source of truth, right? We shouldn't be making this describe call, that should be on the AWSNodeTemplate Controller.

pkg/cloudprovider/suite_test.go
```diff
@@ -215,10 +218,14 @@ var _ = Describe("CloudProvider", func() {
...
 	Context("Node Drift", func() {
 		var validAMI string
+		var validSecurityGroup string
 		var selectedInstanceType *corecloudproivder.InstanceType
 		var instance *ec2.Instance
+		var nodeTemplateController corecontroller.Controller
 		BeforeEach(func() {
 			validAMI = fake.ImageID()
+			validSecurityGroup = fake.SecurityGroupID()
+			nodeTemplateController = nodetemplate.NewController(env.Client, awsEnv.SubnetProvider, awsEnv.SecurityGroupProvider, awsEnv.AMIProvider)
```
njtran, line 228: We probably don't need this controller and a fake API call, we can probably just set the status to what we want it to be, since this is really about testing the cloud provider functions, not the node template controller.

pkg/cloudprovider/suite_test.go
```diff
@@ -314,6 +327,42 @@ var _ = Describe("CloudProvider", func() {
...
+				},
+			})
+			// Instance is a reference to what we return in the GetInstances call
+			instance.SecurityGroups = []*ec2.GroupIdentifier{{GroupId: aws.String(fake.SecurityGroupID())}}
+			ExpectReconcileSucceeded(ctx, nodeTemplateController, client.ObjectKeyFromObject(nodeTemplate))
+			isDrifted, err := cloudProvider.IsMachineDrifted(ctx, machine)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isDrifted).To(BeTrue())
+		})
+		It("should not return drifted if the securitygroup is valid", func() {
```
njtran, line 349: Can you do this with multiple security groups?

test/suites/drift/suite_test.go
```diff
@@ -148,20 +148,87 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+		env.ExpectSettingsOverridden(map[string]string{
+			"featureGates.driftEnabled": "true",
+		})
+		By("creating a new securitygroup")
+		securitygroup := env.GetSecurityGroups(map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName})
+		Expect(len(securitygroup)).To(BeNumerically("==", 2))
+
+		createSecurityGroup := &ec2.CreateSecurityGroupInput{
+			GroupName:   awssdk.String("security-group-drift"),
+			Description: awssdk.String("Drift Test"),
```
njtran, line 161: Since we're creating a resource here, we should be as descriptive about this so that we can easily identify this and clean this up later, if it leaks. 

Can you add something on top of this that infers something special about the environment and put it in the group name and in the description?
```suggestion
			Description: awssdk.String("End-to-end Drift Test"),
```

test/suites/drift/suite_test.go
```diff
@@ -148,20 +148,87 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+					},
+				},
+			},
+		}
+		newSecurityGroup, err := env.EC2API.CreateSecurityGroup(createSecurityGroup)
+		Expect(err).To(BeNil())
+		DeferCleanup(func() {
+			By("deleting the new securitygroup")
+			// // Need to make sure that all instance launched with the security group are terminated
+			// time.Sleep(time.Minute)
```
njtran, line 180: Remove?

test/suites/drift/suite_test.go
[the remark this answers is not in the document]
engedaam: The Security group will not delete because it is dependent on the network interface to be detached from the instance

engedaam (commented):
/karpenter snapshot

pkg/cloudprovider/drift.go
```diff
@@ -73,3 +83,34 @@ func (c *CloudProvider) isAMIDrifted(ctx context.Context, machine *v1alpha5.Mach
 	}
 	return !lo.Contains(lo.Keys(mappedAMIs), instance.ImageID), nil
 }
+
+// Checks if the security groups are drifted, by comparing the AWSNodeTemplate.Status.SecurityGroups
+// to the ec2 instance security groups
+func (c *CloudProvider) areSecurityGroupsDrifted(ec2Instance *instance.Instance, nodeTemplate *v1alpha1.AWSNodeTemplate) (bool, error) {
+	// nodeTemplate.Spec.SecurityGroupSelector can be nill if the user is using a launchTemplateName to define SecurityGroups
```
njtran, line 90:
```suggestion
	// nodeTemplate.Spec.SecurityGroupSelector can be nil if the user is using a launchTemplateName to define SecurityGroups
```

pkg/cloudprovider/drift.go
```diff
@@ -73,3 +83,34 @@ func (c *CloudProvider) isAMIDrifted(ctx context.Context, machine *v1alpha5.Mach
...
+	// Karpenter will not drift on changes to securitygroup in the launchTemplateName
+	if nodeTemplate.Spec.SecurityGroupSelector == nil && nodeTemplate.Spec.LaunchTemplateName != nil {
+		return false, nil
+	}
```
njtran, line 94: This should be caught by validation, right? AFAIK, a user can't have both of these nil.

pkg/cloudprovider/suite_test.go
```diff
@@ -314,6 +323,104 @@ var _ = Describe("CloudProvider", func() {
...
+					},
+				},
+			})
+			// Instance is a reference to what we return in the GetInstances call
+			instance.SecurityGroups = []*ec2.GroupIdentifier{{GroupId: aws.String(fake.SecurityGroupID())}, {GroupId: aws.String(validSecurityGroup)}}
+			isDrifted, err := cloudProvider.IsMachineDrifted(ctx, machine)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isDrifted).To(BeTrue())
+		})
+		It("should return drifted if more AWSNodeTemplate securitygroup are present then instance securitygroup", func() {
```
njtran, line 362:
```suggestion
		It("should return drifted if more AWSNodeTemplate securitygroups are present than instance securitygroups", func() {
```

test/suites/drift/suite_test.go
```diff
@@ -148,20 +149,89 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+					v1alpha5.DoNotEvictPodAnnotationKey: "true",
+				},
+			},
+		})
+
+		env.ExpectCreated(pod, provider, provisioner)
+		env.EventuallyExpectHealthy(pod)
+		env.ExpectCreatedNodeCount("==", 1)
+		By("updating the provider securitygroup")
+		node := env.Monitor.CreatedNodes()[0]
```
njtran, line 223: I think you can use the following to be more deterministic about which node we're talking about.
`node := env.ExpectScheduled(pod)`

We don't want any other accidentally created nodes messing with this test.

njtran (commented):
Nice work! Pretty close, just a couple of comments

test/suites/drift/suite_test.go
engedaam: Isn't `env.ExpectScheduled(pod)` only for our unit testing?

website/content/en/preview/concepts/deprovisioning.md
```diff
@@ -170,7 +170,7 @@ Read the [Drift Design](https://github.com/aws/karpenter-core/pull/366/files) fo
 |AWSNodeTemplate Fields      | Static | Dynamic | Behavioral | Implemented |
 |----------------------------|  :---: |  :---:  |   :---:    |    :---:    |
 | Subnet Selector            |        |    x    |            |             |
-| Security Group Selector    |        |    x    |            |             |
```
jonathan-innis, line 173: Any reason this is getting modified?

engedaam (commented):
/karpenter snapshot

test/suites/drift/suite_test.go
```diff
@@ -148,20 +150,84 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+		output, err := env.EKSAPI.DescribeCluster(&eks.DescribeClusterInput{Name: awssdk.String(settings.FromContext(env.Context).ClusterName)})
+		Expect(err).To(BeNil())
+
+		By("looking for security groups")
+		securitygroups := env.GetSecurityGroups(map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName})
+		testSecurityGroup, found := lo.Find(securitygroups, func(sg aws.SecurityGroup) bool {
+			return awssdk.StringValue(sg.GroupName) == "security-group-drift"
+		})
+
+		if !found {
```
jonathan-innis, line 164: nit: This technically has a race where it could be created in-between. Not a big deal for a test scenario but I think it might be simpler to just attempt to create it and then ignore the AlreadyExists error if that's what comes back from the API

pkg/cloudprovider/suite_test.go
```diff
@@ -314,6 +323,104 @@ var _ = Describe("CloudProvider", func() {
...
+					},
+				},
+			})
+			// Instance is a reference to what we return in the GetInstances call
+			instance.SecurityGroups = []*ec2.GroupIdentifier{{GroupId: aws.String(fake.SecurityGroupID())}, {GroupId: aws.String(validSecurityGroup)}}
+			isDrifted, err := cloudProvider.IsMachineDrifted(ctx, machine)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isDrifted).To(BeTrue())
+		})
+		It("should return drifted if more AWSNodeTemplate securitygroups are present than instance securitygroups", func() {
```
jonathan-innis, line 362: should we have a test case for the inverse of this? `should return drifted if more instance security groups are present than AWSNodeTemplate security groups`

pkg/cloudprovider/suite_test.go
```diff
@@ -314,6 +323,104 @@ var _ = Describe("CloudProvider", func() {
...
+					},
+				},
+			})
+			// Instance is a reference to what we return in the GetInstances call
+			instance.SecurityGroups = []*ec2.GroupIdentifier{{GroupId: aws.String(fake.SecurityGroupID())}}
+			isDrifted, err := cloudProvider.IsMachineDrifted(ctx, machine)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isDrifted).To(BeTrue())
+		})
+		It("should return drifted if multiple instance securitygroup do not match AWSNodeTemplate securitygroups", func() {
```
jonathan-innis, line 344: This test case just checks if they don't match in general, IIUC? I can't tell from just reading the description for the test case here so I have to dig into the code to understand what's going on

pkg/cloudprovider/suite_test.go
```diff
@@ -314,6 +323,104 @@ var _ = Describe("CloudProvider", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(isDrifted).To(BeTrue())
 		})
+		It("should return drifted if the AWSNodeTemplate securitygroup do not match the machine", func() {
```
jonathan-innis, line 326: This checks the empty case IIUC, so maybe we should describe the test case as checking that

jonathan-innis, line 326: A sanity check for the empty case. It's not possible to have an empty set of security groups, right?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -194,7 +198,11 @@ func (c *CloudProvider) IsMachineDrifted(ctx context.Context, machine *v1alpha5.
 	if err != nil {
 		return false, client.IgnoreNotFound(fmt.Errorf("resolving node template, %w", err))
 	}
-	return c.isNodeTemplateDrifted(ctx, machine, provisioner, nodeTemplate)
+	drifted, err := c.isNodeTemplateDrifted(ctx, machine, provisioner, nodeTemplate)
+	if err != nil {
+		return false, cloudprovider.IgnoreMachineNotFoundError(err)
```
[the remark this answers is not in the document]
njtran: I think we should keep it as is here. If the instance is missing and the node is orphaned, we should let the garbage collection handle this, rather than trying to get the Drift node controller handle it.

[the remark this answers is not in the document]
jonathan-innis: Right, but my thought is that you should return errors from the interface as transparently as possible. If the instance is NotFound, that's an error that the client should be aware of and should be able to handle how they like. Think if we have some other package that is calling this package. Should we obfuscate that the instance isn't found and that's why we are returning false? IMO, I don't think we should.

pkg/cloudprovider/drift.go
```diff
@@ -73,3 +84,33 @@ func (c *CloudProvider) isAMIDrifted(ctx context.Context, machine *v1alpha5.Mach
...
+func (c *CloudProvider) areSecurityGroupsDrifted(ec2Instance *instance.Instance, nodeTemplate *v1alpha1.AWSNodeTemplate) (bool, error) {
+	// nodeTemplate.Spec.SecurityGroupSelector can be nil if the user is using a launchTemplateName to define SecurityGroups
+	// Karpenter will not drift on changes to securitygroup in the launchTemplateName
+	if nodeTemplate.Spec.LaunchTemplateName != nil {
+		return false, nil
+	}
+
+	securityGroupIds := sets.New(lo.Map(nodeTemplate.Status.SecurityGroups, func(sg v1alpha1.SecurityGroup, _ int) string { return sg.ID })...)
+	if len(securityGroupIds) == 0 {
+		return false, fmt.Errorf("no security groups exist given constraints")
```
njtran, line 99: This is the same error message as in the AWSNodeTemplate controller, right? We're not actually computing constraints here, just querying from the status. Can it be something like the following?
```suggestion
		return false, fmt.Errorf("no security groups exist in the AWSNodeTemplate Status")
```

pkg/cloudprovider/suite_test.go
njtran: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-security-groups.html#:~:text=When%20you%20launch%20an%20instance%20in%20a%20VPC%2C%20you%20must%20specify%20a%20security%20group%20that%27s%20created%20for%20that%20VPC. Looks like an instance has to have security groups, and the node template can only be empty if the ANT controller hasn't updated the status due to race or an error.

njtran: AFAIK, this shouldn't return drifted right? https://github.com/aws/karpenter/pull/3949/files#diff-24b89fbce3fa5104d7328f74ecdfe7e57d275728825e337fa0307f94f8b071c3R99 says we'll return false.

test/suites/drift/suite_test.go
njtran: You're absolutely right 🤦 , never mind!

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
engedaam: Yeah, that sounds good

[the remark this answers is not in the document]
engedaam: https://github.com/aws/karpenter-core/pull/373

pkg/cloudprovider/suite_test.go
engedaam: This is not checking the empty case. This is checking a normal drifted case

engedaam: In the empty case, drift will just error out. We will not drift

pkg/cloudprovider/suite_test.go
```diff
@@ -314,6 +323,137 @@ var _ = Describe("CloudProvider", func() {
...
+					},
+				},
+			})
+			// Instance is a reference to what we return in the GetInstances call
+			instance.SecurityGroups = []*ec2.GroupIdentifier{{GroupId: aws.String(fake.SecurityGroupID())}}
+			isDrifted, err := cloudProvider.IsMachineDrifted(ctx, machine)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isDrifted).To(BeTrue())
+		})
+		It("should return drifted if instance securitygroup is not contained AWSNodeTemplateStatus", func() {
```
njtran, line 344: nit on wording
```suggestion
		It("should return drifted if the instance securitygroups do not match the AWSNodeTemplateStatus", func() {
```

pkg/cloudprovider/suite_test.go
```diff
@@ -314,6 +323,137 @@ var _ = Describe("CloudProvider", func() {
...
+					ID:   fake.SecurityGroupID(),
+					Name: "test-securitygroup",
+				},
+			}
+			ExpectApplied(ctx, env.Client, nodeTemplate)
+			isDrifted, err := cloudProvider.IsMachineDrifted(ctx, machine)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isDrifted).To(BeTrue())
+		})
+		It("should return drifted if more instance security groups are present than AWSNodeTemplate security groups", func() {
```
njtran, line 389: This is actually the same test case as L344, right? This one tests 2 vs 3, whereas the L344 tests 1 vs 2 SGs. I don't know if this gives us anymore coverage.

test/suites/drift/suite_test.go
```diff
@@ -148,20 +150,77 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+				},
+			},
+		}
+		_, _ = env.EC2API.CreateSecurityGroup(createSecurityGroup)
+
+		By("looking for security groups")
+		securitygroups := env.GetSecurityGroups(map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName})
+		testSecurityGroup, _ := lo.Find(securitygroups, func(sg aws.SecurityGroup) bool {
+			return awssdk.StringValue(sg.GroupName) == "security-group-drift"
+		})
```
njtran, line 181: Can we do some sort of eventually here? I can imagine since we just created this that there could be some races.

pkg/cloudprovider/drift.go
```diff
@@ -73,3 +84,33 @@ func (c *CloudProvider) isAMIDrifted(ctx context.Context, machine *v1alpha5.Mach
 	}
 	return !lo.Contains(lo.Keys(mappedAMIs), instance.ImageID), nil
 }
+
+// Checks if the security groups are drifted, by comparing the AWSNodeTemplate.Status.SecurityGroups
+// to the ec2 instance security groups
+func (c *CloudProvider) areSecurityGroupsDrifted(ec2Instance *instance.Instance, nodeTemplate *v1alpha1.AWSNodeTemplate) (bool, error) {
+	// nodeTemplate.Spec.SecurityGroupSelector can be nil if the user is using a launchTemplateName to define SecurityGroups
+	// Karpenter will not drift on changes to securitygroup in the launchTemplateName
+	if nodeTemplate.Spec.LaunchTemplateName != nil {
```
jonathan-innis, line 93: I was thinking about this more and I'm wondering if we should encapsulate this behavior in the AWSNodeTemplate controller. Meaning, that we should set the resolved security groups to `nil` when a LaunchTemplateName is defined so that we can just evaluate using L98 and we don't have think about this specific case in this context as well

test/suites/drift/suite_test.go
jonathan-innis: This is a good callout. Does `CreateSecurityGroup` not just give us the securityGroup back if it's already created or is it only an error. If it's the latter, then I think the `Eventually` makes sense.

test/suites/drift/suite_test.go
```diff
@@ -148,20 +150,77 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+							Value: awssdk.String(settings.FromContext(env.Context).ClusterName),
+						},
+					},
+				},
+			},
+		}
+		_, _ = env.EC2API.CreateSecurityGroup(createSecurityGroup)
+
+		By("looking for security groups")
+		securitygroups := env.GetSecurityGroups(map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName})
```
jonathan-innis, line 178: Is there a way to get the security group by name or id so you don't have to do all this processing to get the security group ID?

test/suites/drift/suite_test.go
engedaam: `CreateSecurityGroup` only provides the SG id

engedaam: `CreateSecurityGroup` does not give us back an ID if it's already created. That's why don't think an `Eventually` will work here

test/suites/drift/suite_test.go
engedaam: No don't have a way for that. If the `CreateSecurityGroup` creates the SG, we then would have a SG ID. However, when `CreateSecurityGroup` does not create the SG will not have any information.

pkg/cloudprovider/drift.go
engedaam: Currently, AWSNodeTemplate controller does place a `nil` for the status when LaunchTemplateName is set. We should't have errors, if the user use LaunchTemplateName. In L98 we will error if the status does not have any SGs

engedaam (commented):
/karpenter snapshot

njtran (approved):

njtran:
Great work on responding to feedback and getting this in!

PR #3951, opened 2023-05-27
fix: Don't Delete Instances if Orphaned

jonathan-innis:
Fixes # 

**Description**

The Machine Linker Controller shouldn't delete instances that have the `v1alpha5.ProvisionerName` label but do not have Provisioners present on the cluster. These instances should either be assumed to be actively deleting (in which case a delete is unnecessary) or be assumed to have been orphaned when the provisioner is deleted. If the Node/Instance was orphaned when the Provisioner was deleted, we shouldn't take any action on the instance.

* `make presubmit`
* `/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

### DIFF
--- a/pkg/controllers/machine/link/controller.go
+++ b/pkg/controllers/machine/link/controller.go
@@ -25,7 +25,6 @@ import (
 	"github.com/samber/lo"
 	"go.uber.org/multierr"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


jonathan-innis (commented):
/karpenter snapshot

PR #3953, opened 2023-05-29
fix: Prevent Machine Link Controller workqueue clogging

Fedosin:
**Description**
Now, to prevent requeueing of requests in the machine link controller we return `reconcile.Result{RequeueAfter: math.MaxInt64}`, which results in an ever-growing queue and memory leakage, as it only delayed the requeueing requests with an extremely long wait time.

This PR improves the handling of requests requeueing by returning an empty `reconcile.Result{}`, which already has Requeue flag set to false.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/machine/link/controller.go
+++ b/pkg/controllers/machine/link/controller.go
@@ -17,7 +17,6 @@ package link
 import (
 	"context"
 	"fmt"
-	"math"
 	"time"
 
 	"github.com/patrickmn/go-cache"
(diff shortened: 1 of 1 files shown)


jonathan-innis:
@Fedosin I'm curious if you were seeing memory leakage from the controller. This controller implements our [singleton controller design](https://github.com/aws/karpenter-core/blob/main/pkg/operator/controller/singleton.go#L97) which is an extension of the base controller-runtime event-based handling. This controller runs a single loop and constantly re-reconciles when it either receives a `requeueAfter` or an empty `reconcile.Result`. Thus, if we returned an empty `reconcile.Result`, we would just get an immediate requeue here.

Granted, I think we should probably update the [singleton controller design](https://github.com/aws/karpenter-core/blob/main/pkg/operator/controller/singleton.go#L97) to not requeue on am empty result (we'd be open to taking a contribution to update that) but, at present, I think this change causes more requeues than neccesary.

njtran:
Closing, feel free to reopen if you are able to work on this.

PR #3955, opened 2023-05-29
feat: create kubernetes events in case of insufficient capacity errors

Fedosin:
Fixes # 

**Description**
We need to have improved visibility of ICE events when karpenter encounters them.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Karpenter creates Kubernetes events in case of insufficient capacity errors.
```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -47,6 +47,7 @@ import (
 	coreapis "github.com/aws/karpenter-core/pkg/apis"
 	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
 	"github.com/aws/karpenter-core/pkg/cloudprovider"
+	"github.com/aws/karpenter-core/pkg/events"
 )
 
 func init() {
@@ -61,14 +62,16 @@ type CloudProvider struct {
 	instanceProvider     *instance.Provider
 	kubeClient           client.Client
 	amiProvider          *amifamily.Provider
+	recorder             events.Recorder
 }
 
-func New(instanceTypeProvider *instancetype.Provider, instanceProvider *instance.Provider, kubeClient client.Client, amiProvider *amifamily.Provider) *CloudProvider {
+func New(instanceTypeProvider *instancetype.Provider, instanceProvider *instance.Provider, kubeClient client.Client, amiProvider *amifamily.Provider, recorder events.Recorder) *CloudProvider {
 	return &CloudProvider{
 		instanceTypeProvider: instanceTypeProvider,
 		instanceProvider:     instanceProvider,
 		kubeClient:           kubeClient,
 		amiProvider:          amiProvider,
+		recorder:             recorder,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


pkg/cloudprovider/cloudprovider.go
```diff
@@ -85,10 +88,26 @@ func (c *CloudProvider) Create(ctx context.Context, machine *v1alpha5.Machine) (
...
+			Type:           v1.EventTypeWarning,
+			Reason:         "InsufficientCapacityError",
+			Message:        fmt.Sprintf("Machine %s event: all requested instance types were unavailable during launch", machine.Name),
+			DedupeValues:   []string{machine.Name},
+		})
 		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("all requested instance types were unavailable during launch"))
 	}
 	instance, err := c.instanceProvider.Create(ctx, nodeTemplate, machine, instanceTypes)
 	if err != nil {
+		if cloudprovider.IsInsufficientCapacityError(err) {
```
jonathan-innis, line 102: Is it possible for us to surface this over in karpenter-core? I'm wondering if [this](https://github.com/aws/karpenter-core/blob/main/pkg/controllers/machine/lifecycle/launch.go#L111) might be a good place to surface the event when launch fails generally for this reason.

jonathan-innis (commented):
Thanks for the contribution! Just one thought on making this apply generally.

Fedosin:
@jonathan-innis Thank you for your suggestion! I moved this code to karpenter-core: https://github.com/aws/karpenter-core/pull/356

Fedosin:
@jonathan-innis I'm going to close this one, so we can continue discussion in https://github.com/aws/karpenter-core/pull/356

pkg/cloudprovider/cloudprovider.go
jonathan-innis: > the ability to see the instance family, instance type to the event when karpenter

What's your ultimate goal here? Is it to track down all the instance types and offerings that are ICEd at a given time? If that's the case, then I suspect something like [this](https://github.com/aws/karpenter/issues/2596) might be a better mechanism to surface that. 

I'm not opposed to surfacing those details also in the K8s event, I'm just curious what your ultimate goal is here.

Fedosin: @jonathan-innis we are buildin a kubernetes controller around these things, so prometheus metrics are not really helpful for us. We'd like to have a kubernetes event object that we can react to.
Technically I don't have preferences where to create it - in [karpenter](https://github.com/aws/karpenter/pull/3955) or in [karpenter-core](https://github.com/aws/karpenter-core/pull/356). Both ways work for us :)

jonathan-innis: Ah, makes sense. I'm wondering if we should just create a generic event for each individual ICE error based on the instance type rather than creating the event directly on the Machine. It seems like what you are interested in here is getting the high-level eventing detail that there has been a generic ICE error on an instance-type/AZ combination and you aren't necessarily interested that a Machine launch failed generically due to ICE. 

What are your thoughts on firing an event with no involved object that details the instance type and az of the instance type that received the ICE?

tasdikrahman: > It seems like what you are interested in here is getting the high-level eventing detail that there has been a generic ICE error on an instance-type/AZ combination and you aren't necessarily interested that a Machine launch failed generically due to ICE.

Hello there, @jonathan-innis . Thanks for the summary, yes, we are trying to have some more visibility added for the ICE event on the AZ/instance type combination.

What we are thinking in the direction further with this event is that, with that visibility, we would want our controller to stop recommending the instance family which we are trying to decide as a scheduling constraint for the application spec. To prevent a scenario where the application would stay in pending state due to non-availability of the instance-types in that instance family.

bwagner5:
It looks like other PRs are going to implement this change, so closing this one.

PR #4030, opened 2023-06-12
fix bottlerocket node-labels merge in user-data

bwagner5:
Fixes # 
https://github.com/aws/karpenter/issues/4013

**Description**
 - Fixes Node Labels merging in Bottlerocket's user-data.
 - Fixes invalid toml in docs

* tested manually by:
   * adding labels that should be overridden like `karpenter.sh/capacity-type` and observed it being properly overridden. 
   * adding labels that should not be overridden and observing them properly merged
* make test

- [ x Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/amifamily/bootstrap/bottlerocket.go
+++ b/pkg/providers/amifamily/bootstrap/bottlerocket.go
@@ -21,6 +21,7 @@ import (
 
 	"knative.dev/pkg/ptr"
 
+	"github.com/imdario/mergo"
 	"github.com/samber/lo"
 
 	"github.com/aws/karpenter-core/pkg/utils/resources"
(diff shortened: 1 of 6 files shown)


jonathan-innis (approved):
Seems like this was missed because our `...golden` was wrong. Should we assert that we have valid TOML in our test cases so that this doesn't creep in with something else?

bwagner5:
> Seems like this was missed because our `...golden` was wrong. Should we assert that we have valid TOML in our test cases so that this doesn't creep in with something else?

wdym? The golden was valid toml, the docs were not

jonathan-innis:
> wdym? The golden was valid toml, the docs were not

Ooop. I got it. I misread the original issue. Yep, makes sense and looks good 👍

PR #4069, opened 2023-06-16
test: fix node created count

njtran:
Fixes # 

**Description**
When a test would create and delete nodes at the same time, this would just check the difference, rather than actually returning how many nodes were created. 

* `TEST_SUITE=scale ENABLE_CLOUDWATCH=true FOCUS="single consolidation replace" make e2etests`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/test/pkg/environment/common/monitor.go
+++ b/test/pkg/environment/common/monitor.go
@@ -101,7 +101,7 @@ func (m *Monitor) NodeCountAtReset() int {
 
 // CreatedNodeCount returns the number of nodes created since the last reset
 func (m *Monitor) CreatedNodeCount() int {
-	return m.NodeCount() - m.NodeCountAtReset()
+	return len(m.CreatedNodes())
 }
 
 // NodesAtReset returns a slice of nodes that the monitor saw at the last reset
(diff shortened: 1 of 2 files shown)


test/pkg/environment/common/monitor.go
```diff
@@ -101,7 +101,7 @@ func (m *Monitor) NodeCountAtReset() int {
 
 // CreatedNodeCount returns the number of nodes created since the last reset
 func (m *Monitor) CreatedNodeCount() int {
```
jonathan-innis, line 103: Doesn't this still fail since we don't have persistent tracking of all the nodes that we have created? I think CreatedNodes() still assumes there are no deletions

jonathan-innis: Discussed offline. These two operations are different because one uses the raw data to determine which of the current set of nodes didn't exist at the reset while the other just blindly subtracts. We should favor using the raw data here.

jonathan-innis (dismissed):
LGTM 🚀

PR #4112, opened 2023-06-22
docs: Updated latest release version to v0.28.1

ellistarn:
Fixes #N/A 

**Description**

Our docs process is broken for cherry-picked patch releases. I had to do this manually 👨‍🍳 

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/hack/release/prepare-website.sh
+++ b/hack/release/prepare-website.sh
@@ -6,7 +6,7 @@ source "${SCRIPT_DIR}/common.sh"
 
 config
 
-GIT_TAG=$(git describe --exact-match --tags || echo "none")
+GIT_TAG=${GIT_TAG:-$(git describe --exact-match --tags || echo "none")}
 if [[ $(releaseType "$GIT_TAG") != $RELEASE_TYPE_STABLE ]]; then
   echo "Not a stable release. Missing required git tag."
   exit 1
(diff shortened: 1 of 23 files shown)


jonathan-innis (commented):
Are all these docs changes supposed to go in since the v0.28.1 should just have the single commit? I think you mainly just need to update the latest release version in the patch version update

hack/release/prepare-website.sh
```diff
@@ -6,7 +6,7 @@ source "${SCRIPT_DIR}/common.sh"
 
 config
 
-GIT_TAG=$(git describe --exact-match --tags || echo "none")
+GIT_TAG=${GIT_TAG:-$(git describe --exact-match --tags || echo "none")}
```
bwagner5, line 9:
```suggestion
GIT_TAG=$(git describe --exact-match --tags || echo "none")
```

bwagner5 (approved):
lgtm

bwagner5:
> Are all these docs changes supposed to go in since the v0.28.1 should just have the single commit? I think you mainly just need to update the latest release version in the patch version update

the minor version docs are static now, so these updates are needed. latest-release only affects preview now. We need to get rid of /docs/ (got distracted and haven't tried the symlink yet) and then this will be a little more straight forward (less changes at least).

PR #4128, opened 2023-06-26
fix: only warn for ssm query errors

everpcpc:
**Description**
In some region without specific AMI such as `cn-northwest-1` , launching would fail with errors like:

```
2023-06-26T08:13:14.481Z	ERROR	controller.provisioner	launching machine, creating cloud provider instance, creating instance, getting launch template configs, getting launch templates, getting ssm parameter "/aws/service/bottlerocket/aws-k8s-1.26-nvidia/arm64/latest/image_id", ParameterNotFound: 	{"commit": "698f22f-dirty"}
```
This error should be ignored since we do not need this kind of AMIs.

Tested with  AWS `cn-northwest-1`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/controllers/nodetemplate/suite_test.go
+++ b/pkg/controllers/nodetemplate/suite_test.go
@@ -573,8 +573,85 @@ var _ = Describe("AWSNodeTemplateController", func() {
 						},
 					},
 				},
-			},
-			))
+			}))
+		})
+		It("should resolve amiSelector AMis and requirements into status when all SSM aliases don't resolve", func() {
+			version := lo.Must(awsEnv.AMIProvider.KubeServerVersion(ctx))
+			// This parameter set doesn't include any of the Nvidia AMIs
+			awsEnv.SSMAPI.Parameters = map[string]string{
+				fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/x86_64/latest/image_id", version): "ami-id-123",
+				fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/arm64/latest/image_id", version):  "ami-id-456",
+			}
+			nodeTemplate.Spec.AMIFamily = &v1alpha1.AMIFamilyBottlerocket
+			nodeTemplate.Spec.AMISelector = nil
+			awsEnv.EC2API.DescribeImagesOutput.Set(&ec2.DescribeImagesOutput{
+				Images: []*ec2.Image{
+					{
+						Name:         aws.String("test-ami-1"),
+						ImageId:      aws.String("ami-id-123"),
+						CreationDate: aws.String(time.Now().Format(time.RFC3339)),
+						Architecture: aws.String("x86_64"),
+						Tags: []*ec2.Tag{
+							{Key: aws.String("Name"), Value: aws.String("test-ami-1")},
+							{Key: aws.String("foo"), Value: aws.String("bar")},
+						},
+					},
+					{
+						Name:         aws.String("test-ami-2"),
+						ImageId:      aws.String("ami-id-456"),
+						CreationDate: aws.String(time.Now().Add(time.Minute).Format(time.RFC3339)),
+						Architecture: aws.String("arm64"),
+						Tags: []*ec2.Tag{
+							{Key: aws.String("Name"), Value: aws.String("test-ami-2")},
+							{Key: aws.String("foo"), Value: aws.String("bar")},
+						},
+					},
+				},
+			})
+			ExpectApplied(ctx, env.Client, nodeTemplate)
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(nodeTemplate))
+			nodeTemplate = ExpectExists(ctx, env.Client, nodeTemplate)
+			sortRequirements(nodeTemplate.Status.AMIs)
+			Expect(nodeTemplate.Status.AMIs).To(ContainElements([]v1alpha1.AMI{
+				{
+					Name: "test-ami-1",
+					ID:   "ami-id-123",
+					Requirements: []v1.NodeSelectorRequirement{
+						{
+							Key:      v1.LabelArchStable,
+							Operator: v1.NodeSelectorOpIn,
+							Values:   []string{v1alpha5.ArchitectureAmd64},
+						},
+						{
+							Key:      v1alpha1.LabelInstanceGPUCount,
+							Operator: v1.NodeSelectorOpDoesNotExist,
+						},
+						{
+							Key:      v1alpha1.LabelInstanceAcceleratorCount,
+							Operator: v1.NodeSelectorOpDoesNotExist,
+						},
+					},
+				},
+				{
+					Name: "test-ami-2",
+					ID:   "ami-id-456",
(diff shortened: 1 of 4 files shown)


pkg/providers/amifamily/ami.go
```diff
@@ -160,10 +160,14 @@ func (p *Provider) getDefaultAMIFromSSM(ctx context.Context, nodeTemplate *v1alp
 	for _, ssmOutput := range ssmRequirements {
 		amiID, err := p.fetchAMIsFromSSM(ctx, ssmOutput.Query)
 		if err != nil {
-			return nil, err
+			logging.FromContext(ctx).With("query", ssmOutput.Query).Warnf("discover ami from ssm failed: %v", err)
```
jonathan-innis, line 163: I think you still want this as an error here rather than a warning.

pkg/providers/amifamily/ami.go
```diff
@@ -160,10 +160,14 @@ func (p *Provider) getDefaultAMIFromSSM(ctx context.Context, nodeTemplate *v1alp
...
+			continue
 		}
 		amis = append(amis, AMI{AmiID: amiID, Requirements: ssmOutput.Requirements})
 	}
+	if len(amis) == 0 {
```
jonathan-innis, line 168: I think you can let the outer caller handle this condition rather than trying to handle it within the calling function directly

jonathan-innis (commented):
Just a couple comments; otherwise, LGTM

jonathan-innis:
Would you be willing to add a functional unit test for this new scenario where the SSM alias isn't available but some other alias is available within the `ami_test.go` file as well?

jonathan-innis:
> Would you be willing to add a functional unit test for this new scenario where the SSM alias isn't available but some other alias is available within the `ami_test.go` file as well?

Went ahead and added some functional tests around this functionality

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

pkg/providers/amifamily/ami_test.go
```diff
@@ -12,178 +12,346 @@ See the License for the specific language governing permissions and
...
-			aws.String("0123456789"),
-			aws.String("self"),
-		}))
-		Expect(filters).Should(ConsistOf([]*ec2.Filter{
-			{
-				Name:   aws.String("name"),
-				Values: aws.StringSlice([]string{"my-ami"}),
+		amis, err := awsEnv.AMIProvider.Get(ctx, nodeTemplate, &amifamily.Options{})
+		Expect(err).ToNot(HaveOccurred())
+		Expect(amis).To(HaveLen(6))
```
njtran, line 160: Same here for 6, shouldn't it be 4?

pkg/providers/amifamily/ami_test.go
```diff
@@ -12,178 +12,346 @@ See the License for the specific language governing permissions and
...
+		// No GPU AMI exists for AM64 here
+		awsEnv.SSMAPI.Parameters = map[string]string{
+			fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/x86_64/latest/image_id", version):        amd64AMIName,
+			fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s-nvidia/x86_64/latest/image_id", version): amd64NvidiaAMIName,
+			fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s/arm64/latest/image_id", version):         arm64AMIName,
+		}
+		// Only 4 of the requirements sets for the SSM aliases will resolve
+		amis, err := awsEnv.AMIProvider.Get(ctx, nodeTemplate, &amifamily.Options{})
+		Expect(err).ToNot(HaveOccurred())
+		Expect(amis).To(HaveLen(4))
```
njtran, line 217: Same here, confused why this isn't 3.

pkg/providers/amifamily/ami_test.go
```diff
@@ -12,178 +12,346 @@ See the License for the specific language governing permissions and
...
-		}))
+		})
+		awsEnv.SSMAPI.Parameters = map[string]string{
+			fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2/recommended/image_id", version):       amd64AMIName,
+			fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2-gpu/recommended/image_id", version):   amd64NvidiaAMIName,
+			fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2-arm64/recommended/image_id", version): arm64AMIName,
+		}
+		amis, err := awsEnv.AMIProvider.Get(ctx, nodeTemplate, &amifamily.Options{})
+		Expect(err).ToNot(HaveOccurred())
+		Expect(amis).To(HaveLen(4))
```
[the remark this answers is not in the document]
jonathan-innis: Because there are 4 separate requirements sets that can't be combined. We can't have a single requirements set that says you can have "this or this" so they have to be separate requirements sets to fit the "OR" semantic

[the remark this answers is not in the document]
jonathan-innis: See: https://github.com/aws/karpenter/blob/main/pkg/providers/amifamily/al2.go#L37

pkg/controllers/nodetemplate/suite_test.go
```diff
@@ -573,8 +573,85 @@ var _ = Describe("AWSNodeTemplateController", func() {
...
+							{Key: aws.String("Name"), Value: aws.String("test-ami-2")},
+							{Key: aws.String("foo"), Value: aws.String("bar")},
+						},
+					},
+				},
+			})
+			ExpectApplied(ctx, env.Client, nodeTemplate)
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(nodeTemplate))
+			nodeTemplate = ExpectExists(ctx, env.Client, nodeTemplate)
+			sortRequirements(nodeTemplate.Status.AMIs)
```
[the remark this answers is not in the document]
jonathan-innis: You are correct sir, I forgot about that 🎉

pkg/providers/amifamily/ami_test.go
jonathan-innis: Same response as above. See the requirements sets. We have to create disparate sets of the requirements that can be ORed together

pkg/providers/amifamily/ami_test.go
jonathan-innis: Same response as the others

njtran (approved):

pkg/controllers/nodetemplate/suite_test.go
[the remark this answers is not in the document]
jonathan-innis: I take it back. It's not possible to do this since the `requirements` are inside of the `AMI` and are converted from a map in code, which means that they will come in different orders each time if we don't sort them

njtran (approved):

everpcpc:
Sorry for missing notification on it, thanks a lot for helping went this ahead!

PR #4211, opened 2023-07-06
docs: [RFC] Add Unmanaged Launch Template Removal RFC

jonathan-innis:
Fixes #N/A 

**Description**

This PR adds a RFC design doc recommending that Karpenter drop the `spec.launchTemplateName` field from the new `v1beta1/EC2NodeClass` and provides the rationale behind the decision, including alternative options.

N/A

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/designs/unmanaged-launch-template-removal.md
+++ b/designs/unmanaged-launch-template-removal.md
@@ -1,0 +1,59 @@
+# RFC: Unmanaged LaunchTemplate Removal
+
+## Overview
+
+EC2 fleet requires a launch template to be created and referenced in the CreateFleet call prior to the launch of an instance. Karpenter supports two methods for the user to reference this launch template:
+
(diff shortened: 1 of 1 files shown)


sftim:
If we want to support custom launch templates, I would add a whole new custom resource, eg AWSCustomNodeTemplate. For example, we could have that as beta-grade, tested, best-effort code and an alpha API version.

To highlight the alpha state, bump the alpha API version every _n_ months and deprecate old-enough alpha APIs.

jonathan-innis:
>  I would add a whole new custom resource

This is a really interesting idea. If we had enough users doing this, I could imagine us doing something like this. This at least draws the line better between the two concepts so that we don't have two api concepts that have different incompatibilities with each other that make it hard to reason about.

ellistarn (approved):

PR #4252, opened 2023-07-11
[DRAFT] Add capacity memory overhead generation

jonathan-innis:
Fixes #N/A 

**Description**

- [] Yes, issue opened: # 
- [x] No


jonathan-innis:
Interesting thing to note here is that each AMI appears to be reporting slightly different capacity overhead values (at least testing Bottlerocket and AL2 I can see that they are different). It might be worth just taking the min of all the known AMIFamily values and using that to make sure that we don't overschedule to nodes like we have seen in some other issues.

PR #4258, opened 2023-07-11
test: Add SG and OIDC Clean-up

engedaam:
Fixes #N/A 

**Description**
- Add clean up of SG and leaked OICD 

- Manually tested 

- [] Yes, issue opened: # 
- [x] No


.github/workflows/sweeper.yaml
```diff
@@ -9,14 +9,18 @@ permissions:
 jobs:
   sweeper:
     if: github.repository == 'aws/karpenter' || github.event_name == 'workflow_dispatch'
+    strategy:
```
jonathan-innis, line 12: Nice 🎉

test/hack/cleanup/main.go
```diff
@@ -105,6 +132,22 @@ func main() {
 	if err := fireMetric(ctx, cloudWatchClient, "LaunchTemplatesDeleted", float64(deleted)); err != nil {
 		logger.With("name", "LaunchTemplatesDeleted").Errorf("firing metric, %v", err)
 	}
+
+	names = getOldOICD(ctx, iamClient, expirationTime)
```
jonathan-innis, line 136:
```suggestion
	names = getOldOIDCProviders(ctx, iamClient, expirationTime)
```

test/hack/cleanup/main.go
```diff
@@ -71,10 +76,32 @@ func main() {
 
 	// Terminate any old stacks that were provisioned as part of testing
 	// We execute these in serial since we will most likely get rate limited if we try to delete these too aggressively
+	sgInAccount := lo.Must(ec2Client.DescribeSecurityGroups(ctx, &ec2.DescribeSecurityGroupsInput{
```
jonathan-innis, line 79: Can we just delete these security groups if they are beyond the expiration time like the other things?

jonathan-innis, line 79: Also, a follow-up question: is this just a back-stop for the security group not deleting as part of the actual test run? Can we make changes so that, in the general case, we delete the security group and successfully delete the VPC but we only use the sweeper if we fail to do so for some other reason

test/hack/cleanup/main.go
```diff
@@ -208,3 +251,23 @@ func getOldLaunchTemplates(ctx context.Context, ec2Client *ec2.Client, expiratio
 	}
 	return names
 }
+
+func getOldOICD(ctx context.Context, iamClient *iam.Client, expirationTime time.Time) (names []string) {
+	testSuite := []string{"upgrade", "chaos", "consolidation", "drift", "integration", "interruption", "ipv6", "machine", "scale", "utilization"}
```
jonathan-innis, line 256: Is it possible for us to identify this OIDC provider by some other tag that is only added by a test run? Otherwise, we are going to have to keep expanding this list as we add tests and we are going to forget to maintain this list

test/hack/cleanup/main.go
```diff
@@ -105,6 +132,22 @@ func main() {
 	if err := fireMetric(ctx, cloudWatchClient, "LaunchTemplatesDeleted", float64(deleted)); err != nil {
```
jonathan-innis, line 132: Side task: can we configure these metrics to fire to timestream, since we are using this for our metrics storage now?

test/hack/cleanup/main.go
```diff
@@ -105,6 +132,22 @@ func main() {
...
 		logger.With("name", "LaunchTemplatesDeleted").Errorf("firing metric, %v", err)
 	}
+
+	names = getOldOICD(ctx, iamClient, expirationTime)
+	deleted = 0
+	for i := range names {
```
jonathan-innis, line 138:
```suggestion
	for i := range arns {
```
If this is ARNs, can we name the variable as such?

test/hack/cleanup/main.go
```diff
@@ -105,6 +132,22 @@ func main() {
...
+		if _, err := iamClient.DeleteOpenIDConnectProvider(ctx, &iam.DeleteOpenIDConnectProviderInput{
+			OpenIDConnectProviderArn: lo.ToPtr(names[i]),
+		}); err != nil {
+			logger.With("name", names[i]).Errorf("deleting test cluster oidc, %v", err)
```
jonathan-innis, line 142:
```suggestion
			logger.With("arn", arns[i]).Errorf("deleting test cluster oidc provider, %v", err)
```

test/hack/cleanup/main.go
```diff
@@ -105,6 +132,22 @@ func main() {
...
+		} else {
+			logger.With("name", names[i]).Infof("deleted test cluster oidc")
```
jonathan-innis, line 144:
```suggestion
			logger.With("arn", names[i]).Infof("deleted test cluster oidc provider")
```

.github/workflows/sweeper.yaml
```diff
@@ -9,14 +9,18 @@ permissions:
...
+      fail-fast: false
+      matrix:
+        region: [us-east-2, us-west-2]
```
jonathan-innis, line 15: Should we go ahead and add the `eu-west` region ahead of us expanding the conformance tests over there?

test/hack/cleanup/main.go
jonathan-innis: What about the VPC? Can we validate the age of the owning VPC and see if this is beyond the timestamp?

test/hack/cleanup/main.go
jonathan-innis: Any chance we can discover these names by introspecting the `.github/workflows` file or by reading the directory names. I have some concern that if we specify the test suites like this we are going to continually forget to update these values.

jonathan-innis: Another option is to do the inverse i.e. any OIDC provider that doens't have the `do-not-delete` tag gets cleaned up.

test/hack/cleanup/main.go
engedaam: The VPC does not have creation timestamp as well

test/hack/cleanup/main.go
engedaam: We don't create the OIDCs, so I'm not sure how we can tag them

test/hack/cleanup/main.go
engedaam: For the sake of scope, we can do that when we handle resource leak monitoring

test/hack/cleanup/main.go
```diff
@@ -68,13 +80,37 @@ func main() {
 			}
 		}
 	}
+}
 
-	// Terminate any old stacks that were provisioned as part of testing
-	// We execute these in serial since we will most likely get rate limited if we try to delete these too aggressively
+// Terminate any old stacks that were provisioned as part of testing
+// We execute these in serial since we will most likely get rate limited if we try to delete these too aggressively
+func cleanupOldstack(ctx context.Context, ec2Client *ec2.Client, cloudWatchClient *cloudwatch.Client, cloudFormationClient *cloudformation.Client, expirationTime time.Time, logger *zap.SugaredLogger) {
```
jonathan-innis, line 87: I feel like we should try to separate out deleting the security groups in the VPC from deleting the Stacks if we can. I'm not a big fan of us tying these two operations together in this one function.

.github/actions/e2e/cleanup/action.yaml
```diff
@@ -30,6 +30,18 @@ runs:
     - uses: ./.github/actions/e2e/install-eksctl
       with:
         eksctl_version: v0.147.0
+    - name: delete-security-group
+      shell: bash
+      run: | 
+        aws ec2 describe-security-groups \
```
jonathan-innis, line 36: Can we describe why we delete the security group here. Either in the YAML or in the script as a comment?

.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -125,6 +125,16 @@ runs:
         wellKnownPolicies:
           ebsCSIController: true
       EOF
+  - name: tag oidc provider of the cluster
+    shell: bash
+    run: |
+      for arn in $(aws iam list-open-id-connect-providers --query "OpenIDConnectProviderList[*].{ARN:Arn}" --output text); do
```
jonathan-innis, line 131: Nice! Can we tag it with the github run url like we do in the clusterConfig so that these tags are consistent?

test/cloudformation/iam_cloudformation.yaml
```diff
@@ -66,12 +66,15 @@ Resources:
               - iam:GetRole
               - iam:GetRolePolicy
               - iam:ListAttachedRolePolicies
+              - iam:ListOpenIDConnectProviders
+              - iam:TagOpenIDConnectProvider
```
jonathan-innis, line 70: Can we add these policies to the list of OpenID permissions? I believe there's a block that already covers OIDC permissions right now

test/hack/cleanup/main.go
```diff
@@ -68,13 +85,42 @@ func main() {
...
+	if err != nil {
+		logger.With("error", err).Error("getting security groups")
+	}
+	logger.With("ids", ids, "count", len(ids)).Infof("discovered test security groups to delete")
+	deleted := 0
 
-	// Terminate any old stacks that were provisioned as part of testing
-	// We execute these in serial since we will most likely get rate limited if we try to delete these too aggressively
-	names := getOldStacks(ctx, cloudFormationClient, expirationTime)
+	workqueue.ParallelizeUntil(ctx, 10, len(ids), func(i int) {
```
jonathan-innis, line 98: Is it possible for us to get throttled if we do this this way?

test/hack/cleanup/main.go
```diff
@@ -51,10 +56,22 @@ func main() {
 	ec2Client := ec2.NewFromConfig(cfg)
 	cloudFormationClient := cloudformation.NewFromConfig(cfg)
 	cloudWatchClient := cloudwatch.NewFromConfig(cfg)
+	iamClient := iam.NewFromConfig(cfg)
 
-	// Terminate any old instances that were provisioned by Karpenter as part of testing
-	// We execute these in serial since we will most likely get rate limited if we try to delete these too aggressively
-	ids := getOldInstances(ctx, ec2Client, expirationTime)
+	cleanupInstances(ctx, ec2Client, cloudWatchClient, expirationTime, logger)
```
jonathan-innis, line 61: When I was thinking about parallelizing, I was thinking that you would just parallelize each of these individual functions so that the categories are parallelized. I think you actually save more time that way and you're less likely to get throttled from what I could think.

test/suites/drift/suite_test.go
```diff
@@ -171,6 +171,10 @@ var _ = Describe("Drift", Label("AWS"), func() {
 							Key:   awssdk.String("karpenter.sh/discovery"),
 							Value: awssdk.String(settings.FromContext(env.Context).ClusterName),
 						},
+						{
+							Key:   awssdk.String("creation-date"),
```
jonathan-innis, line 175: Nice!

.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -125,6 +125,17 @@ runs:
         wellKnownPolicies:
           ebsCSIController: true
       EOF
+  - name: tag oidc provider of the cluster
```
jonathan-innis, line 128: Can we make sure that we do an `always()` on this so that we always attempt to tag the OIDC even if the previous step doesn't succeed?

jonathan-innis: That will give us a lower chance of leaking OIDC providers if something happens with the `eksctl create cluster`

test/cloudformation/iam_cloudformation.yaml
```diff
@@ -90,6 +91,8 @@ Resources:
             Action:
               - iam:CreateOpenIDConnectProvider
               - iam:DeleteOpenIDConnectProvider
+              - iam:ListOpenIDConnectProviders
```
jonathan-innis, line 94: Is there a permission that is associated with listing the OIDC provider tags since we are also doing that?

test/cloudformation/iam_cloudformation.yaml
```diff
@@ -72,6 +72,7 @@ Resources:
               - !Sub "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/eksctl-*"
               - !Sub "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/karpenter-irsa-*"
               - !Sub "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/prometheus-irsa-*"
+              - !Sub "arn:${AWS::Partition}:iam::${AWS::AccountId}:oidc-provider/*"
```
jonathan-innis, line 75: Why do we need to add this?

test/hack/cleanup/main.go
```diff
@@ -35,9 +39,15 @@ const (
 
 	karpenterProvisionerNameTag = "karpenter.sh/provisioner-name"
 	karpenterLaunchTemplateTag  = "karpenter.k8s.aws/cluster"
+	karpenterSecurityGroupTag   = "karpenter.sh/discovery"
 	githubRunURLTag             = "github.com/run-url"
 )
 
+type resources interface {
+	cleanupResources(ctx context.Context)
```
jonathan-innis, line 47:
```suggestion
	cleanup(ctx context.Context)
```

test/hack/cleanup/main.go
```diff
@@ -35,9 +39,15 @@ const (
...
+	getOldResources(ctx context.Context) (ids []string, err error)
```
jonathan-innis, line 48:
```suggestion
	getExpired(ctx context.Context) (ids []string, err error)
```

test/hack/cleanup/main.go
```diff
@@ -51,59 +61,171 @@ func main() {
...
+		&stack{cloudFormationClient: cloudFormationClient, cloudWatchClient: cloudWatchClient, expirationTime: expirationTime, logger: logger},
+		&launchtemplate{ec2Client: ec2Client, cloudWatchClient: cloudWatchClient, expirationTime: expirationTime, logger: logger},
+		&oidc{iamClient: iamClient, cloudWatchClient: cloudWatchClient, expirationTime: expirationTime, logger: logger},
+	}
 
-	// Terminate any old instances that were provisioned by Karpenter as part of testing
-	// We execute these in serial since we will most likely get rate limited if we try to delete these too aggressively
-	ids := getOldInstances(ctx, ec2Client, expirationTime)
-	logger.With("ids", ids, "count", len(ids)).Infof("discovered test instances to delete")
+	workqueue.ParallelizeUntil(ctx, 10, len(resources), func(i int) {
```
jonathan-innis, line 74: Why do we need 10 workers?

test/hack/cleanup/main.go
jonathan-innis: Why does this interface even need this method if it's not being called by some external thing?

test/hack/cleanup/main.go
```diff
@@ -51,59 +61,171 @@ func main() {
...
 	}
+}
 
-	// Terminate any old stacks that were provisioned as part of testing
-	// We execute these in serial since we will most likely get rate limited if we try to delete these too aggressively
-	names := getOldStacks(ctx, cloudFormationClient, expirationTime)
-	logger.With("names", names, "count", len(names)).Infof("discovered test stacks to delete")
+type securitygroup struct {
+	ec2Client        *ec2.Client
+	cloudWatchClient *cloudwatch.Client
```
jonathan-innis, line 110: This cloudwatch client is common to all of these? Is there a way to standardize this across the interfaces so each one doesn't need this?

test/hack/cleanup/main.go
```diff
@@ -51,59 +61,171 @@ func main() {
...
+	expirationTime   time.Time
```
jonathan-innis, line 111: Same comment on expiration time? Each of these resource structs has this value so is there a good way to standardize this so it's more DRY

test/hack/cleanup/main.go
```diff
@@ -25,19 +25,35 @@ import (
...
+	karpenterMetricNamespace = "testing.karpenter.sh/Cleanup"
 
 	karpenterProvisionerNameTag = "karpenter.sh/provisioner-name"
 	karpenterLaunchTemplateTag  = "karpenter.k8s.aws/cluster"
+	karpenterSecurityGroupTag   = "karpenter.sh/discovery"
 	githubRunURLTag             = "github.com/run-url"
 )
 
+type CleanableResourceType interface {
+	Type() string
```
jonathan-innis, line 47: I still feel like we could reduce a lot of repetition if we use a `Get()` method and then just use the get method to call the logger and then fire the metric after we do the cleanup. You could even return the items that you were able to successfully cleanup and fire a metric on that.

test/hack/cleanup/main.go
```diff
@@ -25,19 +25,35 @@ import (
...
 	"github.com/samber/lo"
+	"go.uber.org/multierr"
 	"go.uber.org/zap"
+	"k8s.io/client-go/util/workqueue"
 )
 
 const (
 	expirationTTL            = time.Hour * 12
-	karpenterMetricNamespace = "testing.karpenter.sh/cleanup"
+	karpenterMetricNamespace = "testing.karpenter.sh/Cleanup"
```
jonathan-innis, line 38: Why the change here?

engedaam (commented):
/karpenter snapshot

jonathan-innis (approved):
Awesome work 🚀

PR #4260, opened 2023-07-12
fix: Add support for CFS quota field in Bottlerocket

domgoodwin:
**Description**
Adds support to set the cfs quota field and have it passed through to Bottlerocket config. Currently `spec.kubeletConfiguration.cpuCFSQuota: false` is ignored on Bottlerocket hosts.

Deployed this image into a test cluster - this meant when `spec.kubeletConfiguration.cpuCFSQuota: false` was set this properly was fed down to the Kubelet on a Bottlerocket host.

- [] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/amifamily/bootstrap/bottlerocket.go
+++ b/pkg/providers/amifamily/bootstrap/bottlerocket.go
@@ -74,6 +74,9 @@ func (b Bottlerocket) Script() (string, error) {
 		if b.KubeletConfig.ImageGCLowThresholdPercent != nil {
 			s.Settings.Kubernetes.ImageGCLowThresholdPercent = lo.ToPtr(strconv.FormatInt(int64(*b.KubeletConfig.ImageGCLowThresholdPercent), 10))
 		}
+		if b.KubeletConfig.CPUCFSQuota != nil {
+			s.Settings.Kubernetes.CPUCFSQuota = b.KubeletConfig.CPUCFSQuota
+		}
 	}
 
 	s.Settings.Kubernetes.NodeTaints = map[string][]string{}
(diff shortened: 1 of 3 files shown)


pkg/providers/amifamily/bootstrap/bottlerocket.go
```diff
@@ -74,6 +74,9 @@ func (b Bottlerocket) Script() (string, error) {
 		if b.KubeletConfig.ImageGCLowThresholdPercent != nil {
```
jonathan-innis, line 74: @domgoodwin Do you mind valiadating that this pass-through works in functional testing and adding the setting pass-through in the E2E testing as well.

Here's some pointers to each:
1. Functional Testing: https://github.com/aws/karpenter/blob/main/pkg/providers/launchtemplate/suite_test.go#L1135
2. E2E Testing: https://github.com/aws/karpenter/blob/main/test/suites/integration/kubelet_config_test.go#L51

jonathan-innis (commented):
LGTM 🚀 Do you mind adding testing alongside these changes?

jonathan-innis:
Added the functional test to your code. From some brief testing, it seems like BR does not respond well to settings that it doesn't know (i.e. it won't startup if it has unknown settings passed to it). This causes issues if there are users that are pinned to an earlier version of BR (< 1.14.0) but we try to pass this setting through. [Having an open discussion in the #karpenter-dev Kubernetes slack](https://kubernetes.slack.com/archives/C04JW2J5J5P/p1689368471539629) to discuss how we should handle cases like this.

jonathan-innis:
> [Having an open discussion in the #karpenter-dev Kubernetes slack](https://kubernetes.slack.com/archives/C04JW2J5J5P/p1689368471539629) to discuss how we should handle cases like this.

Discussed in Slack and came to the conclusion that this shouldn't be an issue so removing the blocked label.

jonathan-innis (approved):
LGTM 🚀

PR #4268, opened 2023-07-13
test: Stop notification for conformance testing

engedaam:
Fixes #N/A 

**Description**
- Stop notification for conformance testing

- Tested inside of forked repo 

- [] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -118,7 +118,7 @@ jobs:
           TEST_SUITE="${{ inputs.suite }}" ENABLE_METRICS=${{ inputs.enable_metrics }} METRICS_REGION=${{ vars.TIMESTREAM_REGION }} GIT_REF="$(git rev-parse HEAD)" make e2etests
       - name: notify slack of success or failure
         uses: ./.github/actions/e2e/slack/notify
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-conformance-trigger.yaml
```diff
@@ -7,6 +7,7 @@ jobs:
   conformance:
     if: github.repository == 'aws/karpenter' || github.event_name == 'workflow_dispatch'
     strategy:
+      max-parallel: 1
```
jonathan-innis, line 10: Will this cause timeouts?

engedaam: It should not timeout. By default, each conformance job will have 6 hours to complete. The workflow has 35 days before timeout.

jonathan-innis: Ok, sounds good. I'd be more apt to increase our API limits first before restricting this parallelism. 1 is going to mean that all of these jobs are going to take a really long time to run so I think we should do a little validation to tweak the value to something that makes sense after we get our limits/quotas bumped.

jonathan-innis (approved):
🚀

PR #4274, opened 2023-07-17
docs: [RFC] Add v1beta1 API RFC

jonathan-innis:
Fixes #N/A 

**Description**

This PR proposes the `v1beta1` APIs and full Change List into the design directory

- [] Yes, issue opened: # 
- [x] No

### DIFF
--- a/designs/v1beta1-api.md
+++ b/designs/v1beta1-api.md
@@ -8,26 +8,41 @@ API changes create a user migration burden that should be weighed against the be
 
 ## Migration
 
-**Tenet:** Customers will be able to migrate from v1alpha5 to v1beta1 in a single cluster using any Karpenter version from the time that we release v1beta1 up until we release v1.
-
-Kubernetes custom resources have built in support for API version compatibility. CRDs with multiple versions must define a “storage version”, which controls the data stored in etcd. Other versions are views onto this data and converted using [conversion webhooks](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion). However, there is a fundamental limitation that [all versions must be safely round-trippable through each other](https://book.kubebuilder.io/multiversion-tutorial/api-changes.html)[.](https://book.kubebuilder.io/multiversion-tutorial/api-changes.html) This means that it must be possible to define a function that converts a v1alpha5 Provisioner into a v1beta1 Provisioner and vise versa.
+Kubernetes custom resources have built-in support for API version compatibility. CRDs with multiple versions must define a “storage version”, which controls the data stored in etcd. Other versions are views onto this data and converted using [conversion webhooks](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion). However, there is a fundamental limitation that [all versions must be safely round-trippable through each other](https://book.kubebuilder.io/multiversion-tutorial/api-changes.html)[.](https://book.kubebuilder.io/multiversion-tutorial/api-changes.html) This means that it must be possible to define a function that converts a v1alpha5 Provisioner into a v1beta1 Provisioner and vise versa.
 
 Unfortunately, multiple proposed changes in v1beta1 are not round-trippable. Below, we propose deprecations of legacy fields in favor more modern mechanisms that have seen adoption in v1alpha5. These changes remove sharp edges that regularly cause users surprises and production pain.
 
-To workaround the limitation of round-trippability, we are proposing a rename of both the API group and Kind (`NodePool`) that the CRDs exist within. This allows both CRDs to exist alongside each other simultaneously and gives users a natural migration path to move through.
+To workaround the limitation of round-trippability, we are proposing a rename of the Kinds (`NodePool`, `NodeClaim`, and `EC2NodeClass`) that the CRDs exist within. This allows both CRDs to exist alongside each other simultaneously and gives users a natural migration path to move through.
 
 ### Migration Path
 
-#### Aggressive Scale-Down
+Below describes a few migration paths at a high-level. These paths are not comprehensive, but offer good guidance through which users might migrate between the v1alpha5 APIs and the v1beta1 APIs.
 
-1. Spin up a new `v1beta1/NodePool` and `v1beta1/EC2NodeClass` resource matching their `v1alpha5` counterparts
-2. Delete the `v1alpha5/Provisioner` and `v1alpha5/AWSNodeTemplate` with [cascading foreground deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion) to delete the nodes and have Karpenter roll the nodes to `v1beta1`
+#### Periodic Rolling with Drift
 
-#### Slow Scale-Down
+For each Provisioner in your cluster, perform the following actions:
 
-1. Spin up a new `v1beta1/NodePool` and `v1beta1/EC2NodeClass` resource matching their `v1alpha5` counterparts
-2. Set the `spec.limits` on your `v1alpha5/Provisioner` resource to `cpu=0` to stop provisioning
-3. Manually delete Nodes one-by-one to have Karpenter roll the nodes over to `v1beta1`
+1. Create a NodePool/NodeClass in your cluster that is the v1beta1 equivalent of the v1alpha5 Provisioner/AWSNodeTemplate
+2. Add a taint to the old Provisioner such as `karpenter.sh/legacy=true:NoSchedule`
+3. Karpenter drift will mark all machines/nodes owned by that Provisioner as drifted
+4. Karpenter drift will launch replacements for the nodes in the new NodePool resource
+   1. Currently, Karpenter only supports rolling of one node at a time, which means that it may take some time for Karpenter to completely roll all nodes under a single Provisioner
+
+#### Forced Deletion
+
+For each Provisioner in your cluster, perform the following actions:
+
+1. Create a NodePool/NodeClass in your cluster that is the v1beta1 equivalent of the v1alpha5 Provisioner/AWSNodeTemplate
+2. Delete the old Provisioner with `kubectl delete provisioner <provisioner-name> --cascade=foreground`
+   1. Karpenter will delete each Node that is owned by the Provisioner, draining all nodes simultaneously and will launch nodes for the newly pending pods as soon as the Nodes enter a draining state
+
+#### Manual Rolling
+
+For each Provisioner in your cluster, perform the following actions:
+
+1. Create a NodePool/NodeClass in your cluster that is the v1beta1 equivalent of the v1alpha5 Provisioner/AWSNodeTemplate
+2. Add a taint to the old Provisioner such as `karpenter.sh/legacy=true:NoSchedule`
+3. Delete each node one-at-time owned by the Provisioner by running `kubectl delete node <node-name>`
 
 ## APIs
 
@@ -41,6 +56,8 @@ To help clearly define where configuration should live within Karpenter’s API,
     2. Cloudprovider-specific behavior-based fields for configuring Karpenter’s scheduling and deprovisioning decision-making (e.g. interruption-based disruption, allocation strategy)
 3. `NodeClaim`
     1. A Karpenter management object that fully manages the lifecycle of a single node including: configuring and launching the node, monitoring the node health (including disruption conditions), and handling the deprovisioning and termination of the node
+
+With these boundaries defined, below shows each API, with all fields specified, with values filled in as examples.
 
 ### `karpenter.sh/NodePool`
 
@@ -60,7 +77,7 @@ spec:
       nodeClass:
         name: default
         kind: EC2NodeClass
-        apiVersion: compute.k8s.aws/v1beta1
+        apiVersion: karpenter.k8s.aws/v1beta1
       taints:
         - key: example.com/special-taint
           effect: NoSchedule
@@ -119,10 +136,10 @@ status:
      ephemeral-storage: "100Gi"
 ```
 
-### `compute.k8s.aws/EC2NodeClass`
+### `karpenter.k8s.aws/EC2NodeClass`
 
 ```
-apiVersion: compute.k8s.aws/v1beta1
+apiVersion: karpenter.k8s.aws/v1beta1
 kind: EC2NodeClass
 metadata:
   name: default
@@ -292,15 +309,15 @@ status:
     memory: 942684Ki
     pods: "4"
   conditions:
-  - type: NodeDrifted
+  - type: Drifted
     status: "True"
     severity: Warning
   - status: "True"
-    type: NodeInitialized
+    type: Initialized
   - status: "True"
-    type: NodeLaunched
+    type: Lanched
   - status: "True"
-    type: NodeRegistered
+    type: Registered
   - status: "True"
     type: Ready
   nodeName: ip-192-168-62-137.us-west-2.compute.internal
@@ -317,23 +334,22 @@ status:
 4. `karpenter.sh/capacity-type`
 5. `karpenter.sh/do-not-disrupt`
 
-#### `compute.k8s.aws`
+#### `karpenter.k8s.aws`
 
-1. `compute.k8s.aws/instance-hypervisor`
-2. `compute.k8s.aws/instance-encryption-in-transit-supported`
-3. `compute.k8s.aws/instance-category`
-4. `compute.k8s.aws/instance-family`
-5. `compute.k8s.aws/instance-generation`
-6. `compute.k8s.aws/instance-local-nvme`
-7. `compute.k8s.aws/instance-size`
-8. `compute.k8s.aws/instance-cpu`
-9. `compute.k8s.aws/instance-memory`
-10. `compute.k8s.aws/instance-network-bandwidth`
-11. `compute.k8s.aws/instance-pods`
-12. `compute.k8s.aws/instance-gpu-name`
-13. `compute.k8s.aws/instance-gpu-manufacturer`
-14. `compute.k8s.aws/instance-gpu-count`
-15. `compute.k8s.aws/instance-gpu-memory`
-16. `compute.k8s.aws/instance-accelerator-name`
-17. `compute.k8s.aws/instance-accelerator-manufacturer`
-18. `compute.k8s.aws/instance-accelerator-count`
+1. `karpenter.k8s.aws/instance-hypervisor`
+2. `karpenter.k8s.aws/instance-encryption-in-transit-supported`
+3. `karpenter.k8s.aws/instance-category`
+4. `karpenter.k8s.aws/instance-family`
+5. `karpenter.k8s.aws/instance-generation`
+6. `karpenter.k8s.aws/instance-local-nvme`
+7. `karpenter.k8s.aws/instance-size`
+8. `karpenter.k8s.aws/instance-cpu`
+9. `karpenter.k8s.aws/instance-memory`
+10. `karpenter.k8s.aws/instance-network-bandwidth`
+11. `karpenter.k8s.aws/instance-gpu-name`
+12. `karpenter.k8s.aws/instance-gpu-manufacturer`
+13. `karpenter.k8s.aws/instance-gpu-count`
+14. `karpenter.k8s.aws/instance-gpu-memory`
+15. `karpenter.k8s.aws/instance-accelerator-name`
+16. `karpenter.k8s.aws/instance-accelerator-manufacturer`
+17. `karpenter.k8s.aws/instance-accelerator-count`
--- a/designs/v1beta1-api.md
+++ b/designs/v1beta1-api.md
@@ -1,0 +1,338 @@
+# Karpenter v1beta1 APIs
+
+This document formalizes the [v1beta1 laundry list](https://github.com/aws/karpenter/issues/1327) and describes the high-level migration strategy for users moving to v1beta1. It shows the full API specs, including Group/Kind names and label names. This document does not go into explicit details on each of the individual changes in v1beta1. For details on these individual changes, see [Karpenter v1beta1  Full Change List](./v1beta1-full-changelist.md).
+
+## Bake Time
+
+API changes create a user migration burden that should be weighed against the benefits of the breaking changes. Batching breaking changes into a single version bump **helps to minimize this burden**. The v1alpha5 API has seen broad adoption over the last year, and resulted in a large amount of feedback. We see this period to have been a critical maturation process for the Karpenter project, and has given us confidence that the changes in v1beta1 will be sufficient to promote after a shorter feedback period.
+
+## Migration
+
+**Tenet:** Customers will be able to migrate from v1alpha5 to v1beta1 in a single cluster using any Karpenter version from the time that we release v1beta1 up until we release v1.
+
+Kubernetes custom resources have built in support for API version compatibility. CRDs with multiple versions must define a “storage version”, which controls the data stored in etcd. Other versions are views onto this data and converted using [conversion webhooks](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion). However, there is a fundamental limitation that [all versions must be safely round-trippable through each other](https://book.kubebuilder.io/multiversion-tutorial/api-changes.html)[.](https://book.kubebuilder.io/multiversion-tutorial/api-changes.html) This means that it must be possible to define a function that converts a v1alpha5 Provisioner into a v1beta1 Provisioner and vise versa.
+
+Unfortunately, multiple proposed changes in v1beta1 are not round-trippable. Below, we propose deprecations of legacy fields in favor more modern mechanisms that have seen adoption in v1alpha5. These changes remove sharp edges that regularly cause users surprises and production pain.
+
+To workaround the limitation of round-trippability, we are proposing a rename of both the API group and Kind (`NodePool`) that the CRDs exist within. This allows both CRDs to exist alongside each other simultaneously and gives users a natural migration path to move through.
+
+### Migration Path
+
+#### Aggressive Scale-Down
+
+1. Spin up a new `v1beta1/NodePool` and `v1beta1/NodeTemplate` resource matching their `v1alpha5` counterparts
+2. Delete the `v1alpha5/Provisioner` and `v1alpha5/AWSNodeTemplate` with [cascading foreground deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion) to delete the nodes and have Karpenter roll the nodes to `v1beta1`
+
+#### Slow Scale-Down
+
+1. Spin up a new `v1beta1/NodePool` and `v1beta1/NodeTemplate` resource matching their `v1alpha5` counterparts
+2. Set the `spec.limits` on your `v1alpha5/Provisioner` resource to `cpu=0` to stop provisioning
+3. Manually delete Nodes one-by-one to have Karpenter roll the nodes over to `v1beta1`
+
+## APIs
+
+To help clearly define where configuration should live within Karpenter’s API, we should define the logical boundary between each Kind in the project.
+
+1. `NodePool`
+    1. Neutral Node configuration-based fields that affect the **compatibility between Nodes and Pods during scheduling** (e.g. requirements, taints, labels)
+    2. Neutral behavior-based fields for configuring Karpenter’s scheduling and deprovisioning decision-making
+2. `NodeTemplate`
+    1. Cloudprovider-specific Node configuration-based fields that affect launch and bootstrap process for that Node including: configuring startup scripts (including kubelet configuration), volume mappings, metadata settings, etc.
+    2. Cloudprovider-specific behavior-based fields for configuring Karpenter’s scheduling and deprovisioning decision-making (e.g. interruption-based disruption, allocation strategy)
+3. `Machine`
+    1. A Karpenter management object that fully manages the lifecycle of a single node including: configuring and launching the node, monitoring the node health (including disruption conditions), and handling the deprovisioning and termination of the node
+
+### `karpenter.sh/NodePool`
+
+```
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+metadata:
+  name: default
+spec:
+  template:
+    metadata:
+      labels:
+        billing-team: my-team
+      annotations:
+        example.com/owner: "my-team"
+    spec:
+      nodeTemplateRef:
+        name: default
+      taints:
+        - key: example.com/special-taint
+          effect: NoSchedule
+      startupTaints:
+        - key: example.com/another-taint
+          effect: NoSchedule
+      requirements:
+        - key: "karpenter.k8s.aws/instance-category"
+          operator: In
+          values: ["c", "m", "r"]
+      resources: # Most users wouldn't set this field but it's in the MachineSpec
+        requests:
+          cpu: "1"
+          memory: "100Mi"
+  deprovisioning:
+    ttlAfterUninitialized: 10m
+    ttlAfterUnderutilized: 30s
+    ttlUntilExpired: 30d
+    consolidation:
+      enabled: true/false
+  weight: 10
+  limits:
+    cpu: "1000"
+    memory: 1000Gi
+  kubeletConfiguration:
+    clusterDNS: ["10.0.1.100"]
+    containerRuntime: containerd
+    systemReserved:
+      cpu: 100m
+      memory: 100Mi
+      ephemeral-storage: 1Gi
+    kubeReserved:
+      cpu: 200m
+      memory: 100Mi
+      ephemeral-storage: 3Gi
+    evictionHard:
+      memory.available: 5%
+      nodefs.available: 10%
+      nodefs.inodesFree: 10%
+    evictionSoft:
+      memory.available: 500Mi
+      nodefs.available: 15%
+      nodefs.inodesFree: 15%
+    evictionSoftGracePeriod:
+      memory.available: 1m
+      nodefs.available: 1m30s
+      nodefs.inodesFree: 2m
+    evictionMaxPodGracePeriod: 60
+    imageGCHighThresholdPercent: 85
+    imageGCLowThresholdPercent: 80
+    cpuCFSQuota: true
+    podsPerCore: 2
+    maxPods: 20
+status:
+  resources:
+     cpu: "2"
+     memory: "100Mi"
+     ephemeral-storage: "100Gi"
+```
+
+### `compute.k8s.aws/NodeTemplate`
+
+```
+apiVersion: compute.k8s.aws/v1beta1
+kind: NodeTemplate
+metadata:
+  name: default
+spec:
+  amiFamily: AL2
+  amiSelector:
+    - tags: 
+        key: value
+    - id: abc-123
+    - name: foo
+      owner: amazon
+    - ssm: "/my/custom/parameter"
+  subnetSelector:
+    - tags:
+        compute.k8s.aws/discovery: cluster-name
+    - id: subnet-1234
+  securityGroupSelector:
+    - tags:
+        compute.k8s.aws/discovery: cluster-name
+    - name: default-security-group
+  role: karpenter-node-role
+  userData: |
+    echo "this is custom user data"
+  tags:
+    custom-tag: custom-value
+  metadataOptions:
+    httpEndpoint: enabled
+    httpProtocolIPv6: disabled
+    httpPutResponseHopLimit: 2
+    httpTokens: required
+  blockDeviceMappings:
+    - deviceName: /dev/xvda
+      ebs:
+        volumeSize: 20Gi
+        volumeType: gp3
+        encrypted: true
+  detailedMonitoring: true
+status:
+  subnets:
+    - id: subnet-0a462d98193ff9fac
+      zone: us-east-2b
+    - id: subnet-0322dfafd76a609b6
+      zone: us-east-2c
+    - id: subnet-0727ef01daf4ac9fe
+      zone: us-east-2b
+    - id: subnet-00c99aeafe2a70304
+      zone: us-east-2a
+    - id: subnet-023b232fd5eb0028e
+      zone: us-east-2c
+    - id: subnet-03941e7ad6afeaa72
+      zone: us-east-2a
+  securityGroups:
+    - id: sg-041513b454818610b
+      name: ClusterSharedNodeSecurityGroup
+    - id: sg-0286715698b894bca
+      name: ControlPlaneSecurityGroup-1AQ073TSAAPW
+  amis:
+    - id: ami-05a05e85b17bb60d7
+      name: amazon-eks-node-1.24-v20230703
+      requirements:
+        - key: karpenter.k8s.aws/instance-accelerator-count
+          operator: DoesNotExist
+        - key: kubernetes.io/arch
+          operator: In
+          values:
+            - amd64
+        - key: karpenter.k8s.aws/instance-gpu-count
+          operator: DoesNotExist
+    - id: ami-0d849ef1e65103147
+      name: amazon-eks-gpu-node-1.24-v20230703
+      requirements:
+        - key: kubernetes.io/arch
+          operator: In
+          values:
+            - amd64
+        - key: karpenter.k8s.aws/instance-accelerator-count
+          operator: Exists
+    - id: ami-0d849ef1e65103147
+      name: amazon-eks-gpu-node-1.24-v20230703
+      requirements:
+        - key: kubernetes.io/arch
+          operator: In
+          values:
+           - amd64
+        - key: karpenter.k8s.aws/instance-gpu-count
+          operator: Exists
+    - id: ami-0c3487f30d003deb3
+      name: amazon-eks-arm64-node-1.24-v20230703
+      requirements:
+        - key: kubernetes.io/arch
+          operator: In
+          values:
+            - arm64
+        - key: karpenter.k8s.aws/instance-gpu-count
+          operator: DoesNotExist
+        - key: karpenter.k8s.aws/instance-accelerator-count
+          operator: DoesNotExist
+```
+
+### `karpenter.sh/Machine`
+
+```
+apiVersion: karpenter.sh/v1beta1
+kind: Machine
+metadata:
+  name: default
+  labels:
+    billing-team: my-team
+  annotations:
+    example.com/owner: "my-team"
+spec:
+  nodeTemplateRef:
+    name: default
+  taints:
+    - key: example.com/special-taint
+      effect: NoSchedule
+  startupTaints:
+    - key: example.com/another-taint
+      effect: NoSchedule
+  requirements:
+    - key: "karpenter.k8s.aws/instance-category"
+      operator: In
+      values: ["c", "m", "r"]
+  resources:
+    requests:
+      cpu: "1"
+      memory: "100Mi"
+  kubeletConfiguration:
+    clusterDNS: ["10.0.1.100"]
+    containerRuntime: containerd
+    systemReserved:
+      cpu: 100m
+      memory: 100Mi
+      ephemeral-storage: 1Gi
+    kubeReserved:
+      cpu: 200m
+      memory: 100Mi
+      ephemeral-storage: 3Gi
+    evictionHard:
+      memory.available: 5%
+      nodefs.available: 10%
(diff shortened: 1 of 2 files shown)


ellistarn (commented):
Awesome work on this!!!

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+### `compute.k8s.aws/NodeClass`
+
+```
+apiVersion: compute.k8s.aws/v1beta1
+kind: NodeClass
+metadata:
+  name: default
+spec:
+  amiFamily: AL2
+  amiSelector:
```
sftim, line 132: This isn't really a selector (not in the Kubernetes sense). Can we call it a filter instead?

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+    - tags: 
+        key: value
+    - id: abc-123
+    - name: foo
+      owner: amazon
+    - ssm: "/my/custom/parameter"
+  subnetSelector:
```
sftim, line 139: This isn't really a selector (not in the Kubernetes sense). Can we call it a filter instead?

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+    - tags:
+        compute.k8s.aws/discovery: cluster-name
+    - id: subnet-1234
+  securityGroupSelector:
```
sftim, line 143: This isn't really a selector (not in the Kubernetes sense). Can we call it a filter instead?

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+    - id: subnet-023b232fd5eb0028e
+      zone: us-east-2c
+    - id: subnet-03941e7ad6afeaa72
+      zone: us-east-2a
+  securityGroups:
+    - id: sg-041513b454818610b
+      name: ClusterSharedNodeSecurityGroup
+    - id: sg-0286715698b894bca
+      name: ControlPlaneSecurityGroup-1AQ073TSAAPW
+  amis:
```
sftim, line 183: What if the list of matched AMIs is quite big (could that ever happen?)

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+metadata:
+  name: default
+  labels:
+    billing-team: my-team
+  annotations:
+    example.com/owner: "my-team"
+spec:
+  nodeClass:
+    name: default
+    kind: NodeClass
```
sftim, line 240: It's slightly more accurate to specify the API path (eg `nodeclasses`) than the kind (kinds can clash; API paths never do)

designs/v1beta1-full-changelist.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Update Kind/Group Naming
+
+As part of the bump to v1beta1, to allow the v1alpha5 APIs to exist alongside the v1beta1 APIs while users go through a migration process, the following renames are being proposed:
+
+1. API Group Renames
+    1. `karpenter.k8s.aws` → `compute.k8s.aws`
+2. Kind Renames
+    1. `Provisioner` → `NodePool`
+    2. `Machine` -> `NodeClaim`
```
sftim, line 13: How accurate is the “claim” analogy? It's not like I can adopt an existing running instance (cf. PersistentVolumeClaim, ResourceClaim), right? 

AIUI Karpenter's internal model is that it creates a Machine and then drives towards the desired state (healthy Node) by setting up instances.
You could even imagine a NodePool that has manual [ClickOps](https://www.lastweekinaws.com/blog/clickops/) provisioning with a team that should know better following some runbook to bring up a new node using the `poor-life-choice` node class.

designs/v1beta1-full-changelist.md
```diff
@@ -0,0 +1,200 @@
...
+    3. `AWSNodeTemplate` → `NodeClass`
+
+We see the renames as opportunities to better align our API groups and kinds with upstream concepts as well as reducing confusion between other Kubernetes API concepts. Specifically, the word `Provisioner` (on its own) has become overloaded in Kubernetes, [particularly in the area of storage provisioning](https://kubernetes.io/docs/concepts/storage/storage-classes/#the-storageclass-resource). We want to get completely away from this naming, while also prefixing all of our kinds that apply to nodes with `Node` for better alignment and clarity across the project. 
+
+This gives the following naming to API types within the Karpenter project
+
+1. `karpenter.sh/NodePool`
+2. `karpenter.sh/NodeClaim`
+3. `compute.k8s.aws/NodeClass`
```
sftim, line 22: If I have a cluster that uses AWS and (custom, not EKS Anywhere) on-prem nodes, what do I call the two different kinds? Are they both NodeClass in different API groups?

designs/v1beta1-full-changelist.md
```diff
@@ -0,0 +1,200 @@
...
+      taints: ...
+      startupTaints: ...
+      requirements: ...
+      providerRef: ...
+  deprovisioning:
+    ttlUntilExpired: ...
+    ttlAfterUnderutilized: ...
+```
+
+#### `provisioner.Spec.ttl...` → `nodePool.Spec.deprovisioning.ttl...`
```
sftim, line 75:
```suggestion
#### `provisioner.spec.ttl...` → `nodePool.spec.deprovisioning.ttl...`
```

designs/v1beta1-full-changelist.md
```diff
@@ -0,0 +1,200 @@
...
+
+If `consolidationPolicy` is not set, Karpenter will implicitly default to `WhenUnderutilized`.
+
+#### `provisioner.Spec.ttlSecondsAfterEmpty` → `nodePool.spec.deprovisioning.consolidationTTL`
+
+While the `consolidationPolicy` offers one mechanism for users to control the aggressiveness of deprovisioning, users that enable a `consolidationPolicy` of `WhenEmpty` or `WhenUnderutilized` may still want to dictate the speed at which nodes are deemed underutilized. This is particularly true on clusters that are large in size and have a large amount of churn. To support this, Karpenter will surface a `consolidationTTL` which will allow you to define a per-node TTL to define the time that Karpenter can begin deprovisioning the node after first seeing that the node is eligible for consolidation.
```
sftim, line 94: Depending on how much we like CEL, we could allow people to specify a _consolidation eligibility expression_.
(for early beta, that could be further constrained to say something like

“OK, but _for now_ the only supported expressions look like `node.secondsSinceEmpty >= *`, more later”

designs/v1beta1-full-changelist.md
```diff
@@ -0,0 +1,200 @@
...
+
+```
+amiSelector:
+- name: foo
+  id: abc-123
+  owner: amazon
+  ssm: "/al2/alias"
+  tags: 
+    key: value
+# Selectors are ORed
```
sftim, line 118: Not in Kubernetes they're not! I'd avoid calling this a selector.

designs/v1beta1-full-changelist.md
```diff
@@ -0,0 +1,200 @@
...
+    key: value
+```
+
+#### Remove `nodeClass.spec.launchTemplate`
+
+Direct launch template support is problematic for many reasons, outlined in the design [Unmanaged LaunchTemplate Support for Karpenter](./unmanaged-launch-template-removal.md). Customers continue to run into issues when directly using launch templates. Rather than continue to maintain these sharp edges and give users a half-baked experience of Karpenter, we should remove this field, considering that we can always add it back later if there is enough ask from users to do so.
+
+#### `nodeTemplate.spec.instanceProfile` → `nodeClass.spec.role`
+
+Currently, Karpenter uses an `instanceProfile` in the `AWSNodeTemplate` that is referenced to determine the profile that the EC2 node should launch with. Instance profiles are IAM entities that are specific to EC2 and do not have a lot of detail built around them (including console support); users are generally more familiar with the concept of IAM roles. As a result, we can support a `role` in the new `NodeClass` and allow Karpenter to provision the instance profile `ad-hoc` with the `role`  specified attached to it.
```
sftim, line 165: If I want to use an instance profile and _not_ allow Karpenter to manage that, can I still have that option?

designs/v1beta1-api.md
sftim, line 1: Not part of the RFC, but related: we (Karpenter) could provide some sample ValidatingAdmissionPolicies. 

Examples:
1. a policy  that ensures that all AWSNodeTemplates (or whatever we replace them with) are setting a required tag.
1. a policy that requires containerd as a runtime
1. a policy to require that nodes run Bottlerocket

designs/v1beta1-api.md
jonathan-innis: I'm curious why you say this isn't a selector. Agreed that isn't a label selector or a field selector but (from what I can tell), there is prior art with label selectors with `matchExpressions` that allows you to select on multiple ORed things.

designs/v1beta1-api.md
jonathan-innis: Commented above.

designs/v1beta1-api.md
jonathan-innis: Commented above.

designs/v1beta1-full-changelist.md
jonathan-innis: Yes, we would expect that as we get more CloudProviders we would add more API groups but each API group would have the name `NodeClass`. This allows us to version each group separate from each other and give each of the cloudproviders some autonomy.

designs/v1beta1-full-changelist.md
jonathan-innis: > we could allow people to specify a __consolidation eligibility expression__

I'm curious what you're thinking here. Like define their own custom selector logic based on creationTimestamp or the lastTransitionTime of a condition field?

designs/v1beta1-full-changelist.md
jonathan-innis: Good callout. Is this the reason that they are called `NodeSelectorTerms` in [`nodeAffinity`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#nodeselector-v1-core:~:text=NodeSelectorTerm%20array)? Maybe we should model this the same way with `amiSelectorTerms`?

designs/v1beta1-full-changelist.md
jonathan-innis: What would be the reason that you would want to use your own instance profile. I think we can add this back if we hear enough asks for it but there doesn't seem to be a strong enough reason to keep it when there is a 1-1 mapping between a role and an instance profile

designs/v1beta1-api.md
sftim: It's not a selector because Kubernetes selectors have certain qualities (eg: always a logical AND), that this doesn't align with.

If we call it a selector, people may assume it works like other selectors - but it may not.

designs/v1beta1-api.md
sftim: Pathological case is something like:
- `/apis/cloud-provider.example/v1beta1/providernodeclasses/foo`
- `/apis/cloud-provider.example/v1beta1/nodeclasses/foo`

both say that their kind is NodeClass (yes, I know, don't do this).

Using `providernodeclasses` rather than NodeClass would  avoid that risk.

designs/v1beta1-full-changelist.md
sftim: I'm half convinced, but I still prefer Machine.

designs/v1beta1-full-changelist.md
sftim: Annoying for people who want to `kubectl get NodeClasses` though. It won't work how most people would like, I imagine.

designs/v1beta1-full-changelist.md
sftim: Not a selector, but broadly yes. However this is mostly an aside / tangent.

designs/v1beta1-full-changelist.md
sftim: Even `nodeSelectorTerms` is an AND expression. I would use a word other than selector, especially as the backing API doesn't look (or work) like a Kubernetes one.

If we call it a selector, people who know a different mental model - ie, actual selectors - may assume it works just like those ones.
There's a related concern. Imagine that Kubernetes one day adds a logical-OR mechanism. Once we stabilise this API it's difficult to revise it, so the Karpenter “selectors“ might be stuck working similarly to actual selectors, but frustratingly not the same.

designs/v1beta1-full-changelist.md
sftim: I might want to use a role that some other code has made, along with an instance profile.

designs/v1beta1-api.md
ellistarn: These selectors are absolutely a logical AND. All selector terms are applied to filter down to a set of selected objects. For objects that can use multiple (e.g. security groups), we use them all. For objects that are singleton (e.g. sunset/ami) we pick one. 

This is no different than node selector requirements that use an AND to filter down to a known list, and then we pick one (e.g. instance type, zone).

designs/v1beta1-api.md
jonathan-innis: Huh. I didn't realize that you could create a Kind that had a different path than the actual kind name. Regardless if that's the case, I don't think that people should be doing this and (if someone is doing this) we shouldn't try to build support around our API to respect it.

sftim: SIG Auth has to think about the risks here. For Karpenter I think it's fine. Anyway, I wanted to make sure we made a decision with good context.

designs/v1beta1-api.md
sftim: OK, that's covered then.

designs/v1beta1-full-changelist.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+#### Update `nodeClass.spec.securityGroupSelector`
+
+The same logic for `subnetSelectorTerms` applies to `securityGroupSelectorTerms`. We should have a similar parity to the `amiSelectorTerms` to improve the ease-of-use around this selector.
+
+```
+securityGroupSelectorTerms:
+- id: abc-123
+  name: default-secuirty-group # Not the same as the name tag
```
sftim, line 149: (typo)

designs/v1beta1-full-changelist.md
```diff
@@ -0,0 +1,200 @@
...
+  tags:
+    key: value
+# Selectors are ORed
```
sftim, line 152: I reread how the actual [`DescribeSecurityGroups`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSecurityGroups.html) API works. You can specify multiple _values_ (eg, ask for two different tag values for the same key, and accept either) but the “key” part requirements are ANDed together.

designs/v1beta1-full-changelist.md
sftim: As a thought experiment, try expressing these requirements first using EC2's `Filter` API and then using a Kubernetes-style selector:

- AMI must **not** be public
- AMI must be for AMD64 CPU architecture
- AMI must have a tag set on it, `example.com/kubelet-minor-version`
- Volume type must be `gp2` or `gp3`

I think that's a reasonable filter and it's one you can specify using the underlying API. So, we should consider allowing it. But we can't make that API look enough like a Kubernetes selector (even the [set-based requirement](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement) kind), because there are different ergonomics here. The problem actually isn't the OR expression, now I think about it; it's what we set the keys to. Actual node selectors rely on automatic labelling, but I'm worried about whether it's feasible to map AWS tags to look like label keys.

---

Calling it a filter (and, to be clear, automating the migration for people using the older API) avoids people having to learn the subtle traps here.

More importantly, I think: a future change to the EC2 `DescribeImages` operation might make us want to expose more ways to filter images. The more that we try to adapt that API to look like Kubernetes, when it isn't really, the bigger the risk of future trouble - and problems that AWS' internal API change review process is unlikely to account for.

designs/v1beta1-full-changelist.md
ellistarn: Could be solved with shortnames? Agree that, it's a touch awkward to have overlap, but I suspect it to be pretty rare. I'd hate the lose the simplicity in naming to account for a theoretical use case being more awkward. 

Frankly, kubectl should support overlapping kinds and return both!

designs/v1beta1-full-changelist.md
jonathan-innis: Yep. Maybe what's confusing is how I worded the comment. What I _meant_ was that the selector terms are ORed but the actual selector itself is ANDed.

designs/v1beta1-api.md
jonathan-innis: I think there's some confusion here around what is getting ANDed and what is getting ORed. I've changed the name to `amiSelectorTerms` since I agree that this better aligns with `nodeSelectorTerms`. Each requirement in a given term will be ANDed together. For example if we have an `amiSelectorTerm`

```yaml
amiSelectorTerms:
- tags:
     key: value
   owner: amazon
```

This implies that we will look across all AMIs that have `amazon` as their owner AND also have the tag `key: value`. Each of these terms is ORed together to create the full list of available AMIs.

designs/v1beta1-full-changelist.md
jonathan-innis: Still not sure that I get the full use-case. It doesn't seem likely to me that you would have a hard requirement to utilize a specific instance profile that must be manually specified. To me, it seems like you would enforce this restriction at the role level.

designs/v1beta1-full-changelist.md
jonathan-innis: > Even `nodeSelectorTerms` is an AND expression

The [K8s documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#:~:text=NodeSelectorRequirement%20v1%20core) mentions that these terms are ORed

Screenshot 2023-07-30 at 10 35 36 PM

jonathan-innis: > I'm worried about whether it's feasible to map AWS tags to look like label keys

I think, generally, the user has complete control over these tags (if they are using tag-based selection) so ideally we shouldn't hit conflicts with label selection mechanisms in Kubernetes. I'm still leaning towards calling it a `selector` with terms since it's more K8s-native and the purpose of these API fields is a "selection"-based mechanism

sftim: I do tech writing (and cloud consultancy) and I try to avoid APIs where you end up having to write things like “**NOTE** This looks like a selector in Kubernetes but isn't the same, here are a list of differences that might catch you out”.

Given that experience sometimes I look upstream to see where hard-to-document things are popping up. That's all the context I can offer, anyway. I think this will be hard to document and I worry that there will be a steady stream of slightly confused people adopting this API.

designs/v1beta1-full-changelist.md
sftim: If the instance profile already exists, what behavior / error do I see?
If I stop using that role, does Karpenter try to delete the instance profile that is no longer needed?

designs/v1beta1-full-changelist.md
jonathan-innis: > This looks like a selector in Kubernetes but isn't the same, here are a list of differences that might catch you out

I'm slightly confused by this statement. I'm missing what exactly is different about this mechanism other than the service that it is interacting with. In all other ways, it should behave the same as selectors that interact directly with the K8s API server.

designs/v1beta1-full-changelist.md
jonathan-innis: Karpenter will surface the instance profile through a discovery mechanism and resolve this instance profile into the AWS `NodeClass` status. If the role changes, Karpenter should go and cleanup the old role assignment and update it with the new one. 

We should be generating the instanceProfile with a similar mechanism to `generateName` so that it is sufficiently random that we are unlikely to have collisions; however, if we do have a collision, we should just try to regenerate a random value again.

designs/v1beta1-api.md
ellistarn: @jonathan-innis set me straight. This is a list (ORed) of terms (ANDed), identical to node selector.

jonathan-innis: Correct.

designs/v1beta1-full-changelist.md
jonathan-innis: > kubectl should support overlapping kinds and return both

I feel like this starts to get complex when it comes to duplicate kind names from different API groups. Like, if you have two different nodeClasses from two different API groups with the same name, what happens when you do `kubectl edit nodeclass default`. It might start to get confusing to the user, at which point, a user just has to resort to specify the scoped resource name giving the full API group with `kubectl edit nodeclasses.compute.k8s.aws default` or some other cloudprovider to get the clarity they are desiring. 

I do doubt that we will have a lot of cross-cloud clusters using Karpenter (at least on the outset or maybe into the future) but the prospect of these naming conflicts does scare me a bit since they don't appear to be well-supported in K8s and fully based on the discovery order of the client.

jonathan-innis: @sftim Did you have a recommendation on what to do here? I know projects like CAPI have resorted to just prefixing their kinds with the cloudprovider name like `AWSNodeClass`. Though I find that naming a little annoying as well.

sftim: > AWSNodeClass

That's what I'd do; I think it'd match end user expectations.

designs/v1beta1-full-changelist.md
```diff
@@ -0,0 +1,200 @@
...
+
+### `karpenter.sh/do-not-evict` →  `karpenter.sh/do-not-disrupt`
+
+Karpenter validates disruption across NodeClaims and determines which NodeClaims/Nodes it is allowed to disrupt as part of the deprovisioning flow. While eviction is part of the termination process, it’s more accurate to say that the `karpenter.sh/do-not-evict` annotation actually prevents Karpenter’s disruption of the NodeClaim/Node rather than the eviction of it.
+
+###  `karpenter.sh/do-not-consolidate`  → `karpenter.sh/do-not-disrupt`
+
+Karpenter currently surfaces the `karpenter.sh/do-not-consolidate` annotation to block consolidation actions against individual nodes without having to make changes to the owning provisioner. We have found this is useful for users that have one-off scenarios for blocking consolidation, including debugging failures on nodes. 
+
+While this feature is useful for consolidation, it should be expanded out to all disruption mechanisms, so that we have both pod-level and node-level control to block disruption using the `karpenter.sh/do-not-disrupt` annotation.
```
garvinp-stripe, line 42: Is it possible to add this annotation also at the Provisioner/ NodePool level? This would allow an entire pool of nodes to be temporarily frozen for certain operation.

designs/v1beta1-full-changelist.md
garvinp-stripe: Expanding on this, it would be cool if eventually deprovisioning for all 5 types of deprovisioning can be controlled. Such as turning on and off each type of deprovisioning and the TTL.

designs/v1beta1-full-changelist.md
njtran: Hey @garvinp-stripe, there's a separate effort to include time-based controls that come with a `maxDisruptions` type field. The end state in conjunction with this proposal is that each TTL field can be enabled/disabled so that users can define which types of deprovisioning they want for a Provisioner/NodePool, and then in the same Provisioner/NodePool, users can define a schedule that determine how many nodes can be deprovisioned at a time, irrespective of the method. You would be able to set `maxDisruptions:0` to effectively freeze the Provisioner/NodePool.

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+        - key: karpenter.k8s.aws/instance-gpu-count
+          operator: DoesNotExist
+        - key: karpenter.k8s.aws/instance-accelerator-count
+          operator: DoesNotExist
+```
+
+### `karpenter.sh/NodeClaim`
+
+```
+apiVersion: karpenter.sh/v1beta1
```
rothgar, line 229: Why is NodePool and NodeClaim `karpenter.sh/v1beta1` but NodeClass is `compute.k8s.aws/v1beta1`?

jonathan-innis: We don't want to be managing the CloudProvider-specific CRDs in our top-level API Group, since according to upstream guidance, this means that they would have to be versioned and managed together. 

Guidance from upstream and also from @sftim was that these should be split out into their own cloudprovider-specific API groups so that each cloudprovider is free to manage the CRDs and version them as they see fit.

rothgar: Is there a way for users to validate the support matrix of karpenter, NodeClass, and their cloud provider's APIs?

sftim: > Is there a way for users to validate the support matrix of karpenter, NodeClass, and their cloud provider's APIs?

In general? It would depend on the provider. The design here anticipates having more than once cloud provider integrations.

(Personally, I would call the AWS node class kind an AWSNodeClass, in case someone is running a hybrid cluster with eg a mix of AWSNodeClass and ExampleBareMetalNode objects)

ellistarn: One option I discussed with Jonathan is to call it EC2NodeClass but make a shortname for nodeclass. Alternatively, you could make the CRD NodeClass, and make the shortname ec2nodeclass. 

Ideally, we could just let kubectl support multiple objects with the same name but different group. Would be useful for Knative Services, etc.

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+```
+
+### `compute.k8s.aws/NodeClass`
+
+```
+apiVersion: compute.k8s.aws/v1beta1
+kind: NodeClass
+metadata:
+  name: default
+spec:
```
tallaxes, line 130: Have you considered using `spec.template` for configuration-based fields, same as `NodePool`?

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+
+## APIs
+
+To help clearly define where configuration should live within Karpenter’s API, we should define the logical boundary between each Kind in the project.
+
+1. `NodePool`
+    1. Neutral Node configuration-based fields that affect the **compatibility between Nodes and Pods during scheduling** (e.g. requirements, taints, labels)
+    2. Neutral behavior-based fields for configuring Karpenter’s scheduling and deprovisioning decision-making
+2. `NodeClass`
+    1. Cloudprovider-specific Node configuration-based fields that affect launch and bootstrap process for that Node including: configuring startup scripts (including kubelet configuration), volume mappings, metadata settings, etc.
```
tallaxes, line 40: "Including kubelet configuration" may be a little confusing here, given the example that follows clearly shows `kubeletConfiguration` field in `NodePool`

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+        - key: example.com/another-taint
+          effect: NoSchedule
+      requirements:
+        - key: "karpenter.k8s.aws/instance-category"
+          operator: In
+          values: ["c", "m", "r"]
+      resources:
+        requests:
+          cpu: "1"
+          memory: "100Mi"
```
tallaxes, line 77: What would resource requests mean here?

designs/v1beta1-api.md
sftim: > Ideally, we could just let kubectl support multiple objects with the same name but different group. Would be useful for Knative Services, etc.

Sure, but: do we want to hold off v1beta1 until that's done? I think not.

I like the idea of `ec2nodeclass` and also `ec2nc` as a short name. The Helm chart could plausibly let you decide on whether to make just `nodeclass` also be a short name (defaults to off; AWS docs would encourage flipping that to on).

designs/v1beta1-full-changelist.md
```diff
@@ -0,0 +1,200 @@
...
+
+#### Remove `nodePool.spec.provider`
+
+We’ve recommended that customers leverage spec.providerRef in favor of spec.provider since Q2 2022. Documentation for this feature has been removed since Q3 2022. We will take the opportunity to remove the feature entirely to minimize code bugs/complexity and user confusion.
+
+### `NodeClass` Changes
+
+#### Update `nodeClass.spec.amiSelector`
+
+The current `amiSelector` has two primary limitations that restrict user’s ability to specify the AMIs that they want Karpenter to use:
```
sftim, line 104:
```suggestion
The alpha API `amiSelector` has two primary limitations that restrict user’s ability to specify the AMIs that they want Karpenter to use:
```

designs/v1beta1-full-changelist.md
```diff
@@ -0,0 +1,200 @@
...
+securityGroupSelectorTerms:
+- id: abc-123
+  name: default-security-group # Not the same as the name tag
+  tags:
+    key: value
+# Selector Terms are ORed
+- id: abc-123
+  name: custom-security-group # Not the same as the name tag
+  tags:
+    key: value
```
sftim, line 156: Could do this:
```suggestion
- id: abc-123
  tags:
    use: private-subnet
# Selector Terms are ORed
- name: custom-security-group-b # Not the same as the Name tag
  tags:
    use: private-subnet
- tags:
    use: private-subnet
    Name: custom-security-group-c # not the same as the name field
```

designs/v1beta1-full-changelist.md
sftim: OK, cool. There need not be a 1-to-1 mapping though. _n_ instance profiles can map to 1 role (right?)

Given that Karpenter is going to manage instance profiles, I might want:
- to specify a custom `path` for the instance profile
- to set custom tags on the instance profile

More likely: I restrict the IAM role that Karpenter runs as. The restrictions enforce tag-on-create for instance profiles, and expect certain tags to be set. Karpenter is then only allowed to delete or update instance profiles that have the appropriate tag.

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+
+Unfortunately, multiple proposed changes in v1beta1 are not round-trippable. Below, we propose deprecations of legacy fields in favor more modern mechanisms that have seen adoption in v1alpha5. These changes remove sharp edges that regularly cause users surprises and production pain.
+
+To workaround the limitation of round-trippability, we are proposing a rename of both the API group and Kind (`NodePool`) that the CRDs exist within. This allows both CRDs to exist alongside each other simultaneously and gives users a natural migration path to move through.
+
+### Migration Path
```
sftim, line 19: Any plan to provide a tool to help people migrate to NodePool?

designs/v1beta1-full-changelist.md
garvinp-stripe: Make sense, thanks

designs/v1beta1-api.md
sftim: Anyway, +1 for EC2NodeClass.

garvinp-stripe:
Great work! I had some time to think, this isn't necessarily a strong opinion but I think there is some value in considering. I don't know what the timeline looks like and maybe what I am thinking doesn't make sense at v1beta1 however I think there is an opportunity to drive some usability improvement since there is going to be significant work in changing the custom resources. 

I wanted to see we can take some inspiration from https://gateway-api.sigs.k8s.io/ which is somewhat new and have been through quite a bit of changes in its overall design. Specifically the way it carved out custom resources that is able to abstract discrete configuration depending on who is using it. I think this way of thinking can yield quite a bit of insight into how to structure custom resources so that Cluster operators can easily provide application developers the appropriate level of control.

Starting with `NodePool`. I think it has a lot of abstraction leaks between App dev vs Cluster Operator. Ideally App devs can specify their pods -> node, maybe some taints and label for their nodes (pretty much all the parameters in spec). However App dev shouldn't be able to specify kubelet configurations since this is usually managed by cluster operator. I think if possible this can be carved out to something like `NodeProvider` which are associated to `NodePool`. Similar to `HTTPRoutes` to `Gateway` in GatewayAPI. I think this would significantly improve the flexibility of how Cluster operators manage Provisioners/ NodePool. 

I don't think there is much suggestion for `NodeClass` but I think something nice would be "version" similar to LaunchTemplate version. This allows Cluster Operators to rollout and back nodes with specific changes safely and have the necessary controls when managing many NodeClasses. Ideally each NodeClasses version should be immutable. 

Thanks

designs/v1beta1-api.md
jonathan-innis: AFAIK, you can't template CRDs in helm charts so I think it would have to be on by default.

designs/v1beta1-api.md
jonathan-innis: We liked using `spec.template` for `NodePool` since it works very similarly to a Deployment's `spec.template` where it templates another Kubernetes object. In the case of deployment, the deployment templates pods. In the case of `NodePool`, it templates `NodeClaims`.

I'm not sure if `NodeClass` works the same way since it's mainly providing the interface through which the cloudprovider can understand the settings for how it launches nodes; it's not templating out another Kubernetes object. WDYT>

designs/v1beta1-api.md
jonathan-innis: These requests are used by the NodeClaim to understand the minimum set of pod requests that are required for the node to be launched. By default, we aren't going to allow this value to be set in the `NodePool` (but the field will still be available).

I could imagine that this field could be available in the future for the top-level `NodePool` to be a secondary mechanism to define the minimum set of requests required by a NodeClaim that Karpenter launches from that pool similar to the current `Gt`/`Lt` operators.

designs/v1beta1-api.md
jonathan-innis: Yeah, I think this is reasonable and we should definitely support some kind of migration tool.

jonathan-innis:
> I think something nice would be "version" similar to LaunchTemplate version. This allows Cluster Operators to rollout and rollback changes to nodes safely and have the necessary controls when managing many NodeClasses. Ideally each NodeClasses version should be immutable. Right now it is difficult to expand on `NodeClass` for nodes outside of AWS. It may be valuable to split NodeClasses, I don't have a good name, such that we have something that is at the cloud provider level and one that just defines nodes specification specific to a cloud provider. (Similar to gatewayclass and gateway in Gateway API). The Cloud provider layer resource defines the controller for creating nodes for that cloud provider and the nodeclass themselves work in the context of each cloud provider.

I'm still thinking on the other comment that you made around the Gateway API, but, for this one, the idea was that the `NodeClass` _is_ the cloudprovider extensibility mechanism; it's effectively the interface to define how Karpenter can actually bring up a Node in a given Cloudprovider. If another Cloudprovider came along and wanted to support Karpenter, I would expect that they would create their own NodeClass and this NodeClass would define its own interface with how to launch Nodes in that Cloudprovider.

Generally, we wanted to avoid creating too many disparate concepts that have to be individually reasoned about (i.e. more API kinds). We effectively have taken the stance that if something is common enough that it should be useful to _all_ cloudproviders, then it should be hoisted up into the `NodePool`; otherwise, it belongs within the cloudprovider-specific `NodeClass`.

Is there a specific use-case that you have in mind where you think splitting out the details of the `NodeClass` could be useful beyond hoisting common cross-cloudprovider values into the `NodePool`?

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+  resources:
+     cpu: "2"
+     memory: "100Mi"
+     ephemeral-storage: "100Gi"
+```
+
+### `compute.k8s.aws/NodeClass`
+
+```
+apiVersion: compute.k8s.aws/v1beta1
```
garvinp-stripe, line 126: > > I think something nice would be "version" similar to LaunchTemplate version. This allows Cluster Operators to rollout and rollback changes to nodes safely and have the necessary controls when managing many NodeClasses. Ideally each NodeClasses version should be immutable. Right now it is difficult to expand on `NodeClass` for nodes outside of AWS. It may be valuable to split NodeClasses, I don't have a good name, such that we have something that is at the cloud provider level and one that just defines nodes specification specific to a cloud provider. (Similar to gatewayclass and gateway in Gateway API). The Cloud provider layer resource defines the controller for creating nodes for that cloud provider and the nodeclass themselves work in the context of each cloud provider.
> 
> I'm still thinking on the other comment that you made around the Gateway API, but, for this one, the idea was that the `NodeClass` _is_ the cloudprovider extensibility mechanism; it's effectively the interface to define how Karpenter can actually bring up a Node in a given Cloudprovider. If another Cloudprovider came along and wanted to support Karpenter, I would expect that they would create their own NodeClass and this NodeClass would define its own interface with how to launch Nodes in that Cloudprovider.
> 
> Generally, we wanted to avoid creating too many disparate concepts that have to be individually reasoned about (i.e. more API kinds). We effectively have taken the stance that if something is common enough that it should be useful to _all_ cloudproviders, then it should be hoisted up into the `NodePool`; otherwise, it belongs within the cloudprovider-specific `NodeClass`.
> 
> Is there a specific use-case that you have in mind where you think splitting out the details of the `NodeClass` could be useful beyond hoisting common cross-cloudprovider values into the `NodePool`?

Sorry moving this to here instead because github threading is kinda bad on main comment. 

Ah Ok I think i am getting the picture more now thanks for the explanation. I think I agree with you NodeClass is a good interface and I think likely don't need to split Nodeclasses out. 

I wonder if there is value as part of this to separate out Karpenter's AWS NodeClass controller out from the main project and reorganize Karpenter vs Karpenter core. Karpenter-core feels more like a library atm. I think there was part of my confusion in understanding how NodeClass may work since right now AWS' pieces feels so integrated with Karpenter as a whole.

designs/v1beta1-api.md
```diff
@@ -0,0 +1,341 @@
...
+
+1. Spin up a new `v1beta1/NodePool` and `v1beta1/NodeClass` resource matching their `v1alpha5` counterparts
+2. Set the `spec.limits` on your `v1alpha5/Provisioner` resource to `cpu=0` to stop provisioning
+3. Manually delete Nodes one-by-one to have Karpenter roll the nodes over to `v1beta1`
+
+## APIs
+
+To help clearly define where configuration should live within Karpenter’s API, we should define the logical boundary between each Kind in the project.
+
+1. `NodePool`
```
garvinp-stripe, line 36: > Starting with NodePool. I think it has a lot of abstraction leaks between App dev vs Cluster Operator. Ideally app devs can specify their pods -> node, maybe some taints and label for their nodes (pretty much all the parameters in spec). However app dev shouldn't be able to specify kubelet configurations since this is usually managed by cluster operator. I think if possible this can be carved out to something like NodeProviders which are associated to NodePool. Similar to HTTPRoutes to Gateway in GatewayAPI. I think this would significantly improve the flexibility of how cluster operators manage Provisioners/ NodePool since often times there are much more app dev teams to cluster operator teams and most cluster operators are currently creating provisioners per app team.

To further expand on this thought, I think there should be a separation between App Dev configuration and cluster operator. 
- An app dev entity that defines Node <-> pod compatibility on requirements, taints, label.
- A cluster operator entity that defines cloud-provider agnostic configuration for nodes (kubelet config), scheduling and deprovisioning decision making. Since these usually have deeper company wide polices such as SOX compliance, node lifetime policies, daemonset/ OS overhead, networking setup and such. Ideally an app dev doesn't even have to know this is "kubernetes"

sftim: Here's how I'd solve that, **if needed**:
- don't put any behavior into Karpenter itself
- provide a contrib controller (or a configuration for `shell-operator`, or whatever) that allows someone to manage a subset of the settings for a NodePool

_or_

- a whole new controller that manages scheduling behavior and node outcomes, per namespace. Again, not Karpenter.

designs/v1beta1-api.md
jonathan-innis: > I wonder if there is value as part of this to separate out Karpenter's AWS NodeClass controller out from the main project and reorganize Karpenter vs Karpenter core

It's funny because we've thought about this at a few different points in time and I think in a lot of ways we are really close to being able to do this right now. I think it really depends on the value that we get from separating it out like this. When there are more cloudproviders in place, I could see karpenter-core being a "common controller" that handles machine scheduling and hands off Machine launch and deprovisioning process off to the Cloudprovider through Create/Delete on the API server.

That pretty much leaves InstanceTypes as the final piece to separate out into a separate controller, which we have also thought about at different points in time could be communicated through a `compute.k8s.aws/InstanceType` CRD. 

All of this is of course ignoring the fact that you could just write the CloudProvider interface to be gRPC instead of library/in-code without even re-architecting anything

designs/v1beta1-api.md
jonathan-innis: To add to what @sftim said, there's tools like [Backstage](https://backstage.io/docs/features/kubernetes/) that effectively allow cluster admins to "wrap" Kubernetes with a simpler interface that they can hand off to their app teams. It's effectively templating the manifests but in a managed way. That might be a good solution here if there's a need to draw a line between what service teams should be owning and what cluster admins should be owning.

designs/v1beta1-api.md
sftim: I'm still a bit dubious about

```
amiSelectorTerms:
- tags:
     key: value
   owner: amazon
```
but since this isn't Karpenter core, I really don't think it should block the beta.

designs/v1beta1-api.md
```diff
@@ -0,0 +1,340 @@
...
+  subnetSelectorTerms:
+    - tags:
+        compute.k8s.aws/discovery: cluster-name
+    - id: subnet-1234
+  securityGroupSelectorTerms:
+    - tags:
+        compute.k8s.aws/discovery: cluster-name
+    - name: default-security-group
+  role: karpenter-node-role
+  userData: |
```
sftim, line 147: :thought_balloon: Can we support a case where the user data at API level is 16 KiB of raster image data?

If that's important, we could allow:
```yaml
userData:
  textValue: |
    echo "this is custom user data"
```
and
```yaml
userData:
  data: ZWNobyAidGhpcyBpcyBjdXN0b20gdXNlciBkYXRhIgo=
```

designs/v1beta1-api.md
sftim: Ah yep.

designs/v1beta1-api.md
garvinp-stripe: interesting, let me look into that

garvinp-stripe: Out of curiosity, is there a reason why not to split it at the Karpenter level? Is this just adding too much to the API surface?

jonathan-innis: Yeah, I think it's mainly trying to address too many concerns in one place. IMO, these should be layers of abstraction that users that need it can add as they please.

ellistarn (approved):

designs/v1beta1-api.md
jonathan-innis: > you can't template CRDs in helm charts so I think it would have to be on by default

Because of the concern for conflicts right now, we're going to punt on short-naming the `EC2NodeClass` with `nodeclass`. We can always support this down-the-line and it might be nice for `kubectl`/apiserver to support multiple objects with the same shortname and to print them in a way that is friendly for that.

designs/v1beta1-full-changelist.md
```diff
@@ -4,22 +4,19 @@ This document formalizes the [v1beta1 laundry list](https://github.com/aws/karpe
 
 ### Update Kind/Group Naming
 
-As part of the bump to v1beta1, to allow the v1alpha5 APIs to exist alongside the v1beta1 APIs while users go through a migration process, the following renames are being proposed:
+As part of the bump to v1beta1, to allow the v1alpha5 APIs to exist alongside the v1beta1 APIs while users go through a migration process, the following kind names are being proposed:
 
-1. API Group Renames
-    1. `karpenter.k8s.aws` → `compute.k8s.aws`
```
tallaxes: @jonathan-innis, I am curios what prompted the reversal of the original plan?

designs/v1beta1-api.md
```diff
@@ -317,23 +334,22 @@ status:
...
+8. `karpenter.k8s.aws/instance-cpu`
+9. `karpenter.k8s.aws/instance-memory`
+10. `karpenter.k8s.aws/instance-network-bandwidth`
+11. `karpenter.k8s.aws/instance-gpu-name`
+12. `karpenter.k8s.aws/instance-gpu-manufacturer`
+13. `karpenter.k8s.aws/instance-gpu-count`
+14. `karpenter.k8s.aws/instance-gpu-memory`
+15. `karpenter.k8s.aws/instance-accelerator-name`
+16. `karpenter.k8s.aws/instance-accelerator-manufacturer`
+17. `karpenter.k8s.aws/instance-accelerator-count`
```
tallaxes, line 355: @jonathan-innis, I am curios what prompted the reversal of the original plan?

PR #4300, opened 2023-07-19
feat: Add Windows custom userdata support

tzifudzi:
Fixes https://github.com/aws/karpenter/issues/4081

**Description**
This PR adds support for the specifying of XML custom user data for Windows on AWS EKS.

- As with user data for other platforms, any user data specified will be merged with the Karpenter managed user data, such that the custom user data runs before the EKS bootstrap script. 
- If merging of user data fails, code logic will fall back to use the Karpenter managed user data so bootstrapping will not be blocked.

- Added 3 unit tests in `pkg/providers/amifamily/bootstrap/windows_test.go`. All unit tests are passing.
- Added 3 integration tests in `pkg/providers/launchtemplate/suite_test.go`. All integration tests are passing.


tzifudzi:
@rawahars Please join the review for this PR

rawahars (commented):
The PR looks really good. Thanks Tatenda!

My only concern is about the fallback mechanism on Karpenter managed userdata, if the customer provided user data fails. Consider a scenario where a customer has alarms configured based on the failed nodes or insufficient capacity. If we go ahead with fallback mechanism then such a customer will see that the capacity is there but the actual workload will eventually fail. Or worse, it can be intermittent issue where few pods fail.

I am more in favour of explicit erroring out in case of failed userdata.

tzifudzi:
> My only concern is about the fallback mechanism on Karpenter managed userdata...I am more in favour of explicit erroring out in case of failed userdata.

@rawahars Good point, explicitly failing might be a better way to bring attention to the fact that the custom user data did not bootstrap correctly. Will make the suggested change.

examples/provisioner/launchtemplates/windows-custom-userdata.yaml
```diff
@@ -0,0 +1,38 @@
...
+    karpenter.sh/discovery: "${CLUSTER_NAME}" # replace with your cluster name
+  securityGroupSelector:
+    karpenter.sh/discovery: "${CLUSTER_NAME}" # replace with your cluster name
+  amiFamily: Windows2022
+  metadataOptions:
+    httpProtocolIPv6: disabled
+    httpTokens: required
+  userData: |
+    <powershell>
+    New-Item -Path 'C:\temp\' -ItemType Directory
```
jonathan-innis, line 36: I'm wondering if we should show an example that allows for updating kubeletConfiguration if that's possible on the Windows AMI as well?

pkg/providers/amifamily/bootstrap/windows.go
```diff
@@ -17,16 +17,39 @@ package bootstrap
...
 	"strings"
+
+	"github.com/samber/lo"
 )
 
 type Windows struct {
 	Options
 }
 
+type UserDataXML struct {
```
jonathan-innis, line 32: Just so I understand, these are all of the possible fields that can be specified in the Windows UserData XML?

pkg/providers/amifamily/bootstrap/windows.go
```diff
@@ -44,5 +67,50 @@ func (w Windows) Script() (string, error) {
...
+		}
+	}
+
+	s, err := xml.Marshal(mergedUserDataXML)
+	if err != nil {
+		return "", err
+	}
+
+	// Remove enclosing <userdata></userdata> tags
+	userdata := strings.TrimPrefix(string(s), "<userdata>")
```
jonathan-innis, line 113: Why do we add and then remove these `<userdata>` tags?

pkg/providers/launchtemplate/suite_test.go
```diff
@@ -1696,6 +1696,63 @@ var _ = Describe("LaunchTemplates", func() {
...
 		})
+		Context("Windows Custom UserData", func() {
+			BeforeEach(func() {
+				ctx = settings.ToContext(ctx, test.Settings(test.SettingOptions{
+					EnableENILimitedPodDensity: lo.ToPtr(false),
+				}))
+				provisioner.Spec.Requirements = []v1.NodeSelectorRequirement{{Key: v1.LabelOSStable, Operator: v1.NodeSelectorOpIn, Values: []string{"windows"}}}
+				nodeTemplate.Spec.AMIFamily = &v1alpha1.AMIFamilyWindows2022
+			})
+			It("should merge and bootstrap with custom user data if valid XML user data", func() {
```
jonathan-innis, line 1707: Can we test all the XML parameters that are available with the merging of the userData?

website/content/en/docs/concepts/node-templates.md
```diff
@@ -445,10 +445,6 @@ spec:
 
 For more examples on configuring these fields for different AMI families, see the [examples here](https://github.com/aws/karpenter/blob/main/examples/provisioner/launchtemplates).
 
-{{% alert title="Windows Support Notice" color="warning" %}}
```
jonathan-innis, line 448: Docs points to the latest stable version so we should just update the `/preview` for now

website/content/en/v0.29/concepts/node-templates.md
```diff
@@ -445,10 +445,6 @@ spec:
 
 For more examples on configuring these fields for different AMI families, see the [examples here](https://github.com/aws/karpenter/blob/main/examples/provisioner/launchtemplates).
 
-{{% alert title="Windows Support Notice" color="warning" %}}
```
jonathan-innis, line 448: Likewise, this merge will lead to a feature minor version bump so we shouldn't be updating old versions of the docs with this change.

pkg/providers/launchtemplate/suite_test.go
jonathan-innis: From what I can tell we can be testing the `Script`, `Detach`, and `Persist` tags

examples/provisioner/launchtemplates/windows-custom-userdata.yaml
tzifudzi: I agree its a good idea to have a more practical use case specified as an example such as updating the kubelet configuration. Unfortunately, at the moment, the bootstrap only supports arguments being passed directly to it as the script is invoked e.g. `.\Start-EKSBootstrap.ps1 -KubeletExtraArgs "--max-pods=110"`.

pkg/providers/amifamily/bootstrap/windows.go
tzifudzi: With the exception of one more arg `runAsLocalSystem`, the list is exhaustive yes. I will add the remaining field `runAsLocalSystem`.

pkg/providers/amifamily/bootstrap/windows.go
tzifudzi: The reason of wrapping with the userdata string tags is that XML documents [must contain one root element](https://www.w3.org/TR/xml/#:~:text=There%20is%20exactly%20one%20element%2C%20called%20the%20root). As such the `encoding/xml` function `xml.Unmarshal` requires a root element to properly map the XML string data into a struct. The workaround was therefore this rather hacky way you see here.

Other options using a non-standard XML parsing library which is more flexible to parse XML strings without a root.  Another option might be to not map the string into a struct and manipulate the strings directly. Both solutions weren't preferable but happy to discuss options.

pkg/providers/launchtemplate/suite_test.go
tzifudzi: Removed the flexibility to specify EC2Launch XML elements. In the future if the need arises we can implement further changes to support this.

pkg/providers/amifamily/bootstrap/windows.go
tzifudzi: Removed the flexibility to specify EC2Launch XML elements. In the future if the need arises we can implement further changes to support this.

.golangci.yaml
```diff
@@ -57,7 +57,7 @@ linters-settings:
       limitations under the License.
 issues:
   fix: true
-  exclude: ['declaration of "(err|ctx)" shadows declaration at']
+  exclude: ['declaration of "(err|ctx)" shadows declaration at', 'getInstanceAttribute` - `attribute` always receives `"userData"']
```
tzifudzi, line 60: CI tests are failing because there is a method which has userdata as the arg at all time. We can either exclude the lint check or otherwise refactor the method so it does not always receive the same argument.

pkg/providers/amifamily/bootstrap/windows.go
tzifudzi: Removed the flexibility to specify EC2Launch XML elements. In the future if the need arises we can implement further changes to support this.

jonathan-innis (dismissed):
LGTM 🚀

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (approved):
LGTM 🚀

jonathan-innis (commented):
/karpenter snapshot

PR #4308, opened 2023-07-20
chore: Adding AWSNodeTemplate Hash Annotation

engedaam:
Fixes #N/A 

**Description**
- Karpenter will create a provisioner hash of the fields that the team considers to be statically drifted
- The will contain the values of each of those fields

- `make battletest`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/apis/v1alpha1/awsnodetemplate.go
+++ b/pkg/apis/v1alpha1/awsnodetemplate.go
@@ -15,6 +15,9 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"fmt"
+
+	"github.com/mitchellh/hashstructure/v2"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -79,7 +82,7 @@ type AWSNodeTemplateSpec struct {
 	AWS      `json:",inline"`
 	// AMISelector discovers AMIs to be used by Amazon EC2 tags.
 	// +optional
-	AMISelector map[string]string `json:"amiSelector,omitempty"`
+	AMISelector map[string]string `json:"amiSelector,omitempty" hash:"ignore"`
 	// DetailedMonitoring controls if detailed monitoring is enabled for instances that are launched
 	// +optional
 	DetailedMonitoring *bool `json:"detailedMonitoring,omitempty"`
@@ -95,6 +98,16 @@ type AWSNodeTemplate struct {
 
 	Spec   AWSNodeTemplateSpec   `json:"spec,omitempty"`
 	Status AWSNodeTemplateStatus `json:"status,omitempty"`
+}
+
+func (a *AWSNodeTemplate) Hash() string {
+	hash, _ := hashstructure.Hash(a.Spec, hashstructure.FormatV2, &hashstructure.HashOptions{
+		SlicesAsSets:    true,
+		IgnoreZeroValue: true,
+		ZeroNil:         true,
+	})
+
+	return fmt.Sprint(hash)
 }
 
 // AWSNodeTemplateList contains a list of AWSNodeTemplate
--- a/pkg/apis/v1alpha1/provider.go
+++ b/pkg/apis/v1alpha1/provider.go
@@ -24,7 +24,7 @@ import (
 type AWS struct {
 	// TypeMeta includes version and kind of the extensions, inferred if not provided.
 	// +optional
-	metav1.TypeMeta `json:",inline"`
+	metav1.TypeMeta `json:",inline" hash:"ignore"`
 	// AMIFamily is the AMI family that instances use.
 	// +optional
 	AMIFamily *string `json:"amiFamily,omitempty"`
@@ -37,10 +37,10 @@ type AWS struct {
 	InstanceProfile *string `json:"instanceProfile,omitempty"`
 	// SubnetSelector discovers subnets by tags. A value of "" is a wildcard.
 	// +optional
-	SubnetSelector map[string]string `json:"subnetSelector,omitempty"`
+	SubnetSelector map[string]string `json:"subnetSelector,omitempty" hash:"ignore"`
 	// SecurityGroups specify the names of the security groups.
 	// +optional
-	SecurityGroupSelector map[string]string `json:"securityGroupSelector,omitempty"`
+	SecurityGroupSelector map[string]string `json:"securityGroupSelector,omitempty" hash:"ignore"`
 	// Tags to be applied on ec2 resources like instances and launch templates.
 	// +optional
 	Tags map[string]string `json:"tags,omitempty"`
@@ -53,7 +53,7 @@ type LaunchTemplate struct {
 	// NOTE: This field is for specifying a custom launch template and is exposed in the Spec
 	// as `launchTemplate` for backwards compatibility.
 	// +optional
-	LaunchTemplateName *string `json:"launchTemplate,omitempty"`
+	LaunchTemplateName *string `json:"launchTemplate,omitempty" hash:"ignore"`
 	// MetadataOptions for the generated launch template of provisioned nodes.
 	//
 	// This specifies the exposure of the Instance Metadata Service to
--- a/pkg/apis/v1alpha1/register.go
+++ b/pkg/apis/v1alpha1/register.go
@@ -107,6 +107,7 @@ var (
 	LabelInstanceAcceleratorName              = LabelDomain + "/instance-accelerator-name"
 	LabelInstanceAcceleratorManufacturer      = LabelDomain + "/instance-accelerator-manufacturer"
 	LabelInstanceAcceleratorCount             = LabelDomain + "/instance-accelerator-count"
+	AnnotationNodeTemplateHash                = LabelDomain + "/nodetemplate-hash"
 )
 
 var (
--- a/pkg/apis/v1alpha1/suite_test.go
+++ b/pkg/apis/v1alpha1/suite_test.go
@@ -16,10 +16,12 @@ package v1alpha1_test
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"testing"
 
 	"github.com/Pallinder/go-randomdata"
+	"github.com/mitchellh/hashstructure/v2"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	. "knative.dev/pkg/logging/testing"
@@ -27,7 +29,10 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/aws/aws-sdk-go/aws"
+
 	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+	"github.com/aws/karpenter/pkg/test"
 )
 
 var ctx context.Context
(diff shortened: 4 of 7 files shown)


pkg/apis/v1alpha1/register.go
```diff
@@ -107,6 +107,7 @@ var (
 	LabelInstanceAcceleratorName              = LabelDomain + "/instance-accelerator-name"
 	LabelInstanceAcceleratorManufacturer      = LabelDomain + "/instance-accelerator-manufacturer"
 	LabelInstanceAcceleratorCount             = LabelDomain + "/instance-accelerator-count"
+	AnnotationAWSNodeTemplateHash             = LabelDomain + "/awsnodetemplate-hash"
```
njtran, line 110: can we call it providerRefHash or something similar? Ideally the machine should have this hash, but the label value should be expandable to other provider implementations.
```suggestion
	AnnotationAWSNodeTemplateHash             = LabelDomain + "/awsnodetemplate-hash"
```

pkg/apis/v1alpha1/suite_test.go
```diff
@@ -115,4 +120,96 @@ var _ = Describe("Validation", func() {
 			Expect(ant.Validate(ctx)).To(Not(Succeed()))
 		})
 	})
+	var _ = Describe("AWSNodeTemplate Annotation", func() {
```
njtran, line 123: nit: I'd name it Hash
```suggestion
	var _ = Describe("AWSNodeTemplate Hash", func() {
```

pkg/apis/v1alpha1/provider.go
njtran, line 1: I thought we weren't going to be drifting on launch templates. It's not included in website docs

pkg/apis/v1alpha1/suite_test.go
```diff
@@ -115,4 +120,96 @@ var _ = Describe("Validation", func() {
...
+		})
+		It("should expect two provisioner with the same spec to have the same provisioner hash", func() {
+			awsnodetemplateTwo := &v1alpha1.AWSNodeTemplate{
+				ObjectMeta: metav1.ObjectMeta{Name: strings.ToLower(randomdata.SillyName())},
+			}
+			awsnodetemplateTwo.Spec = awsnodetemplatespec
+
+			Expect(awsnodetemplate.Hash()).To(Equal(awsnodetemplateTwo.Hash()))
+		})
+		It("should expect hashes that are reordered to not produce a new hash", func() {
```
njtran, line 200:
```suggestion
		It("should expect fields that are reordered to not produce a new hash", func() {
```

pkg/apis/v1alpha1/provider.go
engedaam: Yes, we should not be drifting on `LaunchTemplateName`

pkg/apis/v1alpha1/suite_test.go
```diff
@@ -115,4 +120,96 @@ var _ = Describe("Validation", func() {
...
+				},
+				UserData:           aws.String("userdata-test-1"),
+				DetailedMonitoring: aws.Bool(false),
+			}
+			awsnodetemplate = test.AWSNodeTemplate(awsnodetemplatespec)
+		})
+		It("should change hash when static fields are updated", func() {
+			expectedHash := awsnodetemplate.Hash()
+
+			// Change one static field for 5 provisioners
```
njtran, line 156: nit: technically 6, but should be fine to do like this:
```suggestion
			// Create an AWSNodeTemplate for each static field
```

njtran (dismissed):
LGTM with one comment nit

pkg/apis/v1alpha1/awsnodetemplate.go
```diff
@@ -97,6 +100,16 @@ type AWSNodeTemplate struct {
...
 }
 
+func (a *AWSNodeTemplate) Hash() string {
+	hash, _ := hashstructure.Hash(a.Spec, hashstructure.FormatV2, &hashstructure.HashOptions{
+		SlicesAsSets:    true,
+		IgnoreZeroValue: true,
+		ZeroNil:         true,
+	})
+
+	return fmt.Sprint(hash)
```
jonathan-innis, line 110:
```suggestion
	return string(hash)
```

It think it's basically the same but probably better to not rely on `fmt` here

pkg/apis/v1alpha1/provider.go
```diff
@@ -24,23 +24,23 @@ import (
...
+	metav1.TypeMeta `json:",inline" hash:"ignore"`
 	// AMIFamily is the AMI family that instances use.
 	// +optional
-	AMIFamily *string `json:"amiFamily,omitempty"`
+	AMIFamily *string `json:"amiFamily,omitempty" hash:"ignore"`
 	// Context is a Reserved field in EC2 APIs
 	// https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html
 	// +optional
-	Context *string `json:"context,omitempty"`
+	Context *string `json:"context,omitempty" hash:"ignore"`
```
jonathan-innis, line 34: Why are we ignoring context? _Should_ we be ignoring context here?

pkg/apis/v1alpha1/provider.go
```diff
@@ -53,7 +53,7 @@ type LaunchTemplate struct {
 	// NOTE: This field is for specifying a custom launch template and is exposed in the Spec
 	// as `launchTemplate` for backwards compatibility.
 	// +optional
-	LaunchTemplateName *string `json:"launchTemplate,omitempty"`
+	LaunchTemplateName *string `json:"launchTemplate,omitempty" hash:"ignore"`
```
jonathan-innis, line 56: Any reason not to drift on this if someone changes the name here? If someone changes the name, then we know that they have definitely drifted.

pkg/apis/v1alpha1/register.go
```diff
@@ -107,6 +107,7 @@ var (
 	LabelInstanceAcceleratorName              = LabelDomain + "/instance-accelerator-name"
 	LabelInstanceAcceleratorManufacturer      = LabelDomain + "/instance-accelerator-manufacturer"
 	LabelInstanceAcceleratorCount             = LabelDomain + "/instance-accelerator-count"
+	AnnotationProviderHash                    = LabelDomain + "/provider-hash"
```
jonathan-innis, line 110:
```suggestion
	AnnotationNodeTemplateHash                    = LabelDomain + "/nodetemplate-hash"
```

pkg/cloudprovider/cloudprovider.go
```diff
@@ -87,6 +87,9 @@ func (c *CloudProvider) Create(ctx context.Context, machine *v1alpha5.Machine) (
 	if err != nil {
 		return nil, fmt.Errorf("resolving node template, %w", err)
 	}
+	machine.Annotations = lo.Assign(machine.Annotations, map[string]string{
```
jonathan-innis, line 90: Should this be resolved at the end with all the other `toMachine` details?

pkg/apis/v1alpha1/provider.go
engedaam: Since it is a deprecated field, I was thinking we should not build any new capability to it

pkg/cloudprovider/cloudprovider.go
engedaam: No, we should not because we want to only resolve this on creation. `toMachine` is utilized on GET and LIST calls

pkg/apis/v1alpha1/awsnodetemplate.go
engedaam: Since we are converting from `uint64` to a string, the ci-test will not pass

jonathan-innis: Oh! this is a uint64. My bad. Didn't realize it wasn't a string.

pkg/apis/v1alpha1/provider.go
```diff
@@ -24,10 +24,10 @@ import (
 type AWS struct {
 	// TypeMeta includes version and kind of the extensions, inferred if not provided.
 	// +optional
-	metav1.TypeMeta `json:",inline"`
+	metav1.TypeMeta `json:",inline" hash:"ignore"`
 	// AMIFamily is the AMI family that instances use.
 	// +optional
-	AMIFamily *string `json:"amiFamily,omitempty"`
+	AMIFamily *string `json:"amiFamily,omitempty" hash:"ignore"`
```
jonathan-innis, line 30: What happens if `AMIFamily` changes but AMISelector doesn't change?

jonathan-innis: i.e. Changing from `Bottlerocket` AMIFamily to `Custom` or vise-versa?

pkg/apis/v1alpha1/provider.go
jonathan-innis: Yeah, that's probably the right call. SGTM 👍

pkg/apis/v1alpha1/suite_test.go
```diff
@@ -115,4 +120,98 @@ var _ = Describe("Validation", func() {
...
+							},
+						},
+					},
+				},
+				UserData:           aws.String("userdata-test-1"),
+				DetailedMonitoring: aws.Bool(false),
+			}
+			awsnodetemplate = test.AWSNodeTemplate(awsnodetemplatespec)
+		})
+		It("should change hash when static fields are updated", func() {
```
jonathan-innis, line 154: `DescribeTable`?

pkg/cloudprovider/cloudprovider.go
jonathan-innis: I don't think this value gets propagated back to the returned Machine if you do it this way. You should validate this with E2E testing but I'm surprised if it is.

jonathan-innis, line 90: We should be adding tests around `cloudprovider_test.go` to ensure that the hash is actually returned on `Create`

pkg/apis/v1alpha1/provider.go
engedaam: That will be picked up by the AMI dynamic drift. We the AMIFaimly will alter what the resolved list of AMI IDs will be, so It will drift.

pkg/apis/v1alpha1/suite_test.go
```diff
@@ -115,4 +120,93 @@ var _ = Describe("Validation", func() {
...
+		})
+		It("should expect two provisioner with the same spec to have the same provisioner hash", func() {
+			awsnodetemplateTwo := &v1alpha1.AWSNodeTemplate{
+				ObjectMeta: metav1.ObjectMeta{Name: strings.ToLower(randomdata.SillyName())},
+			}
+			awsnodetemplateTwo.Spec = awsnodetemplatespec
+
+			Expect(awsnodetemplate.Hash()).To(Equal(awsnodetemplateTwo.Hash()))
+		})
+		It("should expect fields that are reordered to not produce a new hash", func() {
```
jonathan-innis, line 197: `DescribeTable` I think here too. I think if you find yourself iterating over what seems like a test case in an `It` block, we should be using a `DescribeTable`

pkg/apis/v1alpha1/suite_test.go
```diff
@@ -16,18 +16,23 @@ package v1alpha1_test
 
 import (
```
jonathan-innis, line 17: Really nice test coverage! I feel like this covers a comprehensive set of cases!

pkg/cloudprovider/cloudprovider.go
```diff
@@ -101,7 +101,11 @@ func (c *CloudProvider) Create(ctx context.Context, machine *v1alpha5.Machine) (
 	instanceType, _ := lo.Find(instanceTypes, func(i *cloudprovider.InstanceType) bool {
 		return i.Name == instance.Type
 	})
-	return c.instanceToMachine(instance, instanceType), nil
+	outputMachine := c.instanceToMachine(instance, instanceType)
```
jonathan-innis, line 104:
```suggestion
	m := c.instanceToMachine(instance, instanceType)
```

nit: Maybe just consider shortening

pkg/apis/v1alpha1/provider.go
jonathan-innis: Not necessarily. Imagine someone uses an AMISelector. Now the AMIFamily does not affect the set of resolved AMIs. Now it only affects how we treat the userData and some details of node launch like allocatable details.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -101,7 +101,11 @@ func (c *CloudProvider) Create(ctx context.Context, machine *v1alpha5.Machine) (
 	instanceType, _ := lo.Find(instanceTypes, func(i *cloudprovider.InstanceType) bool {
 		return i.Name == instance.Type
 	})
-	return c.instanceToMachine(instance, instanceType), nil
+	m := c.instanceToMachine(instance, instanceType)
+	m.Annotations = lo.Assign(machine.Annotations, map[string]string{
```
jonathan-innis, line 105:
```suggestion
	m.Annotations = lo.Assign(m.Annotations, map[string]string{
```

jonathan-innis (approved):
LGTM 🚀

PR #4320, opened 2023-07-21
chore:Add default securityContext and podSecurityContext to Karpenter deployment #4278

### LINKED ISSUE #4278: Add default `securityContext` and `podSecurityContext` to Karpenter deployment
### Description

Currently, Karpenter allows you to set the `securityContext` and `podSecurityContext` within the Helm chart values file but doesn't provide default values that would be "secure by default". We should enforce default values that enforce this by doing things like:

```yaml
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
    - ALL
  readOnlyRootFilesystem: true
  runAsNonRoot: true
  seccompProfile:
    type: RuntimeDefault
```

bhargaje:
Fixes #4278 

**Description**
This change is to set the default securityContext for controller pod within the Helm chart values file to enforce secure by default.

Tried scaling the pods and Karpenter was able to provision the new nodes as expected.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter/README.md
+++ b/charts/karpenter/README.md
@@ -45,10 +45,6 @@ helm upgrade --install --namespace karpenter --create-namespace \
 | controller.metrics.port | int | `8000` | The container port to use for metrics. |
 | controller.outputPaths | list | `["stdout"]` | Controller outputPaths - default to stdout only |
 | controller.resources | object | `{}` | Resources for the controller pod. |
-| controller.securityContext | object | `{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"readOnlyRootFilesystem":true}` | SecurityContext for the container. |
-| controller.securityContext.allowPrivilegeEscalation | bool | `false` | Disabling privilege escalation for the container. |
-| controller.securityContext.capabilities | object | `{"drop":["ALL"]}` | Dropping all capabilities for the container. |
-| controller.securityContext.readOnlyRootFilesystem | bool | `true` | Set root filesystem to read only for the container. |
 | controller.sidecarContainer | list | `[]` | Additional sidecarContainer config |
 | controller.sidecarVolumeMounts | list | `[]` | Additional volumeMounts for the sidecar - this will be added to the volume mounts on top of extraVolumeMounts |
 | dnsConfig | object | `{}` | Configure DNS Config for the pod |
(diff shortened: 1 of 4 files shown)


bhargaje:
https://github.com/aws/karpenter/issues/4278

jonathan-innis (commented):
/karpenter snapshot

charts/karpenter/values.yaml
```diff
@@ -103,7 +103,20 @@ controller:
     # -- SHA256 digest of the controller image.
     digest: sha256:bac5ea470c09df21eb3742cf9448e9b806138ed5b6321d4e04697bbdf122eac6
   # -- SecurityContext for the controller container.
-  securityContext: {}
+  securityContext:
```
jonathan-innis, line 106: Should we just set all the defaults for this at the `podSecurityContext` level?

bhargaje: Hi @jonathan-innis, Since allow privilege escalation, capabilities ReadOnly filesystem are available only for container security context, I modified all these under container security context so that it would be ease for someone to change. Let me your thoughts on this?

jonathan-innis: Ah, got it. I didn't realize that some of the parameters were not available at the pod-level. Yep, that makes sense.

jonathan-innis: Can we also set `runAsGroup` and `runAsUser` values at the pod-security context level on top of the other parameters?

bhargaje: I have added the runAsGroup and runAsUser now.

jonathan-innis (dismissed):
LGTM 🚀

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

charts/karpenter/README.md
```diff
@@ -45,7 +45,10 @@ helm upgrade --install --namespace karpenter --create-namespace \
 | controller.metrics.port | int | `8000` | The container port to use for metrics. |
 | controller.outputPaths | list | `["stdout"]` | Controller outputPaths - default to stdout only |
 | controller.resources | object | `{}` | Resources for the controller pod. |
-| controller.securityContext | object | `{}` | SecurityContext for the controller container. |
+| controller.securityContext | object | `{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"readOnlyRootFilesystem":true}` | SecurityContext for the container. |
```
bwagner5, line 48: I think we could just hardcode the security related parameters since I don't see a use-case for a user changing these. wdyt?

jonathan-innis: @bhargaje I think it probably makes sense to hardcode this since I can't think of a good reason a user would want to make the karpenter containers less secure.

jonathan-innis: Pushed this down into static values in the deployment

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

bwagner5 (approved):
worth calling out in the upgrade guide too

PR #4339, opened 2023-07-25
chore: Update Controller to Include Hash

engedaam:
Fixes #N/A 

**Description**
- Added AWSNodeTemplate hash to be included in the AWSNodeTemplate Annotations 

- Unit and manual testing

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/controllers/nodetemplate/controller.go
+++ b/pkg/controllers/nodetemplate/controller.go
@@ -22,6 +22,7 @@ import (
 
 	"go.uber.org/multierr"
 	"golang.org/x/time/rate"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/client-go/util/workqueue"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -61,14 +62,21 @@ func NewController(kubeClient client.Client, subnetProvider *subnet.Provider, se
 func (c *Controller) Reconcile(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate) (reconcile.Result, error) {
 	stored := nodeTemplate.DeepCopy()
 
+	nodeTemplate.Annotations = lo.Assign(nodeTemplate.ObjectMeta.Annotations, map[string]string{v1alpha1.AnnotationNodeTemplateHash: nodeTemplate.Hash()})
 	err := multierr.Combine(
 		c.resolveSubnets(ctx, nodeTemplate),
 		c.resolveSecurityGroups(ctx, nodeTemplate),
 		c.resolveAMIs(ctx, nodeTemplate),
 	)
 
-	if patchErr := c.kubeClient.Status().Patch(ctx, nodeTemplate, client.MergeFrom(stored)); patchErr != nil {
-		err = multierr.Append(err, client.IgnoreNotFound(patchErr))
+	if !equality.Semantic.DeepEqual(stored, nodeTemplate) {
+		statusCopy := nodeTemplate.DeepCopy()
+		if patchErr := c.kubeClient.Patch(ctx, nodeTemplate, client.MergeFrom(stored)); patchErr != nil {
+			err = multierr.Append(err, client.IgnoreNotFound(patchErr))
+		}
+		if patchErr := c.kubeClient.Status().Patch(ctx, statusCopy, client.MergeFrom(stored)); patchErr != nil {
+			err = multierr.Append(err, client.IgnoreNotFound(patchErr))
+		}
 	}
 
 	return reconcile.Result{RequeueAfter: 5 * time.Minute}, err
--- a/charts/karpenter/templates/clusterrole.yaml
+++ b/charts/karpenter/templates/clusterrole.yaml
@@ -42,5 +42,5 @@ rules:
     resourceNames: ["defaulting.webhook.karpenter.k8s.aws"]
   # Write
   - apiGroups: ["karpenter.k8s.aws"]
-    resources: ["awsnodetemplates/status"]
+    resources: ["awsnodetemplates", "awsnodetemplates/status"]
     verbs: ["patch", "update"]
--- a/pkg/controllers/nodetemplate/suite_test.go
+++ b/pkg/controllers/nodetemplate/suite_test.go
@@ -720,6 +720,55 @@ var _ = Describe("AWSNodeTemplateController", func() {
 			))
 		})
 	})
+	Context("AWSNodeTemplate Static Drift Hash", func() {
+		DescribeTable("should update the static drift hash when nodeTemplate static field is updated", func(awsnodetemplatespec v1alpha1.AWSNodeTemplateSpec) {
+			updatedAWSNodeTemplate := test.AWSNodeTemplate(*nodeTemplate.Spec.DeepCopy(), awsnodetemplatespec)
+			updatedAWSNodeTemplate.ObjectMeta = nodeTemplate.ObjectMeta
+			updatedAWSNodeTemplate.Annotations = map[string]string{v1alpha1.AnnotationNodeTemplateHash: updatedAWSNodeTemplate.Hash()}
+
+			ExpectApplied(ctx, env.Client, nodeTemplate)
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(nodeTemplate))
+			nodeTemplate = ExpectExists(ctx, env.Client, nodeTemplate)
+
+			expectedHash := nodeTemplate.Hash()
+			Expect(nodeTemplate.ObjectMeta.Annotations[v1alpha1.AnnotationNodeTemplateHash]).To(Equal(expectedHash))
+
+			ExpectApplied(ctx, env.Client, updatedAWSNodeTemplate)
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(nodeTemplate))
+			nodeTemplate = ExpectExists(ctx, env.Client, nodeTemplate)
+
+			expectedHashTwo := nodeTemplate.Hash()
+			Expect(expectedHash).ToNot(Equal(expectedHashTwo))
+			Expect(nodeTemplate.ObjectMeta.Annotations[v1alpha1.AnnotationNodeTemplateHash]).To(Equal(expectedHashTwo))
+		},
+			Entry("InstanceProfile Drift", v1alpha1.AWSNodeTemplateSpec{AWS: v1alpha1.AWS{InstanceProfile: aws.String("profile-2")}}),
+			Entry("UserData Drift", v1alpha1.AWSNodeTemplateSpec{UserData: aws.String("userdata-test-2")}),
+			Entry("Tags Drift", v1alpha1.AWSNodeTemplateSpec{AWS: v1alpha1.AWS{Tags: map[string]string{"keyTag-test-3": "valueTag-test-3"}}}),
+			Entry("MetadataOptions Drift", v1alpha1.AWSNodeTemplateSpec{AWS: v1alpha1.AWS{LaunchTemplate: v1alpha1.LaunchTemplate{MetadataOptions: &v1alpha1.MetadataOptions{HTTPEndpoint: aws.String("test-metadata-2")}}}}),
+			Entry("BlockDeviceMappings Drift", v1alpha1.AWSNodeTemplateSpec{AWS: v1alpha1.AWS{LaunchTemplate: v1alpha1.LaunchTemplate{BlockDeviceMappings: []*v1alpha1.BlockDeviceMapping{{DeviceName: aws.String("map-device-test-3")}}}}}),
+			Entry("Context Drift", v1alpha1.AWSNodeTemplateSpec{AWS: v1alpha1.AWS{Context: aws.String("context-2")}}),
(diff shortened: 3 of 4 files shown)


pkg/controllers/nodetemplate/controller.go
```diff
@@ -30,48 +29,60 @@ import (
...
-	subnetProvider        *subnet.Provider
-	securityGroupProvider *securitygroup.Provider
-	amiProvider           *amifamily.Provider
+	kubeClient client.Client
+	status     *Status
+	hash       *Hash
 }
 
-func NewController(kubeClient client.Client, subnetProvider *subnet.Provider, securityGroups *securitygroup.Provider, amiprovider *amifamily.Provider) corecontroller.Controller {
+func NewController(kubeClient client.Client, subnetprovider *subnet.Provider, securitygroupprovider *securitygroup.Provider, amiprovider *amifamily.Provider) corecontroller.Controller {
```
jonathan-innis, line 52: Why did you decide to remove the camel casing here?

pkg/controllers/nodetemplate/controller.go
```diff
@@ -30,48 +29,60 @@ import (
...
+		res, err := reconciler.Reconcile(ctx, nodeTemplate)
+		errs = multierr.Append(errs, err)
+		results = append(results, res)
+	}
 
-	if patchErr := c.kubeClient.Status().Patch(ctx, nodeTemplate, client.MergeFrom(stored)); patchErr != nil {
-		err = multierr.Append(err, client.IgnoreNotFound(patchErr))
+	if !equality.Semantic.DeepEqual(stored, nodeTemplate) {
+		statusCopy := nodeTemplate.DeepCopy()
+		if patchErr := c.kubeClient.Patch(ctx, nodeTemplate, client.MergeFrom(stored)); patchErr != nil {
```
jonathan-innis, line 77: Should this just be the patch error or should we also append the other errors that we have to this error as well?

pkg/controllers/nodetemplate/hash.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+	"github.com/samber/lo"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type Hash struct{}
+
+func (s *Hash) Reconcile(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate) (reconcile.Result, error) {
+	nodeTemplateHash := nodeTemplate.Hash()
+	nodeTemplate.ObjectMeta.Annotations = lo.Assign(nodeTemplate.ObjectMeta.Annotations, map[string]string{v1alpha1.AnnotationAWSNodeTemplateHash: nodeTemplateHash})
```
jonathan-innis, line 29:
```suggestion
	nodeTemplate.Annotations = lo.Assign(nodeTemplate.Annotations, map[string]string{v1alpha1.AnnotationAWSNodeTemplateHash: nodeTemplateHash})
```

pkg/controllers/nodetemplate/status.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+	"github.com/aws/karpenter/pkg/providers/amifamily"
+	"github.com/aws/karpenter/pkg/providers/securitygroup"
+	"github.com/aws/karpenter/pkg/providers/subnet"
+	"github.com/samber/lo"
+	"go.uber.org/multierr"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// Drift is a machine sub-controller that adds or removes status conditions on drifted machines
```
jonathan-innis, line 33: This comment header looks incorrect

pkg/controllers/nodetemplate/hash.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/samber/lo"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+)
+
+type Hash struct{}
+
+func (h *Hash) Reconcile(_ context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate) (reconcile.Result, error) {
+	nodeTemplateHash := nodeTemplate.Hash()
```
jonathan-innis, line 29: nit: Why not inline this?

pkg/controllers/nodetemplate/hash_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+		expectedHash := nodeTemplate.Hash()
+		Expect(nodeTemplate.ObjectMeta.Annotations[v1alpha1.AnnotationNodeTemplateHash]).To(Equal(expectedHash))
+
+		nodeTemplate.Spec.UserData = aws.String("test-userData-2")
+		ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(nodeTemplate))
+		nodeTemplate = ExpectExists(ctx, env.Client, nodeTemplate)
+
+		expectedHashTwo := nodeTemplate.Hash()
+		Expect(nodeTemplate.ObjectMeta.Annotations[v1alpha1.AnnotationNodeTemplateHash]).To(Equal(expectedHashTwo))
```
jonathan-innis, line 42: nit: Can we also expect that this hash is changed from the old hash?

pkg/controllers/nodetemplate/hash.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/samber/lo"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+)
+
+type Hash struct{}
+
+func (h *Hash) Reconcile(_ context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate) (reconcile.Result, error) {
+	nodeTemplate.Annotations = lo.Assign(nodeTemplate.ObjectMeta.Annotations, map[string]string{v1alpha1.AnnotationNodeTemplateHash: nodeTemplate.Hash()})
```
njtran, line 29: Since there's no chance for this to fail, I don't think this needs its own reconciliation loop. You can just in-line this and remove this file.

pkg/controllers/nodetemplate/controller.go
```diff
@@ -30,48 +29,60 @@ import (
...
-	err := multierr.Combine(
-		c.resolveSubnets(ctx, nodeTemplate),
-		c.resolveSecurityGroups(ctx, nodeTemplate),
-		c.resolveAMIs(ctx, nodeTemplate),
-	)
+	var results []reconcile.Result
+	var errs error
+	reconcilers := []nodeTemplateReconciler{
+		c.status,
+		c.hash,
```
njtran, line 67: likewise for my other comment, this has no chance to fail, so I think you may be over-abstracting this. Personally, I think you should be able to leave it as it was and just always idempotently add in the hash annotation.

pkg/controllers/nodetemplate/controller.go
```diff
@@ -30,48 +29,60 @@ import (
...
+		if patchErr := c.kubeClient.Patch(ctx, nodeTemplate, client.MergeFrom(stored)); patchErr != nil {
+			errs = multierr.Append(errs, patchErr)
+		}
+		if patchErr := c.kubeClient.Status().Patch(ctx, statusCopy, client.MergeFrom(stored)); patchErr != nil {
+			errs = multierr.Append(errs, patchErr)
+		}
 	}
 
-	return reconcile.Result{RequeueAfter: 5 * time.Minute}, err
+	return result.Min(results...), errs
```
njtran, line 85: Does result.Min() return 0 if there are no errors? We should ensure that we're still requeue-ing every 5 minutes regardless. Maybe something like 
```suggestion
	return result.Min(reconcile.Result{RequeueAfter: 5 * time.Minute}, results...), errs
```

pkg/controllers/nodetemplate/hash_test.go
```diff
@@ -0,0 +1,64 @@
...
+		ExpectApplied(ctx, env.Client, nodeTemplate)
+		ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(nodeTemplate))
+		nodeTemplate = ExpectExists(ctx, env.Client, nodeTemplate)
+
+		expectedHash := nodeTemplate.Hash()
+		Expect(nodeTemplate.ObjectMeta.Annotations[v1alpha1.AnnotationNodeTemplateHash]).To(Equal(expectedHash))
+
+		nodeTemplate.Spec.SubnetSelector = map[string]string{"aws-ids": "subnet-test1"}
+		nodeTemplate.Spec.SecurityGroupSelector = map[string]string{"aws-ids": "sg-test1"}
+		nodeTemplate.Spec.AMISelector = map[string]string{"ami-test-key": "ami-test-value"}
```
njtran, line 56: Make sure you add in behavioral fields in this test too.

pkg/controllers/nodetemplate/hash_test.go
```diff
@@ -0,0 +1,64 @@
...
+var _ = Describe("AWSNodeTemplate Static Drift Hash", func() {
+	It("should update the static drift hash when nodeTemplate static field is updated", func() {
+		ExpectApplied(ctx, env.Client, nodeTemplate)
+		ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(nodeTemplate))
+		nodeTemplate = ExpectExists(ctx, env.Client, nodeTemplate)
+
+		expectedHash := nodeTemplate.Hash()
+		Expect(nodeTemplate.ObjectMeta.Annotations[v1alpha1.AnnotationNodeTemplateHash]).To(Equal(expectedHash))
+
+		nodeTemplate.Spec.UserData = aws.String("test-userData-2")
```
njtran, line 37: Make sure you test all static fields here.

pkg/controllers/nodetemplate/status.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/aws/karpenter/pkg/providers/subnet"
+)
+
+type Status struct {
+	subnetProvider        *subnet.Provider
+	securityGroupProvider *securitygroup.Provider
+	amiProvider           *amifamily.Provider
+}
+
+func (s *Status) Reconcile(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate) (reconcile.Result, error) {
```
njtran, line 40: I appreciate the desire to abstract, but I think you could've kept as is.

pkg/controllers/nodetemplate/status_test.go
```diff
@@ -0,0 +1,669 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+
+	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
+	. "github.com/aws/karpenter-core/pkg/test/expectations"
+
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+)
+
+var _ = Describe("AWSNodeTemplate Status", func() {
```
njtran, line 38: I think you can just keep these two suite_test files as one file in the same folder with different Describe() functions, I think it's actually easier to reason about since these two files are both pretty short compared to other suite_test files.

pkg/controllers/nodetemplate/hash_test.go
engedaam: These are all the fields that will not be drifted

pkg/controllers/nodetemplate/controller.go
```diff
@@ -61,14 +62,21 @@ func NewController(kubeClient client.Client, subnetProvider *subnet.Provider, se
...
 		c.resolveSubnets(ctx, nodeTemplate),
 		c.resolveSecurityGroups(ctx, nodeTemplate),
 		c.resolveAMIs(ctx, nodeTemplate),
 	)
 
-	if patchErr := c.kubeClient.Status().Patch(ctx, nodeTemplate, client.MergeFrom(stored)); patchErr != nil {
-		err = multierr.Append(err, client.IgnoreNotFound(patchErr))
+	if !equality.Semantic.DeepEqual(stored, nodeTemplate) {
+		statusCopy := nodeTemplate.DeepCopy()
+		if patchErr := c.kubeClient.Patch(ctx, nodeTemplate, client.MergeFrom(stored)); patchErr != nil {
```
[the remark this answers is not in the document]
engedaam: No, it does not patch the status. So we need to do both patchs

pkg/controllers/nodetemplate/controller.go
```diff
@@ -61,14 +62,21 @@ func NewController(kubeClient client.Client, subnetProvider *subnet.Provider, se
 func (c *Controller) Reconcile(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate) (reconcile.Result, error) {
 	stored := nodeTemplate.DeepCopy()
 
+	nodeTemplate.Annotations = lo.Assign(nodeTemplate.ObjectMeta.Annotations, map[string]string{v1alpha1.AnnotationNodeTemplateHash: nodeTemplate.Hash()})
```
jonathan-innis, line 65: Do we have an E2E test ensuring the presence of this annotation? I don't think anything else would validate this since this is just there for observability right now

jonathan-innis (dismissed):
LGTM 🚀

jonathan-innis (approved):
LGTM 🚀

PR #4415, opened 2023-08-11
chore: Add conversion utils for `v1beta1/NodeClass`

jonathan-innis:
Fixes #N/A 

**Description**

This PR adds conversion functions to convert `NodeTemplate` to `NodeClass` The long-term goal of these changes is to enable us to treat every internal in-memory object in Karpenter as a `v1beta1` object (this is our model) and be able to convert both from the `v1beta1` APIs (View 1) and the `v1alpha5` APIs (View 2) into this internal model.

These conversion functions will enable us to create a shim layer around each of our reconciliations to convert between the alpha APIs and the beta APIs while still allowing both APIs to exist simultaneously at the apiserver for the migration path.

`make presubmit`

- [ ] Yes, issue opened: # 
- [x] No


jonathan-innis (commented):
/karpenter snapshot

ellistarn (approved):

PR #4433, opened 2023-08-14
chore: AssumeRoleDuration to `time.Duration`

engedaam:
Fixes #N/A 

**Description**
- Utilize time.Duration instead of metav1.Duration

- `/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No


engedaam (commented):
/karpenter snapshot

pkg/apis/settings/settings.go
```diff
@@ -158,6 +157,24 @@ func AsTypedString[T ~string](key string, target *T) configmap.ParseFunc {
 	}
 }
 
+// AsMetaDuration parses the value at key as a time.Duration into the target, if it exists.
+func AsMetaDuration(key string, target **metav1.Duration) configmap.ParseFunc {
+	return func(data map[string]string) error {
+		if raw, ok := data[key]; ok {
+			if raw == "" {
+				*target = &metav1.Duration{Duration: time.Minute * 15}
```
ellistarn, line 165: I don't know if we want to default implicitly for every duration.

Also -- why metaDuration rather than just duration?

jonathan-innis: IIRC, `metav1.Duration` was used originally for json serialization/deserialization. We may be able to remove its usage now in favor of `time.Duration` since we don't need the string serialization any longer.

sftim:
@engedaam: how do we know that 15 minutes is a good choice?

pkg/apis/settings/settings.go
engedaam: - Putting `metav1.Duration` for consistency across all Duration values on the in settings interface. 
- We would not be using these changes in coresettings as users need to set `batchMaxDuration` and `batchIdleDuration`

engedaam:
> @engedaam: how do we know that 15 minutes is a good choice?

@sftim By default assumeRoleDuration is set at 15min: https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.44.326/aws/credentials/stscreds#NewCredentials

jonathan-innis (approved):
LGTM 🚀

PR #4447, opened 2023-08-16
fix: provisioning when a misconfigured awsnodetemplate selects 0 subnets

bwagner5:
Fixes #N/A 
https://github.com/aws/karpenter/issues/4441

**Description**
 - Before this fix, a misconfigured AWSNodeTemplate that selects 0 subnets would cause all Karpenter provisioning to stop, even when there are properly configured Provisioners. The fix removes an error returned from GetInstanceTypes when a selector matches 0 subnets. A log is still produced via the AWSNodeTemplate controller to alert the user of a potential misconfiguration. 

 - Tested in a live cluster to reproduce and added a suite_test that failed before the fix and passes after the fix.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/instancetype/instancetype.go
+++ b/pkg/providers/instancetype/instancetype.go
@@ -177,7 +177,7 @@ func (p *Provider) getInstanceTypeZones(ctx context.Context, nodeTemplate *v1alp
 		return nil, err
 	}
 	if len(subnets) == 0 {
-		return nil, fmt.Errorf("no subnets matched selector %v", nodeTemplate.Spec.SubnetSelector)
+		return nil, nil
 	}
 	zones := sets.NewString(lo.Map(subnets, func(subnet *ec2.Subnet, _ int) string {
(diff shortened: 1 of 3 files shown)


pkg/providers/instancetype/instancetype.go
```diff
@@ -177,7 +177,7 @@ func (p *Provider) getInstanceTypeZones(ctx context.Context, nodeTemplate *v1alp
 		return nil, err
 	}
 	if len(subnets) == 0 {
-		return nil, fmt.Errorf("no subnets matched selector %v", nodeTemplate.Spec.SubnetSelector)
+		return nil, nil
```
ellistarn, line 180: Assuming we log as part of the nodetemplatecontroller?

bwagner5: yep, I logged here first and then realized how many times this things is called

jonathan-innis (approved):
LGTM 🚀

PR #4453, opened 2023-08-17
chore: Permission for leasegarbagecollection controller 

engedaam:
Fixes #N/A 

**Description**
- Adding permissions to run `leasegarbagecollection` controller 
- Bumping core to include https://github.com/aws/karpenter-core/pull/471

- /kapenter snapshot
 
- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/suites/integration/lease_garbagecollection_test.go
+++ b/test/suites/integration/lease_garbagecollection_test.go
@@ -1,0 +1,44 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration_test
+
+import (
+	"time"
+
+	coordinationsv1 "k8s.io/api/coordination/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/aws/karpenter-core/pkg/test"
+
+	. "github.com/onsi/ginkgo/v2"
+)
+
+var _ = Describe("Lease Garbage Collection", func() {
+	var badLease *coordinationsv1.Lease
+	BeforeEach(func() {
+		badLease = &coordinationsv1.Lease{
+			ObjectMeta: v1.ObjectMeta{
+				CreationTimestamp: v1.Time{Time: time.Now().Add(-time.Hour * 2)},
+				Name:              "new-lease",
+				Namespace:         "kube-node-lease",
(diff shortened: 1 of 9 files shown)


engedaam (commented):
/karpenter snapshot

charts/karpenter/templates/clusterrole-core.yaml
```diff
@@ -29,6 +29,9 @@ metadata:
   {{- end }}
 rules:
   # Read
+  - apiGroups: ["coordination.k8s.io"]
+    resources: ["leases"]
+    verbs: ["list", "watch"]
```
jonathan-innis, line 34: I think you need "get" too, right?

charts/karpenter/templates/clusterrole-core.yaml
```diff
@@ -29,6 +29,9 @@ metadata:
   {{- end }}
```
jonathan-innis, line 29: Can we add an E2E test to validate this functionality?

charts/karpenter/templates/clusterrole-core.yaml
engedaam: I was able to run the it, and all the e2etests passed without the `get`

test/suites/integration/lease_garbagecollection_test.go
```diff
@@ -0,0 +1,44 @@
...
+		badLease = &coordinationsv1.Lease{
+			ObjectMeta: v1.ObjectMeta{
+				CreationTimestamp: v1.Time{Time: time.Now().Add(-time.Hour * 2)},
+				Name:              "new-lease",
+				Namespace:         "kube-node-lease",
+				Labels:            map[string]string{test.DiscoveryLabel: "unspecified"},
+			},
+		}
+	})
+	It("should delete node lease that does not contain an OwnerReference", func() {
```
jonathan-innis, line 40: Can we make the test more realistic? e.g. we have a use-case from the issue of this happening (i.e. make the gracefulTermination period for the node sufficiently long) so can we use that as a valid E2E test case for this problem?

charts/karpenter/templates/clusterrole-core.yaml
jonathan-innis: 🤔 Might as well just add it. It's effectively the same. I think you are "technically" right that you don't need it since the "Get()" from controller-runtime is actually based on the in-memory cache that relies on the List/Watch. Still, I don't see a reason to differentiate this from our other read permissions.

charts/karpenter/templates/clusterrole-core.yaml
```diff
@@ -29,6 +29,9 @@ metadata:
...
 rules:
   # Read
+  - apiGroups: ["coordination.k8s.io"]
```
jonathan-innis, line 32: We also don't need clusterrole on this. We just need permission from the single namespace. Is it possible to only give ourselves permission to the single namespace where we need the permission?

engedaam (commented):
/karpenter snapshot

jonathan-innis (approved):
LGTM 🚀 @engedaam is going to follow-up this PR with an E2E test case repro-ing the issue

jonathan-innis (approved):
LGTM 🚀

PR #4479, opened 2023-08-22
Cluster ca bundle setting

neerajvshah:
**Description**
Adding clusterCABundle as a Karpenter setting. This is needed for our use case where a Karpenter controller needs to communicate with a cluster, but does not use TLS itself - instead it talks to a local proxy over HTTP. Without this change, the CA bundle would be unresolvable and the CA bundle passed to the node would be empty, preventing that node from joining the cluster.

One call out I had was that this was failing silently for our use case before I caught this - this is because the validation below assumes that the rest config is TLS-enabled. Should we add a separate check to verify that the CAData string isn't empty before returning?

https://github.com/neerajvshah/karpenter/blob/main/pkg/operator/operator.go#L210-L223

EDIT - technically Kubernetes api-servers can be configured to ignore CA data from nodes with `--insecure-skip-tls-verify`, so it makes sense to not have the above check.

I tested this locally and saw that the CA bundle set by the global config map was ingested by Karpenter.

- [ ] Yes, issue opened: # 
- [] No

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -136,7 +136,7 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 		amiResolver,
 		securityGroupProvider,
 		subnetProvider,
-		lo.Must(getCABundle(operator.GetConfig())),
+		lo.Must(getCABundle(ctx, operator.GetConfig())),
 		operator.Elected(),
 		kubeDNSIP,
 		clusterEndpoint,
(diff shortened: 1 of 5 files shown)


ellistarn (approved):
💯 

> One call out I had was that this was failing silently for our use case before I caught this - this is because the validation below assumes that the rest config is TLS-enabled. Should we add a separate check to verify that the CAData string isn't empty before returning?

I think we have to allow users to set cabundle as empty, to support cases of communicating with an API server over HTTP.

jonathan-innis (approved):
Can we add a docs update to surface this setting as well?

pkg/operator/operator.go
```diff
@@ -207,11 +207,14 @@ func ResolveClusterEndpoint(ctx context.Context, eksAPI eksiface.EKSAPI) (string
 	return *out.Cluster.Endpoint, nil
 }
 
-func getCABundle(restConfig *rest.Config) (*string, error) {
+func getCABundle(ctx context.Context, restConfig *rest.Config) (*string, error) {
 	// Discover CA Bundle from the REST client. We could alternatively
 	// have used the simpler client-go InClusterConfig() method.
 	// However, that only works when Karpenter is running as a Pod
 	// within the same cluster it's managing.
+	if caBundle := settings.FromContext(ctx).ClusterCABundle; caBundle != "" {
```
jonathan-innis, line 215: If you wanted to support the empty case, you could make it a pointer and have the empty string mean that you want to disable CABundle discovery by default.

PR #4491, opened 2023-08-23
test: Prevent Drift from deleting nodes for unhealthy replacement nodes

engedaam:
Fixes #N/A 

**Description**
- Testing to ensure that expiration and drift do not continue if a machine fails to initialize and register 
- E2E testing for https://github.com/aws/karpenter-core/pull/451

- `/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/pkg/environment/common/expectations.go
+++ b/test/pkg/environment/common/expectations.go
@@ -364,10 +364,52 @@ func (env *Environment) ExpectCreatedNodeCount(comparator string, count int) []*
 	return createdNodes
 }
 
+func MachineNames(machines []*v1alpha5.Machine) []string {
+	return lo.Map(machines, func(m *v1alpha5.Machine, index int) string {
+		return m.Name
+	})
+}
+
 func NodeNames(nodes []*v1.Node) []string {
 	return lo.Map(nodes, func(n *v1.Node, index int) string {
 		return n.Name
 	})
+}
+
+func (env *Environment) ConsistentlyExpectNodeCount(comparator string, count int, duration string) []*v1.Node {
+	GinkgoHelper()
+	By(fmt.Sprintf("expecting nodes to be %s to %d for %s", comparator, count, duration))
+	nodeList := &v1.NodeList{}
+	Consistently(func(g Gomega) {
+		g.Expect(env.Client.List(env, nodeList, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
+		g.Expect(len(nodeList.Items)).To(BeNumerically(comparator, count),
+			fmt.Sprintf("expected %d nodes, had %d (%v) for %s", count, len(nodeList.Items), NodeNames(lo.ToSlicePtr(nodeList.Items)), duration))
+	}, duration).Should(Succeed())
+	return lo.ToSlicePtr(nodeList.Items)
+}
+
+func (env *Environment) ConsistentlyExpectMachineCount(comparator string, count int, duration string) []*v1alpha5.Machine {
+	GinkgoHelper()
+	By(fmt.Sprintf("expecting machines to be %s to %d for %s", comparator, count, duration))
+	machineList := &v1alpha5.MachineList{}
+	Consistently(func(g Gomega) {
+		g.Expect(env.Client.List(env, machineList, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
+		g.Expect(len(machineList.Items)).To(BeNumerically(comparator, count),
+			fmt.Sprintf("expected %d machines, had %d (%v) for %s", count, len(machineList.Items), MachineNames(lo.ToSlicePtr(machineList.Items)), duration))
+	}, duration).Should(Succeed())
+	return lo.ToSlicePtr(machineList.Items)
+}
+
+func (env *Environment) EventuallyExpectCordonedNodeCount(comparator string, count int) []*v1.Node {
+	GinkgoHelper()
+	By(fmt.Sprintf("waiting for cordoned nodes to be %s to %d", comparator, count))
+	nodeList := &v1.NodeList{}
+	Eventually(func(g Gomega) {
+		g.Expect(env.Client.List(env, nodeList, client.MatchingFields{"spec.unschedulable": "true"})).To(Succeed())
+		g.Expect(len(nodeList.Items)).To(BeNumerically(comparator, count),
+			fmt.Sprintf("expected %d cordoned nodes, had %d (%v)", count, len(nodeList.Items), NodeNames(lo.ToSlicePtr(nodeList.Items))))
+	}).Should(Succeed())
+	return lo.ToSlicePtr(nodeList.Items)
 }
 
 func (env *Environment) EventuallyExpectNodeCount(comparator string, count int) []*v1.Node {
@@ -380,6 +422,18 @@ func (env *Environment) EventuallyExpectNodeCount(comparator string, count int) 
 			fmt.Sprintf("expected %d nodes, had %d (%v)", count, len(nodeList.Items), NodeNames(lo.ToSlicePtr(nodeList.Items))))
 	}).Should(Succeed())
 	return lo.ToSlicePtr(nodeList.Items)
+}
+
+func (env *Environment) EventuallyExpectMachineCount(comparator string, count int) []*v1alpha5.Machine {
+	GinkgoHelper()
+	By(fmt.Sprintf("waiting for machines to be %s to %d", comparator, count))
+	machineList := &v1alpha5.MachineList{}
+	Eventually(func(g Gomega) {
+		g.Expect(env.Client.List(env, machineList, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
+		g.Expect(len(machineList.Items)).To(BeNumerically(comparator, count),
+			fmt.Sprintf("expected %d machines, had %d (%v)", count, len(machineList.Items), MachineNames(lo.ToSlicePtr(machineList.Items))))
+	}).Should(Succeed())
+	return lo.ToSlicePtr(machineList.Items)
 }
 
 func (env *Environment) EventuallyExpectNodeCountWithSelector(comparator string, count int, selector labels.Selector) []*v1.Node {
--- a/test/suites/drift/suite_test.go
+++ b/test/suites/drift/suite_test.go
@@ -337,6 +337,121 @@ var _ = Describe("Drift", Label("AWS"), func() {
 		Entry("DetailedMonitoring Drift", "DetailedMonitoring", v1alpha1.AWSNodeTemplateSpec{DetailedMonitoring: awssdk.Bool(true)}),
 		Entry("AMIFamily Drift", "AMIFamily", v1alpha1.AWSNodeTemplateSpec{AWS: v1alpha1.AWS{AMIFamily: awssdk.String(v1alpha1.AMIFamilyBottlerocket)}}),
 	)
+	Context("Drift Failure", func() {
+		It("should not continue to drift if a node never registers", func() {
+			// launch a new machine
+			var numPods int32 = 2
+			dep := test.Deployment(test.DeploymentOptions{
+				Replicas: 2,
+				PodOptions: test.PodOptions{
+					ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"app": "inflate"}},
+					PodAntiRequirements: []v1.PodAffinityTerm{{
+						TopologyKey: v1.LabelHostname,
+						LabelSelector: &metav1.LabelSelector{
+							MatchLabels: map[string]string{"app": "inflate"},
+						}},
+					},
+				},
+			})
+			env.ExpectCreated(dep, nodeTemplate, provisioner)
+			env.EventuallyExpectNodeCount("==", 2)
+
+			// Drift the machine with bad configuration
+			parameter, err := env.SSMAPI.GetParameter(&ssm.GetParameterInput{
+				Name: awssdk.String("/aws/service/ami-amazon-linux-latest/amzn-ami-hvm-x86_64-ebs"),
+			})
+			Expect(err).ToNot(HaveOccurred())
+			nodeTemplate.Spec.AMISelector = map[string]string{"aws::ids": *parameter.Parameter.Value}
+			env.ExpectCreatedOrUpdated(nodeTemplate)
+
+			// Should see the machine has drifted
+			statingMachineState := env.EventuallyExpectCreatedMachineCount("==", int(numPods))
+			Eventually(func(g Gomega) {
+				for _, machine := range statingMachineState {
+					g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(machine), machine)).To(Succeed())
+					g.Expect(machine.StatusConditions().GetCondition(v1alpha5.MachineDrifted).IsTrue()).To(BeTrue())
+				}
+			}).Should(Succeed())
+
+			// Expect nodes To get cordoned
+			cordonedNodes := env.EventuallyExpectCordonedNodeCount("==", 1)
+
+			// Drift should fail and the original node should be uncordoned
+			// TODO: reduce timeouts when deprovisioning waits are factored out
+			Eventually(func(g Gomega) {
+				g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(cordonedNodes[0]), cordonedNodes[0]))
+				g.Expect(cordonedNodes[0].Spec.Unschedulable).To(BeFalse())
+			}).WithTimeout(11 * time.Minute).Should(Succeed())
+
+			endMachineState := &v1alpha5.MachineList{}
+			Eventually(func(g Gomega) {
+				g.Expect(env.Client.List(env, endMachineState, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
+				g.Expect(len(endMachineState.Items)).To(BeNumerically("==", int(numPods)))
+			}).WithTimeout(6 * time.Minute).Should(Succeed())
+
+			Consistently(func(g Gomega) {
+				g.Expect(lo.EveryBy(statingMachineState, func(sm *v1alpha5.Machine) bool {
+					g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(sm), sm)).To(Succeed())
+					return lo.ContainsBy(endMachineState.Items, func(em v1alpha5.Machine) bool {
+						g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(&em), &em)).To(Succeed())
+						return sm.Name == em.Name
+					})
+				})).To(BeTrue())
+			}, "2m").Should(Succeed())
+		})
+		It("should not continue to drift if a node registers but never becomes initialized", func() {
+			// launch a new machine
+			var numPods int32 = 2
+			dep := test.Deployment(test.DeploymentOptions{
+				Replicas: 2,
+				PodOptions: test.PodOptions{
+					ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"app": "inflate"}},
+					PodAntiRequirements: []v1.PodAffinityTerm{{
+						TopologyKey: v1.LabelHostname,
+						LabelSelector: &metav1.LabelSelector{
+							MatchLabels: map[string]string{"app": "inflate"},
+						}},
+					},
+				},
+			})
+			env.ExpectCreated(dep, nodeTemplate, provisioner)
+			statingNodeState := env.EventuallyExpectNodeCount("==", int(numPods))
+
+			// Drift the machine with bad configuration
+			provisioner.Spec.StartupTaints = []v1.Taint{{Key: "example.com/taint", Effect: v1.TaintEffectPreferNoSchedule}}
+			env.ExpectCreatedOrUpdated(provisioner)
+
+			// Should see the machine has drifted
+			machines := env.EventuallyExpectCreatedMachineCount("==", int(numPods))
+			Eventually(func(g Gomega) {
+				for _, machine := range machines {
+					g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(machine), machine)).To(Succeed())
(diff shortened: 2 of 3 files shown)


test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,111 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+		By(fmt.Sprintf("AMI: %s", *parameter.Parameter.Value))
+		nodeTemplate.Spec.AMISelector = map[string]string{"aws::ids": *parameter.Parameter.Value}
+		env.ExpectCreatedOrUpdated(nodeTemplate)
+
+		// Should see the machine has drifted
+		EventuallyWithOffset(1, func(g Gomega) {
+			for _, machine := range env.EventuallyExpectCreatedMachineCount("==", int(numPods)) {
+				g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(machine), machine)).To(Succeed())
+				g.Expect(machine.StatusConditions().GetCondition(v1alpha5.MachineDrifted)).ToNot(BeNil())
+				g.Expect(machine.StatusConditions().GetCondition(v1alpha5.MachineDrifted).IsTrue()).To(BeTrue())
```
njtran, line 372: I believe the IsTrue() call should also check if it's nil, so you probably don't need L371.

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,111 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+		}).Should(Succeed())
+
+		// Expect machine To get cordoned
+		env.EventuallyExpectMachineCount("==", 3)
+		nodes := env.EventuallyExpectNodeCount("==", int(numPods))
+		Eventually(func(g Gomega) {
+			g.Expect(lo.SomeBy(nodes, func(node *v1.Node) bool {
+				g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(node), node))
+				return node.Spec.Unschedulable == true
+			})).To(BeTrue())
```
njtran, line 383: I think we should only assert that one of them is tainted, right?

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,111 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+		}).Should(Succeed())
+
+		// Drift should fail and original node should be get uncoreded
```
njtran, line 386:
```suggestion
		// Drift should fail and original node should be uncordoned
```

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,111 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+		Eventually(func(g Gomega) {
+			g.Expect(lo.EveryBy(nodes, func(node *v1.Node) bool {
+				g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(node), node))
+				return node.Spec.Unschedulable == false
+			})).To(BeTrue())
+		}).WithTimeout(20 * time.Minute).Should(Succeed())
```
njtran, line 392: I wish there was some way we could make this quicker, as a 20 minute timeout is just really unfortunate. 
Can you add a TODO here to consider lowering this when we have factored out the waits here?

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,111 @@ var _ = Describe("Drift", Label("AWS"), func() {
 		Entry("DetailedMonitoring Drift", "DetailedMonitoring", v1alpha1.AWSNodeTemplateSpec{DetailedMonitoring: awssdk.Bool(true)}),
 		Entry("AMIFamily Drift", "AMIFamily", v1alpha1.AWSNodeTemplateSpec{AWS: v1alpha1.AWS{AMIFamily: awssdk.String(v1alpha1.AMIFamilyBottlerocket)}}),
 	)
+	It("should not continue to drift if a machine fails to register", func() {
```
njtran, line 340: nit: consider renaming to include something about the delete
```suggestion
	It("should delete a replacement machine from drift if it fails to register", func() {
```

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,111 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+		// Expect nodes To be cordoned
+		nodes := env.EventuallyExpectNodeCount("==", 3)
+		Eventually(func(g Gomega) {
+			g.Expect(lo.SomeBy(nodes, func(node *v1.Node) bool {
+				g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(node), node))
+				return node.Spec.Unschedulable == true
+			})).To(BeTrue())
+		}).Should(Succeed())
+
+		// Drift should fail and original node should be get uncoreded
```
njtran, line 436:
```suggestion
		// Drift should fail and original node should be uncordoned
```

njtran (commented):
Nice work, some comments

engedaam (commented):
/karpenter snapshot

engedaam (commented):
/karpenter snapshot

njtran (dismissed):
LGTM

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+		})
+		env.ExpectCreated(dep, nodeTemplate, provisioner)
+		env.EventuallyExpectNodeCount("==", 2)
+
+		// Drift the machine with bad configuration
+		parameter, err := env.SSMAPI.GetParameter(&ssm.GetParameterInput{
+			Name: awssdk.String("/aws/service/ami-amazon-linux-latest/amzn-ami-hvm-x86_64-ebs"),
+		})
+		Expect(err).ToNot(HaveOccurred())
+		By(fmt.Sprintf("AMI: %s", *parameter.Parameter.Value))
```
jonathan-innis, line 363: What's this `By` statement for? If we are going to be adding `By` statements, we should generally make them more descriptive for debugging purposes

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+		nodeTemplate.Spec.AMISelector = map[string]string{"aws::ids": *parameter.Parameter.Value}
+		env.ExpectCreatedOrUpdated(nodeTemplate)
+
+		// Should see the machine has drifted
+		EventuallyWithOffset(1, func(g Gomega) {
+			for _, machine := range env.EventuallyExpectCreatedMachineCount("==", int(numPods)) {
```
jonathan-innis, line 369: Shouldn't this eventually statement be before the `Eventually` that you are trying to check here? Ideally, we should be waiting for one condition to complete and then starting the next one

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+		parameter, err := env.SSMAPI.GetParameter(&ssm.GetParameterInput{
+			Name: awssdk.String("/aws/service/ami-amazon-linux-latest/amzn-ami-hvm-x86_64-ebs"),
+		})
+		Expect(err).ToNot(HaveOccurred())
+		By(fmt.Sprintf("AMI: %s", *parameter.Parameter.Value))
+		nodeTemplate.Spec.AMISelector = map[string]string{"aws::ids": *parameter.Parameter.Value}
+		env.ExpectCreatedOrUpdated(nodeTemplate)
+
+		// Should see the machine has drifted
+		EventuallyWithOffset(1, func(g Gomega) {
```
jonathan-innis, line 368:
```suggestion
		Eventually(func(g Gomega) {
```

As a general rule we should not be using `...WithOffset` anymore, in favor of `GinkgoHelper()`. In this case, we need neither since this is a top-level call

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+			for _, machine := range env.EventuallyExpectCreatedMachineCount("==", int(numPods)) {
+				g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(machine), machine)).To(Succeed())
+				g.Expect(machine.StatusConditions().GetCondition(v1alpha5.MachineDrifted).IsTrue()).To(BeTrue())
+			}
+		}).Should(Succeed())
+
+		// Expect machine To get cordoned
+		env.EventuallyExpectMachineCount("==", 3)
```
jonathan-innis, line 376: Do we need to ensure this again? At this point, we know we have 3 machines (and ideally we should know that we have 3 nodes). We should ensure this check 1 time and then just use the node/machine names to grab an updated version of the data like you do on L380

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+		nodes := env.EventuallyExpectNodeCount("==", int(numPods))
+		Eventually(func(g Gomega) {
+			g.Expect(len(lo.Filter(nodes, func(node *v1.Node, _ int) bool {
+				g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(node), node))
+				return node.Spec.Unschedulable == true &&
+					lo.ContainsBy(node.Spec.Taints, func(taint v1.Taint) bool { return taint.Key == v1.TaintNodeUnschedulable })
```
jonathan-innis, line 382: Do we need to check both of these conditions? AFAIK, these conditions occur together so ideally we should just be able to check if the node is unschedulable.

njtran (dismissed):

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+			}))).To(BeNumerically("==", 1))
+		}).Should(Succeed())
+
+		// Drift should fail and original node should be uncordoned
+		// TODO: reduce timeouts when deprovisioning waits are factored out
+		machineList := &v1alpha5.MachineList{}
+		Eventually(func(g Gomega) {
```
jonathan-innis, line 389: Do we need to check all nodes here or just the node that we care about? I feel like we are doing a lot of repetitive checks here that make the code harder to read as opposed to indexing for simplicity.

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
 		Entry("DetailedMonitoring Drift", "DetailedMonitoring", v1alpha1.AWSNodeTemplateSpec{DetailedMonitoring: awssdk.Bool(true)}),
 		Entry("AMIFamily Drift", "AMIFamily", v1alpha1.AWSNodeTemplateSpec{AWS: v1alpha1.AWS{AMIFamily: awssdk.String(v1alpha1.AMIFamilyBottlerocket)}}),
 	)
+	It("should delete a replacement machine from drift if it fails to register", func() {
```
jonathan-innis, line 340: Can we create a `Context` for these to sit inside of so that we know these are our failure scenarios?

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+			g.Expect(lo.EveryBy(nodes, func(node *v1.Node) bool {
+				g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(node), node))
+				return node.Spec.Unschedulable == false
+			})).To(BeTrue())
+			g.Expect(env.Client.List(env, machineList, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
+			g.Expect(len(machineList.Items)).To(BeNumerically("==", int(numPods)))
+		}).WithTimeout(20 * time.Minute).Should(Succeed())
+		env.EventuallyExpectNodeCount("==", int(numPods))
+	})
+	It("should not continue to drift if a machine fails to become initialized", func() {
```
jonathan-innis, line 399:
```suggestion
	It("should not continue to drift if a node registers but never becomes initialized", func() {
```

test/suites/drift/suite_test.go
jonathan-innis, line 340:
```suggestion
	It("should not continue to drift if a node never registers", func() {
```

jonathan-innis: I think we are mainly validating what we do if the node never registers, whether the node gets cleaned up when it doesn't register is handled by a test inside of the machine test suite

test/suites/integration/expiration_test.go
```diff
@@ -174,4 +181,113 @@ var _ = Describe("Expiration", func() {
 		env.EventuallyExpectCreatedNodeCount("==", 1)
 		env.EventuallyExpectHealthyPodCount(selector, int(numPods))
 	})
+	It("should delete a replacement machine from expiration if it fails to register", func() {
```
jonathan-innis, line 184: Another set of test suites where I would create a separate `Context` for the failure scenarios and then give a name to the test that describes more specifically what we are trying to validate here.

test/suites/integration/expiration_test.go
```diff
@@ -174,4 +181,113 @@ var _ = Describe("Expiration", func() {
...
+		machineList := &v1alpha5.MachineList{}
+		Eventually(func(g Gomega) {
+			g.Expect(lo.EveryBy(nodes, func(node *v1.Node) bool {
+				g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(node), node))
+				return node.Spec.Unschedulable == false
+			})).To(BeTrue())
+			g.Expect(env.Client.List(env, machineList, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
+			g.Expect(len(machineList.Items)).To(BeNumerically("==", int(numPods)))
+		}).WithTimeout(20 * time.Minute).Should(Succeed())
+		env.EventuallyExpectNodeCount("==", int(numPods))
```
jonathan-innis, line 241: I think we _might_ need to do a `Consistently` check here to validate that we aren't proceeding with scale down here. e.g. After the timeout expires and the node is uncordoned, I should ensure that deprovisioning doesn't keep scaling me down

engedaam (commented):
/karpenter snapshot

engedaam (commented):
/karpenter snapshot

test/pkg/environment/common/expectations.go
```diff
@@ -364,12 +364,30 @@ func (env *Environment) ExpectCreatedNodeCount(comparator string, count int) []*
...
+
 func NodeNames(nodes []*v1.Node) []string {
 	return lo.Map(nodes, func(n *v1.Node, index int) string {
 		return n.Name
 	})
 }
 
+func (env *Environment) ConsistentlyExpectNodeCount(comparator string, count int, duration string) []*v1.Node {
+	GinkgoHelper()
+	By(fmt.Sprintf("waiting for nodes to be %s to %d", comparator, count))
```
jonathan-innis, line 381: We need to make sure all of this messaging is consistent when we change the function to do something different. This `By` statement should say something like `expecting nodes to be %s to %d for %s", comparator, count, duration.String()`

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+					g.Expect(machine.StatusConditions().GetCondition(v1alpha5.MachineDrifted).IsTrue()).To(BeTrue())
+				}
+			}).Should(Succeed())
+
+			// Expect machine To get cordoned
+			cordonedNodes := []*v1.Node{}
+			nodes := env.EventuallyExpectNodeCount("==", int(numPods))
+			Eventually(func(g Gomega) {
+				cordonedNodes = lo.Filter(nodes, func(node *v1.Node, _ int) bool {
+					g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(node), node))
```
jonathan-innis, line 381: This is going to replace the data that the thing is pointed to which seems problematic if this `Get` call fails. You need a local copy of the node to ensure that you don't overwrite the data that you want to continually loop on

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+				for _, machine := range machines {
+					g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(machine), machine)).To(Succeed())
+					g.Expect(machine.StatusConditions().GetCondition(v1alpha5.MachineDrifted).IsTrue()).To(BeTrue())
+				}
+			}).Should(Succeed())
+
+			// Expect machine To get cordoned
+			cordonedNodes := []*v1.Node{}
+			nodes := env.EventuallyExpectNodeCount("==", int(numPods))
+			Eventually(func(g Gomega) {
```
jonathan-innis, line 379: It _might_ be worth make this a function like `EventuallyExpectCordonedNodeCount` and then return the cordoned nodes once that function succeeds. Particularly if we are using this in multiple places like we appear to be across all of these functions

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+					return node.Spec.Unschedulable == true
+				})
+				g.Expect(len(cordonedNodes)).To(BeNumerically("==", 1))
+			}).Should(Succeed())
+
+			// Drift should fail and original node should be uncordoned
+			// TODO: reduce timeouts when deprovisioning waits are factored out
+			machineList := &v1alpha5.MachineList{}
+			Eventually(func(g Gomega) {
+				g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(cordonedNodes[0]), cordonedNodes[0]))
```
jonathan-innis, line 391: This is another case where you need a local copy. If this call ever fails, you risk mutating the data that you are trying to use as your static set of data with the dynamic set of data that you are testing against.

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+				g.Expect(cordonedNodes[0].Spec.Unschedulable).To(BeFalse())
+				g.Expect(env.Client.List(env, machineList, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
+				g.Expect(len(machineList.Items)).To(BeNumerically("==", int(numPods)))
+			}).WithTimeout(20 * time.Minute).Should(Succeed())
+			env.ConsistentlyExpectNodeCount(">=", int(numPods), "20s")
```
jonathan-innis, line 396: Checking the node count doesn't do much since this test already assumes that the node never registers. We should check the Machine count.

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+					g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(node), node))
+					return node.Spec.Unschedulable == true
+				})
+				g.Expect(len(cordonedNodes)).To(BeNumerically("==", 1))
+			}).Should(Succeed())
+
+			// Drift should fail and original node should be uncordoned
+			// TODO: reduce timeouts when deprovisioning waits are factored out
+			machineList := &v1alpha5.MachineList{}
+			Eventually(func(g Gomega) {
```
jonathan-innis, line 390: These are two separate checks that can be separated out, right? One check is going to validate that we uncordon, the other is going to validate that we deprovision the old machine

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,115 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+				g.Expect(len(cordonedNodes)).To(BeNumerically("==", 1))
+			}).Should(Succeed())
+
+			// Drift should fail and original node should be uncordoned
+			// TODO: reduce timeouts when deprovisioning waits are factored out
+			Eventually(func(g Gomega) {
+				g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(cordonedNodes[0]), cordonedNodes[0]))
+				g.Expect(cordonedNodes[0].Spec.Unschedulable).To(BeFalse())
+			}).WithTimeout(15 * time.Minute).Should(Succeed())
+			env.ConsistentlyExpectNodeCount("==", 3, "20s")
```
jonathan-innis, line 446: Same comments as on the other test. We should be validating the Machine count, not the node count to ensure that we are not continuing to provision or scale-down.

test/suites/drift/suite_test.go
engedaam: Found a trail in the Get call showing that the client will not update the object on an error.

- https://github.com/kubernetes-sigs/controller-runtime/blob/c20ea143a236a34fb331e6c04820b75aac444e7d/pkg/client/client.go#L343

- https://github.com/kubernetes-sigs/controller-runtime/blob/c20ea143a236a34fb331e6c04820b75aac444e7d/pkg/client/typed_client.go#L136

- https://github.com/kubernetes/client-go/blob/6ab07f775da9a5a8cf441431c1651b6952996495/rest/request.go#L1319

test/suites/drift/suite_test.go
```diff
@@ -337,6 +337,121 @@ var _ = Describe("Drift", Label("AWS"), func() {
...
+			}).WithTimeout(11 * time.Minute).Should(Succeed())
+
+			endMachineState := &v1alpha5.MachineList{}
+			Eventually(func(g Gomega) {
+				g.Expect(env.Client.List(env, endMachineState, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
+				g.Expect(len(endMachineState.Items)).To(BeNumerically("==", int(numPods)))
+			}).WithTimeout(6 * time.Minute).Should(Succeed())
+
+			Consistently(func(g Gomega) {
+				g.Expect(lo.EveryBy(statingMachineState, func(sm *v1alpha5.Machine) bool {
```
jonathan-innis, line 393: How about a List() call to reduce the number of Gets here

jonathan-innis (approved):
Lgtm

engedaam (commented):
/karpenter snapshot

PR #4499, opened 2023-08-24
chore: stop passing --container-runtime to EKS bootstrap for K8s >= 1.27

### LINKED ISSUE #4241: Stop passing --container-runtime in the bootstrap for K8s >= 1.27
### Description

**Observed Behavior**:

We pass the container runtime flag to the bootstrap script.

**Expected Behavior**:

We shouldn't pass it at all.  It currently works as the bootstrap ignores it, but we should remove it now to prevent problems in the future.

**Versions**:
- Chart Version: 
- Kubernetes Version (`kubectl version`): 

jmdeal:
Fixes #4241

**Description**
No longer passes the `--container-runtime` flag to the EKS bootstrap script when the kubernetes server version is >= 1.27. The flag is currently ignored and is going to be removed in a future release.

`make test`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/amifamily/resolver.go
+++ b/pkg/providers/amifamily/resolver.go
@@ -17,6 +17,7 @@ package amifamily
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/version"
 	"net"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -74,7 +75,7 @@ type LaunchTemplate struct {
 // AMIFamily can be implemented to override the default logic for generating dynamic launch template parameters
 type AMIFamily interface {
 	DefaultAMIs(version string) []DefaultAMIOutput
-	UserData(kubeletConfig *v1alpha5.KubeletConfiguration, taints []core.Taint, labels map[string]string, caBundle *string, instanceTypes []*cloudprovider.InstanceType, customUserData *string) bootstrap.Bootstrapper
+	UserData(kubeletConfig *v1alpha5.KubeletConfiguration, kubeServerVersion *version.Version, taints []core.Taint, labels map[string]string, caBundle *string, instanceTypes []*cloudprovider.InstanceType, customUserData *string) bootstrap.Bootstrapper
 	DefaultBlockDeviceMappings() []*v1alpha1.BlockDeviceMapping
 	DefaultMetadataOptions() *v1alpha1.MetadataOptions
 	EphemeralBlockDevice() *string
@@ -128,6 +129,11 @@ func (r Resolver) Resolve(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTem
 	if len(mappedAMIs) == 0 {
 		return nil, fmt.Errorf("no instance types satisfy requirements of amis %v,", amis)
 	}
+	kubeVersionStr, err := r.amiProvider.KubeServerVersion(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("getting kubernetes version %w", err)
+	}
+	kubeVersion := version.MustParseGeneric(kubeVersionStr)
 	var resolvedTemplates []*LaunchTemplate
 	for amiID, instanceTypes := range mappedAMIs {
 		maxPodsToInstanceTypes := lo.GroupBy(instanceTypes, func(instanceType *cloudprovider.InstanceType) int {
@@ -150,6 +156,7 @@ func (r Resolver) Resolve(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTem
 				Options: options,
 				UserData: amiFamily.UserData(
 					r.defaultClusterDNS(options, kubeletConfig),
+					kubeVersion,
 					append(machine.Spec.Taints, machine.Spec.StartupTaints...),
 					options.Labels,
 					options.CABundle,
--- a/pkg/providers/amifamily/resolver.go
+++ b/pkg/providers/amifamily/resolver.go
@@ -124,13 +124,9 @@ func (r Resolver) Resolve(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTem
 		return nil, err
 	}
 
-	kubeVersionStr, err := r.amiProvider.KubeServerVersion(ctx)
+	kubeVersion, err := r.amiProvider.KubeServerVersion(ctx)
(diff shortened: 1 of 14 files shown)


pkg/providers/amifamily/resolver.go
```diff
@@ -117,17 +119,20 @@ func New(amiProvider *Provider) *Resolver {
...
 	amiFamily := GetAMIFamily(nodeTemplate.Spec.AMIFamily, options)
-	amis, err := r.amiProvider.Get(ctx, nodeTemplate, options)
+	mappedAMIs, err := r.getMappedAMIs(ctx, nodeTemplate, instanceTypes, options)
 	if err != nil {
 		return nil, err
 	}
-	if len(amis) == 0 {
-		return nil, fmt.Errorf("no amis exist given constraints")
+
+	kubeVersionStr, err := r.amiProvider.KubeServerVersion(ctx)
```
jonathan-innis, line 127: `KubeServerVersion` should already have these structured values from within the `Discover().ServerVersion()` call. I wonder if we can just always return the structured version and then convert it to a string when it's needed elsewhere

pkg/providers/amifamily/resolver.go
```diff
@@ -117,17 +119,20 @@ func New(amiProvider *Provider) *Resolver {
 // Multiple ResolvedTemplates are returned based on the instanceTypes passed in to support special AMIs for certain instance types like GPUs.
 func (r Resolver) Resolve(ctx context.Context, nodeTemplate *v1alpha1.AWSNodeTemplate, machine *v1alpha5.Machine, instanceTypes []*cloudprovider.InstanceType, options *Options) ([]*LaunchTemplate, error) {
 	amiFamily := GetAMIFamily(nodeTemplate.Spec.AMIFamily, options)
-	amis, err := r.amiProvider.Get(ctx, nodeTemplate, options)
+	mappedAMIs, err := r.getMappedAMIs(ctx, nodeTemplate, instanceTypes, options)
```
jonathan-innis, line 122: Really nice code simplification 🎉  I'd still recommend that we do this change separately from this PR so that all the code changes make sense in context. I'm not sure this change was needed in order to get the bug fix in.

pkg/providers/launchtemplate/testdata/al2_userdata_merged.golden
```diff
@@ -12,8 +12,7 @@ Content-Type: text/x-shellscript; charset="us-ascii"
 
 #!/bin/bash -xe
```
jonathan-innis, line 13: I could go either way but is it worth just creating an entirely separate file for the new version of the generated userData so that a reader doesn't have to reason about more template params

pkg/providers/amifamily/ubuntu.go
```diff
@@ -52,7 +54,7 @@ func (u Ubuntu) DefaultAMIs(version string) []DefaultAMIOutput {
 }
 
 // UserData returns the default userdata script for the AMI Family
-func (u Ubuntu) UserData(kubeletConfig *v1alpha5.KubeletConfiguration, taints []v1.Taint, labels map[string]string, caBundle *string, _ []*cloudprovider.InstanceType, customUserData *string) bootstrap.Bootstrapper {
+func (u Ubuntu) UserData(kubeletConfig *v1alpha5.KubeletConfiguration, _ *version.Version, taints []v1.Taint, labels map[string]string, caBundle *string, _ []*cloudprovider.InstanceType, customUserData *string) bootstrap.Bootstrapper {
```
jonathan-innis, line 57: AFAIK Ubuntu uses the same eks-bootstrap script that contains the removal of the `--container-runtime` parameter so we should check if this is also something that we should remove from this AMIFamily as well.

jonathan-innis (commented):
Great work 🎉  I think that we should just generally to a pass across all AMIFamilies to make sure to check if this change applies to any other AMIs besides AL2 starting on 1.27.x

pkg/providers/amifamily/ubuntu.go
jmdeal: It didn't look like any AMI types other than AL2 did, I can take a look and make sure though. The reason I didn't remove it for AMIFamily as a whole is `dockerd` is still supported by K8s 1.23.

pkg/providers/amifamily/resolver.go
jmdeal: I agree it doesn't really make sense in this PR. I included it because I was failing tests due to function complexity and this change got it through. I can try getting the structured version of the kubeVersion directly and see if that makes a difference.

pkg/providers/launchtemplate/testdata/al2_userdata_merged.golden
jmdeal: I was on the fence about this as well.  I liked this approach since it could be expanded to deal with other version dependent / optional arguments without needing a file for each combination. Since that's not the case now though I'd agree that multiple files may be the better way to go from a readability standpoint.

pkg/providers/amifamily/resolver.go
jonathan-innis: Ah, that makes more sense 👍 If that's blocking, this is fine to me

jonathan-innis: Might be helpful in the future to self-comment this on the PR so the change is well-understood

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jonathan-innis:
Discussed offline: This parameter is only being passed through to the AL2 AMIFamily right now. That AMIFamily currently has correct handling for not passing the `--container-runtime` flag to the `kubelet` binary unless the K8s version is below 1.27. In the case that the K8s version _is_ below 1.27, then the AL2 bootstrap script will only pass through `--container-runtime=remote`. 

Ref: https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L593

We should just be able to outright remove this field from the Provisioner `KubeletConfiguration` when the min K8s version that EKS supports is bumped to 1.24, at which point `containerd` will be the only supported container runtime and it doesn't make sense to allow users to pass this through.

PR #4504, opened 2023-08-26
feat: template custom userdata

ekristen:
Fixes #N/A 

**Description**

Allows Custom User Data for amiFamily Custom to be templated. This basically allows labels and taints to be templated into the user data which is useful for custom deployments. For example I am using RKE1 (from Rancher) with Rancher on EC2 instances and bootstrap the ec2 nodes using custom data.

Example template (pseudo code, not full)

```console
sudo docker run -d --privileged --restart=unless-stopped --net=host -v /etc/kubernetes:/etc/kubernetes -v /var/run:/var/run rancher/rancher-agent:v2.6.9 \
      --server https://rancher2 \
      --token token \
      --ca-checksum checksum \
      {{ with .Labels }}{{ range $k, $v := . }}--label {{ $k }}={{ $v }} {{ end }} \{{ end }}
      --worker
```

Deployment into clusters, struggled to get the test suite working with lack of docs (or couldn't locate it)

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/amifamily/bootstrap/custom.go
+++ b/pkg/providers/amifamily/bootstrap/custom.go
@@ -27,9 +27,8 @@ type Custom struct {
 }
 
 type TemplateData struct {
-	Taints        []v1.Taint        `hash:"set"`
-	Labels        map[string]string `hash:"set"`
-	InstanceTypes []string
+	Taints []v1.Taint        `hash:"set"`
+	Labels map[string]string `hash:"set"`
 }
 
 func (e Custom) Script() (string, error) {
@@ -48,9 +47,8 @@ func (e Custom) templateUserData(rawUserData string) (string, error) {
 	}
 
 	data := TemplateData{
-		Taints:        e.Options.Taints,
-		Labels:        e.Options.Labels,
-		InstanceTypes: e.Options.InstanceTypes,
+		Taints: e.Options.Taints,
+		Labels: e.Options.Labels,
 	}
 
 	var buf bytes.Buffer
--- a/pkg/providers/amifamily/bootstrap/custom.go
+++ b/pkg/providers/amifamily/bootstrap/custom.go
@@ -15,15 +15,49 @@ limitations under the License.
 package bootstrap
 
 import (
+	"bytes"
 	"encoding/base64"
-
-	"github.com/aws/aws-sdk-go/aws"
+	"github.com/samber/lo"
+	v1 "k8s.io/api/core/v1"
+	"text/template"
 )
 
 type Custom struct {
 	Options
 }
 
+type TemplateData struct {
+	Taints        []v1.Taint        `hash:"set"`
+	Labels        map[string]string `hash:"set"`
+	InstanceTypes []string
+}
+
 func (e Custom) Script() (string, error) {
-	return base64.StdEncoding.EncodeToString([]byte(aws.StringValue(e.Options.CustomUserData))), nil
+	userData, err := e.templateUserData(lo.FromPtr(e.Options.CustomUserData))
+	if err != nil {
+		return "", err
+	}
+
+	return base64.StdEncoding.EncodeToString([]byte(userData)), nil
+}
+
+func (e Custom) templateUserData(rawUserData string) (string, error) {
+	tmpl, err := template.New("custom").Parse(rawUserData)
+	if err != nil {
+		return "", err
+	}
+
+	data := TemplateData{
+		Taints:        e.Options.Taints,
+		Labels:        e.Options.Labels,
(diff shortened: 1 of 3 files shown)


pkg/providers/amifamily/bootstrap/custom.go
```diff
@@ -15,15 +15,47 @@ limitations under the License.
...
+	"github.com/samber/lo"
+	v1 "k8s.io/api/core/v1"
+	"text/template"
 )
 
 type Custom struct {
 	Options
 }
 
+type TemplateData struct {
```
jonathan-innis, line 29: This is really cool! I'm curious on a couple of things:
1. Do you think that this kind of templating might be useful for other AMIFamilies within Karpenter?
2. Karpenter currently has a "node registration flow" which adds all of the labels that Karpenter is tracking on its Machine CustomResource onto the Node when the Node joins the cluster. Is this mechanism sufficient to get the labels onto the node or do you also need these labels passed through into the userData?

ekristen: It could be useful for sure. It all depends on how each registers though. I’m not as familiar with the others.

ekristen: Looking at like bottlerocket, it wouldn't be useful because it's already passing in taints and labels via it's special toml configuration https://github.com/aws/karpenter/blob/main/pkg/providers/amifamily/bootstrap/bottlerocket.go#L47-L49

So I feel it's likely to be implementation specific.

jonathan-innis: How do you think this templating plays with the ideas presented here: https://github.com/aws/karpenter/issues/4459? One thought that we had was to drop a file onto the filesystem that you could modify and pull data from rather than having to template the userData like this.

ekristen: I think that's a different problem entirely as it assumes EKS. This simple just makes the data available to the custom user data. At least with the custom user data it's expected to be passed through the CRD definition. For the others the file already exists or is generated it seems depending on the family.

ellistarn: On first glance, this is a brilliant idea imo. It gives the user access to karpenters state (runtime interface) at launch time, without assuming anything about the file system. It's even user data format agnostic, i.e. could work with bash or toml. 

To Jonathan's point, I want us to be very careful about what exactly we expose in this template, as it will become something we support forever. 

Specifically on labels and taints, we've been hoping to shift label and taints registration to the controller, instead of the bootstrap. Does that mean you should be able to know the labels at bootstrap time? I don't know -- maybe.

ekristen:
@ellistarn let me know. it would be nice to have.

ellistarn:
@jonathan-innis to take the lead here

pkg/providers/amifamily/bootstrap/custom.go
jonathan-innis: To Ellis's point @ekristen, do you need these taints and labels added in the userData script if they are added to the Node at node-join time when the the node first starts up?

ekristen: Ah. I guess in my initial testing this wasn't happening but for different reasons at least labels work. I'll need to verify taints.

I don't use cloud provider operators so I had to write a shim to get the cloud provider id to be set so it would link up to karpenter.

jonathan-innis: > I don't use cloud provider operators so I had to write a shim to get the cloud provider id to be set so it would link up to karpenter. 

I could still see this being _extremely_ useful for things like the kubeletConfiguration if you are trying to inject the kubelet args that Karpenter is aware of into your userData script. I think, in general, I'm supportive of this templating mechanism since we can expand it out to other areas of Karpenter configuration that might affect userData.

pkg/providers/amifamily/bootstrap/custom.go
```diff
@@ -15,15 +15,47 @@ limitations under the License.
...
+	Taints []v1.Taint        `hash:"set"`
```
jonathan-innis, line 30: Sanity check, what are the `hash` annotations here for?

pkg/providers/amifamily/bootstrap/custom.go
```diff
@@ -15,15 +15,47 @@ limitations under the License.
 package bootstrap
```
jonathan-innis, line 15: Can we add some test cases into the `User Data` section of the launch template file (see https://github.com/aws/karpenter/blob/main/pkg/providers/launchtemplate/suite_test.go#L842)

PR #4509, opened 2023-08-28
chore: Add validation to ensure `id` field is mutually exclusive in selectors

jonathan-innis:
Fixes #N/A 

**Description**

This PR adds validation to ensure that the `id` field that can be specified in selectors through `aws-ids` and `aws::ids` is mutually exclusive with other fields that can be specified in the selector like `tags`, `owners`, and `names`

`make presubmit`
`/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/apis/v1alpha1/awsnodetemplate_validation.go
+++ b/pkg/apis/v1alpha1/awsnodetemplate_validation.go
@@ -78,6 +78,7 @@ func (a *AWSNodeTemplateSpec) validateAMIFamily() (errs *apis.FieldError) {
 	return errs
 }
 
+//nolint:gocyclo
 func (a *AWSNodeTemplateSpec) validateAMISelector() (errs *apis.FieldError) {
 	if a.AMISelector == nil {
 		return nil
@@ -85,11 +86,13 @@ func (a *AWSNodeTemplateSpec) validateAMISelector() (errs *apis.FieldError) {
 	if a.LaunchTemplateName != nil {
 		errs = errs.Also(apis.ErrMultipleOneOf(amiSelectorPath, launchTemplatePath))
 	}
+	var idFilterKeyUsed string
 	for key, value := range a.AMISelector {
 		if key == "" || value == "" {
 			errs = errs.Also(apis.ErrInvalidValue("\"\"", fmt.Sprintf("%s['%s']", amiSelectorPath, key)))
 		}
-		if key == "aws-ids" {
+		if key == "aws-ids" || key == "aws::ids" {
+			idFilterKeyUsed = key
 			for _, amiID := range functional.SplitCommaSeparatedString(value) {
 				if !amiRegex.MatchString(amiID) {
 					fieldValue := fmt.Sprintf("\"%s\"", amiID)
@@ -98,6 +101,9 @@ func (a *AWSNodeTemplateSpec) validateAMISelector() (errs *apis.FieldError) {
 				}
 			}
 		}
+	}
+	if idFilterKeyUsed != "" && len(a.AMISelector) > 1 {
+		errs = errs.Also(apis.ErrGeneric(fmt.Sprintf("%q filter is mutually exclusive, cannot be set with a combination of other filters in", idFilterKeyUsed), amiSelectorPath))
 	}
 	return errs
 }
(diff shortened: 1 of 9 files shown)


jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

pkg/apis/v1beta1/nodeclass_validation.go
```diff
@@ -112,10 +119,16 @@ func (in *NodeClassSpec) validateAMISelectorTerms() (errs *apis.FieldError) {
...
 
+//nolint:gocyclo
 func (in *AMISelectorTerm) validate() (errs *apis.FieldError) {
 	errs = errs.Also(validateTags(in.Tags).ViaField("tags"))
 	if len(in.Tags) == 0 && in.ID == "" && in.Name == "" && in.SSM == "" {
 		errs = errs.Also(apis.ErrGeneric("expect at least one, got none", "tags", "id", "name", "ssm"))
+	} else if len(in.Tags) > 0 && in.ID != "" ||
+		in.Name != "" && in.ID != "" ||
+		in.SSM != "" && in.ID != "" ||
+		in.Owner != "" && in.ID != "" {
```
njtran, line 130: I think this would look cleaner:
```suggestion
	} else if in.ID != "" && (len(in.Tags) > 0 ||
		in.Name != "" ||
		in.SSM != "" ||
		in.Owner != "") {
```

pkg/apis/v1beta1/nodeclass_validation.go
```diff
@@ -97,10 +99,15 @@ func (in *NodeClassSpec) validateSecurityGroupSelectorTerms() (errs *apis.FieldE
...
 
+//nolint:gocyclo
 func (in *SecurityGroupSelectorTerm) validate() (errs *apis.FieldError) {
 	errs = errs.Also(validateTags(in.Tags).ViaField("tags"))
 	if len(in.Tags) == 0 && in.ID == "" && in.Name == "" {
 		errs = errs.Also(apis.ErrGeneric("expect at least one, got none", "tags", "id", "name"))
+	} else if len(in.Tags) > 0 && in.ID != "" || in.Name != "" && in.ID != "" {
+		errs = errs.Also(apis.ErrGeneric(`"id" is mutually exclusive, cannot be set with a combination of other fields in`))
+	} else if len(in.Tags) > 0 && in.Name != "" || in.ID != "" && in.Name != "" {
+		errs = errs.Also(apis.ErrGeneric(`"name" is mutually exclusive, cannot be set with a combination of other fields in`))
```
njtran, line 110: I think it's cleaner to start with the common condition and to || the differences. If not, I think it's worth making a new block:
```suggestion
	} else if in.ID != "" && (len(in.Tags) > 0 || in.Name != "") {
		errs = errs.Also(apis.ErrGeneric(`"id" is mutually exclusive, cannot be set with a combination of other fields in`))
	} else if in.Name != "" && (len(in.Tags) > 0 || in.ID != "") {
		errs = errs.Also(apis.ErrGeneric(`"name" is mutually exclusive, cannot be set with a combination of other fields in`))
```

or 

```suggestion
	} else if in.ID != "" {
	     if len(in.Tags) > 0 || in.Name != "" {
		    errs = errs.Also(apis.ErrGeneric(`"id" is mutually exclusive, cannot be set with a combination of other fields in`))
	     }	
	} else if in.Name != "" {
	    if len(in.Tags) > 0 || in.ID != "" {
		errs = errs.Also(apis.ErrGeneric(`"name" is mutually exclusive, cannot be set with a combination of other fields in`))
		}
```

pkg/apis/v1beta1/nodeclass_validation.go
jonathan-innis: Good call 👍

pkg/apis/v1alpha1/awsnodetemplate_validation.go
```diff
@@ -78,18 +78,21 @@ func (a *AWSNodeTemplateSpec) validateAMIFamily() (errs *apis.FieldError) {
 	return errs
 }
 
+//nolint:gocyclo
```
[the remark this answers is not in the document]
jonathan-innis: It is, yes. I agree that this isn't a long function but it was getting flagged nonetheless

njtran (approved):

PR #4512, opened 2023-08-29
fix: emit events when node template fails to resolve

njtran:
Fixes #N/A 

**Description**
When a NodeTemplate fails to resolve in the cloudprovider code for a NotFoundError, Karpenter ignores it. Since this may happen very quickly and will not resolve itself, Karpenter now emits an event for this. 

Additionally, this change will not include cloudprovider.InstanceTypes() if they have no offerings. In the case of `karpenter-core`'s scheduling and deprovisioning logic, this means that Karpenter will consider less instance types in some cases when the zones fail to resolve. 

This is also the buddy change for https://github.com/aws/karpenter-core/pull/491

- make presubmit && snapshot tests.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/PuerkitoBio/goquery v1.8.1
 	github.com/avast/retry-go v3.0.0+incompatible
 	github.com/aws/aws-sdk-go v1.44.328
-	github.com/aws/karpenter-core v0.30.0-rc.0.0.20230824190041-a4c05cede32f
+	github.com/aws/karpenter-core v0.30.0-rc.0.0.20230829175231-8ad12e5df1cb
 	github.com/imdario/mergo v0.3.16
 	github.com/mitchellh/hashstructure/v2 v2.0.2
 	github.com/onsi/ginkgo/v2 v2.11.0
@@ -107,5 +107,3 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
-
-replace github.com/aws/karpenter-core => github.com/njtran/karpenter-core v0.0.0-20230829002458-d8c86f22e7e6
--- a/go.mod
+++ b/go.mod
@@ -107,3 +107,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/aws/karpenter-core => /Users/nichotr/workplace/go/src/github.com/aws/karpenter-core
--- a/go.mod
+++ b/go.mod
@@ -108,4 +108,4 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace github.com/aws/karpenter-core => /Users/nichotr/workplace/go/src/github.com/aws/karpenter-core
+replace github.com/aws/karpenter-core => github.com/njtran/karpenter-core v0.0.0-20230829002458-d8c86f22e7e6
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
+	"github.com/aws/karpenter-core/pkg/events"
 	"github.com/aws/karpenter-core/pkg/utils/functional"
 	"github.com/aws/karpenter/pkg/apis"
 	"github.com/aws/karpenter/pkg/apis/v1alpha1"
@@ -40,6 +41,7 @@ import (
 	"knative.dev/pkg/logging"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	cloudproviderevents "github.com/aws/karpenter/pkg/cloudprovider/events"
 	"github.com/aws/karpenter/pkg/providers/amifamily"
 	"github.com/aws/karpenter/pkg/providers/instance"
 	"github.com/aws/karpenter/pkg/providers/instancetype"
@@ -65,9 +67,10 @@ type CloudProvider struct {
 	amiProvider           *amifamily.Provider
 	securityGroupProvider *securitygroup.Provider
 	subnetProvider        *subnet.Provider
+	recorder              events.Recorder
 }
 
-func New(instanceTypeProvider *instancetype.Provider, instanceProvider *instance.Provider,
+func New(instanceTypeProvider *instancetype.Provider, instanceProvider *instance.Provider, recorder events.Recorder,
 	kubeClient client.Client, amiProvider *amifamily.Provider, securityGroupProvider *securitygroup.Provider, subnetProvider *subnet.Provider) *CloudProvider {
 	return &CloudProvider{
 		instanceTypeProvider:  instanceTypeProvider,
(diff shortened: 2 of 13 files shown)


njtran (commented):
/karpenter snapshot

ellistarn:
Can we drive these events via status conditions?

go.mod
```diff
@@ -107,3 +107,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/aws/karpenter-core => github.com/njtran/karpenter-core v0.0.0-20230829002458-d8c86f22e7e6
```
jonathan-innis, line 111: Do you just want to bump directly since this update should be merged now?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -40,6 +41,7 @@ import (
 	"knative.dev/pkg/logging"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	cloudproviderevents "github.com/aws/karpenter/pkg/cloudprovider/events"
```
jonathan-innis, line 44: I've changed my mind a bit about the utility of this being in a separate package. What do you think about collapsing the events package into the same package and suffixing the function call with `Event`?

pkg/cloudprovider/events/events.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
+	"github.com/aws/karpenter-core/pkg/events"
+)
+
+func ProvisionerFailedToResolveNodeTemplate(provisioner *v1alpha5.Provisioner) events.Event {
+	return events.Event{
+		InvolvedObject: provisioner,
+		Type:           v1.EventTypeWarning,
+		Message:        "Failed to resolve AWSNodeTemplate",
```
jonathan-innis, line 28:
```suggestion
		Message:        "Failed resolving AWSNodeTemplate",
```

nit: IMO, this sounds more natural

pkg/cloudprovider/events/events.go
```diff
@@ -0,0 +1,40 @@
...
+		DedupeValues:   []string{provisioner.Name},
+	}
+}
+
+func MachineFailedToResolveNodeTemplate(machine *v1alpha5.Machine) events.Event {
+	return events.Event{
+		InvolvedObject: machine,
+		Type:           v1.EventTypeWarning,
+		Message:        "Failed to resolve AWSNodeTemplate",
```
jonathan-innis, line 37:
```suggestion
		Message:        "Failed resolving AWSNodeTemplate",
```

nit: IMO, this sounds more natural

pkg/cloudprovider/events/events.go
```diff
@@ -0,0 +1,40 @@
...
+		DedupeValues:   []string{provisioner.Name},
+	}
+}
+
+func MachineFailedToResolveNodeTemplate(machine *v1alpha5.Machine) events.Event {
+	return events.Event{
+		InvolvedObject: machine,
+		Type:           v1.EventTypeWarning,
```
jonathan-innis, line 36: Is this a Warning? I think this should probably be an `Error` event?

pkg/cloudprovider/events/events.go
```diff
@@ -0,0 +1,40 @@
...
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
+	"github.com/aws/karpenter-core/pkg/events"
+)
+
+func ProvisionerFailedToResolveNodeTemplate(provisioner *v1alpha5.Provisioner) events.Event {
+	return events.Event{
+		InvolvedObject: provisioner,
+		Type:           v1.EventTypeWarning,
```
jonathan-innis, line 27: Is this a Warning? I think this should probably be an Error event?

pkg/providers/instancetype/suite_test.go
```diff
@@ -1135,6 +1136,42 @@ var _ = Describe("Instance Types", func() {
 			}
 		})
 		It("shouldn't report more resources than are actually available on instances", func() {
+			awsEnv.EC2API.DescribeSubnetsOutput.Set(&ec2.DescribeSubnetsOutput{
```
jonathan-innis, line 1139: Why did you need to add this?

jonathan-innis (commented):
Nice work! Just a few thoughts

pkg/cloudprovider/cloudprovider.go
njtran: I'd like to follow the semantic of how this is already done, and then opt to re-factor it in a separate PR, if that's alright with you. Why have you changed your mind?

pkg/cloudprovider/events/events.go
njtran: There are only Warning and Normal events, right? https://pkg.go.dev/k8s.io/api@v0.26.6/core/v1#:~:text=const%20(%0A%09//%20Information%20only%20and%20will%20not%20cause%20any%20problems%0A%09EventTypeNormal%20string%20%3D%20%22Normal%22%0A%09//%20These%20events%20are%20to%20warn%20that%20something%20might%20go%20wrong%0A%09EventTypeWarning%20string%20%3D%20%22Warning%22%0A)

pkg/providers/instancetype/suite_test.go
njtran: Since we now reject and remove an instance types that have 0 offerings, we need to make sure the describe subnets returns a zone so that we can get an instance type that has at least one offering for the test below.

pkg/cloudprovider/cloudprovider.go
jonathan-innis: > Why have you changed your mind

Creates extra packages for almost no reason. We've already started moving to this semantic in other PRs. Fair point though, I'm fine taking this separately.

pkg/cloudprovider/events/events.go
```diff
@@ -0,0 +1,40 @@
...
+		DedupeValues:   []string{provisioner.Name},
+	}
+}
+
+func MachineFailedToResolveNodeTemplate(machine *v1alpha5.Machine) events.Event {
+	return events.Event{
+		InvolvedObject: machine,
+		Type:           v1.EventTypeWarning,
+		Message:        "Failed resolving AWSNodeTemplate",
+		DedupeValues:   []string{machine.Name},
```
jonathan-innis, line 38: Should this be the name or the UID?

pkg/cloudprovider/events/events.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
+	"github.com/aws/karpenter-core/pkg/events"
+)
+
+func ProvisionerFailedToResolveNodeTemplate(provisioner *v1alpha5.Provisioner) events.Event {
+	return events.Event{
+		InvolvedObject: provisioner,
+		Type:           v1.EventTypeWarning,
+		Message:        "Failed resolving AWSNodeTemplate",
+		DedupeValues:   []string{provisioner.Name},
```
jonathan-innis, line 29: Should this be the name or the UID?

jonathan-innis (approved):
LGTM 🚀

njtran (commented):
/karpenter snapshot

njtran:
@ellistarn Definitely. I would see this as a new controller, but think this is a great opportunity for more observability into healthiness of various Karpenter components. Made an issue for this: https://github.com/aws/karpenter-core/issues/493

Issue #4168, opened 2023-06-29
Leader election 


nalshamaajc:
### Description

**Observed Behavior**:
Randomly getting the below errors. The incident involved frequent deprovisioning and service outage noting that PDBs are set.
```
E0626 20:00:18.625437       1 leaderelection.go:367] Failed to update lock: Put "[https://172.20.0.1:443/apis/coordination.k8s.io/v1/namespaces/karpenter/leases/karpenter-leader-election](https://172.20.0.1/apis/coordination.k8s.io/v1/namespaces/karpenter/leases/karpenter-leader-election)": context deadline exceeded
E0626 19:38:37.928683       1 leaderelection.go:330] error retrieving resource lock karpenter/karpenter-leader-election: Get "[https://172.20.0.1:443/apis/coordination.k8s.io/v1/namespaces/karpenter/leases/karpenter-leader-election](https://172.20.0.1/apis/coordination.k8s.io/v1/namespaces/karpenter/leases/karpenter-leader-election)": net/http: TLS handshake timeout
E0626 19:35:58.115419       1 leaderelection.go:330] error retrieving resource lock karpenter/karpenter-leader-election: Get "[https://172.20.0.1:443/apis/coordination.k8s.io/v1/namespaces/karpenter/leases/karpenter-leader-election](https://172.20.0.1/apis/coordination.k8s.io/v1/namespaces/karpenter/leases/karpenter-leader-election)": http2: client connection lost
```
![Pasted Graphic 9](https://github.com/aws/karpenter/assets/114491638/1394e56b-1381-47bd-bd7d-19beaa2cbe26)

and another set of logs that are weird and are asking me to file a report

```
"level":"ERROR","time":"2023-06-29T14:15:21.662Z","logger":"controller.pod_state","message":"invariant violated registering host port usage, IP=0.0.0.0 Port=8125 Proto=UDP conflicts with existing HostPort configuration IP=0.0.0.0 Port=8125 Proto=UDP, please file an issue","commit":"d7e22b1-dirty","pod”:”xxxx/xxxx-djkh9"}

2
[...]

PR #4530, opened 2023-08-31
feat: Add FlowSchemas to the Karpenter helm chart

jonathan-innis:
Relates to #4168 

**Description**

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter/README.md.gotmpl
+++ b/charts/karpenter/README.md.gotmpl
@@ -15,7 +15,7 @@ You can follow the detailed installation instruction in the [documentation](http
 helm upgrade --install --namespace karpenter --create-namespace \
   karpenter oci://public.ecr.aws/karpenter/{{ template "chart.name" . }} \
   --version v{{ template "chart.version" . }} \
-  --set serviceAccount.annotations.eks\.amazonaws\.com/role-arn=${KARPENTER_IAM_ROLE_ARN} \
+  --set "serviceAccount.annotations.eks\.amazonaws\.com/role-arn=${KARPENTER_IAM_ROLE_ARN}" \
   --set settings.clusterName=${CLUSTER_NAME} \
   --set settings.clusterEndpoint=${CLUSTER_ENDPOINT} \
   --set settings.interruptionQueue=${CLUSTER_NAME} \
--- a/charts/karpenter/README.md.gotmpl
+++ b/charts/karpenter/README.md.gotmpl
@@ -12,10 +12,10 @@ For full Karpenter documentation please checkout [https://karpenter.sh](https://
 You can follow the detailed installation instruction in the [documentation](https://karpenter.sh/docs/getting-started/getting-started-with-karpenter/#install) which covers the Karpenter prerequisites and installation options. The outcome of these instructions should result in something like the following command.
 
 ```bash
-helm upgrade --install --namespace karpenter --create-namespace \
+helm upgrade --install --namespace kube-system --create-namespace \
   karpenter oci://public.ecr.aws/karpenter/{{ template "chart.name" . }} \
   --version v{{ template "chart.version" . }} \
-  --set "serviceAccount.annotations.eks\.amazonaws\.com/role-arn=${KARPENTER_IAM_ROLE_ARN}" \
+  --set-string serviceAccount.annotations.eks\.amazonaws\.com/role-arn=${KARPENTER_IAM_ROLE_ARN} \
   --set settings.clusterName=${CLUSTER_NAME} \
   --set settings.clusterEndpoint=${CLUSTER_ENDPOINT} \
   --set settings.interruptionQueue=${CLUSTER_NAME} \
--- a/charts/karpenter/templates/flowschema.yaml
+++ b/charts/karpenter/templates/flowschema.yaml
@@ -1,4 +1,4 @@
-{{ if .Values.flowSchema.leaderElection.create }}
+{{ if .Values.flowSchema.create }}
 ---
 apiVersion: flowcontrol.apiserver.k8s.io/v1beta3
 kind: FlowSchema
@@ -13,7 +13,7 @@ metadata:
 spec:
   distinguisherMethod:
     type: ByUser
-  matchingPrecedence: {{ .Values.flowSchema.leaderElection.matchingPrecedence }}
+  matchingPrecedence: 200
   priorityLevelConfiguration:
     name: leader-election
   rules:
@@ -33,8 +33,6 @@ spec:
           serviceAccount:
             name: {{ include "karpenter.serviceAccountName" . }}
             namespace: {{ .Release.Namespace }}
-{{- end -}}
-{{ if .Values.flowSchema.workload.create }}
 ---
 apiVersion: flowcontrol.apiserver.k8s.io/v1beta3
 kind: FlowSchema
@@ -49,7 +47,7 @@ metadata:
 spec:
   distinguisherMethod:
     type: ByUser
-  matchingPrecedence: {{ .Values.flowSchema.workload.matchingPrecedence }}
+  matchingPrecedence: 1000
   priorityLevelConfiguration:
     name: workload-high
   rules:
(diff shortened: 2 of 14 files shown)


charts/karpenter/templates/flowschema.yaml
```diff
@@ -0,0 +1,87 @@
...
+    name: leader-election
+  rules:
+    - resourceRules:
+        - apiGroups:
+            - ""
+          namespaces:
+            - '*'
+          resources:
+            - endpoints
+            - configmaps
```
sftim, line 27: Does Karpenter actually use these legacy leader election mechanisms?

jonathan-innis: Good callout 👍

jonathan-innis (commented):
/karpenter scale

jonathan-innis (commented):
/karpenter scale

jonathan-innis (commented):
/karpenter scale

paurosello:
Hello @jonathan-innis we are seeing issues with karpenter on big clusters too and are considering adding the FlowSchema, any reason not to merge this PR?

charts/karpenter/templates/flowschema.yaml
```diff
@@ -0,0 +1,76 @@
+{{ if .Values.flowSchema.leaderElection.create }}
+---
+apiVersion: flowcontrol.apiserver.k8s.io/v1beta3
+kind: FlowSchema
+metadata:
+  name: karpenter-leader-election
```
mengqiy, line 6: I don't think you need this one given upstream already has this flowSchema.

charts/karpenter/templates/flowschema.yaml
```diff
@@ -0,0 +1,76 @@
...
+  {{- with .Values.additionalAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  distinguisherMethod:
+    type: ByUser
+  matchingPrecedence: {{ .Values.flowSchema.workload.matchingPrecedence }}
+  priorityLevelConfiguration:
+    name: workload-high
```
mengqiy, line 54: This flowSchema makes sense. I'm supportive for this one.

charts/karpenter/templates/flowschema.yaml
mengqiy: This `karpenter-leader-election` makes sense now.

jonathan-innis:
> any reason not to merge this PR

We're doing a double-check on our side that this change won't have any negative effects on performance for other components on the cluster, specifically those in `workload-high` and in the `leader-election` PLC. From what I can tell, there shouldn't be an issue here.

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

charts/karpenter/templates/flowschema.yaml
```diff
@@ -0,0 +1,76 @@
...
+  labels:
+    {{- include "karpenter.labels" . | nindent 4 }}
+  {{- with .Values.additionalAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  distinguisherMethod:
+    type: ByUser
+  matchingPrecedence: {{ .Values.flowSchema.leaderElection.matchingPrecedence }}
```
mengqiy, line 16: Do you see a need from CX to customize this field?
If no, I'd suggest to use a fixed value. It can reduce the chance a user does something wrong with it.

charts/karpenter/templates/flowschema.yaml
```diff
@@ -0,0 +1,76 @@
...
+  labels:
+    {{- include "karpenter.labels" . | nindent 4 }}
+  {{- with .Values.additionalAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  distinguisherMethod:
+    type: ByUser
+  matchingPrecedence: {{ .Values.flowSchema.workload.matchingPrecedence }}
```
mengqiy, line 52: Same comment for this one

charts/karpenter/templates/flowschema.yaml
jonathan-innis: I think it makes sense for users to be able to customize this. Not every user may have the same precedence values so we shouldn't assume that they do, particularly if they aren't using a default EKS control plane

mengqiy:
LGTM
I'd defer the final approval to someone from the Karpenter team.

charts/karpenter/README.md.gotmpl
```diff
@@ -15,7 +15,7 @@ You can follow the detailed installation instruction in the [documentation](http
 helm upgrade --install --namespace karpenter --create-namespace \
   karpenter oci://public.ecr.aws/karpenter/{{ template "chart.name" . }} \
   --version v{{ template "chart.version" . }} \
-  --set serviceAccount.annotations.eks\.amazonaws\.com/role-arn=${KARPENTER_IAM_ROLE_ARN} \
+  --set "serviceAccount.annotations.eks\.amazonaws\.com/role-arn=${KARPENTER_IAM_ROLE_ARN}" \
```
njtran, line 18: Do we need this? I think this has always worked for me. 

If anything, would this work?
```suggestion
  --set-string serviceAccount.annotations.eks\.amazonaws\.com/role-arn=${KARPENTER_IAM_ROLE_ARN} \
```

charts/karpenter/templates/flowschema.yaml
```diff
@@ -0,0 +1,76 @@
+{{ if .Values.flowSchema.leaderElection.create }}
+---
+apiVersion: flowcontrol.apiserver.k8s.io/v1beta3
```
njtran, line 3: Should we create an issue to track this graduating to stable in v1.29?

njtran (approved):

charts/karpenter/README.md.gotmpl
jonathan-innis: Looks like this was updated here: #5025. I was trying to understand the case where this fails and couldn't come up with anything good. I don't want to revert it per-se since we merged the PR, but I'd like to understand when this fails with `--set` in general

charts/karpenter/templates/flowschema.yaml
jonathan-innis: Yep, good callout

charts/karpenter/README.md.gotmpl
jonathan-innis: I think we can try to hear back from the original contributor but I don't want to block this PR on this

charts/karpenter/templates/flowschema.yaml
jonathan-innis: Added #5060

charts/karpenter/README.md.gotmpl
jonathan-innis: From what I can understand, I'm assuming that it has to do with [globbing](https://www.shellcheck.net/wiki/SC2086) of the variable if it's not put into a string from the bash side.

jonathan-innis: See #5061

jonathan-innis:
Closing this one in favor of #5124. The new docs are now up for deploying these FlowSchemas through manual configuration: https://karpenter.sh/docs/getting-started/getting-started-with-karpenter/#preventing-apiserver-request-throttling. Our recommendation going forward from `v0.33.x` is going to be to install karpenter into the `kube-system` namesapce so that you get these FlowSchemas assigned to the correct PLCs for free.

We belive that Karpenter is a cluster-critical component and, as such, it should end up in the `kube-system` namespace along other cluster critical components. If you still want to install karpenter in a custom namespace (like `karpenter` which was previously recommended) you can still do so and just apply the FlowSchemas mentioned in the docs here https://karpenter.sh/docs/getting-started/getting-started-with-karpenter/#preventing-apiserver-request-throttling to enable the correct PLC mapping for Karpenter in that namespace.

PR #4531, opened 2023-09-01
fix: Pending Karpenter pods when less than 2 non-Karpenter nodes exist

integrii:
**Description**
This change downgrades the affinity rules that require Karpetner pods to provision to non-karpenter nodes from a requirement to a preference.  In some situations, there may be less than 2 karpenter provisioned nodes.  In this situation, one (or both) Karpenter pods will never be allowed to schedule, and autoscaling in the cluster will break.  With this change, Karpenter pods will schedule on non-karpenter nodes if they must.

We run a lot of Karpenter enabled EKS clusters.  Our EKS clusters only have one non-karpenter node (to bootstrap the cluster).  These clusters get stuck with one dead Karpenter pod because provisioning any node would not be sufficient for Karpetner to get scheduled.  The requirement was manually changed to a preference, and everything immediately scheduled as expected.

- [x] No

### DIFF
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -64,7 +64,7 @@ nodeSelector:
 # -- Affinity rules for scheduling the pod. If an explicit label selector is not provided for pod affinity or pod anti-affinity one will be created from the pod selector labels.
 affinity:
   nodeAffinity:
-    requiredDuringSchedulingIgnoredDuringExecution:
+    preferredDuringSchedulingIgnoredDuringExecution:
       nodeSelectorTerms:
         - matchExpressions:
             - key: karpenter.sh/provisioner-name


jonathan-innis:
> These clusters get stuck with one dead Karpenter pod because provisioning any node would not be sufficient for Karpetner to get scheduled

The value in the `values.yaml` file should be overridable to set a preferred affinity over the required affinity if this better fits your use-case. We've generally taken the stance that if you are running Karpenter in HA mode, then you want the pods to be able to tolerate an outage to a Node without facing any real impact so enforcing that the pods run on separate nodes feels like the right way to go for this critical cluster application. Thoughts?

integrii:
Thanks for the reply Jonathan! I'm not suggesting that we remove the pod anti-affinity that keeps Karpenter pods on different nodes.  I think that's a great idea.  This PR weakens the rule that Karpenter pods can never schedule onto karpenter provisioned nodes.  It would still be a requirement that they are on different nodes.

I suppose the questions are:
- Is there a reason that Karpenter pods should not run on Karpenter nodes? (such as insufficient permissions?)
- The value is currently overridable if you set the following `values.yaml` file (which was not easy to figure out).
- This may make this PR seem useless, but my suggestion is that unless there are permissions issues with Karpetner pods running on Karpenter nodes that I don't know about, this should be the default mode of operation for optimal durability and functionality.

```yaml
  affinity:
    nodeAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
          - matchExpressions:
                operator: DoesNotExist
      requiredDuringSchedulingIgnoredDuringExecution: null
```

jonathan-innis:
> this should be the default mode of operation for optimal durability

Our reasoning is precisely because of durability that we don't want Karpenter nodes running on Karpenter-provisioned capacity. Imagine there's a single Karpenter pod running on a Karpenter provisioned node and that node gets reclaimed or removed for some reason. A bunch of pods go pending (including the Karpenter pods) that were running on that Node, but the cluster can't be auto-scaled up because the Karpenter pod is not running on _any_ capacity, so you've effectively gotten yourself into a situation where Karpenter has deadlocked itself. You could extend this from one replica out to many replicas, but the same philosophy still holds: "we don't want something taking a dependency on itself". 

That exact philosophy is the reason why we have a hard stance against Karpenter running on Karpenter-provisioned nodes. They should run on some other capacity management solution so that the auto-scaler is resilient to failure.

integrii:
I understand your thinking.  However, in the situation where Karpenter pods are pending in a sea of other pending pods, Kubernetes will evict pods in order to make room for Karpenter.  This is because of the [preemption](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#preemption) functionality set for pods that have a priority class of `system-cluster-critical`, which has `preemptionPolicy: PreemptLowerPriority` set on it.

After preemption makes room on the remaining capacity in the cluster (for us, we run one non-karpenter node in all clusters), then Karpenter can scale out.  Once a second node exists, the change to this helm chart will allow the second karpetner node to get provisioned.

As long as `system-cluster-critical` is set on Karpenter pods, then it should never be an issue that the pods live on karpenter provisioned pods, unless for some reason the karpenter provisioned pods are configured with insufficient IAM permissions to run Karpenter.

jonathan-innis:
> As far as durability goes, I believe this is nothing but an improvement. With the current configuration, you could hypothetically have a deadlock in the other direction, where for some reason the ASG nodes are dead or broken, and Karpenter has no ability to run on karpenter-provisioned nodes - thus deadlocking the cluster autoscaling as well.

Yeah, honestly this idea is pretty cool! I think our general reservation about letting this be the default is mainly the concern that some users may not have a deep enough understanding when installing this component that they need the additional capacity to run Karpenter. Realistically, there probably aren't many users that are bootstrapping Karpenter, letting Karpenter run to launch capacity and then removing the original capacity, but setting this affinity to be preferred by default allows someone to do that without having any awareness that they shouldn't be doing that. 

I feel like if you understand what you are doing and you have a sophisticated use-case like you do, you can make the change in the helm chart by overriding the defaults, but we are trying to make sure that the default experience avoids as many pitfalls as possible.

integrii:
Personally, I don't think that people would often be bootstrapping their cluster and then removing their ASG later on, but I suppose it could happen and in that case things would take time to break, rather than breaking right away, which could potentially be more confusing. That default setting is your judgement call as a project maintainer.  Hopefully, AWS will add Karpenter to the EKS control plane and this concern can be a thing of the past!

I'll close my PR and just stick with downstream overrides.  Thanks for all your time thinking about this and writing me back cohesive logic.

PR #4541, opened 2023-09-01
chore: Add conversion from `v1beta1/NodeClass` to `v1alpha1/AWSNodeTemplate`

jonathan-innis:
Fixes #N/A 

**Description**

Adds conversion from a `v1beta1/NodeClass` to a `v1alpha5/AWSNodeTemplate` for the controller and hash annotation handling. This allows the business logic for the controllers and the CloudProvider to only be written with the `v1beta1` APIs in mind.

`make presubmit`

- [ ] Yes, issue opened: # 
- [x] No


pkg/test/expectations/expectations.go
```diff
@@ -0,0 +1,115 @@
...
+
+	. "github.com/onsi/gomega" //nolint:revive,stylecheck
+	"github.com/samber/lo"
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+	"github.com/aws/karpenter/pkg/apis/v1beta1"
+)
+
+func ExpectBlockDeviceMappingsEqual(bdm1 []*v1alpha1.BlockDeviceMapping, bdm2 []*v1beta1.BlockDeviceMapping) {
```
njtran, line 28: Just curious, should we be asserting anything about the ordering of this? Don't we consider root volumes/default volume size by the ordering in this list? Should we care about that?

pkg/utils/nodeclass/nodeclass.go
```diff
@@ -36,66 +37,79 @@ func New(nodeTemplate *v1alpha1.AWSNodeTemplate) *v1beta1.NodeClass {
...
+			Subnets:        NewSubnets(nodeTemplate.Status.Subnets),
+			SecurityGroups: NewSecurityGroups(nodeTemplate.Status.SecurityGroups),
+			AMIs:           NewAMIs(nodeTemplate.Status.AMIs),
 		},
 		IsNodeTemplate: true,
 	}
 }
 
-func NewSubnetSelectorTerms(subnetSelector map[string]string) []v1beta1.SubnetSelectorTerm {
+func NewSubnetSelectorTerms(subnetSelector map[string]string) (terms []v1beta1.SubnetSelectorTerm) {
```
njtran, line 65: nit: named return values

pkg/utils/nodeclass/nodeclass.go
```diff
@@ -36,66 +37,79 @@ func New(nodeTemplate *v1alpha1.AWSNodeTemplate) *v1beta1.NodeClass {
...
+		terms = append(terms, v1beta1.SubnetSelectorTerm{
+			Tags: tags,
+			ID:   id,
+		})
 	}
+	return terms
 }
 
-func NewSecurityGroupSelectorTerms(securityGroupSelector map[string]string) []v1beta1.SecurityGroupSelectorTerm {
+func NewSecurityGroupSelectorTerms(securityGroupSelector map[string]string) (terms []v1beta1.SecurityGroupSelectorTerm) {
```
njtran, line 90: Same here

pkg/utils/nodeclass/nodeclass.go
```diff
@@ -173,6 +196,43 @@ func NewMetadataOptions(mo *v1alpha1.MetadataOptions) *v1beta1.MetadataOptions {
 	}
 }
 
+func NewSubnets(subnets []v1alpha1.Subnet) []v1beta1.Subnet {
+	if subnets == nil {
```
njtran, line 200: Just sanity checking that lo.Map() doesn't take in a nil list?

pkg/test/expectations/expectations.go
jonathan-innis: Ah, that's a good point. Yes, we should be asserting the order here.

pkg/utils/nodeclass/nodeclass.go
jonathan-innis: I generally take the approach of naming the return parameters when the return value is a list and we are going to have to declare it in the body of the function anyways with a `var` statement

pkg/utils/nodeclass/nodeclass.go
jonathan-innis: I generally take the approach of naming the return parameters when the return value is a list and we are going to have to declare it in the body of the function anyways with a `var` statement so I personally think that this is ok here.

pkg/utils/nodeclass/nodeclass.go
njtran: Ahhh makes sense. Thanks!

njtran (approved):

PR #4546, opened 2023-09-03
perf: Reduce memory by using `DescribeImages` paginated call

jonathan-innis:
Fixes #N/A 

**Description**

This PR reduces the memory utilization from the `DescribeImages` call for calls that are wide-open and grab an extremely high number of images from the API >> 1000 images. This changes the call to be paginated and to cap the number of results in a page at 500 images. Each page only stores an image if the set of requirements for the image is different from the existing set of image requirements OR if the requirements already exist but the image is newer. This reduces the amount of things that are stored in the cache for a request.

`make presubmit`

For the following test `AWSNodeTemplate`, these are the memory footprints before and after the PR

```yaml
kind: AWSNodeTemplate
metadata:
  name: default
spec:
  subnetSelector:                             
    karpenter.sh/discovery: <>
  securityGroupSelector:          
    karpenter.sh/discovery: <>
  amiSelector:
    aws::owners: self,amazon
```

#### Before PR

Screenshot 2023-09-03 at 12 24 33 AM

#### After PR

Screenshot 2023-09-03 at 3 08 58 PM

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/cloudprovider/drift.go
+++ b/pkg/cloudprovider/drift.go
@@ -84,9 +84,9 @@ func (c *CloudProvider) isAMIDrifted(ctx context.Context, nodeClaim *corev1beta1
 	if len(amis) == 0 {
 		return "", fmt.Errorf("no amis exist given constraints")
 	}
-	mappedAMIs := amifamily.MapInstanceTypes(amis, []*cloudprovider.InstanceType{nodeInstanceType})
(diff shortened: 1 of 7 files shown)


jonathan-innis (commented):
/karpenter snapshot

bwagner5 (approved):
lgtm

Issue #3182, opened 2023-01-09
Add support for Host Resource Groups


cspargo:
### Tell us about your request

We use Dedicated Hosts provisioned with Host Resource Groups (a.k.a "host" tenancy).  These require some additional settings in the launch template data  to set the tenancy, the ARN of the host resource group and the ARN of the license configuration. These settings are mentioned [here](https://docs.aws.amazon.com/autoscaling/ec2/userguide/create-launch-template.html#advanced-settings-for-your-launch-template). 

In cloudformation, it would look like this
```
   Type: AWS::EC2::LaunchTemplate
    Properties:
      LaunchTemplateName: foo
      LaunchTemplateData:
        Placement:
          Tenancy: host
          HostResourceGroupArn: arn:aws:resource-groups:us-east-1:123456789012:group/my-hrg-name
        LicenseSpecifications:
          - LicenseConfigurationArn: arn:aws:license-manager:eu-east-1:123456789012:license-configuration:lic-edf7f9e241f5e16f29996c842111f448
```

It would mean adding full or partial support for these to the aws node template

- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-placement.html (only need "Tenancy" and "HostResourceGroupArn" for this use case)
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-licensespecification.html

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

We would like to be able to provision instances onto our dedicated hosts with Karpenter. This blocks us from using it.

There was a previous issue regarding dedicated hosts https://github.com/aws/karpenter/issues/2556. The recommendation there was to use custom launch templates, however I believe support for these is or will be removed.

### Are you currently working around this issue?

we don't have any workaround

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ellistarn:
Seems like a good feature! Happy to accept a contribution if you're interested.

cspargo:
sure, i'll put something together

mra89:
[...]

PR #4553, opened 2023-09-04
feat: dedicated hosts selectors

preflightsiren:
Fixes: #3182

**Description**
Wanted to create a draft PR to start a conversation about the approach. I believe following https://karpenter.sh/v0.27.3/contributing/design-guide/ that adding to the AWSNodeTemplate best follows "Identify an opinionated default that solves the majority of use cases."
Of course I'd appreciate any and all help to mould this design.

This design focuses on a simplified configuration, just the minimum needed to use host resource groups.

```
kind: AWSNodeTemplate
metadata:
  name: default
spec:
  subnetSelector: { ... }        # required, discovers tagged subnets to attach to instances
  securityGroupSelector: { ... } # required, discovers tagged security groups to attach to instances
  licenseSelector: 
    name: myLicense
  hostResourceGroup: 
    name: my-hrg-name
```

- [ ] Yes, issue opened: # 
- [ ] No


preflightsiren:
Created this to split away from https://github.com/aws/karpenter/pull/3859 since the approach has significantly changed from when it was first opened.

designs/dedicatedhosts.md
```diff
@@ -0,0 +1,107 @@
...
+kind: AWSNodeTemplate
+metadata:
+  name: default
+spec:
+  dedicatedHostConfig:
+    licenseConfiguration: arn:aws:license-manager:eu-east-1:123456789012:license-configuration:lic-edf7f9e241f5e16f29996c842111f448 # optional, arn of the license configuration
+    resourceHostGroup: arn:aws:resource-groups:us-east-1:123456789012:group/my-hrg-name #option, arn of the HRG 
+```
+
+4. Implement Selectors for all relevant fields
```
jonathan-innis, line 75: This looks really nice to me! You could open up the `hostResourceGroupSelector` as well to allow it to select on ID as well!

preflightsiren:
Functionality wise this change is pretty complete, it's all wired up through the nodeclass controller, and extends the amifamily resolver to dynamicaly generate the Launch Template. 

Todo:
- update the website docs
- fix/review the code comments
- clean up any unused references to licenProvider
- Add placement group provider

pkg/providers/license/license.go
```diff
@@ -0,0 +1,93 @@
...
+	return &Provider{
+		licensemanager: lmapi,
+		cm:             pretty.NewChangeMonitor(),
+		// TODO: Remove cache for v1beta1, utilize resolved subnet from the AWSNodeTemplate.status
+		// Subnets are sorted on AvailableIpAddressCount, descending order
+		cache: cache,
+	}
+}
+
+func (p *Provider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) ([]string, error) {
```
preflightsiren, line 51: rename to `Get`

pkg/providers/license/license.go
```diff
@@ -0,0 +1,93 @@
...
+	p.Lock()
+	defer p.Unlock()
+
+	// Get selectors from the nodeClass, exit if no selectors defined
+	selectors := nodeClass.Spec.LicenseSelectorTerms
+	if selectors == nil {
+		logging.FromContext(ctx).
```
preflightsiren, line 58: remove

pkg/utils/nodeclass/nodeclass.go
```diff
@@ -37,31 +37,53 @@ func New(nodeTemplate *v1alpha1.AWSNodeTemplate) *v1beta1.EC2NodeClass {
...
+	return []v1beta1.HostResourceGroupSelectorTerm{
+		{Name: hrgSelector["name"]},
+	}
+}
+
+func NewHostResourceGroup(hrgs []string) *v1beta1.HostResourceGroup {
+	if len(hrgs) == 0 {
+		return nil
+	}
+	return &v1beta1.HostResourceGroup{ARN: hrgs[0]}
```
preflightsiren, line 84: refactor

website/content/en/preview/concepts/node-templates.md
```diff
@@ -24,20 +24,27 @@ kind: AWSNodeTemplate
...
+  securityGroupSelector: { ... }     # required, discovers tagged security groups to attach to instances
+  instanceProfile: "..."             # optional, overrides the node's identity from global settings
+  amiFamily: "..."                   # optional, resolves a default ami and userdata
+  amiSelector: { ... }               # optional, discovers tagged amis to override the amiFamily's default
+  userData: "..."                    # optional, overrides autogenerated userdata with a merge semantic
+  tags: { ... }                      # optional, propagates tags to underlying EC2 resources
+  metadataOptions: { ... }           # optional, configures IMDS for the instance
+  blockDeviceMappings: [ ... ]       # optional, configures storage devices for the instance
+  detailedMonitoring: "..."          # optional, configures detailed monitoring for the instance
+  licenseSelector: { ... }           #optional, discovers resource group license configurations
```
preflightsiren, line 37: update docs

pkg/providers/hostresourcegroup/hostresourcegroup.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+func (p *Provider) Get(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (*v1beta1.HostResourceGroup, error) {
+	p.Lock()
+	defer p.Unlock()
+    var nextToken *string = nil
+	var groups []*resourcegroups.GroupIdentifier
+	for {
+		resp, err := p.resourcegroups.ListGroupsWithContext(ctx, &resourcegroups.ListGroupsInput{ NextToken: nextToken})
```
jonathan-innis, line 53: There's `ListGroupsPagesWithContext` function that you can use to get this `nextToken` iteration logic here for free

pkg/providers/instancetype/zz_generated.bandwidth.go
```diff
@@ -42,6 +42,7 @@ var (
 		"t4g.medium":        256,
 		"m7a.medium":        390,
 		"m7i-flex.large":    390,
+		"r7a.medium":        390,
```
jonathan-innis, line 45: I don't think that these changes should be included here. Can you do a `go checkout main <path>` on this to revert it? If we need to update these generated files, we can do so in a separate PR

pkg/providers/hostresourcegroup/hostresourcegroup.go
```diff
@@ -0,0 +1,94 @@
...
+		cache: cache,
+	}
+}
+
+func (p *Provider) Get(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (*v1beta1.HostResourceGroup, error) {
+	p.Lock()
+	defer p.Unlock()
+    var nextToken *string = nil
+	var groups []*resourcegroups.GroupIdentifier
+	for {
```
jonathan-innis, line 52: Can we consider filtering these out as we are going through the pages. That way, we are using less memory for what we are actually storing?

pkg/providers/pricing/zz_generated.pricing.go
```diff
@@ -18,9 +18,9 @@ package pricing
 
 import "time"
 
-// generated at 2023-09-11T13:06:38Z for us-east-1
+// generated at 2023-08-21T14:24:21Z for us-east-1
```
jonathan-innis, line 21: Same comment as the generated instance types. I think we can avoid running `make docgen` or `make codegen` for this PR

pkg/providers/placementgroup/placementgroup.go
```diff
@@ -0,0 +1,94 @@
...
+	return &Provider{
+		ec2api: ec2api,
+		cm:     pretty.NewChangeMonitor(),
+		// TODO: Remove cache for v1beta1, utilize resolved subnet from the AWSNodeTemplate.status
+		// Subnets are sorted on AvailableIpAddressCount, descending order
+		cache: cache,
+	}
+}
+
+func (p *Provider) Get(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (*ec2.PlacementGroup, error) {
```
jonathan-innis, line 51: From looking at the code, I see that we have a `placementGroupSelector` but we aren't actually using this in the launch template when we generate the placement. Should we be using this as well?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -262,6 +267,29 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, options *amifamily.
...
+        return nil
+    }
+    return &ec2.LaunchTemplatePlacementRequest{
+        HostResourceGroupArn: aws.String(placement.HostResourceGroup),
+    }
+
+}
+
+func generateLicenseSpecification(licenses []string) []*ec2.LaunchTemplateLicenseConfigurationRequest {
+    result := []*ec2.LaunchTemplateLicenseConfigurationRequest{}
```
jonathan-innis, line 281: Generally favor the `var` initialization of slices due to: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

jonathan-innis: This also means that you don't have to explicitly return nil when iterating since the var value will be `nil` by default

pkg/providers/hostresourcegroup/hostresourcegroup.go
```diff
@@ -0,0 +1,94 @@
...
+        for i := range resp.GroupIdentifiers {
+            groups = append(groups, resp.GroupIdentifiers[i])
+        }
+		if nextToken == nil {
+			break
+		}
+	}
+
+	// filter resp to only include those that match the hostResourceGroupSelector Name
+	for i := range groups {
```
jonathan-innis, line 70: Same comment on using pages for this rather than building out everything in memory and then iterating through it later. That'll help memory performance

pkg/providers/hostresourcegroup/hostresourcegroup.go
```diff
@@ -0,0 +1,94 @@
...
+		group := groups[i]
+		for x := range nodeClass.Spec.HostResourceGroupSelectorTerms {
+			selector := nodeClass.Spec.HostResourceGroupSelectorTerms[x]
+				logging.FromContext(ctx).
+					With("group", group).
+					With("selector", selector).
+					Debugf("checking for host resource group match")
+			if *group.GroupName == selector.Name {
+				match := &v1beta1.HostResourceGroup{ARN: *group.GroupArn, Name: *group.GroupName}
+				logging.FromContext(ctx).
```
jonathan-innis, line 80: We could maybe consider just putting the `group name` or the entire `ARN`? I'm not sure that printing out the entire struct here is neccessary

pkg/providers/hostresourcegroup/hostresourcegroup.go
```diff
@@ -0,0 +1,94 @@
...
+					With("Matched hrg", match).
+					Debugf("discovered host resource group configuration")
+
+				return match, nil
+			}
+		}
+	}
+	logging.FromContext(ctx).
```
jonathan-innis, line 88: IMO, we should only log here when we discover something. A NoOp for matching anything here should probably be implicit.

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -152,6 +176,30 @@ type AMISelectorTerm struct {
 	Owner string `json:"owner,omitempty"`
 }
 
+// LicenseSelectorTerm defines selection logic for a licenseConfigurationSpecification used to launch nodes
+// If multiple fields are used for selection, the requirements are ANDed.
+type LicenseSelectorTerm struct {
```
jonathan-innis, line 181: Should we consider doing ID or ARN for any of these? I'm curious generally if we are going to make all of these terms if there's a use-case for allowing the LaunchTemplate request from a set of options, rather than just specifying a single option. 

I'm not sure I have enough depth on the topic right now to make that call, though I'm happy to look into it a bit deeper if we need help making that call here

jonathan-innis (commented):
High-level question: We are thinking of supporting only v1beta1 and dropping support for v1alpha5 in the near-term. Does it make sense to just release these fields _only_ under the v1beta1 API, marking them as alpha fields so that we don't have to reason about the conversion logic in this PR. Up to this point, we've just been handling conversion logic for existing API and trying to push new features into the soon-to-be released beta API

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -38,6 +38,15 @@ type EC2NodeClassSpec struct {
 	// AMIFamily is the AMI family that instances use.
 	// +optional
 	AMIFamily *string `json:"amiFamily,omitempty"`
+	// LicenseSelectorTerms is a list of LicenseSelectors. The terms are ORed.
```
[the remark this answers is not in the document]
preflightsiren: I'm mostly following the example of the other selectors like subnet, amifamily and security groups. In these selectors they use an array of selectorTerms. License Configurations may contain multiple results, however the other (hrg / placement group) new selectors can only use a single result. 

The ami selector I feel behaves the same way (https://github.com/aws/karpenter/blob/7eb291579a7a813a5b70c5a541ab73d4f1c151bf/pkg/apis/v1beta1/ec2nodeclass.go#L37) where it accepts multiple `AMISelectorTerm`s however only a single AMI can be selected.

If you don't want to continue that structure, I'd be happy to change it, as I'm not wed to any particular direction.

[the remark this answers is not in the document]
preflightsiren: cool approach, I'm not sure how I would implement this - kubernetes appears to use a custom type `featureGate` to implement this annotation. I can't see anything equivalent within this project, only in karpenter-core.

Do you have an idea how we would implement feature gating this off? does it require coordinating changes with karpenter-core?

As far as documenting fields within the api as alpha, I'm totally fine adding any disclaimers around this. I have some intuition that these features will not be widely used.

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -38,6 +38,15 @@ type EC2NodeClassSpec struct {
...
+	// +optional
+	LicenseSelectorTerms []LicenseSelectorTerm `json:"licenseSelectorTerms,omitempty" hash:"ignore"`
+    // HostResourceGroupSelectorTerms is a list of HostResourceGroupSelectors. The terms are ORed.
```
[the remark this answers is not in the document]
preflightsiren: same as https://github.com/aws/karpenter/pull/4553/files#r1329558288

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -38,6 +38,15 @@ type EC2NodeClassSpec struct {
...
+    // +optional
+    HostResourceGroupSelectorTerms []HostResourceGroupSelectorTerm `json:"hostResourceGroupSelectorTerms,omitempty" hash:"ignore"`
+    // PlacementGroupSelectorTerms is a list of PlacementGroupSelector. The terms are ORed.
+    // +optional
+    PlacementGroupSelectorTerms []PlacementGroupSelectorTerm `json:"placementGroupSelectorTerms,omitempty" hash:"ignore"`
```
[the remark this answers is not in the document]
preflightsiren: same as https://github.com/aws/karpenter/pull/4553/files#r1329558288

pkg/apis/v1beta1/ec2nodeclass.go
preflightsiren: I've stayed away from IDs and ARNs as I feel they're counter to the ethos of the selector pattern. If we want different clusters / regions / accounts to be able to share the same configuration, name is the most portable selector.

My currently thinking is we should implement the selectors with name only first, and then we can easily extend them to select based upon different identifiers as needed.

pkg/apis/v1beta1/ec2nodeclass_status.go
```diff
@@ -63,4 +73,13 @@ type EC2NodeClassStatus struct {
 	// cluster under the AMI selectors.
 	// +optional
 	AMIs []AMI `json:"amis,omitempty"`
+    // Licenses contains the license arns
+    // +optional
+    Licenses []string `json:"licenses,omitempty"`
```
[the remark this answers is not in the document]
preflightsiren: The reason I've selected ARN is to highlight the aws account id that's the owner of the license. In our environment the cluster and the license owner are in different aws accounts. I don't know if that's a pattern we'll see in many other environments.

pkg/providers/instancetype/zz_generated.bandwidth.go
preflightsiren: reverted.

pkg/providers/hostresourcegroup/hostresourcegroup.go
preflightsiren: refactored and updated to return the first match.

pkg/providers/pricing/zz_generated.pricing.go
preflightsiren: reverted

pkg/providers/placementgroup/placementgroup.go
preflightsiren: `selectors := nodeClass.Spec.LicenseSelectorTerms` should have been `selectors := nodeClass.Spec.placementGroupSelector`.
fixed.

pkg/providers/amifamily/resolver.go
```diff
@@ -187,7 +183,7 @@ func (r Resolver) Resolve(ctx context.Context, nodeClass *v1beta1.EC2NodeClass,
 				DetailedMonitoring:  aws.BoolValue(nodeClass.Spec.DetailedMonitoring),
 				AMIID:               amiID,
 				InstanceTypes:       instanceTypes,
-				Licenses:            licenses,
+                Licenses:            nodeClass.Status.Licenses,
```
preflightsiren, line 186: @jonathan-innis What are your thoughts over this, should we use the resolved values from the nodeClass status? or load it again via the provider `r.licenseProvider.Get()`

preflightsiren:
Ok I think this PR is now essentially complete or very near to, and is ready for a thorough review.
I have addressed all the feedback so far and re-tested in our test clusters to ensure the selector logic works as expected.
I have added some test cases around the 3 different providers, updating all the test infrastructure to load fake APIs for the aws services, and plumb them into the providers should anyone want to expand the test usage. I have also added some sanity checks into the utils v1alpha1 <-> v1beta1 conversion to ensure they'll maintain compatibility.

jonathan-innis:
@preflightsiren Did you get a chance to see this comment here on the v1beta1 API? I think it could save you a lot of conversion logic if you just supported the single, newer API

> High-level question: We are thinking of supporting only v1beta1 and dropping support for v1alpha5 in the near-term. Does it make sense to just release these fields only under the v1beta1 API, marking them as alpha fields so that we don't have to reason about the conversion logic in this PR. Up to this point, we've just been handling conversion logic for existing API and trying to push new features into the soon-to-be released beta API

preflightsiren:
> @preflightsiren Did you get a chance to see this comment here on the v1beta1 API? I think it could save you a lot of conversion logic if you just supported the single, newer API
> 
> > High-level question: We are thinking of supporting only v1beta1 and dropping support for v1alpha5 in the near-term. Does it make sense to just release these fields only under the v1beta1 API, marking them as alpha fields so that we don't have to reason about the conversion logic in this PR. Up to this point, we've just been handling conversion logic for existing API and trying to push new features into the soon-to-be released beta API

Thanks Jon, I must have only replied in my mind and not in github! reposting what I commented in slack:
1. My team currently uses the v1alpha1.AWSNodeTemplate resources today, and would require an uplift to EC2NodeClass.
2. The conversion logic is already implemented, is there value to removing it ahead of the v1alpha1 deprecations

jonathan-innis:
Discussed offline. We agreed to remove the v1alpha1 support for this feature and only take it in the newer, v1beta1 API.

preflightsiren:
As per our conversation: I've removed the `v1alpha1` and conversion code for `awsnodetemplates`, all the placement selectors and status fields only apply to the `v1beta1.ec2nodeclass` resource.

I've documented the fields in the ec2nodeclass documents in `preview`

jonathan-innis (commented):
/karpenter snapshot

preflightsiren:
sorry @jonathan-innis, don't know how I missed that amongst all the `make presubmit` checks performed :(

pkg/providers/license/license.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if cached, ok := p.cache.Get(fmt.Sprint(hash)); ok {
+		return cached.([]string), nil
+	}
+
+	var licenses []string
+	// Look up all License Configurations
+	output, err := p.licensemanager.ListLicenseConfigurationsWithContext(ctx, &licensemanager.ListLicenseConfigurationsInput{})
```
jonathan-innis, line 72: How big is this output? Is there a pages version that we can use to reduce the memory overhead?

pkg/apis/v1beta1/ec2nodeclass_status.go
```diff
@@ -49,6 +49,16 @@ type AMI struct {
 	Requirements []v1.NodeSelectorRequirement `json:"requirements"`
 }
 
+// HostResourceGroup contains the resolved host resource group name and arn for node launch
+type HostResourceGroup struct {
+	// Name of the HRG
+	// +optional
+	Name string `json:"name,omitempty"`
```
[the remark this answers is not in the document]
preflightsiren: I believe so, I can change this to only return the ARN.

pkg/providers/amifamily/resolver.go
```diff
@@ -174,6 +193,25 @@ func (r Resolver) Resolve(ctx context.Context, nodeClass *v1beta1.EC2NodeClass,
 	return resolvedTemplates, nil
 }
 
+func (r Resolver) resolvePlacement(nodeClass *v1beta1.EC2NodeClass) *Placement {
+	var placement *Placement
+	hrg := nodeClass.Status.HostResourceGroup
```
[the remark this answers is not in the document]
preflightsiren: I guess it all depends on how _eventually_ consistent the system will be. I don't believe it's an issue unless the consistency goes beyond the observed behaviour (reconciled in seconds).

If only 1 property is set, provisioning will error and the reconciliation loop with start afresh.

I believe this is all a binary operation that will fail or succeed, and far less likely to partially succeed.

pkg/providers/amifamily/resolver.go
```diff
@@ -161,6 +178,8 @@ func (r Resolver) Resolve(ctx context.Context, nodeClass *v1beta1.EC2NodeClass,
 				DetailedMonitoring:  aws.BoolValue(nodeClass.Spec.DetailedMonitoring),
 				AMIID:               amiID,
 				InstanceTypes:       instanceTypes,
+				Licenses:            nodeClass.Status.Licenses,
```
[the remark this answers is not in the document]
preflightsiren: I'm happy to leave this eventually consistent; but I'm happy to take  input on this direction.

pkg/providers/placementgroup/placementgroup.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if cached, ok := p.cache.Get(fmt.Sprint(hash)); ok {
+		return cached.(*ec2.PlacementGroup), nil
+	}
+
+	var match *ec2.PlacementGroup
+	// Look up all ec2 placement groups
+	output, err := p.ec2api.DescribePlacementGroupsWithContext(ctx, &ec2.DescribePlacementGroupsInput{})
```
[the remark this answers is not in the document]
preflightsiren: I can't imagine it being large, and perhaps the API owners felt the same, hence no paged function - https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DescribePlacementGroups

pkg/providers/placementgroup/placementgroup.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		logging.FromContext(ctx).Errorf("discovering placement groups, %w", err)
+		return nil, err
+	}
+	for i := range output.PlacementGroups {
```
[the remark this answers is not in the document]
preflightsiren: not sure I understand your question. here: we're iterating through the list of placementgroups that is the output of the api response. We can only use a single Placement Group in any of our LaunchTemplate configurations, so we resolve this list down to a single placement group.

pkg/providers/license/license.go
preflightsiren: typically small - there's no paged option https://docs.aws.amazon.com/sdk-for-go/api/service/licensemanager/#LicenseManager.ListLicenseConfigurationsWithContext

pkg/providers/placementgroup/placementgroup.go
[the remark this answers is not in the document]
jonathan-innis: Right, but the question was more if the API should be represented as "terms" or if it need only be represented as a selector. Is there a need to have the flexibility to select between two placement groups in the API? From what you are saying, I don't think so. We may only need a `placementGroupSelector` if we don't need the terms component

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -193,6 +193,33 @@ spec:
...
+                    logic for host Resource groups that are used to launch nodes.
+                    If multiple fields are used for selection, the requirements are
+                    ANDed
+                  properties:
+                    name:
+                      description: Name of the hrg to be selected
+                      type: string
+                  type: object
+                type: array
+              licenseSelectorTerms:
```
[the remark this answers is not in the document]
cspargo: there's two options:
1. Associate it with each AMI
2. Specify it in launch template

In some cases it makes sense to associate with AMIs, in others it does not, such as when the an AMI is used both with & without license rules or where doing so just adds unnecessary workflow steps.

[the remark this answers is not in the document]
cspargo: so dedicated host resource groups are a peculiar thing. They are implemented within license manager, but they don't have to have anything to do with software licenses https://docs.aws.amazon.com/license-manager/latest/userguide/host-resource-groups.html

>You can use host resource groups to separate hosts by purpose, for example, development test hosts versus production, organizational unit, or license constraint.

We use them to get a dynamically allocated pool of dedicated hosts that can be shared across accounts.

preflightsiren:
After reviewing the work needed to finish this feature, I can't justify it. I'm happy for others to build off its current state.

PR #4556, opened 2023-09-04
feat: Add cloudformation describe stack step on create cluster failure

### LINKED ISSUE #4467: Run `aws cloudformation describe-stack-events` on a Create Cluster failure in E2E
### Description

Currently, if we experience a failure in the [create-cluster](https://github.com/aws/karpenter/blob/main/.github/actions/e2e/create-cluster/action.yaml#L61) step in the E2E testing suite flow, we don't get any detail on why the CloudFormation stack failed to deploy. We should automatically run `aws cloudformation describe-stack-events <stack-id>` after a failure to ensure that we get all the information directly in the logs.

aman0408:
Fixes #4467 

**Description**
Added a new step in E2E test workflow to describe cloudformation stack events on failure to "create eks cluster" step. This will enable us to view the failure reason in the logs itself.

Ran E2E workflow while ensuring the create eks cluster step failed. Also tried creating a dummy ec2 stack creation and made it fail.

- [ ] Yes, issue opened: #
- [x] No


.github/workflows/e2e.yaml
```diff
@@ -101,6 +102,9 @@ jobs:
           k8s_version: ${{ inputs.k8s_version }}
           ip_family: ${{ inputs.suite == 'IPv6' && 'IPv6' || 'IPv4' }} # Set the value to IPv6 if IPv6 suite, else IPv4
           git_ref: ${{ inputs.git_ref }}
+      - name: cloudformation describe stack events
+        if: ${{ failure() && steps.create-eks-cluster.outcome != 'success' }}
```
jonathan-innis, line 106:
```suggestion
        if: ${{ steps.create-eks-cluster.outcome != 'success' }}
```

Do we need both of these checks or can we just rely on the second one?

.github/workflows/e2e.yaml
```diff
@@ -101,6 +102,9 @@ jobs:
...
+        run: aws cloudformation describe-stack-events --stack-name=iam-${{ env.CLUSTER_NAME }}
```
jonathan-innis, line 107: I believe there are many stacks that are spawned by the `eksctl` command. Did we make sure that we handle describe stack events if _any_ of these different stacks fails, not just the IAM stack.

jonathan-innis: Maybe we can list stacks by tag and then describe stack events across all stacks on failure?

.github/workflows/e2e.yaml
aman0408: We would need both as just the second one skips the steps because of a default `sucess()` check being applied.
[Docs state that](https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions): 

> A default status check of success() is applied unless you include one of these functions.

website/content/en/preview/getting-started/getting-started-with-karpenter/createec2.yaml
aman0408, line 1: Test stack file. To be removed

.github/workflows/e2e.yaml
aman0408: Updated this

.github/workflows/e2e.yaml
jonathan-innis: Doesn't that mean that if you don't have an `if` condition, that it would use `success()`? I think in this case, since we have an `if` statement, we can just assume that we can check a single step to see if it failed and base whether we run this step on that

jonathan-innis (commented):
Nice! 🎉 Just one nit

.github/workflows/e2e.yaml
jonathan-innis: Talked offline on Slack. Github has a section that details that both of these conditions need to be there: https://docs.github.com/en/actions/learn-github-actions/expressions#failure-with-conditions

jonathan-innis (dismissed):
LGTM 🚀

jonathan-innis (approved):
LGTM 🚀

jonathan-innis (approved):
/karpenter snapshot

PR #4571, opened 2023-09-06
feat: support kubelet graceful shutdown for Bottlerocket

bwagner5:
Fixes #N/A 
https://github.com/bottlerocket-os/bottlerocket/pull/2930

**Description**
 - Support Kubelet graceful shutdown for Bottelrocket in BRs kubernetes configuration block. 

 - Manually on EKS cluster. 

- [ ] Yes, issue opened: # 
- [x] No


jonathan-innis (commented):
Can we make an update to the E2E test? Or do we already validate that passing these settings through succeeds on BR?

bwagner5:
> Can we make an update to the E2E test? Or do we already validate that passing these settings through succeeds on BR?

Today we just validate that some params pass thru and test the merge logic.  If we wanted to actually test, we'd need to add all BR params to the suite-test. Not sure if it makes sense to test each param like that though. We don't test the other pass thru params (outside of the kubernetes block)

jonathan-innis (approved):
Ah got it. This isn't being added to our Provisioner, this is just adding support for the user to be able to specify it in the `AWSNodeTemplate` merge logic

PR #4575, opened 2023-09-06
feat: Add instance profile generation for `v1beta1/NodeClass`

jonathan-innis:
Fixes #N/A 

**Description**

This change adds instance profile generation when specifying a role in the `v1beta1/NodeClass`. This will not change the ways that instance profiles can be specified with the `v1alpha5/Provisioner`.

Starting in `v1beta1`, you will be required to specify a role in your `EC2NodeClass` like

```yaml
kind: EC2NodeClass
metadata:
  name: default
spec:
  role: KarpenterNodeRole-karpenter-test
...
```

Karpenter will provision an instance profile that uses this role and propagate that profile into the `EC2NodeClass` status

```yaml
kind: EC2NodeClass
metadata:
  name: default
spec:
  role: KarpenterNodeRole-karpenter-test
...
status:
  instanceProfile: karpenter-test_1234567890123456789
```

To ensure that instances that are using this instance profile are able to safely terminate, we now enforce that any `v1beta1/NodeClaim` that is using the `EC2NodeClass` must be removed before the `EC2NodeClass` will remove the instance profile and terminate. You can view this through the eventing logic like

```console
➜  karpenter git:(v1beta1-testing) k describe ec2nodeclasses default
...
Events:
  Type    Reason                        Age   From       Message
  ----    ------                        ----  ----       -------
  Normal  WaitingOnNodeClaimTermination  18s   karpenter  Waiting on NodeClaim termination for default-8l2tz, default-frz5j, default-bfwnx, default-kzwhq, default-jp8dn and 5 other(s)
```

`make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
+++ b/pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
@@ -341,6 +341,10 @@ spec:
                   - requirements
                   type: object
                 type: array
+              instanceProfile:
+                description: InstanceProfile contains the resolved instance profile
+                  for the role
+                type: string
               securityGroups:
                 description: SecurityGroups contains the current Security Groups values
                   that are available to the cluster under the SecurityGroups selectors.
--- a/pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
+++ b/pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
@@ -223,7 +223,10 @@ spec:
               role:
                 description: Role is the AWS identity that nodes use. This field is
                   immutable. Marking this field as immutable avoids concerns around
-                  terminating managed instance profiles from running instances.
+                  terminating managed instance profiles from running instances. This
+                  field may be made mutable in the future, assuming the correct garbage
+                  collection and drift handling is implemented for the old instance
+                  profiles on an update.
                 type: string
               securityGroupSelectorTerms:
                 description: SecurityGroupSelectorTerms is a list of or security group
--- a/pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
+++ b/pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
@@ -221,7 +221,8 @@ spec:
                     type: string
                 type: object
               role:
-                description: Role is the AWS identity that nodes use.
+                description: Role is the AWS identity that nodes use. This field is
+                  immutable.
                 type: string
               securityGroupSelectorTerms:
                 description: SecurityGroupSelectorTerms is a list of or security group
--- a/pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
+++ b/pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
@@ -222,7 +222,8 @@ spec:
                 type: object
               role:
                 description: Role is the AWS identity that nodes use. This field is
-                  immutable.
+                  immutable. Marking this field as immutable avoids concerns around
+                  terminating managed instance profiles from running instances.
                 type: string
               securityGroupSelectorTerms:
                 description: SecurityGroupSelectorTerms is a list of or security group
--- a/pkg/apis/v1beta1/ec2nodeclass_status.go
+++ b/pkg/apis/v1beta1/ec2nodeclass_status.go
@@ -63,4 +63,7 @@ type EC2NodeClassStatus struct {
 	// cluster under the AMI selectors.
 	// +optional
 	AMIs []AMI `json:"amis,omitempty"`
+	// InstanceProfile contains the resolved instance profile for the role
+	// +optional
+	InstanceProfile string `json:"instanceProfile,omitempty"`
 }
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -51,6 +51,10 @@ func init() {
 	)
 }
 
+const (
+	TerminationFinalizer = Group + "/termination"
+)
+
 var (
 	CapacityTypeSpot       = ec2.DefaultTargetCapacityTypeSpot
 	CapacityTypeOnDemand   = ec2.DefaultTargetCapacityTypeOnDemand
@@ -99,6 +103,8 @@ var (
 	ResourceHabanaGaudi        v1.ResourceName = "habana.ai/gaudi"
 	ResourceAWSPodENI          v1.ResourceName = "vpc.amazonaws.com/pod-eni"
 	ResourcePrivateIPv4Address v1.ResourceName = "vpc.amazonaws.com/PrivateIPv4Address"
+
+	LabelNodeClass = Group + "/nodeclass"
 
 	LabelInstanceHypervisor                   = Group + "/instance-hypervisor"
 	LabelInstanceEncryptionInTransitSupported = Group + "/instance-encryption-in-transit-supported"
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -97,7 +97,8 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *corev1beta1.NodeC
 		if errors.IsNotFound(err) {
 			c.recorder.Publish(cloudproviderevents.NodeClaimFailedToResolveNodeClass(nodeClaim))
 		}
-		return nil, fmt.Errorf("resolving node class, %w", err)
+		// We treat a failure to resolve the NodeClass as an ICE since this means there is no capacity possibilities for this NodeClaim
+		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("resolving node class, %w", err))
 	}
 	instanceTypes, err := c.resolveInstanceTypes(ctx, nodeClaim, nodeClass)
 	if err != nil {
@@ -241,6 +242,10 @@ func (c *CloudProvider) resolveNodeClassFromNodeClaim(ctx context.Context, nodeC
 	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: nodeClaim.Spec.NodeClass.Name}, nodeClass); err != nil {
 		return nil, err
 	}
+	// For the purposes of NodeClass CloudProvider resolution, we treat deleting NodeClasses as NotFound
+	if !nodeClass.DeletionTimestamp.IsZero() {
+		return nil, errors.NewNotFound(v1beta1.SchemeGroupVersion.WithResource("ec2nodeclasses").GroupResource(), nodeClass.Name)
+	}
 	return nodeClass, nil
 }
 
@@ -260,6 +265,10 @@ func (c *CloudProvider) resolveNodeClassFromNodePool(ctx context.Context, nodePo
 	nodeClass := &v1beta1.EC2NodeClass{}
 	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: nodePool.Spec.Template.Spec.NodeClass.Name}, nodeClass); err != nil {
 		return nil, err
+	}
+	// For the purposes of NodeClass CloudProvider resolution, we treat deleting NodeClasses as NotFound
+	if !nodeClass.DeletionTimestamp.IsZero() {
+		return nil, errors.NewNotFound(v1beta1.SchemeGroupVersion.WithResource("ec2nodeclasses").GroupResource(), nodeClass.Name)
 	}
 	return nodeClass, nil
 }
--- a/pkg/controllers/nodeclass/controller.go
+++ b/pkg/controllers/nodeclass/controller.go
@@ -24,9 +24,11 @@ import (
 	"golang.org/x/time/rate"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/client-go/util/workqueue"
+	"knative.dev/pkg/logging"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -38,30 +40,36 @@ import (
 	"github.com/aws/karpenter/pkg/apis/v1alpha1"
 	"github.com/aws/karpenter/pkg/apis/v1beta1"
 	"github.com/aws/karpenter/pkg/providers/amifamily"
+	"github.com/aws/karpenter/pkg/providers/instanceprofile"
 	"github.com/aws/karpenter/pkg/providers/securitygroup"
 	"github.com/aws/karpenter/pkg/providers/subnet"
 	nodeclassutil "github.com/aws/karpenter/pkg/utils/nodeclass"
 )
 
 type Controller struct {
-	kubeClient            client.Client
-	subnetProvider        *subnet.Provider
-	securityGroupProvider *securitygroup.Provider
-	amiProvider           *amifamily.Provider
+	kubeClient              client.Client
+	subnetProvider          *subnet.Provider
+	securityGroupProvider   *securitygroup.Provider
+	amiProvider             *amifamily.Provider
+	instanceProfileProvider *instanceprofile.Provider
 }
 
-func NewController(kubeClient client.Client, subnetProvider *subnet.Provider,
-	securityGroupProvider *securitygroup.Provider, amiProvider *amifamily.Provider) *Controller {
+func NewController(kubeClient client.Client, subnetProvider *subnet.Provider, securityGroupProvider *securitygroup.Provider,
+	amiProvider *amifamily.Provider, instanceProfileProvider *instanceprofile.Provider) *Controller {
 	return &Controller{
-		kubeClient:            kubeClient,
-		subnetProvider:        subnetProvider,
-		securityGroupProvider: securityGroupProvider,
-		amiProvider:           amiProvider,
+		kubeClient:              kubeClient,
+		subnetProvider:          subnetProvider,
+		securityGroupProvider:   securityGroupProvider,
+		amiProvider:             amiProvider,
+		instanceProfileProvider: instanceProfileProvider,
 	}
 }
 
 func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
 	stored := nodeClass.DeepCopy()
+	if !nodeClass.IsNodeTemplate {
+		controllerutil.AddFinalizer(nodeClass, v1beta1.TerminationFinalizer)
+	}
 	nodeClass.Annotations = lo.Assign(nodeClass.Annotations, nodeclassutil.HashAnnotation(nodeClass))
 	err := multierr.Combine(
 		c.resolveSubnets(ctx, nodeClass),
@@ -78,6 +86,33 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
 		}
 	}
 	return reconcile.Result{RequeueAfter: 5 * time.Minute}, err
+}
+
+func (c *Controller) Finalize(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
+	stored := nodeClass.DeepCopy()
+	if !controllerutil.ContainsFinalizer(nodeClass, v1beta1.TerminationFinalizer) {
+		return reconcile.Result{}, nil
+	}
+	ids, err := c.instanceProfileProvider.AssociatedInstances(ctx, nodeClass)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("terminating instance profile, %w", err)
+	}
+	if len(ids) > 0 {
+		if time.Since(nodeClass.DeletionTimestamp.Time) > time.Minute*5 {
+			logging.FromContext(ctx).With("ids", ids).Debugf("waiting to terminate instance profile until all associated instances are terminated")
+		}
+		return reconcile.Result{Requeue: true}, nil
+	}
+	if err = c.instanceProfileProvider.Delete(ctx, nodeClass); err != nil {
+		return reconcile.Result{}, fmt.Errorf("terminating instance profile, %w", err)
+	}
+	controllerutil.RemoveFinalizer(nodeClass, v1beta1.TerminationFinalizer)
+	if !equality.Semantic.DeepEqual(stored, nodeClass) {
+		if err = nodeclassutil.Patch(ctx, c.kubeClient, stored, nodeClass); err != nil {
+			return reconcile.Result{}, client.IgnoreNotFound(fmt.Errorf("removing termination finalizer, %w", err))
+		}
+	}
+	return reconcile.Result{}, nil
 }
 
 func (c *Controller) resolveSubnets(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) error {
@@ -152,15 +187,17 @@ func (c *Controller) resolveAMIs(ctx context.Context, nodeClass *v1beta1.EC2Node
 	return nil
 }
 
+var _ corecontroller.FinalizingTypedController[*v1beta1.EC2NodeClass] = (*NodeClassController)(nil)
+
 //nolint:revive
 type NodeClassController struct {
 	*Controller
 }
 
-func NewNodeClassController(kubeClient client.Client, subnetProvider *subnet.Provider,
-	securityGroupProvider *securitygroup.Provider, amiProvider *amifamily.Provider) corecontroller.Controller {
+func NewNodeClassController(kubeClient client.Client, subnetProvider *subnet.Provider, securityGroupProvider *securitygroup.Provider,
+	amiProvider *amifamily.Provider, instanceProfileProvider *instanceprofile.Provider) corecontroller.Controller {
 	return corecontroller.Typed[*v1beta1.EC2NodeClass](kubeClient, &NodeClassController{
-		Controller: NewController(kubeClient, subnetProvider, securityGroupProvider, amiProvider),
+		Controller: NewController(kubeClient, subnetProvider, securityGroupProvider, amiProvider, instanceProfileProvider),
 	})
 }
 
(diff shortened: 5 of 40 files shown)


pkg/apis/v1beta1/labels.go
```diff
@@ -51,6 +51,10 @@ func init() {
 	)
 }
 
+const (
+	TerminationFinalizer = Group + "/termination"
```
njtran, line 55: Had we mistakenly removed this?

pkg/controllers/nodeclass/controller.go
```diff
@@ -80,6 +87,13 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
 	return reconcile.Result{RequeueAfter: 5 * time.Minute}, err
 }
 
+func (c *Controller) Finalize(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
+	if !controllerutil.ContainsFinalizer(nodeClass, v1beta1.TerminationFinalizer) {
+		return reconcile.Result{}, nil
+	}
+	return reconcile.Result{}, c.instanceProfileProvider.Delete(ctx, nodeClass)
```
njtran, line 94: Does this fail if there are other instances using the same instance profile but owned by a different node class? That would require user intervention if I understand correctly, and is still possible even if we suffix each instance profile we generate with the name of the node class.

njtran: Thinking further, since IAM is global, any clusters that have the same nodeclass names/setups will also be relying on this instance profile generation. Can we be sure the instance profile is safe to delete here if other instances are using it? Same with creation. Is this idempotent?

pkg/test/environment.go
```diff
@@ -78,13 +83,15 @@ func NewEnvironment(ctx context.Context, env *coretest.Environment) *Environment
 	launchTemplateCache := cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval)
 	subnetCache := cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval)
 	securityGroupCache := cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval)
+	instanceProfileCache := cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval)
 	fakePricingAPI := &fake.PricingAPI{}
```
njtran, line 87: If you're going to add instantiation functions for the fakes, can you do it for this too?

pkg/providers/instanceprofile/instanceprofile.go
```diff
@@ -0,0 +1,128 @@
...
+			return profileName, nil
+		}
+		if _, err = p.iamapi.RemoveRoleFromInstanceProfileWithContext(ctx, &iam.RemoveRoleFromInstanceProfileInput{
+			InstanceProfileName: aws.String(profileName),
+			RoleName:            instanceProfile.Roles[0].RoleName,
+		}); err != nil {
+			return "", fmt.Errorf("removing role %q from instance profile %q, %w", aws.StringValue(instanceProfile.Roles[0].RoleName), profileName, err)
+		}
+	}
+	if _, err := p.iamapi.AddRoleToInstanceProfileWithContext(ctx, &iam.AddRoleToInstanceProfileInput{
```
njtran, line 84: What is the implication if role removal succeeds, but role addition fails? Does this just mean the instance profile has a period of time where it has no role? How does that impact existing instances?

pkg/operator/suite_test.go
```diff
@@ -85,7 +85,7 @@ var _ = Describe("Operator", func() {
 		ctx = settings.ToContext(ctx, test.Settings(test.SettingOptions{
 			ClusterEndpoint: lo.ToPtr(""),
 		}))
-		fakeEKSAPI.DescribeClusterBehaviour.Output.Set(
+		fakeEKSAPI.DescribeClusterBehavior.Output.Set(
```
njtran, line 88: 🇺🇸

njtran (commented):
Looks good. I have some questions about how we're planning on managing these instance profiles now.

pkg/controllers/nodeclass/controller.go
jonathan-innis: > since IAM is global

That's a good callout. This should be region prefixed so that we are not hitting conflicts with other regions when creating deleting this instance profile for this role. In terms of the safety of deleting this instance profile if there are instances using it, I think it probably makes sense to hold EC2NodeClass deletion until all instances are dissociated from the instance profile.

I think there should be an implicit assumption here that the instance profile that is being generated here is for the explicit use and management of the Karpenter controller. You _can_ attach other instance types to the instance profile, but (if you do so), you will get failures when we try to remove the EC2NodeClass, which you will see inside of your error logs on the termination and you won't be able to fully remove the EC2NodeClass until you resolve those instance association errors (which you would have to deal with by doing an instance deletion)

njtran: > This should be region prefixed so that we are not hitting conflicts with other regions when creating deleting this instance profile for this role. 

Can we make sure that the limit on instanceProfiles here is sufficiently large? IIUC from this change, Karpenter will now create/update/delete instanceProfiles for the number of EC2NodeClasses for an account across all regions. I can imagine for users with multiple clusters per region and multiple EC2NodeClasses per cluster that this may be a lot of instance profiles.

njtran: > you will get failures when we try to remove the EC2NodeClass, which you will see inside of your error logs on the termination and you won't be able to fully remove the EC2NodeClass until you resolve those instance association errors (which you would have to deal with by doing an instance deletion)

Is there someway we can limit use of this instanceProfile? It seems odd that if a user created an instance (not necessarily connected to a cluster) with the instanceProfile that it would prevent cascade deletion of the nodeClass. Also having this IAM dependency now means that we can't do any type of Forceful deletion of this object without orphaning/leaking resources, which may be a sharp edge for karpenter uninstallations or cluster deletes.

pkg/operator/suite_test.go
jonathan-innis: 😆

pkg/apis/v1beta1/labels.go
jonathan-innis: No, we weren't using a termination finalizer in the `karpenter.k8s.aws` group up to this point

pkg/providers/instanceprofile/instanceprofile.go
```diff
@@ -0,0 +1,154 @@
...
+	p.cache.SetDefault(string(nodeClass.UID), nil)
+	return profileName, nil
+}
+
+func (p *Provider) AssociatedInstances(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) ([]string, error) {
+	profileName := p.GetProfileName(ctx, nodeClass)
+
+	// Get all instances that are using our instance profile name and are not yet terminated
+	var ids []string
+	if err := p.ec2api.DescribeInstancesPagesWithContext(ctx, &ec2.DescribeInstancesInput{
```
njtran, line 105: QQ: does this automatically get included into our batcher?

pkg/providers/instanceprofile/instanceprofile.go
```diff
@@ -0,0 +1,154 @@
...
+	if _, err = p.iamapi.DeleteInstanceProfileWithContext(ctx, &iam.DeleteInstanceProfileInput{
+		InstanceProfileName: aws.String(profileName),
+	}); err != nil {
+		return awserrors.IgnoreNotFound(fmt.Errorf("deleting instance profile %q, %w", profileName, err))
+	}
+	return nil
+}
+
+func (p *Provider) GetProfileName(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) string {
+	return fmt.Sprintf("%s/%d", settings.FromContext(ctx).ClusterName, lo.Must(hashstructure.Hash(fmt.Sprintf("%s/%s", nodeClass.Name, p.region), hashstructure.FormatV2, &hashstructure.HashOptions{})))
```
njtran, line 153: Does it make sense to not hash this? I can see this being a nice observability pro to know which node class each Instance profile refers to easily.

pkg/controllers/nodeclass/nodeclass_test.go
```diff
@@ -790,4 +792,106 @@ var _ = Describe("NodeClassController", func() {
...
+			Expect(awsEnv.IAMAPI.InstanceProfiles).To(HaveLen(0))
+
+			Expect(env.Client.Delete(ctx, nodeClass)).To(Succeed())
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+			Expect(awsEnv.IAMAPI.InstanceProfiles).To(HaveLen(0))
+			ExpectNotFound(ctx, env.Client, nodeClass)
+		})
+		It("should not delete the instance profile until all associated instances are terminated", func() {
+			var ids []string
+			for i := 0; i < 10; i++ {
```
njtran, line 849: Probably easier to do just 2 instances here since we're just going to be doing the following, and how many instances over 1 doesn't matter. 
- Create n instances
- Try to delete
- Make it 1 instance left
- Try to delete
- Make 0 instances
- Delete successful

pkg/fake/iamapi.go
```diff
@@ -0,0 +1,146 @@
...
+
+func NewIAMAPI() *IAMAPI {
+	return &IAMAPI{InstanceProfiles: map[string]*iam.InstanceProfile{}}
+}
+
+// Reset must be called between tests otherwise tests will pollute
+// each other.
+func (s *IAMAPI) Reset() {
+	s.GetInstanceProfileBehavior.Reset()
+	s.InstanceProfiles = map[string]*iam.InstanceProfile{}
```
njtran, line 59: Is there a reason we shouldn't just reset all the behavior funcs?

pkg/providers/instanceprofile/instanceprofile.go
jonathan-innis: So if we don't hash it, then we have to truncate it, at which point we can't guarantee the name is sufficiently random. This is part of the reason that we add the tags to the instance profile so that users can track down the associated instance profile if they need to.

jonathan-innis: There's probably also room here down the line to do the async reconciliation of this instance profile, where we propagate the name into the status, at which point that would be the observability piece that we are looking for here. But the real limitation here is:
1. The EKS cluster name limit is 100 chars
2. The Node Class name limit is 64 chars
3. The Instance profile name limit is 128 chars

We necessarily can't capture all the information here. We could consider _just_ using the NodeClass UUID as another option.

pkg/fake/iamapi.go
jonathan-innis: Miss on my side. Good catch1

pkg/providers/instanceprofile/instanceprofile.go
jonathan-innis: Good callout 👍 I'm not sure the batcher would really have a tangible impact here but I'm fine going down the same codepath to be more DRY.

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -221,7 +221,8 @@ spec:
                     type: string
                 type: object
               role:
-                description: Role is the AWS identity that nodes use.
+                description: Role is the AWS identity that nodes use. This field is
```
njtran, line 224: Can we add something in the docs/definition here that explains why we chose to make this immutable? I want to reduce the questions we get about this.

pkg/controllers/nodeclass/controller.go
```diff
@@ -80,6 +91,31 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
+	if !controllerutil.ContainsFinalizer(nodeClass, v1beta1.TerminationFinalizer) {
+		return reconcile.Result{}, nil
+	}
+	ids, err := c.instanceProfileProvider.AssociatedInstances(ctx, nodeClass)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("terminating instance profile, %w", err)
+	}
+	if len(ids) > 0 {
+		c.recorder.Publish(WaitingOnInstanceTerminationEvent(nodeClass, instanceprofile.GetProfileName(ctx, nodeClass), ids))
+		return reconcile.Result{Requeue: true}, nil
```
njtran, line 105: I know this does a exponential backoff, but thoughts on just making this a `RequeueAfter: 10s` by default? Would prevent us from hammering the API calls, and since this is finalizing, we probably don't need to be super quick about this either.

pkg/controllers/nodeclass/nodeclass_test.go
```diff
@@ -790,4 +793,138 @@ var _ = Describe("NodeClassController", func() {
...
+			}
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+			Expect(awsEnv.IAMAPI.InstanceProfiles).To(HaveLen(1))
+
+			Expect(env.Client.Delete(ctx, nodeClass)).To(Succeed())
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+			Expect(awsEnv.IAMAPI.InstanceProfiles).To(HaveLen(0))
+			ExpectNotFound(ctx, env.Client, nodeClass)
+		})
+		It("should succeed to delete the instance profile when it doesn't exist", func() {
```
njtran, line 837: nit: think this semantically makes more sense as a sentence :) 
```suggestion
		It("should succeed to delete the node class when the instance profile doesn't exist", func() {
```

pkg/fake/iamapi.go
```diff
@@ -0,0 +1,150 @@
...
+	s.DeleteInstanceProfileBehavior.Reset()
+	s.AddRoleToInstanceProfileBehavior.Reset()
+	s.RemoveRoleFromInstanceProfileBehavior.Reset()
+	s.InstanceProfiles = map[string]*iam.InstanceProfile{}
+}
+
+func (s *IAMAPI) GetInstanceProfileWithContext(_ context.Context, input *iam.GetInstanceProfileInput, _ ...request.Option) (*iam.GetInstanceProfileOutput, error) {
+	return s.GetInstanceProfileBehavior.Invoke(input, func(*iam.GetInstanceProfileInput) (*iam.GetInstanceProfileOutput, error) {
+		s.Lock()
+		defer s.Unlock()
```
njtran, line 69: Definitely good practice to add these locks/unlocks into our mock APIs. Not saying you need to do in this PR, but we should probably do it.

pkg/fake/iamapi.go
```diff
@@ -0,0 +1,150 @@
...
+
+		if i, ok := s.InstanceProfiles[aws.StringValue(input.InstanceProfileName)]; ok {
+			var newRoles []*iam.Role
+			for _, role := range i.Roles {
+				if aws.StringValue(role.RoleName) != aws.StringValue(input.RoleName) {
+					newRoles = append(newRoles, role)
+				}
+			}
```
njtran, line 141: nit: Is there a lo.Reject() you can do here?

pkg/fake/iamapi.go
```diff
@@ -0,0 +1,150 @@
...
+			if len(i.Roles) == len(newRoles) {
+				return nil, awserr.New(iam.ErrCodeNoSuchEntityException, fmt.Sprintf("The role with name %s cannot be found", aws.StringValue(input.RoleName)), nil)
```
njtran, line 143: nit: Can probably shorten, since it's inferred that we're referring to the role by name
```suggestion
				return nil, awserr.New(iam.ErrCodeNoSuchEntityException, fmt.Sprintf("Role %s cannot be found", aws.StringValue(input.RoleName)), nil)
```

pkg/providers/instanceprofile/instanceprofile.go
```diff
@@ -0,0 +1,159 @@
...
+		cache:  cache,
+	}
+}
+
+func (p *Provider) Create(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (string, error) {
+	tags := lo.Assign(nodeClass.Spec.Tags, map[string]string{
+		fmt.Sprintf("kubernetes.io/cluster/%s", settings.FromContext(ctx).ClusterName): "owned",
+		corev1beta1.ManagedByAnnotationKey:                                             settings.FromContext(ctx).ClusterName,
+		v1beta1.LabelNodeClass:                                                         nodeClass.Name, v1.LabelTopologyRegion: p.region,
+	})
```
njtran, line 65: musing: It would be nice if we could have a nodeclassutils.GetKarpenterTags() that would return this for us so we don't have to do it each time in multiple places in the code. Would be nice to align this elsewhere too, like in instance.go and launchtemplate.go

pkg/providers/instanceprofile/instanceprofile.go
```diff
@@ -0,0 +1,159 @@
...
+	profileName := GetProfileName(ctx, nodeClass)
+
+	// An instance profile exists for this NodeClass
+	if _, ok := p.cache.Get(string(nodeClass.UID)); ok {
```
njtran, line 69: Don't we need to check the status here rather than the cache?

pkg/providers/instanceprofile/instanceprofile.go
```diff
@@ -0,0 +1,159 @@
...
+			Tags:                lo.MapToSlice(tags, func(k, v string) *iam.Tag { return &iam.Tag{Key: aws.String(k), Value: aws.String(v)} }),
+		})
+		if err != nil {
+			return "", fmt.Errorf("creating instance profile %q, %w", profileName, err)
+		}
+		instanceProfile = o.InstanceProfile
+	} else {
+		instanceProfile = out.InstanceProfile
+	}
+	if len(instanceProfile.Roles) == 1 {
```
njtran, line 90: Don't we need to check if the role is the correct role?

pkg/providers/instanceprofile/instanceprofile.go
```diff
@@ -0,0 +1,159 @@
...
+				Values: aws.StringSlice([]string{profileName}),
+			},
+			instanceStateFilter,
+		},
+	}, func(page *ec2.DescribeInstancesOutput, _ bool) bool {
+		for _, res := range page.Reservations {
+			ids = append(ids, lo.Map(res.Instances, func(i *ec2.Instance, _ int) string {
+				return aws.StringValue(i.InstanceId)
+			})...)
+		}
```
njtran, line 121: nit: can do lo.Map()

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -355,6 +360,12 @@ func (p *Provider) getInstanceProfile(ctx context.Context, nodeClass *v1beta1.EC
 	if nodeClass.Spec.InstanceProfile != nil {
 		return aws.StringValue(nodeClass.Spec.InstanceProfile), nil
 	}
+	if nodeClass.Spec.Role != "" {
+		if nodeClass.Status.InstanceProfile == "" {
+			return "", cloudprovider.NewNodeClassNotReadyError(fmt.Errorf("instance profile hasn't resolved for role"))
```
njtran, line 365: Generally I think it should be easy to add this check in for subnets, security groups, and AMIs. Thoughts on adding this in here or just opting for fast follow?

pkg/fake/iamapi.go
jonathan-innis: I was mainly just copying the formatting of the error messages from the actual IAM apis. If you attempt these error paths on these APIs, this is the error you get.

pkg/providers/instanceprofile/instanceprofile.go
jonathan-innis: Agreed. I could see this coming in a follow-on.

pkg/providers/instanceprofile/instanceprofile.go
jonathan-innis: Not a huge fan of it since it's a double `lo.Map` which means you also have to do a `lo.Flatten` at the end of it all

pkg/providers/launchtemplate/launchtemplate.go
jonathan-innis: Agreed. I think it's out of the scope of this PR so it makes sense as a fast-follow.

pkg/fake/iamapi.go
njtran: Ah if this is the same error as the APIs, sounds good to me 👍

njtran (approved):

pkg/controllers/nodeclass/controller.go
```diff
@@ -80,6 +95,31 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
 }
 
+func (c *Controller) Finalize(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
+	stored := nodeClass.DeepCopy()
+	if !controllerutil.ContainsFinalizer(nodeClass, v1beta1.TerminationFinalizer) {
+		return reconcile.Result{}, nil
+	}
+	nodeClaimList := &corev1beta1.NodeClaimList{}
+	if err := c.kubeClient.List(ctx, nodeClaimList, client.MatchingFields{"spec.nodeClass.name": nodeClass.Name}); err != nil {
+		return reconcile.Result{}, fmt.Errorf("listing NodeClaims using NodeClass, %w", err)
```
njtran, line 105: nit on capitalization
```suggestion
		return reconcile.Result{}, fmt.Errorf("listing node claims using node class, %w", err)
```

pkg/controllers/nodeclass/controller.go
```diff
@@ -80,6 +95,31 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
+	}
+	if len(nodeClaimList.Items) > 0 {
+		c.recorder.Publish(WaitingOnNodeClaimTerminationEvent(nodeClass, lo.Map(nodeClaimList.Items, func(nc corev1beta1.NodeClaim, _ int) string { return nc.Name })))
+		return reconcile.Result{RequeueAfter: time.Minute * 10}, nil // periodically fire the event
```
njtran, line 109: We probably don't need this 10s anymore, right?

pkg/controllers/nodeclass/controller.go
```diff
@@ -172,6 +226,16 @@ func (c *NodeClassController) Builder(_ context.Context, m manager.Manager) core
...
 		For(&v1beta1.EC2NodeClass{}).
+		Watches(
+			&source.Kind{Type: &corev1beta1.NodeClaim{}},
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
+				nc := o.(*corev1beta1.NodeClaim)
+				if nc.Spec.NodeClass == nil {
+					return nil
+				}
+				return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: nc.Spec.NodeClass.Name}}}
+			}),
```
njtran, line 237: We only need to watch NodeClaims delete events, right? Or is it update because the finalizer would be removed?

pkg/providers/instanceprofile/instanceprofile.go
```diff
@@ -0,0 +1,131 @@
...
+		if aws.StringValue(instanceProfile.Roles[0].RoleName) == nodeClass.Spec.Role {
+			return profileName, nil
+		}
+		if _, err = p.iamapi.RemoveRoleFromInstanceProfileWithContext(ctx, &iam.RemoveRoleFromInstanceProfileInput{
+			InstanceProfileName: aws.String(profileName),
+			RoleName:            instanceProfile.Roles[0].RoleName,
+		}); err != nil {
+			return "", fmt.Errorf("removing role %q for instance profile %q, %w", aws.StringValue(instanceProfile.Roles[0].RoleName), profileName, err)
+		}
+	}
```
njtran, line 90: Do we need an else if len(instanceProfile.Roles) > 1?

pkg/providers/instanceprofile/instanceprofile.go
```diff
@@ -0,0 +1,131 @@
...
+	if _, err = p.iamapi.AddRoleToInstanceProfileWithContext(ctx, &iam.AddRoleToInstanceProfileInput{
+		InstanceProfileName: aws.String(profileName),
+		RoleName:            aws.String(nodeClass.Spec.Role),
+	}); err != nil {
+		return "", fmt.Errorf("adding role %q to instance profile %q, %w", nodeClass.Spec.Role, profileName, err)
+	}
+	p.cache.SetDefault(string(nodeClass.UID), nil)
```
njtran, line 97: i thought we were going to do something other than the UID here?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -97,7 +97,8 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *corev1beta1.NodeC
 		if errors.IsNotFound(err) {
 			c.recorder.Publish(cloudproviderevents.NodeClaimFailedToResolveNodeClass(nodeClaim))
 		}
-		return nil, fmt.Errorf("resolving node class, %w", err)
+		// We treat a failure to resolve the NodeClass as an ICE since this means there is no capacity possibilities for this NodeClaim
+		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("resolving node class, %w", err))
```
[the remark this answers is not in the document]
jonathan-innis: No impact there. We don't do any updates to the ICE cache on returning this type of error. This error type is here to describe cases where you attempt to launch an instance with the Create() call but the cloudprovider comes back and tells you that there's no capacity there to do so for one reason or another. In this case, it's because we couldn't resolve the config information. 

From the NodeClaim launch side, this causes a delete of the NodeClaim and the provisioner will pick-up and try again, this time realizing that it can't use that Provisioner due to the NotFound returning no instance types

pkg/controllers/nodeclass/controller.go
jonathan-innis: I didn't do 10s here, I did 10m. I was thinking about it and thought that a periodic event to keep that event fresh on the NodeClass was important. Otherwise, you'd only get the event once and then it would be dropped by etcd after an hour

pkg/controllers/nodeclass/controller.go
jonathan-innis: You're right. I can constrain this to deletes

pkg/providers/instanceprofile/instanceprofile.go
jonathan-innis: UID makes sense for the cache, since we don't want previous versions of entries in the cache with the same nodeclass name to overlap

pkg/apis/v1beta1/ec2nodeclass_status.go
```diff
@@ -63,4 +63,7 @@ type EC2NodeClassStatus struct {
 	// cluster under the AMI selectors.
 	// +optional
 	AMIs []AMI `json:"amis,omitempty"`
+	// InstanceProfile contains the resolved instance profile for the role
```
ellistarn, line 66: Worth making as deprecated?

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -221,7 +221,9 @@ spec:
...
+                  immutable. Marking this field as immutable avoids concerns around
```
ellistarn, line 225: Worth mentioning this could be made mutable in the future.

pkg/apis/v1beta1/ec2nodeclass_status.go
jonathan-innis: This isn't deprecated. This is the resolved value into the status, not in the spec

njtran (approved):

PR #4586, opened 2023-09-08
feat(crds): Migrate from a webhook to CEL validation for aws nodetemplates

### LINKED ISSUE #4563: Allow for Usage of Cert-Manager in helm chart for webhook certificate
### Description

I'd like to be able to use cert-manager to manage: https://github.com/aws/karpenter/blob/main/charts/karpenter/templates/secret-webhook-cert.yaml rather than relying on a pre-sync hook or some other mechanism.

Reasonably important. Allows us to control the issuer, renewal timing, and other features of the certificate.

knechtionscoding:
Fixes #4563 

**Description**

Adds support for CEL validation of the AWS Templates

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/pkg/apis/crds/karpenter.k8s.aws_awsnodetemplates.yaml
+++ b/pkg/apis/crds/karpenter.k8s.aws_awsnodetemplates.yaml
@@ -46,6 +46,9 @@ spec:
                   type: string
                 description: AMISelector discovers AMIs to be used by Amazon EC2 tags.
                 type: object
+                x-kubernetes-validations:
+                - message: Only one of UserData and AMISelector can be provided
+                  rule: '!has(self.AMISelector) || has(self.AMISelector) != has(self.UserData)'
               apiVersion:
                 description: 'APIVersion defines the versioned schema of this representation
                   of an object. Servers should convert recognized schemas to the latest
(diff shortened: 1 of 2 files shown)


knechtionscoding:
I have a couple of questions that I'm not sure about:

I'm not sure of the intent behind:
- https://github.com/aws/karpenter/blob/main/pkg/apis/v1alpha1/awsnodetemplate_validation.go#L105
- https://github.com/aws/karpenter/blob/main/pkg/apis/v1alpha1/awsnodetemplate_validation.go#L91-L93

I also don't love how CEL handles nested fields. It would be better if we could provide the validation closer to the actual spec of the AWSTemplate. Right now I have to write a bunch of rules at the top level which feels very wrong and like I should be able to write them at the spec level. Not sure where that's defined though or if the spec is pulled in from a different package.

njtran:
Hey @knechtionscoding, thanks for the PR. Would you be able to make changes to this to enable this only for the v1beta1 APIs? We're actively migrating the APIs over, where the alpha APIs will be soon deprecated.

knechtionscoding:
Sure! Is there a good way to test this change? And I'm still not sure this is the right place for some of the validation. Some the aws parts might be better in karpenter-core?

engedaam:
> Sure! Is there a good way to test this change? And I'm still not sure this is the right place for some of the validation. Some the aws parts might be better in karpenter-core?

@knechtionscoding Which validation are thinking would be better in karpenter-core?

engedaam:
@knechtionscoding Have you made any progress on CEL for v1beta1 APIs? I have some work going toward the effort and just wanted to know your progress

ellistarn:
This is being worked on by @engedaam

PR #4590, opened 2023-09-08
test: use default instance family / generation for Windows tests

jmdeal:
Fixes #N/A 

**Description**
Windows tests will now use the default instance family and generations as they would have if the provisioners were not configured for Windows. This is to address test flaking caused by provisioning instances with extremely slow boot times (e.g. t2.nano / micro).

`go test`

- [ ] Yes, issue opened: # 
- [x] No


bwagner5 (approved):
lgtm!

Issue #4228, opened 2023-07-07
Dynamic instance tagging (from node labels)


cebernardi:
### Description

**What problem are you trying to solve?**

We have a Provisioner that is configured to launch instances both in arm64 and amd64. 
We would like to dynamically tag the ec2 instances with the architecture.

Karpenter tags the instances with a default set of "dynamic" tags, described at the beginning of this documentation section: https://karpenter.sh/docs/concepts/node-templates/#spectags

```
Name: karpenter.sh/provisioner-name/<provisioner-name>
karpenter.sh/provisioner-name: <provisioner-name>
kubernetes.io/cluster/<cluster-name>: owned
```

All the rest of tagging options, are static ([here](https://karpenter.sh/docs/concepts/node-templates/#spectags) and [here](https://karpenter.sh/docs/concepts/settings/#configmap)).

We would like to have a way to specify dynamic tags, for example:
```
spec:
  tags:
    InternalAccountingTag: 1234
    dev.corp.net/app: Calculator
  tagsFromLabels:
  - kubernetes.io/arch
  - node.kubernetes.io/instance-type
```

it could be also something like:

```
spec:
  tags:
    InternalAccountingTag: 1234
    dev.corp.net/app: Calculator
    arch: "{kubernetes.io/arch}"
    instance-type: "node.kubernetes.io/instance-type"
```

while `InternalAccountingTag` and `dev.corp.net/app` are "static" tags, `kubernetes.io/arch` and `node.kubernetes.io/instance-type` are known labels, that could be used as tags as well. 

**How important is this feature to you?**
It would enable closer monitoring of costs and migrations, which are blocked now, or possible with workarounds.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

cebernardi:
I would be interested to contribute to this feature, should it be accepted.

njtran:
@cebernardi if you're open to contributing, please reach out to us in the [#karpenter-dev channel](https://kubernetes.slack.com/archives/C04JW2J5J5P)! 

In your head, this would land in the AWSNodeTemplate right? 

[...]

Issue #2783, opened 2022-11-02
Dynamic tags support based on resource annotation


AlonAvrahami:
### Tell us about your request

Today Karpenter support defining tags only via the Provisioner configurations (`spec.tags`) or via the AWSNodeTemplate configurations (`spec.tags`).
When running many types of workload, it would be awesome to support a way to set EC2 instance (and all other relevant resources) tags dynamically by providing the relevant configuration as resource annotation.
When providing such an annotation, Karpenter will merge the tags with the tags from both Provisioner and AWSNodeTemplate configurations.

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

Instead of defining multiple providers with different tags (per application for example) it would be awesome to add support for dynamic tags by using the resource annotation.
With the feature i can use fewer Provisioners and give a better visibility and control over the resource allocation and usage, mainly for cost analysis. (when you run an application and you would like to analyze how much it cost to run this specific application)

### Are you currently working around this issue?

Currently I'm using terraform for defining all of the Karpenter provisioners, so I'm creating many provisioners for different applications where the only differences between them are the resource tags.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jonathan-innis:
Do you have any details on exactly which tags are being used that differ per-provisioner with an example as well as what you had in mind for how we might be able to enable these tags to be generated dynamicaly?

AlonAvrahami:
> Do you have any details on exactly which tags are being used that differ per-provisioner with an example as well as what you had in mind for how we might be able to enable these tags to be generated dynamicaly?

Yes, for example if i want to request a new instance in a specific region, i want it to add the relevant region tag to the EC2.
A more specific use case could be an ML application that get a batch of tasks, and i want it to add a tag of this specific ML application in order to calculate how much resources have been used in order to perform this batch of tasks.

Currently I'm trying to use one provisioner that will provision nodes for all types of ML application, if an application using a specific node i would like to add a tag that will mark that node X is being used by application A. so in the end result i can add a tag: `Application A = True`.

In anyway, I'm also trying to install KubeCost that should help with the cost visibility.

[...]

PR #4611, opened 2023-09-11
feat: instance tagging controller

### LINKED ISSUE #3342: How to remove "name" tag or customize it dynamically?
### Is an existing page relevant?

_No response_

### What karpenter features are relevant?

labels and tags.

### How should the docs be improved?

Rather than spin up ec2 instances with a static EC2 tag name of the provisioner, it would be really useful if the EC2 tag name matched the node's name, which is currently the private IP for us in EKS.  Is this possible?  It may require karpenter to update the name after the node is provisioned.

If it is not, is it possible to just not tag the node's name at all?  We currently use datadog that automatically imports EC2 tags into metric tags, but aren't getting much use out of `name` since they all have the same name tag.

Thanks!

### Community Note

jmdeal:
Fixes #3342 

**Description**
This PR introduces a new controller for v1beta1 responsible for tagging ec2 instances with their in-cluster name. This replaces the previous behavior where each instance's name corresponded to it's `Provisioner` (e.g. `karpenter.sh/provisioner/default`). It is not possible to do this as part of the `CreateFleet` call since the node name is not yet known. Instead, this controller observes `NodeClaim`s, waiting for their node name to be populated. 

Additionally, while this PR does not introduce custom dynamic tags, it does open up a path for them in the future (see issues #2783 and #4228).

`make test` and `make e2e`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/controllers/nodeclaim/tagging/controller.go
+++ b/pkg/controllers/nodeclaim/tagging/controller.go
@@ -1,0 +1,141 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tagging
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"knative.dev/pkg/logging"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/samber/lo"
+
+	"github.com/aws/karpenter/pkg/apis/v1beta1"
+	"github.com/aws/karpenter/pkg/providers/instance"
+	"github.com/aws/karpenter/pkg/utils"
+
+	corev1beta1 "github.com/aws/karpenter-core/pkg/apis/v1beta1"
+	"github.com/aws/karpenter-core/pkg/cloudprovider"
+	corecontroller "github.com/aws/karpenter-core/pkg/operator/controller"
+)
+
+const (
+	TagNodeClaim = corev1beta1.Group + "/nodeclaim"
+	TagName      = "Name"
+)
+
+type Controller struct {
+	kubeClient       client.Client
+	instanceProvider *instance.Provider
+}
+
+func NewController(kubeClient client.Client, instanceProvider *instance.Provider) corecontroller.Controller {
+	return corecontroller.Typed[*corev1beta1.NodeClaim](kubeClient, &Controller{
+		kubeClient:       kubeClient,
+		instanceProvider: instanceProvider,
+	})
+}
+
+func (c *Controller) Name() string {
+	return "nodeclaim.tagging"
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClaim *corev1beta1.NodeClaim) (reconcile.Result, error) {
+	stored := nodeClaim.DeepCopy()
+
+	if !isTaggable(nodeClaim) {
+		return reconcile.Result{}, nil
+	}
+
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("providerID", nodeClaim.Status.ProviderID))
+	id, err := utils.ParseInstanceID(nodeClaim.Status.ProviderID)
+	if err != nil {
+		// We don't throw an error here since we don't want to retry until the ProviderID has been updated.
+		logging.FromContext(ctx).Errorf("failed to parse instance ID, %w", err)
+		return reconcile.Result{}, nil
+	}
+
+	if err := c.tagInstance(ctx, nodeClaim, id); err != nil {
+		return reconcile.Result{}, cloudprovider.IgnoreNodeClaimNotFoundError(err)
+	}
+
+	nodeClaim.Annotations = lo.Assign(nodeClaim.Annotations, map[string]string{v1beta1.AnnotationInstanceTagged: "true"})
+	if !equality.Semantic.DeepEqual(nodeClaim, stored) {
+		if err := c.kubeClient.Patch(ctx, nodeClaim, client.MergeFrom(stored)); err != nil {
+			return reconcile.Result{}, client.IgnoreNotFound(err)
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontroller.Builder {
+	return corecontroller.Adapt(
+		controllerruntime.
+			NewControllerManagedBy(m).
+			For(&corev1beta1.NodeClaim{}).
+			WithEventFilter(predicate.NewPredicateFuncs(func(o client.Object) bool {
+				return isTaggable(o.(*corev1beta1.NodeClaim))
+			})),
+	)
+}
+
+func (c *Controller) tagInstance(ctx context.Context, nc *corev1beta1.NodeClaim, id string) error {
+	tags := map[string]string{
+		TagName:      nc.Status.NodeName,
+		TagNodeClaim: nc.Name,
+	}
+
+	// Remove tags which have been already populated
+	instance, err := c.instanceProvider.Get(ctx, id)
+	if err != nil {
+		return fmt.Errorf("tagging nodeclaim, %w", err)
+	}
+	tags = lo.OmitByKeys(tags, lo.Keys(instance.Tags))
+	if len(tags) == 0 {
+		return nil
+	}
+
+	// Ensures that no more than 1 CreateTags call is made per second. Rate limiting is required since CreateTags
+	// shares a pool with other mutating calls (e.g. CreateFleet).
+	defer time.Sleep(time.Second)
+	if err := c.instanceProvider.CreateTags(ctx, id, tags); err != nil {
+		return fmt.Errorf("tagging nodeclaim, %w", err)
+	}
+	return nil
+}
+
+func isTaggable(nc *corev1beta1.NodeClaim) bool {
+	// Instance has already been tagged
+	if val := nc.Annotations[v1beta1.AnnotationInstanceTagged]; val == "true" {
+		return false
+	}
+	// Node name is not yet known
+	if nc.Status.NodeName == "" {
+		return false
+	}
+	// NodeClaim is currently terminating
+	if !nc.DeletionTimestamp.IsZero() {
+		return false
+	}
+	return true
+}
--- a/pkg/controllers/nodeclaim/tagging/controller.go
+++ b/pkg/controllers/nodeclaim/tagging/controller.go
@@ -66,7 +66,7 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *corev1beta1.NodeC
 		return reconcile.Result{}, nil
 	}
 
-	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("providerID", nodeClaim.Status.ProviderID))
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("provider-id", nodeClaim.Status.ProviderID))
 	id, err := utils.ParseInstanceID(nodeClaim.Status.ProviderID)
 	if err != nil {
 		// We don't throw an error here since we don't want to retry until the ProviderID has been updated.
--- a/pkg/controllers/nodeclaim/tagging/suite_test.go
+++ b/pkg/controllers/nodeclaim/tagging/suite_test.go
@@ -1,0 +1,230 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tagging_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/samber/lo"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	. "github.com/aws/karpenter-core/pkg/test/expectations"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	. "knative.dev/pkg/logging/testing"
+
+	coresettings "github.com/aws/karpenter-core/pkg/apis/settings"
+	corev1beta1 "github.com/aws/karpenter-core/pkg/apis/v1beta1"
+	coretest "github.com/aws/karpenter-core/pkg/test"
+	"github.com/aws/karpenter/pkg/apis"
+	"github.com/aws/karpenter/pkg/apis/settings"
+	"github.com/aws/karpenter/pkg/apis/v1beta1"
+	"github.com/aws/karpenter/pkg/controllers/nodeclaim/tagging"
+	"github.com/aws/karpenter/pkg/fake"
+	"github.com/aws/karpenter/pkg/providers/instance"
+	"github.com/aws/karpenter/pkg/test"
+
+	"github.com/aws/karpenter-core/pkg/operator/controller"
+	"github.com/aws/karpenter-core/pkg/operator/scheme"
+)
+
+var ctx context.Context
+var awsEnv *test.Environment
+var env *coretest.Environment
+var taggingController controller.Controller
+
+func TestAPIs(t *testing.T) {
+	ctx = TestContextWithLogger(t)
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "TaggingController")
+}
+
+var _ = BeforeSuite(func() {
+	env = coretest.NewEnvironment(scheme.Scheme, coretest.WithCRDs(apis.CRDs...))
+	ctx = coresettings.ToContext(ctx, coretest.Settings())
+	ctx = settings.ToContext(ctx, test.Settings())
+	awsEnv = test.NewEnvironment(ctx, env)
+	taggingController = tagging.NewController(env.Client, awsEnv.InstanceProvider)
+})
+var _ = AfterSuite(func() {
+	Expect(env.Stop()).To(Succeed(), "Failed to stop environment")
+})
+
+var _ = BeforeEach(func() {
+	awsEnv.Reset()
+})
+
+var _ = AfterEach(func() {
+	ExpectCleanedUp(ctx, env.Client)
+})
+
+var _ = Describe("TaggingController", func() {
+	var ec2Instance *ec2.Instance
+
+	BeforeEach(func() {
+		ec2Instance = &ec2.Instance{
+			State: &ec2.InstanceState{
+				Name: aws.String(ec2.InstanceStateNameRunning),
+			},
+			Tags: []*ec2.Tag{
+				{
+					Key:   aws.String(fmt.Sprintf("kubernetes.io/cluster/%s", settings.FromContext(ctx).ClusterName)),
+					Value: aws.String("owned"),
+				},
+				{
+					Key:   aws.String(corev1beta1.NodePoolLabelKey),
+					Value: aws.String("default"),
+				},
+				{
+					Key:   aws.String(corev1beta1.ManagedByAnnotationKey),
+					Value: aws.String(settings.FromContext(ctx).ClusterName),
+				},
+			},
+			PrivateDnsName: aws.String(fake.PrivateDNSName()),
+			Placement: &ec2.Placement{
+				AvailabilityZone: aws.String(fake.DefaultRegion),
+			},
+			InstanceId:   aws.String(fake.InstanceID()),
+			InstanceType: aws.String("m5.large"),
+		}
+
+		awsEnv.EC2API.Instances.Store(*ec2Instance.InstanceId, ec2Instance)
+	})
+
+	It("shouldn't tag instances without a Node", func() {
+		nodeClaim := coretest.NodeClaim(corev1beta1.NodeClaim{
+			Status: corev1beta1.NodeClaimStatus{
+				ProviderID: fake.ProviderID(*ec2Instance.InstanceId),
+			},
+		})
+
+		ExpectApplied(ctx, env.Client, nodeClaim)
+		ExpectReconcileSucceeded(ctx, taggingController, client.ObjectKeyFromObject(nodeClaim))
+		Expect(nodeClaim.Annotations).To(Not(HaveKey(v1beta1.AnnotationInstanceTagged)))
+		Expect(lo.ContainsBy(ec2Instance.Tags, func(tag *ec2.Tag) bool {
+			return *tag.Key == tagging.TagName
+		})).To(BeFalse())
+	})
+
+	It("shouldn't tag nodeclaim with a malformed provderID", func() {
+		nodeClaim := coretest.NodeClaim(corev1beta1.NodeClaim{
+			Status: corev1beta1.NodeClaimStatus{
+				ProviderID: "Bad providerID",
+				NodeName:   "default",
+			},
+		})
+
+		ExpectApplied(ctx, env.Client, nodeClaim)
+		ExpectReconcileSucceeded(ctx, taggingController, client.ObjectKeyFromObject(nodeClaim))
+		Expect(nodeClaim.Annotations).To(Not(HaveKey(v1beta1.AnnotationInstanceTagged)))
+		Expect(lo.ContainsBy(ec2Instance.Tags, func(tag *ec2.Tag) bool {
+			return *tag.Key == tagging.TagName
+		})).To(BeFalse())
+	})
+
+	It("should gracefully handle missing NodeClaim", func() {
+		nodeClaim := coretest.NodeClaim(corev1beta1.NodeClaim{
+			Status: corev1beta1.NodeClaimStatus{
+				ProviderID: fake.ProviderID(*ec2Instance.InstanceId),
+				NodeName:   "default",
+			},
+		})
+
+		ExpectApplied(ctx, env.Client, nodeClaim)
+		ExpectDeleted(ctx, env.Client, nodeClaim)
+		ExpectReconcileSucceeded(ctx, taggingController, client.ObjectKeyFromObject(nodeClaim))
+	})
+
+	It("should gracefully handle missing instance", func() {
+		nodeClaim := coretest.NodeClaim(corev1beta1.NodeClaim{
+			Status: corev1beta1.NodeClaimStatus{
+				ProviderID: fake.ProviderID(*ec2Instance.InstanceId),
+				NodeName:   "default",
+			},
+		})
+
+		ExpectApplied(ctx, env.Client, nodeClaim)
+		awsEnv.EC2API.Instances.Delete(*ec2Instance.InstanceId)
+		ExpectReconcileSucceeded(ctx, taggingController, client.ObjectKeyFromObject(nodeClaim))
+		Expect(nodeClaim.Annotations).To(Not(HaveKey(v1beta1.AnnotationInstanceTagged)))
+	})
+
+	It("shouldn't tag nodeclaim with deletion timestamp set", func() {
+		nodeClaim := coretest.NodeClaim(corev1beta1.NodeClaim{
+			Status: corev1beta1.NodeClaimStatus{
+				ProviderID: fake.ProviderID(*ec2Instance.InstanceId),
+				NodeName:   "default",
+			},
+			ObjectMeta: v1.ObjectMeta{
+				Finalizers: []string{"testing/finalizer"},
+			},
+		})
+
+		ExpectApplied(ctx, env.Client, nodeClaim)
+		Expect(env.Client.Delete(ctx, nodeClaim)).To(Succeed())
+		ExpectReconcileSucceeded(ctx, taggingController, client.ObjectKeyFromObject(nodeClaim))
+		Expect(nodeClaim.Annotations).To(Not(HaveKey(v1beta1.AnnotationInstanceTagged)))
+		Expect(lo.ContainsBy(ec2Instance.Tags, func(tag *ec2.Tag) bool {
+			return *tag.Key == tagging.TagName
+		})).To(BeFalse())
+	})
+
+	DescribeTable(
+		"should tag taggable instances",
+		func(customTags ...string) {
+			nodeClaim := coretest.NodeClaim(corev1beta1.NodeClaim{
+				Status: corev1beta1.NodeClaimStatus{
+					ProviderID: fake.ProviderID(*ec2Instance.InstanceId),
+					NodeName:   "default",
+				},
+			})
+
+			for _, tag := range customTags {
+				ec2Instance.Tags = append(ec2Instance.Tags, &ec2.Tag{
+					Key:   aws.String(tag),
+					Value: aws.String("custom-tag"),
+				})
+			}
+			awsEnv.EC2API.Instances.Store(*ec2Instance.InstanceId, ec2Instance)
+
+			ExpectApplied(ctx, env.Client, nodeClaim)
+			ExpectReconcileSucceeded(ctx, taggingController, client.ObjectKeyFromObject(nodeClaim))
+			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+			Expect(nodeClaim.Annotations).To(HaveKey(v1beta1.AnnotationInstanceTagged))
+
+			expectedTags := map[string]string{
+				tagging.TagName:      nodeClaim.Status.NodeName,
+				tagging.TagNodeClaim: nodeClaim.Name,
+			}
+			instanceTags := instance.NewInstance(ec2Instance).Tags
+			for tag, value := range expectedTags {
+				if lo.Contains(customTags, tag) {
+					value = "custom-tag"
+				}
+				Expect(instanceTags).To(HaveKeyWithValue(tag, value))
+			}
+		},
+		Entry("with only karpenter.k8s.aws/nodeclaim tag", tagging.TagName),
+		Entry("with only Name tag", tagging.TagNodeClaim),
+		Entry("with both Name and karpenter.k8s.aws/nodeclaim tags"),
+		Entry("with nothing to tag", tagging.TagName, tagging.TagNodeClaim),
+	)
+})
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -103,23 +103,10 @@ func (p *Provider) Create(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, 
 }
 
 func (p *Provider) Link(ctx context.Context, id, provisionerName string) error {
-	_, err := p.ec2api.CreateTagsWithContext(ctx, &ec2.CreateTagsInput{
-		Resources: aws.StringSlice([]string{id}),
-		Tags: []*ec2.Tag{
-			{
-				Key:   aws.String(v1alpha5.MachineManagedByAnnotationKey),
-				Value: aws.String(settings.FromContext(ctx).ClusterName),
-			},
-			{
-				Key:   aws.String(v1alpha5.ProvisionerNameLabelKey),
-				Value: aws.String(provisionerName),
-			},
-		},
-	})
-	if err != nil {
-		if awserrors.IsNotFound(err) {
-			return cloudprovider.NewNodeClaimNotFoundError(fmt.Errorf("linking tags, %w", err))
-		}
+	if err := p.CreateTags(ctx, id, map[string]string{
+		v1alpha5.MachineManagedByAnnotationKey: settings.FromContext(ctx).ClusterName,
+		v1alpha5.ProvisionerNameLabelKey:       provisionerName,
+	}); err != nil {
 		return fmt.Errorf("linking tags, %w", err)
 	}
 	return nil
@@ -185,6 +172,22 @@ func (p *Provider) Delete(ctx context.Context, id string) error {
 			err = multierr.Append(err, e)
 		}
 		return fmt.Errorf("terminating instance, %w", err)
+	}
+	return nil
+}
+
+func (p *Provider) CreateTags(ctx context.Context, id string, tags map[string]string) error {
+	ec2Tags := lo.MapToSlice(tags, func(key, value string) *ec2.Tag {
+		return &ec2.Tag{Key: aws.String(key), Value: aws.String(value)}
+	})
+	if _, err := p.ec2api.CreateTagsWithContext(ctx, &ec2.CreateTagsInput{
+		Resources: aws.StringSlice([]string{id}),
+		Tags:      ec2Tags,
+	}); err != nil {
+		if awserrors.IsNotFound(err) {
+			return cloudprovider.NewNodeClaimNotFoundError(fmt.Errorf("tagging instance, %w", err))
+		}
+		return fmt.Errorf("tagging instance, %w", err)
 	}
 	return nil
 }
@@ -259,9 +262,6 @@ func getTags(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, nodeClaim *co
 			v1alpha5.MachineManagedByAnnotationKey:                                         settings.FromContext(ctx).ClusterName,
 		}
 	} else {
-		overridableTags = map[string]string{
-			"Name": fmt.Sprintf("%s/%s", corev1beta1.NodePoolLabelKey, nodeClaim.Labels[corev1beta1.NodePoolLabelKey]),
-		}
 		staticTags = map[string]string{
 			fmt.Sprintf("kubernetes.io/cluster/%s", settings.FromContext(ctx).ClusterName): "owned",
 			corev1beta1.NodePoolLabelKey:       nodeClaim.Labels[corev1beta1.NodePoolLabelKey],
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -119,4 +119,5 @@ var (
 	LabelInstanceAcceleratorManufacturer      = Group + "/instance-accelerator-manufacturer"
 	LabelInstanceAcceleratorCount             = Group + "/instance-accelerator-count"
 	AnnotationNodeClassHash                   = Group + "/nodeclass-hash"
+	AnnotationInstanceTagged                  = Group + "/tagged"
 )
--- a/pkg/fake/ec2api.go
+++ b/pkg/fake/ec2api.go
@@ -230,9 +230,13 @@ func (e *EC2API) CreateTagsWithContext(_ context.Context, input *ec2.CreateTagsI
 			instance := raw.(*ec2.Instance)
 
 			// Upsert any tags that have the same key
-			newTagKeys := sets.New(lo.Map(input.Tags, func(t *ec2.Tag, _ int) string { return aws.StringValue(t.Key) })...)
-			instance.Tags = lo.Filter(input.Tags, func(t *ec2.Tag, _ int) bool { return newTagKeys.Has(aws.StringValue(t.Key)) })
-			instance.Tags = append(instance.Tags, input.Tags...)
+			tagsToMap := func(tag *ec2.Tag) (string, string) {
+				return *tag.Key, *tag.Value
+			}
+			tags := lo.Assign(lo.SliceToMap(instance.Tags, tagsToMap), lo.SliceToMap(input.Tags, tagsToMap))
+			instance.Tags = lo.MapToSlice(tags, func(key, value string) *ec2.Tag {
+				return &ec2.Tag{Key: aws.String(key), Value: aws.String(value)}
+			})
 		}
 		return nil, nil
 	})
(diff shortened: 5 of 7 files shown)


pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,179 @@
...
+	// Rate limit to one API call per tag period
+	<-c.timer.C
+	_, err := c.ec2.CreateTags(&ec2.CreateTagsInput{
+		Resources: []*string{&resourceID},
+		Tags: []*ec2.Tag{{
+			Key:   aws.String("Name"),
+			Value: &nodeClaim.Status.NodeName,
+		}},
+	})
+	c.timer.Reset(tagPeriod)
```
jonathan-innis, line 172: `time.Sleep(time.Second)`?

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,179 @@
...
+			WithEventFilter(onlyTaggableInstances()),
+	)
+}
+
+func onlyTaggableInstances() predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(object client.Object) bool {
+		nodeClaim := object.(*corev1beta1.NodeClaim)
+
+		// Instance has already been tagged
+		if val, ok := nodeClaim.Annotations[v1beta1.AnnotationInstanceTagged]; ok && val == "true" {
```
jonathan-innis, line 108: This doesn't work in all cases where there is already another update event that has made it through the filter while the event is being processed. There should be a check after the `Get` in the reconcile that validates this same condition as the check on the event filter to reduce the calls as much as possible

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	nodeClaim := &corev1beta1.NodeClaim{}
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, nodeClaim); err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// Extract resource ID from node spec
+	resourceID := path.Base(nodeClaim.Status.ProviderID)
```
jonathan-innis, line 70: There's a `utils` function right now that parses out this ProviderID for you

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	// Check if the instance has the default tag (karpenter.sh/provisioner-name/*). If so it has either:
+	//   a) Already been tagged by us but it is not reflected on the API server
+	//   b) Tagged by some other process. In this case we will respect that tag.
+	if !c.defaultTagRgx.MatchString(tag) {
+		nodeClaim.Annotations[v1beta1.AnnotationInstanceTagged] = "true"
+		err := c.kubeClient.Update(ctx, nodeClaim)
```
jonathan-innis, line 132: We should consider using a `Patch` operation here since this call can fail due to a conflict with [optimistic locking](https://www.learnsteps.com/what-is-optimistic-concurrency-how-does-is-it-help-to-scale-the-kubernetes-cluster/). See the [NodeTemplate controller](https://github.com/aws/karpenter/blob/main/pkg/controllers/nodeclass/controller.go#L71) for a good example of how we can check if things are changed and then do a Patch operation if things updated.

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,179 @@
...
+		return true, client.IgnoreNotFound(err)
+	}
+
+	return false, nil
+}
+
+func (c *Controller) getTagValue(resourceID string, tagKey string) (string, error) {
```
jonathan-innis, line 139: Is there a way that we could reference some of the existing tag calls or abstract some of the existing tag calls using the instance provider?

jonathan-innis: Also, if we do a `Describe` call like this, we don't get the benefit of our current batching logic on our API calls. If we use our [`Get` call on our instance provider](https://github.com/aws/karpenter/blob/main/pkg/providers/instance/instance.go#L128), we are re-using existing logic _and_ if we ever decide to increase the number of `DescribeInstance` calls we do later, we get the benefit of batching

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,179 @@
...
+		timer:         time.NewTimer(tagPeriod),
+	}
+}
+
+func (c *Controller) Name() string {
+	return "ec2-instance-tagger"
+}
+
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	nodeClaim := &corev1beta1.NodeClaim{}
```
jonathan-innis, line 64: We should filter out cases here where:
1. The NodeClaim has an annotation that it has already been tagged
2. The NodeClaim doesn't have its `status.nodeName` value resolved yet (so the value is equal to `""`

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,179 @@
...
+		return true
+	})
+	if err != nil {
+		return "", err
+	}
+
+	return lo.Ternary(tag == nil, "", *tag.Value), nil
+}
+
+func (c *Controller) tagInstance(ctx context.Context, nodeClaim *corev1beta1.NodeClaim, resourceID string) error {
```
jonathan-innis, line 162: I think the intention of tagging the instance is that we tag the instance with the node name instead of the nodeClaim name. These values will be different. I think, ideally, we can add both to the instance so that we get the following tags on the instance

```yaml
tags:
  karpenter.sh/nodeclaim: nodeClaim.Status.Name
  Name: nodeClaim.Status.NodeName
```

pkg/controllers/nodeclaim/tagging/controller.go
jonathan-innis: Since we are looking at updating this annotation in two different places in two different functions, we should maybe even look at pulling out the `Patch` operation from this function and putting it at the end of the `Reconcile` loop so that we are patching once at the end of the reconcile.

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,179 @@
...
+func (c *Controller) isInstanceTagged(ctx context.Context, nodeClaim *corev1beta1.NodeClaim, resourceID string) (bool, error) {
+	tag, err := c.getTagValue(resourceID, "Name")
+	if err != nil {
+		return false, err
+	}
+
+	// Check if the instance has the default tag (karpenter.sh/provisioner-name/*). If so it has either:
+	//   a) Already been tagged by us but it is not reflected on the API server
+	//   b) Tagged by some other process. In this case we will respect that tag.
+	if !c.defaultTagRgx.MatchString(tag) {
```
jonathan-innis, line 130: Do we need to check the value with regex at all? Is it enough to check if there is a value set?

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,144 @@
...
+					Operator: v1.NodeSelectorOpIn,
+					Values:   []string{"nitro"},
+				},
+			},
+			Limits: v1.ResourceList{},
+		}
+	})
+
+	It("should overwrite default Name tags", func() {
+		nodeTemplate := awstest.AWSNodeTemplate(nodeTemplateOpts)
```
jonathan-innis, line 70: Can we create the entire nodeTemplate and the entire Provisioner at the top-level and then just make field-specific overrides in each It block when we need to? I think this should make the overall test look a bit cleaner with less setup happening at the beginning of each It block

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,144 @@
...
+		provisioner := test.Provisioner(provisionerOpts, test.ProvisionerOptions{ProviderRef: &v1alpha5.MachineTemplateRef{
```
jonathan-innis, line 71: Should we be validating this works for Provisioners or only just for NodePools?

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,179 @@
+/*
```
jonathan-innis, line 1: Can we create a `suite_test.go` for this controller. Looking at `pkg/controllers/nodeclass/suite_test.go` might be a good start to look at what the testing should look like and how the mocks can get setup

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,144 @@
...
+		env.ExpectCreated(provisioner, nodeTemplate, pod)
+		env.EventuallyExpectCreatedMachineCount("==", 1)
+		env.EventuallyExpectCreatedNodeCount("==", 1)
+		node := env.EventuallyExpectInitializedNodeCount("==", 1)[0]
+		nodeName := client.ObjectKeyFromObject(node)
+
+		Eventually(func(g Gomega) {
+			node = &v1.Node{}
+			lo.Must0(env.Client.Get(env.Context, nodeName, node))
+			g.Expect(node.Annotations["karpenter.sh/instance-tagged"]).To(Equal("true"))
```
jonathan-innis, line 85:
```suggestion
			g.Expect(node.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationInstanceTagged, "true"))
```

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+func getResourceID(providerID string) (string, error) {
+	id := path.Base(providerID)
+	if id == "." || id == "/" {
+		return "", fmt.Errorf("failed to extract resourceID from providerID")
+	}
+	return id, nil
+}
+
+func getResourceTags(resourceID string) map[string]string {
```
jonathan-innis, line 130: Checkout the `test/pkg/environment/aws/expectations.go` file. I think there may be some functions in there (like `env.GetInstance`) that should help you be able to get some of this information without having to build-out your own functions

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,144 @@
...
+			g.Expect(node.Annotations["karpenter.sh/instance-tagged"]).To(Equal("true"))
+		}, 30*time.Second)
+
+		resourceID := lo.Must(getResourceID(node.Spec.ProviderID))
+		tags := getResourceTags(resourceID)
+		Expect(tags["Name"]).To(Equal("custom-name"))
+	})
+})
+
+func getResourceID(providerID string) (string, error) {
```
jonathan-innis, line 122: `utils.ParseInstanceID`

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,144 @@
...
+			lo.Must0(env.Client.Get(env.Context, nodeName, node))
+			g.Expect(node.Annotations["karpenter.sh/instance-tagged"]).To(Equal("true"))
+		}, 30*time.Second)
+
+		resourceID := lo.Must(getResourceID(node.Spec.ProviderID))
+		tags := getResourceTags(resourceID)
+		Expect(tags["Name"]).To(Equal(node.Name))
+	})
+
+	It("shouldn't overwrite custom Name tags", func() {
```
jonathan-innis, line 93: Can we add a test case for validating that we don't tag instances that are provisioned from Provisioners/Machines?

jonathan-innis (commented):
Overall, super solid! Most of my comments have to do with looking around for existing functions and frameworks that do a lot of the same stuff that you are doing so that we stay DRY in the codebase. Great work 🎉

jonathan-innis (commented):
/karpenter snapshot

pkg/controllers/nodeclaim/tagging/controller.go
jmdeal: I guess I made the assumption that even if we removed our default naming strategy it would still be assigned a default name by AWS. The idea here was to ensure that it was set to our default name `karpenter.sh/nodepool/.*` rather than a custom name that may have been set through some other mechanism.

pkg/controllers/nodeclaim/tagging/controller.go
jmdeal: Right now I think it is being tagged with the node name, the entire `nodeClaim` is passed in to also update it's annotation but like you suggested in a different comment I'm going to refactor that out into the reconcile loop. Would we be able to annotate with the node claim name as part of the create fleet call rather than as part of this controller?

test/suites/integration/instance_tag_test.go
jmdeal: Sure, I'm working on changing the tests to use a `NodePool` and `NodeClass` now and have made that change.

test/suites/integration/instance_tag_test.go
jmdeal: Really only NodePools, I'm updating the tests to reflect that. You mentioned in another comment a test that verifies it doesn't tag instances backed by machines rather than node claims which seems like a good idea so a provisioner will stick around at least for that test.

pkg/controllers/nodeclaim/tagging/controller.go
jonathan-innis: With this change, we should remove the tagging of the instance in the `CreateFleet` call for `NodeClaims`. That way, the instance won't be tagged with a `Name` by default and then we'll know that if it isn't tagged, we should tag it

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,158 @@
...
+	//   b) Tagged by some other process. In this case we will respect that tag.
+	if !c.defaultTagRgx.MatchString(tag) {
+		return true, client.IgnoreNotFound(err)
+	}
+	return false, nil
+}
+
+func (c *Controller) tagInstance(nodeClaim *corev1beta1.NodeClaim, resourceID string) error {
+	// Rate limit to one API call per tag period
+	<-c.timer.C
```
jonathan-innis, line 135: Why are we introducing a timer concept here and not using something more native like `time.Sleep`

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,158 @@
...
+	defer c.timer.Reset(time.Second)
+
+	_, err := c.ec2.CreateTags(&ec2.CreateTagsInput{
```
jonathan-innis, line 138: Can you delegate this creation to the instance provider?

jonathan-innis: That allows us to abstract out this call and potentially batch and cache things if we need to

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,158 @@
...
+		Resources: []*string{&resourceID},
+		Tags: []*ec2.Tag{{
+			Key:   aws.String("Name"),
+			Value: &nodeClaim.Status.NodeName,
+		}},
+	})
+	return err
+}
+
+func isNodeClaimTaggable(nc *corev1beta1.NodeClaim) bool {
```
jonathan-innis, line 148: I'm not opposed to breaking this out, but I also think it's easier if this is just inlined. It's a trade-off between someone having to do an extra hop to validate what the `isNodeClaimTaggable` function does vs. just looking at the logic inline for filtering

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,158 @@
...
+			NewControllerManagedBy(m).
+			For(&corev1beta1.NodeClaim{}).
+			WithEventFilter(predicate.NewPredicateFuncs(func(o client.Object) bool {
+				nc := o.(*corev1beta1.NodeClaim)
+				return isNodeClaimTaggable(nc)
+			})),
+	)
+}
+
+func (c *Controller) isInstanceTagged(ctx context.Context, resourceID string) (bool, error) {
```
jonathan-innis, line 113: Nice 🎉

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,158 @@
...
+	if !isTagged {
+		err = c.tagInstance(nodeClaim, resourceID)
+		if err != nil {
+			return reconcile.Result{Requeue: true}, client.IgnoreNotFound(err)
+		}
+	}
+
+	storedNodeClaim := nodeClaim.DeepCopy()
+	nodeClaim.Annotations[v1beta1.AnnotationInstanceTagged] = "true"
+	err = c.kubeClient.Patch(ctx, nodeClaim, client.MergeFrom(storedNodeClaim))
```
jonathan-innis, line 93: We should only patch, if we know that the annotations actually changed here, so we should do something like `if !equality.semantic.DeepEqual` like in the other controllers that perform patch

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,158 @@
...
+		defaultTagRgx:    regexp.MustCompile(corev1beta1.NodePoolLabelKey + "/.*"),
+		timer:            time.NewTimer(0),
+	}
+}
+
+func (c *Controller) Name() string {
+	return "ec2-instance-tagger"
+}
+
+func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
```
jonathan-innis, line 61: Look at using the `TypedController` which will automatically inject the typed object into the `Reconcile` function. See: https://github.com/aws/karpenter/blob/main/pkg/controllers/nodeclass/controller.go#L149

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,158 @@
...
+	kubeClient       client.Client
+	instanceProvider *instance.Provider
+	ec2              *ec2.EC2
+	defaultTagRgx    *regexp.Regexp
+	timer            *time.Timer
+}
+
+func NewController(kubeClient client.Client, sess *session.Session, instanceProvider *instance.Provider) *Controller {
+	return &Controller{
+		kubeClient:       kubeClient,
```
jonathan-innis, line 49: I think if we collapse a lot of our logic down, we can collapse this constructor down to just take in the `instanceProvider` and the `kubeClient`

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/samber/lo"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
+	corev1beta1 "github.com/aws/karpenter-core/pkg/apis/v1beta1"
+	"github.com/aws/karpenter-core/pkg/test"
```
jonathan-innis, line 28:
```suggestion
	coretest "github.com/aws/karpenter-core/pkg/test"
```

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/aws/karpenter/pkg/apis/settings"
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+	"github.com/aws/karpenter/pkg/apis/v1beta1"
+	"github.com/aws/karpenter/pkg/providers/instance"
+	awstest "github.com/aws/karpenter/pkg/test"
```
jonathan-innis, line 33:
```suggestion
     "github.com/aws/karpenter/pkg/test"
```

We've been aliasing this one as test and the one that references the core package as `coretest` elsewhere. It would be nice to keep this for consistency

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,135 @@
...
+		pod := test.Pod()
+
+		env.ExpectCreated(nodePool, nodeClass, pod)
+		env.EventuallyExpectCreatedNodeCount("==", 1)
+		node := env.EventuallyExpectInitializedNodeCount("==", 1)[0]
+		nodeName := client.ObjectKeyFromObject(node)
+
+		Eventually(func(g Gomega) {
+			node = &v1.Node{}
+			lo.Must0(env.Client.Get(env.Context, nodeName, node))
```
jonathan-innis, line 98: Use `g.Expect` here; otherwise, a failure on `lo.Must0` will cause a panic and cause the entire test case to crash

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,135 @@
...
```
jonathan-innis, line 128: Same comment here about `lo.Must0` in eventually statements

pkg/controllers/nodeclaim/tagging/controller.go
jmdeal: The intention was to ensure that there was at least one second between each call across reconciliations but I suppose since concurrent reconciliations == 1 `time.Sleep` would have pretty much the same effect. It would overall make each reconciliation longer but I don't think there would really be any consequences to that.

pkg/controllers/nodeclaim/tagging/controller.go
jmdeal: The idea was to break it out since it's used both for the event filter and at the beginning of each reconciliation loop. I don't think it would be a good idea to duplicate that inline in case conditions ever change.

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,167 @@
...
+
+func NewController(kubeClient client.Client, instanceProvider *instance.Provider) corecontroller.Controller {
+	return corecontroller.Typed[*corev1beta1.NodeClaim](kubeClient, &Controller{
+		kubeClient:       kubeClient,
+		instanceProvider: instanceProvider,
+	})
+}
+
+func (c *Controller) Name() string {
+	return "ec2-instance-tagger"
```
jonathan-innis, line 57:
```suggestion
	return "nodeclaim.tagging"
```

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClaim *corev1beta1.NodeClaim) (reconcile.Result, error) {
+	if !isNodeClaimTaggable(nodeClaim) {
```
jonathan-innis, line 61:
```suggestion
	if !isTaggable(nodeClaim) {
```

nit: optional since this is already scoped to `NodeClaims` through the args

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,167 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	// Extract resource ID from node spec
+	// TODO: Should we retry here? If the ProviderID is malformed is there a case where it is later repaired?
+	resourceID, err := utils.ParseInstanceID(nodeClaim.Status.ProviderID)
```
jonathan-innis, line 67: We shouldn't retry, but we should log the error. Backing-off won't do us any good until the value is updated

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Check if instance has already been tagged
+	isTagged, err := c.isInstanceTagged(ctx, resourceID)
```
jonathan-innis, line 73: I could see all of this check if it's tagging logic and getting the tags to be collapsed into a single function

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if len(lo.PickByValues(isTagged, []bool{false})) != 0 {
+		tags := getTags(isTagged, nodeClaim)
+		err = c.tagInstance(ctx, tags, resourceID)
+		if err != nil {
+			if client.IgnoreNotFound(err) != nil {
```
jonathan-innis, line 82: This doesn't handle the `NotFound` error that would come from AWS. We should probably return something from the `CreateTags` call like `cloudprovider.NewNodeClaimNotFoundError` and then ignore that error here

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,167 @@
...
+				return reconcile.Result{Requeue: true}, err
```
jonathan-innis, line 83:
```suggestion
				return reconcile.Result{}, err
```

No need to specify `Requeue: true` if you return an error. It does the same thing. Logic is here: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/controller/controller.go#L317

pkg/providers/instance/instance.go
```diff
@@ -103,17 +103,14 @@ func (p *Provider) Create(ctx context.Context, nodeClass *v1beta1.NodeClass, nod
 }
 
 func (p *Provider) Link(ctx context.Context, id, provisionerName string) error {
-	_, err := p.ec2api.CreateTagsWithContext(ctx, &ec2.CreateTagsInput{
```
jonathan-innis, line 106: Nice!

pkg/providers/instance/instance.go
```diff
@@ -189,6 +186,14 @@ func (p *Provider) Delete(ctx context.Context, id string) error {
 	return nil
 }
 
+func (p *Provider) CreateTags(ctx context.Context, id string, tags []*ec2.Tag) error {
```
jonathan-innis, line 189:
```suggestion
func (p *Provider) CreateTags(ctx context.Context, id string, tags map[string]string) error {
```

Consider taking in a map here

pkg/providers/instance/instance.go
```diff
@@ -259,9 +264,7 @@ func getTags(ctx context.Context, nodeClass *v1beta1.NodeClass, nodeClaim *corev
 			v1alpha5.MachineManagedByAnnotationKey:                                         settings.FromContext(ctx).ClusterName,
 		}
 	} else {
-		overridableTags = map[string]string{
-			"Name": fmt.Sprintf("%s/%s", corev1beta1.NodePoolLabelKey, nodeClaim.Labels[corev1beta1.NodePoolLabelKey]),
-		}
+		overridableTags = map[string]string{}
```
jonathan-innis, line 267: You can just remove this variable altogether if nothing is going to be in it anymore

jmdeal: Didn't realize `lo.Assign` could take a `nil` map, looks like it is safe though so will do.

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,137 @@
...
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("nodeclaim", nodeClaim.Name))
+	storedNodeClaim := nodeClaim.DeepCopy()
+
+	if !isTaggable(nodeClaim) {
+		return reconcile.Result{}, nil
+	}
+
+	id, err := utils.ParseInstanceID(nodeClaim.Status.ProviderID)
+	if err != nil {
+		logging.FromContext(ctx).Errorf("failed to parse instance ID, %w", err)
```
jonathan-innis, line 71: Maybe a comment on why we don't return the error here

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		if !cloudprovider.IsNodeClaimNotFoundError(err) {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
+
+	nodeClaim.Annotations[v1beta1.AnnotationInstanceTagged] = "true"
+	if !equality.Semantic.DeepEqual(nodeClaim, storedNodeClaim) {
+		err = c.kubeClient.Patch(ctx, nodeClaim, client.MergeFrom(storedNodeClaim))
```
jonathan-innis, line 85: Prefer to one-line this one if you have no need to store the error. e.g.

```go
if err := c.kubeClient.Patch(ctx, nodeClaim, client.MergeFrom(storedNodeClaim)); err != nil {
  return reconcile.Result{}, client.IgnoreNotFound(err)
}
```

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,137 @@
...
+		instanceProvider: instanceProvider,
+	})
+}
+
+func (c *Controller) Name() string {
+	return "nodeclaim.tagging"
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeClaim) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("nodeclaim", nodeClaim.Name))
```
jonathan-innis, line 62: nit: This is already injected by the typed controller so you don't need this line anymore

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,137 @@
...
+		tagName:      nc.Status.NodeName,
+		tagNodeClaim: nc.Name,
+	}
+
+	// Remove tags which have been already populated
+	instance, err := c.instanceProvider.Get(ctx, id)
+	if err != nil {
+		return err
+	}
+	tags = lo.OmitByKeys(tags, lo.Keys(instance.Tags))
```
jonathan-innis, line 117: Nice 🚀

jonathan-innis (commented):
This looks really good to me! Just the functional testing in here and this looks good to go!

jmdeal (commented):
/karpenter snapshot

pkg/providers/instance/instance.go
```diff
@@ -189,6 +177,23 @@ func (p *Provider) Delete(ctx context.Context, id string) error {
 	return nil
 }
 
+func (p *Provider) CreateTags(ctx context.Context, id string, tags map[string]string) error {
+	ec2Tags := lo.MapToSlice(tags, func(key, value string) *ec2.Tag {
+		return &ec2.Tag{Key: aws.String(key), Value: aws.String(value)}
+	})
+	_, err := p.ec2api.CreateTagsWithContext(ctx, &ec2.CreateTagsInput{
```
jonathan-innis, line 184: You could collapse this line with the if statement on L188

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,181 @@
...
+		}})
+
+		nodePool = coretest.NodePool(corev1beta1.NodePool{
+			Spec: corev1beta1.NodePoolSpec{
+				Template: corev1beta1.NodeClaimTemplate{
+					Spec: corev1beta1.NodeClaimSpec{
+						NodeClass: &corev1beta1.NodeClassReference{
+							Name: nodeClass.Name,
+						},
+						Requirements: []v1.NodeSelectorRequirement{
```
jonathan-innis, line 61: This seems like a lot of requirements to be setting for this test case. Is there a reason that you needed to set this many requirements to get the test working?

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,181 @@
...
+							},
+						},
+					},
+				},
+				Limits: corev1beta1.Limits{},
+			},
+		})
+	})
+
+	It("should overwrite default Name tags", func() {
```
jonathan-innis, line 90:
```suggestion
	It("should tag with karpenter.k8s.was/nodeclaim and Name tags", func() {
```

jonathan-innis (commented):
Really nice testing coverage 🎉 Just a few comments!

test/suites/integration/instance_tag_test.go
```diff
@@ -0,0 +1,181 @@
+/*
```
jonathan-innis, line 1: Since we already have a `tags_test.go` file, should we just use this same file with a different `Context` block for these tests?

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,136 @@
...
+
+	if len(tags) != 0 {
+		err = c.instanceProvider.CreateTags(ctx, id, tags)
+		time.Sleep(time.Second)
```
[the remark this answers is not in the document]
njtran: Is there a better solution here and have we seen this happening in testing? It strikes me as an unnecessary precaution and slowness for users that may not run into the rate limiting here.

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Remove tags which have been already populated
+	instance, err := c.instanceProvider.Get(ctx, id)
+	if err != nil {
+		return err
+	}
+	tags = lo.OmitByKeys(tags, lo.Keys(instance.Tags))
+
+	if len(tags) != 0 {
+		err = c.instanceProvider.CreateTags(ctx, id, tags)
```
njtran, line 119: Creating tags should be idempotent, which means we shouldn't need to do another instanceProvider.Get() call to get the tags here. Can we just do the `createTags` call here without a `instanceProvider.Get()` since we're already doing a short circuit in the controller with the node annotation?

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,136 @@
...
+		time.Sleep(time.Second)
+		return err
+	}
+	return nil
+}
+
+func isTaggable(nc *corev1beta1.NodeClaim) bool {
+	// Instance has already been tagged
+	if val := nc.Annotations[v1beta1.AnnotationInstanceTagged]; val == "true" {
```
njtran, line 128: Is it possible that we'll want to dynamically tag for a nodeClaim multiple times in the future? For instance, let's say we want to make sure that instance tags sync up with a specific set of values, I'm assuming we'll probably have to remove this check on the annotation, and then probably remove the annotation altogether. I'd rather not add an annotation if we know we're going to remove it anyway.

pkg/providers/instance/instance.go
```diff
@@ -189,6 +177,23 @@ func (p *Provider) Delete(ctx context.Context, id string) error {
...
+		Resources: aws.StringSlice([]string{id}),
+		Tags:      ec2Tags,
+	})
+	if err != nil {
+		if awserrors.IsNotFound(err) {
+			return cloudprovider.NewNodeClaimNotFoundError(fmt.Errorf("tagging resource, %w", err))
+		}
+		return fmt.Errorf("tagging resource, %w", err)
```
njtran, line 192: should it be "tagging instance" since we know it's an instance each time? 
```suggestion
			return cloudprovider.NewNodeClaimNotFoundError(fmt.Errorf("tagging instance, %w", err))
		}
		return fmt.Errorf("tagging instance, %w", err)
```

pkg/providers/instance/instance.go
```diff
@@ -259,9 +264,6 @@ func getTags(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, nodeClaim *co
 			v1alpha5.MachineManagedByAnnotationKey:                                         settings.FromContext(ctx).ClusterName,
 		}
 	} else {
-		overridableTags = map[string]string{
-			"Name": fmt.Sprintf("%s/%s", corev1beta1.NodePoolLabelKey, nodeClaim.Labels[corev1beta1.NodePoolLabelKey]),
-		}
```
njtran, line 264: Was this intentional? Will this break users?

jmdeal: Yes it was intentional. Since this only applies to `NodePool`s it won't break users who are still on v1alpha5. The main intention of this PR is to change the value of the `Name` tag to reflect the name of the node rather than the nodepool so that would be a breaking change as part of v1beta1.

pkg/controllers/nodeclaim/tagging/controller.go
jmdeal: We don't want to overwrite tags that the user specifies. A use case scenario for this would be if a user specifies a `Name` tag as part of the `NodeClass` or if they have a separate process that applies the tag. Unless we want to say that Karpenter owns the `Name` tag and it can't be configured by users we need some mechanism to determine if we should tag it. In this case, we check if it has already been tagged.

pkg/controllers/nodeclaim/tagging/controller.go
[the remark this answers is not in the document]
jmdeal: I haven't tested this controller at scale yet but it would very likely be an issue in a large scale-up. The reason we're being cautious here is because `CreateTags` shares a bucket with other mutating API calls and we'd rather this be slow than get rate limited on other, more important operations (e.g. `CreateFleet`). It would be good to get some data to see if this would actually be an issue in practice though.

pkg/controllers/nodeclaim/tagging/controller.go
jmdeal: That seems reasonable to me. The intention behind using an annotation rather than tracking in memory was so it would persist through controller restarts. In a large scale cluster this will result in a high number of `DescribeInstance` calls right at startup but I don't know if that would actually be an issue.

pkg/controllers/nodeclaim/tagging/controller.go
njtran: Could we just check if the user has defined any tags we don't want to override in their NodeClass and then just remove that from our CreateTags call, rather than doing Get() and then Create()? Trying to compare it to the underlying instance means we have to sync and union the tags each time, when I think we can save on API calls by just using the information that Karpenter already knows and not adding that to the createTags call in the first place.

pkg/controllers/nodeclaim/tagging/controller.go
njtran: Our DescribeInstances is [batched](https://github.com/aws/karpenter/blob/d1d1371ae2e1552b8fdded7d343bf24ea18bee31/pkg/batcher/describeinstances.go#L36), so I don't think the number of calls here would be a problem, if anything it would be a problem to have a large queue depth, but that's an issue with other controllers too.

pkg/controllers/nodeclaim/tagging/controller.go
jmdeal: The case that doesn't necessarily cover is if an external process is responsible for tagging. In that case, maybe users should specify a 'default' tag as part of the NodeClass that way we know not to tag it? That seems like a reasonable solution to me, but maybe a little messy?

pkg/controllers/nodeclaim/tagging/controller.go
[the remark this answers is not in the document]
njtran: Can we try to rate-limit this as part of the underlying queue mechanism rather than injecting a sleep into our code?

test/suites/integration/instance_tag_test.go
jmdeal: Nope, probably not. Honestly don't remember why they were there at this point but default requirements should be fine.

pkg/controllers/nodeclaim/tagging/controller.go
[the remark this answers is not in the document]
jmdeal: To my understanding the rate limiter is only used for retries, not for overall requests. If there is a way to get it working like that I absolutely agree that would be better.

pkg/controllers/nodeclaim/tagging/controller.go
njtran: What if we just make sure that we're only adding tags with a Karpenter prefix? That way they can add the Name tag if they want to, but it'll be a different name tag than the one that we set (maybe something like `karpenter.sh/node-name`)

jmdeal: I think that would be reasonable. However, if in the future we want to support additional dynamic tags would it make sense for those to live in the `karpenter.sh` namespace? Also, it would still be nice to support the `Name` tag since that's the visible name for the instance in the AWS console. Maybe we could special case ownership for `Name`? Thoughts @jonathan-innis?

pkg/controllers/nodeclaim/tagging/controller.go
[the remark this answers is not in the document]
njtran: Yeah, after discussing with @jonathan-innis, it seems like this the sleep is the best way to do this 👍

pkg/controllers/nodeclaim/tagging/controller.go
jmdeal: Aligned offline to keep the annotation to minimize API calls upon controller restart as well as minimize the time new requests have to wait after restart.

pkg/controllers/nodeclaim/tagging/controller.go
jmdeal: Alligned offline to keep original implementation (not overwriting existing tags). This model ensures that user specified tags are not overwritten while also giving us the flexibility to create tags not within the karpenter namespace (e.g. `Name`).

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	id, err := utils.ParseInstanceID(nodeClaim.Status.ProviderID)
+	if err != nil {
+		// We don't throw an error here since we don't want to retry until the ProviderID has been updated.
+		logging.FromContext(ctx).Errorf("failed to parse instance ID, %w", err)
+		return reconcile.Result{}, nil
+	}
+
+	err = c.tagInstance(ctx, nodeClaim, id)
```
jonathan-innis, line 75: nit: Collapse into a single line?

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,133 @@
...
+		instanceProvider: instanceProvider,
+	})
+}
+
+func (c *Controller) Name() string {
+	return "nodeclaim.tagging"
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClaim *corev1beta1.NodeClaim) (reconcile.Result, error) {
+	storedNodeClaim := nodeClaim.DeepCopy()
```
jonathan-innis, line 62:
```suggestion
	stored := nodeClaim.DeepCopy()
```

nit: naming convention for other controllers

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,133 @@
...
+func (c *Controller) tagInstance(ctx context.Context, nc *corev1beta1.NodeClaim, id string) error {
+	tags := map[string]string{
+		TagName:      nc.Status.NodeName,
+		TagNodeClaim: nc.Name,
+	}
+
+	// Remove tags which have been already populated
+	instance, err := c.instanceProvider.Get(ctx, id)
+	if err != nil {
+		return err
```
jonathan-innis, line 110:
```suggestion
		return fmt.Errorf("tagging nodeclaim, %w", err)
```

nit: contextualize the error

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	tags = lo.OmitByKeys(tags, lo.Keys(instance.Tags))
+
+	if len(tags) != 0 {
+		err = c.instanceProvider.CreateTags(ctx, id, tags)
```
jonathan-innis, line 115: nit: consider contextualizing this error with `fmt.Errorf()`

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClaim *corev1beta1.NodeClaim) (reconcile.Result, error) {
+	storedNodeClaim := nodeClaim.DeepCopy()
+
+	if !isTaggable(nodeClaim) {
+		return reconcile.Result{}, nil
+	}
+
+	id, err := utils.ParseInstanceID(nodeClaim.Status.ProviderID)
```
jonathan-innis, line 68: Should we add more logging context here 

```go
ctx = logging.WithContext(logging.FromContext(ctx).With("provider-id", nodeClaim.Status.providerID))
```

pkg/controllers/nodeclaim/tagging/suite_test.go
```diff
@@ -0,0 +1,186 @@
...
+var ctx context.Context
+var awsEnv *test.Environment
+var env *coretest.Environment
+var taggingController controller.Controller
+var opts options.Options
+
+func TestAPIs(t *testing.T) {
+	ctx = TestContextWithLogger(t)
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Tagging")
```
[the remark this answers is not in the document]
jonathan-innis: You resolved this nit but didn't change it. Are you rejecting the suggestion or did you just miss changing it?

pkg/controllers/nodeclaim/tagging/suite_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+		ExpectApplied(ctx, env.Client, nodeClaim)
+		ExpectReconcileSucceeded(ctx, taggingController, client.ObjectKeyFromObject(nodeClaim))
+		Expect(nodeClaim.Annotations).To(Not(HaveKey(v1beta1.AnnotationInstanceTagged)))
+		Expect(lo.ContainsBy(ec2Instance.Tags, func(tag *ec2.Tag) bool {
+			return *tag.Key == tagging.TagName
+		})).To(BeFalse())
+	})
+
+	It("should gracefully handle missing NodeClaim", func() {
```
jonathan-innis, line 142: I was thinking that the case where the instance doesn't exist might be more complex here. Should we check for that as well?

jonathan-innis: This is actually a good callout from the test case. Should we ignore tagging instances from NodeClaims that are actively deleting? So check the DeletionTimestamp is set and ignore it if it is

pkg/controllers/nodeclaim/tagging/suite_test.go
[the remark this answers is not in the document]
jmdeal: Missed changing it, thought it was a different nit when I resolved it and didn't come back to it.

pkg/controllers/nodeclaim/tagging/suite_test.go
jmdeal: Both seem reasonable to me 👍

jmdeal (commented):
/karpenter snapshot

pkg/controllers/nodeclaim/tagging/controller.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClaim *corev1beta1.NodeClaim) (reconcile.Result, error) {
+	stored := nodeClaim.DeepCopy()
+
+	if !isTaggable(nodeClaim) {
+		return reconcile.Result{}, nil
+	}
+
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("providerID", nodeClaim.Status.ProviderID))
```
jonathan-innis, line 69:
```suggestion
	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("provider-id", nodeClaim.Status.ProviderID))
```

To match naming elsewhere throughout the codebase

jonathan-innis (approved):
LGTM 🚀 Nice work!

PR #4618, opened 2023-09-13
fix: use k8s version from kubernetes endpoint

njtran:
Fixes #N/A 

**Description**
This changes the version provider to make an https call to the kubernetes api endpoints to get k8s version from all APIServers. In the case where APIServers have differing versions, this will ensure that Drift doesn't flap around, and that the result is consistent. 

- [ ] Yes, issue opened: # 
- [x] No


pkg/providers/version/version.go
```diff
@@ -30,35 +36,94 @@ const (
...
+	if err := p.kubeClient.Get(ctx, types.NamespacedName{Name: "kubernetes", Namespace: "default"}, &endpointSlice); err != nil {
+		return "", fmt.Errorf("getting endpoints, %w", err)
+	}
+	// This error should never occur
+	minVersion, err := version.ParseGeneric("v99.99.99")
+	if err != nil {
+		return "", fmt.Errorf("parsing version, %w", err)
+	}
+	for _, address := range getAddresses(endpointSlice) {
+		resp, err := p.httpClient.Get(address)
```
jonathan-innis, line 90: Does it make sense to use the default config and just update the host to be equal to the IP and port of the APIServer so that we are still using the same config and clientSet concepts rather than building the http client ourselves: https://github.com/kubernetes/client-go/blob/v0.28.1/rest/config.go#L57

pkg/providers/version/version.go
```diff
@@ -30,35 +36,94 @@ const (
...
+		}
+		if v.LessThan(minVersion) {
+			minVersion = v
+		}
+	}
+	// Return only the first two version numbers
+	return strings.Join(strings.Split(minVersion.String(), ".")[0:2], "."), nil
+}
+func getAddresses(endpoints v1.EndpointSlice) []string {
+	ports := []string{}
```
jonathan-innis, line 111:
```suggestion
	var ports []string
```

nit: this is preferred by https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

pkg/providers/version/version.go
```diff
@@ -30,35 +36,94 @@ const (
...
+	// If there are no ports, it's the same as defining all ports.
+	if len(endpoints.Ports) == 0 {
+		ports = []string{""}
+	} else {
+		for _, port := range endpoints.Ports {
+			ports = append(ports, fmt.Sprintf(":%s", strconv.Itoa(int(*port.Port))))
+		}
+	}
+	ret := []string{}
```
jonathan-innis, line 120:
```suggestion
	var ret []string
```

This is preferred by https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

pkg/providers/version/version.go
```diff
@@ -30,35 +36,94 @@ const (
...
+			minVersion = v
+		}
+	}
+	// Return only the first two version numbers
+	return strings.Join(strings.Split(minVersion.String(), ".")[0:2], "."), nil
+}
+func getAddresses(endpoints v1.EndpointSlice) []string {
+	ports := []string{}
+	// If there are no ports, it's the same as defining all ports.
+	if len(endpoints.Ports) == 0 {
```
jonathan-innis, line 113: optional: Initialize ports to be equal to this value and then only iterate if the length is greater than 0?

```go
ports := []string{""}
if len(endpoints.Ports) > 0 {
  ports = lo.Map(endpoints.Ports, func(p *v1.Port, _ int) string { return fmt.Sprintf(":%s", strconv.Itoa(int(*port.Port))) }
}
```

pkg/providers/version/version.go
```diff
@@ -30,35 +36,94 @@ const (
...
+// GetMinKubernetesVersion ensures that we query all known APIServers for the K8s version.
+// This to handle any scenarios where there may be multiple APIServer reporting different
+// K8s versions.
+func (p *Provider) GetMinKubernetesVersion(ctx context.Context) (string, error) {
+	var endpointSlice v1.EndpointSlice
+	if err := p.kubeClient.Get(ctx, types.NamespacedName{Name: "kubernetes", Namespace: "default"}, &endpointSlice); err != nil {
+		return "", fmt.Errorf("getting endpoints, %w", err)
+	}
+	// This error should never occur
+	minVersion, err := version.ParseGeneric("v99.99.99")
```
jonathan-innis, line 85: Is it ever possible that the endpoint slice returns an empty list? In that case, we are returning this version. It seems like in that case, we should error for safety, saying something like "failed to discover the version since we didn't get any apiserver endpoints"

charts/karpenter/templates/clusterrole.yaml
```diff
@@ -29,6 +29,9 @@ metadata:
   {{- end }}
 rules:
   # Read
+  - apiGroups: ["discovery.k8s.io"]
```
jonathan-innis, line 32: Is it possible to constraint this by resource name or namespace? I'm assuming that this permission might be too wide right now since it's cluster-wide.

pkg/providers/version/version.go
njtran: I think there'd have to be no APIServer endpoints, but you bring up a good point. Let me fix this

charts/karpenter/templates/clusterrole.yaml
njtran: Scoped it down by resource name. To constrain it by namespace, I think I'd need to create a new role in the default namespace for karpenter. Thoughts?

pkg/providers/version/version.go
njtran: We need to use the config that's initialized as part of the Operator IIUC. 
> We could alternatively have used the simpler client-go InClusterConfig() method. However, that only works when Karpenter is running as a Pod within the same cluster it's managing.

njtran (commented):
/karpenter snapshot

njtran (commented):
/karpenter snapshot

charts/karpenter/templates/role.yaml
```diff
@@ -48,6 +48,23 @@ rules:
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
+metadata:
```
jonathan-innis, line 51: Is it worth adding comments on these roles around why we need them if it isn't immediately clear what we are doing with these things?

charts/karpenter/templates/rolebinding.yaml
```diff
@@ -20,6 +20,26 @@ subjects:
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
+metadata:
```
jonathan-innis, line 23: Same comment here bout a comment on the RoleBinding

pkg/operator/operator.go
```diff
@@ -127,7 +128,8 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 		ec2api,
 		*sess.Config.Region,
 	)
-	versionProvider := version.NewProvider(operator.KubernetesInterface, cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval))
+	versionProvider := version.NewProvider(operator.KubernetesInterface, cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval),
+		operator.GetClient(), lo.Must(getHTTPClient(operator.GetConfig())))
```
jonathan-innis, line 132: Should this be part of building the operator? Like `operator.GetHTTPClient()`

pkg/providers/version/version.go
```diff
@@ -51,14 +67,77 @@ func (p *Provider) Get(ctx context.Context) (string, error) {
 	if version, ok := p.cache.Get(kubernetesVersionCacheKey); ok {
 		return version.(string), nil
 	}
-	serverVersion, err := p.kubernetesInterface.Discovery().ServerVersion()
-	if err != nil {
-		return "", err
+	var version string
+	var err error
+	// If we're running locally, these environment variables will be empty.
+	host, port := os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT")
```
jonathan-innis, line 73: Nice! Auto-discovery! These will always be set in-cluster? Docs to it?

pkg/providers/version/version.go
```diff
@@ -51,14 +67,77 @@ func (p *Provider) Get(ctx context.Context) (string, error) {
...
 	if p.cm.HasChanged("kubernetes-version", version) {
 		logging.FromContext(ctx).With("version", version).Debugf("discovered kubernetes version")
 	}
 	return version, nil
 }
+
+// GetMinKubernetesVersion ensures that we query all known APIServers for the K8s version.
+// This to handle any scenarios where there may be multiple APIServer reporting different
+// K8s versions.
+func (p *Provider) GetMinKubernetesVersion(ctx context.Context) (string, error) {
```
jonathan-innis, line 98: Should `Get()` just handle this for me so I don't have to make different calls to the provider to understand the difference in this behavior?

pkg/providers/version/version.go
```diff
@@ -51,14 +67,77 @@ func (p *Provider) Get(ctx context.Context) (string, error) {
...
+	if err := p.kubeClient.Get(ctx, types.NamespacedName{Name: "kubernetes", Namespace: "default"}, &endpointSlice); err != nil {
+		return "", fmt.Errorf("getting endpoints, %w", err)
+	}
+	var minVersion *version.Version
+	for _, address := range getAddresses(endpointSlice) {
+		resp, err := p.httpClient.Get(address)
+		if err != nil {
+			return "", fmt.Errorf("sending get request, %w", err)
+		}
+		defer resp.Body.Close()
```
jonathan-innis, line 109: Maybe worth a comment similar to the one that we have in the `webhookChecker` just so we don't remove this later on:

```go
		// Close the body to avoid leaking file descriptors
		// Always read the body so we can re-use the connection: https://stackoverflow.com/questions/17948827/reusing-http-connections-in-go

```

jonathan-innis, line 109: Defer closes the response at the end of the function, which I don't think is what we want here. This isn't a closure, so I think this leaks resources. https://stackoverflow.com/questions/45617758/proper-way-to-release-resources-with-defer-in-a-loop

pkg/providers/version/version.go
njtran: Yeah this is only needed if we're using the http client. I'll make it not public so that users don't have to understand this. Sound good?

pkg/providers/version/version.go
njtran: Ah yes, good catch!

pkg/operator/operator.go
njtran: Since we only need it for the version provider, I think it's sufficient to just construct it inline and then pass it into the one provider. If there's other providers that need the HTTP client, I can imagine us pulling this out in the future.

pkg/providers/version/version.go
njtran: Yeah, specifically this is to figure out if we're running within a pod in the same cluster that Karpenter's managing: https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#directly-accessing-the-rest-api , similar to the InClusterConfig().

charts/karpenter/templates/role.yaml
```diff
@@ -46,6 +46,25 @@ rules:
     resources: ["configmaps"]
     verbs: ["create"]
 ---
+# We need this role so that Karpenter can discover the Kubernetes APIServer Endpoints
```
jonathan-innis, line 49:
```suggestion
# Karpenter uses this role to discover the Kubernetes APIServer Endpoints
```

charts/karpenter/templates/role.yaml
```diff
@@ -46,6 +46,25 @@ rules:
...
+# This is important for getting consensus on the API Version in the cluster.
```
jonathan-innis, line 50:
```suggestion
# This is important for getting consensus on the Kubernetes Version in the cluster.
```

charts/karpenter/templates/rolebinding.yaml
```diff
@@ -18,6 +18,28 @@ subjects:
     name: {{ template "karpenter.serviceAccountName" . }}
     namespace: {{ .Release.Namespace }}
 ---
+# We need this role so that Karpenter can discover the Kubernetes APIServer Endpoints
```
jonathan-innis, line 21: Probably just apply the same wording changes from above in here if you make the change

jonathan-innis (commented):
Nice! A few docs nits

ellistarn:
Were you able to reproduce the flapping?

charts/karpenter/templates/role.yaml
```diff
@@ -46,6 +46,25 @@ rules:
...
+  labels:
+    {{- include "karpenter.labels" . | nindent 4 }}
+  {{- with .Values.additionalAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+rules:
+  # Read
+  - apiGroups: ["discovery.k8s.io"]
+    resources: ["endpointslices"]
```
ellistarn, line 65: This should be scoped.

charts/karpenter/templates/role.yaml
ellistarn: Why make a separate role?

njtran (commented):
/karpenter snapshot

charts/karpenter/templates/role.yaml
njtran: To establish a watch on a resource, it has to be on a collection, so we can't select by name here IIUC. At the moment it's scoped by namespace, and at least in my cluster, I only see kubernetes endpoints here.

jonathan-innis:
Closing this for now since we aren't currently sure the best direction for ensuring that we get the correct Kubernetes control-plane version every time.

PR #4625, opened 2023-09-14
feat: Add new configuration for service CIDR to enable Windows bootstrapping

### LINKED ISSUE #4088: Automatically include ServiceCIDR as an argument to Windows nodes bootstrap

After Windows support is implemented, it will be ideal and preferable to automatically include `ServiceCIDR` as an argument to bootstrap file for Windows nodes. It is possible we can retrieve this value along with the `DNSClusterIP` as is part of the DescribeCluster request operation against the EKS API. The value will be set in `serviceIpv4Cidr` in the `kubernetesNetworkConfig`. 

If the ServiceCIDR is not automatically retrieved or provided, the default value will be used as described in EKS docs [Bootstrap script configuration parameters](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html#bootstrap-script-configuration-parameters) and will default to 172.20.0.0/16 or 10.100.0.0/16 based on the IP address of the primary interface.

In a case where a custom service CIDR was used, pods behind a Kubernetes service not have network connectivity as it would have been misconfigured with the default values. If this issue is not fixed, it would become a blocker for anyone wishing to adopt Karpenter as an autoscaler for Windows nodes in a case where they need a custom service CIDR.

The only workaround is to use the default IP address range for the service CIDR.

### Additional Context

 At the current time of opening the issue, support for Windows is still yet to be released but has been merged in PR https://github.com/aws/karpenter/pull/3698.

### Attachments

_No response_

### Community Note

tzifudzi:
Fixes #4088

**Description**
TODO

TODO

### DIFF
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -172,6 +172,8 @@ settings:
     # -- Cluster endpoint. If not set, will be discovered during startup (EKS only)
     clusterEndpoint: ""
     # -- The default instance profile to use when launching nodes
+    clusterServiceIpv4Cidr: ""
+    # -- The CIDR block to assign Kubernetes service IP addresses from for ipv4
     defaultInstanceProfile: ""
     # -- If true then instances that support pod ENI will report a vpc.amazonaws.com/pod-eni resource
     enablePodENI: false
--- a/pkg/apis/settings/settings.go
+++ b/pkg/apis/settings/settings.go
@@ -34,6 +34,7 @@ var defaultSettings = &Settings{
 	ClusterCABundle:            "",
 	ClusterName:                "",
 	ClusterEndpoint:            "",
+	ClusterServiceIpv4Cidr:     "",
 	DefaultInstanceProfile:     "",
 	EnablePodENI:               false,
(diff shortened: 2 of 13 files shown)


charts/karpenter/values.yaml
```diff
@@ -172,6 +172,8 @@ settings:
     # -- Cluster endpoint. If not set, will be discovered during startup (EKS only)
     clusterEndpoint: ""
     # -- The default instance profile to use when launching nodes
+    clusterServiceIpv4Cidr: ""
```
jonathan-innis, line 175: Is it possible to discover this value from the cluster rather than specifying it within the `karpenter-global-settings`?

tzifudzi: Yes its possible. Currently it is discovering the value from the EKS cluster if the value is not specified. Perhaps we can skip specifying it in `karpenter-global-settings` altogether?

jonathan-innis:
Discussed offline: Looking into some options for users to specify this from environment variables in the `userData` or to get this info directly from the APIServer. We want to avoid users having to specify this at a global level in the `karpenter-global-settings`

PR #4648, opened 2023-09-19
ci: Increase`WARM_ENI_TARGET` to avoid timeouts for binding pods on E2E scale tests

engedaam:
Fixes #N/A 

**Description**
- Increase the the ENI allocated to k8s VPC CNI to prevent long wait on binding pods to nodes on the E2E scale tests 

- Forked repo 

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.github/actions/e2e/create-cluster/action.yaml
+++ b/.github/actions/e2e/create-cluster/action.yaml
@@ -114,8 +114,12 @@ runs:
             roleOnly: true
         withOIDC: true
       addons:
+      # For our scale testing, warm ENIs to help prevent timeouts on binding pods 
+      # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html 
       - name: vpc-cni
         permissionsBoundary: "arn:aws:iam::${{ inputs.account_id }}:policy/GithubActionsPermissionsBoundary"
+        configurationValues: |-
+          {"env":{"WARM_ENI_TARGET":"30"}}
       - name: coredns
         permissionsBoundary: "arn:aws:iam::${{ inputs.account_id }}:policy/GithubActionsPermissionsBoundary"
       - name: kube-proxy


.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -115,7 +115,10 @@ runs:
         withOIDC: true
       addons:
       - name: vpc-cni
+        version: 1.15.0
```
jonathan-innis, line 118: We're going to have to keep up with this and update it, which doesn't seem sustainable. Can we just use the default here?

.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -115,7 +115,10 @@ runs:
...
         permissionsBoundary: "arn:aws:iam::${{ inputs.account_id }}:policy/GithubActionsPermissionsBoundary"
+        configurationValues: |-
+          {"env":{"WARM_ENI_TARGET":"30"}}
```
jonathan-innis, line 121: Rationale for 30 here?

test/cloudformation/iam_cloudformation.yaml
```diff
@@ -157,6 +157,7 @@ Resources:
             Action:
               - eks:DeleteAddon
               - eks:DescribeAddon
+              - eks:DescribeAddonVersions
```
jonathan-innis, line 160: I'm not sure if we should be specifying the version here anyways

.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -114,8 +114,13 @@ runs:
             roleOnly: true
         withOIDC: true
       addons:
+      # For our scale testing, warm ENIs to help prevent timeouts on binding pods
```
jonathan-innis, line 117: AFAIK, there isn't an assignment of 1 pod:1 ENI. An ENI can support a certain number of Private IPs (this is specified on a per-instance type basis) and a new ENI is only launched once the number of IPs on a given ENI is exhausted.

PR #4652, opened 2023-09-19
feat: add basic priority support for aws on-demand capacity

### LINKED ISSUE #4136: Custom Pricing Overrides for RIs, Savings Plans, and other strategies
### Description

Karpenter should be able to understand custom pricing strategies by allowing users to manually override instance type/offering pricing for instance types that Karpenter chooses from. This would allow users to effectively modify the consolidation and provisioning behavior to prioritize certain instance types which they could mark as "cheaper" if they had discounts on certain instance types due to different Amazon EC2 billing strategies

guswynn:
Partially resolves #4136 #N/A 

**Description**

Our service performs significantly better on rarer instance types, and we would prefer to schedule it on those instances before falling back to cheaper ones. We thought about using provisioner weights to do this, but our understanding is that overlapping provisioners are problematic.

So, this pr implements _consciously constrained_  priority support for on-demand aws capacity. The `AwsNodeTemplate` is enriched with a priority map (not the `Provisioner`, as this is ec2-specific), which changes the AWS Fleet capacity allocation strategy to `"prioritized"` when it actually applies.

I was unsure exactly how the `instancetype` module works, but my understanding is that `getOverrides` ALWAYS produces overrides based on what instance types are available in the relevant az's. Because of this, I just added the `priority` field to these overrides, if applicable. If its possible for Karpenter to produce empty overrides (i.e. `unwrappedOfferings` is empty), then let me know, and I can adjust the code to ensure priority overrides are produced in these cases. (like i said, from reading the code, i don't *think* this is the case).

Also, the commit has some adjustments to indentation to some generated crd yaml's. I think this is just a newer version of `kubebuilder` changing things (i am running on a apple silicon mac)?

Tested in a staging environment in a real eks cluster, as well as an added unit test.

- [x] Yes, PR includes docs updates, but their may need to be additional color added to the 0.31 faq
- [ ] Yes, issue opened: # 
- [] No


guswynn:
cc @jonathan-innis I had discussed changing crds with you here: https://kubernetes.slack.com/archives/C04JW2J5J5P/p1694535628089169

jonathan-innis:
Also, for an a large behavioral change like this with API surface, we'd generally recommend an design doc/RFC to get the conversation going around what the use-cases this feature would support as well as the design recommendations and alternatives. 

I'm not necessarily convinced that this solution is the right direction that we should be going in or that we should be defining priorities like this in the AWSNodeTemplate but I'd be interested to hear the use-case more. Happy to discuss here or over in the K8s slack.

guswynn:
@jonathan-innis I went ahead and tested using provisioner weights instead and it worked great! Thanks for the help!

PR #4667, opened 2023-09-21
fix: configure serviceMonitor condition for monitoring API

cesarokuti:
Fixes
**Description**
Some installations we configure Karpenter before prometheus-operator and our argo-cd Waves broke and cannot continue the installation we implement the same structure as argocd [helm-chart](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-application-controller/servicemonitor.yaml#L1C1-L1C70) to check if the monitoring API is available.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter/templates/servicemonitor.yaml
+++ b/charts/karpenter/templates/servicemonitor.yaml
@@ -1,3 +1,4 @@
+{{- if .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" }}
 {{- if.Values.serviceMonitor.enabled -}}
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
@@ -26,4 +27,5 @@ spec:
     {{- with .Values.serviceMonitor.endpointConfig }}
       {{- toYaml . | nindent 6 }}
     {{- end }}
+{{- end -}}
 {{- end -}}


jonathan-innis:
> configure Karpenter before prometheus-operator

When you say before, I'm assuming that you mean the order that each set of manifests got deployed in. I believe that Argo has a dependency feature. Could you consider using that here to order the fact that the prometheus operator should be installed before the Karpenter installation is enabled? Or at least that the CRDs go on before the Karpenter installation?

Separately, if you want to disable the ServiceMonitor entirely, you could just consider setting that enabled value to `false`

cesarokuti:
> > configure Karpenter before prometheus-operator
> 
> When you say before, I'm assuming that you mean the order that each set of manifests got deployed in. I believe that Argo has a dependency feature. Could you consider using that here to order the fact that the prometheus operator should be installed before the Karpenter installation is enabled? Or at least that the CRDs go on before the Karpenter installation?
> 
> Separately, if you want to disable the Service monitor entirely, you could just consider setting that enabled value to `false`

Our automation using eks-blueprints addons and argocd and gitOps it's not a flexible way to disable when we install and re-enable it later. We have 2 solutions for now, the first one is configure this exception and the another one is create a new service monitor separately and managed out of community helm chart.

jonathan-innis:
> it's not a flexible way to disable when we install and re-enable it later

I'm not getting this. Isn't the `false` flag exactly what you want? It drops certain resources from the helm chart so that you can manage and control the ordering of the installation of them separately?

jonathan-innis:
Closing this for now since it's not clear to me how to proceed. Feel free to re-open this if you feel like this was closed by mistake.

PR #4686, opened 2023-09-25
docs: Add Carbon Efficient design document

JacobValdemar:
Fixes #N/A 

**Description**
*I would really appreciate your support, feedback, and collaboration on this design document. Your involvement would be immensely valuable.*

*As described in RFC aws/karpenter-core#675 and discussed on Slack, I am documenting a proposed design. This is a living document, and I expect your comments to necessitate changes to the current design before it is merged. I am still a new contributor to Karpenter, so while I have the best intentions, I lack complete technical insight; therefore, do not hesitate to correct me.*

*Together, we can make a significant impact on the sustainability of Kubernetes clusters!*

- [ ] Yes, issue opened: # 
- [x] No


designs/carbon-aware.md
```diff
@@ -0,0 +1,133 @@
...
+- **Embodied carbon emissions**: Manufacturing emissions (CO2e) amortized over instance lifetime (usually 4 years) divided by how long we use the instance
+- **Operational carbon emissions**: Carbon emitted by electricity grid to produce electricity for the instance in the region where it is used, multiplied by PUE
+
+There is a lot more to Green Software. If you want to learn more, I recommend you to visit [Green Software Practitioner](https://learn.greensoftware.foundation/) (a Green Software Foundation project - an affiliate of the Linux Foundation).
+
+## Solution
+
+### Feature Gate
+
+The feature is proposed to be controlled using a [feature gate](https://github.com/aws/karpenter/blob/main/website/content/en/docs/concepts/settings.md#feature-gates).
```
sftim, line 26:
```suggestion
The feature is proposed to be controlled using a [feature gate](https://karpenter.sh/docs/concepts/settings/#feature-gates).
```

designs/carbon-aware.md
```diff
@@ -0,0 +1,133 @@
...
+I'm currently working on my master's thesis in Computer Engineering (M.Sc.Eng) at Aarhus University located in Denmark. The objective of the thesis is to enable Karpenter to minimize carbon emissions from Kubernetes clusters that run on cloud infrastructure (focus is AWS).
+
+RFC: https://github.com/aws/karpenter/issues/4630
+
+## Fundamentals of Green Software
+I'll try to keep it simple. The reader should be familiar with the following.
+
+A cluster's emissions is made of two elements: embodied emissions and operational emissions. To get the total emissions, one can add them togeather.
+
+- **Embodied carbon emissions**: Manufacturing emissions (CO2e) amortized over instance lifetime (usually 4 years) divided by how long we use the instance
```
sftim, line 17: Do any cloud providers publish this data? Without it, the optimisation is hard to do.

designs/carbon-aware.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Solution
+
+### Feature Gate
+
+The feature is proposed to be controlled using a [feature gate](https://github.com/aws/karpenter/blob/main/website/content/en/docs/concepts/settings.md#feature-gates).
+
+| **Feature**  | **Default** |         **Config Key**          | **Stage** |    **Since**    | **Until** |
+| :----------: | :---------: | :-----------------------------: | :-------: | :-------------: | :-------: |
+| Carbon Aware |    false    | featureGates.carbonAwareEnabled |   Alpha   | v0.31.0/v0.32.0 |           |
```
sftim, line 30: Not sure, but I'd expect:
```suggestiion
| CarbonAware |    false    | featureGates.carbonAwareEnabled |   Alpha   | v0.31.0/v0.32.0 |           |
```

designs/carbon-aware.md
```diff
@@ -0,0 +1,133 @@
...
+
+I assume that we only want to use "static" data so we don't have to go out and make requests over the Internet to get real time data, since that would be expensive for performance.
+
+### Changes to consolidation (karpenter-core)
+
+Single Machine Consolidation (`singlemachineconsolidation.go`) and Multi Machine Consolidation (`multimachineconsolidation.go`) as well as `consolidation.go` is currently consolidating nodes to reduce costs. We want to change this when Carbon Aware is enabled. They should consolidate to minimize carbon emissions. 
+
+I have identified two potential solutions.
+
+Recommendation: solution 1.
```
sftim, line 47: How I'd approach this:
- set a price per tonne (or kg) of CO₂e
  - maybe have a config option for this, defaulting to $0. A good starting value for taking environmental impact into account is $0.25 / kg
- adjust the price algorithm to account for the pollution cost supplement

That could be consolidation option 3.

designs/carbon-aware.md
```diff
@@ -0,0 +1,133 @@
...
+
+Currently, provisioning (roughly) filter instances based on requirements, sort instances by price, and launch the cheapest instance. We want to change this when Carbon Aware is enabled. It should sort instances by carbon emissions and launch the instance which has the lowest Global Warming Potential[^1].
+
+#### Solution 1
+
+In karpenter-core, create a new method `types.go/OrderByCarbonEmissions` and use that in `nodeclaimtemplate.go/ToMachine` and `nodeclaimtemplate.go/ToNodeClaim` if Carbon Aware is enabled.
+
+In karpenter, create a new method `CarbonAwareCreate` in `pkg/providers/instance/instance.go` that is used in `pkg/cloudprovider/cloudprovider.go/Create` when Carbon Aware is enabled.
+
+[^1]: The potential impact of greenhouse gases on global warming. Measured in terms of CO2e.
```
sftim, line 133: (nit)
```suggestion
[^1]: The potential impact of greenhouse gases on global warming. Measured in terms of CO₂e.
```

designs/carbon-aware.md
sftim, line 17:
```suggestion
- **Embodied carbon emissions**: Manufacturing emissions (CO₂e) amortized over instance lifetime (usually 4 years) divided by how long we use the instance
```

designs/carbon-aware.md
sftim: Also, the same pricing choices should influence instance launch for reasons other than consolidation.

designs/carbon-aware.md
```diff
@@ -0,0 +1,133 @@
...
+- Less code copy-paste
+- Improvements to original consolidation methods also improve the Carbon Aware feature
+
+Disadvanteges:
+- Has a risk of breaking undocumented invariants
+- Adds complexity to the original consolidation methods
+
+### Changes to Provisioning
+
+Currently, provisioning (roughly) filter instances based on requirements, sort instances by price, and launch the cheapest instance. We want to change this when Carbon Aware is enabled. It should sort instances by carbon emissions and launch the instance which has the lowest Global Warming Potential[^1].
```
sftim, line 125: See https://github.com/aws/karpenter/pull/4686/files#r1335705803

designs/carbon-aware.md
JacobValdemar: Unfortunately, cloud providers generally do not publish high-quality carbon emission data. However, as noted in the _Carbon emissions data source_ section, there are still valid methods to obtain carbon emission data. An example is Boaviztapi which you can interact with in the [data demo](https://dataviz.boavizta.org/cloudimpact) and read about their methodology in [this article](https://www.boavizta.org/en/blog/empreinte-de-la-fabrication-d-un-serveur).

designs/carbon-aware.md
JacobValdemar: @sftim That is actually a very good idea!! Have not thought of that before. That way one could "simulate" a CO₂e tax/price on the machines.

JacobValdemar: @sftim I have added your idea as option 3. Interesting approach, I like it

designs/carbon-aware.md
```diff
@@ -0,0 +1,147 @@
...
+- [Cloud Carbon Footprint](https://github.com/cloud-carbon-footprint/ccf-coefficients). Extract data and create calculations.
+- [Boavizta](https://github.com/Boavizta/boaviztapi). Extract data and port calculations (for AWS). [Data demo](https://datavizta.boavizta.org/cloudimpact).
+
+I assume that we only want to use "static" data so we don't have to go out and make requests over the Internet to get real time data, since that would be expensive for performance.
+
+### Changes to consolidation (karpenter-core)
+
+Single Machine Consolidation (`singlemachineconsolidation.go`) and Multi Machine Consolidation (`multimachineconsolidation.go`) as well as `consolidation.go` is currently consolidating nodes to reduce costs. We want to change this when Carbon Aware is enabled. They should consolidate to minimize carbon emissions. 
+
+I have identified two potential solutions.
```
sftim, line 45:
```suggestion
I have identified three potential solutions.
```

JacobValdemar: Good catch 😁

designs/carbon-aware.md
```diff
@@ -0,0 +1,158 @@
...
+
+I assume that we only want to use "static" data so we don't have to go out and make requests over the Internet to get real time data, since that would be expensive for performance.
+
+#### Limitations
+For both options above, there is a discrepancy between the available instances known to Karpenter and instances know to the carbon emissions data source. This means that as it is right now, it is not possible to get carbon emissions data for all instances types. This is mostly the case for new instance types such as m7g. Unfurtunately this seems to extend to around 300 out of 700 instance types. See full comparison in [this Gist](https://gist.github.com/JacobValdemar/e1342013c0f5c980126f6a1feb66b4a1).
+
+I will attempt to eleminate this discrepancy, but it might not be possible. It will probably not always be possible to have an updated list of estimated carbon emissions for all instances as AWS continue to release new instance types. We should consider what to do with instance types that we do not have carbon emission estimates for.
+
+Approaches to handle this:
```
ellistarn, line 46: One option would be to model emissions as a price factor. e.g.,
1. If we don't know, multiply by 1
2. If known, multiply price by something > 1

You might want need to tune it such that 1 == your worst relative emissions machine.

designs/carbon-aware.md
```diff
@@ -0,0 +1,158 @@
...
+A cluster's emissions is made of two elements: embodied emissions and operational emissions. To get the total emissions, one can add them togeather.
+
+- **Embodied carbon emissions**: Manufacturing emissions (CO₂e) amortized over instance lifetime (usually 4 years) divided by how long we use the instance
+- **Operational carbon emissions**: Carbon emitted by electricity grid to produce electricity for the instance in the region where it is used, multiplied by PUE
+
+There is a lot more to Green Software. If you want to learn more, I recommend you to visit [Green Software Practitioner](https://learn.greensoftware.foundation/) (a Green Software Foundation project - an affiliate of the Linux Foundation).
+
+## Solution
+
+### Feature Gate
```
ellistarn, line 24: Worth exploring an approach where you can configure instance prices separately from Karpenter. This would allow karpenter to be unaware of "carbon aware mode", and you simply provide it the updated prices. One option for prices would be a configmap.

```
priceOverrides:
  - name: m5.large
     zone: us-west-2a
     capacity-type: spot
     price: .001
  - name: m5.xlarge
     zone: us-west-2a
     capacity-type: spot
     price: .002
```

designs/carbon-aware.md
```diff
@@ -0,0 +1,158 @@
...
+- 100% weighting to the inferred pollution cost
+- 0% weighting to the billed cost from AWS
+
+Also applies to provisioning.
+
+### Changes to Provisioning
+
+Currently, provisioning (roughly) filter instances based on requirements, sort instances by price, and launch the cheapest instance. We want to change this when Carbon Aware is enabled. It should sort instances by carbon emissions and launch the instance which has the lowest Global Warming Potential[^1].
+
+#### Solution 1
```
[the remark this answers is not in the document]
JacobValdemar: I agree, that would be a good approach to succeed with our own prioritization. Standard Karpenter would still just use `lowest-price` right?

One thing I have a hard time understanding is how `pkg/cloudprovider/types.go/OrderByPrice` in `karpenter-core` is related to the ordering performed in `karpenter`. Would it even be necessary to do something with `OrderByPrice` within `karpenter-core`?

designs/carbon-aware.md
JacobValdemar: Interesting idea - that would generally give users a lot of flexibility. 

However, if it is an override, it would not be possible to combine carbon emission price with actual price as [suggested by sftim](https://github.com/aws/karpenter/pull/4686#discussion_r1335703216) (although that is not a № 1 priority for the feature). Also, it would make it hard get precise operational carbon emissions as the [carbon intensity](https://learn.greensoftware.foundation/carbon-awareness#carbon-intensity) of the energy grid depends on the geographic region of the data center. I'll add it as an option tomorrow (it's 22.30 PM here)

designs/carbon-aware.md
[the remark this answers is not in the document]
ellistarn: No -- standard Karpenter would just use Prioritized, and the code would sort by price (regardless of price overrides)

designs/carbon-aware.md
JacobValdemar: I have added it as option 4

designs/carbon-aware.md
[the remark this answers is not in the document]
JacobValdemar: @ellistarn I am unsure how spot instances would fit into all and any of the options. What do you think?

JacobValdemar:
@ellistarn and @sftim I have made some changes based on your feedback. You may re-review the design document if you want to :)

designs/carbon-aware.md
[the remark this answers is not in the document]
ellistarn: Generational challenging. I'd suggest removing it from the scope of your MVP feature.

designs/carbon-aware.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Option 4: Enable custom instance price overrides
+Enable administrators to configure custom instance price overrides, e.g. in a ConfigMap. A configuration using emission factors (varying with region and instance type) masked as prices can be pre-generated. Administrators then copy-paste a Carbon Aware `priceOverride` into their environment.
+
+```yaml
+priceOverrides:
+  - instanceType: "m5.large"
+    region: "eu-west-1"
+    capacityType: OnDemand
+    price: 0.007712
```
jonathan-innis, line 156: I'd want to lean toward calling this `cost` now. We are starting to talk more about a "cost-optimization" function vs. a "price-optimization" function, especially if we are going to reason in terms of things that definitely aren't price

JacobValdemar: I agree. The term `price` does not concisely express what the variable has/can/will become.

designs/carbon-aware.md
```diff
@@ -0,0 +1,197 @@
...
+I am currently working on my master's thesis in Computer Engineering (Master of Science in Engineering) at Aarhus University located in Denmark. The objective of the thesis is to enable Karpenter to minimize carbon emissions from Kubernetes clusters that run on cloud infrastructure (scoped to AWS).
+
+RFC: https://github.com/aws/karpenter/issues/4630
+
+## Fundamentals of Green Software
+I will try to keep it simple. The reader should be familiar with the following.
+
+A cluster's emissions is made of two elements: embodied emissions and operational emissions. To get the total emissions, one can add them togeather.
+
+- **Embodied carbon emissions**: Manufacturing emissions (CO₂e) amortized over instance lifetime (usually 4 years) divided by how long we use the instance
```
jackfrancis, line 16: Can we clarify that "instance" in this sense refers to a percentage of a physical machine?

designs/carbon-aware.md
```diff
@@ -0,0 +1,197 @@
...
+| :---------: | :---------: | :-----------------------------: | :-------: | :-------------: | :-------: |
+| CarbonAware |    false    | featureGates.carbonAwareEnabled |   Alpha   | v0.32.0/v0.33.0 |           |
+
+### Carbon emissions data source
+Currently the best option is to create estimates based on the methodology used in [Boaviztapi](https://github.com/Boavizta/boaviztapi).
+
+[Try out Boaviztapi on the Datavizta demo website](https://datavizta.boavizta.org/cloudimpact).
+
+#### Licensing
+Boaviztapi is licensed under [`GNU Affero General Public License v3.0`](https://github.com/Boavizta/boaviztapi/blob/main/LICENSE). Therefore, as far as I know, we must license their data under the same license if used in the Karpenter repository.
```
jackfrancis, line 36: Will there be conflicts in licensing their data according to these requirements and the CNCF guidance for licensing the karpenter-core source itself?

designs/carbon-aware.md
```diff
@@ -0,0 +1,197 @@
...
+
+#### Limitations
+There is a discrepancy between the available instances known to Karpenter and instances know to Boaviztapi. This means that as it is right now, it is not possible to get carbon emissions data for all instances types. This is mostly the case for new instance types such as m7g. Around 290 out of 700 instance types is missing data. See full comparison in [this Gist](https://gist.github.com/JacobValdemar/e1342013c0f5c980126f6a1feb66b4a1).
+
+I will attempt to eleminate this discrepancy, but it might not be possible. It will probably not always be possible to have an updated list of estimated carbon emissions for all instances as AWS continue to release new instance types. We should consider what to do with instance types that we do not have carbon emission estimates for.
+
+Approaches to handle this:
```
jackfrancis, line 43: Why not exclude instances that are not emission-priced when the `carbonAwareEnabled` feature gate is enabled? As a user if I were strict about this accounting I would not want the risk of polluting my reckoning with non-determinate inputs.

jackfrancis:
I don't want to get philosophical, but should we consider the following?

Adding to karpenter the ability to further optimize for carbon-friendly outcomes may have the side-effect of preventing less efficient systems (I'm imagining legacy operational platforms that don't have auto-scaling and consolidation features) from the availability of the most carbon-friendly instance types (already reserved by karpenter-powered systems).

Another consideration is that karpenter-powered systems would be demonstrably more economical to a customer's infra spending, and that particular carbon-friendly instance types may emerge in the cloud provider's data as being especially unprofitable; conversely non-karpenter-powered systems might be more likely to reserve carbon-inefficient instances, making those instance types more attractice to a cloud provider, and thus more likely for continued expansion across the cloud as a whole.

Do we want to include considerations like the above in our design thinking, or is that more abstract (I'm sure it'll go into your thesis! :))

designs/carbon-aware.md
```diff
@@ -0,0 +1,197 @@
...
+#### Considerations
+1. 👍 Less code copy-paste.
+1. 👍 Improvements to original consolidation methods also improve the Carbon Aware feature.
+1. 👎 Has a risk of breaking undocumented invariants.
+1. 👎 Adds complexity to the original consolidation methods.
+
+### Option 3: Override instance price with carbon price (recommended)
+Minimize carbon emissions by defining a price per kgCO₂e and override the instance price with the carbon price (USD/kgCO₂e). Using the `prioritized` launch strategy, carbon emissions will be minimized during provisioning. Consolidation will unknowingly consolidate to minimize carbon emissions.
+
+The carbon price will depend on on `region` and `instanceType` and assume constant resource utilization (e.g. always 80% utilization). The carbon price will be generated in a "hack" and included as consts (same method as used for generating initial pricing[^2]). The carbon price / emission estimates can be updated with new versions.
```
jackfrancis, line 138: Can you clarify how "assume constant resource utilization" concept affects carbon price?

JacobValdemar: This is one place I am constrained in my knowledge about Karpenter, so please correct me if this is incorrect: as far as I know, Karpenter can not provide information about current resource utilization (CPU/RAM) nor about expected utilization for nodes considered to be created from provisioning or consolidation.

To calculate the operational Global Warming Potential (kgCO2e) we must know how much power (W) that an instance ("server") consume. The power consumption is not static, but depends on how large the workload (%) is (e.g. see image). Power consumption is not a linear function of the workload. Therefore, CPU Power Consumption profiles (based on empirical data, aka. server testing) are used to estimate how much power a CPU consume depending on the workload. When calculating the operational carbon emissions we therefore should know what the current/expected workload is to be able to accurately estimate the operational footprint. However, if we do not know the workload (%) we must make an assumption about how large it is to complete the calculation.

I can elaborate further on this in the call today (Wednesday) if you would like, because the calculation is a bit complex to explain in a comment ;) You can read more about the calculation in the [Boavizta API (v.dev) docs here](https://dev.doc.api.boavizta.org/Explanations/components/cpu/#consumption-profile). 

See an illustration of the concept in this picture (AGPL-3.0, Boavizta API): 
billede

designs/carbon-aware.md
JacobValdemar: @jackfrancis So I am in no way a lawyer, but I guess that there wont be licensing issues. We probably just have to put a different license notice in the top of files that is based on their works. Currently, I would expect the data with that license only to be in cloud provider repos (e.g. https://github.com/aws/karpenter) and not karpenter-core.

designs/carbon-aware.md
JacobValdemar: @jackfrancis Good point. That is totally not obvious when reading the text. When we say instance lifetime we of course mean the lifetime of the physical machine that the instance is part of 👍

JacobValdemar: a7134dd

designs/carbon-aware.md
JacobValdemar: @jackfrancis I agree. I also think an exclusion is the best option in case we can't estimate the emissions accurately. The method I thought about excluding them was to assigning them an absurd high price so they will never be picked voluntarily. However, I want to improve the dataset that we depend on, so that we can use as many instance types as possible. See this issue for reference https://github.com/Boavizta/boaviztapi/issues/232

JacobValdemar:
>I don't want to get philosophical, but should we consider the following?
>
>Adding to karpenter the ability to further optimize for carbon-friendly outcomes may have the side-effect of preventing less efficient systems (I'm imagining legacy operational platforms that don't have auto-scaling and consolidation features) from the availability of the most carbon-friendly instance types (already reserved by karpenter-powered systems).
>
>Another consideration is that karpenter-powered systems would be demonstrably more economical to a customer's infra spending, and that particular carbon-friendly instance types may emerge in the cloud provider's data as being especially unprofitable; conversely non-karpenter-powered systems might be more likely to reserve carbon-inefficient instances, making those instance types more attractice to a cloud provider, and thus more likely for continued expansion across the cloud as a whole.
>
>Do we want to include considerations like the above in our design thinking, or is that more abstract (I'm sure it'll go into your thesis! :))

@jackfrancis So if I understand you correctly, you are saying that

1. Making Karpenter more carbon-efficient may increase demand for carbon-efficient instance types. This may prevent carbon-inefficient deployments from migrating to carbon-efficient instance types.
1. If carbon-efficient instance types are less profitable than carbon-inefficient instance types cloud providers will not invest in carbon-efficient instance types (you know better than me if that is true)

I am not sure I follow you completely. Hopefully you can elaborate on it. I think a quote from a paper I read earlier explains very well why optimizing for carbon emissions is important[^1]:

> Optimizing embodied carbon is important because it motivates operators to choose suppliers with a lower carbon footprint, as well as increase their own infrastructure’s efficiency and lifetime to better amortize its embodied carbon. Such optimizations can influence the entire supply chain by indirectly incentivizing every downstream supplier to also reduce their carbon footprint – both operational and embodied – as well as increase their product lifetime. By comparison, while optimizing operational carbon is more direct, as it reduces a single operator’s emissions, it does not affect incentives up and down supply chain.

[^1]: N. Bashir, D. Irwin, and P. Shenoy, “On the Promise and Pitfalls of Optimizing Embodied Carbon,” in Proceedings of the 2nd Workshop on Sustainable Computer Systems, in HotCarbon ’23. New York, NY, USA: Association for Computing Machinery, Aug. 2023, pp. 1–6. doi: [10.1145/3604930.3605710](https://doi.org/10.1145/3604930.3605710).

designs/carbon-aware.md
ellistarn: > So I am in no way a lawyer, but I guess that there wont be licensing issues

Famous last words.

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+There is a lot more to Green Software. If you want to learn more, I recommend you to visit [Green Software Practitioner](https://learn.greensoftware.foundation/) (a Green Software Foundation project - an affiliate of the Linux Foundation).
+
+## Solution
+
+### Feature Gate
+The feature is proposed to be controlled using a [feature gate](https://karpenter.sh/docs/concepts/settings/#feature-gates).
+
+| **Feature** | **Default** |         **Config Key**          | **Stage** |    **Since**    | **Until** |
+| :---------: | :---------: | :-----------------------------: | :-------: | :-------------: | :-------: |
+| CarbonAware |    false    | featureGates.carbonAwareEnabled |   Alpha   | v0.32.0/v0.33.0 |           |
```
jonathan-innis, line 30: Small detail: we're planning to drop the config key so this feature flag would probably have the same format a-la Kubernetes upstream. If we take the last solution, this feature flag might look like `PricingOverridesEnabled=true`

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+
+RFC: https://github.com/aws/karpenter/issues/4630
+
+## Fundamentals of Green Software
+I will try to keep it simple. The reader should be familiar with the following.
+
+A cluster's emissions is made of two elements: embodied emissions and operational emissions. To get the total emissions, one can add them togeather.
+
+- **Embodied carbon emissions**: Manufacturing emissions (CO₂e) amortized over instance lifetime (usually 4 years) divided by how long we use the instance
+- **Operational carbon emissions**: Carbon emitted by electricity grid to produce electricity for the instance in the region where it is used, multiplied by PUE
```
njtran, line 17: Would this be different for regions that focus on more sustainable energy sources like hydroelectric or geothermal energy? Or are you forgoing the energy sources, and talking purely about the byproduct of the heat and other leftover elements in electricity grids?

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+2. Estimate zero emissions
+
+### Launch strategy
+To enable emission based priotization, the launch strategy should be changed from `lowest-price` to `prioritized`.
+
+### Changes to consolidation (karpenter-core)
+Single Machine Consolidation (`singlemachineconsolidation.go`) and Multi Machine Consolidation (`multimachineconsolidation.go`) as well as `consolidation.go` is currently consolidating nodes to reduce costs. We want to change this when Carbon Aware is enabled. They should consolidate to minimize carbon emissions. 
+
+### Changes to Provisioning
+Currently, provisioning (roughly) filter instances based on requirements, sort instances by price, and launch the cheapest instance. We want to change this when Carbon Aware is enabled. It should sort instances by carbon emissions and launch the instance which has the lowest Global Warming Potential[^1].
```
ellistarn, line 56: Might we want a balanced approach to this?

designs/carbon-aware.md
jonathan-innis: We'd definitely need to check with AWS legal for this (while AWS owns it) and consider how this interacts with the CNCF guidance since this would probably live in the karpenter-core repo if it was gen-ed. I could also see this living in its own separate repo that provided configuration plugins to the pricing overrides that we are thinking about here

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+
+#### Limitations
+There is a discrepancy between the available instances known to Karpenter and instances know to Boaviztapi. This means that as it is right now, it is not possible to get carbon emissions data for all instances types. This is mostly the case for new instance types such as m7g. Around 290 out of 700 instance types is missing data. See full comparison in [this Gist](https://gist.github.com/JacobValdemar/e1342013c0f5c980126f6a1feb66b4a1).
+
+I will attempt to eleminate this discrepancy, but it might not be possible. It will probably not always be possible to have an updated list of estimated carbon emissions for all instances as AWS continue to release new instance types. We should consider what to do with instance types that we do not have carbon emission estimates for.
+
+Approaches to handle this:
+1. Estimate extremely high emissions to effectively filter out unknown instance types (recommended)
+2. Estimate zero emissions
```
jonathan-innis, line 47: Zero is probably wrong. There may be a third option here which would be to do some estimation function. What on? I haven't thought through it that hard; however, I agree with @jackfrancis that the easy way out is to just exclude for now.

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Launch strategy
+To enable emission based priotization, the launch strategy should be changed from `lowest-price` to `prioritized`.
```
jonathan-innis, line 50: We probably want to do `capacity-optimized-prioritized` to properly handle spot

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+1. 👎 Adds complexity to the *price* concept. Price is not *just* price, but rather becomes an optimization function.
+1. 👎 Depending on implementation, the `karpenter_cloudprovider_instance_type_price_estimate` metric *may* represent more than just price when Carbon Aware is enabled.
+
+### Option 4: Enable custom instance price overrides
+Enable administrators to configure custom instance price overrides, e.g. in a ConfigMap. A configuration using emission factors (varying with region and instance type) masked as prices can be pre-generated. Administrators then copy-paste a Carbon Aware `priceOverride` into their environment.
+
+```yaml
+priceOverrides:
+  - instanceType: "m5.large"
+    region: "eu-west-1"
```
ellistarn, line 156: I think there's some misunderstanding of region/zone going on.

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+  - instanceType: "m5.xlarge"
+    region: "eu-west-1"
+    capacityType: OnDemand
+    price: 0.015424
+```
+
+<details>
+<summary>Alternative interface</summary>
+
+Alternatively, a more flexible interface could be:
```
ellistarn, line 168: Oddly, I think of this as more rigid.

jonathan-innis (commented):
To me it seems like option 3 is a natural extension of 4. I could see us going with 4 and (if we saw this being a native-enough concept that users wanted to have built-in in the future), it could turn into 3

designs/carbon-aware.md
JacobValdemar: I think it could be interesting, like the carbon tax discussed in option 3, but I think that I something we should consider it something we add later

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+      capacityType: OnDemand
+      price: 0.007712
+    - instanceType: "m5.xlarge"
+      region: "eu-west-1"
+      capacityType: OnDemand
+      price: 0.015424
+```
+</details>
+
+A ConfigMap with price overrides for all combinations of instance types and regions will be very huge. 632 instances * 29 regions = 18,328 pairs. Four lines per pair gives a file with 73,312 lines. The file/configmap will approximately have a size of 2 MB. That exceeds the [`1 MiB` limit on ConfigMap size in Kubernetes](https://kubernetes.io/docs/concepts/configuration/configmap/#motivation).
```
ellistarn, line 185: EC2 launches new instance types regularly, and we should expect this to grow.

designs/carbon-aware.md
jackfrancis: The particular license that Boaviztapi chooses to use suggests that it's using the "copyleft malware" approach (for Good, not Evil, of course) to ensure that its definitions of freedom (as in both beer and speech) are enforced upon downstream projects. So that's the only reason to bring this up. AWS and CNCF are probably going to combine w/ such a license in their own unique interesting ways. Non-zero chance there will be friction.

designs/carbon-aware.md
jackfrancis: My personal take is that if I were a customer I would prefer not to use something at all that isn't carbon-priced.

1. I would be biased to suspect that such a thing was more likely than not to be carbon-inefficient (i.e., a previous generation of tech)
2. I think that exclusion is a better forcing function for incentivizing such things to do the work of calculating cost, and/or marginalizing them from the marketplace (i.e., replacing them w/ newer, more carbon-efficient generations of tech)

sftim:
@JacobValdemar what are your plans for moving this PR forward?

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+### Launch strategy
+To enable emission based priotization, the launch strategy should be changed from `lowest-price` to `prioritized`.
+
+### Changes to consolidation (karpenter-core)
+Single Machine Consolidation (`singlemachineconsolidation.go`) and Multi Machine Consolidation (`multimachineconsolidation.go`) as well as `consolidation.go` is currently consolidating nodes to reduce costs. We want to change this when Carbon Aware is enabled. They should consolidate to minimize carbon emissions. 
+
+### Changes to Provisioning
+Currently, provisioning (roughly) filter instances based on requirements, sort instances by price, and launch the cheapest instance. We want to change this when Carbon Aware is enabled. It should sort instances by carbon emissions and launch the instance which has the lowest Global Warming Potential[^1].
+
+### Option 1: Use Carbon Aware provisioning and concolidation methods
```
sftim, line 58:
```suggestion
### Option 1: Use Carbon Aware provisioning and consolidation methods
```

JacobValdemar:
@sftim I am not really sure. In the meeting we talked about selecting option 4. However, I thought that that wouldn't make a good solution for the master's project. So since than I have been working on option 3 on a [fork](https://github.com/JacobValdemar/carbon-aware-karpenter/tree/carbon-aware-feature). It is currently working for eu-west-1, but I need to develop some more before it is finished. When it is done, I will consider requesting that it is merged into karpenter. If the result is something the maintainers like, they can merge it.
However, based on previous dialogue I don't expect it to be approved (argument being that it require too much maintenance considering that there currently is little demand for the feature), in which case the feature will only exist in the fork.

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+### Carbon emissions data source
+Currently the best option is to create estimates based on the methodology used in [Boaviztapi](https://github.com/Boavizta/boaviztapi).
+
+[Try out Boaviztapi on the Datavizta demo website](https://datavizta.boavizta.org/cloudimpact).
+
+#### Licensing
+Boaviztapi is licensed under [`GNU Affero General Public License v3.0`](https://github.com/Boavizta/boaviztapi/blob/main/LICENSE). Therefore, as far as I know, we must license their data under the same license if used in the Karpenter repository.
+
+#### Limitations
+There is a discrepancy between the available instances known to Karpenter and instances know to Boaviztapi. This means that as it is right now, it is not possible to get carbon emissions data for all instances types. This is mostly the case for new instance types such as m7g. Around 290 out of 700 instance types is missing data. See full comparison in [this Gist](https://gist.github.com/JacobValdemar/e1342013c0f5c980126f6a1feb66b4a1).
```
chrisnegus, line 41:
```suggestion
There is a discrepancy between the available instances known to Karpenter and instances know to Boaviztapi. This means that as it is right now, it is not possible to get carbon emissions data for all instances types. This is mostly the case for new instance types such as m7g. Around 290 out of 700 instance types are missing data. See full comparison in [this Gist](https://gist.github.com/JacobValdemar/e1342013c0f5c980126f6a1feb66b4a1).
```

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+
+I will attempt to eleminate this discrepancy, but it might not be possible. It will probably not always be possible to have an updated list of estimated carbon emissions for all instances as AWS continue to release new instance types. We should consider what to do with instance types that we do not have carbon emission estimates for.
```
chrisnegus, line 43:
```suggestion
I will attempt to eleminate this discrepancy, but it might not be possible. It will probably not always be possible to have an updated list of estimated carbon emissions for all instances as AWS continues to release new instance types. We should consider what to do with instance types that we do not have carbon emission estimates for.
```
Ignore this comment if this is for a British audience.

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+In `karpenter`, create a new method `CarbonAwareCreate` in `pkg/providers/instance/instance.go` that is used in `pkg/cloudprovider/cloudprovider.go/Create` instead of `pkg/providers/instance/instance.go/Create` when Carbon Aware is enabled.
+
+#### Considerations
+1. 👍 Current consolidation methods are unaffected.
+1. 👎 There might be copy-paste of code from the original consolidation methods to the carbon aware consolidators.
+
+### Option 2: Use Carbon Aware filtering/sorting methods
+
+#### Consolidation
+Create carbon aware implementations of low-level functions like `filterByPrice`, `filterOutSameType`, `getCandidatePrices`, etc. that is used when Carbon Aware is enabled. Usage of aforementioned functions might assume that it is price that they are getting, but in reality it is data about carbon emissions.
```
chrisnegus, line 126:
```suggestion
Create carbon aware implementations of low-level functions like `filterByPrice`, `filterOutSameType`, `getCandidatePrices`, etc. that are used when Carbon Aware is enabled. Usage of aforementioned functions might assume that it is price that they are getting, but in reality it is data about carbon emissions.
```

designs/carbon-aware.md
```diff
@@ -0,0 +1,199 @@
...
+#### Considerations
+1. 👍 Less code copy-paste.
+1. 👍 Improvements to original consolidation methods also improve the Carbon Aware feature.
+1. 👎 Has a risk of breaking undocumented invariants.
+1. 👎 Adds complexity to the original consolidation methods.
+
+### Option 3: Override instance price with carbon price (recommended)
+Minimize carbon emissions by defining a price per kgCO₂e and override the instance price with the carbon price (USD/kgCO₂e). Using the `prioritized` launch strategy, carbon emissions will be minimized during provisioning. Consolidation will unknowingly consolidate to minimize carbon emissions.
+
+The carbon price will depend on on `region` and `instanceType` and assume constant resource utilization (e.g. always 80% utilization). The carbon price will be generated in a "hack" and included as consts (same method as used for generating initial pricing[^2]). The carbon price / emission estimates can be updated with new versions.
```
chrisnegus, line 140:
```suggestion
The carbon price will depend on `region` and `instanceType` and assume constant resource utilization (e.g. always 80% utilization). The carbon price will be generated in a "hack" and included as consts (same method as used for generating initial pricing[^2]). The carbon price / emission estimates can be updated with new versions.
```

chrisnegus (commented):
A few small wording suggestions.

JacobValdemar:
### Status update

I am handing in my master's thesis within the next couple of days. Within 3 weeks I will publicly defend it at Aarhus University. After that I will share the master's thesis report with you. Therein you can read about the implementation of option 3 and an evaluation of the impact of using Carbon-Efficient Karpenter. Based on that we can discuss the next steps we want to take.

The work we are doing can significantly reduce the environmental footprint of cloud-based Kubernetes clusters. I hope you are all still motivated, because I am! :D

JacobValdemar:
Dear Karpenter community,

**I am happy to say that I can finally share my master's thesis with you.**

In this thesis, a carbon efficient cluster autoscaler for Kubernetes is developed by adapting Karpenter to minimize cloud instance carbon emission estimates from BoaviztAPI. Specifically, this is an implementation of _Option 3: Override instance price with carbon price_ described in this design document. The developed Carbon Efficient Karpenter shows a statistically significant median reduction in cloud instance carbon emissions when compared to the original Karpenter.

I recommend that you read Chapter 7 (Design and Implementation) and Chapter 8 (Experiments and Results).

Check it out! 🚀 

https://github.com/JacobValdemar/carbon-efficient-karpenter-thesis

JacobValdemar:
Hey all. What would you think about merging the feature described in my master's thesis into current Karpenter Provider AWS? It will enable users to opt in for carbon efficient cluster autoscaling with no impact on existing users, very limited impact on the Karpenter design and almost no maintenance cost for Karpenter developers. Considering the potential described in my master's thesis, what do you think about this idea?

designs/carbon-aware.md
```diff
@@ -0,0 +1,197 @@
...
+1. 👎 Adds complexity to the *price* concept. Price is not *just* price, but rather becomes an optimization function.
+1. 👎 Depending on implementation, the `karpenter_cloudprovider_instance_type_price_estimate` metric *may* represent more than just price when Carbon Efficient is enabled.
+
+### Option 4: Enable custom instance price overrides
+Enable administrators to configure custom instance price overrides, e.g. in a ConfigMap. A configuration using emission factors (varying with region and instance type) masked as prices can be pre-generated. Administrators then copy-paste a Carbon Efficient `priceOverride` into their environment.
+
+```yaml
+priceOverrides:
+  - instanceType: "m5.large"
+    region: "eu-west-1"
```
Bryce-Soghigian, line 154: Price overrides is interested in multiple respects. For example there are reserved instances customers will reserve on azure and get a better contract for. So its actually cheaper for karpenter ot use those instances they got at a discount than the typical "retail pricing api" pricing data for the instance price. 


So this field of pricing overrides can be used and extended to other usecases.

designs/carbon-aware.md
```diff
@@ -0,0 +1,197 @@
...
++			deprovisioners: []Deprovisioner{
++				NewExpiration(clk, kubeClient, cluster, provisioner, recorder),
++				NewDrift(kubeClient, cluster, provisioner, recorder),
++				NewEmptiness(clk),
++				NewEmptyMachineConsolidation(clk, cluster, kubeClient, provisioner, cp, recorder),
++				NewCarbonEfficientMultiMachineConsolidation(clk, cluster, kubeClient, provisioner, cp, recorder),
++				NewCarbonEfficientSingleMachineConsolidation(clk, cluster, kubeClient, provisioner, cp, recorder),
++			},
++		}
++	}
```
Bryce-Soghigian, line 89: Code gets outdated in a design doc, this bit already is outdated with the karpenter-core reference.

designs/carbon-aware.md
```diff
@@ -0,0 +1,197 @@
...
+In `karpenter`, create a new method `CarbonEfficientCreate` in `pkg/providers/instance/instance.go` that is used in `pkg/cloudprovider/cloudprovider.go/Create` instead of `pkg/providers/instance/instance.go/Create` when Carbon Efficient is enabled.
+
+#### Considerations
+1. 👍 Current consolidation methods are unaffected.
+1. 👎 There might be copy-paste of code from the original consolidation methods to the carbon efficient consolidators.
+
+### Option 2: Use Carbon Efficient filtering/sorting methods
+
+#### Consolidation
+Create carbon efficient implementations of low-level functions like `filterByPrice`, `filterOutSameType`, `getCandidatePrices`, etc. that are used when Carbon Efficient is enabled. Usage of aforementioned functions might assume that it is price that they are getting, but in reality it is data about carbon emissions.
```
Bryce-Soghigian, line 124: Can you go into further depth where the data consolidation will use or consider? How do cloud providers provide this carbon aware data to karpenter core?

designs/carbon-aware.md
JacobValdemar: Yes, I have heard in multiple occasions that price overrides could be useful for various reasons. 

In my [Master's Thesis](https://github.com/JacobValdemar/carbon-efficient-karpenter-thesis), overriding the price of all existing instance types is actually what I did. It is feasible and maintainable with enough automation. I think it makes sense in _this_ context, but might be infeasible in other use cases.

JacobValdemar:
Really no motivation for adding a carbon-efficient feature to Karpenter?

Issue #3860, opened 2023-05-05
Account for AWS Savings Plan when choosing to deploy on-demand or spot


lksvenoy-r7:
### Tell us about your request

We need Karpenter to take into account discounts applied on compute as a result of AWS Savings Plans, and other types of discounts. 

Having some mechanism to pass in configuration to inform Karpenter of discounts applicable to the specific environment it is running in, or have it somehow figure that out, would allow Karpenter to make better decisions with more accurate pricing information.

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

Currently, Karpenter does not respect discounts that may be applicable as a result of AWS Savings Plan (and other discounts). Starting in April, we've noticed a significant increase in spot prices for certain regions, with for example us-east-1 spot instances that we are utilizing costing on average 50% more than 2 months prior. 

This puts us in an awkward spot where running on on-demand may be significantly cheaper than running on spot pricing, due to discounts applied as a result of AWS Savings Plans. 

To illustrate, consider the instance type m5d.12xlarge in us-east-1:

On the 16th of March, the spot price for this instance type was **$1.06/hr.** Since then, the price has been steadily increasing, currently at **$1.56/hr**.

This same instance type using a Compute Savings Plan, with a term length of 3 years, partial upfront, puts the price at **$1.264/hr**. 

This is a ~21% difference in price.

Excluding other types of discounts, this means that we are now running into scenarios where running our workloads on on-demand is significantly cheaper than running on spot, not to mention more stable.

### Are you currently working around this issue?

Short-term, we are seriously considering moving our workloads to on-demand, as this would significantly reduce the cost of our infrastructure. Long-term, we would like to see Karpenter be able to make better decisions around price performance.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

runningman84:
The workaround assumes on demand pricing instead of having saving plan or private pricing.

iandelahorne:
We're interested in this as AWS promotes savings plans heavily and we've noticed that spot pricing for SKUs we use have trended upward a lot recently. 

[...]

PR #4697, opened 2023-09-26
feat: Adding capacityType discount as a percentage

rkirk-r7:
Fixes #N/A 

**Description**
Reimplementation of this PR: https://github.com/aws/karpenter/pull/3896

Added code to allow a custom multiplier to be applied to on-demand and spot pricing.
This allows you to add a multiplier to prices that will be taken into consideration on which instance to use.
This take's the same approach as [autospotting](https://github.com/LeanerCloud/AutoSpotting/blob/master/START.md#configuration-of-autospotting).

For Enterprise customers, instance prices can come from the Payer account which is not accessible outside of this account - this leads to a problem as we cannot get the accurate price data in AWS accounts running the Kubernetes clusters where it may be that running on-demand instance types would be cheaper than spot instance types.

When Karpenter retrieves the up-to-date pricing from AWS, the multiplier will be applied to the returned data.
The instance `getCapacityType` function has been updated so that it will be price aware - only if
 `offering.CapacityType == corev1beta1.CapacityTypeSpot && offering.Price == cheapestOffering` will the capacityType be set to Spot

- [x] Tests written around the settings change
- [x] Tests written around the multiplier being applied
- [ ] Tested in a Kubernetes cluster:
    - [x] Verify multiplier set to default value does not modify prices and karpenter launched expected capacity
    - [x] Verify when multiplier set prices were correctly modified and karpenter launched expected capacity
    - [x] Verify when multiplier set prices were correctly modified and karpenter launched on-demand capacity type when cheaper than spot pricing
    - [ ] Verify spot instances are consolidated to cheaper on-demand
    - [x] Verify on-demand instances are consolidated to cheaper spot

* [x] Yes, PR includes docs updates 
* [ ] Yes, issue opened: # 
* [ ] No

### DIFF
--- a/designs/discount-pricing.md
+++ b/designs/discount-pricing.md
@@ -1,0 +1,55 @@
+# Discounted Pricing Support
+
+## Overview
+
+Karpenter is currently unaware of any discounted pricing, such as volume discounts or reserved instances/savings plans, which can lead to more expensive instances being chosen. For example an on demand instance with a savings plan discount may cost less than a spot instance of the same type. This pricing is apparently only available from the "payer" account, not any other child account API's for pricing.
+
+This was made explicit recently - the price of spot instances rose sharply at the beginning of Q2 2023. For users on default pricing this may not be noticeable, however if there is any discount for on-demand instances in an account, it could begin to become cheaper to use on-demand instances instead of spot.
+
+## User Stories
+
+* Karpenter will automatically prioritise the cheapest node capacity type in an account based on personal modifications to EC2 pricing
+* Karpenter will allow me to configure discounted pricing at the account level
+
+## Background
+
+[Conversation on Slack](https://kubernetes.slack.com/archives/C02SFFZSA2K/p1684246928553159)
+
+## How Will Karpenter Handle Discounted Pricing
+
+A multiplier will be applied to the price to allow any discounts to be applied and determine the real cost of an EC2 instance.
+For example, a multiplier of 0.9 would apply a 10% discount
+
+Separate multiplier values for Spot and On Demand pricing will be allowed to allow for accounts which have different pricing discounts for each type.
+
+The multiplier will default to a value of 1 so no discount will be applied unless explicitly enabled. 
+
+### Spot pricing
+
+The Spot price will be multiplied with the SpotPriceMultiplier to determine the real cost for Spot instances.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: karpenter-global-settings
+  namespace: karpenter
+data:
+  # Spot Price Multiplier for including volume discounts etc. for spot prices. The spot price will be multiplied with the spotPriceMultiplier to determine the real cost
+  aws.spotPriceMultiplier: "0.95"
+```
+
+### On Demand pricing
+
+The On Demand price will be multiplied with the OnDemandPriceMultiplier to determine the real cost for On Demand instances.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: karpenter-global-settings
+  namespace: karpenter
+data:
+  # On Demand Multiplier for including volume discounts etc. to ensure choosing the cheapest available instance. The ondemand price will be multiplied with the onDemandPriceMultiplier to determine the real cost
+  aws.onDemandPriceMultiplier: "0.60"
+```
--- a/pkg/apis/settings/settings.go
+++ b/pkg/apis/settings/settings.go
@@ -42,6 +42,8 @@ var defaultSettings = &Settings{
 	InterruptionQueueName:      "",
 	Tags:                       map[string]string{},
 	ReservedENIs:               0,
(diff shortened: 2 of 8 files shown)


rkirk-r7:
When testing consolidation to verify karpenter would consolidate more expensive spot instance types to on-demand instance types we noticed that the following event log message was output
```
Normal   Unconsolidatable         6s                 karpenter              Can't replace with a cheaper node
```

Which we believe is from this https://github.com/aws/karpenter-core/blob/cbc03e98ef1457bf4c7e4397a67cc892fa4719b1/pkg/controllers/deprovisioning/helpers.go#L226-L252.
This appears to only consider the spot pricing if the provisioner supports both capacity types, rather than the cheaper price for on-demand, which this change would allow us to launch as a replacement node.

Looking for some advice on best way to proceed here, should we be submitting a additional PR to change the pricing decision making in the deprovisioning module too?
Is there any other areas of the code which we should also be reviewing/checking that might prioritise spot instances rather than the cheapest instance type?

rkirk-r7:
@njtran Yes of course, do I just need to attend one of the regular meetings listed here https://karpenter.sh/docs/contributing/working-group ?

njtran:
Hey @rkirk-r7, checking in here. Trying to remember if you have a path forward here for your issue? Was there a bug?

rkirk-r7:
@njtran Hey, sorry for being a bit tardy have just put a thread into the [slack channel](https://kubernetes.slack.com/archives/C04JW2J5J5P/p1696443497636939) to detail the issues we had run into hopefully more clearly than on the call.  Since the working group call we also went back and revisited the change and also put an alternative idea into that thread as well that we think would be a much less invasive change but will wait on some feedback about that approach before updating this PR

designs/discount-pricing.md
```diff
@@ -0,0 +1,55 @@
+# Discounted Pricing Support
+
+## Overview
+
+Karpenter is currently unaware of any discounted pricing, such as volume discounts or reserved instances/savings plans, which can lead to more expensive instances being chosen. For example an on demand instance with a savings plan discount may cost less than a spot instance of the same type. This pricing is apparently only available from the "payer" account, not any other child account API's for pricing.
```
sftim, line 5:
```suggestion
Karpenter is currently unaware of any discounted pricing, such as volume discounts or reserved instances/savings plans, which can lead to more expensive instances being chosen. For example an on demand instance with a savings plan discount may cost less than a spot instance of the same type. On AWS, this pricing is apparently only available from the "payer" account, not any other child account API's for pricing.
```

:thought_balloon: could Karpenter learn to assume an IAM role into the AWS account where discounts are visible?

designs/discount-pricing.md
```diff
@@ -0,0 +1,55 @@
...
+A multiplier will be applied to the price to allow any discounts to be applied and determine the real cost of an EC2 instance.
+For example, a multiplier of 0.9 would apply a 10% discount
+
+Separate multiplier values for Spot and On Demand pricing will be allowed to allow for accounts which have different pricing discounts for each type.
+
+The multiplier will default to a value of 1 so no discount will be applied unless explicitly enabled. 
+
+### Spot pricing
+
+The Spot price will be multiplied with the SpotPriceMultiplier to determine the real cost for Spot instances.
```
sftim, line 29:
```suggestion
The Spot price will be multiplied with the `aws.spotPriceMultiplier` to determine the real cost for Spot instances.
```

ellistarn (commented):
As discussed in the latest working group, I think we want to pursue an instance type overlay approach.

njtran:
@rkirk-r7, if I remember correctly, we discussed the instance overlay approach rather than the spot discount factor as this PR explores, and cutting a bug issue/fix related to pricing if you were able to. I'm noticing that this PR is close to being staled out as I don't think this was the agreed upon direction. Did you want to discuss more here? Or is this good to close?

rkirk-r7:
@njtran I can update this as I split out some defects this was trying to solve as well as the new feature  (https://github.com/aws/karpenter/issues/4923 / https://github.com/aws/karpenter/pull/4924). So does need cleaned up now.  We were asked to follow up with our TAM in last working group we attended and after meeting with EKS team there did seem to be some interest in the multiplier approach so I am waiting to hear more from that so would like to keep open whilst those discussions are on going if thats okay?

njtran:
> there did seem to be some interest in the multiplier approach 

Are you talking about from other users? Would be great if we could track these conversations somewhere.  Maybe https://github.com/aws/karpenter/issues/3860 ?

rkirk-r7:
> > there did seem to be some interest in the multiplier approach
> 
> Are you talking about from other users? Would be great if we could track these conversations somewhere. Maybe #3860 ?

Sorry for late reply, stuck with day job work. It was call with our TAM which we were advised to setup from the last working group call I attended

rkirk-r7:
Have removed the changes related to launching as #4924 has been raised instead.  Will need to revisit to update for the v1beta1 changes I had not realised the global settings had been changed so much

PR #4700, opened 2023-09-26
ci: Combine E2E clean-up with sweeper

engedaam:
Fixes #N/A 

**Description**
- Consolidate the clean-up logic for the E2E tests to the clean-up script that is used by the sweeper, such that all clean-up logic is in one place.

- Tested in a forked account 

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.github/actions/e2e/cleanup/action.yaml
+++ b/.github/actions/e2e/cleanup/action.yaml
@@ -27,71 +27,17 @@ runs:
     - uses: ./.github/actions/e2e/install-eksctl
       with:
         version: ${{ inputs.eksctl_version }}
-    - name: delete-instance-profiles
-      shell:
-      run: |
-        for name in $(aws iam list-instance-profiles --query "InstanceProfiles[*].{Name:InstanceProfileName}" --output text); do
-          tags=$(aws iam list-instance-profile-tags --instance-profile-name $name --output json || true)
-          if [[ $(echo $tags | jq -r '.Tags[] | select(.Key == "testing/cluster") | .Value') == "${{ inputs.cluster_name }}" ]]; then
-              echo "Deleting instance profile '$name'..."
-              roleName=$(aws iam get-instance-profile --instance-profile-name $name --query "InstanceProfile.Roles[*].{Name:RoleName}" --output text)
-              aws iam remove-role-from-instance-profile --instance-profile-name $name --role-name $roleName
-              aws iam delete-instance-profile --instance-profile-name $name
-          fi
-        done
     - name: delete-cluster
       shell: bash
       run: |
         eksctl delete cluster --name ${{ inputs.cluster_name }} --timeout 60m --wait || true
-    - name: delete-network-interfaces
-      shell: bash
+    - uses: actions/setup-go@v4
+      with:
+        go-version-file: test/hack/cleanup/go.mod
+        cache-dependency-path: test/hack/cleanup/go.sum
+        check-latest: true
+    - name: "Run cleanup script"
       run: |
-        aws ec2 describe-network-interfaces \
-          --filter Name=tag:cluster.k8s.amazonaws.com/name,Values=${{ inputs.cluster_name }} \
-          --query "NetworkInterfaces[*].NetworkInterfaceId" \
-          --output text |
-        xargs \
-          -n 1 \
-          -r \
-          aws ec2 delete-network-interface \
-            --network-interface-id
-    - name: delete-security-group
+        go run main.go ${{ inputs.cluster_name }}
+      working-directory: ./test/hack/cleanup
       shell: bash
-      # For drift testing, we create a security group and need to clean it up here
-      # to avoid leaks if the tests is not fully completed
-      run: |
-        aws ec2 describe-security-groups \
-          --filters Name=group-name,Values=security-group-drift Name=tag:karpenter.sh/discovery,Values=${{ inputs.cluster_name }} \
-          --query "SecurityGroups[*].{ID:GroupId}" \
-          --output text |
-        xargs \
-          -n 1 \
-          -r \
-          aws ec2 delete-security-group \
-            --group-id
-    - name: delete-iam-alpha-policy
-      shell: bash
-      run: |
-        aws iam delete-policy --policy-arn "arn:aws:iam::${{ inputs.account_id }}:policy/KarpenterControllerPolicy-Alpha-${{ inputs.cluster_name }}" || true
-    - name: delete-iam-policies-stack
-      shell: bash
-      run: |
-        aws cloudformation delete-stack --stack-name iam-${{ inputs.cluster_name }}
-        aws cloudformation wait stack-delete-complete --stack-name iam-${{ inputs.cluster_name }}
-    - name: delete-cluster-stack
-      shell: bash
-      run: |
-        aws cloudformation delete-stack --stack-name eksctl-${{ inputs.cluster_name }}-cluster || true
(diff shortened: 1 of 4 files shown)


.github/actions/e2e/cleanup/action.yaml
```diff
@@ -30,67 +30,17 @@ runs:
...
-      run: |
-        aws cloudformation delete-stack --stack-name iam-${{ inputs.cluster_name }}
-        aws cloudformation wait stack-delete-complete --stack-name iam-${{ inputs.cluster_name }}
-    - name: delete-cluster-stack
-      shell: bash
-      run: |
-        aws cloudformation delete-stack --stack-name eksctl-${{ inputs.cluster_name }}-cluster || true
-        aws cloudformation wait stack-delete-complete --stack-name eksctl-${{ inputs.cluster_name }}-cluster || true
-    - name: delete-launch-templates
+    - uses: actions/setup-go@v4
```
jonathan-innis, line 37: I think we do a setup at the beginning of the job so do we need to setup again here?

test/hack/cleanup/main.go
```diff
@@ -54,13 +56,18 @@ type CleanableResourceType interface {
...
+	DeleteByClusterName(context.Context, string) error
 }
 
 type MetricsClient interface {
 	FireMetric(context.Context, string, float64, string) error
 }
 
 func main() {
+	var clusterName string
+	if len(os.Args) != 0 {
```
jonathan-innis, line 68: Should we check if the length of the args is exactly what we expect (0 or 1). Maybe worth using a flag here to be more explicit as opposed to OS args?

test/hack/cleanup/main.go
```diff
@@ -54,13 +56,18 @@ type CleanableResourceType interface {
 	Type() string
 	Get(context.Context, time.Time) ([]string, error)
 	Cleanup(context.Context, []string) ([]string, error)
+	DeleteByClusterName(context.Context, string) error
```
jonathan-innis, line 59: Is it worth scoping our existing `Get()` logic to take in an option that would let you pass in the cluster name? If that's passed in for each of these test cases, add a filter? Alternatively, you could do a `GetByName()` call that got the ids and then call the same Cleanup() call for both sets of logic

.github/actions/e2e/cleanup/action.yaml
```diff
@@ -27,71 +27,12 @@ runs:
...
-        aws ec2 describe-launch-templates \
-          --filter Name=tag:karpenter.k8s.aws/cluster,Values=${{ inputs.cluster_name }} \
-          --query "LaunchTemplates[*].LaunchTemplateId" \
-          --output text |
-        xargs \
-          -n 1 \
-          -r \
-          aws ec2 delete-launch-template \
-            --launch-template-id
+      name: "Run cleanup script"
```
jonathan-innis, line 38: I generally think it reads a little better if the `name` field is sitting above the actual script that it's attached to

.github/actions/install-deps/action.yaml
```diff
@@ -7,14 +7,14 @@ inputs:
...
   steps:
-    - uses: actions/setup-go@v4
-      with:
-        go-version-file: go.mod
-        check-latest: true
-        cache-dependency-path: "**/go.sum"
     # Root path permission workaround for caching https://github.com/actions/cache/issues/845#issuecomment-1252594999
     - run: sudo chown "$USER" /usr/local
       shell: bash
+    - uses: actions/setup-go@v4
```
jonathan-innis, line 13: Why is this moved? Also, changing this `go.mod` file is going to result in some odd behavior and some potential conflicts if the `hack/go.mod` version doesn't match with the regular `go.mod` version

test/hack/cleanup/main.go
```diff
@@ -112,7 +119,7 @@ func (i *instance) Type() string {
 	return "Instances"
 }
 
-func (i *instance) Get(ctx context.Context, expirationTime time.Time) (ids []string, err error) {
+func (i *instance) Get(ctx context.Context, _ string, expirationTime time.Time) (ids []string, err error) {
```
jonathan-innis, line 122: We should be using the cluster-name here, right? Cleanup any instances that have the cluster name and provisioner name? Also, I think that this change needs to be updated to use the new beta label alongside the alpha label

test/hack/cleanup/main.go
```diff
@@ -169,16 +176,29 @@ func (sg *securitygroup) Type() string {
 	return "SecurityGroup"
 }
 
-func (sg *securitygroup) Get(ctx context.Context, expirationTime time.Time) (ids []string, err error) {
+func (sg *securitygroup) Get(ctx context.Context, clusterName string, expirationTime time.Time) (ids []string, err error) {
 	var nextToken *string
+	var ec2Filter []ec2types.Filter
+
+	if clusterName == "" {
+		ec2Filter = []ec2types.Filter{
```
jonathan-innis, line 184: We can't clean-up all the security groups by tag when we are cleaning up at the end of a test? That might be more reliable long-term than relying on name, which we are going to forget to update if we ever change that test

test/hack/cleanup/main.go
```diff
@@ -52,7 +54,7 @@ const (
 
 type CleanableResourceType interface {
 	Type() string
-	Get(context.Context, time.Time) ([]string, error)
+	Get(context.Context, string, time.Time) ([]string, error)
```
jonathan-innis, line 57: The cluster name is global, right? Should this be some value that's updated in a global variable that can be accessed from everywhere? Or maybe some value that's injected into the context?

test/hack/cleanup/main.go
engedaam: I'm not getting what you mean here

jonathan-innis: It looks like we are just looking for a single security group name here and I was wondering if we could make it so that the security groups are tagged a certain way during the test so that we can delete them off of their tag vs. off a specific name that might change over time

jonathan-innis (commented):
The more I look at this, the more I'm thinking that we need to have separate interface methods for handling the `Get()` call when we are trying to get all resources past a certain time and the `Get()` call when we are trying to get resources for a particular cluster. 

You're almost talking about a `Get()` and a `GetExpired()` or a `Get` and a `GetAll` method.

.github/actions/e2e/cleanup/action.yaml
```diff
@@ -27,71 +27,17 @@ runs:
...
-          if [[ $(echo $tags | jq -r '.Tags[] | select(.Key == "testing/cluster") | .Value') == "${{ inputs.cluster_name }}" ]]; then
-              echo "Deleting instance profile '$name'..."
-              roleName=$(aws iam get-instance-profile --instance-profile-name $name --query "InstanceProfile.Roles[*].{Name:RoleName}" --output text)
-              aws iam remove-role-from-instance-profile --instance-profile-name $name --role-name $roleName
-              aws iam delete-instance-profile --instance-profile-name $name
-          fi
-        done
+    - uses: actions/setup-go@v4
+      with:
+        go-version-file: test/hack/cleanup/go.mod
```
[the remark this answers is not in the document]
engedaam: We need to change the `go.mod` and `go.sum`

test/hack/cleanup/main.go
engedaam: The Idea was to delete only the SG from our drift test if that is being orphaned

jonathan-innis (approved):
Nice 🎉

PR #4718, opened 2023-09-27
docs: Fix website local build

dtzar:
**Description**
When I execute `make website` there are two errors:
1. from the troubleshooting markdown file
```
Error: error building site: "$.\karpenter\website\content\en\v0.27\troubleshooting.md:130:4": failed to parse Markdown attributes; you may need to quote the values
```
2. from the template which is resolved in an updated version of docsy
```
ERROR render of "page" failed: "$.\karpenter\website\themes\docsy\layouts\docs\baseof.html:4:7": execute of template failed: template: docs/single.html:4:7: executing "docs/single.html" at <partial "head.html" .>: error calling partial: execute of template failed: html/template:partials/head.html:20:13: no such template "_internal/google_news.html"
```

Locally updated submodule and ran `make website` using local hugo 0.119.0.

- [ ] Yes, issue opened: # 
- [ ] No


jonathan-innis:
This is a nice update, though I'm wondering if we can reduce the scope of this PR to fixing the local build as opposed to also bumping the version of Hugo in the docs. From looking over the preview, I'm thinking that we need to pretty careful about the version bump and should validate very carefully that everything still works since I noticed a few things just quickly moving through the docs:
1. The version dropdown in the header was broken
2. There is a tags field that shows up on the doc headline and in the sidebar that's unclear to me in its relevancy
3. The header style has shifted a bit so some of the elements are now left-aligned
4. The homepage has bigger font all of a sudden
5. Search result suggestions no longer appear when using the searchbar

dtzar:
I'm going to try and see if I can fix the issues with the site since the only-fix-some-issues also turns out to be non-trivial.

dtzar:
I *think* everything is working now, minus the tags.  Theoretically, this should disable it which is in the hugo.config file:
```
disableKinds:
  - taxonomy
```
but for some reason it doesn't take.  Maybe if this is the only thing wrong it's an acceptable trade-off?

jonathan-innis (dismissed):
LGTM, let's create issues for some of the things that we know we need to follow-up in from the upgrade

njtran:
Commenting here from slack for persistence, we're going to cut issues to handle the following issues:
- Tags sidebar, either remove or make sensible
- search bar seems to be cut off and smaller on Chrome, but not on Safari or Edge

Otherwise, everything looks good to me once deploy succeeds and CI succeeds.

Makefile
```diff
@@ -31,6 +31,7 @@ GETTING_STARTED_SCRIPT_DIR = website/content/en/preview/getting-started/getting-
 
 # Common Directories
 MOD_DIRS = $(shell find . -name go.mod -type f | xargs dirname)
+MOD_DIRS=`echo "${MOD_DIRS}" | sed 's/.\/website//g'`
```
njtran, line 34: I don't think we want to remove it from the env var here, since we use this MOD_DIRS for the `make tidy` and `make download` target so that we automatically do the go mod tidy. Can you just do it in the `make verify` target?

Makefile
```diff
@@ -124,6 +125,9 @@ verify: tidy download ## Verify code. Includes dependencies, linting, formatting
 	cp  $(KARPENTER_CORE_DIR)/pkg/apis/crds/* pkg/apis/crds
 	hack/validation/requirements.sh
 	hack/validation/labels.sh
+	if [ "${CI}" = true ]; then\
+		hack/validation/validate.sh;\
+	fi
```
njtran, line 130: I don't think there's a validation/validate.sh path. Did you leave out a new file? I don't think we need any more validations IIUC.

dtzar: Not sure where this came from.  I'll remove.

Makefile
```diff
@@ -124,6 +124,7 @@ verify: tidy download ## Verify code. Includes dependencies, linting, formatting
 	cp  $(KARPENTER_CORE_DIR)/pkg/apis/crds/* pkg/apis/crds
 	hack/validation/requirements.sh
 	hack/validation/labels.sh
+	MOD_DIRS=`echo "${MOD_DIRS}" | sed 's/.\/website//g'`
```
njtran, line 127: Would you be able to copy this to another environment variable so we make sure we're not editing the other calls here? 

Also, can you do the $(shell) syntax so that we follow the other in-line bash commands in the file? Just want to make sure we minimize bash weirdness.
```suggestion
	MOD_DIRS_VERIFY=$(shell echo $(MOD_DIRS) | sed 's/.\/website//g')
```

Makefile
```diff
@@ -124,7 +124,8 @@ verify: tidy download ## Verify code. Includes dependencies, linting, formatting
 	cp  $(KARPENTER_CORE_DIR)/pkg/apis/crds/* pkg/apis/crds
 	hack/validation/requirements.sh
 	hack/validation/labels.sh
-	$(foreach dir,$(MOD_DIRS),cd $(dir) && golangci-lint run $(newline))
+	MOD_DIRS=$(shell echo $(MOD_DIRS) | sed 's/.\/website//g')
```
jonathan-innis, line 127: I feel like we might be able to exclude the website in _all_ cases. When the `go.mod` was included, even in the `go mod tidy` step it was causing issues since `go` doesn't see the usage of the packages so it cleans up the `go.mod` to be empty every time

```makefile
MOD_DIRS = $(shell find . -path "./website" -prune -o -name go.mod -type f -print | xargs dirname
```

Makefile
```diff
@@ -195,7 +195,7 @@ issues: ## Run GitHub issue analysis scripts
 	./hack/github/label_issue_count.py > "karpenter-labels-$(shell date +"%Y-%m-%d").csv"
 
 website: ## Serve the docs website locally
-	cd website && npm install && git submodule update --init --recursive && hugo server
+	cd website && npm install && hugo server
```
jonathan-innis, line 198:
```suggestion
	cd website && npm install && hugo server && hugo mod tidy
```

We should consider running `hugo mod tidy` to ensure that the `go.mod` is up-to-date here

Makefile
```diff
@@ -124,7 +124,7 @@ verify: tidy download ## Verify code. Includes dependencies, linting, formatting
 	cp  $(KARPENTER_CORE_DIR)/pkg/apis/crds/* pkg/apis/crds
 	hack/validation/requirements.sh
 	hack/validation/labels.sh
-	$(foreach dir,$(MOD_DIRS),cd $(dir) && golangci-lint run $(newline))
+	$(foreach dir,$(MOD_DIRS_VERIFY),cd $(dir) && golangci-lint run $(newline))
```
jonathan-innis, line 127:
```suggestion
	$(foreach dir,$(MOD_DIRS),cd $(dir) && golangci-lint run $(newline))
```

dtzar:
Good find @jonathan-innis on disabling the tags.  Had to add the `term` to disable.

We should add this to the improvements.  I don't *think* this is actually causing a problem, but would be nice to get rid of the warning message during the `hugo server` build.

```
WARN  .Path when the page is backed by a file is deprecated. We plan to use Path for a canonical source path and you probably want to check the source is a file. To get the current behaviour, you can use a construct similar to the one below:

  {{ $path := "" }}
  {{ with .File }}
        {{ $path = .Path }}
  {{ else }}
        {{ $path = .Path }}
  {{ end }}  ```

jonathan-innis (approved):
LGTM 🚀

jonathan-innis:
Nice work @dtzar! Thanks for getting us up-to-date here!

PR #4734, opened 2023-10-02
test: move snapshot releases to separate ECR repo

jmdeal:
Fixes #N/A 

**Description**
This PR separates official releases and snapshot releases into separate ECR repositories. This enables the use of lifecycle policies to automatically expire snapshot releases after some given time period.

TBD

- [ ] Yes, issue opened: # 
- [x] No


.github/actions/authenticate-aws/action.yaml
```diff
@@ -1,9 +1,14 @@
 name: AuthenticateAWS
```
jonathan-innis, line 1: We should just drop this action. IMO, this composite action isn't giving us anything different than calling the `aws-actions/configure-aws-credentials` action

hack/release/common.sh
```diff
@@ -8,7 +8,9 @@ config(){
   RELEASE_REPO_ECR=${RELEASE_REPO_ECR:-public.ecr.aws/${ECR_GALLERY_NAME}/}
   RELEASE_REPO_GH=${RELEASE_REPO_GH:-ghcr.io/${GITHUB_ACCOUNT}/karpenter}
 
-  PRIVATE_PULL_THROUGH_HOST="${AWS_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com"
+  PRIVATE_HOST="${AWS_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com"
```
jonathan-innis, line 11:
```suggestion
  SNAPSHOT_HOST="${AWS_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com"
```

nit: naming consistency to show these concepts are related. This helps when someone comes in later and is trying to parse concepts that go together

hack/release/common.sh
```diff
@@ -29,6 +31,20 @@ versionData(){
   RELEASE_MINOR_VERSION="v${RELEASE_VERSION_MAJOR}.${RELEASE_VERSION_MINOR}"
 }
 
+snapshot() {
+  RELEASE_VERSION=$1
+  echo "Release Type: $(releaseType "${RELEASE_VERSION}")
```
jonathan-innis, line 36: Do we need this releaseType function now that we are differentiating between snapshot and official releases with different functions?

hack/release/release.sh
```diff
@@ -8,7 +8,7 @@ SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
 source "${SCRIPT_DIR}/common.sh"
 
 config
-release "$HEAD_HASH" #release a snapshot version
+snapshot "$HEAD_HASH" #release a snapshot version
```
jonathan-innis, line 11: Does it make sense for the release trigger to both create a snapshot and create an official release? I think we just need one. We should already have a snapshot from the merge when we tag

hack/release/common.sh
```diff
@@ -109,7 +126,7 @@ cosignImages() {
 pullPrivateReplica(){
```
jonathan-innis, line 126: I'm assuming that this is happening in a follow-on, but we can drop this function later.

jonathan-innis (commented):
Nice work 🎉  Just a few high-level comments

hack/release/common.sh
jmdeal: I think `PRIVATE_HOST` is appropriate since it's used for both the pull through cache based replicas as well as the snapshot repos. We've discussed dropping the replicas offline so I think it would make the most sense to rename it in a follow up PR.

hack/release/common.sh
jmdeal: I'll check if there's anywhere else its used but probably not. With some more refactoring we can definitely drop it.

hack/release/release.sh
jmdeal: RIght now the `install-karpenter` action relies upon the snapshot release, even for official releases. We could update it so it would check for both but I honestly think its simpler to keep it as is.

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

.github/workflows/release.yaml
```diff
@@ -27,17 +27,31 @@ jobs:
           fetch-depth: 0
           ref: ${{ env.TAG_COMMIT }}
       - uses: ./.github/actions/install-deps
-      - uses: ./.github/actions/authenticate-aws
+      - uses: aws-actions/configure-aws-credentials@v1-node16
```
jonathan-innis, line 30: Why do we need to create a snapshot on release? This commit should already have a snapshot from the initial merge?

hack/release/common.sh
jonathan-innis: I still disagree but not gonna block on it 👍🏻

hack/release/common.sh
```diff
@@ -29,15 +31,29 @@ versionData(){
...
+  buildImages ${SNAPSHOT_REPO_ECR}
+  cosignImages
+  publishHelmChart "karpenter" "${RELEASE_VERSION}" "${SNAPSHOT_REPO_ECR}"
+  publishHelmChart "karpenter-crd" "${RELEASE_VERSION}" "${SNAPSHOT_REPO_ECR}"
+}
+
 release() {
   RELEASE_VERSION=$1
-  echo "Release Type: $(releaseType "${RELEASE_VERSION}")
+  echo "Release Type: stable
```
jonathan-innis, line 50: Not sure that this echo is necessary anymore

jmdeal (commented):
/karpenter snapshot

hack/release/common.sh
```diff
@@ -29,15 +31,29 @@ versionData(){
   RELEASE_MINOR_VERSION="v${RELEASE_VERSION_MAJOR}.${RELEASE_VERSION_MINOR}"
 }
 
+snapshot() {
+  RELEASE_VERSION=$1
+  echo "Release Type: snapshot
```
jonathan-innis, line 36: nit: I think we can drop this echo

hack/release/release.sh
```diff
@@ -6,16 +6,7 @@ HEAD_HASH=$(git rev-parse HEAD)
 
 SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
```
jonathan-innis, line 7: Nice 🎉

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

hack/release/release.sh
jonathan-innis: > action relies upon the snapshot release, even for official releases

I'm not quite getting what you mean here. Install karpenter is only used for snapshot releases so why do we need to handle this for the official release?

.github/actions/e2e/install-karpenter/action.yaml
```diff
@@ -15,6 +15,9 @@ inputs:
     required: true
   git_ref:
     description: "The git commit, tag, or branch to check out. Requires a corresponding Karpenter snapshot release"
+  ecr_repo:
```
jonathan-innis, line 18: Any reason we need to pass the `ecr_repo` value through here as an input? I don't see a place in the workflows or actions where this value changes.

.github/workflows/codegen.yaml
```diff
@@ -21,7 +21,10 @@ jobs:
           git config user.email "APICodeGen@users.noreply.github.com"
           git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }}
           git config pull.rebase false
-      - uses: ./.github/actions/authenticate-aws
+      - uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          role-to-assume: 'arn:aws:iam::071440425669:role/GithubRelease'
+          aws-region: us-east-1
```
jonathan-innis, line 27: Region could be a repository environment variable as well

hack/release/snapshot.sh
```diff
@@ -0,0 +1,16 @@
...
+HEAD_HASH=$(git rev-parse HEAD)
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+source "${SCRIPT_DIR}/common.sh"
+config
+
+# Create a snapshot release
+snapshot "$HEAD_HASH"
+
+git reset --hard
```
jonathan-innis, line 14: I don't think we need these calls anymore if we aren't doing anything after this. Can just finish this script after the call to the snapshot function

.github/actions/e2e/install-karpenter/action.yaml
jmdeal: Not really, I wanted a constant since it was referenced in a couple of places but this probably isn't the way to do it. Is there a better way or would it be better to just hardcode it in at both locations?

hack/release/snapshot.sh
jmdeal: Wouldn't the 'dirty' state persist after this script ends? Since release is being called immediately after I didn't think we could remove it.

jmdeal (commented):
/karpenter snapshot

hack/release/release.sh
jmdeal: Discussed offline, we've separated the release workflow into two separate workflows: release and snapshot.

.github/workflows/release.yaml
jmdeal: Discussed offline, we've separated the release workflow into two separate workflows: release and snapshot.

hack/release/snapshot.sh
jmdeal: Since we've separated into separate workflows, state won't be shared. Resolving.

.github/workflows/merge-release.yaml
```diff
@@ -1,43 +1,30 @@
-name: Release
+name: MergeRelease
```
jonathan-innis, line 1:
```suggestion
name: Release
```

.github/workflows/merge-snapshot.yaml
```diff
@@ -0,0 +1,16 @@
+name: MergeSnapshot
```
jonathan-innis, line 1: Snapshot

.github/workflows/merge-snapshot.yaml
```diff
@@ -0,0 +1,16 @@
...
+    if: github.repository == 'aws/karpenter'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ./.github/actions/install-deps
+      - uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          role-to-assume: 'arn:aws:iam::${{ vars.ECR_ACCOUNT_ID }}:role/${{ vars.ECR_SNAPSHOT_ROLE_NAME }}'
+          aws-region: ${{ vars.ECR_REGION }}
+      - run: make snapshot
+      - run: make docgen
```
jonathan-innis, line 16: We have the `docgen` workflow so I think we can drop it from this file

.github/workflows/pr-snapshot.yaml
```diff
@@ -29,8 +29,11 @@ jobs:
           name: "${{ github.workflow }} / ${{ github.job }} (pull_request_review)"
           git_ref: ${{ env.PR_COMMIT }}
       - uses: ./.github/actions/install-deps
-      - uses: ./.github/actions/authenticate-aws
-      - run: make release
+      - uses: aws-actions/configure-aws-credentials@v1-node16
```
jonathan-innis, line 32: Nice 🎉

.github/workflows/publish-test-tools.yaml
```diff
@@ -18,6 +18,9 @@ jobs:
         with:
           fetch-depth: 0
       - uses: ./.github/actions/install-deps
-      - uses: ./.github/actions/authenticate-aws
+      - uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          role-to-assume: 'arn:aws:iam::${{ vars.ECR_ACCOUNT_ID }}:role/${{ vars.ECR_RELEASE_ROLE_NAME }}'
```
jonathan-innis, line 23:
```suggestion
          role-to-assume: 'arn:aws:iam::${{ vars.ECR_ACCOUNT_ID }}:role/${{ vars.ECR_SNAPSHOT_ROLE_NAME }}'
```

Makefile
```diff
@@ -169,7 +169,10 @@ codegen: ## Auto generate files based on AWS APIs response
 stable-release-pr: ## Generate PR for stable release
 	$(WITH_GOFLAGS) ./hack/release/stable-pr.sh
 
-release: ## Builds and publishes stable release if env var RELEASE_VERSION is set, or a snapshot release otherwise
+snapshot:
```
jonathan-innis, line 172:
```suggestion
snapshot: ## Builds and publishes a snapshot release
```

Comment for what this command does

Makefile
```diff
@@ -169,7 +169,10 @@ codegen: ## Auto generate files based on AWS APIs response
...
+	$(WITH_GOFLAGS) ./hack/release/snapshot.sh
+
+release: ## Builds and publishes stable release if env var RELEASE_VERSION is set
```
jonathan-innis, line 175: Is this statement about the RELEASE_VERSION still true or does it matter whether the current commit is tagged?

jmdeal: It's not actually been true, `release.sh` has been using the git tag to determine if it's a stable release rather than using `RELEASE_VERSION`. The current version will always create a release, though it might break if there isn't a valid git tag. Might be good to add a check for that.

jonathan-innis (approved):
Ship it! 🚀

PR #4735, opened 2023-10-02
feat: add support for mounted instance-store ephemeral storage

alec-rabold:
- Add option to RAID0 instance-store disks for node ephemeral-storage

Relevant issue: https://github.com/aws/karpenter/issues/2723

Something like this would allow instance-store disks to be used for ephemeral-storage and have Karpenter aware of the capacity.

Please let me know your thoughts!


runningman84:
it would be great to have this for bottlerocket too...

arabold-dd:
> it would be great to have this for bottlerocket too...

I believe bottlerocket should have this merged before that's possible to support here: https://github.com/bottlerocket-os/bottlerocket/issues/3060

alec-rabold:
> it would be great to have this for bottlerocket too...

Bottlerocket currently doesn't have native support for this (see: https://github.com/bottlerocket-os/bottlerocket/issues/3060).

With this change, you could still enable `MountNodeEphemeralStorage` to read in epehermal-storage capacity from the instance -- you'd just be responsible for configuring the disks in your user data script yourself.

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -316,6 +319,24 @@ type BlockDevice struct {
 	VolumeType *string `json:"volumeType,omitempty"`
 }
 
+// InstanceStoreConfigurationOption enumerates options for configuring instance store disks.
+// See: https://github.com/awslabs/amazon-eks-ami/blob/master/doc/USER_GUIDE.md
+// +kubebuilder:validation:Enum=MountNodeEphemeralStorage;MountIndividualDisks
+type InstanceStoreConfigurationOption string
```
jonathan-innis, line 325:
```suggestion
type InstanceStorePolicyOption string
```

I wonder if this is really more of a policy in upstream terms rather than a configuration.

pkg/providers/amifamily/bootstrap/eksbootstrap.go
```diff
@@ -78,6 +78,9 @@ func (e EKS) eksBootstrapScript() string {
 	if args := e.kubeletExtraArgs(); len(args) > 0 {
 		userData.WriteString(fmt.Sprintf(" \\\n--kubelet-extra-args '%s'", strings.Join(args, " ")))
 	}
+	if e.InstanceStoreConfiguration != nil {
+		userData.WriteString(fmt.Sprintf(" \\\n--local-disks %s", *e.InstanceStoreConfiguration))
```
jonathan-innis, line 82: Reading through the [EKS AMI docs](https://github.com/awslabs/amazon-eks-ami/blob/master/doc/USER_GUIDE.md), I think this needs to be one of "raid0" or "mount". From looking over this code change, it looks like this is just going to be set to one of the enum values that we have declared as part of our API

pkg/providers/instancetype/types.go
```diff
@@ -206,23 +206,29 @@ func memory(ctx context.Context, info *ec2.InstanceTypeInfo) *resource.Quantity
 	return mem
 }
 
-// Setting ephemeral-storage to be either the default value or what is defined in blockDeviceMappings
-func ephemeralStorage(amiFamily amifamily.AMIFamily, blockDeviceMappings []*v1beta1.BlockDeviceMapping) *resource.Quantity {
-	if len(blockDeviceMappings) != 0 {
+// Setting ephemeral-storage to be either the default value, what is defined in blockDeviceMappings, or the combined size of local store volumes.
+func ephemeralStorage(info *ec2.InstanceTypeInfo, amiFamily amifamily.AMIFamily, nodeClass *v1beta1.EC2NodeClass) *resource.Quantity {
+	// If local store disks have been configured for node ephemeral-storage, use the total size of the disks.
+	if lo.FromPtr(nodeClass.Spec.InstanceStoreConfiguration) == v1beta1.MountNodeEphemeralStorage {
```
jonathan-innis, line 212: Checking this assumption: If a user specifies this value with an AMIFamily that doesn't support this option, we just assume that a user has configured their userData correctly to automatically RAID0 these stores. I think this makes sense; just double-checking my understanding

pkg/utils/nodeclass/nodeclass.go
```diff
@@ -158,6 +159,19 @@ func NewBlockDeviceMappings(bdms []*v1alpha1.BlockDeviceMapping) []*v1beta1.Bloc
 	})
 }
 
+func NewInstanceStoreConfiguration(option *v1alpha1.InstanceStoreConfigurationOption) *v1beta1.InstanceStoreConfigurationOption {
```
jonathan-innis, line 162: Hopefully shouldn't need this now that we are dropping alpha support 🤞

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -316,6 +319,24 @@ type BlockDevice struct {
...
+
+const (
+	// MountNodeEphemeralStorage mounts instance store disks as node ephemeral-storage.
+	// A RAID-0 array is setup that includes all ephemeral NVMe instance storage disks.
+	// The containerd and kubelet state directories (`/var/lib/containerd` and `/var/lib/kubelet`)
+	// will then use the ephemeral storage for more and faster node ephemeral-storage. The node's ephemeral
+	// storage can be shared among pods that request ephemeral storage and container images that are downloaded to the node.
+	MountNodeEphemeralStorage InstanceStoreConfigurationOption = "MountNodeEphemeralStorage"
```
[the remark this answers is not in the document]
ellistarn: FWIW, I don't find either of these names intuitive, and had to read about what was going on to understand them. Specifically, `MountIndividialDisks` only makes sense in the context of the local static provisioner, which in my mind is equivalent to us just ignoring the disks and allowing other systems to use them. 

Thoughts on
```
instanceStorePolicy: RAID0 | Ignore
```

pkg/apis/v1beta1/ec2nodeclass.go
alec-rabold: Policy makes sense to me, changed in https://github.com/aws/karpenter/commit/5cc1de801310e130bf6332da9ac71d94f4dd9228

pkg/providers/amifamily/bootstrap/eksbootstrap.go
alec-rabold: Whoops, previously had these enum values as`raid0`/`mount`. Fixed in https://github.com/aws/karpenter/commit/5cc1de801310e130bf6332da9ac71d94f4dd9228

pkg/providers/instancetype/types.go
alec-rabold: That sounds correct. Though it could get tricky if/when a currently-unsupported AMIFamily starts supporting this in the future. _Maybe_ that warrants another policy option for using the instance-store size here without the explicit bootstrap arg?

pkg/utils/nodeclass/nodeclass.go
alec-rabold: 👍  rebased / dropped alpha files in https://github.com/aws/karpenter/commit/5cc1de801310e130bf6332da9ac71d94f4dd9228

pkg/apis/v1beta1/ec2nodeclass.go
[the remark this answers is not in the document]
alec-rabold: I'm onboard `RAID0 | Ignore` 👍 The "mount" part of the bootstrap script feels outside of Karpenter's scope. Do you think we need the explicit `Ignore` option given that'd be the default behavior?

Only other option I could think of is if a user needs more control over how the disks are formatted / mounted (while still intended to be used for ephemeral-storage). But in that case, a more generic system for "overrides" would probably make sense.

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -85,6 +85,9 @@ type EC2NodeClassSpec struct {
 	// +kubebuilder:validation:MaxItems:=50
 	// +optional
 	BlockDeviceMappings []*BlockDeviceMapping `json:"blockDeviceMappings,omitempty"`
+	// InstanceStoreConfiguration describes how to format and mount instance store disks.
+	// +optional
+	InstanceStoreConfiguration *InstanceStoreConfigurationOption `json:"instanceStoreConfiguration,omitempty"`
```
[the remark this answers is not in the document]
alec-rabold: Good point, TIL -- updated from `Configuration` --> `Policy` in https://github.com/aws/karpenter/commit/5cc1de801310e130bf6332da9ac71d94f4dd9228

ilkinmammadzada:
We are using custom AMI, our own bootstrapping script (+custom script to mount local NVMe as a ephemeral storage). We will just set "instanceStorePolicy: RAID0" and karpenter will consider local storage to calculate host ephemeral-storage. Did I understand right?

alec-rabold:
> We are using custom AMI, our own bootstrapping script (+custom script to mount local NVMe as a ephemeral storage). We will just set "instanceStorePolicy: RAID0" and karpenter will consider local storage to calculate host ephemeral-storage. Did I understand right?

Yep that's correct, it'll use the total size of the instance store volume(s) for ephemeral-storage capacity; only AL2 would RAID them. I wonder if it makes sense to have another policy option for using the instance store size and explicitly _not_ trying to RAID.

jonathan-innis:
> I wonder if it makes sense to have another policy option for using the instance store size and explicitly not trying to RAID

jonathan-innis:
> I wonder if it makes sense to have another policy option for using the instance store size and explicitly not trying to RAID

If you're using a custom AMI and don't need the RAID0 option explicitly passed through to the userData, then this shouldn't matter for you, since the custom AMI doesn't have any userData that's templated out by default. Seems reasonable to me to just start with `RAID0` or `Ignore` and then see if there is a need to expand this out later.

pkg/apis/v1beta1/ec2nodeclass.go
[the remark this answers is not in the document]
jonathan-innis: Agreed. I think we can consider handling that separately. This probably covers 95% of use-cases around how users want to handle these instance store volumes.

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -85,6 +85,9 @@ type EC2NodeClassSpec struct {
 	// +kubebuilder:validation:MaxItems:=50
 	// +optional
 	BlockDeviceMappings []*BlockDeviceMapping `json:"blockDeviceMappings,omitempty"`
+	// InstanceStorePolicy describes how to format and mount instance store disks.
+	// +optional
+	InstanceStorePolicy *InstanceStorePolicy `json:"instanceStorePolicy,omitempty"`
```
jonathan-innis, line 90: How do we handle the case where a user specifies `instanceStorePolicy: RAID0` but then also tries to mount the EBS volume on the same mount that the instance store is getting RAID0-ed and mounted on? Should we just disallow this case?

test/suites/integration/storage_test.go
```diff
@@ -206,6 +207,25 @@ var _ = Describe("Persistent Volumes", func() {
 	})
 })
 
+var _ = Describe("Ephemeral Storage", func() {
+	It("should run a pod with mounted ephemeral storage that exceeds EBS root block device mappings on AL2", func() {
+		nodeClass.Spec.InstanceStorePolicy = lo.ToPtr(v1beta1.InstanceStorePolicyRAID0)
+
+		pod := test.Pod(test.PodOptions{
+			ResourceRequirements: v1.ResourceRequirements{
+				Requests: v1.ResourceList{
```
jonathan-innis, line 216: Nice 🎉

jonathan-innis (commented):
Awesome work here 🎉

jonathan-innis (commented):
/karpenter snapshot

pkg/apis/v1beta1/ec2nodeclass.go
alec-rabold: Good point, not sure if there's a way to explicitly prevent that but it's probably worth calling out the mount point / device name in the docs

alec-rabold: Added this warning: https://github.com/aws/karpenter-provider-aws/blob/fdc8272f7c3ec8d7609dd56105433397b235511a/website/content/en/docs/concepts/nodeclasses.md?plain=1#L574-L576

website/content/en/docs/concepts/nodeclasses.md
```diff
@@ -551,6 +554,36 @@ spec:
...
+#### AL2
+
+On AL2, Karpenter automatically configures the disks through an additional boostrap argument (`--local-disks raid0`). The device name is `/dev/md/0` and its mount point is `/mnt/k8s-disks/0`. You should ensure any additional disk setup does not interfere with these.
+
+#### Others
+
+For all other AMI families, you must configure the disks yourself. Check out the [`setup-local-disks`](https://github.com/awslabs/amazon-eks-ami/blob/master/files/bin/setup-local-disks) script in [amazon-eks-ami](https://github.com/awslabs/amazon-eks-ami) to see how this is done for AL2.
+
+{{% alert title="Tip" color="secondary" %}}
+Since the Kubelet & Containerd will be using the instance-store filesystem, you may consider using a more minimal root volume size.
```
jonathan-innis, line 583: Since this is the case, should we consider setting this more minimal root volume size for users by default? Is there a conservative value that we could set here that would cover the snapshot size for AL2, Bottlerocket, and other AMIs?

pkg/providers/launchtemplate/suite_test.go
```diff
@@ -548,6 +548,16 @@ var _ = Describe("LaunchTemplates", func() {
 			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
 			ExpectNotScheduled(ctx, env.Client, pod)
 		})
+		It("should pack pods if the sum of pod ephemeral-storage and overhead exceeds node capacity and instance storage is mounted", func() {
+			nodeClass.Spec.InstanceStorePolicy = lo.ToPtr(v1beta1.InstanceStorePolicyRAID0)
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			pod := coretest.UnschedulablePod(coretest.PodOptions{ResourceRequirements: v1.ResourceRequirements{
```
jonathan-innis, line 554: nit: comment on how much storage this adds up to with overhead

pkg/providers/launchtemplate/suite_test.go
```diff
@@ -548,6 +548,16 @@ var _ = Describe("LaunchTemplates", func() {
...
+				Requests: map[v1.ResourceName]resource.Quantity{
+					v1.ResourceEphemeralStorage: resource.MustParse("19Gi"),
+				}}})
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
+			ExpectScheduled(ctx, env.Client, pod)
```
jonathan-innis, line 559: Worth checking the instance type that this scheduled to to ensure that it's an instance store instance type?

pkg/providers/instancetype/suite_test.go
```diff
@@ -654,6 +654,37 @@ var _ = Describe("InstanceTypes", func() {
...
+		ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+		pod := coretest.UnschedulablePod(coretest.PodOptions{
+			ResourceRequirements: v1.ResourceRequirements{
+				Requests: v1.ResourceList{v1.ResourceEphemeralStorage: resource.MustParse("5000Gi")},
+			},
+		})
+		ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
+		ExpectNotScheduled(ctx, env.Client, pod)
+	})
+	It("should not set pods to 110 if using ENI-based pod density", func() {
```
jonathan-innis, line 680: nit: looks like this test got duplicated

pkg/providers/launchtemplate/suite_test.go
jonathan-innis, line 559: We should probably also consider checking that we passed the correct userData through when we are using the AL2 ami here

pkg/providers/instancetype/suite_test.go
```diff
@@ -654,6 +654,37 @@ var _ = Describe("InstanceTypes", func() {
...
+		nodeClass.Spec.AMIFamily = lo.ToPtr(v1beta1.AMIFamilyAL2)
+		nodeClass.Spec.InstanceStorePolicy = lo.ToPtr(v1beta1.InstanceStorePolicyRAID0)
+		ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+		pod := coretest.UnschedulablePod(coretest.PodOptions{
+			ResourceRequirements: v1.ResourceRequirements{
+				Requests: v1.ResourceList{v1.ResourceEphemeralStorage: resource.MustParse("5000Gi")},
+			},
+		})
+		ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
+		ExpectScheduled(ctx, env.Client, pod)
```
jonathan-innis, line 667: Worth checking which instance type that this scheduled to and the size of the ephemeral storage on that instance type? Same as the launch template testing?

pkg/providers/launchtemplate/suite_test.go
alec-rabold: Makes sense to me, added check for the instance-store instance & ephemeral-storage capacity in https://github.com/aws/karpenter-provider-aws/pull/4735/commits/4bf56c65efa9007b3aee0871e6d554e370fa549b

alec-rabold: Good call, added https://github.com/aws/karpenter-provider-aws/pull/4735/commits/4bf56c65efa9007b3aee0871e6d554e370fa549b

pkg/providers/instancetype/suite_test.go
alec-rabold: added in https://github.com/aws/karpenter-provider-aws/pull/4735/commits/4bf56c65efa9007b3aee0871e6d554e370fa549b

pkg/providers/launchtemplate/suite_test.go
alec-rabold: actually this explicitly checks `m6idn.32xlarge`, can change that to be more dynamic if preferred

pkg/providers/launchtemplate/suite_test.go
alec-rabold: added in https://github.com/aws/karpenter-provider-aws/commit/4bf56c65efa9007b3aee0871e6d554e370fa549b

website/content/en/docs/concepts/nodeclasses.md
alec-rabold: Wasn't 100% sure on this without having the actual data to support it. Looks like AL2 is about 8 GiB so not all that much lower from the default `20GiB`. But if folks were previously bumping that up for ephemeral-storage and are now using instance-store, then that EBS volume would be unused.

Do you think that tip should instead say to use the default root volume size?

pkg/providers/launchtemplate/suite_test.go
jonathan-innis: Sounds good 👍🏼 The current state of the code lgtm

website/content/en/docs/concepts/nodeclasses.md
jonathan-innis: I think the way that you have written it is fine. If we start hearing reports that we should be dropping down the default volume size after this goes in, we can always change it later.

jonathan-innis:
@alec-rabold One thing I did notice as I was testing this out is that it seems like there is a discrepancy between what the the instance store volume reports (and therefore what the NodeClaim reports) and what the actual capacity of the node is for ephemeral storage when it comes up. Just doing some testing, this is what I got when I tried to launch a bunch of pods that requested `ephemeral-storage: 40Gi`.

### NodeClaim

```yaml
  allocatable:
    cpu: 940m
    ephemeral-storage: "52026258176"
    memory: 1392Mi
    pods: "8"
    vpc.amazonaws.com/pod-eni: "4"
  capacity:
    cpu: "1"
    ephemeral-storage: 59G
    memory: 1835Mi
    pods: "8"
    vpc.amazonaws.com/pod-eni: "4"
```

### Node

```yaml
  allocatable:
    cpu: 940m
    ephemeral-storage: "51969867689"
    hugepages-1Gi: "0"
    hugepages-2Mi: "0"
    hugepages-32Mi: "0"
    hugepages-64Ki: "0"
    memory: 1429916Ki
    pods: "8"
  capacity:
    cpu: "1"
    ephemeral-storage: 57556000Ki
    hugepages-1Gi: "0"
    hugepages-2Mi: "0"
    hugepages-32Mi: "0"
    hugepages-64Ki: "0"
    memory: 1883548Ki
```

Any idea what might be causing the discrepancy? Looks like it's off by almost 4Gi.

alec-rabold:
> Any idea what might be causing the discrepancy? Looks like it's off by almost 4Gi.

It looks like there is a difference but I think it's closer than `4Gi`:
```
57556000 Ki (Kibi) --> 58.937344 G (Giga)

57556000 Ki (Kibi) --> 54.88967895508 Gi (Gibi)
```

The `ephemeral-storage: 59G` in `NodeClaim` would be the former conversion; did you get a `4Gi` difference by converting it like the latter?

For `0.062656 G` difference could it be due to disk partitions? TBH I'm not totally sure how cAdvisor calculates the root-dir size. That's also assuming that the instance-store's advertised [`TotalSizeInGB`](https://github.com/aws/karpenter-provider-aws/blob/1b896293c4969424235d9465cdebac090b7e4d7b/pkg/providers/instancetype/types.go#L210) is 100% accurate

jonathan-innis:
> The ephemeral-storage: 59G in NodeClaim would be the former conversion; did you get a 4Gi difference by converting it like the latter

Oh yep. You're totally right. I just blindly assumed that the ephemeral storage was coming back in `Gi` rather than `G`. Looking back at it, the data is correct. Seems completely reasonable to me 👍🏼

website/content/en/docs/concepts/nodeclasses.md
```diff
@@ -56,6 +56,9 @@ spec:
   amiSelectorTerms:
```
jonathan-innis, line 56: One thing to callout in our docs changes. This shouldn't go in `docs` (I know it's confusing) because this is the most recently _released_ version of our docs. This belongs in preview since we promote that version of the docs to the latest release version when we are ready to ship it.

jonathan-innis:
One comment on documentation, but otherwise this LGTM to ship 🚢

jonathan-innis (commented):
/karpenter snapshot

website/content/en/docs/concepts/nodeclasses.md
alec-rabold: Gotcha, I follow :+1: moved into `/preview` in https://github.com/aws/karpenter-provider-aws/commit/bb62cc9c65b4fd39dd5b3fcd26059e352a592028. There's still that new [example](https://github.com/aws/karpenter-provider-aws/blob/bb62cc9c65b4fd39dd5b3fcd26059e352a592028/examples/v1beta1/instance-store-ephemeral-storage.yaml) though -- should it be moved/removed/commented?

jonathan-innis: The example is fine where it's at 👍🏼

jonathan-innis (approved):
LGTM 🚀

jonathan-innis (commented):
/karpenter snapshot

ajayOO8:
We were testing this feature and found a bit random behavior.  For same instance-type, the allocatable capacity values of ephemeral-storage comes out different. 

> {"level":"INFO","time":"2024-02-22T03:43:44.467Z","logger":"controller.nodeclaim.lifecycle","message":"launched nodeclaim","commit":"17d6c05","nodeclaim":"default-uswest1cstageg-rtr24","provider-id":"aws:///us-west-1c/i-foobar","instance-type":"m5ad.4xlarge","zone":"us-west-1c","capacity-type":"spot","allocatable":{"cpu":"13890m","ephemeral-storage":"114176Mi","memory":"55169334Ki","pods":"200","vpc.amazonaws.com/pod-eni":"54"}}

the correct one
> {"level":"INFO","time":"2024-02-22T02:19:00.132Z","logger":"controller.nodeclaim.lifecycle","message":"launched nodeclaim","commit":"17d6c05","nodeclaim":"default-uswest1astagef-5xpmd","provider-id":"aws:///us-west-1a/i-fooobar","instance-type":"m5ad.4xlarge","zone":"us-west-1a","capacity-type":"spot","allocatable":{"cpu":"13890m","ephemeral-storage":"537852516352","memory":"55169334Ki","pods":"200","vpc.amazonaws.com/pod-eni":"54"}}

We also have blockDeviceMappings config in nodeClass configuration if it helps:

> apiVersion: karpenter.k8s.aws/v1beta1
> kind: EC2NodeClass
> metadata:
>   name: default-useast1a
> spec:
>   amiFamily: Custom
>   instanceStorePolicy: RAID0
>     tags:
>   blockDeviceMappings:
>   - deviceName: /dev/sda1
>     ebs:
>       deleteOnTermination: true
>       volumeSize: 64Gi
>       volumeType: gp2

what could be the possible reason for this? Although this isn't but could lead to bad decision making by karpenter if there are enough wrong ones?

PR #4744, opened 2023-10-03
feat: v1beta1

jonathan-innis:
Fixes #N/A 

**Description**

This PR enables the v1beta1 `v1beta1/NodeClass` controllers as well as enabling the v1beta1 webhooks into Karpenter. 

- [ ] Yes, issue opened: # 
- [x] No


website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
```diff
@@ -143,7 +142,7 @@ Resources:
                   "aws:ResourceTag/kubernetes.io/cluster/${ClusterName}": "owned"
                 },
                 "StringLike": {
-                  "aws:ResourceTag/karpenter.sh/provisioner-name": "*"
+                  "aws:ResourceTag/karpenter.sh/nodepool": "*"
```
gtomolillo, line 145: This should keep the support for both `karpenter.sh/provisioner-name` and `karpenter.sh/nodepool` in this release. It is necessary to facilitate the transition and allow karpenter to deprovision nodes created with the previous release.

charts/karpenter/values.yaml
gtomolillo, line 1: Shouldn't `featureGates.driftEnabled` be set to `true`?

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
jonathan-innis: Correct. The CF that we're going to include in the preview docs isn't going to have both of these permissions. The easiest way that I was thinking the migration could occur is if a user applies a separate policy (which would be this new policy) and then drops the old policy when they are done.

jonathan-innis: That keeps things separate rather than having to unwind all of the permissions if they are collapsed into the same policy.

charts/karpenter/values.yaml
```diff
@@ -153,8 +152,28 @@ webhook:
...
+  errorOutputPaths:
+    - stderr
+  # -- Log encoding - defaults to console - must be one of 'json', 'console'
+  logEncoding: console
+  # -- Component-based log configuration
+  logLevel:
+    # -- Global log level, defaults to 'debug'
+    global: debug
+    # -- Controller log level, defaults to 'debug'
+    controller: debug
```
sftim, line 174: For beta, maybe we'd like a more production-friendly default?

website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml
sftim, line 1: Is this still needed?
```
  KarpenterNodeInstanceProfile:
    Type: "AWS::IAM::InstanceProfile"
    Properties:
      InstanceProfileName: !Sub "KarpenterNodeInstanceProfile-${ClusterName}"
      Path: "/"
      Roles:
        - !Ref "KarpenterNodeRole"
```

If it is, would we tag it (eg as deprecated)?

jonathan-innis: I dropped it. Thanks for the reminder 👍

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

cmd/controller/main.go
```diff
@@ -64,6 +64,7 @@ func main() {
 			op.SubnetProvider,
 			op.SecurityGroupProvider,
 			op.InstanceProfileProvider,
+			op.InstanceProvider,
```
njtran, line 67: Was this erroneously removed?

hack/docs/configuration_gen_docs.go
```diff
@@ -48,11 +48,12 @@ func main() {
...
 	bottomDoc := fmt.Sprintf("\n%s%s", genEnd, endDocSections[1])
 
-	opts := options.New()
+	fs := flag.NewFlagSet("karpenter", flag.ContinueOnError)
+	(&options.Options{}).AddFlags(fs)
 
 	envVarsBlock := "| Environment Variable | CLI Flag | Description |\n"
 	envVarsBlock += "|--|--|--|\n"
-	opts.FlagSet.VisitAll(func(f *flag.Flag) {
+	fs.VisitAll(func(f *flag.Flag) {
```
njtran, line 56: Not looking into the package, but this like this is just another way to do the same thing. Is there a reason this change was made?

pkg/providers/instance/instance.go
```diff
@@ -136,15 +135,11 @@ func (p *Provider) Get(ctx context.Context, id string) (*Instance, error) {
...
-	tagKeys := []string{v1alpha5.ProvisionerNameLabelKey}
-	if nodepoolutil.EnableNodePools {
-		tagKeys = append(tagKeys, corev1beta1.NodePoolLabelKey)
-	}
 	err := p.ec2api.DescribeInstancesPagesWithContext(ctx, &ec2.DescribeInstancesInput{
 		Filters: []*ec2.Filter{
 			{
 				Name:   aws.String("tag-key"),
-				Values: aws.StringSlice(tagKeys),
+				Values: aws.StringSlice([]string{v1alpha5.ProvisionerNameLabelKey, corev1beta1.NodePoolLabelKey}),
```
njtran, line 142: Is this an OR or an AND?

pkg/webhooks/webhooks.go
```diff
@@ -60,4 +61,5 @@ func NewCRDValidationWebhook(ctx context.Context, _ configmap.Watcher) *controll
 var Resources = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
 	v1alpha1.SchemeGroupVersion.WithKind("AWSNodeTemplate"): &v1alpha1.AWSNodeTemplate{},
 	corev1alpha5.SchemeGroupVersion.WithKind("Provisioner"): &v1alpha5.Provisioner{},
+	v1beta1.SchemeGroupVersion.WithKind("EC2NodeClass"):     &v1beta1.EC2NodeClass{},
```
njtran, line 64: No NodePool needed here?

website/content/en/preview/reference/metrics.md
```diff
@@ -50,6 +50,26 @@ Amount of time required for a replacement machine to become initialized.
 ### `karpenter_deprovisioning_replacement_machine_launch_failure_counter`
 The number of times that Karpenter failed to launch a replacement node for deprovisioning. Labeled by deprovisioner.
 
+## Disruption Metrics
+
+### `karpenter_disruption_actions_performed_total`
+Number of disruption methods performed. Labeled by disruption type.
```
njtran, line 56: does this make more sense?
```suggestion
Number of disruption actions performed. Labeled by disruption method.
```

charts/karpenter/values.yaml
jonathan-innis: We are going to set this value to `true` in `v0.33.0` as mentioned in the blogpost and the v1beta1 reference.

hack/docs/configuration_gen_docs.go
jonathan-innis: It's not the same way, we need that flagset so that we can add our own internal flags to the call; I don't think there is another way to do this than the way that we are doing it here

pkg/providers/instance/instance.go
jonathan-innis: This is an OR

jonathan-innis: It's: there must be a value from this set of things

pkg/webhooks/webhooks.go
jonathan-innis: Nope! Those webhook validations were added in the core operator package

website/content/en/preview/reference/metrics.md
jonathan-innis: Ooops. Yep, should probably change this in core, then

jonathan-innis: Updated

cmd/controller/main.go
jonathan-innis: Nope! This function changed when we added the tagging controller since that is the only controller that uses the instance provider

jonathan-innis (commented):
/karpenter snapshot

njtran (approved):

PR #4746, opened 2023-10-03
feat: Adding CEL validation for the `EC2NodeClass`

engedaam:
Fixes #N/A 

**Description**
- Adding CEL validation in the effort of moving away from webhook validation.
- https://github.com/aws/karpenter-core/issues/103

- `make test`

- [ ] Yes, issue opened: # 
- [x] No


pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -77,9 +77,21 @@ spec:
                       description: Tags is a map of key/value tags used to select
                         subnets Specifying '*' for a value selects all values for
                         a given tag key.
+                      maxProperties: 30
                       type: object
+                      x-kubernetes-validations:
+                      - message: empty tag keys and values aren't supported
```
jonathan-innis, line 83:
```suggestion
                      - message: empty tag keys or values aren't supported
```

nit: this is either/or not both

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -77,9 +77,21 @@ spec:
                       description: Tags is a map of key/value tags used to select
                         subnets Specifying '*' for a value selects all values for
                         a given tag key.
+                      maxProperties: 30
```
jonathan-innis, line 80: Can we do this in reverse order, with this value being 20 and the number of terms being 30. I'd rather support more terms and less tag keys/values here

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -143,18 +155,41 @@ spec:
                             and sc1: 125-16,384 \n * standard: 1-1,024"
                           pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                           x-kubernetes-int-or-string: true
+                          x-kubernetes-validations:
+                          - rule: self.endsWith('Gi') || self.endsWith('Ti') || self.endsWith('G')
```
jonathan-innis, line 159: Would it be simpler to enforce this with pure regex? Your call but I think you might have an easier time writing the rule vs doing 4 or 5 rules here that do conversions into doubles. Regex should also be less computationally expensive  then the current ruleset that you have

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -143,18 +155,41 @@ spec:
...
+                      x-kubernetes-validations:
+                      - message: snapshotID or volumeSize must be defined
+                        rule: has(self.snapshotID) || has(self.volumeSize)
                     rootVolume:
                       description: RootVolume is a flag indicating if this device
                         is mounted as kubelet root dir. You can configure at most
                         one root volume in BlockDeviceMappings.
                       type: boolean
                   type: object
+                maxItems: 15
```
jonathan-innis, line 192: Just curious. Is there a maximum number of block device mappings that are enforced by EC2? If so, should we enforce the limit that they impose? I think 15 may be a reasonable amount but curious if we just adopt what they say.

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -203,6 +244,10 @@ spec:
                       but this parameter is not specified, the default value is 2.
                     format: int64
                     type: integer
+                    x-kubernetes-validations:
```
jonathan-innis, line 247: It it possible to disable this field? If so, then 0 should be disabled.

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -228,6 +276,9 @@ spec:
                   collection and drift handling is implemented for the old instance
                   profiles on an update.
                 type: string
+                x-kubernetes-validations:
+                - message: Immutable field changed
```
jonathan-innis, line 280:
```suggestion
                - message: immutable field changed
```

The rest of the errors have been lowercase

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -250,9 +301,25 @@ spec:
                       description: Tags is a map of key/value tags used to select
                         subnets Specifying '*' for a value selects all values for
                         a given tag key.
+                      maxProperties: 30
```
jonathan-innis, line 304: Same comment as amiSelectorTerms. I'd like to see us be more restrictive on the number of tags and less restrictive on the number of terms

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -250,9 +301,25 @@ spec:
...
                       type: object
+                      x-kubernetes-validations:
+                      - message: tags cannot be empty
```
jonathan-innis, line 307: Does this rule apply if the field is ommitted?

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -250,9 +301,25 @@ spec:
...
+                        rule: self.size() != 0
+                      - message: empty tag keys and values aren't supported
```
jonathan-innis, line 309:
```suggestion
                      - message: empty tag keys or values aren't supported
```
nit

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -250,9 +301,25 @@ spec:
...
                   type: object
+                maxItems: 20
                 type: array
+                x-kubernetes-validations:
+                - message: securityGroupSelectorTerms cannot be empty
+                  rule: self.size() != 0
+                - message: '''id'' is mutually exclusive, cannot be set with a combination
+                    of other fields in'
+                  rule: '!self.all(x, has(x.id) && (has(x.tags) || has(x.name)))'
+                - message: '''name'' is mutually exclusive, cannot be set with a combination
```
jonathan-innis, line 320: Seems like you didn't finish the message here

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -250,9 +301,25 @@ spec:
...
+                        rule: self.size() != 0
+                      - message: empty tag keys and values aren't supported
+                        rule: self.all(k, k != '' && self[k] != '')
                   type: object
+                maxItems: 20
                 type: array
+                x-kubernetes-validations:
+                - message: securityGroupSelectorTerms cannot be empty
+                  rule: self.size() != 0
+                - message: '''id'' is mutually exclusive, cannot be set with a combination
```
jonathan-innis, line 317: Seems like you didn't finish the message here

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -250,9 +301,25 @@ spec:
...
+                      - message: tags cannot be empty
+                        rule: self.size() != 0
+                      - message: empty tag keys and values aren't supported
+                        rule: self.all(k, k != '' && self[k] != '')
                   type: object
+                maxItems: 20
                 type: array
+                x-kubernetes-validations:
+                - message: securityGroupSelectorTerms cannot be empty
+                  rule: self.size() != 0
```
jonathan-innis, line 316: Does this rule apply when the field is ommitted?

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -271,15 +338,39 @@ spec:
...
                 description: Tags to be applied on ec2 resources like instances and
                   launch templates.
                 type: object
+                x-kubernetes-validations:
+                - message: empty tag keys aren't supported
+                  rule: self.all(k, k != '')
+                - message: tag contains in restricted tag matching kubernetes.io/cluster/
+                  rule: self.all(k, !k.startsWith('kubernetes.io/cluster') )
+                - message: tag contains in restricted tag matching karpenter.sh/provisioner-name
+                  rule: self.all(k, !k.startsWith('karpenter.sh/provisioner-name'))
```
jonathan-innis, line 369: I think everything here except for the `kubernetes.io/cluster` key are equals checks rather than a `startsWith` check

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
engedaam: I was not able to find an exact number. From what I can find the maximum number of volumes that can be attached for an instance can vary on the type.

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
engedaam: The field can't be omitted, its required in the CRD

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
engedaam: We can't use Regex on an object that is not a string

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
engedaam: What do you mean by disable? 0 will not be considered as null value

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
jonathan-innis: Ah yes, you're right :)

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
jonathan-innis: Isn't this a string? Or does it fail because this is an int-or-string. I'd assume that this is being stored by the apiserver on the backend as a string. You could also override this to be a string type since I don't think we are going to support small values like integers anyways

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -133,22 +145,35 @@ spec:
                           format: int64
                           type: integer
                         volumeSize:
+                          allOf:
+                          - pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                          - pattern: ^((?:[1-9][0-9]{0,3}|[1-4][0-9]{4}|[5][0-8][0-9]{3}|59000)(G|Gi)|([0-9]||[0-9][0-7]|58)(T|Ti))$
```
jonathan-innis, line 150: Discussing offline. I think we could potentially separate out expressions for G and T to be closer to their maximums

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -250,9 +289,25 @@ spec:
...
+                      x-kubernetes-validations:
+                      - message: empty tag keys or values aren't supported
+                        rule: self.all(k, k != '' && self[k] != '')
                   type: object
+                maxItems: 30
                 type: array
+                x-kubernetes-validations:
+                - message: securityGroupSelectorTerms cannot be empty
+                  rule: self.size() != 0
+                - message: must contain tags, id, or name
```
jonathan-innis, line 303: This is different from the style of the messages in the `amiSelectorTerms`. Can we make sure that we are consistent with the messages that we are using 

```
- message: expected at least one, got none, ['tags', 'id', 'name']
```

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -271,15 +326,39 @@ spec:
...
+                      x-kubernetes-validations:
+                      - message: empty tag keys oe values aren't supported
+                        rule: self.all(k, k != '' && self[k] != '')
                   type: object
+                maxItems: 30
                 type: array
+                x-kubernetes-validations:
+                - message: subnetSelectorTerms cannot be empty
+                  rule: self.size() != 0
+                - message: must contain tags or id
```
jonathan-innis, line 340: Same comment here on message style. Can we take the same formatting that we did for `amiSelectorTerms` in the message and apply that everywhere?

pkg/apis/v1beta1/ec2nodeclass_validation_cel_test.go
```diff
@@ -0,0 +1,576 @@
...
+	"github.com/aws/karpenter/pkg/apis/v1beta1"
+	"github.com/aws/karpenter/pkg/test"
+)
+
+var _ = Describe("CEL/Validation", func() {
+	var nc *v1beta1.EC2NodeClass
+
+	BeforeEach(func() {
+		v := lo.Must1(awsEnv.VersionProvider.Get(ctx))
+		if v == "1.23" || v == "1.24" {
```
jonathan-innis, line 41: Can we just check this from the environment in the same way that we check it in `karpenter-core`? Not sure that we need to utilize the version provider here

pkg/apis/v1beta1/ec2nodeclass_validation_cel_test.go
```diff
@@ -0,0 +1,576 @@
...
+		})
+		It("should succeed if tags aren't in restricted tag keys", func() {
+			nc.Spec.Tags = map[string]string{
+				"karpenter.sh/custom-key": "value",
+				"karpenter.sh/managed":    "true",
+				"kubernetes.io/role/key":  "value",
+			}
+			Expect(env.Client.Create(ctx, nc)).To(Succeed())
+		})
+		It("should succeed by validating that regex is properly escaped", func() {
```
jonathan-innis, line 83: We're not using regex anymore so I'm not sure if this test is needed anymore

pkg/apis/v1beta1/ec2nodeclass_validation_cel_test.go
```diff
@@ -0,0 +1,576 @@
...
+			Expect(hash).To(Equal(updatedHash))
+		})
+		It("should expect two provisioner with the same spec to have the same provisioner hash", func() {
+			otherNodeClass := test.EC2NodeClass(v1beta1.EC2NodeClass{
+				Spec: nodeClass.Spec,
+			})
+			Expect(nodeClass.Hash()).To(Equal(otherNodeClass.Hash()))
+		})
+	})
+	Context("BlockDeviceMappings", func() {
```
jonathan-innis, line 539: Can we validate some things in here about the new blockDeviceMapping logic now that we are doing Regex here instead of what we were doing before

pkg/apis/v1beta1/ec2nodeclass_validation_cel_test.go
```diff
@@ -0,0 +1,576 @@
...
+			updatedHash := nodeClass.Hash()
+			Expect(hash).To(Equal(updatedHash))
+		})
+		It("should expect two provisioner with the same spec to have the same provisioner hash", func() {
+			otherNodeClass := test.EC2NodeClass(v1beta1.EC2NodeClass{
+				Spec: nodeClass.Spec,
+			})
+			Expect(nodeClass.Hash()).To(Equal(otherNodeClass.Hash()))
+		})
+	})
```
jonathan-innis, line 538: I don't see any checks here for the MetadataOptions. Can we add some tests to validate them and ensure things like the enum are working?

pkg/apis/v1beta1/ec2nodeclass_validation_cel_test.go
```diff
@@ -0,0 +1,576 @@
...
+	"knative.dev/pkg/apis"
+
+	"github.com/aws/aws-sdk-go/aws"
+
+	"github.com/aws/karpenter/pkg/apis/v1alpha1"
+	"github.com/aws/karpenter/pkg/apis/v1beta1"
+	"github.com/aws/karpenter/pkg/test"
+)
+
+var _ = Describe("CEL/Validation", func() {
```
jonathan-innis, line 36: I don't see a check here for validating that amiSelector is sepcfied when AMIFamily is "Custom"

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
engedaam: We can't have multiple patterns here, without manually editing the crds. Kubebuilder does not support that

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -271,15 +326,39 @@ spec:
                       description: Tags is a map of key/value tags used to select
                         subnets Specifying '*' for a value selects all values for
                         a given tag key.
+                      maxProperties: 20
                       type: object
+                      x-kubernetes-validations:
+                      - message: empty tag keys oe values aren't supported
```
jonathan-innis, line 332:
```suggestion
                      - message: empty tag keys or values aren't supported
```

nit

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -271,15 +326,39 @@ spec:
...
+                x-kubernetes-validations:
+                - message: empty tag keys aren't supported
+                  rule: self.all(k, k != '')
+                - message: tag contains in restricted tag matching kubernetes.io/cluster/
+                  rule: self.all(k, !k.startsWith('kubernetes.io/cluster') )
+                - message: tag contains in restricted tag matching karpenter.sh/provisioner-name
+                  rule: self.all(k, k != 'karpenter.sh/provisioner-name')
+                - message: tag contains in restricted tag matching karpenter.sh/nodepool
+                  rule: self.all(k, k != 'karpenter.sh/nodepool')
+                - message: tag contains in restricted tag matching karpenter.sh/managed-by
```
jonathan-innis, line 360:
```suggestion
                - message: tag contains a restricted tag matching karpenter.sh/managed-by
```

jonathan-innis (approved):
LGTM 🚀

engedaam (commented):
/karpenter snapshot

PR #4763, opened 2023-10-06
ci: disable china price gen

jmdeal:
Fixes #N/A 

**Description**
Separates out US and China price generation. This enables temporarily disabling china price gen.

`make codegen`

- [ ] Yes, issue opened: # 
- [x] No


pkg/providers/pricing/pricing.go
```diff
@@ -415,6 +416,14 @@ func (p *Provider) Reset() {
 	p.onDemandPrices = staticPricing
 	// default our spot pricing to the same as the on-demand pricing until a price update
 	p.spotPrices = populateInitialSpotPricing(staticPricing)
-	p.onDemandUpdateTime = initialPriceUpdate
-	p.spotUpdateTime = initialPriceUpdate
+	p.onDemandUpdateTime = getInitialPriceUpdateByRegion(p.region)
+	p.spotUpdateTime = getInitialPriceUpdateByRegion(p.region)
+}
+
+func getInitialPriceUpdateByRegion(region string) time.Time {
```
jonathan-innis, line 423: nit: `lo.Ternary` for this?

pkg/providers/pricing/pricing.go
```diff
@@ -405,6 +405,7 @@ func populateInitialSpotPricing(pricing map[string]float64) map[string]zonal {
 }
 
 func (p *Provider) Reset() {
+	initialOnDemandPrices := lo.Assign(initialOnDemandPricesUS, initialOnDemandPricesUS)
```
jonathan-innis, line 408: You're assigning US here twice so CN isn't included

jonathan-innis: Since this wasn't caught by our testing, can we make sure to add a test to validate that we get the right data from the china region?

jonathan-innis, line 408: Store this in a top-level var? Then you don't have to get this from a local value, you just always respect the global.

hack/codegen.sh
```diff
@@ -19,11 +19,11 @@ bandwidth() {
...
 
 pricing() {
-  GENERATED_FILE="pkg/providers/pricing/zz_generated.pricing.go"
-  NO_UPDATE=$' pkg/providers/pricing/zz_generated.pricing.go | 4 ++--\n 1 file changed, 2 insertions(+), 2 deletions(-)'
+  GENERATED_FILE="pkg/providers/pricing/zz_generated.pricing_us.go"
+  NO_UPDATE=$' pkg/providers/pricing/zz_generated.pricing_us.go | 4 ++--\n 1 file changed, 2 insertions(+), 2 deletions(-)'
   SUBJECT="Pricing"
 
-  go run hack/code/prices_gen.go -- "${GENERATED_FILE}"
+  go run hack/code/prices_gen.go --region us --output "${GENERATED_FILE}"
```
jonathan-innis, line 26:
```suggestion
  go run hack/code/prices_gen.go --partition us --output "${GENERATED_FILE}"
```

I think we are actually talking about the partition here. These are the partition names, it's something to think about if we should just match the file naming and our CLI arg alignment with the well-known partition naming: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html

jmdeal: Seems reasonable, I can split it out into the three partitions (`aws`, `aws-cn`, and `aws-us-gov`) rather than `us` and `cn`.

pkg/providers/pricing/pricing.go
jmdeal: Wow that's a miss...I guess that's why we do reviews. Are you thinking that test would just validate that the regions we expect data to exist for, `us-east-1`, `us-gov-east-1`, `us-gov-west-1`, and `cn-north-1`, all do?

jmdeal: I could go either way, if it was here alone for sure. It's also used [here](https://github.com/aws/karpenter/blob/2758afa7fe86b40475193c3322ef70fd0a1c391f/pkg/providers/pricing/pricing.go#L152) though so I separated it out since a ternary seemed a little messy there.

pkg/providers/pricing/pricing.go
jmdeal: Unless we don't plan on adding `cn-northwest-1` to this, I think this is cleaner than a ternary, especially since this is used in two places.

jonathan-innis: To me, that's over-optimizing a bit for the future rather than just building for now. We don't know if we will or won't. I'm fine either way but I'd generally caution against optimizing a function for something that we _think_ we might do as a philosophy.

pkg/providers/pricing/pricing.go
jonathan-innis: "If I do a get against pricing for each region without an update", do I get the data I expect from the price update map?

jonathan-innis: Advantage is that there's no reason to scope it. I don't think you're polluting the global context by doing this. This really _should_ be a global since it's just combining two other globals

pkg/providers/pricing/pricing.go
jmdeal: That's fair. With the change two the 3 AWS partitions vs the 2 unofficial partitions, I think this is quite a bit cleaner so I'll keep this for now.

pkg/providers/pricing/pricing.go
jmdeal: You sold me on lifting it to global scope, but idk if its really possible with go's package init order. Since globals in the same package are initialized in ascending alphabetical order by their file name, the generated files are always initialized last. We could continue initializing the var in `Reset`, but at that point I'd rather just keep it as a local.

hack/code/prices_gen.go
```diff
@@ -39,12 +39,50 @@ import (
...
+
+func NewOptions() *Options {
+	o := &Options{}
+	flag.StringVar(&o.partition, "partition", "aws", "The partition to generate prices for. Valid options are \"aws\", \"aws-us-gov\", and \"aws-cn\".")
+	flag.StringVar(&o.output, "output", "pkg/providers/pricing/zz_generated.pricing_us.go", "The destination for the generated go file.")
 	flag.Parse()
-	if flag.NArg() != 1 {
-		log.Fatalf("Usage: %s pkg/providers/pricing/zz_generated.pricing.go", os.Args[0])
+	if !lo.Contains([]string{"aws", "aws-us-gov", "aws-cn"}, o.partition) {
+		panic("invalid partition: must be \"aws\", \"aws-us-gov\", or \"aws-cn\"")
```
jonathan-innis, line 79:
```suggestion
		log.Fatal("invalid partition: must be \"aws\", \"aws-us-gov\", or \"aws-cn\"")
```

nit: prefer `log.Fatal` over `panic` for consistency and convention

hack/code/prices_gen.go
```diff
@@ -39,12 +39,50 @@ import (
...
+
+type Options struct {
+	partition string
+	output    string
+}
+
+func NewOptions() *Options {
+	o := &Options{}
+	flag.StringVar(&o.partition, "partition", "aws", "The partition to generate prices for. Valid options are \"aws\", \"aws-us-gov\", and \"aws-cn\".")
+	flag.StringVar(&o.output, "output", "pkg/providers/pricing/zz_generated.pricing_us.go", "The destination for the generated go file.")
```
jonathan-innis, line 76:
```suggestion
	flag.StringVar(&o.output, "output", "pkg/providers/pricing/zz_generated.pricing_aws.go", "The destination for the generated go file.")
```

nit: fix this default value

pkg/providers/pricing/zz_generated.pricing_aws_cn.go
```diff
@@ -0,0 +1,157 @@
...
+limitations under the License.
+*/
+
+package pricing
+
+import "time"
+
+// generated at 2023-09-18T13:06:44Z for cn-north-1
+
+var InitialPriceUpdateCN, _ = time.Parse(time.RFC3339, "2023-09-18T13:06:44Z")
```
jonathan-innis, line 23:
```suggestion
var InitialPriceUpdateTimeCN, _ = time.Parse(time.RFC3339, "2023-09-18T13:06:44Z")
```

pkg/providers/pricing/zz_generated.pricing_aws_us_gov.go
```diff
@@ -0,0 +1,385 @@
...
+limitations under the License.
+*/
+
+package pricing
+
+import "time"
+
+// generated at 2023-09-18T13:06:44Z for us-east-1
+
+var InitialPriceUpdateUSGov, _ = time.Parse(time.RFC3339, "2023-09-18T13:06:44Z")
```
jonathan-innis, line 23:
```suggestion
var InitialPriceUpdateTimeUSGov, _ = time.Parse(time.RFC3339, "2023-09-18T13:06:44Z")
```

pkg/providers/pricing/zz_generated.pricing_aws.go
```diff
@@ -0,0 +1,375 @@
...
+limitations under the License.
+*/
+
+package pricing
+
+import "time"
+
+// generated at 2023-09-18T13:06:44Z for us-east-1
+
+var InitialPriceUpdateAWS, _ = time.Parse(time.RFC3339, "2023-09-18T13:06:44Z")
```
jonathan-innis, line 23:
```suggestion
var InitialPriceUpdateTimeAWS, _ = time.Parse(time.RFC3339, "2023-09-18T13:06:44Z")
```

Better to be specific that this is a time

pkg/providers/pricing/pricing.go
```diff
@@ -415,6 +417,23 @@ func (p *Provider) Reset() {
 	p.onDemandPrices = staticPricing
 	// default our spot pricing to the same as the on-demand pricing until a price update
 	p.spotPrices = populateInitialSpotPricing(staticPricing)
-	p.onDemandUpdateTime = initialPriceUpdate
-	p.spotUpdateTime = initialPriceUpdate
+	p.onDemandUpdateTime = getInitialPriceUpdateByRegion(p.region)
+	p.spotUpdateTime = getInitialPriceUpdateByRegion(p.region)
+}
+
+func getInitialPriceUpdateByRegion(region string) time.Time {
```
jonathan-innis, line 424: I didn't really love this function and the more I looked at this, the more I was asking the question around why we need the update time in the first place for the generated file data. Why does it matter to us at all? It feels a little clunky to me that we would be using this for any of our logic or our logging if this is just static data that's being leveraged as a fallback.

Looking at pulling it out, it's actually unnecessary that we keep track of the update times. We have some logic that is built around it but those things seem to be easily change-able. I'd opt to just drop keeping track of these update times all together so we can drop this partition mapping that we are doing here.

jonathan-innis: This is also something that I respect to be done in a follow-on. I have a staged commit that I could show you where I dropped it and happy to convert that into a PR if you'd rather merge this as-is withe the drop coming afterwards.

jonathan-innis: Link to the commit of the proposed change: https://github.com/jonathan-innis/karpenter/commit/b3b08bbe019c3466c3905e4c9cd9e9ea06a7990a

jmdeal: Took a look at your commit, that looks good to me. I can go either way on putting it in this PR or a follow-up, though since you've already done the work a follow-up seems easiest.

jonathan-innis (approved):
LGTM 🚀

PR #4768, opened 2023-10-06
chore: Add handling for v1beta1 logging configuration migration

jonathan-innis:
Fixes #N/A 

**Description**

This change adds logging configuration values in the helm chart under the `logConfig` values block. You can now selectively disable the `logging-config` ConfigMap with `logConfig.enabled=false`. The global `logLevel` helm value now sets the `LOG_LEVEL` environment variable to configure the Karpenter controller log level. 

Also, with this change, combined with https://github.com/aws/karpenter-core/pull/553, Karpenter will now configure the logging ConfigMap to be mounted as a volumeMount in the container by default.

`make presubmit`
`/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter/templates/_helpers.tpl
+++ b/charts/karpenter/templates/_helpers.tpl
@@ -171,10 +171,17 @@ Flatten Settings Map using "." syntax
 {{/*
 Flatten the stdout logging outputs from args provided
 */}}
-{{- define "karpenter.controller.outputPathsList" -}}
+{{- define "karpenter.outputPathsList" -}}
 {{ $paths := list -}}
 {{- range .Values.controller.outputPaths -}}
-    {{- $paths = printf "%s" . | quote  | append $paths -}}
+    {{- if not (has (printf "%s" . | quote) $paths) -}}
+        {{- $paths = printf "%s" . | quote  | append $paths -}}
+    {{- end -}}
+{{- end -}}
+{{- range .Values.logConfig.outputPaths -}}
+    {{- if not (has (printf "%s" . | quote) $paths) -}}
+        {{- $paths = printf "%s" . | quote  | append $paths -}}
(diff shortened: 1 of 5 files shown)


jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

charts/karpenter/templates/_helpers.tpl
```diff
@@ -171,21 +171,35 @@ Flatten Settings Map using "." syntax
...
 Flatten the stdout logging outputs from args provided
 */}}
-{{- define "karpenter.controller.outputPathsList" -}}
+{{- define "karpenter.outputPathsList" -}}
 {{ $paths := list -}}
 {{- range .Values.controller.outputPaths -}}
-    {{- $paths = printf "%s" . | quote  | append $paths -}}
+    {{- if not (has (printf "%s" . | quote) $paths) -}}
+        {{- $paths = printf "%s" . | quote  | append $paths -}}
+    {{- end -}}
```
njtran, line 179: QQ: hoping you can help educate me here. Don't we want to include these paths if they're defined? It looks like the code is saying "if not has string defined" then add it in, which I don't think makes sense. What am I missing here?

njtran (approved):
approved with one question

charts/karpenter/templates/_helpers.tpl
jonathan-innis: This is saying: "If this string doesn't already exist in the set of paths, then add it." It's uniqueness criteria.

njtran: Ahhhh gotcha. Thanks!

njtran (approved):

PR #4789, opened 2023-10-10
chore: karpenter-global-settings migration

jmdeal:
Fixes #N/A 

**Description**
This PR is the first step in removing the `karpenter-global-settings` configmap as part of v1beta1. It adds the additional fields to the cli flags / env vars, but does not remove the configmap. Currently both are supported, but values set by cli flags / env vars will take precedent.

`karpenter snapshot` + manual testing

- [ ] Yes, issue opened: # 
- [x] No


jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

pkg/providers/launchtemplate/nodeclass_test.go
```diff
@@ -742,7 +749,10 @@ var _ = Describe("EC2NodeClass/LaunchTemplates", func() {
 		It("should calculate memory overhead based on eni limited pods when ENI limited", func() {
 			ctx = settings.ToContext(ctx, test.Settings(test.SettingOptions{
```
jonathan-innis, line 750: Do we need to explicitly set this in the settings if this is the default?

pkg/providers/launchtemplate/nodetemplate_test.go
```diff
@@ -707,7 +708,10 @@ var _ = Describe("NodeTemplate/LaunchTemplates", func() {
 		It("should calculate memory overhead based on eni limited pods when ENI limited", func() {
 			ctx = settings.ToContext(ctx, test.Settings(test.SettingOptions{
```
jonathan-innis, line 709: Should this be set to false here? If so, do we need a TODO to note that we should drop this when it's no longer supported?

pkg/providers/launchtemplate/nodetemplate_test.go
```diff
@@ -718,7 +722,10 @@ var _ = Describe("NodeTemplate/LaunchTemplates", func() {
 		It("should calculate memory overhead based on eni limited pods when not ENI limited", func() {
 			ctx = settings.ToContext(ctx, test.Settings(test.SettingOptions{
```
jonathan-innis, line 723: Same comments here and everywhere else we explicitly set the settings to false.

pkg/providers/instancetype/nodeclass_test.go
```diff
@@ -645,16 +646,8 @@ var _ = Describe("NodeClass/InstanceTypes", func() {
...
-					"aws.clusterName": "karpenter-cluster",
-				},
-			})
-			Expect(err).To(BeNil())
-
-			s := settings.FromContext(ctx)
-			ctx = settings.ToContext(ctx, test.Settings(test.SettingOptions{
-				VMMemoryOverheadPercent: &s.VMMemoryOverheadPercent,
+			ctx = options.ToContext(ctx, test.Options(test.OptionsFields{
+				ClusterName: lo.ToPtr("CLUSTER_NAME"),
```
jonathan-innis, line 650: Just curious: any specific reason to change this value?

pkg/providers/instancetype/types.go
```diff
@@ -240,7 +241,7 @@ func ephemeralStorage(amiFamily amifamily.AMIFamily, blockDeviceMappings []*v1be
 func awsPodENI(ctx context.Context, name string) *resource.Quantity {
 	// https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html#supported-instance-types
 	limits, ok := Limits[name]
-	if awssettings.FromContext(ctx).EnablePodENI && ok && limits.IsTrunkingCompatible {
+	if settings.FromContext(ctx).EnablePodENI && ok && limits.IsTrunkingCompatible {
```
jonathan-innis, line 244: Is this always guaranteed to be set even if the settings ConfigMap doesn't exist>?

test/suites/chaos/suite_test.go
```diff
@@ -69,8 +69,8 @@ var _ = Describe("Chaos", func() {
 			defer cancel()
 
 			provider := awstest.AWSNodeTemplate(v1alpha1.AWSNodeTemplateSpec{AWS: v1alpha1.AWS{
-				SecurityGroupSelector: map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName},
-				SubnetSelector:        map[string]string{"karpenter.sh/discovery": settings.FromContext(env.Context).ClusterName},
+				SecurityGroupSelector: map[string]string{"karpenter.sh/discovery": lo.Must(os.LookupEnv("CLUSTER_NAME"))},
```
jonathan-innis, line 72: Is there a way to lookup this value once in the environment and then refer to it that way rather than having to continually do an `os.LookupEnv`?

test/suites/integration/extended_resources_test.go
```diff
@@ -195,8 +194,8 @@ var _ = Describe("Extended Resources", func() {
 				},
 			},
 		})
-		provider.Spec.UserData = lo.ToPtr(fmt.Sprintf(string(rawContent), settings.FromContext(env.Context).ClusterName,
-			settings.FromContext(env.Context).ClusterEndpoint, env.ExpectCABundle(), provisioner.Name))
+		provider.Spec.UserData = lo.ToPtr(fmt.Sprintf(string(rawContent), lo.Must(os.LookupEnv("CLUSTER_NAME")),
+			lo.Must(os.LookupEnv("CLUSTER_ENDPOINT")), env.ExpectCABundle(), provisioner.Name))
```
jonathan-innis, line 198: Same comment here on the cluster endpoint and storing it in the environment?

hack/code/prices_gen.go
```diff
@@ -34,7 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
-	"github.com/aws/karpenter/pkg/apis/settings"
+	awsoptions "github.com/aws/karpenter/pkg/operator/options"
```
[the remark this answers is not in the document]
jmdeal: Not sure, it can be dropped.

pkg/operator/options/options.go
```diff
@@ -0,0 +1,141 @@
...
+	fs.Float64Var(&o.VMMemoryOverheadPercent, "vm-memory-overhead-percent", env.WithDefaultFloat64("VM_MEMORY_OVERHEAD_PERCENT", 0.075), "The VM memory overhead as a percent that will be subtracted from the total memory for all instance types.")
+	fs.StringVar(&o.InterruptionQueueName, "interruption-queue-name", env.WithDefaultString("INTERRUPTION_QUEUE_NAME", ""), "Interruption queue is disabled if not specified. Enabling interruption handling may require additional permissions on the controller service account. Additional permissions are outlined in the docs.")
+	fs.IntVar(&o.ReservedENIs, "reserved-enis", env.WithDefaultInt("RESERVED_ENIS", 0), "Reserved ENIs are not included in the calculations for max-pods or kube-reserved. This is most often used in the VPC CNI custom networking setup https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html.")
+}
+
+func (o *Options) Parse(fs *flag.FlagSet, args ...string) error {
+	if err := fs.Parse(args); err != nil {
+		if errors.Is(err, flag.ErrHelp) {
+			os.Exit(0)
```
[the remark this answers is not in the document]
jmdeal: The idea is that `flag` prints a help message and the program immediately exists without an error code. This isn't really an error, its just how `flag` handles someone passing the `--help` flag. I think this is the correct behavior.

test/suites/chaos/suite_test.go
jmdeal: A couple of options there:
1) We could introduce a function along the lines of `env.GetClusterName()`
2) We could do something similar to what was done before, inject `options` with `ClusterName` populated.
I'd lean toward the first option as it doesn't couple this to Karpenter's cli flags.

pkg/operator/options/options_validation.go
```diff
@@ -0,0 +1,78 @@
...
+		o.validateEndpoint(),
+		o.validateClusterName(),
+		o.validateVMMemoryOverheadPercent(),
+		o.validateAssumeRoleDuration(),
+	).ViaField("aws")
+}
+
+func (o Options) validateAssumeRoleDuration() (errs *apis.FieldError) {
+	// TODO: Remove with karpenter-global-settings
+	if !o.setFlags["aws-assume-role-arn"] {
```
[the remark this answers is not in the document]
jmdeal: You're right, it doesn't need to be on all of them, only `ClusterName` needs that validation.

pkg/providers/instancetype/nodeclass_test.go
jmdeal: Miss on my part, idk if it matters but I'll change it back to karpenter-cluster.

pkg/providers/launchtemplate/nodeclass_test.go
jmdeal: We don't need to, but one could argue it makes the test more readable. I could go either way on dropping it.

pkg/providers/launchtemplate/nodetemplate_test.go
jmdeal: Do these tests need to be modified to match v1beta1 behavior or will v1beta1 always behave as though this option is set to true? If it's the latter, I think we keep the test while we still support the configmap and drop the tests later.

pkg/providers/instancetype/types.go
jmdeal: Nope, discussed offline that this should be modified to return the default value in the case that it doesn't exist.

pkg/providers/instancetype/types.go
```diff
@@ -403,7 +404,7 @@ func pods(ctx context.Context, info *ec2.InstanceTypeInfo, amiFamily amifamily.A
 	switch {
 	case kc != nil && kc.MaxPods != nil:
 		count = int64(ptr.Int32Value(kc.MaxPods))
-	case awssettings.FromContext(ctx).EnableENILimitedPodDensity && amiFamily.FeatureFlags().SupportsENILimitedPodDensity:
+	case settings.FromContext(ctx).EnableENILimitedPodDensity && amiFamily.FeatureFlags().SupportsENILimitedPodDensity:
```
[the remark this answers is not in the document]
jmdeal: Resolved, settings always exists in context. It will have its default value if the configmap doesn't exist.

hack/code/prices_gen.go
```diff
@@ -93,7 +93,7 @@ func main() {
 	os.Setenv("AWS_SDK_LOAD_CONFIG", "true")
 	os.Setenv("AWS_REGION", region)
 	ctx := context.Background()
-	ctx = settings.ToContext(ctx, test.Settings())
+	ctx = options.ToContext(ctx, test.Options())
```
jonathan-innis, line 96: Ah, I see you why you renamed this the first time now. This isn't blocking and I'm fine either way that you choose this, but I'd personally prefer that we keep the struct and function `Options` and `NewOptions`

pkg/apis/settings/settings.go
```diff
@@ -92,15 +97,18 @@ func (*Settings) Inject(ctx context.Context, cm *v1.ConfigMap) (context.Context,
...
+}
+
 func ToContext(ctx context.Context, s *Settings) context.Context {
 	return context.WithValue(ctx, ContextKey, s)
 }
 
 func FromContext(ctx context.Context) *Settings {
 	data := ctx.Value(ContextKey)
 	if data == nil {
-		// This is developer error if this happens, so we should panic
```
jonathan-innis, line 102: Why change this?

pkg/operator/options/options_validation.go
```diff
@@ -0,0 +1,67 @@
...
+func (o Options) validateAssumeRoleDuration() (errs *apis.FieldError) {
+	if o.AssumeRoleDuration < time.Minute*15 {
+		return errs.Also(apis.ErrInvalidValue("assume-role-duration cannot be less than 15 Minutes", "assume-role-duration"))
+	}
+	return nil
+}
+
+func (o Options) validateClusterName() (errs *apis.FieldError) {
+	// TODO: Remove with karpenter-global-settings
+	if !o.setFlags["aws-cluster-name"] {
```
jonathan-innis, line 43:
```suggestion
	if !o.setFlags["cluster-name"] {
```

This is the wrong setFlag key, right?

pkg/operator/options/options_validation.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"fmt"
+	"net/url"
+	"time"
+
+	"knative.dev/pkg/apis"
+)
+
+func (o Options) Validate() (errs *apis.FieldError) {
```
jonathan-innis, line 25: Minor thing, but we could probably replace the error type from `apis.FieldError` into just a standard error and merge the errors together with `mutlierr.Combine` and `multierr.Append`

pkg/operator/options/options.go
```diff
@@ -0,0 +1,137 @@
...
+	mergeField(&o.ClusterCABundle, s.ClusterCABundle, o.setFlags["cluster-ca-bundle"])
+	mergeField(&o.ClusterName, s.ClusterName, o.setFlags["cluster-name"])
+	mergeField(&o.ClusterEndpoint, s.ClusterEndpoint, o.setFlags["cluster-endpoint"])
+	mergeField(&o.IsolatedVPC, s.IsolatedVPC, o.setFlags["isolated-vpc"])
+	mergeField(&o.VMMemoryOverheadPercent, s.VMMemoryOverheadPercent, o.setFlags["vm-memory-overhead-percent"])
+	mergeField(&o.InterruptionQueueName, s.InterruptionQueueName, o.setFlags["interruption-queue-name"])
+	mergeField(&o.ReservedENIs, s.ReservedENIs, o.setFlags["reserved-enis"])
+
+	// Temporarily revalidate ClusterName, since default value is invalid
+	o.setFlags["cluster-name"] = true
```
jonathan-innis, line 114: Just to confirm? This is to validate that required fields are set? Should we create a function like `validateRequiredFields()`?

test/pkg/environment/aws/environment.go
```diff
@@ -90,3 +91,15 @@ func GetTimeStreamAPI(session *session.Session) timestreamwriteiface.TimestreamW
 	}
 	return &NoOpTimeStreamAPI{}
 }
+
+func GetClusterName() string {
+	return lo.Must(os.LookupEnv("CLUSTER_NAME"))
```
jonathan-innis, line 96: Why do we need to re-get this value every time? Why not get it at the beginning of the test suite during the setup in the `test/pkg/environment/aws` Environment struct and then store it in the environment so users can get these values with `env.ClusterName` or `env.ClusterEndpoint`

hack/code/prices_gen.go
jmdeal: This didn't actually end up being the issue, I just left it after I changed it since it seemed more descriptive anyway. I'm perfectly happy to change it back.

pkg/apis/settings/settings.go
jmdeal: I think its hidden since I've squashed a lot of commits into 1 but this was `return nil` for most of the lifetime of this PR. I'll restore the original panic message / comment.

test/pkg/environment/aws/environment.go
jmdeal: There's not really any reason not to, I can go ahead and change that.

hack/code/prices_gen.go
jmdeal: So a problem is multiple files in `hack/code` have an `Options` struct and they're all in the main package so we have name conflicts. I'm not sure what a good solution here is other than:
a) Separate out hack files into separate folders
b) Add qualifiers the names (e.g. PriceGenOptions, VPCLimitsOptions)

pkg/operator/options/options.go
jmdeal: That is what its doing, but I'm not sure it's necessary to separate it out into its own function. It's only necessary to do here for cluster name and this additional check won't be required once the configmap is removed (handled by `options.Validate` as part of `Parse`).

pkg/operator/options/options.go
```diff
@@ -0,0 +1,142 @@
...
+	return nil
+}
+
+func (o *Options) ToContext(ctx context.Context) context.Context {
+	return ToContext(ctx, o)
+}
+
+func (o *Options) MergeSettings(ctx context.Context) {
+	s := settings.FromContext(ctx)
+	if s == nil {
```
jonathan-innis, line 99: Will this ever be the case now that we are always injecting this value?

pkg/operator/options/options.go
```diff
@@ -0,0 +1,142 @@
...
+// Note: Separated out to help with cyclomatic complexity check
+func mergeField[T any](dest *T, src T, isDestSet bool) {
+	if !isDestSet {
+		*dest = src
+	}
+}
+
+func (o *Options) checkRequiredFields() error {
+	if o.ClusterName == "" {
+		return fmt.Errorf("missing field, cluster-name")
```
jonathan-innis, line 139:
```suggestion
		return fmt.Errorf("missing required field, cluster-name")
```

pkg/operator/options/options_validation.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	return nil
+}
+
+func (o Options) validateClusterName() error {
+	// TODO: Remove with karpenter-global-settings
+	if !o.setFlags["cluster-name"] {
+		return nil
+	}
+	if o.ClusterName == "" {
```
jonathan-innis, line 46: We shouldn't fail here, right? Since it can be set in either place?

test/pkg/environment/common/environment.go
```diff
@@ -58,6 +58,9 @@ type Environment struct {
 	Monitor    *Monitor
 
 	StartingNodeCount int
+
+	ClusterName     string
```
jonathan-innis, line 62: These are AWS-specific and may not be part of the common environment practically, we should move this over to the AWS environment

jmdeal: Sounds good, I honestly wasn't sure where they should live 👍

pkg/operator/options/options_validation.go
jmdeal: That's why we return `nil` if the cluster name wasn't set. The only case it would fail here is if someone explicitly set cluster name to `""`. If they don't specify it as a cli flag or env var it will fall back to settings.

pkg/operator/options/options.go
jmdeal: Good point, it won't be

pkg/operator/options/options.go
jmdeal: It's wrapped further up by "checking required fields, %w" so I thought that might be redundant. Thoughts?

pkg/operator/options/options_validation.go
jmdeal: Discussed offline, while this behavior was correct it didn't need to happen both pre and post merge. For now required field checks will only take place post merge with settings.

pkg/operator/options/options.go
```diff
@@ -0,0 +1,134 @@
...
+		_, ok := os.LookupEnv(envName)
+		o.setFlags[f.Name] = ok || cliFlags.Has(f.Name)
+	})
+
+	if err := o.Validate(); err != nil {
+		return fmt.Errorf("validating options, %w", err)
+	}
+
+	// Parse options which can't be directly parsed by flag
+	o.IsolatedVPC = (o.isolatedVPC == "true")
```
jonathan-innis, line 92: Can you explain this? Or link to a file or description of this behavior? This isn't going to be immediately evident to anyone reading it, so it should have some link to documentation if this is needed for the behavior.

pkg/operator/options/options_validation.go
```diff
@@ -0,0 +1,82 @@
...
+func (o Options) validateAssumeRoleDuration() error {
+	if o.AssumeRoleDuration < time.Minute*15 {
+		return fmt.Errorf("assume-role-duration cannot be less than 15 minutes")
+	}
+	return nil
+}
+
+func (o Options) validateIsolateVPC() error {
+	if o.isolatedVPC != "true" && o.isolatedVPC != "false" {
+		return fmt.Errorf("%q is not a valid value for isolate-vpc, options are true or false", o.isolatedVPC)
```
jonathan-innis, line 44:
```suggestion
		return fmt.Errorf("%q is not a valid value for isolated-vpc, options are true or false", o.isolatedVPC)
```

pkg/operator/options/options_validation.go
```diff
@@ -0,0 +1,82 @@
...
+	// sure it's a real URL
+	if err != nil || !endpoint.IsAbs() || endpoint.Hostname() == "" {
+		return fmt.Errorf("%q not a valid cluster-endpoint URL", o.ClusterEndpoint)
+	}
+	return nil
+}
+
+func (o Options) validateVMMemoryOverheadPercent() error {
+	if o.VMMemoryOverheadPercent < 0 {
+		return fmt.Errorf("vm-memory-overhead-percent may not be negative")
```
jonathan-innis, line 64:
```suggestion
		return fmt.Errorf("vm-memory-overhead-percent cannot be negative")
```

pkg/operator/options/options_validation.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return nil
+}
+
+func (o Options) validateReservedENIs() error {
+	if o.ReservedENIs < 0 {
+		return fmt.Errorf("reserved-enis may not be negative")
```
jonathan-innis, line 71:
```suggestion
		return fmt.Errorf("reserved-enis cannot be negative")
```

pkg/operator/options/options_validation.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (o Options) validateEndpoint() error {
+	if o.ClusterEndpoint == "" {
+		return nil
+	}
+	endpoint, err := url.Parse(o.ClusterEndpoint)
+	// url.Parse() will accept a lot of input without error; make
+	// sure it's a real URL
+	if err != nil || !endpoint.IsAbs() || endpoint.Hostname() == "" {
+		return fmt.Errorf("%q not a valid cluster-endpoint URL", o.ClusterEndpoint)
```
jonathan-innis, line 57:
```suggestion
		return fmt.Errorf("%q is not a valid cluster-endpoint URL", o.ClusterEndpoint)
```

pkg/operator/options/suite_test.go
```diff
@@ -0,0 +1,118 @@
...
+		opts = &options.Options{}
+		opts.AddFlags(fs)
+
+		// Inject default settings
+		var err error
+		ctx, err = (&settings.Settings{}).Inject(ctx, nil)
+		Expect(err).To(BeNil())
+	})
+
+	Context("Merging", func() {
```
jonathan-innis, line 53: Does it make sense to have testing to make sure that things succeed when only the settings are set OR only the options are set? I could see some edge-cases that might crop up there based on how the logic is written.

pkg/operator/options/options.go
jmdeal: Do you mean why this can't be parsed via BoolVar and it has to be parsed as a string and then converted? The go docs didn't really have a good description of it, I can see if I can find one but worst case I found some examples.

pkg/operator/options/suite_test.go
jmdeal: Ya we probably should, that's easy enough to implement.

pkg/operator/options/options.go
jmdeal: I think I got caught up by `BoolVar` being the only one of the `TypeVar` functions on `flag` that doesn't take an argument (without calling it out), but I guess `--isolated-vpc` probably shouldn't take an argument (i.e. `--isolated-vpc` instead of `--isolated-vpc true`.

jmdeal: A counterpoint to this would be that there is no longer an explicit way to set `isolated-vpc` to false via the CLI flags. This isn't an issue if we consider environment variables and cli flags as mutually exclusive (a user will only specify one and not the other), but otherwise it makes it impossible to explicitly set `isolated-vpc` to false when `ISOLATED_VPC=true` exists as an env var. I'm honestly not sure if this is contrived / a non-issue or if it's worth considering.

jmdeal: After sleeping on it I think the correct solution is to use a `StringVar` for `IsolatedVPC` so we can explicitly set it to true or false via the CLI flags. I think the consistent merge semantics with environment variables outweigh breaking consistency with how bools are normally handled with flags.

jmdeal: This is being addressed by a wrapper to `FlagSet` implemented in karpenter-core.

pkg/operator/options/suite_test.go
```diff
@@ -0,0 +1,265 @@
...
+				ClusterCABundle:         "settings-bundle",
+				ClusterName:             "settings-cluster",
+				ClusterEndpoint:         "https://settings-cluster",
+				IsolatedVPC:             true,
+				VMMemoryOverheadPercent: 0.05,
+				InterruptionQueueName:   "settings-cluster",
+				ReservedENIs:            8,
+			})
+			opts.MergeSettings(ctx)
+			expectOptionsEqual(opts, test.Options(test.OptionsFields{
```
jonathan-innis, line 118: Should this have been a public `ExpectOptionsEqual` function since we are calling it both in `aws/karpenter-core` and `aws/karpenter`

test/pkg/environment/aws/environment.go
```diff
@@ -80,6 +85,10 @@ func NewEnvironment(t *testing.T) *Environment {
 		EKSAPI:        eks.New(session),
 		SQSProvider:   interruption.NewSQSProvider(sqs.New(session)),
 		TimeStreamAPI: GetTimeStreamAPI(session),
+
+		ClusterName:           lo.Must(os.LookupEnv("CLUSTER_NAME")),
+		ClusterEndpoint:       lo.Must(os.LookupEnv("CLUSTER_ENDPOINT")),
+		InterruptionQueueName: lo.Must(os.LookupEnv("INTERRUPTION_QUEUE_NAME")),
```
jonathan-innis, line 91:
```suggestion
		InterruptionQueueName: lo.Must(os.LookupEnv("INTERRUPTION_QUEUE")),
```

pkg/operator/options/suite_test.go
jmdeal: The two structs they're comparing are different, and unless I dive into reflection I don't think it can be made generic.

jonathan-innis (dismissed):
LGTM 🚀

jmdeal (commented):
/karpenter snapshot

jonathan-innis (dismissed):
LGTM 🚀

jmdeal (commented):
/karpenter snapshot

jonathan-innis (approved):
LGTM 🚀

PR #4834, opened 2023-10-14
ci: Fix sweeper bug with instance profiles

jonathan-innis:
Fixes #N/A 

**Description**

Fix a bug where we were comparing a date value that could set an instance profile that was created at `23:59:59` to be created at `00:00:00` due to it only capturing the date value with the expiration time. If we got unlucky and the instance profile was created late in the day, the instance profile could get deleted while the tests were running.

- [ ] Yes, issue opened: # 
- [x] No


engedaam (approved):
LGTM 🚀

PR #4865, opened 2023-10-20
docs: Fix region being undefined while applying beta policy

jonathan-innis:
Fixes #N/A 

**Description**

Some users did not define `$AWS_REGION` when adding the beta controller policy to their existing controller role. This change ensures that the `$AWS_REGION` is set be running `set -euo pipefail` at the beginning of the script. On top of this, the `envsubst` that is injecting environment variables into the controller policy will now fallback to the `$AWS_DEFAULT_REGION`

- Applying the beta controller policy to the controller role

- [ ] Yes, issue opened: # 
- [ ] No


engedaam (approved):
LGTM 🚀

PR #4869, opened 2023-10-20
chore: touch up v1beta1 examples

ellistarn:
Fixes #N/A 

**Description**
* Tags were overindented
* Shifted "disruption" forward in the example, since it's the focal point (it's also how the keys will be ordered from kubectl get)
* Shifted "nodeClassRef" forward to make it crystal clear to users that there's another entity in play. This is a question/misunderstanding I get a lot. (it's also how the keys will be ordered from kubectl get)

- [ ] Yes, issue opened: # 
- [] No

### DIFF
--- a/examples/v1beta1/100-cpu-limit.yaml
+++ b/examples/v1beta1/100-cpu-limit.yaml
@@ -11,6 +11,8 @@ metadata:
 spec:
   template:
     spec:
+      nodeClassRef:
+        name: default
       requirements:
         - key: kubernetes.io/arch
           operator: In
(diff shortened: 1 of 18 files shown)


examples/v1beta1/100-cpu-limit.yaml
```diff
@@ -43,7 +43,7 @@ spec:
   role: "KarpenterNodeRole-${CLUSTER_NAME}" # replace with your cluster name
   subnetSelectorTerms:
     - tags:
```
jonathan-innis, line 45: Tags aren't over-indented. this looks incorrect to me

PR #4921, opened 2023-10-25
BREAKING: Fix helm ordering of values and set default log encoding to `json`

jonathan-innis:
Fixes #N/A 

**Description**

This change updates the ordering of helm values for evaluation in the chart to ensure that the old values are always considered first. This ensures that the defaults that are specified in the default `values.yaml` file do not override any existing values were set by the user on a previous version.

Note that this PR also contains a breaking change to defaults for log encoding. If you did not previously set the log encoding for Karpenter and relied on the default log encoding of `console`, this will now be changed in the helm chart to `json`, unless this value is updated by specifying the `logConfig.logEncoding` value like `--set logConfig.logEncoding=console`

- [ ] Yes, issue opened: # 
- [x] No


jonathan-innis (commented):
/karpenter snapshot

njtran (approved):

Issue #4923, opened 2023-10-25
Launches spot instance types when on-demand instance type is cheaper for mixed capacity types


rkirk-r7:
### Description

**Observed Behavior**: Launches spot instance types when on-demand instance type is cheaper for mixed capacity types provisioner

**Expected Behavior**: Should launch on demand instance type if cheaper than spot instance types for mixed capacity types provisioner.  
When discounts are applied to ec2 pricing we find that on demand can often be cheaper than spot pricing for the same instance type.

**Reproduction Steps** (Please include YAML):
I don't have the ability to run in my payee account so have made the following change to affect the pricing to show the issue to force on demand pricing to be seen as cheaper than spot.
```
diff --git a/pkg/providers/pricing/pricing.go b/pkg/providers/pricing/pricing.go
index 77795975..78aa9bbb 100644
--- a/pkg/providers/pricing/pricing.go
+++ b/pkg/providers/pricing/pricing.go
@@ -299,7 +299,8 @@ func (p *Provider) onDemandPage(prices map[string]float64) func(output *pricing.
 					if err != nil || price == 0 {
 						continue
 					}
-					prices[pItem.Product.Attributes.InstanceType] = price
+					prices[pItem.Product.Attributes.InstanceType] = price * 0.5
+					logging.FromContext(context.Background()).Infof("rkirk ondemand %s - %f", pItem.Product.Attributes.InstanceType, prices[pItem.Product.Attributes.InstanceType])
 				}
 			}
 		}
@@ -334,7 +335,8 @@ func (p *Provider) UpdateSpotPricing(ctx context.Context) error {
 			if !ok {
 				prices[instanceType] = map[string]float64{}
 			}
-			prices[instanceType][az] = spotPrice
+			prices[instanceType][az] = spotPrice * 5.0
+			logging.FromContext(ctx).Infof("rkirk spot %s - %f", instanceType, prices[instanceType][az])
 			InstancePriceEstimate.With(prometheus.Labels{
 				InstanceTypeLabel: instanceType,
 				CapacityTypeLabel: ec2.UsageClassTypeSpot,
```

This results in the following logging showing a spot instance is launched even though the on demand price for the selected instance type is cheaper
```
INFO    controller.pricing      rkirk spot m4.10xlarge - 4.960000       {"commit": "895504c-dirty"}
INFO    controller.pricing      rkirk spot m4.10xlarge - 4.740500       {"commit": "895504c-dirty"}
[...]

PR #4924, opened 2023-10-25
fix: Allow on demand instance type to be launched if cheaper than spot price

rkirk-r7:
Fixes 4923

https://github.com/aws/karpenter/issues/4923

**Description**

Ensure the cheapest instance type is launched for mixed capacity type configurations, if spot is cheapest it should be launched or if on demand is cheapest it should be launched. Currently for mixed capacity type provisioners regardless of price spot instances will be launched if they are available to be launched, on demand instances would only be launched if there was no availability for spot for the selected instance type(s)

This change modifies the pricing api to return an error when the spot price for an instance type/az is more expensive than the on demand price for the same instance type.  This means these spot instance types will not be considered available to launch, but the on demand instance type can still be launched if it is the most appropriate.   

- [x] Added unit tests for change to pricing lookup
- [x] Added unit tests to ensure correct instance type is launched when on demand is cheaper than spot pricing based on aws pricing api responses
- [x] Extended existing unit tests to ensure correct instance type is launched when on demand is more expensive than spot pricing based on aws pricing api responses
- [x] Manual testing

For manual testing I scaled up a deployment to force a new ec2 node to be launched by karpenter and check the correct capacity type is launched

```
2023-10-25T09:51:41.140Z        INFO    controller.pricing      rkirk spot m4.10xlarge - 4.960000       {"commit": "895504c-dirty"}
2023-10-25T09:51:41.145Z        INFO    controller.pricing      rkirk spot m4.10xlarge - 4.740500       {"commit": "895504c-dirty"}
2023-10-25T09:51:41.349Z        INFO    controller.pricing      rkirk spot m4.10xlarge - 4.714000       {"commit": "895504c-dirty"}
2023-10-25T09:51:41.350Z        INFO    controller.pricing      rkirk spot m4.10xlarge - 4.962500       {"commit": "895504c-dirty"}
2023-10-25T09:51:41.546Z        INFO    controller.pricing      rkirk spot m4.10xlarge - 4.434500       {"commit": "895504c-dirty"}
2023-10-25T09:51:41.549Z        INFO    controller.pricing      rkirk spot m4.10xlarge - 4.245000       {"commit": "895504c-dirty"}
{"level":"info","ts":1698227502.2016387,"logger":"fallback","caller":"pricing/pricing.go:303","msg":"rkirk ondemand m4.10xlarge - 1.000000"}
2023-10-25T10:16:53.176Z        INFO    controller.provisioner  found provisionable pod(s)      {"commit": "895504c-dirty", "pods": "apps/curl-deployment-f467cbbd7-qd7qf, apps/curl-deployment-f467cbbd7-wjh95, apps/curl-deployment-f467cbbd7-97kdb, apps/curl-deployment-f467cbbd7-58lkf, apps/curl-deployment-f467cbbd7-bst74 and 13 other(s)", "duration": "183.677533ms"}
2023-10-25T10:16:53.176Z        INFO    controller.provisioner  computed new machine(s) to fit pod(s)   {"commit": "895504c-dirty", "machines": 1, "pods": 18}
2023-10-25T10:16:53.190Z        INFO    controller.provisioner  created machine {"commit": "895504c-dirty", "provisioner": "karpenter-on-demand", "machine": "karpenter-on-demand-qxr79", "requests": {"cpu":"39055m","memory":"144536680Ki","pods":"27"}, "instance-types": "c5.24xlarge, c5.metal, c5a.24xlarge, c5ad.24xlarge, c5d.24xlarge and 95 other(s)"}
2023-10-25T10:16:55.893Z        INFO    controller.machine.lifecycle    launched machine        {"commit": "895504c-dirty", "machine": "karpenter-on-demand-qxr79", "provisioner": "karpenter-on-demand", "provider-id": "aws:///us-east-1b/i-00e9d1b2657485a92", "instance-type": "m4.10xlarge", "zone": "us-east-1b", "capacity-type": "on-demand", "allocatable": {"cpu":"39830m","ephemeral-storage":"52926258176","memory":"148623Mi","pods":"234"}}
```

The following modification was made for the manual testing as I cannot run this in my payee account where on demand is typically cheaper than spot pricing due to discounts that are applied.
```
diff --git a/pkg/providers/pricing/pricing.go b/pkg/providers/pricing/pricing.go
index 77795975..78aa9bbb 100644
--- a/pkg/providers/pricing/pricing.go
+++ b/pkg/providers/pricing/pricing.go
@@ -299,7 +299,8 @@ func (p *Provider) onDemandPage(prices map[string]float64) func(output *pricing.
 					if err != nil || price == 0 {
 						continue
 					}
-					prices[pItem.Product.Attributes.InstanceType] = price
+					prices[pItem.Product.Attributes.InstanceType] = price * 0.5
+					logging.FromContext(context.Background()).Infof("rkirk ondemand %s - %f", pItem.Product.Attributes.InstanceType, prices[pItem.Product.Attributes.InstanceType])
 				}
 			}
 		}
@@ -334,7 +335,8 @@ func (p *Provider) UpdateSpotPricing(ctx context.Context) error {
 			if !ok {
 				prices[instanceType] = map[string]float64{}
 			}
-			prices[instanceType][az] = spotPrice
+			prices[instanceType][az] = spotPrice * 5.0
+			logging.FromContext(ctx).Infof("rkirk spot %s - %f", instanceType, prices[instanceType][az])
 			InstancePriceEstimate.With(prometheus.Labels{
 				InstanceTypeLabel: instanceType,
 				CapacityTypeLabel: ec2.UsageClassTypeSpot,

```

- [ ] Yes, issue opened: # 
- [x] No


mgrundie-r7 (approved):

pkg/providers/pricing/pricing.go
```diff
@@ -135,7 +136,9 @@ func (p *Provider) SpotPrice(instanceType string, zone string) (float64, bool) {
 			return val.defaultPrice, true
 		}
 		if price, ok := p.spotPrices[instanceType].prices[zone]; ok {
-			return price, true
+			if onDemandPrice, onDemandOk := p.onDemandPrices[instanceType]; (onDemandOk && price <= onDemandPrice) || !onDemandOk {
```
njtran, line 139: As I understand, this, in the current code, this should always return true, since the onDemand price will always be available if the spot price is available, and with no discounts, this is the guarantee of spot vs on-demand (as there'd be no benefit to spot if on-demand was cheaper).

njtran (commented):
Hey @rkirk-r7, sorry for the delay here. 

I understand you've tested this with your code to modify the Spot prices to be 5x more expensive, and On-demand prices 2x cheaper, essentially making every spot price more expensive than its OD counterpart, unless the spot discount is > 90%. 

Is the intent of this PR to be merged after your previous one? If that PR isn't merged, which I believe we discussed some alternatives in WG, then I don't think we should be merging this one. It adds more functionality that is an invariant with spot/OD pricing. 

What are your thoughts?

pkg/providers/pricing/pricing.go
rkirk-r7: Yes it would always be true in the case with no discounts to lower the on-demand pricing

rkirk-r7:
> Hey @rkirk-r7, sorry for the delay here.
> 
> I understand you've tested this with your code to modify the Spot prices to be 5x more expensive, and On-demand prices 2x cheaper, essentially making every spot price more expensive than its OD counterpart, unless the spot discount is > 90%.

Yes this was just a way to be able to test and change the pricing, it's not reflection of the discounted price we pay.

> Is the intent of this PR to be merged after your previous one? If that PR isn't merged, which I believe we discussed some alternatives in WG, then I don't think we should be merging this one. It adds more functionality that is an invariant with spot/OD pricing.
> 
> What are your thoughts?

The intent was this would be safe to merge regardless of the price multiplier PR and would not change the behaviour for the non-discounted pricing where spot would be cheaper than on demand.  Whatever approach we take regarding getting more accurate pricing/discount aware pricing be it the multiplier, per instancetype config, assumable IAM role, a.n.other approach then I believe this change would be required.  Whilst the price filtering handles determining the correct pricing between spot and on demand types, the launching code later does not check the price and if the filtered instancetypes have spot Offerings then it will default to launching spot regardless of price.  This change is intended to filter out the spot Offerings which are more expensive than the on demand price so we can launch spot where it is cheapest but fallback to ondemand when it is cheapest.

As an example, if this was the following was the output of filtering the instancetypes by price
```
{
    "instancetypes": [
        {
            "instancetype": "m6a.xl",
            "offerings": [
                {
                    "capacity-type": "spot",
                    "price": 2.0
                },
                {
                    "capacity-type": "on-demand",
                    "price": 1.0
                }
            ]
        }
    ]
}
```
Then the current code would launch spot because there is a spot offering for the filtered instancetype(s) without rechecking the pricing.

This change would end up with the filtered list looking like
```
{
    "instancetypes": [
        {
            "instancetype": "m6a.xl",
            "offerings": [
                {
                    "capacity-type": "on-demand",
                    "price": 1.0
                }
            ]
        }
    ]
}
```
Which would lead to on-demand type being launched.

Another example would be if the discount means spot is sometimes cheaper, but sometimes on-demand is, so this type of scenario
```
{
    "instancetypes": [
        {
            "instancetype": "m6a.xl",
            "offerings": [
                {
                    "capacity-type": "spot",
                    "avilability-zone": "az-1",
                    "price": 1.0
                },
                {
                    "capacity-type": "spot",
                    "avilability-zone": "az-2",
                    "price": 1.5
                },
                {
                    "capacity-type": "on-demand",
                    "price": 1.2
                }
            ]
        }
    ]
}
```
In this case spot would be launched but we would allow it to be launched in either az so could get the good price and could get the more expensive price. If the requirement was for launching in az-2 then the more expensive spot would be used.
This change would end up with the following, which would allow the cheap spot instance if we can launch in az-1 or would then drop back to on demand
```
{
    "instancetypes": [
        {
            "instancetype": "m6a.xl",
            "offerings": [
                {
                    "capacity-type": "spot",
                    "avilability-zone": "az-1",
                    "price": 1.0
                },
                {
                    "capacity-type": "on-demand",
                    "price": 1.2
                }
            ]
        }
    ]
}
```

njtran:
@rkirk-r7 got it, I agree that this would be necessary for when custom pricing overrides are implemented, but I'm wondering if this PR should be blocked on more progress and consensus for https://github.com/aws/karpenter/pull/4697. As I understand it, even though you are paying a lower cost, Karpenter's pricing API isn't aware of this, and will continue to think On-demand is always more expensive than spot, never triggering the error that you've implemented. 

I totally understand your desire to get this in as a stepping stone for https://github.com/aws/karpenter/pull/4697, but do you think this PR is valuable to get in if the other doesn't?

rkirk-r7:
> @rkirk-r7 got it, I agree that this would be necessary for when custom pricing overrides are implemented, but I'm wondering if this PR should be blocked on more progress and consensus for #4697. As I understand it, even though you are paying a lower cost, Karpenter's pricing API isn't aware of this, and will continue to think On-demand is always more expensive than spot, never triggering the error that you've implemented.
> 
> I totally understand your desire to get this in as a stepping stone for #4697, but do you think this PR is valuable to get in if the other doesn't?

I think this PR would work however we get more accurate pricing information, whether thats via #4697 or per instance overlay configuration etc. Kinda feel like i'm a bit stuck with this, because if I can get karpenter to be aware of more accuate pricing then it will not honour that pricing when launching, but can't fix the launching because the current price view always sees spot as the cheapest if its available. So i'm not really sure how to try and take this forward at this point.

njtran:
I totally understand, and I'm sorry that you've gone through the effort for this, but haven't been able to make progress. I think this is more indicative of the much larger design space that's been teased at here within our discussion and in Working Group. One of the reasons we ask for a design when considering features is to ensure that we can align on a direction before the developers spend their time and energy on implementing. 

We can _definitely_ improve on communicating these contributing requirements and guidelines better so it's clearer when a feature has the go-ahead.

> Kinda feel like i'm a bit stuck with this, because if I can get karpenter to be aware of more accuate pricing then it will not honour that pricing when launching, but can't fix the launching because the current price view always sees spot as the cheapest if its available.

I think this pretty much shows that this is a stepping stone for https://github.com/aws/karpenter-provider-aws/pull/4697. Correct me if I'm wrong, but since this isn't a bug with the current code, and we don't know how the overarching feature will be implemented, I don't think we should merge this. @rkirk-r7 thoughts?

njtran:
Closing this because of the above comment. Please feel free to reach out to us on slack if you have more questions.

grosser:
PR is closed but [issue](https://github.com/aws/karpenter-provider-aws/issues/4923) is closed too ... do we have anything tracking this feature/bug ? @njtran

PR #4941, opened 2023-10-26
ci: Fix boolean check with GHA

jonathan-innis:
Fixes #N/A 

**Description**

Booleans need to be directly compared with `true` when used in `if` conditionals; otherwise, they are treated as a "truthy" value regardless

- [ ] Yes, issue opened: # 
- [x] No


engedaam (approved):
LGTM 🚀

PR #4946, opened 2023-10-27
chore: Handle Provisioners with provider field for `karpenter-convert`

jonathan-innis:
Fixes #N/A 

**Description**

This PR adds handling for Provisioners that contain `spec.provider` by converting their fields into a `EC2NodeClass` and returning this additional object alongside the `NodePool`

`make presubmit`
`karpenter-convert`

- [ ] Yes, issue opened: # 
- [x] No


tools/karpenter-convert/pkg/convert/convert.go
```diff
@@ -169,43 +164,55 @@ func dropFields(buffer bytes.Buffer) string {
...
+
+		var providerObj runtime.Object
+		var err error
+		if provider := provisioner.Spec.Provider; provider != nil {
+			providerObj, err = convertProvider(provider.Raw, provisioner.Name)
+			if err != nil {
+				return nil, fmt.Errorf("converting spec.provider for Provisioner, %w", err)
+			}
+			provisioner.Spec.ProviderRef = &corev1alpha5.MachineTemplateRef{
+				Name: provisioner.Name,
```
njtran, line 186: should this be providerObj.Name? looks like we make the Provider the provisioner name anyway in the `convertProvider` call, but would make more sense if we just use the providerObj name.

njtran (approved):

tools/karpenter-convert/pkg/convert/convert.go
jonathan-innis: Sure, makes sense

njtran (approved):

PR #4954, opened 2023-10-30
chore: Simplify the SQS provider with no caching

jonathan-innis:
Fixes #N/A 

**Description**

Drop the caching layers from the SQS provider since there queue name can't change over the lifetime of the binary. This means that we grab the queueURL in the constructor of the queue and hold onto it for the lifetime of the executable.

`make presubmit`
`/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No


jonathan-innis (commented):
/karpenter snapshot

pkg/providers/sqs/sqs.go
```diff
@@ -0,0 +1,98 @@
...
+type Provider struct {
+	client sqsiface.SQSAPI
+
+	name string
+	url  string
+}
+
+func NewProvider(ctx context.Context, client sqsiface.SQSAPI) (*Provider, error) {
+	ret, err := client.GetQueueUrlWithContext(ctx, &sqs.GetQueueUrlInput{
+		QueueName: aws.String(options.FromContext(ctx).InterruptionQueue),
```
ellistarn, line 38: This constructor's method signature feels a bit magical. Any reason to not pass the queue name in as an argument?

jonathan-innis: Agreed, good callout. I was going back and forth on it, but I think your suggestion is best

test/pkg/environment/aws/environment.go
```diff
@@ -90,12 +90,11 @@ func NewEnvironment(t *testing.T) *Environment {
 		IAMAPI:        iam.New(session),
 		FISAPI:        fis.New(session),
 		EKSAPI:        eks.New(session),
-		SQSProvider:   interruption.NewSQSProvider(sqs.New(session)),
+		SQSProvider:   lo.Must(sqs.NewProvider(env.Context, servicesqs.New(session), lo.Must(os.LookupEnv("INTERRUPTION_QUEUE")))),
```
njtran, line 93: I think I might've asked this before, but wouldn't this panic if the queue doesn't exist? 

https://github.com/aws/karpenter/pull/4954/files#diff-89f1d90b65fce5d8e3d30df41441d721cecea16149000f221ceb29db20761354R38-R40

Should we be putting a lo.Must() around a function that we know can fail? If it can't, can we remove the error in the function itself?

jonathan-innis: I think it's reasonable to expect that the queue exists as a pre-requisite for the provider to be created. In this case, every E2E test is assumed to have a queue. Are you thinking that we should only create the provider if we are doing interruption testing?

jonathan-innis (commented):
/karpenter snapshot

test/pkg/environment/aws/environment.go
ellistarn: We have customers who aren't using the queue, and Karpenter shouldn't crash if they are not using interruption. We've taken issues on this before.

njtran (approved):

PR #4994, opened 2023-10-31
fix: deprioritize metal-24xl and metal-48xl in addition to non-suffixed metal instance types

bwagner5:
Fixes #N/A 

**Description**
 - Karpenter should deprioritize metal instance types since they are the same price as non-metal and take a lot longer to startup. EC2 previously only had instance types of the form `x.metal`. EC2 recently added metal sized variants `x.metal-24xl` and `x.metal-48xl`. The size variants were not being deprioritized properly, only `x.metal` was being deprioritized. This change checks if `metal` is contained anywhere in the size.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"math"
 	"sort"
+	"strings"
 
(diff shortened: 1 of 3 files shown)


njtran (approved):
LGTM

PR #4996, opened 2023-11-01
ci: Enable E2E Soak testing

engedaam:
Fixes #N/A 

**Description**
- Enable Soak testing to monitor the resource usage to identify CPU, Memory, and cluster resources leaks. 

- Forked Repo 

- [ ] Yes, issue opened: # 
- [x] No


.github/workflows/e2e-sock-trigger.yaml
```diff
@@ -0,0 +1,67 @@
...
+      with:
+       role-to-assume: arn:aws:iam::${{ vars.ACCOUNT_ID }}:role/${{ vars.ROLE_NAME }}
+       aws-region: ${{ inputs.region || 'us-east-1' }}
+       role-duration-seconds: 21600
+    - uses: ./.github/actions/e2e/install-eksctl
+      with:
+        version: v0.163.0
+    - id: create_cluster
+      run: |
+        export PREEXISTING=$(eksctl get cluster -o json | jq '.[].Name' | grep soak)
```
njtran, line 47: this assumes that there's only one cluster that has soak in its name. Is there some way to persist values of these clusters so that we can refer to them in the future? Maybe we can create clusters with name of the date they were created in, so that our soak test knows for sure what the name of the cluster should be?

.github/workflows/e2e-sock-trigger.yaml
```diff
@@ -0,0 +1,67 @@
...
+        echo "Found existing cluster name \"$PREEXISTING\""
+        if [[ $PREEXISTING != '' ]]; then 
+            echo PREEXISTING=$PREEXISTING >> $GITHUB_OUTPUT
+            echo CREATE_CLUSTER=false >> $GITHUB_OUTPUT
+        else
+            echo CREATE_CLUSTER=true >> $GITHUB_OUTPUT
```
njtran, line 53: Can't we just use the value of PREEXISTING as a check for if we need to create a cluster?

.github/workflows/e2e-sock-trigger.yaml
```diff
@@ -0,0 +1,67 @@
...
+        fi
+  sock:
+    needs: [reslove_cluster]
+    uses: ./.github/workflows/e2e.yaml
+    with:
+      suite: Beta/Integration
```
njtran, line 59: Shouldn't need the Beta/ Prefix anymore. 
```suggestion
      suite: Integration
```

.github/workflows/e2e-sock-trigger.yaml
```diff
@@ -0,0 +1,67 @@
...
+      region: ${{ inputs.region || 'us-east-1' }}
+      workflow_trigger: "soak"
+      create_cluster: ${{ needs.reslove_cluster.outputs.CREATE_CLUSTER == 'true' }}
```
njtran, line 62:
```suggestion
      create_cluster: ${{ needs.resolve_cluster.outputs.CREATE_CLUSTER == 'true' }}
```

.github/workflows/e2e-sock-trigger.yaml
```diff
@@ -0,0 +1,67 @@
...
+      preexisiting_cluster: ${{needs.reslove_cluster.outputs.PREEXISTING || '' }}
```
njtran, line 63:
```suggestion
      preexisiting_cluster: ${{needs.resolve_cluster.outputs.PREEXISTING || '' }}
```

.github/workflows/e2e.yaml
```diff
@@ -35,13 +35,19 @@ on:
...
           - "1.28"
         default: "1.28"
-      enable_metrics:
+      preexisiting_cluster:
+        type: string
+      create_cluster:
         type: boolean
-        default: false
+        default: true
+        required: true
```
njtran, line 43: Can we collapse this into one environment variable? When would I have a preexisting cluster but want to create a new one? and vice versa?

test/cloudformation/iam_cloudformation.yaml
```diff
@@ -249,6 +249,21 @@ Resources:
           - Effect: Allow
             Action: timestream:DescribeEndpoints
             Resource: "*"
+  GithubActionsPolicySecond:
```
njtran, line 252: Why does this need to be a new policy?

njtran: If we need to make it, can we rename it to be more descriptive?

test/hack/cleanup/main.go
```diff
@@ -33,11 +37,11 @@ import (
 	"github.com/samber/lo"
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
-	"k8s.io/client-go/util/workqueue"
 )
 
 const (
 	expirationTTL            = time.Hour * 12
+	sockExpirationTTL        = time.Hour * 24
```
njtran, line 44:
```suggestion
	soakExpirationTTL        = time.Hour * 24
```

test/hack/cleanup/main.go
```diff
@@ -76,12 +80,26 @@ func main() {
 	logger := lo.Must(zap.NewProduction()).Sugar()
 
 	expirationTime := time.Now().Add(-expirationTTL)
+	sockExpirationTime := time.Now().Add(-sockExpirationTTL)
```
njtran, line 83:
```suggestion
	soakExpirationTime := time.Now().Add(-sockExpirationTTL)
```

test/hack/cleanup/main.go
```diff
@@ -76,12 +80,26 @@ func main() {
...
 	cloudFormationClient := cloudformation.NewFromConfig(cfg)
 	iamClient := iam.NewFromConfig(cfg)
+	eksClient := eks.NewFromConfig(cfg)
+
+	if clusterName != "" {
+		continueWithCleanup, err := cleanUpCluster(ctx, clusterName, eksClient, sockExpirationTime)
+		if err == nil && continueWithCleanup {
+			logger.Infof("sock testing cluster (%s) will not need to be cleanup", clusterName)
+			return
+		} else if err != nil {
```
njtran, line 97:
```suggestion
	eksClient := eks.NewFromConfig(cfg)

	if clusterName != "" {
		continueWithCleanup, err := cleanUpCluster(ctx, clusterName, eksClient, soakExpirationTime)
		if err == nil && continueWithCleanup {
			logger.Infof("soak testing cluster (%s) will not need to be cleanup", clusterName)
			return
		} else if err != nil {
```

.github/workflows/e2e-sock-trigger.yaml
```diff
@@ -0,0 +1,67 @@
...
+      git_ref:
+        type: string
+      region:
+        required: true
+        default: 'us-east-1'
+        type: choice
+        options:
+          - "us-east-1"
+          - "us-west-2"
+      cleanup:
```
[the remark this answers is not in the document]
engedaam: I believe those metrics are only used for scale testing. Those metrics should be the Timestream data

.github/actions/e2e/cleanup/action.yaml
```diff
@@ -27,10 +27,6 @@ runs:
     - uses: ./.github/actions/e2e/install-eksctl
       with:
         version: ${{ inputs.eksctl_version }}
-    - name: delete-cluster
-      shell: bash
-      run: |
-        eksctl delete cluster --name ${{ inputs.cluster_name }} --timeout 60m --wait || true
```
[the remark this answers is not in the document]
engedaam: No, we are just moving the `eksctl` delete to the go clean-up script

.github/workflows/e2e-sock-trigger.yaml
```diff
@@ -0,0 +1,67 @@
...
+      PREEXISTING: ${{ steps.create_cluster.outputs.PREEXISTING }}
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ inputs.git_ref }}
+    - name: configure aws credentials
+      uses: aws-actions/configure-aws-credentials@v4.0.1
+      with:
+       role-to-assume: arn:aws:iam::${{ vars.ACCOUNT_ID }}:role/${{ vars.ROLE_NAME }}
+       aws-region: ${{ inputs.region || 'us-east-1' }}
```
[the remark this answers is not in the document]
engedaam: On scheduled events, we will not get values passed

test/cloudformation/iam_cloudformation.yaml
engedaam: We have run out of character space on the policy, so we will need a second policy to add new permissions

.github/workflows/e2e-sock-trigger.yaml
```diff
@@ -0,0 +1,57 @@
...
+        fi
+  sock:
+    needs: [resolve_cluster]
+    uses: ./.github/workflows/e2e.yaml
+    with:
+      suite: Integration
+      region: ${{ inputs.region || 'us-east-1' }}
+      workflow_trigger: "soak"
+      preexisiting_cluster: ${{needs.resolve_cluster.outputs.PREEXISTING || '' }}
+      cleanup: true
```
njtran, line 55: Do we always want to clean up on soak testing? I don't think so right?

.github/workflows/e2e.yaml
```diff
@@ -91,15 +95,23 @@ jobs:
...
           # Creating jitter so that we can stagger cluster creation to avoid throttling
           sleep $(( RANDOM % 300 + 1 ))
       - id: generate-cluster-name
+        name: generate-cluster-name
+        if: always() && inputs.preexisiting_cluster == ''
         run: |
-          CLUSTER_NAME="$(echo ${{ inputs.suite }}-"$RANDOM$RANDOM" | awk '{print tolower($0)}' | tr / -)"
-          echo Using cluster name "$CLUSTER_NAME"
+          if [[ ${{ inputs.workflow_trigger || 'create-cluster' }} == 'soak' ]]; then 
+            CLUSTER_NAME=${{ inputs.workflow_trigger }}-periodic
```
njtran, line 107:
```suggestion
            CLUSTER_NAME=${{ inputs.workflow_trigger }}-soak-periodic
```

.github/workflows/e2e.yaml
```diff
@@ -137,10 +151,10 @@ jobs:
...
           fi
-          aws eks update-kubeconfig --name ${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}
+          aws eks update-kubeconfig --name ${{ steps.generate-cluster-name.outputs.CLUSTER_NAME || inputs.preexisiting_cluster }}
           TEST_SUITE="$TEST_SUITE" ENABLE_METRICS=${{ inputs.enable_metrics }} METRICS_REGION=${{ vars.TIMESTREAM_REGION }} GIT_REF="$(git rev-parse HEAD)" \
-            CLUSTER_NAME="${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}" CLUSTER_ENDPOINT="$(aws eks describe-cluster --name ${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }} --query "cluster.endpoint" --output text)" \
-            INTERRUPTION_QUEUE="${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}" make e2etests
+            CLUSTER_NAME="${{ steps.generate-cluster-name.outputs.CLUSTER_NAME || inputs.preexisiting_cluster }}" CLUSTER_ENDPOINT="$(aws eks describe-cluster --name ${{ steps.generate-cluster-name.outputs.CLUSTER_NAME || inputs.preexisiting_cluster }} --query "cluster.endpoint" --output text)" \
+            INTERRUPTION_QUEUE="${{ steps.generate-cluster-name.outputs.CLUSTER_NAME || inputs.preexisiting_cluster }}" make e2etests
```
njtran, line 157: Consider defining a CLUSTER_NAME env var that is the result of `${{ steps.generate-cluster-name.outputs.CLUSTER_NAME || inputs.preexisiting_cluster }}` so that you dont have to keep referencing it all over

.github/workflows/e2e.yaml
```diff
@@ -156,15 +170,15 @@ jobs:
           account_id: ${{ vars.ACCOUNT_ID }}
           role: ${{ vars.ROLE_NAME }}
           region: ${{ inputs.region }}
-          cluster_name: ${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}
-      - name: cleanup karpenter and cluster '${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}' resources
+          cluster_name: ${{ inputs.preexisiting_cluster || steps.generate-cluster-name.outputs.CLUSTER_NAME }}
```
njtran, line 173: Consider using something like this in `run the test suite`

test/cloudformation/iam_cloudformation.yaml
```diff
@@ -249,6 +249,24 @@ Resources:
           - Effect: Allow
             Action: timestream:DescribeEndpoints
             Resource: "*"
+  # This second is need to add more permission
+  # GithubActionsPolicy has hit the maximum character size limit for managed policies of 6,144
+  # https://repost.aws/knowledge-center/iam-increase-policy-size#
```
njtran, line 254: Are we absolutely sure that we can't reduce the other policy? What if the other policy is smaller in the future, how do we reason about which policy should go where? If we absolutely need it, we should think of two reasonable policies groups, especially if we need to add more permissions to the other policy. If the current groups are the other one and this one, that's fine, but I'd like us to be more deliberate about this rather than just appending it to the list.

test/hack/cleanup/main.go
```diff
@@ -93,3 +105,26 @@ func main() {
 		}
 	}
 }
+
+func cleanUpCluster(ctx context.Context, name string, client *eks.Client, log *zap.SugaredLogger, expirationTime time.Time) {
+	if name != "" {
+		if strings.HasPrefix(name, "soak-") {
```
njtran, line 111: needs to be this, right?
```suggestion
		if strings.HasPrefix(name, "soak-periodic") {
```

test/hack/cleanup/main.go
```diff
@@ -93,3 +105,26 @@ func main() {
...
+			cluster, err := client.DescribeCluster(ctx, &eks.DescribeClusterInput{Name: aws.String(name)})
+			if err != nil {
+				log.Errorf("%v", err)
+			}
+
+			if !lo.FromPtr(cluster.Cluster.CreatedAt).Before(expirationTime) {
+				log.Infof("soak testing cluster (%s) will not need to be cleanup", name)
```
njtran, line 118: Can you also add some info on when the cluster _should_ be cleaned up?
```suggestion
				log.Infof("soak testing cluster (%s) does not need to be cleaned up", name)
```

test/hack/cleanup/main.go
```diff
@@ -93,3 +105,26 @@ func main() {
...
+				os.Exit(0)
+			}
+		}
+
+		deleteCluster := exec.Command("eksctl", "delete", "cluster", "--name", name, "--timeout", "60m", "--wait")
+		fmt.Println(deleteCluster.String())
+		if out, err := deleteCluster.Output(); err != nil {
+			fmt.Println(string(out))
+			log.Errorf("%v", err)
+		}
```
njtran, line 128: Will this always work? I feel like we probably need our cleanup logic to refer to the cloud formation stacks so that it cleans up all associate resources too.

.github/workflows/e2e-sock-trigger.yaml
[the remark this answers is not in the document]
engedaam: `e2e-soak-trigger.yaml` file contains the highest level abstraction for the tests. I think that would be consistent with the rest of the files

.github/workflows/e2e-sock-trigger.yaml
engedaam: We eventually do want to cleanup the soak testing clusters. The check of when we cleanup the cluster is now held in the sweeper logic

test/hack/cleanup/main.go
engedaam: So EKSCTL does cleanup all the stacks. EKSCTL is aware of the dependency between stacks. For example, we will need to delete the stack that creates the NodeGroups first and wait for that to be deleted completely before we can delete the cluster. I think it would be cleaner to allow EKSCTL to handle that logic.

test/cloudformation/iam_cloudformation.yaml
engedaam: I will look into a way to split up the first policy into multiple, to prevent us from hit the limit

.github/workflows/e2e-sock-trigger.yaml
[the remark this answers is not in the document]
njtran: Ah yeah, just meant to rename sock to soak, thanks!

test/hack/cleanup/main.go
njtran: Ah you're right, got my wires crossed. Still, this gets me thinking, what's the retry rate on this? Just the recurrence of the cron?

.github/workflows/e2e-sock-trigger.yaml
njtran: So what does this cleanup refer to?

engedaam: This cleanup references us to run the cleanup script. The field is required

test/hack/soak/main.go
```diff
@@ -0,0 +1,59 @@
...
+	for _, cluster := range clusters.Clusters {
+		clusterDetails := lo.Must(eksClient.DescribeCluster(ctx, &eks.DescribeClusterInput{Name: aws.String(cluster)}))
+		if clusterDetails.Cluster.CreatedAt.YearDay() == time.Now().YearDay() {
+			createNewCluster = false
+		}
+
+		if strings.HasPrefix(cluster, "soak-periodic-") {
+			outputList = append(outputList, cluster)
+		}
+	}
```
njtran, line 51: Is this worth paginating? Can you see what the max items in a page is?

test/hack/soak/main.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if createNewCluster {
+		outputList = append(outputList, "createNewCluster")
+	}
+
+	output, _ := json.Marshal(outputList)
+	fmt.Println(string(output))
```
njtran, line 58: Should just return the list of cluster names and it just doesn't do anything for an empty/non-empty list.

.github/workflows/e2e.yaml
```diff
@@ -156,15 +174,15 @@ jobs:
...
+      - name: cleanup karpenter and cluster '${{ inputs.preexisiting_cluster || steps.generate-cluster-name.outputs.CLUSTER_NAME }}' resources
```
njtran, line 178: should inputs be the second parameter for scheduled triggers?

.github/workflows/e2e.yaml
```diff
@@ -112,6 +124,7 @@ jobs:
           private_cluster: ${{ inputs.suite == 'PrivateCluster' }}
           git_ref: ${{ inputs.git_ref }}
       - name: install prometheus
+        if: always() && inputs.preexisiting_cluster == 'createNewCluster'
```
njtran, line 127:
```suggestion
        if: always() && inputs.preexisiting_cluster == ''
```

.github/workflows/e2e.yaml
```diff
@@ -121,6 +134,7 @@ jobs:
           workspace_id: ${{ vars.WORKSPACE_ID }}
           git_ref: ${{ inputs.git_ref }}
       - name: install karpenter
+        if: always() && inputs.preexisiting_cluster == 'createNewCluster'
```
njtran, line 137:
```suggestion
        if: always() && inputs.preexisiting_cluster == ''
```

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,64 @@
...
+        PREEXISTING_CLUSTERS="$(go run main.go)" 
+        echo PREEXISTING_CLUSTERS="$PREEXISTING_CLUSTERS" >> "$GITHUB_OUTPUT"
+      working-directory: ./test/hack/soak
+      shell: bash
+  soak:
+    needs: [resolve_cluster]
+    strategy:
+      fail-fast: false
+      matrix: 
+        cluster: ${{ fromJson(needs.resolve_cluster.outputs.PREEXISTING_CLUSTERS) }}
```
njtran, line 55: I think we should be able to pass in an empty string.

.github/workflows/e2e.yaml
njtran, line 1: Let's factor out the install tasks from the test run tasks, so that we can use each of these items more piecemeal

njtran: talked offline, doing another PR to factor this out, which this existing PR will rebase off of.

.github/workflows/e2e-soak-trigger.yaml
engedaam: We can't pass an empty string, however the value that is passed to e2e.yaml will be an empty string

test/hack/soak/main.go
engedaam: We would need some way of indicating that we need to create a cluster. This would be the best way to go forward. To make the e2e logic cleaner, that will treat `createNewCluster` as an empty string

test/hack/cleanup/main.go
```diff
@@ -44,12 +50,18 @@ func main() {
 	logger := lo.Must(zap.NewProduction()).Sugar()
 
 	expirationTime := time.Now().Add(-expirationTTL)
+	soakExpirationTime := time.Now().Add(-soakExpirationTTL)
```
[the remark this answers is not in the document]
engedaam: I was thinking it would be more readable setting the expiration out expliceitly

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,64 @@
+name: E2ESoakTrigger
+on:
+  schedule:
+    - cron: '0 */3 * * *'
```
jonathan-innis, line 4: If this is every 3h, is this going to cause issues if any of the previous tests go over?

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,64 @@
...
+  workflow_run:
```
jonathan-innis, line 5: Should we have a `workflow_run` trigger for this one?

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,64 @@
...
+    strategy:
+      fail-fast: false
+      matrix: 
+        cluster: ${{ fromJson(needs.resolve_cluster.outputs.PREEXISTING_CLUSTERS) }}
+    uses: ./.github/workflows/e2e.yaml
+    with:
+      suite: Integration
+      region: ${{ inputs.region || 'us-east-1' }}
+      workflow_trigger: "soak"
+      preexisiting_cluster: ${{ matrix.cluster == 'createNewCluster' && '' || matrix.cluster }}
```
jonathan-innis, line 61: Should this just be the `cluster_name`? If you pass through the cluster_name, then you just use the cluster name. Otherwise, we generate and create the cluster.

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,64 @@
...
+      cleanup: true
```
jonathan-innis, line 62: We don't want to cleanup, right? We want to keep the cluster around so we should disable this

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - uses: actions/setup-go@v4
+      with:
+        go-version-file: test/hack/soak/go.mod
+        cache-dependency-path: test/hack/soak/go.sum
+        check-latest: true
+        cache: false
+    - id: list_clusters 
+      name: "Run list clusters script"
+      run: |
+        PREEXISTING_CLUSTERS="$(go run main.go)"
```
jonathan-innis, line 46: Why do we need a go script for this? This is another thing that we have to maintain. Can we just `eks describe-clusters` to get the information that we need on the cluster names

.github/workflows/e2e-soak-trigger.yaml
engedaam: All the tests have a 3h timeout. The thinking being here, tests should never have any overlap

test/hack/cleanup/main.go
```diff
@@ -93,3 +107,30 @@ func main() {
...
+			return false
+		}
+
+		if !lo.FromPtr(cluster.Cluster.CreatedAt).Before(soakExpirationTime) {
+			logger.Infof("soak testing cluster (%s) does not need to be cleaned up until %s", name, soakExpirationTime)
+			return true
+		}
+	}
+
+	deleteCluster := exec.Command("eksctl", "delete", "cluster", "--name", name, "--timeout", "60m", "--wait")
```
jonathan-innis, line 128: I don't think that we should be `exec`-ing out to `eksctl`. We should work to make this soak testing portable, so reduce the dependency sprawl that's required on the system that's running it.

test/hack/cleanup/main.go
```diff
@@ -18,20 +18,26 @@ import (
...
 	"github.com/aws/karpenter/test/hack/cleanup/metrics"
 	"github.com/aws/karpenter/test/hack/cleanup/resourcetypes"
+
+	"github.com/aws/aws-sdk-go-v2/service/iam"
+	"github.com/samber/lo"
+	"go.uber.org/zap"
 )
 
 const expirationTTL = time.Hour * 12
+const soakExpirationTTL = time.Hour * 168 // 7 Days
```
jonathan-innis, line 40: AFAIK, everything should be the same in terms of resource cleanup except for the tagging that's associated with the cluster resources. Can we just make this code work as-is; the only difference being that we have a tag like `testing/type=soak`

test/hack/soak/main.go
```diff
@@ -0,0 +1,64 @@
...
+	ctx := context.Background()
+	cfg := lo.Must(config.LoadDefaultConfig(ctx))
+	eksClient := eks.NewFromConfig(cfg)
+
+	var outputList []string
+	createNewCluster := true
+
+	var nextToken *string
+	for {
+		clusters := lo.Must(eksClient.ListClusters(ctx, &eks.ListClustersInput{NextToken: nextToken, MaxResults: aws.Int32(50)}))
```
jonathan-innis, line 40: Can we enable this logic with bash calls? I'd like to avoid maintaining additional scripting files unless we absolutely need it. I think all you need to do is to get clusters that are created with specific tags.

.github/actions/e2e/cleanup/action.yaml
```diff
@@ -27,10 +27,6 @@ runs:
     - uses: ./.github/actions/e2e/install-eksctl
       with:
         version: ${{ inputs.eksctl_version }}
-    - name: delete-cluster
```
jonathan-innis, line 30: I'm not quite getting the reason why we are removing this.

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,64 @@
+name: E2ESoakTrigger
```
jonathan-innis, line 1: How frequently did we say that we are creating the clusters? I'd like to decouple the clusters creation from the cluster running in the case of soak testing. Ideally, we can just create a separate workflow that runs something like daily where we can run the `create-cluster` step

.github/workflows/e2e.yaml
```diff
@@ -97,10 +101,20 @@ jobs:
...
+              CLUSTER_NAME=$(echo ${{ inputs.suite }}-$RANDOM$RANDOM | awk '{print tolower($0)}' | tr / -)
+            fi 
+          else 
+            CLUSTER_NAME=${{ inputs.preexisiting_cluster }}
+          fi
+          echo "Using cluster name \"$CLUSTER_NAME\""
           echo CLUSTER_NAME="$CLUSTER_NAME" >> "$GITHUB_OUTPUT"
       - name: create eks cluster '${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}'
+        if: always() && steps.generate-cluster-name.outputs.CLUSTER_NAME == ''
         uses: ./.github/actions/e2e/create-cluster
```
jonathan-innis, line 118:
```suggestion
        uses: ./.github/actions/e2e/setup-cluster
```

Does this become `setup-cluster` since now we are doing more than just cluster creation?

test/cloudformation/iam_cloudformation.yaml
```diff
@@ -39,6 +39,112 @@ Resources:
     Type: AWS::IAM::ManagedPolicy
     Properties:
       ManagedPolicyName: GithubActionsPolicy
+      PolicyDocument:
```
jonathan-innis, line 42: What was the need for moving this here? The current IAM policy gets all of the permissions from the permission boundary already

.github/workflows/e2e.yaml
```diff
@@ -36,13 +36,15 @@ on:
           - "1.27"
           - "1.28"
         default: "1.28"
-      enable_metrics:
-        type: boolean
-        default: false
+      cluster_name:
```
jonathan-innis, line 39: Consider adding a description to this argument to the workflow_call. Something that states that if this value is an empty string or isn't specified, that implies that we are going to create a cluster. Else, if it is specified, we are going to use the existing cluster

.github/workflows/e2e.yaml
```diff
@@ -123,13 +135,12 @@ jobs:
           if [[ inputs.suite == 'PrivateCluster' ]]; then
             TEST_SUITE="Integration"
           fi
-          aws eks update-kubeconfig --name ${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}
-          TEST_SUITE="$TEST_SUITE" ENABLE_METRICS=${{ inputs.enable_metrics }} METRICS_REGION=${{ vars.TIMESTREAM_REGION }} GIT_REF="$(git rev-parse HEAD)" \
-            CLUSTER_NAME="${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}" CLUSTER_ENDPOINT="$(aws eks describe-cluster --name ${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }} --query "cluster.endpoint" --output text)" \
-            INTERRUPTION_QUEUE="${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}" make e2etests
+          aws eks update-kubeconfig --name "${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}"
```
jonathan-innis, line 138: Not a big deal but any reason this expression got re-ordered here

.github/workflows/e2e.yaml
```diff
@@ -123,13 +135,12 @@ jobs:
...
+          INTERRUPTION_QUEUE="${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}" CLUSTER_NAME="${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}" TEST_SUITE="$TEST_SUITE" ENABLE_METRICS=${{ inputs.enable_metrics }} METRICS_REGION=${{ vars.TIMESTREAM_REGION }} GIT_REF="$(git rev-parse HEAD)" \
+            CLUSTER_ENDPOINT="$(aws eks describe-cluster --name "${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}" --query "cluster.endpoint" --output text)"  make e2etests
       - name: notify slack of success or failure
         uses: ./.github/actions/e2e/slack/notify
-        if: (success() || failure()) && github.event_name != 'workflow_run' && inputs.workflow_trigger != 'versionCompatibility'
+        if: (success() || failure()) && github.event_name != 'workflow_run' && inputs.workflow_trigger != 'versionCompatibility' && inputs.workflow_trigger != 'soak'
```
jonathan-innis, line 143: We said that we wanted to notify on slack for the soak testing to start. We just need to designate a separate channel to get the notifications. Should we specify and add that here?

.github/workflows/sweeper.yaml
```diff
@@ -25,6 +25,7 @@ jobs:
           go-version-file: test/hack/resource/go.mod
           check-latest: true
           cache-dependency-path: "test/hack/resource/go.sum"
-      - run: go run main.go
+      # us-east-1 contains the soak tests and resources should be sweept if they are older then 8 days
+      - run: go run main.go ${{ matrix.region == 'eu-north-1' && '-expiration 192h' || '' }}
```
jonathan-innis, line 29: Probably just easier to specify this using bash. Something like

```bash
if [[ "${{ matrix.region }}" == "eu-north-1" ]]; then
  go run main.go --expiration 192h
else
  go run main.go --expiration 12h
fi
```

Also, I think it's reasonable to have `12h` be the default expiration time in the script, but it might make sense to just specify it here in the bash cleanup script for clarity so that it's easier to reason about.

test/hack/resource/clean/main.go
```diff
@@ -31,19 +31,22 @@ import (
...
 
-const expirationTTL = time.Hour * 12
 const sweeperCleanedResourcesTableName = "sweeperCleanedResources"
 
 func main() {
-	var clusterName string
-	if len(os.Args) == 2 {
-		clusterName = os.Args[1]
-	}
+	expiration := flag.String("expiration", "12h", "define the expirationTTL of the resources")
```
jonathan-innis, line 37:
```suggestion
	expiration := flag.String("--expiration", "12h", "define the expirationTTL of the resources")
```

More standard syntax

test/hack/soak/main.go
```diff
@@ -0,0 +1,72 @@
...
+func main() {
+	ctx := context.Background()
+	cfg := lo.Must(config.LoadDefaultConfig(ctx))
+	eksClient := eks.NewFromConfig(cfg)
+
+	var outputList []*cluster
+	createNewCluster := true
+	expirationTime := time.Now().Add(-expirationTTL)
+
+	var nextToken *string
```
jonathan-innis, line 46: Do we expect to have a `nextToken` and pages with such a small result set (I see below that we are limiting the response to 50 values)

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      PREEXISTING_CLUSTERS: ${{ steps.list_clusters.outputs.PREEXISTING_CLUSTERS }}
+    steps:
+    - uses: actions/checkout@v4
+    - name: configure aws credentials
+      uses: aws-actions/configure-aws-credentials@010d0da01d0b5a38af31e9c3470dbfdabdecca3a
+      with:
+       role-to-assume: arn:aws:iam::${{ vars.ACCOUNT_ID }}:role/${{ vars.ROLE_NAME }}
+       aws-region: eu-north-1
```
njtran, line 17: Do we need to parameterize this?

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,49 @@
...
+        check-latest: true
+        cache: false
+    # Grab all the names of the soak testing clusters, and identify clusters that will need to be deleted
+    # Empty cluster_name will indicate that a new cluster will be created
+    - id: list_clusters 
+      name: "Run list clusters script"
+      run: |
+        PREEXISTING_CLUSTERS="$(go run main.go)" 
+        echo PREEXISTING_CLUSTERS="$PREEXISTING_CLUSTERS" >> "$GITHUB_OUTPUT"
+      working-directory: ./test/hack/soak
```
njtran, line 31: can we rename the function main.go to something like `get_cluster_names.go` that is more indicative of what it's actually doing?

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,49 @@
...
+    permissions:
+      id-token: write # aws-actions/configure-aws-credentials@v4.0.1
+    needs: [resolve_cluster]
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.resolve_cluster.outputs.PREEXISTING_CLUSTERS) }}
+    uses: ./.github/workflows/e2e.yaml
+    with:
+      suite: Integration
+      region: eu-north-1
```
njtran, line 43: same here on parameterization

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,49 @@
...
+  soak:
+    permissions:
+      id-token: write # aws-actions/configure-aws-credentials@v4.0.1
+    needs: [resolve_cluster]
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.resolve_cluster.outputs.PREEXISTING_CLUSTERS) }}
+    uses: ./.github/workflows/e2e.yaml
+    with:
+      suite: Integration
```
njtran, line 42: Is there value in factoring this out to be one or multiple suites? I can imagine us wanting to do more here, especially since the name Integration is a little misleading since it doesn't test integration of all things.

.github/workflows/e2e.yaml
```diff
@@ -131,7 +146,8 @@ jobs:
         uses: ./.github/actions/e2e/slack/notify
         if: (success() || failure()) && github.event_name != 'workflow_run' && inputs.workflow_trigger != 'versionCompatibility'
         with:
-          url: ${{ secrets.SLACK_WEBHOOK_URL }}
+          cluster_name: ${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}
+          url: ${{ inputs.workflow_trigger == 'soak' || secrets.SLACK_WEBHOOK_SOAK_URL && secrets.SLACK_WEBHOOK_URL }}
```
njtran, line 150: I'm not sure I understand the need of having another slack url just for soak. Are we creating two separate notification channels? Can you explain to me what this line does and add it into a comment?

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,49 @@
...
+      region: eu-north-1
+      workflow_trigger: "soak"
+      cluster_name: ${{ matrix.cluster_name }}
+      cleanup: ${{ matrix.cluster_cleanup }}
+    secrets:
+      SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      SLACK_WEBHOOK_SOAK_URL: ${{ secrets.SLACK_WEBHOOK_SOAK_URL }}
```
njtran, line 49: not quite understanding why we need two urls here.

.github/workflows/e2e.yaml
```diff
@@ -64,9 +66,14 @@ on:
         required: true
       workflow_trigger:
         type: string
+      cluster_name:
+        type: string
+        description: If this cluster_name is an empty string or isn't specified, a new cluster will be created. A specified cluster_name, will run tests on an existing cluster
```
njtran, line 71: What do you think about this?
```suggestion
        description: If cluster_name is empty, a new cluster will be created. Otherwise, tests will run on an existing cluster
```

.github/workflows/e2e.yaml
```diff
@@ -97,10 +104,18 @@ jobs:
...
-          CLUSTER_NAME="$(echo ${{ inputs.suite }}-"$RANDOM$RANDOM" | awk '{print tolower($0)}' | tr / -)"
-          echo Using cluster name "$CLUSTER_NAME"
+          CLUSTER_NAME='${{ inputs.cluster_name }}'
+          if [[ '${{ inputs.cluster_name }}' == '' ]]; then 
+            if [[ '${{ inputs.workflow_trigger }}' == 'soak' ]]; then 
+              CLUSTER_NAME=$(echo soak-periodic-$RANDOM$RANDOM | awk '{print tolower($0)}' | tr / -)
+            else 
+              CLUSTER_NAME=$(echo ${{ inputs.suite }}-$RANDOM$RANDOM | awk '{print tolower($0)}' | tr / -)
+            fi 
+          fi
```
njtran, line 114: nit: I think the way you did it in action.yaml looks cleaner, but up to you. https://github.com/aws/karpenter/pull/4996/files#diff-4eb536776cca4973b074324c28a1f76a816f488a2bd4c1de81f8b1585fab402bL23-L27

```
if a & b  { 
  set cluster name
} 
elif a & !b {
   set cluster name
}
```

.github/workflows/e2e.yaml
```diff
@@ -97,10 +104,18 @@ jobs:
...
+          echo "Using cluster name \"$CLUSTER_NAME\""
           echo CLUSTER_NAME="$CLUSTER_NAME" >> "$GITHUB_OUTPUT"
       - name: setup eks cluster '${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}'
+        if: inputs.cluster_name == ''
```
njtran, line 118: Looks like e2e/setup-cluster assumes that cluster name is set. When do we actually generate the cluster name? 

Required: https://github.com/aws/karpenter/blob/main/.github/actions/e2e/setup-cluster/action.yaml#L25-L27
Used as input to a function: https://github.com/aws/karpenter/blob/main/.github/actions/e2e/setup-cluster/action.yaml#L66
same pt2: https://github.com/aws/karpenter/blob/main/.github/actions/e2e/setup-cluster/action.yaml#L83

.github/workflows/sweeper.yaml
```diff
@@ -25,6 +25,12 @@ jobs:
           go-version-file: test/hack/resource/go.mod
           check-latest: true
           cache-dependency-path: "test/hack/resource/go.sum"
-      - run: go run main.go
+      # eu-north-1 contains the soak tests and resources should be sweept if they are older then 8 days
```
njtran, line 28:
```suggestion
      # eu-north-1 contains the soak tests and resources should be swept if they are older than 8 days
```

.github/workflows/sweeper.yaml
```diff
@@ -25,6 +25,12 @@ jobs:
...
+      - run: |
+          if [[ "${{ matrix.region }}" == "eu-north-1" ]]; then
```
njtran, line 30: Is it possible for users to run other tests in eu-north-1? If so, does this mean that the resources take a lot longer to get swept up? Is there a way we can make sure this doesn't happen?

.github/workflows/sweeper.yaml
```diff
@@ -25,6 +25,12 @@ jobs:
...
+            go run main.go --expiration 192h
+          else
+            go run main.go --expiration 12h
```
njtran, line 33: Looks like the default is 12h. Can we just not set it here, or not make a default?

test/hack/resource/clean/main.go
```diff
@@ -31,19 +31,22 @@ import (
...
-const expirationTTL = time.Hour * 12
 const sweeperCleanedResourcesTableName = "sweeperCleanedResources"
 
 func main() {
-	var clusterName string
-	if len(os.Args) == 2 {
-		clusterName = os.Args[1]
-	}
+	expiration := flag.String("--expiration", "12h", "define the expirationTTL of the resources")
+	clusterName := flag.String("--cluster-name", "", "define cluster name to cleanup")
```
njtran, line 38: Can we make this required?

test/hack/resource/clean/main.go
```diff
@@ -31,19 +31,22 @@ import (
...
+
 	ctx := context.Background()
 	cfg := lo.Must(config.LoadDefaultConfig(ctx))
 
 	logger := lo.Must(zap.NewProduction()).Sugar()
 
+	expirationTTL, err := time.ParseDuration(lo.FromPtr(expiration))
+	if err != nil {
+		logger.Fatalln("need a valid expiration duration", err)
+	}
```
njtran, line 49: What do you think about also validating the cluster name?

test/hack/resource/clean/main.go
```diff
@@ -73,10 +76,10 @@ func main() {
 		resourceLogger := logger.With("type", resourceTypes[i].String())
 		var ids []string
 		var err error
-		if clusterName == "" {
+		if lo.FromPtr(clusterName) == "" {
 			ids, err = resourceTypes[i].GetExpired(ctx, expirationTime)
 		} else {
```
njtran, line 81: Can you leave a comment here so it's clearer what's happening? Maybe it was just tougher to figure out what was going on because I had to click through a couple of links not in the PR to find what this was doing. 

i.e. If there's no cluster defined, clean up all resources. otherwise, only cleanup the resources associated with the cluster.

test/hack/resource/clean/main.go
```diff
@@ -88,7 +91,7 @@ func main() {
 				resourceLogger.Errorf("%v", err)
 			}
 			// Should only fire metrics if the resource have expired
-			if clusterName == "" {
+			if lo.FromPtr(clusterName) == "" {
```
njtran, line 94: Shouldn't we fire a metric regardless of it was from a specific cluster? IIUC, this should fire if L89 doesn't fail, regardless of if it was cleaned up for a specific cluster or not.

test/hack/soak/main.go
njtran, line 1: Would be great to have a comment here to know what this file does, or just rename it to be more indicative of what it does.

njtran (commented):
Nice work! Great to see progress being made.

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,49 @@
+name: E2ESoakTrigger
+on:
+  schedule:
+    - cron: '0 */3 * * *'
+jobs:
+  resolve_cluster:
+    if: github.repository == 'aws/karpenter'
```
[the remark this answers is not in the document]
jonathan-innis:
```suggestion
    if: github.repository == 'aws/karpenter-provider-aws'
```
Yep

.github/workflows/e2e.yaml
jonathan-innis: The plan here is that we can monitor the soak testing on a separate notification channel on the outset until we have proper monitoring and clean-up in place. I don't think we want to send this to the other notification channel since this is going to cause us to get notifications every 3h and is going to spam quite a bit, taking away from the other notifications that are due to periodics and repo merges.

.github/workflows/sweeper.yaml
jonathan-innis: I had asked to specify the expiration explicitly here. I think that it makes it clearer. Having a default probably still makes sense for anyone running the sweeper manually.

.github/actions/e2e/slack/notify/action.yaml
```diff
@@ -20,8 +23,10 @@ runs:
     - id: get-run-name
       shell: bash
       run: |
-        if [[ ${{ github.event_name }} == "schedule" ]]; then
+        if [[ ${{ github.event_name }} == "schedule" && inputs.suite != "soak" ]]; then
           RUN_NAME="${{ inputs.suite }}-periodic"
+        elif [[ ${{ github.event_name }} == "schedule" && inputs.suite == "soak" ]]; then
+          RUN_NAME="$soak-periodic(${{ inputs.cluster_name }})"
```
jonathan-innis, line 29: I wonder how much value there is in putting the cluster name in the channel since we can find the run url to track-down the failure if we see that this has failed in Slack

.github/workflows/e2e.yaml
engedaam: ^^

.github/actions/e2e/slack/notify/action.yaml
```diff
@@ -18,8 +21,10 @@ runs:
     - id: get-run-name
       shell: bash
       run: |
-        if [[ ${{ github.event_name }} == "schedule" ]]; then
+        if [[ ${{ github.event_name }} == "schedule" && inputs.suite != "soak" ]]; then
           RUN_NAME="${{ inputs.suite }}-periodic"
+        elif [[ ${{ github.event_name }} == "schedule" && inputs.suite == "soak" ]]; then
```
jonathan-innis, line 26: optional: You could make a simplifying assumption here since if the value doesn't equal soak, it will already be caught by the previous `if`. The way you have it is fine, but it's generally easier to start from more specific to less specific here when it comes to if conditionals, so something like this:

```bash
if [[ ${{ github.event_name }} == "schedule" && inputs.suite == "soak" ]]; then
  ....
elif [[ ${{ github.event_name }} == "schedule" ]]; then
  ....
else
  ....
fi
```

.github/workflows/e2e-soak-trigger.yaml
```diff
@@ -0,0 +1,51 @@
...
+    outputs:
+      PREEXISTING_CLUSTERS: ${{ steps.list_clusters.outputs.PREEXISTING_CLUSTERS }}
+    steps:
+    - uses: actions/checkout@v4
+    - name: configure aws credentials
+      uses: aws-actions/configure-aws-credentials@010d0da01d0b5a38af31e9c3470dbfdabdecca3a
+      with:
+       role-to-assume: arn:aws:iam::${{ vars.ACCOUNT_ID }}:role/${{ vars.ROLE_NAME }}
+       aws-region: eu-north-1
+    - uses: actions/setup-go@v4
```
jonathan-innis, line 20: We're not pinning our actions to SHAs and we should be here

jonathan-innis: I think that there's also a miss on the `ResourceCount` workflow on this

.github/workflows/e2e.yaml
```diff
@@ -97,10 +104,18 @@ jobs:
       - id: generate-cluster-name
         name: generate cluster name
         run: |
-          CLUSTER_NAME="$(echo ${{ inputs.suite }}-"$RANDOM$RANDOM" | awk '{print tolower($0)}' | tr / -)"
-          echo Using cluster name "$CLUSTER_NAME"
+          CLUSTER_NAME=''
+          if [[ '${{ inputs.cluster_name }}' == '' ]] && [[ '${{ inputs.workflow_trigger }}' == 'soak' ]]; then
```
jonathan-innis, line 108: optional: maybe it's just me but I find this more confusing than the last iteration

.github/workflows/e2e-soak-trigger.yaml
engedaam: Checked `ResourceCount`, and seems to be updated

jonathan-innis (approved):
LGTM 🚀

.github/workflows/e2e.yaml
njtran: I agree. Sorry If I said otherwise, I honestly don't remember, but I think it would look better like the following

```
if cluster_name is unset {
   if soak { cluster_name = soak }
   else { cluster_name = not_soak }
} else {
  cluster_name is existing
}
```

njtran (approved):
LGTM. One optional as well

njtran (dismissed):

test/hack/resource/pkg/resourcetypes/instanceprofile.go
```diff
@@ -47,6 +49,9 @@ func (ip *InstanceProfile) GetExpired(ctx context.Context, expirationTime time.T
 
 	errs := make([]error, len(out.InstanceProfiles))
 	for i := range out.InstanceProfiles {
+		if ip.Global() && !strings.Contains(lo.FromPtr(out.InstanceProfiles[i].Arn), lo.Must(config.LoadDefaultConfig(ctx)).Region) {
```
jonathan-innis, line 52: What is this for?

test/hack/resource/pkg/resourcetypes/oidc.go
```diff
@@ -47,6 +49,9 @@ func (o *OIDC) GetExpired(ctx context.Context, expirationTime time.Time) (names
 
 	errs := make([]error, len(out.OpenIDConnectProviderList))
 	for i := range out.OpenIDConnectProviderList {
+		if o.Global() && !strings.Contains(lo.FromPtr(out.OpenIDConnectProviderList[i].Arn), lo.Must(config.LoadDefaultConfig(ctx)).Region) {
```
jonathan-innis, line 52: What is this for? Don't we know that this resource is global so why do we need to check it within the method?

test/suites/integration/scheduling_test.go
```diff
@@ -416,7 +416,7 @@ var _ = Describe("Scheduling", Ordered, ContinueOnFailure, func() {
 		env.ExpectCreated(pod, nodeClass, nodePoolLowPri, nodePoolHighPri)
 		env.EventuallyExpectHealthy(pod)
 		env.ExpectCreatedNodeCount("==", 1)
-		Expect(ptr.StringValue(env.GetInstance(pod.Spec.NodeName).InstanceType)).To(Equal("c4.large"))
+		Expect(ptr.StringValue(env.GetInstance(pod.Spec.NodeName).InstanceType)).To(Equal("c5.large"))
```
jonathan-innis, line 419: Why are these changes worked in here? How do these apply to the PR?

engedaam: `eu-north-1` does not contain these instance types.

test/hack/resource/pkg/resourcetypes/oidc.go
engedaam: We will have a case were a sweeper running for  us-west-2 can clean up resources meant for eu-north-1 as those resource will be discovered as well.

test/hack/resource/pkg/resourcetypes/instanceprofile.go
engedaam: The cleanup will discover instancesProfiles that are intended for resources outside of sweepers region. This is to scope the clean up to the region the sweeper was intended.

jonathan-innis (approved):
LGTM 🚀

PR #5025, opened 2023-11-06
docs: Fix unmarshalling error of helm chart  README.md

Stuckinrecoverymode:
**Description**
I've made a small change to the README field of the Helm chart. Specifically, I added the missing double quotes to ensure correct formatting.

This change doesn't impact code functionality, so it doesn't require specific testing.

- [ ] No
- [ ] Yes, issue opened: # 


charts/karpenter/README.md
```diff
@@ -16,7 +16,7 @@ You can follow the detailed installation instruction in the [documentation](http
 helm upgrade --install --namespace karpenter --create-namespace \
   karpenter oci://public.ecr.aws/karpenter/karpenter \
   --version v0.32.1 \
-  --set serviceAccount.annotations.eks\.amazonaws\.com/role-arn=${KARPENTER_IAM_ROLE_ARN} \
+  --set serviceAccount.annotations."eks\.amazonaws\.com/role-arn"=${KARPENTER_IAM_ROLE_ARN} \
```
jonathan-innis, line 19:
```suggestion
  --set "serviceAccount.annotations.eks\.amazonaws\.com/role-arn=${KARPENTER_IAM_ROLE_ARN}" \
```

Should we double-quote around the whole thing?

Stuckinrecoverymode: @jonathan-innis well, maybe it will work, but that worked for me so.

Stuckinrecoverymode: I mean double quote only that particular part.

ellistarn: +1 to this recommendation.

ellistarn (approved):

jonathan-innis:
@Stuckinrecoverymode Coming back around this to understand which value for IAM Role caused this to fail.

Just trying to get the justification so we know why we explicitly cast this to a string going forward.

Stuckinrecoverymode:
@jonathan-innis Yes, i think so. i will share my policy so you can look at it.
`{
    "Statement": [
        {
            "Action": [
                "ssm:GetParameter",
                "iam:PassRole",
                "ec2:DescribeImages",
                "ec2:RunInstances",
                "ec2:DescribeSubnets",
                "ec2:DescribeSecurityGroups",
                "ec2:DescribeLaunchTemplates",
                "ec2:DescribeInstances",
                "ec2:DescribeInstanceTypes",
                "ec2:DescribeInstanceTypeOfferings",
                "ec2:DescribeAvailabilityZones",
                "ec2:DeleteLaunchTemplate",
                "ec2:CreateTags",
                "ec2:CreateLaunchTemplate",
                "ec2:CreateFleet",
                "ec2:DescribeSpotPriceHistory",
                "pricing:GetProducts",
                "iam:GetInstanceProfile"
            ],
            "Effect": "Allow",
            "Resource": "*",
            "Sid": "Karpenter"
        },
        {
            "Action": "ec2:TerminateInstances",
            "Condition": {
                "StringEquals": {
                    "ec2:ResourceTag/karpenter.sh/managed-by": "true"
                },
                "StringLike": {
                    "ec2:ResourceTag/Name": "*karpenter*"
                }
            },
            "Effect": "Allow",
            "Resource": "*",
            "Sid": "ConditionalEC2Termination"
        }
    ],
    "Version": "2012-10-17"
}`

Stuckinrecoverymode:
And also i tried with cloudformation stack >> https://raw.githubusercontent.com/aws/karpenter/main/website/content/en/v0.32/getting-started/getting-started-with-karpenter/cloudformation.yaml

Stuckinrecoverymode:
@jonathan-innis  now i got it, I think it is about that regular expression is containing dots, so reason why we used double quote here. Without the quotes, YAML might interpret the dots as a structure separator, causing parsing errors.

jonathan-innis:
> now i got it, I think it is about that regular expression is containing dots, so reason why we used double quote here

Which shell are you using? This doesn't happen in my shell but I suspect there's a difference in interpretation from the various shell implementations.

Stuckinrecoverymode:
> > now i got it, I think it is about that regular expression is containing dots, so reason why we used double quote here
> 
> Which shell are you using? This doesn't happen in my shell but I suspect there's a difference in interpretation from the various shell implementations.

I'm using zsh on linux 6.5.5. Maybe  that is  why we have this error.

jonathan-innis:
Huh, I'm also using `/bin/zsh`. I'm not _exactly_ sure what's the specific difference here, but I think that we should probably be quoting all of our `--set` calls anyways to avoid globbing when we have environment variables: #5061

Stuckinrecoverymode:
Yeah :) sometimes weird things happens.

PR #5037, opened 2023-11-06
test: fix flake for nodeclaim test

njtran:
Fixes #N/A 

**Description**
Add back instance profile creation for Garbage collection test. 

make presubmit

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/pkg/environment/aws/expectations.go
+++ b/test/pkg/environment/aws/expectations.go
@@ -115,23 +115,17 @@ func (env *Environment) ExpectExperimentTemplateDeleted(id string) {
 	Expect(err).ToNot(HaveOccurred())
 }
 
-func (env *Environment) EventuallyExpectInstanceProfileExists(profileName string) {
+func (env *Environment) EventuallyExpectInstanceProfileExists(profileName string) iam.InstanceProfile {
 	GinkgoHelper()
-	By(fmt.Sprintf("expecting instance profile %s to exist", profileName))
-	var instanceProfile iam.InstanceProfile
+	By(fmt.Sprintf("eventually expecting instance profile %s to exist", profileName))
 	Eventually(func(g Gomega) {
-		instanceProfile = env.ExpectInstanceProfileExists(profileName)
+		out, err := env.IAMAPI.GetInstanceProfileWithContext(env.Context, &iam.GetInstanceProfileInput{
+			InstanceProfileName: aws.String(profileName),
+		})
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(out.InstanceProfile).ToNot(BeNil())
+		g.Expect(out.InstanceProfile.InstanceProfileName).ToNot(BeNil())
 	}).WithTimeout(20 * time.Second).Should(Succeed())
-	Expect(instanceProfile.InstanceProfileName).ToNot(BeNil())
-}
-
-func (env *Environment) ExpectInstanceProfileExists(profileName string) iam.InstanceProfile {
-	GinkgoHelper()
-	out, err := env.IAMAPI.GetInstanceProfileWithContext(env.Context, &iam.GetInstanceProfileInput{
-		InstanceProfileName: aws.String(profileName),
-	})
-	Expect(err).ToNot(HaveOccurred())
-	Expect(out.InstanceProfile).ToNot(BeNil())
 	return lo.FromPtr(out.InstanceProfile)
 }
 
@@ -318,7 +312,7 @@ func (env *Environment) GetCustomAMI(amiPath string, versionOffset int) string {
 
 func (env *Environment) ExpectRunInstances(instanceInput *ec2.RunInstancesInput) *ec2.Reservation {
 	GinkgoHelper()
-	env.EventuallyExpectInstanceProfileExists(aws.StringValue(instanceInput.IamInstanceProfile.Name))
+	_ = env.EventuallyExpectInstanceProfileExists(aws.StringValue(instanceInput.IamInstanceProfile.Name))
 	// implement IMDSv2
 	instanceInput.MetadataOptions = &ec2.InstanceMetadataOptionsRequest{
 		HttpEndpoint: aws.String("enabled"),
--- a/test/pkg/environment/aws/expectations.go
+++ b/test/pkg/environment/aws/expectations.go
@@ -312,17 +312,19 @@ func (env *Environment) GetCustomAMI(amiPath string, versionOffset int) string {
 	return *parameter.Parameter.Value
 }
 
-func (env *Environment) ExpectRunInstances(instanceInput *ec2.RunInstancesInput) *ec2.Reservation {
+func (env *Environment) EventuallyExpectRunInstances(instanceInput *ec2.RunInstancesInput) *ec2.Reservation {
 	GinkgoHelper()
 	// implement IMDSv2
 	instanceInput.MetadataOptions = &ec2.InstanceMetadataOptionsRequest{
 		HttpEndpoint: aws.String("enabled"),
 		HttpTokens:   aws.String("required"),
 	}
-
-	out, err := env.EC2API.RunInstances(instanceInput)
-	Expect(err).ToNot(HaveOccurred())
-
+	var out *ec2.Reservation
+	var err error
+	Eventually(func(g Gomega) {
+		out, err = env.EC2API.RunInstances(instanceInput)
+		g.Expect(err).ToNot(HaveOccurred())
+	}).WithTimeout(5 * time.Minute).Should(Succeed())
 	return out
 }
 
--- a/test/pkg/environment/aws/expectations.go
+++ b/test/pkg/environment/aws/expectations.go
@@ -324,7 +324,7 @@ func (env *Environment) EventuallyExpectRunInstances(instanceInput *ec2.RunInsta
 	Eventually(func(g Gomega) {
 		out, err = env.EC2API.RunInstances(instanceInput)
 		g.Expect(err).ToNot(HaveOccurred())
-	}).WithTimeout(5 * time.Minute).Should(Succeed())
+	}).WithTimeout(30 * time.Second).WithPolling(5 * time.Second).Should(Succeed())
 	return out
 }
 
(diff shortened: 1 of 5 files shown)


test/suites/nodeclaim/garbage_collection_test.go
```diff
@@ -147,3 +150,44 @@ var _ = Describe("NodeClaimGarbageCollection", func() {
 		env.EventuallyExpectNotFound(node)
 	})
 })
+
+func ExpectInstanceProfileCreated(instanceProfileName *string) {
```
jonathan-innis, line 154:
```suggestion
func ExpectInstanceProfileCreated(name string) {
```

Why does it need to be a string pointer?

test/suites/nodeclaim/garbage_collection_test.go
```diff
@@ -147,3 +150,44 @@ var _ = Describe("NodeClaimGarbageCollection", func() {
...
+	By("creating an instance profile")
+	createInstanceProfile := &iam.CreateInstanceProfileInput{
+		InstanceProfileName: instanceProfileName,
+		Tags: []*iam.Tag{
```
jonathan-innis, line 158: Is there a common set of tags that we should add to all of our resources outside of our common tags?

test/suites/nodeclaim/garbage_collection_test.go
```diff
@@ -147,3 +150,44 @@ var _ = Describe("NodeClaimGarbageCollection", func() {
...
+		InstanceProfileName: instanceProfileName,
+		RoleName:            aws.String(fmt.Sprintf("KarpenterNodeRole-%s", env.ClusterName)),
+	}
+	_, err = env.IAMAPI.AddRoleToInstanceProfile(addInstanceProfile)
+	Expect(ignoreAlreadyContainsRole(err)).ToNot(HaveOccurred())
+}
+
+func ignoreAlreadyExists(err error) error {
+	if err != nil {
+		if strings.Contains(err.Error(), "EntityAlreadyExists") {
```
jonathan-innis, line 178: Does IAM already have a constant for this value? Is there room here to convert this to the aws-typed error like we do in the errors package? I'm not against using a string comparison, but it's probably more correct to do an error conversion into the typed thing

test/suites/nodeclaim/garbage_collection_test.go
```diff
@@ -147,3 +150,44 @@ var _ = Describe("NodeClaimGarbageCollection", func() {
...
+			return nil
+		}
+	}
+	return err
+}
+
+func ignoreAlreadyContainsRole(err error) error {
+	if err != nil {
+		if strings.Contains(err.Error(), "Cannot exceed quota for InstanceSessionsPerInstanceProfile") {
```
jonathan-innis, line 187: Same comment here. Should we have handling that types the error?

test/suites/nodeclaim/garbage_collection_test.go
```diff
@@ -86,13 +88,14 @@ var _ = Describe("NodeClaimGarbageCollection", func() {
 			MaxCount: aws.Int64(1),
 		}
 	})
-	It("should succeed to garbage collect an Instance that was launched by a NodeClaim but has no Instance mapping", func() {
+	FIt("should succeed to garbage collect an Instance that was launched by a NodeClaim but has no Instance mapping", func() {
```
jonathan-innis, line 91:
```suggestion
	It("should succeed to garbage collect an Instance that was launched by a NodeClaim but has no Instance mapping", func() {
```

test/pkg/environment/aws/expectations.go
```diff
@@ -115,24 +115,20 @@ func (env *Environment) ExpectExperimentTemplateDeleted(id string) {
 	Expect(err).ToNot(HaveOccurred())
 }
 
-func (env *Environment) EventuallyExpectInstanceProfileExists(profileName string) {
+func (env *Environment) EventuallyExpectInstanceProfileExists(profileName string) iam.InstanceProfile {
```
jonathan-innis, line 118: Does this still need to be an Eventually check or can we be assured that the creation of the instance profile is effectively synchronous and we shouldn't see issues with calls to RunInstances

test/suites/nodeclaim/garbage_collection_test.go
njtran: What tags were you thinking? All we need is the one tag to discover for cleanup, since this is only being used for the one-off.

test/pkg/environment/aws/expectations.go
njtran: The reason I cut the initial PR was that there's a chance for there to be a delay. I don't think an eventually hurts here, just makes it more likely that we don't fail from a fleeting error.

test/suites/nodeclaim/garbage_collection_test.go
```diff
@@ -93,6 +95,13 @@ var _ = Describe("NodeClaimGarbageCollection", func() {
 		instanceInput.UserData = lo.ToPtr(base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf(string(rawContent), env.ClusterName,
 			env.ClusterEndpoint, env.ExpectCABundle(), nodePool.Name))))
 
+		instanceProfileName := fmt.Sprintf("KarpenterNodeInstanceProfile-%s", env.ClusterName)
+		ExpectInstanceProfileCreated(instanceProfileName)
+		// Sleep 10 seconds to allow the instance profile to handle eventual consistency delays.
+		time.Sleep(10 * time.Second)
```
jonathan-innis, line 101: Do we need this now that we have an eventually consistent check for existence?

test/suites/nodeclaim/garbage_collection_test.go
jonathan-innis: I was mainly curious if we had any other tags to at we were adding. If not, then this is fine with me

test/suites/nodeclaim/garbage_collection_test.go
```diff
@@ -147,3 +156,58 @@ var _ = Describe("NodeClaimGarbageCollection", func() {
...
+	return err
+}
+
+func ignoreAlreadyContainsRole(err error) error {
+	if err != nil {
+		if strings.Contains(err.Error(), "Cannot exceed quota for InstanceSessionsPerInstanceProfile") {
+			return nil
+		}
+	}
+
```
jonathan-innis, line 211: Cut

test/suites/nodeclaim/garbage_collection_test.go
```diff
@@ -147,3 +156,58 @@ var _ = Describe("NodeClaimGarbageCollection", func() {
...
+	Expect(err).To(BeNil())
+
+	deleteInstanceProfile := &iam.DeleteInstanceProfileInput{
+		InstanceProfileName: aws.String(instanceProfileName),
+	}
+	_, err = env.IAMAPI.DeleteInstanceProfile(deleteInstanceProfile)
+	Expect(ignoreAlreadyExists(err)).ToNot(HaveOccurred())
+}
+
+func ignoreAlreadyExists(err error) error {
```
jonathan-innis, line 198: Worth it to create this method in the errors package?

test/suites/nodeclaim/garbage_collection_test.go
```diff
@@ -147,3 +156,58 @@ var _ = Describe("NodeClaimGarbageCollection", func() {
...
+		RoleName:            aws.String(fmt.Sprintf("KarpenterNodeRole-%s", env.ClusterName)),
+	}
+	_, err := env.IAMAPI.RemoveRoleFromInstanceProfile(removeRoleFromInstanceProfile)
+	Expect(err).To(BeNil())
+
+	deleteInstanceProfile := &iam.DeleteInstanceProfileInput{
+		InstanceProfileName: aws.String(instanceProfileName),
+	}
+	_, err = env.IAMAPI.DeleteInstanceProfile(deleteInstanceProfile)
+	Expect(ignoreAlreadyExists(err)).ToNot(HaveOccurred())
```
jonathan-innis, line 195: Should we ignore this in the delete scenario? I'm assuming that we should be ignoring if it doesn't exist, not if it already exists

test/suites/nodeclaim/garbage_collection_test.go
```diff
@@ -147,3 +156,58 @@ var _ = Describe("NodeClaimGarbageCollection", func() {
...
+	Expect(ignoreAlreadyContainsRole(err)).ToNot(HaveOccurred())
+}
+
+func ExpectInstanceProfileDeleted(instanceProfileName string) {
+	By("deleting an instance profile")
+	removeRoleFromInstanceProfile := &iam.RemoveRoleFromInstanceProfileInput{
+		InstanceProfileName: aws.String(instanceProfileName),
+		RoleName:            aws.String(fmt.Sprintf("KarpenterNodeRole-%s", env.ClusterName)),
+	}
+	_, err := env.IAMAPI.RemoveRoleFromInstanceProfile(removeRoleFromInstanceProfile)
```
jonathan-innis, line 188: Should we ignore NotExists here?

test/suites/nodeclaim/garbage_collection_test.go
```diff
@@ -93,6 +95,13 @@ var _ = Describe("NodeClaimGarbageCollection", func() {
 		instanceInput.UserData = lo.ToPtr(base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf(string(rawContent), env.ClusterName,
 			env.ClusterEndpoint, env.ExpectCABundle(), nodePool.Name))))
 
+		instanceProfileName := fmt.Sprintf("KarpenterNodeInstanceProfile-%s", env.ClusterName)
+		ExpectInstanceProfileCreated(instanceProfileName)
+		// Sleep 10 seconds to allow the instance profile to handle eventual consistency delays.
+		time.Sleep(10 * time.Second)
+		DeferCleanup(func() {
+			defer ExpectInstanceProfileDeleted(instanceProfileName)
+		})
```
ellistarn, line 104: These deferred cleanup methods seem a bit scary and leak prone. When does this block execute?

njtran: > When DeferCleanup is called in BeforeEach, JustBeforeEach, It, AfterEach, or JustAfterEach the registered callback will be invoked when the spec completes (i.e. it will behave like an AfterEach node)

jonathan-innis (approved):
LGTM

PR #5038, opened 2023-11-06
ci: Use `mattn/goveralls` instead of `shogo821/goveralls`

engedaam:
Fixes #N/A 

**Description**
- Update to use https://github.com/mattn/goveralls directly

- `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -16,6 +16,7 @@ jobs:
       with:
         k8sVersion: ${{ matrix.k8sVersion }}
     - run: K8S_VERSION=${{ matrix.k8sVersion }} make ci-test
-    - uses: shogo82148/actions-goveralls@v1
-      with:
-        path-to-profile: coverage.out
+    - name: Send coverage
+      env:
+        COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-test.yaml
```diff
@@ -16,6 +16,9 @@ jobs:
       with:
         k8sVersion: ${{ matrix.k8sVersion }}
     - run: K8S_VERSION=${{ matrix.k8sVersion }} make ci-test
-    - uses: shogo82148/actions-goveralls@v1
-      with:
-        path-to-profile: coverage.out
+    - name: Install goveralls
+      run: go install github.com/mattn/goveralls@latest
```
njtran, line 20: Can we install this as part of `.github/actions/install-deps`? Is there a good reason to do latest here rather than pinning a version?

njtran (commented):
Should we do this as part of Karpenter-core too?

.github/workflows/ci-test.yaml
jonathan-innis: Same question, should this be part of our `toolchain.sh`?

ellistarn: +1 on toolchain

njtran (approved):

PR #5068, opened 2023-11-11
feat: support `vpc.amazonaws.com/efa` resource

### LINKED ISSUE #3127: Add support for Elastic Fabric Adapter (EFA) to Karpenter
Problem Statement:

Some workloads [distributed training, simulations, HPC applications] require high performance networking on AWS provided by instances that are enabled with [Elastic Fabric Adapter](https://aws.amazon.com/hpc/efa/). The specific instance types are documented [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa.html#efa-instance-types). Currently Karpenter does not recognize EFA resource requests or limits specified in Kubernetes manifests as described below.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: inflate
spec:
  replicas: 1
  selector:
    matchLabels:
      app: inflate
  template:
    metadata:
      labels:
        app: inflate
    spec:
      terminationGracePeriodSeconds: 0
      containers:
        - name: inflate
          image: public.ecr.aws/eks-distro/kubernetes/pause:3.2
          resources:
            requests:
              vpc.amazonaws.com/efa: 1
            limits:
              vpc.amazonaws.com/efa: 1
```

When such a manifest is applied to a Karpenter-enabled cluster, the Karpenter controller produces an error like the following:

```log
ERROR   controller.provisioning Could not schedule pod, incompatible with provisioner "default", no instance type satisfied resources {"pods":"1","vpc.amazonaws.com/efa":"1"} and requirements karpenter.k8s.aws/instance-generation Exists >2, karpenter.sh/provisioner-name In [default], karpenter.sh/capacity-type In [on-demand spot], kubernetes.io/os In [linux], kubernetes.io/arch In [amd64], karpenter.k8s.aws/instance-category In [c m r]   {"commit": "f290d37-dirty", "pod": "default/inflate-fd9bc9f9b-xmkq2"}
```

Feature Request:

Add capability in Karpenter to recognize resource `vpc.amazonaws.com/efa`, identify, and provision a suitable EC2 instance type with EFA enabled.

jmdeal:
Fixes #3127 

**Description**
This PR adds support for the `vpc.amazonaws.com/efa` resource. No additional configuration in either the `NodePool` or `NodeClass` is required, if an unschedulable pod is requesting the resource, Karpenter will launch an EC2 instance with all possible EFAs configured.

`make test` & `make e2etest` (TODO)

**TODO**
- [ ] add integration tests
- [ ] add tests for launch template configuration

- [ ] Yes, issue opened: # 
- [x] No


pkg/providers/instance/types.go
```diff
@@ -53,11 +54,14 @@ func NewInstance(out *ec2.Instance) *Instance {
 		}),
 		SubnetID: aws.StringValue(out.SubnetId),
 		Tags:     lo.SliceToMap(out.Tags, func(t *ec2.Tag) (string, string) { return aws.StringValue(t.Key), aws.StringValue(t.Value) }),
+		EFAEnabled: lo.ContainsBy(out.NetworkInterfaces, func(ni *ec2.InstanceNetworkInterface) bool {
+			return ni != nil && lo.FromPtr(ni.InterfaceType) == "efa"
```
jonathan-innis, line 58: Can we check if `efa` is a constant value that's represented somewhere by the `ec2` package. If not, can we consider creating our own constant?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -307,8 +307,14 @@ func (c *CloudProvider) instanceToNodeClaim(i *instance.Instance, instanceType *
 				labels[key] = req.Values()[0]
 			}
 		}
-		nodeClaim.Status.Capacity = functional.FilterMap(instanceType.Capacity, func(_ v1.ResourceName, v resource.Quantity) bool { return !resources.IsZero(v) })
-		nodeClaim.Status.Allocatable = functional.FilterMap(instanceType.Allocatable(), func(_ v1.ResourceName, v resource.Quantity) bool { return !resources.IsZero(v) })
+		resourceFilter := func(n v1.ResourceName, v resource.Quantity) bool {
```
jonathan-innis, line 310: Consider a comment describing the purpose of this filter

pkg/cloudprovider/cloudprovider.go
```diff
@@ -307,8 +307,14 @@ func (c *CloudProvider) instanceToNodeClaim(i *instance.Instance, instanceType *
...
+			if !i.EFAEnabled && n == v1beta1.ResourceEFA {
```
jonathan-innis, line 311: I could go either way, but I personally find the below filter a bit clearer
```go
if resources.IsZero(v) {
  return false
}
if n == v1beta1.ResourceEFA {
  return i.EFAEnabled
}
return true
```

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -286,6 +286,18 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, capacityType string
 // This is done to help comply with AWS account policies that require explicitly setting that field to 'false'.
 // https://github.com/aws/karpenter/issues/3815
 func (p *Provider) generateNetworkInterface(options *amifamily.LaunchTemplate) []*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest {
+	if options.EFACount != 0 {
+		return lo.Times(options.EFACount, func(i int) *ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest {
+			return &ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest{
+				// You may only associate a public IP if there is a single network interface.
```
jonathan-innis, line 292: What's this case here? Trying to understand the logic for handling this edge case

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -286,6 +286,18 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, capacityType string
...
+				AssociatePublicIpAddress: lo.Ternary(options.EFACount == 1, options.AssociatePublicIPAddress, nil),
+				NetworkCardIndex:         lo.ToPtr(int64(i)),
+				DeviceIndex:              lo.ToPtr(lo.Ternary[int64](i == 0, 0, 1)),
```
jonathan-innis, line 295: Comment to describe this? Is this what delineates whether this is the primary or secondary ENI?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -286,6 +286,18 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, capacityType string
...
+				InterfaceType:            lo.ToPtr("efa"),
```
jonathan-innis, line 296: Is there a constant for this value?

pkg/providers/amifamily/resolver.go
```diff
@@ -130,21 +131,32 @@ func (r Resolver) Resolve(ctx context.Context, nodeClass *v1beta1.EC2NodeClass,
 	}
 	var resolvedTemplates []*LaunchTemplate
 	for amiID, instanceTypes := range mappedAMIs {
-		maxPodsToInstanceTypes := lo.GroupBy(instanceTypes, func(instanceType *cloudprovider.InstanceType) int {
-			return int(instanceType.Capacity.Pods().Value())
+		type launchTemplateParams struct {
+			efaCount int
+			maxPods  int
+		}
+		paramsToInstanceTypes := lo.GroupBy(instanceTypes, func(instanceType *cloudprovider.InstanceType) launchTemplateParams {
```
jonathan-innis, line 138: How much does this increase the number of launch templates that we create in the worst-case? Would be good to do some analysis on how much this is going to increase our TPS toward `CreateLaunchTempmlate`

pkg/providers/amifamily/resolver.go
```diff
@@ -130,21 +131,32 @@ func (r Resolver) Resolve(ctx context.Context, nodeClass *v1beta1.EC2NodeClass,
 	}
 	var resolvedTemplates []*LaunchTemplate
 	for amiID, instanceTypes := range mappedAMIs {
```
jonathan-innis, line 133: Is it worth breaking out some of this logic into a separate function now that the launch template generation logic has gotten more complicated?

test/suites/integration/extended_resources_test.go
```diff
@@ -208,6 +208,48 @@ var _ = Describe("Extended Resources", func() {
...
+					Labels: map[string]string{"app": "efa-app"},
+				},
+				Tolerations: []v1.Toleration{
+					{
+						Key:      "aws.amazon.com/efa",
+						Operator: v1.TolerationOpExists,
+					},
+				},
+				ResourceRequirements: v1.ResourceRequirements{
+					Requests: v1.ResourceList{
```
jonathan-innis, line 238: Any to reason to have a test that requests more than one EFA? Are there any edge-cases that we should make sure that we test when we request multiple EFAs which would launch multiple network interfaces?

pkg/providers/instance/types.go
jmdeal: I don't believe it is but I'll verify. Where do you think it would be best for this constant to live if we define it ourself?

test/suites/integration/extended_resources_test.go
jmdeal: Good call, there should be one to verify that multiple network card launch templates are created correctly.

pkg/providers/instance/types.go
jonathan-innis: Maybe next to the resource type in the `apis` directory? Seems reasonable to me for these to live alongside each other for now until there's a more natural home for it long-term.

jmdeal (commented):
/karpenter snapshot

pkg/providers/launchtemplate/launchtemplate.go
jmdeal: Not really. This solves a bit of a nuanced issue that can cause traffic leaving the VPC to be dropped. Let's say we have an instance with four network cards and we want to send a packet to cloudflare DNS (1.1.1.1). If all of them are on device index 0 a random ENI will be selected for the source address and if it doesn't have a public IP that packet will get dropped. One solution to this would be to associate a public IP with each network interface (not great), or by changing the device index for all but one we ensure that the primary ENI has the highest priority. That's my understanding at the very least. I'll leave a comment but I'll try and keep it briefer than this, probably something to the extent of "some networking black magic to ensure that public traffic is correctly routed".

pkg/providers/amifamily/resolver.go
jonathan-innis: I'd suspect that we're roughly doubling our launch template count here

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -286,6 +286,19 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, capacityType string
 // This is done to help comply with AWS account policies that require explicitly setting that field to 'false'.
 // https://github.com/aws/karpenter/issues/3815
 func (p *Provider) generateNetworkInterface(options *amifamily.LaunchTemplate) []*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest {
```
jonathan-innis, line 288:
```suggestion
func (p *Provider) generateNetworkInterfaces(options *amifamily.LaunchTemplate) []*ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest {
```

nit

jmdeal (commented):
/karpenter snapshot

pkg/providers/launchtemplate/launchtemplate.go
jmdeal: Discussed some more offline, we've agreed to drop this field for the time being for EFAs.

pkg/providers/amifamily/resolver.go
jmdeal: Threw together a little script to test this and here are the results:
```
# With EFAs
EFA Count: 0, Total Max Pod Values: 21, Max Pod Values: [44 437 59 35 10 78 118 394 17 20 234 737 29 18 38 89 58 4 8 12 11]
EFA Count: 1, Total Max Pod Values: 6, Max Pod Values: [737 198 234 58 321 345]
EFA Count: 2, Total Max Pod Values: 1, Max Pod Values: [345]
EFA Count: 4, Total Max Pod Values: 1, Max Pod Values: [737]
EFA Count: 8, Total Max Pod Values: 1, Max Pod Values: [247]
EFA Count: 16, Total Max Pod Values: 1, Max Pod Values: [247]
EFA Count: 32, Total Max Pod Values: 1, Max Pod Values: [100]
Total Possible Launch Templates: 32
```
```
# Without EFAs
Total Max Pod Values: 26, Max Pod Values: [58 89 737 29 198 118 8 234 12 4 78 18 247 17 10 59 35 11 321 20 100 394 345 44 437 38]
Total Possible Launch Templates: 26
```
We already discussed this offline but the increase of 6 is pretty negligible. We've got a burst limit of 200 requests for create calls. This is based off instances available in `us-east-1` but if anything there will be less instance types in other regions (generally).

jmdeal (commented):
/karpenter snapshot

jonathan-innis (dismissed):
LGTM 🚀 Nice work!

jonathan-innis (approved):
LGTM 🚀

jmdeal (commented):
/karpenter snapshot

PR #5087, opened 2023-11-14
ci: Move prometheus and karpenter install to create_cluster action

engedaam:
Fixes #N/A 

**Description**
- Move installing prometheus and karpenter in create cluster for better abstraction 

- Manually tested in a forked repo

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.github/actions/e2e/create-cluster/action.yaml
+++ b/.github/actions/e2e/create-cluster/action.yaml
@@ -4,11 +4,23 @@ inputs:
   account_id:
     description: "Account ID to access AWS"
     required: true
+  ecr_account_id:
+    description: "Account ID to access ECR Repository"
+    required: true
+  prometheus_workspace_id:
+    description: Workspace ID for the Prometheus workspace
+    required: true
   role:
     description: "Role to access AWS"
     required: true
   region:
     description: "Region to access AWS"
+    required: true
+  ecr_region:
+    description: "Region to access ECR Repository"
+    required: true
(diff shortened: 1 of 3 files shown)


.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -4,12 +4,24 @@ inputs:
...
     required: true
   region:
     description: "Region to access AWS"
     required: true
+  ecr_region:
+    description: "Region to access ECR Repository"
+    required: true
+  prometheus_region:
+    description: Region to access Prometheus 
+    required: true
```
njtran, line 24: Is this because we have the prometheus in one region but run tests in multiple?

.github/workflows/e2e.yaml
```diff
@@ -95,6 +95,7 @@ jobs:
           # Creating jitter so that we can stagger cluster creation to avoid throttling
           sleep $(( RANDOM % 300 + 1 ))
       - id: generate-cluster-name
+        name: Generate Cluster Name
```
njtran, line 98: I'd follow what I see elsewhere in the file which is all lowercase, or change them all to upper case :)
```suggestion
        name: generate cluster name
```

njtran: Also worth making the same change in e2e-upgrade if you're going to make it.

njtran (approved):

PR #5098, opened 2023-11-15
ci: Add Resource count to Timestream

engedaam:
Fixes #N/A 

**Description**
- To better track our resources usage, a workflow will count the number of resource in our account over a period of time

- Manually tested

- [ ] Yes, issue opened: # 
- [x] No


test/hack/resource-management/clean-resources/main.go
```diff
@@ -32,6 +32,7 @@ import (
 )
 
 const expirationTTL = time.Hour * 12
+const karpenterMetricTableName = "sweeperCleanedResources"
```
njtran, line 35: consider naming the variable to be the same as the variable value
```suggestion
const sweeperCleanedResourcesTableName = "sweeperCleanedResources"
```

test/hack/resource-management/count-resources/main.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/iam"
+	"github.com/samber/lo"
+	"go.uber.org/zap"
+
+	"github.com/aws/karpenter/test/hack/cleanup/metrics"
+	"github.com/aws/karpenter/test/hack/cleanup/resourcetypes"
+)
+
+const karpenterMetricTableName = "resourceCount"
```
njtran, line 32: Same here

test/hack/resource-management/count-resources/main.go
njtran, line 1: Have we considered doing this through the AWS SDK? Looks like there's something like this here: https://docs.aws.amazon.com/cli/latest/reference/configservice/list-aggregate-discovered-resources.html

test/hack/resource-management/resourcetypes/eni.go
njtran, line 1: Seems like we're going to be grabbing all resources with the "tag-key" with k8sClusterTag. Let's see if https://docs.aws.amazon.com/cli/latest/reference/configservice/list-aggregate-discovered-resources.html can do this so that we don't have to make a new file with a new resource count every time we want to track a new resource.

test/hack/resource-management/resourcetypes/instance.go
```diff
@@ -71,6 +71,39 @@ func (i *Instance) GetExpired(ctx context.Context, expirationTime time.Time) (id
...
+func (i *Instance) GetCount(ctx context.Context) (count int, err error) {
+	var nextToken *string
+
+	for {
+		out, err := i.ec2Client.DescribeInstances(ctx, &ec2.DescribeInstancesInput{
+			Filters: []ec2types.Filter{
+				{
+					Name:   lo.ToPtr("instance-state-name"),
+					Values: []string{string(ec2types.InstanceStateNameRunning)},
+				},
```
njtran, line 83: Do we want other instances too? What about Pending, Shutting-down, and stopped? 
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html

test/hack/resource-management/resourcetypes/instanceprofile.go
```diff
@@ -64,6 +64,15 @@ func (ip *InstanceProfile) GetExpired(ctx context.Context, expirationTime time.T
 	return names, multierr.Combine(errs...)
 }
 
+func (ip *InstanceProfile) GetCount(ctx context.Context) (count int, err error) {
+	out, err := ip.iamClient.ListInstanceProfiles(ctx, &iam.ListInstanceProfilesInput{})
```
njtran, line 68: Do we need pagination?

test/hack/resource-management/resourcetypes/launchtemplate.go
```diff
@@ -66,6 +66,32 @@ func (lt *LaunchTemplate) GetExpired(ctx context.Context, expirationTime time.Ti
 	return names, err
 }
 
+func (lt *LaunchTemplate) GetCount(ctx context.Context) (count int, err error) {
+	var nextToken *string
+	for {
+		out, err := lt.ec2Client.DescribeLaunchTemplates(ctx, &ec2.DescribeLaunchTemplatesInput{
+			Filters: []ec2types.Filter{
```
njtran, line 73: Do we need to add tag filters on the testing tag?

engedaam (commented):
/karpenter snapshot

test/hack/resource-management/count-resources/main.go
engedaam: Considered using AWS Config, however the cost of the service out ways any benefit that is made on monitoring resources.

.github/workflows/resource-count-generator.yaml
```diff
@@ -0,0 +1,30 @@
+name: ResourceCountGenerator
+on:
+  schedule:
+    - cron: '*/30 * * * *' # every 30 minute
```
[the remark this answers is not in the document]
engedaam: Pushed it to every hour

test/hack/resource-management/pkg/metrics/metrics.go
```diff
@@ -25,13 +25,12 @@ import (
...
-	karpenterMetricRegion    = "us-east-2"
-	karpenterMetricDatabase  = "karpenterTesting"
-	karpenterMetricTableName = "sweeperCleanedResources"
+	karpenterMetricRegion   = "us-east-2"
+	karpenterMetricDatabase = "karpenterTesting"
 )
 
 type Client interface {
-	FireMetric(context.Context, string, float64, string) error
+	FireMetric(context.Context, string, string, float64, string) error
```
jonathan-innis, line 33: Can we name these parameters now? It's getting a bit difficult to reason about what parameter is what

test/hack/resource-management/pkg/resourcetypes/resourcetypes.go
```diff
@@ -42,4 +42,6 @@ type Type interface {
 	// In general, if all resources can't be deleted by id with a single API call (like with DeleteInstances)
 	// you should call the requests synchronously to avoid rate limiting against the number of requests made
 	Cleanup(ctx context.Context, ids []string) (cleaned []string, err error)
+	// GetCount returns the number of resource of the type in the account
+	GetCount(ctx context.Context) (count int, err error)
```
jonathan-innis, line 46: nit: consider naming this `CountAll()` to convey that we are counting _all_ of the resources in the account, and not those just associated with testing

test/hack/resource-management/go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/aws/karpenter/test/hack/cleanup
+module github.com/aws/karpenter/test/hack/resource-management
```
jonathan-innis, line 1: There's a little write-up on naming go packages and modules: https://go.dev/blog/package-names. I'd agree that naming is hard here, but I think the second you are throwing hyphens and underscores into the mix, you could probably shorten or rename. 

Consider the following renames:
1. Module: `github.com/aws/karpenter/test/hack/resources`
2. Cleanup Script: `github.com/aws/karpenter/test/hack/resources/clean`
3. Count Script: `github.com/aws/karpenter/test/hack/resources/count`

.github/workflows/resource-count-generator.yaml
```diff
@@ -0,0 +1,30 @@
+name: ResourceCountGenerator
```
jonathan-innis, line 1:
```suggestion
name: ResourceCount
```

nit: Do we need to throw the word generator in there?

PR #5115, opened 2023-11-19
ci: pin gha versions to hashes

### LINKED ISSUE #5112: Pin all third-party GHA to a static hash
### Description

Github Action best-practices recommend that you pin third-party actions to a static hash for security hardening and then ref the exact tag of that action over in the comment to that static hash: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions. 

An example of how you can configure this can be seen here: https://github.com/aws/karpenter-core/blob/main/.github/workflows/presubmit.yaml#L16.

We should pin across all of our third-party actions, including those that are in the composite actions directly inside of the `.github/actions` directory.

jmdeal:
Fixes #5112

**Description**
This PR pins action versions to specific commits SHAs as per [GitHub's security recommendations](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

`make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,18 +4,18 @@ on:
     tags: [ 'v*.*.*' ]
 permissions:
   id-token: write # aws-actions/configure-aws-credentials@v4.0.1
-  contents: write # marvinpinto/action-automatic-releases@latest
+  contents: write # marvinpinto/action-automatic-releases@v1.2.1
   pull-requests: write # name: Create PR
 jobs:
   release:
     if: github.repository == 'aws/karpenter'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
         with:
           fetch-depth: 0
       - name: Create GitHub Release
-        uses: 'marvinpinto/action-automatic-releases@latest'
+        uses: marvinpinto/action-automatic-releases@919008cf3f741b179569b7a6fb4d8860689ab7f0 # v1.2.1
         with:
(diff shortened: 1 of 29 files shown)


jmdeal (commented):
/karpenter snapshot

.github/workflows/release.yaml
```diff
@@ -11,11 +11,11 @@ jobs:
     if: github.repository == 'aws/karpenter'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
         with:
           fetch-depth: 0
       - name: Create GitHub Release
-        uses: 'marvinpinto/action-automatic-releases@latest'
+        uses: marvinpinto/action-automatic-releases@919008cf3f741b179569b7a6fb4d8860689ab7f0 # latest
```
jonathan-innis, line 18: Let's use a pinned version rather than something that's on latest that will change over time

ellistarn:
Possible to configure dependabot to automate this for us, and let it configure this?

.github/workflows/release.yaml
jmdeal: I was hesitant to downgrade since the last tagged release (v1.2.1) was from over two years ago and there have been several fixes since. I suppose we'll see in the karpenter-core release if there are any issues with v1.2.1 though and it should be easy enough to revert if there are.

jmdeal:
> Possible to configure dependabot to automate this for us, and let it configure this?

Dependabot should still update pinned versions and should update the tag comment for us: https://github.com/dependabot/dependabot-core/issues/4691

jonathan-innis (approved):
LGTM 🚀

PR #5122, opened 2023-11-20
fix: match LabelDomainExceptions correctly as label domains

splichy:
Fixes #N/A

**Description**
Match `node-restriction.kubernetes.io` as the label domain instead of the exact match.

update launchtemplate.go to remove `[*].node-restriction.kubernetes.io` from user-data - now it matches only labels that start with `node-restriction.kubernetes.io`.

Requires: https://github.com/aws/karpenter-core/pull/808


pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -156,7 +156,8 @@ func (p *Provider) createAMIOptions(ctx context.Context, nodeClass *v1beta1.EC2N
 	// Remove any labels passed into userData that are prefixed with "node-restriction.kubernetes.io" since the kubelet can't
 	// register the node with any labels from this domain: https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction
 	for k := range labels {
-		if strings.HasPrefix(k, v1.LabelNamespaceNodeRestriction) {
+		labelDomain := corev1beta1.GetLabelDomain(k)
+		if strings.HasSuffix(labelDomain, v1.LabelNamespaceNodeRestriction) {
```
jonathan-innis, line 160: I'm going to extend this to check for the `kops.k8s.io` label exception. Looks like we should be filtering this one out as well but we aren't doing that today

test/suites/integration/scheduling_test.go
```diff
@@ -270,10 +270,12 @@ var _ = Describe("Scheduling", Ordered, ContinueOnFailure, func() {
 		test.ReplaceRequirements(nodePool,
```
jonathan-innis, line 270: Going to add scheduling tests for _all_ of the label domain exceptions rather than just this single one

pkg/providers/launchtemplate/launchtemplate.go
splichy: :+1: good point

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (approved):
LGTM 🚀

jmdeal (approved):
LGTM 🚀

engedaam (commented):
LGTM 🚀

PR #5142, opened 2023-11-22
docs: remove containerRuntime from examples as it was dropped

lordz-md:
Fixes #N/A 

**Description**

- [ ] Yes, issue opened: # 
- [ ] No


chrisnegus (commented):
There are notes after each of the places where you cut `containerRuntime` that say: "You can specify the container runtime to be either `dockerd` or `containerd`. By default, `containerd` is used." Should you remove those notes if you can no longer specify using anything other than containerd?

website/content/en/docs/concepts/nodepools.md
```diff
@@ -92,7 +92,6 @@ spec:
       # These are all optional and provide support for additional customization and use cases.
```
jonathan-innis, line 92: Can you make sure this update is made in the `/preview` path as well? This is how we are actually carrying docs changes forward when we release a new minor version

lordz-md: @jonathan-innis done

lordz-md:
> There are notes after each of the places where you cut `containerRuntime` that say: "You can specify the container runtime to be either `dockerd` or `containerd`. By default, `containerd` is used." Should you remove those notes if you can no longer specify using anything other than containerd?

Fixed with a note about `spec.containerRuntime` being dropped

website/content/en/preview/concepts/nodepools.md
```diff
@@ -283,7 +281,7 @@ kubelet:
   maxPods: 20
 ```
 
-You can specify the container runtime to be either `dockerd` or `containerd`. By default, `containerd` is used.
+* `containerd` is the default container runtime on the latest EKS versions, the `spec.template.spec.kubelet.containerRuntime` field was dropped.
```
jonathan-innis, line 284: Looking at the docs; considering we have dropped the field entirely, I don't see a point in calling out the container runtime anymore. I think it's fine to drop this asterisk statement as well as the next one that says "`containerd` is the only valid container runtime..." WDYT?

lordz-md: @jonathan-innis fixed

website/content/en/v0.32/concepts/nodepools.md
```diff
@@ -283,8 +281,6 @@ kubelet:
   maxPods: 20
 ```
 
-You can specify the container runtime to be either `dockerd` or `containerd`. By default, `containerd` is used.
-
 * `containerd` is the only valid container runtime when using the `Bottlerocket` AMIFamily or when using Kubernetes version 1.24+ and the `AL2`, `Windows2019`, or `Windows2022` AMIFamilies.
```
jonathan-innis, line 284: Can we drop this bullet here too?

lordz-md: @jonathan-innis not sure I get it, you mean removing the whole
`* `containerd` is the only valid container runtime when using the `Bottlerocket` AMIFamily or when using Kubernetes version 1.24+ and the `AL2`, `Windows2019`, or `Windows2022` AMIFamilies.` block?

jonathan-innis: Yep, since you can't specify the field at all now, I don't think that it makes sense to have mention of it.

lordz-md: > Yep, since you can't specify the field at all now, I don't think that it makes sense to have mention of it.

LOL, I initially did that and then reverted back to. Fixed

jonathan-innis (approved):
LGTM 🚀

jonathan-innis:
Thanks for pushing this update through @lordz-md! Nice work 🎉

lordz-md:
> Thanks for pushing this update through @lordz-md! Nice work 🎉

Thanks @jonathan-innis.

Issue #5171, opened 2023-11-27
On-Demand/Spot Ratio Split now working (always get spot only)


xubofei1983:
### Description

**Observed Behavior**:
I have two nodepools:
```
apiVersion: karpenter.sh/v1beta1
kind: NodePool
metadata:
  annotations:
    karpenter.sh/nodepool-hash: "2808749311332697570"
    meta.helm.sh/release-name: karpenter-config
    meta.helm.sh/release-namespace: default
  creationTimestamp: "2023-11-27T20:16:44Z"
  generation: 1
  labels:
    app.kubernetes.io/managed-by: Helm
  name: default-workflow-spot
  resourceVersion: "89758"
  uid: bc4fd30d-3b48-4d08-b1b7-9304005e3c76
spec:
  disruption:
    consolidateAfter: 30s
    consolidationPolicy: WhenEmpty
    expireAfter: 720h
  limits:
    cpu: 6400
  template:
    spec:
      nodeClassRef:
        apiVersion: karpenter.k8s.aws/v1beta1
        kind: EC2NodeClass
        name: default-workflow
      requirements:
      - key: kubernetes.io/arch
        operator: In
        values:
        - amd64
      - key: node.kubernetes.io/instance-type
        operator: In
        values:
        - c6id.8xlarge
        - c5d.9xlarge
        - m6id.8xlarge
        - m5d.8xlarge
        - r6id.8xlarge
        - r5d.8xlarge
      - key: karpenter.sh/capacity-type
        operator: In
        values:
        - spot
      - key: node-role.kariusdx.io/workflows
        operator: In
        values:
        - "true"
      - key: capacity-spread
        operator: In
        values:
        - "2"
        - "3"
        - "4"
        - "5"
      taints:
      - effect: NoSchedule
        key: node-role.kariusdx.io/workflows
        value: "true"
status: {}
```

and

```
apiVersion: karpenter.sh/v1beta1
kind: NodePool
metadata:
  annotations:
    karpenter.sh/nodepool-hash: "2808749311332697570"
    meta.helm.sh/release-name: karpenter-config
    meta.helm.sh/release-namespace: default
  creationTimestamp: "2023-11-27T20:16:44Z"
  generation: 1
  labels:
    app.kubernetes.io/managed-by: Helm
  name: default-workflow-on-demand
  resourceVersion: "76338"
  uid: da8e0d44-cf7b-42a9-8d2f-2c38ffcdeb16
spec:
  disruption:
[...]

PR #5174, opened 2023-11-28
docs: add label selector to workload topology spread constraint example

jmdeal:
Fixes #N/A 

**Description**
This PR adds the `labelSelector` field to the topology spread constraint example under advanced scheduling in the docs. Although it opts to not define a selector, indicating that the field should exist avoid confusion in issues like #5171.

N/A

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/website/content/en/docs/concepts/scheduling.md
+++ b/website/content/en/docs/concepts/scheduling.md
@@ -602,4 +602,6 @@ topologySpreadConstraints:
 - maxSkew: 1
   topologyKey: capacity-spread
   whenUnsatisfiable: DoNotSchedule
+  labelSelector:
+     ...
 ```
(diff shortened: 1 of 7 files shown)


website/content/en/docs/concepts/scheduling.md
```diff
@@ -602,4 +602,6 @@ topologySpreadConstraints:
 - maxSkew: 1
   topologyKey: capacity-spread
   whenUnsatisfiable: DoNotSchedule
+  labelSelector:
```
jonathan-innis, line 605: This needs to be added to `v0.32` docs as well

jonathan-innis, line 605: Are there any other places that we should specify this labelSelector throughout our docs?

jmdeal: Should it be backported to all the current docs or just v0.32? It applies to all versions but not sure what the standard practice for docs is.

jmdeal: For `topologySpreadConstraints` at least this looks to be the only place.

jonathan-innis (approved):
LGTM 🚀

PR #5193, opened 2023-11-30
Test: Docker image for tests

### LINKED ISSUE #4545: E2E Tests should use a Docker Image as opposed to installing tooling at runtime
### Description

Currently, our E2E tests run local actions to install tooling at runtime (like `helm`, `helm diff`, `prometheus`, `eksctl`, etc.). We should create an image that we can use to run our testing so that we don't have to install these things at runtime, reduce our test time, and have more reliability in our tests which sometimes fail due to these installation failures.

aman0408:
Fixes #4545 

**Description**

- [ ] Yes, issue opened: # 
- [x] No


test/Dockerfile
```diff
@@ -59,4 +60,7 @@ RUN go version
 # Cache modules to speed up test execution
 RUN git clone https://github.com/aws/karpenter.git /karpenter
 WORKDIR /karpenter
-RUN GOPROXY=direct go mod tidy
+
+RUN K8S_VERSION="1.28.x" make toolchain
```
jonathan-innis, line 64: Do we need to run the full `make toolchain`? IIRC, we should be fine just defining a subset of dependencies in this Dockerfile and then using those directly.

hack/toolchain.sh
```diff
@@ -3,6 +3,7 @@ set -euo pipefail
 
 K8S_VERSION="${K8S_VERSION:="1.28.x"}"
 KUBEBUILDER_ASSETS="/usr/local/kubebuilder/bin"
+USER=$(whoami)
```
jonathan-innis, line 6: What's the need for adding this to the script?

jonathan-innis:
Discussed offline a bit. Given the state of the current code and the value prop of adding the Dockerfile, I'm more inclined to forego this change in favor of just installing some of the tooling at runtime. If we start feeling like the overhead of running all of the installs at the beginning of the E2E is too onerous, then I could see us adding this, but I'm not sure it's worth the maintenance burden right now.

PR #5202, opened 2023-12-01
ci: Adjust regions for metrics of resource counter and sweeper

engedaam:
Fixes #N/A 

**Description**
- Differentiate  between global and regional resources for the sweeper and counter for the metrics that are pushed to Timestream.  

- N/A

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/hack/resource/pkg/resourcetypes/resourcetypes.go
+++ b/test/hack/resource/pkg/resourcetypes/resourcetypes.go
@@ -34,6 +34,8 @@ const (
 type Type interface {
 	// String is the string representation of the type
 	String() string
+	// Global determines if a resource is globally located in an account
+	Global() bool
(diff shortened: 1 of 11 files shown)


test/hack/resource/pkg/resourcetypes/resourcetypes.go
```diff
@@ -34,6 +34,8 @@ const (
 type Type interface {
 	// String is the string representation of the type
 	String() string
+	// Region returns the region the resource is located
+	Region() string
```
jonathan-innis, line 38:
```suggestion
	Global() bool
```

Just a thought. I'm fine either way. But this would mean that we wouldn't have to pass the region down into every resourcetype client and can just check if something is a global resource or not

jonathan-innis (approved):
LGTM 🚀

PR #5213, opened 2023-12-03
feat: Add build_info metric with version, sha, golang_version labels

nantiferov:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/665

**Description**

As described in the issue, it's useful to have metric with at least version label, which makes easier to find what version was running at each time to correlate issues related to update, etc. Example https://www.robustperception.io/exposing-the-software-version-to-prometheus/

So far this implementation might be a little bit "naive", since I tried to adopt existing code in project to add one more metric. If this should be done in a different way, I'm glad to have a feedback.

Also, I added only version and golang_version labels, because they're already available. In theory more can be added, like branch or sha. Not sure if they're very useful though.

I checked contribution guide and tried to follow it.

 Did `make apply` with pushing image to temporary ECR registry and deploying karpenter to my test EKS cluster.
All works fine, example of metric output:
```
# HELP karpenter_build_info A metric with a constant '1' value labeled by version from which karpenter was built.
# TYPE karpenter_build_info gauge
karpenter_build_info{golang_version="go1.21.4",sha="e19bcd7-dirty",version="e19bcd70"} 1
```

- [ ] Yes, issue opened: # 
- [x] No

Not sure, probably need to update docs, but closer to PR be ready to merge. 


pkg/operator/operator.go
```diff
@@ -67,10 +71,20 @@ import (
 	"github.com/aws/karpenter/pkg/providers/version"
 )
 
+var BuildInfo = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Namespace: metrics.Namespace,
+		Name:      "build_info",
+		Help:      "A metric with a constant '1' value labeled by version from which karpenter was built.",
+	},
+	[]string{"version", "golang_version"})
```
jonathan-innis, line 80:
```suggestion
	[]string{"version", "golang_version", "sha"})
```

pkg/operator/operator.go
```diff
@@ -67,10 +71,20 @@ import (
...
+
 func init() {
 	lo.Must0(apis.AddToScheme(scheme.Scheme))
 	v1alpha5.NormalizedLabels = lo.Assign(v1alpha5.NormalizedLabels, map[string]string{"topology.ebs.csi.aws.com/zone": corev1.LabelTopologyZone})
 	corev1beta1.NormalizedLabels = lo.Assign(corev1beta1.NormalizedLabels, map[string]string{"topology.ebs.csi.aws.com/zone": corev1.LabelTopologyZone})
+	crmetrics.Registry.MustRegister(BuildInfo)
+	BuildInfo.WithLabelValues(operator.Version, runtime.Version()).Set(1)
```
jonathan-innis, line 87: You could consider using `changeset.Get()` to grab the SHA that is injected into the logger here into the metric: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/operator/logging/logging.go#L91

nantiferov: Thank you for review and suggestion 👍 

I just yesterday noticed `sha` somewhere in karpenter logs, missed it initially when checked code.

nantiferov: Done, run tests in test EKS to check how this new metric works, all looks good to me.

pkg/operator/operator.go
```diff
@@ -66,9 +71,19 @@ import (
 	"github.com/aws/karpenter-provider-aws/pkg/providers/version"
 )
 
+var BuildInfo = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Namespace: metrics.Namespace,
+		Name:      "build_info",
+		Help:      "A metric with a constant '1' value labeled by version from which karpenter was built.",
+	},
+	[]string{"version", "golang_version", "sha"})
```
jonathan-innis, line 80:
```suggestion
	[]string{"version", "goversion", "sha"})
```

Looking at this a little more closely, should we follow the prior art here that Prometheus has in terms of labels?

nantiferov: Yep, you're right, missed that.
Thank you!

PR #5216, opened 2023-12-04
ci: Use tag name from context

jonathan-innis:
Fixes #N/A 

**Description**

This PR updates the `release.yaml` PR to use the tag from context rather than pulling it from the describe call

Manual validation on another repo

- [ ] Yes, issue opened: # 
- [x] No


njtran (approved):
LGTM. We should also check if there are other places where we use javascript and use the same pattern.

PR #5229, opened 2023-12-05
test: expanded storage integration tests

jmdeal:
Fixes #N/A 

**Description**
This PR expands upon the existing storage integration tests in light of https://github.com/kubernetes-sigs/karpenter/pull/783.

`go test` and `/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No


jmdeal (commented):
/karpenter snapshot

test/suites/integration/storage_test.go
```diff
@@ -16,104 +16,212 @@ package integration_test
...
+			env.ExpectCreatedNodeCount("==", 1)
 		})
-
-		pvc := test.PersistentVolumeClaim(test.PersistentVolumeClaimOptions{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: "ebs-claim",
-			},
-			StorageClassName: aws.String(storageClassName),
-			Resources:        v1.ResourceRequirements{Requests: v1.ResourceList{v1.ResourceStorage: resource.MustParse("5Gi")}},
+		It("should run a pod with a pre-bound persistent volume (non-existent storage class)", func() {
```
jonathan-innis, line 57: Should we test the case with us safely handling the case where someone is using a dynamically created PVC with a non-existent storage class? I know we discussed this edge-case and it might be nice to have some handling for it in our tests. Also happy to punt on this one and save it for another PR but it's good to know that this is a possibility

test/suites/integration/storage_test.go
```diff
@@ -16,104 +16,212 @@ package integration_test
...
+					Name: pvc.Spec.VolumeName,
+				},
+				StorageClassName: "non-existent-storage-class",
+				Zones:            []string{shuffledAZs[0]},
+			})
+			pod := test.Pod(test.PodOptions{
+				PersistentVolumeClaims: []string{pvc.Name},
+			})
+			env.ExpectCreated(nodeClass, nodePool, pv, pvc, pod)
+			env.EventuallyExpectHealthy(pod)
```
jonathan-innis, line 93: So in this case, should we unschedule the pod, then create a new pod and expect that pod to be healthy, use the PV, and schedule to the same zone?

test/suites/integration/storage_test.go
```diff
@@ -16,104 +16,212 @@ package integration_test
...
+					StorageClassName: &storageClass.Name,
+				})
+			})
+			pods := lo.Map(pvcs, func(pvc *v1.PersistentVolumeClaim, _ int) *v1.Pod {
+				return test.Pod(test.PodOptions{
+					PersistentVolumeClaims: []string{pvc.Name},
+				})
+			})
+			env.ExpectCreated(nodeClass, nodePool, storageClass, pvcs[0], pods[0])
+			env.EventuallyExpectHealthy(pods[0])
```
jonathan-innis, line 173: Consider a comment here on why we have this ordering

test/suites/integration/storage_test.go
```diff
@@ -16,104 +16,212 @@ package integration_test
...
+				}
+			}
+			storageClass = test.StorageClass(test.StorageClassOptions{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-storage-class",
+				},
+				Provisioner:       aws.String("ebs.csi.aws.com"),
+				VolumeBindingMode: lo.ToPtr(storagev1.VolumeBindingWaitForFirstConsumer),
+			})
+			ExpectSetEBSDriverLimit(1)
```
jonathan-innis, line 120: Just curious: Does this need to be set in each test case or just in the ones where we care about the volume limit?

test/suites/integration/storage_test.go
jmdeal: We could do that but I don't think it's necessary to test the topology logic. The PVC needs to bind to the PV before it can schedule, once it binds the pod gains the PVs topology constraints. IIUC these constraints would cause the pod to fail to schedule if Karpenter launched a node in the wrong AZ.

test/suites/integration/storage_test.go
jmdeal: Do you mean with an ephemeral persistent volume (that makes sense to add). I don't think a dynamic PV with a non-existent storage class makes sense to test because it would fail to schedule. This should be covered in the functional testing.

jonathan-innis (approved):
Awesome work here 🎉

jonathan-innis (approved):
LGTM 🚀

PR #5298, opened 2023-12-11
docs: add faq for pod autoscaling

jmdeal:
Fixes #N/A 

**Description**
Adds an FAQ about pod autoscaling being outside of Karpenter's scope and gives recommendations.

N/A

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/website/content/en/docs/faq.md
+++ b/website/content/en/docs/faq.md
@@ -92,9 +92,9 @@ Yes, Karpenter supports provisioning metal instance types when a NodePool's `nod
 
 ### How does Karpenter dynamically select instance types?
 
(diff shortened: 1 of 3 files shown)


website/content/en/docs/faq.md
```diff
@@ -185,6 +185,10 @@ amiSelectorTerms:
     - name: Windows_Server-2022-English-Full-EKS_Optimized-{{< param "latest_k8s_version" >}}*
 ```
 
+### Can I use Karpenter to scale my workloads?
```
jonathan-innis, line 188: I think this question is a little ambigouous. A natural question is here is "scale what?". Maybe something a little more deliberate might help here. Not exactly sure on the wording (we can iron it out) but maybe something like, "Can I use Karpenter to scale my container resources or replica count?"

jmdeal: Agree it's a bit ambiguous. Maybe "Can I use Karpenter to horizontally scale my workloads?" To me that means horizontal pod autoscaling but I'm not sure if that term is overloaded in any other way that still makes it ambiguous.

jmdeal: Discussed offline, settled on "Can I use Karpenter to scale my workload's pods" to encapsulate both vertical and horizontal pod autoscaling.

jonathan-innis (approved):
LGTM 🚀

PR #5303, opened 2023-12-12
ci: Drop unneeded calls to pull-through cache

jonathan-innis:
Fixes #N/A 

**Description**

Calls to the pull-through cache are unneeded since this was only do local cache hydration into the account. Since we are pulling snapshots from a different ECR anyways, there's no need to do pull-through cache hydration here on release.

- [ ] Yes, issue opened: # 
- [x] No


hack/release/common.sh
```diff
@@ -123,16 +123,6 @@ cosignImages() {
         "${CONTROLLER_IMG}"
 }
 
-pullPrivateReplica(){
```
jmdeal, line 126: There's still a call to this in `release()`, other than that looks good 👍

jonathan-innis: Good call. I recalled I had originally scoped this into the change but look like somehow I managed to miss it.

jmdeal (approved):
LGTM 🚀

PR #5307, opened 2023-12-12
feat: Placement Group Support

### LINKED ISSUE #3324: Support for AWS placement group strategies

Support to schedule node based on placement group strategies. As example when you launch a new EC2 instance, the EC2 service attempts to place the instance in such a way that all of your instances are spread out across underlying hardware to minimize correlated failures. You can use placement groups to influence the placement of a group of interdependent instances to meet the needs of your workload.

Need to provision ETCD cluster in the same AZ.

Using ASG with launch template configured to use placement group.

### Additional Context

_No response_

### Attachments

_No response_

### Community Note

lorrrrrrrenzo:
Fixes #3324

**Description**
Allows for the configuration of a Placement Group when defining an `AWSNodeTemplate` resource. Based on existing work by @preflightsiren on PR #4553. 

- [x] Yes, PR includes docs updates (in progress!) 
- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/pkg/apis/v1beta1/ec2nodeclass.go
+++ b/pkg/apis/v1beta1/ec2nodeclass.go
@@ -105,6 +105,9 @@ type EC2NodeClassSpec struct {
 	// +kubebuilder:default={"httpEndpoint":"enabled","httpProtocolIPv6":"disabled","httpPutResponseHopLimit":2,"httpTokens":"required"}
 	// +optional
 	MetadataOptions *MetadataOptions `json:"metadataOptions,omitempty"`
+	// PlacementGroupSelectorTerms is a list of PlacementGroupSelector. The terms are ORed.
+	// +optional
+	PlacementGroupSelectorTerms []PlacementGroupSelectorTerm `json:"placementGroupSelectorTerms,omitempty" hash:"ignore"`
 	// Context is a Reserved field in EC2 APIs
 	// https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html
 	// +optional
@@ -165,6 +168,14 @@ type AMISelectorTerm struct {
 	// You can specify a combination of AWS account IDs, "self", "amazon", and "aws-marketplace"
 	// +optional
 	Owner string `json:"owner,omitempty"`
+}
+
+// PlacementGroupSelectorTerm defines the selection logic for ec2 placement groups
+// that are used to launch nodes. If multiple fields are used for selection, the requirements are ANDed
+type PlacementGroupSelectorTerm struct {
+	// Name of the placement group to be selected
+	// +optional
+	Name string `json:"name,omitempty"`
 }
 
 // MetadataOptions contains parameters for specifying the exposure of the
--- a/pkg/apis/v1beta1/ec2nodeclass_status.go
+++ b/pkg/apis/v1beta1/ec2nodeclass_status.go
@@ -63,6 +63,9 @@ type EC2NodeClassStatus struct {
 	// cluster under the AMI selectors.
 	// +optional
 	AMIs []AMI `json:"amis,omitempty"`
+	// PlacementGroups contains the ec2 placement group arns
+	// +optional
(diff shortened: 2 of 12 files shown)


pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -167,6 +170,14 @@ type AMISelectorTerm struct {
 	Owner string `json:"owner,omitempty"`
 }
 
+// PlacementGroupSelectorTerm defines the selection logic for ec2 placement groups
+// that are used to launch nodes. If multiple fields are used for selection, the requirements are ANDed
+type PlacementGroupSelectorTerm struct {
+	// Name of the placement group to be selected
+	// +optional
+	Name string `json:"name,omitempty"`
```
jonathan-innis, line 178: I see that it's possible to also use tags here as well. I wonder if we also support tagging initially similar to the other selectors

pkg/apis/v1beta1/ec2nodeclass_status.go
```diff
@@ -63,6 +63,9 @@ type EC2NodeClassStatus struct {
 	// cluster under the AMI selectors.
 	// +optional
 	AMIs []AMI `json:"amis,omitempty"`
+	// PlacementGroups contains the ec2 placement group arns
+	// +optional
+	PlacementGroups []string `json:"placementGroups,omitempty"`
```
jonathan-innis, line 68: Should we resolve the strategy into the status block here? If the placement group type is `partition`, should we consider putting the number of partitions that we are using into the status as well?

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -167,6 +170,14 @@ type AMISelectorTerm struct {
 	Owner string `json:"owner,omitempty"`
 }
 
+// PlacementGroupSelectorTerm defines the selection logic for ec2 placement groups
+// that are used to launch nodes. If multiple fields are used for selection, the requirements are ANDed
+type PlacementGroupSelectorTerm struct {
```
jonathan-innis, line 175: There's a question here of how we should pass requirements down when using something like `cluster` placement groups OR when using something like `spread` with `rack`. 

When using the `spread` with `rack`, the EC2 documentation mentions that you can't launch more than 7 instances in a single AZ (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/placement-groups.html#:~:text=A%20rack%20spread%20placement%20group%20supports%20a%20maximum%20of%20seven%20running%20instances%20per%20Availability%20Zone), which means that we are restricted with our node limits when using this type of placement group. One option here is that we could constrain our requirements when using this type of placement group so that when we run out of instances to launch in a single AZ, we just get rid of that AZ from our requirements and start launching in other AZs. Another option here is that we allow to select on multiple placement groups and if we run out of space with one of the `spread` placement groups we just move to the other one.

This handle differently, though, when we are dealing with `cluster` or `partition` placement groups where there is no limit on the number of instances that can be launched and it makes less sense to allow for multiple placement groups to be selected on. 

Based on all of these considerations, I think we should think about writing up a small design that considers the use-cases around placement groups, how they meld with requirements and what the API surface should look like.

pkg/apis/v1beta1/ec2nodeclass.go
jonathan-innis: One thing that's nice about tagging and naming is that it works consistently across accounts, so if you have similar set-ups for different accounts, everything just works out-of-the-box. 

This is the general philosophy that Karpenter has taken towards selectors in the project, ensuring that we can move Karpenter from one account to another or from one region to another and everything "just works."

The trade-off of opening up this selector to tags is that now multiple placement groups can be returned, so how do you pick between them? There probably needs to be a consistent ordering and the selection between them shouldn't be random, particularly when it comes to the interaction of this feature with the drift feature.

unityabir:
Why this PR was closed? @jonathan-innis

Issue #5225, opened 2023-12-05
Consider storing InstanceTypes only once in the cache from the EC2 GetInstanceTypes call rather than per-NodeClass


jonathan-innis:
### Description

Right now, we are storing InstanceTypes in the cache multiple times. More specifically, we differentiate on different forms of the InstanceType by using the cache key that can be found [here](https://github.com/aws/karpenter/blob/62d5bc8ee2607cfe1095fcccf16e9a9448cdcc14/pkg/providers/instancetype/instancetype.go#L118). This cache key causes us to store different instance types whenever we have a different NodeClass, different Kubelets from the NodePools and different resolved subnet values from the NodeClass. 

This can cause us to have a _lot_ of duplication of these InstanceType structs stored in our instance type cache, which can cause us to use a lot of unnecessary memory (see https://github.com/aws/karpenter/issues/5196). I'd like to explore if it's possible for us to only return InstanceTypes one time from the CloudProvider, store this in memory, and then see if we can layer calculate an InstanceType based off a NodeClass at runtime through some sort of layered mechanism like `GetInstanceTypeForNodeClass(instanceType *InstanceType, nodePool *v1beta1.NodePool)`. This would allow us to only store one implementation of the struct in the cache while also allowing us to differentiate the InstanceTypes on a per-NodePool (and per-NodeClass) basis if needed.

As part of this work, it's probably also worth exploring if we can just reduce the cache size generally. That is, are we unnecessarily creating cache keys per-NodeClass when they could be defined at a more granular level?

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #5318, opened 2023-12-13
perf: Reduce Memory Usage from the InstanceType Provider 

### LINKED ISSUE #5196: Karpenter memory utilisation balloons after upgrade to 0.32.2
### Description

**Observed Behavior**:
After upgrading Karpenter Helm Chart from v0.31.1 to v0.32.2 the memory usage more than doubles in some cases. In all clusters a dramatic change in memory utilisation can be seen.

This cluster with 33 nodes + 7 Fargate nodes running approximately 1000 pods
Screenshot 2023-11-30 at 17 17 36

This cluster with 12 nodes + 7 Fargate nodes running approximately 250 pods
Screenshot 2023-11-30 at 17 19 37

**Expected Behavior**:
Memory utilisation should remain the same where
* No changes occurred with the work load on the cluster
* No new Karpenter features were activated during/prior to this upgrade

If this is expected, a note about this in the Changelog would be good to apprise users

Upgrade to Karpenter 0.32.2 from 0.31.1
Monitor resource utilisation of the pods/deployment

**Versions**:
- Chart Version:
0.32.2

**Update:** 
The issue starts in version 0.32.0 and is present in later versions as above.

Reverting back to 0.31.2 from 0.32.2 seems to restore original memory usage

image

engedaam:
Fixes #5196, #5225

**Description**
- Clearing out the cache for the instance types provider on expired items. 
- Save items based on the EC2NodeClass resolution instead of EC2Nodeclass UUID

**E2E  Integration Periodic Run** 
**Memory Utilization** 
Screenshot 2023-12-14 at 4 08 27 PM

**CPU Utilization** 
Screenshot 2023-12-14 at 5 20 22 PM

**E2E Integration with PR Run** 
**Memory Utilization** 
Screenshot 2023-12-14 at 9 09 01 PM

**CPU Utilization** 
Screenshot 2023-12-14 at 9 14 07 PM

- E2E integration was used comparing the memory usage  

- `/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -35,5 +35,5 @@ const (
 
 const (
 	// DefaultCleanupInterval triggers cache cleanup (lazy eviction) at this interval.
-	DefaultCleanupInterval = 10 * time.Minute
+	DefaultCleanupInterval = time.Minute
 )
--- a/pkg/providers/instancetype/instancetype.go
+++ b/pkg/providers/instancetype/instancetype.go
@@ -112,8 +112,8 @@ func (p *Provider) List(ctx context.Context, kc *corev1beta1.KubeletConfiguratio
 	// Compute fully initialized instance types hash key
 	subnetHash, _ := hashstructure.Hash(subnets, hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: true})
 	kcHash, _ := hashstructure.Hash(kc, hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: true})
-	key := fmt.Sprintf("%d-%d-%d-%s-%016x-%016x", p.instanceTypesSeqNum, p.instanceTypeOfferingsSeqNum, p.unavailableOfferings.SeqNum, nodeClass.UID, subnetHash, kcHash)
-
+	blockDeviceMappingsHash, _ := hashstructure.Hash(nodeClass.Spec.BlockDeviceMappings, hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: true})
+	key := fmt.Sprintf("%d-%d-%d-%016x-%016x-%016x-%s", p.instanceTypesSeqNum, p.instanceTypeOfferingsSeqNum, p.unavailableOfferings.SeqNum, subnetHash, kcHash, blockDeviceMappingsHash, aws.StringValue(nodeClass.Spec.AMIFamily))
 	if item, ok := p.cache.Get(key); ok {
 		return item.([]*cloudprovider.InstanceType), nil
 	}
@@ -221,10 +221,12 @@ func (p *Provider) getInstanceTypeOfferings(ctx context.Context) (map[string]set
 		}); err != nil {
 		return nil, fmt.Errorf("describing instance type zone offerings, %w", err)
 	}
-	if p.cm.HasChanged("instance-type-count", len(instanceTypeOfferings)) {
+	if p.cm.HasChanged("instance-type-offering", instanceTypeOfferings) {
+		// Only update instanceTypesSeqNun with the instance type offerings  have been changed
+		// This is to not create new keys with duplicate instance type offerings option
+		atomic.AddUint64(&p.instanceTypeOfferingsSeqNum, 1)
 		logging.FromContext(ctx).With("instance-type-count", len(instanceTypeOfferings)).Debugf("discovered offerings for instance types")
 	}
-	atomic.AddUint64(&p.instanceTypeOfferingsSeqNum, 1)
 	p.cache.SetDefault(InstanceTypeOfferingsCacheKey, instanceTypeOfferings)
 	return instanceTypeOfferings, nil
 }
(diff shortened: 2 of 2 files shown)


engedaam (commented):
/karpenter snapshot

engedaam (commented):
/karpenter scale

pkg/providers/instancetype/instancetype.go
```diff
@@ -275,5 +274,16 @@ func (p *Provider) GetInstanceTypes(ctx context.Context) ([]*ec2.InstanceTypeInf
 	}
 	atomic.AddUint64(&p.instanceTypesSeqNum, 1)
 	p.cache.SetDefault(InstanceTypesCacheKey, instanceTypes)
+	p.cleanOldInstanceTypesFromCache()
```
jonathan-innis, line 277: A few questions here:
1. Should the cache that stores the instance types just be its own cache, separate from the global data values that are being cached?
2. If we have a separate cache here, I think we can just leverage cache.Reset() to reset this data when we want to clear out old data rather than having to call delete across all of the keys
3. How does this affect calls where `GetInstanceTypes()` may be called multiple times within a loop (like a `Create()` loop). Re-getting the instance types might result in inconsistent data, though I think that this is the case today. We should probably improve this story though.

engedaam (commented):
/karpenter snapshot

engedaam (commented):
/karpenter snapshot

pkg/providers/instancetype/instancetype.go
```diff
@@ -88,6 +88,7 @@ func NewProvider(region string, cache *cache.Cache, ec2api ec2iface.EC2API, subn
 }
 
 func (p *Provider) List(ctx context.Context, kc *corev1beta1.KubeletConfiguration, nodeClass *v1beta1.EC2NodeClass) ([]*cloudprovider.InstanceType, error) {
+	p.cache.DeleteExpired()
```
jonathan-innis, line 91: How expensive is this call? Maybe consider benchmarking? If we call it on every list operation, that adds a non-trivial amount of time to the List() call

pkg/providers/instancetype/instancetype.go
```diff
@@ -112,11 +113,7 @@ func (p *Provider) List(ctx context.Context, kc *corev1beta1.KubeletConfiguratio
 	// Compute fully initialized instance types hash key
 	subnetHash, _ := hashstructure.Hash(subnets, hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: true})
 	kcHash, _ := hashstructure.Hash(kc, hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: true})
-	key := fmt.Sprintf("%d-%d-%d-%s-%016x-%016x", p.instanceTypesSeqNum, p.instanceTypeOfferingsSeqNum, p.unavailableOfferings.SeqNum, nodeClass.UID, subnetHash, kcHash)
-
-	if item, ok := p.cache.Get(key); ok {
```
jonathan-innis, line 117: We dropped the caching altogether?

engedaam (commented):
/karpenter snapshot

engedaam (commented):
/karpenter snapshot

engedaam (commented):
/karpenter scale

pkg/cache/cache.go
```diff
@@ -36,4 +36,6 @@ const (
 const (
 	// DefaultCleanupInterval triggers cache cleanup (lazy eviction) at this interval.
 	DefaultCleanupInterval = 10 * time.Minute
```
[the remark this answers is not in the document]
engedaam: Expired items are not used any of the cache. So deceased the overall cleanup to 1 minute

pkg/providers/instancetype/instancetype.go
engedaam: Removed as the cache should be dumped every minute now

engedaam (commented):
/karpenter snapshot

engedaam (commented):
/karpenter snapshot

jonathan-innis (commented):
Looks good. Just a couple of questions

engedaam (commented):
/karpenter snapshot

pkg/cache/cache.go
```diff
@@ -35,5 +35,5 @@ const (
 
 const (
 	// DefaultCleanupInterval triggers cache cleanup (lazy eviction) at this interval.
-	DefaultCleanupInterval = 10 * time.Minute
+	DefaultCleanupInterval = 2 * time.Minute
```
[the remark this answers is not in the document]
engedaam: All expired items will be cleanup after a minute. Other providers will not have as much churn as the instancetype provider, resulting in manual effect.

engedaam (commented):
/karpenter snapshot

engedaam (commented):
/karpenter scale

jonathan-innis (approved):
This all seems good to me. If we are confident that this cleanup interval isn't going to tangibly hurt our performance and improve our memory footprint, then 🚀

Issue #5239, opened 2023-12-05
Unable to provision nodes in other zone types


STollenaar:
I have a requirement to schedule some nodes in a local zone (In this case eu-north-1-cph-1a). I currently have the 1 self managed EKS node running in that zone, and Karpenter successfully starts up. However it cannot provision any new nodes through Karpenter. 

Is there something with my config below that is preventing Karpenter from creating these nodes?

snippet of logs with the error and other startup info:
```
2023-12-05T20:07:33.976Z	DEBUG	controller.nodeclass	discovered subnets	{"commit": "56aaec1", "ec2nodeclass": "company-cluster", "subnets": ["subnet-id (eu-north-1-cph-1a)"]}
2023-12-05T20:07:34.025Z	DEBUG	controller.pricing	updated spot pricing with instance types and offerings	{"commit": "56aaec1", "instance-type-count": 730, "offering-count": 1013}
2023-12-05T20:07:34.135Z	DEBUG	controller.nodeclass	discovered security groups	{"commit": "56aaec1", "ec2nodeclass": "company-cluster", "security-groups": ["sg-id"]}
2023-12-05T20:07:34.189Z	DEBUG	controller.nodeclass	discovered kubernetes version	{"commit": "56aaec1", "ec2nodeclass": "company-cluster", "version": "1.28"}
2023-12-05T20:07:34.625Z	DEBUG	controller.nodeclass	discovered amis	{"commit": "56aaec1", "ec2nodeclass": "company-cluster", "ids": "ami-0bf75e19885e443e6, ami-09389d7cecdb39fba, ami-09389d7cecdb39fba, ami-0f6a90c148c8a59fd", "count": 4}
2023-12-05T20:07:34.730Z	INFO	controller.provisioner	pod kube-system/coredns-d9694c9bb-5758f has a preferred Anti-Affinity which can prevent consolidation	{"commit": "56aaec1"}
2023-12-05T20:07:35.199Z	DEBUG	controller.pricing	updated on-demand pricing	{"commit": "56aaec1", "instance-type-count": 361}
2023-12-05T20:07:35.281Z	DEBUG	controller.disruption	discovered instance types	{"commit": "56aaec1", "count": 361}
2023-12-05T20:07:35.403Z	DEBUG	controller.disruption	discovered offerings for instance types	{"commit": "56aaec1", "instance-type-count": 362}
[...]

PR #5321, opened 2023-12-14
ci: E2E testing for local zones

### LINKED ISSUE #4513: Add a Local Zone to e2e Tests
### Description

Some sort of local zone verification should be automatically tested on a cadence. I think it will be pretty trivial to add a local zone to our current e2e tests. 

https://aws.amazon.com/about-aws/global-infrastructure/localzones/

jmdeal:
Fixes #4513 

**Description**
In light of a recent regression (#5239) this PR adds support for end to end tests in local zones. Since local zones come with a number of caveats, rather than adapting the entire test suite to run in local zones, a new suite is added and will be run in it's own region with a local zone enabled. This test suite may be expanded in the future but for the time being it will at least guarantee that we can launch instances in local zones.

- [x] New tests validated using `go test` against a personal EKS cluster
- [ ] Workflow updates validated via workflow run in personal fork

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.github/actions/e2e/setup-cluster/action.yaml
+++ b/.github/actions/e2e/setup-cluster/action.yaml
@@ -40,6 +40,10 @@ inputs:
   git_ref:
     description: "The git commit, tag, or branch to check out. Requires a corresponding Karpenter snapshot release"
     required: false
+  enable_local_zones:
+    description: "Whether to include local zones in the VPC created for the cluster."
+    required: false
+    default: 'false'
 runs:
   using: "composite"
   steps:
@@ -81,14 +85,15 @@ runs:
       IP_FAMILY: ${{ inputs.ip_family }}
       PRIVATE_CLUSTER: ${{ inputs.private_cluster }}
       GIT_REF:  ${{ inputs.git_ref }}
+      ENABLE_LOCAL_ZONES: ${{ inputs.enable_local_zones }}
     run: |
-      if [[ "$GIT_REF" == '' ]]; then 
+      if [[ "$GIT_REF" == '' ]]; then
         GIT_REF=$(git rev-parse HEAD)
       fi
 
-      # Disable Pod Identity for Karpenter on K8s 1.23. Pod Identity is not supported on K8s 1.23 
+      # Disable Pod Identity for Karpenter on K8s 1.23. Pod Identity is not supported on K8s 1.23
       # https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html#pod-id-considerations
-      if [[ "$K8S_VERSION" == '1.23' ]]; then 
+      if [[ "$K8S_VERSION" == '1.23' ]]; then
         KARPENTER_IAM="""
           - metadata:
               name: karpenter
@@ -104,14 +109,14 @@ runs:
             serviceAccountName: karpenter
             roleName: karpenter-irsa-${{ inputs.cluster_name }}
             permissionsBoundaryARN: "arn:aws:iam::${{ inputs.account_id }}:policy/GithubActionsPermissionsBoundary"
-            permissionPolicyARNs: 
+            permissionPolicyARNs:
               - "arn:aws:iam::${{ inputs.account_id }}:policy/KarpenterControllerPolicy-${{ inputs.cluster_name }}""""
         POD_IDENTITY="""- name: eks-pod-identity-agent
         permissionsBoundary: "arn:aws:iam::$ACCOUNT_ID:policy/GithubActionsPermissionsBoundary"
         configurationValues: |
           tolerations:
             - operator: Exists"""
-      fi 
+      fi
 
       # Create or Upgrade the cluster based on whether the cluster already exists
       cmd="create"
@@ -178,6 +183,13 @@ runs:
       $POD_IDENTITY
       EOF
 
+      if [[ $ENABLE_LOCAL_ZONES == "true" ]]; then
+        local_zones=$(AWS_REGION=$REGION aws ec2 describe-availability-zones | yq '.AvailabilityZones | filter(.ZoneType == "local-zone") | [.[].ZoneName] | join(" ")')
+        for zone in $local_zones; do
+          yq -i ".localZones += [\"$zone\"]" clusterconfig.yaml
+        done
+      fi
+
       if [[ $PRIVATE_CLUSTER == 'true' ]]; then
         yq -i '.privateCluster.enabled=true' clusterconfig.yaml
         yq -i '.managedNodeGroups[0].privateNetworking=true' clusterconfig.yaml
@@ -211,9 +223,9 @@ runs:
       CLUSTER_NAME: ${{ inputs.cluster_name }}
       GIT_REF: ${{ inputs.git_ref }}
     run: |
-      if [[ "$GIT_REF" == '' ]]; then 
+      if [[ "$GIT_REF" == '' ]]; then
         GIT_REF=$(git rev-parse HEAD)
-      fi 
+      fi
       oidc_id=$(aws eks describe-cluster --name "$CLUSTER_NAME" --query "cluster.identity.oidc.issuer" --output text | cut -d '/' -f 3,4,5)
       arn="arn:aws:iam::$ACCOUNT_ID:oidc-provider/${oidc_id}"
       aws iam tag-open-id-connect-provider --open-id-connect-provider-arn $arn \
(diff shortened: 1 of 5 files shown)


test/suites/localzone/suite_test.go
```diff
@@ -0,0 +1,81 @@
...
+	AfterSuite(func() {
+		env.Stop()
+	})
+	RunSpecs(t, "Chaos")
+}
+
+var _ = BeforeEach(func() {
+	env.BeforeEach()
+	nodeClass = env.DefaultEC2NodeClass()
+	// local zones don't support gp3 volumes, use gp2 instead
```
njtran, line 52: I assume this should change eventually. Worth a note with a reference to know when we can remove this?

test/suites/localzone/suite_test.go
```diff
@@ -0,0 +1,81 @@
...
+})
+var _ = AfterEach(func() { env.Cleanup() })
+var _ = AfterEach(func() { env.AfterEach() })
+
+var _ = Describe("LocalZone", func() {
+	It("should successfully launch a node in a local zone", func() {
+		pod := test.UnschedulablePod()
+		env.ExpectCreated(nodeClass, nodePool, pod)
+		env.EventuallyExpectHealthy(pod)
+	})
```
njtran, line 80: Can we do a simple disruption case in this test as well?

test/suites/localzone/suite_test.go
```diff
@@ -0,0 +1,81 @@
...
+	nodePool.Spec.Template.Spec.Requirements = append(nodePool.Spec.Template.Spec.Requirements, v1.NodeSelectorRequirement{
+		Key:      v1.LabelTopologyZone,
+		Operator: v1.NodeSelectorOpIn,
+		Values:   lo.Keys(lo.PickByValues(env.GetZones(), []string{"local-zone"})),
+	})
+})
+var _ = AfterEach(func() { env.Cleanup() })
+var _ = AfterEach(func() { env.AfterEach() })
+
+var _ = Describe("LocalZone", func() {
```
njtran, line 75: Since this is in its own test suite, this will create another cluster. Is it possible for us to include this in another suite so that we can reduce the number of clusters we have to create? It seems unnecessary to create a whole cluster just to test a pod coming online.

.github/workflows/e2e-matrix.yaml
```diff
@@ -31,6 +34,11 @@ on:
           - "us-west-2"
           - "eu-west-1"
         default: "us-east-2"
+      local_zone_region:
+        type: choice
+        options:
+          - "us-west-2"
+        default: "us-west-2"
```
njtran, line 41: If we're only trying to do us-west-2, do we need to parameterize this? Much like soak tests, we can just ensure that we're using the right input here rather than making a separate variable.

njtran (commented):
Nice work. A couple of comments.

test/suites/localzone/suite_test.go
jmdeal: I agree that it seems excessive, the rationale was that this would be the only cluster with local zones enabled. Rather than modifying the existing suites to work around the limitations of local zones, or requiring each NodePool or NodeClass to filter out the zones (either by a topology selector or the subnet selector), this allows us to have a single cluster with local zone support.

test/suites/localzone/suite_test.go
jmdeal: Probably, I'd have to check if it's on a roadmap somewhere. If it is I'll link that here.

.github/workflows/e2e-matrix.yaml
jmdeal: My thought was for people running in this in PRs that may be running against different regions but it probably doesn't matter. Either way you'd have to manually extend the drop-down. Probably fine to hard-code this now and pull it out down the road if needed.

.github/actions/e2e/setup-cluster/action.yaml
```diff
@@ -160,6 +165,12 @@ runs:
             - operator: Exists
       EOF
 
+      if [[ $ENABLE_LOCAL_ZONES == "true" ]]; then
+        while IFS= read -r zone; do
```
jonathan-innis, line 169: Is there a less indirect way for us to handle this? Can we read the items into an array object and then iterate through it? I see what you are doing here, but you can look down at the private cluster VPC endpoint logic to see how this is handled there with subnets and security groups like on L187

.github/workflows/e2e-matrix.yaml
jonathan-innis: +1, let's simplify this now unless we think there is a reason to add the choice here

test/suites/localzone/suite_test.go
jonathan-innis: This makes sense to me since cluster creation is pretty cheap here and I expect our tests to scope down in this package when we have Kwok. This seems resonable to me given the trade-offs

.github/actions/e2e/setup-cluster/action.yaml
jmdeal: I'll preface this with I'm not a bash expert, the way I know how to do it though requires bash 4+ IIRC so wouldn't be very portable. I can take a look and see if there's a simpler way to iterate over the output here.

test/suites/localzone/suite_test.go
jmdeal: Looks like this isn't actually the case in LAX where we'll be testing.
> Q: What EBS volume types are available in AWS Local Zones?
 Local Zones in Los Angeles (us-west-2-lax-1a, and us-west-2-lax-1b), Dallas (us-east-1-dfw-2a), and  Phoenix (us-west-2-phx-2a) provide gp2, gp3, io1, st1, and sc1 EBS volumes. All other Local Zones provide gp2 volume.

I've not been able to find any roadmap for wider gp3 support but I will add a not and link to the existing documentation about what local zones support what. I feel like it does make sense to leave this in though even though we technically don't need to since we're testing in LAX. This way ensures we have some flexibility about what local zones we're running in.

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal:
All previous snapshot failures appear to either be flakes or unrelated to this PR, rerunning after a rebase.

jmdeal (commented):
/karpenter snapshot

.github/workflows/e2e.yaml
```diff
@@ -38,6 +39,10 @@ on:
         default: "1.28"
       cluster_name:
         type: string
+      enable_local_zones:
```
jonathan-innis, line 42: If we have a suite name for this different set of testing, can we just check the suite name rather than having a separate parameter here?

jmdeal (commented):
/karpenter snapshot

jmdeal:
Validated cluster creation with local zones in my personal fork: https://github.com/jmdeal/karpenter-provider-aws/actions/runs/7428846425/job/20216686975. The tests all failed here but that's expected since no snapshot image of karpenter existed for this e2e run. The important part to look at is the availability zones used in the cluster creation process.

jmdeal:
The results of running in my local fork are linked [here](https://github.com/jmdeal/karpenter-provider-aws/actions/runs/7428846425). A couple tests failed due to resource limits in my account but the local zone test was successful.

test/suites/localzone/suite_test.go
```diff
@@ -0,0 +1,104 @@
...
+			}(),
+			VolumeType: lo.ToPtr("gp2"),
+			Encrypted:  lo.ToPtr(false),
+		},
+	})
+	nodePool = env.DefaultNodePool(nodeClass)
+	nodePool.Spec.Template.Spec.Requirements = append(nodePool.Spec.Template.Spec.Requirements, v1.NodeSelectorRequirement{
+		Key:      v1.LabelTopologyZone,
+		Operator: v1.NodeSelectorOpIn,
+		Values:   lo.Keys(lo.PickByValues(env.GetZones(), []string{"local-zone"})),
```
jonathan-innis, line 72: optional: Is there an EC2 constant value for this?

jonathan-innis (approved):
LGTM 🚀 

One optional comment

jonathan-innis:
Quick sanity check: We got the local zone enabled in the CI account?

test/suites/localzone/suite_test.go
jmdeal: Not that I can find, grepped `api.go` and the ec2 go docs with no luck.

jmdeal:
> Quick sanity check: We got the local zone enabled in the CI account?

Yep, only LAX but that's enough here

PR #5335, opened 2023-12-14
chore: remove --container-runtime from AL2 user data

jmdeal:
Fixes #N/A 

**Description**
In light of `--container-runtime` being ignored by the EKS bootstrap script and v1beta1 APIs removing the option to set container runtime, this PR drops the field from the AL2 AMI.

- `make test`

- [ ] Yes, issue opened: # 
- [ ] No


website/content/en/docs/concepts/nodeclasses.md
```diff
@@ -29,21 +29,21 @@ metadata:
   name: default
```
jonathan-innis, line 29: These changes look good to me, but looks like you have a bunch of whitespace changes in files where you actually didn't make any edits.

jmdeal: Are you referring to the two docs files? Both of these have edits, it's just a single edit to remove the `--container-runtime` flag from the default user data section.

jonathan-innis (approved):
LGTM 🚀

website/content/en/docs/concepts/nodeclasses.md
jonathan-innis: Got it. I just did a cursory look over it so I missed that change due to all of the whitespace updates. Not sure if there is a way to reduce the number of whitespace updates that are propagated from your editor. For now, I just enabled the "Hide Whitespace" setting and was able to see it more clearly.

PR #5338, opened 2023-12-15
fix: remove event filtering generation predicate for nodeclass controller

njtran:
Fixes #N/A 

**Description**
This removes the Generation Predicate filter which drops events that don't change the spec. If a user changes the nodeclass hash, the controller won't be requeued until there's another event for this nodeclass, effectively stopping nodeclass drift until the controller is triggered again.

tested manually

- [ ] Yes, issue opened: # 
- [x] No


jonathan-innis (approved):
LGTM 🚀

PR #5339, opened 2023-12-15
docs: update docs to include budgets

njtran:
Fixes #N/A 

**Description**
Adds documentation for budgets 

make website

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/website/content/en/preview/concepts/disruption.md
+++ b/website/content/en/preview/concepts/disruption.md
@@ -183,43 +183,17 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
 
 ### Disruption Budgets
 
-You can rate limit Karpenter's disruption through the `NodePool.Spec.Disruption.Budgets`. If undefined, Karpenter will default to one budget with `nodes: 10%`. Budgets will take into account nodes that are being deleted for any reason, and will only block Karpenter from disrupting nodes voluntarily through expiration, drift, emptiness, and consolidation.
+You can rate limit Karpenter's disruption through the NodePool's `spec.disruption.budgets`. If undefined, Karpenter will default to one budget with `nodes: 10%`. Budgets will consider nodes that are actively being deleted for any reason, and will only block Karpenter from disrupting nodes voluntarily through expiration, drift, emptiness, and consolidation.
 
-#### budget.Duration + budget.Schedule
-Duration and Schedule can only be defined together. When omitted, the budget is always active. When defined, the schedule determines when a budget is active, and the duration determines how long from that point the budget is active. 
+### Nodes
+When calculating if a budget will block nodes from disruption, Karpenter lists the total number of nodes owned by a NodePool, subtracting out the nodes owned by that NodePool that are currently being deleted. If the number of nodes being deleted by Karpenter or any other processes is greater than the number of allowed disruptions, disruption for this node will not proceed.
 
-##### Schedule
-Schedule is a cronjob schedule. Generally, the cron syntax is five space-delimited values with options below. 
-Follow the [Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#writing-a-cronjob-spec) for how to follow the cron syntax. 
-
-```bash
-# ┌───────────── minute (0 - 59)
-# │ ┌───────────── hour (0 - 23)
-# │ │ ┌───────────── day of the month (1 - 31)
-# │ │ │ ┌───────────── month (1 - 12)
-# │ │ │ │ ┌───────────── day of the week (0 - 6) (Sunday to Saturday;
-# │ │ │ │ │                                   7 is also Sunday on some systems)
-# │ │ │ │ │                                   OR sun, mon, tue, wed, thu, fri, sat
-# │ │ │ │ │
-# * * * * *
-```
-
-{{% alert title="Note" color="primary" %}}
-Timezones are not supported. Most images default to UTC, but it is best practice to ensure this is the case when considering how to define your budgets. 
-{{% /alert %}}
-
-##### Duration
-Duration is a metav1.Duration, which allows compound durations with minutes and hours values such as `10h5m` or `30m` or `160h`. Since cron syntax does not accept denominations smaller than minutes, users can only define minutes or hours. 
-
-#### Nodes
-When calculating if a budget will reject a node from disruption, Karpenter will list the total number of nodes provisioned for a NodePool, and the number of those nodes currently being deleted. 
-
-If the budget is configured with a percentage value, such as `20%`, Karpenter will calculate the number of allowed disruptions as `allowed_disruptions = roundup(total * percentage) - total_deleting`. If otherwise defined as a non-percentage value, Karpenter will simply subtract the number of nodes from the total `(total * percentage) - total_deleting`. For multiple budgets in a NodePool, Karpenter will take the minimum value of each of the budgets.
+If the budget is configured with a percentage value, such as `20%`, Karpenter will calculate the number of allowed disruptions as `allowed_disruptions = roundup(total * percentage) - total_deleting`. If otherwise defined as a non-percentage value, Karpenter will simply subtract the number of nodes from the total `(total * percentage) - total_deleting`. For multiple budgets in a NodePool, Karpenter will take the minimum value (most restrictive) of each of the budgets.
 
 For example, the following NodePool with the following three budgets defines:
-- During the first 10 minutes of the day, 0 disruptions are allowed, due to the third budget.
-- When the NodePool has 25 or more nodes, only 5 disruptions are allowed, since 20% of a number larger than 25 would result in 6 nodes, due to rounding up, which is larger than 5. 
-- When the NodePool has 5 through 9 nodes, 2 disruptions would be allowed, since 20% of 9 is 1.8 which rounds up to 2. 
+- When the NodePool has less than 25 nodes nodes, 20% disruptions would be allowed. For instance, if there were 19 nodes owned by the NodePool, 4 disruptions would be allowed, rounding up from `19 * .2 = 3.8`.
+- When the NodePool has 25 or more nodes, only 5 disruptions are allowed, acting as a ceiling. When comparing to the 20% budget, 5 will be more restrictive than any percentage value, resulting in 5 allowed disruptions. 
+- The last budget selectively blocks disruptions during the first 10 minutes of the day, where 0 disruptions are allowed.
 
 ```yaml
 apiVersion: karpenter.sh/v1beta1
@@ -237,6 +211,33 @@ spec:
       schedule: "@daily"
       duration: 10m
 ```
+
+#### Schedule
+Schedule is a cronjob schedule. Generally, the cron syntax is five space-delimited values with options below, with additional special macros like `@yearly`, `monthly`, `@weekly`, `@daily`, `@hourly`.
+Follow the [Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#writing-a-cronjob-spec) for more information on how to follow the cron syntax. 
+
+```bash
+# ┌───────────── minute (0 - 59)
+# │ ┌───────────── hour (0 - 23)
+# │ │ ┌───────────── day of the month (1 - 31)
+# │ │ │ ┌───────────── month (1 - 12)
+# │ │ │ │ ┌───────────── day of the week (0 - 6) (Sunday to Saturday;
+# │ │ │ │ │                                   7 is also Sunday on some systems)
+# │ │ │ │ │                                   OR sun, mon, tue, wed, thu, fri, sat
+# │ │ │ │ │
+# * * * * *
+```
+
+{{% alert title="Note" color="primary" %}}
+Timezones are not supported. Most images default to UTC, but it is best practice to ensure this is the case when considering how to define your budgets. 
+{{% /alert %}}
+
+#### Duration
+Duration allows compound durations with minutes and hours values such as `10h5m` or `30m` or `160h`. Since cron syntax does not accept denominations smaller than minutes, users can only define minutes or hours.
+
+{{% alert title="Note" color="primary" %}}
+Duration and Schedule can only be defined together. When omitted, the budget is always active. When defined, the schedule determines when a budget is active, and the duration determines how long from that point the budget is active. 
+{{% /alert %}}
 
 ### Pod-Level Controls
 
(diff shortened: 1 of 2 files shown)


website/content/en/preview/concepts/disruption.md
```diff
@@ -177,6 +181,63 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
 
 ## Controls
 
+### Disruption Budgets
+
+You can rate limit Karpenter's disruption through the `NodePool.Spec.Disruption.Budgets`. If undefined, Karpenter will default to one budget with `nodes: 10%`. Budgets will take into account nodes that are being deleted for any reason, and will only block Karpenter from disrupting nodes voluntarily through expiration, drift, emptiness, and consolidation.
```
jonathan-innis, line 186:
```suggestion
You can rate limit Karpenter's disruption through the NodePool's `spec.disruption.budgets` field. If undefined, Karpenter will default to one budget with `nodes: 10%`. Budgets will consider nodes that are actively being deleted for any reason, and will only block Karpenter from disrupting nodes voluntarily through expiration, drift, emptiness, and consolidation.
```

website/content/en/preview/concepts/disruption.md
```diff
@@ -177,6 +181,63 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
...
+
+#### budget.Duration + budget.Schedule
```
jonathan-innis, line 188: Should this just be a note or a warning statement rather than a sub-section? I think our depth here is getting pretty deep and I could just see us having the `#### Schedule` and `#### Duration` sections

website/content/en/preview/concepts/disruption.md
```diff
@@ -177,6 +181,63 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
...
+Duration and Schedule can only be defined together. When omitted, the budget is always active. When defined, the schedule determines when a budget is active, and the duration determines how long from that point the budget is active. 
+
+##### Schedule
+Schedule is a cronjob schedule. Generally, the cron syntax is five space-delimited values with options below.
```
jonathan-innis, line 192: Should we use the word "generally" here? This is always true, right? Or am I missing a different way that crons can be written out?

website/content/en/preview/concepts/disruption.md
```diff
@@ -177,6 +181,63 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
...
+# │ │ │ ┌───────────── month (1 - 12)
+# │ │ │ │ ┌───────────── day of the week (0 - 6) (Sunday to Saturday;
+# │ │ │ │ │                                   7 is also Sunday on some systems)
+# │ │ │ │ │                                   OR sun, mon, tue, wed, thu, fri, sat
+# │ │ │ │ │
+# * * * * *
+```
+
+{{% alert title="Note" color="primary" %}}
+Timezones are not supported. Most images default to UTC, but it is best practice to ensure this is the case when considering how to define your budgets.
```
jonathan-innis, line 208: This statement is confusing to me. Can we not guarantee that we are _always_ using UTC when we are doing the cron syntax? If this isn't always the case and we aren't using the same time format across images, we're going to get customer issues for disrupting when we shouldn't be

website/content/en/preview/concepts/disruption.md
```diff
@@ -177,6 +181,63 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
...
+# │ │ │ │ │
+# * * * * *
+```
+
+{{% alert title="Note" color="primary" %}}
+Timezones are not supported. Most images default to UTC, but it is best practice to ensure this is the case when considering how to define your budgets. 
+{{% /alert %}}
+
+##### Duration
+Duration is a metav1.Duration, which allows compound durations with minutes and hours values such as `10h5m` or `30m` or `160h`. Since cron syntax does not accept denominations smaller than minutes, users can only define minutes or hours.
```
jonathan-innis, line 212: To me, this is fairly standard syntax to me. How much do we need to explain how to write this out? For the other fields that use duration, it's assumed so I wonder if we can shorten this

website/content/en/preview/concepts/disruption.md
```diff
@@ -177,6 +181,63 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
...
+
+{{% alert title="Note" color="primary" %}}
+Timezones are not supported. Most images default to UTC, but it is best practice to ensure this is the case when considering how to define your budgets. 
+{{% /alert %}}
+
+##### Duration
+Duration is a metav1.Duration, which allows compound durations with minutes and hours values such as `10h5m` or `30m` or `160h`. Since cron syntax does not accept denominations smaller than minutes, users can only define minutes or hours. 
+
+#### Nodes
+When calculating if a budget will reject a node from disruption, Karpenter will list the total number of nodes provisioned for a NodePool, and the number of those nodes currently being deleted.
```
jonathan-innis, line 215:
```suggestion
When calculating if a budget will block nodes from disruption, Karpenter lists the total number of nodes owned by a NodePool, subtracting out the nodes owned by that NodePool that are currently being deleted. If the number of nodes being deleted by Karpenter or any other processes is greater than the number of allowed disruptions, disruption for this node will not proceed.```

File: website/content/en/preview/concepts/disruption.md [hunk 7]
```diff
@@ -177,6 +181,63 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
...
+Timezones are not supported. Most images default to UTC, but it is best practice to ensure this is the case when considering how to define your budgets. 
+{{% /alert %}}
+
+##### Duration
+Duration is a metav1.Duration, which allows compound durations with minutes and hours values such as `10h5m` or `30m` or `160h`. Since cron syntax does not accept denominations smaller than minutes, users can only define minutes or hours. 
+
+#### Nodes
+When calculating if a budget will reject a node from disruption, Karpenter will list the total number of nodes provisioned for a NodePool, and the number of those nodes currently being deleted. 
+
+If the budget is configured with a percentage value, such as `20%`, Karpenter will calculate the number of allowed disruptions as `allowed_disruptions = roundup(total * percentage) - total_deleting`. If otherwise defined as a non-percentage value, Karpenter will simply subtract the number of nodes from the total `(total * percentage) - total_deleting`. For multiple budgets in a NodePool, Karpenter will take the minimum value of each of the budgets.
```
jonathan-innis, line 217:
```suggestion
If the budget is configured with a percentage value, such as `20%`, Karpenter will calculate the number of allowed disruptions as `allowed_disruptions = roundup(total * percentage) - total_deleting`. If otherwise defined as a non-percentage value, Karpenter will simply subtract the number of nodes from the total `(total * percentage) - total_deleting`. For multiple budgets in a NodePool, Karpenter will take the minimum value (most restrictive) across the budgets.
```

website/content/en/preview/concepts/disruption.md
```diff
@@ -177,6 +181,63 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
...
+
+For example, the following NodePool with the following three budgets defines:
+- During the first 10 minutes of the day, 0 disruptions are allowed, due to the third budget.
+- When the NodePool has 25 or more nodes, only 5 disruptions are allowed, since 20% of a number larger than 25 would result in 6 nodes, due to rounding up, which is larger than 5.
```
jonathan-innis, line 221: Personal preference, I think this works better logically if you order the budgets below in the way that you step through the explanation here.

Also, I think that describing how the percentage field works first as a baseline is easer to reason about. Then, you can step into talking about how to create a ceiling for this percentage value, saying that you can set a budget of 5 to block it going any higher. Finally, you can use your example of blocking disruption during certain times using the schedule/duration syntax.

website/content/en/preview/concepts/nodepools.md
```diff
@@ -138,6 +138,16 @@ spec:
     # You can choose to disable expiration entirely by setting the string value 'Never' here
     expireAfter: 720h
 
+    # Budgets control the speed at which Karpenter can scale down nodes.
```
jonathan-innis, line 141:
```suggestion
    # Budgets control the speed Karpenter can scale down nodes.
```

shorter

website/content/en/preview/concepts/disruption.md
njtran: Nope, you can use the @yearly or @daily https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#:~:text=Other%20than%20the%20standard%20syntax%2C%20some%20macros%20like%20%40monthly%20can%20also%20be%20used%3A

website/content/en/preview/concepts/disruption.md
njtran: From what I can tell, it isn't guaranteed that all images are using UTC. Anywhere I could find, it all pointed to using the default UTC, but I want to have a note here for anyone specifying timezones within their images. Thoughts?

website/content/en/preview/concepts/disruption.md
jonathan-innis: Is it worth just calling out the different ways if there is a minimal set? Rather than saying something like "most of the time" and then not offering the alternatives. If the alternative set is too large, that's fine, just seems like the statement is asking for someone asking, "but what about the non-general case?"

website/content/en/preview/concepts/disruption.md
jonathan-innis: > I could imagine users who are specifying timezone in their image be very surprised if Karpenter logs in that timezone, but then uses UTC for calculations

I don't think this is surprising to anyone, really. I think a better question is probably how upstream handles this for the cronjob-controller. My expectation was that this time was always reasoned in terms of UTC

website/content/en/preview/concepts/disruption.md
njtran: Good callout.

website/content/en/preview/concepts/disruption.md
njtran: I was thinking the link in the markdown doc has this information, so I didn't want to inflate the doc too much with essentially duplicated information. I added the small list and modified the text here.

website/content/en/preview/concepts/disruption.md
njtran: I cut down a little bit, but I was thinking two lines couldn't hurt for users coming straight to this. Thoughts?

website/content/en/preview/concepts/disruption.md
```diff
@@ -177,6 +181,64 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
...
+
+### Nodes
+When calculating if a budget will block nodes from disruption, Karpenter lists the total number of nodes owned by a NodePool, subtracting out the nodes owned by that NodePool that are currently being deleted. If the number of nodes being deleted by Karpenter or any other processes is greater than the number of allowed disruptions, disruption for this node will not proceed.
+
+If the budget is configured with a percentage value, such as `20%`, Karpenter will calculate the number of allowed disruptions as `allowed_disruptions = roundup(total * percentage) - total_deleting`. If otherwise defined as a non-percentage value, Karpenter will simply subtract the number of nodes from the total `(total * percentage) - total_deleting`. For multiple budgets in a NodePool, Karpenter will take the minimum value (most restrictive) of each of the budgets.
+
+For example, the following NodePool with the following three budgets defines:
```
jonathan-innis, line 193:
```suggestion
For example, the following NodePool with three budgets defines:
```

redundant

website/content/en/preview/concepts/disruption.md
```diff
@@ -177,6 +181,64 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
...
+- When the NodePool has less than 25 nodes nodes, 20% disruptions would be allowed. For instance, if there were 19 nodes owned by the NodePool, 4 disruptions would be allowed, rounding up from `19 * .2 = 3.8`.
```
jonathan-innis, line 194:
```suggestion
- When the NodePool has less than 25 nodes, 20% disruptions would be allowed. For instance, if there were 19 nodes owned by the NodePool, 4 disruptions would be allowed, rounding up from `19 * .2 = 3.8`.
```

website/content/en/preview/concepts/disruption.md
jonathan-innis: Discussed offline. The container filesystem itself should contain information on which timezone that Karpenter is running in. By default, this timezone should be set as UTC but this overridable through a volume mount. It makes sense to me that we would just adhere to the container timezone and stick with UTC for the vast majority of use-cases: https://stackoverflow.com/questions/57607381/how-do-i-change-timezone-in-a-docker-container

website/content/en/preview/concepts/disruption.md
```diff
@@ -177,6 +181,64 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
...
+
+{{% alert title="Note" color="primary" %}}
+Timezones are not supported. Most images default to UTC, but it is best practice to ensure this is the case when considering how to define your budgets. 
+{{% /alert %}}
+
+#### Duration
+Duration allows compound durations with minutes and hours values such as `10h5m` or `30m` or `160h`. Since cron syntax does not accept denominations smaller than minutes, users can only define minutes or hours.
+
+{{% alert title="Note" color="primary" %}}
+Duration and Schedule can only be defined together. When omitted, the budget is always active. When defined, the schedule determines when a budget is active, and the duration determines how long from that point the budget is active.
```
jonathan-innis, line 239:
```suggestion
Duration and Schedule must be defined together. When omitted, the budget is always active. When defined, the schedule determines a starting point where the budget will begin being enforced and the duration determines how long from that starting point the budget will be enforced.
```

You don't have to use this exact wording but I think saying something like the fact that the schedule is a starting point is important

website/content/en/preview/concepts/disruption.md
```diff
@@ -177,6 +181,64 @@ To enable interruption handling, configure the `--interruption-queue-name` CLI a
...
+
+#### Nodes
+When calculating if a budget will block nodes from disruption, Karpenter lists the total number of nodes owned by a NodePool, subtracting out the nodes owned by that NodePool that are currently being deleted. If the number of nodes being deleted by Karpenter or any other processes is greater than the number of allowed disruptions, disruption for this node will not proceed.
+
+If the budget is configured with a percentage value, such as `20%`, Karpenter will calculate the number of allowed disruptions as `allowed_disruptions = roundup(total * percentage) - total_deleting`. If otherwise defined as a non-percentage value, Karpenter will simply subtract the number of nodes from the total `(total * percentage) - total_deleting`. For multiple budgets in a NodePool, Karpenter will take the minimum value (most restrictive) of each of the budgets.
+
+For example, the following NodePool with three budgets defines:
```
jonathan-innis, line 193:
```suggestion
For example, the following NodePool with three budgets defines the following requirements:
```

Typically when you have lingo with a colon like this, you want the wording to naturally flow into each line.

jonathan-innis (approved):
LGTM 🚀 one optional comment

jonathan-innis (approved):
LGTM 🚀

PR #5357, opened 2023-12-18
docs: add more upgrade notes about disruption budgets

njtran:
Fixes #N/A 

**Description**
Adding more about how budgets can change the parallelism in v0.34+

make website

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/website/content/en/preview/upgrading/upgrade-guide.md
+++ b/website/content/en/preview/upgrading/upgrade-guide.md
@@ -34,7 +34,11 @@ kubectl apply -f https://raw.githubusercontent.com/aws/karpenter{{< githubRelRef
 ```
 
 ### Upgrading to v0.34.0+
-* Karpenter now supports `nodepool.spec.disruption.budgets`, which allows users to control the speed of disruption in the cluster. This will allow Karpenter to disrupt multiple batches of nodes simultaneously, which can result in overall quicker scale-down of your cluster. Karpenter will default to disrupting at most 10% of your nodes simultaneously per NodePool.   Since this requires an update to the Custom Resource, before upgrading, you should re-apply the new updates to the CRDs. Check out [Disruption Budgets]({{<ref "../concepts/disruption#disruption-budgets" >}}) for more. 
+* Karpenter now supports `nodepool.spec.disruption.budgets`, which allows users to control the speed of disruption in the cluster. Since this requires an update to the Custom Resource, before upgrading, you should re-apply the new updates to the CRDs. Check out [Disruption Budgets]({{<ref "../concepts/disruption#disruption-budgets" >}}) for more. 
+* With Disruption Budgets, Karpenter will disrupt multiple batches of nodes simultaneously, which can result in overall quicker scale-down of your cluster. Before v0.34, Karpenter had a hard-coded parallelism limit for each type of disruption. In v0.34, Karpenter will now disrupt at most 10% of nodes for a given NodePool. There is no setting that will be perfectly equivalent with the behavior prior to v0.34. When considering how to configure your budgets, please refer to the following limits for versions prior to v0.34:
+  * `Empty Expiration / Empty Drift / Empty Consolidation`: infinite parallelism 
+  * `Non-Empty Expiration / Non-Empty Drift / Single-Node Consolidation`: one node at a time
+  * `Multi-Node Consolidation`: max 100 nodes
 
 ### Upgrading to v0.33.0+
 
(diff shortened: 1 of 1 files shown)


website/content/en/preview/upgrading/upgrade-guide.md
```diff
@@ -34,7 +34,7 @@ kubectl apply -f https://raw.githubusercontent.com/aws/karpenter{{< githubRelRef
 ```
 
 ### Upgrading to v0.34.0+
-* Karpenter now supports `nodepool.spec.disruption.budgets`. Since this requires an update to the Custom Resource, before upgrading, you should re-apply the new updates to the CRDs. Check out [Disruption Budgets]({{<ref "../concepts/disruption#disruption-budgets" >}}) for more.
+* Karpenter now supports `nodepool.spec.disruption.budgets`, which allows users to control the speed of disruption in the cluster. This will allow Karpenter to disrupt multiple batches of nodes simultaneously, which can result in overall quicker scale-down of your cluster. Karpenter will default to disrupting at most 10% of your nodes simultaneously per NodePool.   Since this requires an update to the Custom Resource, before upgrading, you should re-apply the new updates to the CRDs. Check out [Disruption Budgets]({{<ref "../concepts/disruption#disruption-budgets" >}}) for more.
```
jonathan-innis, line 37:
```suggestion
* Karpenter now supports `nodepool.spec.disruption.budgets`, which allows users to control the speed of disruption in the cluster. This will allow Karpenter to disrupt multiple batches of nodes simultaneously, which can result in overall quicker scale-down of your cluster. Karpenter will default to disrupting at most 10% of your nodes simultaneously per NodePool. Since this requires an update to the Custom Resource, before upgrading, you should re-apply the new updates to the CRDs. Check out [Disruption Budgets]({{<ref "../concepts/disruption#disruption-budgets" >}}) for more. 
```

Not sure why it spaced so much here.

jonathan-innis, line 37: One other thing: I'm not sure this makes it super clear IMO that when they apply this new behavior, this will be different from the existing behavior. To me, I'd make a callout that the existing behavior is that we disrupted a single node at a time and then tell them how they can keep the existing behavior using a budget that models that behavior

njtran: There's no setting that would be perfectly equivalent, considering it was previously artificially bounded per-disruption method. I can call out which methods would be different. WDYT?

jonathan-innis: I wonder if we just add detail here, then, that describes what the old behavior was here in a little more detail. Mainly want to callout here that when you upgrade to this newer version, you will see different behavior which will result in more aggressive disruption in general (unless we are talking about emptiness).

website/content/en/preview/upgrading/upgrade-guide.md
```diff
@@ -34,7 +34,11 @@ kubectl apply -f https://raw.githubusercontent.com/aws/karpenter{{< githubRelRef
 ```
 
 ### Upgrading to v0.34.0+
-* Karpenter now supports `nodepool.spec.disruption.budgets`. Since this requires an update to the Custom Resource, before upgrading, you should re-apply the new updates to the CRDs. Check out [Disruption Budgets]({{<ref "../concepts/disruption#disruption-budgets" >}}) for more.
+* Karpenter now supports `nodepool.spec.disruption.budgets`, which allows users to control the speed of disruption in the cluster. Since this requires an update to the Custom Resource, before upgrading, you should re-apply the new updates to the CRDs. Check out [Disruption Budgets]({{<ref "../concepts/disruption#disruption-budgets" >}}) for more. 
+* With Disruption Budgets, Karpenter will disrupt multiple batches of nodes simultaneously, which can result in overall quicker scale-down of your cluster. Before v0.34, Karpenter had a hard-coded parallelism limit for each type of disruption. In v0.34, Karpenter will now disrupt at most 10% of nodes for a given NodePool. There is no setting that will be perfectly equivalent with the behavior prior to v0.34. When considering how to configure your budgets, please refer to the following limits for versions prior to v0.34:
```
jonathan-innis, line 38: nit: I question whether these should be separate bullets since they apply to the same concept

jonathan-innis (approved):

website/content/en/preview/upgrading/upgrade-guide.md
njtran: I think they make sense as separate bullets since they're two separate implications/considerations. One is to upgrade CRDs, one is to account for how it changes disruption, which I think is more clearly shown as two separate things if you have two separate bullet points.

Issue #2649, opened 2022-10-11
JSON IAM Controller Policy


johnwesley:
### Is an existing page relevant?

[getting started](https://karpenter.sh/v0.17.0/getting-started/)

### What karpenter features are relevant?

Install and getting started documentation.

### How should the docs be improved?

By having the karpenter-controller IAM policy, and a description of, readily available in `json` format for a user to review and maybe take away for a more custom install. Perhaps https://github.com/aws/karpenter/tree/main/hack/iam would be a good place?

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ellistarn:
We're wary of exposing something like this, since users will take a dependency on it and we will have an implicit contract to maintain it. We currently expose this information via the getting started guide, and are aiming to keep that as the source of truth to minimize the surface of things we need to maintain/test.

FernandoMiguel:
@ellistarn i disagree.
folks, like myself, community members, already try to keep up with PRs here to update https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/2cd89ecc/modules/kubernetes-addons/karpenter/data.tf#L7-L25 and https://github.com/terraform-aws-modules/terraform-aws-iam/blob/c6adf42/modules/iam-role-for-service-accounts-eks/policies.tf#L508
when new policies like [this](https://github.com/aws/karpenter/blob/9f083523bad1ab2430bd45ca6e0d937d2ca6d9a0/website/content/en/preview/getting-started/getting-started-with-eksctl/cloudformation.yaml#L67) appear , we have to scramble to update those.

having a source of truth in code would be helpful to refer too.

Nuru:
@ellistarn I disagree, also. The Getting started guide just says to create an IRSA, it does not specify what policy to create or even give a reference to where to find the required permissions. I found this issue because the code I am using is outdated, Karpenter is getting "access denied" errors, and I'm still looking for what I need to add. 

For example, [the `eksctl` section says](https://karpenter.sh/v0.18.0/getting-started/getting-started-with-eksctl/#create-the-karpentercontroller-iam-role)

```
eksctl create iamserviceaccount \
[...]

PR #5361, opened 2023-12-18
docs: Update v1beta1-controller-policy.json

huynhj93:
Fixed docs by adding `arn:${AWS::Partition}:ec2:${AWS::Region}:*:spot-instances-request/*` resources for AllowScopedResourceCreationTagging and AllowScopedEC2InstanceActionsWithTags

**Description**
With 0.33, the controller policy in the upgrading guide needs to be updated. Specifically `arn:${AWS::Partition}:ec2:${AWS::Region}:*:spot-instances-request/*` needs to be added as a resource to certain policy statements.

Referenced https://github.com/aws/karpenter-provider-aws/blob/v0.33.0/website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml#L61 and https://github.com/aws/karpenter-provider-aws/blob/v0.33.0/website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml#L86. Let me know if i need to change anything else.

### DIFF
--- a/website/content/en/v0.33/upgrading/v1beta1-controller-policy.json
+++ b/website/content/en/v0.33/upgrading/v1beta1-controller-policy.json
@@ -25,7 +25,8 @@
         "arn:${AWS_PARTITION}:ec2:${AWS_REGION}:*:instance/*",
         "arn:${AWS_PARTITION}:ec2:${AWS_REGION}:*:volume/*",
         "arn:${AWS_PARTITION}:ec2:${AWS_REGION}:*:network-interface/*",
-        "arn:${AWS_PARTITION}:ec2:${AWS_REGION}:*:launch-template/*"
+        "arn:${AWS_PARTITION}:ec2:${AWS_REGION}:*:launch-template/*",
+        "arn:${AWS::Partition}:ec2:${AWS::Region}:*:spot-instances-request/*"
       ],
       "Action": [
         "ec2:RunInstances",
@@ -49,7 +50,8 @@
         "arn:${AWS_PARTITION}:ec2:${AWS_REGION}:*:instance/*",
         "arn:${AWS_PARTITION}:ec2:${AWS_REGION}:*:volume/*",
         "arn:${AWS_PARTITION}:ec2:${AWS_REGION}:*:network-interface/*",
-        "arn:${AWS_PARTITION}:ec2:${AWS_REGION}:*:launch-template/*"
+        "arn:${AWS_PARTITION}:ec2:${AWS_REGION}:*:launch-template/*",
+        "arn:${AWS::Partition}:ec2:${AWS::Region}:*:spot-instances-request/*"
       ],
       "Action": "ec2:CreateTags",
       "Condition": {


jonathan-innis:
@huynhj93 This policy isn't intended for continual usage. It's only intended for usage within the migration path onto v0.32.x. For an up-to-date policy, please refer to https://github.com/aws/karpenter-provider-aws/blob/main/website/content/en/docs/getting-started/getting-started-with-karpenter/cloudformation.yaml. 

As an aside, we might consider renaming the policy here since the fact that it just says that it's the `v1beta1-controller-policy.json` is confusing people. If we renamed it to something like `v1beta1-0.32.x-controller-policy.json

john-kajabi:
Ah, then yes I'd propose to get rid of it, rename it as suggested, or update it along with the cloudformation. I followed the migration guide for 0.32 and stumbled upon that json policy, and thought it was intuitive to do the same thing for 0.33. I would propose the last change to update it along with cloudformation because json is more universal.

huynhj93:
Ah, then yes I'd propose to get rid of it, rename it as suggested, or update it along with the cloudformation. I followed the migration guide for 0.32 and stumbled upon that json policy, and thought it was intuitive to do the same thing for 0.33. I would propose the last change to update it along with cloudformation because json is more universal. Renaming it to `website/content/en/v0.33/upgrading/v1beta1-032.x-controller-policy-.json` is downright confusing, so I'd propose to just get rid of it in that case.

jonathan-innis:
> Ah, then yes I'd propose to get rid of it, rename it as suggested, or update it along with the cloudformation

ksquaredkey:
Well, I lost over a day to this landmine. I was going from v0.19 to v0.33.1 and taking the opportunity to move my v1alpha5 resources to v1beta1.  I don't use CloudFormation. Been doing AWS for 13+ years and got burned a decade ago when CloudFormation was basically abandon-ware and grossly lagged behind API changes.  Been using pure boto scripts and then Terraform and never looked back.  Most of my co-workers over the years have been the same. 

Anyways, whether or not the policy was only intended for v0.32 or not, it is already in the "v0.33" directory, as well as "preview" directory, as well as the "docs" directory.  And it is in v0.33 release tag as the policy to use when migration to v1beta1.  So regardless that this may have been intended for only v0.32-specific upgrades, it is NOT currently documented as such.  

Personally, I think the correct IAM policy JSON for v1beta1 should be documented as a standalone reference in the mainline installation documentation, independent of management tool a user might be using and independent of
whether it is a fresh installation or a migration. JSON is the lingua franca for IAM Policies, so having the canonical
fragment embedded in the CoudFormation YAML is just wrong.  Sure, anyone to Kubernetes and karpenter is going to be conversant in YAML, but that just isn't how IAM policies are documented in the vast majority of AWS documentation.

Bottom line - this needs to be fixed  before more people step on this landmine!

jonathan-innis:
>  I think the correct IAM policy JSON for v1beta1 should be documented as a standalone reference in the mainline installation documentation, independent of management tool a user might be using and independent of
whether it is a fresh installation or a migration

jonathan-innis:
>  I think the correct IAM policy JSON for v1beta1 should be documented as a standalone reference in the mainline installation documentation, independent of management tool a user might be using and independent of
whether it is a fresh installation or a migration

There's a discussion about adding JSON support for the policy here #2649. I agree that it would be nice to have the go-to JSON policy documented in our upstream docs, more in the original issue for context.

For now (until we get a JSON version of the policy documented and the automation built around it), I think there's enough here to indicate that we should drop the v1beta1 migration and the controller policy from newer versions of the docs and use redirects for anyone trying to access the migration guide in a newer version of the docs to avoid any breaking changes here.

ksquaredkey:
>We're not advising users to move from an older version (other than v0.32.x) to v0.33.x outright since v0.32.x is the only version that would be able to reason in both the alpha and beta APIs so I'm not sure that it makes sense to document this detail in versions outside of v0.32.x anyways (since this is the version that we expect everyone to move through during the migration).

If that is truly the expectation, the documentation has failed to make that clear and seems to imply the opposite by having the policy file in a directory named "v0.33".  Personally, I see no reason for force users to go through v0.32 in order to go beyond.  I am going to flag-day it from 0.19 to v0.33, now that I have the correct policy that corresponds to v0.33.1 that I was able to include in my Terraform.

I read through #2649 and I can agree with having a single source of truth for the current IAM policy needed for the IRSA Role that the user will be creating is a great idea.  Having that single source of truth hidden in the middle of a *example* YAML file is NOT that.  It should be it's own standalone JSON file, just as it is for the vast majority of other things that need policies for IRSAs defined (see documentation for [aws-efs-csi-driver](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/docs/iam-policy-example.json), [aws-ebs-csi-driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/example-iam-policy.json) , etc, etc).  If anything, the CloudFormation file should REFERENCE that file to get the canonical policy value, keeping the single source of truth.

jonathan-innis:
> If anything, the CloudFormation file should REFERENCE that file to get the canonical policy value, keeping the single source of truth

Agree, as mentioned in #2649, we'd definitely accept a change to our docs that pointed us in that direction, so long as we can make sure to have a single place where the policy is listed 👍🏼

ksquaredkey:
For those, following the bouncing conversation, I've replied and continued discussion of single canonical source for IAM policy in #2649

jonathan-innis:
I opened #5419 to execute on a few things that we discussed, including calling out a bit more clearly in the upgrade guide that v0.32 is the alpha/beta migration version and should be moved-to ahead of the other versions after it

jonathan-innis:
Closing this one since #5419 merged.

PR #5363, opened 2023-12-18
docs: update v1beta1 upgrade guide to use expiration instead of drift

njtran:
Fixes #N/A 

**Description**
For users that were upgrading from before v0.31 to then using the migration guide, it was possible that nodes would not properly drift since the feature that the rolling relied on was only introduced for nodes created with a Karpenter version v0.30. 

This changes Drift to use expiration instead so that it can work for all users, regardless of where they're beginning their migration path.

make website

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/website/content/en/docs/upgrading/v1beta1-migration.md
+++ b/website/content/en/docs/upgrading/v1beta1-migration.md
@@ -144,7 +144,7 @@ The [`karpenter-convert`](https://github.com/aws/karpenter/tree/release-v0.32.x/
 
     1. Periodic Rolling with [Expiration]({{< relref "../concepts/disruption#automated-methods" >}}):
        - Add the following taint to the old Provisioner: `karpenter.sh/legacy=true:NoSchedule`
-       - Set the Expiration in your Provisioner to a small value like `ttlSecondsUntilExpired: 300` to mark all nodes older than 5 minutes as expired.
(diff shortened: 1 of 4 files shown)


website/content/en/docs/upgrading/v1beta1-migration.md
```diff
@@ -142,12 +142,12 @@ The [`karpenter-convert`](https://github.com/aws/karpenter/tree/release-v0.32.x/
 
 12. Roll over nodes: With the new NodePool yaml in hand, there are several ways you can begin to roll over your nodes to use the new NodePool:
 
-    1. Periodic Rolling with [Drift]({{< relref "../concepts/disruption#drift" >}}): Enable [drift]({{< relref "../reference/settings/#feature-gates" >}}) in Karpenter, then do the following:
+    1. Periodic Rolling with [Expiration]({{< relref "../concepts/disruption#automated-methods" >}}):
```
jmdeal, line 145: Would it be possible to link to the Provisioner docs from a previous version here rather than the current NodePool docs? I believe the answer is no, but the reason I ask is it may be confusing to say later to set `ttlSecondsUntilExpired` and then link to the NodePool docs which have a different set of fields for disruption.

njtran: Adding a note to make sure there's less confusion about how to do this.

jmdeal (approved):
LGTM 🚀

jonathan-innis (approved):
LGTM 🚀

PR #5388, opened 2023-12-25
fix: enable webhook as default.

HenItzhaki:
Fixes #N/A 

**Description**

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -134,7 +134,7 @@ controller:
     port: 8081
 webhook:
   # -- Whether to enable the webhooks and webhook permissions.
-  enabled: false
+  enabled: true
   # -- The container port to use for the webhook.
   port: 8443
   metrics:


jonathan-innis:
@HenItzhaki Do you have context around why you want to make this change? We made this change after we dropped alpha API since for newer versions of Kubernetes (>= 1.25) CommonExpressionLanguage (CEL) validation for CustomResourceDefinitions is enabled, which means we have no need for the webhooks. If you are still using an earlier version of Kubernetes, you can still enable the webhooks through `--set webhook.enabled=true`

jonathan-innis:
@HenItzhaki I'm going to close this for now since I don't think we'll be taking this change. If you want to enable the webhooks for your specific configuration, you can always do so as described above with `--set webhook.enabled=true`

ricardclau:
Hi, I also encountered this problem and had to enable this value again in our deployments. We are running EKS 1.27 but karpenter was still trying to hit the webhook which obviously is disabled by default. I have checked the docs and chart values but I did not find anything to make karpenter use CEL (or at least, stop trying to hit the webhook) and could not find anything. Can you provide some light on what extra steps are needed to use CEL?

PR #5389, opened 2023-12-25
docs: Placement Group Design Document

lorrrrrrrenzo:
Addresses #3324 
Requested here: https://github.com/aws/karpenter-provider-aws/pull/5307#discussion_r1424284463

**Description**
This is a design document for the addition of EC2 Placement Group support to Karpenter, as requested by @jonathan-innis. 

N/A

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/designs/placement-groups.md
+++ b/designs/placement-groups.md
@@ -1,0 +1,65 @@
+# Placement Group Configuration
+
+## Goals
+* Allowing a user to configure EC2 Placement Group strategies from an `EC2NodeClass` resource
+* Allow the Placement Group strategies to be developed independently from one another
+
+## Background
+At present, there is no way to configure an [EC2 Placement Group](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/placement-groups.html) with Karpenter. The addition of this feature will expose an important feature of EC2 to Karpenter users, expanding the set of use-cases the tool is appropriate for.
+
+## Proposed Design
+To enable the configuration of EC2 Placement Groups with Karpenter, I propose the addition of a `.spec.placementStrategy` value in the `karpenter.k8s.aws/v1beta1/EC2NodeClass` resource. This new value will have three child fields --  `cluster`, `partition`, and `spread` -- only one of which can be enabled for a given `EC2NodeClass` instance, which is in line with the EC2 API. 
+
+Given the simple and self-contained nature of EC2 Placement Group configurations, my sense is that it makes litle sense to enable the import of preexisting Placement Group configurations. This would add implementation complexity without actually enabling additional features for the user. Instead, Karpenter should completely manage the lifecycle of EC2 Placement Group resources, tying their lifetimes to the lifetime of the `.spec.placementStrategy` value in the supervising EC2NodeClass resource.
+
+
+The following YAML snippets depict how this would look in practice, including one case where the `EC2NodeClass` resource *should* be rejected by the Karpenter server.
+
+```yaml
+---
+apiVersion: karpenter.k8s.aws/v1beta1
+kind: EC2NodeClass
+metadata:
+  name: example-cluster
+spec:
+  # ...
+  placementStrategy:
+    cluster: { enabled: true } # No configuration ields
+---
+apiVersion: karpenter.k8s.aws/v1beta1
+kind: EC2NodeClass
+metadata:
+  name: example-spread
+spec:
+  # ...
+  placementStrategy:
+    spread:
+      enabled: true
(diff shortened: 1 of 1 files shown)


engedaam (commented):
/karpenter snapshot

amanueng (commented):
/karpenter snapshot

designs/placement-groups.md
```diff
@@ -0,0 +1,65 @@
...
+---
+apiVersion: karpenter.k8s.aws/v1beta1
+kind: EC2NodeClass
+metadata:
+  name: example-spread
+spec:
+  # ...
+  placementStrategy:
+    spread:
+      enabled: true
```
jmdeal, line 37: What is the purpose of the enabled field? I would assume specifying the placement strategy would implicitly enable it, but I'm going to guess it's needed for cluster since it wouldn't have any other fields?  Or does enabled refer to something different here?

designs/placement-groups.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Proposed Design
+To enable the configuration of EC2 Placement Groups with Karpenter, I propose the addition of a `.spec.placementStrategy` value in the `karpenter.k8s.aws/v1beta1/EC2NodeClass` resource. This new value will have three child fields --  `cluster`, `partition`, and `spread` -- only one of which can be enabled for a given `EC2NodeClass` instance, which is in line with the EC2 API. 
+
+Given the simple and self-contained nature of EC2 Placement Group configurations, my sense is that it makes litle sense to enable the import of preexisting Placement Group configurations. This would add implementation complexity without actually enabling additional features for the user. Instead, Karpenter should completely manage the lifecycle of EC2 Placement Group resources, tying their lifetimes to the lifetime of the `.spec.placementStrategy` value in the supervising EC2NodeClass resource.
```
engedaam, line 13: Have you made any consideration as to how drift will operate for this field? Seems like as soon as a user adds a value here they will automatically be drifted, even if there is not any change to the placementStrategy

designs/placement-groups.md
```diff
@@ -0,0 +1,65 @@
...
+  placementStrategy:
+    cluster: { enabled: true } # No configuration fields
+---
+apiVersion: karpenter.k8s.aws/v1beta1
+kind: EC2NodeClass
+metadata:
+  name: example-spread
+spec:
+  # ...
+  placementStrategy:
```
engedaam, line 35: Have you considered using `.spec.placementStrategy` as a string? such that the only values that can be are `cluster`, `partition`, and `spread`

designs/placement-groups.md
```diff
@@ -0,0 +1,65 @@
...
+      enabled: true
+      spreadLevel: host # constraints: rack | host
+---
+apiVersion: karpenter.k8s.aws/v1beta1
+kind: EC2NodeClass
+metadata:
+  name: example-partition
+spec:
+  # ...
+  placementStrategy:
```
engedaam, line 46: Will the field have a default value? if so what will that be

designs/placement-groups.md
```diff
@@ -0,0 +1,65 @@
+# Placement Group Configuration
+
+## Goals
```
jonathan-innis, line 3: I think we should write this design doc in a way that discusses the use-cases that we are trying to address through adding support for placement groups to the EC2NodeClass. From that, we can derive what the best API looks like here. 

Without the use-cases here, how can we say that users don't want to bring their own placement groups and want Karpenter to fully manage them?

designs/placement-groups.md
jonathan-innis, line 13: Can you expand on this a bit? I don't think it's immediately clear to the reader why that's the case. I can see the argument that placement groups (at least for spread and partition) assume you are defining spread at the instance-level rather than having the flexibility of containers which allows you to spread your application out on a level just below the instance-level.

jonathan-innis, line 13: I think I'd like to see a larger discussion about options here given the use-cases that we should write-up up top. Realistically, placement groups is a single EC2 feature, but the use-cases for cluster placement groups look _really_ different from the use-cases for spread and partition. Maybe it's worth a larger background section that describes what the use-cases are for each of these different types, what the user problem is that we are trying to solve, and then consider options in Karpenter that could make the best UX for solving this. Happy to discuss offline more to drive towards something that works for everyone here.

sgundapu:
support already available in Karpenter release v1.11.0

https://github.com/aws/karpenter-provider-aws/releases/tag/v1.11.0

PR #5419, opened 2024-01-02
docs: Only put v1beta1 migration on v0.32 docs

jonathan-innis:
Relates to and supersedes the change here: #5361

**Description**

This updates the docs so that we only publish the v1beta1 migration guide as part of the v0.32 version of the docs. This clarifies to users who may look at the website content in the source that this controller policy is only for the migration version (v0.32) and is not meant to be applied to subsequent versions of the controller release.

This also adds more clarifications through alerts in the upgrade guide that v0.32 is intended to be the "train stop" version for everyone to move through on the way to v0.32 and that you shouldn't run past v0.32 without first going through v0.32 to migrate your alpha apis to beta.

- [ ] Yes, issue opened: # 
- [ ] No


jmdeal (approved):
LGTM 🚀

PR #5437, opened 2024-01-08
feat: Add AssociatePublicIpAddress to EC2NodeClass

myaser:
Fixes aws/karpenter#2026

**Description**
Add an AssignPublicIpAddress field to the `nodeclass` 

unit and integration tests were created;

- [ ] Yes, issue opened: # 
- [ ] No


pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -282,9 +282,10 @@ func (p *Provider) generateNetworkInterfaces(options *amifamily.LaunchTemplate)
 			return &ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest{
 				NetworkCardIndex: lo.ToPtr(int64(i)),
 				// Some networking magic to ensure that one network card has higher priority than all the others (important if an instance needs a public IP w/o adding an EIP to every network card)
-				DeviceIndex:   lo.ToPtr(lo.Ternary[int64](i == 0, 0, 1)),
-				InterfaceType: lo.ToPtr(ec2.NetworkInterfaceTypeEfa),
-				Groups:        lo.Map(options.SecurityGroups, func(s v1beta1.SecurityGroup, _ int) *string { return aws.String(s.ID) }),
+				DeviceIndex:              lo.ToPtr(lo.Ternary[int64](i == 0, 0, 1)),
+				InterfaceType:            lo.ToPtr(ec2.NetworkInterfaceTypeEfa),
+				Groups:                   lo.Map(options.SecurityGroups, func(s v1beta1.SecurityGroup, _ int) *string { return aws.String(s.ID) }),
+				AssociatePublicIpAddress: options.AssignPublicIpAddress,
```
jmdeal, line 288: Something to consider here is that this can cause instance launches to fail if AssociatePublicIpAddress is true and the instance was configured with multiple EFAs. The two options I see are allowing the launch to fail or ignoring the specified AssociatePublicIpAddress value when launching instances with multiple network interfaces. Either way we should probably add a warning in the docs about this configuration footgun.

website/content/en/preview/concepts/nodeclasses.md
```diff
@@ -1,4 +1,4 @@
----
+ ---
```
jmdeal, line 1: Once this is ready to merge doc changes will also need to be reflected in the `website/content/en/docs` path. No rush to add them there, they'll just have to be before merge.

pkg/providers/launchtemplate/launchtemplate.go
jmdeal: Discussed with some people offline and feel that failing the launch is probably the right move here. Users should be able to use a separate `EC2NodeClass` for EFA workloads so as long as this edge case is called out appropriately in the docs I'm not too worried about it. I think this is more expected than the option of ignoring the flag and successfully launching.

website/content/en/preview/concepts/nodeclasses.md
myaser: I see, is it to version `v0.33`?

jmdeal: Since this is a new feature it wouldn't go in until the next minor version bump so no. There should be two paths where doc changes are reflected, `website/content/en/preview` (what you have now), and `website/content/en/docs`.

pkg/providers/pricing/zz_generated.pricing_aws.go
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 package pricing
 
-// generated at 2023-12-11T13:07:33Z for us-east-1
```
jmdeal, line 19: Note: we'll want to drop these generated pricing changes by the time it's ready to merge.

jmdeal (commented):
/karpenter snapshot

pkg/providers/launchtemplate/suite_test.go
```diff
@@ -1523,7 +1523,22 @@ var _ = Describe("LaunchTemplates", func() {
 				Expect(*input.LaunchTemplateData.NetworkInterfaces[0].AssociatePublicIpAddress).To(BeFalse())
 			})
 
-			It("should not explicitly set 'AssignPublicIPv4' when the subnets are configured to assign public IPv4 addresses", func() {
+			It("should overwrite 'AssociatePublicIPAddress' to true when specified by user in the EC2NodeClass", func() {
```
jmdeal, line 1526: Can we also include tests for the `false` and `nil` cases here?

website/content/en/preview/concepts/nodeclasses.md
jmdeal: Sorry, I was mistaken here. I thought `website/content/en/docs` is an alias for preview but it's actually an alias for the current version (v0.33). You only need the changes in preview.

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -107,11 +107,11 @@ func (p *Provider) EnsureAll(ctx context.Context, nodeClass *v1beta1.EC2NodeClas
 	p.Lock()
 	defer p.Unlock()
 
-	options, err := p.createAMIOptions(ctx, nodeClass, lo.Assign(nodeClaim.Labels, map[string]string{corev1beta1.CapacityTypeLabelKey: capacityType}), tags)
+	opts, err := p.createAMIOptions(ctx, nodeClass, lo.Assign(nodeClaim.Labels, map[string]string{corev1beta1.CapacityTypeLabelKey: capacityType}), tags)
```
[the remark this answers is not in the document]
myaser: but it collides with the imported package named options

anyway, I am ok keeping it as options

website/content/en/preview/concepts/nodeclasses.md
```diff
@@ -306,6 +310,16 @@ spec:
...
 ```
 
+## spec.associatePublicIpAddress
+
+a boolean field to control whether the instances created by karpenter for this node class will have an associated public IP address. 
+when not specified, the node will have public IP adders if it is launched in a subnet with public IP auto-assign enabled `MapPublicIpOnLaunch=true`.
+
+{{% alert title="Note" color="warning" %}}
+setting this field to true can cause instance launches to fail if the instance was configured with multiple EFAs. the two options can not be used together
+{{% /alert %}}
```
jmdeal, line 320:
```suggestion
A boolean field that controls whether instances created by Karpenter for this EC2NodeClass will have an associated public IP address. This overrides the `MapPublicIpOnLaunch` setting applied to the subnet the node is launched in. If this field is not set, the `MapPublicIpOnLaunch` field will be respected.

{{% alert title="Note" color="warning" %}}
If `associatePublicIPAddress` is set to `true` and a NodePool referencing that EC2NodeClass is used to provision a Node for a NodeClaim requesting multiple `vpc.amazonaws.com/efa` resources, the instance launch will fail. A public IP address may only be associated with a node at launch if a single network interface is configured. 
{{% /alert %}}
```

Could you also move this block to the end of this section so the order matches the EC2NodeClass example at the top of the page?

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -41,6 +41,9 @@ type EC2NodeClassSpec struct {
 	// +kubebuilder:validation:MaxItems:=30
 	// +required
 	SecurityGroupSelectorTerms []SecurityGroupSelectorTerm `json:"securityGroupSelectorTerms" hash:"ignore"`
+	// AssociatePublicIPAddress controls if public IP addresses are assigned to instances that are launched by the launch template.
```
njtran, line 44: Does this make sense? the fact that we generate a launch template and is passed through aws is more of an implementation detail.
```suggestion
	// AssociatePublicIPAddress controls if public IP addresses are assigned to instances that are launched with the nodeclass.
```

website/content/en/preview/concepts/nodeclasses.md
```diff
@@ -859,6 +864,14 @@ spec:
   detailedMonitoring: true
 ```
 
+## spec.associatePublicIPAddress
+
+A boolean field that controls whether instances created by Karpenter for this EC2NodeClass will have an associated public IP address. This overrides the `MapPublicIpOnLaunch` setting applied to the subnet the node is launched in. If this field is not set, the `MapPublicIpOnLaunch` field will be respected.
+
+{{% alert title="Note" color="warning" %}}
+If `associatePublicIPAddress` is set to `true` and a NodePool referencing that EC2NodeClass is used to provision a Node for a NodeClaim requesting multiple `vpc.amazonaws.com/efa` resources, the instance launch will fail. A public IP address may only be associated with a node at launch if a single network interface is configured.
```
ellistarn, line 872: Why can't we assign a public IP to the primary ENI, and use secondary enis for the EFAs?

myaser: I think it could be done!
I thought this was undesirable from design perspective
https://github.com/aws/karpenter-provider-aws/pull/5437#discussion_r1445231762

jmdeal: I think we could do that but I'm not sure that solves the problem laid out in my original comment. My understanding is that when launching with multiple ENIs (regardless of EFA) you can not set this field to true. Would there be any other advantage of configuring the primary ENI to be a non-EFA ENI?

jmdeal: The only reason it's called out specifically for EFA is currently this is the only path for Karpenter to launch an instance with multiple ENIs.

test/suites/integration/network_interface_test.go
```diff
@@ -0,0 +1,66 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/samber/lo"
+	"sigs.k8s.io/karpenter/pkg/test"
+
+	"github.com/aws/karpenter-provider-aws/pkg/apis/v1beta1"
+)
+
+var _ = Describe("NetworkInterfaces", func() {
+
+	DescribeTable(
```
jmdeal, line 28: Sorry to waffle on what's included here but after discussing with @jonathan-innis offline we thought the time trade-off may not be worth the additional test coverage here to test each permutation. Since the functional tests already cover that I think it would be reasonable to reduce this to just testing a single permutation, e.g. private subnets with a public IP.

website/content/en/preview/concepts/nodeclasses.md
jmdeal: I'm having a hard time finding docs to back this at the moment but this is the error message from launching multiple non-EFA ENIs:
```
"error":"launching nodeclaim, creating instance, with fleet error(s), InvalidParameterCombination: The associatePublicIPAddress parameter cannot be specified when launching with multiple network interfaces."
```

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -282,15 +285,16 @@ func (p *Provider) generateNetworkInterfaces(options *amifamily.LaunchTemplate)
 			return &ec2.LaunchTemplateInstanceNetworkInterfaceSpecificationRequest{
 				NetworkCardIndex: lo.ToPtr(int64(i)),
 				// Some networking magic to ensure that one network card has higher priority than all the others (important if an instance needs a public IP w/o adding an EIP to every network card)
-				DeviceIndex:   lo.ToPtr(lo.Ternary[int64](i == 0, 0, 1)),
-				InterfaceType: lo.ToPtr(ec2.NetworkInterfaceTypeEfa),
-				Groups:        lo.Map(options.SecurityGroups, func(s v1beta1.SecurityGroup, _ int) *string { return aws.String(s.ID) }),
+				DeviceIndex:              lo.ToPtr(lo.Ternary[int64](i == 0, 0, 1)),
+				InterfaceType:            lo.ToPtr(ec2.NetworkInterfaceTypeEfa),
+				Groups:                   lo.Map(options.SecurityGroups, func(s v1beta1.SecurityGroup, _ int) *string { return aws.String(s.ID) }),
+				AssociatePublicIpAddress: options.AssociatePublicIPAddress,
```
jonathan-innis, line 291: Let's confirm that this doesn't break behavior with EFA when this is set to false. Since we auto-set this when all of the subnets are private in the subnetSelector, I want to make sure that this isn't a breaking change where we all of sudden start failing for users that are leveraging EFA

jonathan-innis: If this _does_ work, we should add a comment here on why this is alright to do with `CheckAnyPublicIPAssociations`. I'd really like to see that function get pulled out sooner than later since it can cause some funky interactions here with things like EFA

website/content/en/preview/concepts/nodeclasses.md
myaser: if I understand this correctly, this happens when all interfaces are configured with `AssociatePublicIPAddress` = true

I think it is not a problem when only the primary one is set to have `AssociatePublicIPAddress` while the rest are not. I will try to validate this path

ellistarn: SGTM!

myaser: I was wrong 
I created a test case where I set the AssociatePublicIPAddress to the first network interface and the create fleet instance failed
also, it is mentioned in the documentation [here](https://docs.aws.amazon.com/cli/latest/reference/ec2/request-spot-fleet.html
)
Screenshot 2024-01-25 at 14 28 33

pkg/providers/launchtemplate/launchtemplate.go
myaser: I created an e2e test case to validate this, and it failed. so I decided to remove this line.

if the pod is requesting EFA, the associatePublicIPAddress is completely ignored

website/content/en/preview/concepts/nodeclasses.md
myaser: I also updated the docs

test/suites/integration/network_interface_test.go
```diff
@@ -0,0 +1,89 @@
...
+			if useEFA {
+				Expect(instance.NetworkInterfaces).To(HaveLen(2))
+				Expect(instance.NetworkInterfaces[0].InterfaceType).To(HaveValue(Equal("efa")))
+				Expect(instance.NetworkInterfaces[1].InterfaceType).To(HaveValue(Equal("efa")))
+			}
+		},
+		Entry("AssociatePublicIPAddress not specified by the user while using a private subnet", false, true, nil),
+		Entry("AssociatePublicIPAddress set true by user while using a private subnet", false, true, lo.ToPtr(true)),
+
+		Entry("AssociatePublicIPAddress not specified by the user while using a private subnet and multiple EFAs", true, true, nil),
```
myaser, line 86: @jmdeal and @ellistarn 

I reduced the e2e cases to these 3

website/content/en/preview/concepts/nodeclasses.md
```diff
@@ -859,6 +864,15 @@ spec:
   detailedMonitoring: true
 ```
 
+## spec.associatePublicIPAddress
+
+A boolean field that controls whether instances created by Karpenter for this EC2NodeClass will have an associated public IP address. This overrides the `MapPublicIpOnLaunch` setting applied to the subnet the node is launched in. If this field is not set, the `MapPublicIpOnLaunch` field will be respected.
+
+{{% alert title="Note" color="warning" %}}
+for a NodeClaim requesting `vpc.amazonaws.com/efa` resources, the `associatePublicIPAddress` field is completely ignored
+A public IP address may only be associated with a node at launch if a single network interface is configured.
```
jmdeal, line 873:
```suggestion
If a `NodeClaim` requests `vpc.amazonaws.com/efa` resources, the `associatePublicIPAddress` field is ignored.
A public IP address may only be associated with a node at launch if a single network interface is configured. 
This is inherently incompatible with instances configured for EFA workloads since Karpenter will configure an EFA for each network card on the instance.
```

website/content/en/preview/concepts/nodeclasses.md
jmdeal: I'm good with ignoring this for EFA but I think we should still allow users to set this to false. Some users have reported this as a requirement for security policies and from personal testing I can confirm that you can set the value to either false or nil with multiple network interfaces. Basically would just need to change true to nil and respect the value otherwise.

jmdeal:
Other than this piece of [feedback](https://github.com/aws/karpenter-provider-aws/pull/5437#discussion_r1468098700) this looks good to me! I'm definitely open to discussion on that point but I think it's reasonable to enable users to still disable public IPs in public subnets for EFA instances if they want. This is relevant for instance types with a single network card and thus a single EFA.

myaser:
> Other than this piece of [feedback](https://github.com/aws/karpenter-provider-aws/pull/5437#discussion_r1468098700) this looks good to me! I'm definitely open to discussion on that point but I think it's reasonable to enable users to still disable public IPs in public subnets for EFA instances if they want. This is relevant for instance types with a single network card and thus a single EFA.

thanks, would it be possible to add this piece of requirements iteratively? I do not think we are compromising the UX of developers with this PR, so we can consider this as the next step.

I am just afraid this is taking more and more time and we are in need of this feature, so incremental improvements would help here

jmdeal:
Sure, we can include this in a follow-up PR. I do think it's important but it is a gap in the original EFA implementation. I'll take one more review pass since it's been a couple weeks since I've last looked and then this should be good to go.

jmdeal (commented):
/karpenter snapshot

myaser:
@jmdeal 
what is the status now :) can we merge this?

jmdeal:
Since I made a couple of small updates I can't (or at least shouldn't) approve, I've reached out to other members of the Karpenter team to give it a quick review. It looks good to me, this should be able to go in before the next release.

jonathan-innis (approved):
LGTM 🚀

PR #5445, opened 2024-01-10
ci: Add IRSA Annotation for k8s 1.23 Karpenter Deployment 

engedaam:
Fixes #N/A 

**Description**
-  For k8s 1.23, the E2E test relay on IRSA instead of pod identity. Therefore, the karpenter service account will need the karpenter role name annotated 
 
- Forked account 

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.github/actions/e2e/install-karpenter/action.yaml
+++ b/.github/actions/e2e/install-karpenter/action.yaml
@@ -63,10 +63,12 @@ runs:
         WEBHOOK_ENABLED=true
       fi
 
+      # Remove service account annotation when dropping support for 1.23
       helm upgrade --install karpenter "oci://$ECR_ACCOUNT_ID.dkr.ecr.$ECR_REGION.amazonaws.com/karpenter/snapshot/karpenter" \
         -n kube-system \
         --version "v0-$(git rev-parse HEAD)" \
         --set webhook.enabled=${WEBHOOK_ENABLED} \
+        --set serviceAccount.annotations."eks\.amazonaws\.com/role-arn"="arn:aws:iam::$ACCOUNT_ID:role/karpenter-irsa-$CLUSTER_NAME" \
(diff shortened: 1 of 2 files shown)


.github/workflows/snapshot.yaml
```diff
@@ -8,7 +8,6 @@ jobs:
   release:
     permissions:
       id-token: write # aws-actions/configure-aws-credentials@v4.0.1
-    if: github.repository == 'aws/karpenter-provider-aws'
```
njtran, line 11: What's the reasoning behind removing the repo check? Is it implied that we're in the right repo?

.github/actions/e2e/install-karpenter/action.yaml
```diff
@@ -83,7 +83,10 @@ runs:
         --set "serviceMonitor.endpointConfig.relabelings[2].replacement=$(git describe --abbrev=0 --tags)" \
         --set "serviceMonitor.endpointConfig.relabelings[3].targetLabel=commitsAfterTag" \
         --set "serviceMonitor.endpointConfig.relabelings[3].replacement=\"$(git describe --tags | cut -d '-' -f 2)\"" \
-        --wait
```
jmdeal, line 86: Was this flag meant to be dropped? I'm guessing no since the previous line still has the trailing '\'.

.github/actions/e2e/install-karpenter/action.yaml
```diff
@@ -83,7 +83,10 @@ runs:
...
+
+      if [[ "$K8S_VERSION" == '1.23' ]]; then
```
jonathan-innis, line 87: Comment on why we are adding this annotation, just for context going forward. I don't expect us to have to keep this around for very long since we are about to drop support for 1.23 but would just be good to mark it down for now.

jmdeal (approved):
LGTM 🚀

PR #5453, opened 2024-01-11
Delete Launch templates on deletion of EC2NodeClass

### LINKED ISSUE #995: Launch Template Removal After Uninstalling Karpenter
## Version

**Karpenter**: v0.5.2

**Kubernetes**: v1.21

## Expected Behavior
After uninstalling Karpenter from the cluster or completely deleting the cluster, cleanup process should remove all resources related to Karpenter including Launch Templates those are created for Provisioners.

## Actual Behavior
After removing Karpenter from the cluster there are some orphaned Launch Templates left in the AWS account that everyone should manually perform the cleanup...

- Install Karpenter,
- Add some Provisioner CRDs,
- Deploy some sample applications and let Karpenter provision new nodes using Launch Templates,
- Remove the applications,
- Remove Karpenter provisioner CRDs,
- Uninstall Karpenter,
- Delete the Kubernetes cluster,
- Check EC2 Launch Templates.

jigisha620:
Fixes #995 

**Description**
Added a method to delete the launch templates on deletion of EC2NodeClass. Added tests for the same.

Deployed the changes on my local EKS cluster. Steps followed -
1. Install Karpenter
2. Created Nodepool that references a Nodeclass
3. Deployed sample applications
4. Removed sample applications
5. Removed the Nodeclass object

Performing the nodeclass object deletion will trigger a flow that deletes the launch templates.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/controllers/nodeclass/suite_test.go
+++ b/pkg/controllers/nodeclass/suite_test.go
@@ -833,9 +833,23 @@ var _ = Describe("NodeClassController", func() {
 			ExpectReconcileFailed(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
 			ExpectExists(ctx, env.Client, nodeClass)
 		})
-		It("should succeed to delete the launch template", func() {
+		It("should not delete the launch template not associated with the nodeClass", func() {
 			launchTemplateName := aws.String(fake.LaunchTemplateName())
 			awsEnv.EC2API.LaunchTemplates.Store(launchTemplateName, ec2.LaunchTemplate{LaunchTemplateName: launchTemplateName, LaunchTemplateId: aws.String(fake.LaunchTemplateID()), Tags: []*ec2.Tag{&ec2.Tag{Key: aws.String("karpenter.k8s.aws/cluster"), Value: aws.String("test-cluster")}}})
+			_, ok := awsEnv.EC2API.LaunchTemplates.Load(launchTemplateName)
+			Expect(ok).To(BeTrue())
+			ExpectApplied(ctx, env.Client, nodeClass)
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+
+			Expect(env.Client.Delete(ctx, nodeClass)).To(Succeed())
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+			_, ok = awsEnv.EC2API.LaunchTemplates.Load(launchTemplateName)
+			Expect(ok).To(BeTrue())
+			ExpectNotFound(ctx, env.Client, nodeClass)
+		})
+		It("should succeed to delete the launch template", func() {
+			launchTemplateName := aws.String(fake.LaunchTemplateName())
+			awsEnv.EC2API.LaunchTemplates.Store(launchTemplateName, ec2.LaunchTemplate{LaunchTemplateName: launchTemplateName, LaunchTemplateId: aws.String(fake.LaunchTemplateID()), Tags: []*ec2.Tag{&ec2.Tag{Key: aws.String("karpenter.k8s.aws/cluster"), Value: aws.String("test-cluster")}, {Key: aws.String("karpenter.k8s.aws/ec2nodeclass"), Value: aws.String(nodeClass.Name)}}})
 			_, ok := awsEnv.EC2API.LaunchTemplates.Load(launchTemplateName)
 			Expect(ok).To(BeTrue())
 			ExpectApplied(ctx, env.Client, nodeClass)
--- a/pkg/controllers/nodeclass/suite_test.go
+++ b/pkg/controllers/nodeclass/suite_test.go
@@ -30,7 +30,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/iam"
-
 	corev1beta1 "sigs.k8s.io/karpenter/pkg/apis/v1beta1"
 	"sigs.k8s.io/karpenter/pkg/events"
 	corecontroller "sigs.k8s.io/karpenter/pkg/operator/controller"
@@ -69,7 +68,7 @@ var _ = BeforeSuite(func() {
 	ctx = options.ToContext(ctx, test.Options())
 	awsEnv = test.NewEnvironment(ctx, env)
 
-	nodeClassController = nodeclass.NewNodeClassController(env.Client, events.NewRecorder(&record.FakeRecorder{}), awsEnv.SubnetProvider, awsEnv.SecurityGroupProvider, awsEnv.AMIProvider, awsEnv.InstanceProfileProvider)
+	nodeClassController = nodeclass.NewNodeClassController(env.Client, events.NewRecorder(&record.FakeRecorder{}), awsEnv.SubnetProvider, awsEnv.SecurityGroupProvider, awsEnv.AMIProvider, awsEnv.InstanceProfileProvider, awsEnv.LaunchTemplateProvider)
 })
 
 var _ = AfterSuite(func() {
@@ -820,6 +819,33 @@ var _ = Describe("NodeClassController", func() {
 		var profileName string
 		BeforeEach(func() {
 			profileName = instanceprofile.GetProfileName(ctx, fake.DefaultRegion, nodeClass)
+		})
+		It("should fail to delete the launch template", func() {
+			launchTemplateName := aws.String(fake.LaunchTemplateName())
+			awsEnv.EC2API.LaunchTemplates.Store(launchTemplateName, ec2.LaunchTemplate{LaunchTemplateName: launchTemplateName, LaunchTemplateId: aws.String(fake.LaunchTemplateID()), Tags: []*ec2.Tag{&ec2.Tag{Key: aws.String("karpenter.k8s.aws/cluster"), Value: aws.String("test-cluster")}}})
+			_, ok := awsEnv.EC2API.LaunchTemplates.Load(launchTemplateName)
+			Expect(ok).To(BeTrue())
+			ExpectApplied(ctx, env.Client, nodeClass)
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+
+			Expect(env.Client.Delete(ctx, nodeClass)).To(Succeed())
+			awsEnv.EC2API.NextError.Set(fmt.Errorf("delete Launch Template Error"))
+			ExpectReconcileFailed(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+			ExpectExists(ctx, env.Client, nodeClass)
+		})
+		It("should succeed to delete the launch template", func() {
+			launchTemplateName := aws.String(fake.LaunchTemplateName())
+			awsEnv.EC2API.LaunchTemplates.Store(launchTemplateName, ec2.LaunchTemplate{LaunchTemplateName: launchTemplateName, LaunchTemplateId: aws.String(fake.LaunchTemplateID()), Tags: []*ec2.Tag{&ec2.Tag{Key: aws.String("karpenter.k8s.aws/cluster"), Value: aws.String("test-cluster")}}})
+			_, ok := awsEnv.EC2API.LaunchTemplates.Load(launchTemplateName)
+			Expect(ok).To(BeTrue())
+			ExpectApplied(ctx, env.Client, nodeClass)
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+
+			Expect(env.Client.Delete(ctx, nodeClass)).To(Succeed())
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+			_, ok = awsEnv.EC2API.LaunchTemplates.Load(launchTemplateName)
+			Expect(ok).To(BeFalse())
+			ExpectNotFound(ctx, env.Client, nodeClass)
 		})
 		It("should succeed to delete the instance profile with no NodeClaims", func() {
 			awsEnv.IAMAPI.InstanceProfiles = map[string]*iam.InstanceProfile{
--- a/pkg/controllers/nodeclass/suite_test.go
+++ b/pkg/controllers/nodeclass/suite_test.go
@@ -822,7 +822,7 @@ var _ = Describe("NodeClassController", func() {
 		})
 		It("should fail to delete the launch template", func() {
 			launchTemplateName := aws.String(fake.LaunchTemplateName())
-			awsEnv.EC2API.LaunchTemplates.Store(launchTemplateName, ec2.LaunchTemplate{LaunchTemplateName: launchTemplateName, LaunchTemplateId: aws.String(fake.LaunchTemplateID()), Tags: []*ec2.Tag{&ec2.Tag{Key: aws.String("karpenter.k8s.aws/cluster"), Value: aws.String("test-cluster")}}})
+			awsEnv.EC2API.LaunchTemplates.Store(launchTemplateName, &ec2.LaunchTemplate{LaunchTemplateName: launchTemplateName, LaunchTemplateId: aws.String(fake.LaunchTemplateID()), Tags: []*ec2.Tag{&ec2.Tag{Key: aws.String("karpenter.k8s.aws/cluster"), Value: aws.String("test-cluster")}}})
 			_, ok := awsEnv.EC2API.LaunchTemplates.Load(launchTemplateName)
 			Expect(ok).To(BeTrue())
 			ExpectApplied(ctx, env.Client, nodeClass)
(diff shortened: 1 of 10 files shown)


jmdeal:
NIce work 🎉 A high level thing, since each cluster can have multiple EC2NodeClasses we really only want to delete the LaunchTemplates which are 'owned' by that NodeClass. We just need to keep in mind there's not a 1-1 mapping between launch templates, some EC2NodeClasses could share launch templates and some could have multiple.

jigisha620:
> NIce work 🎉 A high level thing, since each cluster can have multiple EC2NodeClasses we really only want to delete the LaunchTemplates which are 'owned' by that NodeClass. We just need to keep in mind there's not a 1-1 mapping between launch templates, some EC2NodeClasses could share launch templates and some could have multiple.

Thank you. Also, that makes sense. I have updated the LT deletion in such a way that it will check for a tag that corresponds to the nodeclass it is associated with. Also the tag would be added to LT on creation now. The hashed LT name now also contains clusterName and nodeClassName.

pkg/controllers/nodeclass/suite_test.go
```diff
@@ -821,6 +820,47 @@ var _ = Describe("NodeClassController", func() {
 		BeforeEach(func() {
 			profileName = instanceprofile.GetProfileName(ctx, fake.DefaultRegion, nodeClass)
 		})
+		It("should fail to delete the launch template", func() {
```
jmdeal, line 823: An idea for an additional test: ensure multiple identical EC2NodeClasses create unique launch templates with the correct tags. Probably shouldn't be here in the nodeclass functional testing, maybe in the launchtemplate functional testing or in an E2E test.

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -389,3 +390,28 @@ func (p *Provider) getInstanceProfile(nodeClass *v1beta1.EC2NodeClass) (string,
 	}
 	return "", errors.New("neither spec.instanceProfile or spec.role is specified")
 }
+
+func (p *Provider) DeleteLaunchTemplates(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) error {
```
jmdeal, line 394: note: remember to remove debug logging you don't want to keep before merge

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -389,3 +390,28 @@ func (p *Provider) getInstanceProfile(nodeClass *v1beta1.EC2NodeClass) (string,
...
+	clusterName := options.FromContext(ctx).ClusterName
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("tag-key", karpenterManagedTagKey, "tag-value", clusterName))
+	if err := p.ec2api.DescribeLaunchTemplatesPagesWithContext(ctx, &ec2.DescribeLaunchTemplatesInput{
+		Filters: []*ec2.Filter{{Name: aws.String(fmt.Sprintf("tag:%s", karpenterManagedTagKey)), Values: []*string{aws.String(clusterName)}},
+			{Name: aws.String(fmt.Sprintf("tag:%s", v1beta1.LabelNodeClass)), Values: []*string{aws.String(nodeClass.Name)}}},
+	}, func(output *ec2.DescribeLaunchTemplatesOutput, _ bool) bool {
+		for _, lt := range output.LaunchTemplates {
+			if _, err := p.ec2api.DeleteLaunchTemplate(&ec2.DeleteLaunchTemplateInput{LaunchTemplateName: lt.LaunchTemplateName}); err != nil {
```
jonathan-innis, line 402: Is it possible for us to delete multiple launch templates at a time or does EC2 require that we delete these individually

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -389,3 +390,28 @@ func (p *Provider) getInstanceProfile(nodeClass *v1beta1.EC2NodeClass) (string,
...
+				logging.FromContext(ctx).With("launch-template", lt.LaunchTemplateName).Errorf("failed to delete launch template, %v", err)
```
jonathan-innis, line 403: Rather than logging here, we should wrap this using a `multierr` to ensure that we don't return without an error in a case where we fail to delete the launch templates. My understanding from reading this code through right now is that we could succeed to describe the launch templates, but then fail the delete operation on one of them and we would just log an error but still succeed to remove the finalizer.

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -389,3 +390,28 @@ func (p *Provider) getInstanceProfile(nodeClass *v1beta1.EC2NodeClass) (string,
...
+				return false
+			}
+			logging.FromContext(ctx).With(
+				"id", lt.LaunchTemplateName,
+				"name", lt.LaunchTemplateName,
+			).Debugf("deleted launch template")
+		}
+		return true
+	}); err != nil {
+		return fmt.Errorf("unable to fetch the AWS launch templates, %w", err)
```
jonathan-innis, line 413:
```suggestion
		return fmt.Errorf("fetching launch templates, %w", err)
```

nit: Style for error messages in golang is to use the gerund syntax to start. Also, this is the AWS provider, so I think that we can drop any mention of AWS in errors

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -389,3 +390,28 @@ func (p *Provider) getInstanceProfile(nodeClass *v1beta1.EC2NodeClass) (string,
...
+	} else {
+		return nil
```
jonathan-innis, line 415: nit: No need for an else statement if the previous statement returns an argument as well

pkg/fake/ec2api.go
```diff
@@ -382,6 +382,42 @@ func (e *EC2API) DescribeLaunchTemplatesWithContext(_ context.Context, input *ec
 	return output, nil
 }
 
+func (e *EC2API) DescribeLaunchTemplatesPagesWithContext(_ context.Context, input *ec2.DescribeLaunchTemplatesInput, callbck func(*ec2.DescribeLaunchTemplatesOutput, bool) bool, _ ...request.Option) error {
```
jonathan-innis, line 385: If you look at the way that other calls are defined in the fake api, you can see that we have defined the standard call, in this case we have defined a definition for `DescribeLaunchTemplatesWithContext` elsewhere in the file, and then you can create a `DescribeLaunchTemplatesPagesWithContext` function that calls out to the `DescribeLaunchTemplatesWithContext` directly so you don't have to rewrite the same logic for essentially the same function twice.

Take a look at `DescribeImagesPagesWithContext` for more context

pkg/providers/launchtemplate/launchtemplate.go
jonathan-innis, line 402: We should use `DeleteLaunchTemplateWithContext` so that we make sure that we cancel the function call if the context ever gets cancelled.

jonathan-innis (commented):
Nice work 🎉

pkg/providers/launchtemplate/launchtemplate.go
jigisha620: I tried looking for that option but seems like EC2 wants us to delete LTs individually.

pkg/controllers/nodeclass/suite_test.go
jigisha620: Thanks for the idea. I have added a test to the launchtemplate functional testing.

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -389,3 +392,30 @@ func (p *Provider) getInstanceProfile(nodeClass *v1beta1.EC2NodeClass) (string,
...
 }
+
+func (p *Provider) DeleteLaunchTemplates(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) error {
+	clusterName := options.FromContext(ctx).ClusterName
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("tag-key", karpenterManagedTagKey, "tag-value", clusterName))
+	var deleteLTerr error
+	describeErr := p.ec2api.DescribeLaunchTemplatesPagesWithContext(ctx, &ec2.DescribeLaunchTemplatesInput{
+		Filters: []*ec2.Filter{{Name: aws.String(fmt.Sprintf("tag:%s", karpenterManagedTagKey)), Values: []*string{aws.String(clusterName)}},
+			{Name: aws.String(fmt.Sprintf("tag:%s", v1beta1.LabelNodeClass)), Values: []*string{aws.String(nodeClass.Name)}}},
+	}, func(output *ec2.DescribeLaunchTemplatesOutput, _ bool) bool {
```
jmdeal, line 403: Super optional nit, but this is pretty dense and a little hard for me to read so a small formatting suggestion:
```suggestion
	describeErr := p.ec2api.DescribeLaunchTemplatesPagesWithContext(ctx, &ec2.DescribeLaunchTemplatesInput{
		Filters: []*ec2.Filter{
			{Name: aws.String(fmt.Sprintf("tag:%s", karpenterManagedTagKey)), Values: []*string{aws.String(clusterName)}},
			{Name: aws.String(fmt.Sprintf("tag:%s", v1beta1.LabelNodeClass)), Values: []*string{aws.String(nodeClass.Name)}}
		},
	}, func(output *ec2.DescribeLaunchTemplatesOutput, _ bool) bool {
```

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -389,3 +392,30 @@ func (p *Provider) getInstanceProfile(nodeClass *v1beta1.EC2NodeClass) (string,
 	}
 	return "", errors.New("neither spec.instanceProfile or spec.role is specified")
 }
+
+func (p *Provider) DeleteLaunchTemplates(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) error {
+	clusterName := options.FromContext(ctx).ClusterName
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("tag-key", karpenterManagedTagKey, "tag-value", clusterName))
```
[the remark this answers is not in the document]
jonathan-innis: +1

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -389,3 +392,30 @@ func (p *Provider) getInstanceProfile(nodeClass *v1beta1.EC2NodeClass) (string,
...
+	var deleteLTerr error
+	describeErr := p.ec2api.DescribeLaunchTemplatesPagesWithContext(ctx, &ec2.DescribeLaunchTemplatesInput{
+		Filters: []*ec2.Filter{{Name: aws.String(fmt.Sprintf("tag:%s", karpenterManagedTagKey)), Values: []*string{aws.String(clusterName)}},
+			{Name: aws.String(fmt.Sprintf("tag:%s", v1beta1.LabelNodeClass)), Values: []*string{aws.String(nodeClass.Name)}}},
+	}, func(output *ec2.DescribeLaunchTemplatesOutput, _ bool) bool {
+		for _, lt := range output.LaunchTemplates {
+			if _, err := p.ec2api.DeleteLaunchTemplateWithContext(ctx, &ec2.DeleteLaunchTemplateInput{LaunchTemplateName: lt.LaunchTemplateName}); err != nil {
+				logging.FromContext(ctx).With("launch-template", lt.LaunchTemplateName).Errorf("failed to delete launch template, %v", err)
```
jonathan-innis, line 406: Don't log if you are going to return the error. controller-runtime will take care of printing the error for us if we return the error back up the stack.

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -389,3 +392,30 @@ func (p *Provider) getInstanceProfile(nodeClass *v1beta1.EC2NodeClass) (string,
 	}
 	return "", errors.New("neither spec.instanceProfile or spec.role is specified")
 }
+
+func (p *Provider) DeleteLaunchTemplates(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) error {
+	clusterName := options.FromContext(ctx).ClusterName
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("tag-key", karpenterManagedTagKey, "tag-value", clusterName))
+	var deleteLTerr error
+	describeErr := p.ec2api.DescribeLaunchTemplatesPagesWithContext(ctx, &ec2.DescribeLaunchTemplatesInput{
```
jonathan-innis, line 400: QQ: Would it be better to just describe all of the launch templates and store the names in one action and then delete all of the names that we have stored in a separate action? It might improve the way you do error handling and help out with readability

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -389,3 +392,30 @@ func (p *Provider) getInstanceProfile(nodeClass *v1beta1.EC2NodeClass) (string,
...
+			{Name: aws.String(fmt.Sprintf("tag:%s", v1beta1.LabelNodeClass)), Values: []*string{aws.String(nodeClass.Name)}}},
+	}, func(output *ec2.DescribeLaunchTemplatesOutput, _ bool) bool {
+		for _, lt := range output.LaunchTemplates {
+			if _, err := p.ec2api.DeleteLaunchTemplateWithContext(ctx, &ec2.DeleteLaunchTemplateInput{LaunchTemplateName: lt.LaunchTemplateName}); err != nil {
+				logging.FromContext(ctx).With("launch-template", lt.LaunchTemplateName).Errorf("failed to delete launch template, %v", err)
+				deleteLTerr = multierr.Append(deleteLTerr, err)
+				return false
+			}
+		}
+		logging.FromContext(ctx).Debugf("deleted %v launch templates", len(output.LaunchTemplates))
```
jonathan-innis, line 411: Consider using `util.PrettySlice()` here to print out some of the names for you

pkg/providers/launchtemplate/launchtemplate.go
jonathan-innis: Sounds good. I think I remember this as well. It's annoying we use up a bunch of TPS to do this, but if there's no other way around it, that's what we have to do

jonathan-innis:
Do we have an E2E test to validate all of the launch templates are gone when we delete the EC2NodeClass? Take a look at https://github.com/aws/karpenter-provider-aws/blob/main/test/suites/integration/instance_profile_test.go#L47 for a similar style of test

pkg/providers/launchtemplate/launchtemplate.go
jigisha620: Sure, I will remove this.

pkg/providers/launchtemplate/launchtemplate.go
jigisha620: Yes, can be done that way instead to improve readability.

jigisha620:
> Do we have an E2E test to validate all of the launch templates are gone when we delete the EC2NodeClass? Take a look at https://github.com/aws/karpenter-provider-aws/blob/main/test/suites/integration/instance_profile_test.go#L47 for a similar style of test

No, there's no E2E test to validate that. I will add it.

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

pkg/controllers/nodeclass/suite_test.go
jonathan-innis, line 823:
```suggestion
		It("should not delete the NodeClass if launch template deletion fails", func() {
```

This seems like it's more ensuring that we don't delete the NodeClass as long as the launch template deletion fails

pkg/controllers/nodeclass/suite_test.go
```diff
@@ -821,6 +820,47 @@ var _ = Describe("NodeClassController", func() {
...
+			ExpectApplied(ctx, env.Client, nodeClass)
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+
+			Expect(env.Client.Delete(ctx, nodeClass)).To(Succeed())
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+			_, ok = awsEnv.EC2API.LaunchTemplates.Load(launchTemplateName)
+			Expect(ok).To(BeTrue())
+			ExpectNotFound(ctx, env.Client, nodeClass)
+		})
+		It("should succeed to delete the launch template", func() {
```
jonathan-innis, line 850: When we're deleting here, should we just validate that this works with multiple launch templates here with the same tag key associated with the NodeClass for better coverage?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -389,3 +390,29 @@ func (p *Provider) getInstanceProfile(nodeClass *v1beta1.EC2NodeClass) (string,
...
+	}, func(output *ec2.DescribeLaunchTemplatesOutput, _ bool) bool {
+		for _, lt := range output.LaunchTemplates {
+			ltNames = append(ltNames, lt.LaunchTemplateName)
+		}
+		return true
+	}); err != nil {
+		return fmt.Errorf("fetching launch templates, %w", err)
+	}
+
+	for _, name := range ltNames {
```
jonathan-innis, line 411: I wonder if we should best-effort try to delete the launch templates and then return a multi-err here that way we delete as many launch templates as possible here.

test/suites/integration/launch_template_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+var _ = Describe("Launch Template Deletion", func() {
+	It("should remove the generated Launch Templates when deleting the NodeClass", func() {
+		pod := coretest.Pod()
+		env.ExpectCreated(nodePool, nodeClass, pod)
+		env.EventuallyExpectHealthy(pod)
+		env.ExpectCreatedNodeCount("==", 1)
+
+		env.ExpectDeleted(nodePool, nodeClass)
+		Eventually(func(g Gomega) {
```
jmdeal, line 38: `Eventually` has a default polling period of 10ms, is this going to be an issue with rate limiting? Probably a good idea to set it higher, just to be safe.

jigisha620 (commented):
/karpenter snapshot

jonathan-innis (dismissed):
LGTM 🚀

jigisha620 (commented):
/karpenter snapshot

jmdeal (approved):
LGTM 🚀

PR #3330, opened 2023-01-31
fix: Round up when launching a node with block device mappings

jonathan-innis:
Fixes #3302

**Description**

- Round up when launching a node with block device mappings
- Include the blockDeviceMapping volumeSize requests in the hashing to notice a change in the volume size

**How was this change tested?**

* `make presubmit`

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
None
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5454, opened 2024-01-11
fix: Hash volume size quantity value as a string

### LINKED ISSUE #5447: A blockDeviceMapping change in EC2NodeClass does not trigger drift replace
### Description

**Observed Behavior**:

When changing `spec.blockDeviceMapping[0].ebs.volumeSize` no drift replace is triggered. Also, the `karpenter.k8s.aws/ec2nodeclass-hash` annotation remains unchanged.

However, when I flip `spec.detailedMonitoring` a drift replace is triggered. And the `karpenter.k8s.aws/ec2nodeclass-hash` annotation gets updated.

Since the hash value does not get updated for block device mapping changes , I assume the bug is inside the hashing code.

**Expected Behavior**:

Changing `spec.blockDeviceMapping[0].ebs.volumeSize` triggers a drift replace.

- Create a EC2NodeClass resource and merge in this spec:

```
spec:
  blockDeviceMappings:
  - deviceName: /dev/xvda
    ebs:
      encrypted: true
      volumeSize: 50Gi
      volumeType: gp3
  detailedMonitoring: true
```

- Make note of the `karpenter.k8s.aws/ec2nodeclass-hash` annotation on the EC2NodeClass resource
- Now edit the EC2NodeClass resource and change the volumeSize.
- Check the annotation again. For me the value is unchanged (which is unexpected).

To ensure Karpenter drift does work:

- Edit the EC2NodeClass resource again and flip spec.detailedMonitoring
- Check the annotation. You should see it changed.

**Versions**:
- Chart Version: v0.33.0
- Kubernetes Version (`kubectl version`): v1.27.7

jonathan-innis:
Fixes #5447

**Description**

This change applies the change that was added in #3330 and ensures that we won't regress on this issue again by adding the relevant testing into functional and E2E testing. Hashing the resource quantity value as a string was missed with the changes were ported over and duplicated from the AWS provider struct in v1alpha1.

`make presubmit`
`/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/apis/v1beta1/ec2nodeclass_hash_test.go
+++ b/pkg/apis/v1beta1/ec2nodeclass_hash_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/imdario/mergo"
 	"github.com/samber/lo"
+	"sigs.k8s.io/karpenter/pkg/utils/resources"
 
 	"github.com/aws/karpenter-provider-aws/pkg/apis/v1beta1"
 	"github.com/aws/karpenter-provider-aws/pkg/test"
@@ -66,7 +67,21 @@ var _ = Describe("Hash", func() {
 		Entry("UserData Drift", "588756456110800812", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{UserData: aws.String("userdata-test-2")}}),
 		Entry("Tags Drift", "2471764681523766508", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{Tags: map[string]string{"keyTag-test-3": "valueTag-test-3"}}}),
 		Entry("MetadataOptions Drift", "11030161632375731908", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{MetadataOptions: &v1beta1.MetadataOptions{HTTPEndpoint: aws.String("test-metadata-2")}}}),
-		Entry("BlockDeviceMappings Drift", "436753305915039702", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{BlockDeviceMappings: []*v1beta1.BlockDeviceMapping{{DeviceName: aws.String("map-device-test-3")}}}}),
+		Entry("BlockDeviceMappings Drift", "2884646775270747279", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{BlockDeviceMappings: []*v1beta1.BlockDeviceMapping{
+			{
+				DeviceName: aws.String("map-device-test-3"),
+				EBS: &v1beta1.BlockDevice{
+					DeleteOnTermination: aws.Bool(true),
+					Encrypted:           aws.Bool(true),
+					IOPS:                aws.Int64(1000),
+					KMSKeyID:            aws.String("kmskeyid"),
+					SnapshotID:          aws.String("snapshotid"),
+					Throughput:          aws.Int64(1000),
+					VolumeSize:          resources.Quantity("100Gi"),
+					VolumeType:          aws.String("volumetype"),
+				},
+			},
+		}}}),
 		Entry("Context Drift", "3729470655588343019", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{Context: aws.String("context-2")}}),
 		Entry("DetailedMonitoring Drift", "17892305444040067573", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{DetailedMonitoring: aws.Bool(true)}}),
 		Entry("AMIFamily Drift", "9493798894326942407", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{AMIFamily: aws.String(v1beta1.AMIFamilyBottlerocket)}}),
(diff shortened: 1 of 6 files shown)


jonathan-innis:
This change is blocked until we have the ability to version the hashes that we are using to evaluate drift since this would be a breaking change to the existing hash versioning mechanism since a previously unconsidered field would now get hashed with a value.

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

pkg/apis/v1beta1/ec2nodeclass_hash_test.go
```diff
@@ -66,7 +67,21 @@ var _ = Describe("Hash", func() {
 		Entry("UserData Drift", "588756456110800812", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{UserData: aws.String("userdata-test-2")}}),
 		Entry("Tags Drift", "2471764681523766508", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{Tags: map[string]string{"keyTag-test-3": "valueTag-test-3"}}}),
 		Entry("MetadataOptions Drift", "11030161632375731908", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{MetadataOptions: &v1beta1.MetadataOptions{HTTPEndpoint: aws.String("test-metadata-2")}}}),
-		Entry("BlockDeviceMappings Drift", "436753305915039702", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{BlockDeviceMappings: []*v1beta1.BlockDeviceMapping{{DeviceName: aws.String("map-device-test-3")}}}}),
+		Entry("BlockDeviceMappings Drift", "2884646775270747279", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{BlockDeviceMappings: []*v1beta1.BlockDeviceMapping{
```
engedaam, line 70: I think we should expanded this out make sure that making any changes to every field will cause the cache to change. I don't think this test would have helped us catch the issue we are trying to fix here.

test/suites/drift/suite_test.go
```diff
@@ -680,7 +681,7 @@ var _ = Describe("Drift", func() {
 			{
 				DeviceName: awssdk.String("/dev/xvda"),
 				EBS: &v1beta1.BlockDevice{
-					VolumeSize: resource.NewScaledQuantity(20, resource.Giga),
+					VolumeSize: resources.Quantity("20Gi"),
```
engedaam, line 684: I think we should just test each of these fields separately and this should also eliminate the need to have a separate test just for `volumeSize`

jonathan-innis: Going to leave this the way it is since it catches a regression specifically on this volumeSize issue which cropped up a couple times only due to the "stringification" piece here. I agree that we should test each field, but I think that we can do so exclusively through the functional testing to avoid extending the time of this E2E too much

pkg/apis/v1beta1/ec2nodeclass_hash_test.go
jonathan-innis: > I don't think this test would have helped us catch the issue we are trying to fix here

Test would have helped catch this specific issue, but I agree that we can expand this out to add additional testing at the field-level.

jonathan-innis: Updated with the new tests.

pkg/apis/v1beta1/ec2nodeclass_hash_test.go
```diff
@@ -27,76 +28,137 @@ import (
...
+		Entry("MetadataOptions HTTPPutResponseHopLimit", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{MetadataOptions: &v1beta1.MetadataOptions{HTTPPutResponseHopLimit: lo.ToPtr(int64(10))}}}),
+		Entry("MetadataOptions HTTPTokens", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{MetadataOptions: &v1beta1.MetadataOptions{HTTPTokens: lo.ToPtr("required")}}}),
+		Entry("BlockDeviceMapping DeviceName", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{BlockDeviceMappings: []*v1beta1.BlockDeviceMapping{{DeviceName: lo.ToPtr("map-device-test-3")}}}}),
+		Entry("BlockDeviceMapping RootVolume Drift", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{BlockDeviceMappings: []*v1beta1.BlockDeviceMapping{{RootVolume: true}}}}),
+		Entry("BlockDeviceMapping DeleteOnTermination Drift", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{BlockDeviceMappings: []*v1beta1.BlockDeviceMapping{{EBS: &v1beta1.BlockDevice{DeleteOnTermination: lo.ToPtr(true)}}}}}),
+		Entry("BlockDeviceMapping Encrypted Drift", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{BlockDeviceMappings: []*v1beta1.BlockDeviceMapping{{EBS: &v1beta1.BlockDevice{Encrypted: lo.ToPtr(true)}}}}}),
```
engedaam, line 147:
```suggestion
		Entry("BlockDeviceMapping Encrypted", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{BlockDeviceMappings: []*v1beta1.BlockDeviceMapping{{EBS: &v1beta1.BlockDevice{Encrypted: lo.ToPtr(true)}}}}}),
```

PR #5480, opened 2024-01-16
test: Add testing for PodDisrputionBudgets in E2E testing

jonathan-innis:
Fixes #N/A 

**Description**

Add testing for ensuring PodDisruptionBudgets block drift and expiration in E2E testing

`/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No


jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

test/pkg/environment/common/expectations.go
```diff
@@ -267,22 +267,33 @@ func (env *Environment) ExpectExists(obj client.Object) {
 	}).WithTimeout(time.Second * 5).Should(Succeed())
 }
 
+func (env *Environment) EventuallyExpectBound(pods ...*v1.Pod) {
+	GinkgoHelper()
+	Eventually(func(g Gomega) {
+		for _, pod := range pods {
+			g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(pod), pod)).To(Succeed())
+			g.Expect(pod.Spec.NodeName).ToNot(BeEmpty())
```
njtran, line 275: I'm assuming this is the same as == ""?

test/pkg/environment/common/expectations.go
```diff
@@ -373,6 +384,23 @@ func (env *Environment) EventuallyExpectPendingPodCount(selector labels.Selector
 	}).Should(Succeed())
 }
 
+func (env *Environment) EventuallyExpectBoundPodCount(selector labels.Selector, numPods int) []*v1.Pod {
+	GinkgoHelper()
+	var res []*v1.Pod
+	Eventually(func(g Gomega) {
+		res = []*v1.Pod{}
+		podList := &v1.PodList{}
+		g.Expect(env.Client.List(env.Context, podList)).To(Succeed())
```
njtran, line 393: Can we input the selector into the list call so that we're not doing a cluster wide list?

test/pkg/environment/common/expectations.go
```diff
@@ -458,6 +486,26 @@ func (env *Environment) ConsistentlyExpectNodeCount(comparator string, count int
 	return lo.ToSlicePtr(nodeList.Items)
 }
 
+func (env *Environment) ConsistentlyExpectNoDisruptions(nodeCount int, duration string) {
```
njtran, line 489: Thoughts on hardcoded duration?

test/pkg/environment/common/expectations.go
jonathan-innis: Well, it's the same as `!= ""` but yeah that's correct.

test/pkg/environment/common/expectations.go
jonathan-innis: I could go either way. Maybe overfactoring too early, but I also think that you naturally ask the "how long" question when it comes to `Consistently` blocks. So, I could realistically see making it a standard that when you add a `Consistently` block, you pass in a duration so that the caller is clear on how long that you are checking for at the maximum

test/pkg/environment/common/expectations.go
jonathan-innis: Good call. Don't know why I didn't do that from the start. That's a lot better here.

jonathan-innis (commented):
/karpenter snapshot

test/pkg/environment/common/expectations.go
njtran: I personally think it's best to do it the two function route. Most calls just use a default, but you can use a configured function if you like. Up to you if you think that's overfactoring.

njtran (dismissed):

test/pkg/environment/common/expectations.go
jonathan-innis: I think I'm happy to leave it the way that it is right now.

njtran (approved):

PR #5525, opened 2024-01-25
Add testing for consolidation budgets e2e

jonathan-innis:
Fixes #N/A 

**Description**

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/pkg/environment/common/expectations.go
+++ b/test/pkg/environment/common/expectations.go
@@ -479,7 +479,7 @@ func (env *Environment) ExpectNodeCount(comparator string, count int) {
 
 func (env *Environment) ExpectNodeClaimCount(comparator string, count int) {
 	GinkgoHelper()
-	
+
 	nodeClaimList := &corev1beta1.NodeClaimList{}
 	Expect(env.Client.List(env, nodeClaimList, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
 	Expect(len(nodeClaimList.Items)).To(BeNumerically(comparator, count))
@@ -525,6 +525,19 @@ func (env *Environment) ConsistentlyExpectNoDisruptions(nodeCount int, duration 
 			g.Expect(ok).To(BeFalse())
 		}
 	}, duration).Should(Succeed())
+}
+
+func (env *Environment) ConsistentlyExpectTaintedNodeCount(comparator string, count int, duration string) []*v1.Node {
+	GinkgoHelper()
+
+	By(fmt.Sprintf("checking for tainted nodes to be %s to %d for %s", comparator, count, duration))
+	nodeList := &v1.NodeList{}
+	Consistently(func(g Gomega) {
+		g.Expect(env.Client.List(env, nodeList, client.MatchingFields{"spec.taints[*].karpenter.sh/disruption": "disrupting"})).To(Succeed())
+		g.Expect(len(nodeList.Items)).To(BeNumerically(comparator, count),
+			fmt.Sprintf("expected %d tainted nodes, had %d (%v)", count, len(nodeList.Items), NodeNames(lo.ToSlicePtr(nodeList.Items))))
+	}, duration).Should(Succeed())
+	return lo.ToSlicePtr(nodeList.Items)
 }
 
 func (env *Environment) EventuallyExpectTaintedNodeCount(comparator string, count int) []*v1.Node {
--- a/test/pkg/environment/common/expectations.go
+++ b/test/pkg/environment/common/expectations.go
@@ -19,6 +19,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io"
+	"math"
 	"strings"
 	"time"
 
@@ -468,6 +469,22 @@ func (env *Environment) ExpectCreatedNodeCount(comparator string, count int) []*
 	return createdNodes
 }
 
+func (env *Environment) ExpectNodeCount(comparator string, count int) {
+	GinkgoHelper()
+
+	nodeList := &v1.NodeList{}
+	Expect(env.Client.List(env, nodeList, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
+	Expect(len(nodeList.Items)).To(BeNumerically(comparator, count))
+}
+
+func (env *Environment) ExpectNodeClaimCount(comparator string, count int) {
+	GinkgoHelper()
+	
+	nodeClaimList := &corev1beta1.NodeClaimList{}
+	Expect(env.Client.List(env, nodeClaimList, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
+	Expect(len(nodeClaimList.Items)).To(BeNumerically(comparator, count))
+}
+
 func NodeNames(nodes []*v1.Node) []string {
 	return lo.Map(nodes, func(n *v1.Node, index int) string {
 		return n.Name
@@ -486,6 +503,9 @@ func (env *Environment) ConsistentlyExpectNodeCount(comparator string, count int
 	return lo.ToSlicePtr(nodeList.Items)
 }
 
+// ConsistentlyExpectNoDisruptions ensures that the state of the cluster is not changed within a passed duration
+// Specifically, we check if the cluster size in terms of nodes is the same as the passed-in size and we validate
+// that no disrupting taints are added throughout the window
 func (env *Environment) ConsistentlyExpectNoDisruptions(nodeCount int, duration string) {
 	GinkgoHelper()
 	Consistently(func(g Gomega) {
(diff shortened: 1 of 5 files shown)


jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

test/pkg/environment/common/expectations.go
```diff
@@ -751,17 +784,56 @@ func (env *Environment) ExpectDaemonSetEnvironmentVariableUpdated(obj client.Obj
 	Expect(env.Client.Patch(env.Context, ds, patch)).To(Succeed())
 }
 
-func (env *Environment) ExpectHealthyPodsForNode(nodeName string) []*v1.Pod {
+// ForcePodsToSpread ensures that currently scheduled pods get spread evenly across all passed nodes by deleting pods off of existing
+// nodes and waiting them to reschedule. This is useful for scenarios where you want to force the nodes be underutilized
+// but you want to keep a consistent count of nodes rather than leaving around empty ones.
+func (env *Environment) ForcePodsToSpread(nodes ...*v1.Node) {
```
njtran, line 790: Really like this function!!

test/pkg/environment/common/expectations.go
```diff
@@ -751,17 +784,56 @@ func (env *Environment) ExpectDaemonSetEnvironmentVariableUpdated(obj client.Obj
...
+		if !found {
+			break
+		}
+		// Set the nodes to unschedulable so that the pods won't reschedule.
+		Expect(env.Client.Get(env.Context, client.ObjectKeyFromObject(node), node)).To(Succeed())
+		stored := node.DeepCopy()
+		node.Spec.Unschedulable = true
+		Expect(env.Client.Patch(env.Context, node, client.MergeFrom(stored))).To(Succeed())
+		for _, pod := range nodePods[maxPodsPerNode:] {
+			env.ExpectDeleted(pod)
```
njtran, line 816: Since this is e2es, can we set the terminationGracePeriodSeconds to 0 for these pod deletes? I know for some of the tests with having a preStop that I'm adding, we'll need to set this to 0 so that we can enforce it through here.

test/pkg/environment/common/expectations.go
```diff
@@ -751,17 +784,56 @@ func (env *Environment) ExpectDaemonSetEnvironmentVariableUpdated(obj client.Obj
...
+
+	// Get the total count of pods across
+	podCount := 0
+	for _, n := range nodes {
+		podCount += len(env.ExpectActivePodsForNode(n.Name))
+	}
+	maxPodsPerNode := int(math.Ceil(float64(podCount) / float64(len(nodes))))
+
+	By(fmt.Sprintf("forcing %d pods to spread across %d nodes", podCount, len(nodes)))
+	for {
```
njtran, line 801: Should we add a timeout here? We're probably the only ones using this function, but if someone does this at large scale, it may take a really long time.

test/pkg/environment/common/expectations.go
```diff
@@ -751,17 +784,56 @@ func (env *Environment) ExpectDaemonSetEnvironmentVariableUpdated(obj client.Obj
...
+		Expect(env.Client.Patch(env.Context, node, client.MergeFrom(stored))).To(Succeed())
+		for _, pod := range nodePods[maxPodsPerNode:] {
+			env.ExpectDeleted(pod)
+		}
+		Eventually(func(g Gomega) {
+			g.Expect(len(env.ExpectActivePodsForNode(node.Name))).To(Or(Equal(maxPodsPerNode), Equal(maxPodsPerNode-1)))
+		}).WithTimeout(5 * time.Second).Should(Succeed())
+	}
+	for _, n := range nodes {
+		stored := n.DeepCopy()
```
njtran, line 823: Can we re-get the nodes?

test/pkg/environment/common/expectations.go
```diff
@@ -751,17 +784,56 @@ func (env *Environment) ExpectDaemonSetEnvironmentVariableUpdated(obj client.Obj
...
 	GinkgoHelper()
 	podList := &v1.PodList{}
 	Expect(env.Client.List(env, podList, client.MatchingFields{"spec.nodeName": nodeName}, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
 
 	// Return the healthy pods
 	return lo.Filter(lo.ToSlicePtr(podList.Items), func(p *v1.Pod, _ int) bool {
-		_, found := lo.Find(p.Status.Conditions, func(cond v1.PodCondition) bool {
-			return cond.Type == v1.PodReady && cond.Status == v1.ConditionTrue
-		})
-		return found
```
njtran, line 764: Why are we removing the check on Ready Pods? Shouldn't we only model pods that are healthy and not deleting?

test/pkg/environment/common/expectations.go
```diff
@@ -507,6 +527,19 @@ func (env *Environment) ConsistentlyExpectNoDisruptions(nodeCount int, duration
 	}, duration).Should(Succeed())
 }
 
+func (env *Environment) ConsistentlyExpectTaintedNodeCount(comparator string, count int, duration string) []*v1.Node {
```
njtran, line 530: One thing I was curious about when I was making my changes to this function is why is this duration a string and not a time.Duration?

test/suites/consolidation/suite_test.go
```diff
@@ -63,6 +65,222 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+			}
+
+			dep.Spec.Replicas = lo.ToPtr[int32](1)
+			By("making the nodes empty")
+			// Update the deployment to only contain 1 replica.
+			env.ExpectUpdated(dep)
+
+			// Ensure that we get two nodes tainted, and they have overlap during the drift
+			nodes = env.EventuallyExpectTaintedNodeCount("==", 2)
+			env.ConsistentlyExpectTaintedNodeCount("==", 2, "5s")
```
njtran, line 129: nit: you should set nodes = second one since it's technically going to be more updated, and also already returns the list of nodes.

test/suites/consolidation/suite_test.go
```diff
@@ -63,6 +65,222 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+			for _, node := range nodes {
+				Expect(env.ExpectTestingFinalizerRemoved(node)).To(Succeed())
+			}
+
+			// After the deletion timestamp is set and all pods are drained
+			// the node should be gone
+			env.EventuallyExpectNotFound(nodes[0], nodes[1])
+
+			// This check ensures that we are consolidating nodes at the same time
+			nodes = env.EventuallyExpectTaintedNodeCount("==", 2)
```
njtran, line 141: This is for the other 2 of the 4 nodes that are empty, right? Can you put that in a comment here? Can we do the same consistently check here too before continuing?

test/suites/consolidation/suite_test.go
```diff
@@ -63,6 +65,222 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+			env.EventuallyExpectHealthyPodCount(selector, int(numPods))
+
+			dep.Spec.Replicas = lo.ToPtr[int32](1)
+			By("making the nodes empty")
+			// Update the deployment to only contain 1 replica.
+			env.ExpectUpdated(dep)
+
+			env.ConsistentlyExpectNoDisruptions(5, "1m")
+		})
+		It("should not allow consolidation if the budget is fully blocking during a scheduled time", func() {
```
njtran, line 247: Thoughts on making this test something like a schedule that is previously blocking that no longer becomes blocking? For instance, you could do 
- `scheduleStart = time.Now() - 2 minutes`
- `scheduleDuration = 4 minutes`
- expect no disruptions for first 2 minutes
- expect disruptions for the 2 minutes after

test/suites/drift/suite_test.go
```diff
@@ -209,36 +210,13 @@ var _ = Describe("Drift", func() {
 			nodeClaims := env.EventuallyExpectCreatedNodeClaimCount("==", 3)
 			nodes := env.EventuallyExpectCreatedNodeCount("==", 3)
 			env.EventuallyExpectHealthyPodCount(selector, int(numPods))
-			env.Monitor.Reset() // Reset the monitor so that we can expect a single node to be spun up after drift
```
njtran, line 212: Was this monitor reset wrong?

njtran (commented):
Nice work! Thanks for doing this!

test/pkg/environment/common/expectations.go
jonathan-innis: > we'll need to set this to 0 so that we can enforce it through here

Not sure that I follow this one. If you want to add a preStop, you're also going to have to extend the terminationGracePeriod seconds. terminationGracePeriod is a hard deadline. preStop can't stop the pod from being terminated beyond it.

test/pkg/environment/common/expectations.go
jonathan-innis: Right, it's a check on active pods. That strikes me as the right behavior over checking if they are healthy, since pending pods that are about to go ready on the node should also go toward the count

test/pkg/environment/common/expectations.go
jonathan-innis: It's a good call. I'm fine going to `time.Duration` if you want

test/suites/consolidation/suite_test.go
jonathan-innis: > you should set nodes = second one since it's technically going to be more updated, and also already returns the list of nodes

Sorry, not getting what this comment means here.

test/suites/drift/suite_test.go
jonathan-innis: You're not doing any checking on the monitor anywhere after this so there's no reason to call the reset from what I could tell.

test/pkg/environment/common/expectations.go
jonathan-innis: Shouldn't need to re-get the nodes if you are doing a patch operation here.

test/pkg/environment/common/expectations.go
jonathan-innis: Discussed offline: `gracePeriod: 0` already exists within the `ExpectDeleted` call

test/pkg/environment/common/expectations.go
jonathan-innis: Discussed offline: This is the behavior that we want so we correctly consider Creating pods.

test/suites/drift/suite_test.go
jonathan-innis: Discussed offline: Confirmed that this monitor reset isn't needed in any of these tests

jonathan-innis (commented):
/karpenter snapshot

njtran (commented):
LGTM

njtran (approved):

PR #5551, opened 2024-01-29
[DRAFT] Ami minimum age

### LINKED ISSUE #5382: Selector Terms for AMI Minimum Age
### Description

** READ BEFORE CONTINUING: If your issue is not specific to AWS, please cut a ticket in [kubernetes-sigs/karpenter](https://github.com/kubernetes-sigs/karpenter/issues/new/choose).

It would be great if I could specify the minimum age of an AMI, so that an AMI can be run in staging for a while before it goes to production.

We just had a production outage due to a [bad AMI](https://github.com/awslabs/amazon-eks-ami/issues/1551).
While we can build a delay mechanism on top using automation/pipeline/etc, a "minimumAge" seems simpler to manage operationally. 

cebernardi:
Fixes #5382 

**Description**
This is a draft PR that proposed a possible implementation for the introduction of a "ami minimum age" parameter.
I opened it as a draft to get early feedback. 

Not sure if this feature is already worked on by the maintainers or if there are other plans about it.

It's not exactly tested (a part for a very small subset of use cases, locally), all the unit tests are missing.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/designs/ami-minimum-age.md
+++ b/designs/ami-minimum-age.md
@@ -8,7 +8,8 @@ This document describes a design for a new feature to allow users to specify a m
 
 Karpenter allows users to specify TTL for nodes, which configures the maximum age of nodes in the cluster. 
 
-This is useful to ensure that nodes are replaced periodically, which helps to ensure that the cluster is running on the latest AMIs available, which simplifies and automates the security patching process.
+This is useful to ensure that nodes are replaced periodically, and helps to ensure that the cluster is running on the latest AMIs available, 
+simplifying and automating the security patching process.
 
 However, a recent issue with a released AMI (see [this issue](https://github.com/awslabs/amazon-eks-ami/issues/1551)) has highlighted that it would be useful to be able to specify a minimum age for AMIs, to ensure that nodes are not replaced with an AMI that is too new.
 
@@ -36,7 +37,7 @@ The parameter applies to AMIs resolved via:
 
 If `amiSelectorTerms` are specified, the AMIs candidates are currently resolved querying the AWS EC2 API for AMIs, applying a filter to match the selector terms.
 
-Once the images are retrieved from the API, the images younger than `amiMinimumAge` are filtered out.
+To implement this feature, once the images are retrieved from the API, the ones that are younger than `amiMinimumAge` are filtered out.
 
 AMIs could be partially resolved, if no candidates matching `amiMinimumAge` are found.
 
@@ -45,7 +46,8 @@ AMIs could be partially resolved, if no candidates matching `amiMinimumAge` are 
 If no `amiSelectorTerms` are specified, the AMIs candidates are currently resolved querying the SSM Parameter Store for the recommended EKS AMI matching the `amiFamily`,
 and then querying the AWS EC2 API for the AMI details (by specific `image-id`s).
 
-One solution would be switching to a broader query, so, instead of querying for the recommended AMI, we query for all the AMIs matching the `amiFamily` and then filter out the images younger than `amiMinimumAge`. 
+To implement this feature, we could switch to a broader query, so, instead of querying for the recommended AMI, we query for all the AMIs matching the `amiFamily` for the specified `eks` version,
+and then we filter out the images younger than `amiMinimumAge`. 
 
 for example, AL2 family, amd64 architecture:
 
--- a/designs/ami-minimum-age.md
+++ b/designs/ami-minimum-age.md
@@ -38,6 +38,8 @@ If `amiSelectorTerms` are specified, the AMIs candidates are currently resolved 
 
 Once the images are retrieved from the API, the images younger than `amiMinimumAge` are filtered out.
 
+AMIs could be partially resolved, if no candidates matching `amiMinimumAge` are found.
+
 ### AmiFamily and AmiMinimumAge
 
 If no `amiSelectorTerms` are specified, the AMIs candidates are currently resolved querying the SSM Parameter Store for the recommended EKS AMI matching the `amiFamily`,
@@ -59,6 +61,8 @@ for example, AL2 family, amd64 architecture:
 }
 ```
 
-We can then query the EC2 API for the AMIs matching the `image-id`s returned by the SSM query, and look for the recommended `image-ids`:
+We can then query the EC2 API for the AMIs matching the `image-id`s returned by the SSM query, and look for the recommended `image-id`s:
 - if the image with the recommended `image-id` is older than `amiMinimumAge`, we keep it as candidate
-- if the image with the recommended `image-id` is younger than `amiMinimumAge`, we look for the next image in the list that matches the `imageLocation` pattern and is older than `amiMinimumAge`
+- if the image with the recommended `image-id` is younger than `amiMinimumAge`, we look for the next image in the list that is older than `amiMinimumAge`
+
+AMIs could be partially resolved, if no candidates matching `amiMinimumAge` are found.
--- a/designs/ami-minimum-age.md
+++ b/designs/ami-minimum-age.md
@@ -1,0 +1,64 @@
+# Ami Minimum Age
+
+This document describes a design for a new feature to allow users to specify a minimum age for AMIs.
+
+(a @jonathan-innis's idea, see [this comment](https://github.com/aws/karpenter-provider-aws/issues/5382#issuecomment-1868068193))
+
+## Background
+
+Karpenter allows users to specify TTL for nodes, which configures the maximum age of nodes in the cluster. 
+
+This is useful to ensure that nodes are replaced periodically, which helps to ensure that the cluster is running on the latest AMIs available, which simplifies and automates the security patching process.
+
+However, a recent issue with a released AMI (see [this issue](https://github.com/awslabs/amazon-eks-ami/issues/1551)) has highlighted that it would be useful to be able to specify a minimum age for AMIs, to ensure that nodes are not replaced with an AMI that is too new.
+
+## Proposed solution
+
+Add a new field `amiMinimumAge` to the `EC2NodeClass`
+
+```
+apiVersion: karpenter.k8s.aws/v1beta1
+kind: EC2NodeClass
+metadata:
+  name: my-node-class
+spec:
+  amiFamily: AL2
+  amiMinimumAge: 2w
+```
+
+and modify the logic that selects the AMI to ensure that the AMI is at least `amiMinimumAge` old.
+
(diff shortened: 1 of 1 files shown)


cebernardi:
Basically the proposed change is the following:

we query SSM for the recommended Amis, and then we check with the `ec2 describe-image` if the creation date complies with the newly introduced `amiminimumage` parameter. If not, we pick the next ami that matches the required minimum age. 

some assumptions here:
- if a recommended ami is more recent than the required `miMinumAge`, we need to retrieve anyway a "valid" image (is this even true? [This test suite](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/providers/amifamily/suite_test.go#L189) seems to suggest that in other scenarios we accept we don't have a valid ami)
- we can't query the SSM parameter history of the recommended amis (I have a very powerful role and I couldn't get the parameter history, so I assumed the parameter was protected)
- the "approved" eks node images are always published with the naming convention of `amazon-eks-arm64-node-<eks-version>-` or similar. It sounds fragile, I hope I'm missing something: the ec2 api doesn't allow to specify the image location in the filter, so I'm wondering if the name of the ami is robust enough.

cebernardi:
For brevity, the code only implements the proposed change for AL2 family, to gather some feedback about the strategy overall, so the other families are missing (on purpose)

njtran:
Hey @cebernardi can you go into a bit of your use case here? For features like these, it would be nice to talk through what you're trying to achieve and if this is the best way to go about it. Usually the best way to do this is to open a RFC (review for comments) or design doc that details the use-cases and the reason you picked this solution.

cebernardi:
I was picking up from [this discussion](https://github.com/aws/karpenter-provider-aws/issues/5382), the `amiMinimumAge` was proposed by @jonathan-innis , and it seemed perfect for the use case discussed on the issue.

I'm happy to open a RFC, but given that's Jonathan's idea, would you still suggest me to open one?

njtran:
I think it'd be best to open a RFC, as it seems like he was just proposing an idea, but there wasn't much discussion on if we want to do this.

cebernardi:
Sounds good!

cebernardi:
@njtran I reverted the code and added a design document. 
Please let me know if I interpreted your request correctly (I couldn't find any reference to RFC format on the documentation)

cebernardi:
should I also change it from draft to a proper PR?

designs/ami-minimum-age.md
```diff
@@ -0,0 +1,70 @@
...
+Add a new field `amiMinimumAge` to the `EC2NodeClass`
+
+```
+apiVersion: karpenter.k8s.aws/v1beta1
+kind: EC2NodeClass
+metadata:
+  name: my-node-class
+spec:
+  amiFamily: AL2
+  amiMinimumAge: 2w
```
njtran, line 27: I feel like this might make more sense to be a sub-field of an AMI Selector Term. I could imagine users wanting to set a amiMinimumAge as another selection logic for each term, rather than prescribing it as an overlay over all selector terms.

designs/ami-minimum-age.md
```diff
@@ -0,0 +1,70 @@
...
+This document describes a design for a new feature to allow users to specify a minimum age for AMIs.
+
+(a @jonathan-innis's idea, see [this comment](https://github.com/aws/karpenter-provider-aws/issues/5382#issuecomment-1868068193))
+
+## Background
+
+Karpenter allows users to specify TTL for nodes, which configures the maximum age of nodes in the cluster. 
+
+This is useful to ensure that nodes are replaced periodically, and helps to ensure that the cluster is running on the latest AMIs available, 
+simplifying and automating the security patching process.
```
njtran, line 12: One thing i'm curious about is that you're looking to use this in response to the AMI issue that you've tagged below. What would you have set on this amiMinimumAge if it was specified?

One of the issues with this is if you have a new AMI released as a CVE patch, or a roll forward from a faulty image, would you still want to wait the amiMinimumAge to get that AMI? I think in almost all cases like that, you'd want to grab the image as soon as possible.

designs/ami-minimum-age.md
```diff
@@ -0,0 +1,70 @@
+# Ami Minimum Age
+
+This document describes a design for a new feature to allow users to specify a minimum age for AMIs.
+
+(a @jonathan-innis's idea, see [this comment](https://github.com/aws/karpenter-provider-aws/issues/5382#issuecomment-1868068193))
+
+## Background
+
+Karpenter allows users to specify TTL for nodes, which configures the maximum age of nodes in the cluster.
```
njtran, line 9: I think you're referring to expiration here, which I don't think really has relevance here. In the case of an AMISelector Term that doesn't use AMI ID, we look periodically at if the AMI has changed every 5 minutes, and roll that out. This minimumAge would be a way to modify our discovery logic to change what that resolved AMI would be.

njtran (commented):
Hey @cebernardi thanks for cutting it.  I've left a few high level comments.  You can take it out of draft if you'd like.

designs/ami-minimum-age.md
cebernardi: my understanding is that a new AMI doesn't trigger a drift, so nodes running on the old AMI won't be recycled. 
So setting an expiration makes sure that the entire cluster is bounced to the new AMI, over a certain period of time.
Is this correct, or when a new AMI is detected, the nodes running the old AMI are marked as drifted?

cebernardi: found this in the documentation: https://karpenter.sh/docs/concepts/disruption/#special-cases-on-drift but not sure if it's related also to configurations where no `amiSelectorTerms` is set?

designs/ami-minimum-age.md
cebernardi: I think we would pick an interval that will comply with the most stringent security SLA that our company has defined, something like 5d.

We sourced the referred bad AMI in our production clusters the same day it was published. If we had this mechanism we would have pulled it in only in our integration environment, where we would (hopefully) detected the issue, and stopped the production clusters from pulling the new AMI until the problem was solved.

In the scenarios you're describing (an urgent fix-forward AMI release) we would have released another version of the EC2NodeClass, maybe with no or a shorter minimum age. That would have triggered a drift and reclaimed the nodes out of compliance (please correct me if I'm wrong).

this is correlated with [this comment](https://github.com/aws/karpenter-provider-aws/pull/5551#discussion_r1481936955) about how a new AMI and drift disruption work. I think I might not have a clear picture of it.

designs/ami-minimum-age.md
cebernardi: I see your point.

Just to double check, if `amiMinimumAge` was a sub-field of AMI Selector Term, it wouldn't be available to users that only want to set `amiFamily`.

Users that only need to specify  the `amiFamily` and a minimum age, would need to use a configuration like the following (assuming AL2 and eks 1.28):

```
amiSelectorTerms:
- name: "amazon-eks*node-1.28*"
   minimumAge: 5d
```
it seems to complicate a bit the simplest configuration scenario (I need the most recent AMI of the specified family, older than `minAge`), but I guess you see `amiMinimumAge` as a sort of "advanced" configuration option that warrants some more maintenance work from users? 
I'm thinking in particular at use cases when eks gets updated. Where before users that only specified `amiFamily` didn't have to upgrade their NodeClasses, now they would have to push a new NodeClasses configuration.

I'm wondering if we could do both?
The way the AMIs are discovered right now is mutually exclusive: https://github.com/aws/karpenter-provider-aws/blob/main/pkg/providers/amifamily/ami.go#L118

if there are `amiSelectorTerms` specified, the `amiFamily` is ignored. 

Can we adopt the same logic and have something like
```
kind: EC2NodeClass
metadata:
  name: my-node-class
spec:
  amiFamily: AL2
  amiMinimumAge: 2w
```

as well as
```
kind: EC2NodeClass
metadata:
  name: my-node-class
spec:
  amiSelectorTerms:
  - name: "amazon-eks*node-1.28*"
     minimumAge: 5d
```

and have the best of the 2 world?

njtran: You bring up a good point, and even if there was a way to have an amiSelectorTerm, it's not _as_ simple to define it within each term. On the topic of having it in two places, I'd really opt against that, as that it may be a best of both worlds for some people, but is likely to have semantic and education confusion later.

cebernardi: So we c/should keep it as first level attribute (`spec.amiMinimumAge`)?

njtran:
Hey @cebernardi, planning on taking another look at this next week, but in the meantime, there's some discussion amongst the maintainers to figure out how to handle the issues that are motivating you to cut this. I'll post here when I have more information on this.

njtran:
Hey @cebernardi, I know I said I would give you an update on this last week, but still figuring some things out. We're currently figuring out if there are any additional guardrails that we can implement in clusters that would allow you to control the rate that new AMIs are rolled out in your cluster. Can you give some more details about how you detected the issue with the AMI you had? How would you have best liked to see the issue detected?

cebernardi:
Hi @njtran 
we detected the issue because we got a spike of applications failing to schedule on the affected nodes.
The spike was happening both in test and production, which was confusing because we couldn't immediately link it to any change we deployed and we were actively monitoring.

We're not testing _specifically_ new AMIs, but we have synthetic use-cases running twice a day in our test environments.

Ideally the spike would have been happening only in test environments, so the higher environments would have been spared.

Does it answer your question?

njtran:
@cebernardi Ah so you use automatic upgrades (no AMISelectorTerms) in both dev and production?

cebernardi:
Yes

> @cebernardi Ah so you use automatic upgrades (no AMISelectorTerms) in both dev and production?

Yes

njtran:
It may be worth your time to read through https://karpenter.sh/docs/tasks/amitasks/#choosing-ami-tasks. We added some more details on what we think is the best way to model your AMI rollouts. Not saying this fixes everything, but this should help a bit while we're still figuring out things on our end.

njtran:
Closing this due to a conversation that @jonathan-innis and @cebernardi had at KubeCon that we don't think minimum age is the right approach here.

cebernardi:
we tested the approach suggested [here](https://github.com/aws/karpenter-provider-aws/pull/5551#issuecomment-1979677824), and we found it adds an (avoidable, in my opinion) overhead on the operational side. 
Please see [this comment](https://github.com/aws/karpenter-provider-aws/issues/5382#issuecomment-2056555016).

Can we reopen this?

PR #5620, opened 2024-02-07
ci: Fix release script to succeed with xargs

jonathan-innis:
Fixes #N/A 

**Description**

Fix `release.sh` script to succeed when running with `xargs`

Running a local `Dockerfile` that uses ubuntu and ensures that the script succeeds 

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/website/content/en/docs/faq.md
+++ b/website/content/en/docs/faq.md
@@ -179,10 +179,10 @@ Yes, see the [KubeletConfiguration Section in the NodePool docs]({{<ref "./conce
 The difference between the Core and Full variants is that Core is a minimal OS with less components and no graphic user interface (GUI) or desktop experience.
 `Windows2019` and `Windows2022` AMI families use the Windows Server Core option for simplicity, but if required, you can specify a custom AMI to run Windows Server Full.
 
-You can specify the [Amazon EKS optimized AMI](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html) with Windows Server 2022 Full for Kubernetes {{< param "latest_k8s_version" >}} by configuring an `amiSelector` that references the AMI name.
+You can specify the [Amazon EKS optimized AMI](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html) with Windows Server 2022 Full for Kubernetes 1.29 by configuring an `amiSelector` that references the AMI name.
 ```yaml
 amiSelectorTerms:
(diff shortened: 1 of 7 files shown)


website/content/en/docs/faq.md
```diff
@@ -179,10 +179,10 @@ Yes, see the [KubeletConfiguration Section in the NodePool docs]({{<ref "./conce
 The difference between the Core and Full variants is that Core is a minimal OS with less components and no graphic user interface (GUI) or desktop experience.
 `Windows2019` and `Windows2022` AMI families use the Windows Server Core option for simplicity, but if required, you can specify a custom AMI to run Windows Server Full.
 
-You can specify the [Amazon EKS optimized AMI](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html) with Windows Server 2022 Full for Kubernetes {{< param "latest_k8s_version" >}} by configuring an `amiSelector` that references the AMI name.
+You can specify the [Amazon EKS optimized AMI](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html) with Windows Server 2022 Full for Kubernetes 1.29 by configuring an `amiSelector` that references the AMI name.
```
jmdeal, line 182: Should this not be the latest k8s version in the latest docs? Latest k8s version should correspond to the latest k8s version supported by Karpenter, not the overall latest. I did change this for some of the older versions since 1.29 is only supported for v0.34+. Seems like a long term item would be to utilize the compatibility matrix and populate with the latest supported version across all doc versions but this should still work here.

jonathan-innis: > Seems like a long term item would be to utilize the compatibility matrix and populate with the latest supported version across all doc versions but this should still work here

We always template over the value when releasing a new minor version of Karpenter. I think, generally, when we are going to be bumping the supported K8s version, we are going to be doing minor version releases of Karpenter which should bump the compat matrix and automatically take care of this. If we are doing a patch version release, the docs gen that we have right now has to manually take care of this anyways, but realistically the latest "versioned" thing should match the "docs" version since this is just meant to be a symlink anyways.

I agree that there's probably room to build automation around the compat matrix. Maybe something that we can take as a separate item?

jmdeal: I think that seems reasonable. Definitely don't need to include the compatibility matrix driven automation here, would be good to include as a separate item in the future.

engedaam (approved):
LGTM 🚀

PR #5640, opened 2024-02-09
feat: Add minValues flexibility to the NodePool requirement.

nikmohan123:
**Description**
This PR adds `minValues` flexibility to the NodePool requirement. The new NodePool requirement will be represented as:
```
type NodeSelectorRequirementWithMinValues struct {
	v1.NodeSelectorRequirement `json:",inline"`
	// MinValues is the minimum number of unique values required to define the flexibility of the specific requirement.
	// +kubebuilder:validation:Minimum:=1
	// +kubebuilder:validation:Maximum:=100
	// +optional
	MinValues *int `json:"minValues,omitempty"`
}
```

`minValues` represents the minimum number of unique values that needs to be considered for a particular requirement. So, this value ensures that the scheduler creates NodeClaims meeting this new requirement.

This PR has a dependency on another PR (Karpenter-sigs) that I am going to send out shortly (Update: Here is the PR: https://github.com/nikmohan123/karpenter/pull/1/files). Most of the changes in this PR is about changing the requirement struct to a new one that has flexibility of `minValues` in it. The main changes are related to lifting the restrictions/truncation logic that we have for the list of InstanceTypes that we send today to CreateFleet and also bypassing the logic that filters the expensive instanceTypes. This is to ensure that the `minValues` flexibility of the requirement is not compromised.

More details in: https://github.com/kubernetes-sigs/karpenter/pull/966

Functional tests added.
Manually tested as well.

- [X] Yes, issue opened: # <TODO> 
- [] No


pkg/providers/instance/instance.go
```diff
@@ -84,8 +84,12 @@ func NewProvider(ctx context.Context, region string, ec2api ec2iface.EC2API, una
...
-	instanceTypes = p.filterInstanceTypes(nodeClaim, instanceTypes)
-	instanceTypes = orderInstanceTypesByPrice(instanceTypes, scheduling.NewNodeSelectorRequirements(nodeClaim.Spec.Requirements...))
+	schedulingRequirements := scheduling.NewNodeSelectorRequirements(nodeClaim.Spec.Requirements...)
+	// Only filter the instances if there are no minValues in the requirement.
+	if !schedulingRequirements.HasMinValues() {
+		instanceTypes = p.filterInstanceTypes(nodeClaim, instanceTypes)
+	}
+	instanceTypes = orderInstanceTypesByPrice(instanceTypes, schedulingRequirements)
 	if len(instanceTypes) > MaxInstanceTypes {
 		instanceTypes = instanceTypes[0:MaxInstanceTypes]
```
jonathan-innis, line 94: Seems like we don't even need to truncate anymore if the upstream project is doing it for us

pkg/providers/instancetype/suite_test.go
```diff
@@ -425,6 +430,162 @@ var _ = Describe("InstanceTypes", func() {
 			Expect(spotPrice).To(BeNumerically("<", cheapestODPrice))
 		}
 	})
+	It("should consider the minValues from any requirement for capping InstanceTypeOptions", func() {
```
jonathan-innis, line 433: I don't think that we need to cap anymore and we can just trust what we are given through the NodeClaim now

pkg/fake/utils.go
```diff
@@ -155,3 +160,66 @@ func matchTags(tags []*ec2.Tag, filter *ec2.Filter) bool {
 	}
 	return false
 }
+
+func MakeFakeInstances() []*ec2.InstanceTypeInfo {
```
jonathan-innis, line 164:
```suggestion
func MakeInstances() []*ec2.InstanceTypeInfo {
```

Do you need fake in the name of any of these packages anymore since the call will always be prefixed with "fake". It's more idiomatic to say `fake.MakeInstances()`. The same applies to the rest of the functions that are shifting to this package

pkg/providers/instance/instance.go
nikmohan123: You are right. Removed this.

pkg/providers/instancetype/suite_test.go
nikmohan123: Yes, this is a old one. We should remove this since we have changed how the results from the scheduler is handled for the minimum requirement.

pkg/fake/utils.go
nikmohan123: Changed. Makes sense.

pkg/providers/instance/instance.go
```diff
@@ -51,7 +51,7 @@ import (
 
 var (
 	// MaxInstanceTypes defines the number of instance type options to pass to CreateFleet
-	MaxInstanceTypes                 = 60
+	MaxInstanceTypes                 = 100
```
jonathan-innis, line 54: Can we just drop this value if we aren't truncating over here anymore?

test/suites/integration/validation_test.go
```diff
@@ -124,6 +129,39 @@ var _ = Describe("Validation", func() {
 			}
 			Expect(env.Client.Create(env.Context, nodePool)).ToNot(Succeed())
 		})
+		It("should error when minValues for a requirement key is negative", func() {
```
jonathan-innis, line 132: Should this test also check 0 since we are also saying that you can't put 0 in that value as part of the kubebuilder spec

pkg/providers/instancetype/suite_test.go
```diff
@@ -425,6 +430,117 @@ var _ = Describe("InstanceTypes", func() {
...
+		for _, ltc := range call.LaunchTemplateConfigs {
+			for _, ovr := range ltc.Overrides {
+				if strings.Contains(aws.StringValue(ovr.InstanceType), "metal") {
+					expensiveInstanceType = true
+				}
+			}
+		}
+		Expect(expensiveInstanceType).To(BeTrue())
+	})
+	It("should not remove expensive metal instanceTypeOptions if any of the requirement with minValues is provided", func() {
```
jonathan-innis, line 476: How is this test different from the test above it?

pkg/providers/instancetype/suite_test.go
```diff
@@ -425,6 +430,117 @@ var _ = Describe("InstanceTypes", func() {
...
+		for _, ltc := range call.LaunchTemplateConfigs {
+			for _, ovr := range ltc.Overrides {
+				if strings.Contains(aws.StringValue(ovr.InstanceType), "metal") {
+					expensiveInstanceType = true
+				}
+			}
+		}
+		Expect(expensiveInstanceType).To(BeTrue())
+	})
+	It("should de-prioritize metal if instance-type requirement is specified but minValues isn't present", func() {
```
jonathan-innis, line 513:
```suggestion
	It("should de-prioritize metal if minValues isn't present", func() {
```

jonathan-innis, line 513: How different is this test from the test after it? "Should deprioritize metal"

pkg/providers/instance/instance.go
nikmohan123: Yeah, I dropped this. But it is also used in a test though to make sure LT would have the 100 cheapest instancetype as part of CreateFleet call since that is still relevant. We just happen to do this upstream now. So, I have just hard coded it for the test.

pkg/providers/instancetype/suite_test.go
nikmohan123: Ah, it isn't. It's redundant. Good catch. Let me remove one.

pkg/providers/instancetype/suite_test.go
nikmohan123: This is a redundant test like you said below. Removed it.

nikmohan123: Same. It's redundant. Removed this. For some reason, I thought this test wasn't available for metal instances and added that. My bad.

pkg/cloudprovider/suite_test.go
```diff
@@ -188,6 +192,302 @@ var _ = Describe("CloudProvider", func() {
...
+			uniqueInstanceTypes := sets.Set[string]{}
+			for _, launchTemplateConfig := range createFleetInput.LaunchTemplateConfigs {
+				for _, override := range launchTemplateConfig.Overrides {
+					uniqueInstanceTypes.Insert(*override.InstanceType)
+				}
+			}
+			// This ensures that we have sent the minimum number of requirements defined in the NodePool.
+			Expect(len(uniqueInstanceTypes)).To(BeNumerically(">=", 2))
+		})
+		It("createfleet input should respect minValues for Exists Operator in requirement from NodePool", func() {
```
jonathan-innis, line 293:
```suggestion
		It("CreateFleet input should respect minValues for Exists Operator in requirement from NodePool", func() {
```

nit

pkg/cloudprovider/suite_test.go
```diff
@@ -188,6 +192,302 @@ var _ = Describe("CloudProvider", func() {
 			Expect(createFleetInput.Context).To(BeNil())
 		})
 	})
+	Context("MinValues", func() {
+		It("createfleet input should respect minValues for In operator requirement from NodePool", func() {
```
jonathan-innis, line 196:
```suggestion
		It("CreateFleet input should respect minValues for In operator requirement from NodePool", func() {
```

nit

pkg/cloudprovider/suite_test.go
```diff
@@ -188,6 +192,302 @@ var _ = Describe("CloudProvider", func() {
...
+			uniqueInstanceTypes := sets.Set[string]{}
+			for _, launchTemplateConfig := range createFleetInput.LaunchTemplateConfigs {
+				for _, override := range launchTemplateConfig.Overrides {
+					uniqueInstanceTypes.Insert(*override.InstanceType)
+				}
+			}
+			// This ensures that we have sent the minimum number of requirements defined in the NodePool.
+			Expect(len(uniqueInstanceTypes)).To(BeNumerically(">=", 2))
+		})
+		It("createfleet input should respect minValues from multiple keys in NodePool", func() {
```
jonathan-innis, line 387:
```suggestion
		It("CreateFleet input should respect minValues from multiple keys in NodePool", func() {
```

nit

jonathan-innis (dismissed):
Just some nits. LGTM 🚀

pkg/cloudprovider/suite_test.go
nikmohan123: Done

pkg/cloudprovider/suite_test.go
nikmohan123: Done

pkg/cloudprovider/suite_test.go
nikmohan123: Done

jonathan-innis (dismissed):
LGTM 🚀

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (approved):
LGTM 🚀

PR #5659, opened 2024-02-13
docs: Remove the word "necessary" from optional fields

jonathan-innis:
Fixes #N/A 

**Description**

This updates the docs to clarify that only one of `spec.role` and `spec.instanceProfile` is needed since the previous wording might have been confusing with respect to this.

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/website/content/en/docs/concepts/nodeclasses.md
+++ b/website/content/en/docs/concepts/nodeclasses.md
@@ -483,7 +483,7 @@ Specify using ids:
 
 ## spec.role
 
-`Role` is an optional field and is necessary to tell Karpenter which identity nodes from this `EC2NodeClass` should assume. You must specify one of `role` or `instanceProfile` when creating a Karpenter `EC2NodeClass`. If using the [Karpenter Getting Started Guide]({{<ref "../getting-started/getting-started-with-karpenter" >}}) to deploy Karpenter, you can use the `KarpenterNodeRole-$CLUSTER_NAME` role provisioned by that process.
+`Role` is an optional field and tells Karpenter which IAM identity nodes should assume. You must specify one of `role` or `instanceProfile` when creating a Karpenter `EC2NodeClass`. If using the [Karpenter Getting Started Guide]({{<ref "../getting-started/getting-started-with-karpenter" >}}) to deploy Karpenter, you can use the `KarpenterNodeRole-$CLUSTER_NAME` role provisioned by that process.
 
 ```yaml
 spec:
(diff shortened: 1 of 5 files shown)


website/content/en/docs/concepts/nodeclasses.md
```diff
@@ -492,7 +492,9 @@ spec:
 
 ## spec.instanceProfile
 
-`InstanceProfile` is an optional field and is necessary to tell Karpenter which identity nodes from this `EC2NodeClass` should assume. You must specify one of `role` or `instanceProfile` when creating a Karpenter `EC2NodeClasss`. If you use the `instanceProfile` field instead of `role`, Karpenter will not manage the InstanceProfile on your behalf.
+`InstanceProfile` is an optional field and tells Karpenter which IAM identity nodes should assume. You must specify one of `role` or `instanceProfile` when creating a Karpenter `EC2NodeClass`. If you use the `instanceProfile` field instead of `role`, Karpenter will not manage the InstanceProfile on your behalf; instead, it expects that you have pre-provisioned an IAM instance profile and assigned it a role..
```
njtran, line 495: ominous....
```suggestion
`InstanceProfile` is an optional field and tells Karpenter which IAM identity nodes should assume. You must specify one of `role` or `instanceProfile` when creating a Karpenter `EC2NodeClass`. If you use the `instanceProfile` field instead of `role`, Karpenter will not manage the InstanceProfile on your behalf; instead, it expects that you have pre-provisioned an IAM instance profile and assigned it a role.
```

jonathan-innis: Lol :D

andrewhibbert:
Would it be possible to clarify this bit:

> For [private clusters](https://docs.aws.amazon.com/eks/latest/userguide/private-clusters.html) that do not have access to the public internet, using spec.instanceProfile is required. spec.role cannot be used since Karpenter needs to access IAM endpoints to manage a generated instance profile. IAM [doesn’t support private endpoints](https://docs.aws.amazon.com/vpc/latest/privatelink/aws-services-privatelink-support.html) to enable accessing the service without going to the public internet.

If you're in a private cluster with access to internet via NAT gateway, you'll reach IAM over the public internet, so it would be good to know the implications of this in terms of cost, does it contact the endpoint just once to generate the profile or is it a repetitive process?

jonathan-innis:
> good to know the implications of this in terms of cost

I'm not sure what you mean in terms of cost. IAM entities don't incur additional cost so I don't think that should be a concern there. In terms of the actual behavior of Karpenter, it is a continuous reconciliation so if you were to delete the instance profile accidentally or remove the role, Karpenter would add it back. You need constant access to IAM if you are using the `spec.role` field.

njtran (approved):

andrewhibbert:
Possibly I'm missing something, if using an instance profile, it wouldn't go over public internet to IAM, whereas with a role it would and do so continuously to reconcile? Sounds like it would cost more

PR #5661, opened 2024-02-14
chore: Drop unneeded scripts and targets from Karpenter

jonathan-innis:
Fixes #N/A 

**Description**

This change removes unneeded scripts from our website. These scripts were calling a combination of other steps but they were never used or referenced during the dev loop so they were not updated or maintained.

This change also removes unneeded targets that we have in our `Makefile`. Again, we have a `make setup` that is never used during the dev loop so it is not maintained by developers.

`make presubmit`

- [ ] Yes, issue opened: # 
- [x] No


Makefile
```diff
@@ -55,31 +52,40 @@ run: ## Run Karpenter controller binary against your local cluster
...
+
 e2etests: ## Run the e2e suite against your local cluster
 	cd test && CLUSTER_ENDPOINT=${CLUSTER_ENDPOINT} \
 		CLUSTER_NAME=${CLUSTER_NAME} \
 		INTERRUPTION_QUEUE=${CLUSTER_NAME} \
 		go test \
 		-p 1 \
 		-count 1 \
-		-timeout ${TEST_TIMEOUT} \
+		-timeout 3h \
```
engedaam, line 77: Why go for hardcode values here?

website/content/en/preview/getting-started/getting-started-with-karpenter/scripts/add-monitoring.sh
```diff
@@ -1,22 +0,0 @@
-#!/bin/bash
```
engedaam, line 1: Should we remove these files from the versioned sections

Makefile
jonathan-innis: I don't think that anyone was ever changing these values, so I didn't see a need for them to be parameterized. If we decide that we want/need them to be parameterized later, we can always add the environment variables back

website/content/en/preview/getting-started/getting-started-with-karpenter/scripts/add-monitoring.sh
jonathan-innis: Updated

engedaam (approved):
LGTM 🚀

PR #5666, opened 2024-02-15
chore: respect AssociatePublicIPAddress with EFA

jmdeal:
Fixes #N/A 

**Description**
Updates network interface generation to respect `AssociatePublicIPAddress` for EFA interfaces. This originally wasn't rejected due to a potential footgun: setting this value to true when launching instances with multiple EFAs will result in launch failure. The following error will be reported:
```
{"level":"ERROR","time":"2024-02-15T05:41:31.594Z","logger":"controller","message":"Reconciler error","commit":"95f42f1-dirty","controller":"nodeclaim.lifecycle","controllerGroup":"karpenter.sh","controllerKind":"NodeClaim","NodeClaim":{"name":"efa-tt6zp"},"namespace":"","name":"efa-tt6zp","reconcileID":"45b919ce-8b3a-480f-b0cd-a10eec71ff91","error":"launching nodeclaim, creating instance, with fleet error(s), InvalidParameterCombination: The associatePublicIPAddress parameter cannot be specified when launching with multiple network interfaces.; UnfulfillableCapacity: Unable to fulfill capacity due to your request configuration. Please adjust your request and try again."}
```
It isn't possible to eliminate this footgun through validation of the EC2NodeClass becuase EFA requests are drive through pod resource requests, not through any field on the EC2NodeClass. 

This left us with a few options:
1) Always respect `AssociatePublicIPAddress`, even when it results in instance launch failure. This is considered user misconfiguration.
2) Only respect `AssociatePublicIPAddress` when set to false. This eliminates the footgun while still enabling the most useful application (that we've heard requested), disabling public IPs in public subnets.
3) Don't respect `AssociatePublicIPAddress`, always set to nil.

After some discussion with @jonathan-innis offline, we decided option 1 was the best option. The following reasons were considered:
- Principle of least surprise, don't silently drop user configured settings
- Enables edge case of enabling on single interface EFA instances
- Error message from EC2 is plenty descriptive

`make test`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/launchtemplate/suite_test.go
+++ b/pkg/providers/launchtemplate/suite_test.go
@@ -1560,7 +1560,7 @@ var _ = Describe("LaunchTemplates", func() {
 				Expect(*input.LaunchTemplateData.ImageId).To(ContainSubstring("test-ami"))
 			})
 		})
-		Context("Subnet-based Launch Template Configration", func() {
+		Context("Public IP Association", func() {
 			It("should explicitly set 'AssociatePublicIPAddress' to false in the Launch Template", func() {
 				nodeClass.Spec.SubnetSelectorTerms = []v1beta1.SubnetSelectorTerm{
 					{Tags: map[string]string{"Name": "test-subnet-1"}},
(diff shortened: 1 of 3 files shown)


website/content/en/preview/concepts/nodeclasses.md
```diff
@@ -872,9 +872,8 @@ spec:
...
 
 {{% alert title="Note" color="warning" %}}
-If a `NodeClaim` requests `vpc.amazonaws.com/efa` resources, the `associatePublicIPAddress` field is ignored.
-A public IP address may only be associated with a node at launch if a single network interface is configured.
-This is inherently incompatible with instances configured for EFA workloads since Karpenter will configure an EFA for each network card on the instance.
+If a `NodeClaim` requests `vpc.amazonaws.com/efa` resources, `spec.associatePublicIPAddress` is respected. However, if this `NodeClaim` requests **multiple** EFA resources and the value for `spec.associatePublicIPAddress` is true, the instance will fail to launch. This is due to an EC2 restriction which
+requires that the field is only set to true when configuring an instance with a single ENI at launch. When using this field, it is advised that users segregate their EFA workload to use a separate `NodePool` / `EC2NodeClass` pair.
```
engedaam, line 876: Should we include the reference for customers?  https://docs.aws.amazon.com/cli/latest/reference/ec2/request-spot-fleet.html

pkg/providers/launchtemplate/suite_test.go
```diff
@@ -1560,7 +1560,7 @@ var _ = Describe("LaunchTemplates", func() {
 				Expect(*input.LaunchTemplateData.ImageId).To(ContainSubstring("test-ami"))
 			})
 		})
-		Context("Subnet-based Launch Template Configration", func() {
+		Context("Public IP Association", func() {
```
engedaam, line 1563: Should we add or change our E2E tests?

jmdeal: So these changes are a little difficult to test in E2E and I'm not sure if there's much advantage to increasing the scope of the existing tests. As long as the lt is correctly generated I'm not worried EFA vs non-EFA instances from an E2E standpoint.

jmdeal: E.G. on the difficult to test point, can't test setting the association to false in a public subnet. There's probably some work we could do by grabbing the provider ID off the nodeclass and describing the instance through the AWS API but I don't think that really gains us anything over the existing tests. The LT tests plus a couple smoke tests seem sufficient to me.

website/content/en/preview/concepts/nodeclasses.md
jmdeal: Looking at this ref again it's actually incorrect. Will probably drop it from the code, it describes the right behavior but for a different API. I've not been able to find any reference of this behavior for the APIs we use. I think it's fine for users to take our word for it, with this change they can try it themselves if they don't believe us 🤷. I would want to link if I could find a reference that actually applies to the APIs we use directly though.

pkg/providers/launchtemplate/suite_test.go
engedaam: Do we do any specific E2E testing for EFA?

jmdeal: Ya we do, in the extended resources tests: https://github.com/aws/karpenter-provider-aws/blob/d9a2d1b95ce39856d93ea68600a38d468333b8c4/test/suites/integration/extended_resources_test.go#L214

engedaam (approved):
LGTM 🚀

PR #5680, opened 2024-02-16
feat: Changed OCI tag format

stevehipwell:
Fixes #N/A 

**Description**
This PR changes the OCI tag pattern to be the SemVer version without the `v` prefix.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/hack/release/common.sh
+++ b/hack/release/common.sh
@@ -88,14 +88,14 @@ publishHelmChart() {
   commit_sha="${4}"
   build_date="${5}"
 
-  ah_config_file_name"${helm_chart}/artifacthub-repo.yaml"
+  ah_config_file_name="${helm_chart}/artifacthub-repo.yaml"
   helm_chart_artifact="${helm_chart}-${version}.tgz"
 
   yq e -i ".appVersion = \"${version}\"" "charts/${helm_chart}/Chart.yaml"
   yq e -i ".version = \"${version}\"" "charts/${helm_chart}/Chart.yaml"
 
   cd charts
-  [[ -s "${ah_config_file_name}" ]] && oras push "${oci_repo}:artifacthub.io" --config /dev/null:application/vnd.cncf.artifacthub.config.v1+yaml "${ah_config_file_name}:application/vnd.cncf.artifacthub.repository-metadata.layer.v1.yaml"
+  [[ -s "${ah_config_file_name}" ]] && oras push "${oci_repo}/${helm_chart}:artifacthub.io" --config /dev/null:application/vnd.cncf.artifacthub.config.v1+yaml "${ah_config_file_name}:application/vnd.cncf.artifacthub.repository-metadata.layer.v1.yaml"
   helm dependency update "${helm_chart}"
   helm lint "${helm_chart}"
   helm package "${helm_chart}" --version "${version}"
@@ -103,8 +103,8 @@ publishHelmChart() {
   rm "${helm_chart_artifact}"
   cd ..
 
-  helm_chart_digest="$(crane digest "${oci_repo}:${version}")"
-  cosignOciArtifact "${version}" "${commit_sha}" "${build_date}" "${oci_repo}:${version}@${helm_chart_digest}"
+  helm_chart_digest="$(crane digest "${oci_repo}/${helm_chart}:${version}")"
+  cosignOciArtifact "${version}" "${commit_sha}" "${build_date}" "${oci_repo}/${helm_chart}:${version}@${helm_chart_digest}"
 }
 
 cosignOciArtifact() {
--- a/hack/release/common.sh
+++ b/hack/release/common.sh
@@ -138,7 +138,7 @@ prepareWebsite() {
   version_parts=(${version//./ })
   short_version="${version_parts[0]}.${version_parts[1]}"
 
-  createNewWebsiteDirectory "${short_version}"
+  createNewWebsiteDirectory "${short_version}" "${version}"
   removeOldWebsiteDirectories
   editWebsiteConfig "${version}"
   editWebsiteVersionsMenu
@@ -199,7 +199,7 @@ editWebsiteVersionsMenu() {
   local versions version
 
   # shellcheck disable=SC2207
-  versions=($(find website/content/en/* -maxdepth 0 -type d -name "*" -print0 | xargs -r -n 1 basename | grep -v "docs\|preview"))
+  versions=($(find website/content/en/* -maxdepth 0 -type d -name "*" -print0 | xargs -0 -r -n 1 basename | grep -v "docs\|preview"))
   versions+=('preview')
 
   yq -i '.params.versions = []' website/hugo.yaml
--- a/hack/release/common.sh
+++ b/hack/release/common.sh
@@ -1,61 +1,48 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-config(){
-  GITHUB_ACCOUNT="aws"
-  ECR_GALLERY_NAME="karpenter"
-  RELEASE_REPO_ECR=${RELEASE_REPO_ECR:-public.ecr.aws/${ECR_GALLERY_NAME}/}
-  RELEASE_REPO_GH=${RELEASE_REPO_GH:-ghcr.io/${GITHUB_ACCOUNT}/karpenter}
+GITHUB_ACCOUNT="aws"
+ECR_GALLERY_NAME="karpenter"
+RELEASE_REPO_ECR="${RELEASE_REPO_ECR:-public.ecr.aws/${ECR_GALLERY_NAME}/}"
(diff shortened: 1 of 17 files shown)


hack/release/common.sh
```diff
@@ -66,136 +53,165 @@ authenticatePrivateRepo() {
...
-  for VERSION in "${VERSIONS[@]}"; do
-    yq -i ".params.versions += \"${VERSION}\"" website/hugo.yaml
+  for version in "${versions[@]}"; do
+    yq -i ".params.versions += \"${version}\"" website/hugo.yaml
   done
 }
+
+# updateKarpenterCoreGoMod bumps the karpenter-core go.mod to the release version so that the
+# karpenter and karpenter-core release versions match
+updateKarpenterCoreGoMod(){
```
jonathan-innis, line 206: I think we can drop this function from the stable PR. We ensure we pin the version through a helm update ahead-of-time so there's no need for this function to exist right now AFAIK

hack/release/common.sh
```diff
@@ -66,136 +53,165 @@ authenticatePrivateRepo() {
...
-    rm "${HELM_CHART_FILE_NAME}"
-    cd ..
-
-    cosignHelmChart "${RELEASE_REPO}/${CHART_NAME}" "${HELM_CHART_VERSION}"
+buildDate() {
+  local date_epoch
+
+  date_epoch="${1}"
+
+  date -u -r "${date_epoch}" "+%Y-%m-%dT%H:%M:%SZ" 2>/dev/null
```
jonathan-innis, line 129: What's the reason/need to change the command that is being called to produce the buildDate from before?

hack/release/stable-pr.sh
```diff
@@ -1,42 +1,25 @@
...
-if [[ $(releaseType $GIT_TAG) != $RELEASE_TYPE_STABLE ]]; then
+git_tag="$(git describe --exact-match --tags || echo "none")"
+if [[ "${git_tag}" != v* ]]; then
   echo "Not a stable release. Missing required git tag."
   exit 1
 fi
 
-versionData "$GIT_TAG"
-updateKarpenterCoreGoMod "$GIT_TAG"
+updateKarpenterCoreGoMod "${git_tag}"
```
jonathan-innis, line 14: I think we should be able to drop this command now

hack/release/release.sh
```diff
@@ -1,20 +1,20 @@
...
-HEAD_HASH=$(git rev-parse HEAD)
-
-SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
+# shellcheck source=hack/release/common.sh
 source "${SCRIPT_DIR}/common.sh"
-config
+
+git_tag="$(git describe --exact-match --tags || echo "no tag")"
+if [[ "${git_tag}" == "no tag" ]]; then
```
jonathan-innis, line 9: What's the reason this conditional check is different from the one that we use for `release-crd.sh`?

hack/release/common.sh
```diff
@@ -66,136 +53,165 @@ authenticatePrivateRepo() {
...
-    SOURCE_DATE_EPOCH=$(git log -1 --format='%ct')
-    echo "$(date -u -r "${SOURCE_DATE_EPOCH}" $DATE_FMT 2>/dev/null)"
+cosignOciArtifact() {
+  local commit_sha build_date artifact
+
+  commit_sha="${1}"
+  build_date="${2}"
+  artifact="${3}"
+
+  cosign sign --yes -a commitSha="${commit_sha}" -a buildDate="${build_date}" "${artifact}"
```
jonathan-innis, line 117: Seems like we aren't attesting the GIT_VERSION anymore?

hack/release/common.sh
```diff
@@ -66,136 +53,165 @@ authenticatePrivateRepo() {
...
-    RELEASE_REPO=$1
-    HELM_CHART_VERSION=$2
-    digest="$(crane digest "${RELEASE_REPO}:${HELM_CHART_VERSION}")"
-    cosign sign --yes "${RELEASE_REPO}:${HELM_CHART_VERSION}@${digest}"
+prepareWebsite() {
+  local version version_parts short_version
+
+  version="${1}"
+  # shellcheck disable=SC2206
+  version_parts=(${version//./ })
```
jonathan-innis, line 137: I'm not a bash expert, can you point me to docs that explain what this does?

hack/release/common.sh
```diff
@@ -66,136 +53,165 @@ authenticatePrivateRepo() {
...
+  mkdir -p "website/content/en/v${short_version}"
+  cp -r website/content/en/preview/* "website/content/en/v${short_version}/"
 
-    # Update parameterized variables in the preview documentation to be statically set in the versioned documentation
-    find "website/content/en/${RELEASE_MINOR_VERSION}/" -type f | xargs perl -i -p -e "s/{{< param \"latest_release_version\" >}}/${RELEASE_VERSION}/g;"
-    find "website/content/en/${RELEASE_MINOR_VERSION}/" -type f | xargs perl -i -p -e "s/{{< param \"latest_k8s_version\" >}}/$(yq .params.latest_k8s_version website/hugo.yaml)/g;"
-    find website/content/en/${RELEASE_MINOR_VERSION}/*/*/*.yaml -type f | xargs perl -i -p -e "s/preview/${RELEASE_MINOR_VERSION}/g;"
-    find "website/content/en/${RELEASE_MINOR_VERSION}/" -type f | xargs perl -i -p -e "s/{{< githubRelRef >}}/\/${RELEASE_VERSION}\//g;"
+  # Update parameterized variables in the preview documentation to be statically set in the versioned documentation
+  find "website/content/en/v${short_version}/" -type f -print0 | xargs perl -i -p -e "s/{{< param \"latest_release_version\" >}}/${RELEASE_VERSION}/g;"
```
jonathan-innis, line 153: I had recalled doing this and this messed with the output format in a way that `xargs` couldn't understand. I'll double-check as well, but you're sure that adding `-print0` here doesn't mess with the success of this command?

jonathan-innis (commented):
Nice work 🎉

hack/release/common.sh
stevehipwell: The git version was being added as an annotation; annotations aren't checked as part of validation unless opted in to. I can add a `version` attribute which can be checked as part of verification if there is the need to correlate the SHA with a version, although I'm not sure that's necessary.

hack/release/common.sh
stevehipwell: The command hasn't changed but I've refactored it to remove the unnecessary `echo` and extract the epoch as the same command was being used in multiple places.

hack/release/common.sh
stevehipwell: The outer `(` & `)` make an array with the contents. The `//./` replacement syntax swaps `.` for ` ` and by not wrapping the variable in `"` it is treated as multiple words so each part of the version is an array entry.

hack/release/common.sh
stevehipwell: Shellcheck is adamant that it's needed but I'll check again.

hack/release/common.sh
stevehipwell: Will do.

hack/release/release.sh
stevehipwell: This is guard clause and the other is a conditional extra function. I didn't modify these checks, just refactored the bash style.

hack/release/stable-pr.sh
stevehipwell: Will do.

hack/release/common.sh
stevehipwell: I've added the `version` attribute.

hack/release/common.sh
jonathan-innis: That's cool and fancy, didn't know that that would work

hack/release/common.sh
jonathan-innis: Sure thing. Let me know. I had also gotten warnings for not adding this before but I recalled that adding it had caused issues

stevehipwell: You're absolutely right, so I've supressed the Shellcheck warnings so it's clear this is the desired outcome.

stevehipwell:
@jonathan-innis I've updated the docs and have standardised the pattern used to refer to Karpenter versions for consistency (see below).

```markdown
Karpenter version: `0.34.1`
Karpenter minimum version: `0.34.0`+
Karpenter latest patch: `0.34.x` (this shouldn't be used where the minimum version would also work)
```

jonathan-innis (dismissed):
LGTM 🚀

jonathan-innis (dismissed):
LGTM 🚀

jmdeal (approved):
LGTM 🚀

PR #5704, opened 2024-02-21
fix: Update spot pricing even in Isolated VPC

liafizan:
Fixes ##5587 

**Description**
Spot pricing uses EC2 api and a check in code base disabled spot pricing check when `--isolated-vpc` flag was used. Removed condition which will allow spot prices to be updated when in `isolated-vpc`

Manual testing. Tried adding a test in pricing suite but it did not help since pricing controller itself was enabled which deemed this test unnecessary.

- [x ] Yes, PR includes docs updates 
- [ ] Yes, issue opened: # 
- [] No


website/content/en/docs/reference/settings.md
```diff
@@ -10,32 +10,32 @@ Karpenter surfaces environment variables and CLI parameters to allow you to conf
...
-| RESERVED_ENIS | \-\-reserved-enis | Reserved ENIs are not included in the calculations for max-pods or kube-reserved. This is most often used in the VPC CNI custom networking setup https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html. (default = 0)|
-| VM_MEMORY_OVERHEAD_PERCENT | \-\-vm-memory-overhead-percent | The VM memory overhead as a percent that will be subtracted from the total memory for all instance types. (default = 0.075)|
-| WEBHOOK_METRICS_PORT | \-\-webhook-metrics-port | The port the webhook metric endpoing binds to for operating metrics about the webhook (default = 8001)|
-| WEBHOOK_PORT | \-\-webhook-port | The port the webhook endpoint binds to for validation and mutation of resources (default = 8443)|
+| Environment Variable | CLI Flag | Description                                                                                                                                                                                                                                                                         |
```
engedaam, line 13: Thank you for making this PR! Just an FYI, this section is automatically generated from our doc generation file`hack/docs/configuration_gen_docs.go`. IsolatedVPC option should be updated as part of the flags https://github.com/aws/karpenter-provider-aws/blob/293f6476a8a6056fc34984419b7a0726ae9f1b5a/pkg/operator/options/options.go#L53 Also if you think we need to make changes to the formatting, please make it so in `configuration_gen_docs.go`

liafizan: Thank you. Ill make this change

pkg/providers/pricing/suite_test.go
```diff
@@ -262,6 +262,57 @@ var _ = Describe("Pricing", func() {
...
+		}))
+		awsEnv.PricingAPI.GetProductsOutput.Set(&awspricing.GetProductsOutput{
+			PriceList: []aws.JSONValue{
+				fake.NewOnDemandPrice("c98.large", 1.20),
+				fake.NewOnDemandPrice("c99.large", 1.23),
+			},
+		})
+		// this should fail as we are not providing spot pricing info which will be
+		// called even when isolated-vpc is set
+		ExpectReconcileFailed(ctx, controller, types.NamespacedName{})
```
engedaam, line 277: Should we just check the `pricingProvider.SpotPrice` error that is returned? so that we can validate the exact error.

engedaam:
Make sure you run `make presubmit` and `make docgen` which should make changes to get the ci-tests to pass

website/content/en/docs/reference/settings.md
engedaam, line 13: I forgot to mention that the docs update should be happening in the review folder. They should only be updated in preview for unreleased karpenter changes. The docs section would show users this change as current version of Karpenter. So all change to this file should be reverted.

charts/karpenter/README.md
```diff
@@ -15,7 +15,7 @@ You can follow the detailed installation instruction in the [documentation](http
 ```bash
   karpenter oci://public.ecr.aws/karpenter/karpenter \
-  --version v0.34.0 \
+  --version 0.34.0 \
```
engedaam, line 18: Did you mean to make this change?

liafizan: Not really. I ran `make presubmit` and `make docgen`. I can revert this!

website/content/en/docs/reference/settings.md
liafizan: I just ran `make docgen`, is that not the right thing to do? 

It seems like it updated this file and helm settings file. I can revert this but for getting docs in review, i do not see any make targets to get docs in review. Is there something I need to do. Sorry for missing this.

engedaam: Just tried `make docgen` and it's changing the helm settings files for me as well. You can just push the changes that are generated by `make docgen` and revert the changes in this file. Sorry for the confusion. After that we should be good for this PR

charts/karpenter/README.md
jonathan-innis: `make docgen` would definitely produce this. I think we can drop this and take this in separately. We should also generate this when we generate a new minor version release automatically.

pkg/operator/options/options.go
```diff
@@ -50,7 +50,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.StringVar(&o.ClusterCABundle, "cluster-ca-bundle", env.WithDefaultString("CLUSTER_CA_BUNDLE", ""), "Cluster CA bundle for nodes to use for TLS connections with the API server. If not set, this is taken from the controller's TLS configuration.")
 	fs.StringVar(&o.ClusterName, "cluster-name", env.WithDefaultString("CLUSTER_NAME", ""), "[REQUIRED] The kubernetes cluster name for resource discovery.")
 	fs.StringVar(&o.ClusterEndpoint, "cluster-endpoint", env.WithDefaultString("CLUSTER_ENDPOINT", ""), "The external kubernetes cluster endpoint for new nodes to connect with. If not specified, will discover the cluster endpoint using DescribeCluster API.")
-	fs.BoolVarWithEnv(&o.IsolatedVPC, "isolated-vpc", "ISOLATED_VPC", false, "If true, then assume we can't reach AWS services which don't have a VPC endpoint. This also has the effect of disabling look-ups to the AWS pricing endpoint.")
+	fs.BoolVarWithEnv(&o.IsolatedVPC, "isolated-vpc", "ISOLATED_VPC", false, "If true, then assume we can't reach AWS services which don't have a VPC endpoint. This also has the effect of disabling look-ups to the AWS pricing endpoint and only spot pricing will be updated.")
```
jonathan-innis, line 53:
```suggestion
	fs.BoolVarWithEnv(&o.IsolatedVPC, "isolated-vpc", "ISOLATED_VPC", false, "If true, then assume we can't reach AWS services which don't have a VPC endpoint. This also has the effect of disabling look-ups to the AWS on-demand pricing endpoint.")
```

nit

pkg/providers/pricing/controller.go
```diff
@@ -50,9 +52,13 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontrolle
 
 func (c *Controller) updatePricing(ctx context.Context) error {
 	work := []func(ctx context.Context) error{
-		c.pricingProvider.UpdateOnDemandPricing,
 		c.pricingProvider.UpdateSpotPricing,
 	}
+
+	if !options.FromContext(ctx).IsolatedVPC {
```
jonathan-innis, line 58: nit: Worth a comment over this to describe why we can fetch spot pricing but not on-demand pricing?

pkg/providers/pricing/suite_test.go
```diff
@@ -262,6 +262,44 @@ var _ = Describe("Pricing", func() {
...
+		awsEnv.PricingAPI.GetProductsOutput.Set(&awspricing.GetProductsOutput{
+			PriceList: []aws.JSONValue{
+				fake.NewOnDemandPrice("c98.large", 1.20),
+				fake.NewOnDemandPrice("c99.large", 1.23),
+			},
+		})
+		ExpectReconcileSucceeded(ctx, controller, types.NamespacedName{})
+		price, ok := awsEnv.PricingProvider.OnDemandPrice("c5.large")
+		Expect(ok).To(BeTrue())
+		Expect(price).To(BeNumerically(">", 0))
```
jonathan-innis, line 296: Should we also test that this is equal to the static pricing value that we have? Only checking that this is above 0 could be returning either the static data OR the data that you set in the output on L287, so it's hard to know if this will catch things if we make a mistake in the code here.

jonathan-innis (commented):
Thanks for adding the change!

jonathan-innis:
@liafizan Did you validate this on your own private cluster?

pkg/providers/pricing/suite_test.go
liafizan: Updated test to check static value

liafizan:
> @liafizan Did you validate this on your own private cluster?

Hey @jonathan-innis. I tested my changes in local cluster. Though not exhaustive but I can validate `--isolated-vpc` cases and see that both are invoked now when isolated-vpc is true or else only spot pricing is updated.

charts/karpenter/README.md
liafizan: I believe this is done, please let me know in case there is still something missing here

pkg/providers/pricing/controller.go
```diff
@@ -49,10 +53,18 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontrolle
 }
 
 func (c *Controller) updatePricing(ctx context.Context) error {
+	logging.FromContext(ctx).Infof("Updating spot price information")
```
jonathan-innis, line 56: We have a log right now that's directly in the pricing provider for [spot](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/providers/pricing/pricing.go#L367) and [on-demand](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/providers/pricing/pricing.go#L207) pricing that indicates when we have made updates to our pricing information. I don't think that we also need info logs that indicate that we are starting the update too

pkg/controllers/controllers.go
```diff
@@ -55,14 +54,10 @@ func NewControllers(ctx context.Context, sess *session.Session, clk clock.Clock,
 		nodeclass.NewController(kubeClient, recorder, subnetProvider, securityGroupProvider, amiProvider, instanceProfileProvider, launchTemplateProvider),
 		nodeclaimgarbagecollection.NewController(kubeClient, cloudProvider),
 		nodeclaimtagging.NewController(kubeClient, instanceProvider),
+		pricing.NewController(pricingProvider),
 	}
 	if options.FromContext(ctx).InterruptionQueue != "" {
 		controllers = append(controllers, interruption.NewController(kubeClient, clk, recorder, lo.Must(sqs.NewProvider(ctx, servicesqs.New(sess), options.FromContext(ctx).InterruptionQueue)), unavailableOfferings))
 	}
-	if options.FromContext(ctx).IsolatedVPC {
-		logging.FromContext(ctx).Infof("assuming isolated VPC, pricing information will not be updated")
```
jonathan-innis, line 63: We've dropped this log, which I think that users would still want to know. Is there a way to move this log into our pricing provider so that we still log this info if we are in an isolated VPC?

pkg/providers/pricing/suite_test.go
```diff
@@ -262,6 +262,44 @@ var _ = Describe("Pricing", func() {
...
+		})
+
+		awsEnv.PricingAPI.GetProductsOutput.Set(&awspricing.GetProductsOutput{
+			PriceList: []aws.JSONValue{
+				fake.NewOnDemandPrice("c98.large", 1.20),
+				fake.NewOnDemandPrice("c99.large", 1.23),
+			},
+		})
+		ExpectReconcileSucceeded(ctx, controller, types.NamespacedName{})
+		price, ok := awsEnv.PricingProvider.OnDemandPrice("c5.large")
```
jonathan-innis, line 294: nit: Should we get an on-demand price for the `c98.large` since we have a price at our API for it, but we want to get the static price. This should give us the confidence that we really aren't hitting our API, since we aren't using the value that we explicitly set in the test-case.

pkg/controllers/controllers.go
liafizan: Changed [here](https://github.com/aws/karpenter-provider-aws/pull/5704/commits/a39ce5dfc06a18579322857498ba7db1a89c0021). Please let me know this still needs change.

pkg/providers/pricing/suite_test.go
liafizan: Done

pkg/providers/pricing/controller.go
liafizan: removed this

pkg/providers/pricing/controller.go
```diff
@@ -50,9 +54,17 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontrolle
...
-		c.pricingProvider.UpdateOnDemandPricing,
 		c.pricingProvider.UpdateSpotPricing,
 	}
+
+	// we try to update on demand pricing only if we are not in isolated VPC, as
+	// AWS Price list api might not be accessible for isolated VPCs
+	if !options.FromContext(ctx).IsolatedVPC {
+		work = append(work, c.pricingProvider.UpdateOnDemandPricing)
+	} else {
+		logging.FromContext(ctx).Infof("assuming isolated VPC, on-demand pricing information will not be updated")
```
jonathan-innis, line 65: Is there a way to reduce spam to ensure that we only log this log once on startup or, minimally, that we only log it once per update to the pricing provider. My worry is that this is going to spam a lot if there is an error, where-as our pricing provider ensures that the log only comes every 12 hours using the change monitor. 

An alternative suggestion: What if we consider encapsulating the isolated VPC logic inside of the pricing provider and still call `c.pricingProvider.UpdateOnDemandPricing()`?

pkg/providers/pricing/suite_test.go
```diff
@@ -262,6 +262,46 @@ var _ = Describe("Pricing", func() {
...
+		awsEnv.PricingAPI.GetProductsOutput.Set(&awspricing.GetProductsOutput{
+			// these are incorrect prices which are here to ensure that
+			// results from only static pricing are used
+			PriceList: []aws.JSONValue{
+				fake.NewOnDemandPrice("c3.2xlarge", 1.20),
+				fake.NewOnDemandPrice("c5.xlarge", 1.23),
+			},
+		})
+		ExpectReconcileSucceeded(ctx, controller, types.NamespacedName{})
+		price, ok := awsEnv.PricingProvider.OnDemandPrice("c3.2xlarge")
```
jonathan-innis, line 296: nice 🎉

pkg/providers/pricing/controller.go
liafizan: Updated based on what I understood from your inputs. I am not sure if this is what you meant. With this change, log line should be added when reconciliation happens ( every 12 hours ). Let me know if this is not what you had in mind.

jonathan-innis (approved):
LGTM 🚀

PR #5770, opened 2024-03-03
feat: Add Versioned for EC2NodeClass Hash to Prevent Drift on EC2NodeClass CRD Upgrade

engedaam:
Fixes #N/A 

**Description**
Add versioned NodePool Hash to prevent Drift from occurring under these scenario:
1. A `EC2NodeClass` adds default values to existing fields that are included in the hash calculations
2. A `EC2NodeClass` field is added to the hash calculation with an already-set value
3. A `EC2NodeClass` field is removed from the hash calculations

RFC: https://github.com/kubernetes-sigs/karpenter/pull/1015
Related Issue: https://github.com/kubernetes-sigs/karpenter/issues/957

- `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No


pkg/controllers/nodeclass/controller.go
```diff
@@ -222,6 +232,39 @@ func (c *Controller) resolveInstanceProfile(ctx context.Context, nodeClass *v1be
...
+				v1beta1.AnnotationEC2NodeClassHashVersion: v1beta1.EC2NodeClassHashVersion,
+			})
+
+			// Any NodeClaim that is already drifted will remain drifted if the karpenter.sh/ec2nodeclass-hash-version doesn't match
+			// Since the hashing mechanism has changed we will not be able to determine if the drifted status of the node has changed
+			if ncList.Items[i].StatusConditions().GetCondition(corev1beta1.Drifted) == nil {
+				ncList.Items[i].Annotations = lo.Assign(ncList.Items[i].Annotations, map[string]string{
+					v1beta1.AnnotationEC2NodeClassHash: nodeClass.Hash(),
+				})
+			}
```
njtran, line 255: Have the same comment here as in the kubernetes-sigs. We probably want to not update the hash version if it's drifted as well

engedaam: We will not update the nodepool-hash on the NodeClaim, if the nodepool-hash-version match the NodePool

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -329,6 +329,12 @@ type EC2NodeClass struct {
 	Status EC2NodeClassStatus `json:"status,omitempty"`
 }
 
+// We need to bump the EC2NodeClassHashVersion under these condition when we make an update to the EC2NodeClass CRD
+// 1. A  adds default values to existing fields that are included in the hash calculations
```
jonathan-innis, line 333: We should have whatever comment that we have here match the comment that we have for the version over in the kubernetes-sigs/karpenter repo

pkg/controllers/nodeclass/suite_test.go
```diff
@@ -889,6 +890,118 @@ var _ = Describe("NodeClassController", func() {
 			nodeClass = ExpectExists(ctx, env.Client, nodeClass)
 			Expect(nodeClass.Annotations[v1beta1.AnnotationEC2NodeClassHash]).To(Equal(expectedHash))
 		})
+		It("should update ec2nodeclass hash version when the ec2nodeclass hash version is out of sync with the controller", func() {
```
jonathan-innis, line 893:
```suggestion
		It("should update ec2nodeclass hash version when the ec2nodeclass hash version is out of sync with the controller hash version", func() {
```

Same comments as in the other repo, we should be really explicit about what we are comparing and is saying is "out of sync" here

test/suites/drift/suite_test.go
```diff
@@ -719,6 +719,23 @@ var _ = Describe("Drift", func() {
 		env.EventuallyExpectNotFound(pod, node)
 		env.EventuallyExpectHealthyPodCount(selector, numPods)
 	})
+	It("should update NodeClass hash on the NodeClass and NodeClaim", func() {
```
jonathan-innis, line 722: Can you be more descriptive here? This isn't just about validating that we are updating the hash on the NodeClass and the NodeClaim, this is about testing this new versioning mechanism and validating that we are doing the repropagation like we think that we are

pkg/controllers/nodeclass/suite_test.go
```diff
@@ -889,6 +890,118 @@ var _ = Describe("NodeClassController", func() {
...
+			ExpectApplied(ctx, env.Client, nodeClass)
+
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+			nodeClass = ExpectExists(ctx, env.Client, nodeClass)
+
+			expectedHash := nodeClass.Hash()
+			Expect(nodeClass.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHash, expectedHash))
+			Expect(nodeClass.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHashVersion, v1beta1.EC2NodeClassHashVersion))
+		})
+		It("should update ec2nodeclass hash versions on all nodeclaims when the hash versions don't match", func() {
```
jonathan-innis, line 904:
```suggestion
		It("should update ec2nodeclass hash versions on all nodeclaims when the hash versions don't match the controller hash version", func() {
```

jonathan-innis, line 904: In this testcase, is the nodeclass hash version set to actually reflect a value that's different than the current hash version? Or is this just working because the hash version doesn't exist yet and that still counts as the version being different?

pkg/controllers/nodeclass/suite_test.go
```diff
@@ -889,6 +890,118 @@ var _ = Describe("NodeClassController", func() {
...
+			nodeClaimOne = ExpectExists(ctx, env.Client, nodeClaimOne)
+			nodeClaimTwo = ExpectExists(ctx, env.Client, nodeClaimTwo)
+
+			expectedHash := nodeClass.Hash()
+			Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHash, expectedHash))
+			Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHashVersion, v1beta1.EC2NodeClassHashVersion))
+			Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHash, expectedHash))
+			Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHashVersion, v1beta1.EC2NodeClassHashVersion))
+		})
+		It("should not update ec2nodeclass hash on all nodeclaims when the hash versions match", func() {
```
jonathan-innis, line 945:
```suggestion
		It("should not update ec2nodeclass hash on all nodeclaims when the hash versions matches the controller hash version", func() {
```

pkg/controllers/nodeclass/suite_test.go
```diff
@@ -889,6 +890,118 @@ var _ = Describe("NodeClassController", func() {
...
+			nodeClaim := coretest.NodeClaim(corev1beta1.NodeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						v1beta1.AnnotationEC2NodeClassHash:        "1234564654",
+						v1beta1.AnnotationEC2NodeClassHashVersion: "test-version",
```
jonathan-innis, line 950: Based on the testcase name, I would have expected this value to be equal to the `v1beta1.EC2NodeClassHashVersion` value

pkg/controllers/nodeclass/suite_test.go
```diff
@@ -889,6 +890,118 @@ var _ = Describe("NodeClassController", func() {
...
+
+			nodeClaim.Annotations[v1beta1.AnnotationEC2NodeClassHash] = "test-hash"
+			ExpectApplied(ctx, env.Client, nodeClass, nodeClaim)
+
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+
+			Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHash, "test-hash"))
+			Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHashVersion, v1beta1.EC2NodeClassHashVersion))
+		})
+		It("should not update ec2nodeclass hash on the nodeclaim if it's drifted", func() {
```
jonathan-innis, line 981:
```suggestion
		It("should not update ec2nodeclass hash on the nodeclaim if it's drifted and the hash version doesn't match the controller hash version", func() {
```

pkg/cloudprovider/suite_test.go
```diff
@@ -770,6 +774,20 @@ var _ = Describe("CloudProvider", func() {
 				Expect(err).NotTo(HaveOccurred())
 				Expect(isDrifted).To(BeEmpty())
 			})
+			It("should not return drifted if karpenter.sh/ec2nodeclass-hash-version annotation version does not match the EC2NodeClass on the NodeClaim", func() {
```
jonathan-innis, line 777:
```suggestion
			It("should not return drifted if the NodeClaim's karpenter.sh/ec2nodeclass-hash-version annotation does not match the EC2NodeClass's", func() {
```

pkg/cloudprovider/suite_test.go
```diff
@@ -770,6 +774,20 @@ var _ = Describe("CloudProvider", func() {
...
+				nodeClass.ObjectMeta.Annotations = map[string]string{
+					v1beta1.AnnotationEC2NodeClassHash:        "test-hash-1",
```
jonathan-innis, line 779: Same comment as the other PR: These values need to be different between the NodeClaim and the EC2NodeClass or you aren't really evaluating anything in your testing

pkg/controllers/nodeclass/suite_test.go
engedaam: Updated the test to better reflect the values

pkg/cloudprovider/suite_test.go
jonathan-innis: This still doesn't look like it was fixed. The hash values for both resources are still the same.

pkg/controllers/nodeclass/controller.go
```diff
@@ -222,6 +232,40 @@ func (c *Controller) resolveInstanceProfile(ctx context.Context, nodeClass *v1be
...
+			// Any NodeClaim that is already drifted will remain drifted if the karpenter.sh/ec2nodeclass-hash-version doesn't match
+			// Since the hashing mechanism has changed we will not be able to determine if the drifted status of the node has changed
+			if nc.StatusConditions().GetCondition(corev1beta1.Drifted) == nil {
+				nc.Annotations = lo.Assign(nc.Annotations, map[string]string{
+					v1beta1.AnnotationEC2NodeClassHash: nodeClass.Hash(),
+				})
+			}
+
+			if !equality.Semantic.DeepEqual(stored, nc) {
+				if err := c.kubeClient.Patch(ctx, &nc, client.MergeFrom(stored)); err != nil {
```
jonathan-innis, line 259: You need to do the same error handling here that you do in the NodePool PR; else, you are going to hit the same situation as before, where one NotFound error is really going to wreak havoc on this logic

pkg/controllers/nodeclass/suite_test.go
```diff
@@ -889,6 +890,125 @@ var _ = Describe("NodeClassController", func() {
...
+			})
+			ExpectApplied(ctx, env.Client, nodeClass, nodeClaim)
+
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+			nodeClass = ExpectExists(ctx, env.Client, nodeClass)
+			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+
+			expectedHash := nodeClass.Hash()
+
+			// Expect NodeClaims to have been updated to the original hash
```
jonathan-innis, line 978:
```suggestion
			// Expect NodeClaims not to have been updated to the NodePool hash
```

This comment seems wrong; I think you mean that you expect the NodeClaims not to get updated

pkg/controllers/nodeclass/suite_test.go
```diff
@@ -889,6 +890,125 @@ var _ = Describe("NodeClassController", func() {
...
+					},
+				},
+			})
+			nodeClaim.StatusConditions().MarkTrue(corev1beta1.Drifted)
+			ExpectApplied(ctx, env.Client, nodeClass, nodeClaim)
+
+			ExpectReconcileSucceeded(ctx, nodeClassController, client.ObjectKeyFromObject(nodeClass))
+			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+
+			// Expect NodeClaims to have been updated to the original hash
```
jonathan-innis, line 1008:
```suggestion
			// Expect NodeClaims not to have been updated to the NodePool hash
```

Same comment here. I think you mean that you expect them _not_ to be updated

pkg/cloudprovider/suite_test.go
engedaam: It was left out due rebase issue. Should be updated now

pkg/cloudprovider/suite_test.go
```diff
@@ -761,7 +765,35 @@ var _ = Describe("CloudProvider", func() {
 				Entry("SecurityGroup Drift", v1beta1.EC2NodeClass{Spec: v1beta1.EC2NodeClassSpec{SecurityGroupSelectorTerms: []v1beta1.SecurityGroupSelectorTerm{{Tags: map[string]string{"sg-key": "sg-value"}}}}}),
 			)
 			It("should not return drifted if karpenter.k8s.aws/nodeclass-hash annotation is not present on the NodeClaim", func() {
```
jonathan-innis, line 767:
```suggestion
			It("should not return drifted if karpenter.k8s.aws/ec2nodeclass-hash annotation is not present on the NodeClaim", func() {
```

nit: parity with other testcase titles?

test/suites/drift/suite_test.go
```diff
@@ -719,6 +719,55 @@ var _ = Describe("Drift", func() {
...
 		env.EventuallyExpectHealthyPodCount(selector, numPods)
 	})
+	It("should update the ec2nodeclass-hash annotation on the ec2nodeclass and nodeclaim when the ec2nodeclass's ec2nodeclass-hash-version annotation does not match the controller hash version", func() {
+		env.ExpectCreated(dep, nodeClass, nodePool)
+		env.EventuallyExpectHealthyPodCount(selector, numPods)
+		nodeClaim := env.EventuallyExpectCreatedNodeClaimCount("==", 1)[0]
+		nodeClass = env.ExpectExists(nodeClass).(*v1beta1.EC2NodeClass)
+		expectedHash := nodeClass.Hash()
+
+		By(fmt.Sprintf("expect nodeclass %s and nodeclaim %s to contain karpenter.k8s.aws/ec2nodeclass-hash and karpenter.k8s.aws/ec2nodeclass-hash-version annotations", nodeClass.Name, nodeClaim.Name))
```
jonathan-innis, line 729:
```suggestion
		By(fmt.Sprintf("expect nodeclass %s and nodeclaim %s to contain %s and %s annotations", nodeClass.Name, nodeClaim.Name, v1beta1.AnnotationEC2NodeClassHash, v1beta1.AnnotationEC2NodeClassHashVersion))
```

test/suites/drift/suite_test.go
```diff
@@ -719,6 +719,55 @@ var _ = Describe("Drift", func() {
...
+		Eventually(func(g Gomega) {
```
jonathan-innis, line 730: You aren't re-grabbing the NodeClass and NodeClaim within the Eventually block, so if you don't get the correct result on the first try, you'll just keep working with the old data and never succeed here

test/suites/drift/suite_test.go
```diff
@@ -719,6 +719,55 @@ var _ = Describe("Drift", func() {
...
+		nodeClaim.Annotations = lo.Assign(nodePool.Annotations, map[string]string{
+			v1beta1.AnnotationEC2NodeClassHash:        "test-hash-2",
+			v1beta1.AnnotationEC2NodeClassHashVersion: "test-hash-version-2",
+		})
+
+		// The nodeclaim will need to be updated first, as the hash controller will only be triggered on changes to the nodeclass
+		env.ExpectUpdated(nodeClaim, nodeClass)
+		expectedHash = nodeClass.Hash()
+
+		// Should expect the all nodeclaims will not be drifted and contain an updated `ec2nodeclass-hash` and `ec2nodeclass-hash-version` annotations
```
jonathan-innis, line 756:
```suggestion
		// Should expect all nodeclaims will not be drifted and contain an updated `ec2nodeclass-hash` and `ec2nodeclass-hash-version` annotation
```

test/suites/drift/suite_test.go
```diff
@@ -719,6 +719,55 @@ var _ = Describe("Drift", func() {
...
+		Eventually(func(g Gomega) {
+			nodeClass = env.ExpectExists(nodeClass).(*v1beta1.EC2NodeClass)
```
jonathan-innis, line 758: This ExpectExists won't be retried if it fails inside this block since it doesn't use the "g" Gomega context that you have declared, it uses the global context. Prefer `g.Expect(env.Client.Get(...)).To(Succeed())`

test/suites/drift/suite_test.go
```diff
@@ -719,6 +719,55 @@ var _ = Describe("Drift", func() {
...
+			nodeClaim = env.ExpectExists(nodeClaim).(*corev1beta1.NodeClaim)
+
+			g.Expect(nodeClass.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHash, expectedHash))
+			g.Expect(nodeClass.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHashVersion, v1beta1.EC2NodeClassHashVersion))
+			g.Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHash, expectedHash))
+			g.Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHashVersion, v1beta1.EC2NodeClassHashVersion))
+		}).WithTimeout(30 * time.Second).Should(Succeed())
+
+		notDriftedDuration := time.Minute
+		By(fmt.Sprintf("consistently expect nodeclaim %s not to be drifted for %s", nodeClaim.Name, notDriftedDuration))
```
jonathan-innis, line 768: Consider putting this `By` statement inside of the `Consistently` function itself

test/suites/drift/suite_test.go
```diff
@@ -719,6 +719,55 @@ var _ = Describe("Drift", func() {
...
+			nodeClass = env.ExpectExists(nodeClass).(*v1beta1.EC2NodeClass)
+			nodeClaim = env.ExpectExists(nodeClaim).(*corev1beta1.NodeClaim)
+
+			g.Expect(nodeClass.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHash, expectedHash))
+			g.Expect(nodeClass.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHashVersion, v1beta1.EC2NodeClassHashVersion))
+			g.Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHash, expectedHash))
+			g.Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.AnnotationEC2NodeClassHashVersion, v1beta1.EC2NodeClassHashVersion))
+		}).WithTimeout(30 * time.Second).Should(Succeed())
+
+		notDriftedDuration := time.Minute
```
jonathan-innis, line 767: If you take the suggestion below and don't put the `By` statement in this context, then you don't need to declare this as a separate variable

jonathan-innis (approved):
LGTM 🚀

PR #5779, opened 2024-03-04
feat: Add ability to select instance by performance mode

### LINKED ISSUE #4367: How does Karpenter deal with the new `m7i-flex` instance type?
Given that the new `m7i-flex` instance type is [now available](https://aws.amazon.com/about-aws/whats-new/2023/08/amazon-ec2-m7i-flex-m7i-instances/) and breaks the current pattern of burstable instances being in a non `m` family; how does Karpenter currently treat these instances and allow them to be filtered out? If this isn't a default pattern what plans are there to improve this?

My "two cents" is that the concept of burstable instances is very similar to the spot instance concept and should probably align with that.

NetanelK:
Fixes #4367 

**Description**
Added ability to select instance by its performance mode, either `flex`, `burstable`, or `standard`. 
By using the label `karpener.k8s.aws/instance-performance-mode`

Ran unit tests

- [ ] Yes, issue opened: # 
- [ ] No


pkg/providers/instancetype/types.go
```diff
@@ -41,12 +42,19 @@ import (
 )
 
 const (
-	MemoryAvailable = "memory.available"
-	NodeFSAvailable = "nodefs.available"
+	MemoryAvailable          = "memory.available"
+	NodeFSAvailable          = "nodefs.available"
+	PerformanceModeBurstable = "burstable"
```
jonathan-innis, line 47: What do you think? Should this be "burstable" or "burst"?

NetanelK: AWS calls those instances "burstable performance instances", so I kept it that way.
But it's up to you and other maintainers to decide.

jonathan-innis: Got it. If this is the nomenclature that EC2 uses, I'd rather not stray from it

pkg/apis/v1beta1/labels.go
```diff
@@ -114,6 +118,7 @@ var (
 	LabelInstanceAcceleratorName              = Group + "/instance-accelerator-name"
 	LabelInstanceAcceleratorManufacturer      = Group + "/instance-accelerator-manufacturer"
 	LabelInstanceAcceleratorCount             = Group + "/instance-accelerator-count"
+	LabelInstancePerformanceMode              = Group + "/instance-performance-mode"
```
jonathan-innis, line 121:
```suggestion
	LabelInstancePerformanceMode              = Group + "/instance-capability-flex"
```

Suggesting this label name in the original issue because of the way that EC2 describes their own naming convention here: https://docs.aws.amazon.com/ec2/latest/instancetypes/instance-type-names.html. Given that page as well, I'm not sure that it makes sense to couple the burstable instance types with the flex ones. Can we treat flex as its own thing here?

NetanelK: Personally I would prefer a value that can be fetched from the EC2 API, instead of relaying on the instance name structure.
Anyhow, it can be done fairly easy.

NetanelK: @jonathan-innis 

I tried to add a label per each instance capability (total 7) and encountered the following error:

`"NodeClaim.karpenter.sh \"spiritslow-94-ywrmgisn0j-p88dm\" is invalid: [spec.requirements: Too many: 31: must have at most 30 items, <nil>: Invalid value: \"null\": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]"`

The cause is [this line](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/apis/v1beta1/nodeclaim.go#L40) from kubernetes-sigs/karpenter, can we increase the limit of requirements?

PR #5093, opened 2023-11-14
feat: publish ICE events with involved object of EC2NodeClass

tasdikrahman:
Fixes https://github.com/aws/karpenter/issues/5004

This PR adds support for publishing ICE events along with the ec2nodeClass involved object. 

**How was this change tested?**

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates 
- [ ] Yes, issue opened: # 
- [x] No

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5785, opened 2024-03-05
[wip]feat: publish ICE events with involved object of NodeClaim

tasdikrahman:
Fixes https://github.com/aws/karpenter/issues/5004, follow up of PR https://github.com/aws/karpenter-provider-aws/pull/5093

This PR adds support for publishing ICE events along with the ec2nodeClass involved object. 

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/cache/events.go
+++ b/pkg/cache/events.go
@@ -16,17 +16,19 @@ package cache
 
 import (
 	"fmt"
+	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
 
 	v1 "k8s.io/api/core/v1"
 
 	"sigs.k8s.io/karpenter/pkg/events"
 )
 
-func UnavailableOfferingEvent(instanceType, availabilityZone, capacityType string) events.Event {
+func UnavailableOfferingEvent(nodeClaim *v1beta1.NodeClaim, instanceType, availabilityZone, capacityType string) events.Event {
 	return events.Event{
-		Type:         v1.EventTypeWarning,
-		Reason:       "UnavailableOffering",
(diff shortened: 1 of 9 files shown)


pkg/cache/events.go
```diff
@@ -0,0 +1,34 @@
...
+	"sigs.k8s.io/karpenter/pkg/events"
+)
+
+func UnavailableOfferingEvent(nodeClaim *v1beta1.NodeClaim, instanceType, availabilityZone, capacityType string) events.Event {
+	return events.Event{
+		InvolvedObject: nodeClaim,
+		Type:           v1.EventTypeWarning,
+		Reason:         "UnavailableOffering",
+		Message:        fmt.Sprintf(`UnavailableOffering for {"instanceType": %q, "availabilityZone": %q, "capacityType": %q}`, instanceType, availabilityZone, capacityType),
+		DedupeValues:   []string{instanceType, availabilityZone, capacityType},
```
jonathan-innis, line 32: We'd probably also want to rate limit this type of event generally. This is going to be really expensive to fire these events every time we get a spot insufficient capacity error, especially if they come in bursts, which I would generally expect them to.

jonathan-innis:
@tasdikrahman Can you remind me again why metrics wasn't sufficient for your use-case? I know that you wanted to do watches on the events and react to it with a controller, but I'm wondering if you can achieve a similar behavior with metrics here. The reason I'm calling this out is that I think that metrics are a better, global mechanism to handle this kind of thing now. We now have the [`karpenter_cloudprovider_instance_type_offering_available`](https://karpenter.sh/preview/reference/metrics/#karpenter_cloudprovider_instance_type_offering_available) metric which would allow you to see when instance type offerings are oscillating between available and unavailable.

engedaam:
@tasdikrahman Any updates here?

PR #5795, opened 2024-03-07
Add karpenter dashboards to helm-chart

### LINKED ISSUE #5791: Adding karpenter dashboards to helm chart
### Description

There are a few karpenter dashboards under getting-started (https://github.com/aws/karpenter-provider-aws/tree/main/website/content/en/docs/getting-started/getting-started-with-karpenter), it would be great if these would be bundled into the helm-chart, so deploying these to your cluster is as easy as setting values.yaml (f.e.: `grafana.dashboards.enabled: true`) to get latest updated charts with the last helm-chart being deployed.

Currently you have to copy the json file, add it to a ConfigMap and then either apply it, or use postRenderer in helm to add into the same release.

tvonhacht-apple:
Fixes #5791 

**Description**

Adding karpenter dashboards to helm chart, so we can improve them over time and rolling them out easily to users deploying karpenter via helm-chart.

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -29,6 +29,11 @@ serviceMonitor:
   additionalLabels: {}
   # -- Endpoint configuration for the ServiceMonitor.
   endpointConfig: {}
+grafanaDashboards:
+  # -- Specifies whether grafana dashboards should be created.
+  enabled: false
+  # -- Additional labels for the grafana dashboards.
+  additionalLabels: {}
 # -- Number of replicas.
 replicas: 2
 # -- The number of old ReplicaSets to retain to allow rollback.
--- a/charts/karpenter/grafana-dashboards/karpenter-capacity-dashboard.json
+++ b/charts/karpenter/grafana-dashboards/karpenter-capacity-dashboard.json
@@ -1,0 +1,1440 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "graphTooltip": 2,
+  "id": 6,
+  "links": [],
+  "liveNow": true,
+  "panels": [
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${datasource}"
+      },
+      "description": "",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "axisLabel": "",
+            "axisPlacement": "auto",
+            "barAlignment": 0,
+            "drawStyle": "line",
+            "fillOpacity": 10,
+            "gradientMode": "none",
+            "hideFrom": {
+              "legend": false,
+              "tooltip": false,
+              "viz": false
+            },
+            "lineInterpolation": "linear",
+            "lineWidth": 1,
+            "pointSize": 5,
+            "scaleDistribution": {
+              "type": "linear"
+            },
+            "showPoints": "auto",
+            "spanNulls": false,
+            "stacking": {
+              "group": "A",
+              "mode": "none"
+            },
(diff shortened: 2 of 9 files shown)


jonathan-innis:
If we have this change in, does this [change our docs around deploying these dashboards](https://karpenter.sh/docs/getting-started/getting-started-with-karpenter/#monitoring-with-grafana-optional)?

charts/karpenter/templates/configmap-grafana-dashboards-karpenter-capacity-dashboard.yaml
```diff
@@ -0,0 +1,1459 @@
...
+  {{- with .Values.grafana.dashboards.additionalLabels }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with .Values.additionalAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+data:
+  # https://github.com/aws/karpenter-provider-aws/blob/main/website/content/en/docs/getting-started/getting-started-with-karpenter/karpenter-capacity-dashboard.json
+  karpenter-capacity-dashboard.json: |
```
jonathan-innis, line 18: Can we consider storing the JSON in a separate JSON file and [then having Helm inject the file into the ConfigMap](https://helm.sh/docs/chart_template_guide/accessing_files/#basic-example)? IMO, that'll make each of these files easier to maintain.

charts/karpenter/values.yaml
```diff
@@ -29,6 +29,12 @@ serviceMonitor:
   additionalLabels: {}
   # -- Endpoint configuration for the ServiceMonitor.
   endpointConfig: {}
+grafana:
```
jonathan-innis, line 32: Do we expect to have other grafana-style configurations that we should put in this stanza or do we just collapse it down like we did to the serviceMonitor to `grafanaDashboards`? I'd be interested to see prior-art for other helm charts that also deploy Grafana dashboards. Maybe @stevehipwell has some insight here too?

charts/karpenter/templates/configmap-grafana-dashboards-karpenter-controllers-allocation.yaml
```diff
@@ -0,0 +1,352 @@
...
+  {{- with .Values.grafana.dashboards.additionalLabels }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with .Values.additionalAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+data:
+  # https://github.com/aws/karpenter-provider-aws/blob/main/website/content/en/docs/getting-started/getting-started-with-karpenter/karpenter-controllers-allocation.json
+  karpenter-controllers-allocation.json: |
```
jonathan-innis, line 18: I'm less familiar with the Grafana configuration, but do these need to be separate ConfigMaps or can they all be in the same ConfigMap but under different data keys?

stevehipwell: It's probably best to have them as separate config maps, due to the maximum object size and to reduce churn.

charts/karpenter/templates/configmap-grafana-dashboards-karpenter-capacity-dashboard.yaml
stevehipwell: This would generally be done via a `range` so a single template could handle all of the dashboards in a folder but keep them as separate ConfigMap resources.

charts/karpenter/values.yaml
stevehipwell: I think in the context of Helm you want to restrict the actual dashboard implementation to be as closely scoped as possible; think hidden variables for things such as namespace which are dynamic for Helm but fixed once Karpenter has been deployed. I'm not sure if any of these require additional "Grafana" configuration but I'd be tempted to use `grafana.dashboards` in case there ever was.

charts/karpenter/templates/configmap-grafana-dashboards-karpenter-capacity-dashboard.yaml
tvonhacht-apple: ok yeah, I should have thought about that! will make change

charts/karpenter/values.yaml
tvonhacht-apple: I can do either way, I just started it so we can have this discussion. please let me know what you both decide on

universam1:
@tvonhacht-apple I see this PR is closed, I'd love to see it to land. Any reason for?

PR #5808, opened 2024-03-08
docs: fix topology zone list value quote issue in docs

BingchenLi:
Fixes #N/A 

**Description**
This change is to fix the quote mistake in topology zone value list in docs.
Different zone values need to be separate strings, thus should be quoted separately in the list.

This only includes the docs change. 

- [ ] Yes, issue opened: # 
- [] No


jonathan-innis (approved):
LGTM 🚀 

Thanks for catching this!

PR #5834, opened 2024-03-11
chore: fix the clean-up for instanceProfile to catch InstanceProfileName

engedaam:
Fixes #N/A 

**Description**
- The clean-up script was not grabbing the correct tag to get the instance profile name. Fixed the tags and switched to using data from the `ListInstanceProfileTags` API call
- Also changed to using the instance profile tag for region to properly identify instance profiles to clean-up

- Manually tested

- [ ] Yes, issue opened: # 
- [x] No


test/hack/resource/pkg/resourcetypes/instanceprofile.go
```diff
@@ -63,10 +58,16 @@ func (ip *InstanceProfile) GetExpired(ctx context.Context, expirationTime time.T
...
+		clusterName, foundClusterName := lo.Find(profiles.Tags, func(tag iamtypes.Tag) bool {
+			return lo.FromPtr(tag.Key) == karpenterTestingTag
 		})
-		if found && slices.Contains(excludedClusters, lo.FromPtr(clusterName.Value)) {
+		// Checking to make sure we are only list resources in the given region
+		region, foundRegion := lo.Find(profiles.Tags, func(tag iamtypes.Tag) bool {
+			return lo.FromPtr(tag.Key) == resourceRegionTag
+		})
+
+		if foundClusterName && slices.Contains(excludedClusters, lo.FromPtr(clusterName.Value)) ||
```
jonathan-innis, line 69: So, if we didn't find the region, the region shouldn't be equal to the region (it'll be `""`). The same is true for the excluded clusters. I wonder if we even need the found check here.

test/hack/resource/pkg/resourcetypes/resourcetypes.go
```diff
@@ -27,6 +27,7 @@ const (
 	karpenterTestingTag        = "testing/cluster"
 	k8sClusterTag              = "cluster.k8s.amazonaws.com/name"
 	githubRunURLTag            = "github.com/run-url"
+	resourceRegionTag          = "topology.kubernetes.io/region"
```
jonathan-innis, line 30: This an upstream label. Should we just use that constant rather than copying the tag into our own environment?

test/hack/resource/pkg/resourcetypes/instanceprofile.go
engedaam: Yeah, that's a fair point. Removed them

jonathan-innis (dismissed):
LGTM 🚀

jonathan-innis (approved):
LGTM 🚀

Issue #5306, opened 2023-12-12
Add ArtifactHub Karpenter Provovider AWS repo


stevehipwell:
### Description

**What problem are you trying to solve?**
I'd like there to be an official Karpenter Provider AWS repo in [ArtifactHub](https://artifacthub.io/).

@bwagner5 added a [repo](https://artifacthub.io/packages/helm/karpenter/karpenter) for the legacy release process which wasn't adopted (but still could be). I've added a repo for the new release process which can either be adopted (see ArtifactHub docs for [ownership claim](https://artifacthub.io/docs/topics/repositories/#ownership-claim)) or I can delete when you're ready to create a repo if you want to use a new name (I used `aws-karpenter` but am aware that `karpenter-provider-aws` might have been a better name). You could adopt first and then delete & replace afterwards.

I think you also need to verify as a publisher ([docs](https://artifacthub.io/docs/topics/repositories/#verified-publisher)) and set the chart as official ([docs](https://artifacthub.io/docs/topics/repositories/#official-status)).

I'd be happy to open a PR to support adding and updating the metadata file in the automation.

**How important is this feature to you?**
This should be important to the project for discoverability. I've fixed the automation tooling issues by adding a repo to support this but it's not official so won't help discoverability in the same way that an official repo would.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jonathan-innis:
We'd be open to surfacing Karpenter here. What's the maintenance burden to maintain the config on ArtifactHub? Mainly just making sure the work to update this release-to-release isn't too much to have to reason about for the maintainers.

stevehipwell:
@jonathan-innis the maintenance cost is either trivial (sync an almost static file to the OCI registry with `oras` on each release) or a manual one off (I'd advise against this).

The initial one off cost is to set up an ArtifactHub organisation if you don't already have one, then setup the repo, and finally fill out the form to mark the chart as official after it's been verified.

I'm happy to open a PR to cover the maintenance and/or jump on a call to discuss and walk you through the process.

FernandoMiguel:
here's a tiny example of how argocd updates artifacthub
[...]

PR #5843, opened 2024-03-12
chore: Added ArtifactHub Helm metadata

stevehipwell:
Fixes #N/A 

**Description**
This PR adds Artifact Hub metadata to the Helm charts which will flag them as being verified and will support removing the invalid releases from being processed (this could/should also be handled manually or with a cleanup rule?). It also adds the placeholder to set the email address of the person claiming the repository for AWS (see #5306).

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter-crd/artifacthub-repo.yaml
+++ b/charts/karpenter-crd/artifacthub-repo.yaml
@@ -1,0 +1,7 @@
+repositoryID: fda7ffc4-4672-4218-8264-321ec3b4e3cc
+owners: []
+#   - name: awsadmin1
+#     email: artifacthub1@aws.com
+ignore:
+  - name: karpenter-crd
+    version: (?:^\d+$)|(?:^v?0\.0\.0)|(?:^v?\d+\-)
--- a/charts/karpenter/artifacthub-repo.yaml
+++ b/charts/karpenter/artifacthub-repo.yaml
@@ -1,0 +1,7 @@
+repositoryID: 356cb63f-9ee3-4956-9c20-003e416715c7
+owners: []
+#   - name: awsadmin1
+#     email: artifacthub1@aws.com
+ignore:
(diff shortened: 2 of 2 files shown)


stevehipwell:
CC @jonathan-innis

charts/karpenter-crd/artifacthub-repo.yaml
```diff
@@ -0,0 +1,7 @@
+repositoryID: fda7ffc4-4672-4218-8264-321ec3b4e3cc
+owners: []
+#   - name: awsadmin1
+#     email: artifacthub1@aws.com
+ignore:
+  - name: karpenter-crd
+    version: (?:^\d+$)|(?:^v?0\.0\.0)|(?:^v?\d+\-)
```
jonathan-innis, line 7: Can we add a comment here to indicate why we want to ignore ArtifactHub indexing these versions? It's not immediately apparent to me why we are ignoring these versions just reading through it.

stevehipwell: Am I right in thinking that this repo should only ever get stable versions now? If so then this is to exclude old non-stable versions. This could be solved without the regex by just removing all non-stable versions.

jonathan-innis: Got it. So this was produced by looking through all of the tags in our repository for the chart? What about the tags for chart signatures? Do we also need to ignore these? I believe that they start with `sha256-`

stevehipwell: Only the Helm chart OCI media types are processed so signatures are fine, it's just the old non-production charts which need ignoring (the `yyyyddmm` ones actually error).

jonathan-innis (approved):
LGTM 🚀

PR #5855, opened 2024-03-13
chore: Add event on nodeClaim for instance type filtering

jigisha620:
Fixes #N/A 

**Description**
We filter instance types at various levels but currently don't log it which makes it difficult to debug in case of any errors. In this PR we instead added an event that would describe the filtering that we do.

`/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/cloudprovider/events/events.go
+++ b/pkg/cloudprovider/events/events.go
@@ -15,7 +15,11 @@ limitations under the License.
 package events
 
 import (
+	"fmt"
+
 	v1 "k8s.io/api/core/v1"
+
+	"github.com/aws/karpenter-provider-aws/pkg/utils"
 
 	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
 	"sigs.k8s.io/karpenter/pkg/events"
@@ -36,5 +40,15 @@ func NodeClaimFailedToResolveNodeClass(nodeClaim *v1beta1.NodeClaim) events.Even
 		Type:           v1.EventTypeWarning,
 		Message:        "Failed resolving NodeClass",
 		DedupeValues:   []string{string(nodeClaim.UID)},
+	}
+}
+
+func FilteredInstanceTypes(nodeClaim *v1beta1.NodeClaim, filteredITsRequirement int, filteredExoticITs int, filteredExpensiveSpot int, filteredITs []string) events.Event {
+	return events.Event{
+		InvolvedObject: nodeClaim,
+		Type:           v1.EventTypeNormal,
+		Message: fmt.Sprintf("Filtered %d instance types on requirements, filtered %d instance types from exotic instance types, filtered %d instance types with spot offerings more expensive than cheapest on-demand. Remaining instance types:: %s",
+			filteredITsRequirement, filteredExoticITs, filteredExpensiveSpot, utils.PrettySlice(filteredITs, 5)),
+		DedupeValues: []string{string(nodeClaim.UID)},
 	}
 }
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -22,6 +22,10 @@ import (
 	"sort"
 	"strings"
 
+	"sigs.k8s.io/karpenter/pkg/events"
+
+	cloudproviderevents "github.com/aws/karpenter-provider-aws/pkg/cloudprovider/events"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -66,10 +70,11 @@ type Provider struct {
 	subnetProvider         *subnet.Provider
 	launchTemplateProvider *launchtemplate.Provider
 	ec2Batcher             *batcher.EC2API
+	recorder               events.Recorder
 }
 
 func NewProvider(ctx context.Context, region string, ec2api ec2iface.EC2API, unavailableOfferings *cache.UnavailableOfferings,
-	instanceTypeProvider *instancetype.Provider, subnetProvider *subnet.Provider, launchTemplateProvider *launchtemplate.Provider) *Provider {
+	instanceTypeProvider *instancetype.Provider, subnetProvider *subnet.Provider, launchTemplateProvider *launchtemplate.Provider, recorder events.Recorder) *Provider {
 	return &Provider{
 		region:                 region,
 		ec2api:                 ec2api,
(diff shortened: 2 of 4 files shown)


pkg/cloudprovider/events/events.go
```diff
@@ -38,3 +42,13 @@ func NodeClaimFailedToResolveNodeClass(nodeClaim *v1beta1.NodeClaim) events.Even
 		DedupeValues:   []string{string(nodeClaim.UID)},
 	}
 }
+
+func FilteredInstanceTypes(nodeClaim *v1beta1.NodeClaim, resolvedNodeClassITs []string, filteredITs []string) events.Event {
+	return events.Event{
+		InvolvedObject: nodeClaim,
+		Type:           v1.EventTypeNormal,
+		Message: fmt.Sprintf("Resolved instance types from nodeclass: %s, filtered instance types after removing exotic types and spot instance types costlier than cheapest on-demand: %s", utils.PrettySlice(resolvedNodeClassITs, 5),
```
jonathan-innis, line 50:
```suggestion
		Message: fmt.Sprintf("Filtered instance types after removing exotic types and spot instance types costlier than cheapest on-demand: %s", utils.PrettySlice(resolvedNodeClassITs, 5),
```

I'm not sure that you need the NodeClass bit

jonathan-innis, line 50: Is it possible for us to create a message that shows how many instance types we filtered at each stage and then the last part of the message prints the remaining instance types using the pretty slice?

pkg/providers/instance/instance.go
```diff
@@ -78,15 +83,18 @@ func NewProvider(ctx context.Context, region string, ec2api ec2iface.EC2API, una
 		subnetProvider:         subnetProvider,
 		launchTemplateProvider: launchTemplateProvider,
 		ec2Batcher:             batcher.EC2(ctx, ec2api),
+		recorder:               recorder,
 	}
 }
 
 func (p *Provider) Create(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, nodeClaim *corev1beta1.NodeClaim, instanceTypes []*cloudprovider.InstanceType) (*Instance, error) {
 	schedulingRequirements := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...)
+	resolvedNodeClassITs := instanceTypes
```
jonathan-innis, line 92: I'm not 100% sure, but since this is a slice, I think this just stores the pointer to the front of instanceTypes, so doing something like this doesn't actually copy the slice like I think you intend

jonathan-innis: Further, I'm not sure if you need to copy the slice here. How were you thinking this would help the eventing?

jigisha620: Hmmm, well the reason I tried to store that is because we are not publishing an event until we filter the exotic types. But we update the instance types [here](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/providers/instance/instance.go#L88) and if I don't store it before that then we will not get the previously filtered instance types and like I mentioned earlier, the ITs passed to this function have also gone through one round of [filtering](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/cloudprovider/cloudprovider.go#L247).

pkg/cloudprovider/events/events.go
jigisha620: @jonathan-innis Do you mean something like - `x instances were filtered/removed at a stage, y instances were filtered/removed at b stage, remaining instance types after filtering - u, v, w and 5 others`

jigisha620: Even before filtering the exotic types, we filter instanceTypes [here](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/cloudprovider/cloudprovider.go#L247). I believe we would want to have that as part of the event as well?

jigisha620: Also, just for reference, with the current changes the event looks like this - `Resolved instance types from nodeclass: r6in.metal, c6i.32xlarge, m6id.32xlarge, i4i.metal, m6in.32xlarge and 23 other(s), filtered instance types after removing exotic types and spot instance types costlier than cheapest on-demand: c6i.32xlarge, m6id.32xlarge, m6in.32xlarge, m6i.32xlarge, r6id.32xlarge and 9 other(s)`

jonathan-innis: Right, my point was that we don't do that filtering on the "nodeclass", it's done on the "nodeclaim". And the name of the NodeClaim is already included in the event since the involved object here is the NodeClaim

pkg/providers/instance/instance.go
jonathan-innis: Right, but I'm not sure that you actually need the previously filtered instance types, I think you just need the result at the end and you also need to capture the number of instance types that are filtered at each round

pkg/cloudprovider/events/events.go
```diff
@@ -38,3 +42,13 @@ func NodeClaimFailedToResolveNodeClass(nodeClaim *v1beta1.NodeClaim) events.Even
 		DedupeValues:   []string{string(nodeClaim.UID)},
 	}
 }
+
+func FilteredInstanceTypes(nodeClaim *v1beta1.NodeClaim, filteredITsRequirement int, filteredExoticITs int, filteredExpensiveSpot int, filteredITs []string) events.Event {
+	return events.Event{
+		InvolvedObject: nodeClaim,
+		Type:           v1.EventTypeNormal,
+		Message: fmt.Sprintf("Filtered %d instance types on requirements, filtered %d instance types from exotic instance types, filtered %d instance types with spot offerings more expensive than cheapest on-demand. Remaining instance types:: %s",
+			filteredITsRequirement, filteredExoticITs, filteredExpensiveSpot, utils.PrettySlice(filteredITs, 5)),
```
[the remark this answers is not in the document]
jonathan-innis: +1 to this. If we can build this so that it only adds the "filtered" statement when it actually filters something out, it's going to make this statement a lot more concise

pkg/providers/instance/instance.go
```diff
@@ -78,15 +83,19 @@ func NewProvider(ctx context.Context, region string, ec2api ec2iface.EC2API, una
 		subnetProvider:         subnetProvider,
 		launchTemplateProvider: launchTemplateProvider,
 		ec2Batcher:             batcher.EC2(ctx, ec2api),
+		recorder:               recorder,
 	}
 }
 
 func (p *Provider) Create(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, nodeClaim *corev1beta1.NodeClaim, instanceTypes []*cloudprovider.InstanceType) (*Instance, error) {
+	filteredITsRequirement := len(instanceTypes)
```
jonathan-innis, line 91: I feel like we still want to understand the instance types that were filtered out since they didn't match the requirements, or the ones that weren't large enough with the `resource.Fits` calculation, or the ones that didn't have any compatible offerings that were available when we went to go resolve the instance types. Are there ways that we could break up the `resolveInstanceTypes` function in a way that would also allow us to understand if we filtered anything out there. Otherwise, we are flying a little blind around why we might have filtered out an instance type when we went to first resolve them

jonathan-innis: This number is also wrong, right? This isn't the number of instance types that were filtered out after requirements, this is the number of instance types that work after the requirements has filtered them out?

jonathan-innis: Musing: Is there one function that we could build to combine all of the filtering logic into one place so that capturing all of this information and creating an event is much easier and is self-contained to just that function?

pkg/cloudprovider/events/events.go
```diff
@@ -38,3 +42,13 @@ func NodeClaimFailedToResolveNodeClass(nodeClaim *v1beta1.NodeClaim) events.Even
 		DedupeValues:   []string{string(nodeClaim.UID)},
 	}
 }
+
+func FilteredInstanceTypes(nodeClaim *v1beta1.NodeClaim, filteredITsRequirement int, filteredExoticITs int, filteredExpensiveSpot int, filteredITs []string) events.Event {
```
jonathan-innis, line 46: Musing: Consider storing all of this information together in a struct rather than making this info into separate arguments to the function and stored in different variables.

PR #5877, opened 2024-03-15
docs: pin amis for getting started guides

njtran:
Fixes #N/A 

**Description**
This changes the recommended ec2nodeclass in the getting started guides to use pinned AMIs. 

make website

- [ ] Yes, issue opened: # 
- [x] No


website/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/aws/karpenter-provider-aws/website
 
-go 1.22
```
jmdeal: So I stopped running into this and don't think we should pin if possible. Have you reinstalled go? If we do pin we should increase the pin to 1.22.1, IIRC 1.22.0 has some CVEs related to MIME parsing.

website/content/en/docs/getting-started/getting-started-with-karpenter/scripts/step12-add-nodepool.sh
```diff
@@ -45,4 +45,8 @@ spec:
   securityGroupSelectorTerms:
     - tags:
         karpenter.sh/discovery: "${CLUSTER_NAME}" # replace with your cluster name
+  amiSelectorTerms:
+    - id: "${ARM_AMI_ID}"
+    - id: "${AMD_AMI_ID}"
+#   - name: amazon-eks-node-"${K8S_VERSION}"-* # <- automatically upgrade when a new AL2 EKS Optimized AMI is released. This is unsafe for production workloads. Validate AMIs in lower environments before deploying them to production.
```
jmdeal, line 51: I don't think the double quotes should be inside the name, no?
```suggestion
#   - name: "amazon-eks-node-${K8S_VERSION}-*" # <- automatically upgrade when a new AL2 EKS Optimized AMI is released. This is unsafe for production workloads. Validate AMIs in lower environments before deploying them to production.
```

website/content/en/docs/getting-started/getting-started-with-karpenter/scripts/step01-config.sh
```diff
@@ -3,3 +3,5 @@ export CLUSTER_NAME="${USER}-karpenter-demo"
 export AWS_DEFAULT_REGION="us-west-2"
 export AWS_ACCOUNT_ID="$(aws sts get-caller-identity --query Account --output text)"
 export TEMPOUT="$(mktemp)"
+export ARM_AMI_ID="$(aws ssm get-parameter --name /aws/service/eks/optimized-ami/${K8S_VERSION}/amazon-linux-2-arm64/recommended/image_id --query Parameter.Value --output text)"
```
jmdeal, line 6: Should we also include the GPU AMI?

njtran: I decided not to, since currently there's no requirements field in the ami from the GPU that would let us know which of these AMIs to choose. WDYT?

jmdeal: Ah I see what you mean, maybe it could be included in a similar way to name based selection, commented by default?

jmdeal (approved):
LGTM 🚀

PR #5894, opened 2024-03-18
fix:  create tag script  error (InvalidID) when multiple sunbets for eks nodegroup

wttat:
Fixes #N/A 

**Description**

The output of the subcommand should not be wrapped in `$()`. This means that the output of the subcommand will be treated as multiple separate arguments. Therefore, if the subcommand outputs multiple subnet IDs separated by spaces, they will be treated as separate arguments passed to `--resources`. This resolves the previous "InvalidID" error if we have multiple subnets for eks nodegroup.


engedaam:
Thank you for this PR, would be able to make this changes in all version of the docs including preview and docs?

wttat:
> Thank you for this PR, would be able to make this changes in all version of the docs including `preview` and `docs`?

@engedaam  Added for preview and docs

engedaam:
@wttat Can we also include v0.34, and v0.36? Those are the only version of the docs left

wttat:
https://github.com/aws/karpenter-provider-aws/pull/6073
changed to a new pr

PR #5917, opened 2024-03-22
test: Add unit test to validate the instancetype cache

engedaam:
Fixes #N/A 

**Description**
Add unit test to validate the instancetype cache will give different set of instance type based on input
The tested value currently are 
- kubelet.kubeReserved
- kubelet.systemReserved
- kubelet.evictionHard
- kubelet.evictionSoft
- kubelet.maxPods
- nodeClass.instanceStorePolicy
- nodeClass.amiFamily
- nodeClass.blockDeviceMapping.rootVolume
- nodeClass.blockDeviceMapping.volumeSize
- nodeClass.blockDeviceMapping.deviceName

- N/A

- [ ] Yes, issue opened: # 
- [x] No


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/aws/karpenter-provider-aws
 go 1.22
 
 require (
+	dario.cat/mergo v1.0.0
```
jonathan-innis, line 6: What's this new dep and why do we need it?

pkg/providers/instancetype/suite_test.go
```diff
@@ -1716,6 +1718,159 @@ var _ = Describe("InstanceTypes", func() {
...
+			// kubelet.evictionSoft
+			// kubelet.maxPods
+			nodePool.Spec.Template.Spec.Kubelet = &corev1beta1.KubeletConfiguration{
+				KubeReserved:   v1.ResourceList{v1.ResourceCPU: resource.MustParse("1")},
+				SystemReserved: v1.ResourceList{v1.ResourceCPU: resource.MustParse("1")},
+				EvictionHard:   map[string]string{"memory.available": "5%"},
+				EvictionSoft:   map[string]string{"nodefs.available": "10%"},
+				MaxPods:        aws.Int32(10),
+			}
+			kubeletChanges := []*corev1beta1.KubeletConfiguration{
```
jonathan-innis, line 1736: Does it make sense to make this a `DescribeTable` so each test is sectioned out rather than a single test that iterates through each thing?

pkg/providers/instancetype/suite_test.go
```diff
@@ -1716,6 +1718,159 @@ var _ = Describe("InstanceTypes", func() {
...
+				instanetypes, err := awsEnv.InstanceTypesProvider.List(ctx, nodePool.Spec.Template.Spec.Kubelet, nodeClass)
+				Expect(err).To(BeNil())
+				sort.Slice(instanetypes, func(x int, y int) bool {
+					return instanetypes[x].Name < instanetypes[y].Name
+				})
+				instanceTypeResult = append(instanceTypeResult, instanetypes)
+			}
+
+			// We expect that the updated kubelet configuration will result in a subset of the full set instance types thats offered
+			// The goal is to make sure that we don't have a cache crash and have a fewer set of instancesTypes then expected
```
jonathan-innis, line 1769: What is a cache crash? Not sure I understand this comment

pkg/providers/instancetype/suite_test.go
```diff
@@ -1716,6 +1718,159 @@ var _ = Describe("InstanceTypes", func() {
...
+			// nodeClass.blockDeviceMapping.volumeSize
+			// nodeClass.blockDeviceMapping.deviceName
+			nodeClass.Spec.BlockDeviceMappings = []*v1beta1.BlockDeviceMapping{
+				{
+					DeviceName: lo.ToPtr("/dev/xvda"),
+					EBS:        &v1beta1.BlockDevice{VolumeSize: resource.NewScaledQuantity(10, resource.Giga)},
+					RootVolume: false,
+				},
+			}
+			nodeClassChanges := []*v1beta1.EC2NodeClass{
```
jonathan-innis, line 1801: Same comment as above: Should this be a `DescribeTable`?

go.mod
engedaam: Got that confused with `github.com/imdario/mergo`

pkg/providers/instancetype/suite_test.go
engedaam: I opted to not make it a DescribeTable call to guarantee that the `awsEnv` cache for instanceTypes is not cleared out in between test runs

engedaam: Realized I was BeforeEach` is not run on `DescribeTable` https://github.com/onsi/ginkgo/issues/378

engedaam: We will need to generate all the different configuration and compare the list of instancetypes that are contained in the cache, so for that reason they will need to be in the same `IT` block.

pkg/providers/instancetype/suite_test.go
```diff
@@ -1716,6 +1718,159 @@ var _ = Describe("InstanceTypes", func() {
...
+
+			sorted := nodePool.DeepCopy()
+			for _, change := range kubeletChanges {
+				nodePool = sorted.DeepCopy()
+				Expect(mergo.Merge(nodePool.Spec.Template.Spec.Kubelet, change, mergo.WithOverride, mergo.WithSliceDeepCopy)).To(BeNil())
+				// Calling the provider and storing the instance type list to the instancetype provider cache
+				_, err := awsEnv.InstanceTypesProvider.List(ctx, nodePool.Spec.Template.Spec.Kubelet, nodeClass)
+				Expect(err).To(BeNil())
+				// We are making sure to pull from the cache
+				instanetypes, err := awsEnv.InstanceTypesProvider.List(ctx, nodePool.Spec.Template.Spec.Kubelet, nodeClass)
```
njtran, line 1760:
```suggestion
				instancetypes, err := awsEnv.InstanceTypesProvider.List(ctx, nodePool.Spec.Template.Spec.Kubelet, nodeClass)
```

njtran: How does this also make sure we're pulling from the cache? Why do we need to be pulling again here instead of just relying on what's in the output at L1757?

pkg/providers/instancetype/suite_test.go
```diff
@@ -1716,6 +1718,159 @@ var _ = Describe("InstanceTypes", func() {
...
+				Expect(err).To(BeNil())
+				sort.Slice(instanetypes, func(x int, y int) bool {
+					return instanetypes[x].Name < instanetypes[y].Name
```
njtran, line 1763:
```suggestion
				sort.Slice(instancetypes, func(x int, y int) bool {
					return instancetypes[x].Name < instancetypes[y].Name
```

pkg/providers/instancetype/suite_test.go
```diff
@@ -1716,6 +1718,159 @@ var _ = Describe("InstanceTypes", func() {
...
+				})
+				instanceTypeResult = append(instanceTypeResult, instanetypes)
+			}
+
+			// We expect that the updated kubelet configuration will result in a subset of the full set instance types thats offered
+			// The goal is to make sure that we don't have different configuration collapsing to the same cache key and have a fewer set of instancesTypes then expected
```
njtran, line 1769: Should it be a subset? Or should the allocatables be different? Either way, I assume that the deep equals returning false is the right check here so it generally applies to all instance types.

pkg/providers/instancetype/suite_test.go
```diff
@@ -1716,6 +1718,159 @@ var _ = Describe("InstanceTypes", func() {
...
+					if x == y {
+						continue
+					}
+					key := reflect.DeepEqual(instanceTypeResult[x], instanceTypeResult[y])
+					if key {
+						fmt.Printf("failed x: %d, y: %d", x, y)
+					}
+					Expect(reflect.DeepEqual(instanceTypeResult[x], instanceTypeResult[y])).To(BeFalse())
+				}
+			}
```
njtran, line 1785: is this something that can be done with `lo.Uniq()`? Or does that not do DeepEquals?

pkg/providers/instancetype/suite_test.go
njtran: Seems to me like we can make this a table. Each iteration of the table is a different application of the kubeletChanges. It would change the For loop on L1753

njtran (commented):
Is there some sort of reflection we can do to make sure that each of the values that are added into the InstanceTypesProvider.List() has a table entry? It seems like this does the job, but there's no test that would fail if someone just finds + replaces all the instances of the List() function definition to add a new parameter.

pkg/providers/instancetype/suite_test.go
engedaam: We want to make sure that are getting values stored in the cache as opposed to the cloudprovider

pkg/providers/instancetype/suite_test.go
engedaam: Yeah, but then we need to make sure that each set of instance types is compared against each other in the For loop on L1773.

pkg/providers/instancetype/suite_test.go
engedaam: `lo.Uniq()` needs a type that's comparable and []InstanceType is not a really comparable type so we need to do a `DeepEquals`

pkg/providers/instancetype/suite_test.go
jonathan-innis: Since it's non-obvious why we did this in the first place, should we consider adding a comment that explains why we are testing these things in a single It() block so that it's clearer to the reader?

pkg/providers/instancetype/suite_test.go
```diff
@@ -1716,8 +1718,147 @@ var _ = Describe("InstanceTypes", func() {
...
+			// nodeClass.blockDeviceMapping.deviceName
+			nodeClass.Spec.BlockDeviceMappings = []*v1beta1.BlockDeviceMapping{
+				{
+					DeviceName: lo.ToPtr("/dev/xvda"),
+					EBS:        &v1beta1.BlockDevice{VolumeSize: resource.NewScaledQuantity(10, resource.Giga)},
+					RootVolume: false,
+				},
+			}
+			nodeClassChanges := []*v1beta1.EC2NodeClass{
+				{},
```
jonathan-innis, line 1788: optional nit: If I was reading through this code, I think I'd miss this. What do you think about a comment here to describe what the empty update means.

pkg/providers/instancetype/suite_test.go
```diff
@@ -1716,8 +1718,147 @@ var _ = Describe("InstanceTypes", func() {
...
+			// kubelet.maxPods
+			nodePool.Spec.Template.Spec.Kubelet = &corev1beta1.KubeletConfiguration{
+				KubeReserved:   map[string]string{string(v1.ResourceCPU): "1"},
+				SystemReserved: map[string]string{string(v1.ResourceCPU): "1"},
+				EvictionHard:   map[string]string{"memory.available": "5%"},
+				EvictionSoft:   map[string]string{"nodefs.available": "10%"},
+				MaxPods:        aws.Int32(10),
+			}
+			kubeletChanges := []*corev1beta1.KubeletConfiguration{
+				{},
```
jonathan-innis, line 1738: optional nit: If I was reading through this code, I think I'd miss this. What do you think about a comment here to describe what the empty update means.

jonathan-innis (approved):
LGTM 🚀

Issue #5934, opened 2024-03-26
Karpenter 0.35.2 Grafana dashboards do not work properly


youwalther65:
### Description

**Observed Behavior**:

Karpenter Grafana dashboard "Karpenter Capacity" contains metric "karpenter_deprovisioning_actions_performed" which is not available anymore according to [Metrics](https://karpenter.sh/v0.35/reference/metrics/).

Karpenter Grafana dashboard "Karpenter Capacity" uses controller-runtime based metrics "controller_runtime_reconcile_time_seconds_bucket" and "controller_runtime_reconcile_total" which are used by other K8s components like "aws-load-balancer-controller", so a job label to filter for job="karpenter" is necessary to avoid ingesting controllers of these K8s components as well.

**Expected Behavior**:
Usage of metrics "karpenter_disruption_nodes_disrupted_total" and "karpenter_disruption_eligible_nodes" in Capacity dashboard will provide the needed insides into Karpenter action, consolidation_type and method.

Using job="karpenter" in Performance dashboard will show the correct Karpenter related controllers

**Reproduction Steps** (Please include YAML):
Original Grafana dashboards installed as described in [Monitoring with Grafana (optional)](https://karpenter.sh/v0.35/getting-started/getting-started-with-karpenter/#monitoring-with-grafana-optional)

I have submitted PR [#5935](https://github.com/aws/karpenter-provider-aws/pull/5935) for proposed changes.

**Versions**: 0.35.2
- Chart Version: karpenter-0.35.2
- Kubernetes Version (`kubectl version`): 1.29.2

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #5935, opened 2024-03-26
Modification of Karpenter Grafana dashboards capacity and performance

youwalther65:
Fixes [Karpenter 0.35.2 Grafana dashboards do not work properly #5934](https://github.com/aws/karpenter-provider-aws/issues/5934)

**Description**

- Capacity dashboard: removed deprecated metric "karpenter_deprovisioning_actions_performed" and provides new visualizations using metrics "karpenter_disruption_nodes_disrupted_total" and "karpenter_disruption_eligible_nodes"
- Performance dashboard: filtered for job="karpenter" to avoid getting controllers from other scarep jobsof K8s components using controller-runtime metrics as well

- Karpenter serviceMonitor object for kube-prometheus-stack created, which creates a scarep job "karpenter"
- provisioned dashboards to local Grafana instance

- [ ] Yes, issue opened: # 
- [x] No


njtran (commented):
I haven't pulled this down and applied it to my cluster, but I'd expect to see instances of `karpenter_deprovisioning_actions_performed` replaced with `karpenter_disruption_actions_performed`. But I don't see that. Can we just do a find all replace with the existing deprovisioning metrics and add the new metrics/changes in another PR?

youwalther65:
@njtran I used metric [karpenter_disruption_nodes_disrupted_total](https://karpenter.sh/docs/reference/metrics/#karpenter_disruption_nodes_disrupted_total) instead of metric [karpenter_disruption_actions_performed_total](https://karpenter.sh/docs/reference/metrics/#karpenter_disruption_actions_performed_total).
Both provide drill-down by label disruption action, method, and consolidation type, but [karpenter_disruption_nodes_disrupted_total](https://karpenter.sh/docs/reference/metrics/#karpenter_disruption_nodes_disrupted_total) gives these details even by NodePool, which I believe is a clear advantage over [karpenter_disruption_actions_performed_total](https://karpenter.sh/docs/reference/metrics/#karpenter_disruption_actions_performed_total)

njtran:
Hey @youwalther65, the number of actions taken is different from the number of nodes disrupted. Actions taken can provide information on the number of batches of disruptions. Do you mind adding the number of disruption actions back in as well?

youwalther65:
Hey @njtran , that's interesting, thank you. So it would look something like that:
Screenshot 2024-04-09 at 12 19 25
with the other Visualizations from the originla still in place (and not visible in the screenshot I provided). If you like that I will go ahead and put it into my PR.

jmdeal:
While disruption can be controlled at the nodepool level (TTLs, budgets, etc) the actual disruption actions are NodePool agnostic. For example, a single multi-node consolidation action could act upon nodes owned by multiple NodePools. We don't consider any action tied to a particular nodepool for that reason.

jmdeal:
Sorry it took me a moment to get around to pulling this down and reviewed, overall really like the changes!

Some nits on the chart naming for consistency / spelling:
- Nodes eligable for Disruptions -> Nodes Eligible for Disruption
- Number of disruption actions performed -> Disruption Actions Performed
- Node Disruptions -> Voluntary Node Disruptions

I was going to suggest "Nodes Disrupted" for the last one to be consistent with the other naming, but I realized that's really ambiguous since we also have "Nodes Terminated". Since these dashboards are really intended for newcomers to Karpenter, I don't think we want to rely on terminology that requires knowledge of Karpenter's node lifecycle to differentiate between. I think voluntary node disruptions hits the mark since it filters out involuntary disruptions which aren't driven by the disruption controller (e.g. spot interruption, kubectl delete node, etc). Definitely open to other ideas here though.

Some other nitty feedback:
- IMO nodes eligible for disruption shouldn't be stacked since the same node can be counted in multiple categories. Maybe I'm wrong but I've always read stacked graphs as the different sections being mutually exclusive.
- Now that we have multiple charts regarding disruption here, I'm wondering if it would be good to add a dashboard filter for the disruption method? Seems like it would be a nice addition if you're interested in adding it.

I did also notice some funky numbers for nodes eligible for disruption, turns out there's a bug in this metric that should be fixed with the following PR: https://github.com/kubernetes-sigs/karpenter/pull/1187.

youwalther65:
@jmdeal Thank you for the feedback. I like the ideas and have incorporated all of them in a new version. Please check again, thx.

jmdeal:
/unstale I'll try and take a look by the end of the day or tomorrow, I got caught up with some other tasks last week and this slipped through the cracks. Need to get the upstream metric fix in as well.

jmdeal:
Sorry for the delays here, two last comments and this looks good to me:
- Looks like there's still a typo in "Nodes Eligibale for Disruption" (should be Eligible)
- We should probably copy these dashboards into the v0.35, v0.36, and preview doc folders as well. Shouldn't need any additional changes there, just copies.

youwalther65:
@jmdeal Thx, will follow up after my vacation end of  May.

youwalther65:
Done with the requested changes @jmdeal

website/content/en/docs/getting-started/getting-started-with-karpenter/karpenter-capacity-dashboard.json
```diff
@@ -1,1440 +1,1743 @@
...
+				"AliasAttributes": [
+					"email",
+					"preferred_username"
+				],
+				"UserPoolName": {
+					"Fn::Sub": "${AWS::StackName}-UserPool"
+				}
+			}
+		}
+	}
```
jmdeal, line 1742: I'm curious where this came from, seems like it was added in this latest commit. I can't figure out what, if anything, this affects in the dashboard but it seems out of place and looks more like cloudformation json than anything.

youwalther65: That's really kind of strange. I am not aware how this happened, sorry for that. Removed it , should work now.

jmdeal (approved):
LGTM, thanks for the contribution 🚀

PR #5946, opened 2024-03-29
fix: filter out a1 instance types for AL2023

jmdeal:
Fixes #N/A 

**Description**
Adds a filter in the instance type provider to remove incompatible instance types from the cloudprovider GetInstanceTypes call. Specifically, this PR adds a filter for the a1 instance family when the EC2NodeClass is using the AL2023 AMI family. AL2023 AMIs are explicitly incompatible with the instance family, and due to it being an architectural requirement this will not change in the future. 

ref: https://docs.aws.amazon.com/linux/al2023/ug/system-requirements.html

`make test`

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/pkg/cloudprovider/suite_test.go
+++ b/pkg/cloudprovider/suite_test.go
@@ -50,6 +50,7 @@ import (
 	"sigs.k8s.io/karpenter/pkg/events"
 	coreoptions "sigs.k8s.io/karpenter/pkg/operator/options"
 	"sigs.k8s.io/karpenter/pkg/operator/scheme"
+	"sigs.k8s.io/karpenter/pkg/scheduling"
 	coretest "sigs.k8s.io/karpenter/pkg/test"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -491,7 +492,6 @@ var _ = Describe("CloudProvider", func() {
 	Context("NodeClaim Drift", func() {
 		var armAMIID, amdAMIID string
 		var validSecurityGroup string
-		var selectedInstanceType *corecloudproivder.InstanceType
 		var instance *ec2.Instance
 		var validSubnet1 string
 		var validSubnet2 string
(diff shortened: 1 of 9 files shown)


pkg/providers/instancetype/instancetype.go
```diff
@@ -143,6 +143,7 @@ func (p *Provider) List(ctx context.Context, kc *corev1beta1.KubeletConfiguratio
 
 		return NewInstanceType(ctx, i, kc, p.region, nodeClass, p.createOfferings(ctx, i, instanceTypeOfferings[aws.StringValue(i.InstanceType)], allZones, subnetZones))
 	})
+	result = filterIncompatibleInstanceTypes(nodeClass, result)
```
engedaam, line 146: Could we just do the filtering when we are creating the instance type list?

pkg/providers/instancetype/suite_test.go
```diff
@@ -823,6 +823,29 @@ var _ = Describe("InstanceTypes", func() {
...
+			})
+			pod := coretest.UnschedulablePod()
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass, pod)
+			instanceTypes, err := cloudProvider.GetInstanceTypes(ctx, nodePool)
+			Expect(err).To(BeNil())
+			Expect(lo.ContainsBy(instanceTypes, func(it *corecloudprovider.InstanceType) bool {
+				return lo.Contains(it.Requirements.Get(v1beta1.LabelInstanceFamily).Values(), "a1")
+			})).To(Equal(amiFamily != v1beta1.AMIFamilyAL2023))
+		},
+		Entry("AL2", v1beta1.AMIFamilyAL2),
```
jonathan-innis, line 841: You could go either way here (I just thought of it since you have an argument here that matches the title): You can specify `nil` as the first argument and the test will just print the arguments that are using for the entry if it fails. 

Minimally, this keeps you from having to maintain consistency between the description and the arguments in this case

pkg/cloudprovider/suite_test.go
```diff
@@ -572,7 +572,12 @@ var _ = Describe("CloudProvider", func() {
 			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
 			instanceTypes, err := cloudProvider.GetInstanceTypes(ctx, nodePool)
 			Expect(err).ToNot(HaveOccurred())
-			selectedInstanceType = instanceTypes[0]
+			selectedInstanceType, found := lo.Find(instanceTypes, func(it *corecloudproivder.InstanceType) bool {
```
jonathan-innis, line 575: Why do we need this additional handling here? Feels odd that we would need to add additional logic to all test cases because we added filtering to `a1`. Maybe this speaks to the fact that we should statically select the instance type here. This might be necessary, hard for me to say, but on first look: It seems like a strange side-effect of these changes.

jmdeal: This could be a static selection, it just so happened that the first instance type is now an a1 instance. IIRC this caused a failure when computing drift since `a1` instances are excluded from the `GetInstanceTypes` call for AL2023. I'm actually wondering if this could pose a problem for users upgrading, I'm not sure what happens if a NodeClass has an instance type option that is no longer surfaced by `GetInstanceTypes`. I'm not sure we need to worry about the existing node case since it wouldn't be working anyway.

jmdeal:
Closing this out for the time being since it violates the tenant of NodePool driving scheduling decisions. If we want the EC2NodeClass to affect scheduling decisions there are a number of additional changes we should probably incorporate (e.g. restricting instance architecture based on AMI availability). Will open another PR with just a docs warning.

PR #5949, opened 2024-04-01
BREAKING CHANGE: Refactor NodeClass controller

engedaam:
Fixes #N/A 

**Description**
- **BREAKING CHANGE:** the names for log keys produced by the NodeClass controller will change as such: `nodeclass -> nodeclass.status, nodeclass.hash, nodeclass.termination`
- Refactored the NodeClass controller to be split into three controllers (termination, status, and hash) 
- This change will keep functionality of each controllers to one operation and easier to reason through 
- All logic of the NodeClass will remain consistent

- `make presubmit` 
- `/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No


engedaam (commented):
/karpenter snapshot

engedaam (commented):
/karpenter snapshot

pkg/controllers/controllers.go
```diff
@@ -52,7 +54,9 @@ func NewControllers(ctx context.Context, sess *session.Session, clk clock.Clock,
 	pricingProvider *pricing.Provider, amiProvider *amifamily.Provider, launchTemplateProvider *launchtemplate.Provider) []controller.Controller {
 
 	controllers := []controller.Controller{
-		nodeclass.NewController(kubeClient, recorder, subnetProvider, securityGroupProvider, amiProvider, instanceProfileProvider, launchTemplateProvider),
+		hash.NewController(kubeClient),
```
jonathan-innis, line 57:
```suggestion
		nodeclasshash.NewController(kubeClient),
```

Like the other controllers, I think that it's a bit easier to read if we prefix these with the `nodeclass` package name as well

pkg/controllers/controllers.go
```diff
@@ -52,7 +54,9 @@ func NewControllers(ctx context.Context, sess *session.Session, clk clock.Clock,
...
+		status.NewController(kubeClient, subnetProvider, securityGroupProvider, amiProvider, instanceProfileProvider),
```
jonathan-innis, line 58:
```suggestion
		nodeclassstatus.NewController(kubeClient, subnetProvider, securityGroupProvider, amiProvider, instanceProfileProvider),
```

pkg/controllers/controllers.go
```diff
@@ -52,7 +54,9 @@ func NewControllers(ctx context.Context, sess *session.Session, clk clock.Clock,
...
+		termination.NewController(kubeClient, recorder, instanceProfileProvider, launchTemplateProvider),
```
jonathan-innis, line 59:
```suggestion
		nodeclasstermination.NewController(kubeClient, recorder, instanceProfileProvider, launchTemplateProvider),
```

pkg/controllers/nodeclass/hash/controller.go
```diff
@@ -0,0 +1,115 @@
...
+		if err := c.kubeClient.Patch(ctx, nodeClass, client.MergeFrom(stored)); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) Name() string {
+	return "nodeclass.hash"
```
jonathan-innis, line 68: This change is technically a breaking change to the controller name for logging purposes. We should make a note in the description and in the upgrade guide that this component name is changing with this update so anyone relying on these log keys isn't surprised

pkg/controllers/nodeclass/termination/controller.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
+	stored := nodeClass.DeepCopy()
+	if controllerutil.ContainsFinalizer(nodeClass, v1beta1.TerminationFinalizer) {
+		return reconcile.Result{}, nil
+	}
+	controllerutil.AddFinalizer(nodeClass, v1beta1.TerminationFinalizer)
+	err := c.kubeClient.Patch(ctx, nodeClass, client.MergeFrom(stored))
+
+	return reconcile.Result{}, err
```
jonathan-innis, line 79: This doesn't ignore NotFound scenarios. Consider something like

```go
return reconcile.Result{}, client.IgnoreNotFound(c.kubeClient.Patch(ctx, nodeClass, client.MergeFrom(stored))
```

pkg/operator/operator.go
```diff
@@ -176,6 +176,10 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 		launchTemplateProvider,
 	)
 
+	if cidrErr := launchTemplateProvider.ResolveClusterCIDR(ctx); cidrErr != nil {
```
jonathan-innis, line 179: This will cause crashing now for anyone who isn't using DescribeCluster, which is a major breaking change. 

The reason this existed before only as an error and was gated for only AL2023 is because we took the stance that users using AL2023 should just give us DescribeCluster permission. However, not everyone is using EKS and certainly not everyone is giving us DescribeCluster. We should shift this back to be gated only by AL2023.

jonathan-innis (commented):
Nice refactor. I really like the changes! Assuming the logic is remaining consistent everywhere with the changes and the tests haven't changed, I like it.

Didn't look at the tests because I assumed that they were straight-copies from the different `Context` blocks inside of the `nodeclass` package `suite_test.go`. Let me know if this assumption doesn't hold and I can look at specific files.

pkg/operator/operator.go
engedaam: Sounds good, add this back into the status controller and gated to be only resolved on AL2023. I also added a comment just to outline why we do it as reconcile loop and not on start-up.

engedaam (commented):
/karpenter snapshot

pkg/operator/operator.go
jonathan-innis: Thanks. Adding a comment is a good call, since I'm realizing after having this convo with you that only @jmdeal and I were aware of exactly why this line existed where it did

engedaam (commented):
/karpenter snapshot

pkg/controllers/nodeclass/status/ami.go
```diff
@@ -0,0 +1,58 @@
...
+		return reconcile.Result{}, err
+	}
+	if len(amis) == 0 {
+		nodeClass.Status.AMIs = nil
+		return reconcile.Result{}, fmt.Errorf("no amis exist given constraints")
+	}
+	nodeClass.Status.AMIs = lo.Map(amis, func(ami amifamily.AMI, _ int) v1beta1.AMI {
+		reqs := ami.Requirements.NodeSelectorRequirements()
+		sort.Slice(reqs, func(i, j int) bool {
+			if len(reqs[i].Key) != len(reqs[j].Key) {
```
jonathan-innis, line 46: Not something that you need to change in this PR, but seems odd to me that we compare the length here first before ordering them alphabetically. Do you know why we had originally decided to do that?

pkg/controllers/nodeclass/status/controller.go
```diff
@@ -0,0 +1,105 @@
...
+	})
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
+	stored := nodeClass.DeepCopy()
+
+	var results []reconcile.Result
+	var errs error
+	for _, reconciler := range []nodeClassStatusReconciler{
+		c.ami,
```
jonathan-innis, line 74: Consider in a follow-up: I'm not fully convinced that this pattern is actually a good idea -- where we have an explicit ordering of operations that rely on each other. I don't even think that these things need ordering, but if something _does_ need explicit ordering, seems better to update a status condition in a separate controller to decouple the ordering.

pkg/controllers/nodeclass/termination/controller.go
```diff
@@ -0,0 +1,149 @@
...
+		kubeClient:              kubeClient,
+		recorder:                recorder,
+		instanceProfileProvider: instanceProfileProvider,
+		launchTemplateProvider:  launchTemplateProvider,
+	})
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
+	stored := nodeClass.DeepCopy()
+	if controllerutil.ContainsFinalizer(nodeClass, v1beta1.TerminationFinalizer) {
```
jonathan-innis, line 73: This is going to cause leaks. The finalizer needs to be directly tied to the operation that it's blocking in its termination flow. In the way that you have it here, it wouldn't happen often, but it's possible that the finalizer never gets patched on, there was a NodeClaim launch, and then the EC2NodeClass is terminated. This will never wait for the launch templates to cleanup or the instance profiles to go away.

website/content/en/preview/upgrading/upgrade-guide.md
```diff
@@ -49,6 +49,8 @@ WHEN CREATING A NEW SECTION OF THE UPGRADE GUIDANCE FOR NEWER VERSIONS, ENSURE T
 
 * Karpenter changed the name of the `karpenter_cloudprovider_instance_type_price_estimate` metric to `karpenter_cloudprovider_instance_type_offering_price_estimate` to align with the new `karpenter_cloudprovider_instance_type_offering_available` metric. The `region` label was also dropped from the metric, since this can be inferred from the environment that Karpenter is running in.
 
+* Karpenter changed the log keys produced by the NodeClass controller as such: `nodeclass -> nodeclass.status, nodeclass.hash, nodeclass.termination`
```
jonathan-innis, line 52:
```suggestion
* Karpenter updated the NodeClass controller naming in the following way: `nodeclass` -> `nodeclass.status`, `nodeclass.hash`, `nodeclass.termination`
```

pkg/controllers/nodeclass/termination/controller.go
engedaam: Discussed offline, we will be adding the NodeClass finalizer as part of the instanceProfile Creation. Since the instanceProfile creation is current operation that gates instance launch, that will prevent the race of when EC2NodeClass does not have a finalizer and gets deleted.

pkg/controllers/nodeclass/status/controller.go
engedaam: We can potentially do these operation in parallel, I can make that change as part of the readiness work

pkg/controllers/nodeclass/status/ami.go
engedaam: We wanted to make sure that architecture came before gpu requirements

pkg/controllers/nodeclass/status/instanceprofile.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/aws/karpenter-provider-aws/pkg/apis/v1beta1"
+	"github.com/aws/karpenter-provider-aws/pkg/providers/instanceprofile"
+)
+
+type InstanceProfile struct {
+	instanceProfileProvider instanceprofile.Provider
+}
+
+func (ip *InstanceProfile) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
+	if controllerutil.ContainsFinalizer(nodeClass, v1beta1.TerminationFinalizer) {
```
jonathan-innis, line 34: Can this just be added at the top-level for now and patched at the beginning of the reconcile. This code still has a race since you can still go and create the instance profile, but fail to update the Kubernetes object

engedaam: Moved it to the top level of the reconcile loop. Seems like this race is currently present, no? Since we can fail to update the k8s object even in the current NodeClass controller

pkg/controllers/nodeclass/status/controller.go
```diff
@@ -0,0 +1,113 @@
...
+		ami:             &AMI{amiProvider: amiProvider},
+		subnet:          &Subnet{subnetProvider: subnetProvider},
+		securitygroup:   &SecurityGroup{securityGroupProvider: securityGroupProvider},
+		instanceprofile: &InstanceProfile{instanceProfileProvider: instanceProfileProvider},
+		launchtemplate:  &LaunchTemplate{launchTemplateProvider: launchTemplateProvider},
+	})
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
+	stored := nodeClass.DeepCopy()
```
jonathan-innis, line 70: You should create this stored object after you patch, else you are going to be trying to patch the finalizer on two separate occassions

jonathan-innis (dismissed):
LGTM 🚀

engedaam (commented):
/karpenter snapshot

jonathan-innis (approved):
LGTM 🚀

PR #5679, opened 2024-02-16
changes in getting-started-with-karpenter doc to fix #5677

santosh-at-github:
Fixes #5677

**Description**
The Karpenter pods fails to come up when the helm command provided in the documentation used for deploying Karpenter pods on Fargate. I have suggested changes in helm command which would fix it.

**How was this change tested?**
In my test account EKS deployment.

**Does this change impact docs?**
- [x] Yes, PR includes docs updates 
- [ ] Yes, issue opened: # 
- [ ] No

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5961, opened 2024-04-02
fix: missing step in step08-apply-helm-chart.sh re serviceaccount annotation

dspaxton:
Missing step from step08-apply-helm.chart.sh which configures an annotation to the service account to allow IAM access otherwise karpenter fails to deploy.

Fixes #N/A 

The latest (v0.35) version of the docs includes this step:

# Logout of helm registry to perform an unauthenticated pull against the public ECR
helm registry logout public.ecr.aws

helm upgrade --install karpenter oci://public.ecr.aws/karpenter/karpenter --version "${KARPENTER_VERSION}" --namespace "${KARPENTER_NAMESPACE}" --create-namespace \
  --set "settings.clusterName=${CLUSTER_NAME}" \
  --set "settings.interruptionQueue=${CLUSTER_NAME}" \
  --set controller.resources.requests.cpu=1 \
  --set controller.resources.requests.memory=1Gi \
  --set controller.resources.limits.cpu=1 \
  --set controller.resources.limits.memory=1Gi \
  --wait

This step does not include annotation of the service account to points to the ${KARPENTER_IAM_ROLE} so that IRSA can make appropriate calls to the backend services. If using Fargate for Karpenter then IRSA will be a requirement. 

I've tested this on a cluster of my own.

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/website/content/en/preview/getting-started/getting-started-with-karpenter/scripts/step08-apply-helm-chart.sh
+++ b/website/content/en/preview/getting-started/getting-started-with-karpenter/scripts/step08-apply-helm-chart.sh
@@ -1,9 +1,10 @@
 # Logout of helm registry to perform an unauthenticated pull against the public ECR
 helm registry logout public.ecr.aws
 
-helm upgrade --install karpenter oci://public.ecr.aws/karpenter/karpenter --version "${KARPENTER_VERSION}" --namespace "${KARPENTER_NAMESPACE}" --create-namespace \
-  --set "settings.clusterName=${CLUSTER_NAME}" \
-  --set "settings.interruptionQueue=${CLUSTER_NAME}" \
+helm upgrade --install karpenter oci://public.ecr.aws/karpenter/karpenter --version ${KARPENTER_VERSION} --namespace karpenter --create-namespace \
+  --set serviceAccount.annotations."eks\.amazonaws\.com/role-arn"=${KARPENTER_IAM_ROLE_ARN} \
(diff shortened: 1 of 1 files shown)


engedaam:
@dspaxton As part of the getting started guide, Karpenter is installed using pod Identity, so users should not need the annotation unless they are using Fargate or running a version of k8s < 1.24. There is a notice on the getting started page bellow on installing Karpenter step https://karpenter.sh/docs/getting-started/getting-started-with-karpenter/#4-install-karpenter

dspaxton:
The documentation is somewhat misleading then since the cluster configuration calls out that fargate doesn't support pod identity but if a customer wishes to use Karpenter on fargate then they have the stanza for the cluster configuration but will run into issues since the helm install will fail as it currently stands since it doesn't include the serviceaccount.annotation step.

engedaam:
The team has heard this feedback before. There is currently a PR in progress that will make separate install commands for both Fargate and MNG. The team was thinking this would a better way to go instead of adding the SA annotation. That PR being: https://github.com/aws/karpenter-provider-aws/pull/5679. Would you be willing to leave feedback on that PR? since you also see value in that change.

dspaxton:
Absolutely - thank you

engedaam:
Also, here is a preview of the proposed changes in that PR https://deploy-preview-5679--karpenter-docs-prod.netlify.app/docs/getting-started/getting-started-with-karpenter/. Would you be okay closing this PR out in favor of that one?

PR #5979, opened 2024-04-03
chore: Wait for instance termination before deleting nodeclaim

jigisha620:
Fixes #N/A 

**Description**
The termination controller calls `cloudProvider.Delete(nodeClaim)` which tries to terminate the underlying instance. It will remove the finalizer once the cloudProvider.Delete() call succeeds. However, the current implementation does not wait until the instance state is actually `terminated` and proceeds to remove the finalizer once the `ec2.TerminateInstance()` call has succeeded. 

The change in this PR will check if the instance state is not terminated (check the instance state code)and will return a retryable error that will effectively re-enqueue the node into the reconciliation loop of the termination controller.

- [ ] Yes, issue opened: # 
- [x] No


go.mod
```diff
@@ -115,3 +115,5 @@ require (
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
+
+replace sigs.k8s.io/karpenter => github.com/jigisha620/karpenter v0.0.0-20240403195503-a1d90e788c2a
```
engedaam, line 119: We should remove this line

jigisha620: I am going to remove this once the upstream PR gets merged. This has been added because of the dependency on upstream changes for which I have a PR out - https://github.com/kubernetes-sigs/karpenter/pull/1164
The RetryableError added in upstream PR is being referenced in this one.

jigisha620: Removed

pkg/providers/instance/instance.go
```diff
@@ -166,6 +167,11 @@ func (p *Provider) Delete(ctx context.Context, id string) error {
 		}
 		return fmt.Errorf("terminating instance, %w", err)
 	}
+	for _, instance := range response.TerminatingInstances {
+		if instance.CurrentState.Name != aws.String(ec2.InstanceStateNameTerminated) {
+			return cloudprovider.NewRetryableError(fmt.Errorf("terminating instance '%s'", aws.StringValue(instance.CurrentState.Name)))
```
engedaam, line 172: How ofter are we logging this error to customers? Also this doesn't seem like an error, but more debug information

engedaam (commented):
I think we should add and E2E test to validate this behavior on top of the unit test

jigisha620 (commented):
/karpenter snapshot

test/suites/nodeclaim/nodeclaim_test.go
```diff
@@ -377,4 +377,45 @@ var _ = Describe("StandaloneNodeClaim", func() {
...
+			},
+		})
+		env.ExpectCreated(nodeClass, nodeClaim)
+		node := env.EventuallyExpectInitializedNodeCount("==", 1)[0]
+		nodeClaim = env.EventuallyExpectCreatedNodeClaimCount("==", 1)[0]
+
+		instanceID := env.ExpectParsedProviderID(node.Spec.ProviderID)
+		env.GetInstance(node.Name)
+
+		env.ExpectDeleted(node)
```
engedaam, line 411: Why are we deleting the node here? Shouldnt we delete the NodeClaim?

test/suites/nodeclaim/nodeclaim_test.go
```diff
@@ -377,4 +377,45 @@ var _ = Describe("StandaloneNodeClaim", func() {
...
+		Eventually(func(g Gomega) {
```
engedaam, line 412: Shouldn't we check that finalizer is still on the NodeClaim when the instance is in the `shuting-down` pahse

test/suites/nodeclaim/nodeclaim_test.go
```diff
@@ -377,4 +377,45 @@ var _ = Describe("StandaloneNodeClaim", func() {
...
+			g.Expect(lo.FromPtr(env.GetInstanceByID(instanceID).State.Name)).To(Equal("shutting-down"))
+		}, time.Second*10).Should(Succeed())
+
+		env.EventuallyExpectNotFound(nodeClaim, node)
+		Eventually(func(g Gomega) {
```
engedaam, line 417: Why are we checking that the instance is terminated here?

engedaam (commented):
Seems like there are failures on the E2E tests

test/suites/nodeclaim/nodeclaim_test.go
jigisha620: Made the change.

test/suites/nodeclaim/nodeclaim_test.go
jigisha620: Added

test/suites/nodeclaim/nodeclaim_test.go
jigisha620: My though behind this - the termination finalizer on nodeClaim is only removed when the instance is terminated. And if the nodeClaim is deleted, we verify that the instance state is also terminated.

jigisha620 (commented):
/karpenter snapshot

pkg/controllers/nodeclaim/garbagecollection/controller.go
```diff
@@ -90,8 +90,8 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 
 func (c *Controller) garbageCollect(ctx context.Context, nodeClaim *v1beta1.NodeClaim, nodeList *v1.NodeList) error {
 	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("provider-id", nodeClaim.Status.ProviderID))
-	if err := c.cloudProvider.Delete(ctx, nodeClaim); err != nil {
-		return cloudprovider.IgnoreNodeClaimNotFoundError(err)
+	if err := c.cloudProvider.Delete(ctx, nodeClaim); cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil && cloudprovider.IgnoreRetryableError(err) != nil {
```
jonathan-innis, line 93: This was part of my concern with returning an error when the operation succeeds. Now we have to handle this error wherever Delete() is called. It's not well-understood why we are handling it here, because the reason we are firing the error only relates to a narrow case in the NodeClaim termination flow

pkg/providers/instance/instance.go
```diff
@@ -152,9 +152,10 @@ func (p *Provider) List(ctx context.Context) ([]*Instance, error) {
 }
 
 func (p *Provider) Delete(ctx context.Context, id string) error {
-	if _, err := p.ec2Batcher.TerminateInstances(ctx, &ec2.TerminateInstancesInput{
+	response, err := p.ec2Batcher.TerminateInstances(ctx, &ec2.TerminateInstancesInput{
```
jonathan-innis, line 155:
```suggestion
	res, err := p.ec2Batcher.TerminateInstances(ctx, &ec2.TerminateInstancesInput{
```

nit: This is the name use elsewhere. I think we may also use `out` but you can check

pkg/providers/instance/instance.go
```diff
@@ -166,6 +167,11 @@ func (p *Provider) Delete(ctx context.Context, id string) error {
 		}
 		return fmt.Errorf("terminating instance, %w", err)
 	}
+	for _, instance := range response.TerminatingInstances {
```
jonathan-innis, line 170: What are the possible states for the response here? Do we sometimes get a length of 1 response and sometimes get a length of 0? Or is always 1?

pkg/providers/instance/instance.go
```diff
@@ -166,6 +167,11 @@ func (p *Provider) Delete(ctx context.Context, id string) error {
...
+		if *instance.CurrentState.Name != ec2.InstanceStateNameTerminated {
+			return cloudprovider.NewRetryableError(fmt.Errorf("terminating instance '%s'", aws.StringValue(instance.CurrentState.Name)))
```
jonathan-innis, line 172: IMO this error should be more verbose about why it is "retryable". Maybe something like, `fmt.Errorf("received terminate instance call but instance is still in state %q, instance.CurrentState.Name")`.

Also, you can use the `%q` formatting directive to easily quote variables

pkg/providers/instance/instance.go
jigisha620: I think we will always get 1 here.

pkg/providers/instance/instance.go
jigisha620: Haven't really used this elsewhere in this function but `out` seems good.

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

jigisha620 (commented):
/karpenter snapshot

test/suites/expiration/suite_test.go
```diff
@@ -586,13 +586,6 @@ var _ = Describe("Expiration", func() {
 			// TODO: reduce timeouts when deprovisioning waits are factored out
 			env.EventuallyExpectNodesUntaintedWithTimeout(11*time.Minute, taintedNodes...)
 
-			// The nodeclaims that never registers will be removed
```
jmdeal, line 589: Why are we removing this expectation vs increasing the timeout? Is the validation just not worth the additional E2E time or do the upstream changes break an assumption here that I'm not realizing?

jigisha620: There's no change in upstream that breaks an assumption here. This expectation takes too long to succeed. In fact I have already increased the default time out to 16 minutes [here](https://github.com/aws/karpenter-provider-aws/pull/5979/files#diff-e50fbe8dc3a4e43a11436cbd3e0c86d0fefcb7713d7357cf67c5a566dcbca958R80). But the test would still fail when I tried running it on my local cluster with `[FAILED] Timed out after 960.000s.`

jmdeal (approved):
LGTM 🚀

PR #5982, opened 2024-04-04
chore: Forward Container logs for E2E tests

engedaam:
Fixes #N/A 

**Description**
- Add log forwarding for containers to CloudWatch

- Tested in a forked repo

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.github/actions/e2e/setup-cluster/action.yaml
+++ b/.github/actions/e2e/setup-cluster/action.yaml
@@ -30,7 +30,7 @@ inputs:
     default: "1.29"
   eksctl_version:
     description: "Version of eksctl to install"
-    default: v0.169.0
+    default: v0.175.0
   ip_family:
     description: "IP Family of the cluster. Valid values are IPv4 or IPv6"
     default: "IPv4"
@@ -152,11 +152,9 @@ runs:
           minSize: 2
           maxSize: 2
           iam:
+            withAddonPolicies:
+              cloudWatch: true
             instanceRolePermissionsBoundary: "arn:aws:iam::$ACCOUNT_ID:policy/GithubActionsPermissionsBoundary"
-          taints:
-          - key: CriticalAddonsOnly
-            value: "true"
-            effect: NoSchedule
       cloudWatch:
         clusterLogging:
           enableTypes: ["*"]
@@ -175,6 +173,8 @@ runs:
         $KARPENTER_IAM
         withOIDC: true
       addons:
+      - name: amazon-cloudwatch-observability
+        permissionsBoundary: "arn:aws:iam::$ACCOUNT_ID:policy/GithubActionsPermissionsBoundary"
       - name: vpc-cni
         permissionsBoundary: "arn:aws:iam::$ACCOUNT_ID:policy/GithubActionsPermissionsBoundary"
       - name: coredns
@@ -211,6 +211,11 @@ runs:
       else
         eksctl ${cmd} cluster -f clusterconfig.yaml
       fi
+
+      # Adding taints after all necessary pods have scheduled to the manged node group nodes
+      # amazon-cloudwatch-observability pods do no not tolerate CriticalAddonsOnly=true:NoSchedule and 
+      # amazon-cloudwatch-observability addons does not allow to add tolerations to the addon pods as part of the advanced configuration
+      kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --all
   - name: tag oidc provider of the cluster
     if: always()
     shell: bash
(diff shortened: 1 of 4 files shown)


.github/actions/e2e/setup-cluster/action.yaml
```diff
@@ -269,3 +273,13 @@ runs:
       k8s_version: ${{ inputs.k8s_version }}
       git_ref: ${{ inputs.git_ref }}
       private_cluster: ${{ inputs.private_cluster }}
+  - name: log retention policy
```
jonathan-innis, line 276: Do we have log forwarding and a retention policy setup for our private cluster? If not, we should configure that so that we can access all of the logs in one place

.github/actions/e2e/setup-cluster/action.yaml
```diff
@@ -269,3 +273,13 @@ runs:
...
+    shell: bash
+    if: ${{ inputs.private_cluster == 'false' }}
+    env:
+      CLUSTER_NAME: ${{ inputs.cluster_name }}
+    run: |
+      aws logs put-retention-policy --log-group-name /aws/containerinsights/$CLUSTER_NAME/application --retention-in-days 30
```
jonathan-innis, line 282: What's the default retention policy that CloudWatch assigns to us? Is there a way to configure the retention directly through the addon config with eksctl?

jonathan-innis (commented):
Can you post the result of getting these logs back in the description? I also don't think this is going to give us Karpenter logs out of the box. I'm imagining that there's a bunch of additional configuration that you have to do the fluentbit container so that we ensure that it ships logs specifically from our application

jonathan-innis:
I think that the CWA will send back metrics as well by default. How do we ensure that we are not double-sending metrics with our current prometheus agent and the CWA? Also, maybe something to look into, is there a way to integrate the CWA with AMP so that we don't need the separate prometheus instance running on the cluster?

.github/actions/e2e/setup-cluster/action.yaml
engedaam: There is no retention policy set on the log stream and there is currently no way to configure the retention using eksctl.

.github/actions/e2e/setup-cluster/action.yaml
engedaam: Added and Updated

.github/actions/e2e/setup-cluster/action.yaml
```diff
@@ -30,7 +30,7 @@ inputs:
     default: "1.29"
   eksctl_version:
     description: "Version of eksctl to install"
-    default: v0.169.0
+    default: v0.175.0
```
jonathan-innis, line 33: Can we create an issue for auto-updating this through automation?

.github/workflows/e2e-upgrade.yaml
```diff
@@ -90,7 +90,7 @@ jobs:
           region: ${{ inputs.region }}
           cluster_name: ${{ steps.generate-cluster-name.outputs.CLUSTER_NAME }}
           k8s_version: ${{ inputs.k8s_version }}
-          eksctl_version: v0.169.0
+          eksctl_version: v0.175.0
```
jonathan-innis, line 93: Same comment here

.github/actions/e2e/setup-cluster/action.yaml
```diff
@@ -211,6 +211,11 @@ runs:
       else
         eksctl ${cmd} cluster -f clusterconfig.yaml
       fi
+
+      # Adding taints after all necessary pods have scheduled to the manged node group nodes
+      # amazon-cloudwatch-observability pods do no not tolerate CriticalAddonsOnly=true:NoSchedule and 
+      # EKS addons does not allow to add tolerations to the addon pods as part of the advanced configuration 
+      kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --all
```
jonathan-innis, line 218: Rather than banking on cluster creation and then tainting immediately after the fact, can this just be part of the ClusterConfig and then anything that doesn't tolerate it by default out-of-the-box, we just add additional addon configuration to?

.github/actions/e2e/setup-cluster/action.yaml
engedaam: Created the issue: https://github.com/aws/karpenter-provider-aws/issues/6096

.github/actions/e2e/setup-cluster/action.yaml
engedaam: Currently, observability addon configuration does not support adding tolerations

jonathan-innis (approved):
LGTM 🚀

PR #5992, opened 2024-04-06
BREAKING CHANGE: make AMISelectorTerms required

njtran:
Fixes #N/A 

**Description**
Makes the AMISelectorTerms required. @jmdeal is taking this over. 

- [ ] Yes, issue opened: # 
- [x] No


cmd/controller/main.go
```diff
@@ -43,6 +47,27 @@ func main() {
 	lo.Must0(op.AddHealthzCheck("cloud-provider", awsCloudProvider.LivenessProbe))
 	cloudProvider := metrics.Decorate(awsCloudProvider)
 
+	client := op.Manager.GetAPIReader()
+	if client == nil {
```
jonathan-innis, line 51: Is this possible?

cmd/controller/main.go
```diff
@@ -43,6 +47,27 @@ func main() {
...
+		panic("panic: failed to initialize read-client on startup")
+	}
+	nodeClassList := v1beta1.EC2NodeClassList{}
+	err := client.List(ctx, &nodeClassList)
+	if err != nil {
+		panic(fmt.Sprintf("panic: failed to list ec2nodeclasses on startup, %s", err.Error()))
```
jonathan-innis, line 57: I question whether these should be `logging.FromContext().Fatal` vs. `panic`. Panics are typically reserved for programming errors that we don't expect to happen. These are more like startup conditions aren't met, which feels more semantically correct to "fatal" out on

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -48,8 +48,8 @@ type EC2NodeClassSpec struct {
 	// +kubebuilder:validation:XValidation:message="expected at least one, got none, ['tags', 'id', 'name']",rule="self.all(x, has(x.tags) || has(x.id) || has(x.name))"
 	// +kubebuilder:validation:XValidation:message="'id' is mutually exclusive, cannot be set with a combination of other fields in amiSelectorTerms",rule="!self.all(x, has(x.id) && (has(x.tags) || has(x.name) || has(x.owner)))"
 	// +kubebuilder:validation:MaxItems:=30
-	// +optional
-	AMISelectorTerms []AMISelectorTerm `json:"amiSelectorTerms,omitempty" hash:"ignore"`
+	// +required
+	AMISelectorTerms []AMISelectorTerm `json:"amiSelectorTerms" hash:"ignore"`
```
jonathan-innis, line 52: We should validate that you can't still specify an empty slice here and have it be valid

cmd/controller/main.go
jmdeal: After reading through the source from controller-runtime the answer appears to be no, we can remove this check.

jmdeal (commented):
/karpenter snapshot

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -65,6 +65,23 @@ spec:
                     AMISelectorTerm defines selection logic for an ami used by Karpenter to launch nodes.
                     If multiple fields are used for selection, the requirements are ANDed.
                   properties:
+                    eksOptimized:
+                      description: EKSOptimized test
```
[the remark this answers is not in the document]
jmdeal: Fixed locally, staged to push 👍

website/content/en/preview/concepts/nodeclasses.md
```diff
@@ -449,14 +454,36 @@ amiSelectorTerms:
...
+        karpenter.sh/discover: "${CLUSTER_NAME}"
+        environment: test
+```
+
+```yaml
+  amiSelectorTerms:
+    - eksOptimzed:
+      family: AL2023
+    # Can not specify other terms with eksOptimzed
+    - id: ami-123
```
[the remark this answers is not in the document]
jmdeal: Discussed offline, agreed to tentatively keep this validation on the CRD / webhook level but to not make this assumption in code. Even though we can allow this from a technical standpoint, the latest AMI from any given term would be used as normal, since we know that users are not in control of the release cycle of EKS optimized AMIs mixing this term with others could result in unexpected drift. We can change this in the future if good use-case arguments are made, but for now we'll choose to block a potential footgun. 

The reason we're just doing this on the validation side and not treating this as an assumed invariant in code is there are users who will be able to circumvent this validation. The main case I'm considering is users on 1.23 and 1.24 (no CEL) who haven't deployed the validation webhook, but we can also consider users who edited their CRD. In both of these cases, I believe amiSelectorTerms should work consistently with it's definition, amis resolved from different terms being ored together and the latest compatible AMI being chosen.

cmd/controller/main.go
```diff
@@ -43,6 +47,20 @@ func main() {
 	lo.Must0(op.AddHealthzCheck("cloud-provider", awsCloudProvider.LivenessProbe))
 	cloudProvider := metrics.Decorate(awsCloudProvider)
 
+	client := op.Manager.GetAPIReader()
```
jonathan-innis, line 50: Can we consider scoping this into a function to make it easier to grok what this is validating, something like `ValidateEC2NodeClassAMISelectorTerms`?

cmd/controller/main.go
```diff
@@ -43,6 +47,20 @@ func main() {
...
+	nodeClassList := v1beta1.EC2NodeClassList{}
+	err := client.List(ctx, &nodeClassList)
+	if err != nil {
+		logging.FromContext(ctx).Fatalf("failed to list EC2NodeClasses on startup, %s", err.Error())
```
jonathan-innis, line 54:
```suggestion
		logging.FromContext(ctx).Fatalf("listing EC2NodeClasses on startup, %s", err.Error())
```

nit: consider using the gerund since we use this everywhere for wrapping errors in the codebase

cmd/controller/main.go
```diff
@@ -43,6 +47,20 @@ func main() {
 	lo.Must0(op.AddHealthzCheck("cloud-provider", awsCloudProvider.LivenessProbe))
 	cloudProvider := metrics.Decorate(awsCloudProvider)
 
+	client := op.Manager.GetAPIReader()
+	nodeClassList := v1beta1.EC2NodeClassList{}
+	err := client.List(ctx, &nodeClassList)
```
jonathan-innis, line 52: nit: Collapse this and the subsequent line

cmd/controller/main.go
```diff
@@ -43,6 +47,20 @@ func main() {
...
+	if err != nil {
+		logging.FromContext(ctx).Fatalf("failed to list EC2NodeClasses on startup, %s", err.Error())
+	}
+
+	invalidNodeClasses := lo.FilterMap(nodeClassList.Items, func(nc v1beta1.EC2NodeClass, _ int) (string, bool) {
+		return nc.Name, len(nc.Spec.AMISelectorTerms) == 0
+	})
+	if len(invalidNodeClasses) != 0 {
+		logging.FromContext(ctx).With("ec2nodeclasses", strings.Join(invalidNodeClasses, ",")).Fatalf("detected EC2NodeClasses with un-set AMISelectorTerms. Upgrade your EC2NodeClasses to include AMISelectorTerms to continue.")
```
jonathan-innis, line 61: Should we make this error message more verbose about why this change was made and what's required to make the change here? From just reading this statement, it's not clear to me what I need to do e.g. "Upgrade your EC2NodeClasses". If I was just reading this, I would think that it just means that I need to bump the CRD version, which won't do anything here. Can we consider pointing to docs?

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -456,6 +481,11 @@ spec:
                 this.
               rule: (has(oldSelf.role) && has(self.role)) || (has(oldSelf.instanceProfile)
                 && has(self.instanceProfile))
+            - message: amiFamily must match amiSelectorTerms[].eksOptimized.family
+                or be 'Custom'
+              rule: 'self.amiFamily == ''Custom'' || (self.amiSelectorTerms.exists(x,
+                has(x.eksOptimized)) ? self.amiSelectorTerms.exists(x, has(x.eksOptimized)
+                && x.eksOptimized.family == self.amiFamily) : true)'
```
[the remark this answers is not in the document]
jonathan-innis: Can you remind me what the `exists` operator does again? It's just useful on lists?

pkg/apis/v1beta1/ec2nodeclass_validation.go
```diff
@@ -128,6 +134,14 @@ func (in *SecurityGroupSelectorTerm) validate() (errs *apis.FieldError) {
 }
 
 func (in *EC2NodeClassSpec) validateAMISelectorTerms() (errs *apis.FieldError) {
+	if len(in.AMISelectorTerms) == 0 {
+		errs = errs.Also(apis.ErrMissingOneOf())
```
jonathan-innis, line 138: Does `apis.ErrMissingOneOf()` receive an argument here? Or does it just work out of box since we are already inside of the amiSelectorTerms?

pkg/test/nodeclass.go
```diff
@@ -38,6 +38,13 @@ func EC2NodeClass(overrides ...v1beta1.EC2NodeClass) *v1beta1.EC2NodeClass {
 	if options.Spec.AMIFamily == nil {
 		options.Spec.AMIFamily = &v1beta1.AMIFamilyAL2
 	}
+	if options.Spec.AMISelectorTerms == nil {
+		options.Spec.AMISelectorTerms = append(options.Spec.AMISelectorTerms, v1beta1.AMISelectorTerm{
+			EKSOptimized: &v1beta1.EKSOptimized{
+				Family: v1beta1.AMIFamilyAL2,
```
jonathan-innis, line 44: (consider for a follow-on) Should we update this from AL2 to AL2023 to match our E2E testing

website/content/en/preview/concepts/nodeclasses.md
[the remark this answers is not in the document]
jonathan-innis: > The reason we're just doing this on the validation side and not treating this as an assumed invariant in code is there are users who will be able to circumvent this validation

IMO, we should assume all of the validations happen and continue to assume these invariants in code. Yes, it's technically possible that users disable webhooks and are on an early version of K8s, but we shouldn't build our code around these users. It creates too much sprawl; if we are doing something that assumes this isn't the case in code, I'd argue that we should make changes to ensure our code is simpler by assuming this invariant (a good example of us doing this in other ways is assuming that the NodeClassReference is always set)

website/content/en/preview/concepts/nodeclasses.md
```diff
@@ -1157,15 +1191,40 @@ status:
...
+    name: custom-ami-arm64
+    requirements:
+    - key: kubernetes.io/arch
+      operator: In
+      values:
+      - arm64
+```
 
-Default AMIs resolved from the AL2 AMIFamily:
+However, specifying an `eksOptimized` term may allow Karpenter to discover more requirements.
```
jonathan-innis, line 1221: May allow? Seems like it does allow.

pkg/providers/launchtemplate/suite_test.go
```diff
@@ -1755,6 +1773,7 @@ var _ = Describe("LaunchTemplate Provider", func() {
 			)
 			It("should fail to create launch templates if cluster CIDR is unresolved", func() {
 				awsEnv.LaunchTemplateProvider.ClusterCIDR.Store(nil)
+				fmt.Printf("nodeClass: %s\n", string(lo.Must(json.Marshal(nodeClass))))
```
jonathan-innis, line 1776: Do we still need to print here?

pkg/apis/v1beta1/ec2nodeclass_validation_cel_test.go
```diff
@@ -318,6 +319,129 @@ var _ = Describe("CEL/Validation", func() {
...
+		It("should fail when amiSelectorTerms is nil", func() {
+			nc.Spec.AMISelectorTerms = nil
+			Expect(env.Client.Create(ctx, nc)).ToNot(Succeed())
+		})
+		It("should fail when len(amiSelectorTerms) is zero", func() {
+			nc.Spec.AMISelectorTerms = []v1beta1.AMISelectorTerm{}
+			Expect(env.Client.Create(ctx, nc)).ToNot(Succeed())
+		})
+		DescribeTable(
+			"should succeed with a valid eksOptimized familiy",
```
jonathan-innis, line 331:
```suggestion
			"should succeed with a valid eksOptimized family",
```

nit: spelling

pkg/apis/v1beta1/ec2nodeclass_validation_cel_test.go
```diff
@@ -318,6 +319,129 @@ var _ = Describe("CEL/Validation", func() {
...
+							AMISelectorTerms: []v1beta1.AMISelectorTerm{{
+								EKSOptimized: &v1beta1.EKSOptimized{Family: families[j]},
+							}},
+						},
+					})
+					Expect(env.Client.Create(ctx, nc)).ToNot(Succeed())
+				}
+			}
+		})
+		It("should succed when the eksOptimized family matches amiFamily", func() {
```
jonathan-innis, line 405:
```suggestion
		It("should succeed when the eksOptimized family matches amiFamily", func() {
```

nit: spelling

pkg/apis/v1beta1/ec2nodeclass_validation_webhook_test.go
```diff
@@ -330,6 +331,129 @@ var _ = Describe("Webhook/Validation", func() {
...
+							AMISelectorTerms: []v1beta1.AMISelectorTerm{{
+								EKSOptimized: &v1beta1.EKSOptimized{Family: families[j]},
+							}},
+						},
+					})
+					Expect(nc.Validate(ctx)).ToNot(Succeed())
+				}
+			}
+		})
+		It("should succed when the eksOptimized family matches amiFamily", func() {
```
jonathan-innis, line 417:
```suggestion
		It("should succeed when the eksOptimized family matches amiFamily", func() {
```

nit: spelling

pkg/apis/v1beta1/ec2nodeclass_validation_webhook_test.go
```diff
@@ -330,6 +331,129 @@ var _ = Describe("Webhook/Validation", func() {
...
+		It("should fail when amiSelectorTerms is nil", func() {
+			nc.Spec.AMISelectorTerms = nil
+			Expect(nc.Validate(ctx)).ToNot(Succeed())
+		})
+		It("should fail when len(amiSelectorTerms) is zero", func() {
+			nc.Spec.AMISelectorTerms = []v1beta1.AMISelectorTerm{}
+			Expect(nc.Validate(ctx)).ToNot(Succeed())
+		})
+		DescribeTable(
+			"should succeed with a valid eksOptimized familiy",
```
jonathan-innis, line 343:
```suggestion
			"should succeed with a valid eksOptimized family",
```

nit: spelling

website/content/en/preview/concepts/nodeclasses.md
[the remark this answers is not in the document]
jmdeal: I'd agree but I don't think there's any impact on code complexity whether we consider the invariant or not. The difference is returning one of two slices or appending them together. Since there's no downside from a code complexity / maintainability perspective I think it's really just a question of what is the better UX for users that circumvent the CRD validation, whether that's intentionally or not. IMO it makes more sense to stay consistent with the defined semantics for `amiSelectorTerms` than only include the `eksOptimized` term or all other terms ored together.

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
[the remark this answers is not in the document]
jmdeal: As far as I know it's just for lists though there may be more use cases I'm unaware of. It's pretty much just `lo.Contains`, return true if the predicate applied to any element returns true.

website/content/en/preview/concepts/nodeclasses.md
jmdeal: It depends on the AMI family, which is why I opted for "may allow". For example, AL2023 only discovers architecture which we would be able to discover using other AMI selector terms.

cmd/controller/main.go
jmdeal: I agree, this should probably point to the docs. I still need to make the update to the upgrade guide but I'll have it point there.

pkg/apis/v1beta1/ec2nodeclass_validation.go
jmdeal: Should have it out of the box since the path is injected at the call site for validateAMISelectorTerms.

pkg/providers/amifamily/ami.go
```diff
@@ -117,18 +117,18 @@ func NewDefaultProvider(versionProvider version.Provider, ssm ssmiface.SSMAPI, e
...
-	} else {
-		amis, err = p.getAMIs(ctx, nodeClass.Spec.AMISelectorTerms)
+	amis, err := p.getAMIs(ctx, nodeClass.Spec.AMISelectorTerms)
+	if err != nil {
+		return nil, err
+	}
+	if term, ok := lo.Find(nodeClass.Spec.AMISelectorTerms, func(term v1beta1.AMISelectorTerm) bool {
+		return term.EKSOptimized != nil
+	}); ok {
+		optimizedAMIs, err := p.getOptimizedAMIs(ctx, term.EKSOptimized.Family, options)
```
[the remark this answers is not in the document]
jmdeal: IMO the two functions go about getting AMIs in very different ways and make sense to keep separate. I could see renaming `getAMIs` to be more specific since it currently implies get all AMIs. Having trouble thinking of a good name though, maybe `getFilterAndOwnerSetAMIs`?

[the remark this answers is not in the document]
njtran: That's fair, I'm happy to keep as separate calls. Personally I think having the check in the order how it previously was makes a little bit more code sense to me. If I'm reading line by line, I'd personally be surprised that p.getAMIs() would be succeeded by another type of GetAMI call. Up to you though.

website/content/en/preview/concepts/nodeclasses.md
```diff
@@ -28,8 +28,8 @@ kind: EC2NodeClass
 metadata:
   name: default
 spec:
-  # Required, resolves a default ami and userdata
-  amiFamily: AL2
+  # Required, configures the bootstrapping mode
```
njtran, line 31: wdyt about being more specific here? 
```suggestion
  # Required, configures the userdata that Karpenter injects into the bootstrapping script
```

website/content/en/preview/concepts/nodeclasses.md
```diff
@@ -79,6 +79,11 @@ spec:
     - name: my-ami
     - id: ami-123
 
+    # Automatically select the latest EKS optimzed AMI for the given AMI family.
+    # Note: This term is mutually exclusive and may not be specified with other AMI selector terms
```
njtran, line 83: Let's be more specific here in the notes as well so people read it about the dangers of enabling this in production. I know it's elsewhere, but we should put it everywhere we can.

website/content/en/preview/concepts/nodeclasses.md
```diff
@@ -429,13 +430,17 @@ spec:
 
 ## spec.amiSelectorTerms
 
-AMI Selector Terms are used to configure custom AMIs for Karpenter to use, where the AMIs are discovered through ids, owners, name, and [tags](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html). **When you specify `amiSelectorTerms`, you fully override the default AMIs that are selected on by your EC2NodeClass [`amiFamily`]({{< ref "#specamifamily" >}}).**
+AMI Selector Terms are used to select which AMIs Karpenter will use when launching nodes with this `EC2NodeClass`. You can opt-in to the latest EKS optimized AMIs through an `eksOptimzed` term or AMIs can be discovered through ids, ownders, name, and [tags](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html).
```
njtran, line 433:
```suggestion
AMI Selector Terms are used to select which AMIs Karpenter will use when launching nodes with this `EC2NodeClass`. You can opt-in to the latest EKS optimized AMIs by using the `eksOptimized` term or AMIs can be discovered through ids, owners, name, and [tags](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html).
```

website/content/en/preview/upgrading/upgrade-guide.md
```diff
@@ -39,6 +39,18 @@ WHEN CREATING A NEW SECTION OF THE UPGRADE GUIDANCE FOR NEWER VERSIONS, ENSURE T
...
+{{% /alert %}}
+
+{{% alert title="Warning" color="warning" %}}
+   v0.36.x introduces update to drift that restricts rollback. When rolling back from >=v0.36.0, note that v0.32.9+, v0.33.4+, v0.34.5+, v0.35.4+ are the patch versions that support rollback. If Karpenter is rolled back to an older patch version, Karpenter can potentially drift all the nodes in the cluster.
+{{% /alert %}}
+
+{{% alert title="Warning" color="warning" %}}
+Starting at `v0.37.0`, the `ec2nodeclass.spec.amiSelectorTerms` field is required. When upgrading from previous Karpetner versions, the controller will continue to panic until `amiSelectorTerms` have been specified for all `EC2NodeClasses` applied to the cluster. Please refer to the [`AMISelectorTerms` documentation]({{<ref "../concepts/nodeclasses#spec-amiFamily" >}}) for more details.
```
njtran, line 51:
```suggestion
Starting at `v0.37.0`, the `ec2nodeclass.spec.amiSelectorTerms` field is required. When upgrading from previous Karpenter versions, the controller will continue to panic until `amiSelectorTerms` have been specified for all `EC2NodeClasses` applied to the cluster. Please refer to the [`AMISelectorTerms` documentation]({{<ref "../concepts/nodeclasses#spec-amiFamily" >}}) for more details.
```

website/content/en/preview/concepts/nodeclasses.md
jmdeal: Right now it still configures more than just UserData, but also the default BlockDeviceMappings. I could go either way on adding more depth here or just expanding later in the docs.

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -65,6 +65,23 @@ spec:
                     AMISelectorTerm defines selection logic for an ami used by Karpenter to launch nodes.
                     If multiple fields are used for selection, the requirements are ANDed.
                   properties:
+                    eksOptimized:
```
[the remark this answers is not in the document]
jmdeal: Yes, aligned with `eksOptimized` as of today 👍

pkg/test/nodeclass.go
jonathan-innis: Yep, still probably good to be consistent. Can we backlog this and have it be something small that gets picked up?

website/content/en/preview/concepts/nodeclasses.md
jonathan-innis: Well, Karpenter is always allowed to discover more requirements. It just doesn't if they don't exist. I'm arguing semantics, but I'd generally avoid words like "may" since it make things less clear to people. It leads to the question, "Does it or doesn't it?"

jmdeal: Good point, I'll see what I can come up with to reword. Basically just want to make sure it's clear that not all `eksOptimized` families will be able to discover GPU AMIs and such.

pkg/providers/amifamily/ami.go
```diff
@@ -243,6 +242,8 @@ func GetFilterAndOwnerSets(terms []v1beta1.AMISelectorTerm) (res []FiltersAndOwn
 	idFilter := &ec2.Filter{Name: aws.String("image-id")}
 	for _, term := range terms {
 		switch {
+		case term.EKSOptimized != nil:
```
jonathan-innis, line 245: musing: What if the Optimized AMI just mutating the selector terms in a way that it injected an additional id requirement? Then, you could just use the common function for this and wouldn't have to have separate handling here.

Consider taking this in a follow-on and backlogging it.

website/content/en/preview/tasks/managing-amis.md
```diff
@@ -91,23 +102,17 @@ These examples show two different ways to identify the same AMI:
 
 ```yaml
 amiSelectorTerms:
-- tags:
-    karpenter.sh/discovery: "${CLUSTER_NAME}"
-    environment: prod
-- name: al2023-ami-2023.3.20240219.0-kernel-6.1-x86_64
+  - name: al2023-ami-2023.3.20240219.0-kernel-6.1-x86_64
```
jonathan-innis, line 105: Why did we drop tags here?

website/content/en/preview/tasks/managing-amis.md
```diff
@@ -91,23 +102,17 @@ These examples show two different ways to identify the same AMI:
...
 
 or
 
 ```yaml
 amiSelectorTerms:
-- tags:
-    karpenter.sh/discovery: "${CLUSTER_NAME}"
-    environment: prod
-- id: ami-052c9ea013e6e3567
+  - id: ami-052c9ea013e6e3567
```
jonathan-innis, line 112: Same question here? Are we telling customers not to use tags?

website/content/en/preview/upgrading/upgrade-guide.md
```diff
@@ -39,6 +39,18 @@ WHEN CREATING A NEW SECTION OF THE UPGRADE GUIDANCE FOR NEWER VERSIONS, ENSURE T
 
 ### Upgrading to `0.37.0`+
 
+{{% alert title="Warning" color="warning" %}}
+`0.33.0`+ _only_ supports Karpenter v1beta1 APIs and will not work with existing Provisioner, AWSNodeTemplate or Machine alpha APIs. Do not upgrade to `0.36.0`+ without first [upgrading to `0.32.x`]({{<ref "#upgrading-to-0320" >}}). This version supports both the alpha and beta APIs, allowing you to migrate all of your existing APIs to beta APIs without experiencing downtime.
```
jonathan-innis, line 43:
```suggestion
`0.33.0`+ _only_ supports Karpenter v1beta1 APIs and will not work with existing Provisioner, AWSNodeTemplate or Machine alpha APIs. Do not upgrade to `0.37.0`+ without first [upgrading to `0.32.x`]({{<ref "#upgrading-to-0320" >}}). This version supports both the alpha and beta APIs, allowing you to migrate all of your existing APIs to beta APIs without experiencing downtime.
```

website/content/en/preview/upgrading/upgrade-guide.md
```diff
@@ -39,6 +39,18 @@ WHEN CREATING A NEW SECTION OF THE UPGRADE GUIDANCE FOR NEWER VERSIONS, ENSURE T
...
+{{% /alert %}}
+
+{{% alert title="Warning" color="warning" %}}
+   v0.36.x introduces update to drift that restricts rollback. When rolling back from >=v0.36.0, note that v0.32.9+, v0.33.4+, v0.34.5+, v0.35.4+ are the patch versions that support rollback. If Karpenter is rolled back to an older patch version, Karpenter can potentially drift all the nodes in the cluster.
+{{% /alert %}}
+
+{{% alert title="Warning" color="warning" %}}
```
jonathan-innis, line 50: If we have too many warnings, it starts to all be noise. Are there ways to scope down the number of warning messages that we have here?

jmdeal: I had a similar thought. Ideally, I think we should probably have a common set of warnings at the top of the page. I think the original concern with this apporach was that users would go right to the version and miss these top level warnings. Maybe if we link to these top level warnings from the version specific section that would be a good compromise?

website/content/en/preview/upgrading/upgrade-guide.md
jmdeal: I don't think we should change this, it does apply to every version post v0.36.0.

pkg/providers/amifamily/ami.go
jmdeal: Definitely an interesting idea, but I don't think we'd want to go with that specific for a couple of reasons:
- We shouldn't be calling describe-images if we don't need to. IIUC we call describe-images for ids for two reasons: to validate that the image exists and the user has access and to discover the architecture requirement. Since we already know the requirements and know that these are public AMIs this wouldn't be worth calling.
- We'd have to reinject the requirements that we know from the SSM paths after.

IMO there isn't a massive code-readability benefit and I think these drawbacks are significant enough that it's not worth pursuing. Personally, I think from a readability standpoint the only thing missing is a better collective name for the non-eksOptimized terms so we can be more specific than `getAMIs`.

website/content/en/preview/upgrading/upgrade-guide.md
jmdeal: Going to decouple this for the moment, worth consideration though.

PR #6022, opened 2024-04-11
fix(chart): Updated release to not duplicate AH config

stevehipwell:
Fixes #N/A 

**Description**
This PR fixes the duplication of the Artifact Hub config file for each chart release resulting in the  `artifacthub.io` tag having a new digest after each release. I've also added a README and fixed the CRDs chart config.

There should be an untagged image for each chart release with AH config prior to the current release, these should be deleted unless there is an automated cleanup policy in place.

I switched to this pattern for my own Helm chart logic and have tested the logic there.

- [ ] Yes, issue opened: # 
- [x] No


stevehipwell:
CC @jonathan-innis

hack/release/common.sh
```diff
@@ -110,6 +104,41 @@ publishHelmChart() {
...
+    echo "{}" > "${image_config_path}"
+
+    image_config="${image_config_path}:application/vnd.cncf.artifacthub.config.v1+yaml"
+    media_type="application/vnd.cncf.artifacthub.repository-metadata.layer.v1.yaml"
+    oci_repository="${oci_repo}${helm_chart}"
+    oci_image="${oci_repository}:artifacthub.io"
+
+    old_config_digest="$(crane digest "${oci_image}" || true)"
+
+    if [[ -n "${old_config_digest}" ]]; then
```
jonathan-innis, line 129: Out of curiosity: What's the technical problem if each new release produces a new digest for the ArtifactHub config file?

stevehipwell: The config file digest stays the same, but the manifest file has a build date which obviously changes every time (I can't find a reproducible build flag although explicitly setting the annotation might work). So the outcome is an orphaned image every time the chart releases, as the `artifacthub.io` tag is moved.

hack/release/common.sh
```diff
@@ -110,6 +104,41 @@ publishHelmChart() {
   cosignOciArtifact "${version}" "${commit_sha}" "${build_date}" "${oci_repo}${helm_chart}:${version}@${helm_chart_digest}"
 }
 
+updateAhConfig() {
+  local oci_repo helm_chart ah_config_path image_config_path image_config media_type oci_repository oci_image old_config_digest blob_digest
+
+  oci_repo="${1}"
+  helm_chart="${2}"
+
+  ah_config_path="./charts/${helm_chart}/artifacthub-repo.yml"
```
jonathan-innis, line 113:
```suggestion
  ah_config_path="./charts/${helm_chart}/artifacthub-repo.yaml"
```

This name doesn't match the current filename

jonathan-innis:
Otherwise, changes look reasonable to me

hack/release/common.sh
stevehipwell: @jonathan-innis this should be resolved now.

jonathan-innis (approved):
LGTM 🚀

jonathan-innis:
@stevehipwell Looks like the helm charts still have updated digests whenever they get pushed with the same tag. I haven't done a ton of looking but there's perhaps some work that we can do here to get a reproducible build so that the digest is always the same here.

stevehipwell:
@jonathan-innis do you mean the digest of the Helm chart created from the same raw files and uploaded with a `helm push`? If so then I suspect that this is an ORAS limitation as that's what Helm is using under the hood too. I'll take a look at if I can manually set the OCI metadata for ORAS and solve it at that layer. For Helm I can open a PR there for reproducible builds (although this is likely reproducible pushes.

For Karpenter you (probably) could manually implement the `helm push` with `oras` as I've done for the AH config to solve this today.

PR #6037, opened 2024-04-14
feat(chart): expose health port via svc

owais-maswala:
**Description**

Expose health port via Service

```yaml
apiVersion: v1
kind: Service
metadata:
  name: karpenter
  namespace: dev
  labels:
    helm.sh/chart: karpenter-0.36.0
    app.kubernetes.io/name: karpenter
    app.kubernetes.io/instance: karpenter
    app.kubernetes.io/version: "0.36.0"
    app.kubernetes.io/managed-by: Helm
spec:
  type: ClusterIP
  ports:
    - name: http
      port: 8081
      targetPort: http
      protocol: TCP
    - name: http-metrics
      port: 8000
      targetPort: http-metrics
      protocol: TCP
  selector:
    app.kubernetes.io/name: karpenter
    app.kubernetes.io/instance: karpenter
```

tested on k8s cluster, allows reaching `karpenter.dev.svc.cluster.local:8081/healthz`

- [ ] Yes, issue opened: # 
- [x] No


engedaam (approved):
LGTM 🚀

ellistarn:
What use case does this meet? It's a bit odd to me that we would loadbalance a health port.

owais-maswala:
> What use case does this meet? It's a bit odd to me that we would loadbalance a health port.

I'd like to monitor the health/availability of Karpenter using the Blackbox Exporter/Prometheus Probes

ellistarn:
Is it not possible to scrape the pods directly?

owais-maswala:
> Is it not possible to scrape the pods directly?

Sure its possible, but pods are ephemeral, and would involve using Endpoints to dynamically update our Probe config at regular intervals

Probing the Service instead allows for a stable endpoint and abstraction away from individual pods. I am setting up SLO's as opposed to alerting (which we already have at a pod level)

We have the same setup for 20+ other services on our clusters, but Karpenter is the exception as it does not currently have its health port exposed via the Service

jonathan-innis:
> It's a bit odd to me that we would loadbalance a health port

PR #6042, opened 2024-04-15
feat: Add option to configure extra pod capacity for alternate cnis

### LINKED ISSUE #5780: Karpenter pod capacity exceeds available IPs from ENIs with other CNIs
### Description

**Observed Behavior**: We've observed Karpenter creates nodes with a pod capacity higher than is actually supported by the instance type's ENI's. We observed this most recently with a "g4ad.4xlarge" instance node, which with our Karpenter configuration (reserved ENIs = 1) was assigned a pod capacity of "20" pods, however, we had a pod that was not able to start, and after investigating, it was unable to be assigned an IP from Cilium.

We run clusters with Cilium and use Cilium to replace kube-proxy, so there are no kube-proxy pods that run on our nodes. When we were debugging this behavior we noticed that the calculation Karpenter uses for the node's pod capacity is based on this:

```
#
# Mapping is calculated from AWS EC2 API using the following formula:
# * First IP on each ENI is not used for pods
# * +2 for the pods that use host-networking (AWS CNI and kube-proxy)
#
#   # of ENI * (# of IPv4 per ENI - 1) + 2
#
# https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI
#
```

This does come out to 20 when supplying the variables for the g4ad.4xlarge: `(3 ENIs - 1 Reserved) * (10 - 1) + 2 = 20`. So Karpenter is following this formula, however, if there's no kube-proxy pod and only 1 CNI pod, then would that make this math wrong?

**Expected Behavior**: Expected the node's pod capacity to match the number of pods that can run on the node.

**Reproduction Steps** (Please include YAML): N/A

**Versions**: 
- Chart Version: v0.34.0
- Kubernetes Version (`kubectl version`): v1.26.12-eks-5e0fdde

cnmcavoy:
Fixes #5780

**Description**
 Adds an option to configure extra pod capacity for alternative cni runtimes. Karpenter hardcodes 2 extra pods beyond the ENI IP limits, to account for aws-cni and kube-proxy. This exposes that value as a configurable option while leaving the default untouched.

 We want to be able to use [Cilium w/kube-proxy replacement mode enabled](https://docs.cilium.io/en/stable/network/kubernetes/kubeproxy-free/), and so we need to be able to set this to "1" to account for 1 fewer host networked pod.

`make presubmit`

- [x] Yes, issue opened: #5780
- [x] No

### DIFF
--- a/pkg/operator/options/options.go
+++ b/pkg/operator/options/options.go
@@ -44,6 +44,7 @@ type Options struct {
 	VMMemoryOverheadPercent float64
 	InterruptionQueue       string
 	ReservedENIs            int
+	MaxPodsExtraCapacity    int
 }
 
 func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
@@ -56,6 +57,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.Float64Var(&o.VMMemoryOverheadPercent, "vm-memory-overhead-percent", utils.WithDefaultFloat64("VM_MEMORY_OVERHEAD_PERCENT", 0.075), "The VM memory overhead as a percent that will be subtracted from the total memory for all instance types.")
 	fs.StringVar(&o.InterruptionQueue, "interruption-queue", env.WithDefaultString("INTERRUPTION_QUEUE", ""), "Interruption queue is the name of the SQS queue used for processing interruption events from EC2. Interruption handling is disabled if not specified. Enabling interruption handling may require additional permissions on the controller service account. Additional permissions are outlined in the docs.")
 	fs.IntVar(&o.ReservedENIs, "reserved-enis", env.WithDefaultInt("RESERVED_ENIS", 0), "Reserved ENIs are not included in the calculations for max-pods or kube-reserved. This is most often used in the VPC CNI custom networking setup https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html.")
+	fs.IntVar(&o.MaxPodsExtraCapacity, "max-pods-extra-capacity", env.WithDefaultInt("MAX_PODS_EXTRA_CAPACITY", 2), "Extra pod capacity allocatable on nodes due to host networked pods that do not consume ENIs. The default is for aws-cni and kube-proxy pods, which are not included in the calculations for max-pods or kube-reserved.")
 }
 
 func (o *Options) Parse(fs *coreoptions.FlagSet, args ...string) error {
--- a/pkg/operator/options/options_validation.go
+++ b/pkg/operator/options/options_validation.go
@@ -28,6 +28,7 @@ func (o Options) Validate() error {
 		o.validateVMMemoryOverheadPercent(),
 		o.validateAssumeRoleDuration(),
 		o.validateReservedENIs(),
+		o.validateMaxPodsExtraCapacity(),
 		o.validateRequiredFields(),
 	)
 }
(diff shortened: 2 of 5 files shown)


njtran:
Sorry this closed and slipped through the cracks. I wanted to truly understand the use-cases here before we make this configurable, but I agree that if the default is the same, this should be easy to reason about.

cnmcavoy:
> Sorry this closed and slipped through the cracks. I wanted to truly understand the use-cases here before we make this configurable, but I agree that if the default is the same, this should be easy to reason about.

The hardcoded assumption that there are "2" pods of capacity that can be assumed extra on each node, because of the deamonsets is the core change here we are asking for. With Cilium in kube-proxyless mode, that assumption does not hold.

njtran:
@cnmcavoy does this solve your issue? https://github.com/kubernetes-sigs/karpenter/pull/1305

@ellistarn can add more details here, but this could potentially add this as an overlay on the node. 

Do you think that this expected pod overhead differs by instance type? or should the overhead be the same across the cluster since it's related to the networking setup for the cluster?

Bryce-Soghigian:
I have some relevant context as we just had the discussion as to whether or not we should lower the ip configurations we generate in the azure provider based on the Static Host Networking addons.

I was just looking at IP allocation for static addons in our provider. We had a similar value "StaticAzureCNIHostNetworkAddons", and we decided to just respect the max pods values we pass into kubelet and not add the decifit at all. Im not sure how aws provider generates ip configurations or how eni works at all, but maybe this is a useful datapoint.  

See this commit for more [details](https://github.com/Azure/karpenter-provider-azure/pull/365/commits/cba91f6e300f9445615ce20c0abd47be42f3fa45)

cc: @paulgmiller who is our networking lead at AKS who might have more context.

Bryce-Soghigian:
" NOTE: There are some pods that we see on every node for networking like kube-proxy, and ip-masq-agent). We could be subtracting ips for these pods, and we do for AKSEngine. In the case of AKS However, we have a dynamic number of host networking addon pods. So we are explicitly choosing to not subtract these IPs like AKS engine and some parts of AKS do. The AKS networking team regrets adding this behavior and karpenter is a chance to do better so we shall."

cnmcavoy:
@Bryce-Soghigian 

Interesting, if I am following your context with AKS, you are suggesting Karpenter should remove the hardcoded static limit entirely and present the stated ec2 pod capacity. The consequences of that would be that the pod capacity would not match the AWS eni capacity of the node, for those of us using ENI IPAM. What do you do when pods try to start on a node and can't receive an IP, because the pod capacity != CNI capacity? I opened this ticket because we hit this in production @ Indeed.

@njtran 

I do not see how node-overlays addresses this bug report. As a user of Karpenter, I want the pod capacity of node's to be correct. I do not want to set my own custom pod capacity. I want the node pod capacity to match the number of ENI's that can be attached to a node. Currently it is not because there is hardcoded magic numbers in the math that assumes too much. Does this request make sense to you?

ellistarn:
There's a broader challenge here that host-networked pods are not correctly accounted for. We make a bad assumption that there will be 2 (e.g. kubeproxy/vpccni), but in reality, there are often far more than this due to other daemons, which artificially restricts nodes from having fewer pods than they could.

Another way to achieve your feature request is via a negative overhead [node overlay](https://github.com/kubernetes-sigs/karpenter/pull/1305).

```
kind: NodeOverlay
metadata:
  name: host-networking
spec:
  overhead:
    pods: -1
```

This will select all nodes, and tell the scheduler that there's an additional pod slot available

israelbgf:
> There's a broader challenge here that host-networked pods are not correctly accounted for. We make a bad assumption that there will be 2 (e.g. kubeproxy/vpccni), but in reality, there are often far more than this due to other daemons, which artificially restricts nodes from having fewer pods than they could.
> 
> Another way to achieve your feature request is via a negative overhead [node overlay](https://github.com/kubernetes-sigs/karpenter/pull/1305).
> 
> ```
> kind: NodeOverlay
> metadata:
>   name: host-networking
> spec:
>   overhead:
>     pods: -1
> ```
> 
> This will select all nodes, and tell the scheduler that there's an additional pod slot available.

How can I achieve this with the current state of NodeOverlays? Based on the current docs I can't see a way to change the pod count for scheduling purposes. @ellistarn

PR #6057, opened 2024-04-18
chore: Add an Subnet Controller to Asynchronously Hydrate Subnet Data

engedaam:
Fixes #N/A 

**Description**
- Asynchronously hydrate with subnet data. The controller will attempt to refresh the data every 5 minutes, and implement a exponential backoff for delay: 1ms and max: 1000s

- `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No


engedaam (commented):
/karpenter snapshot

pkg/providers/subnet/subnet.go
```diff
@@ -64,41 +65,59 @@ func NewDefaultProvider(ec2api ec2iface.EC2API, cache *cache.Cache) *DefaultProv
...
-	defer p.Unlock()
+	p.RLock()
+	defer p.RUnlock()
+
 	filterSets := getFilterSets(nodeClass.Spec.SubnetSelectorTerms)
 	if len(filterSets) == 0 {
 		return []*ec2.Subnet{}, nil
 	}
-	hash, err := hashstructure.Hash(filterSets, hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: true})
+	hash, err := subnetHash(filterSets)
```
jonathan-innis, line 75: When can this hashing function fail? Is this a case where it only fails when there is a developer error and we could just use `lo.Must` here to panic if it fails?

pkg/controllers/providers/subnet/controller.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func NewController(kubeClient client.Client, subnetProvider subnet.Provider) corecontroller.Controller {
+	return corecontroller.Typed[*v1beta1.EC2NodeClass](kubeClient, &Controller{
+		subnetProvider: subnetProvider,
+	})
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
+	return reconcile.Result{RequeueAfter: 5 * time.Minute}, c.subnetProvider.UpdateSubnets(ctx, nodeClass)
```
jonathan-innis, line 46: controller-runtime will yell at you if you return both a valid result and an error at the same time

pkg/providers/subnet/subnet.go
```diff
@@ -64,41 +65,59 @@ func NewDefaultProvider(ec2api ec2iface.EC2API, cache *cache.Cache) *DefaultProv
 }
 
 func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) ([]*ec2.Subnet, error) {
```
jonathan-innis, line 67: How do we make sure that we don't hold subnets in our cache forever when there is a delete operation that has been performed on a NodeClass? Thinking about it a little more -- if you hold the subnets in the cache like this, you basically have to do tracking to see which NodeClasses own the entry and respond to the delete operations for the NodeClasses, which sounds like a lot of work and sounds like it could create complicated semantics.

I'm thinking your original thought about relying on the status outright right now may actually be fine -- and we can move over to readiness conditions shortly. If we don't have subnets resolve -- like you were saying, we can just return no instance types and let the scheduler know that we can't use this NodePool. You could even consider the EKSNodeClass instanceProfile, security group, and subnet resolution as gates to return back valid values from the instance types call.

jonathan-innis: The only problem then becomes how you ensure that if you are consistently re-reconciling the NodeClass -- that you still don't call the API too frequently when there is a request for data. Given that there is back-off built-into the NodeClass controller out-of-the-box are we good to just have each controller re-reconcile its own object and then hit the API cache to prevent duplicate requests (same filters on AMIs, subnets, or security groups) from causing a higher number of API requests.

engedaam (commented):
/karpenter snapshot

pkg/providers/instancetype/instancetype.go
```diff
@@ -101,27 +101,26 @@ func (p *DefaultProvider) List(ctx context.Context, kc *corev1beta1.KubeletConfi
 	defer p.muInstanceTypeInfo.RUnlock()
 	defer p.muInstanceTypeOfferings.RUnlock()
 
+	if kc == nil {
+		kc = &corev1beta1.KubeletConfiguration{}
+	}
+	if nodeClass == nil {
```
jonathan-innis, line 107: musing: Is there a way that we can remove this nil check in this List call long-term

pkg/providers/instancetype/suite_test.go
```diff
@@ -155,27 +155,81 @@ var _ = Describe("InstanceTypeProvider", func() {
 				},
 			},
 		})
-		nodeClass.Status.SecurityGroups = []v1beta1.SecurityGroup{
-			{
-				ID: "sg-test1",
+
+		nodeClass.Status = v1beta1.EC2NodeClassStatus{
+			InstanceProfile: "test-profile",
+			AMIs: []v1beta1.AMI{
```
jonathan-innis, line 161: Do we need the AMIs section here?

pkg/providers/launchtemplate/suite_test.go
```diff
@@ -146,15 +147,64 @@ var _ = Describe("LaunchTemplate Provider", func() {
 				},
 			},
 		})
-		nodeClass.Status.SecurityGroups = []v1beta1.SecurityGroup{
-			{
-				ID: "sg-test1",
+		nodeClass.Status = v1beta1.EC2NodeClassStatus{
+			InstanceProfile: "test-profile",
+			AMIs: []v1beta1.AMI{
```
jonathan-innis, line 152: Do we need the AMIs in the status here or should this be added in the AMI PR?

pkg/providers/subnet/subnet.go
```diff
@@ -107,54 +107,42 @@ func (p *DefaultProvider) CheckAnyPublicIPAssociations(ctx context.Context, node
...
 }
 
 // ZonalSubnetsForLaunch returns a mapping of zone to the subnet with the most available IP addresses and deducts the passed ips from the available count
-func (p *DefaultProvider) ZonalSubnetsForLaunch(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, instanceTypes []*cloudprovider.InstanceType, capacityType string) (map[string]*ec2.Subnet, error) {
-	subnets, err := p.List(ctx, nodeClass)
-	if err != nil {
-		return nil, err
-	}
+func (p *DefaultProvider) ZonalSubnetsForLaunch(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, instanceTypes []*cloudprovider.InstanceType, capacityType string) (map[string]v1beta1.Subnet, error) {
+	subnets := nodeClass.Status.Subnets
```
jonathan-innis, line 119: Why reassign to another variable here?

pkg/providers/subnet/subnet.go
```diff
@@ -107,54 +107,42 @@ func (p *DefaultProvider) CheckAnyPublicIPAssociations(ctx context.Context, node
...
 	sort.Slice(subnets, func(i, j int) bool {
-		iIPs := aws.Int64Value(subnets[i].AvailableIpAddressCount)
-		jIPs := aws.Int64Value(subnets[j].AvailableIpAddressCount)
-		// override ip count from ec2.Subnet if we've tracked launches
-		if ips, ok := p.inflightIPs[*subnets[i].SubnetId]; ok {
-			iIPs = ips
-		}
-		if ips, ok := p.inflightIPs[*subnets[j].SubnetId]; ok {
-			jIPs = ips
+		if p.inflightIPs[nodeClass.Status.Subnets[i].ID] != p.inflightIPs[nodeClass.Status.Subnets[j].ID] {
```
jonathan-innis, line 128: Maybe add a comment that we assume inflightIPs was pre-hydrated before this

pkg/providers/subnet/subnet.go
```diff
@@ -87,7 +87,7 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
 		}
 		for i := range output.Subnets {
 			subnets[lo.FromPtr(output.Subnets[i].SubnetId)] = output.Subnets[i]
-			delete(p.inflightIPs, lo.FromPtr(output.Subnets[i].SubnetId)) // remove any previously tracked IP addresses since we just refreshed from EC2
+			p.inflightIPs[lo.FromPtr(output.Subnets[i].SubnetId)] = lo.FromPtr(output.Subnets[i].AvailableIpAddressCount) // remove any previously tracked IP addresses since we just refreshed from EC2
```
jonathan-innis, line 90: Throw a TODO on this that says, "we know that we leak here, but we are accepting it because of..."

pkg/providers/subnet/subnet.go
```diff
@@ -107,54 +107,42 @@ func (p *DefaultProvider) CheckAnyPublicIPAssociations(ctx context.Context, node
...
-		iIPs := aws.Int64Value(subnets[i].AvailableIpAddressCount)
-		jIPs := aws.Int64Value(subnets[j].AvailableIpAddressCount)
-		// override ip count from ec2.Subnet if we've tracked launches
-		if ips, ok := p.inflightIPs[*subnets[i].SubnetId]; ok {
-			iIPs = ips
-		}
-		if ips, ok := p.inflightIPs[*subnets[j].SubnetId]; ok {
-			jIPs = ips
+		if p.inflightIPs[nodeClass.Status.Subnets[i].ID] != p.inflightIPs[nodeClass.Status.Subnets[j].ID] {
+			return p.inflightIPs[nodeClass.Status.Subnets[i].ID] < p.inflightIPs[nodeClass.Status.Subnets[j].ID]
```
jonathan-innis, line 129: Can we call this `availableIPs`?

pkg/providers/subnet/subnet.go
```diff
@@ -107,54 +107,42 @@ func (p *DefaultProvider) CheckAnyPublicIPAssociations(ctx context.Context, node
...
 	}
 	for _, subnet := range zonalSubnets {
-		predictedIPsUsed := p.minPods(instanceTypes, *subnet.AvailabilityZone, capacityType)
-		prevIPs := *subnet.AvailableIpAddressCount
-		if trackedIPs, ok := p.inflightIPs[*subnet.SubnetId]; ok {
-			prevIPs = trackedIPs
-		}
-		p.inflightIPs[*subnet.SubnetId] = prevIPs - predictedIPsUsed
+		predictedIPsUsed := p.minPods(instanceTypes, subnet.Zone, capacityType)
+		p.inflightIPs[subnet.ID] = p.inflightIPs[subnet.ID] - predictedIPsUsed
```
jonathan-innis, line 138:
```suggestion
		p.inflightIPs[subnet.ID] -= predictedIPsUsed
```

nit

pkg/providers/subnet/subnet.go
```diff
@@ -193,19 +181,19 @@ func (p *DefaultProvider) UpdateInflightIPs(createFleetInput *ec2.CreateFleetInp
...
+		originalSubnet, ok := lo.Find(subnets, func(subnet v1beta1.Subnet) bool {
+			return subnet.ID == *cachedSubnet.SubnetId
 		})
 		if !ok {
 			continue
 		}
 		// If the cached subnet IP address count hasn't changed from the original subnet used to
 		// launch the instance, then we need to update the tracked IPs
-		if *originalSubnet.AvailableIpAddressCount == *cachedSubnet.AvailableIpAddressCount {
+		if p.inflightIPs[originalSubnet.ID] == *cachedSubnet.AvailableIpAddressCount {
```
jonathan-innis, line 192: I think this is wrong (though I need to understand it better). The old code checked if the IP addresses in the cache are the same as the subnets that we stored. This code basically does the opposite of that. I know that this isn't correct, but I'm not sure what the correct behavior is here. We should discuss more and try to work it out.

engedaam: Discussed offline, and we will be keeping the subnet logic consistent.

pkg/providers/instancetype/suite_test.go
engedaam: This is to help with merge conflict, but I can remove it

pkg/providers/instancetype/instancetype.go
engedaam: We will need to make sure that users don't pass in nil values, but I can see us getting rid of this in the future

pkg/cache/cache.go
```diff
@@ -31,6 +31,8 @@ const (
 	InstanceTypesAndZonesTTL = 5 * time.Minute
 	// InstanceProfileTTL is the time before we refresh checking instance profile existence at IAM
 	InstanceProfileTTL = 15 * time.Minute
+	// ExpandedSubnetTTL is time to allow reconciliation to update the subnet data needed before expiration
```
jonathan-innis, line 34: I feel like this name is a little strange. Why not just create variables that match the cache name? e.g. `availableIPAddressTTL` and `associatePublicIPAddressTTL`

jonathan-innis, line 34: Is this description correct? This cache timeout is actually the time that we hold this data in the cache before we drop it. This will always be updated more frequently unless there are no node classes referencing a subnet anymore OR there is an issue with DescribeSubnets

pkg/test/environment.go
```diff
@@ -58,14 +58,16 @@ type Environment struct {
...
-	SubnetCache               *cache.Cache
-	SecurityGroupCache        *cache.Cache
-	InstanceProfileCache      *cache.Cache
+	EC2Cache                      *cache.Cache
+	KubernetesVersionCache        *cache.Cache
+	InstanceTypeCache             *cache.Cache
+	UnavailableOfferingsCache     *awscache.UnavailableOfferings
+	LaunchTemplateCache           *cache.Cache
+	SubnetCache                   *cache.Cache
+	SubnetIPAdressCache           *cache.Cache
```
jonathan-innis, line 67:
```suggestion
	AvailableIPAdressCache           *cache.Cache
```

nit: To match naming?

pkg/providers/subnet/subnet.go
```diff
@@ -103,29 +111,48 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
 
 // CheckAnyPublicIPAssociations returns a bool indicating whether all referenced subnets assign public IPv4 addresses to EC2 instances created therein
 func (p *DefaultProvider) CheckAnyPublicIPAssociations(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (bool, error) {
-	subnets, err := p.List(ctx, nodeClass)
-	if err != nil {
-		return false, err
+	subnets := map[string]bool{}
+	for _, subnet := range nodeClass.Status.Subnets {
+		if subnetAssociatePublicIP, ok := p.associatePublicIPAddressCache.Get(subnet.ID); ok {
+			subnets[subnet.ID] = subnetAssociatePublicIP.(bool)
```
jonathan-innis, line 117: Do you need to build this map like this? Can you just get items from the cache and just return early without building another data structure?

pkg/providers/subnet/subnet.go
```diff
@@ -103,29 +111,48 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
-		return nil, err
-	}
-	if len(subnets) == 0 {
-		return nil, fmt.Errorf("no subnets matched selector %v", nodeClass.Spec.SubnetSelectorTerms)
-	}
 	p.Lock()
 	defer p.Unlock()
-	// sort subnets in ascending order of available IP addresses and populate map with most available subnet per AZ
+
+	if len(nodeClass.Status.Subnets) == 0 {
```
jonathan-innis, line 134: Does this being below the lock change anything?

pkg/providers/subnet/subnet.go
```diff
@@ -103,29 +111,48 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
 	zonalSubnets := map[string]*ec2.Subnet{}
+
+	availableIPAddress := map[string]int64{}
+	for _, subnet := range nodeClass.Status.Subnets {
+		if subnetAvailableIP, ok := p.availableIPAddressCache.Get(subnet.ID); ok {
+			availableIPAddress[subnet.ID] = subnetAvailableIP.(int64)
+		}
+	}
+
+	subnets := lo.Map(nodeClass.Status.Subnets, func(s v1beta1.Subnet, _ int) *ec2.Subnet {
```
jonathan-innis, line 147: Is it worth creating a tightly scoped struct here to store only the details we need rather than using an ec2.Subnet? The only concern that I have here is that someone not really reading the code could think that this has all of the data in the ec2.Subnet, when it reality we are only injecting the data that we care about

pkg/providers/subnet/subnet.go
```diff
@@ -136,6 +163,9 @@ func (p *DefaultProvider) ZonalSubnetsForLaunch(ctx context.Context, nodeClass *
 		if ips, ok := p.inflightIPs[*subnets[j].SubnetId]; ok {
 			jIPs = ips
 		}
+		if iIPs == jIPs {
+			return lo.FromPtr(subnets[i].SubnetId) < lo.FromPtr(subnets[j].SubnetId)
+		}
 		return iIPs < jIPs
 	})
 	for _, subnet := range subnets {
```
jonathan-innis, line 171: I wonder if rather than sorting first and then iterating through everything and assigning the top subnet to the AZ if we just iterate once and then check if the existing subnet has more IPs. If you do this you should also be able to just use a `map[string]int64` to store the IP address count since you can just grab it out of the map when you are iterating through the subnets rather than building a separate data structure

pkg/providers/subnet/subnet.go
```diff
@@ -182,26 +212,26 @@ func (p *DefaultProvider) UpdateInflightIPs(createFleetInput *ec2.CreateFleetInp
...
+	// Aggregate all the cached subnets ip address count
+	cachedIPAddressCount := lo.MapEntries(p.availableIPAddressCache.Items(), func(k string, v cache.Item) (string, int64) {
+		return k, v.Object.(int64)
+	})
 
 	// Update the inflight IP tracking of subnets stored in the cache that have not be synchronized since the initial
 	// deduction of IP addresses before the instance launch
-	for _, cachedSubnet := range cachedSubnets {
-		if !lo.Contains(subnetIDsToAddBackIPs, *cachedSubnet.SubnetId) {
+	for cachedSubnetID, cachedIPAddressCount := range cachedIPAddressCount {
```
jonathan-innis, line 222: Should we rename this map? You're calling the key the same thing as the map, which is going to get confusing

pkg/providers/subnet/subnet.go
engedaam: It won't change much, but it would be better to exit on loops sooner if they will fail this check

pkg/providers/subnet/subnet.go
engedaam: But we would only want to track it for the top subnet in a zone, not for all subnets in every zone. I'm not seeing how this would be tracked?

pkg/providers/subnet/subnet.go
```diff
@@ -103,58 +117,74 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
+func (p *DefaultProvider) ZonalSubnetsForLaunch(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, instanceTypes []*cloudprovider.InstanceType, capacityType string) (map[string]*Subnet, error) {
+	if len(nodeClass.Status.Subnets) == 0 {
 		return nil, fmt.Errorf("no subnets matched selector %v", nodeClass.Spec.SubnetSelectorTerms)
 	}
+
 	p.Lock()
 	defer p.Unlock()
+
+	zonalSubnets := map[string]*Subnet{}
+	availableIPAddress := map[string]int64{}
```
jonathan-innis, line 140:
```suggestion
	availableIPAddressCount := map[string]int64{}
```

nit

pkg/cloudprovider/suite_test.go
```diff
@@ -964,28 +1006,34 @@ var _ = Describe("CloudProvider", func() {
...
 			awsEnv.EC2API.DescribeSubnetsOutput.Set(&ec2.DescribeSubnetsOutput{Subnets: []*ec2.Subnet{
 				{SubnetId: aws.String("test-subnet-1"), AvailabilityZone: aws.String("test-zone-1a"), AvailableIpAddressCount: aws.Int64(10),
 					Tags: []*ec2.Tag{{Key: aws.String("Name"), Value: aws.String("test-subnet-1")}}},
 				{SubnetId: aws.String("test-subnet-2"), AvailabilityZone: aws.String("test-zone-1a"), AvailableIpAddressCount: aws.Int64(11),
 					Tags: []*ec2.Tag{{Key: aws.String("Name"), Value: aws.String("test-subnet-2")}}},
 			}})
+			controller := status.NewController(env.Client, awsEnv.SubnetProvider, awsEnv.SecurityGroupProvider, awsEnv.AMIProvider, awsEnv.InstanceProfileProvider, awsEnv.LaunchTemplateProvider)
 			nodePool.Spec.Template.Spec.Kubelet = &corev1beta1.KubeletConfiguration{MaxPods: aws.Int32(1)}
 			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(nodeClass))
```
jonathan-innis, line 1036: sanity check: These reconciles are definitely needed and we can't directly hydrate the status?

engedaam: This is so that we can hydrate the status and sync up all the subnet caches

pkg/providers/subnet/subnet.go
engedaam: But we expect the zonalSubnets to contain subnets that have the most available IPs, If we don't sort first including the inflight IPs, how would know which subnet to pick for each zone? From my understand the above code, just adds all the subnets in the zonalSubnets map

pkg/providers/instancetype/instancetype.go
jonathan-innis: I'm wondering if we can just track down the calls here and determine that we will never get a `nil` call here. I think that I originally built it this way so that we can still resolve InstanceTypes with default values, even if we don't get any offerings, but it may not be worth this configuration or handling anymore.

Anyways, we can punt this to a separate change and come back to it later.

pkg/cloudprovider/suite_test.go
```diff
@@ -113,7 +115,41 @@ var _ = Describe("CloudProvider", func() {
 	var nodePool *corev1beta1.NodePool
 	var nodeClaim *corev1beta1.NodeClaim
 	var _ = BeforeEach(func() {
-		nodeClass = test.EC2NodeClass()
+		nodeClass = test.EC2NodeClass(
```
jonathan-innis, line 118: optional: Did we consider automatically using these values when we are using the default EC2NodeClass from `test.EC2NodeClass`?

pkg/cloudprovider/suite_test.go
```diff
@@ -964,28 +1009,34 @@ var _ = Describe("CloudProvider", func() {
 			Expect(foundNonGPULT).To(BeTrue())
 		})
 		It("should launch instances into subnet with the most available IP addresses", func() {
+			awsEnv.SubnetCache.Flush()
```
jonathan-innis, line 1012: Do we need this if the environment is getting reset anyway at the end of the test?

pkg/cloudprovider/suite_test.go
```diff
@@ -964,28 +1009,34 @@ var _ = Describe("CloudProvider", func() {
...
 			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			ExpectReconcileSucceeded(ctx, controller, client.ObjectKeyFromObject(nodeClass))
 			pod := coretest.UnschedulablePod(coretest.PodOptions{NodeSelector: map[string]string{v1.LabelTopologyZone: "test-zone-1a"}})
 			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
 			ExpectScheduled(ctx, env.Client, pod)
 			createFleetInput := awsEnv.EC2API.CreateFleetBehavior.CalledWithInput.Pop()
 			Expect(fake.SubnetsFromFleetRequest(createFleetInput)).To(ConsistOf("test-subnet-2"))
 		})
 		It("should launch instances into subnet with the most available IP addresses in-between cache refreshes", func() {
+			awsEnv.SubnetCache.Flush()
```
jonathan-innis, line 1029: Same comment here

pkg/providers/subnet/subnet.go
```diff
@@ -103,58 +116,64 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
+			currentZonalSubnetIPAddressCount := v.AvailableIPAddressCount
+			newZonalSubnetIPAddressCount := availableIPAddressCount[subnet.ID]
+			if ips, ok := p.inflightIPs[v.ID]; ok {
+				currentZonalSubnetIPAddressCount = ips
+			}
+			if ips, ok := p.inflightIPs[subnet.ID]; ok {
+				newZonalSubnetIPAddressCount = ips
+			}
+
+			if currentZonalSubnetIPAddressCount < newZonalSubnetIPAddressCount {
```
jonathan-innis, line 155: This statement here on the one on L159 are the same. Is there a way to collapse this into a single line that we call once in either scenario.

jonathan-innis (approved):
Nice work! Just some optional comments.

pkg/cloudprovider/suite_test.go
engedaam: For the cloud provider testing we hydrate the subnet cache with the default subnets in the fake EC2 API, so we will need remove those and add new ones

pkg/cloudprovider/suite_test.go
engedaam: I didn't want it to be surprising to developers that the status is already hydrated and they will need update the API and status.

jonathan-innis (approved):
LGTM 🚀

pkg/cloudprovider/suite_test.go
engedaam: Discussed Offline, will add in a follow-up

engedaam (commented):
/karpenter snapshot

PR #4840, opened 2023-10-15
chore: Passthrough new helm values as environment variables

jonathan-innis:
Fixes #N/A 

**Description**

This PR supports the new Karpenter environment variables that are going to be supported for configuring Karpenter as part of the v1beta1 drop of the Karpenter ConfigMap. The associated changes where the environment variables are implemented are: https://github.com/aws/karpenter-core/pull/541 and https://github.com/aws/karpenter/pull/4789

This PR also collapses all of the controller-wide settings in the `settings.aws` values block into `settings`. The old values will be supported up to v0.33.0, at which point they will be dropped.

**How was this change tested?**

`make apply`
`/karpenter snapshot`

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates 
- [ ] Yes, issue opened: # 
- [x] No

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6062, opened 2024-04-19
fix: Update docs about queue name change

maxforasteiro:
(it doesn't) Fix https://github.com/eksctl-io/eksctl/issues/7697 (but here's an issue caused by this)

**Description**

Seems to have been added in https://github.com/aws/karpenter-provider-aws/pull/4840
I'm just adding the docs because we spent one month having to deal with spot nodes being shutdown in 50s 

no test needed

- [ ] Yes, issue opened: # 
- [ ] No


website/content/en/docs/upgrading/upgrade-guide.md
```diff
@@ -103,6 +103,7 @@ This version includes **dual support** for both alpha and beta APIs to ensure th
 Note that if you are rolling back after upgrading to `0.32.0`, note that `0.31.4` is the only version that supports handling rollback after you have deployed the v1beta1 APIs to your cluster.
 {{% /alert %}}
 
+* Karpenter now uses `settings.InterruptionQueue` instead of `settings.aws.InterruptionQueueName` in its helm chart. The CLI argument also changed to `--interruption-queue`.
```
jonathan-innis, line 106: Looks like we just added this one section to the v0.32 section of the upgrade docs on the latest docs version. This change was covered in this section of the migration guide docs (https://karpenter.sh/v0.32/upgrading/v1beta1-migration/#global-settings). Given that this was the only "true" name change, I'm fine with including this in the upgrade note callouts but we should probably add this to every upgrade guide version so that it pops up in the v0.32 section on every version

maxforasteiro: I was also thinking: does it make sense to generate a warning if a nodepool is configured with `capacity-type: spot` and no `interruptionQueue` set?

maxforasteiro: added in https://github.com/aws/karpenter-provider-aws/pull/6062/commits/45e03b90f870f93f2235657eb02528effd562c3e, let me know what you think, I can revert if you don't like it 😄

pkg/providers/instance/instance.go
```diff
@@ -224,6 +224,9 @@ func (p *DefaultProvider) launchInstance(ctx context.Context, nodeClass *v1beta1
 		},
 	}
 	if capacityType == corev1beta1.CapacityTypeSpot {
+		if options.FromContext(ctx).InterruptionQueue == "" {
```
jonathan-innis, line 227: I don't think that we should be gating the use of spot based on if someone has the interruption queue enabled or not. I'd imagine if we wanted to do this, that we would just require the interruption queue to be set outright since part of Karpenter's value proposition is that it makes it so easy to use spot out of the box. 

Additionally, you can run spot without using the interruption queue (not sure I'd recommend it, but you definitely can, and some people use NTH (Node Termination Handler) rather than using Karpenter's in-built interruption handling.

website/content/en/docs/upgrading/upgrade-guide.md
jonathan-innis: Made a comment on this below next to the change. Probably best to decouple this change from this docs update; I'm also not sure that it's a change that we would want to take. I think there are scenarios where you may not want to run the interruption queue and you are still comfortable using spot.

pkg/providers/instance/instance.go
maxforasteiro: Bear in mind that this is just a Warn message. I'll remove this from the PR for now. Thanks for the quick feedback, really appreciate

jonathan-innis (approved):
LGTM 🚀

PR #6069, opened 2024-04-21
chore: Use Security Group Status Controller to Asynchronously Hydrate Security Group  Data

engedaam:
Fixes #N/A 

**Description**
- Asynchronously hydrate with Security Group data. The controller will attempt to refresh the data every 5 minutes, and implement a exponential backoff for delay: 1ms and max: 1000s

- `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/cloudprovider/drift.go
+++ b/pkg/cloudprovider/drift.go
@@ -54,7 +54,7 @@ func (c *CloudProvider) isNodeClassDrifted(ctx context.Context, nodeClaim *corev
 	if err != nil {
 		return "", fmt.Errorf("calculating ami drift, %w", err)
 	}
-	securitygroupDrifted, err := c.areSecurityGroupsDrifted(ctx, instance, nodeClass)
+	securitygroupDrifted, err := c.areSecurityGroupsDrifted(instance, nodeClass)
 	if err != nil {
 		return "", fmt.Errorf("calculating securitygroup drift, %w", err)
 	}
@@ -118,14 +118,10 @@ func (c *CloudProvider) isSubnetDrifted(ctx context.Context, instance *instance.
 
 // Checks if the security groups are drifted, by comparing the security groups returned from the SecurityGroupProvider
 // to the ec2 instance security groups
-func (c *CloudProvider) areSecurityGroupsDrifted(ctx context.Context, ec2Instance *instance.Instance, nodeClass *v1beta1.EC2NodeClass) (cloudprovider.DriftReason, error) {
-	securitygroup, err := c.securityGroupProvider.List(ctx, nodeClass)
-	if err != nil {
-		return "", err
-	}
-	securityGroupIds := sets.New(lo.Map(securitygroup, func(sg *ec2.SecurityGroup, _ int) string { return aws.StringValue(sg.GroupId) })...)
+func (c *CloudProvider) areSecurityGroupsDrifted(ec2Instance *instance.Instance, nodeClass *v1beta1.EC2NodeClass) (cloudprovider.DriftReason, error) {
+	securityGroupIds := sets.New(lo.Map(nodeClass.Status.SecurityGroups, func(sg v1beta1.SecurityGroup, _ int) string { return sg.ID })...)
 	if len(securityGroupIds) == 0 {
-		return "", fmt.Errorf("no security groups are discovered")
+		return "", fmt.Errorf("no security groups are present in the status")
 	}
 
 	if !securityGroupIds.Equal(sets.New(ec2Instance.SecurityGroupIDs...)) {
--- a/pkg/cloudprovider/suite_test.go
+++ b/pkg/cloudprovider/suite_test.go
@@ -138,6 +138,20 @@ var _ = Describe("CloudProvider", func() {
 				},
 			},
 		})
+		nodeClass.Status.SecurityGroups = []v1beta1.SecurityGroup{
+			{
+				ID:   "sg-test1",
+				Name: "securityGroup-test1",
+			},
+			{
+				ID:   "sg-test2",
+				Name: "securityGroup-test2",
+			},
+			{
+				ID:   "sg-test3",
+				Name: "securityGroup-test3",
+			},
+		}
 		Expect(awsEnv.InstanceTypesProvider.UpdateInstanceTypes(ctx)).To(Succeed())
 		Expect(awsEnv.InstanceTypesProvider.UpdateInstanceTypeOfferings(ctx)).To(Succeed())
 	})
@@ -586,6 +600,11 @@ var _ = Describe("CloudProvider", func() {
 					},
 				},
 			})
+			nodeClass.Status.SecurityGroups = []v1beta1.SecurityGroup{
+				{
+					ID: validSecurityGroup,
+				},
+			}
 			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
 			instanceTypes, err := cloudProvider.GetInstanceTypes(ctx, nodePool)
 			Expect(err).ToNot(HaveOccurred())
@@ -671,7 +690,8 @@ var _ = Describe("CloudProvider", func() {
 			Expect(isDrifted).To(BeEmpty())
 		})
 		It("should return an error if the security groups are empty", func() {
-			awsEnv.EC2API.DescribeSecurityGroupsOutput.Set(&ec2.DescribeSecurityGroupsOutput{SecurityGroups: []*ec2.SecurityGroup{}})
+			nodeClass.Status.SecurityGroups = []v1beta1.SecurityGroup{}
+			ExpectApplied(ctx, env.Client, nodeClass)
 			// Instance is a reference to what we return in the GetInstances call
 			instance.SecurityGroups = []*ec2.GroupIdentifier{{GroupId: aws.String(fake.SecurityGroupID())}}
 			_, err := cloudProvider.IsDrifted(ctx, nodeClaim)
@@ -692,18 +712,17 @@ var _ = Describe("CloudProvider", func() {
 			Expect(isDrifted).To(Equal(cloudprovider.SecurityGroupDrift))
 		})
 		It("should return drifted if more security groups are present than instance security groups then discovered from nodeclass", func() {
-			awsEnv.EC2API.DescribeSecurityGroupsOutput.Set(&ec2.DescribeSecurityGroupsOutput{
-				SecurityGroups: []*ec2.SecurityGroup{
-					{
-						GroupId:   aws.String(validSecurityGroup),
-						GroupName: aws.String("test-securitygroup"),
-					},
-					{
-						GroupId:   aws.String(fake.SecurityGroupID()),
-						GroupName: aws.String("test-securitygroup"),
-					},
+			nodeClass.Status.SecurityGroups = []v1beta1.SecurityGroup{
+				{
+					ID:   validSecurityGroup,
+					Name: "test-securitygroup",
 				},
-			})
+				{
+					ID:   fake.SecurityGroupID(),
+					Name: "test-securitygroup",
+				},
+			}
+			ExpectApplied(ctx, env.Client, nodeClass)
 			isDrifted, err := cloudProvider.IsDrifted(ctx, nodeClaim)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(isDrifted).To(Equal(cloudprovider.SecurityGroupDrift))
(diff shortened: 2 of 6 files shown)


pkg/controllers/nodeclass/status/securitygroup.go
```diff
@@ -29,11 +29,11 @@ import (
 )
 
 type SecurityGroup struct {
-	securityGroupProvider securitygroup.Provider
+	SecurityGroupProvider securitygroup.Provider
 }
 
 func (sg *SecurityGroup) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
-	securityGroups, err := sg.securityGroupProvider.List(ctx, nodeClass)
+	securityGroups, err := sg.SecurityGroupProvider.UpdateSecurityGroups(ctx, nodeClass)
```
jonathan-innis, line 36: Why change the name? This isn't updating security groups; this is still Listing the security groups like before. I think the handling is the same.

pkg/providers/securitygroup/securitygroup.go
```diff
@@ -25,37 +25,47 @@ import (
...
+		cm:         pretty.NewChangeMonitor(),
 		// TODO: Remove cache cache when we utilize the security groups from the EC2NodeClass.status
 		cache: cache,
 	}
 }
 
-func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) ([]*ec2.SecurityGroup, error) {
-	p.Lock()
-	defer p.Unlock()
+func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) ([]v1beta1.SecurityGroup, error) {
```
jonathan-innis, line 60: So, rather than relying on the provider for list operations, why doesn't our launch path just pull security groups from the status? Then you don't rely on the providers in the launch path, you only rely on the providers

jonathan-innis: That also means that your List() call doesn't have to change. From a conceptual perspective: List() on the provider continues to mean that you are listing security groups on the cloudprovider and then the status just always holds the info for the launch path and for drift

jonathan-innis (commented):
High-level comment about paring down the provider

pkg/controllers/nodeclass/status/controller.go
```diff
@@ -63,7 +63,7 @@ func NewController(kubeClient client.Client, subnetProvider subnet.Provider, sec
 
 		ami:             &AMI{amiProvider: amiProvider},
 		subnet:          &Subnet{subnetProvider: subnetProvider},
-		securitygroup:   &SecurityGroup{securityGroupProvider: securityGroupProvider},
+		securitygroup:   &SecurityGroup{SecurityGroupProvider: securityGroupProvider},
```
jonathan-innis, line 66: Do we update our requeue interval with these changes?

pkg/providers/securitygroup/securitygroup.go
```diff
@@ -25,37 +25,47 @@ import (
...
+	if err := p.kubeClient.Get(ctx, types.NamespacedName{Name: nodeClass.Name}, nodeClass); err != nil {
+		return nil, err
+	}
+	return nodeClass.Status.SecurityGroups, nil
+}
 
+// Use security groups provided by the nodeclass status
+func (p *DefaultProvider) UpdateSecurityGroups(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) ([]*ec2.SecurityGroup, error) {
 	// Get SecurityGroups
```
jonathan-innis, line 69: Do we need to `p.mu.Lock()` like we do with instance types and subnets? What's the risk of multiple NodeClasses bypassing the cache all at once and generating a lot more API calls than we expect?

engedaam (commented):
/karpenter snapshot

pkg/apis/v1beta1/ec2nodeclass_status.go
```diff
@@ -68,4 +68,7 @@ type EC2NodeClassStatus struct {
 	// InstanceProfile contains the resolved instance profile for the role
 	// +optional
 	InstanceProfile string `json:"instanceProfile,omitempty"`
+	// The generation observed by the nodeclass controller.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
jonathan-innis, line 73: This should be a per-status condition observedGeneration rather than a status-wide observed generation. See: https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1553. That also allows us to indicate "This thing has been updated for this generation (because it wasn't changed) but this other thing wasn't updated."

jonathan-innis: Maybe just wrap the upstream status conditions library right now? Or somehow coordinate the move-over of the knative status conditions library to operatorpkg?

pkg/cloudprovider/drift.go
```diff
@@ -118,12 +118,11 @@ func (c *CloudProvider) isSubnetDrifted(ctx context.Context, instance *instance.
 
 // Checks if the security groups are drifted, by comparing the security groups returned from the SecurityGroupProvider
 // to the ec2 instance security groups
-func (c *CloudProvider) areSecurityGroupsDrifted(ctx context.Context, ec2Instance *instance.Instance, nodeClass *v1beta1.EC2NodeClass) (cloudprovider.DriftReason, error) {
-	securitygroup, err := c.securityGroupProvider.List(ctx, nodeClass)
-	if err != nil {
-		return "", err
+func (c *CloudProvider) areSecurityGroupsDrifted(ec2Instance *instance.Instance, nodeClass *v1beta1.EC2NodeClass) (cloudprovider.DriftReason, error) {
+	if nodeClass.Generation != nodeClass.Status.ObservedGeneration {
```
jonathan-innis, line 122: If this is per status condition, we can just evaluate if the security group condition has been updated to do this 👍🏼

pkg/controllers/nodeclass/status/securitygroup.go
```diff
@@ -50,5 +50,5 @@ func (sg *SecurityGroup) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2No
 			Name: *securityGroup.GroupName,
 		}
 	})
-	return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
+	return reconcile.Result{RequeueAfter: time.Minute}, nil
```
jonathan-innis, line 53: This is going to increase our call rate across everything right now. What's our impact on call volume with this change? We're calling 5 times as often but we haven't taken things like discover subnets out of the critical path so what's the impact?

pkg/providers/launchtemplate/suite_test.go
```diff
@@ -204,6 +217,7 @@ var _ = Describe("LaunchTemplate Provider", func() {
 	})
 	It("should default to a generated launch template", func() {
 		ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+		ExpectReconcileSucceeded(ctx, statusController, client.ObjectKeyFromObject(nodeClass))
```
jonathan-innis, line 220: So we don't have to reconcile, should we just define status in the beforeEach and set the status to match? I'm fine either way, but just curious what you think about that option vs. directly calling the status controller reconcile

pkg/controllers/nodeclass/status/securitygroup.go
engedaam: Decided to leave this as 5 minutes

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -164,11 +164,10 @@ func (p *DefaultProvider) createAMIOptions(ctx context.Context, nodeClass *v1bet
 		return nil, err
 	}
 	// Get constrained security groups
-	securityGroups, err := p.securityGroupProvider.List(ctx, nodeClass)
-	if err != nil {
-		return nil, err
+	if nodeClass.Generation != nodeClass.Status.SecurityGroupStatus.ObservedGeneration {
+		return nil, fmt.Errorf("waiting until nodeclass is updated")
```
njtran, line 168: The error should describe the error, not the handling of the error
```suggestion
		return nil, fmt.Errorf("detected spec-status generation mismatch")
```

pkg/cloudprovider/drift.go
njtran, line 1: I'm not sure where it would be in this file, but can we add a comment about the eventual consistency delay here with a TODO?

pkg/cloudprovider/suite_test.go
```diff
@@ -674,6 +703,8 @@ var _ = Describe("CloudProvider", func() {
 			awsEnv.EC2API.DescribeSecurityGroupsOutput.Set(&ec2.DescribeSecurityGroupsOutput{SecurityGroups: []*ec2.SecurityGroup{}})
 			// Instance is a reference to what we return in the GetInstances call
 			instance.SecurityGroups = []*ec2.GroupIdentifier{{GroupId: aws.String(fake.SecurityGroupID())}}
+			awsEnv.SecurityGroupCache.Flush()
```
njtran, line 706: Should we just add this as a beforeEach()?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -164,11 +164,7 @@ func (p *DefaultProvider) createAMIOptions(ctx context.Context, nodeClass *v1bet
 		return nil, err
 	}
 	// Get constrained security groups
-	securityGroups, err := p.securityGroupProvider.List(ctx, nodeClass)
-	if err != nil {
-		return nil, err
-	}
-	if len(securityGroups) == 0 {
+	if len(nodeClass.Status.SecurityGroups) == 0 {
 		return nil, fmt.Errorf("no security groups exist given constraints")
```
njtran, line 168: Thoughts on modifying this error so that it explicitly says the status? 
```suggestion
		return nil, fmt.Errorf("no security groups are present in the status")
```

pkg/cloudprovider/suite_test.go
engedaam: No, we need to flush out the cache for this test

pkg/cloudprovider/drift.go
```diff
@@ -118,14 +118,12 @@ func (c *CloudProvider) isSubnetDrifted(ctx context.Context, instance *instance.
...
 // to the ec2 instance security groups
-func (c *CloudProvider) areSecurityGroupsDrifted(ctx context.Context, ec2Instance *instance.Instance, nodeClass *v1beta1.EC2NodeClass) (cloudprovider.DriftReason, error) {
-	securitygroup, err := c.securityGroupProvider.List(ctx, nodeClass)
-	if err != nil {
-		return "", err
-	}
-	securityGroupIds := sets.New(lo.Map(securitygroup, func(sg *ec2.SecurityGroup, _ int) string { return aws.StringValue(sg.GroupId) })...)
+func (c *CloudProvider) areSecurityGroupsDrifted(ec2Instance *instance.Instance, nodeClass *v1beta1.EC2NodeClass) (cloudprovider.DriftReason, error) {
+	// Since we are using security group status, A node is launched and immediately drift due to
+	// consistency delay of when the security group status is updated
```
njtran, line 123: nit: can you add as a TODO so that we know we eventually should fix this

engedaam: Moved it into the launch template file, since that is were the SGs are set during launch

pkg/cloudprovider/suite_test.go
engedaam: We already do that

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -163,28 +163,25 @@ func (p *DefaultProvider) createAMIOptions(ctx context.Context, nodeClass *v1bet
 	if err != nil {
 		return nil, err
 	}
+	// Since we are using security group status, A node is launched and immediately drift due to
+	// consistency delay of when the security group status is updated
+	// TODO @aengeda: add observation check to validate that the status is updated before a node is launched
```
njtran, line 168: This reads a little confusing and vague to me. WDYT:
```suggestion
	// Relying on the status rather than an API call means that Karpenter is subject to a race
	// condition where EC2NodeClass spec changes haven't propagated to the status once a node 
	// has launched. 
	// If a user changes their EC2NodeClass and shortly after Karpenter launches a node,
	// in the worst case, the node could be drifted and re-created.
	// TODO @aengeda: add status generation fields to gate node creation until the status is updated from a spec change
```

njtran (commented):
/lgtm
/approve

njtran (dismissed):

PR #6070, opened 2024-04-21
docs: Clarify conceptual docs around `Exists` operator

### LINKED ISSUE #6058: Better documentation for NodePool "Exists" operator
The Scheduling documentation [for v0.36.0](https://github.com/aws/karpenter-provider-aws/blob/main/website/content/en/v0.36/concepts/scheduling.md) says in part:

> The Exists operator can be used on a NodePool to provide workload segregation across nodes.
> 
> With the requirement on the NodePool, workloads can optionally specify a custom value as a required node affinity or node selector. Karpenter will then label the nodes it launches for these pods which prevents `kube-scheduler` from scheduling conflicting pods to those nodes. This provides a way to more dynamically isolate workloads without requiring a unique NodePool for each workload subset.

I would like elaboration on this topic. 

First, marking a NodePool with a requirement that a certain label key exists suggests that it will *only* launch nodes with that label. That puts it in conflict with the statement "...without requiring a unique NodePool for each workload subset" because I would need to create a unique NodePool for each label. I guess it doesn't actually work this way, but it is non-obvious that a "requirement" for a node label would merely be an option to allow such a node label, and if this is the case, it should be explicitly explained.

Second, how does having a label on a node "prevent `kube-scheduler` from scheduling conflicting pods to those nodes"? AFAIK a node label is not any kind of taint that would prevent pods without a node selector from being deployed to the node. 

Third, what is a "conflicting pod"?

jonathan-innis:
Fixes #6058

**Description**

Clarify some details around the `Exists` operator that caused confusion inside of the conceptual docs.

`make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/website/content/en/docs/concepts/scheduling.md
+++ b/website/content/en/docs/concepts/scheduling.md
@@ -626,19 +626,73 @@ If using Gt/Lt operators, make sure to use values under the actual label values 
 The `Exists` operator can be used on a NodePool to provide workload segregation across nodes.
 
 ```yaml
-...
-requirements:
-- key: company.com/team
-  operator: Exists
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+spec:
+  template:
+    spec:
+      requirements:
+        - key: company.com/team
+          operator: Exists
 ...
 ```
 
-With the requirement on the NodePool, workloads can optionally specify a custom value as a required node affinity or node selector.  Karpenter will then label the nodes it launches for these pods which prevents `kube-scheduler` from scheduling conflicting pods to those nodes.  This provides a way to more dynamically isolate workloads without requiring a unique NodePool for each workload subset.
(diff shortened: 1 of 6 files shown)


Nuru:
@jonathan-innis Thank you for this clarification. 

I still find the statement "you can isolate pods onto different nodes based on their values" a bit ambiguous with regard to pods that do not have node selectors/affinities. In the example given, you are assured that Team A pods and Team B pods are assigned different nodes, but you are not assured that other pods are not also assigned to those nodes, so the Team pods are not as isolated as one may desire.

How about changing the alert:

> If a workload matches the NodePool but doesn't specify a label, Karpenter will generate a random label for the node.

To something like:

> Note that by themselves, the node labels do not exclude workloads. If a workload matches the NodePool but doesn't specify a node selector or affinity, it can be assigned to any node, including the `team-a` or `team-b` nodes in the above example. If there is not an existing labeled node with capacity for the pod, Karpenter will launch a new node with a randomly generated value for the required label key.

jigisha620 (approved):
LGTM

PR #6089, opened 2024-04-23
chore: Use AMI Status Controller to Asynchronously Hydrate AMI Data

engedaam:
Fixes #N/A 

**Description**
- Asynchronously hydrate with AMI data. The controller will attempt to refresh the data every 5 minutes, and implement a exponential backoff for delay: 1ms and max: 1000s

- [ ] Yes, issue opened: # 
- [x] No


pkg/controllers/nodeclass/status/ami.go
```diff
@@ -32,7 +32,7 @@ type AMI struct {
 }
 
 func (a *AMI) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
-	amis, err := a.amiProvider.Get(ctx, nodeClass, &amifamily.Options{})
+	amis, err := a.amiProvider.UpdateAMIs(ctx, nodeClass, &amifamily.Options{})
```
jonathan-innis, line 35: What's the reason we think the name should be changed? The function itself isn't actually doing an update, it's doing a retrieval so I think that Get() or List() makes sense

pkg/providers/amifamily/ami.go
```diff
@@ -92,31 +98,42 @@ func (a AMIs) String() string {
...
 		cm:              pretty.NewChangeMonitor(),
 		versionProvider: versionProvider,
 	}
 }
 
+func (p *DefaultProvider) Get(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) ([]v1beta1.AMI, error) {
+	p.RLock()
+	defer p.RUnlock()
+
+	if err := p.kubeClient.Get(ctx, types.NamespacedName{Name: nodeClass.Name}, nodeClass); err != nil {
```
jonathan-innis, line 129: Do we really need to re-retrieve the NodeClass if we already have the nodeClass as an argument to the function?

pkg/providers/amifamily/resolver.go
```diff
@@ -122,16 +123,24 @@ func NewResolver(amiProvider Provider) *Resolver {
 // Multiple ResolvedTemplates are returned based on the instanceTypes passed in to support special AMIs for certain instance types like GPUs.
 func (r Resolver) Resolve(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, nodeClaim *corev1beta1.NodeClaim, instanceTypes []*cloudprovider.InstanceType, capacityType string, options *Options) ([]*LaunchTemplate, error) {
 	amiFamily := GetAMIFamily(nodeClass.Spec.AMIFamily, options)
-	amis, err := r.amiProvider.Get(ctx, nodeClass, options)
+	if len(nodeClass.Status.AMIs) == 0 {
+		return nil, fmt.Errorf("no amis exist given constraints")
+	}
+	amis, err := r.amiProvider.Get(ctx, nodeClass)
```
jonathan-innis, line 129: Why do we need another Get() if we already have the NodeClass and AMIs in the status above?

pkg/providers/amifamily/resolver.go
```diff
@@ -122,16 +123,24 @@ func NewResolver(amiProvider Provider) *Resolver {
...
 	}
-	if len(amis) == 0 {
-		return nil, fmt.Errorf("no amis exist given constraints")
-	}
-	mappedAMIs := amis.MapToInstanceTypes(instanceTypes)
+	mappedAMIs := MapToInstanceTypes(instanceTypes, amis)
 	if len(mappedAMIs) == 0 {
-		return nil, fmt.Errorf("no instance types satisfy requirements of amis %v", amis)
+		var sb strings.Builder
+		ids := lo.Map(amis, func(a v1beta1.AMI, _ int) string { return a.ID })
```
jonathan-innis, line 136: Is this a separate change to improve the error message that we should take in a separate PR? If not, can we encapsulate this error printing behavior into a function so that it doesn't leak into the core business logic

pkg/providers/amifamily/ami.go
```diff
@@ -117,6 +119,9 @@ func NewDefaultProvider(versionProvider version.Provider, ssm ssmiface.SSMAPI, e
 
 // Get Returning a list of AMIs with its associated requirements
 func (p *DefaultProvider) Get(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, options *Options) (AMIs, error) {
+	p.Lock()
+	defer p.Unlock()
```
njtran, line 123: Do we need to write lock? what are we writing here? It should just be a read lock on L139, right? Can we just do it then?

pkg/providers/launchtemplate/suite_test.go
```diff
@@ -1786,14 +1821,14 @@ var _ = Describe("LaunchTemplate Provider", func() {
...
 				nodeClass.Spec.AMIFamily = &v1beta1.AMIFamilyCustom
-				awsEnv.EC2API.DescribeImagesOutput.Set(&ec2.DescribeImagesOutput{Images: []*ec2.Image{
+				nodeClass.Status.AMIs = []v1beta1.AMI{
 					{
-						Name:         aws.String(coretest.RandomName()),
-						ImageId:      aws.String("ami-123"),
-						Architecture: aws.String("x86_64"),
-						CreationDate: aws.String("2022-08-15T12:00:00Z"),
+						ID: "ami-123",
+						Requirements: scheduling.NewRequirements(
```
njtran, line 1827: Any reason we wouldn't be using NewNodeSelectorRequirements()?

pkg/providers/amifamily/ami.go
```diff
@@ -79,25 +81,25 @@ func (a AMIs) Sort() {
 	})
 }
 
-func (a AMIs) String() string {
+func String(amis []v1beta1.AMI) string {
```
njtran, line 84: Can we rename this to something like PrettyFormat() or something else? Calling it `String()` could make a user think that the intention was to make this a member function.

pkg/providers/amifamily/ami.go
```diff
@@ -79,25 +81,25 @@ func (a AMIs) Sort() {
...
 	var sb strings.Builder
-	ids := lo.Map(a, func(a AMI, _ int) string { return a.AmiID })
-	if len(a) > 25 {
+	ids := lo.Map(amis, func(a v1beta1.AMI, _ int) string { return a.ID })
```
njtran, line 86: What's the point of changing it from this type to the v1beta1 AMI type?

pkg/providers/amifamily/ami.go
engedaam: We need a write lock here because we are writing to the cache, If we don't we could have a cache miss and call AWS more times then needed

pkg/providers/amifamily/ami.go
engedaam: External we will only be using v1beta1 for formatting

pkg/providers/launchtemplate/suite_test.go
engedaam: This mainly to use the `scheduling.NewRequirement`. This would staying consistent with how we define our default AMIs

pkg/providers/amifamily/ami.go
njtran: Do we need this AMI struct anymore then? Seems like we're doing some translation back and forth, where the only difference is the CreationTimestamp. I'm happy to punt to another PR, but think it's worth refactoring eventually.

pkg/providers/amifamily/ami.go
njtran: Yeah let's keep the lock in here

pkg/providers/amifamily/ami.go
njtran, line 1: Let's put the lock back in with the same comment as the instance type file too

pkg/providers/amifamily/ami.go
```diff
@@ -79,25 +79,25 @@ func (a AMIs) Sort() {
...
+		sb.WriteString(fmt.Sprintf(" and %d other(s)", len(amis)-25))
 	} else {
 		sb.WriteString(strings.Join(ids, ", "))
 	}
 	return sb.String()
 }
 
 // MapToInstanceTypes returns a map of AMIIDs that are the most recent on creationDate to compatible instancetypes
-func (a AMIs) MapToInstanceTypes(instanceTypes []*cloudprovider.InstanceType) map[string][]*cloudprovider.InstanceType {
+func MapToInstanceTypes(instanceTypes []*cloudprovider.InstanceType, amis []v1beta1.AMI) map[string][]*cloudprovider.InstanceType {
```
jonathan-innis, line 95: nit: I'd expect that the first argument to a function that's called "MapToInstanceTypes" is the thing that we're mapping. I'd re-order these arguments so that amis come first

pkg/providers/amifamily/ami.go
```diff
@@ -79,25 +79,25 @@ func (a AMIs) Sort() {
...
 	amiIDs := map[string][]*cloudprovider.InstanceType{}
 	for _, instanceType := range instanceTypes {
-		for _, ami := range a {
-			if err := instanceType.Requirements.Compatible(ami.Requirements, scheduling.AllowUndefinedWellKnownLabels); err == nil {
-				amiIDs[ami.AmiID] = append(amiIDs[ami.AmiID], instanceType)
+		for _, ami := range amis {
+			if err := instanceType.Requirements.Compatible(scheduling.NewNodeSelectorRequirementsWithMinValues(ami.Requirements...), scheduling.AllowUndefinedWellKnownLabels); err == nil {
```
jonathan-innis, line 99: Do we know the reason why `minValues` is part of the ami status? Should it be? I can't think of a reason why we would ever have `minValues` passed through to the AMI status. Seems like it really only needs to be in the requirements block of the NodePool and NodeClaim

pkg/providers/amifamily/ami.go
```diff
@@ -79,25 +79,25 @@ func (a AMIs) Sort() {
 	})
 }
```
jonathan-innis, line 81: nit: Should the amiProvider method be called `Get()` or `List()`? I'd recommend `List()` to align the call with the other providers (subnets and security groups)

jonathan-innis: Also, do we need to `Sort()` in the `Get()` call anymore? If we are sorted outside the function and we assume that the status is ordered by creationDate already, do we have a need to do a second sort? Seems like we could just rely on the sort in the nodeClass controller.

engedaam: The status controller does not currently sort the AMIs by creation date

jonathan-innis: But it could, right? Should it? Seems like it would be nice if that was all handled outside the List() call so we don't have an assumed dependency on the data that we return here.

pkg/providers/amifamily/ami.go
```diff
@@ -41,7 +41,7 @@ import (
 )
 
 type Provider interface {
-	Get(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, options *Options) (AMIs, error)
+	List(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, options *Options) (AMIs, error)
```
jonathan-innis, line 44:
```suggestion
	List(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (AMIs, error)
```

Another question: I think we can do this. We don't actually use the passed-through options anywhere.

pkg/controllers/nodeclass/status/ami.go
```diff
@@ -32,7 +32,7 @@ type AMI struct {
 }
 
 func (a *AMI) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
-	amis, err := a.amiProvider.Get(ctx, nodeClass, &amifamily.Options{})
+	amis, err := a.amiProvider.List(ctx, nodeClass)
```
ellistarn, line 35: Totally optional: Check out https://github.com/awslabs/operatorpkg/blob/main/singleton/controller.go to implement this singleton controller.

pkg/providers/amifamily/ami.go
njtran: I had the same thought. Since there's a non min-values one, we should probably just use that function.

pkg/cloudprovider/suite_test.go
```diff
@@ -147,6 +148,38 @@ var _ = Describe("CloudProvider", func() {
 							Zone: "test-zone-1c",
 						},
 					},
+					AMIs: []v1beta1.AMI{
+						{
+							ID: "ami-test1",
+							Requirements: scheduling.NewRequirements(
```
njtran, line 154: Still wondering why we're not using the NewNodeSelectorRequirements here

engedaam: Updated to include that

pkg/providers/amifamily/ami.go
engedaam: Yeah, I think this was just updated as part of the minValues implementation. I'll remove it

pkg/cloudprovider/suite_test.go
```diff
@@ -624,19 +656,36 @@ var _ = Describe("CloudProvider", func() {
...
+					{
+						ID: armAMIID,
+						Requirements: []v1.NodeSelectorRequirement{
+							{Key: v1.LabelArchStable, Operator: v1.NodeSelectorOpIn, Values: []string{corev1beta1.ArchitectureArm64}},
+						},
+					},
+					{
+						ID: amdAMIID,
+						Requirements: []v1.NodeSelectorRequirement{
+							{Key: v1.LabelArchStable, Operator: v1.NodeSelectorOpIn, Values: []string{corev1beta1.ArchitectureArm64}},
```
jonathan-innis, line 686: This requirement doesn't match the AMD ID

pkg/providers/instance/suite_test.go
```diff
@@ -111,6 +112,38 @@ var _ = Describe("InstanceProvider", func() {
...
+							ID: "ami-test2",
+							Requirements: []v1.NodeSelectorRequirement{
+								{Key: v1.LabelArchStable, Operator: v1.NodeSelectorOpIn, Values: []string{corev1beta1.ArchitectureAmd64}},
+								{Key: v1beta1.LabelInstanceGPUCount, Operator: v1.NodeSelectorOpExists},
+							},
+						},
+						{
+							ID: "ami-test3",
+							Requirements: []v1.NodeSelectorRequirement{
+								v1.NodeSelectorRequirement{Key: v1.LabelArchStable, Operator: v1.NodeSelectorOpIn, Values: []string{corev1beta1.ArchitectureAmd64}},
```
jonathan-innis, line 134: This architecture is wrong here. I believe you meant for this to be Arm64

pkg/providers/instancetype/suite_test.go
```diff
@@ -137,6 +137,38 @@ var _ = Describe("InstanceTypeProvider", func() {
...
+							ID: "ami-test2",
+							Requirements: []v1.NodeSelectorRequirement{
+								{Key: v1.LabelArchStable, Operator: v1.NodeSelectorOpIn, Values: []string{corev1beta1.ArchitectureAmd64}},
+								{Key: v1beta1.LabelInstanceGPUCount, Operator: v1.NodeSelectorOpExists},
+							},
+						},
+						{
+							ID: "ami-test3",
+							Requirements: []v1.NodeSelectorRequirement{
+								v1.NodeSelectorRequirement{Key: v1.LabelArchStable, Operator: v1.NodeSelectorOpIn, Values: []string{corev1beta1.ArchitectureAmd64}},
```
jonathan-innis, line 159: This should be Arm64. Looks like these were copied across files, so can we make sure that this is correct everywhere it's used?

jonathan-innis: This is another reason why it might make sense to standardize this in the `test.EC2NodeClass`

engedaam: Yeah, I agree now. I will move the status into a central place. This is meant to mock the default AL2 AMI, Why do you believe this should be Arm64?

pkg/providers/amifamily/resolver.go
```diff
@@ -120,18 +120,14 @@ func NewResolver(amiProvider Provider) *Resolver {
...
-	}
-	if len(amis) == 0 {
+	if len(nodeClass.Status.AMIs) == 0 {
 		return nil, fmt.Errorf("no amis exist given constraints")
 	}
-	mappedAMIs := amis.MapToInstanceTypes(instanceTypes)
+	mappedAMIs := MapToInstanceTypes(instanceTypes, nodeClass.Status.AMIs)
 	if len(mappedAMIs) == 0 {
-		return nil, fmt.Errorf("no instance types satisfy requirements of amis %v", amis)
+		return nil, fmt.Errorf("no instance types satisfy requirements of amis %v", pretty.Slice(lo.Map(nodeClass.Status.AMIs, func(a v1beta1.AMI, _ int) string { return a.ID }), 25))
```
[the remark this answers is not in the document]
engedaam: Users would be able to tag more than 4 AMIs for us to discover, no?

pkg/controllers/nodeclass/status/ami.go
engedaam: I will include that in a follow-up

pkg/controllers/nodeclass/status/ami_test.go
```diff
@@ -328,17 +296,14 @@ var _ = Describe("NodeClass AMI Status Controller", func() {
...
-								Operator: "In",
-								Values: []string{
-									"amd64",
-								},
-							},
+					Requirements: []v1.NodeSelectorRequirement{{
+						Key:      "kubernetes.io/arch",
+						Operator: "In",
+						Values: []string{
+							"amd64",
```
njtran, line 303: Is there any reason these aren't the enums?

njtran (commented):
LGTM just one comment.

njtran (approved):

engedaam (commented):
/karpenter snapshot

jonathan-innis (approved):
LGTM 🚀

PR #6123, opened 2024-05-01
test: fix daemonset sensitive tests

jmdeal:
Fixes #N/A 

**Description**
Fixes several tests which are sensitive to the number of daemonsets in the test environment. This is due to assumptions about the size of node a given pod would fit on. This assumption was broken by #5982 which added an additional daemonset.

`karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/suites/consolidation/suite_test.go
+++ b/test/suites/consolidation/suite_test.go
@@ -560,7 +560,14 @@ var _ = Describe("Consolidation", func() {
 						},
 					},
 					ResourceRequirements: v1.ResourceRequirements{
-						Requests: v1.ResourceList{v1.ResourceCPU: resource.MustParse("1.5")},
+						Requests: v1.ResourceList{
+							v1.ResourceCPU: func() resource.Quantity {
+								dsOverhead := env.GetDaemonSetOverhead(nodePool)
+								base := lo.ToPtr(resource.MustParse("1800m"))
+								base.Sub(*dsOverhead.Cpu())
+								return *base
+							}(),
+						},
 					},
 				},
 			})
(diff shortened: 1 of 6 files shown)


jmdeal (commented):
/karpenter snapshot

test/suites/consolidation/suite_test.go
```diff
@@ -560,7 +560,14 @@ var _ = Describe("Consolidation", func() {
 						},
 					},
 					ResourceRequirements: v1.ResourceRequirements{
-						Requests: v1.ResourceList{v1.ResourceCPU: resource.MustParse("1.5")},
+						Requests: v1.ResourceList{
+							v1.ResourceCPU: func() resource.Quantity {
+								dsOverhead := env.GetDaemonSetOverhead(nodePool)
+								base := lo.ToPtr(resource.MustParse("1800m"))
+								base.Sub(*dsOverhead.Cpu())
```
engedaam, line 567: Why is this only for CPU? shouldn't we do this for all resources?

test/suites/drift/suite_test.go
```diff
@@ -288,34 +281,29 @@ var _ = Describe("Drift", func() {
...
-							Labels: appLabels,
-						},
-						NodeSelector: map[string]string{"test-partition": fmt.Sprintf("%d", i)},
-						// Each xlarge has 4 cpu, so each node should fit no more than 1 pod.
-						ResourceRequirements: v1.ResourceRequirements{
-							Requests: v1.ResourceList{
-								v1.ResourceCPU: resource.MustParse("3"),
-							},
-						},
+			deployment := coretest.Deployment(coretest.DeploymentOptions{
```
engedaam, line 284: This would make 1 deployment with 5 pods, don't we want 5 deployment with 1 pods?

test/suites/consolidation/suite_test.go
jmdeal: Since we're only specifying CPU requests for the pods we're scheduling that should be the only resource under contention. Theoretically daemonset memory requests could grow large enough that they no longer fit on the nodes we're selecting on, but we're nowhere near that memory threshold atm.

test/suites/drift/suite_test.go
jmdeal: It shouldn't make a difference, the 5 deployments allowed us to create 5 different virtual domains to ensure the pods scheduled onto different nodes. Seems like the CPU request was also there for the same purpose. Anti-affinity should have the same result.

test/suites/drift/suite_test.go
```diff
@@ -288,34 +281,29 @@ var _ = Describe("Drift", func() {
 
 			// Make 5 pods all with different deployments and different test partitions, so that each pod can be put
```
engedaam, line 282: Can we just update these comments then?

engedaam (approved):
LGTM 🚀

PR #6125, opened 2024-05-01
chore: add nodeclass name as label to nodeclaim

### LINKED ISSUE #5178: Nodeclass / ec2nodeclass not in labels
### Description

It is difficult to find out which nodeclass is used for a given instance. Please add a human readable label to the instances 

nice to have

rschalo:
Fixes #5178

**Description**
Adds the node class name to node claim.
Unit testing.

- [ ] Yes, issue opened: # 
- [x] No


pkg/cloudprovider/cloudprovider.go
```diff
@@ -317,6 +317,9 @@ func (c *CloudProvider) instanceToNodeClaim(i *instance.Instance, instanceType *
 	}
 	labels[v1.LabelTopologyZone] = i.Zone
 	labels[corev1beta1.CapacityTypeLabelKey] = i.CapacityType
+	if v, ok := i.Tags[v1beta1.LabelNodeClass]; ok {
```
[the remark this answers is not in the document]
engedaam: Miss understood on my end

test/suites/integration/tags_test.go
```diff
@@ -103,6 +103,7 @@ var _ = Describe("Tags", func() {
 			nodeInstance := instance.NewInstance(lo.ToPtr(env.GetInstance(node.Name)))
 			Expect(nodeInstance.Tags).To(HaveKeyWithValue("Name", node.Name))
 			Expect(nodeInstance.Tags).To(HaveKey("karpenter.sh/nodeclaim"))
+			Expect(nodeInstance.Tags).To(HaveKeyWithValue("karpenter.k8s.aws/ec2nodeclass", node.Name))
```
engedaam, line 106:
```suggestion
			Expect(nodeInstance.Tags).To(HaveKeyWithValue("karpenter.k8s.aws/ec2nodeclass", nodeClass.Name))
```

Does that need to be the EC2NodeClass name?

rschalo: Yes - went a bit too fast with the autocomplete.

test/suites/integration/tags_test.go
```diff
@@ -86,7 +86,7 @@ var _ = Describe("Tags", func() {
 	})
 
 	Context("Tagging Controller", func() {
-		It("should tag with karpenter.sh/nodeclaim and Name tag", func() {
+		It("should tag with karpenter.sh/nodeclaim, karpenter.k8s.aws/ec2nodeclass, and Name tag", func() {
```
engedaam, line 89: Do you mind expanding this tests to include all the tags we expect on the instance? The only tags that are left are the `karpenter.sh/nodepool` and `kubernetes.io/cluster/$ClusterName`

engedaam: Also for this PR, I was thinking we could have another test to validate the NodeClaim has the label. Can most likely go here https://github.com/aws/karpenter-provider-aws/blob/main/test/suites/nodeclaim/nodeclaim_test.go

pkg/cloudprovider/cloudprovider.go
jonathan-innis, line 320: Does this take into account the scheduling behavior of this label being there? What if someone has an affinity for a certain NodeClass label? Are we going to take this into account?

rschalo: Added in latest commit

test/suites/integration/tags_test.go
rschalo: Added.

pkg/providers/instancetype/instancetype.go
```diff
@@ -131,6 +131,7 @@ func (p *DefaultProvider) List(ctx context.Context, kc *corev1beta1.KubeletConfi
 		blockDeviceMappingsHash,
 		aws.StringValue((*string)(nodeClass.Spec.InstanceStorePolicy)),
 		aws.StringValue(nodeClass.Spec.AMIFamily),
+		nodeClass.Name,
```
njtran, line 134: I'm a bit worried with how this aligns with other changes we've made in the past that modifies our controller's memory usage + cache hits. Realistically, if every other detail is the same, and only the nodeClass name is different, we shouldn't need to maintain two cache values, especially since the data we're storing is pretty large.

njtran: I'm not sure if adding this requirement into the cloudprovider.InstanceType is the right way to do this in the long run, especially as the `nodeClass` requirement is applicable for all instance types in the NodePool. I'd like to explore the idea of adding in a `karpenter.sh/nodeclass-name` label in the upstream code, so that we can better model this in the scheduling code. @jonathan-innis @engedaam thoughts?

jonathan-innis: > aligns with other changes we've made in the past that modifies our controller's memory usage + cache hits

IMO, this is the strongest point. This will definitely increase our memory footprint for storing these instance types in a cache. Given this trade-off, I'd agree that the neutral code is probably the best way to attack this, though I'd recommend that we take a format more akin to `<nodeClassRef.group>/<nodeClassRef.kind>: <nodeClassRef.name>`

PR #6126, opened 2024-05-01
pref: Reduce the number of `DescribeImages` calls

### LINKED ISSUE #5997: Caching DescribeImages results (amiSelectorTerms)
### Description

We have multiple NodePools/NodeClasses and but amiSelectorTerms field is identical for them. I'm seeing that karpenter sends `DescribeImages` requests for them separately and it causes `Client.RequestLimitExceeded` errors:

"requestParameters": {"nextToken": "x", "maxResults": 500, "executableBySet": {}, "imagesSet": {}, "ownersSet": {"items": [{"owner": "self"}, {"owner": "amazon"}]}, "filterSet": {"items": [{"name": "name", "valueSet": {"items": [{"value": "k8s-ami_*"}]}}, {"name": "tag:active", "valueSet": {"items": [{"value": "true"}]}}]}}

We are getting 3k+ errors in an hour. It isn't super critical but would be nice to solve. 

engedaam:
Fixes #5997

**Description**
- Karpenter currently does not have any locking mechanism during its read and write operation for AMIs that are stored in the cache. Adding in locks to synchronize the cache for identical AMISelectorTerms
- The DescribeImages API does not filter out AMIs before fetching AMIs by tags. This will result in the DescribeImages API sending empty pages(This is currently expected behavior). The best mitigation here will for Karpenter to increases page sizes to reduce the number of request that are made.

- Result from testing 
- I had 20 EC2NodeClass with Identical AMISelectorTerms set 
```
  amiSelectorTerms:
  - tags:
      testing: aengeda
```
- Number of DescribeImages call from main: 1604
- Number of DescribeImages call from with Improvements: 79

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/amifamily/ami.go
+++ b/pkg/providers/amifamily/ami.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -45,6 +46,7 @@ type Provider interface {
 }
 
 type DefaultProvider struct {
+	sync.RWMutex
 	cache           *cache.Cache
 	ssm             ssmiface.SSMAPI
 	ec2api          ec2iface.EC2API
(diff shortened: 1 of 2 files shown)


pkg/providers/amifamily/ami.go
```diff
@@ -45,6 +46,7 @@ type Provider interface {
 }
 
 type DefaultProvider struct {
+	sync.RWMutex
```
jonathan-innis, line 49:
```suggestion
	sync.Mutex
```

This doesn't need to be RW if we're not using the read lock, right? Are there other mutexes that we should just convert to a standard mutex?

engedaam: Added that in. Only found one that should be using the standard mutex

jonathan-innis (approved):
LGTM 🚀

engedaam (commented):
/karpenter snapshot

PR #6137, opened 2024-05-03
ci: Find all of instance-profiles in the CI account

engedaam:
Fixes #N/A 

**Description**
- Instance-profiles API is paginated, and requires us to go through all the pages. This is causing to track and clean-up only the first page of instance-profiles. This PR will allow us to go though all the pages.

- N/A

- [ ] Yes, issue opened: # 
- [x] No


test/hack/resource/pkg/resourcetypes/instanceprofile.go
```diff
@@ -44,39 +44,40 @@ func (ip *InstanceProfile) Global() bool {
...
-			InstanceProfileName: out.InstanceProfiles[i].InstanceProfileName,
+			InstanceProfileName: instanceProfiles[i].InstanceProfileName,
 		})
 		if err != nil {
 			errs[i] = err
 			continue
 		}
 
-		clusterName, _ := lo.Find(profiles.Tags, func(tag iamtypes.Tag) bool {
+		clusterName, foundClusterName := lo.Find(profiles.Tags, func(tag iamtypes.Tag) bool {
```
jonathan-innis, line 62: nit: I can't remember. Maybe not a huge deal, but I recalled that we said we were fine with not doing the "found" check since the result of `lo.FromPtr` was going to be an empty string anyways, which isn't valid for a cluster name or for a region name

test/hack/resource/pkg/resourcetypes/instanceprofile.go
```diff
@@ -44,39 +44,40 @@ func (ip *InstanceProfile) Global() bool {
...
 			continue
 		}
 
 		for _, t := range profiles.Tags {
 			// Since we can only get the date of the instance profile (not the exact time the instance profile was created)
 			// we add a day to the time that it was created to account for the worst-case of the instance profile being created
 			// at 23:59:59 and being marked with a time of 00:00:00 due to only capturing the date and not the time
-			if lo.FromPtr(t.Key) == karpenterTestingTag && out.InstanceProfiles[i].CreateDate.Add(time.Hour*24).Before(expirationTime) {
-				names = append(names, lo.FromPtr(out.InstanceProfiles[i].InstanceProfileName))
+			if lo.FromPtr(t.Key) == karpenterTestingTag && instanceProfiles[i].CreateDate.Add(time.Hour*24).Before(expirationTime) {
```
jonathan-innis, line 78: Seems like we're doing a bit of double-work here. We first already check whether the cluster name is in our set of excluded clusters which means we should have been able to validate that this is a "karpenter-owned" resource above, so we really don't need to check this again here. We could really avoid all of the iteration through this tags section and just reduce all of this to a few "found" checks.

test/hack/resource/pkg/resourcetypes/instanceprofile.go
```diff
@@ -142,3 +142,24 @@ func (ip *InstanceProfile) Cleanup(ctx context.Context, names []string) ([]strin
 	}
 	return deleted, errs
 }
+
+func (ip *InstanceProfile) getAllInstanceProfiles(ctx context.Context) (instanceprofile []iamtypes.InstanceProfile, err error) {
+	var nextToken *string
+	for {
+		out, err := ip.iamClient.ListInstanceProfiles(ctx, &iam.ListInstanceProfilesInput{
```
jonathan-innis, line 149: optional: Consider looking at [aws-sdk-go-v2 pagination](https://aws.github.io/aws-sdk-go-v2/docs/making-requests/#using-paginators)

njtran (approved):
LGTM

jmdeal (approved):
LGTM 🚀

PR #6138, opened 2024-05-03
ci: Fix docgen by calling providers directly

jonathan-innis:
Fixes #N/A 

**Description**

This change fixes calling `make docgen` and the instance types generation by calling the providers directly.

`make docgen`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/instancetype/instancetype.go
+++ b/pkg/providers/instancetype/instancetype.go
@@ -104,10 +104,6 @@ func (p *DefaultProvider) List(ctx context.Context, kc *corev1beta1.KubeletConfi
 	if kc == nil {
 		kc = &corev1beta1.KubeletConfiguration{}
 	}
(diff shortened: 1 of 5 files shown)


pkg/providers/instancetype/instancetype.go
```diff
@@ -104,10 +104,6 @@ func (p *DefaultProvider) List(ctx context.Context, kc *corev1beta1.KubeletConfi
 	if kc == nil {
```
engedaam, line 104: Should we just remove this and always pass a kubleteConfiguration now?

jonathan-innis: It can still be nil. We just pass it transparently, so it's possible someone hasn't set it.

engedaam (approved):
LGTM 🚀

Issue #5161, opened 2023-11-25
Discover Instance Type Capacity Memory Overhead Instead of `vmMemoryOverheadPercent`


jonathan-innis:
### Tell us about your request

We could consider a few options to discover the expected capacity overhead for a given instance type:
1. We could store the instance type capacity in memory once a version of that type has been launched and use that as the capacity value after the initial launch rather than basing our calculations off of some heuristic
2. We could launch instance types, check their capacity and diff the reported capacity from the actual capacity in a generated file that can be shipped with Karpenter on release so we always have accurate measurements on the instance type overhead.

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

Calculating the difference between the EC2-reported memory capacity and the actual capacity of the instance as reported by kubelet.

### Are you currently working around this issue?

Using a heuristic `vmMemoryOverheadPercent` value right now that is tunable by users and passed through `karpenter-global-settings`

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jonathan-innis:
This is a "transferred" version of the issue in `kubernetes-sigs/karpenter`: https://github.com/kubernetes-sigs/karpenter/issues/716

sosheskaz:
Could a simple mitigation be to add a static setting? e.g. `vmMemoryOverheadOffset`, which would take a static amount of memory?

When running compute which is highly heterogeneous, but relatively large, this would simplify the practice of managing this setting, because a sufficiently large static overhead would be able to outweigh the percentage, and allow for less waste.

BEvgeniyS:
The `vmMemoryOverheadOffset` value is clearly not one-size-fits-all, so making it per-nodepool would be a great start.

I prefer approach #<zws>1 because it doesn't tie Karpenter to cloud provider updates and can work even when new instance types appear before an update.

Here's how I imagine this could be implemented:
a) Convert it to something like "initVmMemoryOverheadOffset" and make it a per-nodepool setting.
[...]

PR #6146, opened 2024-05-03
feat: per-instance type VM memory overhead percentages

pH14:
Bandaid for https://github.com/aws/karpenter-provider-aws/issues/5161, without trying to solve the problem globally for all EC2 instance types

**Description**

As originally discussed in https://github.com/kubernetes-sigs/karpenter/issues/716 the global `vmMemoryOverheadPercent` option is limiting when running on a heterogeneous set of nodes. For instance, a small EC2 instance may require ~7% overhead, but the larger nodes may only require 1-2% overhead. Undersizing this values risks provisioning nodes that cannot actually run the pods it's intended for, and oversizing is inefficient / expensive.

This PR is a strawman proposal for solving this problem: allow Karpenter to accept a mapping of instance type to memory overhead percentage, and fall back to the global `vmMemoryOverheadPercent` if no specific mapping exists. For our deployment this would allow us to create a mapping for the instance types we care about, and potentially sets up the suggestion in #5161 of "We could launch instance types, check their capacity and diff the reported capacity from the actual capacity in a generated file that can be shipped with Karpenter on release so we always have accurate measurements on the instance type overhead"

It hasn't been yet! Before going further on the tests, I wanted to gather input into whether this change might be accepted at all first.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/operator/options/options.go
+++ b/pkg/operator/options/options.go
@@ -20,6 +20,8 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strconv"
+	"strings"
 	"time"
 
 	coreoptions "sigs.k8s.io/karpenter/pkg/operator/options"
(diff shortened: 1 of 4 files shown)


engedaam:
@pH14  Thank you for the contribution! I have a few questions and suggestions on the next steps you can take. The team mainly hopes to have an RFC for long standing problems such as this one. Within the RFC, I hope that you can address questions such as; What are all the options for solving this problem? What would be the best short and long terms solutions? What is preventing us from implementing the long term solution? Why should Karpenter have a bandaid solution for the short term? Do we want to expose API surface to customers if we don’t intend for us to keep it long term? The team is heading to v1 of the project and we are critical thinking about API surface that we want to expose to user to avoid the need of introducing a breaking change.

The team is definitely open to accepting contributions for VM memory overhead percentages, but we just want to make sure we have evaluated all the possible solutions.  Here are examples of RFC we have used in the past: https://github.com/kubernetes-sigs/karpenter/tree/main/designs and https://github.com/aws/karpenter-provider-aws/tree/main/designs. I also recommended coming to the working group meetings to discuss your ideas for possible solutions. Here is where you can find the schedule for the meetings: https://karpenter.sh/docs/contributing/working-group/.

PR #6149, opened 2024-05-04
ci: bump upgrade suite source commit

jmdeal:
Fixes #N/A 

**Description**
Bumps the source version for the upgrade suite to be after the cloudwatch infrastructure changes.

N/A (can't test top-level workflow changes w/ `/karpenter snapshot`)

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.github/actions/e2e/setup-cluster/action.yaml
+++ b/.github/actions/e2e/setup-cluster/action.yaml
@@ -215,7 +215,8 @@ runs:
       # Adding taints after all necessary pods have scheduled to the manged node group nodes
       # amazon-cloudwatch-observability pods do no not tolerate CriticalAddonsOnly=true:NoSchedule and
       # amazon-cloudwatch-observability addons does not allow to add tolerations to the addon pods as part of the advanced configuration
-      kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --all
+      # Overwrite existing taints to ensure we don't fail here on upgrade
+      kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --all --overwrite
 
       # We delete DaemonSets that we don't care about because it causes inconsistencies in scheduling due to
       # dcgm-exporter and neuron-monitor selecting on specific instance types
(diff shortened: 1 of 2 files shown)


jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

.github/actions/e2e/setup-cluster/action.yaml
```diff
@@ -215,19 +215,23 @@ runs:
...
-      kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --all
+      # Overwrite existing taints to ensure we don't fail here on upgrade
+      kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --all --overwrite
 
-      # We delete DaemonSets that we don't care about because it causes inconsistencies in scheduling due to
-      # dcgm-exporter and neuron-monitor selecting on specific instance types
-      # See https://github.com/kubernetes-sigs/karpenter/issues/715 for more detail
-      kubectl delete daemonsets -n amazon-cloudwatch dcgm-exporter neuron-monitor
+      # Only modify cloudwatch daemonsets if the addon was installed. This is not guaranteed in the upgrade test since only the control plane is upgraded.
+      if [[ "$(aws eks list-addons --cluster "${CLUSTER_NAME}" | yq '.addons | contains(["amazon-cloudwatch-observability"])')" == "true" ]]; then
```
jonathan-innis, line 222: Doesn't this mean that we won't have the CWA addon in the upgrade test? Should we modify our code so that we still get the CWA installed even in the upgrade scenario? Maybe call specific calls on eksctl in the upgrade test to ensure that they get installed if they aren't already there.

jmdeal (commented):
/karpenter snapshot

.github/actions/e2e/setup-cluster/action.yaml
jmdeal: I think I agree that this is the better approach, I've updated the PR. Not sure it matters too much in the long run since eventually the source cluster will have the cloudwatch addon as well but this probably makes sense for us to ensure generally anyway.

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

engedaam (approved):
LGTM 🚀

PR #6159, opened 2024-05-06
chore: Update `associatePublicIPAddress` to not be set in the launch templets

engedaam:
Fixes #N/A 

**Description**
- Before this change https://github.com/aws/karpenter-provider-aws/pull/6057, If a customer did not set `associatePublicIPAddress` in their nodeclass, Karpenter would auto discover setting `associatePublicIPAddress` on launch templates by seeing if any of the subnets can add public IPs. If not, the field would be set to `nil` on launch templates. After PR 6057, we would set `associatePublicIPAddress` to false in the launch template. This PR looks to make feature parity by setting `associatePublicIPAddress` on launch template to `nil`.

- N/A

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/subnet/subnet.go
+++ b/pkg/providers/subnet/subnet.go
@@ -37,7 +37,7 @@ import (
 type Provider interface {
 	LivenessProbe(*http.Request) error
 	List(context.Context, *v1beta1.EC2NodeClass) ([]*ec2.Subnet, error)
-	CheckAnyPublicIPAssociations(context.Context, *v1beta1.EC2NodeClass) (bool, error)
+	AssociatePublicIPAddressValue( *v1beta1.EC2NodeClass) (*bool)
 	ZonalSubnetsForLaunch(context.Context, *v1beta1.EC2NodeClass, []*cloudprovider.InstanceType, string) (map[string]*Subnet, error)
 	UpdateInflightIPs(*ec2.CreateFleetInput, *ec2.CreateFleetOutput, []*cloudprovider.InstanceType, []*Subnet, string)
 }
@@ -116,14 +116,17 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
 	return lo.Values(subnets), nil
 }
 
-// CheckAnyPublicIPAssociations returns a bool indicating whether all referenced subnets assign public IPv4 addresses to EC2 instances created therein
-func (p *DefaultProvider) CheckAnyPublicIPAssociations(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (bool, error) {
+// associatePublicIPAddressValue validates whether we know the association value for all subnets AND
+// that all subnets don't have associatePublicIP set. If both of these are true, we set the value explicitly to false
+// For more detail see: https://github.com/aws/karpenter-provider-aws/pull/3814
+func (p *DefaultProvider) AssociatePublicIPAddressValue(nodeClass *v1beta1.EC2NodeClass) *bool {
 	for _, subnet := range nodeClass.Status.Subnets {
-		if subnetAssociatePublicIP, ok := p.associatePublicIPAddressCache.Get(subnet.ID); ok && subnetAssociatePublicIP.(bool) {
-			return true, nil
+		subnetAssociatePublicIP, ok := p.associatePublicIPAddressCache.Get(subnet.ID)
+		if !ok || subnetAssociatePublicIP.(bool) {
+			return nil
 		}
 	}
-	return false, nil
+	return lo.ToPtr(false)
 }
 
 // ZonalSubnetsForLaunch returns a mapping of zone to the subnet with the most available IP addresses and deducts the passed ips from the available count
--- a/pkg/providers/subnet/suite_test.go
+++ b/pkg/providers/subnet/suite_test.go
@@ -219,19 +219,20 @@ var _ = Describe("SubnetProvider", func() {
 			}, subnets)
 		})
 	})
-	Context("CheckAnyPublicIPAssociations", func() {
-		It("should note that no subnets assign a public IPv4 address to EC2 instances on launch", func() {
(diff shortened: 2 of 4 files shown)


pkg/providers/subnet/subnet.go
```diff
@@ -117,7 +117,11 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
 // CheckAnyPublicIPAssociations returns a bool indicating whether all referenced subnets assign public IPv4 addresses to EC2 instances created therein
 func (p *DefaultProvider) CheckAnyPublicIPAssociations(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (bool, error) {
```
jonathan-innis, line 118: Does it even make sense to return an error from this function?

engedaam: Yes, we should return an error, because we want to differentiate between when we can't find the data vs when none of the subnets can add a public IPs

jonathan-innis: Why not just return the associatePublicIPAddressValue directly? e.g. 

```go
options.AssociatePublicIPAddress = p.associatePublicIPAddressValue()

// associatePublicIPAddressValue validates whether we know the association value for all subnets AND
// that all subnets don't have associatePublicIP set. If both of these are true, we set the value explicitly to false
// For more detail see: https://github.com/aws/karpenter-provider-aws/pull/3814
func (p *DefaultProvider) associatePublicIPAddressValue() {
  for _, subnet := range nodeClass.Status.Subnets
    subnetAssociatePublicIP, ok := p.associatePublicIPAddressCache.Get(subnet.ID)
    if !ok || subnetAssociatePublicIP.(bool) {
      return nil
    }
  }
  return false
}
```

engedaam: After attempting you suggestion above, It fails on the return value 
```
pkg/providers/subnet/subnet.go:1: : # github.com/aws/karpenter-provider-aws/pkg/providers/subnet
pkg/providers/subnet/subnet.go:137:11: cannot use nil as bool value in return statement (typecheck)
/*
```

jonathan-innis: Why not use a `*bool`?

pkg/providers/subnet/subnet.go
```diff
@@ -114,14 +114,17 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
 	return lo.Values(subnets), nil
 }
 
-// CheckAnyPublicIPAssociations returns a bool indicating whether all referenced subnets assign public IPv4 addresses to EC2 instances created therein
-func (p *DefaultProvider) CheckAnyPublicIPAssociations(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (bool, error) {
+// associatePublicIPAddressValue validates whether we know the association value for all subnets AND
+// that all subnets don't have associatePublicIP set. If both of these are true, we set the value explicitly to false
+// For more detail see: https://github.com/aws/karpenter-provider-aws/pull/3814
+func (p *DefaultProvider) CheckAnyPublicIPAssociations(nodeClass *v1beta1.EC2NodeClass) *bool {
```
jonathan-innis, line 120:
```suggestion
func (p *DefaultProvider) associatePublicIPAddressValue(nodeClass *v1beta1.EC2NodeClass) *bool {
```

I don't think we changed the name

pkg/providers/subnet/suite_test.go
```diff
@@ -225,9 +225,10 @@ var _ = Describe("SubnetProvider", func() {
 					Tags: map[string]string{"foo": "bar"},
 				},
 			}
-			onlyPrivate, err := awsEnv.SubnetProvider.CheckAnyPublicIPAssociations(ctx, nodeClass)
+			_, err := awsEnv.SubnetProvider.List(ctx, nodeClass)
```
jonathan-innis, line 228: Should we add a test here to validate what we do when we don't have the data?

pkg/providers/subnet/suite_test.go
```diff
@@ -219,19 +219,20 @@ var _ = Describe("SubnetProvider", func() {
...
 			Expect(err).To(BeNil())
-			Expect(onlyPrivate).To(BeFalse())
+			onlyPrivate := awsEnv.SubnetProvider.AssociatePublicIPAddressValue(nodeClass)
+			Expect(lo.FromPtr(onlyPrivate)).To(BeFalse())
 		})
-		It("should note that at least one subnet assigns a public IPv4 address to EC2instances on launch", func() {
+		It("should be nil that at least one subnet assigns a public IPv4 address to EC2instances on launch", func() {
```
jonathan-innis, line 235:
```suggestion
		It("should be nil when at least one subnet assigns a public IPv4 address to EC2instances on launch", func() {
```

nit

pkg/providers/subnet/suite_test.go
```diff
@@ -245,9 +246,24 @@ var _ = Describe("SubnetProvider", func() {
...
+				},
+			}
+			nodeClass.Status.Subnets = []v1beta1.Subnet{
+				{
+					ID:   "subnet-test2",
+					Zone: "test-zone-1b",
+				},
+			}
+			awsEnv.SubnetCache.Flush() // remove any subnet data that might be in the subnetCache
+			onlyPrivate := awsEnv.SubnetProvider.AssociatePublicIPAddressValue(nodeClass)
```
jonathan-innis, line 265: nit: Should this variable still be called "onlyPrivate" in all tests?

jonathan-innis (approved):
Two optional nits

jonathan-innis (dismissed):
LGTM 🚀  Nice work!

njtran (approved):

njtran (approved):

engedaam (commented):
/karpenter snapshot

PR #6167, opened 2024-05-08
fix: Ensure shallow copy of data when returning back cached data

jonathan-innis:
Fixes #N/A 

**Description**

Ensure the slices are deep copied before returning back cached data from the caller. This ensures if the slice of the data is manipulated by the caller (like sorted), it doesn't affect the original copy that is stored and cached.

This can particularly be a problem when sorts are occurring on the original data in a multi-threaded manner. Because of swaps, this can cause the same data to appear multiple times or to be altered in ways that are inconsistent.

`make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/amifamily/suite_test.go
+++ b/pkg/providers/amifamily/suite_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"sync"
 	"testing"
 	"time"
 
@@ -74,7 +75,7 @@ var _ = BeforeEach(func() {
 			{
 				Name:         aws.String(amd64AMI),
 				ImageId:      aws.String("amd64-ami-id"),
-				CreationDate: aws.String(time.Now().Format(time.RFC3339)),
+				CreationDate: aws.String(time.Time{}.Format(time.RFC3339)),
 				Architecture: aws.String("x86_64"),
 				Tags: []*ec2.Tag{
 					{Key: aws.String("Name"), Value: aws.String(amd64AMI)},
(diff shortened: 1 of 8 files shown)


pkg/providers/amifamily/suite_test.go
```diff
@@ -196,6 +197,56 @@ var _ = Describe("AMIProvider", func() {
...
+		nodeClass.Spec.AMISelectorTerms = []v1beta1.AMISelectorTerm{
+			{
+				ID: "amd64-ami-id",
+			},
+			{
+				ID: "arm64-ami-id",
+			},
+		}
+		wg := sync.WaitGroup{}
+		for i := 0; i < 100000; i++ {
```
njtran, line 210: is 100k really necessary here? I'm not really sure how this impacts performance of our tests, so it might be fine.

pkg/providers/amifamily/suite_test.go
```diff
@@ -196,6 +197,56 @@ var _ = Describe("AMIProvider", func() {
...
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				defer GinkgoRecover()
+				images, err := awsEnv.AMIProvider.Get(ctx, nodeClass, &amifamily.Options{})
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(images).To(HaveLen(2))
+				// Sort everything in parallel and ensure that we don't get data races
+				sort.Slice(images, func(i, j int) bool {
```
njtran, line 220: Can we just use the function in code rather than redefining it here?

njtran (commented):
LGTM just two questions

pkg/providers/amifamily/suite_test.go
jonathan-innis: Can do 10,000. I was able to repro it at that scale. This was still only taking ~4-6s so 100,000 is also fine.

pkg/providers/amifamily/suite_test.go
jonathan-innis: Sure, I didn't have any strong opinions one way or the other.

jonathan-innis (commented):
/karpenter snapshot

njtran (approved):

vitaly-dt:
Hi @jonathan-innis 
This fix has been communicated by AWS Health Dashboard as a critical fix.
Can maybe elaborate on what is this pull request solving?

PR #6172, opened 2024-05-09
docs: update karpenter docs around soft affinity behavior

cnmcavoy:
Fixes #N/A 

**Description**

Documents Karpenter behavior around soft affinities.

Docs only change.

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/website/content/en/docs/concepts/scheduling.md
+++ b/website/content/en/docs/concepts/scheduling.md
@@ -192,6 +192,16 @@ Then the pod can declare that custom label.
 
 See [nodeSelector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) in the Kubernetes documentation for details.
 
+## Preferences
+
+Karpenter is aware of preferences (node affinity, pod affinity, pod anti-affinity, and pod topology) and treats them as requirements in most circumstances. Karpenter uses these preferences when determining if a pod can schedule on a node (absent topology requirements), or when determining if a pod can be shifted to a new node.
+
+Karpenter starts by treating preferred affinities as required affinities when constructing requirements for a pod. When these requirements can not be met, the pod's preferences are relaxed one-at-a-time by ascending weight (lowest weight is relaxed first) where possible and the remaining requirements are tried again.
+
+{{% alert title="Warning" color="warning" %}}
+Karpenter does not interpret preferred affinities as required when constructing topology requirements for scheduling to a node. If these preferences are necessary, required affinities should be used [as documented in Node Affinity](#node-affinity).
+{{% /alert %}}
+
 ### Node affinity
 
 Examples below illustrate how to use Node affinity to include (`In`) and exclude (`NotIn`) objects.
@@ -200,6 +210,10 @@ When setting rules, the following Node affinity types define how hard or soft ea
 
 * **requiredDuringSchedulingIgnoredDuringExecution**: This is a hard rule that must be met.
 * **preferredDuringSchedulingIgnoredDuringExecution**: This is a preference, but the pod can run on a node where it is not guaranteed.
+
+{{% alert title="Note" color="primary" %}}
+Preferred affinities on pods can result in more nodes being created than expected because Karpenter will prefer to create new nodes to satisfy preferences, [see the preferences documentation](#preferences) for details.
+{{% /alert %}}
 
 The `IgnoredDuringExecution` part of each tells the pod to keep running, even if conditions change on the node so the rules no longer matched.
 You can think of these concepts as `required` and `preferred`, since Kubernetes never implemented other variants of these rules.
@@ -261,7 +275,7 @@ If they all fail, Karpenter will fail to provision the pod.
 Karpenter will backoff and retry over time.
 So if capacity becomes available, it will schedule the pod without user intervention.
 
-## Taints and tolerations
+### Taints and tolerations
 
 Taints are the opposite of affinity.
 Setting a taint on a node tells the scheduler to not run a pod on it unless the pod has explicitly said it can tolerate that taint. This example shows a NodePool that was set up with a taint for only running pods that require a GPU, such as the following:
@@ -308,9 +322,14 @@ spec:
 ```
 See [Taints and Tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) in the Kubernetes documentation for details.
 
-## Topology Spread
+### Topology Spread
 
 By using the Kubernetes `topologySpreadConstraints` you can ask the NodePool to have pods push away from each other to limit the blast radius of an outage. Think of it as the Kubernetes evolution for pod affinity: it lets you relate pods with respect to nodes while still allowing spread.
+
+{{% alert title="Note" color="primary" %}}
+Preferred topology spread (`ScheduleAnyway`) can result in more nodes being created than expected because Karpenter will prefer to create new nodes to satisfy spread constraints, [see the preferences documentation](#preferences) for details.
+{{% /alert %}}
+
 For example:
 
 ```yaml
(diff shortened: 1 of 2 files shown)


website/content/en/docs/concepts/scheduling.md
```diff
@@ -260,6 +260,14 @@ If they all fail, Karpenter will fail to provision the pod.
 Karpenter will backoff and retry over time.
 So if capacity becomes available, it will schedule the pod without user intervention.
 
+#### Preferred Affinities
+
+Karpenter treats preferred affinities as hard affinities in almost all circumstances.
```
jonathan-innis, line 265:
```suggestion
Karpenter treats preferred affinities as required affinities in almost all circumstances.
```

nit

website/content/en/docs/concepts/scheduling.md
```diff
@@ -260,6 +260,14 @@ If they all fail, Karpenter will fail to provision the pod.
 Karpenter will backoff and retry over time.
 So if capacity becomes available, it will schedule the pod without user intervention.
 
+#### Preferred Affinities
```
jonathan-innis, line 263: I wonder if there should just be a section on "Preferences" rather than one specifically on "Preferred Affinities". I think a lot of this behavior that you are detailing below applies generally to how Karpenter thinks about preferences. 

For more detail, see: https://github.com/kubernetes-sigs/karpenter/issues/666

website/content/en/docs/concepts/scheduling.md
```diff
@@ -260,6 +260,14 @@ If they all fail, Karpenter will fail to provision the pod.
...
+
+Karpenter treats preferred affinities as hard affinities in almost all circumstances.
+
+Karpenter interprets preferred affinities as a soft affinity when building up topology requirements for scheduling to a node. Preferred affinities are ignored in regards to topology zones and hard affinities should be used as documented above.
```
jonathan-innis, line 267: Oh, this is interesting. I think it's a little unclear without an example, but I do get what you are saying: Since TSC consider nodeAffinity when considering which domains to consider, you are calling out that this doesn't cause Karpenter to unnecessarily constrain the TSC when these are listed as preferences. It's definitely bespoke. Maybe worth calling this out as an `{{% alert %}}` since this is more like edge behavior

website/content/en/docs/concepts/scheduling.md
```diff
@@ -260,6 +260,14 @@ If they all fail, Karpenter will fail to provision the pod.
...
+
+In all other circumstances, Karpenter treats the required and preferred affinities as if they were required affinities. When determining if a pod can schedule on a node (absent topology requirements), or if a pod can be shifted to a new node, soft affinities are treated as if hard affinities. This includes when determining if a node can be safely disrupted, a node with pods that contain unsatisfied preferred affinities will not be disrupted if there are no matching nodes that meet it's preferred requirements.
```
jonathan-innis, line 269: There's a caveat to this statement. Specifically, if we can't satisfy the hard affinity in any way, a preference will cause us to fallback from treating the affinity as strictly required to removing it altogether.

jonathan-innis, line 269: What do you mean "contain unsatisfied preferred affinities"? You mean if there is a pod on a node that doesn't satisfy its current preference, we also won't shift the pod to another node that also doesn't satisfy this preference? -- If that's the case that's actually interesting behavior. Not surprising based on how I understand the scheduling behavior, but slightly unexpected. We could probably do better here, so it might be worth opening an issue if this is the case.

jonathan-innis:
Thanks for the docs contribution @cnmcavoy! Got a few thoughts.

website/content/en/docs/concepts/scheduling.md
wmgroot: I think https://github.com/kubernetes-sigs/karpenter/issues/1204 may be an example of this behavior. I haven't been able to locate any special handling for the controlplane label in Karpenter's code.

website/content/en/docs/concepts/scheduling.md
cnmcavoy: Added a warning alert.

website/content/en/docs/concepts/scheduling.md
cnmcavoy: @jonathan-innis I updated the language of this to be clearer and match the behavior around relaxing requirements.

@wmgroot I don't think that's evidence of this, I believe what is happening there is no valid nodeclaim can be generated because of the restricted domain.

website/content/en/docs/concepts/scheduling.md
cnmcavoy: I restructured this change to add a 2nd level "Preferences" section, and moved the topics that you called out in https://github.com/kubernetes-sigs/karpenter/issues/666 as sub headings. Let me know if it doesn't match what you expected.

website/content/en/docs/concepts/scheduling.md
cnmcavoy: Updated to include this language.

website/content/en/docs/concepts/scheduling.md
```diff
@@ -191,6 +191,14 @@ Then the pod can declare that custom label.
 
 See [nodeSelector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) in the Kubernetes documentation for details.
 
+## Preferences
+
+Karpenter is aware of preferences (node affinity, pod affinity, pod antiaffinity, and pod topology) and treats them as requirements in almost all circumstances. As a result, Karpenter also treats preferred affinities as required affinities when constructing requirements for a pod. When these requirements can not be met, preferences are relaxed where possible and the remaining requirements are tried again.
```
GnatorX, line 196: Not sure where to add this but this is only true for preferences on pod spec and not scheduler configs

jonathan-innis: This may be too much detail. We don't respect scheduler config at all, so it might be worth just having a different section on scheduler config

jonathan-innis:
```suggestion
Karpenter is aware of preferences (node affinity, pod affinity, pod antiaffinity, and pod topology) and treats them as requirements in almost all circumstances. As a result, Karpenter also starts by treating preferred affinities as required affinities when constructing requirements for a pod. When these requirements can not be met, preferences are relaxed one-at-a-time by ascending weight (lowest weight is relaxed first) and the remaining requirements are tried again.
```

Small wording change, but this just implies that it starts in one way and then relaxes them over time

website/content/en/docs/concepts/scheduling.md
```diff
@@ -200,6 +208,10 @@ When setting rules, the following Node affinity types define how hard or soft ea
 * **requiredDuringSchedulingIgnoredDuringExecution**: This is a hard rule that must be met.
 * **preferredDuringSchedulingIgnoredDuringExecution**: This is a preference, but the pod can run on a node where it is not guaranteed.
 
+{{% alert title="Note" color="primary" %}}
+Karpenter uses preferred as if it was required when determining if a pod can schedule on a node (absent topology requirements), or when determining if a pod can be shifted to a new node. When preferred affinities can't be satisfied by any possible nodeclaim configuration then Karpenter will relax the preferences and only consider required affinities. In practice, this behavior can result in more nodes being created than expected, as Karpenter will prefer to create new nodes to satisfy preferences.
```
jonathan-innis, line 212:
```suggestion
Karpenter treats preferred affinities as if they are required when determining if a pod can schedule on a node (absent topology requirements), or when determining if a pod can be shifted to a new node. When preferred affinities can't be satisfied by any possible nodeclaim configuration then Karpenter will relax the preferences and only consider required affinities. In practice, this behavior can result in more nodes being created than expected, as Karpenter will prefer to create new nodes to satisfy preferences.
```

Does it make sense to keep this statement shortened and then link up to the preferences section in this doc? That way -- you have a single source of truth for talking about preferences in the scheduling doc and don't have to keep referencing it across the different scheduling semantics that we cover

website/content/en/docs/concepts/scheduling.md
```diff
@@ -307,7 +319,7 @@ spec:
 ```
 See [Taints and Tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) in the Kubernetes documentation for details.
 
-## Topology Spread
+### Topology Spread
```
jonathan-innis, line 322: Do we want to talk about preferred topology spread and create an alert block like you did over in the affinity section? You could keep it short and sweet and link it out like a proposed up above

cnmcavoy: Added.

website/content/en/docs/concepts/scheduling.md
cnmcavoy: Great idea. The docs in the Preferences section has the content, I've updated the alerts to reduce their verbiage and to reference back to it.

cnmcavoy:
I hope I managed to address everything, many good suggestions.

jonathan-innis (approved):
LGTM 🚀

GnatorX (commented):
lgtm

PR #6174, opened 2024-05-09
test: fix race in flakey drift test

jmdeal:
Fixes #N/A 

**Description**
Fixes a flake in a drift test. The test eventually expects the original replacement nodeclaim to be deleted, and checks for this by expecting on the number of nodeclaims. The problem is a new replacement nodeclaim may be created between checks in the eventually block. 

`/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/suites/drift/suite_test.go
+++ b/test/suites/drift/suite_test.go
@@ -838,41 +838,19 @@ var _ = Describe("Drift", func() {
 
 			env.EventuallyExpectDrifted(startingNodeClaimState...)
 
-			// Only drift a single node at a time due to default budgets
-			driftedNodeClaimState := env.EventuallyExpectCreatedNodeClaimCount("==", int(numPods)+1)
-
-			// Expect nodes To get tainted
+			// Expect nodes To get tainted, expect only one node is disrupted due to default disruption budgets
 			taintedNodes := env.EventuallyExpectTaintedNodeCount("==", 1)
 
 			// Drift should fail and the original node should be untainted
 			// TODO: reduce timeouts when disruption waits are factored out
 			env.EventuallyExpectNodesUntaintedWithTimeout(11*time.Minute, taintedNodes...)
 
-			// We give another 6 minutes here to handle the deletion at the 15m registration timeout
-			Eventually(func(g Gomega) {
-				nodeClaims := &corev1beta1.NodeClaimList{}
-				g.Expect(env.Client.List(env, nodeClaims, client.HasLabels{coretest.DiscoveryLabel})).To(Succeed())
-				replacementNodeClaims := lo.FilterMap(driftedNodeClaimState, func(nc *corev1beta1.NodeClaim, _ int) (types.UID, bool) {
-					if lo.ContainsBy(startingNodeClaimState, func(snc *corev1beta1.NodeClaim) bool {
-						return snc.UID == nc.UID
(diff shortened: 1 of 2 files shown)


jmdeal (commented):
/karpenter snapshot

test/suites/drift/suite_test.go
```diff
@@ -849,7 +852,17 @@ var _ = Describe("Drift", func() {
 			Eventually(func(g Gomega) {
```
jonathan-innis, line 852: Do we even care that the NodeClaim count is the same or do we just care that the tainted node is deleted? I think we just want to make sure that we get through the loop and that we still maintain the same set of nodes after the full disruption loop occurs. Realistically, I don't even know if we really need to validate that we remove the old node. Ideally, we could run a check throughout the whole test to validate that we never delete our nodes and we maintain static stability rather than just checking at the end.

jmdeal: That's a good point, it is valuable to test this IMO but that can be left to a registration TTL test. Plus removing this wait should trim ~4 minutes off of the drift suite.

jmdeal: I do wonder if there's value in checking the nodes consistently throughout the test vs just testing that they still exist in the end. It definitely doesn't hurt, but introducing a goroutine to poll cluster state throughout the test is more complicated than what we have today. Do we expect there to be a scenario where we delete a node and provision a replacement with the same name?

jonathan-innis: Yeah, that's a good solid point. Since we are checking on UUIDs, there's really no difference.

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

test/suites/drift/suite_test.go
```diff
@@ -838,31 +838,24 @@ var _ = Describe("Drift", func() {
...
+			// Assert this over several minutes to ensure a subsequent disruption controller pass doesn't
+			// successfully schedule the evicted pods to the in-flight nodeclaim and disrupt the original node
 			Consistently(func(g Gomega) {
-				nodeClaims := &corev1beta1.NodeClaimList{}
-				g.Expect(env.Client.List(env, nodeClaims, client.HasLabels{coretest.DiscoveryLabel})).To(Succeed())
-
-				startingNodeClaimUIDs := lo.Map(startingNodeClaimState, func(nc *corev1beta1.NodeClaim, _ int) types.UID { return nc.UID })
-				nodeClaimUIDs := lo.Map(nodeClaims.Items, func(nc corev1beta1.NodeClaim, _ int) types.UID { return nc.UID })
-				g.Expect(sets.New(nodeClaimUIDs...).IsSuperset(sets.New(startingNodeClaimUIDs...))).To(BeTrue())
+				nodeClaims := env.ExpectNodeClaimCount(">=", int(numPods))
```
jonathan-innis, line 852: Calling Expect inside of a Gomega Eventually/Consistently call won't have the "g" context which means the first failure will cause the test to exit

jmdeal: Oh right, I'll revert that half of the changes then. Guess there's no real way around dupeing here.

jonathan-innis (approved):

PR #5716, opened 2024-02-23
docs: RFC Supporting ODCR in Karpenter

garvinp-stripe:
Starts the fix but doesn't completely fixes #3042 

**Description**
Narrowly scoped design doc for support ODCR in Karpenter.

**How was this change tested?**

**Does this change impact docs?**
- [x] Yes, PR includes docs updates 
- [ ] Yes, issue opened: # 
- [ ] No

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6198, opened 2024-05-14
feat: RFC Implementation Supporting ODCR

tvonhacht-apple:
This is a collaboration on implementing the RFC #5716 Supporting ODCRs

## Progress

- EC2NodeClass
  - [x] Capacity Reservation
    - [X] Selector Terms
    - [X] Status update
    - [X] Provisioning
      - [X] with available Capacity Reservation
      - [X] fallback to On-Demand without available Capacity Reservation
    - [ ] Consolidation
      - https://github.com/kubernetes-sigs/karpenter/pull/1263 (outdated)

## TODOs

- [] Agree on `instanceMatchCriteria` vs `type`
- [] Remove all log statements

**Description**

### Supporting associating ODCR to EC2NodeClass

Add a new field `capacityReservationSelectorTerms` to `EC2NodeClass` 
```yaml
kind: EC2NodeClass
metadata:
  name: example-node-class
spec:
  capacityReservationSelectorTerms:
    - # The Availability Zone of the Capacity Reservation
      availabilityZone: String | None
      # The platform of operating system for which the Capacity Reservation reserves capacity
      id: String | None
      # The type of operating system for which the Capacity Reservation reserves capacity
      instanceType: String | None
      # The ID of the Amazon Web Services account that owns the Capacity Reservation
      ownerId: String | None
      # Tags is a map of key/value tags used to select subnets
      # Specifying '*' for a value selects all values for a given tag key.
      tags: Map | None
      # Indicates the type of instance launches that the Capacity Reservation accepts. The options include:
      #    - open:
      #       The Capacity Reservation accepts all instances that have
      #       matching attributes (instance type, platform, and Availability
      #       Zone). Instances that have matching attributes launch into the
      #       Capacity Reservation automatically without specifying any
      #       additional parameters.
      #    - targeted:
      #       The Capacity Reservation only accepts instances that
      #       have matching attributes (instance type, platform, and
      #       Availability Zone), and explicitly target the Capacity
      #       Reservation. This ensures that only permitted instances can use
      #             the reserved capacity.
      type: String | None
status:
  capacityReservations:
    - # AvailabilityZone of the Capacity Reservation
      availabilityZone: String
      # Available Instance Count of the Capacity Reservation
      availableInstanceCount: Integer
      # The date and time at which the Capacity Reservation expires. When a Capacity
      # Reservation expires, the reserved capacity is released and you can no longer
      # launch instances into it. The Capacity Reservation's state changes to expired
      # when it reaches its end date and time.
      endDate: String | None
      # Indicates the way in which the Capacity Reservation ends. A Capacity Reservation
      # can have one of the following end types:
      #   * unlimited - The Capacity Reservation remains active until you explicitly
      #   cancel it.
      #   * limited - The Capacity Reservation expires automatically at a specified
      #   date and time.
      endDateType: String
      # ID of the Capacity Reservation
      id: String
      # Indicates the type of instance launches that the Capacity Reservation accepts. The options include:
      #   - open:
      #       The Capacity Reservation accepts all instances that have
      #       matching attributes (instance type, platform, and Availability
      #       Zone). Instances that have matching attributes launch into the
      #       Capacity Reservation automatically without specifying any
      #       additional parameters.
      #   - targeted:
      #       The Capacity Reservation only accepts instances that
      #       have matching attributes (instance type, platform, and
      #       Availability Zone), and explicitly target the Capacity
      #       Reservation. This ensures that only permitted instances can use
      #       the reserved capacity.
      instanceMatchCriteria: String
      # Instance Platform of the Capacity Reservation
      instancePlatform: String
      # Instance Type of the Capacity Reservation
      instanceType: String
      # Owner Id of the Capacity Reservation
      ownerId: String
      # The date and time at which the Capacity Reservation was started.
      startDate: String
      # Total Instance Count of the Capacity Reservation
      totalInstanceCount: Integer
```

### Adding new label `karpenter.k8s.aws/capacity-reservation-id` nodeClaim/node

In order for Karpenter (and admins) to understand if a NodeClaim/Node is part of an ODCR, Karpenter is adding the annotation `karpenter.k8s.aws/capacity-reservation-id` containing the Capacity Reservation Id (for example `cr-12345678`)

This follows closely (does not implement all fields) how EC2 [DescribeCapacityReservations](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeCapacityReservations.html) can filter.
- only exception `instanceMatchCriteria` is called `type`

Karpenter will perform validation against the spec to ensure there isn't any violation prior to creating the LaunchTemplates.

- [ ] Yes, issue opened: # 
- [ ] No


pkg/providers/amifamily/resolver.go
```diff
@@ -154,12 +157,63 @@ func (r Resolver) Resolve(ctx context.Context, nodeClass *v1beta1.EC2NodeClass,
 				maxPods: int(instanceType.Capacity.Pods().Value()),
 			}
 		})
+
+		zones := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...).Get(v1.LabelTopologyZone)
+		capacityReservations := []v1beta1.CapacityReservation{}
```
tvonhacht-apple, line 162: I think we can handle this within the if below, we dont need to create this variable beforehand

pkg/providers/instance/instance.go
```diff
@@ -289,6 +296,9 @@ func (p *DefaultProvider) getLaunchTemplateConfigs(ctx context.Context, nodeClas
 	var launchTemplateConfigs []*ec2.FleetLaunchTemplateConfigRequest
 	launchTemplates, err := p.launchTemplateProvider.EnsureAll(ctx, nodeClass, nodeClaim, instanceTypes, capacityType, tags)
 	if err != nil {
+		if cloudprovider.IsInsufficientCapacityError(err) {
```
jonathan-innis, line 299: Same comment here about error wrapping and checking. You shouldn't need to do this as long as the error that's returned from L297 is an ICE error

pkg/providers/instance/instance.go
```diff
@@ -193,17 +193,22 @@ func (p *DefaultProvider) CreateTags(ctx context.Context, id string, tags map[st
 	return nil
 }
 
-func (p *DefaultProvider) launchInstance(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, nodeClaim *corev1beta1.NodeClaim, instanceTypes []*cloudprovider.InstanceType, tags map[string]string) (*ec2.CreateFleetInstance, error) {
+// workaround until capacity-reservation natively supported by EC2 API to return capacity type
+func (p *DefaultProvider) launchInstance(ctx context.Context, nodeClass *v1beta1.EC2NodeClass, nodeClaim *corev1beta1.NodeClaim, instanceTypes []*cloudprovider.InstanceType, tags map[string]string) (*ec2.CreateFleetInstance, string, error) {
```
jonathan-innis, line 197: should we just shift `NewInstanceFromFleet` inside of this function so that we return an `instance.Instance` and then we don't have to return capacity type back

pkg/providers/instance/instance.go
```diff
@@ -193,17 +193,22 @@ func (p *DefaultProvider) CreateTags(ctx context.Context, id string, tags map[st
...
 	if err != nil {
-		return nil, fmt.Errorf("getting subnets, %w", err)
+		return nil, "", fmt.Errorf("getting subnets, %w", err)
 	}
 
 	// Get Launch Template Configs, which may differ due to GPU or Architecture requirements
 	launchTemplateConfigs, err := p.getLaunchTemplateConfigs(ctx, nodeClass, nodeClaim, instanceTypes, zonalSubnets, capacityType, tags)
 	if err != nil {
-		return nil, fmt.Errorf("getting launch template configs, %w", err)
+		if cloudprovider.IsInsufficientCapacityError(err) {
```
jonathan-innis, line 208: Same comment here

pkg/providers/instancetype/instancetype.go
```diff
@@ -251,11 +269,22 @@ func (p *DefaultProvider) UpdateInstanceTypeOfferings(ctx context.Context) error
...
+) []cloudprovider.Offering {
 	var offerings []cloudprovider.Offering
+
+	// workaround until ec2 supports "capacity-reservation" as a supported class natively
+	supportedUsageClasses := sets.NewString(append(aws.StringValueSlice(instanceType.SupportedUsageClasses), "capacity-reservation")...)
+
 	for zone := range zones {
 		// while usage classes should be a distinct set, there's no guarantee of that
-		for capacityType := range sets.NewString(aws.StringValueSlice(instanceType.SupportedUsageClasses)...) {
+		for capacityType := range supportedUsageClasses {
```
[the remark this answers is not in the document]
jonathan-innis: Do we need to model an entirely new capacity type here or is it possible for us to consider this an on-demand capacity type that has a different scheduling requirement, something like `karpenter.k8s.aws/capacity-reservation-id`. Users could select on this by using an Exists requirement. The only leftover detail that we'd have to handle is prioritizing the capacity reservation offerings over the spot ones if we have them as part of our launch options.

pkg/errors/errors.go
```diff
@@ -48,6 +48,7 @@ var (
 		"UnfulfillableCapacity",
 		"Unsupported",
 		"InsufficientFreeAddressesInSubnet",
+		"ReservationCapacityExceeded",
```
jonathan-innis, line 51: When we return this type of ICE error, should we short-circuit and update the capacity reservation that we launched with in-place so we don't have to wait for another iteration of the capacity reservation polling to update the instance availability.

If we didn't want to directly update this to 0, we could also use this as a trigger to re-call DescribeCapacityReservations since we know that something has changed since we made the launch decision originally

pkg/apis/v1beta1/ec2nodeclass_status.go
```diff
@@ -38,6 +38,34 @@ type SecurityGroup struct {
...
+	ID string `json:"id"`
+	// AvailabilityZone of the Capacity Reservation
+	// +required
+	AvailabilityZone string `json:"availabilityZone"`
+	// Available Instance Count of the Capacity Reservation
+	// +required
+	AvailableInstanceCount int `json:"availableInstanceCount"`
+	// Instance Match Criteria of the Capacity Reservation
+	// +required
+	InstanceMatchCriteria string `json:"instanceMatchCriteria"`
```
jonathan-innis, line 54: What is instance match criteria?

pkg/apis/v1beta1/ec2nodeclass_status.go
```diff
@@ -38,6 +38,34 @@ type SecurityGroup struct {
...
+	// Instance Platform of the Capacity Reservation
+	// +required
+	InstancePlatform string `json:"instancePlatform"`
```
jonathan-innis, line 57: What's the instance platform?

pkg/apis/v1beta1/ec2nodeclass_status.go
```diff
@@ -38,6 +38,34 @@ type SecurityGroup struct {
...
+	// Instance Type of the Capacity Reservation
+	// +required
+	InstanceType string `json:"instanceType"`
+	// Owner Id of the Capacity Reservation
+	// +required
+	OwnerID string `json:"ownerId"`
```
jonathan-innis, line 63: Do we need to add OwnerID here?

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -175,6 +179,39 @@ type AMISelectorTerm struct {
 	Owner string `json:"owner,omitempty"`
 }
 
+// CapacityReservationSelectorTerm defines selection logic for a Capacity Reservation used by Karpenter to launch nodes.
+// If multiple fields are used for selection, the requirements are ANDed.
+type CapacityReservationSelectorTerm struct {
```
jonathan-innis, line 184: How many of these selector fields do we think we truly need for the initial run? We can always add more fields later but the second we introduce something, we're stuck with it. High-level, I can see tags, id, and instance type making sense. Do we think the other fields would be commonly used?

pkg/apis/v1beta1/ec2nodeclass_status.go
tvonhacht-apple: targeted or open

pkg/apis/v1beta1/ec2nodeclass_status.go
tvonhacht-apple: it helps to identify if the capacity reservation is from own account or a shared one

pkg/apis/v1beta1/ec2nodeclass_status.go
tvonhacht-apple: instancePlatform: Linux/UNIX

pkg/cloudprovider/cloudprovider.go
```diff
@@ -96,6 +96,9 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *corev1beta1.NodeC
 	}
 	instance, err := c.instanceProvider.Create(ctx, nodeClass, nodeClaim, instanceTypes)
 	if err != nil {
+		if cloudprovider.IsInsufficientCapacityError(err) {
```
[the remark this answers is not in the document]
tvonhacht-apple: from my testing that wasn't working, as it becomes a normal error see line 102 where its just a string at that point

website/content/en/preview/getting-started/migrating-from-cas/scripts/step04-controller-iam.sh
```diff
@@ -71,6 +71,12 @@ cat << EOF > controller-policy.json
             "Resource": "arn:${AWS_PARTITION}:eks:${AWS_REGION}:${AWS_ACCOUNT_ID}:cluster/${CLUSTER_NAME}",
             "Sid": "EKSClusterEndpointLookup"
         },
+        {
+            "Effect": "Allow",
+            "Action": "eks:DescribeCapacityReservations",
+            "Resource": "*",
```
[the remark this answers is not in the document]
tvonhacht-apple: done

website/content/en/preview/getting-started/migrating-from-cas/scripts/step04-controller-iam.sh
```diff
@@ -71,6 +71,12 @@ cat << EOF > controller-policy.json
             "Resource": "arn:${AWS_PARTITION}:eks:${AWS_REGION}:${AWS_ACCOUNT_ID}:cluster/${CLUSTER_NAME}",
             "Sid": "EKSClusterEndpointLookup"
         },
+        {
+            "Effect": "Allow",
+            "Action": "eks:DescribeCapacityReservations",
```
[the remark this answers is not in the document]
tvonhacht-apple: done

pkg/errors/errors.go
tvonhacht-apple: I think setting it to zero at this time would be ok, then it eventually will refetch anyways, but this would let other evaluation happening at same time, get it immediately too

pkg/providers/amifamily/resolver.go
```diff
@@ -154,12 +157,63 @@ func (r Resolver) Resolve(ctx context.Context, nodeClass *v1beta1.EC2NodeClass,
...
+		if capacityType == "capacity-reservation" {
```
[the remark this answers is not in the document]
tvonhacht-apple: the benefit of this is that it becomes visible in nodeclaims that something is happening?

pkg/errors/errors.go
jonathan-innis: Yep, I guess it all sort-of comes out in the wash. I'm a little skeptical that we can model this with the existing ICE cache on its own. Minimally, we need to update the ICE cache to reflect the fact that the ICE is only for that particular capacity reservation. Without that, we'd either not try the OD instance in that AZ at all OR we wouldn't try other capacity reservations assigned to that instance type in that zone. Either way, we aren't really reacting correctly to what the CreateFleet call is telling us here.

pkg/cloudprovider/drift.go
```diff
@@ -105,6 +110,32 @@ func (c *CloudProvider) isSubnetDrifted(instance *instance.Instance, nodeClass *
...
+	if nodeClaimCapacityReservationId == "" {
+		return "", nil
+	}
+
+	if nodeClaimCapacityReservationId != lo.FromPtr(instance.CapacityReservationID) {
+		// nodeClaim and instance are not synced yet
+		return "", nil
+	}
+
+	for _, capacityReservation := range nodeClass.Status.CapacityReservations {
```
jonathan-innis, line 130: `lo.Find()`?

pkg/controllers/nodeclaim/capacityreservation/controller.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context) (reconcile.Result, error) {
+	cloudProviderNodeClaims, err := c.cloudProvider.List(ctx)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("listing instances, %w", err)
+	}
+
+	providerIDCloudProviderNodeClaimMap := make(map[string]*corev1beta1.NodeClaim, len(cloudProviderNodeClaims))
```
jonathan-innis, line 57:
```suggestion
	providerIDCloudProviderNodeClaimMap := map[string]*corev1beta1.NodeClaim{}
```

nit (style): We tend to be preferring this syntax over the `make` syntax throughout the codebase

pkg/controllers/nodeclaim/capacityreservation/controller.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, cloudProviderNodeClaim := range cloudProviderNodeClaims {
```
jonathan-innis, line 58:
```suggestion
	for _, nc := range cloudProviderNodeClaims {
```

nit: might be worth shortening the name for looping

pkg/controllers/nodeclaim/capacityreservation/controller.go
```diff
@@ -0,0 +1,104 @@
...
+		log.FromContext(ctx).WithValues("cloudProviderNodeClaim", cloudProviderNodeClaim).V(0).Info("reconcile nodeClaim adding cloudProviderNodeClaim")
```
jonathan-innis, line 59: I think we may only want to debug log when we are _updating_ the label. I doubt that we want to debug log on every reconciliation attempt, even if nothing changes

pkg/controllers/nodeclaim/capacityreservation/controller.go
```diff
@@ -0,0 +1,104 @@
...
+		providerIDCloudProviderNodeClaimMap[cloudProviderNodeClaim.Status.ProviderID] = cloudProviderNodeClaim
```
jonathan-innis, line 60: Rather than creating a map in memory, can we use the controller-runtime implementation of this with field indexers?

You can do something like

```
if err := c.kubeClient.List(ctx, nodeClaimList, client.MatchingFields{"status.providerID": cloudProviderNodeClaim.Status.ProviderID}); err != nil {
  ...
}
```

pkg/controllers/nodeclaim/capacityreservation/controller.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	// Find the NodeClaims that don't match
+	// Then patch the label so that it adds or removes the capacity reservation label
+	errs := make([]error, len(nodeClaimList.Items))
+	for i := range nodeClaimList.Items {
+		nodeClaim := nodeClaimList.Items[i]
+		stored := nodeClaim.DeepCopy()
+
+		log.FromContext(ctx).WithValues("nodeClaim", nodeClaim).V(0).Info("reconcile nodeClaim")
```
jonathan-innis, line 75: This might get really verbose -- we tend to opt to reduce our logging so that it's not overly verbose, ideally logging not just when there is a reconciliation, but specifically when it takes action against the object

pkg/controllers/nodeclaim/capacityreservation/controller.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		log.FromContext(ctx).WithValues("nodeClaim", nodeClaim, "cloudProviderNodeClaim", cloudProviderNodeClaim).V(0).Info("reconcile nodeClaim with cloudProviderNodeClaim")
+
+		nodeClaim.Labels = lo.Assign(nodeClaim.Labels, map[string]string{
+			v1beta1.LabelCapactiyReservationID: cloudProviderNodeClaim.Labels[v1beta1.LabelCapactiyReservationID],
+		})
+
+		if !equality.Semantic.DeepEqual(stored, nodeClaim) {
+			log.FromContext(ctx).WithValues("nodeClaim", nodeClaim, "stored", stored).V(0).Info("patch nodeClaim")
```
jonathan-innis, line 89: nit: I think we want to be careful to not log whole objects

pkg/operator/operator.go
```diff
@@ -149,6 +151,7 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 	launchTemplateProvider := launchtemplate.NewDefaultProvider(
 		ctx,
 		cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval),
+		cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval),
```
jonathan-innis, line 154: What's this second cache for, should it be the same as the other cache that you're passing through the capacity reservation provider?

pkg/providers/instance/instance.go
```diff
@@ -298,11 +304,16 @@ func (p *DefaultProvider) getLaunchTemplateConfigs(ctx context.Context, nodeClas
 	if err != nil {
 		return nil, fmt.Errorf("getting launch templates, %w", err)
 	}
-	requirements := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...)
-	requirements[corev1beta1.CapacityTypeLabelKey] = scheduling.NewRequirement(corev1beta1.CapacityTypeLabelKey, v1.NodeSelectorOpIn, capacityType)
+
+	// Pick 1 NodePool (instanceType + AZ) if mulitple ODCRs match it, because CreateFleet does not allow
```
jonathan-innis, line 308: I wonder if there's some work that we can do ahead of time so that we don't even create the duplicate launch templates

pkg/providers/instance/instance.go
```diff
@@ -312,6 +323,8 @@ func (p *DefaultProvider) getLaunchTemplateConfigs(ctx context.Context, nodeClas
 			launchTemplateConfigs = append(launchTemplateConfigs, launchTemplateConfig)
 		}
 	}
+
+	log.FromContext(ctx).WithValues("launchTemplateConfigs", launchTemplateConfigs).V(0).Info("getLaunchTemplateConfigs")
```
jonathan-innis, line 327: nit: Should we be logging this here?

pkg/providers/instance/instance.go
```diff
@@ -386,13 +402,21 @@ func (p *DefaultProvider) getCapacityType(nodeClaim *corev1beta1.NodeClaim, inst
 
 // filterInstanceTypes is used to provide filtering on the list of potential instance types to further limit it to those
 // that make the most sense given our specific AWS cloudprovider.
-func (p *DefaultProvider) filterInstanceTypes(nodeClaim *corev1beta1.NodeClaim, instanceTypes []*cloudprovider.InstanceType) []*cloudprovider.InstanceType {
+func (p *DefaultProvider) filterInstanceTypes(ctx context.Context, nodeClaim *corev1beta1.NodeClaim, instanceTypes []*cloudprovider.InstanceType) []*cloudprovider.InstanceType {
 	instanceTypes = filterExoticInstanceTypes(instanceTypes)
 	// If we could potentially launch either a spot or on-demand node, we want to filter out the spot instance types that
 	// are more expensive than the cheapest on-demand type.
 	if p.isMixedCapacityLaunch(nodeClaim, instanceTypes) {
 		instanceTypes = filterUnwantedSpot(instanceTypes)
```
jonathan-innis, line 410: How does this filtering get affected if there is an ODCR launch? I wonder if we want to catch this filtering before this line -- otherwise, we're going to lose some ODCR options if something just supports spot

pkg/providers/instance/instance.go
```diff
@@ -424,6 +448,30 @@ func (p *DefaultProvider) isMixedCapacityLaunch(nodeClaim *corev1beta1.NodeClaim
...
+	// requirements must allow on-demand
+	if !requirements.Get(corev1beta1.CapacityTypeLabelKey).Has(corev1beta1.CapacityTypeOnDemand) {
+		return false
+	}
+
+	requirements[corev1beta1.CapacityTypeLabelKey] = scheduling.NewRequirement(corev1beta1.CapacityTypeLabelKey, v1.NodeSelectorOpIn, corev1beta1.CapacityTypeOnDemand)
+	requirements[v1beta1.LabelCapactiyReservationID] = scheduling.NewRequirement(v1beta1.LabelCapactiyReservationID, v1.NodeSelectorOpExists)
+	for _, instanceType := range instanceTypes {
+		for _, offering := range instanceType.Offerings {
+			if requirements.Compatible(offering.Requirements, scheduling.AllowUndefinedWellKnownLabels) == nil {
```
jonathan-innis, line 464: I think we just need to ensure that the instance type has a compatible offering. That can be done just by doing the `HasCompatible()` function on the offerings and then creating an `Exists` requirement for the ODCR ID key to filter out any instance types where there is no offering that has requirements that contain that key (no ODCR offering available)

pkg/providers/instance/instance.go
```diff
@@ -450,6 +498,34 @@ func filterUnwantedSpot(instanceTypes []*cloudprovider.InstanceType) []*cloudpro
 	return instanceTypes
 }
 
+// filterOnlyCapacityReservations is used to filter only offerings that are capacity reservations
+func filterOnlyCapacityReservations(instanceTypes []*cloudprovider.InstanceType) []*cloudprovider.InstanceType {
```
jonathan-innis, line 502: I think there may be some helper functions in the `cloudprovider` package `types.go` file that might be able to help you here

pkg/providers/instance/types.go
```diff
@@ -27,28 +27,30 @@ import (
...
-	Tags             map[string]string
-	EFAEnabled       bool
+	LaunchTime            time.Time
+	State                 string
+	ID                    string
+	ImageID               string
+	Type                  string
+	Zone                  string
+	CapacityType          string
+	CapacityReservationID *string
```
jonathan-innis, line 37: I wonder if you just make this a string and treat empty as non-existent

pkg/providers/instancetype/instancetype.go
```diff
@@ -311,6 +335,52 @@ func (p *DefaultProvider) createOfferings(ctx context.Context, instanceType *ec2
...
+	for _, capacityReservation := range capacityReservations {
+		log.FromContext(ctx).WithValues("instanceType", *instanceType.InstanceType, "capacityReservation", capacityReservation.ID).V(0).Info("capacity reservation for instanceType")
+		if capacityReservation.InstanceType != *instanceType.InstanceType {
+			log.FromContext(ctx).WithValues("instanceType", *instanceType.InstanceType, "capacityReservation", capacityReservation.ID).V(0).Info("capacity reservation not applicatble for instanceType")
+			continue
+		}
+		log.FromContext(ctx).WithValues("instanceType", *instanceType.InstanceType, "capacityReservation", capacityReservation.ID).V(0).Info("capacity reservation applicatble for instanceType")
+		price := 0.0
+		isAvailable := capacityReservation.AvailableInstanceCount > 0
+		zone := capacityReservation.AvailabilityZone
```
jonathan-innis, line 349: nit: I think you may be able to inline some of this where you are creating the offering struct

pkg/providers/instancetype/instancetype.go
```diff
@@ -311,6 +335,52 @@ func (p *DefaultProvider) createOfferings(ctx context.Context, instanceType *ec2
...
+				scheduling.NewRequirement(v1.LabelTopologyZone, v1.NodeSelectorOpIn, zone),
+			),
+			Price:     price,
+			Available: available,
+		}
+		if subnet.ZoneID != "" {
+			offering.Requirements.Add(scheduling.NewRequirement(v1beta1.LabelTopologyZoneID, v1.NodeSelectorOpIn, subnet.ZoneID))
+		}
+		// TODO: tvonhacht
+		offering.Requirements.Add(scheduling.NewRequirement(v1beta1.LabelCapactiyReservationID, v1.NodeSelectorOpIn, capacityReservation.ID))
```
jonathan-innis, line 368: How do we make sure that we don't create duplicate offerings for multiple ODCRs that have the same zone and instance type?

pkg/providers/instancetype/instancetype.go
```diff
@@ -311,6 +335,52 @@ func (p *DefaultProvider) createOfferings(ctx context.Context, instanceType *ec2
...
+		log.FromContext(ctx).WithValues("instanceType", *instanceType.InstanceType, "capacityReservation", capacityReservation.ID, "offering", offering).V(0).Info("offering for capacity reservation and instanceType")
+		offerings = append(offerings, offering)
+		instanceTypeOfferingAvailable.With(prometheus.Labels{
```
jonathan-innis, line 371: I wonder if we punt on metrics for capacity reservations for instance type availability/price right now. This gets trickier since there's now no longer a concept of "global" availability and price but now availability is dictated on a per NodePool-basis, since the EC2NodeClass dictates the offerings that are surfaced for each one. Maybe we need to add a NodePool dimension to make this more accurate, but this is going to increase the cardinality even more than we already have (and this is a pretty high cardinality metric to begin with)

pkg/providers/instancetype/instancetype.go
```diff
@@ -311,6 +335,52 @@ func (p *DefaultProvider) createOfferings(ctx context.Context, instanceType *ec2
...
+			instanceTypeLabel: fmt.Sprintf("%s-%s", *instanceType.InstanceType, "capacity-reservation"),
+			capacityTypeLabel: capacityType,
+			zoneLabel:         zone,
+		}).Set(float64(lo.Ternary(available, 1, 0)))
+		instanceTypeOfferingPriceEstimate.With(prometheus.Labels{
```
jonathan-innis, line 376: Same comment here. Maybe worth punting on the metric since it's going to make things more challenging to reason about

pkg/providers/instancetype/metrics.go
```diff
@@ -64,6 +64,19 @@ var (
 			zoneLabel,
 		},
 	)
+	instanceTypeOfferingAvailableCapacityReservation = prometheus.NewGaugeVec(
```
jonathan-innis, line 67: I wonder if we punt on this too. Ideally, we can just model this through our standard offering availabiliyt

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -58,43 +58,58 @@ type Provider interface {
 	DeleteAll(context.Context, *v1beta1.EC2NodeClass) error
 	InvalidateCache(context.Context, string, string)
 	ResolveClusterCIDR(context.Context) error
+	GetCapacityReservationID(launchTemplateName string) *string
```
jonathan-innis, line 61: I wonder if rather than relying on a launch template lookup if we can just know the map of instance type/az -> capacity reservation id during launch and then know that if a given instance type/az was chosen that it is a particular capacity reservation id. This might reduce cross-provider look-ups and additional dependencies

jmdeal:
Closing in favor of #7726. Thanks a ton for the contribution here though, this PR laid great groundwork for the final PR.

PR #6213, opened 2024-05-16
chore: drop implicit public IP association config

jmdeal:
Fixes #N/A 

**Description**

Drops implicit public IP association as called out in the [v1 roadmap](https://github.com/kubernetes-sigs/karpenter/pull/1222/files#diff-055dd5f24ba5dd20e449cb25f9e8a572200b3bd059ce75b1fbaae860e6284e27R201-R205).

`make presubmit` and `/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/launchtemplate/launchtemplate.go
+++ b/pkg/providers/launchtemplate/launchtemplate.go
@@ -173,30 +173,20 @@ func (p *DefaultProvider) createAMIOptions(ctx context.Context, nodeClass *v1bet
 	if len(nodeClass.Status.SecurityGroups) == 0 {
 		return nil, fmt.Errorf("no security groups are present in the status")
 	}
-	options := &amifamily.Options{
-		ClusterName:         options.FromContext(ctx).ClusterName,
-		ClusterEndpoint:     p.ClusterEndpoint,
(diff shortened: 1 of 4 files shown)


pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -173,32 +173,20 @@ func (p *DefaultProvider) createAMIOptions(ctx context.Context, nodeClass *v1bet
 	if len(nodeClass.Status.SecurityGroups) == 0 {
 		return nil, fmt.Errorf("no security groups are present in the status")
```
jonathan-innis, line 174: How should we track docs changes with these kinds of things?

jmdeal: I couldn't find any explicit documentation for this behavior, so there are no updates required to existing docs. I'm not sure what we want to do for the upgrade guide, I could put in a placeholder not but I think we can also just rely on tracking this in the v1 roadmap and construct the upgrade guide from there.

jonathan-innis (approved):
LGTM 🚀

Issue #3878, opened 2023-05-11
Add az-id label to nodes


sidewinder12s:
### Tell us about your request

Can Karpenter Add an az-id label to nodes?

This currently cannot be done as users need to set the labels using a startup script and/or modify the bootstrap flag for node labels. 

I am not sure if the inability to modify bootstrap command flags is changing with the new machine ownership changes in 0.28

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

Our cost team needs the az-id label for pricing. I could see a lot of other uses too.

### Are you currently working around this issue?

We're not with Karpenter. An alternative is to add some kind of daemonset or controller which adds this label to nodes after join. This exposes some limitations in Prometheus metrics which can cause issues with multiple time series when you modify the labels on that series after creation.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

njtran:
We have an AZ Label on the node already, but as I understand it you want it in the form of `use1-az1` for consistency, right?

sidewinder12s:
Yup, there are a lot of use cases where you might need to get the real physical location for pricing/network/etc.

ellistarn:
Related: https://github.com/kubernetes/cloud-provider-aws/issues/300

njtran:
Would be great if we can just rely on the linked issue above to do this for us. I'm sure we could do this in Karpenter, but if it's going to be implemented in the future, it just means that the label will be redundant in the future.

sftim:
Outside Kubernetes itself, it'd be handy to have AWS define what the label key is going to be. We can get that defined and agreed even before the code to set it is written.

timchenko-a:
Does anybody know any "easy" workaround for this? E.g. getting AZ ID in UserData and passing it to kubelet somehow?

sidewinder12s:
> Does anybody know any "easy" workaround for this? E.g. getting AZ ID in UserData and passing it to kubelet somehow?

[...]

PR #6216, opened 2024-05-16
chore: add label and ability to schedule for az-id

### LINKED ISSUE #4612: Add karpenter.k8s.aws/zoneid with Availability Zone IDs
### Description

AWS maps the physical Availability Zones randomly to the Availability Zone names for each AWS account. 
if we can add `karpenter.k8s.aws/zoneid` to Well-Known Labels with the value from http://169.254.169.254/latest/meta-data/placement/availability-zone-id this will be helpful to identify AZ level issues when your apps are running in multiple AWS accounts.

rschalo:
Fixes #4612 #3878 

**Description**
This change adds the az-id label to nodes and adds the ability to schedule against az-id.
Unit tests but I'm pretty certain an additional e2e test is required but want to get some feedback on implementation.

- [ ] Yes, issue opened: # 
- [x] No


go.mod
```diff
@@ -116,3 +116,5 @@ require (
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
+
+replace sigs.k8s.io/karpenter => /Users/schalor/Documents/github.com/rschalo/karpenter
```
rschalo, line 120: This needs to be changed as soon as the update to the karpenter core is merged.

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

go.mod
jmdeal: Updated this to point directly at your fork so we could run E2E. Will still need to revert the change, and replace it with a version bump once the upstream PR is merged.

pkg/providers/instance/instance.go
```diff
@@ -377,7 +377,7 @@ func (p *DefaultProvider) getCapacityType(nodeClaim *corev1beta1.NodeClaim, inst
 	if requirements.Get(corev1beta1.CapacityTypeLabelKey).Has(corev1beta1.CapacityTypeSpot) {
 		for _, instanceType := range instanceTypes {
 			for _, offering := range instanceType.Offerings.Available() {
-				if requirements.Get(v1.LabelTopologyZone).Has(offering.Zone) && offering.CapacityType == corev1beta1.CapacityTypeSpot {
+				if requirements.Get(v1.LabelTopologyZone).Has(offering.Zone()) && offering.CapacityType() == corev1beta1.CapacityTypeSpot {
```
jmdeal, line 380: We also need to consider zone-id here right?

jmdeal (commented):
/karpenter snapshot

pkg/providers/instance/instance.go
```diff
@@ -377,7 +377,8 @@ func (p *DefaultProvider) getCapacityType(nodeClaim *corev1beta1.NodeClaim, inst
 	if requirements.Get(corev1beta1.CapacityTypeLabelKey).Has(corev1beta1.CapacityTypeSpot) {
 		for _, instanceType := range instanceTypes {
 			for _, offering := range instanceType.Offerings.Available() {
-				if requirements.Get(v1.LabelTopologyZone).Has(offering.Zone) && offering.CapacityType == corev1beta1.CapacityTypeSpot {
+				if requirements.Compatible(offering.Requirements, scheduling.AllowUndefinedWellKnownLabels) == nil &&
```
jonathan-innis, line 380: Does it make sense to just create a set of requirements that only selects on spot here and then just check compatibility rather than having to specifically call out to select it?

pkg/providers/instance/instance.go
```diff
@@ -412,12 +413,13 @@ func (p *DefaultProvider) isMixedCapacityLaunch(nodeClaim *corev1beta1.NodeClaim
...
-				if requirements.Get(v1.LabelTopologyZone).Has(offering.Zone) {
-					if offering.CapacityType == corev1beta1.CapacityTypeSpot {
-						hasSpotOfferings = true
-					} else {
-						hasODOffering = true
-					}
+				if requirements.Compatible(offering.Requirements, scheduling.AllowUndefinedWellKnownLabels) != nil {
+					continue
+				}
+				if offering.Requirements.Get(corev1beta1.CapacityTypeLabelKey).Any() == corev1beta1.CapacityTypeSpot {
```
jonathan-innis, line 419: Same comment here.

pkg/providers/subnet/subnet.go
```diff
@@ -96,20 +104,24 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
 			subnets[lo.FromPtr(output.Subnets[i].SubnetId)] = output.Subnets[i]
 			p.availableIPAddressCache.SetDefault(lo.FromPtr(output.Subnets[i].SubnetId), lo.FromPtr(output.Subnets[i].AvailableIpAddressCount))
 			p.associatePublicIPAddressCache.SetDefault(lo.FromPtr(output.Subnets[i].SubnetId), lo.FromPtr(output.Subnets[i].MapPublicIpOnLaunch))
 			// subnets can be leaked here, if a subnets is never called received from ec2
 			// we are accepting it for now, as this will be an insignificant amount of memory
 			delete(p.inflightIPs, lo.FromPtr(output.Subnets[i].SubnetId)) // remove any previously tracked IP addresses since we just refreshed from EC2
+			// add zone name to ID comparison as subnets are discovered
+			availableZones[aws.StringValue(output.Subnets[i].AvailabilityZone)] = aws.StringValue(output.Subnets[i].AvailabilityZoneId)
 		}
+		p.zoneInfo = availableZones
```
jonathan-innis, line 118: Any reason not to update the data directly rather than creating a new map and assigning it?

pkg/providers/subnet/subnet.go
```diff
@@ -96,20 +104,24 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
 	}
 	p.cache.SetDefault(fmt.Sprint(hash), lo.Values(subnets))
 	if p.cm.HasChanged(fmt.Sprintf("subnets/%s", nodeClass.Name), subnets) {
 		log.FromContext(ctx).
 			WithValues("subnets", lo.Map(lo.Values(subnets), func(s *ec2.Subnet, _ int) string {
-				return fmt.Sprintf("%s (%s)", aws.StringValue(s.SubnetId), aws.StringValue(s.AvailabilityZone))
+				return fmt.Sprintf("%s (zone: %s, zone-id: %s)", aws.StringValue(s.SubnetId), aws.StringValue(s.AvailabilityZone), aws.StringValue(s.AvailabilityZoneId))
```
jonathan-innis, line 124: I wonder if this should just be a json string to structure it now that we are specifying things like `"Node":{"name":"default"}`. Probably nice to have consistent structure here

pkg/providers/subnet/suite_test.go
```diff
@@ -103,10 +104,12 @@ var _ = Describe("SubnetProvider", func() {
 			}
 			subnets, err := awsEnv.SubnetProvider.List(ctx, nodeClass)
 			Expect(err).To(BeNil())
+			fmt.Printf("subnets: %v\n", subnets)
```
jonathan-innis, line 107: Can drop this now

test/suites/nodeclaim/nodeclaim_test.go
```diff
@@ -165,6 +165,7 @@ var _ = Describe("StandaloneNodeClaim", func() {
 		node := env.EventuallyExpectInitializedNodeCount("==", 1)[0]
 		Expect(node.Annotations).To(HaveKeyWithValue("custom-annotation", "custom-value"))
 		Expect(node.Labels).To(HaveKeyWithValue("custom-label", "custom-value"))
+		Expect(node.Labels).To(HaveKey("topology.k8s.aws/zone-id"))
```
jonathan-innis, line 168: Worth checking the value here?

pkg/providers/instancetype/instancetype.go
```diff
@@ -165,7 +167,8 @@ func (p *DefaultProvider) List(ctx context.Context, kc *corev1beta1.KubeletConfi
 		return NewInstanceType(ctx, i, p.region,
 			nodeClass.Spec.BlockDeviceMappings, nodeClass.Spec.InstanceStorePolicy,
 			kc.MaxPods, kc.PodsPerCore, kc.KubeReserved, kc.SystemReserved, kc.EvictionHard, kc.EvictionSoft,
-			amiFamily, p.createOfferings(ctx, i, p.instanceTypeOfferings[aws.StringValue(i.InstanceType)], allZones, subnetZones))
+			amiFamily, p.createOfferings(ctx, i, p.instanceTypeOfferings[aws.StringValue(i.InstanceType)], allZones, subnetZones, p.subnetProvider.ZoneInfo()),
```
jonathan-innis, line 170: Do we need to call ZoneInfo() here or can we just use the data that we get back directly from the subnet provider on the List() call?

pkg/providers/instancetype/types.go
```diff
@@ -78,11 +78,15 @@ func computeRequirements(info *ec2.InstanceTypeInfo, offerings cloudprovider.Off
 		scheduling.NewRequirement(v1.LabelInstanceTypeStable, v1.NodeSelectorOpIn, aws.StringValue(info.InstanceType)),
 		scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, getArchitecture(info)),
 		scheduling.NewRequirement(v1.LabelOSStable, v1.NodeSelectorOpIn, getOS(info, amiFamily)...),
-		scheduling.NewRequirement(v1.LabelTopologyZone, v1.NodeSelectorOpIn, lo.Map(offerings.Available(), func(o cloudprovider.Offering, _ int) string { return o.Zone })...),
+		scheduling.NewRequirement(v1.LabelTopologyZone, v1.NodeSelectorOpIn, lo.Map(offerings.Available(), func(o cloudprovider.Offering, _ int) string {
+			return o.Requirements.Get(v1.LabelTopologyZone).Any()
```
jonathan-innis, line 82: Should it be any or should the semantic be to take all of the values and merge them?

test/suites/integration/scheduling_test.go
```diff
@@ -581,6 +601,81 @@ var _ = Describe("Scheduling", Ordered, ContinueOnFailure, func() {
...
+				},
+			})
+
+			// zoneIDMapping is a mapping from zone to zone-id for all zoneIDMapping available in the provided subnets
+			zoneIDMapping := lo.PickByKeys(env.GetZoneIDMapping(), lo.Keys(env.GetSubnets(map[string]string{"karpenter.sh/discovery": env.ClusterName})))
+			pods := lo.MapToSlice(zoneIDMapping, func(zone, zoneID string) *v1.Pod {
+				return test.Pod(test.PodOptions{
+					NodeRequirements: []v1.NodeSelectorRequirement{
+						{
+							Key:      expectedZoneLabel,
```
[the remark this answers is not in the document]
jmdeal: I wasn't sure if CCM would overwrite the zone-id label if we were incorrect. If it doesn't we could probably go entirely in-memory, but I went this route to be safe.

pkg/apis/v1beta1/labels.go
```diff
@@ -94,6 +95,8 @@ var (
 
 	LabelNodeClass = Group + "/ec2nodeclass"
 
+	LabelTopologyZoneID = "topology.k8s.aws/zone-id"
```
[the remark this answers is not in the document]
jmdeal: Not sure what we would do to quantify if our dependency change is too large. Seems reasonable to depend on CCM to ingest the constant.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -316,6 +316,7 @@ func (c *CloudProvider) instanceToNodeClaim(i *instance.Instance, instanceType *
 		nodeClaim.Status.Allocatable = functional.FilterMap(instanceType.Allocatable(), resourceFilter)
 	}
 	labels[v1.LabelTopologyZone] = i.Zone
+	labels[v1beta1.LabelTopologyZoneID] = c.subnetProvider.ZoneInfo()[i.Zone]
```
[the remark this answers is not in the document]
jmdeal: I didn't think this raced because I thought pointer writes were atomic, but it doesn't look like that's guaranteed by the spec. As for why it's not on Instance, @rschalo can correct me if I'm wrong but I think the main reason is because it's not available as part of the DescribeInstances output. We'd have to populate it from either the EC2NodeClass or the subnet provider inside NewInstance. IMO this is the cleaner approach.

pkg/providers/subnet/suite_test.go
jmdeal: That's my bad, missed dropping this in my latest commit. Will remove.

pkg/providers/instance/instance.go
```diff
@@ -298,9 +298,10 @@ func (p *DefaultProvider) getLaunchTemplateConfigs(ctx context.Context, nodeClas
 	if err != nil {
 		return nil, fmt.Errorf("getting launch templates, %w", err)
 	}
+	requirements := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...)
```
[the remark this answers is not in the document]
jmdeal: That seems reasonable to me. This also gives justification to not inline it since it needs to be mutated.

pkg/providers/instance/instance.go
jmdeal: I was honestly fine leaving it like this since it's slightly more verbose to inject the requirement. Seems fine either way though.

pkg/apis/v1beta1/ec2nodeclass_status.go
```diff
@@ -27,6 +27,9 @@ type Subnet struct {
 	// The associated availability zone
 	// +required
 	Zone string `json:"zone"`
+	// The associated availability zone ID
+	// +optional
+	ZoneID string `json:"zoneId"`
```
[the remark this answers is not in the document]
rschalo: Done.

pkg/apis/v1beta1/labels.go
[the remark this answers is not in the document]
rschalo: It's available here: https://github.com/kubernetes/cloud-provider-aws/blob/389877c071efb56481b4fc0265ea251081696092/pkg/providers/v1/well_known_labels.go#L22. Also not sure what the impact would be to the dependency chain.

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
rschalo: Correct that the ID is not returned as part of the DescribesInstances call. Before we return the instance from the instance provider, we could add the ID there. Rather than reset it every time we describe subnets, would it work better to build out all of the zones for the account? That way, even if the subnets change, the account specific mapping remains.

Additionally, we could make an EC2 feature request for this to be returned as part of this request.

pkg/providers/instance/instance.go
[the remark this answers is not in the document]
rschalo: Inlined.

pkg/providers/instancetype/instancetype.go
rschalo: Done.

pkg/apis/v1beta1/labels.go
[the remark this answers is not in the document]
jmdeal: Decided against taking in the additional dependency since this lablel is not expected to change.

pkg/providers/subnet/subnet.go
jmdeal: Found that this is actually wrong because we overwrite the global provider map each time we list a set of filtered subnets. In reality, this should be merged each time.

pkg/providers/instancetype/types.go
jmdeal: Discussed with @rschalo and @njtran, since we treat this as an invariant it makes sense to continue treating this as an invariant here. If we want to merge here, we should rationalize how to merge throughout the rest of the provider. I think we can decouple this and revisit if there is a need at a later point.

pkg/providers/subnet/subnet.go
```diff
@@ -69,6 +75,8 @@ func NewDefaultProvider(ec2api ec2iface.EC2API, cache *cache.Cache, availableIPA
 		associatePublicIPAddressCache: associatePublicIPAddressCache,
 		// inflightIPs is used to track IPs from known launched instances
 		inflightIPs: map[string]int64{},
+		// availabilityZoneIDs is used to track possible az IDs based on discovered subnets
+		zoneInfo: map[string]string{},
```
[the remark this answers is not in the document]
jmdeal: We need the zone-id mapping in the cloudprovider `Get` and `List` calls when populating labels on the nodeclaim. I've realized this is actually racey even in our current implementation since we don't know for sure if the EC2NodeClasses have resolved yet. Solving this by introducing a `ZoneProvider`, whose sole responsibility is tracking this mapping. This will have minimal API usage impact since we only need to call this once.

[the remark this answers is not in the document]
jmdeal: Actually realizing it's probably fine to leverage the NodeClass in both cases. By doing this, in the case of upgrade we will fail to provision nodes for pods with zone-id requirements. Once the subnet status is re-reconciled after a minute, we will begin to succeed.

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
jmdeal: Came up with a slightly different solution, we're relying completely on resolution from the NodeClass. In the create path (where we really need this label) we've already validated that the nodeclass exists and can continue to leverage it. In the List and Get are only used for garbage collection and node termination respectively, in which case it doesn't matter if we can't populate this label because we can't resolve the nodeclass. IMO this is a good solution which eliminates the need for tracking the zone-id mapping in the subnet provider at all.

pkg/providers/instance/instance.go
jmdeal: The semantic here is slightly different, three total paths. One if we're incompatible, one where we're compatible and the offering is spot, and one where we're compatible and the offering is on-demand.

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

PR #6258, opened 2024-05-22
test: add ko build -race flag for e2e testing

njtran:
Fixes #N/A 

**Description**
Cutting this to see if the race detector would have detected a discovered race

testing here

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -2,3 +2,8 @@ defaultBaseImage: public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-bas
 defaultPlatforms:
   - linux/arm64
   - linux/amd64
+builds: 
+- ldflags:
+  - -race
+  id: main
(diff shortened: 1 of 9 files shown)


njtran (commented):
/karpenter snapshot

.ko.yaml
```diff
@@ -2,3 +2,8 @@ defaultBaseImage: public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-bas
 defaultPlatforms:
   - linux/arm64
   - linux/amd64
+builds:
```
jonathan-innis, line 5: We don't want to build it like this always, right? Only in testing? How do we ensure that this only gets added when we are building a testing snapshot?

jonathan-innis: We may want to actually build a completely separate image from the ones that we vend out to customers for this kind of thing

PR #6267, opened 2024-05-24
docs: add FAQ entry clarifying that startup taints are necessary to avoid excess node provisioning

wmgroot:
**Description**
I spent several days tracking down why Karpenter would create several unnecessary nodes in order to schedule pending pods in our clusters. Ultimately, I discovered this behavior was occurring.

1. Karpenter sees a pending pod that cannot be scheduled without a new node.
2. Karpenter creates a new nodeclaim, followed by a new node joining the cluster.
3. Karpenter's provisioning logic correctly determines that the pending node will satisfy the still pending pod.
4. Cilium's agent daemonset starts up on the new unready node, setting a startup taint to block pods from scheduling until the CNI is ready.
5. Karpenter sees that the still pending pod can no longer be scheduled to the node due to the cilium not-ready taint and provisions a second nodeclaim.
6. The original new node becomes ready and Cilium removes the taint.
7. The pending pod schedules to the original new node.
8. Karpenter removes the unnecessary second node via emptiness/consolidation. 

We found that Karpenter has docs on configuring startupTaints to handle this case.
https://karpenter.sh/docs/concepts/nodepools/#cilium-startup-taint

I'd like to add these documentation updates to the Karpenter FAQ to save users from having to debug the code in order to discover the solution to this problem in the future.

### DIFF
--- a/website/content/en/preview/contributing/documentation-updates.md
+++ b/website/content/en/preview/contributing/documentation-updates.md
@@ -1,0 +1,11 @@
+---
+title: "Documentation Updates"
+linkTitle: "Documentation Updates"
+weight: 50
+description: >
+  Infomration helpful for contributing simple documentation updates.
+---
+
+- Documentation for https://karpenter.sh/docs/ is built under website/content/en/preview/.
+- Documentation updates should be made to the "preview" directory. Your changes will be promoted to website/content/en/docs/ by an automated process after the change has been merged.
+- Previews for your changes are built and available a few minutes after you push. Look for the "netlify Deploy Preview" link in a comment in your PR.
--- a/website/content/en/preview/faq.md
+++ b/website/content/en/preview/faq.md
@@ -138,6 +138,25 @@ For more documentation on enabling IPv6 with the Amazon VPC CNI, see the [docs](
 Windows nodes do not support IPv6.
 {{% /alert %}}
 
+### Why do I see extra nodes get launched to schedule pending pods that remain empty and are later removed?
+
+You might have a daemonset, userData configuration, or some other workload that applies a taint after a node is provisioned. After the taint is applied, Karpenter will detect that the pod cannot be scheduled to this new node due to the added taint. As a result, Karpenter will provision yet another node. Typically, the original node has the taint removed and the pod schedules to it, leaving the extra new node unused and reaped by emptiness/consolidation. If the taint is not removed quickly enough, Karpenter may remove the original node before the pod can be scheduled via emptiness consolidation. This could result in an infinite loop of nodes being provisioned and consolidated without the pending pod ever scheduling.
+
+The solution is to configure [startupTaints]({{<ref "./concepts/nodepools/#cilium-startup-taint" >}}) to make Karpenter aware of any temporary taints that are needed to ensure that pods do not schedule on nodes that are not yet ready to receive them.
+
+Here's an example for Cilium's startup taint.
+```
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+...
+spec:
+  template:
+    spec:
+      startupTaints:
+        - key: node.cilium.io/agent-not-ready
+          effect: NoSchedule
+```
+
 ## Scheduling
 
 ### When using preferred scheduling constraints, Karpenter launches the correct number of nodes at first.  Why do they then sometimes get consolidated immediately?
--- a/website/content/en/preview/concepts/nodepools.md
+++ b/website/content/en/preview/concepts/nodepools.md
@@ -72,7 +72,7 @@ spec:
           operator: In
           values: ["c", "m", "r"]
           # minValues here enforces the scheduler to consider at least that number of unique instance-category to schedule the pods.
-          # This field is ALPHA and can be dropped or replaced at any time 
+          # This field is ALPHA and can be dropped or replaced at any time
           minValues: 2
         - key: "karpenter.k8s.aws/instance-family"
           operator: In
@@ -555,6 +555,8 @@ In order for a pod to run on a node defined in this NodePool, it must tolerate `
 ### Cilium Startup Taint
 
 Per the Cilium [docs](https://docs.cilium.io/en/stable/installation/taints/#taint-effects), it's recommended to place a taint of `node.cilium.io/agent-not-ready=true:NoExecute` on nodes to allow Cilium to configure networking prior to other pods starting.  This can be accomplished via the use of Karpenter `startupTaints`.  These taints are placed on the node, but pods aren't required to tolerate these taints to be considered for provisioning.
+
+Failure to provide accurate `startupTaints` can result in Karpenter continually provisioning new nodes. When the new node joins and the startup taint that Karpenter is unaware of is added, Karpenter now considers the pending pod to be unschedulable to this node. Karpenter will attempt to provision yet another new node to schedule the pending pod.
 
 ```yaml
 apiVersion: karpenter.sh/v1beta1


website/content/en/docs/concepts/nodepools.md
```diff
@@ -556,6 +556,8 @@ In order for a pod to run on a node defined in this NodePool, it must tolerate `
 
 Per the Cilium [docs](https://docs.cilium.io/en/stable/installation/taints/#taint-effects), it's recommended to place a taint of `node.cilium.io/agent-not-ready=true:NoExecute` on nodes to allow Cilium to configure networking prior to other pods starting.  This can be accomplished via the use of Karpenter `startupTaints`.  These taints are placed on the node, but pods aren't required to tolerate these taints to be considered for provisioning.
 
+Failure to provide accurate `startupTaints` can result in Karpenter double provisioning new nodes when it considers a pod unschedulable during the period that a temporary taint untracked by Karpenter is active.
```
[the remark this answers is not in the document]
wmgroot: It's continual provisioning, but in the case of cilium, the taint is removed and the pod schedules before a 3rd cycle can occur.

website/content/en/docs/faq.md
```diff
@@ -138,6 +138,25 @@ For more documentation on enabling IPv6 with the Amazon VPC CNI, see the [docs](
 Windows nodes do not support IPv6.
 {{% /alert %}}
 
+### Why do I see extra nodes get launched to schedule pending pods that remain empty and are later removed?
```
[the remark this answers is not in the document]
wmgroot: If the first node is reaped as you describe, I'd expect the pod to never schedule since that cycle should continue forever. I think it comes down to if the temporary startup taint is removed quickly enough to allow the pending pod to schedule before the node is reaped due to emptiness consolidation.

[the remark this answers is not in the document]
wmgroot: I'll add a note about this case.

website/content/en/docs/faq.md
```diff
@@ -138,6 +138,25 @@ For more documentation on enabling IPv6 with the Amazon VPC CNI, see the [docs](
...
+
+You might have a daemonset, userData configuration, or some other workload that applies a taint after a node is provisioned. After the taint is applied, Karpenter will detect that the pod cannot be scheduled to this new node due to the added taint. As a result, Karpenter will provision yet another node. Typically, the original node has the taint removed and the pod schedules to it, leaving the extra new node unused and reaped by emptiness/consolidation. If the taint is not removed quickly enough, Karpenter may remove the original node before the pod can be scheduled via emptiness consolidation. This could result in an infinite loop of nodes being provisioned and consolidated without the pending pod ever scheduling.
```
jonathan-innis, line 143: Don't need to update this in this statement, but I'd also just briefly mention that you probably don't want to be adding taints after the fact after node creation. You're sort-of relying on the fact that you can be fast enough to beat pods to scheduling. Maybe CNI is fine because the node itself won't go ready without the networking up and ready, but generally speaking you want these taints guaranteed to be there on node creation so that you are just responsible for the removal, not the addition

website/content/en/docs/concepts/nodepools.md
```diff
@@ -72,7 +72,7 @@ spec:
           operator: In
```
jonathan-innis, line 72: Can you update these docs to be in the preview directory as opposed to the `docs` directory. I admit it's a bit confusing (and we need to eventually change it) but updates to the docs go into the preview directory ahead of them being formally released in docs and in the latest version

website/content/en/docs/faq.md
wmgroot: Yep, we had the same conversation and conclusion internally when I understood what was going on.

website/content/en/preview/faq.md
```diff
@@ -26,7 +26,7 @@ Karpenter has multiple mechanisms for configuring the [operating system]({{< ref
...
 
 ### What RBAC access is required?
-All the required RBAC rules can be found in the Helm chart template. See [clusterrole-core.yaml](https://github.com/aws/karpenter/blob{{< githubRelRef >}}charts/karpenter/templates/clusterrole-core.yaml), [clusterrole.yaml](https://github.com/aws/karpenter/blob{{< githubRelRef >}}charts/karpenter/templates/clusterrole.yaml), [rolebinding.yaml](https://github.com/aws/karpenter/blob{{< githubRelRef >}}charts/karpenter/templates/rolebinding.yaml), and [role.yaml](https://github.com/aws/karpenter/blob{{< githubRelRef >}}charts/karpenter/templates/role.yaml) files for details.
+All the required RBAC rules can be found in the Helm chart template. See [clusterrole-core.yaml](https://github.com/aws/karpenter/blob/v0.37.0/charts/karpenter/templates/clusterrole-core.yaml), [clusterrole.yaml](https://github.com/aws/karpenter/blob/v0.37.0/charts/karpenter/templates/clusterrole.yaml), [rolebinding.yaml](https://github.com/aws/karpenter/blob/v0.37.0/charts/karpenter/templates/rolebinding.yaml), and [role.yaml](https://github.com/aws/karpenter/blob/v0.37.0/charts/karpenter/templates/role.yaml) files for details.
```
jonathan-innis, line 29: I think we made more changes here than we intended to make. We need to keep these template parameters around so that we correctly create the new version from the preview directory

website/content/en/preview/faq.md
```diff
@@ -179,10 +198,10 @@ Yes, see the [KubeletConfiguration Section in the NodePool docs]({{<ref "./conce
 The difference between the Core and Full variants is that Core is a minimal OS with less components and no graphic user interface (GUI) or desktop experience.
 `Windows2019` and `Windows2022` AMI families use the Windows Server Core option for simplicity, but if required, you can specify a custom AMI to run Windows Server Full.
 
-You can specify the [Amazon EKS optimized AMI](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html) with Windows Server 2022 Full for Kubernetes {{< param "latest_k8s_version" >}} by configuring an `amiSelector` that references the AMI name.
+You can specify the [Amazon EKS optimized AMI](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html) with Windows Server 2022 Full for Kubernetes 1.29 by configuring an `amiSelector` that references the AMI name.
 ```yaml
 amiSelectorTerms:
-    - name: Windows_Server-2022-English-Full-EKS_Optimized-{{< param "latest_k8s_version" >}}*
+    - name: Windows_Server-2022-English-Full-EKS_Optimized-1.29*
```
jonathan-innis, line 204: Same comment here

website/content/en/preview/faq.md
```diff
@@ -179,10 +198,10 @@ Yes, see the [KubeletConfiguration Section in the NodePool docs]({{<ref "./conce
 The difference between the Core and Full variants is that Core is a minimal OS with less components and no graphic user interface (GUI) or desktop experience.
 `Windows2019` and `Windows2022` AMI families use the Windows Server Core option for simplicity, but if required, you can specify a custom AMI to run Windows Server Full.
 
-You can specify the [Amazon EKS optimized AMI](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html) with Windows Server 2022 Full for Kubernetes {{< param "latest_k8s_version" >}} by configuring an `amiSelector` that references the AMI name.
+You can specify the [Amazon EKS optimized AMI](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-windows-ami.html) with Windows Server 2022 Full for Kubernetes 1.29 by configuring an `amiSelector` that references the AMI name.
```
jonathan-innis, line 201: And here

website/content/en/preview/contributing/documentation-updates.md
```diff
@@ -0,0 +1,11 @@
+---
```
jonathan-innis, line 1: Rather than creating a contributing directory, does it make more sense to create a CONTRIBUTING.md file like in https://github.com/kubernetes-sigs/karpenter/blob/main/CONTRIBUTING.md

jonathan-innis (commented):
Thanks for the docs guidance update! That should be a big help to point people to!

website/content/en/preview/faq.md
wmgroot: Ah, my bad, made an assumption that I could copy the entire file over to the preview directory, I'll get that cleaned up.

website/content/en/preview/contributing/documentation-updates.md
wmgroot: I'm worried that having contribution documentation spread across multiple locations will reduce the chance of users finding the correct documentation.

jonathan-innis (approved):
LGTM 🚀

PR #6294, opened 2024-05-30
feat: Add ability to select instances by additonal capabilities

### LINKED ISSUE #4367: How does Karpenter deal with the new `m7i-flex` instance type?

### LINKED ISSUE #6122: Request: Boolean field for "network-optimized" instances
### Description

Per @jonathan-innis in #4367 , filing to request a boolean field for network-optimized instances.
In #4367 , he suggests the name `karpenter.k8s.aws/instance-capability-network-optimized`.

Currently, one can accomplish something like this via the integer field `karpenter.k8s.aws/instance-network-bandwidth`, but a constraint for `GT 12000` is much trickier and not nearly as user-friendly as a simple boolean. Who wants to parse all of the network-bandwidths from https://karpenter.sh/docs/reference/instance-types/ ? 

It's a nice-to-have.

NetanelK:
Fixes #4367  
Fixes #6122

**Description**
This PR adds 7 instance capabilities selectors based on [AWS Naming Conventions](https://docs.aws.amazon.com/ec2/latest/instancetypes/instance-type-names.html)

As Karpenter supports up to 30 requirements, instancetype suite test doesn't pass, refer to issue kubernetes-sigs/karpenter#1270

- [ ] Yes, issue opened: # 
- [ ] No


jonathan-innis:
@NetanelK I think we want to keep the number of capabilities that we add here to be tightly scoped here to start. It's probably best if we broke out the PRs for each of the "capabilities" that we add so that we can consider them individually. How do you feel about opening separate PRs -- one for flex capability and another for the network optimized capability?

NetanelK:
> @NetanelK I think we want to keep the number of capabilities that we add here to be tightly scoped here to start. It's probably best if we broke out the PRs for each of the "capabilities" that we add so that we can consider them individually. How do you feel about opening separate PRs -- one for flex capability and another for the network optimized capability?

It's reasonable to separate flex capacity from the other ones, I'll do that, but I think that most of the capacities can be added right away instead of waiting for specific requests, as the regex includes them all.

I can drop the `q` and `z` capabilities if you want as they are less used, the other ones like `b`, `d`, `e`, and `n` are used more (in my workplace we need to be able to get the `n` and `d` instances), so I would like to keep them.

jonathan-innis:
> can be added right away instead of waiting for specific requests, as the regex includes them all

I'd prefer not using the implementation as a reason to add externally facing API. I'm agreed that we can add a mechanism (at a minimum) to select (or deselect) flex. I'm still in favor of just scoping this PR to do _that_, then we don't have to block on a bunch of other discussion around the merits of selecting on storage-optimized or network-optimized and the use-cases for them.

jonathan-innis:
I'm trying to do that careful balance of trying not to generate too much churn for you @NetanelK while also making sure that we are accurately understanding and capturing use-cases for new externally-facing API. Apologize that this will create more churn for you, but I do think that it should make the audit trail of each of these features going in easier to track and should facilitate more discussion from the community.

PR #6299, opened 2024-05-30
docs: Add instance-type-pricing design document

awscsober:
**Description**

This feature aims to provide flexibility and precision in cost management by allowing users to specify custom pricing per instance type via a ConfigMap.

### DIFF
--- a/designs/instance-types-pricing.md
+++ b/designs/instance-types-pricing.md
@@ -1,0 +1,88 @@
+## Design Document for Karpenter: Instance Pricing Override via ConfigMap
+
+### Motivation
+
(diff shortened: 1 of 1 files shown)


designs/instance-types-pricing.md
```diff
@@ -0,0 +1,88 @@
...
+The ConfigMap will contain nested key-value pairs where the top-level key represents the instance family, cpu generation and the nested keys represent the instance types with their custom prices. The structure of the ConfigMap will be as follows:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: custom-instance-pricing
+  namespace: karpenter
+data:
+  t:
```
jonathan-innis, line 43: How does this kind of format work for things that are in the format `c6in.xlarge`? Does it make more sense to just do a raw list of instance types with keys "t3.micro", "m5.large", etc. and just remove th indentation?

designs/instance-types-pricing.md
```diff
@@ -0,0 +1,88 @@
+## Design Document for Karpenter: Instance Pricing Override via ConfigMap
```
ellistarn, line 1: I think we likely want to revive the proposal here to meet the pricing use case, among the others within the same problem space: https://github.com/aws/karpenter-provider-aws/pull/2390.

PR #6315, opened 2024-06-03
test: deflake utilization test

jmdeal:
Fixes #N/A 

**Description**
We've been seeing flakes in out utilization test due to instances failing to register with the cluster. We believe this is due to the `t3` instances, so this PR removes this instance specification and adds a CPU count requirement.

`/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/suites/integration/utilization_test.go
+++ b/test/suites/integration/utilization_test.go
@@ -38,9 +38,9 @@ var _ = Describe("Utilization", Label(debug.NoWatch), Label(debug.NoEvents), fun
 		test.ReplaceRequirements(nodePool,
 			corev1beta1.NodeSelectorRequirementWithMinValues{
 				NodeSelectorRequirement: v1.NodeSelectorRequirement{
-					Key:      v1.LabelInstanceTypeStable,
+					Key:      v1beta1.LabelInstanceCPU,
 					Operator: v1.NodeSelectorOpIn,
-					Values:   []string{"t3.small"},
+					Values:   []string{"2"},
 				},
 			},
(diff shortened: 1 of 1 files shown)


jmdeal (commented):
/karpenter snapshot

test/suites/integration/utilization_test.go
```diff
@@ -38,9 +38,9 @@ var _ = Describe("Utilization", Label(debug.NoWatch), Label(debug.NoEvents), fun
 		test.ReplaceRequirements(nodePool,
 			corev1beta1.NodeSelectorRequirementWithMinValues{
 				NodeSelectorRequirement: v1.NodeSelectorRequirement{
-					Key:      v1.LabelInstanceTypeStable,
+					Key:      v1beta1.LabelInstanceCPU,
```
jonathan-innis, line 41: Should we just do CMR?

jonathan-innis: Is there a need to specify any additional requirements at all?

jmdeal: At the very least we don't want `t` instance types. I don't think theres a specific reason to use CMR, but it seems reasonable to stick with our defaults. This test is really just meant to check bin-packing.

jonathan-innis: Given that, can we just drop all overrides? Or do we need the CPU requirement so we only schedule one per node? If we do need it for this, can we just use anti-affinity to achieve the same thing?

PR #6316, opened 2024-06-03
chore: Paginate resources with aws-sdk-go-v2 pagination (aws#6268)

### LINKED ISSUE #6268: Resource package should utilize the `aws-sdk-v2` Pagination
### Description

The karpenter team currently manually paginates list of resource our sweeper and counter scripts. We should update to using [aws-sdk-go-v2 pagination](https://aws.github.io/aws-sdk-go-v2/docs/making-requests/#using-paginators), as opposed to manually paginating each API call. Examples can be found here:
- https://github.com/aws/karpenter-provider-aws/blob/5c3dc152bb1cf71912ae46a1d7c3d0107ab638f7/test/hack/resource/pkg/resourcetypes/launchtemplate.go#L46
- https://github.com/aws/karpenter-provider-aws/blob/5c3dc152bb1cf71912ae46a1d7c3d0107ab638f7/test/hack/resource/pkg/resourcetypes/stack.go#L46

LPetro:
Fixes #6268

**Description**
Updated to using [aws-sdk-go-v2 pagination](https://aws.github.io/aws-sdk-go-v2/docs/making-requests/#using-paginators) from manually paginating each API call.

Local testing for expected count and clean behavior for each resource changed.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/hack/resource/pkg/resourcetypes/eni.go
+++ b/test/hack/resource/pkg/resourcetypes/eni.go
@@ -127,11 +127,8 @@ func (e *ENI) Cleanup(ctx context.Context, ids []string) ([]string, error) {
 	return deleted, errs
 }
 
-// getAllENIs returns all ENIs that match the provided input. We are using a page size of 100 to match our default page size.
 func (e *ENI) getAllENIs(ctx context.Context, params *ec2.DescribeNetworkInterfacesInput) (enis []ec2types.NetworkInterface, err error) {
-	paginator := ec2.NewDescribeNetworkInterfacesPaginator(e.ec2Client, params, func(o *ec2.DescribeNetworkInterfacesPaginatorOptions) {
-		o.Limit = 100
-	})
+	paginator := ec2.NewDescribeNetworkInterfacesPaginator(e.ec2Client, params)
 
 	for paginator.HasMorePages() {
 		page, err := paginator.NextPage(ctx)
--- a/test/hack/resource/pkg/resourcetypes/eni.go
+++ b/test/hack/resource/pkg/resourcetypes/eni.go
@@ -43,95 +43,67 @@ func (e *ENI) Global() bool {
 }
 
 func (e *ENI) GetExpired(ctx context.Context, expirationTime time.Time, excludedClusters []string) (ids []string, err error) {
-	var nextToken *string
-	for {
-		out, err := e.ec2Client.DescribeNetworkInterfaces(ctx, &ec2.DescribeNetworkInterfacesInput{
-			Filters: []ec2types.Filter{
-				{
-					Name:   lo.ToPtr("tag-key"),
-					Values: []string{k8sClusterTag},
-				},
+	enis, err := e.getAllENIs(ctx, &ec2.DescribeNetworkInterfacesInput{
+		Filters: []ec2types.Filter{
+			{
+				Name:   lo.ToPtr("tag-key"),
+				Values: []string{k8sClusterTag},
 			},
-			NextToken: nextToken,
+		},
+	})
+	if err != nil {
+		return ids, err
+	}
+
+	for _, ni := range enis {
+		clusterName, found := lo.Find(ni.TagSet, func(tag ec2types.Tag) bool {
+			return *tag.Key == k8sClusterTag
 		})
+		if found && slices.Contains(excludedClusters, lo.FromPtr(clusterName.Value)) {
+			continue
+		}
+		creationDate, found := lo.Find(ni.TagSet, func(tag ec2types.Tag) bool {
+			return *tag.Key == "node.k8s.amazonaws.com/createdAt"
+		})
+		if !found {
+			continue
+		}
+		creationTime, err := time.Parse(time.RFC3339, *creationDate.Value)
 		if err != nil {
-			return ids, err
+			continue
 		}
-
-		for _, ni := range out.NetworkInterfaces {
-			clusterName, found := lo.Find(ni.TagSet, func(tag ec2types.Tag) bool {
-				return *tag.Key == k8sClusterTag
-			})
-			if found && slices.Contains(excludedClusters, lo.FromPtr(clusterName.Value)) {
-				continue
-			}
-			creationDate, found := lo.Find(ni.TagSet, func(tag ec2types.Tag) bool {
-				return *tag.Key == "node.k8s.amazonaws.com/createdAt"
-			})
-			if !found {
-				continue
-			}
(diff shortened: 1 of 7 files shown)


test/hack/resource/pkg/resourcetypes/securitygroup.go
```diff
@@ -43,96 +43,69 @@ func (sg *SecurityGroup) Global() bool {
...
 func (sg *SecurityGroup) GetExpired(ctx context.Context, expirationTime time.Time, excludedClusters []string) (ids []string, err error) {
-	var nextToken *string
-	for {
-		out, err := sg.ec2Client.DescribeSecurityGroups(ctx, &ec2.DescribeSecurityGroupsInput{
-			Filters: []ec2types.Filter{
-				{
-					Name:   lo.ToPtr("group-name"),
-					Values: []string{"security-group-drift"},
-				},
+	sgs, err := sg.getAllSecurityGroups(ctx, &ec2.DescribeSecurityGroupsInput{
```
engedaam, line 46: Have you considered just passing in `ec2types.Filter` instead of the whone `DescribeSecurityGroupsInput`?

test/hack/resource/pkg/resourcetypes/eni.go
```diff
@@ -43,95 +43,67 @@ func (e *ENI) Global() bool {
...
+		Filters: []ec2types.Filter{
+			{
+				Name:   lo.ToPtr("tag-key"),
+				Values: []string{k8sClusterTag},
 			},
-			NextToken: nextToken,
+		},
+	})
+	if err != nil {
+		return nil, err
```
engedaam, line 55: Why pass `nil` here instead of `ids`?

test/hack/resource/pkg/resourcetypes/eni.go
```diff
@@ -154,3 +126,19 @@ func (e *ENI) Cleanup(ctx context.Context, ids []string) ([]string, error) {
 
 	return deleted, errs
 }
+
+func (e *ENI) getAllENIs(ctx context.Context, params *ec2.DescribeNetworkInterfacesInput) (enis []ec2types.NetworkInterface, err error) {
+	paginator := ec2.NewDescribeNetworkInterfacesPaginator(e.ec2Client, params, func(o *ec2.DescribeNetworkInterfacesPaginatorOptions) {
+		o.Limit = 100
```
engedaam, line 132: Add a comment to why we are setting the page size to 100

test/hack/resource/pkg/resourcetypes/oidc.go
```diff
@@ -110,3 +106,17 @@ func (o *OIDC) Cleanup(ctx context.Context, arns []string) ([]string, error) {
 	}
 	return deleted, errs
 }
+
+// Retrieve all the OpenID Connect (OIDC) providers in the current AWS account.
+// The AWS IAM service has a default soft limit of 100 OIDC providers per account.
+// The ListOpenIDConnectProviders API call used in this function automatically handles pagination,
+// so the function can retrieve more than 100 OIDC providers if needed, without the caller
+// having to worry about pagination.
+func (o *OIDC) getAllOIDCs(ctx context.Context) (oidcs []iamtypes.OpenIDConnectProviderListEntry, err error) {
```
engedaam, line 115: The more I think about it. I don't think this is necessary

test/hack/resource/pkg/resourcetypes/stack.go
```diff
@@ -42,80 +42,51 @@ func (s *Stack) Global() bool {
...
 func (s *Stack) GetExpired(ctx context.Context, expirationTime time.Time, excludedClusters []string) (names []string, err error) {
-	var nextToken *string
-	for {
-		out, err := s.cloudFormationClient.DescribeStacks(ctx, &cloudformation.DescribeStacksInput{
-			NextToken: nextToken,
-		})
-		if err != nil {
-			return names, err
-		}
+	stacks, err := s.getAllStacks(ctx, &cloudformation.DescribeStacksInput{})
```
engedaam, line 45: nit, since we don't filter, do we need to add an Input?
```suggestion
	stacks, err := s.getAllStacks(ctx)
```

engedaam (commented):
Nice Work!!!

test/hack/resource/pkg/resourcetypes/oidc.go
LPetro: Ok, I'll leave oidc.go as it was.

test/hack/resource/pkg/resourcetypes/eni.go
LPetro: My thought was that `ids` hadn't changed at this point, so I was thinking it could be clearer to say `nil` instead of `ids.` Based on the other functions (and other resources) I'll change this to `ids` to match those for consistency.

test/hack/resource/pkg/resourcetypes/securitygroup.go
LPetro: I hadn't because all the `Paginator` functions take in their respective `Input`, and I thought it was clearer to pass `Input` in instead of the struct field `Filter` into the `getAll<Resource>` functions. 

It also felt more general, if down the road, we wanted to `getAll<Resource>` based on other fields of the struct (like `MaxResults`), but I'm not sure if that necessary to consider in these cases.

Do you think it would be better to pass in an optional `filter` across all the resources instead? Something like: 
```
func (sg *SecurityGroup) getAllSecurityGroups(ctx context.Context, filters ...ec2types.Filter) (sgs []ec2types.SecurityGroup, err error) {
    params := &ec2.DescribeSecurityGroupsInput{
        Filters: filters,
    }
```

test/hack/resource/pkg/resourcetypes/stack.go
LPetro: This is related a bit to my design decision from above, from your question about passing in `Filter` instead of `Input`. I could wrap the empty `Input` into the getAllStacks function, since no other function in `stack.go` passes in a non-empty `Input`, I was otherwise just maintaining consistency with the other `getAll<Resource>` functions.

engedaam: I think we can drop it in this context. We should only pass something if we are going to use it. It would be easy to add back in later.

test/hack/resource/pkg/resourcetypes/securitygroup.go
engedaam: That works with me

engedaam (approved):
LGTM 🚀

PR #6319, opened 2024-06-04
feat: use new pod identity based tags

rschalo:
Fixes #N/A 

**Description**
Update to use EKS controller tagging standards

- [ ] Yes, issue opened: # 
- [x] No


ellistarn:
Do we consider this breaking or an implementation detail?

rschalo:
> Do we consider this breaking or an implementation detail?

I think this is an implementation detail.

I tested this by creating a deployment with one replica with the old annotation, waiting for a node to come online, upgrading karpenter to an image with the new annotation, scaling out the deployment to two replicas after the upgrade, and then waiting for the second node to come up. 

Both nodes continued to operate just fine and later when a `SportInterruption` disruption happened to the node with the `karpenter.sh/managed-by` annotation, the node rotated out as expected.

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -356,7 +355,7 @@ func (in *EC2NodeClass) InstanceProfileRole() string {
 func (in *EC2NodeClass) InstanceProfileTags(clusterName string) map[string]string {
 	return lo.Assign(in.Spec.Tags, map[string]string{
 		fmt.Sprintf("kubernetes.io/cluster/%s", clusterName): "owned",
-		corev1beta1.ManagedByAnnotationKey:                   clusterName,
+		EksClusterNameAnnotationKey:                          clusterName,
```
njtran, line 358: nit: follow the Go semantic of capitalizing the acronym
```suggestion
		EKSClusterNameAnnotationKey:                          clusterName,
```

pkg/apis/v1beta1/labels.go
```diff
@@ -69,10 +69,13 @@ var (
 		// https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateCluster.html
 		regexp.MustCompile(`^kubernetes\.io/cluster/[0-9A-Za-z][A-Za-z0-9\-_]*$`),
 		regexp.MustCompile(fmt.Sprintf("^%s$", regexp.QuoteMeta(v1beta1.NodePoolLabelKey))),
-		regexp.MustCompile(fmt.Sprintf("^%s$", regexp.QuoteMeta(v1beta1.ManagedByAnnotationKey))),
+		regexp.MustCompile(fmt.Sprintf("^%s$", regexp.QuoteMeta(EksClusterNameAnnotationKey))),
 		regexp.MustCompile(fmt.Sprintf("^%s$", regexp.QuoteMeta(LabelNodeClass))),
 		regexp.MustCompile(fmt.Sprintf("^%s$", regexp.QuoteMeta(TagNodeClaim))),
 	}
+
+	EksClusterNameAnnotationKey = "eks:eks-cluster-name"
```
njtran, line 77: Same here
```suggestion
	EKSClusterNameAnnotationKey = "eks:eks-cluster-name"
```

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -271,7 +271,11 @@ func (p *DefaultProvider) createLaunchTemplate(ctx context.Context, options *ami
 		TagSpecifications: []*ec2.TagSpecification{
 			{
 				ResourceType: aws.String(ec2.ResourceTypeLaunchTemplate),
-				Tags:         utils.MergeTags(options.Tags, map[string]string{v1beta1.TagManagedLaunchTemplate: options.ClusterName, v1beta1.LabelNodeClass: options.NodeClassName}),
+				Tags: utils.MergeTags(options.Tags, map[string]string{
+					v1beta1.EksClusterNameAnnotationKey: options.ClusterName,
```
njtran, line 275: So this adds the eks cluster annotation tag on now, is this right?

pkg/apis/v1beta1/labels.go
njtran: This isn't viewed as a system tag, right? Are we sure that we'll always be able to do this?

test/hack/resource/pkg/resourcetypes/resourcetypes.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	karpenterClusterNameTag    = "karpenter.sh/managed-by"
+	karpenterClusterNameTag    = "eks:eks-cluster-name"
```
njtran, line 23: Can't rely on the v1beta1 label?

test/suites/integration/tags_test.go
```diff
@@ -86,7 +86,7 @@ var _ = Describe("Tags", func() {
 	})
 
 	Context("Tagging Controller", func() {
-		It("should tag with karpenter.sh/nodeclaim and Name tag", func() {
+		It("should tag with karpenter.sh/nodeclaim, eks:cluster-name, and Name tag", func() {
```
njtran, line 89:
```suggestion
		It("should tag with karpenter.sh/nodeclaim, eks:eks-cluster-name, and Name tag", func() {
```

test/suites/integration/validation_test.go
```diff
@@ -194,7 +194,7 @@ var _ = Describe("Validation", func() {
 			nodeClass.Spec.Tags = map[string]string{"karpenter.sh/nodepool": "custom-value"}
 			Expect(env.Client.Create(env.Context, nodeClass)).ToNot(Succeed())
 
-			nodeClass.Spec.Tags = map[string]string{"karpenter.sh/managed-by": env.ClusterName}
+			nodeClass.Spec.Tags = map[string]string{"eks:eks-cluster-name": env.ClusterName}
```
njtran, line 197: Seems odd that we would define the const but not use it here. Any reason why we're not?

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -80,7 +79,7 @@ type EC2NodeClassSpec struct {
 	// +kubebuilder:validation:XValidation:message="empty tag keys aren't supported",rule="self.all(k, k != '')"
 	// +kubebuilder:validation:XValidation:message="tag contains a restricted tag matching kubernetes.io/cluster/",rule="self.all(k, !k.startsWith('kubernetes.io/cluster') )"
 	// +kubebuilder:validation:XValidation:message="tag contains a restricted tag matching karpenter.sh/nodepool",rule="self.all(k, k != 'karpenter.sh/nodepool')"
-	// +kubebuilder:validation:XValidation:message="tag contains a restricted tag matching karpenter.sh/managed-by",rule="self.all(k, k !='karpenter.sh/managed-by')"
+	// +kubebuilder:validation:XValidation:message="tag contains a restricted tag matching eks:eks-cluster-name",rule="self.all(k, k !='eks:eks-cluster-name')"
```
ellistarn, line 82: Curious where this standard comes from. This stutters the word "eks".

test/suites/integration/validation_test.go
rschalo: No reason, updated.

pkg/providers/launchtemplate/launchtemplate.go
rschalo: Correct, according to @mikestef9, it should be added to all resources Karpenter creates.

PR #6331, opened 2024-06-06
test: get ami ssm path with offset

jmdeal:
Fixes #N/A 

**Description**
Adds a helper function to our environment to get the SSM path for a supported AMI family, with a release offset. Importantly, this allows us to avoid hardcoding old AMIs.

`/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/pkg/environment/aws/expectations.go
+++ b/test/pkg/environment/aws/expectations.go
@@ -17,6 +17,7 @@ package aws
 import (
 	"fmt"
 	"net"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -31,13 +32,16 @@ import (
 	"github.com/mitchellh/hashstructure/v2"
 	"github.com/samber/lo"
 	"go.uber.org/multierr"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"sigs.k8s.io/karpenter/pkg/scheduling"
 	coretest "sigs.k8s.io/karpenter/pkg/test"
 
 	"github.com/aws/karpenter-provider-aws/pkg/apis/v1beta1"
 	awserrors "github.com/aws/karpenter-provider-aws/pkg/errors"
+	"github.com/aws/karpenter-provider-aws/pkg/providers/amifamily"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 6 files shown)


jmdeal (commented):
/karpenter snapshot

test/pkg/environment/aws/expectations.go
```diff
@@ -362,6 +366,41 @@ func (env *Environment) GetAMIBySSMPath(ssmPath string) string {
...
+	Expect(env.SSMAPI.GetParametersByPathPages(&ssm.GetParametersByPathInput{
+		Path:      lo.ToPtr(basePath),
+	}, func(out *ssm.GetParametersByPathOutput, lastPage bool) bool {
+		parameters = append(parameters, lo.Map(out.Parameters, func(p *ssm.Parameter, _ int) string {
+			return lo.FromPtr(p.Name)
+		})...)
+		return lastPage
+	})).To(Succeed())
+
+	// Ensure there are enough published AMIs to satisfy the offset. Note that there are two SSM paths for the latest release.
```
jonathan-innis, line 398: Is that because of the image path having an SSM parameter for recommended or latest?

jonathan-innis: Having to encode that logic in here feels a bit awkward to me. Are there ways that we could workaround that that would be a little more clear to a reader than just adding two to the total number of the offset

test/pkg/environment/aws/expectations.go
```diff
@@ -362,6 +366,41 @@ func (env *Environment) GetAMIBySSMPath(ssmPath string) string {
 	return *parameter.Parameter.Value
 }
 
+// GetSSMPath gets the SSM path for an AMI release specified by the family, k8s version, architecture, and release offset.
+// A release offset of 0 will get you the latest release, 1 the previous release, and so on.
+// Note: this is tighly coupled with the naming conventions for AL2 and AL2023. This is **NOT** compatible with Bottlerocket, Ubuntu, or Windows AMIs.
+func (env *Environment) GetSSMPath(amiFamily string, k8sVersion string, arch string, releaseOffset int) string {
```
jonathan-innis, line 372:
```suggestion
func (env *Environment) GetSSMPathWithOffset(amiFamily string, k8sVersion string, arch string, releaseOffset int) string {
```

Does this need to indicate that it's asking for an offset?

test/pkg/environment/aws/expectations.go
```diff
@@ -362,6 +366,41 @@ func (env *Environment) GetAMIBySSMPath(ssmPath string) string {
 	return *parameter.Parameter.Value
 }
 
+// GetSSMPath gets the SSM path for an AMI release specified by the family, k8s version, architecture, and release offset.
+// A release offset of 0 will get you the latest release, 1 the previous release, and so on.
+// Note: this is tighly coupled with the naming conventions for AL2 and AL2023. This is **NOT** compatible with Bottlerocket, Ubuntu, or Windows AMIs.
```
jonathan-innis, line 371: Why does it have to be tightly coupled? For any given family, shoudn't we be able to get its relative path and then use that to be the list of all the amis published to it, ordering them by date. From what I remember, you can definitely check when the parameter was published and use that as a proxy for determining which one comes before the other one

test/pkg/environment/aws/expectations.go
jonathan-innis, line 372: Since we have cases where offset is just 0, does it make sense to do what we do with K8s version where we have a version that doesn't take in an offset to simplify and then we have an extension of it that takes in an offset when we need it?

jonathan-innis (commented):
Nice work! Really cool to see this and this should help us when we are going between K8s versions in the future!

test/pkg/environment/aws/expectations.go
jmdeal: The update timestamps for the SSM paths didn't line up with the release date, e.g. an SSM path may have been updated after it's release. I can dig in a little more to see if there's a better option here, but this seemed like a reasonable compromise to keep the code simple considering that 99% of the time in our tests we're using AL2023.

test/pkg/environment/aws/expectations.go
jmdeal: Yep, there's the path with the full release name and an alias called recommended. Seems reasonable to reject any path with recommended in the path name, should be a little more readable.

PR #6336, opened 2024-06-07
docs: RFC for disrupted EBS-Backed StatefulSet delays

AndrewSirenko:
Fixes #N/A 

**Description**
Create RFC for disrupted EBS-Backed StatefulSet delays

N/A

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/designs/statefulset-disruption.md
+++ b/designs/statefulset-disruption.md
@@ -27,6 +27,9 @@ This document reviews the desired flow for disrupting of stateful workloads, des
     - [Alternatives Considered](#alternatives-considered-1)
   - [Appendix](#appendix)
     - [Z. Document TODOs](#z-document-todos)
+    - [T. Latency Numbers](#t-latency-numbers)
+      - [Pod termination -\> volumes cleaned up](#pod-termination---volumes-cleaned-up)
+      - [Instance stopped/terminated times](#instance-stoppedterminated-times)
     - [A. Further Reading](#a-further-reading)
     - [B. Terminology](#b-terminology)
     - [C. Related Issues](#c-related-issues)
@@ -52,7 +55,7 @@ From [Karpenter: Disruption](https://karpenter.sh/docs/concepts/disruption/):
 
 For the scope of this document, we will focus on Karpenter's [Node Termination Controller](https://github.com/kubernetes-sigs/karpenter/blob/38b4c32043724f6dafce910669777877295d89ec/pkg/controllers/node/termination/controller.go#L77) and its interactions with the terminating node.
 
-See the following diagram for the relevant sequence of events in the case of only stateless pods:
+See the following diagram for the relevant sequence of events in the case of only stateless workloads:
 
 ![Stateless_Workload_Karpenter_Termination](https://github.com/AndrewSirenko/karpenter-provider-aws/assets/68304519/e629f41a-5b1e-49a2-b7a4-29a2db172394)
 
@@ -65,9 +68,10 @@ Persistent Storage in Kubernetes involves many moving parts, most of which may n
 For the purpose of this document, you should know that:
 - The **Container Storage Interface** (CSI) is a standard way for Container Orchestrators to provision persistent volumes from storage providers and expose block and file storage systems to containers.
 - The **AttachDetach Controller** watches for stateful workloads that are waiting on their storage, and ensures that their volumes are attached to the right node. Also watches for attached volumes that are no longer in use, and ensures they are detached. 
-- The **CSI Controller** attaches/detaches volumes to nodes with pods that require Persistent Volumes. (I.e. Calls EC2 AttachVolume). [^5]
-- The **CSI Node Service** mounts[^3] volumes to make them available for use by pods. Unmounts volumes after pods terminate to ensure they are no longer in use. Runs on each node. The Kubelet's Volume Manager watches for stateful pods and calls csi node service.
-- The CSI Specification states that the container orchestrator must interact with CSI  Plugin through the following flow of Remote Procedure Calls when a workload requires persistent storage: ControllerPublishVolume (i.e. attach volume to node) -> NodeStageVolume (Mount volume to global node mount-point) -> NodePublishVolume (Mount volume to pod's mountpoint) (and when volume no longer in use: NodeUnbpublishVolume -> NodeUnstageVolume -> ControllerUnpublishVolume) 
+- The **CSI Controller** attaches/detaches volumes to nodes with workloads that require Persistent Volumes. (I.e. Calls EC2 AttachVolume). [^5]
+- The **CSI Node Service** mounts[^3] volumes to make them available for use by workloads. Unmounts volumes after workload terminates to ensure they are no longer in use. Runs on each node. The Kubelet's Volume Manager watches for stateful workloads and calls csi node service.
+- `Mounted != Attached`. Attached EBS Volume is visible as a block device by privileged user on node at `/dev/<device-path>`. Mounted volume is visible by workload containers at specified `mountPath`s. See [this StackOverflow post](https://stackoverflow.com/questions/24429949/device-vs-partition-vs-file-system-vs-volume-how-do-these-concepts-relate-to-ea)
+- The [CSI Specification](https://github.com/container-storage-interface/spec/blob/master/spec.md) states that the container orchestrator must interact with CSI  Plugin through the following flow of Remote Procedure Calls when a workload requires persistent storage: ControllerPublishVolume (i.e. attach volume to node) -> NodeStageVolume (Mount volume to global node mount-point) -> NodePublishVolume (Mount volume to pod's mountpoint) (and when volume no longer in use: NodeUnbpublishVolume -> NodeUnstageVolume -> ControllerUnpublishVolume) 
 
 For the purpose of this document, assume volumes have already been created and will never be deleted.
 
@@ -97,9 +101,9 @@ Today, with customers with default Karpenter `v0.37.0` and EBS CSI Driver `v1.31
 
 **Problem A. If step 2 doesn't happen, there will be a 6+ minute delay.**
 
-If volumes are not *confirmed* as unmounted *by CSI Node Service*, Kubernetes cannot confirm volumes are not in use and will wait 6 minutes and confirm node is unhealthy before treating the volume as unmounted. See [EBS CSI 6-minute delay FAQ](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#6-minute-delays-in-attaching-volumes) for more context. [^6] 
+If volumes are not *confirmed* as unmounted *by CSI Node Service*, Kubernetes cannot confirm volumes are not in use and will wait a hard-coded [6 minute MaxWaitForUnmountDuration](https://github.com/kubernetes/kubernetes/blob/8b727956214818a3a5846bca060426a13a578348/pkg/controller/volume/attachdetach/attach_detach_controller.go#L94) and confirm node is unhealthy before treating the volume as unmounted. See [EBS CSI 6-minute delay FAQ](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#6-minute-delays-in-attaching-volumes) for more context. [^6] 
 
-Customers will see the following event (Note the 6+ minute delay): 
+Customers will see the following event on pod object (Note the 6+ minute delay): 
 
 ```
 Warning  FailedAttachVolume      6m51s              attachdetach-controller  Multi-Attach error for volume "pvc-123" Volume is already exclusively attached to one node and can't be attached to another
@@ -107,7 +111,7 @@ Warning  FailedAttachVolume      6m51s              attachdetach-controller  Mul
 
 **Problem B. If step 3 doesn't happen before step 4, there will be a 1+ minute delay**
 
-If Karpenter calls EC2 TerminateInstance **before** EC2 DetachVolume calls from EBS CSI Driver Controller pod finish, then the volumes won't be detached **until the old instance terminates**.This delay depends on how long it takes the underlying instance to enter the `terminated` state, which depends on the instance type. Typically 1 minute for `m5a.large`, up to 15 minutes for certain Metals/GPU/Windows instances. See [appendix D1](#d1-ec2-termination--ec2-detachvolume-relationship-) for more context. 
+If Karpenter calls EC2 TerminateInstance **before** EC2 DetachVolume calls from EBS CSI Driver Controller pod finish, then the volumes won't be detached **until the old instance terminates**.This delay depends on how long it takes the underlying instance to enter the `terminated` state, which depends on the instance type. Typically 1 minute for `m5a.large`,  up to 10 minutes for certain metals instances. See [appendix D1](#d1-ec2-termination--ec2-detachvolume-relationship-) and [instance termination latency measurements](#t-latency-numbers) for more context. 
  
 Customers will see the following events (Note the 1-min delay between `Multi-Attach error` AND `AttachVolume.Attach failed`):
 
@@ -125,9 +129,9 @@ Customers can determine which delay they are suffering from based off of whether
 
 **A2: To solve A today, Karpenter should confirm that volumes are not in use and confirm AttachDetach Controller knows this before deleting the node's finalizer.**
 
-**B1: To solve B today, Karpenter should wait for volumes to detach before terminating the node.**
+**B1: To solve B today, Karpenter should wait for volumes to detach by watching volumeattachment objects before terminating the node.**
 
-See [WIP Kubernetes 1.31 A1 solution in PR #125070](https://github.com/kubernetes/kubernetes/pull/125070)
+See [WIP Kubernetes 1.31/1.32 A1 solution in PR #125070](https://github.com/kubernetes/kubernetes/pull/125070)
 
 See [a proof-of-concept implementation of A2 & B1 in PR #1294](https://github.com/kubernetes-sigs/karpenter/pull/1294)
 
@@ -141,13 +145,13 @@ Finally, we should add the following EBS x Karpenter end-to-end test in karpente
 
 If `ReadWriteOnce` volumes are not unmounted *by CSI Node Service*, Kubernetes cannot confirm volumes are not in use and safe to attach to new node. Kubernetes will wait 6 minutes[^4] and ensure Node is `unhealthy` before treating the volume as unmounted and moving forward with a volume detach. See [EBS CSI 6-minute delay FAQ](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#6-minute-delays-in-attaching-volumes) 
 
-Cluster operator will see a `FailedAttachVolume` event with `Multi-Attach error`
+Cluster operator will see a `FailedAttachVolume` event on pod object with `Multi-Attach error`
 
 ### When does this happen? 
 
 This delay happens when the EBS CSI Node pod is killed before it can unmount all volumes of `terminated` pods. Note that a pod's volumes can only be unmounted **after** the pod enters the `terminated` state. 
 
-The EBS CSI Node pod can be killed in two places, depending on whether it tolerates the `karpenter.sh/disruption=disrupting` taint: If the EBS CSI Node does not tolerate the taint, it will be killed during Karpenter [Terminator's draining process](https://github.com/kubernetes-sigs/karpenter/blob/e58d48e24c051e217b1c0c119bd19b7b29519532/pkg/controllers/node/termination/controller.go#L90) after all pods that are not system-critical Daemonsets enter the `terminated` state.  If the EBS CSI Node Pod does tolerate this Karpenter taint, Karpenter's Terminator will call EC2 TerminateInstances when all intolerant pods are `terminated`. In this case, if [Graceful Shutdown](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#graceful-node-shutdown) is configured on the node, the Kubelet's Node Shutdown Manager will attempt to kill EBS CSI Node Pod after all non-critical pods have entered `terminated` state.  
+The EBS CSI Node pod can be killed in two places, depending on whether it tolerates the `karpenter.sh/disruption=disrupting` taint: If the EBS CSI Node does not tolerate the taint, it will be killed during Karpenter [Terminator's draining process](https://github.com/kubernetes-sigs/karpenter/blob/e58d48e24c051e217b1c0c119bd19b7b29519532/pkg/controllers/node/termination/controller.go#L90) after all pods that are not system-critical Daemonsets enter the `terminated` state.  If the EBS CSI Node Pod does tolerate this Karpenter taint, Karpenter's Terminator will call EC2 TerminateInstances when all intolerant pods are `terminated`. In this case, if [Graceful Shutdown](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#graceful-node-shutdown) is configured on the node, the Kubelet's [Node Shutdown Manager will attempt to kill EBS CSI Node Pod](https://github.com/kubernetes/kubernetes/blob/b3db54ea72a4f7441260982b4d2941f856401c9a/pkg/kubelet/nodeshutdown/nodeshutdown_manager_linux.go#L322-L417) after all non-critical pods have entered `terminated` state.  
 
 As of EBS CSI Driver `v1.31.0`, the EBS CSI Node Pod tolerates all taints be default, therefore we will focus on this second type of race for the following diagram:
 
@@ -183,8 +187,8 @@ Because this solution does not rely on changes in Karpenter, please see [Active 
 
 **Cons:**
 
-- Unavailable until merged in a version of Kubernetes (Likely Kubernetes `v1.31`) (Possibly able to be backported)
-- If Kubelet Shutdown Manager is not able to run 
+- Unavailable until merged in a version of Kubernetes (Likely Kubernetes `v1.31` or `v1.32`) (Possibly able to be backported)
+- If gracefulShutdown period is up BEFORE volumes are unmounted by CSI Node pod, then kubernetes cannot confirm volume was unmounted, and volume will still have delay. (E.g. unmount takes more than 1 minute, which is longer than gracefulShutdown period of 45 seconds.)
 
 #### A2: Taint node as `out-of-service` after termination
 
@@ -196,7 +200,7 @@ In `v1.26`, Kubernetes enabled the [Non-graceful node shutdown handling feature]
 
 Once Karpenter confirms that an instance is terminated, adding this taint to the node object will allow the Attach/Detach Controller to treat the volume as not in use, preventing the 6+ minute delay. 
 
-By modifying Karpenter to apply this taint and wait ~5 seconds until volumes are marked not in use, the following sequence will occur:
+By modifying Karpenter to apply this taint and wait until volumes are marked not in use on node object (~5 seconds), the following sequence will occur:
 
 ![taint_solution](https://github.com/AndrewSirenko/karpenter-provider-aws/assets/68304519/d1bc5704-7592-4aa1-a72e-83d33f824b8b)
 
@@ -211,8 +215,8 @@ See [this commit](https://github.com/kubernetes-sigs/karpenter/pull/1294/commits
 
 **Cons:**
 
-- Unavailable until Kubernetes `v1.26`. Customers running Karpenter on Kubernetes ≤ `v1.25` will require solution B1 to be implemented. 
-- Requires Karpenter-Provider-AWS to not treat `Shutting Down` as terminated (Though Karpenter had already planned on this via [PR #5979](https://github.com/aws/karpenter-provider-aws/pull/5979))
+- Only available in Kubernetes ≥ `v1.26`.
+- Requires Terminator to ensure instance is `terminated` before applying taint.
 - Problem B's delay still occurs because volumes will not be detached until consolidated instance terminates. 
 
 #### Alternatives Considered
@@ -224,10 +228,8 @@ Customers can mitigate 6+ minute delays by configuring their nodes and pod taint
 A quick overview:
 - Configure Kubelet for Graceful Node Shutdown
 - Enable Karpenter Spot Instance interruption handling
-- Use EBS CSI Driver ≥ `v1.x` in order use the [PreStop Lifecycle Hook](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#what-is-the-prestop-lifecycle-hook)
-- Set `.node.tolerateAllTaints=false` when deploying the EBS CSI Driver
-- Confirm that your stateful workloads do not tolerate Karpenter's `disrupting` taint, and any that any system-critical stateful daemonsets have a lower terminationGracePeriod than the EBS CSI Driver.
-- EBS CSI driver should switch to gRPC liveness probe to ensure kubelet trusts CSI Node Pod during graceful node shutdown. ETA EBS CSI Driver v1.33.0. 
+- Use EBS CSI Driver ≥ `v1.28` in order use the [PreStop Lifecycle Hook](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#what-is-the-prestop-lifecycle-hook)
+- Use Karpenter ≥ `v1.0.0`
 
 **Pros:**
 
@@ -245,9 +247,9 @@ Even if we solve the 6+ minute volume-in-use delay, AWS customers may suffer fro
 
 ### When does this happen?
 
-If Karpenter calls EC2 TerminateInstance **before** EC2 DetachVolume calls finish, then the volumes won't be detached **until the old instance terminates**. This delay depends on the instance type. 1 minute for `m5a.large`, 10-15+ minutes for certain GPU/Metal/Windows instances like `m7i.metal-48xl` (making this especially painful for ML/AI or .Net customers). For more context see [Appendix D1](#d1-ec2-termination--ec2-detachvolume-relationship-additional-context)
+If Karpenter calls EC2 TerminateInstance **before** EC2 DetachVolume calls finish, then the volumes won't be detached **until the old instance terminates**. This delay depends on the instance type. 1 minute for `m5a.large`, 2 minutes for large GPU instances like `g4ad.16xlarge`, and 10+ minutes for certain Metal instances like `m7i.metal-48xl`. For more context see [Appendix D1](#d1-ec2-termination--ec2-detachvolume-relationship-additional-context) and [instance termination latencies](#t-latency-numbers)
 
-Operators will see `FailedAttachVolume` events with `Multi-Attach error` and then `AttachVolume.Attach failed` errors. 
+Operators will see `FailedAttachVolume` events on pod object with `Multi-Attach error` and then `AttachVolume.Attach failed` errors. 
 
 ### Solution B1: Wait for detach in Karpenter cloudProvider.Delete
 
@@ -255,7 +257,7 @@ Wait for volumes to detach before terminating the instance.
 
 We can do this by waiting for all volumes of drain-able nodes to be marked as not be in use nor attached before terminating the node in c.cloudProvider.Delete (until a maximum of 20 seconds). See [Appendix D3 for the implementation details of this wait](#d) 
 
-We can detect a volume is detached by ensuring that `volumeattachment` objects associated to relevant PVs are deleted
+We can detect a volume is detached by ensuring that `volumeattachment` objects associated to relevant PVs are deleted. This also implies that the volume was safely unmounted by the CSI Node pod. 
 
 This means that our sequence of events will match the ideal diagram from section [Ideal Graceful Shutdown for Stateful Workloads][#ideal-graceful-shutdown-for-stateful-workloads]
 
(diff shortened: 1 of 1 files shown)


designs/statefulset-disruption.md
```diff
@@ -0,0 +1,739 @@
...
+
+See: [PR #125070](https://github.com/kubernetes/kubernetes/pull/125070) for more information. 
+
+**Pros:** 
+- Other cluster autoscalers will not face this race condition.  
+- Ideal long-term solution
+- Reduces pod migration times
+- Reduces risk of data corruption
+**Cons:**
+- Unavailable until Kubernetes `v1.31`
```
AndrewSirenko, line 168: There is a chance this fix could be back-ported to earlier versions.

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,739 @@
...
+- Configure Kubelet for Graceful Node Shutdown
+- Enable Karpenter Spot Instance interruption handling
+- Use EBS CSI Driver ≥ `v1.x` in order use the [PreStop Lifecycle Hook](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#what-is-the-prestop-lifecycle-hook)
+- Set `.node.tolerateAllTaints=false` when deploying the EBS CSI Driver
+[](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#what-steps-can-be-taken-to-mitigate-this-issue)
+- Confirm that your stateful workloads do not tolerate Karpenter's `disrupting` taint, and any that any system-critical stateful daemonsets have a lower terminationGracePeriod than the EBS CSI Driver.
+
+**Pros:**
+- No code change required in Karpenter. 
+**Cons:**
```
FernandoMiguel, line 212: needs an extra line or a double space after the previous line, in other to render correctly in MD

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,739 @@
...
+
+Operators will see `FailedAttachVolume` events with `Multi-Attach error` and then `AttachVolume.Attach failed` errors. 
+
+### Solution B1: Wait for unmounts in Karpenter cloudProvider.Delete
+
+Wait for volumes to unmount before terminating the instance. 
+
+We can do this by waiting for all volumes of drain-able nodes to be marked as not be in use before terminating the node in c.cloudProvider.Delete (until a maximum of 20 seconds). 
+
+This means that our sequence of events will match the ideal diagram from section [### Ideal Graceful Shutdown for Stateful Workloads]
```
FernandoMiguel, line 230: broken anchor

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,739 @@
...
+
+We can use similar logic to [today's proof-of-concept implementation](https://github.com/kubernetes-sigs/karpenter/pull/1294), but move it to karpenter-provider-aws and check for `node.Status.VolumesInUse` instead of listing volumeattachment objects. A 20 second max wait was sufficient to prevent delays with m5a instance type, but further testing is needed to ensure it is enough for Windows/GPU instance types.
+
+**Pros:**
+- Leaves decision to each cloud/storage provider 
+- Can opt-in to this behavior for specific CSI Drivers (Perhaps via Helm parameter)
+- Only delays termination of nodes with stateful workloads.
+**Cons:**
```
FernandoMiguel, line 238: needs and extra line or double space at the end of the previous line, to properly break line in MD

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,739 @@
...
+
+### Alternatives Considered
+
+**Implement B1 in `kubernetes-sigs/karpenter`**
+
+Instead of solving this inside the [`c.cloudProvider.Delete`](https://github.com/aws/karpenter-provider-aws/blob/9cef47b6df77ec9e0a39dc6f4a4ecd1aab504ae3/pkg/cloudprovider/cloudprovider.go#L179), solve this inside [termination controller's reconciler loop](https://github.com/kubernetes-sigs/karpenter/blob/94d5b41d4711b1c21fc0264f1b29db6a64b95caf/pkg/controllers/node/termination/controller.go#L77) (as is done in [today's proof-of-concept implementation](https://github.com/kubernetes-sigs/karpenter/pull/1294))
+
+**Pros:**
+- Karpenter-provider-aws does not need to know about Kubernetes volume lifecycle
+**Cons:**
```
FernandoMiguel, line 250: needs and extra line or double space at the end of the previous line, to properly break line in MD

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,739 @@
...
+- [Open Question] EBS may be the only storage provider where detaching volumes before terminating the instance matters. If this is the case, the delay before instance termination is not worth it for customers of other storage/cloud providers.
+- Should not hardcode cloud-provider specific CSI Drivers in upstream project. 
+
+**Karpenter Polls EC2 for volume detachments before calling EC2 TerminateInstance**
+
+Karpenter-provider-aws can poll EC2 DescribeVolumes before making an EC2 TerminateInstance call. 
+
+**Pros:**
+- Solves issue
+**Cons:**
```
FernandoMiguel, line 260: needs and extra line or double space at the end of the previous line, to properly break line in MD

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,739 @@
...
+- Kubernetes:
+  - [Node Shutdowns](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/)
+  - [API-initiated Eviction](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/#how-api-initiated-eviction-works)
+
+
+### B. Terminology
+Daemonset: Pod scheduled on every Node. 
+EBS: Elastic Block Store. An Amazon Web Service that 
+Kubelet: 
+StatefulSet: Manages the deployment and scaling of a set of Pods, and provides guarantees about the ordering and uniqueness of these Pods. These uniqueness gauruntees are valuable when your workload needs persistent storage.
```
FernandoMiguel, line 286: needs and extra line or double space at the end of the previous line, to properly break line in MD

FernandoMiguel: or just bullets

FernandoMiguel, line 286: typo: gauruntees

AndrewSirenko:
Thank you @FernandoMiguel for the edits!

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,762 @@
...
+
+
+### B. Terminology
+
+- Daemonset: Pod scheduled on every Node. 
+- EBS: Elastic Block Store. 
+- Kubelet: Primary 'node agent' that runs on each node. Volume Manager service in Kubelet makes gRPC to EBS CSI Node pod.
+- StatefulSet: Manages the deployment and scaling of a set of Pods, and provides gauruntees about the ordering and uniqueness of these Pods. These uniqueness gauruntees are valuable when your workload needs persistent storage. 
+
+### C. Issue Timeline
```
AndrewSirenko, line 311: TODO Summarize and add 

* [VolumeAttachment takes too long to remove](https://github.com/kubernetes-csi/external-attacher/issues/463)
* [Volume still hang on Karpenter Node Consolidation/Termination](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1955)
* [Volume hang on Karpenter Node Consolidation/Termination](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1665)
* [PVC attaching takes much time](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1302)
* [687849515510 Thomson Reuters - it's taking long time to detach and reattach EBS volume](https://t.corp.amazon.com/V974289384/overview)

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,757 @@
...
+
+- Solves 6+ minute delays by default
+- No additional latency before starting instance termination. 
+- Minor latency in deleting Node's finalizer IF karpenter does not treat `shutting down` instance as terminated. (In my tests, a 5 second wait was sufficient for AttachDetach Controller to recognize the out-of-service taint and allow for volume detach)
+- If Kubernetes makes this 6 minute ForceDetach timer infinite by default (As currently planned in Kubernetes v1.32), the out-of-service taint will be the only ensure workload starts on other node 
+
+**Cons:**
+
+- Unavailable until Kubernetes `v1.26`. Customers running Karpenter on Kubernetes ≤ `v1.25` will require solution B1 to be implemented. 
+- Requires Karpenter-Provider-AWS to not treat `Shutting Down` as terminated (Though Karpenter had already planned on this via [PR #5979](https://github.com/aws/karpenter-provider-aws/pull/5979))
```
AndrewSirenko, line 198: I see #5979 was merged but only changed tests. Unsure what status is on this bullet point.

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+### Stateful Workloads Overview
+
+Persistent Storage in Kubernetes involves many moving parts, most of which may not be relevant for the decision at hand. 
+
+For the purpose of this document, you should know that:
+- The **Container Storage Interface** (CSI) is a standard way for Container Orchestrators to provision persistent volumes from storage providers and expose block and file storage systems to containers.
+- The **AttachDetach Controller** watches for stateful workloads that are waiting on their storage, and ensures that their volumes are attached to the right node. Also watches for attached volumes that are no longer in use, and ensures they are detached. 
+- The **CSI Controller** attaches/detaches volumes to nodes with pods that require Persistent Volumes. (I.e. Calls EC2 AttachVolume). [^5]
+- The **CSI Node Service** mounts[^3] volumes to make them available for use by pods. Unmounts volumes after pods terminate to ensure they are no longer in use. Runs on each node. The Kubelet's Volume Manager watches for stateful pods and calls csi node service.
```
jonathan-innis, line 69: Does it just monitor stateful pods or is just anything that carries a PVC/PV that's attached to it?

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+- The CSI Specification states that the container orchestrator must interact with CSI  Plugin through the following flow of Remote Procedure Calls when a workload requires persistent storage: ControllerPublishVolume (i.e. attach volume to node) -> NodeStageVolume (Mount volume to global node mount-point) -> NodePublishVolume (Mount volume to pod's mountpoint) (and when volume no longer in use: NodeUnbpublishVolume -> NodeUnstageVolume -> ControllerUnpublishVolume)
```
jonathan-innis, line 70: Link to the specification?

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+See the following diagram for a more detailed sequence of events.  
+
+![ideal](https://github.com/AndrewSirenko/karpenter-provider-aws/assets/68304519/ae5267e1-fe2d-455f-9199-60cee772f0ae)
+
+### Problems
+
+Today, with customers with default Karpenter `v0.37.0` and EBS CSI Driver `v1.31.0` configurations may experience two different kinds of delays once their disrupted stateful workloads are scheduled on a new node. 
+
+**Problem A. If step 2 doesn't happen, there will be a 6+ minute delay.**
```
njtran, line 98: I'm sure there's a bunch of details here, what was the reason why we have to do 6m here? Does 6+ mean at least 6 minutes of waiting? what's the worst case, infinite?

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+**A2: To solve A today, Karpenter should confirm that volumes are not in use and confirm AttachDetach Controller knows this before deleting the node's finalizer.**
+
+**B1: To solve B today, Karpenter should wait for volumes to detach before terminating the node.**
+
+See [WIP Kubernetes 1.31 A1 solution in PR #125070](https://github.com/kubernetes/kubernetes/pull/125070)
+
+See [a proof-of-concept implementation of A2 & B1 in PR #1294](https://github.com/kubernetes-sigs/karpenter/pull/1294)
+
+Finally, we should add the following EBS x Karpenter end-to-end test in karpenter-provider-aws to catch regressions between releases of Karpenter or EBS CSI Driver:
```
jonathan-innis, line 134: Nice! Really like that we are scoping in testing to this RFC! This is definitely something that we should be actively monitoring!

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+### Solutions
+
+**A1: To solve A long-term, Kubernetes should ensure volumes are unmounted before critical pods like CSI Driver Node pod are terminated.**
+
+**A2: To solve A today, Karpenter should confirm that volumes are not in use and confirm AttachDetach Controller knows this before deleting the node's finalizer.**
+
+**B1: To solve B today, Karpenter should wait for volumes to detach before terminating the node.**
+
+See [WIP Kubernetes 1.31 A1 solution in PR #125070](https://github.com/kubernetes/kubernetes/pull/125070)
```
njtran, line 130: If this is a WIP are we sure this will be in 1.31?

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+### When does this happen? 
+
+This delay happens when the EBS CSI Node pod is killed before it can unmount all volumes of `terminated` pods. Note that a pod's volumes can only be unmounted **after** the pod enters the `terminated` state. 
+
+The EBS CSI Node pod can be killed in two places, depending on whether it tolerates the `karpenter.sh/disruption=disrupting` taint: If the EBS CSI Node does not tolerate the taint, it will be killed during Karpenter [Terminator's draining process](https://github.com/kubernetes-sigs/karpenter/blob/e58d48e24c051e217b1c0c119bd19b7b29519532/pkg/controllers/node/termination/controller.go#L90) after all pods that are not system-critical Daemonsets enter the `terminated` state.  If the EBS CSI Node Pod does tolerate this Karpenter taint, Karpenter's Terminator will call EC2 TerminateInstances when all intolerant pods are `terminated`. In this case, if [Graceful Shutdown](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#graceful-node-shutdown) is configured on the node, the Kubelet's Node Shutdown Manager will attempt to kill EBS CSI Node Pod after all non-critical pods have entered `terminated` state.
```
jonathan-innis, line 150: Is there something that we could link out to that shows us where the Node Shutdown Manager handles the logic to terminate the EBS CSI Node pod after the non-critical pods?

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+Customers can determine which delay they are suffering from based off of whether `AttachVolume.Attach` is in the `FailedAttachVolume` event. 
+
+### Solutions
+
+**A1: To solve A long-term, Kubernetes should ensure volumes are unmounted before critical pods like CSI Driver Node pod are terminated.**
+
+**A2: To solve A today, Karpenter should confirm that volumes are not in use and confirm AttachDetach Controller knows this before deleting the node's finalizer.**
+
+**B1: To solve B today, Karpenter should wait for volumes to detach before terminating the node.**
```
njtran, line 128: Both of these are just status conditions that we have to wait for the node to present, right?

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+Customers will see the following event (Note the 6+ minute delay): 
+
+```
+Warning  FailedAttachVolume      6m51s              attachdetach-controller  Multi-Attach error for volume "pvc-123" Volume is already exclusively attached to one node and can't be attached to another
+```
+
+**Problem B. If step 3 doesn't happen before step 4, there will be a 1+ minute delay**
+
+If Karpenter calls EC2 TerminateInstance **before** EC2 DetachVolume calls from EBS CSI Driver Controller pod finish, then the volumes won't be detached **until the old instance terminates**.This delay depends on how long it takes the underlying instance to enter the `terminated` state, which depends on the instance type. Typically 1 minute for `m5a.large`, up to 15 minutes for certain Metals/GPU/Windows instances. See [appendix D1](#d1-ec2-termination--ec2-detachvolume-relationship-) for more context.
```
AndrewSirenko, line 110: Mention "From my brief experiments, Typically 1 minute for 5ma.large..."

designs/statefulset-disruption.md
AndrewSirenko: So far it is projected to be in Kubernetes 1.31. Awaiting review from appropriate SIG reviewers.

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+1. Deploy statefulset with 1 replica
+2. Consolidate Node
+3. Confirm replica migrated
+4. Confirm replica running within x minutes (where x is high enough to prevent flakes)
+
+## Problem A. Preventing 6+ minute delays 
+
+If `ReadWriteOnce` volumes are not unmounted *by CSI Node Service*, Kubernetes cannot confirm volumes are not in use and safe to attach to new node. Kubernetes will wait 6 minutes[^4] and ensure Node is `unhealthy` before treating the volume as unmounted and moving forward with a volume detach. See [EBS CSI 6-minute delay FAQ](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#6-minute-delays-in-attaching-volumes) 
+
+Cluster operator will see a `FailedAttachVolume` event with `Multi-Attach error`
```
njtran, line 144: Is this a kube-event?

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+![kubelet_shutdown_race](https://github.com/AndrewSirenko/karpenter-provider-aws/assets/68304519/9ccb6c94-ef02-40af-84bf-9471827f3303)
+
+Today, the EBS CSI Driver attempts to work around these races by utilizing a [PreStop hook](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#what-steps-can-be-taken-to-mitigate-this-issue) that tries to keep the Node Service alive for an additional [terminationGracePeriodSeconds](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/73b43995a70b8fea51655577e4b475613ab400f1/charts/aws-ebs-csi-driver/values.yaml#L373) until all volumes are unmounted. We will explore the shortcomings of this solution later in [problem A alternative solutions](#alternatives-considered).  
+
+*Note: In addition to the Kubelet race, this delay can happen if stateful pods out-live the CSI Node Pod. E.g. Operator has a statefulset that tolerates all taints and has a longer terminationGracePeriod than EBS CSI Driver.*
```
jonathan-innis, line 164: What's the delay between the pod termination and the detach timeframe? Does it happen as soon as the pod is terminated or is there a bit of time in there that's required for the pod to go terminated, unmount, and then be good for instance removal

designs/statefulset-disruption.md
AndrewSirenko: One is status condition on node. One is actual `volumeattachment` object

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+#### A2: Taint node as `out-of-service` after termination
+
+While this race should be fixed at the Kubelet level long-term, we still need a solution for earlier versions of Kubernetes.
+
+One solution is to mark terminated nodes via `out-of-service` taint. 
+
+In `v1.26`, Kubernetes enabled the [Non-graceful node shutdown handling feature](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#non-graceful-node-shutdown) by default. This introduces the `node.kubernetes.io/out-of-service` taint, which can be used to mark a node as permanently shut down. See more context in [appendix D2]() 
+
+Once Karpenter confirms that an instance is terminated, adding this taint to the node object will allow the Attach/Detach Controller to treat the volume as not in use, preventing the 6+ minute delay.
```
jonathan-innis, line 197: Oh interesting....so it's an additional signal to the attach/detach controller. That's actually super cool and I get the value in why we'd want to do this

designs/statefulset-disruption.md
AndrewSirenko: 6m was an arbitrary time decided by Kubernetes. Atleast 6 minutes since node is unhealthy until AttachDetach Controller reconciler loop can force a `ControllerUnpublishVolume` detach call. Sometimes it is measured as 7 minutes end-to-end.

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+**Customer configuration**
+
+Customers can mitigate 6+ minute delays by configuring their nodes and pod taint tolerations. See [EBS CSI Driver FAQ: Mitigating 6+ minute delays](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#what-steps-can-be-taken-to-mitigate-this-issue) for an updating list of configuration requirements.
+
+A quick overview:
+- Configure Kubelet for Graceful Node Shutdown
+- Enable Karpenter Spot Instance interruption handling
+- Use EBS CSI Driver ≥ `v1.x` in order use the [PreStop Lifecycle Hook](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#what-is-the-prestop-lifecycle-hook)
+- Set `.node.tolerateAllTaints=false` when deploying the EBS CSI Driver
```
jmdeal, line 228: If graceful node shutdown is configured, wouldn't it be fine for the driver to tolerate Karpenter's disruption taint?

designs/statefulset-disruption.md
AndrewSirenko: This is an event on PVC and Pod object if that answers your question.

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+- Confirm that your stateful workloads do not tolerate Karpenter's `disrupting` taint, and any that any system-critical stateful daemonsets have a lower terminationGracePeriod than the EBS CSI Driver.
```
jmdeal, line 229: IIUC this is only the case if those daemonsets have associated volume attachments right? I'm wondering if it's common for critical daemonsets to have volumes that are detached from one node and reattached to another.

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+- Solves 6+ minute delays by default
+- No additional latency before Karpenter's terminator can start instance termination. 
+- Minor latency in deleting Node's finalizer IF karpenter does not treat `shutting down` instance as terminated. (In my tests, a 5 second wait was sufficient for AttachDetach Controller to recognize the out-of-service taint and allow for volume detach)
+- If Kubernetes makes this 6 minute ForceDetach timer infinite by default (As currently planned in Kubernetes v1.32), and the EBS CSI Node Pod is unable to unmount all volumes before improved Node Shutdown Manager times out, the out-of-service taint will be the only way to ensure workload starts on other node 
+
+**Cons:**
+
+- Unavailable until Kubernetes `v1.26`. Customers running Karpenter on Kubernetes ≤ `v1.25` will require solution B1 to be implemented. 
+- Requires Karpenter-Provider-AWS to not treat `Shutting Down` as terminated (Though Karpenter had already planned on this via [PR #5979](https://github.com/aws/karpenter-provider-aws/pull/5979))
```
jonathan-innis, line 215: This is already merged so I don't think we need to consider this as a consideration anymore

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+**Pros:** 
+
+- Other cluster autoscalers will not face this race condition. 
+- Reduces pod migration times
+- Reduces risk of data corruption because relevant CSI Drivers will perform unmount operations required  
+
+**Cons:**
+
+- Unavailable until merged in a version of Kubernetes (Likely Kubernetes `v1.31`) (Possibly able to be backported)
+- If Kubelet Shutdown Manager is not able to run
```
njtran, line 187: Do we know how often this happens?

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+While this race should be fixed at the Kubelet level long-term, we still need a solution for earlier versions of Kubernetes.
+
+One solution is to mark terminated nodes via `out-of-service` taint. 
+
+In `v1.26`, Kubernetes enabled the [Non-graceful node shutdown handling feature](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#non-graceful-node-shutdown) by default. This introduces the `node.kubernetes.io/out-of-service` taint, which can be used to mark a node as permanently shut down. See more context in [appendix D2]() 
+
+Once Karpenter confirms that an instance is terminated, adding this taint to the node object will allow the Attach/Detach Controller to treat the volume as not in use, preventing the 6+ minute delay. 
+
+By modifying Karpenter to apply this taint and wait ~5 seconds until volumes are marked not in use, the following sequence will occur:
```
njtran, line 199: Is there a signal we can wait for instead of a 5 second sleep?

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+![taint_solution](https://github.com/AndrewSirenko/karpenter-provider-aws/assets/68304519/d1bc5704-7592-4aa1-a72e-83d33f824b8b)
+
+See [this commit](https://github.com/kubernetes-sigs/karpenter/pull/1294/commits/88134e00ad02863d5a7268bba0b639e21f3f5398) for a proof-of-concept implementation. 
+
+**Pros:**
+
+- Solves 6+ minute delays by default
```
jonathan-innis, line 207: Are there any risks to not detaching the volume properly even with the out-of-service taint? For instance, if the volume wasn't properly detached but we know that the node is gone, does this create problems or is this the recommended lifecycle flow by the storage CSI controllers upstream?

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+Customers can mitigate 6+ minute delays by configuring their nodes and pod taint tolerations. See [EBS CSI Driver FAQ: Mitigating 6+ minute delays](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#what-steps-can-be-taken-to-mitigate-this-issue) for an updating list of configuration requirements.
+
+A quick overview:
+- Configure Kubelet for Graceful Node Shutdown
+- Enable Karpenter Spot Instance interruption handling
+- Use EBS CSI Driver ≥ `v1.x` in order use the [PreStop Lifecycle Hook](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#what-is-the-prestop-lifecycle-hook)
+- Set `.node.tolerateAllTaints=false` when deploying the EBS CSI Driver
+- Confirm that your stateful workloads do not tolerate Karpenter's `disrupting` taint, and any that any system-critical stateful daemonsets have a lower terminationGracePeriod than the EBS CSI Driver.
+- EBS CSI driver should switch to gRPC liveness probe to ensure kubelet trusts CSI Node Pod during graceful node shutdown. ETA EBS CSI Driver v1.33.0.
```
jonathan-innis, line 230: What's this option. Seems interesting. Is there a link to it? What would this give us?

designs/statefulset-disruption.md
AndrewSirenko: Well wouldn't daemonsets lifespan be tied to their node, unless customer really wants to re-use a limited set of volumes?

designs/statefulset-disruption.md
AndrewSirenko: Yes, node `volumesInUse` no longer having the relevant PV name.

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+Operators will see `FailedAttachVolume` events with `Multi-Attach error` and then `AttachVolume.Attach failed` errors. 
+
+### Solution B1: Wait for detach in Karpenter cloudProvider.Delete
+
+Wait for volumes to detach before terminating the instance.
+
+We can do this by waiting for all volumes of drain-able nodes to be marked as not be in use nor attached before terminating the node in c.cloudProvider.Delete (until a maximum of 20 seconds). See [Appendix D3 for the implementation details of this wait](#d) 
+
+We can detect a volume is detached by ensuring that `volumeattachment` objects associated to relevant PVs are deleted
```
jonathan-innis, line 258: Does this just give us the DetachVolume stuff for free too (even without the out-of-service taint)?

designs/statefulset-disruption.md
AndrewSirenko: Typically depends on the fix. We will know once relevant SIG reviewer leaves review on PR.

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+This means that our sequence of events will match the ideal diagram from section [Ideal Graceful Shutdown for Stateful Workloads][#ideal-graceful-shutdown-for-stateful-workloads]
+
+We can use similar logic to [today's proof-of-concept implementation](https://github.com/kubernetes-sigs/karpenter/pull/1294), but move it to karpenter-provider-aws and check for `node.Status.VolumesInUse` instead of listing volumeattachment objects. A 20 second max wait was sufficient to prevent delays with m5a instance type, but further testing is needed to ensure it is enough for Windows/GPU instance types.
```
jonathan-innis, line 262: Is there any kind of signal that the EBS CSI driver could give us to know that everything is detached before terminating? One option is that Karpenter does some work from its neutral code; another option is that the EBS CSI driver injects a finalizer and only removes it when it knows that it has finished the work that it needs to do. This, of course, requires us to create a finalizer that we wait for instance termination. There are similar-ish requirements from ALB though with waiting on instance termination before the target group is deregistered and I'm wondering how much overlap that we have here on these two problems

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+**Pros:**
+
+- Leaves decision to each cloud/storage provider 
+- Can opt-in to this behavior for specific CSI Drivers (Perhaps via Helm parameter)
+- Only delays termination of nodes with stateful workloads.
+- Implicitly solves problem A for EBS-backed stateful workloads, if volumeattachment object is deleted before instance is terminated.
```
jonathan-innis, line 269: Ah, here it is, nice!

designs/statefulset-disruption.md
njtran: i'm not sure I understand. Sounds like you're saying that the amount of times that it won't be able to run will be different after the upstream PR is merged? what is the current state like? Mostly just trying to understand what causes this to happen that Karpenter can be aware of.

designs/statefulset-disruption.md
```diff
@@ -0,0 +1,891 @@
...
+
+If Karpenter calls EC2 TerminateInstance **before** EC2 DetachVolume calls finish, then the volumes won't be detached **until the old instance terminates**. This delay depends on the instance type. 1 minute for `m5a.large`, 10-15+ minutes for certain GPU/Metal/Windows instances like `m7i.metal-48xl` (making this especially painful for ML/AI or .Net customers). For more context see [Appendix D1](#d1-ec2-termination--ec2-detachvolume-relationship-additional-context)
+
+Operators will see `FailedAttachVolume` events with `Multi-Attach error` and then `AttachVolume.Attach failed` errors. 
+
+### Solution B1: Wait for detach in Karpenter cloudProvider.Delete
+
+Wait for volumes to detach before terminating the instance.
+
+We can do this by waiting for all volumes of drain-able nodes to be marked as not be in use nor attached before terminating the node in c.cloudProvider.Delete (until a maximum of 20 seconds). See [Appendix D3 for the implementation details of this wait](#d)
```
jonathan-innis, line 256: This is kind of weird IMO. This basically means that the Delete call is going to return an error until we are good to actually delete the instance. I wonder if there's other ways that we could hook into the deletion flow without having to hack the cloudprovider delete call so that it orchestrates the flow that we're looking for

designs/statefulset-disruption.md
AndrewSirenko: For 6+ min delays: In theory yes, in practice I have needed to turn tolerateAllTaints off in order to not see 6+min delays. We can trade notes on this.

designs/statefulset-disruption.md
jonathan-innis: The problem today is that we only call delete once we get a success, which means that you are either going to have to make this synchronous or you are going to have to do some kind of weird erroring so that we back-off until we actually do the termination

designs/statefulset-disruption.md
jmdeal: My point being that you're probably not going to see reattach issues with daemonsets unless that's the case, so this is probably only required for a niche set of users. I still think solving the 1+ minute delay justifies the additional work, but wanted to drill down and see what additional customer configuration is really necessary here.

designs/statefulset-disruption.md
AndrewSirenko: Was looking at wrong line. Shutdown manager will **always** run IF graceful shutdown is enabled properly on node. Will run until gracefulShutdown time is up. If volumes are NOT unmounted by CSI Node pod by then (gracefulShutdown period being up) for some reason, then kubernetes cannot confirm volume was not unmounted, and volume will still have delay.

cartermckinnon:
xref: https://github.com/kubernetes/enhancements/pull/4213

designs/statefulset-disruption.md
AndrewSirenko: From my limited testing:

Pod terminated -> CSI node service unmounts is typically milliseconds (lets call it `<1 second`)

unmount -> EBS Controller plugin sees  `EC2 DetatchVolume` call as success is typically also `<1 second`

unmount -> Volume actually Detached from linux instance is 5-13 seconds. (Should be 4-8 seconds in EBS CSI Driver v1.32 due to polling optimization) 

pod termination -> We're clear for instance termination from my tests is ~10 seconds. (If EC2 DetatchVolume is made enough ahead of EC2TerminateInstances, we are fine. It's only if they're within a few seconds of eachother that we run into the problem B race.)

Will add these timeframes to doc.

designs/statefulset-disruption.md
AndrewSirenko: Will link in doc, it's a brutal read: [spec/spec.md at master · container-storage-interface/spec](https://github.com/container-storage-interface/spec/blob/master/spec.md)

designs/statefulset-disruption.md
AndrewSirenko: Great clarification, EBS CSI Controller monitors pods that rely on EBS-backed PV. Will clarify wording throughout doc.

designs/statefulset-disruption.md
AndrewSirenko: Yes, will link in doc, but here is code https://github.com/kubernetes/kubernetes/blob/b3db54ea72a4f7441260982b4d2941f856401c9a/pkg/kubelet/nodeshutdown/nodeshutdown_manager_linux.go#L322-L417

Note that pod kills happen by priority.

designs/statefulset-disruption.md
AndrewSirenko: TLDR:

For EBS, no, there are *no risks* to detaching volume without having confirmation that volume was unmounted by EBS CSI Node service. 

For other cloud providers or CSI Drivers. Not confirmed yet (though arguably a correctly written CSI Driver should be safe).

See footnotes 5 & 6 in Appendix section G and section Appendix D2: "Is the out-of-service taint safe to use"  for longer discussion. 
---

TLDR of longer discussion: 


EBS CSI Driver has no risk with out of service taint, because EBS will not allow EC2 DetachVolume call to succeed IF volume is still mounted. AND EC2 will flush disk io, auto unmount every mount, and detach all EBS volumes upon instance termination.  

What we can confirm is that all CSI Drivers who are not guaranteed safe, already experience this risk today in all Kubernetes versions, because Kubernetes only waits 6 minutes before issuing force detach. 

Only in Kubernetes v1.30 can cluster operator set a cluster level feature flag to disable this 6 min force detach. There are talks about disabling this 6 min force detach by default in K8s v1.32, which would mean our 6+ problem a delays would turn into infinite delays...

designs/statefulset-disruption.md
cartermckinnon: This variability is really unfortunate. I wonder if you see more consistent behavior with `ec2:StopInstances`? I assume you can detach a volume when an instance is stopped? If so, maybe stopping prior to termination could at least put a more reasonable upper bound on the delay in this case

designs/statefulset-disruption.md
AndrewSirenko: Today, for some reason only with karpenter, even if EBS Node Service container is kept alive due to pre-stop lifecycle hook, liveness probe container is killed, which can lead to Kubelet not relying on EBS node service to unmount (due to failed health check). Jason found out that we can resolve this issue by adding a sleep pre-stop hook to livenessprobe container. Unlike our pre-stop hook, that is a 'dumb' sleep which will delay all node terminations by sleep time. We can avoid having a liveness probe altogether by setting up a gRPC endpoint in our node service that kubelet can health check instead. 

See this section of k8s docs: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe

Thanks to @jmdeal for diving deep and finding sleep workaround, and  Connor knowing about this new feature. 

We want to implement this in next release of driver, which will allow for a real customer workaround. (That doesn't require cx manually editing/patching our daemonset with the sleep prestop for liveness probe)

designs/statefulset-disruption.md
AndrewSirenko: Yes, will bring info from line 269 to here.

designs/statefulset-disruption.md
AndrewSirenko: Thanks for this @cartermckinnon, two notes for you. 

> I assume you can detach a volume when an instance is stopped? If so, maybe stopping prior to termination would at least put a more reasonable upper bound on the delay in this case

1. Great idea on `ec2:StopInstances`, but I'm not sure if the extra step is worth it considering that one can only detach a volume after instance reaches `stopped` state (which would take an extra ~8 seconds).  I measured the difference between stopping/terminating durations for a couple different instance types [here](https://github.com/aws/karpenter-provider-aws/blob/95c2945d8bdb8ea6b3ce6e0d3f1f787e78f7a192/designs/statefulset-disruption.md#instance-stoppedterminated-times). Looks like `stopped` is 10-40 seconds faster depending on instance type or EC2 quirks. 

2. My original document was misleading when talking about GPU/Windows instance termination times. They may be closer to linux termination times than metal instances. I have added an [instance termination timings section](https://github.com/aws/karpenter-provider-aws/blob/95c2945d8bdb8ea6b3ce6e0d3f1f787e78f7a192/designs/statefulset-disruption.md#instance-stoppedterminated-times) where I tested a few instance types. I was relying on hearsay timings when I should have dived deeper before the meeting, my apologies.  

> The guest OS can take a long time to complete shutting down.

I can try digging into this with the folks at EC2.

designs/statefulset-disruption.md
AndrewSirenko: Oh I see what you're saying now. I'll remove this bullet for now because it is un-needed unless customers have a very weird (probably incorrect) configuration.

AndrewSirenko:
/hold

Edits

AndrewSirenko:
@jmdeal Since this change ended up being done at the k-sigs/karpenter level, and we don't plan on making it part of the eventual "pre-node-deletion-hooks" plan, should I move an up-to-date and shortened version of this document to that repo?

jmdeal:
Sorry for the late response @AndrewSirenko, if you wanted to do that that would be great! I'll close this PR out and we can track over in k-sigs.

PR #6343, opened 2024-06-10
test: Enable Windows IPAM at suite level

engedaam:
Fixes #N/A 

**Description**
-  `ExpectUpdated` was running into a race condition of dropping update calls if the client cache has not been updated after an object has been updated in the API server. In the case of windows failures, we have consecutive tests that run one after another that update the same object, `amazon-vpc-cni` configmap. We see that the object is updated in the API server, however the client cache has not been updated. The tests expectation assume the object we currently receive, the old object, is what we the state we would like to update the resource, which results in us dropping an update call. 
- The race condition between the object read and writes for updating the configMap amazon-vpc-cni after each test run. The Update operation for enabling and disabling of the windows IPAM after each test creates a race between the object that's being written and the object that received for updating, causing the windows IPAM not to be enabled.
- Excluded the `c7i-flex` until it is supported by AWS VPC resource controller 

- `/karpenter versionCompatibility`

- [ ] Yes, issue opened: # 
- [x] No


engedaam (commented):
/karpenter versionCompatibility

engedaam (commented):
/karpenter versionCompatibility

engedaam (commented):
/karpenter versionCompatibility

engedaam (commented):
/karpenter versionCompatibility

engedaam (commented):
/karpenter versionCompatibility

engedaam (commented):
/karpenter versionCompatibility

engedaam (commented):
/karpenter versionCompatibility

test/pkg/environment/aws/environment.go
```diff
@@ -62,7 +62,7 @@ var EphemeralInitContainerImage = "alpine"
 // ExcludedInstanceFamilies denotes instance families that have issues during resource registration due to compatibility
 // issues with versions of the VPR Resource Controller.
 // TODO: jmdeal@ remove a1 from exclusion list once Karpenter implicitly filters a1 instances for AL2023 AMI family (incompatible)
-var ExcludedInstanceFamilies = []string{"m7a", "r7a", "c7a", "r7i", "a1"}
+var ExcludedInstanceFamilies = []string{"m7a", "r7a", "c7a", "r7i", "a1", "c7i-flex"}
```
jonathan-innis, line 65: This doesn't seem like something that we would need special handling for just for testing. I'm surprised that this isn't being caught by the instance type provider directly.

test/pkg/environment/common/expectations.go
```diff
@@ -80,9 +79,6 @@ func (env *Environment) ExpectUpdated(objects ...client.Object) {
 		Eventually(func(g Gomega) {
 			current := o.DeepCopyObject().(client.Object)
 			g.Expect(env.Client.Get(env.Context, client.ObjectKeyFromObject(current), current)).To(Succeed())
-			if current.GetResourceVersion() != o.GetResourceVersion() {
```
jonathan-innis, line 83: Why are we choosing to drop this here? I think that it's fine to keep this function as-is, but to just call Update directly for cases when we want to directly handle the Conflict error that's coming back. Anyways, IIRC, we just need to drop the equality check and the blind patch should be enough to fix the issue here

test/pkg/environment/aws/environment.go
engedaam: So this is occurring specifically eu-west-1. This is because it was recently released and is available in us-east-2. https://aws.amazon.com/about-aws/whats-new/2024/05/amazon-ec2-c7i-flex-instances/#:~:text=C7i%2Dflex%20instances%20are%20available,Amazon%20EC2%20C7i%2Dflex%20instances.

engedaam: Are you suggesting Karpenter would filter out unsupported instance types from the vpc-resource-controller?

test/pkg/environment/common/expectations.go
engedaam: Sounds good, I thought we would want to make a build call even if we have a different version then expected.

test/pkg/environment/aws/environment.go
jonathan-innis: Yes, that's what I'm suggesting, at least for Windows instances where it matters that this value exists.

jonathan-innis (approved):

PR #6346, opened 2024-06-10
feat: Reduce hop token count

rschalo:
Fixes #N/A 

**Description**
Disables IMDS access from pods by reducing hop limit to 1

- [ ] Yes, issue opened: # 
- [x] No


jonathan-innis (approved):
Is there any testing around this? If not, should we add some?

rschalo:
Yes I'll look at adding an integration test, still thinking about how we could model this.

rschalo:
Discussed testing offline - moving ahead without for now.

applike-ss:
This will make the aws-load-balancer-controller unable to reach the EC2 Metadata service, which in turn requires to specify the region and vpcId.

Was this taken into consideration?

PR #6354, opened 2024-06-12
fix: default to multizone karpenter deployment

rschalo:
Fixes #N/A 

**Description**
Change default behavior so that both replicas cannot schedule to the same availability zone.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -81,7 +81,7 @@ affinity:
 topologySpreadConstraints:
   - maxSkew: 1
     topologyKey: topology.kubernetes.io/zone
-    whenUnsatisfiable: ScheduleAnyway
+    whenUnsatisfiable: DoNotSchedule
 # -- Tolerations to allow the pod to be scheduled to nodes with taints.
 tolerations:
   - key: CriticalAddonsOnly


charts/karpenter/values.yaml
```diff
@@ -81,7 +81,7 @@ affinity:
 topologySpreadConstraints:
   - maxSkew: 1
     topologyKey: topology.kubernetes.io/zone
-    whenUnsatisfiable: ScheduleAnyway
+    whenUnsatisfiable: DoNotSchedule
```
engedaam, line 84: Just wondering, is it possible for MNG to provide nodes in the same zone? Do we need to include zone spread for nodes launched by MNG?

jmdeal:
I'm on board with changing this to be the default, but should we surface a configuration option in the helm chart to override this default? If we did this from the start I'd defer to user demand, but this would now be a breaking change for customers who may have infra already set up relying on this being a `ScheduleAnyway` TSC.

charts/karpenter/values.yaml
rschalo: The MNG spans every subnet that is specified when creating the group so it's possible a cx provides one and only one subnet. As a result, the MNG only launches into that zone.

rschalo:
So, my thinking is that if leave this as is and don't surface an option to override, there are two kinds of cx that would break. The first are those that would already break given a zonal outage, which would not different behavior with `DoNotSchedule` as the default. The other case, is that Karpenter is running two replicas on the same node, the leader pod goes unhealthy, loses leadership, and then there is not a second node because of the `DoNotSchedule` TSC.

This PR changes the value which can be changed by cx, I think I missed the actual default in the template though, I'll add it.

charts/karpenter/values.yaml
jonathan-innis: DoNotSchdule strikes me as the correct default here. Users have the ability to opt-in to `ScheudleAnyways`, but the vast majority of users are going to want these spread across both multiple nodes and multiple zones. You don't have guaranteed resiliency across zones today, so this seems like a solid change to make 👍🏼

PR #6374, opened 2024-06-17
test: Fix scale test provisioning spike

jigisha620:
Fixes #N/A 

**Description**
We have been seeing spikes in the scale tests for provisioning. That's because there are two different tests that are captured by the same metric and the execution times are different for both.
1. Metric `node-dense` for node dense scale-up and node dense scale-up with minValues
2. Metric `pod-dense` for pod dense scale-up and pod dense scale-up with minValues

Created a different metric for scale up with and without min values. Now we have `node-dense`, `node-dense-min-val`, `pod-dense`, `pod-dense-min-val`.

- [ ] Yes, issue opened: # 
- [x] No


jmdeal (approved):
LGTM 🚀

PR #6379, opened 2024-06-18
feat: Drop karpenter.sh/managed-by annotation

rschalo:
Fixes #N/A 
Addresses concerns in https://github.com/aws/karpenter-provider-aws/pull/6319 and points to correct branch.

**Description**

- [ ] Yes, issue opened: # 
- [x] No


pkg/cloudprovider/cloudprovider.go
```diff
@@ -345,8 +345,8 @@ func (c *CloudProvider) instanceToNodeClaim(i *instance.Instance, instanceType *
 	if v, ok := i.Tags[corev1beta1.NodePoolLabelKey]; ok {
 		labels[corev1beta1.NodePoolLabelKey] = v
 	}
-	if v, ok := i.Tags[corev1beta1.ManagedByAnnotationKey]; ok {
-		annotations[corev1beta1.ManagedByAnnotationKey] = v
+	if v, ok := i.Tags[v1beta1.EKSClusterNameAnnotationKey]; ok {
+		annotations[v1beta1.EKSClusterNameAnnotationKey] = v
```
jonathan-innis, line 349: Confirming: We can use a key value with a colon? IIRC, you mentioned that this wasn't possible and would cause failures?

rschalo: We cannot have a key value with a colon. After looking more, I don't think this annotation is actually used anywhere, wdyt about remove it entirely?

Makefile
```diff
@@ -57,7 +57,8 @@ test: ## Run tests
 		-cover -coverprofile=coverage.out -outputdir=. -coverpkg=./... \
 		--ginkgo.focus="${FOCUS}" \
 		--ginkgo.randomize-all \
-		--ginkgo.vv
+		--ginkgo.v \
+		--ginkgo.fail-fast \
```
njtran, line 61: I'm not sure we want this. We should also end the last line without a \

pkg/controllers/nodeclaim/garbagecollection/controller.go
```diff
@@ -61,18 +61,18 @@ func (c *Controller) Reconcile(ctx context.Context) (reconcile.Result, error) {
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("listing cloudprovider machines, %w", err)
 	}
-	managedRetrieved := lo.Filter(retrieved, func(nc *v1beta1.NodeClaim, _ int) bool {
-		return nc.Annotations[v1beta1.ManagedByAnnotationKey] != "" && nc.DeletionTimestamp.IsZero()
+	managedRetrieved := lo.Filter(retrieved, func(nc *corev1beta1.NodeClaim, _ int) bool {
+		return nc.Annotations[v1beta1.LabelNodeClass] != "" && nc.DeletionTimestamp.IsZero()
```
njtran, line 65: Can we use the nodepool label? I think it should be the same, but I want to make sure we follow the convention in the upstream provider.
```suggestion
		return nc.Annotations[v1beta1.LabelNodePool] != "" && nc.DeletionTimestamp.IsZero()
```

pkg/controllers/nodeclaim/garbagecollection/suite_test.go
```diff
@@ -137,7 +137,8 @@ var _ = Describe("GarbageCollection", func() {
 		awsEnv.EC2API.Instances.Store(aws.StringValue(instance.InstanceId), instance)
 
 		ExpectSingletonReconciled(ctx, garbageCollectionController)
-		_, err := cloudProvider.Get(ctx, providerID)
+		nc, err := cloudProvider.Get(ctx, providerID)
+		fmt.Println(nc)
```
njtran, line 141: Need to remove?

pkg/controllers/nodeclaim/garbagecollection/suite_test.go
```diff
@@ -288,9 +289,9 @@ var _ = Describe("GarbageCollection", func() {
 		Expect(err).NotTo(HaveOccurred())
 	})
 	It("should not delete an instance if it was not launched by a NodeClaim", func() {
-		// Remove the "karpenter.sh/managed-by" tag (this isn't launched by a machine)
+		// Remove the nodepool tag (this isn't launched by a machine)
```
njtran, line 292: You say nodepool, but the label below is nodeclass. Mistake?

PR #6390, opened 2024-06-20
chore: Add soft eviction setting to Bottlerocket kublet config

lolbrish:
Fixes #N/A

**Description**

Issue https://github.com/bottlerocket-os/bottlerocket/issues/1445 is fixed and so soft eviction settings can now be used with Bottlerocket. It looks like old docs have already been removed that used to discuss this limitation.

Tested on a live EKS cluster to provision Bottlerocket nodes, then the kubelet configuration was checked using `http://localhost:8001/api/v1/nodes/$NODE/proxy/configz`.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/providers/amifamily/bootstrap/bottlerocket.go
+++ b/pkg/providers/amifamily/bootstrap/bottlerocket.go
@@ -64,6 +64,18 @@ func (b Bottlerocket) Script() (string, error) {
 		if b.KubeletConfig.EvictionHard != nil {
 			s.Settings.Kubernetes.EvictionHard = b.KubeletConfig.EvictionHard
 		}
+		if b.KubeletConfig.EvictionMaxPodGracePeriod != nil {
+			s.Settings.Kubernetes.EvictionMaxPodGracePeriod = aws.Int(int(ptr.Int32Value(b.KubeletConfig.EvictionMaxPodGracePeriod)))
+		}
+		if b.KubeletConfig.EvictionSoft != nil {
+			s.Settings.Kubernetes.EvictionSoft = b.KubeletConfig.EvictionSoft
+		}
+		if b.KubeletConfig.EvictionSoftGracePeriod != nil {
+			s.Settings.Kubernetes.EvictionSoftGracePeriod = make(map[string]string)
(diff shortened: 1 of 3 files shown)


pkg/providers/amifamily/bottlerocket.go
```diff
@@ -128,17 +128,11 @@ func (b Bottlerocket) EphemeralBlockDevice() *string {
...
-// not currently support the evictionSoft parameter passed through the kubernetes settings TOML userData
-// If a NodePool sets the evictionSoft value when using the Bottlerocket AMIFamily in the provider,
-// evictionSoft will be ignored
-// https://github.com/bottlerocket-os/bottlerocket/issues/1445
-
 func (b Bottlerocket) FeatureFlags() FeatureFlags {
 	return FeatureFlags{
 		UsesENILimitedMemoryOverhead: false,
 		PodsPerCoreEnabled:           false,
-		EvictionSoftEnabled:          false,
```
jonathan-innis, line 141: I don't even think we need the feature flag at all now since I believe they are all `true` now. You can probably just drop the field outright from the struct now

pkg/providers/amifamily/bootstrap/bottlerocket.go
```diff
@@ -64,6 +64,18 @@ func (b Bottlerocket) Script() (string, error) {
 		if b.KubeletConfig.EvictionHard != nil {
 			s.Settings.Kubernetes.EvictionHard = b.KubeletConfig.EvictionHard
 		}
+		if b.KubeletConfig.EvictionMaxPodGracePeriod != nil {
```
jonathan-innis, line 67: Can we add a test (at least a functional one) for the changes that you are making here. There may also be a change that you could make to the current E2E integration testing in our Kubelet Config suite tests that would extend eviction soft testing for BR

jonathan-innis:
Thanks for taking this up! A few comments, mainly about testing 👍🏼

pkg/providers/amifamily/bootstrap/bottlerocket.go
lolbrish: I'll try to take a look at this in the next few days

PR #6404, opened 2024-06-26
adding a new interface for ami resolver

ravisinha0506:
ci: Adding a new interface for AMI resolver

**Description**
Adding a new interface for AMI resolver as this class didn't have any existing interface which makes it tightly coupled with LaunchTemplateProvider. 

I haven't tested this change locally

- [x] No

### DIFF
--- a/pkg/providers/amifamily/resolver.go
+++ b/pkg/providers/amifamily/resolver.go
@@ -34,6 +34,10 @@ import (
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 )
 
+type AMIResolver interface {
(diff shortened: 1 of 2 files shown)


pkg/providers/amifamily/resolver.go
```diff
@@ -34,6 +34,10 @@ import (
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 )
 
+type AMIResolver interface {
```
jmdeal, line 37: Following the naming convention for the other provider interfaces, this should be `Resolver` and we should rename the concrete type `DefaultResolver`.

jmdeal (commented):
One naming nit, otherwise lgtm

PR #6438, opened 2024-07-01
WIP: Add status conditions for nodeClass

jigisha620:
Fixes #N/A 

**Description**
This PR adds status conditions to EC2nodeClass.

/karpenter snapshot

- [ ] Yes, issue opened: # 
- [x] No


jigisha620 (commented):
/karpenter snapshot

pkg/controllers/nodeclass/status/readiness.go
```diff
@@ -1,63 +0,0 @@
...
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-
-	"github.com/aws/karpenter-provider-aws/pkg/apis/v1beta1"
-)
-
-type Readiness struct {
-	launchTemplateProvider launchtemplate.Provider
-}
-
-func (n Readiness) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (reconcile.Result, error) {
```
engedaam, line 35: Can just not move these checks to each reconcile function of each resource? Instead of having a controller for each field we change

pkg/providers/launchtemplate/suite_test.go
```diff
@@ -1958,8 +1962,11 @@ var _ = Describe("LaunchTemplate Provider", func() {
 				}})
 				nodeClass.Spec.AMISelectorTerms = []v1beta1.AMISelectorTerm{{Tags: map[string]string{"*": "*"}}}
 				ExpectApplied(ctx, env.Client, nodeClass)
-				controller := status.NewController(env.Client, awsEnv.SubnetProvider, awsEnv.SecurityGroupProvider, awsEnv.AMIProvider, awsEnv.InstanceProfileProvider, awsEnv.LaunchTemplateProvider)
-				ExpectObjectReconciled(ctx, env.Client, controller, nodeClass)
+				ExpectObjectReconciled(ctx, env.Client, subnet.NewController(env.Client, awsEnv.SubnetProvider), nodeClass)
```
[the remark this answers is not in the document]
jigisha620: Discussed this offline. The base condition `Ready` will not be set to `true` until all the conditions are `true`. That's why we want to run each controller to set it's respective status condition.

pkg/controllers/nodeclass/status/ami/controller.go
```diff
@@ -0,0 +1,123 @@
...
+					Name:         ami.Name,
+					ID:           ami.AmiID,
+					Requirements: reqs,
+				}
+			})
+			nodeClass.StatusConditions().SetTrue(v1beta1.ConditionTypeAMIsReady)
+		}
+	}
+	if !equality.Semantic.DeepEqual(stored, nodeClass) {
+		if err = c.kubeClient.Status().Update(ctx, nodeClass); err != nil {
```
jmdeal, line 97: Why does this need to be an update over a patch? We're only updating the existing status condition right? Do we expect any other controllers to be updating this status condition which we could conflict with? Same comment for all other controllers.

pkg/controllers/nodeclass/status/instance_profile/controller.go
```diff
@@ -77,36 +59,40 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
-		c.ami,
-		c.subnet,
-		c.securitygroup,
-		c.instanceprofile,
-		c.readiness,
-	} {
-		res, err := reconciler.Reconcile(ctx, nodeClass)
-		errs = multierr.Append(errs, err)
-		results = append(results, res)
+	if nodeClass.Spec.Role != "" {
```
jmdeal, line 63: nit: IMO it would be cleaner to pull this logic out into a helper function, and do away with the multierror. This is going to be a theme in my review, but we should avoid multierrors if they're not needed since it's use implies there will be multiple errors IMO. 

```go
func (c *Controller) Reconcile(...) {
    ...
    instanceProfile, err := c.resolveInstanceProfile(nodeClass)
    if err != nil {
        nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeInstanceProfileReady, "InstanceProfileCreateError", "Error creating instance profile")
    } else {
        nodeClass.Status.InstanceProfile = instanceProfile
        nodeClass.StatusConditions().SetTrue(v1beta1.ConditionTypeInstanceProfileReady)
    }

    if !equality.Semantic.DeepEqual(stored, nodeClass) {
        ...
    }
    if err != nil {
    }
    ...
}

func (c *Controller) resolveInstanceProfile(nc *v1beta1.EC2NodeClass) (string, error) { 
    if nodeClass.Spec.Role == "" {
        return lo.FromPtr(nodeClass.Spec.InstanceProfile), nil
    }
    return c.instanceProfileProvider.Create(ctx, nc)
}
```

pkg/controllers/nodeclass/status/launch_template/controller.go
```diff
@@ -0,0 +1,89 @@
...
+			return reconcile.Result{}, err
+		}
+	}
+	// A NodeClass that uses AL2023 requires the cluster CIDR for launching nodes.
+	// To allow Karpenter to be used for Non-EKS clusters, resolving the Cluster CIDR
+	// will not be done at startup but instead in a reconcile loop.
+	if lo.FromPtr(nodeClass.Spec.AMIFamily) == v1beta1.AMIFamilyAL2023 {
+		if err := c.launchTemplateProvider.ResolveClusterCIDR(ctx); err != nil {
+			nodeClass.StatusConditions().SetFalse(status.ConditionReady, "NodeClassNotReady", "Failed to detect the cluster CIDR")
+			if e := c.kubeClient.Status().Update(ctx, nodeClass); e != nil {
```
jmdeal, line 67: nit: Can we pull this out and do our standard deepequal check? Really just looking to reduce nesting, plus the common format across our controllers makes it easy to reason about.

pkg/controllers/nodeclass/status/launch_template/controller.go
```diff
@@ -0,0 +1,89 @@
...
+		if err := c.kubeClient.Patch(ctx, nodeClass, client.MergeFrom(stored)); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+	// A NodeClass that uses AL2023 requires the cluster CIDR for launching nodes.
+	// To allow Karpenter to be used for Non-EKS clusters, resolving the Cluster CIDR
+	// will not be done at startup but instead in a reconcile loop.
+	if lo.FromPtr(nodeClass.Spec.AMIFamily) == v1beta1.AMIFamilyAL2023 {
+		if err := c.launchTemplateProvider.ResolveClusterCIDR(ctx); err != nil {
+			nodeClass.StatusConditions().SetFalse(status.ConditionReady, "NodeClassNotReady", "Failed to detect the cluster CIDR")
```
jmdeal, line 66: Isn't the root condition driven by the other conditions? How does this interact with operatorpkg's mechanism to set this root condition?

pkg/controllers/nodeclass/status/instance_profile/controller.go
```diff
@@ -12,62 +12,44 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package status
+package instance_profile
```
jmdeal, line 15: Package names shouldn't include underscores, same comment for `launch_template` and `security_group`.

pkg/controllers/nodeclass/status/security_group/controller.go
```diff
@@ -0,0 +1,115 @@
...
+		stored := nodeClass.DeepCopy()
+		controllerutil.AddFinalizer(nodeClass, v1beta1.TerminationFinalizer)
+		if err := c.kubeClient.Patch(ctx, nodeClass, client.MergeFrom(stored)); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+	stored := nodeClass.DeepCopy()
+
+	securityGroups, err := c.securityGroupProvider.List(ctx, nodeClass)
+	var errs error
```
jmdeal, line 66: Can we also rework this to not use multierror, maybe by moving into a helper function. In general there's a lot of nesting here, moving it into a helper function with some early returns could help readability.

pkg/controllers/nodeclass/status/security_group/controller.go
```diff
@@ -0,0 +1,115 @@
...
+			if errors.IsConflict(err) {
+				return reconcile.Result{Requeue: true}, nil
+			}
+			return reconcile.Result{}, client.IgnoreNotFound(err)
+		}
+	}
+	if errs != nil {
+		return reconcile.Result{}, errs
+	}
+	if !nodeClass.StatusConditions().IsTrue(v1beta1.ConditionTypeSecurityGroupsReady) {
```
jmdeal, line 99: Like with AMIs I'm not sure we should be rereconciling immediately here. Is there a reason we would fail to resolve security groups once but succeed immediately after (without any changes to the NodeClass).

pkg/controllers/nodeclass/status/subnet/controller.go
```diff
@@ -0,0 +1,118 @@
...
+	if !controllerutil.ContainsFinalizer(nodeClass, v1beta1.TerminationFinalizer) {
+		stored := nodeClass.DeepCopy()
+		controllerutil.AddFinalizer(nodeClass, v1beta1.TerminationFinalizer)
+		if err := c.kubeClient.Patch(ctx, nodeClass, client.MergeFrom(stored)); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+	stored := nodeClass.DeepCopy()
+	subnets, err := c.subnetProvider.List(ctx, nodeClass)
+	var errs error
```
jmdeal, line 66: Same comment I've given elsewhere about multierr and reducing nesting with a helper function.

pkg/controllers/nodeclass/status/subnet/controller.go
```diff
@@ -0,0 +1,118 @@
...
+			if errors.IsConflict(err) {
+				return reconcile.Result{Requeue: true}, nil
+			}
+			return reconcile.Result{}, client.IgnoreNotFound(err)
+		}
+	}
+	if errs != nil {
+		return reconcile.Result{}, errs
+	}
+	if !nodeClass.StatusConditions().IsTrue(v1beta1.ConditionTypeSubnetsReady) {
```
jmdeal, line 102: Same comment I've given on other controllers about the immediate requeue.

pkg/controllers/nodeclass/status/ami/controller.go
jigisha620: I tried to follow what we have done with most of our controllers for status updates. Is there a benefit of using patch over update?

PR #6450, opened 2024-07-03
feat: implement v1 AMI selection

jmdeal:
Fixes #N/A 

**Description**
This PR updates AMISelectorTerms for v1 by implementing the `alias` term.

EKS optimized AMIs can now be selected using the `alias` term. In addition to supporting the same behavior as `v1beta1`, users can also pin to a specific EKS optimized AMI by version.

Example:
```yaml
apiVersion: karpenter.sh/v1
kind: EC2NodeClass
spec:
  amiSelectorTerms:
    - alias: "al2023@latest"
    # Or pin to a specific version
    # - alias: "al2023@v20240625"
```
The following aliases are supported:
- `al2023@<version>`
- `al2@<version>`
- `bottlerocket@<version>`
- `windows2019@<version>`
- `windows2022@<version>`

Additionally, this test:
- Drops support for the Ubuntu AMI family
- Drops `EC2NodeClass.Spec.AMIFamily`

- [ ] Yes, issue opened: # 
- [x] No


pkg/providers/amifamily/ami.go
```diff
@@ -126,163 +82,131 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
+				}
+				requirementSets := func() []scheduling.Requirements {
+					if knownRequirements, ok := query.KnownRequirements[lo.FromPtr(image.ImageId)]; ok {
+						return lo.Map(knownRequirements, func(r scheduling.Requirements, _ int) scheduling.Requirements {
+							r.Add(scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, arch))
+							return r
+						})
+					}
+					return []scheduling.Requirements{scheduling.NewRequirements(scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, arch))}
+				}()
```
njtran, line 168: Wondering why this needs to be its own in-line function. Worth adding as a helper?

pkg/providers/amifamily/types.go
```diff
@@ -0,0 +1,98 @@
...
+	case VariantNeuron:
+		return scheduling.NewRequirements(scheduling.NewRequirement(v1beta1.LabelInstanceGPUCount, v1.NodeSelectorOpExists))
+	}
+	return nil
+}
+
+type AMIQuery struct {
+	Filters           []*ec2.Filter
+	Owners            []string
+	KnownRequirements map[string][]scheduling.Requirements
```
njtran, line 88: Can you add a comment about what the string should be, and when these knownRequirements will be set?
```suggestion
        // Map from ami-id to scheduling requirements
	KnownRequirements map[string][]scheduling.Requirements
```

pkg/providers/amifamily/ami.go
```diff
@@ -126,163 +82,131 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
-	return res
+	images := map[uint64]AMI{}
+	for _, query := range queries {
+		if err = p.ec2api.DescribeImagesPagesWithContext(ctx, query.DescribeImagesInput(), func(page *ec2.DescribeImagesOutput, _ bool) bool {
+			for _, image := range page.Images {
+				arch, ok := v1beta1.AWSToKubeArchitectures[lo.FromPtr(image.Architecture)]
+				if !ok {
+					continue
+				}
+				requirementSets := func() []scheduling.Requirements {
```
njtran, line 160: This adds the architecture requirement for an image and then returns those requirements, right? If there are no requirements, just returns the architecture? I'm having a hard time understanding why it needs to be an in-line funciton and why you need the lo.Map() function on L162.

pkg/providers/amifamily/ami.go
```diff
@@ -126,163 +82,131 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
+						return lo.Map(knownRequirements, func(r scheduling.Requirements, _ int) scheduling.Requirements {
+							r.Add(scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, arch))
+							return r
+						})
+					}
+					return []scheduling.Requirements{scheduling.NewRequirements(scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, arch))}
+				}()
+				for _, reqs := range requirementSets {
+					reqsHash := lo.Must(hashstructure.Hash(reqs.NodeSelectorRequirements(), hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: true}))
+					// If the proposed image is newer, store it so that we can return it
```
njtran, line 171: This comment should really be describing the below, right?
```suggestion
					// If the proposed image has the same set of requirements, replace the older image with the newer image
```

pkg/providers/amifamily/ami.go
```diff
@@ -126,163 +82,131 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
 }
 
-func (p *DefaultProvider) getRequirementsFromImage(ec2Image *ec2.Image) scheduling.Requirements {
-	requirements := scheduling.NewRequirements()
-	// Always add the architecture of an image as a requirement, irrespective of what's specified in EC2 tags.
-	architecture := *ec2Image.Architecture
-	if value, ok := v1beta1.AWSToKubeArchitectures[architecture]; ok {
-		architecture = value
+// MapToInstanceTypes returns a map of AMIIDs that are the most recent on creationDate to compatible instancetypes
+func MapToInstanceTypes(instanceTypes []*cloudprovider.InstanceType, amis []v1beta1.AMI) map[string][]*cloudprovider.InstanceType {
```
njtran, line 200: Thinking about this function, I think we only use this for Drift.  Wondering it would be better if we just returned the requirements of the image, and then compared the instance type directly to those requirements as a function of the instance type.
```
	mappedAMIs := amifamily.MapToInstanceTypes([]*cloudprovider.InstanceType{nodeInstanceType}, nodeClass.Status.AMIs)
	if !lo.Contains(lo.Keys(mappedAMIs), instance.ImageID) {
		return AMIDrift, nil
	}
	return "", nil```

File: pkg/providers/amifamily/bottlerocket.go [hunk 6]
```diff
@@ -37,54 +42,37 @@ type Bottlerocket struct {
...
+		}},
+		KnownRequirements: make(map[string][]scheduling.Requirements),
+	}
+	for rootPath, variants := range map[string][]Variant{
+		fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s", k8sVersion):        []Variant{VariantStandard},
+		fmt.Sprintf("/aws/service/bottlerocket/aws-k8s-%s-nvidia", k8sVersion): []Variant{VariantNeuron, VariantNvidia},
+	} {
+		results, err := ssmProvider.List(ctx, rootPath)
+		if err != nil {
+			log.FromContext(ctx).WithValues("path", rootPath).Error(err, "discovering AMIs from ssm")
```
njtran, line 58: consider strongly typing the ami family here so it's easier to distinguish in logs. Same goes for the other ami families
```suggestion
			log.FromContext(ctx).WithValues("path", rootPath).Error(err, "discovering bottlerocket AMIs from ssm")
```

pkg/providers/amifamily/custom.go
```diff
@@ -39,8 +42,8 @@ func (c Custom) UserData(_ *corev1beta1.KubeletConfiguration, _ []v1.Taint, _ ma
 	}
 }
 
-func (c Custom) DefaultAMIs(_ string) []DefaultAMIOutput {
-	return nil
+func (c Custom) AMIQuery(_ context.Context, _ ssm.Provider, _ string, _ string) (AMIQuery, error) {
```
njtran, line 45: I think it's more clear that this should be nil when this is renamed to SSMPaths, or something of the like.

pkg/providers/amifamily/resolver.go
```diff
@@ -29,11 +30,16 @@ import (
...
 	"github.com/aws/karpenter-provider-aws/pkg/apis/v1beta1"
 	"github.com/aws/karpenter-provider-aws/pkg/providers/amifamily/bootstrap"
+	"github.com/aws/karpenter-provider-aws/pkg/providers/ssm"
 
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 )
 
+const (
+	AMIVersionLatest = "latest"
```
njtran, line 40: Worth calling out in a function what this refers to. Some users reading this might confuse it as the key used for SSM path discovery, which would be wrong when `latest` == `/recommended`.

pkg/providers/amifamily/suite_test.go
```diff
@@ -184,20 +174,21 @@ var _ = Describe("AMIProvider", func() {
...
 		Expect(err).ToNot(HaveOccurred())
 		Expect(amis).To(HaveLen(1))
 	})
-	It("should succeed to resolve AMIs (Custom)", func() {
-		nodeClass.Spec.AMIFamily = &v1beta1.AMIFamilyCustom
-		amis, err := awsEnv.AMIProvider.List(ctx, nodeClass)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(amis).To(HaveLen(0))
-	})
+	// TODO: (jmdeal@) Revisit this test, what is it meant to be doing
```
njtran, line 185: Seems like the test is just making sure that we don't return any default AMIs for custom AMI family, right, but that it doesn't succeed?

pkg/providers/amifamily/types.go
```diff
@@ -0,0 +1,98 @@
...
+	VariantStandard    Variant = "standard"
+	VariantNvidia      Variant = "nvidia"
+	VariantNeuron      Variant = "neuron"
+)
+
+func NewVariant(v string) (Variant, error) {
+	var wellKnownVariants = sets.New(VariantStandard, VariantNvidia, VariantNeuron)
+	variant := Variant(v)
+	if !wellKnownVariants.Has(variant) {
+		return variant, fmt.Errorf("%q is not a well-known variant", variant)
```
njtran, line 65: When can this 😉 invariant 😉 be reached?

pkg/providers/ssm/provider.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func NewDefaultProvider(ssmapi ssmiface.SSMAPI, cache *cache.Cache) *DefaultProvider {
+	return &DefaultProvider{
+		ssmapi: ssmapi,
+		cache: cache,
+	}
+}
+
+func (p *DefaultProvider) List(ctx context.Context, path string) (map[string]string, error) {
```
njtran, line 46: Worth calling out in the comments that List does a recursive list based on the base path, whereas Get just does a one call on the path equality.

pkg/test/nodeclass.go
```diff
@@ -36,8 +36,8 @@ func EC2NodeClass(overrides ...v1beta1.EC2NodeClass) *v1beta1.EC2NodeClass {
 			panic(fmt.Sprintf("Failed to merge settings: %s", err))
 		}
 	}
-	if options.Spec.AMIFamily == nil {
-		options.Spec.AMIFamily = &v1beta1.AMIFamilyAL2
+	if len(options.Spec.AMISelectorTerms) == 0 {
+		options.Spec.AMISelectorTerms = []v1beta1.AMISelectorTerm{{Alias: "al2@latest"}}
```
njtran, line 40: might as well upgrade?
```suggestion
		options.Spec.AMISelectorTerms = []v1beta1.AMISelectorTerm{{Alias: "al2023@latest"}}
```

test/suites/scheduling/suite_test.go
njtran, line 1: loooks like you have some ci checks on test/suites/drift/suite_test.go. Can't comment on the file since it's not in the diff though

njtran (commented):
Cool to see progress on this! Good work!

pkg/providers/amifamily/al2.go
```diff
@@ -28,47 +34,61 @@ import (
...
+			log.FromContext(ctx).WithValues("path", rootPath).Error(err, "discovering AMIs from ssm")
+			continue
+		}
+		for path, value := range results {
+			pathComponents := strings.Split(path, "/")
+			// Only select image_id paths which match the desired AMI version
+			if len(pathComponents) == 8 || pathComponents[7] != "image_id" {
+				continue
+			}
+			if av, err := a.extractAMIVersion(pathComponents[9]); err != nil || av != amiVersion {
```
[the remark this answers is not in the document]
jmdeal: Fixed

pkg/providers/amifamily/al2.go
```diff
@@ -28,47 +34,61 @@ import (
...
+		return AMIQuery{}, fmt.Errorf("failed to discover any AMIs for alias")
+	}
+	return query, nil
+}
+
+func (a AL2) extractAMIVersion(versionStr string) (string, error) {
+	if versionStr == "recommended" {
+		return AMIVersionLatest, nil
+	}
+	rgx := regexp.MustCompile(`^.*(v\d+)$`)
```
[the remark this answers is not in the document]
jmdeal: Good point, I'll update the regex 👍

pkg/providers/amifamily/al2.go
```diff
@@ -28,47 +34,61 @@ import (
...
+	matches := rgx.FindStringSubmatch(versionStr)
+	if len(matches) != 2 {
```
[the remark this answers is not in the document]
jmdeal: I'm sticking with the go naming convention, it's the first match (the entire regex) and then the individual capture groups. So basically it should always be the number of capture groups, plus 1.

pkg/providers/amifamily/al2023.go
```diff
@@ -15,38 +15,72 @@ limitations under the License.
...
+		KnownRequirements: make(map[string][]scheduling.Requirements),
+	}
+	rootPath := fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2023", k8sVersion)
+	results, err := ssmProvider.List(ctx, rootPath)
+	if err != nil {
+		return AMIQuery{}, fmt.Errorf("discovering AMIs from ssm")
+	}
+	for path, value := range results {
+		pathComponents := strings.Split(path, "/")
+		if len(pathComponents) != 11 || pathComponents[10] != "image_id" {
```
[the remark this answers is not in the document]
jmdeal: I think this had a proper bounds check, right? We short circuit if `len(pathComponents) != 11`

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -362,6 +361,35 @@ func (in *EC2NodeClass) InstanceProfileTags(clusterName string) map[string]strin
...
+		}
+	}
+	return AMIFamilyCustom
+}
+
+func (in *EC2NodeClass) AMIVersion() string {
+	if term, ok := lo.Find(in.Spec.AMISelectorTerms, func(t AMISelectorTerm) bool {
+		return t.Alias != ""
+	}); ok {
+		return strings.Split(term.Alias, "@")[1]
```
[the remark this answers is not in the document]
jmdeal: We could do that, but I think we've largely taken the decision that if validation is skipped the user is in undefined behavior land and crashing could be appropriate.

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -362,6 +361,35 @@ func (in *EC2NodeClass) InstanceProfileTags(clusterName string) map[string]strin
...
+			return AMIFamilyAL2023
+		case "bottlerocket":
+			return AMIFamilyBottlerocket
+		case "windows2019":
+			return AMIFamilyWindows2019
+		case "windows2022":
+			return AMIFamilyWindows2022
+		}
+	}
+	return AMIFamilyCustom
```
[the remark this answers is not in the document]
jmdeal: It's a little bit different since this covers both conditional passes.

pkg/fake/ssmapi.go
```diff
@@ -60,6 +67,86 @@ func (a SSMAPI) GetParameterWithContext(_ context.Context, input *ssm.GetParamet
 	}, nil
 }
 
+func (a SSMAPI) GetParametersByPathPagesWithContext(_ context.Context, input *ssm.GetParametersByPathInput, f func(*ssm.GetParametersByPathOutput, bool) bool, _ ...request.Option) error {
+	if a.WantErr != nil {
+		return a.WantErr
+	}
+	if a.GetParametersByPathOutput != nil {
+		f(a.GetParametersByPathOutput, true)
```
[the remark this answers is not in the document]
jmdeal: WDYM we're not going to consume the result?

pkg/fake/ssmapi.go
```diff
@@ -60,6 +67,86 @@ func (a SSMAPI) GetParameterWithContext(_ context.Context, input *ssm.GetParamet
...
+		`\/aws\/service\/bottlerocket\/aws-k8s-.*`: []string{
+			"x86_64/latest/image_id",
+			"arm64/latest/image_id",
+		},
+		`\/aws\/service\/ami-windows-latest`: lo.FlatMap(supportedK8sVersions(), func(version string, _ int) []string {
+			return []string{
+				fmt.Sprintf("Windows_Server-2019-English-Core-EKS_Optimized-%s/image_id", version),
+				fmt.Sprintf("Windows_Server-2022-English-Core-EKS_Optimized-%s/image_id", version),
+			}
+		}),
```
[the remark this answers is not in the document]
jmdeal: There shouldn't be a cyclical dependency, but there also isn't a ton of value in defining these regexes / values in the AMI files. They're really only useful for this fake. 

The reason only Windows needs the k8s version injected is it's the only AMI which doesn't include the k8s version as part of the path hierarchy, but includes it directly in the image version. For all of the other AMIs we handle the k8s version is included in the path we send to `GetParametersByPath`, but not for Windows. This means for the other AMIs we can just append a k8s version independent set of suffixes to the path, but not for windows.

pkg/fake/ssmapi.go
```diff
@@ -60,6 +67,86 @@ func (a SSMAPI) GetParameterWithContext(_ context.Context, input *ssm.GetParamet
...
+		f(a.GetParametersByPathOutput, true)
+		return nil
+	}
+	if len(a.Parameters) != 0 {
+		f(&ssm.GetParametersByPathOutput{
+			Parameters: lo.FilterMap(lo.Entries(a.Parameters), func(p lo.Entry[string, string], _ int) (*ssm.Parameter, bool) {
+				if !strings.HasPrefix(p.Key, lo.FromPtr(input.Path)) {
+					return nil, false
+				}
+				if strings.TrimPrefix(p.Key, lo.FromPtr(input.Path))[0] != '/' {
```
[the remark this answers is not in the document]
jmdeal: Turns out that is not a valid input, you can only pass a subpath.

pkg/providers/amifamily/al2023.go
```diff
@@ -15,38 +15,72 @@ limitations under the License.
...
-			),
-		},
+func (a AL2023) AMIQuery(ctx context.Context, ssmProvider ssm.Provider, k8sVersion string, amiVersion string) (AMIQuery, error) {
+	query := AMIQuery{
+		Filters: []*ec2.Filter{&ec2.Filter{
+			Name: lo.ToPtr("image-id"),
+		}},
+		KnownRequirements: make(map[string][]scheduling.Requirements),
+	}
+	rootPath := fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2023", k8sVersion)
```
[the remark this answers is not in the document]
jmdeal: This approach minimizes the calls to `GetParametersByPath` by getting the results for all variants (including future GPU variants) in a single call. Since AL2 uses different root paths we need to do multiple calls, but we can get away with a single call here. Personally I think this is the approach we should take when possible.

pkg/providers/amifamily/ami.go
```diff
@@ -47,75 +45,33 @@ type Provider interface {
...
-		amis, err = p.getDefaultAMIs(ctx, nodeClass)
-		if err != nil {
-			return nil, err
-		}
-	} else {
-		amis, err = p.getAMIs(ctx, nodeClass.Spec.AMISelectorTerms)
-		if err != nil {
-			return nil, err
-		}
+	queries, err := p.GetAMIQueries(ctx, nodeClass)
```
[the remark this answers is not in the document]
jmdeal: Yep, this was `FiltersAndOwnersSets`, just with the addition of `KnownRequirements`. I agree `DescribeImageQueries` is probably the better name, I'll update.

pkg/providers/amifamily/windows.go
```diff
@@ -41,17 +46,41 @@ type Windows struct {
...
+	query := AMIQuery{
+		Filters: []*ec2.Filter{&ec2.Filter{
+			Name: lo.ToPtr("image-id"),
+		}},
+		KnownRequirements: make(map[string][]scheduling.Requirements),
+	}
+	// SSM aliases are only maintained for the latest Windows AMI releases
+	if amiVersion != AMIVersionLatest {
+		return AMIQuery{}, fmt.Errorf("discovering AMIs for alias, %q is an invalid version for Windows", amiVersion)
+	}
```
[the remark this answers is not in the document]
jmdeal: That's the reason I left this like this, it's unclear how the SSM parameter would be extended to support pinned AMIs.

pkg/test/nodeclass.go
jmdeal: Fair, shouldn't make a difference. If it causes a bunch of other cascading issues based on assumptions tests have around al2 being the AMI family then I think we should decouple, but if it's straightforward might as well.

pkg/providers/amifamily/al2.go
```diff
@@ -28,47 +34,61 @@ import (
...
+	}
+	for rootPath, variants := range map[string][]Variant{
+		fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2", k8sVersion):       []Variant{VariantStandard},
+		fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2-arm64", k8sVersion): []Variant{VariantStandard},
+		fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2-gpu", k8sVersion):   []Variant{VariantNeuron, VariantNvidia},
+	} {
+		results, err := ssmProvider.List(ctx, rootPath)
+		if err != nil {
+			log.FromContext(ctx).WithValues("path", rootPath).Error(err, "discovering AMIs from ssm")
+			continue
```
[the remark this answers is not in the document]
jmdeal: It's a good question. We should only fail on a subset of paths for a couple of reasons:
- A path is removed (e.g. AL2 no longer publishes GPU AMIs)
- Path formatting has changed, but only for a subset of paths
- We loose API connectivity between calls

In the first two situations, I think the current approach of failing open is correct since the information wouldn't change with a retry (also both are unlikely to change, at least not without warning). IMO the last point could be a good reason to fail closed here. Let's say the API call fails when discovering GPU AMIs. Since we fail open, we'll populate our NodeClass' status with all of the AMIs but the GPU AMI and NodePools referencing that NodeClass will be unable to provision nodes for GPU workloads. This should correct itself in 5 minutes (if the connectivity issue resolves) but we may be able to correct the situation quicker if we retry with exponential backoff on an error.

I'm honestly not sure what the correct approach here would be. My gut is that we should continue to fail open, that way changes to the SSM paths don't cripple this version of Karpenter, requiring **all** users to upgrade. Instead, only users relying on AMIs with the updated paths would need to upgrade to a version of Karpenter that supports the new path.

pkg/providers/amifamily/ami.go
jmdeal: Maybe, I can take a look at adding it as a receiver function on the `AMIQuery` type.

pkg/providers/amifamily/ami.go
jmdeal: If there are no known requirements, this returns a single requirement with the architecture. If there are known requirements, it inserts the architecture requirement to each (via the map) and returns the set. There are definitely different ways to implement this, but this was the cleanest I came up with. This could totally be implemented differently, but using [IIFE](https://en.wikipedia.org/wiki/Immediately_invoked_function_expression) is a personal preference of mine and I have seen it elsewhere in the project. I do think this could fit well as a receiver function on `AMIQuery` though so I'll probably make that change.

pkg/providers/amifamily/ami.go
jmdeal: I think that makes sense, I'll take a look 👍

pkg/providers/amifamily/types.go
jmdeal: Discussed offline, shouldn't ever be reached with current SSM paths.

pkg/providers/amifamily/ami.go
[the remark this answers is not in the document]
jmdeal: Also, I don't think `SSMPaths` would be an appropriate name. By the time we're dealing with the `AMIQuery` type, we've already resolved the ami-id from the SSM path so those details have been discarded.

pkg/providers/amifamily/custom.go
jmdeal: Mentioned this in another comment, but `SSMPaths` doesn't really describe what this function returns. It does deal with SSM paths, but only internally. By the time we're returning details, any information about the SSM paths have been discarded. We're returning an `AMIQuery` which has the `image-id` filter and `KnownRequirements` populated.

pkg/providers/amifamily/al2.go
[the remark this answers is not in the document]
njtran: I agree with your assessment. Let's fail open and log.

pkg/providers/amifamily/al2023.go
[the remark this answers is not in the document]
njtran: I think you're right that this is the best we can do now.

pkg/fake/ssmapi.go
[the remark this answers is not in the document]
njtran: I forgot to remove this comment after I re-acquainted myself with this. Ignore :D

pkg/fake/ssmapi.go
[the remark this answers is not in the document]
njtran: Got it. I'm worried about us missing any newly added ssm paths in here if it's not some sort of shared const/var. What do you think we can do to prevent that from happening?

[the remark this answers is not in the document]
jmdeal: Failing to add those should cause any tests relying on those paths to fail. I think that by itself solves the problem.

pkg/providers/ssm/provider.go
jmdeal: Will still call it out, but I'm going to drop `Get` since it is no longer used in the current version of the PR.

pkg/apis/v1beta1/ec2nodeclass.go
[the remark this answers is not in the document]
jmdeal: I'll go ahead and make this change when I migrate over to the v1 API, leaving this unresolved until I do.

pkg/providers/amifamily/ami.go
jmdeal: Ok, so we also use this in the resolver when generating launch templates and we're really just piggybacking off it for drift. I think it's fine to leave as is.

pkg/providers/amifamily/suite_test.go
jmdeal: Ya, I left this comment because I wasn't sure it would still apply, and it really doesn't because there shouldn't be any way to do this anymore. You can only use the `Custom` family when no alias is specified, but there must be some other selector term.

test/suites/scheduling/suite_test.go
jmdeal: Going to wrap this up tonight, need to check and see if we can switch to using aliases all around or if I need to define some custom userdata for some of these tests. Right now there are CI errors because I haven't updated them from using AMIFamily.

pkg/providers/amifamily/ami.go
```diff
@@ -126,163 +81,128 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1beta1.EC2NodeCl
...
+					reqsHash := lo.Must(hashstructure.Hash(reqs.NodeSelectorRequirements(), hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: true}))
+					if v, ok := images[reqsHash]; ok {
+						candidateCreationTime, _ := time.Parse(time.RFC3339, lo.FromPtr(image.CreationDate))
+						existingCreationTime, _ := time.Parse(time.RFC3339, v.CreationDate)
+						if existingCreationTime == candidateCreationTime && lo.FromPtr(image.Name) < v.Name {
+							continue
+						}
+						if candidateCreationTime.Unix() < existingCreationTime.Unix() {
+							continue
+						}
```
njtran, line 173: Seems like the check should be 
1. Check if time A is less than time B
2. If times are equal, check if name A < name B


I think this is the same, but it confuses because we check the higher priority check second.

njtran (commented):
checkpointing

pkg/fake/ssmapi.go
```diff
@@ -60,6 +66,82 @@ func (a SSMAPI) GetParameterWithContext(_ context.Context, input *ssm.GetParamet
 	}, nil
 }
 
+func (a SSMAPI) GetParametersByPathPagesWithContext(_ context.Context, input *ssm.GetParametersByPathInput, f func(*ssm.GetParametersByPathOutput, bool) bool, _ ...request.Option) error {
+	if !lo.FromPtr(input.Recursive) {
+		panic("fake SSM API currently only supports GetParametersByPathPagesWithContext when recursive is true")
```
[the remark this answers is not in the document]
jmdeal: Oh, we had elsewhere I didn't think about it here. I can make the change.

test/suites/drift/suite_test.go
njtran, line 1: adding these back in?

pkg/providers/amifamily/ami.go
njtran: synced offline, going to keep as is

jmdeal (commented):
/karpenter snapshot

pkg/apis/v1/labels.go
```diff
@@ -121,6 +121,7 @@ var (
 	AnnotationEC2NodeClassHash                = apis.Group + "/ec2nodeclass-hash"
 	AnnotationEC2NodeClassHashVersion         = apis.Group + "/ec2nodeclass-hash-version"
 	AnnotationInstanceTagged                  = apis.Group + "/tagged"
+	AnnotationAMIFamilyCompatibility          = apis.CompatibilityGroup + "/v1beta1-ami-family-conversion"
```
njtran, line 124: Why is this different than the v1beta1 label?

pkg/fake/ssmapi.go
```diff
@@ -29,9 +36,10 @@ import (
 
 type SSMAPI struct {
 	ssmiface.SSMAPI
-	Parameters         map[string]string
-	GetParameterOutput *ssm.GetParameterOutput
-	WantErr            error
+	Parameters                map[string]string
+	GetParameterOutput        *ssm.GetParameterOutput
```
njtran, line 40: Thought we didn't need this?

pkg/providers/amifamily/al2.go
```diff
@@ -15,60 +15,85 @@ limitations under the License.
...
+	// - Latest EKS 1.30 Standard Image: /aws/service/eks/optimized-ami/1.30/amazon-linux-2/recommended/image_id
+	// - Specific EKS 1.30 GPU Image: /aws/service/eks/optimized-ami/1.30/amazon-linux-2-gpu/amazon-eks-node-1.30-v20240625/image_id
+	for rootPath, variants := range map[string][]Variant{
+		fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2", k8sVersion):       []Variant{VariantStandard},
+		fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2-arm64", k8sVersion): []Variant{VariantStandard},
+		fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2-gpu", k8sVersion):   []Variant{VariantNeuron, VariantNvidia},
+	} {
+		results, err := ssmProvider.List(ctx, rootPath)
+		if err != nil {
+			log.FromContext(ctx).WithValues("path", rootPath, "family", "AL2").Error(err, "discovering AMIs from ssm")
```
njtran, line 58: probably should just align with the config too
```suggestion
			log.FromContext(ctx).WithValues("path", rootPath, "family", "al2").Error(err, "discovering AMIs from ssm")
```

pkg/providers/amifamily/types.go
```diff
@@ -0,0 +1,119 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/karpenter/pkg/scheduling"
+
+	v1 "github.com/aws/karpenter-provider-aws/pkg/apis/v1"
+)
+
+const (
+	// AMIVersionLatest is the version used in EKS aliases to represent the latest version. This maps to different
+	// values in the SSM path, depending on the AMI type (e.g. "recommended" for AL2/AL2023)).
+	AMIVersionLatest = "latest"
```
njtran, line 35: @ellistarn I think we should keep this as lowercase. Pinging in case you have an opinion on this enum?

engedaam (commented):
Nice work! just a few comments on the conversion webhooks

test/suites/drift/suite_test.go
```diff
@@ -14,954 +14,925 @@ limitations under the License.
...
-			// Drift the nodeclaims
-			nodePool.Spec.Template.Annotations = map[string]string{"test": "annotation"}
-			env.ExpectUpdated(nodePool)
-
-			env.EventuallyExpectDrifted(nodeClaims...)
-			env.ConsistentlyExpectNoDisruptions(int(numPods), time.Minute)
-		})
-	})
-})
+// import (
```
njtran, line 17: should uncomment this out

pkg/apis/v1/ec2nodeclass_conversion_test.go
```diff
@@ -38,8 +43,18 @@ var _ = Describe("Convert v1 to v1beta1 EC2NodeClass API", func() {
...
+		})
+		Expect(v1ec2nodeclass.ConvertTo(ctx, v1beta1ec2nodeclass)).To(Succeed())
+
+		// Remove the compatibility annotations from the EC2NodeClass.
+		v1beta1ec2nodeclass.ObjectMeta.Annotations = lo.OmitByKeys(v1beta1ec2nodeclass.ObjectMeta.Annotations, []string{
+			v1beta1.AnnotationAMIVersionCompatibility,
+		})
+		if len(v1beta1ec2nodeclass.ObjectMeta.Annotations) == 0 {
+			v1beta1ec2nodeclass.ObjectMeta.Annotations = nil
+		}
```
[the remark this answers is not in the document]
jmdeal: Since this does a strict equality check a nil map does not equal an empty map and the test fails. Really just a quirk of Ginkgo I guess.

pkg/apis/apis.go
```diff
@@ -26,7 +26,8 @@ import (
 
 //go:generate controller-gen crd object:headerFile="../../hack/boilerplate.go.txt" paths="./..." output:crd:artifacts:config=crds
 var (
-	Group = "karpenter.k8s.aws"
+	Group              = "karpenter.k8s.aws"
+	CompatibilityGroup = "compatibility.karpenter.k8s.aws"
```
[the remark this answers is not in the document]
jmdeal: This matches the compatibility group added in the upstream project for the kubelet annotation, happy to change it but seems like they should match.

pkg/apis/v1/labels.go
jmdeal: The annotations store two different pieces of information. The v1beta1 label stores the version, whereas this stores the family. @engedaam pointed out though that we don't need the v1beta1 label since we know the storage version is v1 and the version can't be lost.

pkg/fake/ssmapi.go
jmdeal: You're right, I'll drop it.

pkg/apis/v1beta1/labels.go
```diff
@@ -121,6 +121,7 @@ var (
 	AnnotationEC2NodeClassHash                = apis.Group + "/ec2nodeclass-hash"
 	AnnotationEC2NodeClassHashVersion         = apis.Group + "/ec2nodeclass-hash-version"
 	AnnotationInstanceTagged                  = apis.Group + "/tagged"
+	AnnotationAMIVersionCompatibility         = apis.CompatibilityGroup + "/v1-ami-version-conversion"
```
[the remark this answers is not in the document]
jmdeal: This is the label that's applied to the v1beta1 NodeClass which is why it's here. We discussed this offline though, yet to see if it's in this review, but you're right that we can just drop this label since we know the storage version is v1.

pkg/apis/v1/ec2nodeclass.go
```diff
@@ -163,6 +164,17 @@ type SecurityGroupSelectorTerm struct {
...
 type AMISelectorTerm struct {
+	// Alias specifies which EKS optimized AMI to select.
+	// Each alias consists of a family and a version, specified as "family@version".
+	// Valid families include: al2, al2023, bottlerocket, windows2019, and windows2022.
+	// The version can either be pinned to a specific AMI release, with that AMIs version format (ex: "al2023@v20240625" or "bottlerocket@1.10.0").
+	// The version can also be set to "latest" for any family. Setting the version to latest will result in drift when a new AMI is released.
+	// Note: The Windows families do **not** support version pinning, and only latest may be used.
+	// +kubebuilder:validation:XValidation:message="'alias' is improperly formatted, must match the format 'family@version'",rule="self.matches('^[a-zA-Z0-9]*@.*$')"
+	// +kubebuilder:validation:XValidation:message="family is not supported, must be one of the following: 'al2', 'al2023', 'bottlerocket', 'windows2019', 'windows2022'",rule="self.find('^[^@]+') in ['al2','al2023','bottlerocket','windows2019','windows2022']"
+	// +kubebuilder:validation:MaxLength=30
```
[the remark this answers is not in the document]
jmdeal: Yes, without it we exceed the computation cost limit.

pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -57,24 +57,28 @@ spec:
...
                     description: |-
                       AMISelectorTerm defines selection logic for an ami used by Karpenter to launch nodes.
                       If multiple fields are used for selection, the requirements are ANDed.
                     properties:
+                      alias:
+                        description: |-
+                          Alias specifies which EKS optimized AMI to select.
+                          Each alias consistes of a family and a version, specified as "family@version".
+                          Valid families include: al2, al2023, bottlerocket, windows2019, and windows2022.
+                          The version can either be pinned to a specific AMI release, with that AMIs version format (ex: "al2023@v20240625" or "bottlerocket@1.10.0").
```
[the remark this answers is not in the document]
jmdeal: Updated to include the v. I don't think it's necessary to document we support both, just documenting the v for consistency seems fine to me. I still want to support the version without the v though since that is the version string in the ssm path.

pkg/test/nodeclass.go
jmdeal: Didn't make it into this PR, let's track this as a follow-up.

jmdeal (commented):
/karpenter snapshot

test/suites/drift/suite_test.go
```diff
@@ -95,6 +95,10 @@ var _ = Describe("Drift", func() {
 			},
 		})
 		selector = labels.SelectorFromSet(dep.Spec.Selector.MatchLabels)
+
+		rawContent, err := os.ReadFile("testdata/al2023_userdata_input.yaml")
+		Expect(err).ToNot(HaveOccurred())
+		customUserData = lo.ToPtr(fmt.Sprintf(string(rawContent), env.ClusterName, env.ClusterEndpoint, env.ExpectCABundle()))
```
njtran, line 101: is it overkill to marshal, merge, and unmarshal the userdata to do this rather than a magic ordering of string inputs? happy to do what you think is best here.

jmdeal (commented):
/karpenter snapshot

test/suites/drift/suite_test.go
jmdeal: I agree that seems like the better approach, this is something we do in other tests already though. Makes sense to me to make the change in a follow-up PR across the whole project.

njtran (dismissed):
LGTM! Nice work!!

njtran (approved):

PR #6455, opened 2024-07-06
chore: Add status conditions on nodeClass

jigisha620:
Fixes #N/A 

**Description**

This PR adds status conditions to EC2nodeClass.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/controllers/nodeclass/status/ami.go
+++ b/pkg/controllers/nodeclass/status/ami.go
@@ -41,6 +41,7 @@ func (a *AMI) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (r
 	}
 	if len(amis) == 0 {
 		nodeClass.Status.AMIs = nil
+		nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeAMIsReady, "AMINotFound", "AMISelector did not match any AMIs")
 		return reconcile.Result{}, nil
 	}
 	nodeClass.Status.AMIs = lo.Map(amis, func(ami amifamily.AMI, _ int) v1beta1.AMI {
@@ -60,5 +61,6 @@ func (a *AMI) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (r
 			Requirements: reqs,
 		}
 	})
+	nodeClass.StatusConditions().SetTrue(v1beta1.ConditionTypeAMIsReady)
 	return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 }
--- a/pkg/controllers/nodeclass/status/instanceprofile.go
+++ b/pkg/controllers/nodeclass/status/instanceprofile.go
@@ -39,6 +39,6 @@ func (ip *InstanceProfile) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2
 	} else {
 		nodeClass.Status.InstanceProfile = lo.FromPtr(nodeClass.Spec.InstanceProfile)
 	}
-
+	nodeClass.StatusConditions().SetTrue(v1beta1.ConditionTypeInstanceProfileReady)
 	return reconcile.Result{}, nil
 }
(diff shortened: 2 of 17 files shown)


pkg/controllers/nodeclass/status/ami.go
```diff
@@ -41,6 +41,7 @@ func (a *AMI) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2NodeClass) (r
 	}
 	if len(amis) == 0 {
 		nodeClass.Status.AMIs = nil
+		nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeAMIsReady, "UnresolvedAMIs", "Unable to resolve AMI")
```
ellistarn, line 44:
```suggestion
		nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeAMIsReady, "AMINotFound", "amiSelector did not match any AMIs")
```

pkg/controllers/nodeclass/status/instanceprofile.go
```diff
@@ -33,12 +33,13 @@ func (ip *InstanceProfile) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2
 	if nodeClass.Spec.Role != "" {
 		name, err := ip.instanceProfileProvider.Create(ctx, nodeClass)
 		if err != nil {
+			nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeInstanceProfileReady, "InstanceProfileCreateError", "Error creating instance profile")
```
ellistarn, line 36: This error/write pattern is a bit surprising to me. Do we make an Update NodeClass API call in this error path? I'm not sure we should.

Instead, we may simply want to leave this condition as unknown.

pkg/controllers/nodeclass/status/securitygroup.go
```diff
@@ -39,6 +39,7 @@ func (sg *SecurityGroup) Reconcile(ctx context.Context, nodeClass *v1beta1.EC2No
 	}
 	if len(securityGroups) == 0 && len(nodeClass.Spec.SecurityGroupSelectorTerms) > 0 {
 		nodeClass.Status.SecurityGroups = nil
+		nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeSecurityGroupsReady, "UnresolvedSecurityGroups", "Unable to resolve SecurityGroups")
```
ellistarn, line 42: see comment above re: wording

ellistarn:
Nice work! Can you confirm that all of this is happening out-of-band from the launch path? My understanding is that this is a separate "readiness" controller, and once it's marked as ready, it can be used by the provisioning/consolidation control loops.

pkg/controllers/nodeclass/status/instanceprofile.go
jigisha620: Currently we have setup the nodeClass status controller in a way that it would make an [update nodeClass](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/controllers/nodeclass/status/controller.go#L95) api call if the nodeClass is updated after going through reconcile function of each resource. Even if we run into this error path for instance profile, nodeClass object would be updated by other reconcile functions. @jonathan-innis and I were having discussion earlier about decoupling these controllers but everyone on the team does not have an agreement with that idea. And that's why I am holding that [PR](https://github.com/aws/karpenter-provider-aws/pull/6438). With separate controllers we can prevent the update nodeClass API call.

ellistarn: Agree with this approach for now.

jigisha620:
> Nice work! Can you confirm that all of this is happening out-of-band from the launch path? My understanding is that this is a separate "readiness" controller, and once it's marked as ready, it can be used by the provisioning/consolidation control loops.

Yes, this will happen separately from the launch path. NodeClass readiness will be propagated to nodePool which will make nodePool not ready if nodeClass is not ready and we will skip such nodePools when making scheduling decisions.

ellistarn (approved):

PR #6479, opened 2024-07-09
ci: Fix CI failures for golang-ci

jonathan-innis:
Fixes #N/A 

**Description**

Golang CI was failing in my local environment due to an error return value that wasn't assigned. This should have been caught by `golangci-lint` but wasn't caught right now.

`golangci-lint` does not report errors on newer versions (> `v1.57.2`) when there are type check errors. This PR updates the hack directory so that there are no more type check errors due to multiple `main.go` files in the package which surfaces other errors that need to be fixed.

I'm following up with the `golangci-lint` folks to track-down why the linter is not failing all-of-a-sudden on the new versions: https://github.com/golangci/golangci-lint/issues/4869. There's currently a discussion ongoing to not fail open in scenarios where this type checking fails.

`make presubmit`

- [ ] Yes, issue opened: # 
- [x] No


jmdeal (approved):
LGTM 🚀

PR #6488, opened 2024-07-10
chore: drop validation webhooks and enable conversion webhooks

jmdeal:
Fixes #N/A 

**Description**
This PR does the following:
- Updates RBAC permissions for conversion webhooks
- Passes conversion webhooks to the operator
- Removes validation and defaulting webhooks for the EC2NodeClass
- Adds a hack script to add the conversion stanza to CRDs
- Removes the `webhook.enabled` setting from the helm chart.

In-Progress

- [ ] Yes, issue opened: # 
- [x] No


jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

hack/mutation/conversion_webhooks_injection.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+# Add the conversion stanza to the CRD spec to enable conversion via webhook
+yq eval '.spec.conversion = {"strategy": "Webhook", "webhook": {"conversionReviewVersions": ["v1beta1", "v1"], "clientConfig": {"service": {"name": "karpenter", "namespace": "kube-system", "port": 8443}}}}' -i pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
```
jonathan-innis, line 4: This strikes me as something that I'm surprised that we would need. Does kubebuilder not have a way to specify this with markers?

charts/karpenter/templates/clusterrole-core.yaml
```diff
@@ -65,12 +63,9 @@ rules:
   - apiGroups: [""]
     resources: ["pods/eviction"]
     verbs: ["create"]
-{{- if .Values.webhook.enabled }}
-  - apiGroups: ["admissionregistration.k8s.io"]
-    resources: ["validatingwebhookconfigurations"]
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
```
jonathan-innis, line 67: Confirming: We're confident that all of these permissions are strictly required? Do we know _why_ having these permissions is required? Is it to inject the certificate?

pkg/apis/v1/ec2nodeclass_validation.go
```diff
@@ -1,307 +0,0 @@
...
-	"fmt"
-	"strings"
-
-	"github.com/aws/aws-sdk-go/service/ec2"
-	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
-	"knative.dev/pkg/apis"
-)
-
-const (
```
jonathan-innis, line 28: Confirming: we're 100% confident that all of this validation is captured by our current CEL and that there is no runtime validation mechanism performed on this object?

jmdeal: I didn't have to rip out any in-code references, so I believe the answer is no. The only place that I could tell this was used was for the webhook. There was runtime validation for the NodePool in the upstream repo, and that was pretty trivially caught by looking for references to the validation functions.

charts/karpenter/templates/clusterrole-core.yaml
jmdeal: I can do a sanity check without the permissions, but yes it needs to inject the cert

jmdeal: Did the sanity check, we do need it to update the cert.

hack/mutation/conversion_webhooks_injection.sh
njtran: +1 on this, but also kind of makes sense? kubebuilder could do it by appending the conversionReviewVersions references across annotations, but also not sure if it's a feature that's just not possible

pkg/apis/v1/ec2nodeclass_validation.go
njtran: I think the ask is to make sure that our coverage in this file is at least a subset of our current CEL validation, right?

hack/mutation/conversion_webhooks_injection.sh
jmdeal: I don't think there's any way for it to know of the port or service name at generation time. I think in an ideal world we would template our CRDs as part of our helm chart, but this probably isn't worth the squeeze since these will only be included for a single minor version.

njtran (approved):
LGTM

PR #6493, opened 2024-07-11
fix: scheduling zone-id flake

jmdeal:
Fixes #N/A 

**Description**
Fixes a flake in the zone-id scheduling tests, where we might get the node before the label with the expected zone-id has been propagated by the nodeclaim lifecycle controller.

`make e2etests`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/suites/scheduling/suite_test.go
+++ b/test/suites/scheduling/suite_test.go
@@ -644,7 +644,7 @@ var _ = Describe("Scheduling", Ordered, ContinueOnFailure, func() {
 				},
 			})
 			env.ExpectCreated(nodePool, nodeClass, pod)
-			node := env.EventuallyExpectNodeCount("==", 1)[0]
+			node := env.EventuallyExpectInitializedNodeCount("==", 1)[0]
 			Expect(node.Labels[v1.LabelTopologyZone]).To(Equal(subnetInfo[1].Zone))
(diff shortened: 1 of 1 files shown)


jmdeal (commented):
/karpenter snapshot

rschalo (approved):
/lgtm

test/suites/scheduling/suite_test.go
```diff
@@ -644,7 +644,7 @@ var _ = Describe("Scheduling", Ordered, ContinueOnFailure, func() {
 				},
 			})
 			env.ExpectCreated(nodePool, nodeClass, pod)
-			node := env.EventuallyExpectNodeCount("==", 1)[0]
```
jonathan-innis, line 647: I'm assuming this is because the label won't appear until CCM registers it?

jonathan-innis (approved):

test/suites/scheduling/suite_test.go
jmdeal: The label that I saw failures due to is actually one we add, but its added at registration time, not via user data. This is because it's added via a requirement, not a label.

jmdeal: I imagine we could run into the same issue with the CCM added label though.

PR #6532, opened 2024-07-17
chore: fix nodeClaim tests failure

jigisha620:
Fixes #N/A 

**Description**
NodeClaim tests are failing because the standalone nodeClaims do not reference any nodePool. While adding the kubelet compatibility hash if this reference is empty, we should not throw an error and rather fall back to nodeClass kubelet config.

Tested on local cluster.`make test`

- [ ] Yes, issue opened: # 
- [x] No


engedaam (approved):
LGTM 🚀

PR #6546, opened 2024-07-18
tests: fix expiration test failing due to drift when we modify expireAfter

jigisha620:
Fixes #N/A 

**Description**
ExpireAfter has been now moved to nodeClaim. With this change, if we modify expireAfter, nodeClaim will drift causing test failures. Updated the tests to have expireAfter set to a lower value at the beginning of the test.

Tested on local cluster

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/suites/expiration/suite_test.go
+++ b/test/suites/expiration/suite_test.go
@@ -67,7 +67,6 @@ var _ = Describe("Expiration", func() {
 	var numPods int
 	BeforeEach(func() {
 		numPods = 1
-		// Add pods with a do-not-disrupt annotation so that we can check node metadata before we disrupt
 		dep = coretest.Deployment(coretest.DeploymentOptions{
 			Replicas: int32(numPods),
 			PodOptions: coretest.PodOptions{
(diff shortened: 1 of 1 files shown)


test/suites/expiration/suite_test.go
```diff
@@ -102,12 +98,6 @@ var _ = Describe("Expiration", func() {
 			g.Expect(ok).To(BeTrue())
 		}).Should(Succeed())
 
-		// Set the expireAfter to "Never" to make sure new node isn't deleted
-		// This is CRITICAL since it prevents nodes that are immediately spun up from immediately being expired and
-		// racing at the end of the E2E test, leaking node resources into subsequent tests
-		nodePool.Spec.Template.Spec.ExpireAfter.Duration = nil
```
engedaam, line 108: The more I think about it, we should set the limit on the nodepool to zero after one expiration happens. That way we won't delete our current nodeclaims, but we also won't scale up either. This is important to prevent the case of leaked resources

test/suites/expiration/suite_test.go
```diff
@@ -148,12 +137,6 @@ var _ = Describe("Expiration", func() {
...
```
engedaam, line 154: Same here

engedaam (commented):
Small nit

test/suites/expiration/suite_test.go
```diff
@@ -148,10 +142,9 @@ var _ = Describe("Expiration", func() {
...
+		// Set the limit to 0 to make sure we won't delete our current nodeClaims and do not scale up either.
```
engedaam, line 145:
```suggestion
		// Set the limit to 0 to make sure we won't continue to create nodeclaims.
```

engedaam (dismissed):
Optional nits

engedaam (approved):
LGTM 🚀

PR #6548, opened 2024-07-18
feat: Support empty block device mappings

### LINKED ISSUE #6545: Option to disable setting default blockDeviceMappings for AL2 amiFamily
### Description

We would like to define block device mappings directly in our custom AMI builds derived from EKS optimized AMIs without needing to also specify them in our EC2NodeClass configuration. This would allow us to not have to update block device mappings in multiple places, which will simplify deployment pipelines and prevent provisioner errors due to build-time device sizing being larger than what is specified in an EC2NodeClass.

This is very important as we would prefer not to have to move to the Custom amiFamily and thus lose out on the AL2-specific bootstrap configuration, plus any future modifications specific to AL2 AMIs. Ultimately this will help us ease the migration of many clusters to utilize nodes that have custom Level 2 CIS benchmark hardening requirements. One of the requirements is to create new mount points for various directories bound to a separate EBS device. We want to manage the configuration for this EBS device in our AMI build pipeline.

bmendoza820:
Fixes #6545 

**Description**

Updates the v1 EC2NodeClass to allow for specifying a nil `.spec.blockDeviceMappings` value as a way to opt for defaulting the EBS block device mappings. Otherwise, if an empty array is assigned to the field, no block device mappings are configured and only block device mappings defined in the AMI are used.

I have updated the field to a pointer type and updated its assignment in various places in tests. If this is insufficient test coverage, I can add a new test targeting the intended behavior.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/apis/v1/ec2nodeclass.go
+++ b/pkg/apis/v1/ec2nodeclass.go
@@ -106,7 +106,7 @@ type EC2NodeClassSpec struct {
 	// +kubebuilder:validation:XValidation:message="must have only one blockDeviceMappings with rootVolume",rule="self.filter(x, has(x.rootVolume)?x.rootVolume==true:false).size() <= 1"
 	// +kubebuilder:validation:MaxItems:=50
 	// +optional
-	BlockDeviceMappings []*BlockDeviceMapping `json:"blockDeviceMappings,omitempty"`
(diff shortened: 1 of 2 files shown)


bmendoza820:
Commenting to keep this PR active.

PR #6576, opened 2024-07-23
feat: Drop Env Vars for v1

rschalo:
Fixes #N/A 

**Description**
This change drops references for ENV VARs that will no longer be used in with the v1 launch.

Make test

- [ ] Yes, issue opened: # 
- [x] No


Makefile
```diff
@@ -45,11 +45,10 @@ ci-non-test: verify licenses vulncheck ## Runs checks other than tests
 run: ## Run Karpenter controller binary against your local cluster
 	SYSTEM_NAMESPACE=${KARPENTER_NAMESPACE} \
 		KUBERNETES_MIN_VERSION="1.19.0-0" \
-		LEADER_ELECT=false \
+		DISABLE_LEADER_ELECT=true \
 		DISABLE_WEBHOOK=true \
 		CLUSTER_NAME=${CLUSTER_NAME} \
 		INTERRUPTION_QUEUE=${CLUSTER_NAME} \
-		FEATURE_GATES="Drift=true" \
```
jonathan-innis, line 52: As part of the run, should we enable the feature gate for spotToSpot consolidation? I think that we enable this by default when we deploy the helm chart in the `make apply`

Makefile
```diff
@@ -45,11 +45,10 @@ ci-non-test: verify licenses vulncheck ## Runs checks other than tests
 run: ## Run Karpenter controller binary against your local cluster
 	SYSTEM_NAMESPACE=${KARPENTER_NAMESPACE} \
 		KUBERNETES_MIN_VERSION="1.19.0-0" \
-		LEADER_ELECT=false \
+		DISABLE_LEADER_ELECT=true \
```
jonathan-innis, line 48:
```suggestion
		DISABLE_LEADER_ELECTION=true \
```

I'd rename this to `DISABLE_LEADER_ELECTION` instead of leader elect

Makefile
rschalo: We don't do that in apply today, should we?

rschalo: Nm - it's passed as part of helm opts

jonathan-innis (approved):
LGTM 🚀

PR #6589, opened 2024-07-25
feat: Add eks:eks-cluster-name tag and remove karpenter.sh/managed-by annotation

rschalo:
Fixes #N/A 

**Description**
Dropping `karpenter.sh/managed-by` as it is an artifact of a migration path and not something needed in the future. Adding `eks:eks-cluster-name` to support pod identity sessions.

make test

- [ ] Yes, issue opened: # 
- [x] No


pkg/apis/v1/ec2nodeclass.go
```diff
@@ -448,7 +447,7 @@ func (in *EC2NodeClass) InstanceProfileRole() string {
 func (in *EC2NodeClass) InstanceProfileTags(clusterName string) map[string]string {
 	return lo.Assign(in.Spec.Tags, map[string]string{
 		fmt.Sprintf("kubernetes.io/cluster/%s", clusterName): "owned",
-		karpv1.ManagedByAnnotationKey:                        clusterName,
+		"eks:eks-cluster-name":                               clusterName,
```
jonathan-innis, line 450: Can we use the const here?

pkg/apis/v1/labels.go
```diff
@@ -71,7 +71,7 @@ var (
 		// https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateCluster.html
 		regexp.MustCompile(`^kubernetes\.io/cluster/[0-9A-Za-z][A-Za-z0-9\-_]*$`),
 		regexp.MustCompile(fmt.Sprintf("^%s$", regexp.QuoteMeta(karpv1.NodePoolLabelKey))),
-		regexp.MustCompile(fmt.Sprintf("^%s$", regexp.QuoteMeta(karpv1.ManagedByAnnotationKey))),
+		regexp.MustCompile(fmt.Sprintf("^%s$", regexp.QuoteMeta("eks:eks-cluster-name"))),
```
jonathan-innis, line 74: Can we use the `const` here?

pkg/apis/v1beta1/labels.go
```diff
@@ -95,6 +95,8 @@ var (
 	ResourcePrivateIPv4Address corev1.ResourceName = "vpc.amazonaws.com/PrivateIPv4Address"
 	ResourceEFA                corev1.ResourceName = "vpc.amazonaws.com/efa"
 
+	EKSClusterNameTagKey = "eks:eks-cluster-name"
```
jonathan-innis, line 98: Not sure that we need to add this to v1beta1

pkg/cloudprovider/cloudprovider.go
```diff
@@ -358,9 +358,7 @@ func (c *CloudProvider) instanceToNodeClaim(i *instance.Instance, instanceType *
 	labels[karpv1.CapacityTypeLabelKey] = i.CapacityType
 	if v, ok := i.Tags[karpv1.NodePoolLabelKey]; ok {
 		labels[karpv1.NodePoolLabelKey] = v
-	}
-	if v, ok := i.Tags[karpv1.ManagedByAnnotationKey]; ok {
-		annotations[karpv1.ManagedByAnnotationKey] = v
+		annotations[karpv1.NodePoolLabelKey] = v
```
jonathan-innis, line 361: Why do we need the NodePool in the annotations?

pkg/controllers/nodeclaim/garbagecollection/controller.go
```diff
@@ -62,7 +62,7 @@ func (c *Controller) Reconcile(ctx context.Context) (reconcile.Result, error) {
 		return reconcile.Result{}, fmt.Errorf("listing cloudprovider machines, %w", err)
 	}
 	managedRetrieved := lo.Filter(retrieved, func(nc *karpv1.NodeClaim, _ int) bool {
-		return nc.Annotations[karpv1.ManagedByAnnotationKey] != "" && nc.DeletionTimestamp.IsZero()
+		return nc.Annotations[karpv1.NodePoolLabelKey] != "" && nc.DeletionTimestamp.IsZero()
```
jonathan-innis, line 65: Are there cases where we get an instance back on a List() that isn't owned by Karpenter? Just thinking about it at a high-level, I think this check was here from when we did the migration, but I think now we would only return instances that were owned by Karpeneter anyways -- so this check should basically be a no-op

rschalo: Wouldn't we want the `nc.DeletionTimestamp.IsZero()` to be present so we don't attempt to garbage collect multiple times?

pkg/cloudprovider/cloudprovider.go
rschalo: Annotation was to one for one swap with the `managed-by` one and to satisfy the filter in `garbagecollection/controller.go` but if it is no longer used then we can drop it.

pkg/apis/v1beta1/ec2nodeclass.go
```diff
@@ -78,9 +78,9 @@ type EC2NodeClassSpec struct {
 	InstanceProfile *string `json:"instanceProfile,omitempty"`
 	// Tags to be applied on ec2 resources like instances and launch templates.
 	// +kubebuilder:validation:XValidation:message="empty tag keys aren't supported",rule="self.all(k, k != '')"
+	// +kubebuilder:validation:XValidation:message="tag contains a restricted tag matching eks:eks-cluster-name",rule="self.all(k, k !='eks:eks-cluster-name')"
```
jonathan-innis, line 81: Should we change this logic in beta? I think this should only apply to the new API

pkg/apis/v1beta1/ec2nodeclass_validation_cel_test.go
```diff
@@ -79,10 +79,6 @@ var _ = Describe("CEL/Validation", func() {
 				"kubernetes.io/cluster/test": "value",
 			}
 			Expect(env.Client.Create(ctx, nc)).To(Not(Succeed()))
-			nc.Spec.Tags = map[string]string{
```
jonathan-innis, line 82: Why did we drop this from v1beta1 API?

test/pkg/environment/common/environment.go
```diff
@@ -113,10 +113,10 @@ func NewClient(ctx context.Context, config *rest.Config) client.Client {
 		node := o.(*corev1.Node)
 		return []string{strconv.FormatBool(node.Spec.Unschedulable)}
 	}))
-	lo.Must0(cache.IndexField(ctx, &corev1.Node{}, "spec.taints[*].karpenter.sh/disruption", func(o client.Object) []string {
+	lo.Must0(cache.IndexField(ctx, &corev1.Node{}, "spec.taints[*].karpenter.sh/disrupted", func(o client.Object) []string {
```
jonathan-innis, line 116: This is fine to add here to fix things, but I'm wondering if we can decouple into a separate PR since this isn't related

rschalo: Since the testing fails and we already have it here - I think it's fine to leave in.

njtran (approved):
LGTM

pkg/controllers/nodeclaim/garbagecollection/suite_test.go
```diff
@@ -290,9 +290,9 @@ var _ = Describe("GarbageCollection", func() {
 		Expect(err).NotTo(HaveOccurred())
 	})
 	It("should not delete an instance if it was not launched by a NodeClaim", func() {
-		// Remove the "karpenter.sh/managed-by" tag (this isn't launched by a machine)
+		// Remove the "karpv1.NodePoolLabelKey" tag (this isn't launched by a machine)
```
njtran, line 293: should this be adjusted?
```suggestion
		// Remove the "karpenter.sh/nodepool" tag (this isn't launched by a machine)
```

njtran (approved):
LGTM

PR #6603, opened 2024-07-29
test: fix scale testing flakes

jigisha620:
Fixes #N/A 

**Description**
There is a flakiness in provisioning scale tests where we want the created nodeclaim and node count to be equal to an expected value. We can skip this assertion and instead only check if all the pods were successfully provisioned which make sure that our underlying assumption is satisfied as well.

Tested on local cluster

- [ ] Yes, issue opened: # 
- [x] No


jonathan-innis (approved):
Do we have similar expectations elsewhere in this testing that we should look at removing?

jigisha620:
> Do we have similar expectations elsewhere in this testing that we should look at removing?

We do have similar expectations elsewhere. But it is for node dense scale up. I talked to @engedaam about this and we think that we should not remove this check anywhere else.

jonathan-innis (approved):
LGTM 🚀

PR #6626, opened 2024-08-01
chore: Add post install hook

engedaam:
Fixes #N/A 

**Description**
- Adding a post install hook to allow user to configure conversion webhooks on the CRDs on installation.  

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter/templates/post-install-hook.yaml
+++ b/charts/karpenter/templates/post-install-hook.yaml
@@ -1,0 +1,42 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: helm-{{ .Release.Name }}-post-install-hook
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "karpenter.labels" . | nindent 4 }}
+  {{- with .Values.additionalAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": hook-succeeded
+    "helm.sh/hook-delete-policy": hook-failed
+spec:
+  ttlSecondsAfterFinished: 0
+  template:
+    spec:
+      serviceAccountName: {{ include "karpenter.serviceAccountName" . }}
+      restartPolicy: OnFailure
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+      - name: post-install-job
+        image: bitnami/kubectl:latest
+        command:
+        - /bin/sh
+        - -c
+        - |
+          if  "{{ .Values.webhook.enabled }}" == "true"; then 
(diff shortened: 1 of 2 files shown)


charts/karpenter/templates/post-install-hook.yaml
```diff
@@ -0,0 +1,38 @@
...
+    {{- include "karpenter.labels" . | nindent 4 }}
+  {{- with .Values.additionalAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": hook-succeeded
+    "helm.sh/hook-delete-policy": hook-failed
+spec:
+  ttlSecondsAfterFinished: 10
```
njtran, line 16: Is there a reason we care about if it's up after it finished? If we just leave it alone, the user can clean it up when they want to, or it'll just get garbage collected eventually.

charts/karpenter/templates/post-install-hook.yaml
```diff
@@ -0,0 +1,38 @@
...
+            kubectl patch customresourcedefinitions nodeclaims.karpenter.sh  --type='merge' -p '{"spec":{"conversion":{"webhook":{"clientConfig":{"service":{"name":"{{ include "karpenter.fullname" . }}", "port": {{ .Values.webhook.port }} ,"namespace": "{{ .Release.Namespace }}"}}}}}}'
+            kubectl patch customresourcedefinitions ec2nodeclasses.karpenter.k8s.aws  --type='merge' -p '{"spec":{"conversion":{"webhook":{"clientConfig":{"service":{"name":"{{ include "karpenter.fullname" . }}", "port": {{ .Values.webhook.port }} ,"namespace": "{{ .Release.Namespace }}"}}}}}}'
+          else
+            echo "disabled webhooks"
+            kubectl patch customresourcedefinitions nodepools.karpenter.sh  --type='json' -p '[{'op': 'remove', 'path': '/spec/conversion'}]'
+            kubectl patch customresourcedefinitions nodeclaims.karpenter.sh  --type='json' -p '[{'op': 'remove', 'path': '/spec/conversion'}]'
+            kubectl patch customresourcedefinitions ec2nodeclasses.karpenter.k8s.aws  --type='json' -p '[{'op': 'remove', 'path': '/spec/conversion'}]'
+          fi
```
njtran, line 37: This would be in the case where someone had webhooks disabled before migrating to v1 and wanted to migrate back? If this conversion block doesn't exist, they don't run into any issues, right?

charts/karpenter/templates/post-install-hook.yaml
engedaam: From my testing it will not get garbage collected, unless we set the TTL. I can a use case for debug, but if succeed   I think we should remove the pod from the cluster

charts/karpenter/templates/post-install-hook.yaml
engedaam: Yeah, this solves the problem of user needing to remove the block

engedaam: They can run the disable helm and that will remove conversion block

charts/karpenter/templates/post-install-hook.yaml
njtran: Seems like we can just set it to 0 then

njtran (approved):

PR #6628, opened 2024-08-01
test: add lastPodEventTime e2es

njtran:
Fixes #N/A 

**Description**
Adds e2e tests to verify that lastPodEvent is set at the correct times

FOCUS="LastPodEventTime" make e2etests

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/suites/consolidation/suite_test.go
+++ b/test/suites/consolidation/suite_test.go
@@ -66,7 +66,7 @@ var _ = BeforeEach(func() {
 var _ = AfterEach(func() { env.Cleanup() })
 var _ = AfterEach(func() { env.AfterEach() })
 
-var _ = FDescribe("Consolidation", func() {
+var _ = Describe("Consolidation", func() {
 	Context("LastPodEventTime", func() {
 		var nodePool *karpv1.NodePool
 		BeforeEach(func() {
@@ -131,9 +131,8 @@ var _ = FDescribe("Consolidation", func() {
 			podLabels := map[string]string{"app": "regular-app"}
 			pod := test.Pod(test.PodOptions{
 				// use a non-pause image so that we can have a sleep
-				Image:        "alpine:3.20.2",
-				Command:      []string{"/bin/sh", "-c", "sleep 30"},
-				PreStopSleep: lo.ToPtr(int64(30)),
+				Image:   "alpine:3.20.2",
+				Command: []string{"/bin/sh", "-c", "sleep 30"},
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: podLabels,
 				},
--- a/test/suites/consolidation/suite_test.go
+++ b/test/suites/consolidation/suite_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/awslabs/operatorpkg/object"
 	"github.com/samber/lo"
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


njtran (commented):
/karpenter snapshot

test/suites/consolidation/suite_test.go
```diff
@@ -65,7 +66,119 @@ var _ = BeforeEach(func() {
 var _ = AfterEach(func() { env.Cleanup() })
 var _ = AfterEach(func() { env.AfterEach() })
 
-var _ = Describe("Consolidation", func() {
+var _ = FDescribe("Consolidation", func() {
```
jonathan-innis, line 69:
```suggestion
var _ = Describe("Consolidation", func() {
```

test/suites/consolidation/suite_test.go
```diff
@@ -65,7 +66,119 @@ var _ = BeforeEach(func() {
...
+
+			// wait 10 seconds so that we don't run into the de-dupe timeout
+			time.Sleep(10 * time.Second)
+
+			dep.Spec.Replicas = lo.ToPtr[int32](4)
+			By("removing one pod from the node")
+			env.ExpectUpdated(dep)
+
+			Eventually(func(g Gomega) {
+				nodeClaim = env.ExpectExists(nodeClaim).(*karpv1.NodeClaim)
```
jonathan-innis, line 113: Why do you need to do this casting? Isn't this function typed so it shouldn't need an interface cast?

test/suites/consolidation/suite_test.go
```diff
@@ -65,7 +66,119 @@ var _ = BeforeEach(func() {
...
+				g.Expect(nodeClaim.Status.LastPodEventTime.Time).ToNot(BeEquivalentTo(lastPodEventTime.Time))
+			}).WithTimeout(5 * time.Second).WithPolling(1 * time.Second).Should(Succeed())
+		})
+		It("should update lastPodEventTime when pods go terminal", func() {
+			podLabels := map[string]string{"app": "regular-app"}
+			pod := test.Pod(test.PodOptions{
+				// use a non-pause image so that we can have a sleep
+				Image:        "alpine:3.20.2",
+				Command:      []string{"/bin/sh", "-c", "sleep 30"},
+				PreStopSleep: lo.ToPtr(int64(30)),
```
jonathan-innis, line 136: Assuming that we use the preStopSleep to hold the pod around and block it from being marked as succeeded?

jonathan-innis: If that's the case, should we add a comment here to indicate that?

test/suites/consolidation/suite_test.go
```diff
@@ -65,7 +66,119 @@ var _ = BeforeEach(func() {
...
+			env.ExpectCreated(nodeClass, nodePool, job)
+
+			nodeClaims := env.EventuallyExpectCreatedNodeClaimCount("==", 1)
+			env.EventuallyExpectCreatedNodeCount("==", 1)
+			pods := env.EventuallyExpectHealthyPodCount(selector, int(1))
+
+			nodeClaim := env.ExpectExists(nodeClaims[0]).(*karpv1.NodeClaim)
+			lastPodEventTime := nodeClaim.Status.LastPodEventTime
+
+			Eventually(func(g Gomega) {
```
jonathan-innis, line 172: What's the mechanism that we are using here to make sure that we have enough time between the pod creation and the pod completion?

test/suites/consolidation/suite_test.go
jonathan-innis: Actually -- thinking on it more -- should we be using preStopSleep with a job to handle this or should the image of the job itself handle this sleep so that we don't have to handle the preStop hooks?

njtran: I think we actually probably don't need the prestop

test/suites/consolidation/suite_test.go
njtran: I'm not sure I understand. The pod itself is sleeping for 30 seconds, and the timeout is one minute. Isn't that enough? We also only start the timer after its gone healthy, which means it's started running.

test/suites/consolidation/suite_test.go
njtran: It doesn't use generics, uses client.Object, which means it just implements an interface iiuc

jonathan-innis: Seems trivial to change it to use generics?

test/suites/consolidation/suite_test.go
jonathan-innis: I was mainly just trying to grok that and confirm since it's not immediately clear to me from reading through it at a high-level. Makes sense to me. Some in-code documentation might help here but not blocking

jonathan-innis (approved):

test/suites/consolidation/suite_test.go
jonathan-innis: Discussed offline: Can't do generics with member functions so we're pushing this out right now

jonathan-innis (approved):
LGTM 🚀

PR #6630, opened 2024-08-01
chore: Add templated Values for the CRDs

engedaam:
Fixes #N/A 

**Description**
- Added webhook values to allow users to configure webhooks inside of the conversion block

- Manually tested

- [ ] Yes, issue opened: # 
- [x] No


hack/mutation/conversion_webhooks_injection.sh
```diff
@@ -4,3 +4,51 @@
 yq eval '.spec.conversion = {"strategy": "Webhook", "webhook": {"conversionReviewVersions": ["v1beta1", "v1"], "clientConfig": {"service": {"name": "karpenter", "namespace": "kube-system", "port": 8443}}}}' -i pkg/apis/crds/karpenter.k8s.aws_ec2nodeclasses.yaml
 yq eval '.spec.conversion = {"strategy": "Webhook", "webhook": {"conversionReviewVersions": ["v1beta1", "v1"], "clientConfig": {"service": {"name": "karpenter", "namespace": "kube-system", "port": 8443}}}}' -i pkg/apis/crds/karpenter.sh_nodeclaims.yaml
 yq eval '.spec.conversion = {"strategy": "Webhook", "webhook": {"conversionReviewVersions": ["v1beta1", "v1"], "clientConfig": {"service": {"name": "karpenter", "namespace": "kube-system", "port": 8443}}}}' -i pkg/apis/crds/karpenter.sh_nodepools.yaml
+
+# Update to the karpetner-crd charts
```
njtran, line 8:
```suggestion
# Update to the karpenter-crd charts
```

njtran (dismissed):
LGTM

charts/karpenter-crd/values.yaml
```diff
@@ -0,0 +1,7 @@
+webhook:
+  # -- Whether to enable the webhooks and webhook permissions.
+  enabled: true
+  serviceName: karpenter
+  serviceNamespace: kube-system
```
jonathan-innis, line 5: Isn't this just the release namespace? Is there a need to make the namespace configurable?

njtran: These are the default values, right? So this is where someone would change the value to whatever they want.

engedaam: I was thinking that user could install the karpenter-crd in a different namespace then the karpenter deployment

jonathan-innis: Probably not likely, but it probably doesn't matter a ton. I'm fine with it as-is.

jonathan-innis (approved):
LGTM 🚀

engedaam (commented):
/karpenter snapshot

engedaam (commented):
/karpenter snapshot

PR #6636, opened 2024-08-02
chore: Include post upgrade on Helm Hook

engedaam:
Fixes #N/A 

**Description**
- Support upgrade and rollback with Karpenter with the helm chart for the CRDs
- Updated the image to use `public.ecr.aws/bitnami/kubectl:1.30`

- Manually tested

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter/templates/post-install-hook.yaml
+++ b/charts/karpenter/templates/post-install-hook.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     {{- include "karpenter.labels" . | nindent 4 }}
   annotations:
-    "helm.sh/hook": post-install
+    "helm.sh/hook": post-install,post-upgrade,post-rollback
     "helm.sh/hook-delete-policy": hook-succeeded
     "helm.sh/hook-delete-policy": hook-failed
     {{- with .Values.additionalAnnotations }}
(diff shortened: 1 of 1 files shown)


charts/karpenter/templates/post-install-hook.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
     {{- include "karpenter.labels" . | nindent 4 }}
   annotations:
-    "helm.sh/hook": post-install
+    "helm.sh/hook": post-install,post-upgrade,post-rollback
```
jonathan-innis, line 9: This feels dangerous to me. Users installing this chart won't actually get management of their CRDs when they do the upgrade. Helm doesn't upgrade CRDs going from version to version, so I think you only need this on the installation path

jonathan-innis: Discussed offline: After playing with it for a bit, I realized that a user that is going to be executing this hook is going to be going through the process of kubectl applying anyways. By adding the hook in here on rollback and upgrade, this is going to make it so that a user doesn't even have to think about the webhook port or namespace when applying the karpenter helm chart.

jonathan-innis (approved):
LGTM 🚀

PR #6651, opened 2024-08-05
feat: add option to keep the crds with an uninstall

### LINKED ISSUE #6649: add option to karpenter helm chart to set custom annotations on the custom resource definitions
### Description
add an option to make it possible to set annotations on the custom resource definitions, so we can set in our use case the helm keep tag to ensure these never get deleted by an accidential uninstall of the karpenter-crd chart
```
  annotations:
    helm.sh/resource-policy: keep
```

we would like to be able to set cutom annotations to the crds, in our case a the specific helm keep tag to ensure the crds are preserved by an accidental uninstall of the chart and that would prevent node interuptions

martijnvdp:
Fixes #6649 

**Description**
add option to set the `helm.sh/resource-policy: keep` to the crds so it prevents deletion of those crds with an uninstall or redeployment of karpenter
to prevent node disruptions on a redeploy or removal of the karpenter crds

deleted ec2nodeclass crd > deleted nodepool > node disruptions
 
- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter-crd/templates/karpenter.k8s.aws_ec2nodeclasses.yaml
+++ b/charts/karpenter-crd/templates/karpenter.k8s.aws_ec2nodeclasses.yaml
@@ -4,6 +4,9 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.15.0
+{{- if .Values.crds.keep }}
+    helm.sh/resource-policy: keep
+{{- end }}
   name: ec2nodeclasses.karpenter.k8s.aws
 spec:
   group: karpenter.k8s.aws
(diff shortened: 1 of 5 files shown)


charts/karpenter-crd/templates/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -4,6 +4,9 @@ kind: CustomResourceDefinition
 metadata:
```
jonathan-innis, line 4: Why uninstall the `karpenter-crd` chart if you don't want the CRDs to be removed? Is there another reason that you would be removing the chart?

martijnvdp: it is more if it gets uninstalled by accident, like wehn deploying the helm chart with argocd and the argocd deployment gets deleted

martijnvdp: i see othet charts like cert-manager also having an option to set this helm keep annotation to the crds so

PR #6653, opened 2024-08-05
chore: enforce Windows version via CEL

jmdeal:
Fixes #N/A 

**Description**
When using a Windows alias, the only version that may be specified is 'latest'. This PR moves this check from runtime to validation time via a CEL check.

`make presubmit` and `/karpenter snapshot`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter-crd/templates/karpenter.k8s.aws_ec2nodeclasses.yaml
+++ b/charts/karpenter-crd/templates/karpenter.k8s.aws_ec2nodeclasses.yaml
@@ -95,6 +95,8 @@ spec:
                             rule: self.matches('^[a-zA-Z0-9]*@.*$')
                           - message: 'family is not supported, must be one of the following: ''al2'', ''al2023'', ''bottlerocket'', ''windows2019'', ''windows2022'''
                             rule: self.find('^[^@]+') in ['al2','al2023','bottlerocket','windows2019','windows2022']
+                          - message: windows families may only specify version 'latest'
+                            rule: 'self.find(''^[^@]+'') in [''windows2019'',''windows2022''] ? self.split(''@'')[1] == ''latest'': true'
                       id:
                         description: ID is the ami id in EC2
                         pattern: ami-[0-9a-z]+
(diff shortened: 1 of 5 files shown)


jmdeal (commented):
/karpenter snapshot

charts/karpenter-crd/templates/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -95,6 +95,8 @@ spec:
                             rule: self.matches('^[a-zA-Z0-9]*@.*$')
                           - message: 'family is not supported, must be one of the following: ''al2'', ''al2023'', ''bottlerocket'', ''windows2019'', ''windows2022'''
                             rule: self.find('^[^@]+') in ['al2','al2023','bottlerocket','windows2019','windows2022']
+                          - message: windows families may only specify version 'latest'
```
jonathan-innis, line 98: Is this guaranteed to have something after `@`. If I look at the regex above, I can see that I could end the value with `@` which I think would cause this validation to fail with a panic

jmdeal: Looks like a mistake in that regex, it really should be `^[a-zA-Z0-9]+@.+$`. I'll update.

pkg/apis/v1/ec2nodeclass_validation_cel_test.go
```diff
@@ -604,6 +604,15 @@ var _ = Describe("CEL/Validation", func() {
 			nc.Spec.AMISelectorTerms = []v1.AMISelectorTerm{{Alias: "ubuntu@latest"}}
 			Expect(env.Client.Create(ctx, nc)).ToNot(Succeed())
 		})
+		DescribeTable(
```
jonathan-innis, line 607: Can we add a test that would have caught the regex failure that we caught in this PR?

jmdeal: It wasn't really a regex failure, calling it a mistake was a mistake. The mistake was really in this PR, assuming version already had a non-zero length. For non-windows AMIs, version validation is really entirely done at runtime.

jonathan-innis: Still seems like we should add a test, no?

jmdeal: Sure, it makes sense to add a test to ensure this validation invariant holds. I thought you meant there was some other existing failure we should add a test for. I've added it.

jonathan-innis (approved):
LGTM 🚀

PR #6654, opened 2024-08-05
fix: Use NodeClaim annotations instead of NodeClass 

engedaam:
Fixes #N/A 

**Description**
- Carry existing and Add NodeClaim annotations as part of the NodeClaim Create call

- `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No


pkg/cloudprovider/suite_test.go
```diff
@@ -249,6 +249,21 @@ var _ = Describe("CloudProvider", func() {
...
+		nodeClaim.Annotations = map[string]string{
+			"test-key-1": "test-value-1",
+			"test-key-2": "test-value-2",
+			"test-key-3": "test-value-3",
+		}
+		ExpectApplied(ctx, env.Client, nodePool, nodeClass, nodeClaim)
+		cloudProviderNodeClaim, err := cloudProvider.Create(ctx, nodeClaim)
+		Expect(err).To(BeNil())
+		Expect(cloudProviderNodeClaim).ToNot(BeNil())
+		Expect(cloudProviderNodeClaim.ObjectMeta.Annotations["test-key-1"]).To(BeTrue())
```
jonathan-innis, line 262: Is BeTrue() correct here? Why not `HaveKeyWithValue()`?

engedaam: Removed and Updated the test

pkg/cloudprovider/suite_test.go
```diff
@@ -249,6 +249,14 @@ var _ = Describe("CloudProvider", func() {
 		Expect(ok).To(BeTrue())
 		Expect(zoneID).To(Equal(subnet.ZoneID))
 	})
+	It("should expect a strict set of annotation keys", func() {
+		ExpectApplied(ctx, env.Client, nodePool, nodeClass, nodeClaim)
+		cloudProviderNodeClaim, err := cloudProvider.Create(ctx, nodeClaim)
+		Expect(err).To(BeNil())
+		Expect(cloudProviderNodeClaim).ToNot(BeNil())
+		Expect(len(lo.Keys(cloudProviderNodeClaim.Annotations))).To(BeNumerically("==", 3))
+		Expect(lo.Keys(cloudProviderNodeClaim.Annotations)).To(ContainElements(v1.AnnotationKubeletCompatibilityHash, v1.AnnotationEC2NodeClassHash, v1.AnnotationEC2NodeClassHashVersion))
```
jonathan-innis, line 258: Does this actually test anything? Since this would still have passed on the old version of the code. I honestly liked the test that you had before this change better

jonathan-innis: Discussed offline. The `karpenter.sh/managed-by` annotation was dropped, so this is the best that we can do on `main`. We will have to slightly modify this check when we patch it back on the old minor versions

jonathan-innis (approved):
LGTM 🚀

PR #6704, opened 2024-08-12
test: deflake termination grace period tests

njtran:
Fixes #N/A 

**Description**
Deflakes termination Grace Period e2e tests that were failing, and speeds them up by directly deleting the nodeclaim rather than waiting for 90s expiration

make e2etests

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/pkg/environment/common/expectations.go
+++ b/test/pkg/environment/common/expectations.go
@@ -279,6 +279,25 @@ func (env *Environment) EventuallyExpectHealthy(pods ...*corev1.Pod) {
 	env.EventuallyExpectHealthyWithTimeout(-1, pods...)
 }
 
+func (env *Environment) EventuallyExpectTerminating(pods ...*corev1.Pod) {
+	GinkgoHelper()
+	env.EventuallyExpectTerminatingWithTimeout(-1, pods...)
+}
+
+func (env *Environment) EventuallyExpectTerminatingWithTimeout(timeout time.Duration, pods ...*corev1.Pod) {
+	GinkgoHelper()
+	Eventually(func(g Gomega) {
+		for _, pod := range pods {
+			g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(pod), pod)).To(Succeed())
+			g.Expect(pod.Status.Phase).To(Equal(corev1.PodRunning))
+			g.Expect(pod.Status.Conditions).To(ContainElement(And(
+				HaveField("Type", Equal(corev1.DisruptionTarget)),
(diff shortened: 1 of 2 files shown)


njtran (commented):
/karpenter snapshot

njtran (commented):
/karpenter snapshot

test/pkg/environment/common/expectations.go
```diff
@@ -279,6 +279,25 @@ func (env *Environment) EventuallyExpectHealthy(pods ...*corev1.Pod) {
...
+}
+
+func (env *Environment) EventuallyExpectTerminatingWithTimeout(timeout time.Duration, pods ...*corev1.Pod) {
+	GinkgoHelper()
+	Eventually(func(g Gomega) {
+		for _, pod := range pods {
+			g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(pod), pod)).To(Succeed())
+			g.Expect(pod.Status.Phase).To(Equal(corev1.PodRunning))
+			g.Expect(pod.Status.Conditions).To(ContainElement(And(
+				HaveField("Type", Equal(corev1.DisruptionTarget)),
```
jmdeal, line 294: Should we use something other than the disruption condition since this doesn't capture pods which were terminated via deletion? Maybe the disruption target or a non-zero deletion timestamp?

test/suites/termination/termination_grace_period_test.go
```diff
@@ -55,19 +54,28 @@ var _ = Describe("TerminationGracePeriod", func() {
 				return karpv1.IsDisruptingTaint(t)
 			})
 			g.Expect(ok).To(BeTrue())
-		}).Should(Succeed())
+		}).WithTimeout(3 * time.Second).WithPolling(100 * time.Millisecond).Should(Succeed())
+
+		// Check that pod remains healthy until termination grace period
+		// subtract the polling time of the eventually above to reduce any races.
```
jmdeal, line 60: It looks like this currently assumes we'll be successful within a single polling iteration, I'm wondering how often that's true. Could we rely on the TGP annotation instead to determine how long the pod should remain healthy?

test/pkg/environment/common/expectations.go
njtran: non-zero deletion timestamp is probably better, good call

test/suites/termination/termination_grace_period_test.go
jmdeal: Ah got it, makes sense. SGTM

jmdeal (approved):
LGTM 🚀

PR #6727, opened 2024-08-13
chore: Limit parallelism on our GitHub action for version compatibility 

engedaam:
Fixes #N/A 

**Description**
- Limit the number of suites that run as part of our version compatibility testing. Prior to this change, the version compatibility workflow would create 3 suite (using [2 suites from our matrix job](https://github.com/aws/karpenter-provider-aws/blob/c7201b0a5c41befaf13ebe9bc84290878ac01dae/.github/workflows/e2e-matrix.yaml#L60) and upgrade suite). This PR will do two things: first reduce the number tests that are running at any given time from 18 suites to 12 suites, and reorder the test suites to reduce churn from similar operation correlating in one suite. 
- **Reducing the test suites from 18 -> 12:** Reducing the test runs to have at most 12 test suite running at a time. This will be to reduce the amount of transient errors due to throttling or issues caused by high request volume inside the testing account. 
- Re-order test suites: As both the upgrade and integration suite run the same set of tests(also happen to be the most aggressive), re-ordering the matrix will result the test suites to run in a staggered from each other to reduce the churn impact to the account. (should have no impact to the normal integration testing)

- As these changes are to the workflow, we will need to merge in the PR to test

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/.github/workflows/e2e-matrix.yaml
+++ b/.github/workflows/e2e-matrix.yaml
@@ -59,13 +59,15 @@ jobs:
       max-parallel: ${{ inputs.parallelism || 100 }}
       matrix:
         suite:
-          - name: Integration
+          - name: IPv6
             region: ${{ inputs.region }}
           - name: AMI
             region: ${{ inputs.region }}
           - name: Scheduling
             region: ${{ inputs.region }}
           - name: Storage
+            region: ${{ inputs.region }}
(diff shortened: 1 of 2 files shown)


jonathan-innis (approved):
LGTM 🚀

Issue #6765, opened 2024-08-15
Post Install Hook doesn't ever sync via ArgoCD


danielloader:
### Description

**Observed Behavior**:

Due to https://github.com/aws/karpenter-provider-aws/blob/a70356ac2224e93d09372994a790fbdcad9cbc0f/charts/karpenter/templates/post-install-hook.yaml#L15 in conjunction with this issue https://github.com/argoproj/argo-cd/issues/6880#issuecomment-2274495939 means the hook never syncs and installation via ArgoCD isn't possible.

**Expected Behavior**:

Chart syncs via ArgoCD installation.

**Reproduction Steps** (Please include YAML):

Install the karpenter chart via ArgoCD and default values.

**Versions**:
- Chart Version: 1.0.0
- Kubernetes Version (`kubectl version`): 1.30.0

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

Vinaum8:
I installed karpenter in a separate namespace and with a different service name than the default.

**I can't configure the values ​​for the webhook** and it returns the error:

`service: karpenter not found.`

However, **it is obvious** that I did not create the service with this name in the kube-system namespace.

`I am using version 0.37.1 and I installed karpenter with ArgoCD.`

inclemenstv:
I had the same bug when I tried to install Karpenter using Argo cd
Chart Version: 1.0.0

snieg:
> I had the same bug when I tried to install Karpenter using Argo cd Chart Version: 1.0.0

Same here. As a temporary solution, I had to edit each CRD and change the namespace to the proper one.

![image](https://github.com/user-attachments/assets/503abfd8-cd73-4220-8f2a-4b187361d14d)

inclemenstv:
> > I had the same bug when I tried to install Karpenter using Argo cd Chart Version: 1.0.0
> 
> Same here. As a temporary solution, I had to edit each CRD and change the namespace to the proper one.
> 
[...]

PR #6766, opened 2024-08-15
chore: add configurable ttl to post install hook

danielloader:
Fixes https://github.com/aws/karpenter-provider-aws/issues/6765

**Description**

Added configuration option for the ttlSecondsAfterFinished field on the post install job so that Argo CD based installs can complete (as Argo CD needs to reap the job rather than the internal controller).

Forked the chart, installed it with Argo CD with a value of 30 for `.Values.postInstallHook.ttlSecondsAfterFinished`.

- [ ] Yes, issue opened: # 
- [x] No


inclemenstv:
I had the same issues, please fix it

charts/karpenter/templates/post-install-hook.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
       {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
-  ttlSecondsAfterFinished: 0
+  ttlSecondsAfterFinished: {{ .Values.postInstallHook.ttlSecondsAfterFinished }}
```
engedaam, line 15: This mainly looks good, but I think we need to drop the annotation for `hook-succeeded` and `hook-failed` for the `helm.sh/hook-delete-policy`: https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies This will ensure that we are drive the clean-up process as part of the  `ttlSecondsAfterFinished`

danielloader: Makes sense to move these annotations out to a .Values.postInstallHook.annotations map then?

engedaam: Yeah, if user want that level of configuration that would make sense

danielloader:
Closing as the job has been removed.

Issue #5135, opened 2023-11-21
Helm - Add NodePool and EC2NodeClass


mtavaresmedeiros:
### Description

**What problem are you trying to solve?**
Allow specifying NodePool and Ec2NodeClass
**How important is this feature to you?**
It will simplify the installation of the karpenter with helm, allowing specifying a node pool and ec2nodeclass in the initial installation.

Ex: Install the karpenter with the terraform in a Kubernetes cluster with fargate(serverless) or another option with 0 nodes, and install the GitOps tools after it, without the need to create another resource (NodePool and Ec2NodeClass), and has all things work out-of-the-box.

If it is fine, I can do the PR. 

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jonathan-innis:
We had a similar request here: #1956. We are generally recommending that users build a "top-level chart" that take this chart as a dependency elsewhere since we are generally trying to avoid the burden of having to maintain the chart ourselves. given the number of options that we would have to support.

mtavaresmedeiros:
hum... https://github.com/aws/karpenter/pull/1956#issuecomment-1163402365, this commentary really makes sense. closing.

younsl:
@mtavaresmedeiros In my production cluster, I am using my own [karpenter-nodepool](https://github.com/younsl/younsl.github.io/tree/main/content/charts/karpenter-nodepool) chart, integrated into the karpenter chart as a child chart.

[Subchart dependencies](https://helm.sh/docs/helm/helm_dependency/) declared in the `Chart.yaml` of the `karpenter` chart:

```bash
# charts/karpenter/Chart.yaml
dependencies:
  - name: karpenter-nodepool
    version: 0.0.7
    repository: "file://./charts/karpenter-nodepool"
    condition: karpenter-nodepool.enabled
```

Directory structure for `karpenter` chart, including `karpenter-nodepool` chart as subchart:

```bash
karpenter  # parent chart
├── Chart.lock
├── Chart.yaml
├── README.md
├── charts
│   └── karpenter-nodepool  # subchart
├── crds
├── templates
└── values.yaml
```

PR #3018, opened 2022-12-11
feat: add CRD chart to be later used as a dependency

spring1843:
Fixes # 

**Description**

Add CRD chart that publishes to GHCR

**How was this change tested?**

*

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #6791, opened 2024-08-17
feat: add support for extraManifests to create NodePools and EC2NodeClasses during Helm Install

msfidelis:
This PR introduces a new feature to the Karpenter Helm chart that allows users to specify custom extraManifests within the values.yaml file. This enables the creation of NodePools and EC2NodeClasses directly during the Helm apply process.

The feature is useful for users who need to set up specific configurations for their clusters, such as defining instance types, capacity types, and network settings for their nodes, without having to apply these manifests separately.

Fixes #N/A 

**Description**

The change was tested by:

Deploying the Helm chart with extraManifests configured in the values.yaml file.
Verifying that the specified NodePools and EC2NodeClasses were correctly created and applied during the Helm apply process.

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter/templates/extra-manifests.yaml
+++ b/charts/karpenter/templates/extra-manifests.yaml
@@ -1,0 +1,5 @@
+{{- range .Values.extraManifests }}
+{{- $manifest := tpl . $ }}
+{{ $manifest | nindent 0 }}
+---
+{{- end }}
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -194,3 +194,57 @@ settings:
     # -- spotToSpotConsolidation is ALPHA and is disabled by default.
     # Setting this to true will enable spot replacement consolidation for both single and multi-node consolidation.
     spotToSpotConsolidation: false
+extraManifests: []
+  # - |
+  #   apiVersion: karpenter.k8s.aws/v1beta1
+  #   kind: EC2NodeClass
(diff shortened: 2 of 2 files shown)


jonathan-innis:
We've had similar asks in #1956, #3018, and #5135. We've generally taken the stance that this is best managed by a different chart rather than leaking the implementation of the configuration objects into the chart.

msfidelis:
Thank you for the feedback!

I understand the concern about leaking the implementation of configuration objects into the management chart, and I agree that it can become chaotic, especially with a large number of CRDs.

However, I believe that for charts with a limited set of CRDs, having the Helm chart assist in creating them can actually enhance productivity by reducing the number of steps required to deploy a fully configured environment. 

It streamlines the process and makes it easier for users to get started with their deployments without needing separate management for these objects, like an Terraform Pipelines. 

That said, I recognize the potential drawbacks of this approach. What do you think about replacing the extraManifests with dedicated fields for the specific CRDs, like `ec2NodeClasses: []`, `nodePools: []`, etc.? 

This way, we could keep the chart’s structure organized while still providing the convenience of managing these resources directly through Helm.

I would appreciate your thoughts on this approach. I really believe in the potential of this feature. 

> We've had similar asks in #1956, #3018, and #5135. We've generally taken the stance that this is best managed by a different chart rather than leaking the implementation of the configuration objects into the management chart.

PR #6809, opened 2024-08-20
chore: Bump go.mod to use go 1.23

jonathan-innis:
Fixes #N/A 

**Description**

This updates Karpenter Provider to use the latest version of go, `go1.23`

`make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/charts/karpenter-crd/templates/karpenter.k8s.aws_ec2nodeclasses.yaml
+++ b/charts/karpenter-crd/templates/karpenter.k8s.aws_ec2nodeclasses.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.15.0
+    controller-gen.kubebuilder.io/version: v0.16.1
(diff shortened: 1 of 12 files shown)


charts/karpenter-crd/templates/karpenter.k8s.aws_ec2nodeclasses.yaml
```diff
@@ -940,6 +913,9 @@ spec:
                           RootVolume is a flag indicating if this device is mounted as kubelet root dir. You can
                           configure at most one root volume in BlockDeviceMappings.
                         type: boolean
+                    required:
+                      - deviceName
+                      - ebs
```
jmdeal, line 918: I know this came from the new kubebuilder version, but this stands out as a bit of a red flag to me. Have we verified what the consequences of this change are?

jonathan-innis: Good catch -- updated so we don't change the fields that are required here -- though the API may be invalid anyways without those fields being required so we may want to validate that in a subsequent PR

jonathan-innis (commented):
/karpenter snapshot

jmdeal (approved):
LGTM 🚀

PR #6810, opened 2024-08-20
docs: detail how to discover alias versions

jmdeal:
Fixes #N/A 

**Description**
Add details for how to discover available alias versions for AL2, AL2023, and Bottlerocket.

N/A

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/website/content/en/docs/concepts/nodeclasses.md
+++ b/website/content/en/docs/concepts/nodeclasses.md
@@ -56,7 +56,8 @@ spec:
     imageGCLowThresholdPercent: 80
     cpuCFSQuota: true
     clusterDNS: ["10.0.1.100"]
-  # Required, resolves a default ami and userdata
+  # Optional, dictates UserData generation and default block device mappings.
+  # May be ommited when using an `alias` amiSelectorTerm, otherwise required.
   amiFamily: AL2
 
   # Required, discovers subnets to attach to instances
--- a/website/content/en/docs/concepts/nodeclasses.md
+++ b/website/content/en/docs/concepts/nodeclasses.md
@@ -754,6 +754,12 @@ The following commands can be used to determine the versions availble for an ali
   {{% /tab %}}
 {{< /tabpane >}}
 
+{{% alert title="Warning" color="warning" %}}
+Karpenter supports automatic AMI selection and upgrades using the `latest` version pin, but this is **not** recommended for production environments.
+When using `latest`, a new AMI release will cause Karpenter to drift all out-of-date nodes in the cluster, replacing them with nodes running the new AMI.
+We strongly recommend evaluating new AMIs in a lower environment before rolling them out into a production environment.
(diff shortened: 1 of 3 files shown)


website/content/en/docs/concepts/nodeclasses.md
```diff
@@ -731,6 +731,27 @@ alias: bottlerocket@v1.20.4
 ```
 The Windows family does not support pinning, so only `latest` is supported.
 
+The following commands can be used to determine the versions availble for an alias in your region:
+
+{{< tabpane text=true right=false >}}
+  {{% tab "AL2023" %}}
+  ```bash
+  aws ssm get-parameters-by-path --path "/aws/service/eks/optimized-ami/$K8S_VERSION/amazon-linux-2023/" --recursive | jq -c '.Parameters[].Name' | tr -d '"' | grep -v "recommended" | awk -F '/' '{print $10}' | sed -r 's/.*(v[[:digit:]]+)$/\1/' | sort | uniq
```
jonathan-innis, line 739:
```suggestion
  aws ssm get-parameters-by-path --path "/aws/service/eks/optimized-ami/$K8S_VERSION/amazon-linux-2023/" --recursive | jq -c '.Parameters[].Name' -r | grep -v "recommended" | awk -F '/' '{print $10}' | sed -r 's/.*(v[[:digit:]]+)$/\1/' | sort | uniq
```

You can just get the raw output from `jq`

website/content/en/docs/concepts/nodeclasses.md
```diff
@@ -731,6 +731,27 @@ alias: bottlerocket@v1.20.4
 ```
 The Windows family does not support pinning, so only `latest` is supported.
 
+The following commands can be used to determine the versions availble for an alias in your region:
```
jonathan-innis, line 734: Should we ask folks to define K8S_VERSION and use the latest like we do in the geting started guide by default so that the command is straight copy-able

jmdeal: Ah good point, I'll add that 👍

website/content/en/preview/concepts/nodeclasses.md
```diff
@@ -731,6 +731,29 @@ alias: bottlerocket@v1.20.4
 ```
 The Windows family does not support pinning, so only `latest` is supported.
 
+The following commands can be used to determine the versions availble for an alias in your region:
+
+{{< tabpane text=true right=false >}}
+  {{% tab "AL2023" %}}
+  ```bash
+  export K8S_VERSION=1.30
```
jonathan-innis, line 739: Can we reference this as `"{{< param "latest_release_version" >}}"` so that this gets auto-updated whenever we increase this version in the future?

jonathan-innis (approved):
LGTM 🚀

PR #6848, opened 2024-08-22
Fix: add log path options for log paths to deployment

rschalo:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/1572 

**Description**
Bug fix for not referencing log output paths.

make test

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,11 +14,11 @@ require (
 	github.com/imdario/mergo v0.3.16
 	github.com/jonathan-innis/aws-sdk-go-prometheus v0.1.1-0.20240804232425-54c8227e0bab
 	github.com/mitchellh/hashstructure/v2 v2.0.2
-	github.com/onsi/ginkgo/v2 v2.20.0
+	github.com/onsi/ginkgo/v2 v2.20.1
 	github.com/onsi/gomega v1.34.1
 	github.com/patrickmn/go-cache v2.1.0+incompatible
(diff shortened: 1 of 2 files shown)


mariuskimmina (approved):
We were about to open the same PR, we tested this exact change on one of our test clusters and it seems to work.

charts/karpenter/templates/deployment.yaml
```diff
@@ -87,6 +87,14 @@ spec:
           {{- with .Values.logLevel }}
             - name: LOG_LEVEL
               value: "{{ . }}"
+          {{- end }}
+          {{- with .Values.logOutputPaths }}
+            - name: LOG_OUTPUT_PATHS
+              value: "{{ . }}"
```
jonathan-innis, line 93: Does this work? Since it's an array, don't we need some helm magic to convert this into a comma-delimited list?

joshuabaird: Yes, it works.  I believe that happens here:  https://github.com/aws/karpenter-provider-aws/blob/8071346bfe6064d83af8a80023bbcefdbe1196e9/charts/karpenter/templates/_helpers.tpl#L145

rschalo:
Fix implemented in a different PR: https://github.com/aws/karpenter-provider-aws/pull/6875

PR #6875, opened 2024-08-26
Fix log paths

rschalo:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/1572 

**Description**
Bug fix for not referencing log output paths.

make test

- [ ] Yes, issue opened: # 
- [x] No


charts/karpenter/templates/deployment.yaml
```diff
@@ -88,6 +88,10 @@ spec:
             - name: LOG_LEVEL
               value: "{{ . }}"
           {{- end }}
+            - name: LOG_OUTPUT_PATHS
+              value: {{ include "karpenter.outputPathsList" . }}
+            - name: LOG_ERROR_OUTPUT_PATHS
+              value: {{ include "karpenter.errorOutputPathsList" . }}
```
jonathan-innis, line 94: Can we just call the join here if we can just do this without a helper?

rschalo: Yep, I'll make the fix.

jonathan-innis (approved):
LGTM 🚀

PR #6894, opened 2024-08-28
feat: allow setting service annotations via helm chart

jamesduffy:
Fixes #N/A 

**Description**
Allows setting annotations on the karpenter service. We need this so that we can add annotations to setup datadog to monitor karpenter. There are probably other use cases.

Generated a helm chart locally with the additional values:

```yaml
service:
  annotations:
    ad.datadoghq.com/service.checks: |
      {
        "karpenter": {
          "cluster_check": true,
          "init_config": {},
          "instances": [
            {
              "openmetrics_endpoint": "http://%%host%%:8000/metrics",
              "exclude_metrics": [
                "^karpenter\\.go\\..*$"
              ]
            }
          ]
        }
      }
```

- [ ] Yes, issue opened: # 
- [x] No


jmdeal:
What's preventing you from using the existing `additionalAnnotations` field? Unless absolutely necessary, we'd like to avoid additional configuration in the helm chart.

jamesduffy:
> What's preventing you from using the existing `additionalAnnotations` field? Unless absolutely necessary, we'd like to avoid additional configuration in the helm chart.

`additionalAnnotations` adds the annotations to a lot of different things. This would be picked up from the Datadog agent as separate checks and instead of running just one check for Karpenter it would run duplicate checks for each place `additionalAnnotations` is injected. Therefore, we would need a way to add the annotation in a single place. Datadog annotations are often on the service.

jamesduffy:
@jmdeal I would like to move forward with this if my explanation makes sense.

jmdeal:
Sorry this has slipped, could you elaborate on what the consequences of this are?

> This would be picked up from the Datadog agent as separate checks and instead of running just one check for Karpenter it would run duplicate checks for each place additionalAnnotations is injected.

Unless this is strictly blocking, and no other workarounds exist, we don't want to expand the configuration surface for the helm chart we ship.

jamesduffy:
> Sorry this has slipped, could you elaborate on what the consequences of this are?
> 
> > This would be picked up from the Datadog agent as separate checks and instead of running just one check for Karpenter it would run duplicate checks for each place additionalAnnotations is injected.
> 
> Unless this is strictly blocking, and no other workarounds exist, we don't want to expand the configuration surface for the helm chart we ship. 

The datadog agent watches for annotations to configure it's checks/metric collection. Using the existing additionalAnnotations would create annotations on more than one resource. This would configure the datadog agent to run multiple collections of the same metrics making it unusable.

jmdeal:
Got it, I assume for most resources this would be a no-op, but you can get duplicate metrics since it would scrape both the Karpenter pods and the service?

jamesduffy:
Correct.

jmdeal (approved):
Discussed offline with other team members, the use-case seems strong enough. LGTM 🚀

PR #6976, opened 2024-09-10
fix: Correct the AMI selector label requirements for NVIDIA and Neuron variants

bryantbiggs:
Fixes #N/A 

**Description**

- Correct the AMI selector label requirements for NVIDIA and Neuron variants
  - Instances backed with NVIDIA GPUs should have the label `karpenter.sh/instance-gpu-count`
  - Instances backed with Neuron devices should have the label `karpenter.sh/instance-accelerator-count`

- `make test`

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/controllers/nodeclass/status/ami_test.go
+++ b/pkg/controllers/nodeclass/status/ami_test.go
@@ -59,32 +59,62 @@ var _ = Describe("NodeClass AMI Status Controller", func() {
 		awsEnv.EC2API.DescribeImagesOutput.Set(&ec2.DescribeImagesOutput{
 			Images: []*ec2.Image{
 				{
-					Name:         aws.String("test-ami-1"),
-					ImageId:      aws.String("ami-test1"),
+					Name:         aws.String("amd64-standard"),
+					ImageId:      aws.String("ami-amd64-standard"),
 					CreationDate: aws.String(time.Now().Format(time.RFC3339)),
 					Architecture: aws.String("x86_64"),
 					Tags: []*ec2.Tag{
-						{Key: aws.String("Name"), Value: aws.String("test-ami-1")},
+						{Key: aws.String("Name"), Value: aws.String("amd64-standard")},
 						{Key: aws.String("foo"), Value: aws.String("bar")},
 					},
 				},
 				{
-					Name:         aws.String("test-ami-2"),
-					ImageId:      aws.String("ami-test2"),
+					Name:         aws.String("amd64-standard-new"),
+					ImageId:      aws.String("ami-amd64-standard-new"),
 					CreationDate: aws.String(time.Now().Add(time.Minute).Format(time.RFC3339)),
 					Architecture: aws.String("x86_64"),
 					Tags: []*ec2.Tag{
-						{Key: aws.String("Name"), Value: aws.String("test-ami-2")},
+						{Key: aws.String("Name"), Value: aws.String("amd64-standard")},
 						{Key: aws.String("foo"), Value: aws.String("bar")},
 					},
 				},
 				{
-					Name:         aws.String("test-ami-3"),
-					ImageId:      aws.String("ami-test3"),
-					CreationDate: aws.String(time.Now().Add(2 * time.Minute).Format(time.RFC3339)),
+					Name:         aws.String("amd64-nvidia"),
+					ImageId:      aws.String("ami-amd64-nvidia"),
+					CreationDate: aws.String(time.Now().Format(time.RFC3339)),
 					Architecture: aws.String("x86_64"),
 					Tags: []*ec2.Tag{
-						{Key: aws.String("Name"), Value: aws.String("test-ami-3")},
+						{Key: aws.String("Name"), Value: aws.String("amd64-nvidia")},
+						{Key: aws.String("foo"), Value: aws.String("bar")},
+					},
+				},
+				{
+					Name:         aws.String("amd64-neuron"),
+					ImageId:      aws.String("ami-amd64-neuron"),
+					CreationDate: aws.String(time.Now().Format(time.RFC3339)),
+					Architecture: aws.String("x86_64"),
+					Tags: []*ec2.Tag{
+						{Key: aws.String("Name"), Value: aws.String("amd64-neuron")},
+						{Key: aws.String("foo"), Value: aws.String("bar")},
+					},
+				},
+				{
+					Name:         aws.String("arm64-standard"),
+					ImageId:      aws.String("ami-arm64-standard"),
+					CreationDate: aws.String(time.Now().Format(time.RFC3339)),
+					Architecture: aws.String("arm64"),
+					Tags: []*ec2.Tag{
+						{Key: aws.String("Name"), Value: aws.String("arm64-standard")},
+						{Key: aws.String("foo"), Value: aws.String("bar")},
+					},
+				},
+				{
+					Name:         aws.String("arm64-nvidia"),
+					ImageId:      aws.String("ami-arm64-nvidia"),
+					CreationDate: aws.String(time.Now().Format(time.RFC3339)),
+					Architecture: aws.String("arm64"),
+					Tags: []*ec2.Tag{
+						{Key: aws.String("Name"), Value: aws.String("arm64-nvidia")},
 						{Key: aws.String("foo"), Value: aws.String("bar")},
 					},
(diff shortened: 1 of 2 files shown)


jmdeal (approved):
LGTM 🚀 

Looks like this slipped through because our currently supported AMIs don't differentiate between Nvidia and Neuron, so this swap doesn't change the end result. I'd be happy to merge and take a test for that case as a follow-up (separate Nvidia and Neuron AMIs, same AMI family), but also happy to take one here if you wanted to add one.

bryantbiggs:
I'm dreadfully awful with go 😅 - give me 24 hours and let me see if I can add in tests for this

bryantbiggs:
@jmdeal I think this one is outside of my wheelhouse - I tried a naive approach to check that the `scheduling.Requirements` included the appropriate label, but wasn't able to make it work. I'm afraid this one is outside of my wheelhouse 😅

jmdeal:
Sorry for the delay, made a couple additional changes and this should be good to go from my perspective. Since I've touched the PR though I shouldn't be the one to approve, I'm going to get review from someone else on the team. We should be able to merge this shortly

pkg/controllers/nodeclass/status/ami_test.go
```diff
@@ -89,133 +91,426 @@ var _ = Describe("NodeClass AMI Status Controller", func() {
...
+						},
+					},
+				},
+			}))
+			Expect(nodeClass.StatusConditions().IsTrue(v1.ConditionTypeAMIsReady)).To(BeTrue())
+		})
 	})
-	It("should resolve amiSelector AMis and requirements into status when all SSM aliases don't resolve", func() {
-		version := lo.Must(awsEnv.VersionProvider.Get(ctx))
+	It("should resolve amiSelector AMis and requirements into status when all SSM parameters don't resolve", func() {
```
njtran, line 509: nit
```suggestion
	It("should resolve amiSelector AMIs and requirements into status when all SSM parameters don't resolve", func() {
```

pkg/controllers/nodeclass/status/ami_test.go
```diff
@@ -89,133 +91,426 @@ var _ = Describe("NodeClass AMI Status Controller", func() {
...
+					Name: "amd64-nvidia",
+					ID:   "ami-amd64-nvidia",
+					Requirements: []corev1.NodeSelectorRequirement{
+						{
+							Key:      corev1.LabelArchStable,
+							Operator: corev1.NodeSelectorOpIn,
+							Values:   []string{karpv1.ArchitectureAmd64},
+						},
+						{
+							Key:      v1.LabelInstanceGPUCount,
```
njtran, line 195: do we need a Accelerator NotExists in Accelerator here?

pkg/controllers/nodeclass/status/ami_test.go
```diff
@@ -89,133 +91,426 @@ var _ = Describe("NodeClass AMI Status Controller", func() {
...
+					Name: "amd64-neuron",
+					ID:   "ami-amd64-neuron",
+					Requirements: []corev1.NodeSelectorRequirement{
+						{
+							Key:      corev1.LabelArchStable,
+							Operator: corev1.NodeSelectorOpIn,
+							Values:   []string{karpv1.ArchitectureAmd64},
+						},
+						{
+							Key:      v1.LabelInstanceAcceleratorCount,
```
njtran, line 210: same with do we need a NotInGPU here too?

pkg/controllers/nodeclass/status/ami_test.go
```diff
@@ -89,133 +91,426 @@ var _ = Describe("NodeClass AMI Status Controller", func() {
...
+				fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2-gpu/recommended/image_id", k8sVersion):   "ami-amd64-nvidia",
+				fmt.Sprintf("/aws/service/eks/optimized-ami/%s/amazon-linux-2-arm64/recommended/image_id", k8sVersion): "ami-arm64-standard",
+			}
+			nodeClass.Spec.AMISelectorTerms = []v1.AMISelectorTerm{{Alias: "al2@latest"}}
+			ExpectApplied(ctx, env.Client, nodeClass)
+			ExpectObjectReconciled(ctx, env.Client, statusController, nodeClass)
+			nodeClass = ExpectExists(ctx, env.Client, nodeClass)
+
+			Expect(len(nodeClass.Status.AMIs)).To(Equal(4))
+			Expect(nodeClass.Status.AMIs).To(ContainElements([]v1.AMI{
```
njtran, line 249: Do you really need to redefine it? Can't you just use the AMIs that you defined up above in a variable and reference it? Saves on code lines.

pkg/controllers/nodeclass/status/ami_test.go
```diff
@@ -89,133 +91,426 @@ var _ = Describe("NodeClass AMI Status Controller", func() {
...
+						{
+							Key:      v1.LabelInstanceGPUCount,
+							Operator: corev1.NodeSelectorOpExists,
+						},
 					},
-					{
-						Key:      v1.LabelInstanceGPUCount,
-						Operator: corev1.NodeSelectorOpDoesNotExist,
+				},
+				// Note: Bottlerocket uses the same AMI for nvidia and neuron, we use the nvidia AMI here
```
njtran, line 404: so this was only an issue with al2023, yes?

bryantbiggs: looks like this note has a typo `Bottlerocket` -> `AL2`

jmdeal: The only accelerated AMI vended by Bottlerocket is the nvidia variant so this wasn't a typo. We've historically selected the nvidia AMI for both nvidia and inferentium / trainium instances but looking at this [issue](https://github.com/bottlerocket-os/bottlerocket/issues/1995) I'm wondering if that was incorrect.

But yes, the only AMI family we currently support that will have separate AMIs for gpu and accelerator workloads is AL2023.

bryantbiggs: > We've historically selected the nvidia AMI for both nvidia and inferentium / trainium instances but looking at this https://github.com/bottlerocket-os/bottlerocket/issues/1995 I'm wondering if that was incorrect.

Thats correct - only the EKS AL2 GPU AMI can be used on instances with NVIDIA GPUs and AWS Neuron devices. The Bottlerocket NVIDIA variant will only work on NVIDIA GPUs; it won't work on AWS Neuron devices (though that variant is in the works)

pkg/controllers/nodeclass/status/ami_test.go
jmdeal: It wouldn't be incorrect, but I don't think it's necessary. AFAIK the two labels are mutually exclusive; the existence of one implies the absence of the other.

pkg/controllers/nodeclass/status/ami_test.go
jmdeal: When you say up above, are you referring to the expected AMIs in the status conditions from AL2023? In that particular example they're different since AL2023 has different AMIs for neuron and nvidia. Even if they were the same, I don't think we want to couple them in the test since it's really just happenstance. I.e. if the AMIs available for one family changed, that doesn't mean the AMIs for another would change in kind.

pkg/controllers/nodeclass/status/ami_test.go
bryantbiggs: > do you know if this AMI is strictly incompatible? 

Yes - you can only use the Bottlerocket NVIDIA variant on instances with NVIDIA GPUs and vice versa with Neuron devices *if you want to use those GPUs/accelerators. And no, nothing can fix this in user data - the NVIDIA driver and Neuron driver are baked into the OS which is immutable once created

> Either way, I think we can take that as a follow-up rather than changing that here.

Yes - I will defer to you all in terms of how to best tackle this

jmdeal: Discussed with the team today, we'll plan on making this change at v1.1.0. While it shouldn't affect any users, and it is addressing a bug, it's still a breaking change so I'd like to all it out as such. Since we're going to be backporting this PR to 1.0, we'll track that change in a separate PR.

njtran (approved):

PR #6992, opened 2024-09-12
chore: Fix e2e test matrix script syntax

bryantbiggs:
Fixes #N/A 
image

**Description**
- Fix e2e test matrix script syntax

You can reproduce locally with the following:

Current, throws error:
```sh
#!/bin/bash

if (( "$PRIVATE_CLUSTER" == 'true' )); then
  echo "Private cluster is enabled"
else
  echo "Private cluster is disabled"
fi
```

Fixed, no error:
```sh
#!/bin/bash

if (( "$PRIVATE_CLUSTER" == "true" )); then
  echo "Private cluster is enabled"
else
  echo "Private cluster is disabled"
fi
```

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/hack/e2e_scripts/install_karpenter.sh
+++ b/test/hack/e2e_scripts/install_karpenter.sh
@@ -2,7 +2,7 @@ aws eks update-kubeconfig --name "$CLUSTER_NAME"
 
 CHART="oci://$ECR_ACCOUNT_ID.dkr.ecr.$ECR_REGION.amazonaws.com/karpenter/snapshot/karpenter"
 ADDITIONAL_FLAGS=""
-if (( "$PRIVATE_CLUSTER" == 'true' )); then
+if (( "$PRIVATE_CLUSTER" == "true" )); then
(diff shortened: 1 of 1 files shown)


test/hack/e2e_scripts/install_karpenter.sh
```diff
@@ -32,3 +32,4 @@ helm upgrade --install karpenter "${CHART}" \
   --set "serviceMonitor.endpointConfig.relabelings[3].targetLabel=commitsAfterTag" \
   --set "serviceMonitor.endpointConfig.relabelings[3].replacement=\"$(git describe --tags | cut -d '-' -f 2)\"" \
   --wait
+  --timeout 8m0s
```
bryantbiggs, line 35: default is `5m0s` - this could be a red herring and things are failing to install so potentially the timeout is just failing for longer

jonathan-innis: Yeah, I think our builds were just broken right now. Should be fixed now so if you run again, I think that should fix it.

jonathan-innis (approved):
LGTM 🚀

PR #7004, opened 2024-09-12
WIP: Add memory overhead tracking per nodeclass and ami family

jukie:
Fixes #N/A 

**Description**

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/pkg/controllers/instancetypes/memoryoverhead/controller.go
+++ b/pkg/controllers/instancetypes/memoryoverhead/controller.go
@@ -1,0 +1,91 @@
+package memoryoverhead
+
+import (
+	"context"
+	"fmt"
+	v1 "github.com/aws/karpenter-provider-aws/pkg/apis/v1"
+	"github.com/aws/karpenter-provider-aws/pkg/providers/instancetype"
+	"time"
+
+	"github.com/awslabs/operatorpkg/singleton"
+	"github.com/samber/lo"
+	corev1 "k8s.io/api/core/v1"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+
+	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+)
+
+const (
+	fastRequeueThreshold = 20
+	fastRequeueDuration  = 20 * time.Second
+	slowRequeueDuration  = 10 * time.Minute
+)
+
+type Controller struct {
+	kubeClient      client.Client
+	successfulCount uint64 // keeps track of successful reconciles for more aggressive requeueing near the start of the controller
+}
+
+func NewController(kubeClient client.Client) *Controller {
+	return &Controller{
+		kubeClient:      kubeClient,
+		successfulCount: 0,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "instancetypes.memoryoverhead")
+
+	nodeClassList := &v1.EC2NodeClassList{}
+	if err := c.kubeClient.List(ctx, nodeClassList); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	nodeClaimList := &karpv1.NodeClaimList{}
+	if err := c.kubeClient.List(ctx, nodeClaimList); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	nodeList := &corev1.NodeList{}
+	if err := c.kubeClient.List(ctx, nodeList); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	nodeMap := make(map[string]*corev1.Node, len(nodeList.Items))
+	for _, node := range nodeList.Items {
+		nodeMap[node.Name] = &node
+	}
+
+	for _, nodeClass := range nodeClassList.Items {
+
+		// Iterate over NodeClaims to find matching instance types
+		for _, nodeClaim := range nodeClaimList.Items {
+			instType := nodeClaim.Labels[corev1.LabelInstanceTypeStable]
+
+			// Build cache key from combination of NodeClass and instance type
+			key := fmt.Sprintf("%d-%s", nodeClass.Hash(), instType)
+
+			node, exists := nodeMap[nodeClaim.Status.NodeName]
+			if !exists {
+				continue
+			}
+
+			// Store the memory capacity in the map
+			memoryOverheadMib := node.Status.Capacity.Memory().Value() / 1024 / 1024
+			instancetype.MemoryOverheadMebibytes.SetDefault(key, memoryOverheadMib)
+		}
+	}
+	c.successfulCount++
+	return reconcile.Result{RequeueAfter: lo.Ternary(c.successfulCount <= fastRequeueThreshold, fastRequeueDuration, slowRequeueDuration)}, nil
+}
+
+func (c *Controller) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		Named("instancetypes.memoryoverhead").
+		WatchesRawSource(singleton.Source()).
+		Complete(singleton.AsReconciler(c))
+}
--- a/pkg/providers/instancetype/instancetype.go
+++ b/pkg/providers/instancetype/instancetype.go
@@ -165,7 +165,7 @@ func (p *DefaultProvider) List(ctx context.Context, kc *v1.KubeletConfiguration,
 		// so that Karpenter is able to cache the set of InstanceTypes based on values that alter the set of instance types
 		// !!! Important !!!
 		return NewInstanceType(ctx, i, p.region,
-			nodeClass.Spec.BlockDeviceMappings, nodeClass.Spec.InstanceStorePolicy,
+			nodeClass.Hash(), nodeClass.Spec.BlockDeviceMappings, nodeClass.Spec.InstanceStorePolicy,
 			kc.MaxPods, kc.PodsPerCore, kc.KubeReserved, kc.SystemReserved, kc.EvictionHard, kc.EvictionSoft,
 			amiFamily, p.createOfferings(ctx, i, allZones, p.instanceTypeOfferings[aws.StringValue(i.InstanceType)], nodeClass.Status.Subnets),
 		)
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -16,6 +16,7 @@ package controllers
 
 import (
 	"context"
+	"github.com/aws/karpenter-provider-aws/pkg/controllers/instancetypes/memoryoverhead"
 
 	"github.com/awslabs/operatorpkg/controller"
 	"github.com/awslabs/operatorpkg/status"
@@ -67,6 +68,7 @@ func NewControllers(ctx context.Context, mgr manager.Manager, sess *session.Sess
 		controllerspricing.NewController(pricingProvider),
 		controllersinstancetype.NewController(instanceTypeProvider),
 		status.NewController[*v1.EC2NodeClass](kubeClient, mgr.GetEventRecorderFor("karpenter")),
+		memoryoverhead.NewController(kubeClient),
 	}
 	if options.FromContext(ctx).InterruptionQueue != "" {
 		sqsapi := servicesqs.New(sess)
--- a/pkg/providers/instancetype/types.go
+++ b/pkg/providers/instancetype/types.go
@@ -17,10 +17,12 @@ package instancetype
 import (
 	"context"
 	"fmt"
+	"github.com/patrickmn/go-cache"
 	"math"
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -40,16 +42,18 @@ import (
 )
 
 const (
-	MemoryAvailable = "memory.available"
-	NodeFSAvailable = "nodefs.available"
+	MemoryAvailable            = "memory.available"
+	NodeFSAvailable            = "nodefs.available"
+	MemoryOverheadMebibytesTTL = 1 * time.Hour
 )
 
 var (
-	instanceTypeScheme = regexp.MustCompile(`(^[a-z]+)(\-[0-9]+tb)?([0-9]+).*\.`)
+	instanceTypeScheme      = regexp.MustCompile(`(^[a-z]+)(\-[0-9]+tb)?([0-9]+).*\.`)
+	MemoryOverheadMebibytes = cache.New(MemoryOverheadMebibytesTTL, 30*time.Minute)
 )
 
 func NewInstanceType(ctx context.Context, info *ec2.InstanceTypeInfo, region string,
-	blockDeviceMappings []*v1.BlockDeviceMapping, instanceStorePolicy *v1.InstanceStorePolicy, maxPods *int32, podsPerCore *int32,
+	ncHash string, blockDeviceMappings []*v1.BlockDeviceMapping, instanceStorePolicy *v1.InstanceStorePolicy, maxPods *int32, podsPerCore *int32,
 	kubeReserved map[string]string, systemReserved map[string]string, evictionHard map[string]string, evictionSoft map[string]string,
 	amiFamily amifamily.AMIFamily, offerings cloudprovider.Offerings) *cloudprovider.InstanceType {
 
@@ -57,11 +61,11 @@ func NewInstanceType(ctx context.Context, info *ec2.InstanceTypeInfo, region str
 		Name:         aws.StringValue(info.InstanceType),
 		Requirements: computeRequirements(info, offerings, region, amiFamily),
 		Offerings:    offerings,
-		Capacity:     computeCapacity(ctx, info, amiFamily, blockDeviceMappings, instanceStorePolicy, maxPods, podsPerCore),
+		Capacity:     computeCapacity(ctx, info, amiFamily, blockDeviceMappings, instanceStorePolicy, maxPods, podsPerCore, ncHash),
 		Overhead: &cloudprovider.InstanceTypeOverhead{
 			KubeReserved:      kubeReservedResources(cpu(info), pods(ctx, info, amiFamily, maxPods, podsPerCore), ENILimitedPods(ctx, info), amiFamily, kubeReserved),
 			SystemReserved:    systemReservedResources(systemReserved),
-			EvictionThreshold: evictionThreshold(memory(ctx, info), ephemeralStorage(info, amiFamily, blockDeviceMappings, instanceStorePolicy), amiFamily, evictionHard, evictionSoft),
+			EvictionThreshold: evictionThreshold(memory(ctx, info, ncHash), ephemeralStorage(info, amiFamily, blockDeviceMappings, instanceStorePolicy), amiFamily, evictionHard, evictionSoft),
 		},
 	}
 	if it.Requirements.Compatible(scheduling.NewRequirements(scheduling.NewRequirement(corev1.LabelOSStable, corev1.NodeSelectorOpIn, string(corev1.Windows)))) == nil {
@@ -192,11 +196,11 @@ func getArchitecture(info *ec2.InstanceTypeInfo) string {
 
 func computeCapacity(ctx context.Context, info *ec2.InstanceTypeInfo, amiFamily amifamily.AMIFamily,
 	blockDeviceMapping []*v1.BlockDeviceMapping, instanceStorePolicy *v1.InstanceStorePolicy,
-	maxPods *int32, podsPerCore *int32) corev1.ResourceList {
+	maxPods *int32, podsPerCore *int32, ncHash string) corev1.ResourceList {
 
 	resourceList := corev1.ResourceList{
 		corev1.ResourceCPU:              *cpu(info),
-		corev1.ResourceMemory:           *memory(ctx, info),
+		corev1.ResourceMemory:           *memory(ctx, info, ncHash),
 		corev1.ResourceEphemeralStorage: *ephemeralStorage(info, amiFamily, blockDeviceMapping, instanceStorePolicy),
 		corev1.ResourcePods:             *pods(ctx, info, amiFamily, maxPods, podsPerCore),
 		v1.ResourceAWSPodENI:            *awsPodENI(aws.StringValue(info.InstanceType)),
@@ -213,7 +217,7 @@ func cpu(info *ec2.InstanceTypeInfo) *resource.Quantity {
 	return resources.Quantity(fmt.Sprint(*info.VCpuInfo.DefaultVCpus))
 }
 
-func memory(ctx context.Context, info *ec2.InstanceTypeInfo) *resource.Quantity {
+func memory(ctx context.Context, info *ec2.InstanceTypeInfo, ncHash string) *resource.Quantity {
 	sizeInMib := *info.MemoryInfo.SizeInMiB
 	// Gravitons have an extra 64 MiB of cma reserved memory that we can't use
 	if len(info.ProcessorInfo.SupportedArchitectures) > 0 && *info.ProcessorInfo.SupportedArchitectures[0] == "arm64" {
@@ -221,7 +225,12 @@ func memory(ctx context.Context, info *ec2.InstanceTypeInfo) *resource.Quantity 
 	}
 	mem := resources.Quantity(fmt.Sprintf("%dMi", sizeInMib))
 	// Account for VM overhead in calculation
-	mem.Sub(resource.MustParse(fmt.Sprintf("%dMi", int64(math.Ceil(float64(mem.Value())*options.FromContext(ctx).VMMemoryOverheadPercent/1024/1024)))))
+	overhead, ok := MemoryOverheadMebibytes.Get(fmt.Sprintf("%s-%s", ncHash, *info.InstanceType))
+	if ok {
+		mem.Sub(resource.MustParse(fmt.Sprintf("%dMi", overhead.(int64))))
+	} else {
+		mem.Sub(resource.MustParse(fmt.Sprintf("%dMi", int64(math.Ceil(float64(mem.Value())*options.FromContext(ctx).VMMemoryOverheadPercent/1024/1024)))))
+	}
 	return mem
 }
 


pkg/controllers/instancetypes/memoryoverhead/controller.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+
+			// Build cache key from combination of NodeClass hash and instance type
+			instType := nodeClaim.Labels[corev1.LabelInstanceTypeStable]
+			key := fmt.Sprintf("%d-%s", nodeClass.Hash(), instType)
+
+			// Add memory capacity in the map
+			actual := node.Status.Capacity.Memory().Value() / 1024 / 1024
+
+			instancetype.MemoryOverheadMebibytes.SetDefault(key, actual)
```
jukie, line 85: This needs adjusted and will probably rename the controller to reflect it tracking actual

pkg/operator/operator.go
```diff
@@ -164,6 +164,7 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 		subnetProvider,
 		unavailableOfferingsCache,
 		pricingProvider,
+		operator.GetClient(),
```
jukie, line 167: Is this useable or should I move the client into the controller setup in pkg/controllers/providers/instancetype/controller.go?

pkg/providers/instancetype/instancetype.go
```diff
@@ -252,6 +256,60 @@ func (p *DefaultProvider) UpdateInstanceTypeOfferings(ctx context.Context) error
...
+	overheadMap := make(map[string]int64)
+	var mu sync.Mutex
+	workqueue.ParallelizeUntil(ctx, 100, len(nodeList.Items), func(i int) {
+		node := nodeList.Items[i]
+
+		// Only cache capacity info from nodes with current EC2NodeClass hash
+		nodeClass, ok := nodeClassMap[node.Annotations[v1.AnnotationEC2NodeClassHash]]
+		if !ok {
+			return
+		}
```
jukie, line 287: AMI details aren't added to EC2NodeClass Hash but this acts as some soft cache control

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
```diff
@@ -0,0 +1,69 @@
...
+func NewController(kubeClient client.Client, instancetypeProvider instancetype.Provider) *Controller {
+	return &Controller{
+		kubeClient:           kubeClient,
+		instancetypeProvider: instancetypeProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "instancetypes.memoryoverhead")
+	if err := c.instancetypeProvider.UpdateInstanceTypeMemoryOverhead(ctx, c.kubeClient); err != nil {
```
njtran, line 46: We probably want to make a precondition that this is a node that Karpenter provisioned.

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (c *Controller) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		Named("instancetypes.memoryoverhead").
+		For(&corev1.Node{}).
+		WithEventFilter(predicate.Funcs{
+			// Only trigger reconciliation on node creation events
+			CreateFunc: func(_ event.CreateEvent) bool {
+				return true
```
njtran, line 59: If we're making it a precondition for being a karpenter node, we'd probably need to change this to update, rather than create, since we first create the node, but then update it with the correct labels

pkg/providers/instancetype/instancetype.go
```diff
@@ -160,12 +167,17 @@ func (p *DefaultProvider) List(ctx context.Context, kc *v1.KubeletConfiguration,
 			instanceTypeLabel: *i.InstanceType,
 		}).Set(float64(aws.Int64Value(i.MemoryInfo.SizeInMiB) * 1024 * 1024))
 
+		var vmMemoryOverhead int64
+		if cached, ok := p.vmMemoryOverheadCache.Get(fmt.Sprintf("%s-%d", *i.InstanceType, amiHash)); ok {
```
njtran, line 171: We probably want to keep track of the other detail in the nodeclass, specifically from the kubelet

pkg/providers/instancetype/instancetype.go
```diff
@@ -252,6 +264,58 @@ func (p *DefaultProvider) UpdateInstanceTypeOfferings(ctx context.Context) error
...
 }
 
+func (p *DefaultProvider) UpdateInstanceTypeMemoryOverhead(ctx context.Context, kubeClient client.Client) error {
+	nodeClassList := &v1.EC2NodeClassList{}
+	if err := kubeClient.List(ctx, nodeClassList); err != nil {
+		return fmt.Errorf("failed to list nodeclasses: %w", err)
+	}
+
+	nodeClassMap := lo.Associate(nodeClassList.Items, func(nodeClass v1.EC2NodeClass) (string, v1.EC2NodeClass) {
+		return nodeClass.Hash(), nodeClass
```
njtran, line 274: Not sure you want to do the hash, as that encompasses more details than we care about for resource allocation.

pkg/providers/instancetype/instancetype.go
```diff
@@ -252,6 +264,58 @@ func (p *DefaultProvider) UpdateInstanceTypeOfferings(ctx context.Context) error
...
+		}
+
+		amiHash, _ := hashstructure.Hash(nodeClass.Status.AMIs, hashstructure.FormatV2, nil)
+
+		// Calculate memory overhead and store to the map
+		reportedMiB := instanceTypeInfo.MemoryInfo.SizeInMiB
+		actualMib := node.Status.Capacity.Memory().Value() / 1024 / 1024
+		overhead := *reportedMiB - actualMib
+		key := fmt.Sprintf("%s-%d", instanceType, amiHash)
+		p.vmMemoryOverheadCache.SetDefault(key, overhead)
```
njtran, line 314: So we assume that last write wins if somehow the allocatables are different?

njtran (commented):
Thanks for the PR! Looks really good!

pkg/providers/instancetype/instancetype.go
jukie: Yes, in theory we shouldn't see different values for the same instance-type+ami pair but if we do we always want the latest.

pkg/providers/instancetype/instancetype.go
jukie: Further down there's a match against this map and the node's nodeclass hash annotation so that we're only storing cache values for the latest version of the nodeclass. Would you instead prefer keying this map on the nodeclass name and then while looping through the nodes having a check that the node's ami is within the nodeclass status.amis list?

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
jukie: Thanks, good call! Fixed

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
jukie: Does the updated predicate function address this?

jukie:
CC @jmdeal (I think you were the one on the call today?) There's a negligible amount of variance when measuring overhead in Ki but I was unable to find any variance between the same instance type when measuring in Mi which my PR is doing.

Here's a simple script that can be used to verify: https://gist.github.com/jukie/df045af8fec68941f5d119044bf04aee

```
Instance Type: m6i.24xlarge
Variance detected in memory capacities:
 - 389990024
 - 389990016
 - 389990004
 - 389989996
 - 389990000
 - 389990004
 - 389989988
 - 389990000
 - 389990012
 - 389990012
 - 389989988
 - 389989980
 - 389990020
```

pkg/providers/instancetype/instancetype.go
jukie: Updated with a check against existing cache value and only updates if equal to or higher than current value. If non-existent will always write.

pkg/providers/instancetype/instancetype.go
jukie: Specifically here: https://github.com/kubernetes-sigs/karpenter/blob/887761eb533674aa29cb86e84b243781b7b0bf03/pkg/cloudprovider/types.go#L97-L99

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
jukie: I'm going to mark this as resolved since the updated predicate function addresses this when filtering the events.

pkg/providers/instancetype/instancetype.go
jukie: Updated with a lookup of AMI info instead

pkg/providers/instancetype/instancetype.go
```diff
@@ -152,7 +161,12 @@ func (p *DefaultProvider) List(ctx context.Context, nodeClass *v1.EC2NodeClass)
 			}
 		})
 
-		it := p.instanceTypesResolver.Resolve(ctx, i, zoneData, nodeClass)
+		var vmMemoryOverhead int64
+		if cached, ok := p.vmMemoryOverheadCache.Get(fmt.Sprintf("%s-%016x", *i.InstanceType, amiHash)); ok {
+			vmMemoryOverhead = cached.(int64)
+		}
+
+		it := p.instanceTypesResolver.Resolve(ctx, i, vmMemoryOverhead, zoneData, nodeClass)
```
jmdeal, line 169: Making these updates before we insert into the cache isn't ideal. This means we won't use any new discovered overheads until the previous cache entry expires. Unfortunately, there's no trivial way around it without additional changes: modifying the instance types directly after retrieving them from the cache will result in concurrent read-writes, and copying the data will result in significant memory usage.

Some possible solutions:
- Wrap the instance types in a synchronized object so we can mutate the instance capacity after insertion into the cache
- Surface a method on the cloudprovider to get per instance-type capacity overrides that the scheduler of can take advantage of. 
- Accept the 5 minute delay 

Personally I'm leaning toward the second option, it seems like it should integrate nicely with https://github.com/kubernetes-sigs/karpenter/pull/1305 if we end up taking that feature in.

pkg/providers/instancetype/instancetype.go
```diff
@@ -242,6 +256,80 @@ func (p *DefaultProvider) UpdateInstanceTypeOfferings(ctx context.Context) error
...
+
+		instanceTypeInfo, ok := instanceTypeInfoMap[instanceType]
+		if !ok || instanceTypeInfo == nil {
+			return
+		}
+
+		// Calculate memory overhead
+		reportedMiB := instanceTypeInfo.MemoryInfo.SizeInMiB
+		actualMiB := node.Status.Capacity.Memory().Value() / (1024 * 1024)
+		overhead := *reportedMiB - actualMiB
```
jmdeal, line 319: Rather than computing and caching the memory overhead, wouldn't it be simpler to just store the memory directly? I would model this as a capacity override, rather than an override for VM_MEMORY_OVERHEAD_PERCENT.

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
```diff
@@ -0,0 +1,77 @@
...
+			CreateFunc: func(e event.CreateEvent) bool { return false },
+			// Only trigger reconciliation once a node becomes registered
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				oldNode, okOld := e.ObjectOld.(*corev1.Node)
+				newNode, okNew := e.ObjectNew.(*corev1.Node)
+				if !okOld || !okNew {
+					return false
+				}
+				return oldNode.Labels[karpv1.NodeRegisteredLabelKey] == "" && newNode.Labels[karpv1.NodeRegisteredLabelKey] == "true"
+			},
```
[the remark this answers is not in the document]
jukie: Good point, will adjust to checking that registered label exists.

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
```diff
@@ -0,0 +1,77 @@
...
+		instancetypeProvider: instancetypeProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "instancetypes.memoryoverhead")
+	if err := c.instancetypeProvider.UpdateInstanceTypeMemoryOverhead(ctx, c.kubeClient); err != nil {
+		return reconcile.Result{}, fmt.Errorf("updating instancetype, %w", err)
+	}
+	return reconcile.Result{RequeueAfter: 12 * time.Hour}, nil
```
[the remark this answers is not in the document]
jukie: That's mostly to trigger refreshing of the cache before expiration. The function here doesn't actually use the specific node info since UpdateInstanceTypeMemoryOverhead() still lists out nodes on it's own.

pkg/cache/cache.go
```diff
@@ -38,6 +38,9 @@ const (
 	// SSMGetParametersByPathTTL is the time to drop SSM Parameters by path data. This only queries EKS Optimized AMI
 	// releases, so we should expect this to be updated relatively infrequently.
 	SSMGetParametersByPathTTL = 24 * time.Hour
+	// VMMemoryOverheadMebibytesTTL is the time to drop vm memory overhead data per-instance type
+	// if it is not updated by a node creation event or refreshed during controller reconciliation
+	VMMemoryOverheadMebibytesTTL = 36 * time.Hour
```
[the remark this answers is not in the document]
jukie: How would you suggest offloading cached values that are no longer needed? I think having some form of expiration here would be good, would extending this to 30d be sufficient?

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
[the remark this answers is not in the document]
jmdeal: The request we're requeuing is for a specific Node though. This results in an individual reconciliation per registered node every 12 hours.

pkg/cache/cache.go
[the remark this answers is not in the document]
jmdeal: If the AMIs have changed, the key should change as well right? I suppose we should have an expiration period to prevent this cache from growing indefinitely, but it could definitely be longer.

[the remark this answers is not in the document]
jukie: Yep exactly, growing indefinitely is what I was trying to avoid with this. Happy to adjust to whatever's reasonable and will start with 30d for now.

pkg/providers/instancetype/instancetype.go
jukie: As you noted the cache only lives for 5min and I thought that was an acceptable delay. You made a suggestion in another comment about tracking actual capacity. Instead of overhead of passing through the overhead value would it be acceptable to check the capacity cache (would rename from vmMemoryOverheadCache) for each instance-type after doing a lookup on instance-type cache?
```
if item, ok := p.instanceTypesCache.Get(key); ok {
	instanceTypes := append([]*cloudprovider.InstanceType{}, item.([]*cloudprovider.InstanceType)...)
		if cachedCapacity, found := p.vmMemoryOverheadCache.Get(fmt.Sprintf("%s-%016x", instanceType, amiHash)); found {
			instanceType.Capacity[corev1.ResourceMemory] = cachedCapacity.(resource.Quantity)
		}
	}
}
```
Do you have any thoughts of what an extension in the cloudprovider would look like? Probably something [in here](https://github.com/kubernetes-sigs/karpenter/blob/a860293883b67cbda0b8f7917130ccd6857b6cc3/pkg/controllers/provisioning/provisioner.go#L250-L271)
I think having the ability to override would be useful but I also think that cloudproviders should be returning valid information from the beginning rather than treating the override extension to arrive there.

NodeOverlays aren't dynamic though and is why this shouldn't _only_ be housed there in my opinion. Taking the example:
```
# Add memory overhead of 10Mi to all instances with 2Gi memory
# https://github.com/aws/karpenter-provider-aws/issues/5161
kind: NodeOverlay
metadata:
  name: discount
spec:
  selector:
    matchLabels:
      karpenter.k8s.aws/instance-memory: 2048
  overhead:
    memory: 10Mi
```
That would require a user to list out every possible instance-type and configure a per instance-type overhead. As I said having the ability to configure overrides will be useful, but I still think an internal controller is important for accurately tracking instance-type capacity.

pkg/controllers/providers/instancetype/capacitycache/controller.go
```diff
@@ -0,0 +1,78 @@
...
+	if err := c.instancetypeProvider.UpdateInstanceTypeCapacityCache(ctx, c.kubeClient); err != nil {
+		return reconcile.Result{}, fmt.Errorf("updating instancetype capacity cache, %w", err)
+	}
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		Named("instancetypes.capacitycache").
+		WatchesRawSource(singleton.Source()).
```
jukie, line 59: This will trigger an initial reconciliation loop upon startup and then Node events will trigger afterwards

pkg/providers/instancetype/instancetype.go
```diff
@@ -247,6 +263,78 @@ func (p *DefaultProvider) UpdateInstanceTypeOfferings(ctx context.Context) error
...
+		if !ok {
+			return
+		}
+
+		instanceTypeInfo, ok := instanceTypeInfoMap[instanceType]
+		if !ok || instanceTypeInfo == nil {
+			return
+		}
+
+		actualCapacity := node.Status.Capacity.Memory()
```
jukie, line 323: This could later be expanded to caching a v1.ResourceList but should probably be limited to those that are known to be safe overrides.

pkg/providers/instancetype/instancetype.go
jukie: Updated

pkg/cache/cache.go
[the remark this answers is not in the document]
jukie: Configured for 60d

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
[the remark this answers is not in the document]
jukie: Update events happen fairly frequently so rather than trigger reconciliation unnecessarily I've added an initialization event to trigger loading all nodes once upon startup and afterwards only on registration.

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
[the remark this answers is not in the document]
jukie: Should be fixed now

pkg/providers/instancetype/instancetype.go
```diff
@@ -247,6 +261,52 @@ func (p *DefaultProvider) UpdateInstanceTypeOfferings(ctx context.Context) error
...
+	// Ensure AMI is current
+	if !lo.ContainsBy(nodeClass.Status.AMIs, func(ami v1.AMI) bool {
+		return ami.ID == nodeClaim.Status.ImageID
+	}) {
+		return nil
+	}
+
+	instanceType := node.Labels[corev1.LabelInstanceTypeStable]
+
+	actualCapacity := node.Status.Capacity.Memory()
```
jukie, line 296: This could later be expanded to caching a v1.ResourceList but should probably be limited to those that are known to be safe overrides.

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
```diff
@@ -0,0 +1,77 @@
...
+func NewController(kubeClient client.Client, instancetypeProvider *instancetype.DefaultProvider) *Controller {
+	return &Controller{
+		kubeClient:           kubeClient,
+		instancetypeProvider: instancetypeProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "instancetypes.memoryoverhead")
+	if err := c.instancetypeProvider.UpdateInstanceTypeMemoryOverhead(ctx, c.kubeClient); err != nil {
```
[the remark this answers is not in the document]
jukie: Agreed, I've updated the logic to reconcile per-node registration events

pkg/providers/instancetype/instancetype.go
jukie: @jmdeal Rather than adding sync logic I've introduced `vmCapacityCacheSeqNum` to address the issue with waiting for cache to expire. I've also adjusted things to override capacity directly.

pkg/controllers/providers/instancetype/vmcapacitycache/controller.go
```diff
@@ -0,0 +1,105 @@
...
+				newNode, okNew := e.ObjectNew.(*corev1.Node)
+				if !okOld || !okNew {
+					return false
+				}
+				return oldNode.Labels[karpv1.NodeRegisteredLabelKey] == "" && newNode.Labels[karpv1.NodeRegisteredLabelKey] == "true"
+			},
+			CreateFunc:  func(e event.CreateEvent) bool { return false },
+			DeleteFunc:  func(e event.DeleteEvent) bool { return false },
+			GenericFunc: func(e event.GenericEvent) bool { return false },
+		}).
```
jmdeal, line 99: IIUC the purpose of the singleton source is to ensure we hydrate the cache on startup. Rather than having two separate reconciliation paths (one for individual Nodes, and one for startup where we list all nodes) what if we just reconcile against create events for registered Nodes. 
```suggestion
		For(&corev1.Node{}, builder.WithPredicates(predicate.TypedFuncs[client.Object]{
			// Only trigger reconciliation once a node becomes registered. This is an optimization to omit no-op reconciliations and reduce lock contention on the cache.
			UpdateFunc: func(e event.TypedUpdateEvent[client.Object]) bool {
			    if e.ObjectOld.GetLabels()[karpv1.NodeRegisteredLabelKey] != "" {
			        return false
			    }
			    return e.ObjectNew.GetLabels()[karpv1.NodeRegisteredLabelKey] == "true"
			},
			// Reconcile against all Nodes added to the cache in a registered state. This allows us to hydrate the cache on controller startup.
			CreateFunc:  func(e event.TypedCreateEvent[client.Object]) bool { 
			    return e.Object.GetLabels()[karpv1.NodeRegisteredLabelKey] == "true" 
			},
			DeleteFunc:  func(e event.TypedDeleteEvent[client.Object]) bool { return false },
			GenericFunc: func(e event.TypedGenericEvent[client.Object]) bool { return false },
		}).
```

pkg/controllers/providers/instancetype/vmcapacitycache/controller.go
```diff
@@ -0,0 +1,105 @@
...
+	if !c.initialized {
+		return c.reconcileAllNodes(ctx)
+	}
+	if err := c.instancetypeProvider.UpdateVMCapacityCache(ctx, c.kubeClient, r.Name); err != nil {
+		return reconcile.Result{}, fmt.Errorf("updating vm capacity cache, %w", err)
+	}
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) reconcileAllNodes(ctx context.Context) (reconcile.Result, error) {
```
jmdeal, line 61: With the changes to the registration function I suggested in a separate comment, we can eliminate this case altogether.

pkg/controllers/providers/instancetype/vmcapacitycache/controller.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func NewController(kubeClient client.Client, instancetypeProvider *instancetype.DefaultProvider) *Controller {
+	return &Controller{
+		kubeClient:           kubeClient,
+		instancetypeProvider: instancetypeProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, r reconcile.Request) (reconcile.Result, error) {
```
jmdeal, line 50: If we take in the suggested registration changes, this will only be reconciling against Nodes and we can convert it to a typed [`ObjectReconciler`](https://github.com/kubernetes-sigs/controller-runtime/blob/62834f93955d634d84c64159e09a54897c6b7c21/pkg/reconcile/reconcile.go#L127-L132) ([example](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/controllers/nodeclass/status/controller.go)).

jmdeal (commented):
This is looking pretty close. A few more suggestions, but nothing major at this point.

pkg/controllers/providers/instancetype/vmcapacitycache/controller.go
jukie: Thanks, will do!

jukie:
Thanks for the review, I'll get the comments addressed.

pkg/controllers/providers/instancetype/vmcapacitycache/controller.go
jukie: > Reconcile against all Nodes added to the cache in a registered state. This allows us to hydrate the cache on controller startup

Thanks for the education, I though it would still only react on _future_ node creation events.

pkg/providers/instancetype/instancetype.go
```diff
@@ -247,6 +261,52 @@ func (p *DefaultProvider) UpdateInstanceTypeOfferings(ctx context.Context) error
...
+	if err := kubeClient.Get(ctx, client.ObjectKey{Name: nodeClaim.Spec.NodeClassRef.Name}, nodeClass); err != nil {
+		return fmt.Errorf("failed to get ec2nodeclass: %w", err)
+	}
+
+	// Ensure AMI is current
+	if !lo.ContainsBy(nodeClass.Status.AMIs, func(ami v1.AMI) bool {
+		return ami.ID == nodeClaim.Status.ImageID
+	}) {
+		return nil
+	}
```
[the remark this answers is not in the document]
jmdeal: Don't remove it, I definitely think it should be included. And it should work today, using `MapToInstanceTypes` basically just ensures it's future proof.

pkg/controllers/providers/instancetype/memoryoverhead/controller.go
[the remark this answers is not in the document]
jukie: Updated with your suggestion

pkg/providers/instancetype/instancetype.go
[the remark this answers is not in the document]
jukie: @jmdeal could you please explain this a bit more? In order to call MapToInstanceTypes I'd need to also setup the related cloudprovider.InstanceType (unless I'm missing that being available somewhere) which seems a bit more involved than necessary.
My code doesn't consider the AMI requirements and simply checks whether the NodeClaim's AMI exists under the NodeClass and if not, skips it. If additional flexibility comes to AMI requirements wouldn't that just mean more AMIs under nodeClass.Status.AMIs?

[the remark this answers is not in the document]
jukie: Added the below but would like to understand why this is preferable to comparing the NodeClaim ImageID to nodeClass.Status.AMIs.
```
	// Ensure AMI is current
	instanceTypes, err := p.List(ctx, nodeClass)
	amiMap := amifamily.MapToInstanceTypes(instanceTypes, nodeClass.Status.AMIs)
	compatible := amiMap[nodeClaim.Status.ImageID]
	instanceType, found := lo.Find(compatible, func(i *cloudprovider.InstanceType) bool {
		return i.Name == node.Labels[corev1.LabelInstanceTypeStable]
	})
	if !found {
		return nil
	}
```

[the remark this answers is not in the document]
jukie: @jmdeal could you please explain this a bit more? If additional flexibility comes to AMI requirements wouldn't that just mean more AMIs under nodeClass.Status.AMIs?

Added the below to implement MapToInstanceTypes:
```
	// Ensure AMI is current
	instanceTypes, err := p.List(ctx, nodeClass)
	amiMap := amifamily.MapToInstanceTypes(instanceTypes, nodeClass.Status.AMIs)
	compatible := amiMap[nodeClaim.Status.ImageID]
	instanceType, found := lo.Find(compatible, func(i *cloudprovider.InstanceType) bool {
		return i.Name == node.Labels[corev1.LabelInstanceTypeStable]
	})
	if !found {
		return nil
	}
```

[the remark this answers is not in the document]
jukie: Calling List will make things a bit worse without including something like `discoveredCapacityCacheSeqNum` as it will trigger updating instanceTypesCache. Was there a different way you were suggesting to handle this?

pkg/test/environment.go
```diff
@@ -100,6 +100,7 @@ func NewEnvironment(ctx context.Context, env *coretest.Environment) *Environment
 	ec2Cache := cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval)
 	kubernetesVersionCache := cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval)
 	instanceTypeCache := cache.New(awscache.DefaultTTL, awscache.DefaultCleanupInterval)
+	discoveredCapacityCache := cache.New(awscache.DiscoveredCapacityCacheTTL, awscache.DefaultCleanupInterval)
```
jmdeal, line 103: For consistency, let's surface this cache as a public field in the `Environment` struct.

jmdeal (commented):
/karpenter snapshot

pkg/test/environment.go
jukie: Fixed

pkg/controllers/providers/instancetype/discoveredcapacitycache/suite_test.go
```diff
@@ -0,0 +1,142 @@
...
+	awsEnv.EC2API.DescribeInstanceTypeOfferingsOutput.Set(&ec2.DescribeInstanceTypeOfferingsOutput{
+		InstanceTypeOfferings: ec2Offerings,
+	})
+})
+
+var _ = AfterEach(func() {
+	ExpectCleanedUp(ctx, env.Client)
+})
+
+var _ = Describe("CapacityCache", func() {
```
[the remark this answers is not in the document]
jukie: Fixed, added more test cases @jmdeal

jukie:
@jmdeal could you take another look please

jukie:
Fixed the conflicts

jukie:
Any other changes needed here @jmdeal?

pkg/controllers/providers/instancetype/capacity/controller.go
```diff
@@ -0,0 +1,80 @@
...
+		Named("providers.instancetype.capacity").
+		For(&corev1.Node{}, builder.WithPredicates(predicate.TypedFuncs[client.Object]{
+			// Only trigger reconciliation once a node becomes registered. This is an optimization to omit no-op reconciliations and reduce lock contention on the cache.
+			UpdateFunc: func(e event.TypedUpdateEvent[client.Object]) bool {
+				if e.ObjectOld.GetLabels()[karpv1.NodeRegisteredLabelKey] != "" {
+					return false
+				}
+				return e.ObjectNew.GetLabels()[karpv1.NodeRegisteredLabelKey] == "true"
+			},
+			// Reconcile against all Nodes added to the cache in a registered state. This allows us to hydrate the cache on controller startup.
```
jmdeal, line 68: nit: let's differentiate between the informer and discovery caches. I think it's fine elsewhere, just want to differentiate here since they're both referenced in the same comment.
```suggestion
			// Reconcile against all Nodes added to the informer cache in a registered state. This allows us to hydrate the discovered capacity cache on controller startup.
```

jmdeal (commented):
One nit, and I'm going to let E2Es run. Otherwise LGTM.

/karpenter snapshot

jmdeal (commented):
/karpenter snapshot

pkg/controllers/providers/instancetype/capacity/controller.go
jukie: Sounds good, added

jukie:
I think the e2e's need a retry after committing your suggestion but the prior run appeared to pass.

jmdeal (approved):
No need to rerun, the only changes since they last ran was the comment nit and some docs changes you merged in. This LGTM 🚀.

jukie:
Hi @jmdeal I saw 1.0.7 was released but this wasn't included. Will this make it into the next one?

jmdeal:
We weren't planning on including this in a patch release, that's typically reserved for critical bug and security fixes. The plan would be to include this in the next minor version release, `v1.1.0`.

jukie:
Thanks!

PR #7006, opened 2024-09-12
chore: Fix env vars used in e2e Karpenter install and diff test scripts

bryantbiggs:
Fixes #N/A 

Notice that the region is missing from the endpoint that is rendered in the script

image

**Description**
- Fix env vars used in e2e Karpenter install and diff test scripts
  - Since the URI format for ECR is using the ECR DKR endpoint, there shouldn't need to be a difference between endpoints for public cluster vs private cluster (unless the images are located in a different accounts). Therefore the endpoints should be the same between public and private cluster usage which means the same account ID and region. If my understanding of https://github.com/aws/karpenter-provider-aws/blob/c92515867cb869aad4a5781c7f23aeb30063064e/.github/actions/e2e/install-karpenter/action.yaml#L51-L67 is correct - `ACCOUND_ID` and `REGION` are the account ID and region where the clusters are created whereas the `ECR_ACCOUNT_ID` and `ECR_REGION` are the account ID and region where the Helm chart and images are stored within ECR. The prior scripts were switching between a mix of cluster account ID and region and the ECR account ID and region, which not all of those values are provided to the script and resulted in incorrect template rendering

- [ ] Yes, issue opened: # 
- [x] No

### DIFF
--- a/test/hack/e2e_scripts/diff_karpenter.sh
+++ b/test/hack/e2e_scripts/diff_karpenter.sh
@@ -1,7 +1,4 @@
 CHART="oci://$ECR_ACCOUNT_ID.dkr.ecr.$ECR_REGION.amazonaws.com/karpenter/snapshot/karpenter"
-if [[ "$PRIVATE_CLUSTER" == "true" ]]; then
-  CHART="oci://$ACCOUNT_ID.dkr.ecr.$REGION.amazonaws.com/karpenter/snapshot/karpenter"
-fi
 
 helm diff upgrade --namespace kube-system \
 karpenter "${CHART}" \
--- a/test/hack/e2e_scripts/install_karpenter.sh
+++ b/test/hack/e2e_scripts/install_karpenter.sh
@@ -1,10 +1,13 @@
 aws eks update-kubeconfig --name "$CLUSTER_NAME"
 
-CHART="oci://$ECR_ACCOUNT_ID.dkr.ecr.$ECR_REGION.amazonaws.com/karpenter/snapshot/karpenter"
+ECR_URI="oci://$ECR_ACCOUNT_ID.dkr.ecr.$ECR_REGION.amazonaws.com"
+CHART="$ECR_URI/karpenter/snapshot/karpenter"
 ADDITIONAL_FLAGS=""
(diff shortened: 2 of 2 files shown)


bryantbiggs:
cc @jigisha620

jonathan-innis (commented):
/karpenter snapshot

jonathan-innis:
@bryantbiggs Is this still relevant since the code was updated with a fix that resolved the failure that we were running into?

bryantbiggs:
yes, its still relevant 

there are a few parts:
1. Because the endpoint that is used is using the ECR private endpoint format, there shouldn't need to be a difference between the endpoint used for testing a private clusters versus a non-private cluster. The only time you would need to make decision on which endpoint would be if you were pulling images from a public registry like Dockerhub, Public ECR, etc. So the chart URI should be the same regardless of public or private cluster
2. It doesn't look like tests are currently executed against private clusters because not all of the required environment variables are being passed down to the scripts. For example [this script](https://github.com/aws/karpenter-provider-aws/blob/f0dbb2689871d6da851a89090fb738b5476393af/test/hack/e2e_scripts/install_karpenter.sh#L6-L7) needs the `REGION` environment variable but the workflow does [not pass that environment variable](https://github.com/aws/karpenter-provider-aws/blob/f0dbb2689871d6da851a89090fb738b5476393af/.github/actions/e2e/install-karpenter/action.yaml#L53-L58). Likewise, [this script](https://github.com/aws/karpenter-provider-aws/blob/f0dbb2689871d6da851a89090fb738b5476393af/test/hack/e2e_scripts/diff_karpenter.sh#L2-L4) requires both `ACCOUNT_ID` and `REGION` when its a private cluster, but those are [not passed in the workflow](https://github.com/aws/karpenter-provider-aws/blob/f0dbb2689871d6da851a89090fb738b5476393af/.github/actions/e2e/install-karpenter/action.yaml#L63-L65)
3. There seems to be a mixup between `ACCOUNT_ID` and `ECR_ACCOUNT_ID` as well as `REGION` and `ECR_REGION` - I can't say which is right or wrong without knowing the intent, but based on the two issues above - I think the `ECR_*` versions should be used whenever it comes to an OCI artifact coming out of ECR

PR #7012, opened 2024-09-15
chore: Always disable instance metadata tags

jonathan-innis:
Fixes #N/A 

**Description**

This change always disables instance metadata tags. Not passing this setting _explicitly_ to the instance launch can cause an account that [has configured account defaults for instance metadata settings](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html#instance-metadata-options-order-of-precedence) to override the instance default. If the account default has IMDS tags set to `enabled`, EC2 will use this, causing Karpenter to fail to launch instances with

```
  - lastTransitionTime: "2024-09-15T05:00:09Z"
    message: 'creating instance, with fleet error(s), InvalidParameterValue: ''kubernetes.io/cluster/scale-test''
      is not a valid tag key. Tag keys must match pattern ([0-9a-zA-Z\\-_+=,.@:]{1,255}),
      and must not be a reserved name (''.'', ''..'', ''_index''); RequestLimitExceeded:
      Request limit exceeded.'
```

- [ ] Yes, issue opened: # 
- [x] No


jonathan-innis (commented):
/karpenter snapshot

jonathan-innis (commented):
/karpenter snapshot

rschalo (approved):

